PR #7045, opened 2021-05-12
chore: comment engine usage in cli

williamluke4:


Jolg42:
About cwd in Migrate Engine:
https://prisma-company.slack.com/archives/CGMCB82N8/p1621253184006600?thread_ts=1621253015.006400&cid=CGMCB82N8
>If the process is spawned from another directory, all file paths would resolve relative to that instead of the prisma directory — note that it isn't something engines specific but just a process spawning thing.
Paths resolved in engines code include at least:
> - sqlite database paths
> - ssl certificate paths
> The CLI passes in absolute paths to the migrations directory, so that isn't affected by the CWD.

Jolg42:
Replaced by https://github.com/prisma/prisma/pull/10732

PR #7111, opened 2021-05-17
fix(client): process-relative paths

millsp:
With this PR, we make sure that the prisma client is portable. Previously, when bundled, or copied around, the generated client would fail to initialize. At the same time, we improved the way we handle monorepos and symlinked projects with the new non-resolving discovery for `@prisma/client` and `.prisma/client`.

This should ease development on serverless environments and especially when bundling the Prisma client.

e2e: https://github.com/prisma/e2e-tests/actions/runs/893991714


src/packages/client/src/generation/TSClient/common.ts
```diff
@@ -32,10 +32,12 @@ const {
   empty,
   join,
   raw,
-  Decimal
+  Decimal,
+  findSync
 } = require('${runtimePath}')
 
 const path = require('path')
+const fs = require('fs')
```
Jolg42, line 40: Looks extra? We could remove?

src/packages/client/src/runtime/utils/find.ts
```diff
@@ -0,0 +1,290 @@
+/* eslint-disable @typescript-eslint/no-inferrable-types */
+import fs from 'fs'
+import path from 'path'
+
+const readdirAsync = fs.promises.readdir
+const realpathAsync = fs.promises.realpath
+const statAsync = fs.promises.stat
```
Jolg42, line 7: We didn't use `fs.promises` until now because on Node 10 it's marked as experimental and prints warnings in logs.

We can still merge or do it for next release and drop Node 10 at the same time.

src/packages/client/src/runtime/utils/find.ts
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * Like `findSync` but moves to the parent folder if nothing is found
+ * @param root to start from
+ * @param match to match against
+ * @param types to select files, folders, links
+ * @param deep to recurse in the directory tree
+ * @param limit to limit the results
+ * @param filter to further filter results
+ * @param found to add to already found #private
+ * @param seen to add to already seen #private
```
Jolg42, line 138: What does #private means?

src/packages/client/src/runtime/utils/find.ts
```diff
@@ -0,0 +1,290 @@
...
+ * @returns unresolved paths
```
Jolg42, line 139:
```suggestion
 * @returns unresolved paths or matching path
```
maybe?

src/packages/client/src/runtime/utils/find.ts
millsp: It's a field used for recursion mostly, I could remove that as it does not convey much meaning

src/packages/sdk/src/predefinedGeneratorResolvers.ts
```diff
@@ -1,13 +1,13 @@
...
 import fs from 'fs'
 import hasYarn from 'has-yarn'
 import path from 'path'
-import resolvePkg from 'resolve-pkg'
 import { logger } from '.'
 import { getCommandWithExecutor } from './getCommandWithExecutor'
-const debugEnabled = Debug.enabled('generator')
+import { findUpAsync as findUp } from './utils/find'
+
+const load = require
```
Jolg42, line 10: Why aliasing / hiding require in a const?

src/packages/client/src/runtime/utils/find.ts
millsp: Btw that file is duplicated and that is normal, it's related to another issue where we are not able to generate types for cases where we import from another local package. For example, it would fail if `sdk` would export `find` and `find` were imported in `client`. This only fails on functions and is related to `rollup-plugin-dts`. Will open an issue soon.

src/packages/sdk/src/utils/find.ts
```diff
@@ -0,0 +1,290 @@
+/* eslint-disable @typescript-eslint/no-inferrable-types */
```
Jolg42, line 1: Oh so this is a duplicated file 🤔 

Possible to avoid duplication? 

Like having only one in sdk and importing it in client?

src/packages/client/src/generation/TSClient/common.ts
```diff
@@ -32,10 +32,12 @@ const {
   empty,
   join,
   raw,
-  Decimal
+  Decimal,
+  findSync
```
Jolg42, line 36: I guess this is extra too

src/packages/sdk/src/predefinedGeneratorResolvers.ts
millsp: It's to hide it from bundlers. I find it more beautiful than using `eval` which is slower and is not good for type safety.

millsp: I'll add a comment

williamluke4 (approved):
Awesome Work :raised_hands:

src/packages/sdk/src/utils/find.ts
millsp: Just copy pasted my comment above:

> Btw that file is duplicated and that is normal, it's related to another issue where we are not able to generate types for cases where we import from another local package. For example, it would fail if sdk would export find and find is imported in client. This only fails on functions and is related to rollup-plugin-dts. Will open an issue soon.

src/packages/client/src/runtime/utils/find.ts
Jolg42: Is it not possible to import like this from client? That could work I think.
`import something from '@prisma/sdk/dist/find`

williamluke4: Yeah we already use `import something from '@prisma/sdk/dist/xxx` in the runtime

src/packages/client/src/runtime/utils/find.ts
millsp: Ok, will add `promisify` instead

Jolg42 (approved):
💚 Well done!

PR #7144, opened 2021-05-18
fix(migrate): seed - wait for promise (#7139)

Jolg42:


janpio:
Rebasing this should make tests green.

PR #7163, opened 2021-05-19
ci: test napi in buildkite too

Jolg42:
Is that enough?

How to make sure this runs NAPI from the logs?


williamluke4 (approved):

williamluke4:
If you enable "DEBUG=*" it should log out weather it's using NAPI or not

Jolg42:
It worked ✅

PR #7258, opened 2021-05-25
fix(client): disable enumerable on JsonFilter path

### LINKED ISSUE #7136: Disallow passing a string as a path in Postgres
### Bug description

From @Weakky 

> One of the reason why I think we shouldn’t allow a string to be passed is that it can be misinterpreted as a support for JSON path. The reason postgres does not accept a string is because it does NOT support JSON path.
With Postgres, if you want to target a “deep” property, you need to do: path: ["a", "b"]. The equivalent on MySQL is "path": "$.a.b" (because MySQL does support JSON path) 

### How to reproduce

Follow the example in the release notes for 2.23.0 and then write

```js
const posts = await prisma.post.findMany({
    where: {
      content: {
        path: "type",
      },
    },
  })
```

### Expected behavior

- Postgres: `path: string[]`
- MySQL: `path: string`

### Prisma information

```prisma

datasource db {
  provider = "postgresql"
}

generator client {
  previewFeatures = ["filterJson"]
}

model Log {
  id      Int    @id @default(autoincrement())
  level   Level
  message String
  meta    Json
}

enum Level {
  INFO
  WARN
  ERROR
}

```

### Environment & setup

- OS: OSX
- Database: Postgres

### Prisma Version

```
prisma               : 2.23.0-dev.41
@prisma/client       : 2.23.0-dev.41
Current platform     : darwin
Query Engine         : query-engine adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/query-engine-darwin)
Migration Engine     : migration-engine-cli adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine : introspection-core adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary        : prisma-fmt adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash : adf5e8cba3daf12d456d911d72b6e9418681b28b
Studio               : 0.393.0
Preview Features     : filterJson
```

williamluke4:
implements https://github.com/prisma/prisma/issues/7136

### DIFF
--- a/src/packages/client/src/__tests__/types/json-filtering-mysql/index.test-d.ts
+++ b/src/packages/client/src/__tests__/types/json-filtering-mysql/index.test-d.ts
@@ -13,22 +13,21 @@ const prisma = new PrismaClient({
 
 ;(async () => {
   expectError(
-    await prisma.user.create({
-      data: {
+    await prisma.user.findFirst({
+      where: {
         info: {
-          x: new Date(),
+          path: ["any"]
         },
-        email: '',
       },
     }),
   )
   expectError(
-    await prisma.user.create({
-      data: {
+    await prisma.user.findFirst({
+      where: {
         info: {
-          x: /regex/,
+          path: ["any"],
+          gt: 4
         },
-        email: '',
       },
     }),
   )
--- a/src/packages/client/src/__tests__/types/json-filtering-mysql/index.test-d.ts
+++ b/src/packages/client/src/__tests__/types/json-filtering-mysql/index.test-d.ts
@@ -1,7 +1,5 @@
-import { PrismaClient } from '.'
 import { expectError } from 'tsd'
-
-// tslint:disable
+import { PrismaClient } from '.'
 
 const prisma = new PrismaClient({
   datasources: {
@@ -16,7 +14,7 @@ const prisma = new PrismaClient({
     await prisma.user.findFirst({
       where: {
         info: {
-          path: ["any"]
+          path: ['any'],
         },
       },
     }),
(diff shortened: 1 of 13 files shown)


src/packages/client/src/__tests__/types/json-filtering-mysql/index.test-d.ts
```diff
@@ -0,0 +1,34 @@
+import { PrismaClient } from '.'
+import { expectError } from 'tsd'
+
+// tslint:disable
```
Jolg42, line 4: Why is that needed? 👀

williamluke4: No idea was copy/pasted from another test

src/packages/client/src/__tests__/types/json-filtering-mysql/test.ts
```diff
@@ -0,0 +1,29 @@
+import { PrismaClient, Prisma } from '@prisma/client'
+
+// tslint:disable
```
Jolg42, line 3: One more here

src/packages/client/src/__tests__/types/json-filtering-postgres/test.ts
```diff
@@ -0,0 +1,29 @@
+import { PrismaClient, Prisma } from '@prisma/client'
+
+// tslint:disable
```
Jolg42, line 3: and here :)

Jolg42 (approved):
Looks good to me 💯

I think @millsp could also do a quick review

millsp (approved):

PR #7302, opened 2021-05-27
fix: gate mongodb for migrate / db commands

### LINKED ISSUE #7012: Disable "db" and "migrate" commands for "mongodb" provider until we support it.
### Bug description

Reported from a user:

"db push" fails with "collection xy already exists" errors when data already exists. Not sure if db push is supposed to work at all with MongoDB, but this behavior is probably unintentional

### How to reproduce

Follow the MongoDB getting started guide and then try running db push.

### Expected behavior

It should probably say mongodb is not a supported database. We should maybe have an allowlist for providers.

### Prisma information

N/A. User report.

### Environment & setup

N/A. User report.

Jolg42:
Closes https://github.com/prisma/prisma/issues/7012

Message implemented:
```
"mongodb" provider is not supported with this command. For more info see https://www.prisma.io/docs/concepts/database-connectors/mongodb
```

### DIFF
--- a/src/packages/migrate/src/__tests__/MigrateDev.test.ts
+++ b/src/packages/migrate/src/__tests__/MigrateDev.test.ts
@@ -605,14 +605,14 @@ describe('sqlite', () => {
 
     await expect(result).rejects.toMatchInlineSnapshot(`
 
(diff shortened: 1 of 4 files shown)


matthewmueller (approved):
Looks great, thanks @Jolg42 !

Issue #2779, opened 2020-06-16
Parameter for `prisma init` to supply database provider string to be used in `datasource`


janpio:
## Problem
Although I always use a MySQL database, I can not create a project with a `provider = 'mysql'`.

## Suggested solution
Similar to https://github.com/prisma/prisma/issues/2577 it would be neat to be able to set (just) the provider of the datasource via a param `--provider`.

## Alternatives
Do it manually.

SuryaElavazhagan:
I tried to fix this here: #7481

Jolg42:
Thanks to @SuryaElavazhagan for the PR!

I did small changes in https://github.com/prisma/prisma/pull/7557

So how this works?
You can now do with latest internal dev version
```
npx prisma init (default to postgresql)
npx prisma init --provider mysql
npx prisma init --provider sqlserver
npx prisma init --provider sqlite
npx prisma init --provider postgresql
npx prisma init --provider mongodb

the provider is lowercased at runtime so passing MySQL as a provider works too.

npx prisma init --provider invalid 
# will return:
Provider "invalid" is invalid or not supported. Try again with "postgresql", "mysql", "sqlserver" or "sqlite".
```

janpio:
Great work @SuryaElavazhagan - thanks so much! Small feature, but sooo convenient when you are a MySQL user usually.

Jolg42:
We noticed an ambiguity with the new --provider argument.

Since both the generator and datasource blocks have a provider field, it's not clear (just from seeing the argument name --provider) to which of them it applies.

We just removed that --provider from the latest dev version so it doesn't get released to decide if we want to do something to release it later.

PR is https://github.com/prisma/prisma/pull/7671

Jolg42:
This was shipped in 2.26.0 see how to use it in the release notes at https://github.com/prisma/prisma/releases/tag/2.26.0

PR #7481, opened 2021-06-05
feat(cli): Added support for provider argument in `prisma init`

SuryaElavazhagan:
Addresses #2779 issue

### DIFF
--- a/src/packages/cli/src/Init.ts
+++ b/src/packages/cli/src/Init.ts
@@ -68,7 +68,7 @@ export class Init implements Command {
       '--help': Boolean,
       '-h': '--help',
       '--url': String,
-      '--provider': String
+      '--provider': String,
     })
 
     if (isError(args) || args['--help']) {
--- a/src/packages/cli/src/Init.ts
+++ b/src/packages/cli/src/Init.ts
@@ -66,10 +66,16 @@ export const defaultURL = (
   port = defaultPort(provider),
   schema = 'public',
 ) => {
-  const schemaExists = provider === 'postgresql'
-  return `${provider}://johndoe:randompassword@localhost:${port}/mydb${
-    schemaExists ? `?schema=${schema}` : ''
-  }`
+  switch (provider) {
+    case 'postgresql':
+      return `${provider}://johndoe:randompassword@localhost:${port}/mydb?schema=${schema}`
+    case 'mysql':
+      return `${provider}://johndoe:randompassword@localhost:${port}/mydb`
+    case 'sqlserver':
+      return `${provider}://localhost:${port};database=mydb;user=SA;password=randompassword;`
+    case 'sqlite':
+      return 'file:./dev.db'
+  }
 }
 
 export class Init implements Command {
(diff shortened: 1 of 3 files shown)


src/packages/cli/src/Init.ts
```diff
@@ -143,6 +144,11 @@ export class Init implements Command {
       url = args['--url']
     }
 
+    if (args['--provider'] && !args['--url']) {
+      provider = args['--provider']
+      url = `${provider}://johndoe:randompassword@localhost:5432/mydb?schema=public`
```
janpio, line 149: We could differentiate the default port and existance of the schema (only on Postgres) as well.

SuryaElavazhagan: Nice one. Will add them..

SuryaElavazhagan: Please correct me if I'm wrong. I think you can query postgres to check for the existence of the schema only if you have the right credentials. It's very very rare to have the same credentials. I don't think it's worth it. I'd love to hear your thoughts on this.

janpio: What I meant is that on MySQL you do not specify a `schema` as it does not do anything. Only on PostgreSQL you do.

SuryaElavazhagan: Ah, sorry. Got it.

src/packages/cli/src/Init.ts
```diff
@@ -143,6 +167,11 @@ export class Init implements Command {
       url = args['--url']
     }
 
+    if (args['--provider'] && !args['--url']) {
```
Jolg42, line 170:
```suggestion
    } else if (args['--provider']) {
```
Minor suggestion: I think this could be better

Jolg42 (approved):
This looks great!

I think we could add mongodb later.

It's in "Early Access" for now, more info at https://www.prisma.io/docs/concepts/database-connectors/mongodb

And probably we can use this for the default url
```
mongodb://johndoe:randompassword@localhost:27017/mydb?authSource=admin
```

Jolg42 (changes requested):
Actually, this implementation doesn't work for `sqlserver` which is JDBC style like 
```
sqlserver://localhost:5432;database=mydb;user=SA;password=randompassword;"
```
(https://www.prisma.io/docs/concepts/components/preview-features/sql-server/sql-server-connection-string)

src/packages/cli/src/Init.ts
```diff
@@ -49,6 +49,29 @@ export const defaultEnv = (
...
+export const defaultURL = (
+  provider = 'postgresql',
+  port = defaultPort(provider),
+  schema = 'public',
+) => {
+  const schemaExists = provider === 'postgresql'
+  return `${provider}://johndoe:randompassword@localhost:${port}/mydb${
+    schemaExists ? `?schema=${schema}` : ''
+  }`
+}
```
Jolg42, line 73: We could also add a `sqlite` provider here to be complete I think, default could be `file:./dev.db`

https://www.prisma.io/docs/concepts/database-connectors/sqlite

SuryaElavazhagan: Silly me. Thanks. I've responded to your comments btw.

pantharshit00 (approved):

williamluke4 (approved):

Jolg42:
Thanks! 💚 

I added mongodb and small changes/fixes in https://github.com/prisma/prisma/pull/7557

PR #7514, opened 2021-06-07
chore: remove unused script

psavan1655:
style: Removed console log. As i suppose developer forgot to delete after testing it.

### DIFF
--- a/src/scripts/semver.ts
+++ b/src/scripts/semver.ts
@@ -1,4 +1,3 @@
 import semver from 'semver'
 
 const valid = semver.valid('2.0.0-preview017')
-console.log(valid)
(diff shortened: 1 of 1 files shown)


Jolg42:
@psavan1655 The console.log is valid here because it's a manual script so I won't merge this.

Though If you want to edit this PR and actually delete this script that would be nice since it's not used and could be cleaned up.

psavan1655:
> @psavan1655 The console.log is valid here because it's a manual script so I won't merge this.
> 
> Though If you want to edit this PR and actually delete this script that would be nice since it's not used and could be cleaned up.

You mean i can delete the whole file? Please correct me if i am wrong. As it is my first PR to open source. And thankyou for a wonderful response.

Jolg42:
Yes indeed, deleting the whole file would be nice!

psavan1655:
> Yes indeed, deleting the whole file would be nice!

I have deleted the file.

Jolg42 (approved):
Thanks for your first PR!

PR #7558, opened 2021-06-09
ci: change order for Node-API package testing

Jolg42:


Jolg42:
That's a good question.

The new order makes more sense since it's bottom to top dependency, like the publish script does.

What's weird is that this apparently fixes the tests timing out, they were getting stuck after integration-tests for no apparent reason.

PR #7604, opened 2021-06-11
ci: explain how to find out what buildkite test jobs are doing

janpio:


Jolg42 (approved):

Jolg42:
Interesting, flaky? in Node-API only 
Screen Shot 2021-06-11 at 12 48 50
Retried it

janpio:
Yep, saw something like that before and possibly already raised by Pierre in Node-API channel.

PR #7624, opened 2021-06-12
feat: detect local cli installation from global

### LINKED ISSUE #2738: Global Prisma could check for local Prisma and warn or require explicit confirmation
There are two ways to install the Prisma CLI: Global or local installation. Although we strongly suggest local installation in a project, there are reasons why people might ignore that and use a global CLI installation.

This is problematic, when you check out a project with a local installation, but accidentally execute the global `prisma` instead of the local via `npx prisma`.

To avoid this, the global CLI could check for a local one on execution and warn or even require explicit confirmation from the user that this is the right Prisma CLI they want to use.

fnky:
Detects whether a local Prisma installation is in node modules of the current directory and displays a warning, if Prisma CLI was run from a global installation.

### Testing

Since symlinks (i.e. with `npm link`) won't work when testing this feature, you can test
this by packing the cli package and installing the resulting tarball globally.

Install a local version. This can be from directly from npm, just as long as
prisma is in the `node_modules/.bin`).

```sh
# Go to cli package
cd src/packages/cli

# with NPM

npm pack
npm i -g prisma-0.0.0.tgz

# or with Yarn

yarn pack

# note: use `pwd` to get the current working directory
#       and replace <PWD> with that path.
yarn global add file:/<PWD>/prisma-v0.0.0.tgz
```

To remove global packages when you're done testing, simply remove them from each
package manager like so:

```sh
npm rm -g prisma

# or

yarn global remove prisma
```

Closes https://github.com/prisma/prisma/issues/2738

### DIFF
--- a/src/packages/cli/src/detectLocalPrisma.ts
+++ b/src/packages/cli/src/detectLocalPrisma.ts
@@ -14,7 +14,9 @@ export function detectLocalPrisma() {
     console.warn(
       printWarning(`You're running a global installation of Prisma but we detected a local
           installation in your node modules. It's recommended to always run a
-          locally available installation of Prisma.`),
+          locally available installation of Prisma.
+
+          Use \`npx prisma\` to run from local installation.`),
     )
   }
 }
--- a/src/packages/cli/src/detectLocalPrisma.ts
+++ b/src/packages/cli/src/detectLocalPrisma.ts
@@ -6,9 +6,6 @@ import { printWarning } from './prompt/utils/print'
 const isPrismaInstalledGlobally = isCurrentBinInstalledGlobally()
 
 export function detectLocalPrisma() {
-  // TODO: Alternatively we can prompt the user before executing the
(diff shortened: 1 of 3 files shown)


fnky:
Let me know if we need more explicit tests for this. I decided to leave them out as I didn't think it would be a serious enough thing that would break usability.

src/packages/cli/src/detectLocalPrisma.ts
```diff
@@ -0,0 +1,22 @@
+import fs from 'fs'
+import path from 'path'
+import { isCurrentBinInstalledGlobally } from '@prisma/sdk'
+import { printWarning } from './prompt/utils/print'
+
+const isPrismaInstalledGlobally = isCurrentBinInstalledGlobally()
+
+export function detectLocalPrisma() {
+  if (
+    true ||
```
wmadden, line 10: This looks wrong

PR #7654, opened 2021-06-14
fix(migrate): cleanup datasource printing for SQLServer & default to postgres for pg

Jolg42:


src/packages/migrate/src/utils/printDatasource.ts
```diff
@@ -5,15 +5,23 @@ export async function printDatasource(schemaPath: string): Promise<void> {
...
-      chalk.dim(
-        `Datasource "${dbInfo.name}": ${dbInfo.dbType} ${dbInfo.schemaWord} "${
-          dbInfo.dbName
-        }"${dbInfo.schema ? `, schema "${dbInfo.schema}"` : ''} at "${
-          dbInfo.dbLocation
-        }"`,
-      ),
-    )
+    if (dbInfo.dbType === 'PostgreSQL' && dbInfo.dbName === undefined) {
+      dbInfo.dbName = 'postgres'
```
tomhoule, line 9: I think this should default to the username from the URL instead, to match postgres' defaults.

tomhoule: Had a discussion with @Jolg42 , did the investigation, and it turns out quaint (our drivers library) does default to "postgres", while psql defaults to the username. So this is correct.

Jolg42: Actually Tom checked and in quaint the default is always `postgres` (it doesn't depend on the username, while psql does)

Issue #6705, opened 2021-04-22
Transaction with bad input should rollback (when using middleware)


Sytten:
## Bug description

If one query in the transaction fails the input validation, the transaction doesn't rollback and the other queries proceed normally. 

Error:
```
Error: Argument city for data.address.create.city is missing.

    at Document.validate (/app/node_modules/.prisma/client/runtime/index.js:32462:19)
    at NewPrismaClient._executeRequest (/app/node_modules/.prisma/client/runtime/index.js:34399:17)
    at NewPrismaClient._requestWithMiddlewares (/app/node_modules/.prisma/client/runtime/index.js:34350:19)
    at /app/node_modules/.prisma/client/runtime/index.js:34343:53
    at /app/src/common/prisma/retry.js:35:38
    at NewPrismaClient._requestWithMiddlewares (/app/node_modules/.prisma/client/runtime/index.js:34343:16)
    at /app/node_modules/.prisma/client/runtime/index.js:34332:54
    at AsyncResource.runInAsyncScope (async_hooks.js:197:9)
    at NewPrismaClient._request (/app/node_modules/.prisma/client/runtime/index.js:34332:27)
    at Object.requestTransaction (/app/node_modules/.prisma/client/runtime/index.js:34469:39)
```

## How to reproduce

Consider the following schema:
```prisma
model User {
  id   Int    @id @default(autoincrement())
  name String
}
```

Run the following code:
```javascript
const prisma = new PrismaClient();

prisma.$transaction([
  prisma.user.create({ data: { name: 'emile' } })
  prisma.user.create({ data: {} })
])
```

There should be a user in the database as the transaction is **not** rollback.

## Expected behavior

The transaction should be rollback.

## Environment & setup

- OS: Linux
- Database: PostgreSQL
- Node.js version: 14
- Prisma version: 2.21.2

pantharshit00:
Hmm, I wasn't able to reproduce this. I don't get the user in the database. 

I also enabled postgres logging and checked the database logs and those queries didn't even hit the database:
![image](https://user-images.githubusercontent.com/22195362/115692399-91a95d00-a37c-11eb-9f5c-b42597d6bb16.png)


Maybe the faulty details went missing when you abstracted your query for a reproduction?

Sytten:
Yeah most likely, just put an example on top of my head but I will try to reproduce. I think it my be due to a children `create`.

Sytten:
Ok I dig out the issue, it happens when there is a middleware present.
[...]

PR #7662, opened 2021-06-14
fix(client): delete transactionId field in middleware

### LINKED ISSUE #7584: Middleware + $transactions do not play well together
### Bug description

If I use a middleware to Client, then call a `prisma.$transaction`, then the middleware blocks query execution.

### How to reproduce

Assume this schema:

```
model User {
  id Int @id
  name String
}
```

I have a middlware that modifies Client responses to also include some extra parameters with it:

```

  let response = await next(params);
  console.log("middleware", response);
  return {
    params,
    response,
  };
});

  await prisma.$transaction([
    prisma.user.create({
      data: {
        id: "1",
        name: "Sid",
      },
    }),
  ]);
```

However:

1. The query doesn't get executed. I don't see this record in my database.
2. Instead of logging the Client response, the variabke `response` seems to be an async function. If I log `response.toString()`, it looks like calling `next` returns the actual `next` function. If I call this returned value again, like so:

```
  let response = await next(params);
  response = await (response as any)(params);
  console.log("middleware", response);
  return {
    params,
    response,
  };
});

  await prisma.$transaction([
    prisma.user.create({
      data: {
        id: "1",
        name: "Sid",
        age: 0,
      },
    }),
  ])
```

then everything works as expected. I see the correct log and the query gets executed.

### Expected behavior

_No response_

### Prisma information

model User {
  id Int @id
  name String
}

### Environment & setup

- OS: macOS
- Database: Postgres
- Node.js version: v14.17.0

### Prisma Version

This seems to only be happening on `dev`. Prisma 2.24.1 behaves as expected.

I'm on:

```
prisma               : 2.25.0-dev.31
@prisma/client       : 2.25.0-dev.31
Current platform     : darwin
Query Engine         : query-engine 89926e7ecbebb89657ae2b56ab3454364ad7b5c5 (at node_modules/@prisma/engines/query-engine-darwin)
Migration Engine     : migration-engine-cli 89926e7ecbebb89657ae2b56ab3454364ad7b5c5 (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine : introspection-core 89926e7ecbebb89657ae2b56ab3454364ad7b5c5 (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary        : prisma-fmt 89926e7ecbebb89657ae2b56ab3454364ad7b5c5 (at node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash : 89926e7ecbebb89657ae2b56ab3454364ad7b5c5
Studio               : 0.399.0
```

Internal Slack https://prisma-company.slack.com/archives/C016KUHB1R6/p1623327462109600

millsp:
Fixes https://github.com/prisma/prisma/issues/7584

https://github.com/prisma/e2e-tests/actions/runs/937511350


janpio:
Available as `2.25.0-integration-fix-mw-tx.1` @madebysid (to fix Studio)

src/packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -915,45 +915,42 @@ new PrismaClient({
      * @param middlewareIndex
      * @returns
      */
-    private _request(
+    private async _request(
```
Jolg42, line 918: Why this change here?

src/packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -915,45 +915,42 @@ new PrismaClient({
...
-                  ...changedParams,
-                },
-                ++middlewareIndex,
-              ) // recursion happens over here
-            })
-          })
+          const changedInternalParams = { ...internalParams, ...params }
+
+          // TODO remove this, because transactionId should be passed?
+          if (index > 0) delete changedInternalParams['transactionId']
```
Jolg42, line 947: TODO for now?

madebysid:
Can confirm that the version Jan quoted fixes Studio!

src/packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -915,45 +915,42 @@ new PrismaClient({
      * @param middlewareIndex
      * @returns
      */
-    private _request(
+    private async _request(
       internalParams: InternalRequestParams,
-      middlewareIndex = 0,
```
williamluke4, line 920: Why was this removed

src/packages/client/src/runtime/getPrismaClient.ts
Jolg42, line 918:
```suggestion
    private _request(
```

This has performance implications, it's not needed.

src/packages/client/src/runtime/getPrismaClient.ts
millsp: Yes

Jolg42 (approved):

williamluke4 (approved):

src/packages/client/src/runtime/getPrismaClient.ts
millsp: We will need to remove this, because middlewares with transactions makes it not to be executed as a regular transaction (removes `transactionId`)

janpio: Can we make an issue out of this so we can remember and prioritize? Thanks.

millsp: This is preventing the transactions from rolling back. If enabled, transactions do rollback, but introduce those bugs we've seen on studio. https://github.com/prisma/prisma/issues/6705#issuecomment-861055614

pg38scf8dc (commented):
Норм

PR #3353, opened 2020-08-19
fix(generator-helper): missing field types

marcjulian:
* relationFromFields
* isUpdatedAt
* isReadOnly

Example `field` output from a Prisma generator

```json
{
    "name": "user",
    "kind": "object",
    "isList": false,
    "isRequired": true,
    "isUnique": false,
    "isId": false,
    "isReadOnly": false,
    "type": "User",
    "hasDefaultValue": false,
    "relationName": "ProfileToUser",
    "relationFromFields": [
        "userId"
    ],
    "relationToFields": [
        "id"
    ],
    "relationOnDelete": "NONE",
    "isGenerated": false,
    "isUpdatedAt": false
},
```

<details>
<summary>schema.prisma</summary>

```prisma
model User {
  id      Int      @id @default(autoincrement())
  email   String   @unique
  name    String?
  posts   Post[]
  profile Profile?
  role    Role
}

model Profile {
  id     Int     @default(autoincrement()) @id
  bio    String?
  user   User    @relation(fields: [userId], references: [id])
  userId Int     @unique
}

model Post {
  id        Int     @id @default(autoincrement())
  title     String
  content   String?
  published Boolean @default(false)
  author    User?   @relation(fields: [authorId], references: [id])
  authorId  Int?
}

enum Role {
  ADMIN
  USER
}
```
</details>

PR #7668, opened 2021-06-15
fix(generator-helper): missing field types

benkroeger:
Changes from  #3353 must have gotten lost along the way. This PR reintroduces the same props.

- relationFromFields
- isUpdatedAt
- isReadOnly

Example field output from a Prisma generator

```javascript
{
    name: 'createdBy',
    kind: 'object',
    isList: false,
    isRequired: false,
    isUnique: false,
    isId: false,
    isReadOnly: false,
    type: 'User',
    hasDefaultValue: false,
    relationName: 'CreatedQuestions',
    relationFromFields: ['createdById'],
    relationToFields: ['id'],
    relationOnDelete: 'NONE',
    isGenerated: false,
    isUpdatedAt: false,
  }
```


src/packages/generator-helper/src/dmmf.ts
```diff
@@ -75,11 +75,14 @@ export namespace DMMF {
     isList: boolean
     isUnique: boolean
     isId: boolean
+    isReadOnly: boolean
```
millsp, line 78: @Jolg42 any idea about this?

Jolg42: Looks like it's a thing, though I lack context where it's actually "true". Someone with Query Engine knowledge might know more.

Might be set to true on the virtual relations from what I can see, like https://www.prisma.io/docs/concepts/components/prisma-schema/relations

benkroeger: Didn't find where this prop is set - but it surely is [in the tests](https://github.com/prisma/prisma/blob/master/src/packages/client/src/__tests__/__helpers__/dmmf-types.ts#L19)

Jolg42 (approved):

Issue #4039, opened 2020-10-26
Preview feature feedback: `microsoftSqlServer`


2color:
Please share your feedback about the preview of Microsoft SQL Server support that was released in v2.10.0 in this issue.

- If you encounter a bug, please [open a bug report](https://github.com/prisma/prisma/issues/new?assignees=&labels=&template=bug_report.md&title=) in this repo.
- If the feature is working well for you, please share this in a comment below or leave a 👍 on this issue.

If you have any questions, don't hesitate to ask them in the #prisma-client channel in the [Prisma Slack](https://slack.prisma.io).

Related epic: https://github.com/prisma/prisma/issues/4081

chuanqisun:
Hi @2color, does this include Azure SQL? or it's just the on prem MS SQL?

pimeys:
Hi @chuanqisun, we do support Azure SQL due to it being SQL Server 2017 underneath. Most of our tests for now run against the dockerized versions of SQL Server 2017 and 2019, but Azure SQL is a first priority for us.

josh-hemphill:
I created a feature request separately in #6823 , but it probably relates to this.

Even thought SQL Server doesn't support Enums, it would be really nice to still be able to write a schema with enums for the intellisense and type information etc. If I ever have to change the database provider it would also be really nice to only have to worry about native type support/conversions and not about how the schema was structured.

Perhaps there could be some sort of flag to enforce the strong enforcement of enum support that if disabled would create a closest analog in SQL Server that at least expresses the intent of the column being an enum to those working directly with the database.

luomavaltteri:
Howdy!
We are using Prisma with Azure SQL in our project and it has worked well this far! 🚀

Two things we noticed:
- When connecting to Azure SQL, you need to escape at least the user in the connection string (because of the "@" between username and server name). Figuring out how to do that in JDBC connection strings took a bit of time (put the value in curly braces), as that wasn't needed with the SQL server we used locally for development.
- Compared to migration scripts Entity Framework generates, Prisma's migration scripts don't wrap them in transactions or track the migration history in the scripts themselves. I'm not an expert on the topic, so I'm not sure if these things are really needed, but maybe a thing to look at? Below a sample of the EF migration script

```
BEGIN TRANSACTION;
GO

<migrate stuff>

INSERT INTO [__EFMigrationsHistory] ([MigrationId], [ProductVersion])
VALUES (N'20XXXXXXXXXX_AddObject', N'5.0.0');
GO

[...]

PR #7671, opened 2021-06-15
fix(cli): add `--datasource-provider` to `init` command

Jolg42:
Related prisma/prisma#7667

Todo:
- add param in help 
- add param in docs

For release notes

Original Issue 
https://github.com/prisma/prisma/issues/2779

### `prisma init` now accepts a `--datasource-provider` argument

The [`prisma init`](https://www.prisma.io/docs/reference/api-reference/command-reference#init) command now accepts a `--datasource-provider` argument that lets you configure the default `provider` for the initially generated `datasource` block in your Prisma schema. The possible values for this argument are equivalent to the [allowed values](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#fields) for the `provider` field on `datasource` blocks:

- `postgresql`
- `mysql`
- `sqlite`
- `sqlserver` ([Preview](https://github.com/prisma/prisma/issues/4039))

Here's an example that shows how to configure the initial Prisma schema skeleton with a SQLite database:

```
npx prisma init --datasource-provider sqlite
```


millsp (approved):

src/packages/cli/src/Init.ts
```diff
@@ -102,7 +101,7 @@ export class Init implements Command {
       '--help': Boolean,
       '-h': '--help',
       '--url': String,
-      // '--provider': String,
+      '--provider': String,
```
janpio, line 104:
```suggestion
      '--datasource-provider': String,
```

src/packages/cli/src/Init.ts
```diff
@@ -87,13 +87,29 @@ export class Init implements Command {
     return new Init()
   }
 
-  // TODO add --provider & --url
   private static help = format(`
-    Setup a new Prisma project
+  Setup a new Prisma project
```
janpio, line 91: Changing indentation will not have an effect on the CLI output?

janpio (approved):
<3

millsp (approved):

src/packages/cli/src/Init.ts
Jolg42: No, and here is the new output 
Screen Shot 2021-06-25 at 19 35 16

Jolg42: I spotted a misalignment, I'll fix

src/packages/cli/src/Init.ts
```diff
@@ -87,13 +87,31 @@ export class Init implements Command {
...
     
-    ${chalk.bold('Usage')}
+  ${chalk.bold('Usage')}
 
-      ${chalk.dim('$')} prisma init
+    ${chalk.dim('$')} prisma init [options]
+
```
janpio, line 96:
```suggestion

```
😝

Jolg42:
This was shipped in 2.26.0 see how to use it in the release notes at https://github.com/prisma/prisma/releases/tag/2.26.0

PR #7686, opened 2021-06-15
chore: remove unused printDatamodelDiff

Jolg42:
Related https://github.com/prisma/prisma/pull/7646


janpio:
Quite some advanced code here it seems.

Jolg42:
Yes but unused 🤷‍♂️, we can always come back to git history if we need one day.

PR #7730, opened 2021-06-18
chore(NodeEngine): comments about code & tech debt

janpio:
It's now an old PR (June)

Todo:
- [ ] check what is outdated since it's been a while now.

### DIFF
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -542,7 +542,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
       }
     }
 
-    // TODO What is this and why?
+    // overriding the datasource from client constructor (https://github.com/prisma/prisma/pull/7730/files#r657089170)
     if (this.datasources) {
       env.OVERWRITE_DATASOURCES = this.printDatasources()
     }
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -357,7 +357,7 @@ You may have to run ${chalk.greenBright(
       this.incorrectlyPinnedBinaryTarget = this.platform
     }
 
-    // TODO
+    // TODO Fallback to the var above - but why? Shouldn't it be the same already?
     this.platform = this.platform || platform
 
     // TODO Special case for filename `NodeEngine` - Why?
@@ -485,7 +485,7 @@ This means you should very likely pin the platform ${chalk.greenBright(
 ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
     }
 
-    // TODO What and why?
+    // Make executable on non Windows platforms
     if (process.platform !== 'win32') {
       plusX(prismaPath)
     }
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -506,7 +506,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
     return printGeneratorConfig(fixedGenerator)
   }
 
-  // TODO ??
+  // TODO give better name (see usages)
   private printDatasources(): string {
     if (this.datasources) {
       return JSON.stringify(this.datasources)
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -209,7 +209,7 @@ export class NodeEngine implements Engine {
       this.port = Number(url.port)
     }
 
-    // TODO
+    // TODO What is this exactly? I have trouble understanding what the error message really tells me via this conditional.
     if (this.platform) {
       if (
         !knownPlatforms.includes(this.platform as Platform) &&
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -493,7 +493,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
     return prismaPath
   }
 
-  // TODO
+  // TODO Fixed as in "not broken" or fixed as in "written down"? If any of these, why and how and where?
   private getFixedGenerator(): string {
     const fixedGenerator = {
       ...this.generator!,
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -352,7 +352,7 @@ You may have to run ${chalk.greenBright(
     }
 
     const platform = await this.getPlatform()
-    // TODO
+    // TODO How can we describe this in words? Why "pinned" in the variable name?
     if (this.platform && this.platform !== platform) {
       this.incorrectlyPinnedBinaryTarget = this.platform
     }
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -202,7 +202,7 @@ export class NodeEngine implements Engine {
       (e) => !removedFlags.includes(e),
     )
     
-    // custom endpoint
+    // internal only: custom endpoint (see https://prisma-company.slack.com/archives/CEYCG2MCN/p1592992017273200?thread_ts=1592985426.262900&cid=CEYCG2MCN)
     this.engineEndpoint = engineEndpoint
     if (engineEndpoint) {
       const url = new URL(engineEndpoint)
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -542,6 +542,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
       }
     }
 
+    // TODO What is this and why?
     if (this.datasources) {
       env.OVERWRITE_DATASOURCES = this.printDatasources()
     }
--- a/src/packages/engine-core/src/NodeEngine.ts
+++ b/src/packages/engine-core/src/NodeEngine.ts
@@ -43,6 +43,7 @@ import { printGeneratorConfig } from './printGeneratorConfig'
 import { Undici } from './undici'
 import { fixBinaryTargets, getRandomString, plusX } from './util'
 
+
 const debug = Debug('prisma:engine')
 const exists = promisify(fs.exists)
 
@@ -72,6 +73,7 @@ const socketPaths: string[] = []
 
 const MAX_STARTS = process.env.PRISMA_CLIENT_NO_RETRY ? 1 : 2
 const MAX_REQUEST_RETRIES = process.env.PRISMA_CLIENT_NO_RETRY ? 1 : 2
+
 
 export class NodeEngine implements Engine {
   private logEmitter: EventEmitter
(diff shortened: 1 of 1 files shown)


src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -194,17 +198,18 @@ export class NodeEngine implements Engine {
         )}\` were removed, you can now safely remove them from your schema.prisma.`,
       )
     }
-
     this.previewFeatures = this.previewFeatures.filter(
       (e) => !removedFlags.includes(e),
     )
+    
+    // custom endpoint
```
janpio, line 205: What is this exactly?

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -194,17 +198,18 @@ export class NodeEngine implements Engine {
...
     this.engineEndpoint = engineEndpoint
-
     if (engineEndpoint) {
       const url = new URL(engineEndpoint)
       this.port = Number(url.port)
     }
 
+    // TODO
```
janpio, line 212: What is this exactly? I have trouble understanding what the error message really tells me via this conditional.

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -340,21 +346,26 @@ You may have to run ${chalk.greenBright(
   }> {
     const searchedLocations: string[] = []
     let enginePath
+    
     if (this.prismaPath) {
       return { prismaPath: this.prismaPath, searchedLocations }
     }
 
     const platform = await this.getPlatform()
+    // TODO
```
janpio, line 355: How can we describe this in words? Why "pinned" in the variable name?

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -340,21 +346,26 @@ You may have to run ${chalk.greenBright(
...
     if (this.platform && this.platform !== platform) {
       this.incorrectlyPinnedBinaryTarget = this.platform
     }
 
+    // TODO
```
janpio, line 360: Fallback to the var above - but why? Shouldn't it be the same already?

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -340,21 +346,26 @@ You may have to run ${chalk.greenBright(
...
     this.platform = this.platform || platform
 
+    // TODO Special case for filename `NodeEngine` - Why?
```
janpio, line 363: See comment - why?

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -364,18 +375,20 @@ You may have to run ${chalk.greenBright(
       '/tmp/prisma-engines',
     ]
 
+    // TODO Where do we use this and why?
```
janpio, line 378: See comment - why?

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -471,13 +485,15 @@ This means you should very likely pin the platform ${chalk.greenBright(
 ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
     }
 
+    // TODO What and why?
```
janpio, line 488: This is setting the executable flag on non Windows platforms I think.
Does it makes sense to do that _all the time_ and not only if a specific error is raised? I would assume this has a (small) performance impact.

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -471,13 +485,15 @@ This means you should very likely pin the platform ${chalk.greenBright(
...
     if (process.platform !== 'win32') {
       plusX(prismaPath)
     }
 
     return prismaPath
   }
 
+  // TODO
```
janpio, line 496: Fixed as in "not broken" or fixed as in "written down"?
If any of these, why and how and where?

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -490,6 +506,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
     return printGeneratorConfig(fixedGenerator)
   }
 
+  // TODO ??
```
janpio, line 509: Huh?

src/packages/engine-core/src/NodeEngine.ts
Jolg42: Looks like it's not tested nor used.

From this
```
if (engineEndpoint) {
  const url = new URL(engineEndpoint)
  this.port = Number(url.port)
}
```

It looks like engineEndpoint needs to be a valid URL and then it parses the port and uses that port to communicate with the engine.

So a feature to tell the client to use an already existing engine on a defined port?

src/packages/engine-core/src/NodeEngine.ts
Jolg42: Naming is confusing since it doesn't print, it's used for this 
https://github.com/prisma/prisma/blob/eeca05f1d460abd81012c0bbc974d2d71ac6d921/src/packages/engine-core/src/NodeEngine.ts#L532

src/packages/engine-core/src/NodeEngine.ts
janpio: But only the port, not the host (which would make this a way to talk to a hosted Engine running somewhere else maybe)?

src/packages/engine-core/src/NodeEngine.ts
janpio: There I am equally confused what it really does to be honest :/

src/packages/engine-core/src/NodeEngine.ts
Jolg42: Indeed.

Actually, it seems that Tim introduced it for Dom:

Context https://prisma-company.slack.com/archives/CEYCG2MCN/p1592992017273200?thread_ts=1592985426.262900&cid=CEYCG2MCN

>In 2.2.0-dev.2, we now have a new way to override the engine:

```
    __internal: {
      engine: {
        endpoint: `http://localhost:12345`
      }
    }
  } as any)
```

>It’s mostly for Dom for now, so he can start the engine himself, but it might come handy if you need to debug. With that you can now do things like this:
```
  const client1 = new PrismaClient()
  await client1.connect()
  const port = (client1 as any).engine.port
  const engineEndpoint = `http://localhost:${port}`
  const client2 = new PrismaClient({
    __internal: {
      engine: {
        endpoint: engineEndpoint
      }
    }
  } as any)
  const users = await client2.user.findMany()
  console.log({ users })
```

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -542,6 +542,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
       }
     }
 
+    // TODO What is this and why?
     if (this.datasources) {
       env.OVERWRITE_DATASOURCES = this.printDatasources()
     }
```
Jolg42, line 548: It's for overriding the datasource from client constructor https://github.com/prisma/prisma/issues/2443#issuecomment-630679118
```

const client1 = new PrismaClient({ datasources: { db: { url: 'postgres://localhost/db1' }} })
```

src/packages/engine-core/src/NodeEngine.ts
Jolg42: https://github.com/prisma/prisma/pull/7730#discussion_r657089170

src/packages/engine-core/src/NodeEngine.ts
janpio:
```suggestion
    // internal only: custom endpoint (see https://prisma-company.slack.com/archives/CEYCG2MCN/p1592992017273200?thread_ts=1592985426.262900&cid=CEYCG2MCN)
```

src/packages/engine-core/src/NodeEngine.ts
janpio: And then `printDatasources()` is used to actually get it, and then it is submitted to the Engine via `env.OVERWRITE_DATASOURCES`? Do we have an equivalent for this in the Node-API Engine?

Jolg42: Looks like yes https://github.com/prisma/prisma/blob/cc729c77b95c94bf9a57fe89aee6c3c043def278/src/packages/engine-core/src/NAPIEngine.ts#L92-L94

src/packages/engine-core/src/NodeEngine.ts
janpio: https://github.com/prisma/prisma/issues/8898

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -194,17 +198,18 @@ export class NodeEngine implements Engine {
         )}\` were removed, you can now safely remove them from your schema.prisma.`,
       )
     }
-
     this.previewFeatures = this.previewFeatures.filter(
       (e) => !removedFlags.includes(e),
     )
+    
+    // internal only: custom endpoint (see https://prisma-company.slack.com/archives/CEYCG2MCN/p1592992017273200?thread_ts=1592985426.262900&cid=CEYCG2MCN)
```
janpio, line 205: https://github.com/prisma/prisma/issues/8897

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -162,7 +164,10 @@ export class NodeEngine implements Engine {
     this.flags = flags ?? []
     this.previewFeatures = previewFeatures ?? []
     this.activeProvider = activeProvider
+    
     initHooks()
+    
+    // warn for removed preview feature flags
```
janpio, line 170: https://github.com/prisma/prisma/issues/8894

src/packages/engine-core/src/NodeEngine.ts
janpio:
```suggestion
    // TODO What is this exactly? I have trouble understanding what the error message really tells me via this conditional.
```

src/packages/engine-core/src/NodeEngine.ts
janpio, line 355:
```suggestion
    // TODO How can we describe this in words? Why "pinned" in the variable name?
```

src/packages/engine-core/src/NodeEngine.ts
janpio:
```suggestion
    // TODO Fallback to the var above - but why? Shouldn't it be the same already?
```

src/packages/engine-core/src/NodeEngine.ts
janpio:
```suggestion
  // TODO Fixed as in "not broken" or fixed as in "written down"? If any of these, why and how and where?
```

src/packages/engine-core/src/NodeEngine.ts
```diff
@@ -524,6 +542,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
       }
     }
 
+    // TODO What is this and why?
```
janpio, line 545:
```suggestion
    // overriding the datasource from client constructor (https://github.com/prisma/prisma/pull/7730/files#r657089170)
```

src/packages/engine-core/src/NodeEngine.ts
janpio:
```suggestion
  // TODO give better name (see usages)
```

src/packages/engine-core/src/NodeEngine.ts
janpio:
```suggestion
    // Make executable on non Windows platforms
```

PR #7680, opened 2021-06-15
test: run studio test for Node-API too

Jolg42:
For Studio test in CLI

Before enabling testing with Node-API we saw that the CLI test was never ending for we don’t know the reason.

Now that it’s enabled for Node-API
It’s both cli tests, binary and Node-API that never ends
https://github.com/prisma/prisma/pull/7680

```
Jest did not exit one second after the test run has completed.
This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

@madebysid thinks a prisma.$disconnect is probably missing somewhere.

PR #7740, opened 2021-06-18
feat(client, engine-core): Add support for `beforeExit` with Node-API

### LINKED ISSUE #7692: Support and test `beforeExit` hook
The `beforeExit` hook is a way to execute arbitrary code before the Prisma Engine is "shut down": https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/connection-management#exit-hooks

This also needs to be supported and tested for the Node-API library.

williamluke4:


src/packages/cli/package.json
```diff
@@ -110,7 +110,7 @@
     "typescript": "4.3.2"
   },
   "scripts": {
-    "test:commands": "./fixtures/test.sh && jest --maxConcurrency=1",
+    "test:commands": "./fixtures/test.sh && jest --maxConcurrency=1  --forceExit",
```
janpio, line 113: Meh, that is not good. Let's create an issue to track that we need to remove this (and other instances we added) later as it indicates something is not finishing as it should.

src/packages/engine-core/src/NAPIEngine.ts
```diff
@@ -300,9 +301,8 @@ You may have to run ${chalk.greenBright(
     }
   }
 
-  private async runBeforeStop() {
-    await this.executingQueryPromise
```
janpio, line 304: Hm, isn't this scary not to wait for an executing query to stop before we do this?

src/packages/engine-core/src/NAPIEngine.ts
```diff
@@ -355,7 +354,6 @@ You may have to run ${chalk.greenBright(
     if (this.libraryStarted) {
       // eslint-disable-next-line no-async-promise-executor
       this.libraryStoppingPromise = new Promise(async (res) => {
-        await this.runBeforeStop()
```
janpio, line 358: Uh where did this go instead? Where are we even triggering this then?

src/packages/client/src/__tests__/integration/happy/exit-hook/test.ts
```diff
@@ -1,9 +1,6 @@
 import { getTestClient } from '../../../../utils/getTestClient'
 
 test('exit-hook', async () => {
```
janpio, line 3: This test is super specific - it only tests that this is called when `SIGINT` is emitted. But isn't that only part of the use case?

Should we not also add a test case where we just run `$disconnect()` and then the `beforeExit` should be executed as well?

janpio: Tackling via https://github.com/prisma/prisma/pull/7789 to figure out if this is _actually_ the case for NodeEngine that `$disconnect` also triggers this.

src/packages/cli/package.json
williamluke4: I added this information to this [PR](https://github.com/prisma/prisma/pull/7680#issuecomment-863936516), which @madebysid will be working on

src/packages/engine-core/src/NAPIEngine.ts
williamluke4: It was removed, inorder to mirror the functionality of the `NodeEngine`

src/packages/client/src/__tests__/integration/happy/exit-hook/test.ts
williamluke4: As you found out, the Node Engine does not trigger `beforeExit` on `$disconnect`

src/packages/engine-core/src/NAPIEngine.ts
williamluke4: Their is no need

Jolg42: Why? Is it because it was only needed with binary because it is running "side-car"/separate process?
And Node-API is the same process? Though it should still be asynchronous and maybe needed no?

janpio (approved):
Tests pass, so 🪀

src/packages/engine-core/src/NAPIEngine.ts
williamluke4: Mainly because the if logic in the [beforeExitListener](https://github.com/prisma/prisma/pull/7740/files#diff-0e617d7e0e76eca325b6e16cf30d37d499ad42b8f61884a674260dd7d992c2a2R308) includes a query it will execute the [request](https://github.com/prisma/prisma/pull/7740/files#diff-0e617d7e0e76eca325b6e16cf30d37d499ad42b8f61884a674260dd7d992c2a2R402) which includes the required logic.

williamluke4: i.e We don't need to wait for the `executingQueryPromise` here

Jolg42: But `on('beforeExit')` is not the default / usually not used by users.

So by default `runBeforeExit()` does nothing since there will be no listener.

So the default behavior is different from NodeEngine?

In NodeEngine I see "$disconnect" call "stop" which await the currentRequestPromise before stopping
https://github.com/prisma/prisma/blob/feat/napi/beforeExit/src/packages/engine-core/src/NodeEngine.ts#L816-L835

williamluke4: > So by default runBeforeExit() does nothing since there will be no listener.
Correct

> So the default behavior is different from NodeEngine?
No, it is the same you can see that [here](https://github.com/prisma/prisma/blob/master/src/packages/engine-core/src/NAPIEngine.ts#L343)  stop awaits the `executingQueryPromise`

src/packages/client/src/__tests__/integration/happy/exit-hook/test.ts
janpio: Indeed. I am sad.

src/packages/engine-core/src/NAPIEngine.ts
Jolg42: Ok Thanks for the details!

millsp (approved):

PR #7746, opened 2021-06-18
fix(client): use $executeRaw, $queryRaw, $transaction in jsdoc

dimaMachina:
Fix incorrect generated `.prisma/client/index.d.ts` file


Jolg42 (approved):
Nice catch! Thanks 💚

PR #7788, opened 2021-06-21
ci: test migrate with Node-API in Buildkite

Jolg42:


src/scripts/ci/publish.ts
```diff
@@ -870,9 +870,10 @@ async function testPackages(
     order = ['@prisma/client']
   } else if (process.env.BUILDKITE_PARALLEL_JOB === '2') {
     // This is to test N-API
-    console.log('BUILDKITE_PARALLEL_JOB === 2 - running Node API tests for [integration-tests, client, sdk, prisma (cli)]')
+    console.log('BUILDKITE_PARALLEL_JOB === 2 - running Node API tests for [sdk, migrate, client, cli, integration-tests]')
```
janpio, line 873: should we maybe generate the console log output automatically from `order` to make sure this does not get out of sync in the future?

Jolg42: True

Issue #7694, opened 2021-06-15
Polish Node-API library implementation


janpio:
- ~~Clean up NapiEngine https://github.com/prisma/prisma/pull/7595~~
- Engines
	- Change filename of library file (no `napi`, better naming pattern, same file name basis for all platforms [currently `libquery` on linux and macos, but only `query` on windows]) ([internal Slack](https://prisma-company.slack.com/archives/C02232M9CP8/p1621423775006500))
	- Update codebase to use less `napi` but `node-api` (or similar) instead
- CLI/Client
	- Clean up `-v` output: Replace `N-API Query Engine` wording, Suggestion `Query Engine (Node-API)`
	- Feature: Download progress bar should include information if which engine type is being downloaded (for debugging)
	- All occurrences of `N-API` in codebase (NOT `useNAPI` as that will be ripped out soon)
	- Maybe: NodeEngine => BinaryEngine, NapiEngine => LibraryEngine
- Review `prisma:*` log output

pimeys:
what about cleaning up the topic label for node api?

janpio:
All done via the PRs linked above.

PR #7791, opened 2021-06-21
chore: rename `EngineTypes` to `BinaryType`

williamluke4:


Jolg42 (approved):
It indeed makes it easier to read!

One day we can rename "binary" across the codebase because of Node-API

janpio:
https://github.com/prisma/prisma/issues/7694 might be a place to do that :D

Jolg42:
Checked with Will, and this PR makes the code easier to understand because we had a mix of things before. So it's easier to read/understand & rename after this.

janpio:
So the code is currently consistently referring to what is the `Engine` as the `Binary` and now at least it is consistent?

janpio:
https://github.com/prisma/prisma/issues/7802

Jolg42:
Yes. The gist of this is this https://github.com/prisma/prisma/pull/7791/files#diff-51de7854c0cd4d96d3e28f1415061b8b16734f70d4ed34ba71ff2b79757acd81L35-L41

```diff
-export enum EngineTypes {
+export enum BinaryType {
  queryEngine = 'query-engine',
  libqueryEngineNapi = 'libquery-engine-napi',
  migrationEngine = 'migration-engine',
  introspectionEngine = 'introspection-engine',
  prismaFmt = 'prisma-fmt',
}
```

and EngineType is this

https://github.com/prisma/prisma/blob/6a5729f4fab91d78930c1a8f4b3eecc65600f2b2/src/packages/generator-helper/src/types.ts#L82-L87

PR #7797, opened 2021-06-22
fix(client): revert generator path lookup

### LINKED ISSUE #7439: Prisma 2.24.0 throws error `Query engine exited with code 101 - thread 'main' panicked at 'Could not open datamodel file "/schema.prisma"...'`
### Bug description

image

Everything works fine in 2.23.0

I am using [nx](github.com/nrwl/nx) with [next](https://github.com/vercel/next.js) and hosting on [vercel](https://vercel.com)

### How to reproduce

Not sure I'll be able to setup a minimal reproduction, but @janpio has access to our code to reproduce simply by upgrading to 2.24.0

### Expected behavior

_No response_

### Prisma information

Query runs during next build to do SSG. I don't believe the actual schema is relevant

### Environment & setup

- OS: Debian (on vercel)
- Database: PostgreSQL, MySQL, MariaDB or SQLite
- Node.js version: 14.16.0

### Prisma Version

```
2.24.0
```

### LINKED ISSUE #7661: Latest Prisma (2.24.x) version breaks some pnpm projects when generating models
### Bug description

I have a `pnpm` project and was running fine on Prisma 2 v2.23.x and lower.  But when I upgraded to 2.24.1 my project broke and was unable to resolve the models after running `npx prisma generate`.

I will share some simple examples of how to reproduce.  But simply put if you run `npx prisma generate` in a `pnpm` project that has subprojects that by default hoist dependencies, the latest version does not seem to correctly place the generated `@prisma/client` code in the correct place, thus the project does not know any of the generated models anymore.  I'm not sure what broke between versions but it's only in 2.24.x where this happens.

NOTE: I asked about this in the Prisma Slack channel and was advised to post this as a bug in Github.  Someone had thought a change to path resolution in 2.24.x might be the culprit.

### How to reproduce

1.  Switch to Prisma 2.24.x 
2. type `npx prisma generate`

You'll then see the following response (note the main difference in responses from 2.23.x and earlier to 2.24.x is that 2.24.x replies with `.node_modules/xxx` where 2.23.x and earlier response with `../../node_modules/xxx`):

````
➜  api git:(MyGitBranch) ✗ npx prisma generate
✔ Generated Prisma Client (2.24.1) to ./node_modules/@prisma/client in 185ms
```
```
```` 

### Expected behavior

`npx prisma generate` will correctly place the generated file in the root `node_modules` directly like so and no broken links the project will exist:

````
➜  api git:(MyGitBranch) ✗ npx prisma generate
✔ Generated Prisma Client (2.24.1) to ./../../node_modules/.pnpm/@prisma/client@2.24.1_prisma@2.24.1/node_modules/@prisma/client in 640ms
```
```
````

### Prisma information

Not sure this matters...

### Environment & setup

- OS: any
- Database: PostgreSQL
- Node.js version: v14.15.4

### Prisma Version

```
➜  api git:(mygitbranch) ✗ npx prisma -v
prisma               : 2.24.1
@prisma/client       : 2.24.1
Current platform     : darwin
Query Engine         : query-engine 18095475d5ee64536e2f93995e48ad800737a9e4 (at ../../node_modules/.pnpm/@prisma/engines@2.24.1-2.18095475d5ee64536e2f93995e48ad800737a9e4/node_modules/@prisma/engines/query-engine-darwin)
Migration Engine     : migration-engine-cli 18095475d5ee64536e2f93995e48ad800737a9e4 (at ../../node_modules/.pnpm/@prisma/engines@2.24.1-2.18095475d5ee64536e2f93995e48ad800737a9e4/node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine : introspection-core 18095475d5ee64536e2f93995e48ad800737a9e4 (at ../../node_modules/.pnpm/@prisma/engines@2.24.1-2.18095475d5ee64536e2f93995e48ad800737a9e4/node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary        : prisma-fmt 18095475d5ee64536e2f93995e48ad800737a9e4 (at ../../node_modules/.pnpm/@prisma/engines@2.24.1-2.18095475d5ee64536e2f93995e48ad800737a9e4/node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash : 18095475d5ee64536e2f93995e48ad800737a9e4
Studio               : 0.397.0
```

### PR DESCRIPTION
**Pull request by millsp**
I initially wanted to make `@prisma/client` `generator` work for symlinked workspaces like `pnpm`, `yarn`. While my attempt worked, it was too audacious as it did break some of our users projects' that relied on the client being generated in the hoisted modules folder. Despite the fact that a client can be overwritten by another one in symlinked workspaces, I believe that the way forward for this it that in such setups the client `output` needs to be manually set (something to add to the docs instead).

So in a hoisted environment, you would need something like this:
```ts
generator client {
  provider = "prisma-client-js"
  output   = "../node_modules/.prisma/client"
}
```

And you would import the client like the following:
```ts
import { PrismaClient } from '.prisma/client'
```

Fixes #7661 
Fixes #7439


Jolg42:
Note: partial revert of https://github.com/prisma/prisma/pull/7111/files#diff-d22cf9a2c4c5de20b5a5002c968e2141b52fd47042eb3498aa437aacf0848a00R28

src/packages/sdk/src/predefinedGeneratorResolvers.ts
```diff
@@ -1,14 +1,13 @@
...
 import hasYarn from 'has-yarn'
 import path from 'path'
+import resolvePkg from 'resolve-pkg'
 import { logger } from '.'
 import { getCommandWithExecutor } from './getCommandWithExecutor'
-import { findUpAsync as findUp } from './utils/find'
-
-// hide require from bundlers
-const load = require
+const debugEnabled = Debug.enabled('generator')
```
Jolg42, line 10:
```suggestion
const debugEnabled = Debug.enabled('prisma:generator')
```
Better to prefix I think

src/packages/sdk/src/predefinedGeneratorResolvers.ts
```diff
@@ -1,14 +1,13 @@
+import Debug from '@prisma/debug'
 import chalk from 'chalk'
 import execa from 'execa'
 import fs from 'fs'
 import hasYarn from 'has-yarn'
 import path from 'path'
+import resolvePkg from 'resolve-pkg'
 import { logger } from '.'
 import { getCommandWithExecutor } from './getCommandWithExecutor'
-import { findUpAsync as findUp } from './utils/find'
```
Jolg42, line 8: I think this was the only place we used this file/function. So we could then remove it?

Export 
https://github.com/prisma/prisma/blob/master/src/packages/client/src/runtime/index.ts#L33
File
https://github.com/prisma/prisma/blob/master/src/packages/client/src/runtime/utils/find.ts

src/packages/sdk/src/predefinedGeneratorResolvers.ts
```diff
@@ -80,10 +42,12 @@ export const predefinedGeneratorResolvers: PredefinedGeneratorResolvers = {
       `)
   },
   'prisma-client-js': async (baseDir, version) => {
-    let prismaClientDir = await getPrismaClientDir(baseDir)
-
+    let prismaClientDir = resolvePkg('@prisma/client', { cwd: baseDir })
     checkYarnVersion()
     checkTypeScriptVersion()
+    if (debugEnabled) {
+      console.log({ prismaClientDir })
```
Jolg42, line 49:
```suggestion
      console.debug({ prismaClientDir })
```
Would be nice

Jolg42 (changes requested):
Looks good! see comments before we can merge.

Jolg42 (approved):
Nice!

src/packages/sdk/src/predefinedGeneratorResolvers.ts
millsp: We still use this in `client` to ensure that it finds the generated client for contexts where we are not able to resolve it (eg. webpack, nextjs). We don't need it anymore in `sdk`.

Jolg42:
It looks like this PR @millsp broke e2e test 
`● prisma generate › adds @prisma/client to package.json`
Since `2.26.0-dev.19`
https://github.com/prisma/e2e-tests/runs/2888612340?check_suite_focus=true#step:6:97

Also it was broken in integration before merge https://github.com/prisma/e2e-tests/runs/2888181856?check_suite_focus=true#step:6:92

wyrd-code:
I have this issue with pnpm in a monorepo. In one workspace I generate the client, and in others I am just trying to use the types. I tried setting 

```
output = "../node_modules/.prisma/client
```

but I cannot use

```
 import type { User } from '.prisma/client'
 ```
 as it says "cannot fnd module or it's type declarations" then.
 
 Any tips? Should some instructions for this be added to docs perhaps?
 
 Thank you

janpio:
Open a new issue please - we can not help in PR comments.

PR #7813, opened 2021-06-22
chore(PRISMA_FORCE_NAPI): comments and links

janpio:
A few comments and small cleanups when going through all PRISMA_FORCE_NAPI occurrences in codebase.


src/packages/cli/src/__tests__/studio.test.ts
```diff
@@ -21,7 +21,7 @@ let studio: Studio
 const describeIf = (condition: boolean) =>
   condition ? describe : describe.skip
 
-describeIf(process.env.PRISMA_FORCE_NAPI !== 'true')('studio', () => {
+describe('studio', () => {
```
Jolg42, line 24: There is a PR for this here https://github.com/prisma/prisma/pull/7680

janpio: Nice, then merge that. It seems to generally pass here as well, so should not be a problem.

src/packages/client/src/__tests__/integration/happy/restart/test.ts
```diff
@@ -1,24 +1,31 @@
...
+
   const result = await db.user.findMany()
   expect(result.length).toBeGreaterThan(0)
 
-  if (!process.env.PRISMA_FORCE_NAPI) {
-    for (let i = 0; i < 7; i++) {
-      db._engine.child.kill()
-      await new Promise((r) => setTimeout(r, 200))
-    }
+  // kill the binary child process again, moar
```
janpio, line 23:
```suggestion
  // kill the binary child process again, to make sure it also comes back when engine is killed multiple times
```

millsp (approved):

PR #7834, opened 2021-06-23
fix(sdk): client installation when not near cli

millsp:
We noticed that the `@prisma/client` was not necessarily installed into the `node_modules` where the `prisma` cli is installed. This essentially broke our E2E tests because it would not get installed if a higher `node_modules` already had a dep to `@prisma/client` installed -- as it's the case on E2E. This PR fixes that.

Moreover, in the very same scenario, `resolve-pkg` would cache previously found packages. We run it twice, (1) time to check if `@prisma/client` exists, and (2) to see if we installed `@prisma/client` correctly (given that we did not find any in the first place).
*If anything was found in the first step, running `resolvePkg` again would yield the same results later.* But if it found nothing, no cache is created. So even though we've just installed a new package `@prisma/client` in (2), it would still return the previously found `@prisma/client` from the higher `node_modules`.

*PS: To re-activate this lovely puzzle again, replace `path.join('.', baseDir)` with `baseDir` or just do a direct call to `resolvePkg('@prisma/client', { cwd: baseDir })` and log that*

E2E https://github.com/prisma/e2e-tests/runs/2900107799

### DIFF
--- a/src/packages/sdk/src/predefinedGeneratorResolvers.ts
+++ b/src/packages/sdk/src/predefinedGeneratorResolvers.ts
@@ -7,7 +7,7 @@ import path from 'path'
 import resolvePkg from 'resolve-pkg'
 import { logger } from '.'
 import { getCommandWithExecutor } from './getCommandWithExecutor'
-const debugEnabled = Debug.enabled('prisma:generator')
+const debug = Debug('prisma:generator')
 
 export type GeneratorPaths = {
   outputPath: string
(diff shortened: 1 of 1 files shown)


Jolg42:
Note: regression (client not installing) was introduced by https://github.com/prisma/prisma/pull/7797

Jolg42:
Something to note here maybe is that we do not recommend people to use a global `prisma` CLI install but a few percent of users do that.

I'm curious to know what happens in a global CLI install context 🤔

millsp:
`require.resolve` does not resolve global packages by default, unless `NODE_PATH` is set. But in any case, the `prisma` cli gets re-installed in the local folder no matter what :)

Jolg42 (approved):
👍 

So it's about fixing something that didn't work since we introduce the "auto-install", and not a regression here.

williamluke4 (approved):
Awesome :tada: Please double check the e2e tests before merging

janpio:
Will did a PR to disable the binary check that was holding this up: https://github.com/prisma/e2e-tests/pull/1851

A new commit in this branch should create a new release to `integration` which would lead to a new e2e test run that should now succeed.

PR #7857, opened 2021-06-25
fix(cli): warn on min ts version

williamluke4:


Jolg42:
Follows https://github.com/prisma/prisma/pull/7843

src/packages/sdk/src/predefinedGeneratorResolvers.ts
```diff
@@ -44,7 +45,6 @@ export const predefinedGeneratorResolvers: PredefinedGeneratorResolvers = {
   'prisma-client-js': async (baseDir, version) => {
     let prismaClientDir = resolvePkg('@prisma/client', { cwd: baseDir })
     checkYarnVersion()
-    checkTypeScriptVersion()
```
Jolg42, line 47: Why moving this to Generate CLI command?

This TS version should only apply for `prisma-client-js` I think?

For example it doesn't apply for `prisma-client-go`

williamluke4: Good point, still sleeping

Jolg42 (approved):

PR #7859, opened 2021-06-25
chore: clean up error message

williamluke4:


src/packages/sdk/src/predefinedGeneratorResolvers.ts
```diff
@@ -174,11 +174,11 @@ function checkTypeScriptVersion() {
       const currentVersion = pjson.version
       if (semverLt(currentVersion, minVersion)) {
         logger.warn(
-          `Your ${chalk.bold(
+          `Prisma detected that your ${chalk.bold(
             'typescript',
-          )} version is ${currentVersion}, which is outdated. Please update it to ${chalk.bold(
+          )} version ${currentVersion} is outdated. If you want to use Prisma Client with typescript please update it to ${chalk.bold(
```
janpio, line 179:
```suggestion
          )} version ${currentVersion} is outdated. If you want to use Prisma Client with TypeScript please update it to version ${chalk.bold(
```

janpio: Should we maybe output `typescriptPath` in the error message as well to give context what typescript we detected? Maybe we will have many false positives, and this would make it visible to users.

If not in error messages, definitely some debug logging.

src/packages/sdk/src/predefinedGeneratorResolvers.ts
```diff
@@ -174,11 +174,11 @@ function checkTypeScriptVersion() {
       const currentVersion = pjson.version
       if (semverLt(currentVersion, minVersion)) {
         logger.warn(
-          `Your ${chalk.bold(
+          `Prisma detected that your ${chalk.bold(
             'typescript',
```
millsp, line 178:
```suggestion
            'TypeScript',
```

Jolg42 (approved):
👍

src/packages/sdk/src/predefinedGeneratorResolvers.ts
janpio: Second comment is still open I think.

Jolg42: I merged a bit too fast and missed that point, opened an issue here https://github.com/prisma/prisma/issues/7904

PR #7865, opened 2021-06-25
test: simple test for referentialAction cascade with and without flag

Jolg42:
greenfield scenario with flag

Extensive testing is done in prisma-engines.


millsp (approved):
Looks great! We could eventually remove `dev.db`.

Issue #7874, opened 2021-06-25
When a script throws multiple errors, the first one does not output a file path and line numbers


janpio:
When running the same query, that throws an error, multiple times in the same script, the first error message does not include a file path and the line numbers:

```

C:\Users\Jan\Documents\throwaway\errorRepro2 (main -> origin) (errorRepro2@1.0.0)
λ node script.js


1:
 PrismaClientValidationError:
Invalid `prisma.user.findMany()` invocation:

{
  invalidArg: true
  ~~~~~~~~~~
}

Unknown arg `invalidArg` in invalidArg for type User. Did you mean `where`? Available args:
type findManyUser {
  where?: UserWhereInput
  orderBy?: List<UserOrderByInput> | UserOrderByInput
  cursor?: UserWhereUniqueInput
  take?: Int
  skip?: Int
  distinct?: List<UserScalarFieldEnum>
}


    at Document.validate (C:\Users\Jan\Documents\throwaway\errorRepro2\node_modules\@prisma\client\runtime\index.js:32546:19)
    at NewPrismaClient._executeRequest (C:\Users\Jan\Documents\throwaway\errorRepro2\node_modules\@prisma\client\runtime\index.js:34652:17)
    at consumer (C:\Users\Jan\Documents\throwaway\errorRepro2\node_modules\@prisma\client\runtime\index.js:34597:23)
    at C:\Users\Jan\Documents\throwaway\errorRepro2\node_modules\@prisma\client\runtime\index.js:34599:47
    at AsyncResource.runInAsyncScope (async_hooks.js:197:9)
    at NewPrismaClient._request (C:\Users\Jan\Documents\throwaway\errorRepro2\node_modules\@prisma\client\runtime\index.js:34599:25)
    at Object.then (C:\Users\Jan\Documents\throwaway\errorRepro2\node_modules\@prisma\client\runtime\index.js:34707:39) {
  clientVersion: '2.25.0'
}


2:
 PrismaClientValidationError:
Invalid `prisma.user.findMany()` invocation in
C:\Users\Jan\Documents\throwaway\errorRepro2\script.js:13:23

  10 }
  11
  12 try {
→ 13   await prisma.user.findMany({
         invalidArg: true
         ~~~~~~~~~~
       })

Unknown arg `invalidArg` in invalidArg for type User. Did you mean `where`? Available args:
type findManyUser {
  where?: UserWhereInput
  orderBy?: List<UserOrderByInput> | UserOrderByInput
[...]

PR #7876, opened 2021-06-25
tests(multiple-errors-one-client): Cause multiple errors with the same Client

janpio:
Reproduction for https://github.com/prisma/prisma/issues/7874 (Failed, see comment in issue)


src/packages/client/src/__tests__/integration/errors/multiple-errors-one-client/test.ts
```diff
@@ -0,0 +1,70 @@
...
+    console.log(e)
+    expect(e).toMatchInlineSnapshot(`
+
+      Invalid \`prisma.user.findMany()\` invocation in
+      /client/src/__tests__/integration/errors/multiple-errors-one-client/test.ts:8:23
+
+         5 const prisma = new PrismaClient()
+         6 
+         7 try {
+      →  8   await prisma.user.findMany({
```
janpio, line 19: This should be different per https://github.com/prisma/prisma/issues/7874, but somehow the error does not happen here via Jest. Weird.

PR #7898, opened 2021-06-28
refactor(*): Polish Node-API library implementation

williamluke4:
https://github.com/prisma/prisma/issues/7694

### DIFF
--- a/src/packages/engine-core/src/index.ts
+++ b/src/packages/engine-core/src/index.ts
@@ -5,11 +5,12 @@ export {
   PrismaClientUnknownRequestError,
 } from './errors'
 export { getInternalDatamodelJson } from './getInternalDatamodelJson'
-export { NAPIEngine } from './NAPIEngine'
-export { BinaryEngine as Engine } from './BinaryEngine'
+export { LibraryEngine } from './LibraryEngine'
+export { BinaryEngine } from './BinaryEngine'
+export { Engine } from './Engine'
 export {
   printGeneratorConfig,
   getOriginalBinaryTargetsValue,
 } from './printGeneratorConfig'
-export * as NApiEngineTypes from './napi-types'
+export * as NodeAPILibraryTypes from './NodeAPILibraryTypes'
 export { fixBinaryTargets } from './util'
--- a/src/packages/engine-core/src/index.ts
+++ b/src/packages/engine-core/src/index.ts
@@ -6,7 +6,7 @@ export {
 } from './errors'
 export { getInternalDatamodelJson } from './getInternalDatamodelJson'
 export { NAPIEngine } from './NAPIEngine'
-export { NodeEngine as Engine } from './NodeEngine'
+export { BinaryEngine as Engine } from './BinaryEngine'
 export {
   printGeneratorConfig,
   getOriginalBinaryTargetsValue,
--- a/src/packages/fetch-engine/src/download.ts
+++ b/src/packages/fetch-engine/src/download.ts
@@ -220,7 +220,7 @@ function getCollectiveBar(options: DownloadOptions): {
 } {
   const hasNodeAPI = 'libquery-engine' in options.binaries
   const bar = getBar(
-    `Downloading${hasNodeAPI ? ' Node-API ' : ' '}Prisma engines${
+    `Downloading Prisma engines${
       hasNodeAPI ? ' for Node-API' : ''
     } for ${options.binaryTargets?.map((p) => chalk.bold(p)).join(' and ')}`,
   )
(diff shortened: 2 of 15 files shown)


src/packages/sdk/src/utils/load.ts
```diff
@@ -0,0 +1,8 @@
+/** This is a wrapper around `require` */
```
Jolg42, line 1: Would be good to say why?
I think because of https://github.com/vercel/nft?

src/packages/fetch-engine/src/download.ts
```diff
@@ -218,10 +218,11 @@ function getCollectiveBar(options: DownloadOptions): {
   finishBar: () => void
   setProgress: (sourcePath: string) => (progress: number) => void
 } {
+  const hasNodeAPI = 'libquery-engine' in options.binaries
   const bar = getBar(
-    `Downloading Prisma engines for ${options.binaryTargets
-      ?.map((p) => chalk.bold(p))
-      .join(' and ')}`,
+    `Downloading Prisma engines${
+      hasNodeAPI ? ' with Node-API' : ''
```
Jolg42, line 224:
```suggestion
      hasNodeAPI ? ' for Node-API' : ''
```

src/packages/sdk/src/utils/load.ts
williamluke4: I'm not 100% sure, will ask @millsp when he gets in

millsp: This is to avoid `eval` and hide `require` away from bundlers

millsp: While the try catch is just to silence esbuild, I think.

Jolg42:
```suggestion
/** 
 ** This is a wrapper around `require` 
 ** This is to avoid eval and hide require away from bundlers
*/
```

src/packages/engine-core/src/index.ts
```diff
@@ -5,11 +5,12 @@ export {
...
-export { NAPIEngine } from './NAPIEngine'
-export { NodeEngine as Engine } from './NodeEngine'
+export { LibraryEngine } from './LibraryEngine'
+export { BinaryEngine } from './BinaryEngine'
+export { Engine } from './Engine'
 export {
   printGeneratorConfig,
   getOriginalBinaryTargetsValue,
 } from './printGeneratorConfig'
-export * as NApiEngineTypes from './napi-types'
```
millsp, line 14: I prefer using a `d.ts` in those instances

src/packages/sdk/src/engine-commands/getVersion.ts
```diff
@@ -1,23 +1,31 @@
...
+import { load } from '../utils/load'
 
 const debug = Debug('prisma:getVersion')
 
 const MAX_BUFFER = 1_000_000_000
 
 export async function getVersion(
   enginePath?: string,
-  binaryName: BinaryType = BinaryType.queryEngine,
+  binaryName?: BinaryType,
```
millsp, line 15: Maybe call this `binaryType` or rename the type to `binaryName`

src/packages/fetch-engine/src/download.ts
```diff
@@ -218,10 +218,11 @@ function getCollectiveBar(options: DownloadOptions): {
   finishBar: () => void
   setProgress: (sourcePath: string) => (progress: number) => void
 } {
+  const hasNodeAPI = 'libquery-engine' in options.binaries
   const bar = getBar(
-    `Downloading Prisma engines for ${options.binaryTargets
-      ?.map((p) => chalk.bold(p))
-      .join(' and ')}`,
+    `Downloading Prisma engines${
```
millsp, line 223:
```suggestion
    `Downloading${hasNodeAPI ? ' Node-API ' : ' '}Prisma engines${
```

millsp (approved):
Looks good! I've left a bunch of optional suggestions

src/packages/fetch-engine/src/download.ts
```diff
@@ -218,10 +218,11 @@ function getCollectiveBar(options: DownloadOptions): {
   finishBar: () => void
   setProgress: (sourcePath: string) => (progress: number) => void
 } {
+  const hasNodeAPI = 'libquery-engine' in options.binaries
```
janpio, line 221: Is it possible this does not work on Windows @williamluke4?

williamluke4: Don't think so as if you are using NodeAPI then `libquery-engine` will be present in `options.binaries`. Why do you think that this would work on windows?

williamluke4: No this should still work

janpio: See Slack thread for further details and screenshots.

PR #7902, opened 2021-06-28
fix(client): printStack for Node-API

williamluke4:


src/packages/client/src/runtime/utils/printStack.ts
```diff
@@ -77,13 +77,16 @@ function parseStack({
   const trace = stack.reverse().find((t) => {
     return (
       t.file &&
+      t.file !== '<anonymous>' &&
       !t.file.includes('@prisma') &&
       !t.file.includes('getPrismaClient') &&
+      !t.file.startsWith('internal/') &&
       !t.methodName.includes('new ') &&
       !t.methodName.includes('_getCallsite') &&
       t.methodName.split('.').length < 4
```
Jolg42, line 86: These conditions probably all make sense somehow but they are a bit mysterious 🤔 

What is `t.file !== '<anonymous>'`? a Node-API thing only maybe?

And what is `!t.file.startsWith('internal/') &&` I don't remember files starting like this 🤔

millsp: Would also love to know this, @williamluke4

williamluke4: So what I can gather is that `internal/` are internal nodejs files that are returned in the stack trace and `<anonymous>` seems to happen when there is a `try catch`.

Jolg42 (approved):

janpio:
I thought this would fix this issue, https://github.com/prisma/prisma/issues/7874, but seems it is still present even after this PR has been merged 😭 https://github.com/prisma/prisma/issues/7874#issuecomment-873260363

PR #7931, opened 2021-06-29
fix(deps): update undici to 4.0.0

### LINKED ISSUE #7539: Upgrade to undici@4
We have a long standing bug for the combination of Undici and Webpack: https://github.com/prisma/prisma/issues/6899 + https://github.com/prisma/prisma/issues/6564 Unfortunately there is no _full_ workaround.

Undici@4 will most probably fix this, and we should upgrade to that as soon as possible. 

~~To make sure this is possible, we should also try to upgrade to one of the recent release candidates as well and help them get that over the line, and make sure we have the necessary adaptations ready when the stable version is released.~~

michalkvasnicak:
Closes #7539

### DIFF
--- a/src/packages/client/helpers/build.js
+++ b/src/packages/client/helpers/build.js
@@ -1,6 +1,7 @@
 const execa = require('execa')
-const fs = require('fs')
+const fs = require('fs-extra')
 const chalk = require('chalk')
+const path = require('path')
 const { promisify } = require('util')
 const esbuild = require('esbuild')
 const copyFile = promisify(fs.copyFile)
@@ -31,8 +32,15 @@ async function main() {
       target: 'node12',
       outfile: 'generator-build/index.js',
       entryPoints: ['src/generator.ts'],
-      external: ['_http_common']
     }),
+    // copy wasm files, etc necessary for undici
+    fs.copy(
+      path.resolve(
+        __dirname,
+        '../node_modules/@prisma/engine-core/node_modules/undici/lib/llhttp',
+      ),
+      path.resolve(__dirname, '../generator-build/llhttp'),
+    ),
   ])
 
   await Promise.all([
@@ -42,7 +50,6 @@ async function main() {
       target: 'node12',
       outdir: 'runtime',
       entryPoints: ['src/runtime/index.ts'],
-      external: ['_http_common']
     }),
     esbuild.build({
       platform: 'node',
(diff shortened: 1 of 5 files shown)


src/packages/engine-core/src/undici.ts
```diff
@@ -1,33 +1,36 @@
 import getStream = require('get-stream')
-import { Client, Pool } from 'undici'
+import { Client, Dispatcher, Pool } from 'undici'
 import { URL } from 'url'
 export class Undici {
   private pool: Pool
   private closed = false
+  private url: string | URL
```
michalkvasnicak, line 7: not sure about this one. `unidici` requires `origin` in `Dispatcher.DispatchOptions` even if it isn't required in their tests. I can remove this code and use `Omit<Dispatcher.DispatchOptions, 'origin'>` to fix that and get rid of this code.

janpio:
I unblocked the GH Actions run, so you should get test run feedback on this soon @michalkvasnicak

src/packages/client/helpers/build.js
```diff
@@ -31,8 +32,15 @@ async function main() {
       target: 'node12',
       outfile: 'generator-build/index.js',
       entryPoints: ['src/generator.ts'],
-      external: ['_http_common']
     }),
+    // copy wasm files, etc necessary for undici
+    fs.copy(
```
michalkvasnicak, line 37: I'm not familiar enough with esbuild. The problem is that undici uses `path.resolve` to establish a path to wasm files (https://github.com/nodejs/undici/blob/e705509ab22ab80aadf0fc1a394afff7dc014fdf/lib/client.js#L390-L398) and I wasn't able to find a plugin that could resolve that and copy files accordingly.

src/packages/engine-core/package.json
```diff
@@ -51,7 +51,7 @@
     "new-github-issue-url": "^0.2.1",
     "p-retry": "^4.2.0",
     "terminal-link": "^2.1.1",
-    "undici": "3.3.6"
+    "undici": "^4.0.0"
```
michalkvasnicak, line 54: I don't know why but pnpm also keeps the version 3.x.x which causes build of `cli` to fail if I remove `_http_common` from `esbuild.external`.

millsp:
We need to investigate whether `undici@4` will make bundling harder for our users (since it ships with a `lhttp` binary. Thoughts @michalkvasnicak?

hugbubby:
https://media.giphy.com/media/GrUhLU9q3nyRG/source.gif
^ live broadcast of me waiting for this patch since April

millsp:
Closing this in favor of https://github.com/prisma/prisma/pull/8842. Your contribution was included there. Thanks @michalkvasnicak!

src/packages/engine-core/package.json
celador: also experiencing errors when  _http_common is removed from Webpack's externals

PR #7982, opened 2021-07-01
feat(m1): Download and use M1 build

janpio:


williamluke4:
closing as the `fetch-engine` and `get-platform` have been moved to the engines-wrapper

PR #8034, opened 2021-07-03
temp: questions about db seed

janpio:


src/packages/migrate/src/utils/seed.ts
```diff
@@ -172,6 +173,7 @@ To install them run: ${chalk.green(
       // User can customize the `ts-node` command from the package script
       if (scripts?.['ts-node']) {
         tsNodeCommand = scripts['ts-node']
+        // TODO Why does this not need getSeedScript?
```
Jolg42, line 176: Because it's "safer" and you can pass any command there.

Anyway, we are currently looking at renaming and changing the behavior here.

src/packages/migrate/src/utils/seed.ts
```diff
@@ -97,6 +97,7 @@ runSeed()
   })
 `
 
+  // TODO Document why this is necessary for TS
```
Jolg42, line 100: Because TypeScript throws errors if these extra are not passed.

We are actually considering removing that code and only execute as a script to simplify remove complexity and make things more reliable here.

janpio: Interesting, so this is/was for working around that specific properties of TS.

src/packages/migrate/src/utils/seed.ts
janpio: Just for my understanding as I am curious: What makes running the file via the `ts-node` command (Whatever it may be) safer than doing the standard `ts-node --eval`?

Jolg42: `--eval` makes it possible to run this code without having it on the filesystem which was practical for our code snippet doing the execution of export.

Running this snippet with the custom ts-node command means we don't control the command completely and I could see problems depending on the user command input here.

Issue #8033, opened 2021-07-02
`db seed` tests need to run on multiple platforms, not just Ubuntu


janpio:
As we know `db seed` has specific problems on specific platforms, hence it needs to be tested on all platforms. This should not be to big of a deal, as it does not actually talk to databases and so can be extracted from the current test suite in GH Actions and be run on all platforms.

Jolg42:
WIP in https://github.com/prisma/prisma/pull/8037

Jolg42:
Since multi-os testing was added to e2e https://github.com/prisma/e2e-tests/issues/1806

This is not as relevant as before especially if only testing db seed it would not bring more confidence at the tradeoff of more test jobs running 

So basically closing as duplicate of https://github.com/prisma/prisma/issues/7725 for now

PR #8037, opened 2021-07-03
feat(ci): Run migrate seed tests multi os in separate job

janpio:
- Adds a multi OS run of only the `seed` tests in Migrate
	- Does not use Emoji when running in Jest, adapts seed inline snapshots
	- Cleans up indentation of seed inline snapshots in general
	- Add normalization of seed snapshots for cross platform paths via snapshot serializer
- Cleans up the jest snapshot serializers a bit


SirJosh1987 (approved):

Jolg42:
Need to check this with https://github.com/prisma/prisma/pull/8146

janpio:
Yes had I known that I had created multiple stacking PRs to do different things - but most of what I did here I think is still super valuable to improve the Seed tests and tests in general (renaming the serializers and doing that cleanup). 

I would probably go top to bottom and adopt the changes I like to whatever new PR you are working on 👍

Jolg42:
https://github.com/prisma/prisma/issues/8033#issuecomment-913697006

Issue #8058, opened 2021-07-04
SQL Server: `migrate dev` (and `db push`) does not create a database


janpio:
This is the error message when you try to migrate a database that does not exist on SQL Server:

```
C:\Users\Jan\Documents\throwaway\sqlServerCycles>npx prisma migrate dev
Environment variables loaded from .env
Prisma schema loaded from prisma\schema.prisma
Datasource "db" - SQL Server

Error: Migration engine error:
Performing a TLS handshake
Trusting the server certificate without validation.
TLS handshake successful
Turning TLS off after a login. All traffic from here on is not encrypted.
Cannot open database "mydb" requested by the login. The login failed.
Database error
Database does not exist: mydb
```

(We should probably have tests for all databases that ensure this works)

tomhoule:
@Jolg42 I think the new error logging code has a problem, this is not the expected error message. I'll try to look into this today.

tomhoule:
(looking into it currently)

Jolg42:
Todo reconfirm if fixed after it gets to dev in ~20 minutes.

janpio:
No change for `2.27.0-dev.34`:

```
...

+ prisma@2.27.0-dev.34
added 2 packages from 1 contributor and audited 4 packages in 6.057s
found 0 vulnerabilities


C:\Users\Jan\Documents\throwaway\sqlServerCycles>npx prisma db push
Environment variables loaded from .env
Prisma schema loaded from prisma\schema.prisma
Datasource "db" - SQL Server
Error: Migration engine error:
Performing a TLS handshake
Trusting the server certificate without validation.
TLS handshake successful
Turning TLS off after a login. All traffic from here on is not encrypted.
Cannot open database "repro_cycles2" requested by the login. The login failed.
Database error
Database does not exist: repro_cycles2
```

Jolg42:
Sorry I think I posted to the wrong issue I wanted to post to `https://github.com/prisma/prisma/issues/8058`

tomhoule:
So I wasn't planning to look into the SQL server part of this, only the error message handling. And that is done, so the WIP with me assigned is a bit misleading. I'll still take a look tomorrow and see where the problem comes from.

tomhoule:
I did some testing in the context of https://github.com/prisma/prisma-engines/pull/2089 and the ME does try to create a database on `create-database` and returns a proper error when that happens. Maybe there is something in the CLI preventing that code from running?

Jolg42:
[...]

PR #8075, opened 2021-07-05
fix(migrate): Print the expected error message on engine error

tomhoule:
Currently, the CLI will print _all_ the stderr logs in case of regular (non-panic) migration engine errors.

I did some manual testing to confirm that this works as expected.

Relevant issues:
- https://github.com/prisma/prisma/issues/8058

### DIFF
--- a/src/packages/migrate/src/MigrateEngine.ts
+++ b/src/packages/migrate/src/MigrateEngine.ts
@@ -284,9 +284,6 @@ export class MigrateEngine {
 
             this.messages.push(json.fields.message)
 
-            if (json.fields.backtrace) {
-              this.lastError = json.fields
-            }
             if (json.level === 'ERROR') {
               this.lastError = json.fields
             }
--- a/src/packages/migrate/src/MigrateEngine.ts
+++ b/src/packages/migrate/src/MigrateEngine.ts
@@ -238,7 +238,7 @@ export class MigrateEngine {
             const stackTrace = this.messages.join('\n')
             exitWithErr(
               new RustPanic(
-                engineMessage,
+                serializePanic(engineMessage),
                 stackTrace,
                 this.lastRequest,
                 ErrorArea.LIFT_CLI,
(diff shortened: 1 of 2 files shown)


src/packages/migrate/src/MigrateEngine.ts
```diff
@@ -221,40 +223,51 @@ export class MigrateEngine {
 
         this.child.on('error', (err) => {
           console.error('[migration-engine] error: %s', err)
-          reject(err)
           this.rejectAll(err)
+          reject(err)
```
tomhoule, line 227: Unrelated, but it looks wrong. Wdyt?

src/packages/migrate/src/MigrateEngine.ts
```diff
@@ -271,9 +277,6 @@ export class MigrateEngine {
 
             this.messages.push(json.fields.message)
 
-            if (json.fields.backtrace) {
-              this.lastError = json.fields
-            }
```
tomhoule, line 276: This is not necessary.

src/packages/migrate/src/MigrateEngine.ts
```diff
@@ -221,40 +223,44 @@ export class MigrateEngine {
...
-              err = new RustPanic(
-                errorMessage,
-                messages,
+        this.child.on('exit', (code: number | null): void => {
+          const exitWithErr = (err: RustPanic | Error): void => {
+            this.rejectAll(err)
+            reject(err)
+          }
+          const engineMessage =
+            this.lastError?.message || this.messages.join('\n')
```
tomhoule, line 236: out of context bikeshedding, but I find the line length prettier is configured with on this repo shockingly short

src/packages/migrate/src/MigrateEngine.ts
Jolg42: It's easier to read with the `reject` last but since it's asynchronous it should not matter during runtime.

src/packages/migrate/src/MigrateEngine.ts
Jolg42: We agree, we were thinking switching to 100 or so soon!

Jolg42 (approved):
Looks great! Nice and clean ✨

PR #8136, opened 2021-07-07
feat: re-export @prisma/get-platform methods from @prisma/sdk

sdnts:
Need this for Studio.

I use `@prisma/get-platform`, but since it is now released with the engines, I cannot know what version to use. Since we use the correct version in the SDK, I'm re-exporting all functions from it.


Jolg42 (approved):
Looks good!

janpio (approved):

PR #7990, opened 2021-07-02
test(client,sdk,integration-tests): remove --forceExit

millsp:
Figuring why tests are blocking jest from finishing:

- [ ] migrate
- [ ] E2E
- [x] sdk
- [x] client
- [x] integration-tests
- [x] workflows

https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/prisma/e2e-tests%24+forceExit&patternType=literal
https://sourcegraph.com/search?q=context:global+repo:github.com/prisma/prisma+--forceExit+count:1000&patternType=literal

Partially fulfills https://github.com/prisma/prisma/issues/7795

Issue #7795, opened 2021-06-21
Investigate and remove usage of `--forceExit`


janpio:
- [ ] migrate
- [x] E2E
- [x] sdk
- [x] client
- [x] integration-tests
- [x] workflows

https://sourcegraph.com/search?q=context:global+repo:github.com/prisma/prisma+--forceExit+count:1000&patternType=literal
https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/prisma/e2e-tests%24+forceExit&patternType=literal

Especially if introduced for Node-API, this is something we should try to get rid of again before GA.

janpio:
Opened 2 PRs that show what is happening when you do this:
https://github.com/prisma/prisma/pull/7871
https://github.com/prisma/e2e-tests/pull/1853

janpio:
- prisma: the `migrate` tests time out: https://github.com/prisma/prisma/pull/7871/checks?check_run_id=2915976895 + https://github.com/prisma/prisma/pull/7871/checks?check_run_id=2915976946
- e2e-tests: test-runners/jest-with-multi-gen times out for the binary (not node-api): https://github.com/prisma/e2e-tests/pull/1853/checks?check_run_id=2915975499

millsp:
Work has begun over here https://github.com/prisma/prisma/pull/7989

millsp:
Still need to do this for `migrate` and `E2E`.

janpio:
Previous work on e2e: https://github.com/prisma/e2e-tests/pull/1853

Jolg42:
Reopening, still need to investigate something with migrate dev tests.

janpio:
Note-API part done then 🙇‍♀️

PR #8149, opened 2021-07-07
fix(db push): make sure to stop migrate engine when it errors

Jolg42:
Related
https://github.com/prisma/prisma/issues/7795
https://github.com/prisma/prisma/pull/7990

Thanks @millsp for finding the test that uncovered this 👍🏼

### DIFF
--- a/src/packages/migrate/package.json
+++ b/src/packages/migrate/package.json
@@ -74,7 +74,7 @@
     "format": "prettier --write .",
     "lint": "eslint --cache --fix --ext .ts .",
     "lint-ci": "eslint --ext .ts .",
-    "test": "jest --forceExit",
+    "test": "jest",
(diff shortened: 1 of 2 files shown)


millsp (approved):
Awesome!

janpio (changes requested):
Nice, good catch!

Test not passing though yet.

millsp:
If you can't reproduce this locally, let me know.

Jolg42:
Ok I'll merge main later and see what this is about

janpio:
That's the danger of the timeouts I think - they cancel the job, and do not fail it by default.

Jolg42:
Yes... but better that it times out after 10 minutes than 6 hours by default 😄

PR #8195, opened 2021-07-09
fix(init): support SqlServer and MongoDB Preview with flag

### LINKED ISSUE #7900: `prisma init` for datasource providers in Preview/EarlyAccess
## Problem

Providers in Preview, like SQL Server or in Early Access like MongoDB need an extra preview feature flag.

We should add that flag automatically when using `prisma init --datasource-provider sqlserver.

Jolg42:
Add the corresponding preview flag to the default schema automatically.

Changed step 2 for MongoDB from
```
2. Run prisma db pull to turn your database schema into a Prisma schema.
```
to
```
2. Define models in the prisma.schema file.
```

Also check if the default url is good.

Feel free to give feedback here if that makes sense.

Closes https://github.com/prisma/prisma/issues/7900

### DIFF
--- a/src/packages/cli/src/Init.ts
+++ b/src/packages/cli/src/Init.ts
@@ -20,9 +20,25 @@ import path from 'path'
 import { isError } from 'util'
 import { printError } from './prompt/utils/print'
 
-export const defaultSchema = (
-  provider = 'postgresql',
-) => `// This is your Prisma schema file,
+export const defaultSchema = (provider: ConnectorType = 'postgresql') => {
+  if (provider === 'sqlserver' || provider === 'mongodb') {
+    return `// This is your Prisma schema file,
+// learn more about it in the docs: https://pris.ly/d/prisma-schema
+
+datasource db {
+  provider = "${provider}"
+  url      = env("DATABASE_URL")
+}
+
+generator client {
+  provider        = "prisma-client-js"
+  previewFeatures = ["${
+    provider === 'sqlserver' ? 'microsoftSqlServer' : 'mongodb'
+  }"]
+}
+`
+  } else {
+    return `// This is your Prisma schema file,
 // learn more about it in the docs: https://pris.ly/d/prisma-schema
 
 datasource db {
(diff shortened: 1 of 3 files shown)


src/packages/cli/src/Init.ts
```diff
@@ -76,7 +94,7 @@ export const defaultURL = (
     case 'sqlserver':
       return `sqlserver://localhost:${port};database=mydb;user=SA;password=randompassword;`
     case 'mongodb':
-      return `mongodb://johndoe:randompassword@localhost:${port}/mydb?authSource=admin`
+      return `mongodb+srv://root:randompassword@cluster0.ab1cd.mongodb.net/mydb?retryWrites=true&w=majority`
```
millsp, line 97: Is this intended?

millsp (approved):

src/packages/cli/src/__tests__/__snapshots__/init.test.ts.snap
```diff
@@ -8,9 +8,9 @@ Environment variables loaded from .env
 
 Next steps:
 1. Set the DATABASE_URL in the .env file to point to your existing database. If your database has no tables yet, read https://pris.ly/d/getting-started
-2. Set the provider of the datasource block in schema.prisma to match your database: postgresql, mysql, sqlserver or sqlite.
-3. Run prisma db pull to turn your database schema into a Prisma data model.
-4. Run prisma generate to install Prisma Client. You can then start querying your database.
+2. Set the provider of the datasource block in schema.prisma to match your database: postgresql, mysql, sqlite, sqlserver (Preview) or mongodb (Preview).
+3. Run prisma db pull to turn your database schema into a Prisma schema.
```
janpio, line 12: Technically "data model" is correct - it is the part of the schema file with `model`s inside.

But probably fine.

janpio (approved):

src/packages/cli/src/Init.ts
Jolg42: I copy paste from example in the (not yet merged) docs, I guess it's what we want.

PR #8203, opened 2021-07-10
feat(init): Adds minimal .gitignore on prisma init

### LINKED ISSUE #3400: Create a `.gitignore` file on `prisma init`
Quoting from [this](https://github.com/prisma/prisma2/issues/1481) issue:

> Docs + examples should point out that we strongly recommend to not check in the .env file to their VCS

Should we consider creating a `.gitignore` file for the `init` command?

It would probably be as simple as:

**.gitignore**

```
node_modules
# Keep environment variables out of version control
.env
```

ShubhankarKG:
Fixes #3400

Please let me know if anything else is required as well.


src/packages/cli/src/Init.ts
```diff
@@ -82,6 +82,14 @@ export const defaultURL = (
   }
 }
 
+export const defaultGitIgnore = () => {
+  return `
+node_modules
+# Keep environment variables out of version control
+.env
+  `
```
janpio, line 90:
```suggestion
`
```
no trailing spaces in file necessary

src/packages/cli/src/Init.ts
```diff
@@ -82,6 +82,14 @@ export const defaultURL = (
   }
 }
 
+export const defaultGitIgnore = () => {
+  return `
+node_modules
```
janpio, line 87: Does this create the file with an empty line at the beginning? If so, that is probably undesired.

src/packages/cli/src/Init.ts
```diff
@@ -109,8 +117,8 @@ export class Init implements Command {
   
   Setup a new Prisma project and specify the url that will be used
     ${chalk.dim(
-      '$',
-    )} prisma init --url mysql://user:password@localhost:3306/mydb
+    '$',
+  )} prisma init --url mysql://user:password@localhost:3306/mydb
```
janpio, line 121: unwanted indentation change

src/packages/cli/src/Init.ts
```diff
@@ -246,12 +254,17 @@ export class Init implements Command {
         fs.appendFileSync(
           envPath,
           `\n\n` +
-            '# This text is inserted by `prisma init`:\n' +
-            defaultEnv(url),
+          '# This text is inserted by `prisma init`:\n' +
+          defaultEnv(url),
```
janpio, line 258: unwanted indentation change

janpio (commented):
This is great, thank!

Left a few comments.

Can you check if we have any related tests and if so add one that makes sure the file actually gets created? Otherwise it is easy to break this functionality in the future and not notice.

ShubhankarKG:
I added a test that confirms the existence of a `.gitignore`. Can I have some pointers on how to actually run my local build? My main aim is to ensure that a .gitignore with no new lines at the start or end are formed.

janpio:
Our setup should be documented at https://github.com/prisma/prisma/blob/master/CONTRIBUTING.md, if you are missing something let us know, optimally via an issue.

ShubhankarKG:
I added a snapshot test as well! Seems fine to me now, no blank lines at the start or end. Please let me know if anything else is required.

janpio:
I think the line break at the end is actually wanted in files usually, as it is added by just opening and closing them on most OS (Except Windows...) - correct @Jolg42?

ShubhankarKG:
Should I keep the blank line at the end then?

src/packages/cli/src/Init.ts
```diff
@@ -252,6 +258,11 @@ export class Init implements Command {
       }
     }
 
+    fs.writeFileSync(
+      path.join(outputDir, ".gitignore"),
+      defaultGitIgnore()
+    );
+
```
Jolg42, line 265: I think we should try catch this, so that if it fails it doesn't stop.
I would consider writing the .gitignore optional/not required, if we can write it that's ok, if not it's also ok.

src/packages/cli/src/Init.ts
```diff
@@ -82,6 +82,12 @@ export const defaultURL = (
   }
 }
 
+export const defaultGitIgnore = () => {
+  return `node_modules
+# Keep environment variables out of version control
+.env`
```
Jolg42, line 88:
```suggestion
.env
`
```

To have an empty line return at end of file.

janpio (approved):
Looks great to me now.

millsp:
Thanks for your contribution @ShubhankarKG :tada:

PR #8213, opened 2021-07-11
tests(exit-hook): Make exit-hook test a bit clearer

janpio:
This test was not super clear, so a few comments hopefully explain better what was going on.
Also removed outdated comment that said this actually was not implemented - it clearly is.


millsp (approved):
Thanks!

PR #8220, opened 2021-07-11
test(exit-hook): SIGINT should stop running Engine so test should not have to $disconnect manually

janpio:


Jolg42:
Closing as old, we can reopen if needed

PR #8233, opened 2021-07-12
fix(init): mongoDb fixes for prisma init

Jolg42:
Fixes to prisma init after trying it
- use mongoDb instead of lowercase version for preview flag
- missing model suggestion before I guess, I replaced the id in screenshot by the mongo equivalent we suggest in docs
 `id    String  @id @default(dbgenerated()) @map("_id") @db.ObjectId`

### DIFF
--- a/src/packages/sdk/src/getGenerators.ts
+++ b/src/packages/sdk/src/getGenerators.ts
@@ -37,7 +37,10 @@ import { resolveOutput } from './resolveOutput'
 import { extractPreviewFeatures } from './utils/extractPreviewFeatures'
 import { mapPreviewFeatures } from './utils/mapPreviewFeatures'
 import { missingDatasource } from './utils/missingDatasource'
-import { missingModelMessage } from './utils/missingGeneratorMessage'
+import {
+  missingModelMessage,
+  missingModelMessageMongoDB,
+} from './utils/missingGeneratorMessage'
 import { mongoFeatureFlagMissingMessage } from './utils/mongoFeatureFlagMissingMessage'
 import {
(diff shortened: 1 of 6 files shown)


src/packages/sdk/src/getGenerators.ts
```diff
@@ -147,11 +150,16 @@ export async function getGenerators({
   })
 
   if (dmmf.datamodel.models.length === 0) {
+    // Should have a condition if MongoDB datasource to have @id @map("_id")
```
Jolg42, line 153:
```suggestion
```

src/packages/sdk/src/getGenerators.ts
```diff
@@ -147,11 +150,16 @@ export async function getGenerators({
...
+    if (config.datasources.some((d) => d.provider.includes('mongodb'))) {
+      throw new Error(missingModelMessageMongoDB)
+    }
+
     throw new Error(missingModelMessage)
   }
 
   if (
-    config.datasources.some((d) => d.provider.includes('mongoDb')) &&
+    config.datasources.some((d) => d.provider.includes('mongodb')) &&
```
millsp, line 162: Do we lowercase the providers?

millsp (approved):
Looks good! However, I'd like to see `missingGeneratorMessage.ts` split into multiple files (maybe another time).

src/packages/sdk/src/getGenerators.ts
Jolg42: From what I tried uppercase even fails validation in the Prisma extension.

PR #8316, opened 2021-07-16
ci(update-engines): use chore as commit type

steebchen:


janpio (approved):

pantharshit00 (approved):
I matched the message with renovate while writing this(https://github.com/prisma/prisma/pull/8171) but I agree chore makes more sense.

Issue #6329, opened 2021-03-31
Error in PostgreSQL connection: Error { kind: Closed, cause: None }


narciero:
Since upgrading to `2.18.0` (now on `2.19.0`) I have been experiencing a large number (10+/day per ~1k req/day) of the below error, my application is running on Google Cloud Run and I am using Postgres 12 via Cloud SQL.

These errors happen for all different queries/mutations and there is no discernible pattern that triggers the error, they happen regardless of cold-start, mem/cpu usage, etc.

Hi Prisma Team! My Prisma Client just crashed. This is the report:
## Versions

| Name            | Version            |
|-----------------|--------------------|
| Node            | v14.16.0           | 
| OS              | linux-musl         |
| Prisma Client   | 2.19.0             |
| Query Engine    | query-engine c1455d0b443d66b0d9db9bcb1bb9ee0d5bbc511d|
| Database        | postgres 12 (Cloud SQL)|

## Query
```
mutation {
  updateOneBooking(
    where: {
      id: "X"
    }
    data: {
      status: Reserved
      events: {
        create: []
      }
      jobs: {
        upsert: [
          {
            where: {
              id: "X"
            }
            create: {
              id: "X"
              type: Email
              payload: "X"
            }
            update: {

            }
          }
        ]
      }
    }
  ) {
    id
    ...
  }
}
```

## Logs
```
ing the database: connection closed
      at NodeEngine.graphQLToJSError (/usr/src/app/node_modules/@prisma/client/runtime/index.js:29799:14)
      at NodeEngine.request (/usr/src/app/node_modules/@prisma/client/runtime/index.js:29695:24)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
      at async cb (/usr/src/app/node_modules/@prisma/client/runtime/index.js:34744:26)
      at async Object.getCurrentUser (/usr/src/app/dist/services/auth.js:65:43)
      at async ApolloServer.createContext [as context] (/usr/src/app/dist/context.js:43:20) {
    clientVersion: '2.19.0'
  }
}
prisma:engine stdout  Error in PostgreSQL connection: Error { kind: Closed, cause: None }
prisma:engine {
  error: PrismaClientUnknownRequestError2 [PrismaClientUnknownRequestError]: Error in connector: Error querying the database: connection closed
      at NodeEngine.graphQLToJSError (/usr/src/app/node_modules/@prisma/client/runtime/index.js:29799:14)
[...]

PR #8384, opened 2021-07-21
feat(client): interactive transactions

### LINKED ISSUE #1844: API for interactive transactions with dependencies between write-operations
There already is a [GitHub issue](https://github.com/prisma/prisma/issues/2358) asking for a way to submit multiple mutations within the same HTTP request where all mutations are executed as one transaction.

However, this feature does not allow for creating a transaction where the mutations depend on each other. Here is an example where the second operation depends on the first:

```js
prisma.transaction(async tx => {
  const user = await tx.createUser({
    name,
  })
  // compute some stuff
  await tx.updateOrder(user.id)
  await tx.commit()
})
```

This is currently not possible with the Prisma API as it would require having a long-running connection where the results of the operations are sent back and forth between Prisma and the database. 

It should be considered whether it is helpful to add such an API to Prisma, or whether abstractions like nested mutations and the [requested transactional execution of multiple _independent_ transactions](https://github.com/prisma/prisma/issues/2358) are the better approach for Prisma.

### LINKED ISSUE #6705: Transaction with bad input should rollback (when using middleware)

### LINKED ISSUE #7326: Unique constraint error inside a transaction throws unparsed error (but works fine when using Node API)
### Bug description

While trying to auto-generate slugs for user-made pages at [Copyfolio](https://copyfol.io), we noticed that unique constraint violation errors aren’t caught properly inside a `prisma.$transaction`.

### How to reproduce

1. Have a model with a unique constraint, e.g. a `Site` with a globally unique URL `slug`.

2. Run some code which is meant to detect unique `slug` constraint violation errors – and then auto-generate a new slug to try with, but that isn’t part of the code below:

```ts
/* Fill our database with some initial data */
await prisma.site.create({ id: "1", slug: "taken-by-1" });
await prisma.site.create({ id: "2", slug: "another" });

try {
  await prisma.$transaction([
    prisma.site.updateMany({
      where: { id: "2" },
      data: { slug: "taken-by-1" },
    }),

    /* Any other site update is required for the test to fail */
    prisma.site.update({
      where: { id: "2" },
      data: { id: "2" },
    }),
  ]);
} catch (error) {
  if (
    error instanceof PrismaClientKnownRequestError &&
    error.code === "P2002"
  ) {
    console.debug("Unique constraint violation detection works");
  } else {
    console.error("Unknown error, as violation detection fails");
    throw error;
  }
}
```

It should throw an unknown error, which isn’t the expected behavior. However, then using `previewFeatures = ["nApi"]`, the code above works as expected.

### Expected behavior

_No response_

### Prisma information

```prisma
model Site {
  slug String @unique
}
```

### Environment & setup

- OS: mac OS v11.4
- Database: PostgreSQL 12/13
- Node.js version: v14.16.1

### Prisma Version

```
prisma               : 2.23.0
@prisma/client       : 2.23.0
Current platform     : darwin
Query Engine         : query-engine adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/query-engine-darwin)
Migration Engine     : migration-engine-cli adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine : introspection-core adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary        : prisma-fmt adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash : adf5e8cba3daf12d456d911d72b6e9418681b28b
Studio               : 0.393.0
```

### LINKED ISSUE #7399: No PrismaClientKnownRequestError if error in transaction
### Bug description

 If you attempt to delete a row which does not exist in a transaction, you get a 

```
Error: Error occurred during query execution:
InterpretationError("Error for binding \'0\'", Some(QueryGraphBuilderError(RecordNotFound("Record to delete does not exist."))))
```

If you attempt the same without transaction, you get a PrismaClientKnownRequestError with a code P2025, which is expected.

### How to reproduce

With this code:
```ts
 try {
    const user = await dboClient.user.delete({where: {user_id: '9a535076-59f2-4145-9232-7302d82313d0'}});
    console.log(user);
  } catch (e) {
    console.log(e.toString() + '\n' + JSON.stringify(e));
  }
```
Output is the following. Error raised is a PrismaClientKnownRequestError
```
Error:
Invalid `prisma.user.delete()` invocation:
  An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist.
{"code":"P2025","clientVersion":"2.23.0","meta":{"cause":"Record to delete does not exist."}}
```

However, if you wrap this in a transaction,
```ts
  try {
    const [,user] = await dboClient.$transaction([
      dboClient.user.findUnique({where: {user_id: '9a535076-59f2-4145-9232-7302d82313d0'}}),
      dboClient.user.delete({where: {user_id: '9a535076-59f2-4145-9232-7302d82313d0'}})
    ]);
    console.log(user);
  } catch (e) {
    console.log(e.toString() + '\n' + JSON.stringify(e));
  }
```
Output is the following:
```
Error: Error occurred during query execution:
InterpretationError("Error for binding \'0\'", Some(QueryGraphBuilderError(RecordNotFound("Record to delete does not exist."))))
{"clientVersion":"2.23.0"}
```

### Expected behavior

Even in the transaction, Error should be a PrismaClientKnownRequestError with a code P2025

### Prisma information

Schema can be as simple as that
```prisma
model User {
  user_id    String       @id @db.Uuid @default(dbgenerated("gen_random_uuid()"))
}
```

### Environment & setup

- OS: Windows
- Database: PostgreSQL
- Node.js version: v14.16.0

### Prisma Version

```
prisma               : 2.23.0
@prisma/client       : 2.23.0
Query Engine         : query-engine adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules\@prisma\engines\query-engine-windows.exe)
Migration Engine     : migration-engine-cli adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules\@prisma\engines\migration-engine-windows.exe)
Introspection Engine : introspection-core adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary        : prisma-fmt adf5e8cba3daf12d456d911d72b6e9418681b28b (at node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash : adf5e8cba3daf12d456d911d72b6e9418681b28b
Studio               : 0.393.0

```

### LINKED ISSUE #7786: No PrismaClientKnownRequestError when unique constraint fails in transaction
### Bug description

If you attempt to insert a row while in a transaction and a unique constraint fails, the error sent is:
```
ConnectorError(ConnectorError { user_facing_error: Some(KnownError { message: "Unique constraint failed on the fields: (`user_id`)", meta: Object({"target": Array([String("user_id")])}), error_code: "P2002" }), kind: UniqueConstrai
ntViolation { constraint: Fields(["user_id"]) } })
```

If not in a transaction, error is a PrismaClientKnownRequestError with code P2002

### How to reproduce

**Without transaction**
```ts
 try {
    const user = await dboClient.user.create({data: {account_id: '1fe4ed1d-afe7-482c-be6b-430179776c4a', user_id: '9a535076-59f2-4145-9232-7302d82313d0'}});
    console.log(user);
  } catch (e) {
    console.log(e.toString() + '\n' + JSON.stringify(e));
  }
```
Error:
```
Invalid `prisma.user.create()` invocation:
  Unique constraint failed on the fields: (`user_id`)
{"code":"P2002","clientVersion":"2.23.0","meta":{"target":["user_id"]}}

```

**With transaction**
```ts
  try {
    const [,user] = await dboClient.$transaction([
      dboClient.user.findUnique({where: {user_id: '9a535076-59f2-4145-9232-7302d82313d0'}}),
      dboClient.user.create({data: {account_id: '1fe4ed1d-afe7-482c-be6b-430179776c4a', user_id: '9a535076-59f2-4145-9232-7302d82313d0'}})
    ]);
    console.log(user);
  } catch (e) {
    console.log(e.toString() + '\n' + JSON.stringify(e));
  }
```
Error:
```
Error: Error occurred during query execution:
ConnectorError(ConnectorError { user_facing_error: Some(KnownError { message: "Unique constraint failed on the fields: (`user_id`)", meta: Object({"target": Array([String("user_id")])}), error_code: "P2002" }), kind: UniqueConstrai
ntViolation { constraint: Fields(["user_id"]) } })
```

### Expected behavior

Even in the transaction, Error should be a PrismaClientKnownRequestError with a code P2002

### Prisma information

```
model User {
  user_id    String       @id @db.Uuid @default(dbgenerated("gen_random_uuid()"))
}
```

### Environment & setup

- OS: Windows
- Database: PostgreSQL
- Node.js version: v14.16.0

### Prisma Version

```
prisma               : 2.25.0
Query Engine         : query-engine c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\@prisma\engines\query-engine-windows.exe)
Migration Engine     : migration-engine-cli c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\@prisma\engines\migration-engine-windows.exe)
Introspection Engine : introspection-core c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary        : prisma-fmt c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\prisma\node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash : c838e79f39885bc8e1611849b1eb28b5bb5bc922
Studio               : 0.402.0

```

### LINKED ISSUE #7956: LRT: Interactive Transactions Epic
## Problem

Long-running transactions are a very commonly used functionality of database that has influenced how a lot of developers think and build their applications. 

While there are alternative ways to conceive a feature without them, leading to more scalable code, these ways aren't obvious and can complicate the code of applications that wouldn't suffer from long-running transaction shortcomings. 

Moreover, users can't bear the cost of porting pre-existing application heavily relying on long-running transactions to a new paradigm.

## Solution

Our new transaction API will allow you to add arbitrary logic within a transaction. This gives users a simple mental model for working with transactions and reduces the burden of porting existing applications to Prisma.

<details>
<summary>Code Example</summary>

```ts

async function main() {

  const posts = await prisma.$transaction(async (tx) => {
    // we can do our usual queries/await them
    const user = await tx.user.findUnique({
      where: {
        email: 'john@doe.io'
      }
    })

    if (!user) {
      // we can also throw errors of our own
      throw new Error('User does not exist')
    }

    // we can process long running operations
    const copiedPost1 = await fetchData(1)

    // data can be persisted after "long" ops
    const post1 = await tx.post.create({
      data: {
        title: copiedPost1['title'],
        content: copiedPost1['content'],
        author: {
          connect: {
              email: 'john@doe.io'
          }
        }
      }
    })

    // unhandled exceptions trigger rollback
    const post2 = await tx.post.create({
      data: {
        title: undefined as any as string,
        content: undefined as any as string,
        author: {
          connect: {
              email: 'wrong@email.io'
          }
        }
      }
    })

    // we can return any custom return value
    return [post1, post2]
  })
}

// PS: we agree that this is a bad practice
function fetchData(id: string | number) {
  return fetch(`https://myservice.co/api/${id}`)
  .then(response => response.json()) as
  Promise<{title: string, content: string}>
}
```

</details>

## Tasks

- [x] #7957 @dpetrick @millsp 

### TypeScript Client

- [x] #8000 @millsp
- [x] #8090
- [x] #8091 
- [x] #8087
- [x] #8088
- [x] #8092 
- [x] #8093 
- [x] #8094 
- [x] #8095 
- [x] #8096 
- [x] #8097 
- [x] #8098 
- [x] #8099 
- [x] #8100
- [x] #8101
- [x] #8277

### Query Engine
- [x] #8016
- [ ] #8017
- [x] #8018
- [x] #8019

**Note:** The following issues might be easy to fix while implementing LRT, but not necessary for completing the epic.

- [x] https://github.com/prisma/prisma/issues/7786
- [x] https://github.com/prisma/prisma/issues/7399

## References

- Community issue: https://github.com/prisma/prisma/issues/1844
- [Internal Document](https://www.notion.so/prismaio/Long-Running-Transactions-43d624709fbd413e8b17b56bdfd2cbd2)

millsp:
closes https://github.com/prisma/prisma/issues/1844
closes https://github.com/prisma/prisma/issues/7956
closes https://github.com/prisma/prisma/issues/7786
closes https://github.com/prisma/prisma/issues/7399
closes https://github.com/prisma/prisma/issues/6705
closes https://github.com/prisma/prisma/issues/7326

[Internal doc](https://www.notion.so/prismaio/Interactive-Transactions-Design-6b74dddcbf1d4d65b403ff601fc3bf80)


src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
```diff
@@ -0,0 +1,374 @@
...
+    const result = prisma.$transaction((prisma) => {
+      // we accumulate all the forbidden methods and expect to be undefined
+      return forbidden.reduce((acc, item) => acc ?? prisma[item], undefined)
+    })
+
+    await expect(result).resolves.toBe(undefined)
+  })
+
+  /**
+   * If one of the query fails, all queries should cancel
```
Sytten, line 158: Add a test for query failing with bad input (aka not sent to the engine).

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
```diff
@@ -0,0 +1,374 @@
...
+      }),
+    ])
+
+    const users = await prisma.user.findMany()
+
+    expect(users.length).toBe(2)
+  })
+
+  /**
+   * Middlewares should not prevent a rollback
```
Sytten, line 299: Add test with middleware and bad input (aka not sent to the engine).

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
```diff
@@ -0,0 +1,374 @@
...
+      })
+
+      return prisma.user.findMany()
+    })
+
+    expect(result.length).toBe(2)
+  })
+
+  /**
+   * Transactions should fail after the default timeout
```
Sytten, line 30: There doesn't seem to be a test for the maxWait

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
```diff
@@ -0,0 +1,374 @@
...
+
+            Invalid \`prisma.user.findMany()\` invocation:
+
+
+              Transaction API error: Transaction already closed: Transaction is no longer valid. Last state: 'Expired'.
+          `)
+  })
+
+  /**
+   * Transactions should fail and rollback if thrown within
```
Sytten, line 86: Would need a test for the case where the commit or rollback fails.

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
```diff
@@ -0,0 +1,374 @@
...
+  Unique constraint failed on the fields: (\`email\`)
+`)
+
+    const users = await prisma.user.findMany()
+
+    expect(users.length).toBe(0)
+  })
+
+  /**
+   * Minimal example of a long-running transaction & middleware
```
Sytten, line 335: I would add a test with a middleware that call next multiple times (like a retry middleware).

src/packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -907,15 +935,107 @@ new PrismaClient({
...
+     * @param callback
+     * @param options
+     * @returns
+     */
+    private async _transactionWithCallback(
+      callback: (client: Client) => Promise<unknown>,
+      options?: { maxWait: number; timeout: number },
+    ) {
+      // we ask the query engine to open a transaction
+      const info = await this._engine.transaction('start', options)
```
Sytten, line 996: I imagine that this takes a connection for the DB pool when it opens the connection. What happens if that connection is bad? Currently we can retry by having a retry middleware but this middleware wont be called for this start thing, so I imagine this will just throw.
For us this means I need to change all places in code that uses transaction to add retry logic.
I think the retry of failed connection needs some thought put into it.

millsp: That's what `await this.start` just above does. Is that what you meant?

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
millsp: Sounds good! Do you have an example at hand?

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
millsp: There is one `already committed`, is that what you were thinking of?

src/packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -907,15 +935,107 @@ new PrismaClient({
...
+    /**
+     * Execute a batch of requests in a transaction
+     * @param requests
+     * @param options
+     */
+    private async _transactionWithRequests(
+      requests: Array<Promise<unknown>>,
+      options?: { maxWait: number; timeout: number },
+    ) {
+      return this._transactionWithCallback(async (prisma) => {
```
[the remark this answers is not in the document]
millsp: The magic is that it does not add any round-trip. It uses the underlying impl. of LRTs *but* since we don't call `await`, we are able to batch it. If you have a middleware, however, it will do that since you are requesting the result of the next one - can't get around that as I explained before.

src/packages/engine-core/src/common/types/Transaction.ts
```diff
@@ -0,0 +1,8 @@
+export type Options = {
+  maxWait?: number
+  timeout?: number
+}
```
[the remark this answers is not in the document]
millsp: By design on the Rust side, the `timeout` comprises everything. `timeout` includes what would fall under `maxWait` and that is to get a transaction opened. On our side, we don't include connection times into it, the timeout is fully controlled by the Rust Engines.

src/packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -1026,33 +1032,71 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
...
+
+        const result = await Connection.onHttpError(
+          this.connection.post<Tx.Info>('/transaction/start', jsonOptions),
+          transactionHttpErrorHandler,
+        )
+
+        return result.data
+      } else if (action === 'commit') {
+        await Connection.onHttpError(
+          this.connection.post(`/transaction/${arg.id}/commit`),
```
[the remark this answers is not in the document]
millsp: The (http) timeout is the same as any other call. But right, we need to add some retry logic. Thanks.

src/packages/engine-core/src/binary/Connection.ts
```diff
@@ -0,0 +1,141 @@
...
+  data: R
+}
+
+/**
+ * Assertion function to make sure that we have a pool
+ * @param pool
+ */
+function assertHasPool<A>(pool: A): asserts pool is NonNullable<A> {
+  if (pool === undefined) {
+    throw new Error('Connection has not been opened')
```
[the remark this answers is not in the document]
millsp: Thanks! Previously, the client would "possibly" fail on `undefined`.

src/packages/engine-core/src/binary/BinaryEngine.ts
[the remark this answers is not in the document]
millsp: If the connection dies and we cannot reconnect fast enough, the engine will close the tx.

src/packages/client/src/runtime/getPrismaClient.ts
millsp: Marking this resolved in favor of the top conversation https://github.com/prisma/prisma/pull/8384#discussion_r680353340

src/packages/engine-core/src/binary/BinaryEngine.ts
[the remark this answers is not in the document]
millsp: Thanks, makes sense :+1:

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
millsp: Ok, added to my todo list.

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
millsp: Ok, added to my todo list.

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
millsp: Ok, added to my todo list.

src/packages/client/src/runtime/getPrismaClient.ts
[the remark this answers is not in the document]
millsp: However, if you have a middleware with `await next...`, it will do that since you are requesting the result of the next one - can't get around that as I explained before. Our `DataLoader` batches all the queries that happen on the same `tick`... and `await` is causing the queries not to be on the same `tick`. So your while issue with rollbacks is solved, I'm sure that you would still sure that you'll like better batching.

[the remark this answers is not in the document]
millsp: However, if you have a middleware with `await next...`, it will do that since you are requesting the result of the next one - can't get around that as I explained before. Our `DataLoader` batches all the queries that happen on the same `tick`... and `await` is causing the queries not to be on the same `tick`.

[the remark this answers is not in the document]
millsp: I think that this is a good moment to try to change that since we're shipping a this behind a preview flag. I'm thinking of not batching per `tick`, but per time frame (eg. `20ms`) so that we can optimize your use-case. TBI

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
Sytten: Not really, it needs to test the case of a network failure or delay on the commit (aka retry logic I was talking about)

src/packages/engine-core/src/binary/BinaryEngine.ts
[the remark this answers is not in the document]
Sytten: It will close with a rollback I assume?

src/packages/engine-core/src/common/types/Transaction.ts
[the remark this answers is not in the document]
Sytten: Then there might be a race condition for the error if the pool is over used. You would get a timeout error without knowing the cause. If it was 5.5s or something then you would know the cause.

src/packages/client/src/__tests__/integration/happy/long-running-transaction/test.ts
Sytten: I would suggest my retry middleware https://github.com/prisma/prisma/issues/6329#issuecomment-816954832

src/packages/client/src/runtime/getPrismaClient.ts
Sytten: Agreed

src/packages/engine-core/src/binary/BinaryEngine.ts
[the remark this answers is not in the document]
Sytten: If we can make the retry use controllable that. would we best (eventually the same setting could be used for other calls too to eliminate the need for a custom middleware which I currently built).

src/packages/client/src/runtime/getPrismaClient.ts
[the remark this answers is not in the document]
Sytten: Interesting on the batching, I was not aware of that.
My main concern was more that it would be another network call (read another potential for network delay/failure) that could affect currently working code (aka lets try not to disrupt good working production code).

As for batching, I think it probably could support both (basically switch between setTimeout and setImmediate) and offer the choice to the user.
This would be similar has to what GCP PubSub does things, you can control the time, amount of items, total size of payload, etc. but they offer sensible defaults. In general I prefer to offer knobs to advanced users (and I am pushing prisma to adopt that mentality since big co will want that anyway).

src/packages/engine-core/src/common/types/Transaction.ts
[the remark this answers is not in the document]
millsp: I think that it would error with a `maxWait` error, in this case. I'll make sure of that. And good to add retry logic on that.

src/packages/client/src/runtime/getPrismaClient.ts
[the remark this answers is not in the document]
millsp: A forced default for the middlewares + batching use-case would be that we automatically use `setTimeout` to save on round trips. On regular batches (no middlewares), this would just delay the queries, so not interesting there.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -907,15 +935,107 @@ new PrismaClient({
       )
     }
 
-    async $transaction(promises: Array<any>): Promise<any> {
+    /**
+     * @deprecated
+     */
+    async $___transaction(promises: Array<any>): Promise<any> {
```
Jolg42, line 941: Could we find another name for it maybe? 

_ often means private which is not what it means here I think?

`$toBeDeprecatedTransaction` maybe or anything else?

packages/client/src/__tests__/integration/happy/interactive-transactions/test.ts
```diff
@@ -0,0 +1,410 @@
...
+})
+
+beforeAll(async () => {
+  PrismaClient = await getTestClient()
+
+  process.env.PRISMA_FORCE_LRT = 'true'
+})
+
+afterAll(() => {
+  delete process.env.PRISMA_FORCE_LRT
```
Jolg42, line 399: I could not find where `PRISMA_FORCE_LRT` logic is.

Also is it for internal testing only?

If we need to keep it I would recommend updating the name because LRT change to "interactive transactions"

packages/engine-core/src/library/types/Library.ts
```diff
@@ -0,0 +1,50 @@
...
+   * @param request JSON.stringified `QueryEngineRequest | QueryEngineBatchRequest`
+   *
+   * @param headers JSON.stringified `QueryEngineRequestHeaders`
+   */
+  query(
+    request: string,
+    headers: string,
+    transactionId?: string,
+  ): Promise<string>
+  sdlSchema(): Promise<string>
```
Jolg42, line 25: @williamluke4 

Any idea what is sdlSchema?

It’s not documented nor used I think.

packages/engine-core/src/library/types/Library.ts
```diff
@@ -0,0 +1,50 @@
...
+  disconnect(): Promise<void>
+  /**
+   *
+   * @param request JSON.stringified `QueryEngineRequest | QueryEngineBatchRequest`
+   *
+   * @param headers JSON.stringified `QueryEngineRequestHeaders`
+   */
+  query(
+    request: string,
+    headers: string,
```
Jolg42, line 22: Can we rename `headers`? To `headerStr` maybe. 

headers everywhere else is a `Record<string, string>` so from first look it's confusing

packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -100,6 +101,29 @@ export class LibraryEngine implements Engine {
...
+        timeout: arg?.timeout ?? 5000, // default
+      })
+
+      const result = await this.engine?.startTransaction(jsonOptions, '{}')
+      return this.parseEngineResponse<Tx.Info>(result)
+    } else if (action === 'commit') {
+      await this.engine?.commitTransaction(arg.id, '{}')
+    } else if (action === 'rollback') {
+      await this.engine?.rollbackTransaction(arg.id, '{}')
+    }
```
Jolg42, line 122: How is it possible that `engine?` might not be defined?

packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -100,6 +101,29 @@ export class LibraryEngine implements Engine {
     initHooks(this)
   }
 
+  async transaction(action: 'start', options?: Tx.Options): Promise<Tx.Info>
+  async transaction(action: 'commit', info: Tx.Info): Promise<undefined>
+  async transaction(action: 'rollback', info: Tx.Info): Promise<undefined>
```
Jolg42, line 106: void would make more sense from an API point of view?

packages/engine-core/src/common/types/Transaction.ts
```diff
@@ -0,0 +1,8 @@
+export type Options = {
+  maxWait?: number
+  timeout?: number
+}
```
Jolg42, line 4: Should we document the defaults here in a comment maybe?

packages/engine-core/src/common/types/QueryEngine.ts
```diff
@@ -47,36 +51,45 @@ export type QueryEngineConfig = {
...
+
+export type QueryEngineResult<T> = {
+  data: T
+  elapsed: number
+}
+
 export type QueryEngineRequestHeaders = {
   traceparent?: string
+  transactionId?: string
+  fatal?: string // TODO
```
Jolg42, line 73: TODO? Before merging?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1372,29 +1465,62 @@ new PrismaClient({
...
+const forbidden = ['$connect', '$disconnect', '$on', '$transaction', '$use']
+
+/**
+ * Proxy that takes over client promises to pass `transactionId`
+ * @param thing to be proxied
+ * @param transactionId to be passed down to `_query`
+ * @returns
+ */
+function transactionProxy<T>(thing: T, transactionId: string): T {
+  // we only wrap within a proxy if it's posible, if it's an object
```
Jolg42, line 1491:
```suggestion
  // we only wrap within a proxy if it's possible: if it's an object
```

packages/engine-core/src/library/LibraryEngine.ts
millsp: I think that it boils down from the fact that we do async stuff to init the engine. Maybe we should add some type assertion logic into `start` to prevent that.

packages/client/src/__tests__/integration/happy/interactive-transactions/test.ts
millsp: Good catch, thanks!

packages/engine-core/src/library/types/Library.ts
williamluke4: Can't remember, I'll have to ask Julius

williamluke4 (approved):
As always awesome work @millsp :tada:

packages/engine-core/src/binary/Connection.ts
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * Perform a GET request
+   * @param endpoint
+   * @param body
+   * @param headers
+   * @returns
+   */
+  get<R>(path: string, headers?: Client.DispatchOptions['headers']) {
+    return this.raw<R>('GET', path, headers)
+  }
```
Jolg42, line 127:
```suggestion
  get<R>(endpoint: string, headers?: Client.DispatchOptions['headers']) {
    return this.raw<R>('GET', endpoint, headers)
  }
```
to make it consistent with POST

packages/engine-core/src/common/Engine.ts
```diff
@@ -1,25 +1,41 @@
...
+  abstract getConfig(): Promise<GetConfigResult>
+  abstract version(forceRun?: boolean): Promise<string> | string
+  abstract request<T>(
     query: string,
-    headers: Record<string, string>,
-    numTry: number,
-  ): Promise<{ data: T; elapsed: number }>
-  requestBatch<T>(
+    headers?: QueryEngineRequestHeaders,
+    numTry?: number,
```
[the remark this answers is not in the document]
Jolg42: (We can do that in a separate PR)

packages/engine-core/src/library/LibraryEngine.ts
Jolg42: I see comment closed with 👍🏼 but type is still the same 🤔 
Did you want to change it?

packages/engine-core/src/library/types/Library.ts
Jolg42: We can take care of this this in a separate PR

Jolg42 (approved):
Great work ✨

PR #8472, opened 2021-07-28
chore(build, types): cleanup

### LINKED ISSUE #8443: Enable Type Discovery
## Problem

When working on the `prisma/prisma` repository, the different parts of code does not share types properly. This results in poor compilation times and the inability to follow types easily. At the moment, we compile all the types every time we build the project, even when working locally. This should only be necessary during CI.

By enabling automatic discovery of types throughout the prisma/prisma repository, development would be faster as, instead of having to remember or lookup type information, the IDE will automatically provide the information where it's needed.

## Suggested solution
Detailed is a suggested solution from @millsp :

- [x] Define a single typescript file that will be used for all of our packages and type-checks only
- [x] Define a single typescript file that will be used for all of our packages and build only
   - The type-check `tsconfig.json` file will extend the `tsconfig.build.json`
- [x] Define the base TypeScript rules that will apply to all our projects in the monorepo
- [x] Define TS type lookup paths to expose type information accross the monorepo
- [x] Provide each project with a `tsconfig.build.json` and a `tsconfig.json`
- [x] Integrate the new build files into our current build scripts for each package of the monorepo
- [x] Add a new `npm` build (call it `npm dev`) script for local dev that skips slow `tsc`
- [x] Similarly investigate why do we need to recompile all the packages `X` that depend `A` if we have changes in `A`. In a monorepo, this should not be necessary (at least at dev time).
     - Probably has to do with our bundling logic?
- [x] Do the same for `engines-wrapper`

## Additional context

This has been uncovered as a tech debt initiative and prioritized due to its high impact and relatively low complexity and estimated time for completion.

millsp:
With this PR, we can now re-compile the whole project in ~2 seconds with `pnpm run -r dev` and follow types to their source without landing on `dist` types. The compilation is now fully handled by `esbuild`.


millsp:
Test failing on `engine-core`, switching to `esbuild` somehow changed the message

millsp:
We can now compile the whole project in less than 3 seconds with `pnpm run -r dev` and follow types to their source.

tsconfig.json
```diff
@@ -1,8 +1,25 @@
...
+    "baseUrl": "./src/packages",
+    "paths": {
+      "@prisma/client": ["client/src/runtime"],
+      "@prisma/sdk": ["sdk/src"],
+      "@prisma/migrate": ["migrate/src"],
+      "@prisma/generator-helper": ["generator-helper/src"],
+      "@prisma/engine-core": ["engine-core/src"],
+      "@prisma/debug": ["debug/src"],
+      "@prisma/cli": ["cli/src"],
+      "@prisma/get-platform": ["../../../engines-wrapper/packages/get-platform/src"],
```
Jolg42, line 13: That requires us to clone the engines-wrapper repo next to prisma repo on our machine with that exact name to work then?

src/packages/sdk/helpers/build.js
```diff
@@ -0,0 +1,36 @@
...
+  outdir: 'dist',
+  tsconfig: 'tsconfig.build.json'
+}
+
+async function build() {
+  await Promise.all([
+    esbuild.build({
+      ...ESBUILD_DEFAULT,
+      entryPoints: glob.sync('./src/**/*.{j,t}s', {
+        ignore: './src/__tests__/**/*.{j,t}s'
```
Jolg42, line 19:
```suggestion
        ignore: './src/__tests__/**/*'
```

Can probably ignore all files in tests

Jolg42: If that makes sense I can suggest that change in the other places too

Jolg42 (approved):
Amazing! Can't wait to try it locally!

tsconfig.json
millsp: Yes, but if you don't it will fall back on dist types.

millsp:
closed in favor of https://github.com/prisma/prisma/pull/8497

PR #8482, opened 2021-07-28
fix(migrate): fix cases where migrate engine was not stopped after an error

### LINKED ISSUE #7795: Investigate and remove usage of `--forceExit`

Jolg42:
Closes https://github.com/prisma/prisma/issues/7795

### DIFF
--- a/src/packages/client/src/__tests__/integration/__helpers__/migrateDb.ts
+++ b/src/packages/client/src/__tests__/integration/__helpers__/migrateDb.ts
@@ -13,9 +13,11 @@ export async function migrateDb({
   await createDatabase(connectionString)
   const migrate = new Migrate(schemaPath)
 
-  await migrate.push({
-    force: true,
-  })
-
-  migrate.stop()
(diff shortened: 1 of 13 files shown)


janpio (approved):
Seems to do the job!

Adds a lot of localized `migrate.stop()` though - is there maybe a way to handle this on a more global level that automatically runs after the local functionality is done, no matter if with an error or success?

Jolg42:
I agree this is not optimal, we can revisit the implementation in the future.

PR #8497, opened 2021-07-29
chore(prisma): type discovery + fast builds + clean monorepo

### LINKED ISSUE #8443: Enable Type Discovery

millsp:
It's all in the title :point_up: 

closes https://github.com/prisma/prisma/issues/8443


millsp:
Looks like there is a flaky https://buildkite.com/prisma/prisma2-test/builds/8640#5d584656-8612-4d1e-b8f2-531d8cfde31d/229-3354

millsp:
The binary is not fully downloaded before we execute the tests `BINARY_SIZE 200704` (bytes)

millsp:
:tada: Solved thanks to @williamluke4 :tada:

millsp:
thanks @williamluke4 :tada:

Jolg42:
I merged my PR before that one and it looks like it created a conflict 🙈

Jolg42:
I just tried it and so far it works like a charm 💚

packages/client/src/__tests__/types/$transaction/index.test-d.ts
```diff
@@ -0,0 +1,31 @@
...
+  expectError(await prisma.$transaction([prisma.$disconnect()]))
+  expectError(
+    await prisma.$transaction([
+      new Promise((res) => res('You Shall Not Pass')),
+    ]),
+  )
+  expectError(await prisma.$transaction([5]))
+  expectError(await prisma.$transaction(['str']))
+  expectError(await prisma.$transaction([{}]))
+})()
```
Jolg42, line 31: Looks like an unrelated addition, but nice!

packages/debug/src/index.ts
```diff
@@ -24,8 +24,6 @@ Debug.enable = (namespace: string): void => {
 
 Debug.enabled = (namespace: string): boolean => DebugNode.enabled(namespace)
 
-export declare type Debugger = DebugLib.Debugger
```
Jolg42, line 27: Did we use this somewhere?

packages/sdk/src/utils/load.ts
```diff
@@ -1,11 +1,16 @@
...
+ * This is a wrapper around `require`
  * This is to avoid eval and hide require away from bundlers
  */
 export function load<T>(id: string): T {
   try {
     return require(id) as T
   } catch (e) {
+    const { size } = fs.statSync(id)
+
+    console.log('BINARY_SIZE', size)
```
Jolg42, line 13: Looks unrelated to types discovery, what is it for?

scripts/setup.ts
```diff
@@ -84,7 +84,7 @@ has to point to the dev version you want to promote, for example 2.1.0-dev.123`)
           if (
             ['@prisma/migrate', '@prisma/integration-tests'].includes(pkgName)
           ) {
-            run(pkgDir, 'pnpm rebuild')
+            await run(pkgDir, 'pnpm rebuild')
```
Jolg42, line 87: nice catch

Jolg42 (approved):
Love it 💚

packages/sdk/src/utils/load.ts
millsp: Oh yes, that was a debug for tests not passing. Good catch!

packages/client/src/__tests__/types/$transaction/index.test-d.ts
millsp: Thanks! That should be removed

.github/workflows/test.yml
```diff
@@ -18,16 +18,16 @@ on:
...
       - 'src/.buildkite/**'
-      - 'src/.dockerignore'
+      - '.dockerignore'
       - '*.bench.ts'
-      - 'src/scripts/ci/publish.ts'
-      - 'src/graphs/**'
+      - 'scripts/ci/publish.ts'
+      - 'graphs/**'
   pull_request:
     paths-ignore:
```
000gooo, line 26: Pull request

000gooo (commented):
_[]()_

PR #8501, opened 2021-07-30
fix(tar): upgrade to 6.1.2

janpio:


src/packages/sdk/package.json
```diff
@@ -76,7 +76,7 @@
     "string-width": "^4.2.0",
     "strip-ansi": "6.0.0",
     "strip-indent": "3.0.0",
-    "tar": "^6.0.1",
+    "tar": "^6.1.12",
```
madebysid, line 79: 6.1.2 is what you want I think: https://www.npmjs.com/package/tar

src/pnpm-lock.yaml
```diff
@@ -676,7 +676,7 @@ importers:
       string-width: ^4.2.0
       strip-ansi: 6.0.0
       strip-indent: 3.0.0
-      tar: ^6.0.1
+      tar: ^6.1.12
```
madebysid, line 679: I'm not entirely sure why, but we don't commit pnpm-lock updates, I think the release pipeline does that automatically. (Don't quote me on this)

src/packages/sdk/package.json
janpio:
```suggestion
    "tar": "^6.1.2",
```

janpio: lol good catch

sdnts (approved):

PR #8509, opened 2021-07-30
feat: No .gitignore file if already present

### LINKED ISSUE #8496: .gitignore is being overwrittern by `prisma init`
### Bug description

When you run `npx prisma init` everything executes without problems but the `.gitignore` file is being overwritten with the following content mentioned in Issue #3400

### How to reproduce

1. Create an empty `npm` project
2. Create and add some contents to `.gitignore`
3. Run `npx prisma init`
4. Open the `.gitignore` file and notice it was overwritten

### Expected behavior

The expected behavior may be simply not overwrite the `.gitignore` if it already exists, instead just output some message like: *Be sure to exclude node_modules and .env files inside .gitignore*.

### Prisma information

Nothing. I'm just starting prisma

### Environment & setup

- OS: Fedora
- Database: I don't think this is relevant
- Node.js version: v14.17.0

### Prisma Version

```
prisma               : 2.28.0
Current platform     : rhel-openssl-1.1.x
Query Engine         : query-engine 89facabd0366f63911d089156a7a70125bfbcd27 (at node_modules/@prisma/engines/query-engine-rhel-openssl-1.1.x)
Migration Engine     : migration-engine-cli 89facabd0366f63911d089156a7a70125bfbcd27 (at node_modules/@prisma/engines/migration-engine-rhel-openssl-1.1.x)
Introspection Engine : introspection-core 89facabd0366f63911d089156a7a70125bfbcd27 (at node_modules/@prisma/engines/introspection-engine-rhel-openssl-1.1.x)
Format Binary        : prisma-fmt 89facabd0366f63911d089156a7a70125bfbcd27 (at node_modules/@prisma/engines/prisma-fmt-rhel-openssl-1.1.x)
Default Engines Hash : 89facabd0366f63911d089156a7a70125bfbcd27
Studio               : 0.417.0
```

ShubhankarKG:
No tests changed. Only warning emitted if `.gitignore` already exists.

Fixes #8496


src/packages/cli/src/Init.ts
```diff
@@ -275,11 +275,16 @@ export class Init implements Command {
...
-      path.join(outputDir, ".gitignore"),
-      defaultGitIgnore()
-      );
+      if (!fs.existsSync(gitignorePath)) {
+        fs.writeFileSync(
+          gitignorePath,
+          defaultGitIgnore()
+        );
+      } else {
+        warning = `${chalk.yellow('warn')} You already have a .gitignore. Don't forget to exclude .env`
```
pantharshit00, line 286: Add a new line `\n` before this so that this doesn't cram with other warnings.

ShubhankarKG:
A lot of snapshot tests are failing. Any pointers on what I could do to fix them would be great.

pantharshit00:
I think we can remove the `.gitignore` from the init fixture: https://github.com/prisma/prisma/blob/410fb9a3d20770f30dac7c066f233234b03c35e7/src/packages/cli/src/__tests__/fixtures/init/.gitignore

cc @Jolg42 for confirmation.  

But do add one test with `.gitignore`.

src/packages/cli/src/Init.ts
Jolg42: Since warning is a string this assignment here will overwrite a potential previous warning.
We need to find a way to preserve the previous warning.
Maybe switching `let warning` to `let warnings: string[]` and then printing `warnings.join('\n')` will be a good option here.

Jolg42:
Yes indeed we can remove that one!

ShubhankarKG:
After the recent changes, how do I run tests? I tried `cd packages/cli/src/__tests__ && pnpm run test` but i get the following error:
`Error: Cannot find module 'MyFolderPath/packages/cli/build/index.js'`

millsp:
This should work
```
cd packages/cli/
pnpm -r dev
pnpm run test
```

Issue #8467, opened 2021-07-27
`config.datasources[0].provider` from GetConfig needs to be a string and not a string[]


Jolg42:
### Problem

```
const config = await getConfig({
	datamodel: schema,
})

// this `provider` value was a string before 2.22.0 but is now an array of string
config.datasources[0].provider
```

This is the related commit that introduced the regression:
https://github.com/prisma/prisma-engines/commit/b096a80dd515ef7c3a598867d307f729dcc0712b 

In this commit it looks like we removed the array from the datasource but added it to the serialization.
<img width="491" alt="Screen Shot 2021-07-27 at 16 40 35" src="https://user-images.githubusercontent.com/1328733/127189554-c08a04cc-7eab-4e51-8a51-8ddba074d8c2.png">
<img width="468" alt="Screen Shot 2021-07-27 at 16 40 11" src="https://user-images.githubusercontent.com/1328733/127189561-c90e1c0d-b66e-4b25-bdb4-09317d73b37f.png">

Rust Type found in https://github.com/prisma/prisma-engines/blob/master/libs/datamodel/core/src/configuration/datasource.rs#L12-L13
```
    /// The provider string
    pub provider: String,
```

the tests also show it’s currently an array of strings with only one string https://github.com/prisma/prisma-engines/blob/master/libs/datamodel/core/tests/config/datasources.rs#L16

### Suggestion

I think it should be a string and not an array returned here by the Rust code. 

I could monkey patch in the CLI but it seems wrong because we have a datasource[] array and each datasource can only have one provider.
It looks like some legacy from the multi provider array feature deprecated a while ago.

[Internal Slack Thread here](https://prisma-company.slack.com/archives/CC59J6VEY/p1626793129000400)

thebiglabasky:
We'll for now research places where this fix on `getConfig` would have an impact to ensure we don't break unexpectedly shortly before releasing.
Next step is to share the outcome of the research here before deciding how to fix this.
People to speak to: @madebysid (Studio), @millsp and/or @williamluke4

madebysid:
Studio doesn't use `getConfig` directly anymore, so we're good there. If the `getGenerators` function behaves the same as before, Studio will not need any changes.

Jolg42:
Found these lines using it, we can easily fix them after the engine fix:
[...]

PR #8529, opened 2021-08-02
fix(cli): test getConfig provider array fix

Jolg42:
Related https://github.com/prisma/prisma/issues/8467


src/packages/cli/src/bin.ts
```diff
@@ -189,14 +189,26 @@ async function main(): Promise<number> {
     let schemaProvider: string | undefined
     let schemaPreviewFeatures: string[] | undefined
     let schemaGeneratorsProviders: string[] | undefined
+
     try {
       const schema = await getSchema(args['--schema'])
       const config = await getConfig({
         datamodel: schema,
```
Jolg42, line 196:
```suggestion
        datamodel: schema,
        ignoreEnvVarErrors: true,
```

src/packages/cli/src/bin.ts
```diff
@@ -189,14 +189,27 @@ async function main(): Promise<number> {
...
+        ignoreEnvVarErrors: true,
```
Jolg42, line 197: Looks like we forgot to so that a long time ago, without it was throwing when env var was not set for example.

Jolg42: Done in https://github.com/prisma/prisma/pull/8529

Jolg42:
Fix was made in engines, closing
https://github.com/prisma/prisma/issues/8467#issuecomment-892501316

PR #8594, opened 2021-08-05
deep-dive-qe-in-cli

janpio:


janpio:
There were quite a few changes in here that might actually be useful:
https://github.com/prisma/prisma/pull/8594/files#diff-d936061183173fab398741ca06cc917778cd22eda245fb35ee8237831a2f679dL91

PR #8617, opened 2021-08-06
chore: restrict search to prisma-client-js for previewFeatures telemetry

Jolg42:
Does that makes sense to you too?

Maybe it doesn't, for example there is a new community python client https://github.com/RobertCraigie/prisma-client-py


millsp (approved):

packages/cli/src/bin.ts
```diff
@@ -198,15 +198,22 @@ async function main(): Promise<number> {
...
         schemaProvider = config.datasources[0].provider
       }
-      const generator = config.generators.find(
-        (gen) => gen.previewFeatures.length > 0,
-      )
+
+      // restrict the search to prisma-client-js and prisma-client-go previewFeatures
+      const generator = config.generators.find((generator) => {
+        return (
+          ['prisma-client-js', 'prisma-client-go'].includes(generator.name) &&
```
janpio, line 209:
```suggestion
          ['prisma-client-js'].includes(generator.name) &&
```

packages/generator-helper/src/types.ts
```diff
@@ -75,13 +76,16 @@ export type BinaryPaths = {
...
   generator: GeneratorConfig
+  // TODO: what is otherGenerators for?
   otherGenerators: GeneratorConfig[]
   schemaPath: string
   dmmf: DMMF.Document
   datasources: DataSource[]
+  // TODO deprecate datamodel & rename to schema?
   datamodel: string
-  binaryPaths?: BinaryPaths
+  // TODO is it really always version hash? Feature is unclear.
```
janpio, line 86: probably engine version, so git commit hash?

Jolg42: It seems we use it for something and third party users use it for something else

packages/cli/src/bin.ts
Jolg42: The unofficial Python client is named `prisma-client-py` and would be filtered as well as go with this PR/suggestion 

Maybe we just don't restrict the search here?

janpio: I think we only want the preview feature of the official JS Client here. Everything else is out of our control and does not give us a clear picture.

Jolg42: Then we want to keep `prisma-client-go` I would say

janpio: See link above: https://github.com/prisma/prisma/pull/8617#issuecomment-907216484 Go Client does not support any preview features, so if someone is providing them they have no effect.

Jolg42: Good point 👍🏼

packages/cli/src/bin.ts
```diff
@@ -198,15 +198,22 @@ async function main(): Promise<number> {
...
-        (gen) => gen.previewFeatures.length > 0,
-      )
+
+      // restrict the search to prisma-client-js previewFeatures
+      const generator = config.generators.find((generator) => {
+        return (
+          generator.name === 'prisma-client-js' &&
+          generator.previewFeatures.length > 0
+        )
+      })
```
Jolg42, line 212:
```suggestion
      const generator = config.generators.find((generator) => generator.name === 'prisma-client-js' && generator.previewFeatures.length > 0)
```

packages/cli/src/bin.ts
```diff
@@ -198,15 +198,18 @@ async function main(): Promise<number> {
...
       if (config.datasources.length > 0) {
         schemaProvider = config.datasources[0].provider
       }
-      const generator = config.generators.find(
-        (gen) => gen.previewFeatures.length > 0,
-      )
+
+      // restrict the search to prisma-client-js previewFeatures
+	  // this was not scoped to `prisma-client-js` before Prisma 3.0
+      const generator = config.generators.find((generator) => generator.name === 'prisma-client-js' && generator.previewFeatures.length > 0)
```
janpio, line 208: Do we have tests for this?

Jolg42: Nope 🎐

janpio: Scary. We should probably manually test CLI telemetry with a `dev` version then?

janpio: https://prisma-company.slack.com/archives/CC59J6VEY/p1630512455033900

PR #8620, opened 2021-08-06
test(client): namedConstraints @@unique @@id with and without name

Jolg42:


packages/client/src/__tests__/types/namedConstraints/test.ts
```diff
@@ -23,6 +25,25 @@ async function main() {
         },
       },
     })
+
+  // @@unique
+  const result3 = await prisma.atAtUnique.findUnique({
```
janpio, line 30: Should this be `AtAtUniqueNamed` maybe? THe `name` is on that model.

Jolg42: Yes indeed

do4gr: I committed the changes to have the tests rerun

do4gr (approved):
Apart from the issue @janpio  detected this looks good to me.

packages/client/src/__tests__/types/namedConstraints/test.ts
```diff
@@ -23,6 +25,25 @@ async function main() {
         },
       },
     })
+
+  // @@unique
+  const result3 = await prisma.AtAtUniqueNamed.findUnique({
```
janpio, line 30:
```suggestion
  const result3 = await prisma.atAtUniqueNamed.findUnique({
```

Issue #8449, opened 2021-07-23
Cannot access engine inside vercel/pkg


wallkie:
Dear community, 

I'm having a problem to use prisma inside a [vercel/pkg](https://github.com/vercel/pkg) in which i'm getting an `ENOENT` error like:
```
spawn /snapshot/enoent-problem/node_modules/.prisma/client/query-engine-debian-openssl-1.1.x ENOENT 
```
To narrow down the issue i've created a sample application [here](https://github.com/wallkie/enoent-problem)

## To reproduce the issue:
I'm running the packaged app as :
```sh
npm i
npx prisma generate
npx pkg package.json --debug --output bin/exe > log-pkg.log
env -i DEBUG_PKG=1 bin/exe
```
and getting the following error 
```
spawn called cwd: <exec_path>/enoent-problem and dirname: /snapshot/enoent-problem
[Arguments] {
  '0': '/snapshot/enoent-problem/node_modules/.prisma/client/query-engine-debian-openssl-1.1.x',
  '1': [ 'cli', 'get-config' ],
  '2': {
    maxBuffer: 100000000,
    buffer: true,
    stripFinalNewline: true,
    extendEnv: true,
    preferLocal: false,
    localDir: '/snapshot/enoent-problem/node_modules/.prisma/client',
    execPath: '<exec_path>/enoent-problem/bin/exe-node16-linuxstatic',
    encoding: 'utf8',
    reject: true,
    cleanup: true,
    all: false,
    windowsHide: true,
    env: {
      DATABASE_URL: 'postgresql://postgres:postgres@localhost:5432/<table_name>?schema=public',
      PRISMA_DML_PATH: '/snapshot/enoent-problem/node_modules/.prisma/client/schema.prisma',
      RUST_BACKTRACE: '1',
      RUST_LOG: 'info',
      LOG_QUERIES: 'true',
      OVERWRITE_DATASOURCES: '[]'
    },
    cwd: '/snapshot/enoent-problem/node_modules/.prisma/client',
    stdio: [ undefined, undefined, undefined ]
  }
}
spawn called cwd: <exec_path>/enoent-problem and dirname: /snapshot/enoent-problem
[Arguments] {
  '0': '/snapshot/enoent-problem/node_modules/.prisma/client/query-engine-debian-openssl-1.1.x',
  '1': [ '--enable-raw-queries', '--port', '33491' ],
  '2': {
    env: {
[...]

PR #8634, opened 2021-08-06
fix: use root cwd if starting query engine from pkg

hughobrien:
https://github.com/vercel/pkg/issues/342

https://github.com/prisma/prisma/issues/8449


lamualfa:
you don't need this. just follow my instruction here:

https://github.com/prisma/prisma/issues/8449#issuecomment-1147743988

garrensmith:
thanks @lamualfa for the help. I've closing this. We have updated our docs to include this information.

robert-king:
@lamualfa  what about resolving the sqllite dev.db relative to root cwd? for me pkg doesn't work with relative url e.g. file:./dev.db

PR #8659, opened 2021-08-09
test(client): @@unique add namedConstraints test without flag

Jolg42:


packages/client/src/__tests__/integration/happy/namedConstraints-without-preview-flag/schema.prisma
```diff
@@ -0,0 +1,23 @@
...
+
+// named @@unique was possible before namedConstraints flag
+model AtAtUnique {
+  key1 String
+  key2 Int
+
+  @@unique([key1, key2])
+}
+
+model AtAtUniqueNamed {
```
janpio, line 18:
```suggestion
model AtAtUnique {
  key1 String
  key2 Int

  @@unique([key1, key2])
}

// named @@unique was possible before namedConstraints flag
model AtAtUniqueNamed {
```

PR #8698, opened 2021-08-11
fix(generation): stop stringifying strings

FredericEspiau:
https://github.com/prisma/prisma/blob/main/packages/client/src/generation/TSClient/TSClient.ts#L92 will generate code

# Current behavior

```ts
const root = "/path/to/prisma/";

findSync(root, [
  '"prisma/path/to/schema"',
  '"path/to/schema"',
```

will always return `__dirname` even if the path `/path/to/prisma/path/to/schema` exists for example

# New behavior

```ts
const root = "/path/to/prisma/";

findSync(root, [
  'prisma/path/to/schema',
  'path/to/schema",
```

will return `/path/to/prisma/path/to/schema`

# Explanation

If a string is given to `JSON.stringify`, it will be wrapped in double quotes

Example: `JSON.stringify("path/to/b") === '"path/to/b"'`

As the path was used in `String.prototype.includes` right there: https://github.com/prisma/prisma/blob/ba74c81fdbc9e6405946fdc6f9d42d103d008dc2/packages/client/src/runtime/utils/find.ts#L41, this would lead to not detecting that the path was a child of the root

Example: `"/usr/path/to/b".includes(JSON.stringify("path/to/b"))" === false`

What you want is `"/usr/path/to/b".includes("path/to/b")" === true`


Amirault (approved):

millsp:
Good catch, thank you! We use `JSON.stringify` because it escapes `\` for windows paths.

FredericEspiau:
@millsp ah thank you, I didn't know that, good catch

FredericEspiau:
@millsp as you created this https://github.com/prisma/prisma/pull/8702 should I update my PR or you are taking care of this ?

FredericEspiau:
hey @millsp sorry to bother you again, do you want me to update my PR to use `JSON.stringify` or I shouldn't bother because you will work on https://github.com/prisma/prisma/pull/8702 instead ?

millsp:
No worries @FredericEspiau, I will take your PR from there in a few days. Thanks!

FredericEspiau:
alright thanks !

FredericEspiau:
Oups sorry, I thought you didn't need this anymore 😄

FredericEspiau:
This has been done by https://github.com/prisma/prisma/pull/9295

PR #8702, opened 2021-08-11
fix(generation): prevent finding already found paths

millsp:
`seen[realRoot] = true` is a map to mark that a path has already been visited (ie. for circular links). It needs to be enabled to prevent potential infinite loops.

~~Next steps: Before merge, add E2E tests for Next.js Serverless for higher confidence.~~


Jolg42:
Looks like it was already done in another PR but what is the other change here? @millsp 
The uncomment of
```
seen[realRoot] = true
```

Jolg42 (changes requested):
To make this part explicit to not merge before
>Next steps: Before merge, add E2E tests for Next.js Serverless for higher confidence.

Jolg42 (approved):

PR #8720, opened 2021-08-12
fix(client): 2.29.1 fixes

millsp:
Fixes https://github.com/prisma/engines-wrapper/issues/128
Fixes https://github.com/prisma/prisma/issues/8717
Fixes https://github.com/prisma/prisma/issues/8718

PR #8735, opened 2021-08-13
fix(client): 2.29.1 fixes (#8720)

Jolg42:
Backmerge of fixes done in 2.29.x


Jolg42:
Replaced by https://github.com/prisma/prisma/pull/8736 because we needed to remove pnpm override and didn't want to push on 2.29.x branch (reserved for patch commits)

PR #8782, opened 2021-08-17
ci: switch engine type for both cli and client for relavant tests

williamluke4:


Jolg42 (approved):
Thanks @williamluke4 !

PR #8823, opened 2021-08-19
fix(deps): update undici

Jolg42:
Fails with wasm error on Node-API Client tests.

Example:
```
FAIL src/__tests__/integration/happy/browser/test.ts
  ● Test suite failed to run

    RangeError: WebAssembly.Instance(): Out of memory: wasm memory

      at Object.<anonymous> (../../node_modules/.pnpm/undici@4.4.5/node_modules/undici/lib/client.js:400:16)
      at Object.<anonymous> (../../node_modules/.pnpm/undici@4.4.5/node_modules/undici/index.js:3:16)

```

Because we import both LibraryEngine and BinaryEngine files and undici tries to load the wasm even if not instantiated / needed


Jolg42:
Integration available with `2.30.0-integration-undici-4.1`

janpio:
Opened an issue at `undici` asking for input: https://github.com/nodejs/undici/issues/988

janpio:
replaced by https://github.com/prisma/prisma/pull/8842

PR #8841, opened 2021-08-20
feat: Data proxy engine

ghost:
Experimental engine for Data Proxy:

- throws a lot, logs very little
- "beforeExit" event not supported
- transactions not supported
- batched queries result in multiple calls to Data Proxy
- extracting datasource URL is fragile (regex + prismafile)

Main features:
- talk to a DB via Data Proxy
- lazily upload the schema
- retry with backoff if the Proxy misbehaves


cebuladev:
@millsp @Jolg42 @williamluke4 I'd appreciate you folks taking a look at the current state of this PR so we verify I'm not doing something terribly stupid

packages/client/fixtures/dataproxy/prisma/schema.prisma
```diff
@@ -0,0 +1,25 @@
+datasource db {
+  provider = "postgres"
+  url      = env("DATABASE_URL")
+}
+
+generator client {
+  provider        = "prisma-client-js"
+  engineType = "dataproxy"
```
janpio, line 8: nit:
```suggestion
  provider   = "prisma-client-js"
  engineType = "dataproxy"
```

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
+import {
```
millsp, line 1: Could we move this file to `@prisma/engine-core` into a new `data-proxy` folder?

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+
+function backOff(n: number): Promise<number> {
+  const baseDelay = Math.pow(2, n) * BACKOFF_INTERVAL
+  const jitter = Math.ceil(Math.random() * baseDelay) - Math.ceil(baseDelay / 2)
+  const total = baseDelay + jitter
+
+  return new Promise((done) => setTimeout(() => done(total), total))
+}
+
+// TODO: move to @prisma/engine-core
```
millsp, line 25: Just saw this now :+1:

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+  private async uploadSchema() {
+    const res = await fetch(this.url('schema'), {
+      method: 'PUT',
+      headers: this.headers,
+      body: this.schemaBase64,
+    })
+
+    if (!res.ok) {
+      this.logEmitter.emit('warn', 'Could not upload the schema')
+      throw new Error('Could not upload the schema')
```
millsp, line 97: We need to make these errors Prisma errors, you can find them in `engine-core`.

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+      datasources: [
+        {
+          activeProvider: this.config.activeProvider,
+        },
+      ],
+    } as GetConfigResult)
+  }
+
+  private async uploadSchema() {
+    const res = await fetch(this.url('schema'), {
```
millsp, line 89: What issue did you encounter with `undici`?

millsp: Probably Node.js imports, I guess

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    this.schemaText = fs.readFileSync(config.datamodelPath, 'utf8')
+    this.schemaBase64 = Buffer.from(this.schemaText).toString('base64')
+    this.schemaHash = createHash('sha256')
+      .update(this.schemaBase64)
+      .digest('hex')
+
+    const [host, apiKey] = extractHostAndApiKey(this.schemaText)
+    // TODO: how to get the current version the right way?
+    const clientVersion = '2.29.1'
```
Jolg42, line 47: Import it like 
```
import { clientVersion } from './utils/clientVersion'
```

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+        throw err
+      } else {
+        const delay = await backOff(n)
+        this.logEmitter.emit('warn', `Retrying after ${delay}`)
+        return this.request<T>(q, {}, n + 1)
+      }
+    }
+  }
+
+  requestBatch<T>(qs: string[], _hs: any, transact = false) {
```
Jolg42, line 139: Would it be possible to rename the arguments in the code? 
I guess
- `qs` is `querystring` but it's a `string[]` and I also see `q` as `string` so renaming could help maybe? making it more explicit what this is.
- `_hs` is `headers`
- `transact` is `transaction`/`runInTransaction`

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+    ) {
+      const arg = args[0]
+
+      if (arg?.type === 'string_value' && typeof arg.value === 'string') {
+        const envVarName = arg.value.trim()
+
+        if (envVarName.length > 0) {
+          // Datasource URL in environment
+          url = process.env[envVarName] ?? ''
+        }
```
Jolg42, line 208: I think here reading the env var from the `.env` file might be needed as well.

It's actually passed to `_engineConfig` see
https://github.com/prisma/prisma/blob/main/packages/client/src/runtime/getPrismaClient.ts#L396:L396

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  let parsed: URL
+  try {
+    parsed = new URL(url)
+  } catch (err) {
+    throw new Error('Could not parse URL of the datasource')
+  }
+
+  const { protocol, host, searchParams } = parsed
```
Jolg42, line 227: This could be implemented in `uriToCredentials` with tests 
https://github.com/prisma/prisma/blob/main/packages/sdk/src/convertCredentials.ts#L73

Jolg42:
We should add a basic test I think?

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  requestBatch<T>(qs: string[], _hs: any, transact = false) {
+    // TODO: transactions not supported (see below)
+    // TODO: introduce batch endpoint in Data Proxy;
+    // current implementation will result in `qs.length` HTTP calls
+    return transact
+      ? this.transaction()
+      : Promise.all(qs.map((q) => this.request<T>(q)))
```
Jolg42, line 145: Note about Promise.all 
>will resolve when all of the input's promises have resolved. It rejects immediately upon any of the input promises rejecting or non-promises throwing an error, and will reject with this first rejection message / error. 

Not sure how that would be different from the engine `requestBatch` behavior.

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  // TODO: figure out how to support transactions
+  transaction(): Promise<any> {
```
Jolg42, line 149:
```suggestion
  transaction(): void {
```

Jolg42:
Looks like we can close in favor of https://github.com/prisma/prisma/pull/9422 ?

millsp:
Once https://github.com/prisma/prisma/pull/9422 also works on Node.js, I think that we can take it from there. Right now, it's just a build for browser-like environments, but I'm planning on making it work seamlessly on both platforms. That requires some tweaks in the injected code.

Maybe we can leave it open for now if @cebuladev needs to continue experimenting?

cebuladev:
@millsp I already had a chat with @Jolg42 and we agreed to keep it open

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  const datasourceText = strippedToDatasource[0]
+  if (typeof datasourceText !== 'string' || datasourceText.length < 1) {
+    throw new Error('Could not find a valid datasource block in the schema')
+  }
+
+  // Use prismafile to determine the URL
+  let url = ''
+  try {
+    const schemaAst = prismafile.parse(datasourceText)
```
matthewmueller, line 213: I really didn't expect us to start using this officially. Just a note that it's not 100% feature compatible with our current schema.prisma. I think it's missing @@index and a few other things 😅

Maybe instead of using this in-full, we could generate a parser that just specializes in extracting datasources.

E.g. https://github.com/prisma/prismafile/blob/master/src/parser/index.pegjs#L51

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,282 @@
...
+    const [datasource] = schemaAst.blocks.filter(
+      (b) => b.type === 'datasource',
+    ) as any[]
+
+    // Datasource type is not exported from prismafile, so from now on
+    // we're in any-land. Hence the dirtiness and the big try/catch around.
```
matthewmueller, line 219: This would be a very easy fix.

packages/client/src/runtime/DataProxyEngine.ts
```diff
@@ -0,0 +1,282 @@
...
+    throw new Error('Transactions are currently not supported in Data Proxy')
+  }
+}
+
+// Regex + prismafile feels fragile...
+// TODO: build a minimal single-purpose (URL extraction)
+// and well-tested TS schema parser
+function extractHostAndApiKey(schemaText: string) {
+  // Extract the datasource block so we feed prismafile with minimal input
+  const strippedToDatasource = /datasource[ \t]+[^\s]+[ \t]+\{[^}]+}/.exec(
```
matthewmueller, line 198: This feels slightly unnecessary when we parse it a few steps later. I guess it's an optimization, but probably not a very good one since it scans the whole string again anyway.

packages/client/src/runtime/DataProxyEngine.ts
millsp: Could this parsing be tackled by the engines any time soon?

janpio: Problem with doing this in Engines is that it still requires an Engine being present. With the binary and library variants that means downloading a platform dependent file (including it in the generated Client). With WASM this could be a Node package, but we do not have that yet and are making first experiences with `prisma-fmt` only. (This could of course be a good reason to prioritize this work)

millsp: Hmm, I was thinking about using the engines during the generation step to help embedding the URL or the name of the ENV var into the generated code.

cebuladev: I actually think that staying binary-free here makes a lot of sense. I investigated writing a small focused parser for getting `datasource.url` value (or env var name) but `prismafile` does its job just fine. Note that we don't need full parsing, we're just interested in `datasource.url` (hence stripping everything but the datasource block with a regex).

millsp: Via https://github.com/prisma/prisma/pull/9422, the schema is inlined into the generated client during the generation step. That is so that it can be embedded into browser environments (without fs). That also means that it has been pre-validated by the engines. For that reason it is now extracted via a regexp. (cc @steebchen)

steebchen: Please see this thread: https://prisma-company.slack.com/archives/C02FZENHD4N/p1633515267034400

cebuladev: I'm not sure if I follow - how would you get the DMMF w/o the binary? The one that I was getting in the generated clients wouldn't have the URL nor the env var now. What we did here was: trim string to relevant part with a relatively simple readable regex and then parse the datasource (applying rule: parsing > regular expression).

I'm still pretty uncomfortable with the clients, I'd appreciate more detailed explanation for my own understanding.

steebchen (changes requested):
Just wanted to signal this explicitly that everything related to schema parsing should be removed and the URL (or, if provided at run time, the name of the env var) can be accessed at generation time via the DMMF.

cebuladev:
This PR is obsolete as of 3.3.0

PR #8842, opened 2021-08-20
fix(deps): update undici to 4.x

### LINKED ISSUE #6564: Bundling fails on 2.21.0
## Bug description

After upgrading to `2.21.0` and running ESBuild to bundle, a new error occurs:
```
> ../../common/temp/node_modules/.pnpm/@prisma+client@2.21.0_prisma@2.21.0/node_modules/@prisma/client/runtime/index.js:24152:23: 
error: Could not resolve "_http_common" (mark it as external to exclude it from the bundle)
24152 │   var common = require("_http_common");
```
and

```
"var runtimeRequire = 
  typeof __webpack_require__ === 'function' ? 
     __non_webpack_require__ : require // eslint-disable-line"
```

## How to reproduce

 
1. Install Prisma and [ESBuild](https://esbuild.github.io/getting-started/#bundling-for-node) using [PNPM](https://pnpm.io/).
2. Run `prisma generate`
3. Run `esbuild app.js --bundle --platform=node --target=node14`
 
## Expected behavior

Prior to `2.21.0` bundling was successful.

## Prisma information

```
// Datasources
datasource db {
  provider = "postgresql"
}

// Generators
generator client {
  binaryTargets = ["debian-openssl-1.1.x", "linux-musl", "windows", "darwin"]
}
```

## ESBuild config

```js
module.exports = {
  bundle: true,
  minify: false,
  format: "cjs",
  keepNames: true,
  sourcemap: true,
  platform: "node",
  logLevel: "error",
  target: ["node14"],
  external: ["fsevents"],
  outfile: "dist/run.js",
  entryPoints: ["src/index.ts"],
  metafile: false,
  loader: { ".prisma": "file" },
}

```

## Environment & setup

- OS: Windows
- Database:  Postgres
- Node.js version: v14.16.0
- Prisma version: v2.21.0

### LINKED ISSUE #6899: I get "Can't resolve '_http_common'" when I use NestJS with monorepo (webpack)
## Bug description

When I try to composite Prisma with [NestJS monorepo](https://docs.nestjs.com/cli/monorepo#monorepo-mode) project I get the error. I can`t use default prisma client output path because my projects use prisma client and each project overrides before generated client files. I try to use custom output path but I get the error. I found some information about the error -> https://github.com/restify/node-restify/issues/794.

## How to reproduce

1. Create new app as [monorepo project](https://docs.nestjs.com/cli/monorepo#monorepo-mode)
2. Generate client in custom output path. For example "output = ../src/prisma/client"
3. Execute command "build"
4. See error "Can't resolve '_http_common'"

## Expected behavior

Build happy project

## Environment & setup

- OS: Mac OS
- Database: PostgreSQL
- Node.js version: 12.22.1
- Prisma version: 2.21.2

```
datasource db {
  provider	=	"postgresql"
  url	=	env("MAIN_DATABASE_URL")
}

generator client {
  provider = "prisma-client-js"
  output = "../src/prisma/client"
}
```

### LINKED ISSUE #6925: Broken on Next.js 10.2.0 which uses Webpack 5 when running a custom server to support HTTPS locally
## Bug description

When I upgrade Next.js to 10.2.0 which uses Webpack 5 and then run `node server.js` to run my app locally using HTTPS, I get this error:

```
error - ./prisma/client/runtime/index.js:24474:15
Module not found: Can't resolve '_http_common'
  24472 | var require_http_parser = __commonJS2((exports2, module2) => {
  24473 |   "use strict";
> 24474 |   var common = require("_http_common");
        |               ^
  24475 |   if (common.HTTPParser) {
  24476 |     module2.exports = common.HTTPParser;
  24477 |   } else {
```

## How to reproduce

Typically you run Next.js apps locally by running `npm run dev` and that's it, but if you want to use HTTPS locally then you have to do something a bit different which involves creating a file usually called "server.js" which attaches the SSL certificate and then runs the server. I have no idea if this particularly issue is related to that or if it would happen without HTTPS, as I only use this method now. I'm hoping the error message alone might make the issue apparent to anyone who understands what it's saying, and why Webpack 5 would cause this.

## Expected behavior

Everything works just as it would on Next.js <10.2.0 / Webpack 4.

## Environment & setup

- OS: macOS Big Sur
- Database: MySQL / Amazon Aurora
- Node.js version: 14.16.0
- Prisma version:

```
prisma               : 2.22.0
@prisma/client       : 2.22.0
Current platform     : darwin
Query Engine         : query-engine 60cc71d884972ab4e897f0277c4b84383dddaf6c (at node_modules/@prisma/engines/query-engine-darwin)
Migration Engine     : migration-engine-cli 60cc71d884972ab4e897f0277c4b84383dddaf6c (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine : introspection-core 60cc71d884972ab4e897f0277c4b84383dddaf6c (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary        : prisma-fmt 60cc71d884972ab4e897f0277c4b84383dddaf6c (at node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash : 60cc71d884972ab4e897f0277c4b84383dddaf6c
Studio               : 0.379.0

```

### LINKED ISSUE #7539: Upgrade to undici@4

janpio:
Fixes https://github.com/prisma/prisma/issues/6564
Fixes https://github.com/prisma/prisma/issues/6899
Fixes https://github.com/prisma/prisma/issues/6925


janpio:
Available for manual testing as `2.30.0-integration-undici-4.2`

packages/engine-core/package.json
```diff
@@ -55,7 +55,7 @@
     "new-github-issue-url": "0.2.1",
     "p-retry": "4.6.1",
     "terminal-link": "2.1.1",
-    "undici": "3.3.6"
+    "undici": "^4.4.6"
```
millsp, line 58:
```suggestion
    "undici": "4.4.6"
```

packages/cli/helpers/build.js
```diff
@@ -62,6 +62,13 @@ async function main() {
       path.join(require.resolve('open/package.json'), '../xdg-open'),
       './build/xdg-open',
     ),
+    copy({
```
millsp, line 65: Could we remove `fs-extra` ?

packages/client/helpers/build.js
```diff
@@ -1,11 +1,13 @@
...
 const esbuild = require('esbuild')
+const fs = require('fs-extra')
+const path = require('path')
 
 const ESBUILD_DEFAULT = {
   bundle: true,
   platform: 'node',
   target: 'es2018',
-  external: ['_http_common'],
+  // external: ['_http_common'],
```
millsp, line 10:
```suggestion
```

Jolg42:
Note: we founds that undici was using `Blob` https://github.com/nodejs/undici/search?q=require%28%27buffer%27%29&type=code which was only added with Node `v15.7.0` https://nodejs.org/docs/latest-v16.x/api/buffer.html#buffer_class_blob (Node PR https://github.com/nodejs/node/pull/36811)

We need to make sure this doesn't impact our supported versions of Node (12, 14)

packages/engine-core/src/binary/Connection.ts
```diff
@@ -88,7 +91,7 @@ export class Connection {
         ...headers,
       },
       body,
-      bodyTimeout: 0,
+      // bodyTimeout: 0,
```
janpio, line 94: ??

pnpm-lock.yaml
```diff
@@ -360,7 +362,7 @@ importers:
       '@prisma/generator-helper': workspace:*
       '@prisma/get-platform': 2.30.0-22.6ee261c643b2c461a20371f48ffa5090ed578aaf
       '@types/jest': 27.0.1
-      '@types/node': 12.20.19
+      '@types/node': 16.6.2
```
janpio, line 365: unrelated? was lockfile just out of sync?

packages/cli/helpers/build.js
```diff
@@ -62,6 +62,13 @@ async function main() {
...
+      from: path.resolve(__dirname, '../prisma-client/runtime/llhttp'),
```
janpio, line 66: this path depends on some other copy mechanism to already have been run

packages/client/helpers/build.js
```diff
@@ -28,6 +30,22 @@ async function build() {
...
+      path.resolve(__dirname, '../generator-build/llhttp'),
+    ),
+    // copy wasm files, etc necessary for undici
+    fs.copy(
+      path.resolve(
+        __dirname,
+        '../node_modules/@prisma/engine-core/node_modules/undici/lib/llhttp',
+      ),
+      path.resolve(__dirname, '../runtime/llhttp'),
+    ),
```
janpio, line 48: why twice? why both these paths?

pnpm-lock.yaml
Jolg42: Related to undici using `Blob` see previous comment https://github.com/prisma/prisma/pull/8842#issuecomment-903525303

packages/engine-core/src/binary/Connection.ts
Jolg42: I think this option does not exist anymore, to check if it was moved somewhere else and fix it

pnpm-lock.yaml
janpio: Ah, I overlooked the change in `packages/engine-core/package.json`.

packages/engine-core/package.json
```diff
@@ -16,7 +16,7 @@
   "bugs": "https://github.com/prisma/prisma/issues",
   "devDependencies": {
     "@types/jest": "27.0.1",
-    "@types/node": "12.20.19",
+    "@types/node": "16.6.2",
```
janpio, line 19: See https://github.com/prisma/prisma/pull/8842#issuecomment-903525303

packages/client/helpers/build.js
millsp: Somehow the `BinaryEngine` gets bundled into `generator-build`, but does not seem to be used?
So is the copying if `lhttp` really a pre-requisite? Would it throw an error if not found?

packages/engine-core/src/binary/Connection.ts
millsp: It looks like this should still work
https://github.com/nodejs/undici/blob/ffc6341559ae78314a47c81e640c9393f2f86eb9/types/client.d.ts#L34
https://github.com/nodejs/undici/blob/ffc6341559ae78314a47c81e640c9393f2f86eb9/types/pool.d.ts#L16

hugbubby:
https://media.istockphoto.com/photos/senior-man-sitting-on-bench-picture-id499841195

packages/client/helpers/build.js
janpio:
```suggestion
    // copy wasm files, etc necessary for undici
    fs.copy(
      path.resolve(
        __dirname,
        '../node_modules/@prisma/engine-core/node_modules/undici/lib/llhttp',
      ),
      path.resolve(__dirname, '../runtime/llhttp'),
    ),
```

janpio: Let's see: https://github.com/prisma/prisma/pull/8889

janpio: Seems the `generator-build` one was indeed not necessary (at least that is what the PR said), so merged removal of this into here. Should trigger a new integration release.

matthewmueller:
Should close: https://github.com/prisma/prisma/issues/7539

But worth double-checking.

packages/engine-core/src/binary/Connection.ts
millsp: This can now be set at the Pool instantiation.

packages/cli/helpers/build.js
millsp: Done.

millsp (approved):

millsp:
Do not merge yet

janpio:
No worries, with failing tests we won't :trollface: But also put the PR in "draft" state to prevent accidental merge.

millsp:
Open for re-review. A summary of the follow-up changes:
* Merged with `main` to take advantage of the build refactor
* Inlining/bundling the Wasm contents of `undici`'s `lhttp.wasm`

millsp:
Unrelated failure: https://github.com/prisma/prisma/runs/4607037295?check_suite_focus=true

helpers/compile/plugins/replaceWithPlugin.ts
```diff
@@ -1,27 +1,39 @@
...
+    } else {
+      contents = await replacement(regex, contents)
+    }
   }
 
   return contents
 }
 
+// native nodejs imports so that we can filter out
+const nativeDependencies = new Set(Object.keys((process as any).binding('natives')))
```
janpio, line 19: How does this work?

helpers/compile/plugins/replaceWithPlugin.ts
```diff
@@ -1,27 +1,39 @@
...
  * @returns
  */
-export const replaceWithPlugin = (replacements: [RegExp, string][]): esbuild.Plugin => {
+export const replaceWithPlugin = (replacements: Replacement[]): esbuild.Plugin => {
   return {
     name: 'replaceWithPlugin',
     setup(build) {
       build.onLoad({ filter: /.*/ }, async (args) => {
+        if (nativeDependencies.has(args.path)) return {}
+        if (!/.*?(.js|.mjs)$/.exec(args.path)) return {}
```
janpio, line 32: This skips "replacing" I assume? Just looking at this file, `replaceWithPlugin.ts`, is not really enough context to understand.

packages/cli/helpers/build.ts
```diff
@@ -1,7 +1,7 @@
 import type { BuildOptions } from '../../../helpers/compile/build'
 import { run } from '../../../helpers/compile/build'
 import { build } from '../../../helpers/compile/build'
-import { copySync } from 'fs-extra'
+import { copy } from 'fs-extra'
```
janpio, line 4: Is there a deeper reason for that change or just style?

helpers/compile/plugins/replaceWithPlugin.ts
millsp: Yes, it does. If return `{}`, `esbuild` defaults the original content.

packages/cli/helpers/build.ts
```diff
@@ -72,7 +72,7 @@ const cliLifecyclePlugin: esbuild.Plugin = {
 const cliBuildConfig: BuildOptions = {
   entryPoints: ['src/bin.ts'],
   outfile: 'build/index',
-  external: ['@prisma/engines', '_http_common'],
+  external: ['@prisma/engines'],
```
janpio, line 75: 🥳

helpers/compile/plugins/replaceWithPlugin.ts
millsp: I'll add a comment.

packages/cli/src/__tests__/__helpers__/snapshotSerializer.ts
```diff
@@ -18,16 +18,11 @@ function removePlatforms(str) {
 }
```
janpio, line 18: only formatting changes

packages/cli/helpers/build.ts
millsp: `esbuild` plugins are `async`, so it's best not to block the main thread.

packages/cli/src/__tests__/commands/Init.test.ts
```diff
@@ -10,10 +10,7 @@ test('is schema and env written on disk replace', async () => {
   const result = await ctx.cli('init')
```
janpio, line 10: only formatting changes

packages/client/src/__tests__/types/blog/test.ts
```diff
@@ -26,36 +26,21 @@ async function main() {
   })
```
janpio, line 26: only formatting

packages/client/src/__tests__/types/deprecation/test.ts
```diff
@@ -24,36 +24,21 @@ async function main() {
```
janpio, line 24: only formatting

packages/client/package.json
```diff
@@ -94,6 +94,7 @@
     "eslint-plugin-prettier": "4.0.0",
     "execa": "5.1.1",
     "flat-map-polyfill": "0.3.8",
+    "fs-extra": "10.0.0",
```
janpio, line 97: Is this used anywhere? Can not find it.

packages/engine-core/package.json
janpio: What do we think about this comment now?

packages/engine-core/package.json
```diff
@@ -59,7 +59,7 @@
     "new-github-issue-url": "0.2.1",
     "p-retry": "4.6.1",
     "terminal-link": "2.1.1",
-    "undici": "3.3.6"
+    "undici": "4.10.0"
```
janpio, line 62: Undici is at v4.12.1 now - do we want to let these updates happen via Renovate later?

packages/client/helpers/build.ts
```diff
@@ -1,30 +1,45 @@
...
+const inlineUndiciWasm = replaceWithPlugin([
+  [
+    /(await WebAssembly\.compile\().*?'(.*?)'\)\)\)/g,
+    async (regex, contents) => {
+      for (const match of contents.matchAll(regex)) {
+        if (match[2].includes('simd') === false) {
+          // we only bundle the wasm files that are not simd compiled
+          const engineCoreDir = resolve.sync('@prisma/engine-core')
+          const undiciPackage = resolve.sync('undici/package.json', { basedir: engineCoreDir })
+          const lhttpWasmPath = path.join(path.dirname(undiciPackage), 'lib', match[2])
```
aqrln, line 19: should this be `llhttp`?

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -605,9 +605,7 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
               json.fields?.message?.startsWith('Started http server')
```
janpio, line 605: only formatting

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -54,7 +54,7 @@ export class DataProxyEngine extends Engine {
     // That's because we instantiate the client outside of the request handler. This essentially prevents immediate execution of the promise.
```
janpio, line 54: only whitespace

packages/migrate/src/__tests__/DbCommand.test.ts
```diff
@@ -3,9 +3,7 @@ import { DbSeed } from '../commands/DbSeed'
```
janpio, line 3: only formatting

packages/migrate/src/__tests__/MigrateDeploy.test.ts
```diff
@@ -37,12 +37,9 @@ describe('sqlite', () => {
   it('no unapplied migrations', async () => {
```
janpio, line 37: only formatting

pnpm-lock.yaml
```diff
@@ -1755,7 +1757,6 @@ packages:
       express: 4.17.1
       untildify: 4.0.0
     transitivePeerDependencies:
-      - '@prisma/client'
```
janpio, line 1758: Unrelated change? Or what causes this?

helpers/compile/plugins/replaceWithPlugin.ts
millsp: I can't remember where I found this, but it is an internal API that is used all over the Node.js codebase for importing native libraries from C++ if I remember well. The param is a selector to get a lib (eg. `binding('fs')`) or filter all of them with `natives` (for Node's libraries). You'd get something like:
```
  '_http_agent',
  '_http_client',
  '_http_common',
  '_http_incoming',
  '_http_outgoing',
  '_http_server',
  '_stream_duplex',
  '_stream_passthrough',
  '_stream_readable',
  '_stream_transform',
  '_stream_wrap',
  '_stream_writable',
  '_tls_common',
  '_tls_wrap',
  'assert',
  'async_hooks',
  'buffer',
  'child_process',
  'cluster',
  'console',
  'constants',
  'crypto',
  'dgram',
  'dns',
  'domain',
  'events',
  'fs',
  'fs/promises',
  'http',
  'http2',
  'https',
```

janpio: Maybe worth looking at: https://github.com/nodejs/node/issues/3307

packages/engine-core/package.json
millsp: Better to do it know indeed.

helpers/compile/plugins/replaceWithPlugin.ts
millsp: Thanks a lot, I'll give a try to https://nodejs.org/api/module.html#modulebuiltinmodules. Looks more robust.

pnpm-lock.yaml
aqrln: Unrelated, it happens sometimes when running the setup script or when using `pnpm install` in one of the packages (which the setup script essentially does for each package separately). Might be a pnpm bug.

It doesn't really seem to influence anything whether this line is present in the lockfile or not, and even when this change is committed, it returns soon in one of the renovate PRs anyway. So there's no harm when this happens, it's just slightly annoying, but not annoying enough that someone took time and investigated or prepared a bug report for pnpm 😄

packages/engine-core/package.json
millsp: This is reserved for their experimental `fetch` API, which we don't use. This is type-only (to satisfy TS) and `Blob` isn't exported or used, so I consider this change fine.

packages/client/helpers/build.ts
```diff
@@ -1,30 +1,45 @@
...
+import fs from 'fs'
 
-const external = ['_http_common']
+const inlineUndiciWasm = replaceWithPlugin([
+  [
+    /(await WebAssembly\.compile\().*?'(.*?)'\)\)\)/g,
+    async (regex, contents) => {
+      for (const match of contents.matchAll(regex)) {
+        if (match[2].includes('simd') === false) {
+          // we only bundle lhttp wasm files that are not simd compiled
```
aqrln, line 16:
```suggestion
          // we only bundle llhttp wasm files that are not simd compiled
```

packages/client/helpers/build.ts
```diff
@@ -1,30 +1,45 @@
...
+          const engineCoreDir = resolve.sync('@prisma/engine-core')
+          const undiciPackage = resolve.sync('undici/package.json', { basedir: engineCoreDir })
+          const lhttpWasmPath = path.join(path.dirname(undiciPackage), 'lib', match[2])
```
aqrln, line 19:
```suggestion
          const llhttpWasmPath = path.join(path.dirname(undiciPackage), 'lib', match[2])
```

packages/client/helpers/build.ts
```diff
@@ -1,30 +1,45 @@
...
+          const wasmContents = (await fs.promises.readFile(lhttpWasmPath)).toString('base64')
```
aqrln, line 20:
```suggestion
          const wasmContents = (await fs.promises.readFile(llhttpWasmPath)).toString('base64')
```

aqrln (approved):

helpers/compile/plugins/replaceWithPlugin.ts
```diff
@@ -28,7 +26,8 @@ export const replaceWithPlugin = (replacements: Replacement[]): esbuild.Plugin =
     name: 'replaceWithPlugin',
     setup(build) {
       build.onLoad({ filter: /.*/ }, async (args) => {
-        if (nativeDependencies.has(args.path)) return {}
+        // we skip, don't attempt to edit files that aren't js
+        if (builtinModules.includes(args.path)) return {}
         if (!/.*?(.js|.mjs)$/.exec(args.path)) return {}
```
Jolg42, line 31:
```suggestion
        // skip Node.js native modules
        if (builtinModules.includes(args.path)) return {}
        // skip files that are not .js or .mjs
        if (!/.*?(.js|.mjs)$/.exec(args.path)) return {}
```

Jolg42 (approved):
Thank you for pushing this through the door, it might very well be useful for other wasm projects 💚

smasala (approved):

smasala (approved):

Jolg42:
This PR is current blocked because the following action job times out `CI / client (library, ubuntu-latest, 16) (pull_request) Cancelled after 25m`

Some investigation might be needed there first, what I found earlier was that something gets really slow during tests.
Example:
https://github.com/prisma/prisma/runs/4608802724?check_suite_focus=true#step:9:1493
```
 PASS src/__tests__/integration/happy/binary/test.ts (34.847 s)
  ✓ binary (34030 ms)
```
vs on main branch
```
https://github.com/prisma/prisma/runs/4605713109?check_suite_focus=true#step:9:1047
PASS src/__tests__/integration/happy/binary/test.ts (5.121 s)
  ✓ binary (4215 ms)
```

@millsp saw the heap size getting big too, so that might be linked.

juanzgc:
Is there any news on the state of this PR and whether it is still being blocked 😇 ?

I believe this PR would fix the following Issue: https://github.com/prisma/prisma/issues/6899
Which is currently preventing me from using Prisma in Turborepo as a package

helpers/compile/plugins/replaceWithPlugin.ts
```diff
@@ -14,15 +21,44 @@ function applyReplacements(contents: string, replacements: [RegExp, string][]) {
...
+//           const inlineWasm = `${match[1]}(Buffer.from("${wasmContents}", "base64")))`
+
+//           contents = contents.replace(match[0], inlineWasm)
+//         }
+//       }
+
+//       return contents
+//     },
+//   ],
+// ])
```
Jolg42, line 64: What is this example for @millsp ?

packages/engine-core/src/binary/Connection.ts
```diff
@@ -1,14 +1,17 @@
...
-  statusCode: Client.ResponseData['statusCode']
-  headers: Client.ResponseData['headers']
+  statusCode: Dispatcher.ResponseData['statusCode']
+  headers: Dispatcher.ResponseData['headers']
   data: R
 }
 
+// because undici lazily loads llhttp wasm which bloats the memory
+// TODO: hopefully replace with `import` but that causes segfaults
+const undici = () => require('undici')
```
Jolg42, line 13: So that's our current workaround for the segfaults for the Binary? any idea why is this happening?

helpers/compile/plugins/replaceWithPlugin.ts
millsp: It used to show how to inline wasm when the wasm file is not directly imported. However, in the mean time undici changed how they load the wasm and now puts the wasm contents in a .js file instead of doing a `readFileSync` so this is not needed anymore and works well when bundled.

packages/engine-core/src/binary/Connection.ts
millsp: I initially wanted to do `() => import('undici')` and I could not understand why this could cause segfaults in `jest` and only on CI. So I swapped for `() => require('undici')`.

helpers/compile/plugins/replaceWithPlugin.ts
janpio: Should we remove this then in a follow up PR?

packages/engine-core/src/binary/Connection.ts
janpio: Worth opening a PR again to see if it still happens, and if so maybe point it out to Undici/Node ppl?

millsp: I tried this today. I made `undici` not to load lazily like it does by re-wrapping it in `() => `. Somehow, their lazy loading of the wasm (ie. on module load) was also causing memory leaks in jest. It would make sense to debug where it segfaults when doing `import` instead of `require` so that we can report to them.

helpers/compile/plugins/replaceWithPlugin.ts
millsp: We can probably trim it a bit, but I'd like to keep an example for how to use `replaceWithPlugin`.

PR #8889, opened 2021-08-24
fix(deps): update undici to 4.x (remove one of the two copy actions for undici binary files)

janpio:
Seems this is not needed after all.


Jolg42:
Nice!

Jolg42 (approved):

Issue #8894, opened 2021-08-24
Warn on removed preview feature flags (+ PRISMA_HIDE_PREVIEW_FLAG_WARNINGS) + add tests


janpio:
`BinaryEngine` has a feature to warn on usage of outdated preview feature flags:
https://github.com/prisma/prisma/blob/fe3d63cc11678d756732ad14887eea62b2590c36/packages/engine-core/src/binary/BinaryEngine.ts#L174-L208 This does not seem to be present in the `LibraryEngine`.

(This also seems to remove the preview feature flag from the list submitted to the Engine if it is part of the list of removed flags - not sure if this is required/useful)

This functionality also seems to be completely untested.

jkomyno:
This was removed in Prisma 6.

PR #8895, opened 2021-08-24
refactor(BinaryEngine): cleanAndWarnRemovedPreviewFeatureFlags

janpio:
Could become part of https://github.com/prisma/prisma/issues/8894


Jolg42:
Is it something that should be finished @janpio ?

PR #9010, opened 2021-08-30
feat(tests): Run CLI tests on all platforms

janpio:


Jolg42:
Closing since this will require more work than tweaking the os line in GitHub Actions

PR #9030, opened 2021-08-31
fix(LibraryEngine): `enableDebugLogs` feature (`__internal.debug: true`)

janpio:
The Node-API library implementation used the wrong config value here (see BinaryEngine equivalent). As there are 0 tests for this, it unfortunately was not discovered earlier. (We only noticed in the context of https://github.com/prisma/prisma/pull/9031 when we renamed the variable, and this could would make no sense any more at all)

### DIFF
--- a/packages/engine-core/src/library/LibraryEngine.ts
+++ b/packages/engine-core/src/library/LibraryEngine.ts
@@ -98,7 +98,7 @@ export class LibraryEngine extends Engine {
     this.datasourceOverrides = config.datasources
       ? this.convertDatasources(config.datasources)
       : {}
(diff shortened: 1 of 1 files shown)


Jolg42 (approved):
Thanks for the cleanup!

Note: We need to wait for CI to be operational again to merge

Jolg42:
Note: undocumented constructor option is here
https://github.com/prisma/prisma/blob/2c3219ce54cca7330ab3e5eda29d0e16f02e8e43/packages/client/src/runtime/getPrismaClient.ts#L325-L331

PR #9031, opened 2021-08-31
chore(client): Rename enableEngineDebugMode to allowTriggerPanic

janpio:
Although the param of the Rust engine is called `--debug`, what you do in Prisma Client when enabling it, is something very different: You allow usage of the internal trigger panic method.

Renaming this makes sure no more confusion as in https://github.com/prisma/prisma/pull/9030/files can happen.

As this is an internal API, only used in tests, a rename is safe and simple.

### DIFF
--- a/packages/engine-core/src/library/LibraryEngine.ts
+++ b/packages/engine-core/src/library/LibraryEngine.ts
@@ -98,7 +98,7 @@ export class LibraryEngine extends Engine {
     this.datasourceOverrides = config.datasources
       ? this.convertDatasources(config.datasources)
       : {}
-    if (config.enableEngineDebugMode) {
+    if (config.enableDebugLogs) {
       this.logLevel = 'debug'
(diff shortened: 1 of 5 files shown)


packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -98,7 +98,7 @@ export class LibraryEngine extends Engine {
     this.datasourceOverrides = config.datasources
       ? this.convertDatasources(config.datasources)
       : {}
-    if (config.enableEngineDebugMode) {
+    if (config.enableDebugLogs) {
```
janpio, line 101: This change is also available standalone at https://github.com/prisma/prisma/pull/9030

Jolg42 (approved):
Thanks for the cleanup!

Note: We need to wait for CI to be operational again to merge

PR #9033, opened 2021-08-31
fix: see if studio tests fail on windows

eemmiillyy:
Test to see if studio tests will fail on windows

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -438,7 +438,7 @@ jobs:
       fail-fast: false
       matrix:
         queryEngine: ['library', 'binary']
-        os: [ubuntu-20.04] # macos-latest, windows-latest
+        os: [ubuntu-20.04, macos-latest, windows-latest] # macos-latest, windows-latest
         node: [12]
 
     steps:
--- a/packages/cli/src/Studio.ts
+++ b/packages/cli/src/Studio.ts
@@ -141,7 +141,7 @@ ${chalk.bold('Examples')}
           url: true,
         })
       } catch (e) {
-        // Ignore any errors that occur when trying to open the browser, since they should not halt the process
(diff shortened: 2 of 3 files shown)


Jolg42:
and they fail, nice!

The error is weird though
```
    +     data: null,
    +     error: Object {
    +       code: P500,
    +       message: The underlying schema has changed. Please reload Studio.,
    +       stack: ,
    +       type: PrismaClientSchemaDriftedError,
          },
```

eemmiillyy:
Yeah the error is weird :/ @Jolg42, I expect `schemaHash` not to change

Jolg42:
Might be because line ending and windows, always a funny one... 🙈

janpio:
Unfortunately also means that `open`, which is causing the failure on Windows, is not really tested here :/

eemmiillyy:
Yeah this isn't testing whether the `open` package **actually** opens a tab, it's just testing that the call doesn't throw an error. @janpio 
They do something similar in the tests inside the package: https://github.com/sindresorhus/open/blob/main/test.js
But maybe this isn't verifying even that open doesn't show an error because I also expect to see this somewhere: `Cannot open the browser to run studio: + e)` in the logs and I don't

janpio:
Probably swallowed by this: https://github.com/prisma/prisma/blob/cc453c562deb3026c6d96b7c6788ef9e04f9d61d/packages/cli/src/Studio.ts#L138-L145

eemmiillyy:
@janpio I actually changed that line in this PR to throw an error instead of silently fail 
https://github.com/prisma/prisma/pull/9033/files#diff-567c0692a5fdbc9e5017337b5e30c97a1f1b1b39e3a855ec68bd09ee28f65010R144

janpio:
Ah now I know what you mean. And am even more confused.

PR #9062, opened 2021-09-01
fix: studio cli on windows

eemmiillyy:


Jolg42:
version 423 was merged in https://github.com/prisma/prisma/pull/9060

PR #9068, opened 2021-09-01
fix(telemetry): filter generator list by provider for previewFeatures

janpio:
`name` is the name of the generator block (usually `client` or `prisma`, but could be anything), `provider` contains the name of the generator, in this context called `provider`.

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -203,9 +203,13 @@ async function main(): Promise<number> {
         schemaProvider = config.datasources[0].provider
       }
 
-      // restrict the search to previewFeatures of `provider = 'prisma-client-js'` 
+      // restrict the search to previewFeatures of `provider = 'prisma-client-js'`
       // (this was not scoped to `prisma-client-js` before Prisma 3.0)
-      const generator = config.generators.find((generator) => generator.provider === 'prisma-client-js' && generator.previewFeatures.length > 0)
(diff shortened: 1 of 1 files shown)


Jolg42 (approved):
Thanks!

Jolg42:
Needs a fix, I'm on it
```
ERROR: src/bin.ts:208:63 - error TS2367: This condition will always return 'false' since the types 'EnvValue' and 'string' have no overlap.
```

packages/cli/src/bin.ts
```diff
@@ -203,9 +203,13 @@ async function main(): Promise<number> {
         schemaProvider = config.datasources[0].provider
       }
 
-      // restrict the search to previewFeatures of `provider = 'prisma-client-js'` 
+      // restrict the search to previewFeatures of `provider = 'prisma-client-js'`
       // (this was not scoped to `prisma-client-js` before Prisma 3.0)
-      const generator = config.generators.find((generator) => generator.provider === 'prisma-client-js' && generator.previewFeatures.length > 0)
+      const generator = config.generators.find(
+        (generator) =>
+          parseEnvValue(generator.provider) === 'prisma-client-js' &&
```
janpio, line 210: I don't think `provider` is still allowed to be an env var though in reality?

Jolg42: Well I thought so, there was a whole discussion about that, to be double checked then.

PR #9074, opened 2021-09-02
fix(client): json null filtering

### LINKED ISSUE #8399: Can't filter by `null` values on JSONB fields
### Bug description

Can't filter by `null` values on JSONB fields

### How to reproduce

1. Run query with `null` value filter

```typescript
prisma.model.findMany({
  where: {
    NOT: {
      data: {
        path: ["foo"],
        equals: null,
      },
    },
  },
});
```

2. See error:

> Argument equals for where.NOT.0.data.equals must not be null. Please use undefined instead.

### Expected behavior

As JSONB values are allowed, I should be able to find values filtered by `null`.

### Prisma information

```
model model {
  id               String            @id @db.Uuid
  data             Json
}
```

### Environment & setup

- OS: MacOS
- Database: PostgreSQL
- Node.js version: 14.x

### Prisma Version

```
`"@prisma/client": "^2.20.1"`
```

millsp:
fixes https://github.com/prisma/prisma/issues/8399

BREAKING CHANGE


packages/client/src/generation/TSClient/common.ts
```diff
@@ -171,8 +180,16 @@ export type InputJsonObject = {[Key in string]?: JsonValue}
...
-export type InputJsonValue = undefined |  string | number | boolean | null | InputJsonObject | InputJsonArray
- type SelectAndInclude = {
+export type InputJsonValue = string | number | boolean | InputJsonObject | InputJsonArray
+
+/**
+ * Shorthand utilities for JSON filtering
+ */
+export const DbNull: 'DbNull'
+export const JsonNull: 'JsonNull'
+export const AnyNull: 'AnyNull'
```
Jolg42, line 190: Could use some comment or link to doc about what they are maybe

packages/client/src/__tests__/integration/happy/json-filtering-postgres/test.ts
```diff
@@ -208,6 +216,60 @@ describe('json-filtering(postgres)', () => {
...
+        jsonOpt: {
+          equals: PrismaUtil.JsonNull,
+        },
+      },
+      select: {
+        email: true,
+      },
+    })
+
+    expect(user).toMatchSnapshot()
```
Jolg42, line 235: Not sure a snapshot is the best here, we can keep like that and maybe discuss if we should also avoid it sometimes and prefer a simple expect length one and expect email = "email" which is stronger than a snapshot (snapshots can be updated by mistake)

Jolg42 (changes requested):
Looks good! need to resolve conflicts

- not sure why GitHub Action test was not triggered?
- Builkite publish failed a test, but passed in test pipeline?

- Also only merged to engines master now https://github.com/prisma/prisma-engines/pull/2181 so only merge when it's out of the pipeline and updated in this PR

Jolg42:
New engine is ready https://github.com/prisma/prisma/pull/9134

Jolg42:
Merged main in with latest engine

packages/client/src/__tests__/integration/happy/json-filtering-postgres/test.ts
millsp: Some tests return many emails, would you prefer an inline snapshot in this case?

millsp:
* Added sorting on the data to keep snapshots consistent (reason why tests failed).
* Added some documentation hints on the TypeScript level.

Jolg42 (approved):
Nice work! Let's merge this!

PR #9095, opened 2021-09-02
fix(tests): __internal_triggerPanic now also exists for Node-API library

### LINKED ISSUE #7810: triggerPanic for Node-API Library
To be able to test the error reporting mechanisms of the CLI, we will need to also support `triggerPanic` in the Node-API library as we do in the binary.

https://github.com/prisma/prisma/blob/120f04c6f833721ea9109379b299bb67b22dfcbc/packages/client/src/__tests__/integration/errors/error-link/test.ts#L8-L11

https://github.com/prisma/prisma/blob/fc7b272bfa7ee030c2c8dbede73e0da4963f4c67/packages/client/src/runtime/getPrismaClient.ts#L856-L886

janpio:
closes https://github.com/prisma/prisma/issues/7810


Jolg42:
Closing as duplicate of https://github.com/prisma/prisma/pull/9095

janpio:
This is #9095 @Jolg42!?

Jolg42:
My mistake, reopening and closing the other one

janpio:
If I remember correctly: Method exists, but does not achieve the same result to enable us testing the same things.

janpio:
Indeed, current test diff:

```
FAIL src/__tests__/integration/errors/error-link/test.ts
  ● error-link

    expect(received).rejects.toThrowErrorMatchingInlineSnapshot(snapshot)

    Snapshot name: `error-link 1`

    - Snapshot  - 10
    + Received  +  1

    - Query engine debug fatal error, shutting down.
    -
    - This is a non-recoverable error which probably happens when the Prisma Query Engine has a panic.
    -
    - TEST_GITHUB_LINK
    -
    - If you want the Prisma team to look into it, please open the link above ��
    - To increase the chance of success, please post your schema and a snippet of
    - how you used Prisma Client in the issue. 
    -
    + called `Option::unwrap()` on a `None` value

      16 |   })
      17 |
    > 18 |   await expect(db.__internal_triggerPanic(true)).rejects.toThrowErrorMatchingInlineSnapshot(`
         |                                                          ^
      19 |           Query engine debug fatal error, shutting down.
      20 |
      21 |           This is a non-recoverable error which probably happens when the Prisma Query Engine has a panic.

      at Object.toThrowErrorMatchingInlineSnapshot (../../node_modules/.pnpm/expect@27.5.1/node_modules/expect/build/index.js:285:22)
      at Object.<anonymous> (src/__tests__/integration/errors/error-link/test.ts:18:58)
```

So seems the error that it throws is not caught the same way as the binary engine one, which then leads to the "Query engine debug fatal error, shutting down." case handling not being triggered. This might either be a problem in a) the way it throws the error or b) that the handling is just not versatile enough to handle both yet.

PR #9126, opened 2021-09-05
ci(publish): remove last occurrences of 2

millsp:
This PR removes the last occurrences of Prisma 2 in our publish script.

In essence, this removes the `2` out of a filtering `RegExp` for `intergation`, `dev`, and `patch-dev`. I verified the possible side-effects of doing this:
* `getNewPatchDevVersion`: on this, it would allow us to publish a patch on any version (any major)
* `getNewIntegrationVersion`: nothing should change here, we check on which version is latest stable and calculate the next integration version while still filtering with `getAllVersions`
* `getNewDevVersion`: nothing should change here, we check on which version is latest stable and calculate the next integration version while still filtering with `getAllVersions`

### DIFF
--- a/scripts/ci/publish.ts
+++ b/scripts/ci/publish.ts
@@ -421,7 +421,7 @@ async function getNewPatchDevVersion(
 }
 
 function getMaxDevVersionIncrement(versions: string[]): number {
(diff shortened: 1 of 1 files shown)


Jolg42 (approved):
Thanks!

PR #9177, opened 2021-09-08
fix(generate): remove misleading "breaking changes" message for 3.0+

janpio:
close #9171

- This functionality is missing tests (obviously)
- We could also add a warning to review breaking changes and release notes when going from 2.x to 3.x using this functionality

PR #9189, opened 2021-09-09
chore: fix publish on patch if minor = 0

millsp:


millsp (commented):
To be merged later

scripts/ci/publish.ts
```diff
@@ -1286,7 +1289,7 @@ function getPatchBranch() {
     const versions = getSemverFromPatchBranch(process.env.BUILDKITE_BRANCH)
     console.debug('versions from patch branch:', versions)
 
-    if (versions?.minor) {
+    if (versions !== undefined) {
```
janpio, line 1292: Does that do the same?

scripts/ci/publish.ts
```diff
@@ -1321,15 +1324,19 @@ async function sendSlackMessage({
...
 
 \`prisma/prisma-engines\`
-<https://github.com/prisma/prisma-engines/commit/${enginesCommit.hash}|${enginesLines[0]
+<https://github.com/prisma/prisma-engines/commit/${enginesCommit.hash}|${
+      enginesLines[0]
     }\t\t\t\t-  ${enginesCommit.hash.slice(0, 7)}>
-${enginesLines.slice(1).join('\n')}${enginesLines.length > 1 ? '\n' : ''
+${enginesLines.slice(1).join('\n')}${
+      enginesLines.length > 1 ? '\n' : ''
     }Authored by ${enginesCommit.author}`,
```
janpio, line 1340: Isn't the previous formatting so much better?

scripts/ci/publish.ts
millsp: No, if `versions?.minor` returns `0`, it fails the release. Otherwise, yes because the regex would fail otherwise (it has 2 capturing groups).

scripts/ci/publish.ts
millsp: I agree, but prettier knows so much better about readable code :trollface:

packages/cli/src/Generate.ts
```diff
@@ -211,12 +211,14 @@ Please run \`${getCommandWithExecutor('prisma generate')}\` to see the errors.`)
...
+
         if (
           clientVersionBeforeGenerate &&
           typeof clientVersionBeforeGenerate === 'string'
         ) {
-          const minor = clientVersionBeforeGenerate.split('.')[1]
-          if (parseInt(minor, 10) < 12) {
+          const [major, minor] = clientVersionBeforeGenerate.split('.')
+
+          if (parseInt(major) == 2 && parseInt(minor) < 12) {
```
Jolg42, line 221: Done in https://github.com/prisma/prisma/pull/9177

Jolg42:
Looks like a duplicate of https://github.com/prisma/prisma/pull/9177

But changes made to publish.ts are not in main branch

Jolg42 (approved):

PR #9295, opened 2021-09-16
fix: emitted patterns for engine resolution

netlify-team-account-1:
`prisma generate` currently results in these lines to be emitted:

```js
  '"node_modules/.prisma/client"',
  '".prisma/client"',
```

There's one pair of quotes too much, which causes `findSync` to not find anything. Since `dirname` is used to resolve the location of the Prisma engine, it's important that it is passed correct parameters. At the moment, it will never find anything and resort to `__dirname`, breaking deploy for some bundlers.

This can be fixed by removing the outer quotes, as `JSON.stringify` will already emit the double-quotes that mark it as a string literal.

See https://github.com/netlify/zip-it-and-ship-it/issues/630 for reference.

### DIFF
--- a/packages/client/src/generation/TSClient/TSClient.ts
+++ b/packages/client/src/generation/TSClient/TSClient.ts
@@ -90,8 +90,8 @@ export class TSClient implements Generatable {
 
 // folder where the generated client is found
 const dirname = findSync(process.cwd(), [
-  '${JSON.stringify(relativeOutputDir)}',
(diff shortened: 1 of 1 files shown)


Jolg42:
Nice catch and thanks for the PR!

millsp (approved):

FredericEspiau:
https://github.com/prisma/prisma/pull/8698 well I guess I can close mine then

PR #9300, opened 2021-09-16
chore: binaryVersion => engineVersion, Binary => Engine

janpio:
This makes the basic error reports list the engine version as `Engine` instead of `Binary`. It also cleans up a lot of variable names in related code.

### DIFF
--- a/packages/sdk/src/sendPanic.ts
+++ b/packages/sdk/src/sendPanic.ts
@@ -21,7 +21,7 @@ tmp.setGracefulCleanup()
 export async function sendPanic(
   error: RustPanic,
   cliVersion: string,
-  binaryVersion: string,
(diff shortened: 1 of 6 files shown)


packages/sdk/src/sendPanic.ts
```diff
@@ -71,7 +71,7 @@ export async function sendPanic(
       area: error.area,
       kind: ErrorKind.RUST_PANIC,
       cliVersion,
-      binaryVersion,
+      binaryVersion: engineVersion,
```
janpio, line 74: external API of error reporting service that can not be changed here

millsp (approved):

Jolg42 (approved):
Nice!

PR #9422, opened 2021-09-24
feat(client): cfw support

millsp:
This is still a work in progress (at prototype stage), as we don't have any kind of testing yet besides the manual ones we do.

PS: Save yourself the pain of going through the diff, use your favorite git diff tool and watch commit per commit for the changes.

From a higher view, this PR also brings in many necessary changes:
* A refactor of our whole compile pipeline, now with a centralized build system.
* Support for ESM modules (not enabled publicly), but picked up by the bundlers.
* A new runtime build in order to support CFW, with its all node polyfill injections.
* Tree-shaking of CommonJS modules via the compilation of tree-shaken ESM modules.
* Some cleanup here and there in order to make things work more easily.
* A new utility library that is aimed towards performance and type-safety.
* And a bit of documentation.


millsp:
Closing in favor of https://github.com/prisma/prisma/pull/9561

PR #9461, opened 2021-09-27
fix(sdk): check all generators for mongoDb preview flag before throwing

Jolg42:
Context: https://prisma-company.slack.com/archives/CEYCG2MCN/p1632236946009100

Quickly implemented something, to be confirmed it's what we want to do


steebchen (approved):
I can confirm that this fixes the issue 👍

millsp (approved):

Issue #9184, opened 2021-09-09
EPERM: operation not permitted, unlink '...node_modules\prisma\query_engine-windows.dll.node'


RedGlow:
### Bug description

Following the [Getting Started](https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/using-prisma-migrate-typescript-postgres/) guide, when it gets to the point of running the "migrate dev" command, the script tries to install some packages through npm (don't know which ones) and this causes the error in title.

### How to reproduce

1. Follow the [Getting started](https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/using-prisma-migrate-typescript-postgres/) prisma guide up to and included the "npx prisma migrate dev --name init" command.
2. Get the following error:

```
Environment variables loaded from .env
Prisma schema loaded from prisma\schema.prisma
Datasource "db": PostgreSQL database "testprisma", schema "public" at "localhost:5432"

PostgreSQL database testprisma created at localhost:5432

The following migration(s) have been created and applied from new schema changes:

migrations/
  └─ 20210909080511_init/
    └─ migration.sql

Your database is now in sync with your schema.

Running generate... (Use --skip-generate to skip the generators)
npm WARN hello-prisma@1.0.0 No description
npm WARN hello-prisma@1.0.0 No repository field.

npm ERR! code EPERM
npm ERR! syscall unlink
npm ERR! path C:\Users\mbelletti\Documents\Projects\hello-prisma\node_modules\prisma\query_engine-windows.dll.node
npm ERR! errno -4048
npm ERR! Error: EPERM: operation not permitted, unlink 'C:\Users\mbelletti\Documents\Projects\hello-prisma\node_modules\prisma\query_engine-windows.dll.node'
npm ERR!  [OperationalError: EPERM: operation not permitted, unlink 'C:\Users\mbelletti\Documents\Projects\hello-prisma\node_modules\prisma\query_engine-windows.dll.node'] {
npm ERR!   cause: [Error: EPERM: operation not permitted, unlink 'C:\Users\mbelletti\Documents\Projects\hello-prisma\node_modules\prisma\query_engine-windows.dll.node'] {
npm ERR!     errno: -4048,
npm ERR!     code: 'EPERM',
npm ERR!     syscall: 'unlink',
[...]

PR #9513, opened 2021-09-29
fix(getDMMF/getConfig): Unset NodeAPIQueryEngineLibrary after use

janpio:
Playing around with https://github.com/prisma/prisma/issues/9184


Jolg42:
Is this fixing something?

Jolg42:
Ok, got it, thanks!

janpio:
Doesn't work.

PR #9523, opened 2021-09-29
fix(client): merge changed params in middleware (#9522)

### LINKED ISSUE #9522: Changing prisma query params using middleware doesn't work
### Bug description

I try to implement soft delete middlewares reading the official docs but it does not work as expected. Middleware ignores the transformation of query params like `params.action` or `params.args` and throw an exception as if it hadn't transformed anything. 

Prisma v2 works correctly, Prisma v3 no. I already investigated and fixed it, so I try to send a PR https://github.com/prisma/prisma/pull/9523. I noticed that this issue is coming from this [commit](https://github.com/prisma/prisma/commit/f6c69f7d6ac1baec47702253ecaaff94010eb053).

```
Invalid `prisma.account.delete()` invocation:
{
  where: {
    accountId: 'CUST04'
  },
  data: {
  ~~~~
    deletedAt: new Date('2021-09-29T15:53:39.420Z')
  }
}
```

### How to reproduce

1. Create a middleware like the following:
```ts
  if (params.model == 'Foobar') {
    if (params.action == 'delete') {
      params.action = 'update'
      params.args['data'] = { deletedAt: new Date() }
    }
  }
  return next(params)
})
```

2. Delete a Foobar record `await prisma.Foobar.delete({ where: { id: 'test' }})`

### Expected behavior

I expect to find the Foobar record with the field `deletedAt` filled with a DateTime.

### Prisma information

`await prisma.account.delete({ where: { accountId: 'CUST04' } })`

### Environment & setup

- OS: **macOS Big Sur**
- Database: **PostgreSQL**
- Node.js version: **16.4.0**

### Prisma Version

```
Query Engine (Node-API) : libquery-engine c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary           : prisma-fmt c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/prisma-fmt-darwin)

```

cesconix:
Fix #9522


millsp:
Good catch, thank you! We should add tests for this, since it's a feature.

cesconix:
Looking more closely I realized that the issue comes from the way `params` parameter is handled inside the middleware consumer function. 

In my case, using a functional approach, I return a shallow clone of `params` object to avoid reassignment of function parameters (https://eslint.org/docs/rules/no-param-reassign) and I noticed that `params.args` was uptaded successfully (because of reference) but `params.action` no.

So, to make it work, **you have necessarily to mutate the `params` object and return its reference**. In this way, the closure internal params will be mutated correctly (in the private _request method). **Shallow copy is not allowed**. I don't know if it could be useful to describe this situation in the docs.

Maybe the PR could resolve these side effects. Thanks in advance for your feedback!

```
private _request(internalParams: InternalRequestParams): Promise<any> {
  try {
    let index = -1
    // async scope https://github.com/prisma/prisma/issues/3148
    const resource = new AsyncResource('prisma-client-request')
    // make sure that we don't leak extra properties to users
    const params: QueryMiddlewareParams = {
      args: internalParams.args,
      dataPath: internalParams.dataPath,
      runInTransaction: internalParams.runInTransaction,
      action: internalParams.action,
      model: internalParams.model,
    }

    // prepare recursive fn that will pipe params through middlewares
    const consumer = (changedParams: QueryMiddlewareParams) => {
      // if this `next` was called and there's some more middlewares
      const nextMiddleware = this._middlewares.query.get(++index)

      if (nextMiddleware) {
        // we pass the modfied params down to the next one, & repeat
        return nextMiddleware(changedParams, consumer)
      }

      const changedInternalParams = { ...internalParams, ...params }

      // TODO remove this once LRT is the default transaction mode
      if (index > 0 && !this._hasPreviewFlag('interactiveTransactions')) {
        delete changedInternalParams['transactionId']
      }

      // no middleware? then we just proceed with request execution
      return this._executeRequest(changedInternalParams)
    }

    return resource.runInAsyncScope(() => consumer(params))
  } catch (e) {

    throw e
  }
}
```

franky47:
As discussed in https://github.com/prisma/prisma/discussions/10446#discussioncomment-1708582, mutating the params in-place could pose problems for user code that uses shared query objects: the updates would reflect there too.

A solution could be to do a deep clone of the params object before sending it down the middleware chain (this copy can be avoided if there are no middlewares), but there might be others.

matthewmueller:
Next steps: add a test

aqrln:
> As discussed in #10446 (reply in thread), mutating the params in-place could pose problems for user code that uses shared query objects: the updates would reflect there too.

Actually, looking at the code and the existing tests, mutating the `params` object to set `params.args` to a new object (e.g., a deep copy of `params.args` with some field changed) rather than mutating `params.args.data.someField` in-place should be safe and would have worked even before these changes.

That said, there's no reason why passing a completely new `params` object shouldn't work, and judging from the surrounding code and variable names, this, apparently, had been the intention, and `params` instead of `changedParams` was probably a typo.

I have added a new test to this PR that fails without these changes and passes with them.

Jolg42 (approved):
Nice work!

PR #9556, opened 2021-10-01
feat(dbpull): MongoDB

Jolg42:
Note: MongoDB is a preview feature

db pull --url is not supported with MongoDB provider because it's a preview feature (it works only if a local schema has the preview flag in it)

re-introspection does not work (overwrites at the moment)


Jolg42:
We can merge this after a review.

Next to be done https://github.com/prisma/prisma/issues/9557 (I'll check on Monday for this)

millsp (approved):

PR #9652, opened 2021-10-07
ci(publish): take major version into account for calculating patch version

janpio:


Jolg42 (approved):
PR is good for me, that means a new dev version should appear as 3.2.1-dev.1  after merge

PR #9683, opened 2021-10-08
feat(client): add inline docs to generated TS

ifiokjr:
When consuming the generated client TypeScript models the doc string doesn't appear. 

```prisma
model Post {
  /// This will not be seen 😢
  id    String @id @default(uuid())
  /// Neither will this.
  title String
  text  String
}
```

In the following example, hovering over the `post.id` only shows a type, it doesn't pass through the comment. 

```ts
async function createPostAndReturnId(data: { title: string, text: string }) {
  const post = awaitprisma.post.create({ data });
  return post.id;
}
```

This is optional and perhaps there's a reason not to do it, but I feel that inline documentation makes it easier to work within larger teams.

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema/__snapshots__/test.ts.snap
@@ -1764,8 +1764,8 @@ export namespace Prisma {
   }
 
 
-    
-    
+
+
   export type PostGroupByArgs = {
     where?: PostWhereInput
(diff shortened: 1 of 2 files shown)


millsp:
Hey @ifiokjr, thank you for this PR! Would you mind adding a test to make sure this continues working? Thanks

millsp:
Thanks for your contribution. I ported your PR there https://github.com/prisma/prisma/pull/10248 and added doc generation for fields like you wanted. I still have to write some tests.

ImLunaHey:
Is there a reason this implements comments with three `/` instead of the usual two?
Was really confusing when my comments worked but wouldn't show up in the generated types.

PR #9747, opened 2021-10-13
test: fix platform regex for ARM64 macOS

aqrln:
* Sort platform names in the platform regex used in tests by string length in
  descending order. This fixes snapshot tests for error messages failing on M1
  Mac due to `darwin-arm64` being replaced with `TEST_PLATFORM-TEST_PLATFORM64`
  instead of just `TEST_PLATFORM`.

* Move `platformRegex` to `@prisma/sdk` to avoid duplication.


Jolg42 (approved):
Great PR! nice and clean ✨

Issue #6882, opened 2021-05-01
Prisma should allow passing request data to middleware


Sytten:
## Problem

Currently the middleware are essentially stateless. There is one example that uses [global state](https://www.prisma.io/docs/concepts/components/prisma-client/middleware/session-data-middleware) but in practice this is not very common nor desirable. It would be necessary for a tracing middleware.

## Suggested solution

It would be nice if all prisma calls add a context parameter that would be passed down to middlewares.
It could look like:
```typescript
export type TicketCreateArgs = {
    select?: TicketSelect | null
    include?: TicketInclude | null
    data: XOR<TicketCreateInput, TicketUncheckedCreateInput>
    context: Record<string, any>
}
```

```typescript
export type MiddlewareParams = {
    model?: ModelName
    action: PrismaAction
    args: any
    dataPath: string[]
    runInTransaction: boolean
    context: Record<string, any>
}
```

## Alternatives

No good alternative currently if you want to respect typing, unsure if I could pass the context in the data and remove it in the middleware, but I would prefer not too.

janpio:
Similar to https://github.com/prisma/prisma/issues/4217

remioo:
I see a lot of moving around of the milestones on this one. Is this something that is considered for implementation short term?

janpio:
Milestones mean we are looking at things or thinking about them. Things are also taken out of milestones because of changing priorities. Our only communicated roadmap is our Roadmap: http://pris.ly/roadmap It is on another level though, so will not reflect smaller features like this one most of the time.

janpio:
This related issue lists a relevant use case for this and why this might be helpful: https://github.com/prisma/prisma/issues/5128#issuecomment-848890199

dundich:
to provide context for middleware it's better like this, IMHO
```
const prisma = new PrismaClient()

// for per request 
prisma.$context = ({ req }) => ({
    mydata: "foo"
 })

prisma.$use(async (params, next, /* !!! * /context) => {...}
```
like as Apollo

AllanOricil:
related https://github.com/prisma/prisma/issues/10895

blevine:
I would also find this useful.  I'm trying to write a piece of auditing middleware and one of the things I want to store in the audit records is the userid of the user making the request. I can get this information from my GraphQL resolver and then I'd like to pass that as context in each Prisma query where it can then be read by my audit middleware.

KATT:
+1. 

[...]

PR #9759, opened 2021-10-13
feat(client): add context to middleware

reharik:
This pr adds a state variable, _context, and a setter $setContext, to the PrismaClient class.
I don't know much about the over all code base or coding standards or anything else.  This is really just to see if y'all would be open to this feature and if I'm anywhere close to getting it right.
I'm sure there are things I'm missing or preferences that maintainers may have, and I welcome any feedback that would make this feature more acceptable.


AllanOricil:
Would this PR allow me to access `req.user.role` inside a Prisma Middleware? Im currently needing it to be able to enforce the user role before running any queries. I was able to do so, but I kind did a hack https://github.com/prisma/prisma/issues/10895

raman-nbg:
Any update on this PR? What is necessary to accept this?

matthewmueller:
Hey folks, sorry for the delay on this one. Do you mind sharing why you need this? 

I can imagine this being potentially useful for middleware depending on each other, but otherwise I'd suggest passing data in through functions, something like _(untested)_:

```ts
prisma.$use(attachRole("admin"))
```

brxck:
Hey @matthewmueller! The most common use case I've seen for this making middleware to be aware of the current request/session in the context of a webserver, e.g. the current user. 

This is actually one of the examples given in the [Prisma docs](https://www.prisma.io/docs/concepts/components/prisma-client/middleware/session-data-middleware); however, it uses some external piece of state to achieve that. This isn't realistic because one will generally want to create the Prisma client outside the scope of individual requests. 

In this case it's not possible pass context directly to the middleware because calling `prisma.$use()` per request will add multiple middleware instances to the stack, and there is no way to remove existing middleware from the stack.

The discussions in https://github.com/prisma/prisma/issues/6882 and https://github.com/prisma/prisma/issues/10895 are centered around this scenario.

I think this would be a very useful addition!

psikoi:
> Hey folks, sorry for the delay on this one. Do you mind sharing why you need this?

My use case is a bit different, I'll try to hugely simplify it, but still would benefit from metadata:

Let's say I have a `Measurement` table, with `id, type, value, date`, this gets populated every hour with a value and the current date.

I also have another table, `Record`, that holds the current best measurements of each type. There's middleware on "create Measurement", that then does a few comparisons to previous records to determine if this new measurement is an new all-time record.

Now, let's assume that during the creation of Measurement, we computed the value of a local variable `isChange`, that compared the current measurement value, with the previous.

As you can imagine, if the current measure is no different than the previous, there's no point in checking if it's a new record, so it would be very useful if we could pass this extra metadata, that does NOT get saved in the database, and therefore does not exist in the model/middleware params. This way, we could limit "record checks" to only measurements that are actually potential new records.

**Sequelize achieves this through virtual fields, but I am unaware of a proper way to do this using Prisma.**

Creation Example:

```typescript
const isChange = ...

const newMeasurement = await prisma.measurement.create({
  data: {
    type: ...,
    value: ...
  },
  metadata: {
     isChange
  }
})
```

Middleware Example:

```typescript
   if(params.metadata.isChange){
       // check if this new measurement is a new record!
   }
);
```

okomarov:
> Hey folks, sorry for the delay on this one. Do you mind sharing why you need this?
@matthewmueller - I have two use cases:
1. Audit logging
2. Tracing

# Audit logging
You could write to some AuditLog table directly with prisma:
```
  const result = await next(params);

  // Log the changes starting with the user ID of the request
  await prisma.auditLog.create({data: {user_id: REQUEST_USER_ID, ...}}
  return result
}
```
or you could set the current user id in the database, and use databse triggers:

```
  await prisma.$executeRaw(`SET current_user_id = ${REQUEST_USER_ID}`)
  return next(params);
}
```
either way I need to pass the `REQUEST_USER_ID` to the middleware on a **per-request basis** but you should
> Add Prisma middleware **outside the context of the request handler**, otherwise each request adds a new instance of the middleware to the stack

# Tracing
I am aware that proper tracing is being worked on. In the meantime we've been using this:
```
  const { model, action } = params;
  const description = [model, action].filter(Boolean).join('.');
  
  const sentryTransaction = SentryGqlTransaction.getInstance();
  const span = sentryTransaction.startChild({
    op: 'db',
    description,
    data: params,
  });

  const result = await next(params);
  span?.finish();

  return result;
});
``` 
and we would love to pass the sentry transaction, initialised on a per-request-basis, and passed down from the rquest context, but we have to do with a singleton.

rhysawilliams2010:
@matthewmueller is there a reason to not use [continuation-local-storage](https://www.npmjs.com/package/continuation-local-storage)  or [cls-hooked](https://www.npmjs.com/package/cls-hooked) to store the context you need and access it at runtime in your middleware?

okomarov:
> @matthewmueller is there a reason to not use [continuation-local-storage](https://www.npmjs.com/package/continuation-local-storage) or [cls-hooked](https://www.npmjs.com/package/cls-hooked) to store the context you need and access it at runtime in your middleware?

We've gone this direction with Async Local Storage from Node's [async_hooks](https://nodejs.org/api/async_context.html), however, I feel that it's more common in Node to passing app context explicitly, as opposed to e.g. web services in Python (Flask)

trevoro:
@rhysawilliams2010 @okomarov Oh that's interesting; the idea of using async_hooks. 

Do you have a good example of where you've been able to make that work well? Perhaps a tiny snippet? 
We're using it for other parts of our application for logging but haven't been able to get it wired into Prisma properly.

okomarov:
> Do you have a good example of where you've been able to make that work well? Perhaps a tiny snippet?

@trevoro Not sure it's wired in properly: https://gist.github.com/okomarov/8b8d53b8f5d894ef818a6d009861f10d

In our case, Koa middleware first thing injects CLS. There might be some weird order on which you chain Koa middleware and Prisma middleware. Our real case is not pretty and that's why passing context to prisma middleware would make it more robust in general.

matthewmueller:
FYI, there's lots more use cases provided in [this discussion](https://github.com/prisma/prisma/discussions/13366).

millsp:
Related: https://github.com/prisma/prisma/pull/13923

matthewmueller:
More use cases: https://github.com/prisma/prisma/discussions/13366

pedroeckel:
Any update about this feature?

LubomirGeorgiev:
Any progress here?

exAspArk:
> Do you have a good example of where you've been able to make that work well? Perhaps a tiny snippet?

Here is another tiny example of a [Prisma extension](https://www.prisma.io/docs/orm/prisma-client/client-extensions) and context passing by using an [Express middleware](https://expressjs.com/) and native Node.js [async_hooks](https://nodejs.org/api/async_context.html) mentioned by @okomarov 💛 

* https://gist.github.com/exAspArk/d7d5774276b7981931400781608e4770

PR #9792, opened 2021-10-15
feat(sdk): add resolved TypeScript path to logs and error message

### LINKED ISSUE #7904: TypeScript version detection: add typescriptPath to error message or logs
I merged https://github.com/prisma/prisma/pull/7859 maybe too fast and missed this comment from @janpio 

> Should we maybe output typescriptPath in the error message as well to give context what typescript we detected? Maybe we will have many false positives, and this would make it visible to users.
> 
> If not in error messages, definitely some debug logging.

aqrln:
Closes #7904

I took "or" as a non-exclusive or and added it both to logs and the message. IMO, doesn't hurt to have it in logs either way (more so that the surrounding code logs many similar things in the same fashion).

To do later: e2e test for this scenario (as per discussion).

### DIFF
--- a/packages/sdk/src/predefinedGeneratorResolvers.ts
+++ b/packages/sdk/src/predefinedGeneratorResolvers.ts
@@ -200,6 +200,7 @@ async function checkTypeScriptVersion() {
     const typescriptPath = await resolvePkg('typescript', {
       basedir: process.cwd(),
     })
+    debug('typescriptPath', typescriptPath)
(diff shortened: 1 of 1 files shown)


janpio (approved):

janpio:
Seeing the PR like that of course makes clear that having a test here would ultimately be optimal as we could approve the test and snapshots together with the code. Hopefully we get there one day with our test system here.

PR #9872, opened 2021-10-20
fix(build): use os.tmpdir() instead of hardcoded /tmp path

aqrln:
Use `os.tmpdir()` instead of hardcoded `/tmp` path in `fillPlugin.ts`.

This fixes build errors like these on Windows:

```
error: Plugin "fillPlugin" returned a non-absolute path: \tmp\esbuild\buffer83013d5a.js (set a namespace if this is not a file path)
```


millsp (approved):

Jolg42 (approved):
Thanks!

PR #9882, opened 2021-10-21
chore: update pnpm-lock.yaml

aqrln:
In #9858 new entries pertaining to Jest 27.3.1 were added to `pnpm-lock.yaml` but old entries for Jest 27.2.5 weren't removed, which results in this diff after running `pnpm i && pnpm run setup`.


Jolg42 (approved):

janpio:
Is this a bug of Renote? One we should possibly report to them?

PR #10020, opened 2021-10-29
ci: test Node.js v16

Jolg42:
https://nodejs.org/en/about/releases/

v16 is LTS since `2021-10-26`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ on:
 
 env:
   PRISMA_TELEMETRY_INFORMATION: 'prisma test.yml'
-  NODE_OPTIONS: '--max-old-space-size=2048'
+  NODE_OPTIONS: '--max-old-space-size=8096'
 
 jobs:
   detect_jobs_to_run:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -234,7 +234,7 @@ jobs:
           - mysql
           - mariadb
           - mssql
-        node: [12, 16]
+        node: [12]
 
     steps:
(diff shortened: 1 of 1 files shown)


aqrln (approved):

millsp (approved):

janpio:
It seems in the last run it failed because of cancellation after 20 minutes instead:

```
Error: The operation was canceled.
```

aqrln:
@janpio FWIW, I restarted the cancelled pipeline and now it crashed with OOM on Node.js 16 again.

janpio:
Been a long time since I was happy about getting an OOM 🤣

aqrln:
Even 2 GB is not enough, it seems 😄

.github/workflows/test.yml
```diff
@@ -45,6 +45,7 @@ on:
 
 env:
   PRISMA_TELEMETRY_INFORMATION: 'prisma test.yml'
+  NODE_OPTIONS: '--max-old-space-size=8096'
```
aqrln, line 48: Amazing

aqrln (approved):

.github/workflows/test.yml
aqrln: 😂

Jolg42: 😅  I think I might scope this to client test now to avoid leaking this detail to global config

Jolg42: In next PR

peterp:
For what it's worth we're seeing a memory leak in Snaplet on Node16 as well.

Jolg42:
@peterp With which version of Prisma?

PR #10041, opened 2021-11-01
feat: Allow `prisma db push` on mongodb

tomhoule:
It is now implemented in the migration engine, so we should accept
running it in the CLI.

Engines PR: https://github.com/prisma/prisma-engines/pull/2287

The feature gate for mongodb on migrate/db commands is implemented in
ensureDatabaseExists / ensureCanConnectToDatabase, which is not going to
work anymore since we want to act differently in `migrate dev` and other
commands. We will implement a more fine-grained one in the engine
directly.


tomhoule:
See the screenshots here for the same warnings re-implemented engines-side: https://github.com/prisma/prisma-engines/pull/2362

aqrln (approved):

PR #10046, opened 2021-11-01
feat: skip creation of database

maxkomarychev:
Introduce new CLI option `--skip-create-db` into `migrate deploy` that
would avoid creating a new database.

Context: in my case a database is provisioned by a another tool -
terraform. Running `migrate deploy` as part of the application startup
leads to race conditions with terraform and unpredicted deployments.

### DIFF
--- a/packages/migrate/src/commands/MigrateDeploy.ts
+++ b/packages/migrate/src/commands/MigrateDeploy.ts
@@ -27,8 +27,9 @@ ${chalk.bold('Usage')}
 
 ${chalk.bold('Options')}
 
-  -h, --help   Display this help message
-    --schema   Custom path to your Prisma schema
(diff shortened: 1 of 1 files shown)


andersonbloughk:
👍

tomhoule:
We are going to have a discussion about migrate DX and what commands _should_ do next week, this PR will definitely be valuable input.

Jolg42:
Thanks for your PR @maxkomarychev and sorry for the delay.

We decided that we would not merge this for now as we do not fully understand why it's needed in the first place.

What could be helpful is if you could create an issue about this problem/workaround (or lack of), it will also make it easier for other users to manifest their shared interest.

We're happy to revisit this decision in the future if needed.

PR #9893, opened 2021-10-21
chore(client): data proxy improvements

millsp:
* Fixes an issue with webpack5/esbuild (path used for NCCs)
* Improves cold starts considerably (near but not below 1.5s)
* Integrates Jakub's PR for better errors (and adds error codes)
* Simplifies the instantiation logic (originally done with promises)
* Moves the sha/crypto stuff to only happen at generation time
* Makes errors to be thrown in the same way as other engines
* fixes https://github.com/prisma/prisma/issues/9594

But this PR also comes with:
- ~~[+ docs PR](https://github.com/prisma/docs/pull/2481)~~
- [+ E2Es PR](https://github.com/prisma/e2e-tests/pull/2141)

PR #10085, opened 2021-11-03
feat: better errors for DataProxyEngine

ghost:


millsp:
Thanks for your contrib. I've merged it here https://github.com/prisma/prisma/pull/9893.

matthewmueller:
Really cool to see how easy it is to extend Errors in JS these days!

PR #10108, opened 2021-11-04
feat(cli): show engine type when running "prisma generate"

### LINKED ISSUE #9958: Know when the Client is using the data proxy
## Problem

It’s hard to understand/confirm if an app is using the binary or the proxy version of Prisma Client. 

## Suggested solution

**At build time:**

```sh
$ PRISMA_CLIENT_ENGINE_TYPE=dataproxy npx prisma generate
✔ Generated Prisma Client (3.3.0 | dataproxy) to ./node_modules/@prisma/client in 2.00s 
```

**At runtime:**

It would be nice to add a bit of DEBUG logging at runtime as well to see whether or not you're running queries through the data proxy.

aqrln:
Closes #9958


packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -147,4 +136,12 @@ function cleanSnapshot(str: string): string {
     .replace(/\d+ms/g, 'XXms')
     .replace(/\d+s/g, 'XXs')
     .replace(/\(version:.+\)/g, '(version: 0.0.0)')
+    .replace(new RegExp(getClientEngineType(), 'g'), 'TEST_ENGINE_TYPE')
+}
+
+function replaceEngineType(message: string | Error) {
+  if (message instanceof Error) {
+    throw message
+  }
```
Jolg42, line 145: Why a throw here? 

I expected only a replace 🤔

packages/client/src/__tests__/binaryEngine.test.ts
```diff
@@ -1,6 +1,6 @@
 import { BinaryEngine } from '@prisma/engine-core'
+import { ClientEngineType, getClientEngineType } from '@prisma/sdk'
 import path from 'path'
-import { ClientEngineType, getClientEngineType } from '../runtime/utils/getClientEngineType'
```
Jolg42, line 4: What prompted the import changes? Just curious

Jolg42 (changes requested):
I think `engineType` output should be optional as community generators might not need it and would probably look like this (like before)
```
✔ Generated MyCommunityGenerator (0.0.0) to ./mydir/path in XXms
```

packages/client/src/__tests__/binaryEngine.test.ts
aqrln: I needed to use `getClientEngineType` in CLI, and importing it like

```
import { getClientEngineType } from '@prisma/client/src/runtime/utils/getClientEngineType'
```

didn't feel right (even though it technically worked).

packages/cli/src/__tests__/commands/Generate.test.ts
aqrln: `Generate.parse` returns `Promise<string | Error>` and I needed to handle it somehow. I can do something like

```ts
if (message instanceof Error) {
  message = message.stack
}
```

here instead. I think it will be closer to the previous behavior, too (i.e., when `string | Error` was unconditionally matched with the snapshot).

Jolg42: Oh I see 👍🏼 

But we don't really need to handle it since where `replaceEngineType` it's never an error?

I'm ok keeping it if you think it's useful.

packages/client/src/__tests__/binaryEngine.test.ts
Jolg42: Trying to think if there was a reason it was like this 🤔

aqrln: It wasn't like that, this file was only used inside `packages/client` via relative imports. My usage was the first one outside of `@prisma/client`, so I had to move it to sdk (or do the hacky import directly from `@prisma/client` that only worked at run time because we bundle things).

packages/cli/src/__tests__/commands/Generate.test.ts
aqrln: > But we don't really need to handle it

Well, yeah, I didn't think about it, but we can make this function return `string | Error` too and just return the error here as is.

millsp (approved):
Well done!

aqrln:
@Jolg42 I pushed changes to make the engine type output opt-in for generators via the manifest, and enable it for the Prisma Client generator. Is this a good approach?

If yes, optimally this needs new tests and, I guess, updating https://github.com/prisma/specs/tree/master/generators.

aqrln:
Next steps:

* Revert the manifest changes, this is not the right approach. Even though conceptually something like this makes sense for any client generators (JS, Go etc.), the way it was implemented it would only work correctly for `prisma-client-js`, because for other clients the detection logic and available engines would differ from what `getClientEngineType()` returns.
* Hardcode the special format for `prisma-client-js` for now.
  * A general solution can be implemented later if it becomes necessary. Since the generator itself would need to control what to print there, it would probably involve adding a new JSON-RPC method.
    * Arguably, this might have been beneficial even for `prisma-client-js`, because it would have made it possible to preserve code locality and keep `getClientEngineType()` inside `packages/client` and avoid moving it to `packages/sdk`.
    * At the same time, it would add too much complexity than it's worth at this point.
* Update the message for `prisma migrate` as well.
  * Decouple the duplicate code and reuse it in both commands.

Jolg42 (approved):
Well done 🙌🏼 

Nice and clean and removing duplicated code along the way 💯

PR #10155, opened 2021-11-08
fix: do not send content-length to Data Proxy

### LINKED ISSUE #10152: Data Proxy client fails when multi-byte chars are used
### Bug description

Using Prisma client with Data Proxy and multi-byte characters doesn't work.

### How to reproduce

Generate a client so it uses Data Proxy:

```
# generation
PRISMA_CLIENT_ENGINE_TYPE='dataproxy' npx prisma generate

# schema file, in generator section
previewFeatures = ["dataProxy"]
```

Attempt to use multi-byte characters:

```
await prisma.city.create({
  data: {
    name: 'Łódź',
    countryCode: 'PL'
  }
})
```

The call fails with `GraphQL request failed` after 5 retries. Under the hood (log `info` and `query` events) HTTP calls to `/graphql` endpoint of a remote query engine respond with `422 Unprocessable Entity`.

### Expected behavior

Row should be inserted.

### Prisma information

See "How to reproduce". Schema has the Data Proxy feature flag enabled.

### Environment & setup

- OS: Ubuntu LTS
- Database: Postgres
- Node.js version: 16

### Prisma Version

```
prisma                  : 3.4.1
@prisma/client          : 3.4.1
Query Engine (Node-API) : libquery-engine 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : 57771c0558568c7d08bd34c7248af5244ae16bd9
Studio                  : 0.438.0
```

ghost:
Fixes #10152 

Query engine does not require `content-length` request header, but it is verified against request body when provided. Currently the client provides character length but it should provide byte length or skip the header. We decided to do the latter since it works completely fine and is less error prone.

### DIFF
--- a/packages/engine-core/src/data-proxy/utils/request.ts
+++ b/packages/engine-core/src/data-proxy/utils/request.ts
@@ -40,7 +40,6 @@ function buildHeaders(options: RequestOptions): RequestOptions['headers'] {
   return {
     ...options.headers,
     'Content-Type': 'application/json',
-    'Content-Length': `${options.body?.length ?? 0}`,
   }
 }
 


Jolg42 (approved):
👍

Jolg42 (approved):

PR #10157, opened 2021-11-08
fix: Remove upcasting with default generic values on clients

andyrichardson:
# About

Remove unnecessary up-casting to maintain TS language server suggestions.

<details>
<summary>With upcasting (before)</summary>

Screenshot from 2021-11-08 12-09-23

</details>

<details>
<summary>Without upcasting (after)</summary>

Screenshot from 2021-11-08 12-07-56

</details>

Seems like one of those TS nuances but as far as I can tell, there's no need for a default value

### DIFF
--- a/packages/client/src/generation/TSClient/Count.ts
+++ b/packages/client/src/generation/TSClient/Count.ts
@@ -219,7 +219,7 @@ ${indent(
     .map((f) => {
       const fieldTypeName = (f.outputType.type as DMMF.OutputType).name
       return `
-${f.name}<T extends ${getFieldArgName(f)} = {}>(args?: Subset<T, ${getFieldArgName(f)}>): ${getReturnType({
+${f.name}<T extends ${getFieldArgName(f)}>(args?: Subset<T, ${getFieldArgName(f)}>): ${getReturnType({
         name: fieldTypeName,
         actionName: f.outputType.isList ? DMMF.ModelAction.findMany : DMMF.ModelAction.findUnique,
         hideCondition: false,
--- a/packages/client/src/__tests__/types/chaining/index.d.ts
+++ b/packages/client/src/__tests__/types/chaining/index.d.ts
@@ -1,0 +1,1 @@
+export { PrismaClient, User } from '@prisma/client'
--- a/packages/client/src/__tests__/types/chaining/index.test-d.ts
+++ b/packages/client/src/__tests__/types/chaining/index.test-d.ts
@@ -1,0 +1,24 @@
+import { PrismaClient, User } from '.'
+import { expectError } from 'tsd'
+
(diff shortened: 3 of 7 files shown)


aqrln:
Hi @andyrichardson and thank you for your contribution 🚀

I'll take a look in more details later, but on the first glance this looks good to me.

packages/client/src/generation/TSClient/Count.ts
```diff
@@ -219,7 +219,7 @@ ${indent(
     .map((f) => {
       const fieldTypeName = (f.outputType.type as DMMF.OutputType).name
       return `
-${f.name}<T extends ${getFieldArgName(f)} = {}>(args?: Subset<T, ${getFieldArgName(f)}>): ${getSelectReturnType({
+${f.name}<T extends ${getFieldArgName(f)}>(args?: Subset<T, ${getFieldArgName(f)}>): ${getSelectReturnType({
```
aqrln, line 222: Since `args` is an optional parameter, will TypeScript be able to infer anything when the argument is not passed to the function?

aqrln (commented):
Hi @andyrichardson and thanks again for your contribution. I'm sorry it took so long, I was going to take a closer look previously, but ultimately I lost track of it.

packages/client/src/generation/TSClient/Count.ts
```diff
@@ -219,7 +219,7 @@ ${indent(
...
         name: fieldTypeName,
         actionName: f.outputType.isList ? DMMF.ModelAction.findMany : DMMF.ModelAction.findUnique,
```
aqrln, line 224: Another thing to consider is that in both cases `T` is present in the types generated by `getSelectReturnType`. It's worth checking if there's any impact.

packages/client/src/generation/TSClient/Count.ts
aqrln: Another option might have been using `${getFieldArgName(f)}` instead of `{}` as the default value to get all properties by default, but that would probably break the `CheckSelect` check in the return type generated by `getSelectReturnType`.

aqrln: > Since args is an optional parameter, will TypeScript be able to infer anything when the argument is not passed to the function?

Actually it will infer it to the type in the `extends` clause (i.e., `${getFieldArgName(f)}` in this case), yeah.

aqrln:
@millsp I rebased this and resolved the conflicts, and added some smoke tests with tsd, PTAL

millsp (approved):

andyrichardson:
Thanks for taking the time to review this and get it in folks :pray:

Keep up the dope work!

PR #10178, opened 2021-11-09
ci: Document and clarify around `prisma.prismaCommit`

janpio:
When doing `3.4.2` I double checked what the release process code does with `RELEASE_PROMOTE_DEV`, and stumbled over this piece of the code that did not make instant sense to me. Added some documentation, and also renamed a function and changed its signature slightly to make it easier to understand.

### DIFF
--- a/scripts/setup.ts
+++ b/scripts/setup.ts
@@ -27,6 +27,8 @@ has to point to the dev version you want to promote, for example 2.1.0-dev.123`)
     await run('.', `git config --global user.name "prisma-bot"`)
   }
   if (process.env.RELEASE_PROMOTE_DEV) {
+    // Checkout exactly the commit of the `dev` version that is being promoted/released
+    
     const prismaCommit = await getPrismaCommitFromPackageJsonViaUnpkg(process.env.RELEASE_PROMOTE_DEV)
     // TODO: disable the dry run here // TODO 2: What does this mean?
 
--- a/scripts/setup.ts
+++ b/scripts/setup.ts
@@ -27,11 +27,11 @@ has to point to the dev version you want to promote, for example 2.1.0-dev.123`)
     await run('.', `git config --global user.name "prisma-bot"`)
   }
   if (process.env.RELEASE_PROMOTE_DEV) {
-    const versions = await getVersionHashes(process.env.RELEASE_PROMOTE_DEV)
-    // TODO: disable the dry run here
+    const prismaCommit = await getPrismaCommitFromPackageJsonViaUnpkg(process.env.RELEASE_PROMOTE_DEV)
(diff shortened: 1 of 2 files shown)


scripts/setup.ts
```diff
@@ -209,7 +209,7 @@ async function branchExists(dir: string, branch: string): Promise<boolean> {
   return exists
 }
 
-async function getVersionHashes(
+async function getPrismaCommitFromPackageJsonViaUnpkg(
```
janpio, line 212: This rename (and signature change) might be controversial, but this is weird enough that I felt it was useful to surface this is via Pkg and essentially only to get the `prismaCommit` value.

scripts/setup.ts
```diff
@@ -27,11 +27,13 @@ has to point to the dev version you want to promote, for example 2.1.0-dev.123`)
     await run('.', `git config --global user.name "prisma-bot"`)
   }
   if (process.env.RELEASE_PROMOTE_DEV) {
-    const versions = await getVersionHashes(process.env.RELEASE_PROMOTE_DEV)
-    // TODO: disable the dry run here
+    // Checkout exactly the commit of the `dev` version that is being promoted/released
```
janpio, line 30: If/when we start to tag all `dev` versions, this whole complicated thing could be replace by looking up the commit of the tag instead of going via Unpkg and `prisma.prismaCommit`, correct?

scripts/setup.ts
```diff
@@ -27,11 +27,13 @@ has to point to the dev version you want to promote, for example 2.1.0-dev.123`)
...
+    
+    const prismaCommit = await getPrismaCommitFromPackageJsonViaUnpkg(process.env.RELEASE_PROMOTE_DEV)
+    // TODO: disable the dry run here // TODO 2: What does this mean?
```
Jolg42, line 33: 😅

scripts/setup.ts
Jolg42: The function name is much more accurate 👍🏼 
And yes it's a weird part of the release code

scripts/setup.ts
Jolg42: It's also my thinking, and that would be a good thing

Jolg42 (approved):

scripts/setup.ts
janpio: Honestly, do you know?

Jolg42: I think it's about this line
```
await run(`.`, `git checkout ${prismaCommit}`, true)
```

Where true here means dryRun = true

Jolg42: Which means this is never executed? I don't like this

PR #10235, opened 2021-11-12
chore: add docker-compose file for ARM64 machines

aqrln:


millsp:
Are postgres and mongo going to work?

Not sure what this is worth or if it works:
https://hub.docker.com/r/arm64v8/mongo/
https://hub.docker.com/r/arm64v8/postgres/

aqrln:
@millsp everything else besides MySQL and MS SQL supports ARM out-of the box with official images

docker/README.md
```diff
@@ -16,9 +16,14 @@ If they are already used make sure to change the ports like this
...
 ports:
-  - '3307:3306 # only change the first number
+  - '3307:3306' # only change the first number
 ```
 
+## Apple M1 and other ARM64 CPUs
+
+Docker images for MySQL and Microsoft SQL Server only support x86_64 CPUs.
+Copy or link `docker-compose.arm64.yml` to `docker-compose.override.yml` to use ARMv8-compatible drop-in replacements (MariaDB and Azure SQL Edge) instead.
+
```
Jolg42, line 26:
```suggestion
By default, Compose reads two files, a `docker-compose.yml` and an optional `docker-compose.override.yml` file. By convention, the `docker-compose.yml` contains your base configuration. The override file, as its name implies, can contain configuration overrides for existing services or entirely new services.
```

Copy pasted from docs, didn't know this

Jolg42 (approved):
Nice! Thanks!

PR #10295, opened 2021-11-16
fix(client): iTX - disable pipelining for http pool for binary engine

### LINKED ISSUE #8707: interactiveTransactions: 2 concurrent writes to the same row will cause it to hang until expiring
### Bug description

I'm trying to use interactive transactions to build a bank transfer. I'd like to try to achieve the following:

- Alice and Bob have $100
- Concurrently, Alice sends Bob $100 twice
  - One of those requests goes through
  - The other one is rejected saying there's not enough money  
- Alice has $0, Bob has $100

### How to reproduce

**Reproduction Repo:** https://github.com/matthewmueller/interactive-transactions 

Given the following Prisma Schema

```prisma

datasource db {
  provider = "postgresql"
}

generator client {
}

model Account {
  id      Int    @id @default(autoincrement())
  email   String @unique
  balance Int
}
```

And the following script:

```ts
import { PrismaClient } from "@prisma/client"

async function unseed() {
  await prisma.account.deleteMany({
    where: {
      OR: [{ email: "alice@prisma.io" }, { email: "bob@prisma.io" }],
    },
  })
}

async function seed() {
  await prisma.account.create({
    data: {
      email: "alice@prisma.io",
      balance: 100,
    },
  })
  await prisma.account.create({
    data: {
      email: "bob@prisma.io",
      balance: 100,
    },
  })
}

async function transfer(nth: number, from: string, to: string, amount: number) {
  return await prisma.$transaction(
    async (prisma) => {
      console.time("send " + nth)
      const sender = await prisma.account.update({
        data: {
          balance: {
            decrement: amount,
          },
        },
        where: {
          email: from,
        },
      })
      console.timeEnd("send " + nth)
      console.time("throw " + nth)
      if (sender.balance < 0) {
        throw new Error(`${from} doesn't have enough to send ${amount}`)
      }
      console.timeEnd("throw " + nth)
      console.time("recieve " + nth)
      const recipient = prisma.account.update({
        data: {
          balance: {
            increment: amount,
          },
        },
        where: {
          email: to,
        },
      })
      console.timeEnd("recieve " + nth)
      return recipient
    },
    {
      timeout: 20000,
    }
  )
}

async function main() {
  await prisma.$connect()
  await unseed()
  await seed()
  console.time("transfer")
  await Promise.all([
    transfer(1, "alice@prisma.io", "bob@prisma.io", 100),
    transfer(2, "alice@prisma.io", "bob@prisma.io", 100),
  ])
  console.timeEnd("transfer")
}

main()
  .catch(console.error)
  .finally(() => prisma.$disconnect())
```

If you setup a database and then run `ts-node index.ts`, you'll get:

```
send 1: 7.005ms
throw 1: 0.005ms
recieve 1: 0.279ms
send 2: 20.005s
throw 2: 0.367ms
recieve 2: 0.644ms
PrismaClientKnownRequestError3 [PrismaClientKnownRequestError]: 
Invalid `prisma.account.deleteMany()` invocation in
/Users/m/dev/src/github.com/prisma/interactive-transactions/index.ts:5:58

  2 const prisma = new PrismaClient()
  3 
  4 async function unseed() {
→ 5   await prisma.account.deleteMany(
  Transaction API error: Transaction already closed: Transaction is no longer valid. Last state: 'Expired'.
    at RequestHandler.request (/Users/m/dev/src/github.com/prisma/interactive-transactions/node_modules/@prisma/client/runtime/index.js:36361:15)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at PrismaClient._transactionWithCallback (/Users/m/dev/src/github.com/prisma/interactive-transactions/node_modules/@prisma/client/runtime/index.js:36932:18) {
  code: 'P2028',
  clientVersion: '2.30.0-dev.8',
  meta: {
    error: "Transaction already closed: Transaction is no longer valid. Last state: 'Expired'."
  }
}
```

Two notes:
- The 2nd send takes the full 20s
- An FYI that the stack trace is off. It was pointing to that line even when it was commented out.

I think this could be a bug because if I change the code to transfer serially, it works as expected:

```typescript
async function main() {
  await prisma.$connect()
  await unseed()
  await seed()
  await transfer("alice@prisma.io", "bob@prisma.io", 100)
  await transfer("alice@prisma.io", "bob@prisma.io", 100) // Error: alice@prisma.io doesn't have enough to send 100
}
```

### Expected behavior

I'd expect this to work as expected, you should be able to initialize multiple transactions concurrently and let the database sort it out.

### Environment & setup

- OS:  OSX
- Database:  Postgres
- Node.js version:  v14.16.0

### Prisma Version

```
prisma                : 2.30.0-dev.8
@prisma/client        : 2.30.0-dev.8
Current platform      : darwin
Query Engine (Binary) : query-engine 71d96e8bbd21982078694f00add0f51da2056a8b (at node_modules/@prisma/engines/query-engine-darwin)
Migration Engine      : migration-engine-cli 71d96e8bbd21982078694f00add0f51da2056a8b (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine  : introspection-core 71d96e8bbd21982078694f00add0f51da2056a8b (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary         : prisma-fmt 71d96e8bbd21982078694f00add0f51da2056a8b (at node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash  : 71d96e8bbd21982078694f00add0f51da2056a8b
Studio                : 0.419.0
Preview Features      : interactiveTransactions
```

millsp:
fixes https://github.com/prisma/prisma/issues/8707

### DIFF
--- a/packages/engine-core/src/binary/Connection.ts
+++ b/packages/engine-core/src/binary/Connection.ts
@@ -24,7 +24,6 @@ function assertHasPool<A>(pool: A): asserts pool is NonNullable<A> {
  */
 export class Connection {
   private _pool: Pool | undefined
-  url!: string | URL
 
   constructor() {}
 
@@ -54,8 +53,6 @@ export class Connection {
   open(url: string | URL, options?: Pool.Options) {
     if (this._pool) return
 
-    this.url = url
-
     this._pool = new Pool(url, {
       connections: 1000,
       keepAliveMaxTimeout: 600e3,
@@ -80,29 +77,14 @@ export class Connection {
   ) {
     assertHasPool(this._pool)
 
-    const headers2 = {
-      'Content-Type': 'application/json',
-      ...headers,
-    }
-
-    function makeCurlHeaders(headers) {
-      const keys = Object.keys(headers)
-
-      return keys.reduce((acc, key) => {
-        return `${acc} --header '${key}: ${headers[key]}'`
-      }, '')
-    }
-
-    console.log(
-      `curl --location --request ${method} '${this.url}${endpoint}'${makeCurlHeaders(
-        headers2,
-      )} --data-raw ${JSON.stringify(body)}`,
-    )
     const response = await this._pool.request({
       path: endpoint,
       method: method,
-      headers: headers,
-      body: body,
+      headers: {
+        'Content-Type': 'application/json',
+        ...headers,
+      },
+      body,
       bodyTimeout: 0,
     })
 
--- a/packages/engine-core/src/binary/Connection.ts
+++ b/packages/engine-core/src/binary/Connection.ts
@@ -30,7 +30,7 @@ export class Connection {
 
   /**
    * Wrapper to handle HTTP error codes. HTTP errors don't trigger any
-   * execptions because it is optional to handle error status codes.
+   * exceptions because it is optional to handle error status codes.
    * @param response to handle
    * @param handler to execute
    * @returns
@@ -57,8 +57,7 @@ export class Connection {
     this.url = url
 
     this._pool = new Pool(url, {
-      connections: 100,
-      pipelining: 10,
+      connections: 1000,
       keepAliveMaxTimeout: 600e3,
       headersTimeout: 0,
       ...options,
--- a/packages/engine-core/src/binary/Connection.ts
+++ b/packages/engine-core/src/binary/Connection.ts
@@ -24,6 +24,7 @@ function assertHasPool<A>(pool: A): asserts pool is NonNullable<A> {
  */
 export class Connection {
   private _pool: Pool | undefined
+  url!: string | URL
 
   constructor() {}
 
@@ -53,6 +54,8 @@ export class Connection {
   open(url: string | URL, options?: Pool.Options) {
     if (this._pool) return
 
+    this.url = url
+
     this._pool = new Pool(url, {
       connections: 100,
       pipelining: 10,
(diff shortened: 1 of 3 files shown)


packages/engine-core/src/binary/Connection.ts
```diff
@@ -54,8 +54,7 @@ export class Connection {
     if (this._pool) return
 
     this._pool = new Pool(url, {
-      connections: 100,
```
millsp, line 57: Before:
* We instantiated up to 100 Clients
* With each socket sending up to 10 requests

packages/engine-core/src/binary/Connection.ts
```diff
@@ -54,8 +54,7 @@ export class Connection {
...
-      pipelining: 10,
+      connections: 1000,
```
millsp, line 57: After:
* We instantiated up to 1000 Clients
* With each socket sending up to 1 requests

packages/engine-core/src/binary/Connection.ts
```diff
@@ -54,8 +54,7 @@ export class Connection {
     if (this._pool) return
 
     this._pool = new Pool(url, {
-      connections: 100,
-      pipelining: 10,
```
millsp, line 58: The reason for this change is that we cannot pipeline request. In other words, we cannot make requests serial. Why? Because if you have - for instance - a transaction A that depends on B, transaction A will never finish if sent in the wrong order (as the http client waits for a response from the QE).

packages/engine-core/src/binary/Connection.ts
millsp: (discussed with Dom, not an issue)

packages/engine-core/src/binary/Connection.ts
millsp, line 58: The reason for this change is that we cannot pipeline requests. In other words, we cannot make requests serial. Why? Because if you have - for instance - a transaction A that blocks on B, transaction B will never finish (as the http client waits for a response from the QE on transaction A).

Jolg42 (approved):
Nice catch!

aqrln (approved):
LGTM 👍

aqrln:
Restarted the CI because of an unrelated failure (generator test timed out).

```
  ● using cli › should work with a custom generator
726

727
    thrown: "Exceeded timeout of 30000 ms for a test.
728
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
729

730
      26 |   })
731
      27 |
732
    > 28 |   it('should work with a custom generator', async () => {
733
         |   ^
734
      29 |     ctx.fixture('custom-generator')
735
      30 |     const data = await ctx.cli('generate')
736
      31 |
737

738
      at src/__tests__/commands/Generate.test.ts:28:3
739
      at Object.<anonymous> (src/__tests__/commands/Generate.test.ts:7:1)
740
```

Jolg42:
@aqrln not sure why but it's extra flaky these last days

aqrln:
@Jolg42 yeah, not sure if I've seen exactly this one or some other tests failing with timeout like that lately, but the error looks very familiar.

Issue #1750, opened 2020-03-03
`prisma2 generate` output message always says `You can now start using Prisma Client in your code:` with example


Jolg42:
Is this issue a duplicate? I didn't find any.

So here the output is 
<img width="453" alt="Screen Shot 2020-03-03 at 10 56 26" src="https://user-images.githubusercontent.com/1328733/75764215-d53aa500-5d3d-11ea-8b8d-b4b119b797bd.png">

This makes sense on the first run or if you didn't use Prisma Client yet.
After that it's a really noisy message.

Jolg42:
Related #1643

janpio:
Note: Wait with implementation until https://github.com/prisma/specs/issues/427 is decided please.

janpio:
Related: https://github.com/prisma/prisma/issues/3390

janpio:
Duplicate of #3390

PR #10368, opened 2021-11-21
feat(cli): add --quiet/-q option to prisma generate

klausbadelt:
`prisma generate` always outputs "You can now start using Prisma Client in your code...". This gets noisy after some initial use.

This PR adds a `--quiet` and `-q` option to reduce verbosity:
```
$ prisma generate --quiet

✔ Generated Prisma Client (3.4.1) to ./node_modules/@prisma/client in 214ms
```

This addresses #3390 and #1750 in a rather simple way and following an established pattern - many CLIs have a `--quiet` option to reduce verbosity.


matthewmueller:
I'm +1 to this PR. What do you think about adding `--quiet` globally @Jolg42 ?

Jolg42:
I'm not sure if we need `--quiet` in other places at the moment, it sounds ok to me to have it for `generate` only like done here.

@matthewmueller What do you mean exactly by 
> adding `--quiet` globally

matthewmueller:
Sounds good! 

What I meant was that I wonder if `prisma db pull` for example would benefit from a `--quiet` option too. If we add it for others, does it make sense to define it more globally. That would affect this PR.

Either way, I agree that this is a good first start and I think we should merge this one :-)

matthewmueller (approved):

Jolg42:
Note: I'll have a discussion with migration team and then decide what we want to do. 

Thanks for the PR, makes sense to me!

Jolg42:
Feedback from @millsp and @aqrln 

Looks good to merge, the change is relatively simple. Although I would have preferred something more granular like --output none/warn/error/all

I see, I’m not quite sure that granularity works here
because --output=none would be what you want but then output should be empty? 

True, I'm just confused about what we think will be hidden by the --quiet option. Is it just to hide that specific message? What else could be hidden in the future?

the quiet here is just about that message that is only useful when you start using Prisma
but I could see maybe for some migrate commands it could change the output to make it less verbose or so

Because if that message is the issue, we could as well display that on install only, or on prisma init, or only on the first generate. I'd find more value to a --quiet option if it did more than just to hide a message

If it’s on install then yarn hides all the messages
the first generate is maybe too little? but then we need to keep track of the runs somewhere
at the same time first generate might be a good option to consider (no need to track just check if the output file exists)

I’m not opposed to this change, but I agree with Pierre that it’s kinda confusing and maybe --quiet is not the best name for this flag. My intuition would be that --quiet shouldn’t print anything at all unless there is an error.

PR #10380, opened 2021-11-22
fix(client): use proper typing for $disconnect

safareli:
Tested and `$disconnect` is returning `undefined`. 
0 match for this regex search: `= await [a-z.]*\$disconnect`.
It means result of $disconnect shuoldn't be `any` but `void`.


aqrln (approved):
I can confirm that the implementation of `$disconnect()` returns `this._engine.stop()`, and `Engine.stop()` indeed returns `Promise<void>`.

Thanks for the contribution, and nice catch!

Jolg42 (approved):
@safareli Thanks for the PR!

Note: we will merge this once the CLA is signed, see https://github.com/prisma/prisma/pull/10380#issuecomment-975458796

safareli:
Signed

PR #10526, opened 2021-12-01
chore(ci): run tests on multiple operating systems

### LINKED ISSUE #7725: Run (most) tests on all operating system
We should enable the `prisma/prisma` tests for all operating systems. Currently this is not possible, as the tests use databases that are started via Docker: https://github.com/prisma/prisma/pull/7721 (GH Actions can not run Linux containers on Windows, and the MacOS runners do not have Docker installed by default).

A possible solution could be to use `services` in our GH Actions setup again: https://docs.github.com/en/actions/guides/creating-postgresql-service-containers

aqrln:
Closes https://github.com/prisma/prisma/issues/7725


aqrln:
And, finally, everything is green!

image

macOS tests look flaky, so I'm not sure about those

* some tests were failing randomly occasionally and then just worked on the next run, and I couldn't reproduce those failures locally on my mac
* a bunch of tests required much higher timeouts

Currently they are passing, but they may turn out to be a headache later. Should we maybe extract macOS to a separate optional workflow that won't block the PRs? 

Windows tests seem to be pretty robust 🎉

aqrln:
Squashed the 90 commits into one and rebased onto the latest `main`. Let's see if it works now 🤞

.github/workflows/install-mysql.ps1
```diff
@@ -0,0 +1,7 @@
+Invoke-Expression (New-Object System.Net.WebClient).DownloadString('https://get.scoop.sh')
+
+scoop install mysql
```
janpio, line 3: Is this MySQL 5 or MySQL 8?

aqrln: It's MySQL 8

.github/workflows/test.yml
```diff
@@ -561,5 +559,185 @@ jobs:
...
+          name: generator-helper-${{ matrix.os }}
+
+  #
+  # Run all tests on macOS and Windows.
+  #
+  # Unlike the other jobs, this job doesn't use Docker (and thus skips some
+  # tests that require dependencies not easily installable without Docker —
+  # specifically, SQL Server), and runs tests on Node.js 12, 14 and 16 (Linux
+  # jobs only run on Node.js 12 and 16 because Node.js 14 is tested in
+  # Buildkite).
```
aqrln, line 572: Note to self: update this comment. Node.js 12 and 14 were dropped here as per discussion with @millsp and @Jolg42

.github/workflows/test.yml
```diff
@@ -561,5 +559,185 @@ jobs:
...
+          flags: generator-helper,${{ matrix.os }},library,binary
+          name: generator-helper-${{ matrix.os }}
+
+  #
+  # Run all tests on macOS and Windows.
+  #
+  # Unlike the other jobs, this job doesn't use Docker (and thus skips some
+  # tests that require dependencies not easily installable without Docker —
+  # specifically, SQL Server), and runs tests on Node.js 12, 14 and 16 (Linux
+  # jobs only run on Node.js 12 and 16 because Node.js 14 is tested in
```
janpio, line 571: Is running another Node version such high cost that this is worth the difference?
If not, maybe just run all the tests on all node versions.

.github/workflows/test.yml
```diff
@@ -561,5 +559,185 @@ jobs:
...
+  # Buildkite).
+  #
+  # It also runs the tests for different packages sequentially to prevent
+  # combinatorial explosion of the number of parallel jobs, and to minimize the
+  # number of times we need to install MySQL using the package manager
+  # (PostgreSQL and MongoDB are provided by GitHub Actions out of the box).
```
janpio, line 577: Is this correct re Postgres? We seem to run `setup-postgres.sh` below.

.github/workflows/test.yml
aqrln: Kind of. Mostly due to the number of jobs themselves and the time spent waiting for a free runner.

packages/client/src/__tests__/integration/happy/blog-env-mssql/test.ts
```diff
@@ -3,7 +3,9 @@ import { generateTestClient } from '../../../../utils/getTestClient'
 import type { SetupParams } from '../../../../utils/setupMSSQL'
 import { setupMSSQL } from '../../../../utils/setupMSSQL'
 
-describe('blog-env-mssql', () => {
+const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
```
janpio, line 6: Uh, no nicer/built-in way to do this in Jest?

packages/client/src/__tests__/integration/happy/chmod/test.ts
```diff
@@ -4,8 +4,10 @@ import fs from 'fs'
 import path from 'path'
 import { generateTestClient } from '../../../../utils/getTestClient'
 
+const testIf = (condition: boolean) => (condition ? test : test.skip)
+
 // Tests that no error is being thrown when the binary is manually set to chmod 644 because Client fixes that itself
-test('chmod', async () => {
+testIf(process.platform !== 'win32')('chmod', async () => {
```
janpio, line 10: Does this fail on Windows otherwise?

.github/workflows/test.yml
aqrln: On the other hand, we may try merging our Linux jobs into a single one, similar to `no-docker`. Running the setup script takes more time than running tests themselves in all packages except `client` anyway.

packages/client/src/__tests__/integration/happy/signals/test.ts
```diff
@@ -4,9 +4,16 @@ import execa from 'execa'
 import path from 'path'
 import { EXIT_MESSAGE, READY_MESSAGE } from './__helpers__/constants'
 
+const testIf = (condition: boolean) => (condition ? test : test.skip)
+
+const userSignalsSupported = process.platform !== 'win32'
```
janpio, line 9: What is the story here?

.github/workflows/test.yml
aqrln: `setup-postgres.sh` only configures it, but doesn't install

packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -147,7 +150,18 @@ describe('sqlite', () => {
...
+  //
+  // TODO: Windows: test fails with:
+  //
+  // [Error: Failed to create a new migration directory.
+  //    0: migration_core::api::CreateMigration
+  //            with migration_name="xl556ba8iva0gd2qfoyk2fvifsysnq7c766sscsa18rwolofgwo6j1mwc4d5xhgmkfumr8ktberb1y177de7uxcd6v7l44b6fkhlwycl70lrxw0u7h6bdpuf595n046bp9ek87dk59o0nlruto403n7esdq6wgm3o5w425i7svaw557latsslakyjifkd1p21jwj1end" draft=false
+  //              at migration-engine\core\src\api.rs:94
+  // ]
+  //
+  // Probably the file name is too long for Windows?
```
janpio, line 162: That could become a fun bug report - solution is probably to make sure we truncate or make sure string is shorter.

packages/migrate/src/__tests__/MigrateReset.test.ts
```diff
@@ -1,6 +1,9 @@
 process.env.PRISMA_MIGRATE_SKIP_GENERATE = '1'
 process.env.GITHUB_ACTIONS = '1'
 
+// TODO: Windows: some snapshot tests fail on Windows because of emoji.
+const testIf = (condition: boolean) => (condition ? test : test.skip)
```
janpio, line 5: Can't we filter these out from the snapshots? Might be as dumb as really filtering out the x ones we use hard coded.

.github/workflows/test.yml
aqrln: With MySQL and MongoDB we rely on the defaults to keep things simple, but with Postgres it's a little bit more complicated, because there are no good/documented defaults on the macOS runner.

janpio (approved):

packages/client/src/__tests__/integration/happy/blog-env-mssql/test.ts
aqrln: Not that I know of, unfortunately 😞
I would be very glad if someone has better suggestions, but I'm not too optimistic about that since I basically repeated the `testIf` pattern which was already used in the codebase.

packages/client/src/__tests__/integration/happy/chmod/test.ts
aqrln: It failed for unrelated reason (which I can fix just to see what happens, but I didn't try that). I disabled the test instead because it doesn't make much sense on Windows, where there's no executable flag.

Also, the functionality that is being tested here is a no-op on Windows: https://github.com/prisma/prisma/blob/main/packages/engine-core/src/binary/BinaryEngine.ts#L437-L439

packages/client/src/__tests__/integration/happy/signals/test.ts
aqrln: SIGUSR1 and SIGUSR2 don't exist on Windows.

packages/migrate/src/__tests__/MigrateReset.test.ts
aqrln: Yes, we can remove emoji from all snapshots and filter them out in the snapshot serializer to make the tests ignore them on Linux and macOS. Not the other way unfortunately. Should we do that?

janpio: I think that should be fine.

.github/workflows/test.yml
aqrln: Also heads up that this comment is not fully true since we decided to only run Windows and macOS tests with Node.js 16, at least for now: https://github.com/prisma/prisma/pull/10526#discussion_r770799470. It would make sense to expand the matrix if we see that the CI doesn't take too much time though.

aqrln:
I wonder how can this test be non-deterministic and just spontaneously swapping two lines compared to a snapshot on `binary, ubuntu-latest, 12`. I've never seen it fail before on Linux previously, but saw either this or a similar test failing on macOS in a similar way once (except the "Applying migration" line was moved even further down).

```
FAIL src/__tests__/MigrateDeploy.test.ts
  common
    ✓ should fail if no schema file (3 ms)
    ✓ should fail if experimental flag (1 ms)
    ✓ should fail if early access flag (1 ms)
  sqlite
    ✓ no unapplied migrations (341 ms)
    ✕ 1 unapplied migration (577 ms)
    ✓ should throw if database is not empty (263 ms)

  ● sqlite › 1 unapplied migration

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `sqlite 1 unapplied migration 3`

    - Snapshot  - 1
    + Received  + 1

    @@ -3,12 +3,12 @@

      SQLite database dev.db created at file:dev.db

      1 migration found in prisma/migrations

    +
      Applying migration `20201231000000_init`
    -
      Datasource "my_db": SQLite database "dev.db" at "file:dev.db"

      1 migration found in prisma/migrations
      ↵

      79 |
      80 |     expect(ctx.mocked['console.info'].mock.calls.join('\n'))
    > 81 |       .toMatchInlineSnapshot(`
         |        ^
      82 |       Prisma schema loaded from prisma/schema.prisma
      83 |       Datasource "my_db": SQLite database "dev.db" at "file:dev.db"
      84 |

      at Object.<anonymous> (src/__tests__/MigrateDeploy.test.ts:81:8)

 › 1 snapshot failed.
```

Jolg42:
@aqrln likely because this is a log line coming from the engine and not the CLI so the timing can vary slightly

.github/workflows/test.yml
```diff
@@ -561,5 +559,184 @@ jobs:
...
+          node-version: ${{ matrix.node }}
+
+      - run: bash .github/workflows/setup.sh
+        env:
+          CI: true
+          SKIP_GIT: true
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+
+      - name: Test packages/sdk
+        if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') }}
```
millsp, line 632:
```suggestion
        if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') }} || contains(needs.detect_jobs_to_run.outputs.jobs, '-sdk-') }}
```

.github/workflows/test.yml
```diff
@@ -561,5 +559,184 @@ jobs:
...
+        run: pnpm run test
+        working-directory: packages/sdk
+        env:
+          CI: true
+          SKIP_GIT: true
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+
+      - uses: codecov/codecov-action@v2
+        if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') }}
```
millsp, line 641:
```suggestion
        if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') }} || contains(needs.detect_jobs_to_run.outputs.jobs, '-sdk-') }}
```

millsp (changes requested):

.github/workflows/test.yml
aqrln: FWIW, `detect-jobs-to-run.js` never produces `-sdk-`, it will output `-all-` in that case. Still a good idea to include the check here to future-proof it, yeah.

millsp (approved):
Awesome work :tada: :100:.
(PS: let's create an issue for the remaining todos)

packages/migrate/src/__tests__/MigrateReset.test.ts
aqrln: Resolving this comment as that would probably be out of scope of this PR for now. I'll create an issue for the `TODO: Windows` comments. This will be one of the lower-hanging fruits to fix among those.

aqrln:
Need to coordinate merging with https://github.com/prisma/prisma/pull/10772

PR #10546, opened 2021-12-02
test(cli): update engines version in Version.test.ts

aqrln:
Update custom engines version in the "prisma --version" test from an old one that didn't have a build for Apple M1 yet to the latest one to fix the test on ARM64 macOS.


Jolg42 (approved):
Indeed! Thanks!

PR #10551, opened 2021-12-03
fix(studio): Remove schemaHash from test to see if it unblocks Studio tests on Windows

janpio:
Up to @aqrln to decide if this is the right solution or to calculate the hash live in the script.


aqrln:
Thanks!

aqrln:
I believe both behaviors should be tested, and two new tests where we will be sending valid and invalid hashes should be added separately.

PR #10568, opened 2021-12-03
fix(init): Only prevent `init` if existing `prisma` is a folder

janpio:
Currently prevents usage of `prisma init` if there is a binary (e.g. a packages `prisma` CLI via `pkg`) with name `prisma`.

PS: Seems we are currently missing all tests for this functionality of `init` in https://github.com/prisma/prisma/blob/main/packages/cli/src/__tests__/commands/Init.test.ts

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -156,7 +156,7 @@ export class Init implements Command {
       process.exit(1)
     }
 
-    if (fs.existsSync(prismaFolder)) {
+    if (fs.existsSync(prismaFolder) && fs.lstatSync(prismaFolder).isDirectory()) {
       console.log(
         printError(`A folder called ${chalk.bold('prisma')} already exists in your project.
         Please try again in a project that is not yet using Prisma.


packages/cli/src/Init.ts
```diff
@@ -156,7 +156,7 @@ export class Init implements Command {
       process.exit(1)
     }
 
-    if (fs.existsSync(prismaFolder)) {
+    if (fs.existsSync(prismaFolder) && fs.lstatSync(prismaFolder).isDirectory()) {
```
millsp, line 159: Do we want to add an error message for when it's a file that already exists?

janpio: Hm, is that problematic? I was assuming that an existing file should not be a problem and as such should be skipped in the check as the folder should be able to be created then.

aqrln: It might be that some filesystems possibly allow that, but generally and on most platforms, yes, that would be problematic.

```
/tmp
❯ touch prisma

/tmp
❯ mkdir prisma
mkdir: prisma: File exists
```

janpio: Ok, that makes this PR useless indeed.

What we could do instead if be clearer in the error message if it is a file or folder - but that does not really justify a change.

PR #10262, opened 2021-11-15
fix(cli,client): remove DEBUG binding to dotenv enableDebug

Jolg42:
Closes https://github.com/prisma/prisma/issues/9428

PR #10590, opened 2021-12-06
feat(client,migrate): bind debug of dotenv to DOTENV_CONFIG_DEBUG env var

Jolg42:
We remove the binding to `DEBUG` with https://github.com/prisma/prisma/pull/10262

Thinking `DOTENV_CONFIG_DEBUG` would just work but

Seems that using the lib in the code uses the config function https://github.com/motdotla/dotenv/blob/master/lib/main.js#L82
Which needs to have `debug = true` to work

and using it preloaded uses “dotenv/config”
https://github.com/motdotla/dotenv/blob/master/config.js
`DOTENV_CONFIG_<OPTION>=value node -r dotenv/config your_script.js`
which loads the env vars

### DIFF
--- a/packages/sdk/src/utils/tryLoadEnvs.ts
+++ b/packages/sdk/src/utils/tryLoadEnvs.ts
@@ -129,7 +129,7 @@ export function loadEnv(envPath: string | null | undefined): DotenvLoadEnvResult
           //
           // Value needs to be null or undefined, false is truthy
           // https://github.com/motdotla/dotenv/blob/7301ac9be0b2c766f865bbe24280bf82586d25aa/lib/main.js#L89-L91
-          debug: process.env.DOTENV_CONFIG_DEBUG && process.env.DOTENV_CONFIG_DEBUG.length > 0 ? true : undefined,
+          debug: process.env.DOTENV_CONFIG_DEBUG ? true : undefined,
         }),
       ),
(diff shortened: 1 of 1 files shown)


millsp (approved):

packages/sdk/src/utils/tryLoadEnvs.ts
```diff
@@ -122,6 +122,14 @@ export function loadEnv(envPath: string | null | undefined): DotenvLoadEnvResult
...
         dotenv.config({
           path: envPath,
+          // Useful to debug dotenv parsing, prints errors & warnings
+          // Set to any value to enable
+          // Example for empty .env file
+          // [dotenv][DEBUG] did not match key and value when parsing line 1:
+          //
+          // Value needs to be null or undefined, false is truthy
+          // https://github.com/motdotla/dotenv/blob/7301ac9be0b2c766f865bbe24280bf82586d25aa/lib/main.js#L89-L91
+          debug: process.env.DOTENV_CONFIG_DEBUG && process.env.DOTENV_CONFIG_DEBUG.length > 0 ? true : undefined,
```
aqrln, line 132: Is the `process.env.DOTENV_CONFIG_DEBUG.length > 0` check necessary here? An empty string won't evaluate to true anyway.

aqrln (approved):

PR #10645, opened 2021-12-10
fix(client): add stacktrace when rejectOnNotFound is enabled

### LINKED ISSUE #10641: "No X found" error thrown from a findUnique query does not show a stack trace
### Bug description

Error object thrown by prisma has no stack trace and message.
Stack trace and message should be available on all errors in node.js
when I call findUnique with `rejectOnNotFound`

### How to reproduce

Have a code like this:
```ts
import { PrismaClient, Prisma } from '@prisma/client'

const prismaClient = new PrismaClient()


;(async () => {
    await prismaClient.user.findUnique({
        where: {
            id: 20 // make sure this does not exist in your DB
        },
        rejectOnNotFound: true
    })
})()
```

Running this code using `ts-node pr.ts` prints out:
```
[NotFoundError: No User found] { clientVersion: '3.6.0' }
```

### Expected behavior

Stack trace in the error points into the line in user script where query was executed from.
For example an Error thrown should show up in stdout like this:
```
Error: this is how you throw an error
    at /home/capaj/work-repos/test/pr.ts:7:11
    at Object.<anonymous> (/home/capaj/work-repos/test/pr.ts:9:3)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Module.m._compile (/home/capaj/.nvm/versions/node/v16.13.0/lib/node_modules/ts-node/src/index.ts:1371:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Object.require.extensions.<computed> [as .ts] (/home/capaj/.nvm/versions/node/v16.13.0/lib/node_modules/ts-node/src/index.ts:1374:12)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at main (/home/capaj/.nvm/versions/node/v16.13.0/lib/node_modules/ts-node/src/bin.ts:331:12)
```

### Prisma information
```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
  url      = "file:./bugRepro.db"
}

model User {
  email     String   @unique
  name      String?

  role  String @default("USER")
  posts Post[]
}
```
### Environment & setup

- OS: ubuntu
- Database: any
- Node.js version: 17

### Prisma Version

```

```

YassinEldeeb:
Fixes #10641
Closes prisma/docs#2616

Removing the `undefined` from `this.stack` in the custom Error class to allow for stack trace for better development in large code bases.

### Before:
Screenshot 2021-12-11 005248

### with this we can now trace where the error was thrown to debug more efficiently 💪
Screenshot 2021-12-11 005828


capaj:
I confirm this indeed fixes #10641

Jolg42:
Thanks for the PR! This looks good to me 💚 

I think it might be a copy paste of this one, and we could consider doing the same there:
https://github.com/prisma/prisma/blob/main/packages/client/src/generation/generateClient.ts#L28:L28

@aqrln @millsp What do you think? Would also be nice to have an extra test maybe?

aqrln:
Having a small snapshot test would be nice, but LGTM either way.

Jolg42:
The downside of a snapshot test would be that it's different on every machine / OS probably.

aqrln:
Good point. I'm fine with just checking that `error.stack` is not undefined if @YassinEldeeb decides to add a test then.

YassinEldeeb:
I think the only difference in the stack error on each OS would be the separator that's used to express the paths to trace where the error was thrown cause path separator is different between windows and unix based OSs.

so how about checking for these two things in the test:
- If it contains "ErrorType: ErrorMsg" 
- If it contains the function name that the error was thrown from cause this confirms that the stack trace is working

Jolg42:
That sounds great! 

The goal is to keep something which will have the same results on your computer / my computer / CI and this should do.

YassinEldeeb:
I can also help writing that test if you don't mind

Jolg42:
@YassinEldeeb You are very welcome to write the test, let us know if you need help

YassinEldeeb:
I've modified the test suit for `rejectOnNotFound` to work with the added stack trace with the following:
1. not depending on snapshots anymore
2. check for `r (result of prisma query)` to equal `null` if `rejectOnNotFound` wasn't enabled
3. making sure `stack` and `message` on the thrown error are defined
4. wrapping the `prisma.user[method]({..})` in a function with a special name `testRejectionOnNotFound`
5. searching for the function name in the stack trace
6. searching for `ErrorName: ErrorMsg` in the stack trace

and we're now checking for the most important parts in the stack trace if it's correct or not.

YassinEldeeb:
I liked the codebase a lot, It's very organized and very easy to search through.
keep up the good work 💚

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
```diff
@@ -50,21 +50,42 @@ for (const constructorKey of Object.keys(cases.constructor)) {
...
-        // Test Rejection
-        try {
+        // This function name is important cause we're searching
+        // for it's name in the stack bellow
+        const testRejectionOnNotFound = async () => {
           const r = await prisma.user[method]({
             where: { id: 'none' },
             rejectOnNotFound: value,
           })
-          expect(r).toMatchSnapshot()
```
capaj, line 65: @YassinEldeeb *suggestion*: I would suggest we keep the existing snapshots in place. In your branch the expect uses the thrown `error.message` to build the assertion which makes the test less likely to break when the error message changes.  In that case we want the test to fail-error messages should stay the same as some people use them to recognize which error was thrown.

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
```diff
@@ -50,21 +50,42 @@ for (const constructorKey of Object.keys(cases.constructor)) {
...
+          // 1. The "ErrorName: ErrorMsg"
+          // 2. The function name that the error was thrown from
+          const isCustomError = error.name === 'NotFoundError'
+
+          if (isCustomError) {
+            expect(stack.includes(`NotFoundError: ${error.message}`)).toBeTruthy()
+          } else {
+            expect(stack.includes(`Error: ${error.message}`)).toBeTruthy()
+          }
+          expect(stack.includes('at testRejectionOnNotFound')).toBeTruthy()
```
Jolg42, line 88:
```suggestion
          if (isCustomError) {
            expect(stack).toContain(`NotFoundError: ${error.message}`)
          } else {
            expect(stack).toContain(`Error: ${error.message}`)
          }
          expect(stack).toContain('at testRejectionOnNotFound')
```

I think this should work the same and might result in better help from test if falsy

YassinEldeeb: yeah looks better too 😅

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
YassinEldeeb: the snapshot looks like this

```

exports[`rejectOnNotFound | constructor=customError | findFirst=false 1`] = `null`;

exports[`rejectOnNotFound | constructor=customError | findUnique=false 1`] = `null`;

exports[`rejectOnNotFound | constructor=customErrorPerActionPerModel | findFirst=false 1`] = `null`;

exports[`rejectOnNotFound | constructor=customErrorPerActionPerModel | findUnique=false 1`] = `null`;

exports[`rejectOnNotFound | constructor=errorPerAction | findFirst=false 1`] = `null`;

exports[`rejectOnNotFound | constructor=errorPerAction | findUnique=false 1`] = `null`;

exports[`rejectOnNotFound | constructor=false | findFirst=false 1`] = `null`;

exports[`rejectOnNotFound | constructor=false | findFirst=undefined 1`] = `null`;

...
```

so what's the point of it if it's gonna always be equal to null

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
```diff
@@ -50,21 +50,42 @@ for (const constructorKey of Object.keys(cases.constructor)) {
       const value = currentMethod[valueKey]
       test(`rejectOnNotFound | constructor=${constructorKey} | ${method}=${value}`, async () => {
         // It should fail or not
-        expect.assertions(1)
```
aqrln, line 53: If this line is removed, then the comment above has to be as well.

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
YassinEldeeb: checking for null is the same as checking if it's equal to the snapshot in this case

capaj: oh sorry wrong line. Will suggest another one.

aqrln: @YassinEldeeb it doesn't look right, we're basically losing a bunch of tests in this case. These should probably be made snapshot tests for `error.message` then.

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
```diff
@@ -50,21 +50,42 @@ for (const constructorKey of Object.keys(cases.constructor)) {
...
+
+          // If it got to here, then no error was thrown
+          // so we check the value if it's equal to null
+          expect(r).toBeNull()
+        }
+
+        try {
+          await testRejectionOnNotFound()
         } catch (error) {
-          expect(error).toMatchSnapshot()
```
capaj, line 67: this was the line I wanted to suggest we should keep @YassinEldeeb in this comment: https://github.com/prisma/prisma/pull/10645/files#r768610234

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
YassinEldeeb: @aqrln  this is a sample from the actual snapshot, but they're all equal to null anyways

capaj: @YassinEldeeb yes this line doesn't need to be retained :100: , the other one does though

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
YassinEldeeb: @capaj  yeah but @Jolg42 said that the snapshot would be different between the different OSs so It's not good to depend on the error snapshot, cause It won't be the same as for Windows/Linux/..etc

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
```diff
@@ -50,21 +50,42 @@ for (const constructorKey of Object.keys(cases.constructor)) {
...
+          // Checking for the most important parts in the stack trace
+          // 1. The "ErrorName: ErrorMsg"
+          // 2. The function name that the error was thrown from
+          const isCustomError = error.name === 'NotFoundError'
+
+          if (isCustomError) {
+            expect(stack.includes(`NotFoundError: ${error.message}`)).toBeTruthy()
+          } else {
+            expect(stack.includes(`Error: ${error.message}`)).toBeTruthy()
+          }
```
aqrln, line 87: I don't think checking this provides a lot of value since it's testing V8 instead of Prisma, basically. It will also make things more complicated if/when there will be more error types. Just checking `expect(stack).toContain('at testRejectionOnNotFound')` should be enough.

Instead of this, I would prefer to have a snapshot test for the error message, and get the removed snapshots back:

```suggestion
          expect(error).toMatchSnapshot()
```

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
capaj: Fair point, but I think the least we can do is:
`expect(error.message).toMatchSnapshot()`

aqrln: @YassinEldeeb I don't think the error snapshot in Jest includes the stack trace, but in case it does, you may use separate snapshots for `error.name` and `error.message`.

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
capaj: I think jest serializes errors with stack traces and these differ on OSes. Can you confirm this @YassinEldeeb ?

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
aqrln: (I'm pretty sure the error snapshot looks like just `[ErrorName: message]` without any platform-specific things though)

capaj: yeah you are right: https://github.com/facebook/jest/blob/6d101e9b8190c0d265b40df96365dce704fc18d1/packages/pretty-format/src/__tests__/prettyFormat.test.ts#L109

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
YassinEldeeb: @aqrln yeah you're right, that's just extra unnecessary checks that can break in the future

YassinEldeeb: > Instead of this, I would prefer to have a snapshot test for the error message, and get the removed snapshots back:

but the old snapshot have some unnecessary values that equal null that has been replaced with [this line](https://github.com/prisma/prisma/pull/10645/files#diff-6b9e6e1a274c24674bd0fda583cc7f9972b206db55cc958f6f9b5e03f510dddaR68) , so I think It's better updating the snapshot to only include this `expect(error).toMatchSnapshot()` so It looks like this

```

exports[`rejectOnNotFound | constructor=customError | findFirst=() => new Error('FindFirst Custom Error') 1`] = `FindFirst Custom Error`;

exports[`rejectOnNotFound | constructor=customError | findFirst=true 1`] = `No User found`;

exports[`rejectOnNotFound | constructor=customError | findFirst=undefined 1`] = `Constructor Custom Error`;

exports[`rejectOnNotFound | constructor=customError | findUnique=() => new Error('FindUnique Custom Error') 1`] = `FindUnique Custom Error`;

exports[`rejectOnNotFound | constructor=customError | findUnique=true 1`] = `No User found`;

...
```

aqrln: @YassinEldeeb yes, precisely. Sorry, I should have made myself more clear.

aqrln: Also, to clarify just in case, `expect(stack).toContain('at testRejectionOnNotFound')` check should still be there separately.

YassinEldeeb: I've committed the checking for a snapshot instead of all of these unnecessary shenanigans

[commit](https://github.com/prisma/prisma/pull/10645/commits/7cd800ad34a7486a0350fffc3fb3b695b64d7bfe)

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
```diff
@@ -49,21 +49,31 @@ for (const constructorKey of Object.keys(cases.constructor)) {
...
+        }
+
+        try {
+          await testRejectionOnNotFound()
         } catch (error) {
+          const { message, stack } = error
+          expect(stack).toBeDefined()
+          expect(message).toBeDefined()
+
           expect(error).toMatchSnapshot()
```
aqrln, line 77:
```suggestion
          expect(error).toMatchSnapshot()
          expect(stack).toContain('at testRejectionOnNotFound')
```

aqrln: https://github.com/prisma/prisma/pull/10645#discussion_r768650783

aqrln: (FWIW, I would also be fine with just checking that the stack is not undefined, but then it kinda defeats the purpose of all other changes, and they all could be reverted and replaced with a one line addition. Since we already have them anyway, we may assert something useful about the stack trace as well).

YassinEldeeb: oh I've removed it by mistake, thanks for noticing

capaj (approved):
I love it :heart_eyes:

packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts
```diff
@@ -49,22 +49,33 @@ for (const constructorKey of Object.keys(cases.constructor)) {
...
-        expect.assertions(1)
         const PrismaClient = await getTestClient()
         const prisma = new PrismaClient({
           rejectOnNotFound: constructor,
         })
 
-        // Test Rejection
-        try {
+        // This function name is important cause we're searching
+        // for it's name in the stack bellow
```
aqrln, line 58:
```suggestion
        // for its name in the stack below
```

aqrln (approved):
LGTM with a nit

Jolg42 (approved):

PR #10675, opened 2021-12-13
docs(ARCHITECTURE): Warning for DMMF

janpio:
Added small warning that API is internal only, no guarantees.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -24,16 +24,18 @@ See [Prisma Generators](https://prismaio.notion.site/Prisma-Generators-a2cdf2622
 
 ## The `DMMF`, or Data Model Meta Format
 
-What the ... is DMMF? It's the Datamodel Meta Format. It is an AST (abstract syntax tree) of the datamodel in the form of JSON.
+What the ... is DMMF? It's the Datamodel Meta Format. It is an AST (abstract syntax tree) of the datamodel in the form of JSON.  
 The whole Prisma Client is just generated based on the DMMF, which comes from the Rust engines.
+
+> ⚠️ Note:  DMMF is a Prisma ORM internal API with no guarantees for stability to outside users. We might - and do - change the DMMF in potentially breaking ways between minor versions. 🐲
 
 ### Upgrading and debugging
 
 <!-- TODO -->
 
-Oftentimes, the Rust team did a change in DMMF, which you now need to integrate. How to do that?
-The first step is to identify, which new `@prisma/engines` version you want to use.
-Either have a look in the **Versions** tab in https://www.npmjs.com/package/@prisma/engines or check out `npm info @prisma/engines` in your terminal.
+Oftentimes, the Rust team did a change in DMMF, which you now need to integrate. How to do that?  
+The first step is to identify, which new `@prisma/engines` version you want to use.  
+Either have a look in the **Versions** tab in https://www.npmjs.com/package/@prisma/engines or check out `npm info @prisma/engines` in your terminal.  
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -24,16 +24,18 @@ See [Prisma Generators](https://prismaio.notion.site/Prisma-Generators-a2cdf2622
 
 ## The `DMMF`, or Data Model Meta Format
 
-What the ... is DMMF? It's the Datamodel Meta Format. It is an AST (abstract syntax tree) of the datamodel in the form of JSON.
+What the ... is DMMF? It's the Datamodel Meta Format. It is an AST (abstract syntax tree) of the datamodel in the form of JSON.  
 The whole Prisma Client is just generated based on the DMMF, which comes from the Rust engines.
 
+> ⚠️ Note:  DMMF is a Prisma ORM internal API with no guarantees for stability to outside users. We might - and do - change the DMMF in potentially breaking ways between minor versions. 🐲
```
millsp, line 30:
```suggestion
> ⚠️ Note: The DMMF is a Prisma ORM internal API with no guarantees for stability to outside users. We might - and do - change the DMMF in potentially breaking ways between minor versions. 🐲
```

ARCHITECTURE.md
```diff
@@ -24,16 +24,18 @@ See [Prisma Generators](https://prismaio.notion.site/Prisma-Generators-a2cdf2622
 
 ## The `DMMF`, or Data Model Meta Format
 
-What the ... is DMMF? It's the Datamodel Meta Format. It is an AST (abstract syntax tree) of the datamodel in the form of JSON.
+What the ... is DMMF? It's the Datamodel Meta Format. It is an AST (abstract syntax tree) of the datamodel in the form of JSON.
```
millsp, line 27: Is the double space intended?

millsp (changes requested):

ARCHITECTURE.md
Jolg42: Probably yes to have a line return

Jolg42 (approved):
Thanks!

ARCHITECTURE.md
janpio: Yep, the document was written as if someone originally intended there to be line breaks, but GitHub Markdown does only render them with double spaces - so I added them. Alternative would have been to reformat the file itself to not have line breaks that do not render.

janpio:
Hmm, why does the `Detect jobs to run` job not run here but is expected to pass?
I could merge this with admin rights not, but do we have a problem here in our config, maybe because this just edits a `.md` file maybe?

Jolg42:
@janpio I think the current issue with the setup is that "Detect jobs to run" is a required step (useful so automerge doesn't merge before it runs)

But here it cannot run because it's skipped by `paths-ignore` https://github.com/prisma/prisma/blob/main/.github/workflows/test.yml#L32

So maybe we can improve later by doing the logic of skipping in "Detect jobs to run" so it is never skipped

Jolg42:
By the way should we merge this now?

millsp (approved):

aqrln (approved):
To me it doesn't look like those paragraphs were necessarily intended to be rendered with line breaks, some people just prefer putting each sentence on its own line in Markdown sources instead of hard-breaking lines at a specific column, but LGTM either way.

janpio:
Abused admin rights to merge. Please create an issue to fix the underlying issue as it could otherwise block us in the future.

aqrln:
@janpio https://github.com/prisma/prisma/issues/10700

PR #10732, opened 2021-12-16
chore: comments from deep dive #7045

Jolg42:
See previous PR https://github.com/prisma/prisma/pull/7045

### DIFF
--- a/packages/sdk/src/IntrospectionEngine.ts
+++ b/packages/sdk/src/IntrospectionEngine.ts
@@ -269,14 +269,13 @@ export class IntrospectionEngine {
           const binaryPath = await resolveBinary(BinaryType.introspectionEngine)
           debugRpc('starting introspection engine with binary: ' + binaryPath)
 
-          // about `cwd`
-          // If the process is spawned from another directory, all file paths would resolve relative to that instead of the prisma directory
-          // note that it isn't something engines specific but just a process spawning thing.
-          // Paths resolved in engines code include at least:
-          // sqlite database paths
-          // ssl certificate paths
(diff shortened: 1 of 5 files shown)


packages/sdk/src/IntrospectionEngine.ts
```diff
@@ -268,6 +269,7 @@ export class IntrospectionEngine {
           const binaryPath = await resolveBinary(BinaryType.introspectionEngine)
           debugRpc('starting introspection engine with binary: ' + binaryPath)
 
+          // TODO Why does the engine need the cwd
```
Jolg42, line 272:
```suggestion
          // about `cwd`
          // If the process is spawned from another directory, all file paths would resolve relative to that instead of the prisma directory
          // note that it isn't something engines specific but just a process spawning thing.
          // Paths resolved in engines code include at least:
          // sqlite database paths
          // ssl certificate paths
```

packages/sdk/src/IntrospectionEngine.ts
```diff
@@ -269,7 +269,12 @@ export class IntrospectionEngine {
           const binaryPath = await resolveBinary(BinaryType.introspectionEngine)
           debugRpc('starting introspection engine with binary: ' + binaryPath)
 
-          // TODO Why does the engine need the cwd
+          // about `cwd`
```
aqrln, line 272: Would it make sense to move this comment near the `cwd: this.cwd,` line or would the formatting be weird then?

Jolg42: moved ✅

aqrln (approved):

PR #10739, opened 2021-12-17
ci: use setup-node@v2 native cache

Jolg42:
See 
https://github.com/actions/setup-node#caching-packages-dependencies
https://pnpm.io/continuous-integration#github-actions
https://github.com/pnpm/action-setup


Jolg42:
Need to use before `setup-node` to work
```
    - uses: pnpm/action-setup@v2.0.1
      with:
        version: 6.20.3
```

But means we can then remove the manual install of pnpm from our `.buildkite/test|publish/run.sh`

aqrln (approved):

PR #10764, opened 2021-12-19
feat(client): tracing

### LINKED ISSUE #10004: PCO: Pass the trace ID through to an SQL comment
Pull the trace implemented in the application and send it through to the query engine

### LINKED ISSUE #10310: PCO: Pass the trace ID through end-to-end
Pass the application trace ID from Prisma Data Platform server through the Prisma Client, through the Query Engine to SQL Comments

### LINKED ISSUE #11202: Incorrect type for `DMMF.Field.type` field
This field is always of type string.

https://prisma-company.slack.com/archives/C016KUHB1R6/p1638563060145800

millsp:
This PR brings OpenTelemetry tracing support to the Prisma Client.
The feature is in preview an is guarded by the `tracing` feature flag.
I've refactored runtime models while looking at propagation issues.
Note: Waiting to integrate until the engines have tracing support.

closes https://github.com/prisma/prisma/issues/10310

Example project that uses tracing for Prisma Client queries:

```ts
import { trace, context } from "@opentelemetry/api";
import { PrismaClient } from ".prisma/client";
import { otelSetup } from "./otelSetup";

otelSetup();

// set things up for tracing our "app"
const tracer = trace.getTracer("myApp");
const span = tracer.startSpan("mySpan");
const ctx = trace.setSpan(context.active(), span);

context.with(ctx, async () => {
  await main();

  span.end();
});

async function main() {

  // this middleware gets the current span and add info to it
  prisma.$use(function spanModifierMiddleware(params, next) {
    const span = trace.getSpan(context.active())

    span.setAttributes({
      model: params['model'],
      action: params['action']
    });

    return next(params)
  });

  // this middleware self-registers in a span to trace itself
  prisma.$use(function spanCreatingMiddleware(params, next) {
    const span = tracer.startSpan("middleware");
    const ctx = trace.setSpan(context.active(), span)

    const result = context.with(ctx, () => next(params));

    span.end()

    return result
  });

  // here begins the actual prisma client logic to be traced
  const email = Date.now() + "@gmail.com"

  await prisma.user.create({
    data: {
      name: 'John Doe',
      email: email,
      link: {
        create: {
          url: "http://www.google.com",
          shortUrl: "http://www.google.com",
        },
      },
    },
  });

  await prisma.user.findMany({})

  await prisma.$transaction([
    prisma.user.findUnique({
      where: { email }
    }),
    prisma.user.findMany({})
  ])

  await prisma.$disconnect();
}
```

Setup needed to begin tracing with OTEL, `./otelSetup`:

```ts
import { Resource } from "@opentelemetry/resources";
import { SemanticResourceAttributes } from "@opentelemetry/semantic-conventions";
import {
  BasicTracerProvider,
  ConsoleSpanExporter,
  SimpleSpanProcessor,
} from "@opentelemetry/sdk-trace-base";
import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-grpc";
import { registerInstrumentations } from "@opentelemetry/instrumentation";
import { AsyncHooksContextManager } from "@opentelemetry/context-async-hooks";
import * as api from "@opentelemetry/api";

/** SETUP */

export function otelSetup() {
  // a context manager is required to propagate the context
  const contextManager = new AsyncHooksContextManager().enable();

  // it's for node.js for span nesting and ctx propagation
  api.context.setGlobalContextManager(contextManager);

  // a simple exporter that logs the raw data to the console
  const consoleExporter = new ConsoleSpanExporter();

  // exporter that natively works with jaeger without extras
  const otlpTraceExporter = new OTLPTraceExporter();
  // docker run --rm -d -p 13133:13133 -p 16686:16686 -p 4317:55680 jaegertracing/opentelemetry-all-in-one:latest

  // a standard provider that can run on the web and in node
  const provider = new BasicTracerProvider({
    resource: new Resource({
      // we can define some metadata about the trace resource
      [SemanticResourceAttributes.SERVICE_NAME]: "basic-service",
      [SemanticResourceAttributes.SERVICE_VERSION]: "1.0.0",
    }),
  });

  // a provider combines multiple exporters to send the data
  // provider.addSpanProcessor(new SimpleSpanProcessor(consoleExporter));
  provider.addSpanProcessor(new SimpleSpanProcessor(otlpTraceExporter));

  // makes the provider the global tracer provider for telemetry
  provider.register();
}
```

What to expect in the visualizer
Screenshot from 2021-12-19 16-54-57


packages/client/src/__tests__/integration/happy/disconnect-finally/test.ts
```diff
@@ -24,4 +24,6 @@ test('disconnect-finally', async () => {
 
   const data = await run()
   expect(data).toMatchSnapshot()
+
+  await prisma.$disconnect()
```
millsp, line 28: This tests was blocking the CI/jest because doing `await` on the prisma promise re-wakes the engine, even afyter a disconnect - which is the purpose of the test. However, the test did not do a final `$disconnect` after doing that.

packages/client/src/__tests__/integration/happy/disconnect-while-query/test.ts
```diff
@@ -6,10 +6,13 @@ test('disconnect-while-query', async () => {
...
 
-  await a
+  prisma.$disconnect()
+  await a // re-wakes the engine
   await prisma.$disconnect()
+  await a // re-wakes the engine
 
   expect(await a).toMatchInlineSnapshot(`Array []`)
+
+  await prisma.$disconnect()
```
millsp, line 17: This tests was blocking the CI/jest because doing `await` on the prisma promise re-wakes the engine, even afyter a disconnect - which is the purpose of the test. However, the test did not do a final `$disconnect` after doing that.

packages/client/src/__tests__/select.test.ts
```diff
@@ -1,13 +1,13 @@
...
 import { getDMMF } from '../generation/getDMMF'
-import { DMMFClass } from '../runtime/dmmf'
+import { DMMFHelper } from '../runtime/dmmf'
 import { makeDocument } from '../runtime/query'
 
 let dmmf
 beforeAll(async () => {
   const dmmfDocument = await getDMMF({ datamodel: blog })
-  dmmf = new DMMFClass(dmmfDocument)
+  dmmf = new DMMFHelper(dmmfDocument)
```
millsp, line 10: Just a rename.

packages/client/src/generation/TSClient/Output.ts
```diff
@@ -10,15 +10,14 @@ import { wrapComment } from './helpers'
...
+    protected readonly dmmf: DMMFHelper,
     protected readonly field: DMMF.Field,
     protected readonly useNamespace = false,
   ) {}
   public toTS(): string {
     const { field, useNamespace } = this
     // ENUMTODO
-    let fieldType =
-      typeof field.type === 'string' ? GraphQLScalarToJSTypeTable[field.type] || field.type : field.type[0].name
+    let fieldType = GraphQLScalarToJSTypeTable[field.type] || field.type
```
millsp, line 20: `field.type[0].name` can never happen, I forgot to remove this legacy code last time.

packages/client/src/runtime/core/model/aggregates/aggregate.ts
```diff
@@ -0,0 +1,82 @@
+import type { Client } from '../../../getPrismaClient'
```
millsp, line 1: New file for `.aggregate` logic

packages/client/src/runtime/core/model/aggregates/count.ts
```diff
@@ -0,0 +1,22 @@
+import type { Client } from '../../../getPrismaClient'
```
millsp, line 1: New file for `.count` logic

packages/client/src/runtime/core/model/aggregates/groupBy.ts
```diff
@@ -0,0 +1,28 @@
+import type { Client } from '../../../getPrismaClient'
```
millsp, line 1: New file for `.groupBy` logic

packages/client/src/runtime/core/model/aggregates/utils/aggregateMap.ts
```diff
@@ -0,0 +1,7 @@
+export const aggregateMap = {
```
millsp, line 1: Extracted from `getPrismaClient.ts`

packages/client/src/runtime/core/model/applyAggregates.ts
```diff
@@ -0,0 +1,25 @@
+import type { Action, Client } from '../../getPrismaClient'
```
millsp, line 1: New file for choosing the right aggregate logic

packages/client/src/runtime/core/model/applyFluent.ts
```diff
@@ -0,0 +1,115 @@
+import type { Client } from '../../getPrismaClient'
```
millsp, line 1: New file for the fluent api logic

packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -0,0 +1,85 @@
+import type { O, F } from 'ts-toolbelt'
```
millsp, line 1: New file for the runtime model logic (eg. `prisma.user.findMany()`)

packages/client/src/runtime/core/model/applyModels.ts
```diff
@@ -0,0 +1,39 @@
+import type { Client } from '../../getPrismaClient'
```
millsp, line 1: New file for the runtime model emulation (eg. `prisma.user`)

packages/client/src/runtime/core/request/PrismaPromise.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
millsp, line 1: Moved from `getPrismaClient.ts`

packages/client/src/runtime/core/request/createPrismaPromise.ts
```diff
@@ -0,0 +1,54 @@
+import type { PrismaPromise } from './PrismaPromise'
```
millsp, line 1: Moved from `getPrismaClient.ts`

packages/client/src/runtime/core/utils/getCallSite.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
millsp, line 1: Extracted out from `Client` class.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1095,11 +1093,13 @@ new PrismaClient({
 
         if (globalThis.NOT_PRISMA_DATA_PROXY) {
           // async scope https://github.com/prisma/prisma/issues/3148
-          const resource = new AsyncResource('prisma-client-request')
-          return resource.runInAsyncScope(() => consumer(params))
+          return await new AsyncResource('prisma-client-request').runInAsyncScope(() => {
+            return runInChildSpan('request', tracer, internalParams.otelCtx, () => consumer(params))
```
millsp, line 1097: We have to manually pass the previously saved otel context because we can't propagate it automatically through prisma promises.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1060,7 +1056,9 @@ new PrismaClient({
      * @param middlewareIndex
      * @returns
      */
-    private _request(internalParams: InternalRequestParams): Promise<any> {
+    async _request(internalParams: InternalRequestParams): Promise<any> {
+      const tracer = trace.getTracer('prisma')
```
millsp, line 1060: Register an otel tracer for prisma.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -153,6 +158,7 @@ export type InternalRequestParams = {
   headers?: Record<string, string> // TODO what is this
   transactionId?: number // TODO what is this
   unpacker?: Unpacker // TODO what is this
+  otelCtx?: Context // an otel context
```
millsp, line 161: A saved instance of the context at the moment of creation of the prisma promise.

packages/client/src/runtime/core/request/createPrismaPromise.ts
```diff
@@ -0,0 +1,54 @@
...
+ * methods return one, this allows for pre-preparing queries without executing
+ * them until `.then` is called. It's the foundation of Prisma's query batching.
+ * @param callback that will be wrapped within our promise implementation
+ * @see [[PrismaPromise]]
+ * @returns
+ */
+export function createPrismaPromise(
+  callback: (transactionId?: number, runInTransaction?: boolean, otelCtx?: Context) => PrismaPromise<unknown>,
+): PrismaPromise<unknown> {
+  const otelCtx = context.active() // get the context at time of creation
```
millsp, line 17: Capture the otel context here for the first time on creation of the prisma promise.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1118,15 +1118,6 @@ new PrismaClient({
       transactionId,
       unpacker,
     }: InternalRequestParams) {
-      if (action !== 'executeRaw' && action !== 'queryRaw' && !model) {
```
millsp, line 1121: This is not needed anymore since the refactor, it cannot happen.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1208,253 +1199,6 @@ new PrismaClient({
       })
     }
 
-    private _bootstrapClient() {
```
millsp, line 1211: Some new year resolutions.

packages/client/src/runtime/utils/otel/runInChildSpan.ts
```diff
@@ -0,0 +1,26 @@
+import type { Context, Span, Tracer } from '@opentelemetry/api'
```
millsp, line 1: Helper to do our otel stuff.

packages/client/src/runtime/utils/printStack.ts
```diff
@@ -83,7 +83,8 @@ function parseStack({
       !t.file.includes('getPrismaClient') &&
       !t.file.startsWith('internal/') && // We don't want internal nodejs files
       !t.methodName.includes('new ') &&
-      !t.methodName.includes('_getCallsite') &&
+      !t.methodName.includes('getCallSite') &&
+      !t.methodName.includes('Proxy.') &&
```
millsp, line 87: Extra filter needed since the use of proxies.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -1,5 +1,3 @@
-/// <reference lib="webworker" />
```
millsp, line 1: I just moved it around.

packages/generator-helper/src/dmmf.ts
```diff
@@ -76,7 +76,11 @@ export namespace DMMF {
     isReadOnly: boolean
     isGenerated: boolean
     isUpdatedAt: boolean
-    type: string | DMMF.SchemaEnum | DMMF.OutputType | DMMF.SchemaArg
```
millsp, line 79: This is incorrect and is just a type-related change.

packages/client/src/runtime/getPrismaClient.ts
janpio: Worth adding a comment in code?

packages/client/src/runtime/getPrismaClient.ts
janpio: Should this always happen or only if Otel is detected/present?

millsp: Yes, otel is designed to do no-ops if it has not been set up.

packages/client/src/__tests__/integration/happy/aggregations/test.ts
```diff
@@ -116,27 +116,27 @@ describe('aggregations', () => {
...
+            ~~~~~
+          },
+          where: {
+            age: {
+              gt: -1
+            }
+          },
+          skip: 0,
+          take: 10000
+        }
```
janpio, line 134: Why the changed order here?

packages/client/src/__tests__/integration/happy/count/test.ts
```diff
@@ -41,17 +41,21 @@ test('count', async () => {
   } catch (err) {
     expect(err.message).toMatchInlineSnapshot(`
 
-      Invalid \`prisma.user.aggregate()\` invocation:
+      Invalid \`prisma.user.count()\` invocation:
```
janpio, line 44: Previously undetected bug in the error message, that it used the wrong method name?

packages/client/src/runtime/getPrismaClient.ts
millsp: I added it closer to its context in https://github.com/prisma/prisma/pull/10764/files#diff-94c573b7cd12d73c919ef8300bc74e7536fb8cf26682c8409bf5cd8a9e6132f7R18

packages/client/src/__tests__/integration/happy/disconnect-finally/test.ts
janpio: Why did this never cause a problem before?
Or is this one of the flaky tests we had for ages?

packages/client/src/__tests__/integration/happy/disconnect-while-query/test.ts
janpio: The test now does an additinoal `await a` in line 13 - is that necessary if line 15 does that in the `expect` was well?

packages/client/src/__tests__/integration/happy/aggregations/test.ts
millsp: I think it's because of the orders of the creation of the keys when we "de-sugar" user input.

packages/client/src/__tests__/integration/happy/count/test.ts
millsp: Indeed, count is an `aggregate` under the hood.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -658,9 +659,9 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
       query: string | TemplateStringsArray | sqlTemplateTag.Sql,
       ...values: sqlTemplateTag.RawValue[]
     ) {
-      const request = (transactionId?: number, runInTransaction?: boolean) => {
+      const request = (txId?: number, inTx?: boolean, otelCtx?: Context) => {
         try {
-          const promise = this.$executeRawInternal(runInTransaction ?? false, transactionId, query, ...values)
+          const promise = this.$executeRawInternal(txId, inTx, otelCtx, query, ...values)
```
janpio, line 664: What motives difference in param naming here vs. $executeRawInternal?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -821,9 +817,9 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
       query: string | TemplateStringsArray | sqlTemplateTag.Sql,
       ...values: sqlTemplateTag.RawValue[]
     ) {
-      const request = (transactionId?: number, runInTransaction?: boolean) => {
+      const request = (txId?: number, inTx?: boolean, otelCtx?: Context) => {
         try {
-          const promise = this.$queryRawInternal(runInTransaction ?? false, transactionId, query, ...values)
+          const promise = this.$queryRawInternal(txId, inTx, otelCtx, query, ...values)
```
janpio, line 822: Same as https://github.com/prisma/prisma/pull/10764/files#r772661885

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1077,12 +1075,12 @@ new PrismaClient({
           // if this `next` was called and there's some more middlewares
           const nextMiddleware = this._middlewares.query.get(++index)
 
-          if (nextMiddleware) {
-            // we pass the modfied params down to the next one, & repeat
-            return nextMiddleware(changedParams, consumer)
-          }
+          // we pass the modified params down to the next one, & repeat
+          if (nextMiddleware) return nextMiddleware(changedParams, consumer)
+          // calling `next` calls the consumer again with the new params
```
janpio, line 1080: does this comment apply to the line before?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1077,12 +1075,12 @@ new PrismaClient({
...
 
-          const changedInternalParams = { ...internalParams, ...params }
+          // before we send the execution request, we use the changed params
+          const changedInternalParams = { ...internalParams, ...changedParams }
```
janpio, line 1083: intentional change?
no change in functionality for users?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1095,11 +1093,13 @@ new PrismaClient({
 
         if (globalThis.NOT_PRISMA_DATA_PROXY) {
```
janpio, line 1094: Unrelated: This might deserve a comment why special handling for this case (and what that case really is)

packages/client/src/runtime/getPrismaClient.ts
janpio: I see the call of this removed in line 446 as well - but where did the logic go?

packages/client/src/runtime/utils/common.ts
```diff
@@ -65,7 +65,7 @@ export const needNamespace = {
   Decimal: 'Decimal',
 }
 
-export function needsNamespace(fieldType: DMMF.Field['type'], dmmf: DMMFClass): boolean {
+export function needsNamespace(fieldType: DMMF.SchemaField['outputType']['type'], dmmf: DMMFHelper): boolean {
```
janpio, line 68: What is that `SchemaField` change about?

packages/client/src/runtime/getPrismaClient.ts
millsp: We now use JS custom constructor return values
https://github.com/prisma/prisma/pull/10764/files/568f6c1aa293564ba9bf54bd0689462b3ba449a4#diff-09784fc0113a735572cff48990efa35434fb6ec6ec8876845ec9cf423642e5fdR449

packages/client/src/runtime/getPrismaClient.ts
millsp: Good point, I should rename the other instances too. I feel like `txId` and `inTx` are explicit enough with their shorter names.

packages/client/src/runtime/getPrismaClient.ts
millsp: Yes

packages/client/src/runtime/getPrismaClient.ts
millsp: Intentional, yes. It's a bug. Probably went undetected for a while because users mutate `params` in their middleware. But in fact - while that is possible - you should be able to pass a brand new object in the `next(...)` call as well. With the previous logic, it would not be possible.

packages/client/src/runtime/utils/common.ts
millsp: It's a change that I made last time after removing some legacy logic (after talking to Dom). I just realized that the type was wrong (by looking at its usage).

packages/client/src/__tests__/integration/happy/disconnect-while-query/test.ts
```diff
@@ -6,10 +6,13 @@ test('disconnect-while-query', async () => {
 
   await prisma.user.findMany()
   const a = prisma.user.findMany()
-  prisma.$disconnect()
 
-  await a
+  prisma.$disconnect()
+  await a // re-wakes the engine
   await prisma.$disconnect()
+  await a // re-wakes the engine
```
millsp, line 13:
```suggestion
```

packages/client/src/__tests__/integration/happy/disconnect-while-query/test.ts
millsp: Indeed not needed.

packages/client/src/__tests__/integration/happy/disconnect-finally/test.ts
millsp: By calling `disconnect` without await, we were basically pushing it down the event loop. For some reason, that has worked to happen after `await res`. So maybe the refactor has changed the order of these async events, I'm not certain.

millsp: I did a check on the `main` branch:
```ts
  const r = prisma.user.findMany();

  await r
  prisma.$disconnect()
  await r
```
:arrow_up:  Does send the query twice and exits properly (with `binary` of course). While:
```ts
  const r = prisma.user.findMany();

  prisma.$disconnect()
  await r
  await r
```
:arrow_up: Will prevent the program from exiting (still with `binary`). Confirms my theory.

packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -441,11 +441,7 @@ You may have to run ${chalk.greenBright('prisma generate')} for your changes to
     await this.start()
 
     this.lastQuery = JSON.stringify(request)
-    this.executingQueryPromise = this.engine!.query(
-      this.lastQuery,
-      JSON.stringify(headers), // TODO these aren't headers on the engine side
-      headers.transactionId,
-    )
+    this.executingQueryPromise = this.engine!.query(this.lastQuery, JSON.stringify(headers), headers.transactionId)
```
janpio, line 444: I think the comment that those are not _really_ headers was helpful. The `TODO` can go though of course.

matthewmueller:
Stuck it back in Doing to be wrapped up once https://github.com/prisma/prisma/issues/10004 is done.

packages/client/src/runtime/core/model/aggregates/aggregate.ts
```diff
@@ -0,0 +1,82 @@
...
+ * for not having to do things like: `{select: {_avg: {select: {age: true}}}}`.
+ * The goal here is to desugar it into something that is understood by the QE.
+ * @param userArgs to transform
+ * @returns
+ */
+export function desugarUserArgs(userArgs: object) {
+  const _userArgs = desugarCountInUserArgs(userArgs)
+  const userArgsEntries = Object.entries(_userArgs)
+
+  return userArgsEntries.reduce(
```
aqrln, line 16: Since the accumulator here is mutated rather than constructed immutably, it could've been just a `forEach` or a `for` loop with one or two lines of code less, but no strong opinion about this, `reduce` is fine too.

packages/client/src/runtime/core/model/aggregates/aggregate.ts
```diff
@@ -0,0 +1,82 @@
...
+  )
+}
+
+/**
+ * Desugar `userArgs` when it contains `{_count: true}`.
+ * @param userArgs the user input
+ * @returns
+ */
+function desugarCountInUserArgs(userArgs: object) {
+  if (typeof userArgs['_count'] === 'boolean') {
```
aqrln, line 37: Is `{ _count: false }` a valid input here?

packages/client/src/runtime/core/model/applyFluent.ts
```diff
@@ -0,0 +1,124 @@
...
+      has: (_, prop: string) => ownKeys.includes(prop),
+      ownKeys: () => ownKeys,
+    })
+  }
+}
+
+// the only accessible fields are relations to be chained on
+function getOwnKeys(client: Client, dmmfModelName: string) {
+  return client._dmmf.modelMap[dmmfModelName].fields
+    .filter((field) => field.kind === 'object') // relations
```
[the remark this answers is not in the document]
millsp: At least aligns with the original code
```
          // add relation fields
          for (const field of model.fields.filter((f) => f.kind === 'object')) {
```

packages/engine-core/src/library/LibraryEngine.ts
Jolg42:
```suggestion
    // Note that `headers` aren't headers on the engine side
    this.executingQueryPromise = this.engine!.query(this.lastQuery, JSON.stringify(headers), headers.transactionId)
```

packages/generator-helper/src/dmmf.ts
```diff
@@ -207,7 +211,7 @@ export namespace DMMF {
     delete = 'delete',
     deleteMany = 'deleteMany',
     groupBy = 'groupBy',
-    count = 'count',
+    count = 'count', // TODO: count does not actually exist, why?
```
Jolg42, line 214: What do you mean?

This exists https://www.prisma.io/docs/concepts/components/prisma-client/aggregation-grouping-summarizing#count

packages/client/src/runtime/core/model/aggregates/aggregate.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Creates an unpacker that adds sugar to the basic result of the QE. An
+ * unpacker helps to transform a result before returning it to the user.
+ * @param userArgs the user input
+ * @returns
+ */
+export function createSugarUnpacker(userArgs: object) {
+  return (data: object) => {
+    if (typeof userArgs['_count'] === 'boolean') {
+      if (Array.isArray(data)) {
```
Jolg42, line 53: Could add a comment of when it's an array here?

packages/generator-helper/src/dmmf.ts
millsp: I figured out that count is not an operation in the dmmf but we actually use aggregates for count, so maybe we should remove this as it is technically incorrect (might be used an have an actual meaning - to be checked first).

packages/client/src/runtime/core/model/aggregates/aggregate.ts
millsp: Will do. This is shared logic for all aggregates, so arrays only happen when you use group by. Probably not worth saving a few lines of code for the confusion, I'll de-condense.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -559,8 +559,9 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
      * Executes a raw query. Always returns a number
      */
     private $executeRawInternal(
-      runInTransaction: boolean,
```
Jolg42, line 562: Note that runInTransaction is documented in docs 
https://sourcegraph.com/search?q=context:global+repo:github.com/prisma/+runInTransaction+repo:%5Egithub%5C.com/prisma/docs%24+&patternType=literal

So can technically be a breaking change in API?

Jolg42: Or is it just changed inside internals here?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1093,11 +1092,11 @@ new PrismaClient({
           // if this `next` was called and there's some more middlewares
           const nextMiddleware = this._middlewares.query.get(++index)
 
-          if (nextMiddleware) {
-            // we pass the modified params down to the next one, & repeat
-            return nextMiddleware(changedParams, consumer)
-          }
+          // we pass the modified params down to the next one, & repeat
+          if (nextMiddleware) return nextMiddleware(changedParams, consumer)
+          // calling `next` calls the consumer again with the new params
```
Jolg42, line 1097:
```suggestion
          // we pass the modified params down to the next one, & repeat
          // calling `next` calls the consumer again with the new params
          if (nextMiddleware) return nextMiddleware(changedParams, consumer)
```
then this might be easier to read

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1111,11 +1110,13 @@ new PrismaClient({
 
         if (globalThis.NOT_PRISMA_DATA_PROXY) {
           // async scope https://github.com/prisma/prisma/issues/3148
-          const resource = new AsyncResource('prisma-client-request')
-          return resource.runInAsyncScope(() => consumer(params))
+          return await new AsyncResource('prisma-client-request').runInAsyncScope(() => {
```
Jolg42, line 1113: Why is await needed here now?

packages/client/src/runtime/getPrismaClient.ts
Jolg42: Looks ok from https://github.com/prisma/prisma/pull/10764/files#diff-eca352fdbc50201808d520041da8d6a977697fc51eb82750889c0c1ff670a904R38-R46

packages/client/src/runtime/core/model/applyModels.ts
```diff
@@ -0,0 +1,47 @@
...
+      if (modelCache[dmmfModelName] !== undefined) {
+        return modelCache[dmmfModelName]
+      }
+
+      // creates a new model proxy on the fly and caches it
+      if (client._dmmf.modelMap[dmmfModelName] !== undefined) {
+        return (modelCache[dmmfModelName] = applyModel(client, dmmfModelName))
+      }
+
+      // above just failed if the model name is lower cased
```
Jolg42, line 32:
```suggestion
      // above just failed if the model name is lower cased
```
You mean it fails silently?

packages/client/src/runtime/getPrismaClient.ts
janpio: I would probably not understand that when reading the code - but maybe that is just me. Feel free to ignore and resolve comment.

packages/client/src/runtime/getPrismaClient.ts
janpio: GH does not load that link. What makes the removable ok @Jolg42 ?

packages/generator-helper/src/dmmf.ts
Jolg42: Oh right

Jolg42 (approved):
Well done!

So the refactors are really neat using Proxies, it's much easier to read and has nice docs 💚 

About the tracing, someone reading this PR might assume this is a GA feature, so having some preview feature gating would be great before the release.

(+ see comments)

packages/client/src/runtime/getPrismaClient.ts
millsp: Some logic changed, but basically we want async operations to be `try/catch` before the return.

packages/client/src/runtime/core/model/aggregates/aggregate.ts
millsp: It aligns with what we were doing previously. So not a breaking change but I'm curious of the output.

PR #10797, opened 2021-12-21
fix(client): rethrow error for tx

### LINKED ISSUE #9533: interactiveTransactions rollback but don't error out if timeout is reached
### Bug description

When using interactive transactions, the transaction returns successfully even if the timeout was reached and internally the transaction was rolled back.

More context here https://github.com/prisma/prisma/issues/8664#issuecomment-930403241

### How to reproduce

1. Clone https://github.com/awinograd/prisma-include-clause-bug/tree/interactive-transaction-timeout (note it is a specific branch, not main branch)
2. run `yarn dev:database`
3. run `prisma migrate dev`
4. run `yarn dev`

### Expected behavior

Transaction that surpasses `timeout` should throw an error rather than returning the return value as if nothing has happened.

### Prisma information

all included in repro example above

### Environment & setup

- OS: MacOS
- Database: PostgresQL
- Node.js version: 16.9.1

### Prisma Version

```
Query Engine (Node-API) : libquery-engine c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary           : prisma-fmt c22652b7e418506fab23052d569b85d3aec4883f (at node_modules/@prisma/engines/prisma-fmt-darwin)
Preview Features        : interactiveTransactions
```

### LINKED ISSUE #10218: prisma.$transaction: Benchmark produces unexpected results
### Bug description

I have been doing some benchmarks against my project using Prisma. My project uses prisma's transactions and under high load, prisma throws an unmeaningful error
```ts
Error
    at Object.transaction (/.../prisma-issue-tx/starter/node_modules/@prisma/client/runtime/index.js:36083:55)
    at async PrismaClient._transactionWithCallback (/.../prisma-issue-tx/starter/node_modules/@prisma/client/runtime/index.js:39451:9) {
  code: 'StringExpected'
}
```

This error shadows this response from engine (captured from a breakpoint using vscode debugger):
image

I suspect that `StringExpected` error comes from these lines https://github.com/prisma/prisma/blob/main/packages/engine-core/src/library/LibraryEngine.ts#L119-L121, where an id is expected to be sent, but `undefined` is sent really. If engine returns an error here https://github.com/prisma/prisma/blob/main/packages/client/src/runtime/getPrismaClient.ts#L1006, the rest of the code should be avoided? (these lines https://github.com/prisma/prisma/blob/main/packages/client/src/runtime/getPrismaClient.ts#L1008-L1022)

I have been doing some tests with other drivers (`mysql2`), and there have'nt been problem with pressure, so I don't know if it's other additional issue, prisma should work under that pressure? I'm using https://github.com/mcollina/autocannon and the errors start when I use `-p` (`pipeling`). I have not too much experience with benchmarking, so I could be doing some bad assumptions or misunderstanding how I should be doing it. Anyway, that error is a real issue 😅 

### How to reproduce

- Clone https://github.com/tonivj5/prisma-issue
- `npm i`
- `npm run start`
- Once started, `npm run benchmark` it runs autocannon
- It should produces a ton of errors (`StringExpected....`)

### Expected behavior

Most useful errors, no errors (?)

### Prisma information

I have used starter project from Prisma

### Environment & setup

- OS: Linux Mint
- Database: SQLite, MariaDB/MySQL
- Node.js version: 17.0.1

### Prisma Version

```
prisma                  : 3.4.2
@prisma/client          : 3.4.2
Query Engine (Node-API) : libquery-engine 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt 57771c0558568c7d08bd34c7248af5244ae16bd9 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : 57771c0558568c7d08bd34c7248af5244ae16bd9
Studio                  : 0.438.0
Preview Features        : interactiveTransactions
Done in 0.84s.
```

### LINKED ISSUE #10622: Rollback not triggered in an interactive transaction when connection limit is high
### Bug description

I try to execute multiple interactive transactions concurrently and find that when the number of connections is high, throwing an error will not trigger transaction rollback
When the number of connections is small, although there will be errors, it can still be rolled back
(on another computer,  connections 5 (CPU is 2) ,can be rolled back without an error.)  

I'm not sure what caused it or whether it can be reproduced on other computers

### How to reproduce
[code demo](https://github.com/AoiIsMine/test)

### Environment & setup

- OS:  Windows,
- Database: MySQL 8.0
- Node.js version: v14.16.0

### Prisma Version

```
3.6.0
```

millsp:
Fixes https://github.com/prisma/prisma/issues/9533
Fixes https://github.com/prisma/prisma/issues/10218
Fixes https://github.com/prisma/prisma/issues/10622

Adds more tests for postgres.

SET FOR RELEASE 3.9.0


garrensmith (approved):
Nice fix

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
```diff
@@ -0,0 +1,473 @@
...
+
+    await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`you better rollback now`)
+
+    const users = await prisma.user.findMany()
+
+    expect(users.length).toBe(0)
+  })
+
+  /**
+   * A transaction might fail if it's called inside another transaction
```
aqrln, line 132: Does it only depend on the database or something else too?

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
```diff
@@ -0,0 +1,473 @@
...
+      await prisma.$transaction(
+        async (prisma) => {
+          await prisma.user.create({
+            data: {
+              email: 'user_2@website.com',
+            },
+          })
+        },
+        {
+          timeout: 1000,
```
aqrln, line 152: Why is the explicit timeout necessary here?

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
```diff
@@ -0,0 +1,473 @@
...
+    ])
+
+    const users = await prisma.user.findMany()
+
+    expect(users.length).toBe(2)
+  })
+
+  /**
+   * A bad batch should rollback using the interactive transaction logic
+   * // TODO: skipped because output differs from binary to library
```
aqrln, line 269: We may have two copies of this test, each conditionally skipped depending on `getClientEngineType()`

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
```diff
@@ -0,0 +1,473 @@
...
+      prisma.$executeRaw`INSERT INTO User (id, email) VALUES ("1", "user_1@website.com")`,
+      prisma.$executeRaw`INSERT INTO User (id, email) VALUES ("1", "user_1@website.com")`,
+    ])
+
+    await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
+
+                                                                                                Invalid \`prisma.executeRaw()\` invocation:
+
+
+                                                                                                  Raw query failed. Code: \`42601\`. Message: \`db error: ERROR: syntax error at or near "User"\`
```
aqrln, line 324: Is this the right error? I would've expected this to check that a unique constraint failed after the last `INSERT` statement. Was this test copied from the SQLite tests? Maybe the SQL statements need to be slightly adapted.

packages/client/src/__tests__/integration/happy/interactive-transactions-sqlite/test.ts
```diff
@@ -64,7 +54,7 @@ describe('interactive transaction', () => {
    *     Received promise resolved instead of rejected
    *     Resolved to value: [{"email": "user_1@website.com", "id": "0d258eae-1c22-4af1-8c95-68a17e995c2e", "name": null}]
    */
-  testIf(process.platform !== 'darwin' || !process.env.CI)('timeout override', async () => {
+  test('timeout override', async () => {
```
aqrln, line 57: If the reason this test was skipped is no longer an issue, the TODO comment above needs to be removed as well.

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
millsp: I think this comes from the single process nature of sqlite, it will lock the table for the time of the transaction.

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
millsp: I don't think so.

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
millsp: Thanks for the suggestion.

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
millsp: Good catch, thanks!

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
aqrln: Ah, so it only fails on SQLite, got it. That makes sense.

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
```diff
@@ -0,0 +1,473 @@
...
+    ])
+
+    const users = await prisma.user.findMany()
+
+    expect(users.length).toBe(2)
+  })
+
+  /**
+   * Middlewares should not prevent a rollback
+   * // TODO: skipped because output differs from binary to library
```
aqrln, line 362: Here as well: https://github.com/prisma/prisma/pull/10797/files#r783066268

millsp:
Needs https://github.com/prisma/prisma-engines/pull/2619 to be merged on master then tests will pass.

PR #10804, opened 2021-12-21
fix(client): disclude empty string from `findSync` args

### LINKED ISSUE #9435: no such file or directory, open '/path/schema.prisma' since 3.1.1
### Bug description

When I call ` new PrismaClient()`, I get the error 
```bash
(node:12802) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/mnt/d/workspace/path/packages/api/src/schema.prisma'
```
That just happen on 3.1.1

### How to reproduce

1. Upgrade prisma to 3.1.1
2. Execute `prisma db push` command
3. Call ` new PrismaClient()`

### Expected behavior

_No response_

### Environment & setup

- OS: Windows(wsl)
- Database: SQLite
- Node.js version: v14.17.4

### Prisma Version

```
Query Engine (Node-API) : libquery-engine c22652b7e418506fab23052d569b85d3aec4883f (at ../../.yarn/unplugged/@prisma-engines-npm-3.1.0-24.c22652b7e418506fab23052d569b85d3aec4883f-4e2ba773cc/node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli c22652b7e418506fab23052d569b85d3aec4883f (at ../../.yarn/unplugged/@prisma-engines-npm-3.1.0-24.c22652b7e418506fab23052d569b85d3aec4883f-4e2ba773cc/node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core c22652b7e418506fab23052d569b85d3aec4883f (at ../../.yarn/unplugged/@prisma-engines-npm-3.1.0-24.c22652b7e418506fab23052d569b85d3aec4883f-4e2ba773cc/node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt c22652b7e418506fab23052d569b85d3aec4883f (at ../../.yarn/unplugged/@prisma-engines-npm-3.1.0-24.c22652b7e418506fab23052d569b85d3aec4883f-4e2ba773cc/node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
```

Luke-zhang-04:
Closes #9435

- Since every string includes an empty string, findSync was matching everything

```js
function isMatched(string, regexs) {
  for (const regex of regexs) {
    if (typeof regex === "string") {
      if (string.includes(regex)) { // If regex is an empty string, it returns true every time
        return true;
      }
    } else if (regex.exec(string)) {
      return true;
    }
  }
  return false;
}
```


arthurfiorette:
When this will be merged? i'm waiting for a fix.

Luke-zhang-04:
I think we should be good? It's a relatively simple fix.

@Jolg42?

packages/client/src/generation/utils/buildDirname.ts
```diff
@@ -31,13 +31,13 @@ export function buildDirname(clientEngineType: ClientEngineType, relativeOutdir:
...
   const slsRelativeOutputDir = relativeOutdir.split(path.sep).slice(1).join(path.sep)
+  const slsRelativeOutputDirCode = slsRelativeOutputDir ? `\n${JSON.stringify(slsRelativeOutputDir)},` : ""
 
   return `
 const { findSync } = require('${runtimePath}')
 
 const dirname = findSync(process.cwd(), [
-    ${JSON.stringify(relativeOutdir)},
-    ${JSON.stringify(slsRelativeOutputDir)},
+    ${JSON.stringify(relativeOutdir)},${slsRelativeOutputDirCode}
```
arthurfiorette, line 40: If `slsRelativeOutputDirCode` is empty, there will be a trailing comma :)

Luke-zhang-04: trailing comma should be ok, and it was like that before

matthewmueller:
Assigned to @aqrln to review for 3.9.0

aqrln (approved):
Hi @Luke-zhang-04 and sorry for the delay! The change itself LGTM. However, optimally we would want to have a test for this.

Would it be feasible to add a simple test with a minimal directory structure that triggers the bug under `packages/client/src/__tests__/integration/happy` in this PR or should we make it an [e2e test](https://github.com/prisma/e2e-tests/) that involves something like Yarn workspaces?

Also, would you like to add a test yourself or would you prefer me to do this?

Luke-zhang-04:
I'd like to try adding it myself, but would that involve me installing all of the supported databases?

Luke-zhang-04:
ok cool, I'll give it a shot then

Luke-zhang-04:
Ok so I'm not too happy with what I ended up doing, but I couldn't find another way to force the client to generated like that. Also not too fond of the names I chose, but I had some trouble coming up with anything better.

I could also go ahead and add an E2E test if you want

aqrln:
Amazing, thank you! I'll take a closer look soon.

> I could also go ahead and add an E2E test if you want

I'd say it's not required since we now have a test in this PR, which is preferable, but an e2e test might have some additional value on its own, even outside the scope of this specific issue (e.g., if you make it use Yarn workspaces), so it's up to you. You can check https://github.com/prisma/e2e-tests/tree/dev/packagers and see if you want to contribute something there. Interestingly, there's a test called `pnpm-workspaces-custom-output`, but it seems to test a little bit different thing than what's been the issue in this case.

Regarding the PR, one thing I noticed is it currently fails on CI because the type checking happens before the client is generated for the first time:

```ts
[2022-01-13T05:17:20Z]  FAIL  src/__tests__/integration/happy/output-directory/test.ts
--
  | [2022-01-13T05:17:20Z]   ● Test suite failed to run
  | [2022-01-13T05:17:20Z]
  | [2022-01-13T05:17:20Z]     src/__tests__/integration/happy/output-directory/test.ts:14:41 - error TS2307: Cannot find module './prismaClient' or its corresponding type declarations.
  | [2022-01-13T05:17:20Z]
  | [2022-01-13T05:17:20Z]     14   const { PrismaClient } = await import('./prismaClient')
```

You may either replace `await import` with `require`, or add just add `// @ts-ignore` before this line.

`const { PrismaClient } = require('./prismaClient')` is the pattern we use throughout the tests, but I think

```ts
// @ts-ignore
const { PrismaClient } = await import('./prismaClient')
```

is totally fine as well. Eventually we'd want our tests to be ESM compatible, so why not start with this one. Besides, the latter option has an additional advantage of having proper autocompletion and type checking in dev environment, since the `@ts-ignore` directive will only have an effect until the client is generated, unlike `require` that will cause `PrismaClient` to always be `any`.

packages/client/src/utils/generateInFolder.ts
```diff
@@ -50,9 +52,11 @@ export async function generateInFolder({
...
   })
 
-  const outputDir = transpile
-    ? path.join(projectDir, 'node_modules/@prisma/client')
-    : path.join(projectDir, '@prisma/client')
+  const setOutputDir = relativeOutputDir ? path.join(projectDir, relativeOutputDir) : undefined
+
+  const outputDir =
+    setOutputDir ??
+    (transpile ? path.join(projectDir, 'node_modules/@prisma/client') : path.join(projectDir, '@prisma/client'))
```
aqrln, line 59:
```suggestion
  if (relativeOutputDir === undefined) {
    relativeOutputDir = transpile
      ? path.join(projectDir, 'node_modules/@prisma/client')
      : path.join(projectDir, '@prisma/client')
  }

  const outputDir = path.join(projectDir, relativeOutputDir)
```

packages/client/src/utils/generateInFolder.ts
```diff
@@ -50,9 +52,11 @@ export async function generateInFolder({
...
 
   // if (transpile && config.generators[0]?.output) {
   //   outputDir = path.join(path.dirname(schemaPath), config.generators[0]?.output)
```
aqrln, line 62: Making the `output` generator option in `schema.prisma` work in tests instead would've been optimal, but that would require some investigation of why this was commented out in the first place, and being able to override that sounds useful, so let's settle on the solution with `relativeOutputDir` for now.

packages/client/src/__tests__/integration/happy/output-directory/test.ts
```diff
@@ -0,0 +1,19 @@
...
+    projectDir: __dirname,
+    useLocalRuntime: true,
+    transpile: false,
+    useBuiltRuntime: false,
+    relativeOutputDir: './prismaClient',
+  })
+})
+
+test('connect to client generated 1 directory level', async () => {
+  const { PrismaClient } = await import('./prismaClient')
```
aqrln, line 14:
```suggestion
  // @ts-ignore
  const { PrismaClient } = await import('./prismaClient')
```

packages/client/src/__tests__/integration/happy/output-directory/schema.prisma
```diff
@@ -0,0 +1,15 @@
+datasource db {
+    provider = "sqlite"
+    url      = "file:dev.db"
+}
+
+generator client {
+    provider      = "prisma-client-js"
+    binaryTargets = ["native"]
```
aqrln, line 8:
```suggestion
```

packages/client/src/__tests__/integration/happy/output-directory/test.ts
aqrln: @arthurfiorette nope, that won't work. It will fix the CI but break the local workflow as soon as you run that test at least once.

arthurfiorette: ok

packages/client/src/utils/generateInFolder.ts
Luke-zhang-04: oh interesting, that's probably better

Luke-zhang-04:
Ok great. Regarding the esm stuff, I'm a bit confused since static imports are used in tall the other tests, and `require` doesn't mix well with `import`.

Luke-zhang-04:
BTW this wouldn't fix https://github.com/prisma/prisma/issues/9435#issuecomment-946414969, but I feel like that's a bit of a special case. Maybe there should be something to check for a `schema.prisma` file?

aqrln:
It seems that some tests are now failing on CI, presumably due to changes in `generateInFolder`. Oddly, I have a hard time reproducing those failures locally. Can you reproduce them on your environment?

aqrln:
> Regarding the esm stuff, I'm a bit confused since static imports are used in tall the other tests, and require doesn't mix well with import.

Dynamic imports with `require` (and now also with the ESM dynamic import) are what's used in all tests that generate the client on the disk, there's no location to statically import from since the test creates the files that are going to be imported. Regarding `require` not mixing well with ESM, well, in our case it's not an issue since these aren't actually real ES modules, the test files are transpiled down to CommonJS anyway.

aqrln:
> BTW this wouldn't fix #9435 (comment), but I feel like that's a bit of a special case. Maybe there should be something to check for a schema.prisma file?

That's a good point. I would say it might be out of the scope of this PR, but generally needs to be fixed as well. Maybe this PR shouldn't close but only refer to that issue then. Alternatively, that specific comment may be converted to a separate issue.

Luke-zhang-04:
I can't reproduce it either. Maybe it's because I have something already generated, but I don't think the changes to `generateInFolder` should've broken anything.

Luke-zhang-04:
?????????????????????????????? I got it to work https://github.com/Luke-zhang-04/prisma/runs/4819131488 by removing my test so that's definitely it

Luke-zhang-04:
I reproduced the problem. You have to run my test before the blog-dot-env tests and then it will fail.

janpio:
> Alternatively, that specific comment may be converted to a separate issue. It's closely related, but at the same time it's a different bug (and both of them were uncovered at the same time by fixing yet another bug in v3.1.1, which used to mask these). WDYT 

Yes turn this into a new issue please as it will require a separate fix.

Luke-zhang-04:
@aqrln have any ideas? I'm gonna try to take a look again tonight but I suspect this requires more knowledge on how the testing system works.

Luke-zhang-04:
Ok so it seems like I've "fixed it". The solution is less than ideal though, since it's hardcoded, but it's the best I can do.

Luke-zhang-04:
@millsp any plans to merge this soon? Is there anything else I have to do?

I think the failing tests are just flaky or something cause they're just timing out and I didn't touch that part.

Jolg42:
The no-docker job can indeed have flaky tests.

This fail on Windows looks relevant to your changes, and I'm unsure if it's flaky 🤔 
https://github.com/prisma/prisma/runs/4958676029?check_suite_focus=true#step:11:1502
```
FAIL src/__tests__/integration/happy/output-directory/test.ts
  × connect to client generated 1 directory level (13 ms)

  ● connect to client generated 1 directory level

    Cannot find module '..........
    untime/index' from 'src/__tests__/integration/happy/output-directory/prismaClient/index.js'

    Require stack:
      src/__tests__/integration/happy/output-directory/prismaClient/index.js
      src/__tests__/integration/happy/output-directory/test.ts

      15 |   raw,
      16 |   Decimal
    > 17 | } = require('..\..\..\..\..\runtime/index')
         |     ^
      18 |
      19 |
      20 | const Prisma = {}

      at Resolver.resolveModule (../../node_modules/.pnpm/jest-resolve@27.4.6/node_modules/jest-resolve/build/resolver.js:324:11)
      at Object.<anonymous> (src/__tests__/integration/happy/output-directory/prismaClient/index.js:17:5)
```
I just started a rerun to check later.

Thanks for your efforts here, I'll ping my colleagues, though I don't think it would be merged for the release today as it's a change we need to be confident with to avoid a regression. And for that I'll let my colleagues investigate and see what's needed.

Jolg42:
Wndows shows the same issue after a rerun
https://github.com/prisma/prisma/runs/5018382803?check_suite_focus=true#step:11:1568

So this is not flaky and will need to be investigated

packages/client/src/utils/generateInFolder.ts
```diff
@@ -70,7 +74,11 @@ export async function generateInFolder({
         overwrite: true,
       })
     } else {
-      await getPackedPackage('@prisma/client', outputDir)
+      await getPackedPackage(
+        '@prisma/client',
+        outputDir,
+        process.cwd().endsWith('output-directory') ? path.join(__dirname, '..', '..') : undefined,
```
aqrln, line 80: Does it mean that it actually depends on the `relativeOutputDir` maybe?
Either way, I'd prefer to not hardcode the specific test name here.

Luke-zhang-04: Yeah same, I'll keep trying things I guess.

Luke-zhang-04: I think my new implementation is better?

Luke-zhang-04:
I'm not familiar with Windows, so I don't have a clue as to why it's failing.

janpio:
(Quickly fixed a merge conflict so tests run again)

janpio:
The failing Windows tests:
```
FAIL src/__tests__/integration/happy/output-directory/test.ts
  × connect to client generated 1 directory level (12 ms)

  ● connect to client generated 1 directory level

    Cannot find module '..........
    untime/index' from 'src/__tests__/integration/happy/output-directory/prismaClient/index.js'

    Require stack:
      src/__tests__/integration/happy/output-directory/prismaClient/index.js
      src/__tests__/integration/happy/output-directory/test.ts

      15 |   raw,
      16 |   Decimal
    > 17 | } = require('..\..\..\..\..\runtime/index')
         |     ^
      18 |
      19 |
      20 | const Prisma = {}

      at Resolver.resolveModule (../../node_modules/.pnpm/jest-resolve@27.5.1/node_modules/jest-resolve/build/resolver.js:324:11)
      at Object.<anonymous> (src/__tests__/integration/happy/output-directory/prismaClient/index.js:17:5)
```
and
```
FAIL src/__tests__/integration/happy/output-directory/test.ts
  × connect to client generated 1 directory level (12 ms)

  ● connect to client generated 1 directory level

    Cannot find module '..........
    untime/index' from 'src/__tests__/integration/happy/output-directory/prismaClient/index.js'

    Require stack:
      src/__tests__/integration/happy/output-directory/prismaClient/index.js
      src/__tests__/integration/happy/output-directory/test.ts

      15 |   raw,
      16 |   Decimal
    > 17 | } = require('..\..\..\..\..\runtime/index')
         |     ^
      18 |
      19 |
      20 | const Prisma = {}

      at Resolver.resolveModule (../../node_modules/.pnpm/jest-resolve@27.5.1/node_modules/jest-resolve/build/resolver.js:324:11)
      at Object.<anonymous> (src/__tests__/integration/happy/output-directory/prismaClient/index.js:17:5)
```

The `..........(line break)untime` when the code says `'..\..\..\..\..\runtime/index'` indicates that we are somewhere creating "paths" in a way that does not work for Windows - or is different on Windows as it is using `\` instead of the expected `/`. Often that happens when one uses a function that is meant to create file paths (where Windwos uses `\` instead of `/`) but the code actually always expects forward slashes. Usually there is another method that always outputs Linux separators to avoid that problem.

Subtle bug, but pretty common when first migrating things to Windows.

packages/client/src/utils/generateInFolder.ts
```diff
@@ -70,7 +74,17 @@ export async function generateInFolder({
         overwrite: true,
       })
     } else {
-      await getPackedPackage('@prisma/client', outputDir)
+      try {
+        await getPackedPackage('@prisma/client', outputDir)
+      } catch (err) {
+        try {
+          // Fallback to packages/client/package.json
+          await getPackedPackage('@prisma/client', outputDir, path.join(__dirname, '..', '..'))
```
janpio, line 82: This might be a candidate for the Windows failing test. This uses `path.join` which probably uses a OS depending separator, but this really is a "package path" which possibly should always use `/` or something like that?

Luke-zhang-04: oh interesting, I'll give that a shot.

Luke-zhang-04:
looks like that didn't change anything

janpio:
Yep, seems these paths come from somewhere else, nothing changed there:
```
 connect to client generated 1 directory level

    Cannot find module '..........
    untime/index' from 'src/__tests__/integration/happy/output-directory/prismaClient/index.js'

    Require stack:
      src/__tests__/integration/happy/output-directory/prismaClient/index.js
      src/__tests__/integration/happy/output-directory/test.ts

      15 |   raw,
      16 |   Decimal
    > 17 | } = require('..\..\..\..\..\runtime/index')
```

But I am pretty sure that is the underlying problem.

Luke-zhang-04:
@janpio can you try fixing this? I have no idea how to go about it.

Luke-zhang-04:
Seems like that did something, I'm getting new errors now

```
 FAIL src/__tests__/integration/happy/output-directory/test.ts
  ● connect to client generated 1 directory level

    Cannot find module '....alidator
    untime/index' from 'src/__tests__/integration/happy/output-directory/prismaClient/index.js'

    Require stack:
      src/__tests__/integration/happy/output-directory/prismaClient/index.js
      src/__tests__/integration/happy/output-directory/test.ts

      15 |   raw,
      16 |   Decimal
    > 17 | } = require('..\..\validator\runtime/index')
         |     ^
      18 |
      19 |
      20 | const Prisma = {}

      at Resolver.resolveModule (../../node_modules/.pnpm/jest-resolve@27.5.1/node_modules/jest-resolve/build/resolver.js:324:11)
      at Object.<anonymous> (src/__tests__/integration/happy/output-directory/prismaClient/index.js:17:5)
```

janpio:
That looks like the same error to me, probably just coming from some other path creation. Something needs to be similar.

(Yay figuring out that it is the `posix` namespace - that was exactly the one I meant and could not remember)

Luke-zhang-04:
Yeah that's what I meant, different location. I'll fix the rest later.

millsp:
Hey @Luke-zhang-04, sorry to have stepped over, I have resolved the issue via this [PR](https://github.com/prisma/prisma/pull/12000) and added the missing [e2e test](https://github.com/prisma/e2e-tests/pull/2488) for Yarn with PnP. I'm therefore closing your PR but wanted to let you know how thankful we are for your contribution. Your investigative work and motivation has definitely led to the resolution of this issue. So thank you :partying_face:!

PR #10849, opened 2021-12-23
test(*): tweak timeouts

Jolg42:
I could see it failing here https://github.com/prisma/prisma/runs/4618230951?check_suite_focus=true#step:9:795

And I remember I saw it recently as well.

### DIFF
--- a/packages/sdk/src/__tests__/getPackedPackage.test.ts
+++ b/packages/sdk/src/__tests__/getPackedPackage.test.ts
@@ -4,6 +4,8 @@ import fs from 'fs'
 
 describe('getPackedPackage', () => {
   it('test argument vulnerability', async () => {
+    jest.setTimeout(10000)
+    
     const outputDir = '/tmp/some-prisma-target-folder'
     const packageDir = 'foo`touch /tmp/getPackedPackage-exploit`'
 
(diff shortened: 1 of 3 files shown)


packages/sdk/src/__tests__/getPackedPackage.test.ts
```diff
@@ -4,6 +4,8 @@ import fs from 'fs'
 
 describe('getPackedPackage', () => {
   it('test argument vulnerability', async () => {
+    jest.setTimeout(10000)
```
aqrln, line 7: Can `jest.setTimeout` be invoked inside a test? I thought it's only supposed to be called globally before the tests, and for individual tests there's the third argument to `it`/`test`, like this:

```ts
it('test argument vulnerability', async () => { }, 10000)
```

(Not saying `jest.setTimeout(10000)` won't work inside the test though, I never tried doing this)

Jolg42: We use it in both cases and it works the same, though I also like the third argument we don't have a guideline right now which one we should use and if we should use the same one everywhere.

aqrln (approved):

PR #10969, opened 2022-01-05
chore: improve TESTING.md about pnpm run test

Jolg42:
Thanks to @garrensmith for noticing this

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -41,15 +41,17 @@ direnv allow
 
 1. We use the [Jest test framework](https://jestjs.io/). Its CLI is powerful and removes the need for npm scripts mostly. For most cases this is what you need to know:
 
-   ```sh
-   pnpm run jest <fileNamePattern> -t <testNamePattern>
-   ```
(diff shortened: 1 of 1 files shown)


garrensmith (approved):
That is really helpful. Thank you.

TESTING.md
```diff
@@ -41,15 +41,17 @@ direnv allow
 
 1. We use the [Jest test framework](https://jestjs.io/). Its CLI is powerful and removes the need for npm scripts mostly. For most cases this is what you need to know:
 
-   ```sh
-   pnpm run jest <fileNamePattern> -t <testNamePattern>
-   ```
+Note: the following command `pnpm run test` can be used inside the packages folders like `packages/client`. In the base folder you can only run `pnpm run test` without extra arguments.
```
aqrln, line 44: FWIW, this is not entirely true, you can do something like:

```sh
pnpm --filter @prisma/client test -- <jest arguments>
```

Jolg42: Indeed! Nice one to put here I guess

PR #10978, opened 2022-01-05
chore(peer-dep): missing 'encoding' package in @prisma/cli

belgattitude:
> Not sure about this one, so feel free to ask for changes

In the generated output from `prisma/cli`, there's an *optional* dependency on [encoding](https://www.npmjs.com/package/encoding), see note below.

I got an error with yarn 3 and pnpm module linker in strict mode. The module 'encoding' could not be resolved. 

It seems legit while being picky, so I've added it as peerDependency. 

Should work on all pm (yarn, npm, pnpm)... newest npm will install it by default.

## Note

For example look at the `packages/cli/build/child.js`:

```js
let convert;
try {
  convert = require('encoding').convert;
} catch (e) {}

// ...and later on

function convertBody(buffer, headers) {
  if (typeof convert !== 'function') {
    throw new Error('The package `encoding` must be installed to use the textConverted() function');
  }
}
```

## Workaround

For yarn 3.1+ pnpm linker and probably pnp, it's possible to add the deps in the yarnrc.yml

```
packageExtensions:
  "prisma@*":
    peerDependencies:
      encoding: 0.1.13
```

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -132,6 +132,9 @@
   "dependencies": {
     "@prisma/engines": "3.8.0-26.367f1321e79a7ba33fcac73ed17111312b87e526"
   },
+  "peerDependencies": {
+    "encoding": "^0.1.13"
+  },
   "lint-staged": {
     "*.ts": [
       "eslint",


aqrln (changes requested):
Hi @belgattitude! Thank you for your contribution, and thanks for noticing this. While it is a good catch, I don't think this is the right way to solve the problem for the following reasons:

1. `peerDependencies` aren't really appropriate for optional dependencies, they are supposed to be required, but have different semantics and installed differently from the regular dependencies.
2. `encoding` is not a direct dependency of Prisma, optional or not, this code comes from some of our deps in the bundle. We'll also have no way to know if/when it's no longer necessary.
3. If this package is actually needed to run Prisma in some cases, it should be bundled rather than installed as a `dependency`/`peerDependency`.

The right way to approach this would be to investigate if the code path that requires this package can be reached at all, and depending on that either add the dependency to the bundle, or add an empty stub for that package using `fillPlugin`.

Jolg42:
I'll close this PR (see previous comment).

If you could open an issue, it will be the best way to have all the details needed for a reproduction.

After a quick check it seems it's only used like this in the source code, from node-fetch,
```
let convert;
try {
	convert = require('encoding').convert;
} catch (e) {}
```

node-fetch is used to send the optional panic error reports and download the engines binaries.

aqrln:
@Jolg42 I believe in this case the issue was with Yarn 3 and not webpack

belgattitude:
I was wrong guys :smile: 

While it seems true that the generated client might use the js based [encoding](https://www.npmjs.com/package/encoding) library if present, the package manager does not really care (no warn). 

I've tried here https://github.com/belgattitude/nextjs-monorepo-example/pull/1140 and all works.

Sorry for the noise and keep up the good work.

Issue #10747, opened 2021-12-17
Integrate better engine errors into TS client


dpetrick:
Follow up to PR: https://github.com/prisma/prisma-engines/pull/2451

With the engine part merged, it's time for the TS client to integrate it.

janpio:
Please open a new issue @SynergyEvolved - this is an issue about an internal improvement we want to get out. We can not understand your problem in here. Thanks.

Issue #5040, opened 2021-01-13
Help developers better handle SQL errors


Sytten:
## Problem
We sometimes want to handle SQL errors differently to return a proper error to the frontend. For example, we might want to display `email already used` if a unique violation is detected on the field. The current situation is that we have to check against an obscure error code (`P2002` for example for unique validation failure) and then go into `meta.target` to get which field broke. Its never easy to extract meaningful stuff from ORM(-like) tools despite being a very common operation and I think it could be a factor of differentiation for prisma.

## Propositions
I see two possible way of dealing more gracefully with that:
1. Create more custom error types in javascript and use `instanceof` to catch them. This would look like:
```javascript
  try {
    const user = await prisma.user.create({ data });
    return res.status(204).send();
  } catch (err) {
    if (err instanceof PrismaUniqueViolationError) {
      return res.status(400).send('DUPLICATE EMAIL');
    }
    return res.status(400).send('BAD REQUEST');
  }
```
2. Create an enum of errors. This would look like
```javascript
  try {
    const user = await prisma.user.create({ data });
    return res.status(204).send();
  } catch (err) {
    if (err.code === PrismaErrors.UniqueViolation) {
      return res.status(400).send('DUPLICATE EMAIL');
    }
    return res.status(400).send('BAD REQUEST');
  }
```

I also think there should be helpers to extract the problematic fields from the error. Something like:
```javascript
if (err.code === PrismaErrors.UniqueViolation) {
    const duplicateFields = extractErrorFields(err);
    return res.status(400).json({ code: 'unique_violation', fields: duplicateFields});
}
```

brunocasado:
> If possible, I'd like to have something like this which I've initially proposed in [prisma/prisma-client-js#720](https://github.com/prisma/prisma-client-js/issues/720):
> 
> ```ts
> try {
>   await db.users.insert(newUser)
> } catch (err) {
>   if (err instanceof UniqueConstraintViolation) {
>     // not sure if typescript can infer the error, but I think it can
>     if (err.field == "email") {
>       console.log("user already exists with this email")
>       return
>     }
>     console.log("unique constraint violation on another field: " + err.field)
>   }
> }
> ```

So, have we some solutions close to the above? err.field would be great

vinpac:
I published a package to handle this feature https://github.com/vinpac/prisma-error-enum.

I can make a PR to integrate this into prisma. Should I?

```typescript
export const CommonError = {
  /**
[...]

PR #10996, opened 2022-01-07
fix: modify PrismaClientKnownRequestError type from object

dkantereivin:
I didn't open an issue (though tried to raise the point on Slack), but I figure this PR is self-explanatory.

The `meta` property is currently typed as `object` but this seems incorrect since the docs outline many instances where the value of `meta` is not `{}` (for example, [here](https://www.prisma.io/docs/reference/api-reference/error-reference#prismaclientknownrequesterror)). Thus, `Record<string, unknown>` or a more specific type (which is currently undocumented) seems more appropriate, that way things such as `error.meta.target` can be used without a type-error.

If I'm totally wrong, this can of course be closed.


aqrln (approved):
Hi @dkantereivin and sorry for the delay!

This looks good to me, I agree it's a more appropriate/convenient type.

Regarding declaring a more specific type here, it's not trivial because it depends on the specific error and comes from the engine, not from the TypeScript side. I can't provide a timeline for this, but this is something we are working on and will have in the future. These are the relevant issues: https://github.com/prisma/prisma/issues/5040, https://github.com/prisma/prisma/issues/10747.

PR #11003, opened 2022-01-07
chore(migrate): Use the migration engine JSON-RPC printing API

tomhoule:
https://github.com/prisma/prisma-engines/pull/2566
todo: explain better

Note:
https://www.jsonrpc.org/specification


packages/migrate/src/MigrateEngine.ts
```diff
@@ -139,15 +140,11 @@ export class MigrateEngine {
           this.listeners[result.id](result)
           delete this.listeners[result.id]
         }
-      } else {
```
tomhoule, line 142: Deleted because I'm confident this code path is never taken, based on what the ME returns.

Jolg42:
@tomhoule Let's sync about this tomorrow?

tomhoule:
yep, good idea, let's do that :+1:

packages/migrate/src/MigrateEngine.ts
```diff
@@ -136,6 +136,7 @@ export class MigrateEngine {
     } catch (e) {
       console.error(`Could not parse migration engine response: ${response.slice(0, 200)}`)
     }
+
     if (result) {
       if (result.id) {
```
Jolg42, line 141:
```suggestion
      // It's a response
      if (result.id) {
```

packages/migrate/src/MigrateEngine.ts
```diff
@@ -145,15 +146,11 @@ export class MigrateEngine {
...
-        if (result.is_panic) {
-          throw new Error(`Response: ${result.message}`)
-        } else if (result.message) {
-          console.error(chalk.red(`Response: ${result.message}`))
-        } else {
-          console.error(chalk.red(`Response: ${JSON.stringify(result)}`))
-        }
+      } else if (result.method) {
+          // This is a notification.
+          if (result.method === "print" && Boolean(result.params?.content)) {
```
Jolg42, line 151:
```suggestion
          if (result.method === "print" && result.params?.content) {
```

packages/migrate/src/MigrateEngine.ts
```diff
@@ -145,15 +146,11 @@ export class MigrateEngine {
...
+              process.stdout.write(result.params?.content)
```
Jolg42, line 152:
```suggestion
              process.stdout.write(result.params.content)
```

packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -435,6 +435,7 @@ describe('sqlite', () => {
       The migration \`20201231000000_test\` was modified after it was applied.
 
       Applying migration \`20201231000000_test\`
+
       Applying migration \`20201231000000_draft\`
```
Jolg42, line 439: @tomhoule this doesn't look good

packages/migrate/src/commands/MigrateDiff.ts
```diff
@@ -244,6 +244,13 @@ ${chalk.bold('Examples')}
         shadowDatabaseUrl: args['--shadow-database-url'],
       })
     } finally {
+      // We need to wait for the "notification(s)" sent by the engine via JSON-RPC to be printed to console.info
+      // In Jest it manifests also as the following if we don't wait:
+      // Cannot log after tests are done. Did you forget to wait for something async in your test?
+      //
+      // we need to return the value in the response then we can remove this
+      await new Promise((resolve) => setTimeout(resolve, 50))
```
Jolg42, line 252: @tomhoule we really need to clean this up soon and pass the diff in the response 🙈

packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -629,14 +629,14 @@ describe('sqlite', () => {
...
+                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    ⚠️ We found changes that cannot be executed:
```
tomhoule, line 632: Is this expected? I'm a bit puzzled

packages/migrate/src/commands/MigrateDiff.ts
tomhoule: I am pretty sure asynchronously waiting here is not necessary. What we did before is send the notifications and the response on different channels (stderr and stdout), but with the companion PR that is already merged, the print notification is guaranteed to come before the response on stdout, so if the ordering of operations in the CLI matches the ordering of the JSON-RPC messages on stdout, there is no need for extra logic.

If we want to impose more synchronization and delays (e.g. if the log call and the response are reordered by the event loop), we could make the notification a request — the CLI would then send a response (an ACK), and the engine would wait for that before sending its own response. Long term, I think we want to use this mechanism a lot more, not less, because it avoids spreading control flow across both sides of the RPC boundary.

packages/migrate/src/commands/MigrateDeploy.ts
```diff
@@ -121,6 +121,9 @@ ${editedMigrationNames.join('\n')}`,
       const { appliedMigrationNames } = await migrate.applyMigrations()
       migrationIds = appliedMigrationNames
     } finally {
+      // We need to wait for the "notification(s)" sent by the engine via JSON-RPC to be printed to console.info
+      await new Promise((resolve) => setTimeout(resolve, 50))
+      // Stop engine
```
tomhoule, line 126: (see other comment, but I think it's the same here)

packages/migrate/src/commands/MigrateDev.ts
```diff
@@ -175,6 +175,9 @@ ${chalk.bold('Examples')}
       const { appliedMigrationNames } = await migrate.applyMigrations()
       migrationIdsApplied.push(...appliedMigrationNames)
 
+      // We need to wait for the "notification(s)" sent by the engine via JSON-RPC to be printed to console.info
+      await new Promise((resolve) => setTimeout(resolve, 50))
```
tomhoule, line 179: same here

tomhoule (commented):
Since these waits are not nice at all, let's find a solid solution instead. Can you confirm they are really needed with an up-to-date version of engines?

packages/migrate/src/__tests__/MigrateDev.test.ts
Jolg42: Jest snapshots do that for a reason we don't completely understand.
You can ignore

packages/migrate/src/commands/MigrateDiff.ts
millsp: Thanks, that makes a lot of sense. Hopefully, this will resolve most of the flakyness we see today :100:

Jolg42: @tomhoule the tests fail when we do not wait, see
https://github.com/prisma/prisma/runs/5182960010?check_suite_focus=true

Jolg42 (approved):
Thanks for the comments,
Indeed, we can improve the communication of CLI - engine here.

I will merge this now to unblock the pipeline.

PR #11030, opened 2022-01-08
docs: fix typos in docs and comments

hyochan:
While playing around with Prisma, I see some typos and fixed them.


packages/cli/src/utils/prompt/utils/deepExtend.ts
```diff
@@ -1,5 +1,5 @@
 // Taken from https://github.com/unclechu/node-deep-extend/blob/master/lib/deep-extend.js
-// es2017-ified by Tim Suchanek, now it's about 2.5 times faster
+// es2017-Simplified by Tim Suchanek, now it's about 2.5 times faster
```
janpio, line 2: I think this was intentional - it should mean "made es2017"

janpio (approved):

packages/cli/src/utils/prompt/utils/deepExtend.ts
hyochan: Oh, my bad 😅  ~!  Reverted it!

millsp (approved):

PR #11043, opened 2022-01-10
feat: bail in generate when Data Proxy is used with Interactive Transactions

ghost:


Jolg42 (commented):
I don't think it's always valid to assume that someone is using the Data Proxy (with Interactive transactions) if the following previewFeatures flags are found `'dataProxy', 'interactiveTransactions'`

A user could have these flags and actually not use the Data Proxy nor Interactive Transactions

cebuladev:
If they don't use them, then removing the flag shouldn't be an issue. I know this may be slightly breaking but breaking on the build step, not in runtime, and given the current usage we think that it's better to bail and communicate there rather than in the runtime.

packages/sdk/src/get-generators/utils/check-feature-flags/forbiddenTransactionsWithProxyFlagMessage.ts
```diff
@@ -0,0 +1,20 @@
+import chalk from 'chalk'
+import { highlightDatamodel } from '../../../highlight/highlight'
+import { link } from '../../../link'
+
+export const forbiddenTransactionsWithProxyFlagMessage = `\nInteractive Transactions are not supported in the Data Proxy (${chalk.green(
+  'dataProxy',
+)} feature flag).
+Skip the ${chalk.red('interactiveTransactions')} flag:
```
Jolg42, line 8:
```suggestion
Remove ${chalk.red('interactiveTransactions')} from previewFeatures, example:
```

packages/sdk/src/get-generators/utils/check-feature-flags/forbiddenTransactionsWithProxyFlagMessage.ts
```diff
@@ -0,0 +1,20 @@
...
+
+${chalk.bold(
+  highlightDatamodel(`generator client {
+    provider = "prisma-client-js"
+    previewFeatures = ["dataProxy"]
+}`),
+)}
+`
+
+// TODO
```
Jolg42, line 18: Todo?

cebuladev: It's a draft PR - we're still agreeing on the communication, there will be a link. Update coming soon.

cebuladev:
@Jolg42 updated to incorporate your suggestions and Alberto's input. PTAL :pray:

Jolg42:
Nice one!

Could you add a small test, example https://github.com/prisma/prisma/blob/main/packages/sdk/src/__tests__/getGenerators/getGenerators.test.ts#L615

millsp (approved):

Jolg42 (approved):
Thanks, nice one!

PR #11117, opened 2022-01-12
test(cli): add studio tests for schema with a custom filename

eemmiillyy:


packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -48,7 +48,7 @@ describe('studio', () => {
   })
 
   afterAll(async () => {
-    await studio.instance!.stop()
+    studio.instance!.stop()
```
Jolg42, line 51: A bit surprised by the removal of the await 🤔 

Is that intended?

Jolg42 (approved):
Looks good, thanks! 💚

packages/cli/src/__tests__/commands/Studio.test.ts
eemmiillyy: Yeah, the function is not an async promise so the `await` didn't have any effect.

Issue #10896, opened 2021-12-29
Snapshot serializer in CLI tests doesn't replace time durations >= 1s with a placeholder


aqrln:
Snapshot serializer in CLI tests doesn't replace time durations greater than or equal to 1s with a placeholder.

Failed pipeline: https://github.com/prisma/prisma/runs/4657927865?check_suite_focus=true#step:15:662

```
  ● --schema from parent directory › --schema absolute path: should work

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `--schema from parent directory --schema absolute path: should work 1`

    - Snapshot  - 1
    + Received  + 1

      ↵
    - ✔ Generated Prisma Client (0.0.0 | TEST_ENGINE_TYPE) to ./subdirectory/@prisma/client in XXms
    + ✔ Generated Prisma Client (0.0.0 | TEST_ENGINE_TYPE) to ./subdirectory/@prisma/client in 3.34s
      You can now start using Prisma Client in your code. Reference: https://pris.ly/d/client
```

Relevant code: https://github.com/prisma/prisma/blob/c1fdadf96c27509accf4e8d5c0c3ab91ff68afd4/packages/cli/src/__tests__/__helpers__/snapshotSerializer.ts#L37-L39

Jolg42:
It does now so I'll close 

https://github.com/prisma/prisma/blob/main/packages/sdk/src/utils/jestSnapshotSerializer.js#L74

aqrln:
Thanks!

Jolg42:
Thanks! added in https://github.com/prisma/prisma/pull/11123

PR #11123, opened 2022-01-12
test: attempt to fix flaky tests

Jolg42:
4.XXs normalization
https://github.com/prisma/prisma/runs/4790799556?check_suite_focus=true#step:15:770

and

test argument vulnerability

### DIFF
--- a/packages/cli/src/__tests__/commands/Generate.test.ts
+++ b/packages/cli/src/__tests__/commands/Generate.test.ts
@@ -134,7 +134,7 @@ describe('--schema from parent directory', () => {
 function cleanSnapshot(str: string): string {
   return str
     .replace(/\d+ms/g, 'XXms')
-    .replace(/\d+s/g, 'XXms')
+    .replace(/\d+(?>\.\d+)?s/g, 'XXms')
     .replace(/\(version:.+\)/g, '(version: 0.0.0)')
     .replace(new RegExp(getClientEngineType(), 'g'), 'TEST_ENGINE_TYPE')
 }
--- a/packages/cli/src/__tests__/commands/Generate.test.ts
+++ b/packages/cli/src/__tests__/commands/Generate.test.ts
@@ -134,7 +134,7 @@ describe('--schema from parent directory', () => {
 function cleanSnapshot(str: string): string {
   return str
     .replace(/\d+ms/g, 'XXms')
-    .replace(/\d+s/g, 'XXs')
+    .replace(/\d+s/g, 'XXms')
     .replace(/\(version:.+\)/g, '(version: 0.0.0)')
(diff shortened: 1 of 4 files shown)


packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -134,7 +134,7 @@ describe('--schema from parent directory', () => {
 function cleanSnapshot(str: string): string {
   return str
     .replace(/\d+ms/g, 'XXms')
-    .replace(/\d+s/g, 'XXs')
+    .replace(/\d+s/g, 'XXms')
```
millsp, line 137: I think that we had `4.XXs` because that's what the regex matched, it was not designed to match full decimal numbers.
```suggestion
    .replace(/\d+(?>.\d+)?s/g, 'XXms')
```

millsp (changes requested):

packages/cli/src/__tests__/commands/Generate.test.ts
aqrln, line 137: The problem with this function is it's not even called in all but two tests. Other tests rely on the snapshot serializer replacing the milliseconds, so even with this change some tests will sometimes fail: https://github.com/prisma/prisma/issues/10896

Jolg42: I closed the issue since my cleanup refactor PR adressed that here

Jolg42: Indeed, thanks!

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -134,7 +134,7 @@ describe('--schema from parent directory', () => {
 function cleanSnapshot(str: string): string {
   return str
     .replace(/\d+ms/g, 'XXms')
-    .replace(/\d+s/g, 'XXs')
+    .replace(/\d+(?>\.\d+)?s/g, 'XXms')
```
aqrln, line 137: Do we still need these two replacements here? I think we can just remove them and update the corresponding snapshots to use `XXXms` instead of `XXms`, since both replacements are now handled by the snapshot serializer.

packages/cli/src/__tests__/commands/Generate.test.ts
Jolg42: It's an invalid regex, I fixed it in a new commit

packages/cli/src/__tests__/commands/Generate.test.ts
Jolg42: I completely removed that code since it's centralized in jestSnapshotSerializer.js

aqrln (approved):

millsp (changes requested):

millsp (approved):

PR #11156, opened 2022-01-14
chore: 3.8.x

janpio:


helpers/compile/build.ts
```diff
@@ -4,14 +4,15 @@ import { flatten } from '../blaze/flatten'
 import { pipe } from '../blaze/pipe'
 import { map } from '../blaze/map'
 import { transduce } from '../blaze/transduce'
-import glob from 'glob'
+import glob from 'globby'
```
janpio, line 7: Was this import broken before?

helpers/compile/plugins/depCheckPlugin.ts
```diff
@@ -94,25 +90,13 @@ const unusedPlugin: esbuild.Plugin = {
...
-      ignore: ['**/packages/**/*', '**/*.d.ts'],
-      gitignore: true,
-    }),
-    logLevel: 'silent', // there will be errors
-    bundle: true, // we bundle to get everything
-    write: false, // no need to write for analysis
-    outdir: 'out',
-    plugins: [unusedPlugin],
-  })
-  .catch(() => {})
```
janpio, line 118: Where does this logic go?

packages/debug/package.json
```diff
@@ -52,7 +50,8 @@
   ],
   "dependencies": {
     "@types/debug": "4.1.7",
-    "ms": "2.1.3"
+    "ms": "2.1.3",
+    "strip-ansi": "6.0.1"
```
janpio, line 54: Is this really a normal dep? Where is it used, or waht for?

packages/migrate/package.json
```diff
@@ -57,12 +54,16 @@
...
+    "chalk": "4.1.2",
     "execa": "5.1.1",
     "has-yarn": "2.1.0",
     "indent-string": "4.0.0",
     "log-update": "4.0.0",
+    "mariadb": "2.5.5",
+    "mssql": "7.3.0",
     "new-github-issue-url": "0.2.1",
     "open": "7.4.2",
+    "pg": "8.7.1",
```
janpio, line 66: Are these really used outside of tests?

millsp: Same as https://github.com/prisma/prisma/pull/11156#discussion_r784860441

helpers/compile/plugins/depCheckPlugin.ts
millsp: It moved to `prisma/helpers/compile/build.ts`

helpers/compile/build.ts
millsp: Although they are compatible, `globby` had a gitignore option. Thinking about it, it is not even needed anymore, so I could revert that in a later PR.

PR #11235, opened 2022-01-18
[do not merge]  refactor:remove eval

soilSpoon:
I'm thinking of removing eval using [external](https://esbuild.github.io/api/#external)

### DIFF
--- a/packages/cli/scripts/install.js
+++ b/packages/cli/scripts/install.js
@@ -14,7 +14,9 @@ if (process.env.INIT_CWD && process.env.NOW_BUILDER) {
 }
 
 async function ensurePostInstall() {
-  const initPkgPath = require('path').resolve(process.env.INIT_CWD, 'package.json')
+  const initPkgPath = eval(
+    `require('path').resolve(process.env.INIT_CWD, 'package.json')`,
+  )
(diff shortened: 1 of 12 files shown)


Jolg42:
So eval is used because of how bundlers work.
The tests might pass here but it would probably break when using webpack and others.

soilSpoon:
> So eval is used because of how bundlers work. The tests might pass here but it would probably break when using webpack and others.

If it works using eval, doesn't it work without it?

It is beyond my knowledge to understand this phenomenon. Do you have any related documents?

pantharshit00:
I think we did this to prevent asset relocation which bundlers do which is basically changing values of `__dirname`, `process.cwd()` by bundlers. I also remember there were some issues with vercel/ncc which required these annotations. 

@Jolg42 should have more information here though.

soilSpoon:
> I think we did this to prevent asset relocation which bundlers do which is basically changing values of `__dirname`, `process.cwd()` and other variables by bundlers. I also remember there were some issues with vercel/ncc which required these annotations.
> 
> @Jolg42 should have more information here though.

Thank you for a detailed description! If I do more research and the problems you mentioned are resolved, would it be good to reflect the modifications?

PR #11267, opened 2022-01-19
feat(client): add mongo raw ops

### LINKED ISSUE #11136: Integrate MongoDB raw queries
Work needed for https://github.com/prisma/prisma-engines/pull/2394

Additionally, we will need to accommodate mongo in our tests for the first time.

millsp:
See https://github.com/prisma/prisma/issues/11136


docker/docker-compose.yml
```diff
@@ -62,14 +37,9 @@ services:
       - MYSQL_ROOT_PASSWORD=root
       - MYSQL_DATABASE=tests
       - MYSQL_USER=prisma
-      - MYSQL_PASSWORD=prisma
-    # https://stackoverflow.com/questions/55559386/how-to-fix-mbind-operation-not-permitted-in-mysql-error-log
-    cap_add:
-      - SYS_NICE # CAP_SYS_NICE
-    volumes:
-      - mysql:/var/lib/mysql
```
Jolg42, line 70: This seems unrelated and was probably needed for some reason, why removing it now?

.buildkite/publish/docker-compose.yml
```diff
@@ -55,10 +56,12 @@ services:
       - postgres
       - postgres_isolated
       - mysql
+      - mysql_isolated
       - mariadb
       - mssql
       - mongo
-      - mongo-seed
+      - mongo_migrate
+      - mongo_seed
```
Jolg42, line 64: Would be good to rename mongo_seed to mongo_migrate_seed maybe because it could be confusing right now

docker/docker-compose.yml
```diff
@@ -1,37 +1,14 @@
...
-  # - TEST_POSTGRES_URI_MIGRATE=postgres://prisma:prisma@localhost:5432/tests-migrate
-  # - TEST_POSTGRES_URI=postgres://prisma:prisma@localhost:5432/tests
-  # - TEST_COCKROACH_URI=postgresql://prisma@localhost:26257/
-  # - TEST_POSTGRES_ISOLATED_URI=postgres://prisma:prisma@localhost:5435/tests
-  # - TEST_MYSQL_URI=mysql://prisma:prisma@localhost:3306/tests
-  # - TEST_MYSQL_ISOLATED_URI=mysql://prisma:prisma@localhost:3307/tests
-  # - TEST_MARIADB_URI=mysql://prisma:prisma@localhost:4306/tests
-  # - TEST_MSSQL_URI=sqlserver://SA:Pr1sm4_Pr1sm4@localhost:1433/master
-  # - TEST_MSSQL_JDBC_URI=sqlserver://mssql:1433;database=master;user=SA;password=Pr1sm4_Pr1sm4;trustServerCertificate=true;
-  # - TEST_MONGO_URI=mongodb://root:prisma@localhost:27017/tests?authSource=admin
```
Jolg42, line 24: Why removing this now?

docker/mongo-replica/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM mongo:5
+
+ENTRYPOINT mongod --port $PORT --replSet rs0 --bind_ip 0.0.0.0 & MONGOD_PID=$!; \
+INIT_REPL_CMD="rs.initiate({ _id: 'rs0', members: [{ _id: 0, host: '$HOST:$PORT' }] })"; \
+INIT_USER_CMD="db.createUser({ user: 'root', pwd: 'prisma', roles: [ 'root' ] })"; \
+until (mongo admin --port $PORT --eval "$INIT_REPL_CMD && $INIT_USER_CMD"); do sleep 1; done; \
+echo "REPLICA SET ONLINE"; wait $MONGOD_PID; \
```
Jolg42, line 7: While it's still fresh, maybe some comments might help here in the future

docker/docker-compose.yml
millsp: I aligned our config on the one from the engines. Volumes aren't needed and this error does not happen anymore.

.buildkite/publish/docker-compose.yml
millsp: Will do

docker/docker-compose.yml
millsp: Replaced by this, to avoid maintaining multiple strings. I can add a comment that for docker it's not `localhost` but `[docker-service-name]`
`# See https://github.com/prisma/prisma/blob/main/TESTING.md#environment-variables`

docker/mongo-replica/Dockerfile
millsp: Will do

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -100,6 +100,29 @@ function executeRawDefinition(this: PrismaClientClass) {
...
+  if (!this.dmmf.mappings.otherOperations.write.includes('runCommandRaw')) {
+    return '' // https://github.com/prisma/prisma/issues/8189
+  }
+
+  return `
+  /**
+   * Executes a raw mongo command and returns the result of it.
+   * @example
+   * \`\`\`
+   * const user = await (prisma as any).$runCommandRaw({
```
Jolg42, line 114:
```suggestion
   * const user = await prisma.$runCommandRaw({
```
`as any` is suspicious here, is it really needed?

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -100,6 +100,29 @@ function executeRawDefinition(this: PrismaClientClass) {
...
 
+function runCommandRawDefinition(this: PrismaClientClass) {
+  // we do not generate `$executeRaw...` definitions if not supported
+  if (!this.dmmf.mappings.otherOperations.write.includes('runCommandRaw')) {
+    return '' // https://github.com/prisma/prisma/issues/8189
+  }
+
+  return `
+  /**
+   * Executes a raw mongo command and returns the result of it.
```
Jolg42, line 111:
```suggestion
   * Executes a raw MongoDB command and returns the result of it.
```

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -100,6 +100,29 @@ function executeRawDefinition(this: PrismaClientClass) {
   $executeRawUnsafe<T = unknown>(query: string, ...values: any[]): PrismaPromise<number>;`
 }
 
+function runCommandRawDefinition(this: PrismaClientClass) {
+  // we do not generate `$executeRaw...` definitions if not supported
```
Jolg42, line 104:
```suggestion
  // we do not generate `$runCommandRaw` definitions if not supported
```

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -707,6 +706,31 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
 
+    /**
+     * Executes a raw command only for MongoDB
+     *
+     * @param command
+     * @returns
+     */
+    $runCommandRaw(command: object) {
+      if (config.activeProvider !== 'mongodb') {
+        throw new PrismaClientValidationError(`The ${config.activeProvider} provider does not support $runCommandRaw`)
```
Jolg42, line 717:
```suggestion
        throw new PrismaClientValidationError(`The ${config.activeProvider} provider does not support $runCommandRaw. Use the mongodb provider.`)
```

Also, is there a test for this?

docker/docker-compose.yml
```diff
@@ -1,37 +1,14 @@
...
 
+# See https://github.com/prisma/prisma/blob/main/TESTING.md#environment-variables
```
Jolg42, line 3:
```suggestion
# For connection urls to the following instances, see
# https://github.com/prisma/prisma/blob/main/TESTING.md#environment-variables
```

packages/client/src/__tests__/integration/happy/blog-env-mongo/test.ts
```diff
@@ -0,0 +1,279 @@
...
+        // @ts-ignore
+        data: {},
+      })
+
+      expect(false).toBe(true) // The line above needs to throw, so this should never be executed, but if it does (aka the line above did not throw, as expected), it will fail the test
+    } catch (e) {
+      expect(e).not.toBeUndefined()
+      expect(e).toBeInstanceOf(PrismaClientValidationError)
+    }
+  })
```
Jolg42, line 77:
```suggestion
  test('can throw validation errors', async () => {
    expect.assertions(2)

    const {
      Prisma: { PrismaClientValidationError },
    } = require('./node_modules/@prisma/client')

    try {
      await prisma.post.create({
        // @ts-ignore
        data: {},
      })
    } catch (e) {
      expect(e).not.toBeUndefined()
      expect(e).toBeInstanceOf(PrismaClientValidationError)
    }
  })
```

Does the same but in a better way I think

packages/client/src/__tests__/integration/happy/exhaustive-schema/__snapshots__/test.ts.snap
```diff
@@ -7228,6 +7307,10 @@ export namespace Prisma {
    * ManyRequired createMany
    */
   export type ManyRequiredCreateManyArgs = {
+    /**
+     * The data used to create many ManyRequireds.
```
Jolg42, line 7311: The pluralize is a bit weird and we removed it from most places but not here apparently, and `many` in front would be sufficient I think to indicate it's more than one.

Jolg42: I guess we would need to remove pluralize from `getArgFieldJSDoc`

packages/client/src/generation/TSClient/Model.ts
```diff
@@ -68,7 +68,9 @@ export class Model implements Generatable {
         throw new Error(`Oops this must not happen. Could not find field ${fieldName} on either Query or Mutation`)
       }
       if (action === 'updateMany' || action === 'deleteMany' || action === 'createMany') {
-        argsTypes.push(new MinimalArgsType(field.args, model, action as DMMF.ModelAction, this.collector))
+        argsTypes.push(new MinimalArgsType(field.args, this.type, action as DMMF.ModelAction, this.collector))
+      } else if (action === 'findRaw' || action === 'aggregateRaw') {
+        argsTypes.push(new MinimalArgsType(field.args, this.type, action as DMMF.ModelAction, this.collector))
```
Jolg42, line 73: Looks like the 2 conditional could be merged in one with this change

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -100,6 +100,29 @@ function executeRawDefinition(this: PrismaClientClass) {
...
+   * const user = await (prisma as any).$runCommandRaw({
+   *   aggregate: 'User',
+   *   pipeline: [{ $match: { name: 'Bob' } }, { $project: { email: true, _id: false } }],
+   *   explain: false,
+   * })
+   * \`\`\`
+   * 
+   * Read more in our [docs](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/raw-database-access).
+   */
+  $runCommandRaw(command: Prisma.InputJsonValue): PrismaPromise<Prisma.JsonObject>;`
```
Jolg42, line 123: I guess `Prisma.InputJsonValue` doesn't help you much creating a raw command here, and we can't do more than that?

packages/client/src/generation/TSClient/jsdoc.ts
```diff
@@ -326,4 +326,37 @@ const { count } = await ${ctx.method}({
...
+    fields: {
+      pipeline: () =>
+        'A sequence of data aggregation operations or stages ${@link https://docs.mongodb.com/manual/reference/command/aggregate/#command-fields MongoDB Docs}.',
+      options: () =>
+        'Additional options that passes to the aggregate command ${@link https://docs.mongodb.com/manual/reference/command/aggregate/#command-fields MongoDB Docs}.',
+    },
+  },
+  findRaw: {
+    body: (ctx) =>
+      `Find zero or more ${ctx.plural} that matches the filter.
```
Jolg42, line 349:
```suggestion
      `Find zero or more ${ctx.plural} that matches the filter.
```
Plural doesn't work on all names, zero or more is probably enough to convey the meaning here with singular?

packages/client/src/generation/TSClient/jsdoc.ts
```diff
@@ -326,4 +326,37 @@ const { count } = await ${ctx.method}({
...
+@param {${getModelArgName(ctx.model.name, ctx.action)}} args - Select which filters you would like to apply.
+@example
+const ${lowerCase(ctx.mapping.model)} = await ${ctx.method}({
+  filter: { age: { $gt: 25 } } 
+})`,
+    fields: {
+      filter: () =>
+        'Specifies the selection filter using query operators ${@link https://docs.mongodb.com/manual/reference/command/find/#command-fields MongoDB Docs}.',
+      options: () =>
+        'Specifies the fields to return in the documents that match the query filter ${@link https://docs.mongodb.com/manual/reference/command/find/#command-fields MongoDB Docs}.',
```
Jolg42, line 359: It's the same doc link for both.
For the 1st one it's easy, as there is a filter field in the table in the docs.
For options, I could not find where to look in the docs

packages/client/src/generation/TSClient/PrismaClient.ts
millsp: The decision came from the [design doc](https://www.notion.so/prismaio/MongoDB-Query-Raw-a01774807b644b7894c10ecd53060718#3d73000c2d96486380b6dbc9a4bdad50), types change between mongo versions. It aligns with the rest of the raw APIs that are untyped too.

packages/client/src/generation/TSClient/jsdoc.ts
millsp: Ha wait, there's a lot of them :sweat_smile: I'll open an issue instead.

millsp: https://github.com/prisma/prisma/issues/11434

packages/client/src/generation/TSClient/jsdoc.ts
millsp: Updated the docs, first link points to the operator docs itself and the second to the options.

millsp:
:shipit:

millsp:
(Will merge after https://github.com/prisma/prisma-engines/pull/2394 has landed)

Jolg42 (approved):
✨ Some people will be really excited to try this!

Well done ✅ 

And 🏅 for the mongodb replica set setup 💚

shash7:
This is very very much needed. There are cases where you need to set a writeconcern on a specific write and there's no way right now to do it with Prisma.

PR #11279, opened 2022-01-20
test(migrate): attempt to get rid of flaky ci macos log 

Jolg42:


Jolg42:
Closing for now

Jolg42:
Happened recently again https://github.com/prisma/prisma/runs/4883407768?check_suite_focus=true#step:13:438 reopening 👯

pnpm-lock.yaml
```diff
@@ -1694,7 +1694,6 @@ packages:
       express: 4.17.2
       untildify: 4.0.0
     transitivePeerDependencies:
-      - '@prisma/client'
```
Jolg42, line 1697: Didn't expect that one again 🤔

millsp: Was this the result of `setup`?

Jolg42: Yes indeed

packages/migrate/src/commands/MigrateDev.ts
```diff
@@ -284,6 +284,12 @@ ${chalk.bold('Examples')}
       migrate.stop()
     }
 
+    // the log from engine "Applying migration `20201231000000_name`"
+    // should normally be logged around here in time, before the next console
+    // in our macOS tests it shows it happening after "Your database is now in sync with your schema."
+    // the following solves the issue
+    await new Promise((resolve) => setTimeout(resolve, 0))
```
millsp, line 291: For me that is a sign that somewhere something is not `await`ed. From what I understand, the display of `Applying migration 20201231000000_first` should happen before https://github.com/prisma/prisma/blob/bb3001ac10eb1ff9e25bafce7bb7568ace9612bb/packages/migrate/src/commands/MigrateDev.ts/#L179 ? I could not find where this info is printed, maybe directly from the engine ?

Jolg42: It's directly from the engine

millsp: That might a possible explanation, I'm interested into digging a bit more. Are the engine messages displayed in an async manner then, from the engine stdout?

Jolg42: The log comes from stderr, see https://github.com/prisma/prisma/blob/main/packages/migrate/src/MigrateEngine.ts#L239:L241

millsp: Do we rely a lot on this? I mean, how much do we print via this and would it be easy not to rely on it?
In this case, it can probably be easily fixed as we have all the info we need to re-display to same thing.

Jolg42: Right now we do no print a lot through this (only this as far as I know) but we want to use it for more printing in the future.
Not sure what's your idea here.

PR #11299, opened 2022-01-21
chore: update graph dependencies

Jolg42:
And updated the dependency as well because its license was incorrect on npm because update not published: long open issue here https://github.com/glejeune/node-graphviz/issues/40


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "private": true,
+  "name": "dev-env",
```
Jolg42, line 3: Forgot to remove, any other name is good if you have suggestions

millsp (approved):

PR #11338, opened 2022-01-23
fix(cli): Updating link anchor to match change in the docs

dusandz:
After I ran the init script in a new project, I noticed that there's a (half-)dead link in the comments of the generated .env file. Link points to the proper page, but it seems that the anchor has changed. This PR updates the link.

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -47,7 +47,7 @@ export const defaultEnv = (
 ) => {
   let env = comments
     ? `# Environment variables declared in this file are automatically made available to Prisma.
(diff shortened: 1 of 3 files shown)


Jolg42 (approved):
Thanks for the updated anchor!

dusandz:
I'm new to the project, a quick question: are contributors allowed to update the test snapshots, or it it something that only maintainers are authorized to do?

Jolg42:
@dusandz If you could update the test snapshots that would be amazing 💚 So we could merge the PR right after that (because tests need the snapshot update).

dusandz:
Sure!

Jolg42:
And thanks for the PR 💚

PR #11371, opened 2022-01-24
feat(db execute): db execute CLI command

Jolg42:
Related: https://github.com/prisma/migrations-team/issues/313

Notes:
Design doc on notion: https://www.notion.so/prismaio/Design-doc-first-batch-of-migrate-primitives-48be767a7adf4c23b3116d1347f218c8


packages/migrate/src/__tests__/DbExecute.test.ts
```diff
@@ -0,0 +1,702 @@
...
+      }
+    })
+  })
+
+  describe('postgresql', () => {
+    const connectionString = (
+      process.env.TEST_POSTGRES_URI_MIGRATE || 'postgres://prisma:prisma@localhost:5432/tests-migrate'
+    ).replace('tests-migrate', 'tests-migrate-db-execute')
+    // TODO remove when engine doesn't validate datasource anymore by default from schema
+    process.env.TEST_POSTGRES_URI_MIGRATE = connectionString
```
Jolg42, line 223: Related to https://github.com/prisma/prisma-engines/pull/2617

> Only remark is to remove the check for the connectivity on the schema url. But that can be a follow up PR.
> 
> We can't do that without CLI work (the CLI relies on the connectivity check), but that's part of a refactoring scheduled for this cycle (I want to make it happen within prisma/migrations-team#309).

packages/migrate/src/__tests__/DbExecute.test.ts
```diff
@@ -0,0 +1,702 @@
...
+
+    // TODO, it's passing?
+    //  it('should fail with P1001 error with unreachable url with --file --url', async () => {
+    it('should pass with --file --url=file:doesnotexists.db', async () => {
+      ctx.fixture('introspection/sqlite')
+
+      fs.writeFileSync('script.sql', sqlScript)
+      const result = DbExecute.new().parse(['--preview-feature', '--url=file:doesnotexists.db', '--file=./script.sql'])
+      await expect(result).resolves.toMatchInlineSnapshot(`Script executed successfully.`)
+    })
```
Jolg42, line 180: Apparently a non-existing SQLite database file doesn't error, I guess it's auto-created. @tomhoule do you see a problem with that?
Maybe it's ok, I think it’s our default SQLite behavior.

packages/migrate/src/__tests__/DbExecute.test.ts
```diff
@@ -0,0 +1,702 @@
...
+      } catch (e) {
+        // TODO: Code not defined?
+        expect(e.code).toEqual(undefined)
+        expect(e.message).toMatchInlineSnapshot(`
+          Error creating a database connection.
+
+
+        `)
+      }
+    })
```
Jolg42, line 668: SQL Server
Here it's interesting to see that there is no error code, I was expecting `P1001` like in PostgreSQL or MySQL
@tomhoule any idea?

packages/migrate/src/types.ts
```diff
@@ -90,6 +90,24 @@ export namespace EngineArgs {
...
+    url: string
+  }
+  export type DbExecuteDatasourceType = DbExecuteDatasourceTypeSchema | DbExecuteDatasourceTypeUrl
+
+  export interface DbExecuteInput {
+    // The location of the live database to connect to.
+    datasourceType: DbExecuteDatasourceType
+    // The input script.
+    script: string
+  }
```
tomhoule, line 110: :+1:

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -0,0 +1,171 @@
...
+        )
+      }
+    }
+
+    let datasourceType: EngineArgs.DbExecuteDatasourceType
+
+    const schemaPath = await getSchemaPath(args['--schema'])
+    if (!schemaPath) {
+      throw new NoSchemaFoundError()
+    }
```
tomhoule, line 133: (relative to previous comment) this error could also be handled in the engine if that's more convenient. As it is we're reading the script file twice. I'm also wondering if the ENOENT expected error message matches the `NoschemaFoundError` here.

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -0,0 +1,171 @@
...
+        tag: 'url',
+        url: args['--url'],
+      }
+    } else {
+      // Execute command(s) to url from schema
+      datasourceType = {
+        tag: 'schema',
+        schema: schemaPath,
+      }
+    }
```
tomhoule, line 147: I'm not sure I understand the control flow here, but doesn't this `else` means we're defaulting to the URL from the prisma schema in the default location? I don't remember if it's written down in the notion doc, but we had consensus that this was too dangerous — `--url` or `--schema` should always be specified.

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -0,0 +1,171 @@
...
+    try {
+      await migrate.engine.dbExecute({
+        script,
+        datasourceType,
+      })
+    } finally {
+      migrate.stop()
+    }
+
+    return `Script executed successfully.`
```
tomhoule, line 162: I don't see where the value is used — it doesn't look like it's printed.

tomhoule (commented):
Looks good to me :+1: The only important question is the one about the behaviour without `--schema` and without `--url`.

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -0,0 +1,171 @@
...
+    // Read stdin
+    if (args['--stdin']) {
+      script = await getStdin()
+      // If input is empty, stop here
+      if (!script) {
+        throw new Error(
+          `--stdin was passed but the input was empty. See \`${chalk.green(
+            getCommandWithExecutor('prisma db execute -h'),
+          )}\``,
+        )
```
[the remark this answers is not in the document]
Jolg42: I'll remove

packages/migrate/src/__tests__/DbExecute.test.ts
```diff
@@ -0,0 +1,701 @@
...
+      ctx.fixture('empty')
+      expect.assertions(2)
+
+      fs.writeFileSync('script.js', '-- empty')
+      try {
+        await DbExecute.new().parse(['--preview-feature', '--file=./script.sql', '--schema=./doesnoexists.schema'])
+      } catch (e) {
+        expect(e.code).toEqual('ENOENT')
+        expect(e.message).toContain('ENOENT')
+      }
```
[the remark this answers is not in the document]
Jolg42: Indeed, I can remove the reading on CLI side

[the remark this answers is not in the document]
Jolg42: @tomhoule 
But how do I pass the script file path to the engine? Right now the only option is to read the script CLI side.
```
  export interface DbExecuteInput {
    // The location of the live database to connect to.
    datasourceType: DbExecuteDatasourceType
    // The input script.
    script: string
  }
```

packages/migrate/src/commands/DbExecute.ts
Jolg42: Indeed needs a change here, it's copy pasted and I wanted to change but looks like I forgot 😅

packages/migrate/src/commands/DbExecute.ts
Jolg42: There are a few conditions before that ensure the params are all passed, see https://github.com/prisma/prisma/pull/11371/files/3265e18552dbc9c588c62cb816a3a6a6a84567a3#diff-cdc4535306baf77165c5d00cadbdab8c9077c5888b3cdd72af6f4b6b8c119b5eR84-R108

packages/migrate/src/commands/DbExecute.ts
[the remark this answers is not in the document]
do4gr: That's an interesting question. Since we only report back that the command ran successfully users might not be aware if they messed up the stdin passing if we don't fail here. 
On the other hand there is nothing wrong with empty input and other shell commands wouldn't fail in that situation either I think.

packages/migrate/src/commands/DbExecute.ts
do4gr: I think if none or both are present there is an error earlier.

packages/migrate/src/commands/DbExecute.ts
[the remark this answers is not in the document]
tomhoule: I also expect most databases will return an error on empty input

packages/migrate/src/__tests__/DbExecute.test.ts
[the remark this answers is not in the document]
tomhoule: Indeed, I still had a previous API in mind where it was a file path or stdin, but now the engine takes the whole script as a string, so all good. Sorry!

packages/migrate/src/commands/DbExecute.ts
Jolg42: It's printed from here https://github.com/prisma/prisma/blob/integration/cli-db-execute/packages/cli/src/bin.ts#L177:L177

packages/migrate/src/__tests__/DbExecute.test.ts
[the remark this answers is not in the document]
Jolg42: No problem :)

packages/migrate/src/__tests__/DbExecute.test.ts
Jolg42: https://github.com/prisma/prisma/issues/11407

packages/migrate/src/commands/DbExecute.ts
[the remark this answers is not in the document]
Jolg42: Actually it only fails on MySQL, see newest commit with tests named `... with empty --file --schema`
https://github.com/prisma/prisma/pull/11371/commits/aec52d434da92c14d9b80f0e1ecaf79d3f7c4d33

packages/migrate/src/commands/DbExecute.ts
Jolg42: Done in https://github.com/prisma/prisma/pull/11371/commits/abb689dfd368428d33f06f3d8c75fc71b6d9e20b

packages/migrate/src/utils/setupMSSQL.ts
```diff
@@ -25,40 +25,57 @@ function getMSSQLConfig(url: string): mssql.config {
...
-      CREATE DATABASE [tests-migrate-shadowdb]
-      CREATE DATABASE [tests-migrate]
+    if (databaseName === 'tests-migrate') {
+      await connection.query(`
+CREATE DATABASE [tests-migrate-shadowdb]
+CREATE DATABASE [tests-migrate]
+    `)
+    } else {
+      await connection.query(`
+CREATE DATABASE [${databaseName}]
```
millsp, line 43: What is this new case used for?

millsp (approved):
Looks great. Well done!

packages/migrate/src/utils/setupMSSQL.ts
Jolg42: I wanted to create a separate database to avoid concurrency issues during tests here

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -36,21 +40,27 @@ ${chalk.bold('Usage')}
...
-    --file   Path to a file. The content will be sent as the script to be executed.
+
+${chalk.italic('Script input, only 1 must be provided:')}
+   --stdin   Use the terminal standard input as the script to be executed
+    --file   Path to a file. The content will be sent as the script to be executed
 
 ${chalk.bold('Examples')}
  
-  Execute the content of a SQL script file to the database URL from schema.
+  Execute the content of a SQL script file to the datasource URL taken from the schema
```
do4gr, line 54: Maybe execute on  or against instead of to?

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -36,21 +40,27 @@ ${chalk.bold('Usage')}
...
   ${chalk.dim('$')} prisma db execute --preview-feature --file ./script.sql --schema schema.prisma
 
-  Execute the SQL script from stdin to the database url.
+  Execute the SQL script from stdin to the datasource URL specified via the \`DATABASE_URL\` environment variable
```
do4gr, line 57: Same here.

packages/migrate/src/commands/DbExecute.ts
Jolg42:
```suggestion
  Execute the content of a SQL script file on the datasource URL taken from the schema
```

packages/migrate/src/commands/DbExecute.ts
Jolg42:
```suggestion
  Execute the SQL script from stdin on the datasource URL specified via the \`DATABASE_URL\` environment variable
```

packages/migrate/src/commands/DbCommand.ts
```diff
@@ -14,16 +14,20 @@ ${chalk.bold('Usage')}
 
   ${chalk.dim('$')} prisma db [command] [options]
 
+${chalk.bold('Flag')}
+
+--preview-feature   Run Preview Prisma commands
+
```
janpio, line 20: Maybe only after commands, but before Examples? Feels less important than Commands.

packages/migrate/src/commands/DbExecute.ts
janpio: Hm, is this hardcoded to `DATABASE_URL` or is it whatever is defined in `schema.prisma`? Could be harded in there, or be called `FOO_BOOP` instead.

packages/migrate/src/commands/DbCommand.ts
Jolg42: I'll do that in the diff PR

Jolg42: https://github.com/prisma/prisma/pull/11440

PR #11382, opened 2022-01-25
DO NOT MERGE YET: Interactive transaction test

garrensmith:
This adds in a longer concurrent interaction transactions test to make sure the engine doesn't deadlock. 
If I use the engine on main branch the test never completes. If I use the new engine from here https://github.com/prisma/prisma-engines/pull/2619 then the test completes. 

When the query-engine PR is merged I'll updated this PR to remove the tmp commit.


Jolg42:
Seems that GitHub Actions are not running 🤔 Not sure why

millsp:
Closing in favor of https://github.com/prisma/prisma/pull/10797.

PR #11450, opened 2022-01-27
chore(client): dmmfToTypes

### LINKED ISSUE #11295: Internal request: have a way to get types string from dmmf
To summarize, we want expose a tool that does DMMF -> index.d.ts string for the PDP (avoiding generate through the CLI). Not a public / documented API.

millsp:
Adds an internal feature for generating types without needing the CLI or a generator.

### DIFF
--- a/packages/client/src/generation/utils/types/dmmfToTypes.ts
+++ b/packages/client/src/generation/utils/types/dmmfToTypes.ts
@@ -1,7 +1,11 @@
 import type { DMMF } from '../../../runtime/dmmf-types'
 import { TSClient } from '../../TSClient/TSClient'
 
-// Internal, used by the PDP not to need a call to the CLI
+/**
+  * @internal
+  *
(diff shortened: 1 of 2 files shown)


packages/client/src/generation/utils/types/dmmfToTypes.ts
```diff
@@ -0,0 +1,18 @@
+import type { DMMF } from '../../../runtime/dmmf-types'
+import { TSClient } from '../../TSClient/TSClient'
+
+// Internal, used by the PDP not to need a call to the CLI
```
jasonkuhrt, line 4: https://tsdoc.org/pages/tags/internal/

```suggestion
/**
  * @internal
  *
  * @remarks Used by, for example, the PDP to avoid child process calls to the CLI.
  */
```

jasonkuhrt (approved):

PR #11479, opened 2022-01-28
fix(client): prisma promise for count

millsp:
`count` operations were not created as prisma promises.


Jolg42 (commented):
How did you find out?

Next question, how can we test this, so we make sure we don't get a regression for example?

Jolg42 (approved):

jasonkuhrt:
@Jolg42 total fluke I found out this bug because data browser makes count queries on initial load.

PR #11503, opened 2022-01-30
fix(client): Fix typing generation of `<Model>.groupBy` count property

### LINKED ISSUE #10915: Typing problems in groupby method
### Bug description

For `<Proposal>.groupBy` method, the `_count` value has a invalid type, it should only be a boolean as we can see in the doc :
> image
https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#groupby

image

This error let the developer make this type of code: 
image

Witch let thinking that it would only return the count of elements which is in accord with the object definition (only merged proposals).

### How to reproduce

1. Use `<Model>.groupBy` method with `_count` property

### Expected behavior

Only have the optional typing as written in the doc (which make more sense)

### Prisma information

```js
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
}

model Proposal {
  id            Int           @id @default(autoincrement())
  merged        Boolean       @default(false)
  
  // table content removed

  @@map("proposals")
}
```

### Environment & setup

- OS: Docker Node Image based on Ubuntu
- Database: PostgreSQL
- Node.js version: latest

### Prisma Version

```
prisma                  : 3.7.0
@prisma/client          : 3.7.0
Query Engine (Node-API) : libquery-engine 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : 8746e055198f517658c08a0c426c7eec87f5a85f
Studio                  : 0.445.0
```

DraftProducts:
Closes: #10915

Hi guys, here is my first PR on the repo !
Please excuse me in advance for any omissions there may be, I have read the guide-lines well.
Also, I couldn't find what to add as a test with jest, if it's necessary, I would be happy to write them :thumbsup:

### DIFF
--- a/packages/client/src/generation/TSClient/Model.ts
+++ b/packages/client/src/generation/TSClient/Model.ts
@@ -110,8 +110,8 @@ ${indent(
         .filter((f) => f.outputType.location === 'outputObjectTypes')
         .map((f) => {
           if (f.outputType.location === 'outputObjectTypes') {
-            return `${f.name}?: ${getAggregateInputType((f.outputType.type as DMMF.OutputType).name)}${
-              f.name === '_count' ? ' | true' : ''
+            return `${f.name}?: ${
+              f.name === '_count' ? 'true' : getAggregateInputType((f.outputType.type as DMMF.OutputType).name)
             }`
           }
 


aqrln:
Thanks again for your time @DraftProducts! I left the explanation in the issue (https://github.com/prisma/prisma/issues/10915#issuecomment-1066789213), so I'll go ahead and close this PR as well. Don't hesitate to reopen both the issue and the PR if I missed anything!

PR #11516, opened 2022-01-31
test(client): add type snap for mongo

millsp:
I noticed that I did not make `PrismaPromise` the return type of `<model>.findRaw/aggregateRaw`. While valid at runtime, types were off so I added a snapshot. I'm unhappy with the way things are tested with types and I'd like to push forward our test refactor initiative, because duplicating tests is unsustainable (+ types could be tested all at once with the tests).


Jolg42 (approved):
Nice catch!

And I'm all for it! Let's push for that 👍🏼

aqrln (approved):
👍

PR #11521, opened 2022-01-31
fix(cli/migrate): do not load & print env file for all commands

Jolg42:
Not loading and printing .env file is required for `migrate diff` https://github.com/prisma/migrations-team/issues/313 (so you can pipe the output of `migrate diff` to `db execute`)

2 new functions:
- getSchemaPathAndPrint
- loadEnvFileAndPrint

With refactor of some duplicated code

Note that before all commands were loading .env this changes it.
And these do not load the .env file anymore:
- migrate (without subcommand)
- migrate diff (actually it was needed there and added back in a new PR)
- db (without subcommand)
- db execute (actually it was needed there and added back in a new PR)
- init
- version
- telemetry
- format
- dev (legacy that needs to be removed)


packages/migrate/src/utils/getSchemaPathAndPrint.ts
```diff
@@ -0,0 +1,36 @@
+import path from 'path'
+import chalk from 'chalk'
+import { getSchemaPath, logger } from '@prisma/sdk'
+import { NoSchemaFoundError } from './errors'
+
+// TODO move NoSchemaFoundError to sdk and this too
+// then replace the 2 hardcoded errors to NoSchemaFoundError in
+// https://github.com/prisma/prisma/blob/main/packages/sdk/src/cli/getSchema.ts#L383:L383
```
aqrln, line 8: Better to link to the file in a stable commit, otherwise the line numbers will become wrong when the file is changed in `main`.

```suggestion
// https://github.com/prisma/prisma/blob/bbdf1c23653a77b0b5bf7d62efd243dcebea018b/packages/sdk/src/cli/getSchema.ts#L383:L383
```

packages/sdk/src/utils/loadEnvFileAndPrint.ts
```diff
@@ -0,0 +1,12 @@
...
+import { getEnvPaths } from './getEnvPaths'
+
+/**
+ * Read .env file only if next to schema.prisma
+ * .env found: print to console it's relative path
+ */
+export function loadEnvFileAndPrint(schemaPath?: string) {
+  const envPaths = getEnvPaths(schemaPath)
+  const envData = tryLoadEnvs(envPaths, { conflictCheck: 'error' })
+  envData && envData.message && console.info(envData.message)
```
millsp, line 11: Do we need to re-add https://github.com/prisma/prisma/pull/11521/files#diff-f68aae81afb0da1362bf689b66647746ff60979dcf684bd4bd4ff7d6297100f6L78 here?

packages/cli/src/__tests__/dotenv-5-only-root.test.ts
```diff
@@ -1,11 +1,12 @@
...
+import { jestConsoleContext, jestContext, loadEnvFileAndPrint } from '@prisma/sdk'
 
 const ctx = jestContext.new().add(jestConsoleContext()).assemble()
 
 it('should not load root .env file', async () => {
-  process.argv.push('--version')
   ctx.fixture('dotenv-5-only-root')
-  await import('../bin')
-  expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchSnapshot()
+  loadEnvFileAndPrint()
```
millsp, line 7: Does using bin not work anymore, even by passing --schema?

packages/cli/src/__tests__/commands/__snapshots__/Init.test.ts.snap
```diff
@@ -1,7 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`appends when .env present 1`] = `
-Environment variables loaded from .env
```
millsp, line 4: Is this test not loading from env anymore?

packages/cli/src/__tests__/dotenv-5-only-root.test.ts
Jolg42: The env loading logic was done directly in `bin.ts` before and now it's moved to each command because not all commands need to load them (and therefore print)

packages/cli/src/__tests__/commands/__snapshots__/Init.test.ts.snap
Jolg42: Indeed see description of the PR for the list of commands not loading .env anymore

Jolg42: Do you think it's needed? I checked quick but I didn't see how it would be used there, feel free to double check 👍🏼

packages/cli/src/__tests__/dotenv-5-only-root.test.ts
millsp: Thanks, makes sense!

millsp (approved):

packages/migrate/src/commands/MigrateDiff.ts
```diff
@@ -1,6 +1,6 @@
 import path from 'path'
 import type { Command } from '@prisma/sdk'
-import { arg, format, HelpError, isError, getSchemaPath, link, getCommandWithExecutor } from '@prisma/sdk'
+import { arg, format, HelpError, isError, link } from '@prisma/sdk'
 import chalk from 'chalk'
```
tomhoule, line 4: Does the absence of loadEnvAndPrint here mean that we don't load the .env at all? It could be a problem with `--from-schema-datasource`, because there users would expect the env vars from .env to be accessible in the schema.

tomhoule (commented):
Have we considered printing to stderr instead of stdout?

Jolg42:
@tomhoule In which cases? Note that console.error/warn goes to stderr
https://nodejs.org/api/console.html#consoleerrordata-args

packages/migrate/src/commands/MigrateDiff.ts
Jolg42: I think so, shows a missing test, I'll check that now

tomhoule:
not necessarily with `console`, but printing to stderr wouldn't end up getting piped. It would still be visible to users, so I guess we don't want that either (confusing if the first line of your SQL script or diff summary is about env vars)

Jolg42:
So actually

`migrate diff` needs it for `--from-schema-datasource`
`db execute` needs it for `--schema`

PR at https://github.com/prisma/prisma/pull/11546

PR #11524, opened 2022-01-31
test(migrate): tweak setup db scripts

Jolg42:
To avoid this in our test suite (somehow only happening on macOS)
https://github.com/prisma/prisma/runs/5009329376?check_suite_focus=true#step:13:761
```
   ● postgresql › basic introspection

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `postgresql basic introspection 3`

    - Snapshot  - 0
    + Received  + 1

    + Error: P2002: Unique constraint failed on the fields: (`datname`)

```

### DIFF
--- a/packages/migrate/src/__tests__/MigrateDev.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDev.test.ts
@@ -1126,6 +1126,10 @@ describe('mysql', () => {
 
   // Update env var because it's the one that is used in the schemas tested
   process.env.TEST_MYSQL_URI_MIGRATE = connectionString
+  process.env.TEST_MYSQL_SHADOWDB_URI_MIGRATE = connectionString.replace(
+    'tests-migrate-dev',
+    'tests-migrate-dev-shadowdb',
+  )
 
   const setupParams: SetupParams = {
     connectionString,
--- a/packages/migrate/src/__tests__/MigrateDev.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDev.test.ts
@@ -629,14 +629,14 @@ describe('sqlite', () => {
 
     await expect(result).rejects.toMatchInlineSnapshot(`
 
-                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    ⚠️ We found changes that cannot be executed:
+                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                ⚠️ We found changes that cannot be executed:
 
-                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      • Step 0 Made the column \`fullname\` on table \`Blog\` required, but there are 1 existing NULL values.
+                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  • Step 0 Made the column \`fullname\` on table \`Blog\` required, but there are 1 existing NULL values.
 
-                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    You can use prisma migrate dev --create-only to create the migration file, and manually modify it to address the underlying issue(s).
-                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    Then run prisma migrate dev to apply it and verify it works.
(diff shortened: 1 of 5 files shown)


Jolg42:
The flaky windows snapshots should be caused by the fact that the engine implementation sends the result from the RPC request via logs and not in the response, we will change that after the release and can be ignored here.

packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -891,28 +891,22 @@ describe('sqlite', () => {
...
+  const connectionString = (
+    process.env.TEST_POSTGRES_URI_MIGRATE || 'postgres://prisma:prisma@localhost:5432/tests-migrate'
+  ).replace('tests-migrate', 'tests-migrate-dev')
+
+  // Update env var because it's the one that is used in the schemas tested
+  process.env.TEST_POSTGRES_URI_MIGRATE = connectionString
+  process.env.TEST_POSTGRES_SHADOWDB_URI_MIGRATE = connectionString.replace(
+    'tests-migrate-dev',
+    'tests-migrate-dev-shadowdb',
+  )
```
Jolg42, line 903: I'm not happy with this, I think we can improve things here but later

packages/migrate/src/utils/setupMysql.ts
```diff
@@ -13,27 +13,33 @@ export async function setupMysql(options: SetupParams): Promise<void> {
...
     host: credentials.host,
     port: credentials.port,
-    database: credentials.database,
+    // database: credentials.database, // use the default db
     user: credentials.user,
     password: credentials.password,
     multipleStatements: true,
   })
+  await dbDefault.query(`
+CREATE DATABASE IF NOT EXISTS \`${credentials.database}-shadowdb\`;
```
millsp, line 26: No need to drop previous/existing dbs?

packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -1126,17 +1120,22 @@ describe('postgresql', () => {
 })
 
 describe('mysql', () => {
+  const connectionString = (
+    process.env.TEST_MYSQL_URI_MIGRATE || 'mysql://root:root@localhost:3306/tests-migrate'
+  ).replace('tests-migrate', 'tests-migrate-dev')
+
+  // Update env var because it's the one that is used in the schemas tested
+  process.env.TEST_MYSQL_URI_MIGRATE = connectionString
```
millsp, line 1128: I prefer this format https://github.com/prisma/prisma/pull/11524/files#diff-c266a026f47d4de125b7139076b575b33fb6dcb0207110c4a9a0f78304097b41R1335

aqrln (approved):

packages/migrate/src/utils/setupPostgres.ts
```diff
@@ -13,23 +13,26 @@ export async function setupPostgres(options: SetupParams): Promise<void> {
...
-  await createDatabase(connectionString).catch((e) => console.error(e))
-
-  const db = new Client({
-    connectionString: connectionString,
+  // Connect to default db
+  const dbDefault = new Client({
+    connectionString: connectionString.replace(credentials.database!, 'postgres'),
   })
+  await dbDefault.connect()
+  await dbDefault.query(`DROP DATABASE IF EXISTS "${credentials.database}-shadowdb";`)
```
millsp, line 21: Why does PG need to be dropped, but not other DBs?

millsp (approved):

packages/migrate/src/__tests__/MigrateDev.test.ts
aqrln: Yeah, I have no super strong opinion on this, but I personally think the default hardcoded connection strings are not super useful and I'd personally prefer the test to fail if the env var is not set. That the absence of the default value would make the second format (which is shorter and more readable) possible here is a nice bonus.

aqrln: Although that format is possible even with the default value, like this:

```ts
process.env.TEST_MYSQL_URI_MIGRATE = process.env.TEST_MYSQL_URI_MIGRATE?.replace(
  'tests-migrate',
  'tests-migrate-dev'
) ?? 'mysql://root:root@localhost:3306/tests-migrate-dev'
```

Jolg42: Yes I think we can get rid of all the defaults at some point in a separate PR @aqrln, it's hard to maintain them (might not be in sync) and a clear single point of config is better

packages/migrate/src/utils/setupPostgres.ts
Jolg42: It's my way of doing a create table if not exists which doesn't work on PostgreSQL 😢 
So it always starts clean

PR #11529, opened 2022-01-31
fix(client): send valid headers dataproxy

millsp:
Test here (failing before this merge) https://github.com/prisma/e2e-tests/pull/2379

### DIFF
--- a/packages/engine-core/src/data-proxy/utils/request.ts
+++ b/packages/engine-core/src/data-proxy/utils/request.ts
@@ -86,7 +86,6 @@ function nodeFetch(url: string, options: RequestOptions = {}): Promise<RequestRe
   return new Promise((resolve, reject) => {
     const https: typeof _https = eval(`require('https')`)
 
-    console.log(httpsOptions)
     // we execute the https request and build a fetch response out of it
     const request = https.request(url, httpsOptions, (response) => {
       response.on('data', (chunk: Buffer) => incomingData.push(chunk))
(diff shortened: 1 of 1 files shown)


packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -85,6 +86,7 @@ function nodeFetch(url: string, options: RequestOptions = {}): Promise<RequestRe
   return new Promise((resolve, reject) => {
     const https: typeof _https = eval(`require('https')`)
 
+    console.log(httpsOptions)
```
millsp, line 89:
```suggestion
```

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -38,7 +38,8 @@ export async function request(url: string, options: RequestOptions = {}): Promis
  */
 function buildHeaders(options: RequestOptions): RequestOptions['headers'] {
   return {
-    ...options.headers,
+    // this ensures headers will always be valid
+    ...JSON.parse(JSON.stringify(options.headers)),
```
Jolg42, line 42: Which value was passed that prompted this? 👀

millsp: `undefined`

millsp: Discarded correctly everywhere besides the custom nodeFetch

Jolg42: I see https://github.com/prisma/e2e-tests/pull/2379#issuecomment-1026656733

Jolg42 (approved):

PR #11546, opened 2022-02-01
fix(migrate): load .env file for migrate diff and db execute

Jolg42:
Previous PR https://github.com/prisma/prisma/pull/11521 removed global dotenv loading didn't add it for `migrate diff` & `db execute` where it's actually needed 

`migrate diff` needs it for `--from-schema-datasource` and `--to-schema-datasource`
`db execute` needs it for `--schema`


tomhoule (approved):
looks like that should fix it :+1:

packages/migrate/src/commands/MigrateDiff.ts
```diff
@@ -147,6 +147,8 @@ ${chalk.bold('Examples')}
       throw new MigrateDiffNeedsPreviewFeatureFlagError()
     }
 
+    loadEnvFile(args['--schema'], false)
```
tomhoule, line 150: there is no `--schema` argument to diff, is this line right?

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -91,6 +91,8 @@ ${chalk.bold('Examples')}
       throw new DbExecuteNeedsPreviewFeatureFlagError()
     }
 
+    loadEnvFile(args['--schema'], false)
```
tomhoule, line 94: I think same as previous comment

packages/migrate/src/commands/MigrateDiff.ts
Jolg42: Indeed! A bit surprised the type safety didn't help...

PR #11560, opened 2022-02-01
fix(client): do not discard extra args

### LINKED ISSUE #11556: Count does not work with where
### Bug description

When passing the `where` to `prisma.relation.count`, the parameters are not considered, as if they had not been passed

## The code

```
console.log('query without where')

const c = await prisma.user.count()

console.log(`response: ${c}\n\n`)

console.log('query with where id = 1')

const d = await prisma.user.count({
  where: {
    id: 1
  }
})

console.log(`response: ${d}`)
```

## The result

```
query without where
prisma:info Starting a postgresql pool with 5 connections.
prisma:query SELECT COUNT(*) FROM (SELECT "public"."users"."id" FROM "public"."users" WHERE 1=1 OFFSET $1) AS "sub"
response: 24


query with where id = 1
prisma:query SELECT COUNT(*) FROM (SELECT "public"."users"."id" FROM "public"."users" WHERE 1=1 OFFSET $1) AS "sub"
response: 24
```

### Expected behavior

_No response_

### Prisma information

They are in "How to reproduce"

### Environment & setup

- OS: `KDE neon 5.23`
- Database:  `PostgreSQL 13.3`
- Node.js version: `v16.13.2`

### Prisma Version

```
prisma                  : 3.9.0
@prisma/client          : 3.9.0
Query Engine (Node-API) : libquery-engine bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
```

millsp:
We lacked tests on `count` and a regression was introduced. Added more tests and fixed the issue.

### DIFF
--- a/packages/client/src/__tests__/integration/happy/count/test.ts
+++ b/packages/client/src/__tests__/integration/happy/count/test.ts
@@ -65,12 +65,43 @@ describe('count', () => {
     `)
   })
 
-  test('select all', async () => {
+  test('select all true', async () => {
     const value = await prisma.user.count({
       select: true,
     })
 
     expect(value).toMatchInlineSnapshot(`10`)
+  })
+
+  // TODO: this should happen but we always make selects to become true
+  test.skip('select all false', async () => {
+    const value = prisma.user.count({
+      select: false,
+    })
+
+    // atm, succeeds with no error, value === 10 (like the test above)
+    await expect(value).rejects.toThrowErrorMatchingInlineSnapshot(`
+
+            Invalid \`prisma.user.count()\` invocation:
+
+            {
+              select: {
+                _count: {
+                  select: {
+            ?       _all?: true,
+            ?       id?: true,
+            ?       email?: true,
+            ?       age?: true,
+            ?       name?: true
+                  }
+                }
+              }
+            }
+
+
+            The \`select\` statement for type UserCountAggregateOutputType needs at least one truthy value.
+
+          `)
   })
 
   test('select mixed', async () => {
@@ -107,28 +138,28 @@ describe('count', () => {
     } catch (err) {
       expect(err.message).toMatchInlineSnapshot(`
 
-              Invalid \`prisma.user.count()\` invocation:
+                      Invalid \`prisma.user.count()\` invocation:
 
-              {
-                select: {
-                  _count: {
-                    select: {
-              ?       _all?: true,
-              ?       email?: true,
-              ?       age?: true,
-              ?       name?: true,
-                      posts: true,
-                      ~~~~~
-              ?       id?: true
-                    }
-                  }
-                }
-              }
+                      {
+                        select: {
+                          _count: {
+                            select: {
+                      ?       _all?: true,
+                      ?       email?: true,
+                      ?       age?: true,
+                      ?       name?: true,
+                              posts: true,
+                              ~~~~~
+                      ?       id?: true
+                            }
+                          }
+                        }
+                      }
 
 
-              Unknown field \`posts\` for select statement on model UserCountAggregateOutputType. Available options are listed in green. Did you mean \`id\`?
+                      Unknown field \`posts\` for select statement on model UserCountAggregateOutputType. Available options are listed in green. Did you mean \`id\`?
 
-            `)
+                  `)
     }
   })
 })
--- a/packages/client/src/__tests__/integration/happy/count/test.ts
+++ b/packages/client/src/__tests__/integration/happy/count/test.ts
@@ -67,41 +67,18 @@ describe('count', () => {
 
   test('select all true', async () => {
     const value = await prisma.user.count({
-      select: true,
+      select: true, // count with a selection
     })
 
     expect(value).toMatchInlineSnapshot(`10`)
   })
 
-  // TODO: this should happen but we always make selects to become true
-  test.skip('select all false', async () => {
-    const value = prisma.user.count({
-      select: false,
+  test('select all false', async () => {
+    const value = await prisma.user.count({
+      select: false, // count with no selection
     })
 
-    // atm, succeeds with no error, value === 10 (like the test above)
-    await expect(value).rejects.toThrowErrorMatchingInlineSnapshot(`
-
-            Invalid \`prisma.user.count()\` invocation:
-
-            {
-              select: {
-                _count: {
-                  select: {
-            ?       _all?: true,
-            ?       id?: true,
-            ?       email?: true,
-            ?       age?: true,
-            ?       name?: true
-                  }
-                }
-              }
-            }
-
-
-            The \`select\` statement for type UserCountAggregateOutputType needs at least one truthy value.
-
-          `)
+    expect(value).toMatchInlineSnapshot(`10`)
   })
 
   test('select mixed', async () => {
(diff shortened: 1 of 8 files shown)


packages/client/src/__tests__/integration/happy/count/test.ts
```diff
@@ -1,68 +1,165 @@
...
-      ?       id?: true
+  test('select all true', async () => {
+    const value = await prisma.user.count({
+      select: true,
+    })
+
+    expect(value).toMatchInlineSnapshot(`10`)
+  })
+
+  // TODO: this should happen but we always make selects to become true
```
aqrln, line 76: We should create an issue for this 👍

millsp: Definitely. The reason why I chose not to fix this is that it's been behaving like this before this regression even.

packages/client/src/runtime/core/model/aggregates/count.ts
```diff
@@ -9,15 +10,17 @@ import { aggregate } from './aggregate'
  * @param modelAction a callback action that triggers request execution
  * @returns
  */
-export function count(client: Client, userArgs: object | undefined, modelAction: ModelAction) {
+export function count(client: Client, userArgs: UserArgs | undefined, modelAction: ModelAction) {
   // count is an aggregate, we reuse them but hijack their unpacker
 
+  const { select, ..._userArgs } = userArgs ?? {} // exclude select
   if (typeof userArgs?.['select'] === 'object') {
```
aqrln, line 17: `userArgs?.['select']` can now be replaced with just `select` here and below

packages/client/src/runtime/core/model/aggregates/groupBy.ts
```diff
@@ -10,13 +11,15 @@ import { desugarUserArgs as desugarUserArgsAggregate } from './aggregate'
...
+function desugarUserArgs(userArgs: UserArgs) {
   const _userArgs = desugarUserArgsAggregate(userArgs)
 
   // we desugar the array into { [key]: boolean }
-  if (Array.isArray(userArgs?.['by']) === true) {
-    for (const key of userArgs?.['by'] ?? []) {
-      _userArgs['select'][key] = true
+  if (Array.isArray(userArgs['by'])) {
+    for (const key of userArgs['by']) {
+      if (_userArgs['select'] && typeof key === 'string') {
```
aqrln, line 20: Is the first check here to make the type checker happy? As far as I understand, `_userArgs['select']` is always present since it's the result of `desugarUserArgsAggregate`.

packages/client/src/__tests__/integration/happy/count/test.ts
aqrln: yep, that would definitely be out of scope

packages/client/src/runtime/core/model/aggregates/groupBy.ts
millsp: Yep, probably overkill, I adjusted the return type in the parent call instead.

packages/client/src/__tests__/integration/happy/count/test.ts
millsp: Actually, thinking about it, I don't think it's a problem. If a user does not want to make a selection with `select: false` that still means that he wants to count (but without any selection), hence `select: true` in any case.

millsp: (it's confusing from a code logic point of view, but makes sense from a user point of view)

aqrln (approved):

packages/client/src/runtime/core/model/aggregates/groupBy.ts
```diff
@@ -10,13 +11,15 @@ import { desugarUserArgs as desugarUserArgsAggregate } from './aggregate'
...
 
   // we desugar the array into { [key]: boolean }
-  if (Array.isArray(userArgs?.['by']) === true) {
-    for (const key of userArgs?.['by'] ?? []) {
-      _userArgs['select'][key] = true
+  if (Array.isArray(userArgs['by'])) {
+    for (const key of userArgs['by']) {
+      if (typeof key === 'string') {
+        _userArgs['select'][key] = true
+      }
```
Jolg42, line 22: Seems that `if (typeof key === 'string') {` is new here, what is its purpose in this context? (add a comment maybe?)

millsp: It's to keep the checker happy (it was `any` before).

packages/client/src/__tests__/integration/happy/count/test.ts
Jolg42: It would be nice to have a test case with `select all false` that is passing so it changes the test will catch it right away.
With a comment for the test

packages/client/src/runtime/core/model/aggregates/aggregate.ts
```diff
@@ -65,7 +66,7 @@ export function createUnpacker(userArgs: object) {
  * @param modelAction a callback action that triggers request execution
  * @returns
  */
-export function aggregate(client: Client, userArgs: object | undefined, modelAction: ModelAction) {
+export function aggregate(client: Client, userArgs: UserArgs | undefined, modelAction: ModelAction) {
```
Jolg42, line 69:
```suggestion
export function aggregate(client: Client, userArgs: UserArgs, modelAction: ModelAction) {
```

UserArgs type can be undefined I see, so undefined again here is no necessary I think

Jolg42: This change might be needed in other places

packages/client/src/runtime/core/model/UserArgs.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Input that flows from the user into the Client.
+ */
+export type UserArgs = {
+  [K in string]: UserArgsProp | UserArgsProp[]
+}
+
+type UserArgsProp = UserArgs | string | number | boolean | bigint | object | null | undefined
```
Jolg42, line 8: The original type I could see previous code is `object | undefined`, here it expands the possibilities a lot, and I'm unsure if all are needed? (trying to think about it, but a quick chat would help probably)

packages/client/src/runtime/core/model/aggregates/aggregate.ts
millsp: `UserArgs` is not undefined but its properties can.

Jolg42: Ok I'm definitely not at my best these 2 days 😅

Jolg42 (approved):
Let's get this into a patch release now 🚢

Issue #11570, opened 2022-02-02
`prisma db pull` doesn't read `.env` file and errors with Environment variable not found: DATABASE_URL


smoothdvd:
### Bug description

yarn prisma db pull report Environment variable not found: DATABASE_URL.
```
Prisma schema loaded from prisma/schema.prisma
Error: Get config: Schema Parsing P1012

error: Environment variable not found: DATABASE_URL.
  -->  schema.prisma:10
   | 
 9 |   provider = "postgresql"
10 |   url      = env("DATABASE_URL")
   | 

Validation Error Count: 1
```

.env:
DATABASE_URL="postgresql://user:pass@localhost:49154/db?schema=public"

### Expected behavior

read DATABASE_URL from .env correctly

### Prisma information

Environment variables loaded from .env
prisma                  : 3.9.0
@prisma/client          : 3.9.0
Current platform        : darwin-arm64
Query Engine (Node-API) : libquery-engine bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : bcc2ff906db47790ee902e7bbc76d7ffb1893009
Studio                  : 0.457.0

### Environment & setup

- OS: macOS 12.2
- Database: PostgreSQL 13
- Node.js version: v14.18.1

### Prisma Version

```

```

Jolg42:
Hi @smoothdvd 

Thanks for the report!

I can confirm this is a regression, we are working on a fix and are planning a patch release later today 🚢

millsp:
Hey, this issue has been fixed in [3.9.1](https://github.com/prisma/prisma/releases/tag/3.9.1) which was just published. Thanks for reporting this!

cweekly:
Phew! Thought I was losing my mind. Confirmed I saw this in 3.9.0, and thankfully not in 3.9.2.

[...]

PR #11571, opened 2022-02-02
fix(db pull): CLI should load .env file for db pull

Jolg42:
Fixes https://github.com/prisma/prisma/issues/11570

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -410,7 +410,6 @@ jobs:
       - run: pnpm run test
         working-directory: packages/migrate
         env:
-          DEBUG: 'prisma*'
           CI: true
           SKIP_GIT: true
           GITHUB_CONTEXT: ${{ toJson(github) }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -410,6 +410,7 @@ jobs:
       - run: pnpm run test
         working-directory: packages/migrate
         env:
+          DEBUG: '*'
           CI: true
           SKIP_GIT: true
           GITHUB_CONTEXT: ${{ toJson(github) }}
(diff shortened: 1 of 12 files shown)


packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -379,6 +379,45 @@ describe('postgresql', () => {
...
+  test('introspection should load .env file without --print', async () => {
+    ctx.fixture('schema-only-postgresql')
+    expect.assertions(5)
+
+    try {
+      await DbPull.new().parse(['--schema=./prisma/using-dotenv.prisma'])
+    } catch (e) {
+      expect(e.code).toEqual('P1001')
+      expect(e.message).toContain(`fromdotenvdoesnotexist`)
+    }
```
aqrln, line 408: Maybe `expect(...).toThrow()`/`expect(...).toThrowErrorMatchingSnapshot()`? Not sure if it will actually be better than `expect.assertions` + try/catch here, so feel free to ignore the suggestion if you think it won't.

.github/workflows/test.yml
```diff
@@ -410,6 +410,7 @@ jobs:
       - run: pnpm run test
         working-directory: packages/migrate
         env:
+          DEBUG: 'prisma*'
```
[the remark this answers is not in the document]
aqrln: oh, it seems it's already removed, sorry for the noise here

aqrln (approved):

packages/migrate/src/__tests__/DbPull.test.ts
Jolg42: I had that before but here I wanted something that would not be updated like a snapshot. 
Main benefit is that we can't do a `--u` and update things without noticing, it's a stronger test I think, also if the message change around we won't need an update.

Downside is that `expect.assertions` is required, I think we could change that to `expect.hasAssertions()` in most cases.

PR #11589, opened 2022-02-03
feat(client): composites

millsp:
Integrates https://github.com/prisma/prisma-engines/pull/2466


packages/cli/scripts/preinstall.js
```diff
@@ -63,6 +64,7 @@ function parsePackageManagerName(userAgent) {
   // - https://pnpm.js.org/en/3.6/only-allow-pnpm
   // - https://github.com/cameronhunter/npm-config-user-agent-parser
   if (userAgent) {
+    // eslint-disable-next-line no-useless-escape -- ???
```
Jolg42, line 67: I thought I fixed that on main, is this branch up-to-date already?

packages/client/scripts/colors.js
```diff
@@ -3,8 +3,8 @@
 const isObject = (val) => val !== null && typeof val === 'object' && !Array.isArray(val)
 
 // this is a modified version of https://github.com/chalk/ansi-regex (MIT License)
-/* eslint-disable-next-line no-control-regex */
 const ANSI_REGEX =
+  // eslint-disable-next-line no-control-regex
```
Jolg42, line 7: I thought I fixed that on main, is this branch up-to-date already?

Jolg42: and same comments for other places as well

packages/generator-helper/src/dmmf.ts
```diff
@@ -75,8 +75,8 @@ export namespace DMMF {
     isUnique: boolean
     isId: boolean
     isReadOnly: boolean
-    isGenerated: boolean
-    isUpdatedAt: boolean
+    isGenerated?: boolean // does not exist on type
+    isUpdatedAt?: boolean // does not exist on type
```
Jolg42, line 79: Do you know when it exists?

packages/client/src/utils/getTestClient.ts
```diff
@@ -26,11 +26,9 @@ const readFile = promisify(fs.readFile)
...
  */
 export async function getTestClient(schemaDir?: string, printWarnings?: boolean): Promise<any> {
-  if (!schemaDir) {
-    const callsite = parse(new Error('').stack!)
-    schemaDir = path.dirname(callsite[1].file!)
-  }
-  const schemaPath = await getRelativeSchemaPath(schemaDir)
+  const callSite = path.dirname(require.main?.filename ?? '')
+  const absSchemaDir = path.resolve(callSite, schemaDir ?? '')
+  const schemaPath = await getRelativeSchemaPath(absSchemaDir)
```
Jolg42, line 31: Curious to know why the change was needed though it looks good

packages/client/src/runtime/query.ts
```diff
@@ -1050,10 +1050,13 @@ function byToSelect(by: string[]): Record<string, true> {
...
 }
 
-function getDefaultSelection(outputType: DMMF.OutputType) {
+function getDefaultSelection(dmmf: DMMFHelper, outputType: DMMF.OutputType) {
   const acc = Object.create(null)
 
   for (const f of outputType.fields) {
+    if (dmmf.typeMap[(f.outputType.type as DMMF.OutputType).name] !== undefined) {
+      acc[f.name] = true
+    }
```
Jolg42, line 1059: Maybe add a comment about what this does?

packages/client/src/generation/TSClient/Count.ts
```diff
@@ -84,7 +84,7 @@ class CountDelegate implements Generatable {
     if (!mapping) {
       return ''
     }
-    const model = this.dmmf.modelMap[name]
+    const model = this.dmmf.typeModelMap[name]
```
Jolg42, line 87: Maybe could be better if we also rename the const? It can be confusing now I think.
From `const model` to `const modelType` maybe?

packages/client/src/generation/TSClient/Model.ts
```diff
@@ -329,29 +342,36 @@ export class ModelDelegate implements Generatable {
...
-    if (!mapping) {
-      return ''
-    }
-    const model = this.dmmf.modelMap[name]
 
-    const actions = Object.entries(mapping).filter(
-      ([key, value]) => key !== 'model' && key !== 'plural' && key !== 'aggregate' && key !== 'groupBy' && value,
-    )
+    const mapping = this.dmmf.mappingsMap[name] ?? { model: name, plural: `${name}s` }
+    const model = this.dmmf.typeModelMap[name]
```
Jolg42, line 347: Maybe could be better if we also rename the const? It can be confusing now I think.
From `const model` to `const modelType` maybe?

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -144,7 +144,7 @@ ${buildNFTAnnotations(engineType, platforms, relativeOutdir)}
     const collector = new ExportCollector()
 
     const commonCode = commonCodeTS(this.options)
-    const models = Object.values(this.dmmf.modelMap).reduce((acc, model) => {
+    const models = Object.values(this.dmmf.typeModelMap).reduce((acc, model) => {
```
Jolg42, line 147: Maybe could be better if we also rename the const? It can be confusing now I think.
From `const models` to `const modelsType` maybe?

packages/client/src/utils/getTestClient.ts
millsp: I removed the stack trace as `require.main` should be faster. But the real change is about being able to get an in-memory client and specifying a relative path (see tests) or an absolute path. Either absolute or relative works thanks to `path.resolve`.

packages/client/src/generation/TSClient/TSClient.ts
Jolg42: What makes more sense to you, that looks good. Better than keeping the old const name I think

packages/client/src/generation/TSClient/Model.ts
```diff
@@ -54,20 +54,18 @@ export class Model implements Generatable {
   }
   protected get argsTypes(): Generatable[] {
     const { mapping, model } = this
```
Jolg42, line 56:
```suggestion
    const { mapping } = this
```
`model` is unused apparently

packages/client/src/generation/TSClient/Model.ts
```diff
@@ -330,29 +343,36 @@ export class ModelDelegate implements Generatable {
...
-    const mapping = this.dmmf.mappingsMap[name]
-    if (!mapping) {
-      return ''
-    }
-    const model = this.dmmf.modelMap[name]
 
-    const actions = Object.entries(mapping).filter(
-      ([key, value]) => key !== 'model' && key !== 'plural' && key !== 'aggregate' && key !== 'groupBy' && value,
-    )
+    const mapping = this.dmmf.mappingsMap[name] ?? { model: name, plural: `${name}s` }
```
Jolg42, line 347: Pluralizing is often a bad idea, we should remove it in the future (and stop adding more)

packages/client/src/generation/TSClient/common.ts
```diff
@@ -277,7 +277,11 @@ type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
  * XOR is needed to have a real mutually exclusive union type
  * https://stackoverflow.com/questions/42123407/does-typescript-support-mutually-exclusive-types
  */
-type XOR<T, U> = (T | U) extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
+type XOR<T, U> =
+  T extends object ?
+  U extends object ?
+    (Without<T, U> & U) | (Without<U, T> & T)
+  : U : T
```
Jolg42, line 284: The original XOR was a copy pasta from stackoverflow, what is the change about? (add a comment?)

packages/cli/scripts/preinstall.js
millsp: Merged with main

packages/client/scripts/colors.js
millsp: Merged with main that's gone now

packages/generator-helper/src/dmmf.ts
millsp: I updated with a better comment locally

packages/client/src/runtime/query.ts
millsp: Done

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/schema.prisma
```diff
@@ -34,6 +34,33 @@ model User {
...
+  boolean            Boolean
+  embedEmbedList     EmbedEmbed[]
+  requiredEmbedEmbed EmbedEmbed
+  optionalEmbedEmbed EmbedEmbed?
+}
+
+type EmbedEmbed {
+  text    String
+  boolean Boolean
 }
```
Jolg42, line 64: Would be interesting to know how much recursion should be supported / tested.
Introspection default is now -1 = infinite

Jolg42 (approved):
That's quite a bunch of tests!

I think it could be interesting to share the work you did with @aqrln here later

packages/client/src/generation/TSClient/common.ts
millsp: Yep, the original type effectively broke type distribution with `(T | U)`, which is necessary for nested `XOR`s that are unions. It would in turn break in unexpected ways, so probably an issue on the TS side (since required fields would be required but show no error when not provided) - so caused by us, after all.

PR #11636, opened 2022-02-04
test(client): increase 1 generetor test timeout for windows only

Jolg42:
https://www.notion.so/prismaio/client-generator-test-ts-timeouts-Exceeded-timeout-of-30000-ms-for-a-test-34f772ebefc54206864f917a0366f565


Jolg42:
Also fails after 60s 😒

millsp (approved):

PR #11678, opened 2022-02-07
tests(db execute/pull): use some expect.objectContaining and share cli instance

Jolg42:


aqrln (commented):
Is there a reason why the `DbExecute` instance is now shared between tests? I'd prefer not doing this if possible to avoid accidentally sharing state across the tests and introducing dependencies between them.

PR #11682, opened 2022-02-07
fix(build): cleanup intermediary esm files

### LINKED ISSUE #11683: Avoid shipping unused intermediary ESM files
We have ESM builds to achieve tree-shaking. However, this is not proper ESM (still has requires, __dirnames, etc...) and it will fail at runtime. So for now, there is no actual reason to expose this.

https://prisma.github.io/prisma/dev/bench/

millsp:
Fixes https://github.com/prisma/prisma/issues/11683

### DIFF
--- a/helpers/compile/plugins/cleanupPlugin.ts
+++ b/helpers/compile/plugins/cleanupPlugin.ts
@@ -1,5 +1,8 @@
 import type * as esbuild from 'esbuild'
-import { unlink } from 'fs/promises'
+import { promisify } from 'util'
+import fs from 'fs'
+
+const unlink = promisify(fs.unlink)
 
 /**
  * Removes the esm output files after cjs compile.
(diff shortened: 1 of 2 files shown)


helpers/compile/plugins/cleanupPlugin.ts
```diff
@@ -0,0 +1,32 @@
...
+const unlink = promisify(fs.unlink)
+
+/**
+ * Removes the esm output files after cjs compile.
+ * @param build the esbuild build object
+ * @returns
+ */
+function removeIntermediaryEsmFiles(build: esbuild.PluginBuild) {
+  const files = Object.values(build.initialOptions.entryPoints ?? {})
+  const rms = files.map((file) => file.endsWith('.mjs') && unlink(file))
```
Jolg42, line 14:
```suggestion
  const rms = files
  	.filter((file) => file.endsWith('.mjs'))
  	.map((file) => unlink(file))
```

That's probably fewer promises created like this

Jolg42 (approved):
See comment for minor improvement

PR #11693, opened 2022-02-08
ci: fix github actions

Jolg42:
Follows https://github.com/prisma/prisma/pull/10739

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -31,11 +31,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Use Node.js
-        uses: actions/setup-node@v2
+      - uses: pnpm/action-setup@v2.0.1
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark.yml
```diff
@@ -31,11 +31,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Use Node.js
-        uses: actions/setup-node@v2
+      - uses: pnpm/action-setup@v2.0.1
```
millsp, line 34: Why `2.0.1` over `2`?

Jolg42: It's 2 different actions

`actions/checkout@v2` and `pnpm/action-setup@v2.0.1` for the last one, it's the latest version available that we user everywhere and is managed by Renovate

millsp (approved):

PR #11695, opened 2022-02-08
fix(client): itx (3rd round)

### LINKED ISSUE #7918: groupBy is not usable with $transaction
### Bug description

Trying to use the unawaited return value of a prisma.<type>.groupBy(...) in a prisma.$transaction([...]) will result in a typescript Type error:

```
Type 'GetOrderGroupByPayload<{ by: "vendorId"[]; _sum: { price: true; }; }>' is not assignable to type 'PrismaPromise<any>'.
  Property '[prisma]' is missing in type 'Promise<(PickArray<OrderGroupByOutputType, "vendorId"[]> & { _sum: { price: number | null; }; })[]>' but required in type '{ [prisma]: true; }'
```
 It seems that all values passed into $transaction must have a property with a Symbol called prisma set to true. This is not the case for groupBy, but works for other queries/mutate actions like findMany, deleteMany, findUnique, ...

### How to reproduce

Using below minimal demo schema, this code errors:
```ts
 await db.$transaction([
    db.order.groupBy({
      by: ["vendorId"],
      _sum: {
        price: true,
      }
    })
  ])
```

### Expected behavior

$transaction should properly accept groupBy queries, running them as a transaction with other queries.

### Prisma information

```prisma
datasource db {
  provider = "mysql"
}

generator client {
  provider = "prisma-client-js"
}

model Vendor {
  id     Int     @id
  Orders Order[]
}

model Order {
  id       Int    @id
  price    Float
  Vendor   Vendor @relation(fields: [vendorId], references: [id])
  vendorId Int
}
```

### Environment & setup

- OS: Windows 10
- Database: MySQL
- Node.js version: v14.15.4
- Typescript version: Version 4.3.2

### Prisma Version

```
prisma               : 2.25.0
@prisma/client       : 2.25.0
Query Engine         : query-engine c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\@prisma\engines\query-engine-windows.exe)
Migration Engine     : migration-engine-cli c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\@prisma\engines\migration-engine-windows.exe)
Introspection Engine : introspection-core c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary        : prisma-fmt c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash : c838e79f39885bc8e1611849b1eb28b5bb5bc922
Studio               : 0.402.0
```

### LINKED ISSUE #9073: `InteractiveTransaction`: Middleware param `runInTransaction` set to `false`
### Bug description

When using `interactiveTransaction` preview feature alongside with a middleware you won't get the `runInTransaction` parameter set to `true`.

image

Despite the above issue I would want the ability to have access to the `transactionPrismaClient` or an `afterCommit` hook in `params`. That way I can run logic only after the transaction is committed.
Sequelize equivalent of this would be: https://sequelize.org/master/manual/transactions.html#the--code-aftercommit--code--hook

### How to reproduce

```ts
// schema.prisma
generator client {
}

datasource db {
  provider = "postgresql"
}

model Car {
  model String
}

// middleware.ts
export const middleware = (prisma: PrismaClient) => {
 	console.log('params', params) // params.runInTransaction = false -> should be true
  })
}

// prisma.ts
import { middleware } from './middleware'

middleware(prisma)

export { prisma }

// main.js
import { prisma } from './prisma.ts'

const main = async () => {
	await prisma.$transaction(async (tPrisma) => {
		return tPrisma.car.create({ data: { model: 'abc' } })
	})
}
```

### Expected behavior

_No response_

### Prisma information

in `How to reproduce` I wrote an example

### Environment & setup

- OS: Ubuntu
- Database: PostgreSQL
- Node.js version: Node v14

### Prisma Version

```
prisma               : 2.29.1
@prisma/client       : 2.29.1
Current platform     : debian-openssl-1.1.x
Query Engine         : query-engine 1be4cd60b89afa04b192acb1ef47758a39810f3a (at node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x)
Migration Engine     : migration-engine-cli 1be4cd60b89afa04b192acb1ef47758a39810f3a (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine : introspection-core 1be4cd60b89afa04b192acb1ef47758a39810f3a (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary        : prisma-fmt 1be4cd60b89afa04b192acb1ef47758a39810f3a (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash : 1be4cd60b89afa04b192acb1ef47758a39810f3a
Studio               : 0.419.0
Preview Features     : interactiveTransactions
```

### LINKED ISSUE #10615: Using "then ()" in "interactive Transactions" initiates another transaction
### Bug description

In following code, txid1 and txid2 have different value.
```
  await prisma.$transaction(async (prisma) => {
    const txid1 = await prisma.$queryRawUnsafe('SELECT txid_current()')
    const txid2 = await prisma.$queryRawUnsafe('SELECT txid_current()').then()
  })
```

### How to reproduce

Execute following code in postgres
```
  await prisma.$transaction(async (prisma) => {
    const txid1 = await prisma.$queryRawUnsafe('SELECT txid_current()')
    const txid2 = await prisma.$queryRawUnsafe('SELECT txid_current()').then()
  })
```

### Expected behavior

txid1 and txid2 should be same value

### Prisma information
```
datasource db {
  provider = "postgresql"
}

generator client {
  provider = "prisma-client-js"
  binaryTargets = ["native", "rhel-openssl-1.0.x"]
}
```

### Environment & setup

- OS: Mac OS
- Database: PostgreSQL
- Node.js version: v14.17.3

### Prisma Version

```
prisma                  : 3.2.1
@prisma/client          : 3.2.1
Query Engine (Node-API) : libquery-engine b71d8cb16c4ddc7e3e9821f42fd09b0f82d7934c (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli b71d8cb16c4ddc7e3e9821f42fd09b0f82d7934c (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core b71d8cb16c4ddc7e3e9821f42fd09b0f82d7934c (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt b71d8cb16c4ddc7e3e9821f42fd09b0f82d7934c (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : b71d8cb16c4ddc7e3e9821f42fd09b0f82d7934c
Studio                  : 0.435.0
Preview Features        : interactiveTransactions
```

### LINKED ISSUE #11565: `interactiveTransactions` feature breaks long running transactions
### Bug description

When you enable the new `interactiveTransactions` preview feature, it will break all your existing long running transactions.

The following error will be printed to the console:
```
[0] prisma:info Starting a sqlite pool with 9 connections.
[0] PrismaClientKnownRequestError: 
[0] Invalid `prisma_1.prisma.contract.create()` invocation in
[0] /home/livthomas/Projects/livthomas/pawnshop-next/build/main/legacy/contract/importLegacyContractRecords.js:15:41
[0] 
[0]   12 const userPasswordsMap = await (0, createUserHashedPasswordsMap_1.createUserHashedPasswordsMap)(userNames);
[0]   13 const operations = results.map(result => {
[0]   14     const data = (0, convertLegacyRecordToContract_1.convertLegacyRecordToContract)(result, userPasswordsMap, archived);
[0] → 15     return prisma_1.prisma.contract.create(
[0]   Transaction API error: Transaction already closed: Transaction is no longer valid. Last state: 'Expired'.
[0]     at Object.request (/home/livthomas/Projects/livthomas/pawnshop-next/node_modules/@prisma/client/runtime/index.js:39068:15)
[0]     at async PrismaClient._request (/home/livthomas/Projects/livthomas/pawnshop-next/node_modules/@prisma/client/runtime/index.js:40851:18)
[0]     at async Promise.all (index 0)
[0]     at async Proxy._transactionWithCallback (/home/livthomas/Projects/livthomas/pawnshop-next/node_modules/@prisma/client/runtime/index.js:40807:18)
[0]     at async importLegacyContractRecords (/home/livthomas/Projects/livthomas/pawnshop-next/build/main/legacy/contract/importLegacyContractRecords.js:17:5)
[0]     at async reportLegacyTableImportStatus (/home/livthomas/Projects/livthomas/pawnshop-next/build/main/legacy/reportLegacyTableImportStatus.js:34:9)
[0]     at async importLegacyDatabase (/home/livthomas/Projects/livthomas/pawnshop-next/build/main/legacy/importLegacyDatabase.js:27:20)
[0]     at async node:electron/js2c/browser_init:193:563 {
[0]   code: 'P2028',
[0]   clientVersion: '3.9.0',
[0]   meta: {
[0]     error: "Transaction already closed: Transaction is no longer valid. Last state: 'Expired'."
[0]   }
[0] }
```

I thought this would get resolved once #9533 is fixed but nothing really changed with the 3.9.0 release.

### How to reproduce

1. Create `prisma.schema` for an SQLite database
2. Enable `interactiveTransactions` preview feature in `prisma.schema`
2. Run a transaction which takes around 30 seconds to execute using the original API:
    ```ts
    prisma.$transaction([operation1, operation2]);
    ```
3. Wait for errors

### Expected behavior

This new `interactiveTransactions` preview feature should not affect the existing transactions in any way.

### Prisma information

I cannot really share this. But there is nothing special, just 20000 create queries on a single table with a few `connectOrCreate` fields.

### Environment & setup

- OS: Linux 4.19.225-1-MANJARO
- Database: SQLite
- Node.js version: v17.3.0

### Prisma Version

```
prisma                  : 3.9.0
@prisma/client          : 3.9.0
Query Engine (Node-API) : libquery-engine bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Preview Features        : interactiveTransactions
```

### LINKED ISSUE #11574: Go through the existing iTX and explore what needs to be worked on
This involves:
* Looking at existing issues
* Evaluating their severity
* Eventually get a repro working
* Plan what should be tackled

### LINKED ISSUE #11654: Interactive transaction: Sequential Prisma Client operations timeout but work when interactive transactions are disabled
### Bug description

To build a poor man's 'createMany' for SQLite, I use a regular (sequential) transaction

`db.$transaction(arrayOfCreateOperations)`

I have about 20.000 rows. This works fine when interactive transactions are disabled. With the feature activated the transaction times out. I though this was fixed in 3.9.1 but apparently not.

### How to reproduce

1. Create a long running "classic" transaction with `db.$transaction(arrayOfOps)`
2. Active interactive transactions preview feature.
3. Run the same transaction from step 1

### Expected behavior

It should make no difference to "classic" transaction whether interactive transactions are enabled or not.

### Environment & setup

- OS: MacOS
- Database: SQLite
- Node.js version: 16

### Prisma Version

```
prisma                  : 3.9.1
@prisma/client          : 3.9.1
Query Engine (Node-API) : libquery-engine bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
```

millsp:
Fixes https://github.com/prisma/prisma/issues/10615
Fixes https://github.com/prisma/prisma/issues/7918
Fixes https://github.com/prisma/prisma/issues/9073
Fixes https://github.com/prisma/prisma/issues/11654
Fixes https://github.com/prisma/prisma/issues/11565
Closes https://github.com/prisma/prisma/issues/11574


tsconfig.json
```diff
@@ -17,6 +17,6 @@
       "@prisma/engines-version": ["../engines-wrapper/packages/engines-version/src"]
     }
   },
-  "include": ["**/*"],
-  "exclude": ["**/dist", "**/build", "**/node_modules"]
+  "include": ["**/*", ".eslintrc.js"],
+  "exclude": ["**/dist", "**/build", "**/node_modules", "**/packages"]
```
Jolg42, line 21: Excluding all packages with `"**/packages"` doesn't make sense to me, what's the logic here?

packages/client/src/runtime/core/request/PrismaPromise.ts
```diff
@@ -3,23 +3,16 @@
...
     onrejected?: (error: unknown) => R2 | PromiseLike<R2>,
-    transactionId?: number,
+    txId?: string | PromiseLike<number>,
   ): Promise<R1 | R2>
-
-  /**
-   * Called when executing a batch of regular tx
-   * @param id for regular tx ids
-   */
-  requestTransaction?(id: number): PromiseLike<unknown>
```
Jolg42, line 24: Technically, this renaming from `transactionId` to `txId` and type change could be a breaking change for users using `$transaction` which is GA?

Example
https://github.com/juanluispaz/ts-sql-query/blob/0d73da992a8f8ce396044f75fe51095aea2a2ed5/src/queryRunners/PrismaQueryRunner.ts#L286-L337

Also how can it be a string now?

Jolg42: Example
https://github.com/juanluispaz/ts-sql-query/blob/0d73da992a8f8ce396044f75fe51095aea2a2ed5/src/queryRunners/PrismaQueryRunner.ts#L286-L337

Jolg42: Also the removal of `requestTransaction` (to consider as breaking change)

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -32,17 +32,30 @@ export type RequestParams = {
...
+  const inTx = requests[0].runInTransaction
+  const headers = requests[0].headers
+
+  // if the tx has a number for an id, then it's a regular batch tx
+  const _inTx = typeof txId === 'number' && inTx ? true : undefined
+  // if the tx has a string for id, it's an interactive transaction
+  const _txId = typeof txId === 'string' && inTx ? txId : undefined
+
+  return { inTx: _inTx, headers: { transactionId: _txId, ...headers } }
+}
```
Jolg42, line 57: How is it a string and not a number for interactive transaction?

packages/client/src/runtime/core/request/PrismaPromise.ts
Jolg42: about the type string I moved conversation to https://github.com/prisma/prisma/pull/11695/files#r811005220

packages/client/src/runtime/RequestHandler.ts
millsp: It's a GUID

Jolg42: So when interactive transaction go GA, would we have GUIDs everywhere? (and not number like current $transaction GA)

tsconfig.json
millsp: It is necessary because you want `eslint` to only use the `tsconfig.json` of the sub-projects. Right now, it uses the base `tsconfig.json` for all sub-projects. But each project can have it's own custom inclusions and exclusions, or configuration that should be picked up per project/package.

* First, I added this config to the base `.eslintrc`
`project: ['./tsconfig.json', './packages/*/tsconfig.json'],`
Without this, `eslint` would use the base `tsconfig.json` for all packages, but that's not what we want.

* Now `eslint` applies per project, and we can exclude the sub-projects from the root (for performance).

Jolg42: Great! Thanks for the explanation ❤️

packages/client/src/runtime/core/request/PrismaPromise.ts
millsp: Fixed

packages/client/src/runtime/RequestHandler.ts
millsp: So I restored the behavior, tx ids are `string` and continue going through `then` and regular txs have `number` ids and go through `requestTransaction`.

helpers/compile/build.ts
```diff
@@ -170,48 +170,62 @@ export async function build(options: BuildOptions[]) {
...
+    }
   })
 
   return undefined
 }
 
 // Utils ::::::::::::::::::::::::::::::::::::::::::::::::::
 
+// so that only one watcher restarts a full build
+let watchLock = false
```
Jolg42, line 227: Nice one!

packages/client/src/__tests__/batching.test.ts
```diff
@@ -1,21 +1,23 @@
...
   test('basic batching', async () => {
     const dmmf = new DMMFClass(await getDMMF({ datamodel: blog }))
     const batches: any[] = []
     const requests: any[] = []
 
-    const fetcher = new PrismaClientFetcher({
+    const fetcher = new RequestHandler({
       $connect: () => Promise.resolve(),
       _engine: {
+        // @ts-expect-error
```
Jolg42, line 15:
```suggestion
        // @ts-expect-error
```
Why do we need this now? (and in other places) 👀

packages/client/src/__tests__/integration/happy/interactive-transactions-sqlite/test.ts
```diff
@@ -442,6 +443,38 @@ describe('interactive transactions', () => {
...
+            data: {
+              email: 'user_2@website.com',
+            },
+          })
+          .then()
+          .then()
+
+        throw new Error('rollback')
+      })
+      .catch(() => {})
```
Jolg42, line 472: Something like this would be nice
```suggestion
    const transaction = prisma
      .$transaction(async (prisma) => {
        await prisma.user
          .create({
            data: {
              email: 'user_1@website.com',
            },
          })
          .then()

        await prisma.user
          .create({
            data: {
              email: 'user_2@website.com',
            },
          })
          .then()
          .then()

        throw new Error('rollback')
      })
      .catch(() => {})
      await expect(transaction).rejects.toThrowErrorMatchingInlineSnapshot(``)
```

packages/client/src/runtime/core/request/PrismaPromise.ts
```diff
@@ -3,23 +3,22 @@
...
    * @param onfulfilled same as regular promises
    * @param onrejected same as regular promises
-   * @param transactionId for interactive tx ids
+   * @param txId string for iTX & Promise for regular tx
    */
   then<R1 = A, R2 = never>(
     onfulfilled?: (value: A) => R1 | PromiseLike<R1>,
     onrejected?: (error: unknown) => R2 | PromiseLike<R2>,
-    transactionId?: number,
+    txId?: string,
```
Jolg42, line 16: This looks like a breaking change in the public type?

packages/client/src/runtime/core/transaction/utils/getLockCountPromise.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Creates a an exotic promise that opens after x `await`s.
```
Jolg42, line 2:
```suggestion
 * Creates an exotic promise that opens after x `await`s.
```

packages/client/src/runtime/core/request/PrismaPromise.ts
millsp: Actually, it was always typed wrong. That's the tx id for itxs, so it's always a string.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1060,34 +950,30 @@ new PrismaClient({
...
-        const _requests = requests.map((request) => {
-          return new Promise((resolve, reject) => {
-            // each request has already been called with `prisma.<call>`
-            // so we inject `transactionId` by intercepting that promise
-            ;(request as any).then(resolve, reject, transactionId)
-          })
-        })
+    async $transaction(input: any, options?: any) {
+      // TODO: remove this once interactive tx became GA
+      if (!this._hasPreviewFlag('interactiveTransactions')) {
```
Jolg42, line 968: Since preview flags are case-insensitive, I would be curious to know if this condition is still triggered with for example `InTeRaCtiveTransactionS`

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1123,11 +1009,6 @@ new PrismaClient({
           // before we send the execution request, we use the changed params
           const changedInternalParams = { ...internalParams, ...changedParams }
 
-          // TODO remove this once LRT is the default transaction mode
-          if (index > 0 && !this._hasPreviewFlag('interactiveTransactions')) {
-            delete changedInternalParams['transactionId']
-          }
```
Jolg42, line 1129: Do we need to undo this?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1274,25 +1157,21 @@ function transactionProxy<T>(thing: T, transactionId: string): T {
...
       if (typeof target[prop] === 'function') {
-        // we override & handle every function call within the proxy
         return (...args: unknown[]) => {
-          if (prop === 'then') {
-            // this is our promise, we pass it an extra info argument
-            // this will call "our" `then` which will call `_request`
-            return target[prop](...args, transactionId)
-          }
+          // we hijack `then` calls to pass txId to prisma promises
+          if (prop === 'then') return target[prop](args[0], args[1], txId)
```
Jolg42, line 1166: So if there is only a `.catch()` we don't execute. But it's an internal detail maybe and not user facing?

packages/client/src/runtime/getPrismaClient.ts
millsp: No it is not needed anymore. Now both transactions can have a tx id that is either a `string` or a `number`:
* if it's a string, then it's a GUID for an interactive transaction
* if it's a number, then it's an id created by regular transactions

Depending on what kind of transaction it is, we tell different things to the QE:
* if it's an iTX, then the query is sent with a `transactionId` header to the QE
* if it's a rTX, then we send it as a batch to the QE with `runInTransaction = true`

So to answer the question, no I don't think this should be undone.
`delete changedInternalParams['transactionId']`

This could actually cause batching errors because tx would be id-less, see below:
* https://github.com/prisma/prisma/blob/b0af104d16accd2ed92b0ca1ff73847db1cce3b0/packages/client/src/runtime/RequestHandler.ts/#L81
* That means that non-awaited `$transaction` calls, could cause batches to be mixed up
  * It might be hard to test but I have left a todo for myself after this PR

packages/client/src/runtime/getPrismaClient.ts
millsp: That's a good point, I originally modeled this after the existing transactions https://github.com/prisma/prisma/blob/b0af104d16accd2ed92b0ca1ff73847db1cce3b0/packages/client/src/runtime/core/request/createPrismaPromise.ts/#L37-L42
But since the user can actually call `catch` and `finally`, I will fix and add tests for it.
I guess no one has run into this, as most only use `await`, but I'm going to improve this.

packages/client/src/__tests__/batching.test.ts
millsp: `RequestHandler` is now better typed and inherits from the poor internal types that we have for `Client`. That is enough though, as `_engine` is not public API and used to resolve to `any`.

packages/client/src/runtime/getPrismaClient.ts
millsp: I added tests for this.

packages/client/src/runtime/getPrismaClient.ts
millsp: It works :stuck_out_tongue:

Jolg42 (approved):
Nice work! Let's get this in 3.10.0 release 🚢

PR #11696, opened 2022-02-08
ci: Slack Notification on Failure

Jolg42:
With this PR we will now get notified if any of these actions fails in Slack

A `SLACK_WEBHOOK` env var was added to https://github.com/prisma/prisma/settings/secrets/actions


Jolg42:
Example run with failure
https://github.com/prisma/prisma/actions/runs/1812532373

And notification

aqrln (approved):

PR #11720, opened 2022-02-09
fix(client): prisma promise multiple then calls

millsp:
Fixes https://github.com/prisma/prisma/issues/11686

PR #11722, opened 2022-02-09
fix(client): query chaining then catch finally

millsp:
Follows https://github.com/prisma/prisma/pull/11720


Jolg42 (approved):
🚢 Let's get that into a patch release!

PR #11727, opened 2022-02-09
fix(client): allow mutations

### LINKED ISSUE #11628: Cannot extend PrismaClient class
### Bug description

In version 3.9 of the prisma client it's not possible to extend the PrismaClient functionalities:

```ts
class MyClient extends PrismaClient {
     constructor() {
          super()
     }
}
```

Only extending has no side-effects, however every new added property to `MyClass` becomes `readonly`, even though Typescript does not complain on compilation resulting on a runtime TypeError: cannot assign to read only  `property` of object `#<MyClass>`

This behavior is non existent in version 3.8 and one can extend PrismaClient normally.

My problematic code from my actual application:

```ts
export class TenantConnection extends PrismaClient implements IPrismaConnection {
  private hasCleanedUp = false; // <- problematic field, even though any other attempt to change any other property results on the same behavior

  constructor(
    connectionUrl: string,
    private readonly connectionInfo: { name: string; ipAddress?: string },
    private readonly onDisconnect?: (prismaConnection: TenantConnection) => Promise<void>,
  ) {
    super({ datasources: { db: { url: connectionUrl } } });
  }

  get $connectionName() {
    return this.connectionInfo.name;
  }

  get $ipAddress() {
    return this.connectionInfo.ipAddress;
  }

  /**
   *  Executes a function and disconnects the client.
   *  @param fn function to be executed before disconnecting
   */
  async $executeAndDisconnect<T = unknown>(fn: () => Promise<T>): Promise<T> {
    try {
      const result = await fn();
      return result;
    } catch (error) {
      throw error;
    } finally {
      this.$disconnect();
    }
  }

async $disconnect() {
    if (this.onDisconnect && !this.hasCleanedUp) {
      this.hasCleanedUp = true; // <- cannot assign to readonly property, even though it's not readonly
      await this.onDisconnect(this);
      return;
    }
    await super.$disconnect();
  }

}
```

The interface is irrelevant, it only says to implement the `$disconnect` and `executeAndDisconnect` functions

### How to reproduce

Create a PrismaClient in version 3.9.
Create a class that extends PrismaClient with new properties.
Instantiate this new class instead of PrismaClient.
Try to change any properties from the new class.

### Expected behavior

It would be a common assumption in OOP that one could extend a class to override or extend the base functionality of a module, so it can be less dependent on the third party code, my current class for example, existed before I started to use prisma, and it was fundamental for me to be able to switch to prisma without having to refactor thousands lines of code.

### Prisma information

Schema is irrelevant it's not necessary to make any query.

### Environment & setup

- OS: Windows
- Database: MySql
- Node.js version: 16.3

### Prisma Version

```
prisma                  : 3.8.1
@prisma/client          : 3.9.1
Query Engine (Node-API) : libquery-engine 34df67547cf5598f5a6cd3eb45f14ee70c3fb86f (at node_modules\@prisma\engines\query_engine-windows.dll.node)      
Migration Engine        : migration-engine-cli 34df67547cf5598f5a6cd3eb45f14ee70c3fb86f (at node_modules\@prisma\engines\migration-engine-windows.exe)  
Introspection Engine    : introspection-core 34df67547cf5598f5a6cd3eb45f14ee70c3fb86f (at node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary           : prisma-fmt 34df67547cf5598f5a6cd3eb45f14ee70c3fb86f (at node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash    : 34df67547cf5598f5a6cd3eb45f14ee70c3fb86f
Studio                  : 0.452.0
```

### LINKED ISSUE #11655: prisma v3.9.1: Get TypeError when use jest.fn().mockReturnValue.
### Bug description

- I got the below error with `v3.9.1` but didn't get the error with `v3.8.1`.

```js
TypeError: 'get' on proxy: property 'findUnique' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected 'function () {
        return fn.apply(this, arguments);
      }' but got '(userArgs) => {
    const nextDataPath = getNextDataPath(fluentPropName, prevDataPath);
    const nextUserArgs ...<omitted>... }')
```

## whorkwround

- downgrade to `v3.8.1`

### How to reproduce

1. create unit test code like`prisma.tableName.findUnique = jest.fn().mockReturnValue(mockdata)` 
2. run unit test, get the above error.

### Expected behavior

- No error like prisma v3.8.1

### Prisma information

- I think this issue is no relation with schema condition.

### Environment & setup

- OS: Mac OS
- Database: MySQL5,7
- Node.js version: 16.13.2

### Prisma Version

```
prisma                  : 3.9.1
@prisma/client          : 3.9.1
Query Engine (Node-API) : libquery-engine bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
```

millsp:
Fixes https://github.com/prisma/prisma/issues/11655
Fixes https://github.com/prisma/prisma/issues/11628


Jolg42 (approved):

Jolg42 (changes requested):
Blocked until we get these 2 jobs passing again, currently timing out only for binary at 25min

- CI / client (binary, ubuntu-latest, 12) (pull_request) Cancelled after 25m — client (binary, ubuntu-latest, 12)
- CI / client (binary, ubuntu-latest, 16) (pull_request) Cancelled after 25m — client (binary, ubuntu-latest, 16)

Jolg42 (approved):

PR #11802, opened 2022-02-14
fix(migrate): Return responses to migration engine print requests

tomhoule:
We do this to ensure a strict order within a command invocation. With a
full request-response cycle, the engine has to wait for the response to
the `print` request before it can move on with the command it is
executing, ensuring that all requests are processed before the engine
returns.

This makes it unnecessary to handle ME RPC requests asynchronously in
the CLI.

depends on https://github.com/prisma/prisma-engines/pull/2686


packages/migrate/src/MigrateEngine.ts
```diff
@@ -17,9 +17,23 @@ export interface MigrateEngineOptions {
...
-  jsonrpc: string
+/// A JSON-RPC request or response.
+export interface RpcRequestResponse {
+    id: number
+    jsonrpc: "2.0"
+}
+
+interface RpcSuccessResponse<T> extends RpcRequestResponse {
+    result: T
+}
```
tomhoule, line 28: should this be `result: T; error: never`? (and same below)

Jolg42: The specification says 

> result
    This member is REQUIRED on success.
    This member MUST NOT exist if there was an error invoking the method.
    The value of this member is determined by the method invoked on the Server.

> error
    This member is REQUIRED on error.
    This member MUST NOT exist if there was no error triggered during invocation.
    The value for this member MUST be an Object as defined in section 5.1. 

So they are mutually exclusive and one is present to other is not, so no need for that

Jolg42 (approved):

packages/migrate/src/MigrateEngine.ts
tomhoule: The `never` was about enforcing that — with what I wrote, `Success | Error` would accept an object where both are defined, no?

Jolg42: required with never doesn't work and errors with 
```
Type '{ id: number; jsonrpc: "2.0"; result: {}; }' is not assignable to type 'RpcResponse<{}, never>'.
  Property 'error' is missing in type '{ id: number; jsonrpc: "2.0"; result: {}; }' but required in type 'RpcErrorResponse<never>'.(2322)
```

optional kinda work for this 
See [playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgJTG32AEcBXYAZxhKsiSrwG8AoORAEwC44lyAtgCM07OACsqEJFCK8A5ACYAdAAZ5rAL6tWoSLEQo0WXASIBlctlxUqdBkwA8AFQB8cUKiScqZ4mUoae2kmODYOKGpyABsYXmcxNChoAH5eJGAAN1FtVmRUDBw8QmwAUShkqGDGYBd3T2BvXxKSCmpaagcWRIroeLFIqhiYNL4snJ09aHgYAE8wYqJqpzc4AF4-S2tqO06Q2tWAHz9yyuWD1x1sEPgh7dteFr2ax2ZNdw3wrl4ARgBmAAsfwANGJJNJZNgFCp1KCIlFYrw3nC4AB6VEeXpQJGaNEYoQQGAACzg2EwSCQhLgIjgnGA6GQwE4WjgOgQ6DgAAo7jYqMpBsMAJRhMTouDmIkQGKcal4QZdMQ8nb8hEwHQcMUABWSC1gszg8mijQA5sT5LSINQ+FTQAgaHBpHA5gsDRlslB5MpFVZecokqllEakKaiVo2RzuT7lf6oMKvlqdWg5gaY+bOJbfJT4Lb7Y7nXh5G60J7vfc+TGw1cbpjKo8ls8nG8PiKOAgeHB-kCUeCZHIDTD5CixQLEWFNCiYzi8dTCSSyRSqTS6QyMszceGuTGVUNYnHRRiJVLojKafL9j1Ktvhurp9qILrk4aTWaLVasx4QHb4Hn5gWix6vQ4LcYxSQNn1DXJ2U3LE-SxPcNQxO8H31eRU1fTMbU-XMkCdX9XXGACL2gWDKjDIA)

Issue #10512, opened 2021-11-30
Prisma client cannot find datamodel path ("schema.prisma")


0x77dev:
### Bug description

In a monorepo environment, Prisma generates into the `prisma` folder at service root, but when the `PrismaClient` instance is created, it tries to read the schema file in random subdirectories.

That happened after upgrading Prisma to `3.x`.

### How to reproduce

Here is a repo with an example of project configuration:

https://github.com/0x77dev/prisma-datamodel-path

To reproduce:
```shell
yarn install
yarn reproduce
```

### Expected behavior

Schema to be accessed at `join(process.cwd(), 'prisma', 'schema.prisma')`.

### Schema:

```prisma
generator client {
  provider = "prisma-client-js"
  output   = "."
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

model User {
  id String @id @default(uuid()) @db.Uuid
}
```

### PrismaClient
```typescript
export const get = async (): Promise<PrismaClient> => {
    const { dbURL } = init()

    const { PrismaClient } = await import(join(process.cwd(), 'prisma'))

    const prisma = new PrismaClient({
      datasources: { db: { url: dbURL } },
      log:
        process.env.NODE_ENV === 'production' ? [] : ['query', 'info', 'warn'],
      rejectOnNotFound,
    })

    await prisma.$connect()

    return prisma
}
```

### Environment & setup

- OS: Ubuntu (docker image `ubuntu:latest`) (`Linux aef42a236834 5.10.47-linuxkit #1 SMP Sat Jul 3 21:51:47 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`) or macOS Monterey `v12.0.1` (`Darwin 3c7d0a05571a.lan 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64 x86_64`)
- Database: `PostgreSQL 12`
- Node.js version: `v16.13.0`

### Prisma Version

```
prisma                  : 3.5.0
@prisma/client          : 3.5.0
Current platform        : darwin
Query Engine (Node-API) : libquery-engine 78a5df6def6943431f4c022e1428dbc3e833cf8e (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
[...]

PR #11829, opened 2022-02-15
fix(client-generation): avoid empty strings on buildDirNameFind (#10512)

cesconix:
Hi! The issue #10512 seems to be related to the generated js code on prisma client index.js file. I paste the generated part involved (I added a comment):
```js
    "prisma-client",
    "" // 👈 empty string is the cause of the problem! 
```
In `findSync` function the second argument is the "match" parameter (an array of strings or regexes) which is used to find a match with the file paths in the root directory during the generation. In my case this function should returns the fallback `__dirname` but returns another directory path bacause the `findSync` function calls the `isMatched` function that returns `true` for empty string.

```js
function isMatched(string: string, regexs: (RegExp | string)[]) {
  for (const regex of regexs) {
    if (typeof regex === 'string') {
      if (string.includes(regex)) { // 👈 here the problem with the empty string:
        // `string` variable contains a file system path and 
        // `regex` variable contains the empty string: the condition is always true! 💣
        return true
      }
    } else if (regex.exec(string)) {
      return true
    }
  }
  return false
}
```

The root cause is on the generation of the empty string in the buildDirNameFind function that has the generation logic for the serverless environments. I hope this help!

### DIFF
--- a/packages/client/src/generation/utils/buildDirname.ts
+++ b/packages/client/src/generation/utils/buildDirname.ts
@@ -30,15 +30,18 @@ export function buildDirname(clientEngineType: ClientEngineType, relativeOutdir:
  */
 function buildDirnameFind(relativeOutdir: string, runtimePath: string) {
   // potential client location on serverless envs
-  const slsRelativeOutputDir = relativeOutdir.split(path.sep).slice(1).join(path.sep)
+  const jss = [JSON.stringify(relativeOutdir)]
+
+  const dirs = relativeOutdir.split(path.sep)
+  if (dirs.length > 1) {
+    const slsRelativeOutdir = dirs.slice(1).join(path.sep)
+    jss.push(JSON.stringify(slsRelativeOutdir))
+  }
 
   return `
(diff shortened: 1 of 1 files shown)


millsp:
Hey @cesconix, as pointed in this comment https://github.com/prisma/prisma/pull/12000#issuecomment-1049909013, the fix I merged did not resolve https://github.com/prisma/prisma/issues/10512. I am essentially doing the same as you, in a different way. So it looks like we need to keep digging. Thanks :)

jensmeindertsma:
What can I do to get this merged?

millsp:
We will need to further investigate why the issue is happening, we already fixed empty string issues via another PR and it has not resolved https://github.com/prisma/prisma/issues/10512.

millsp:
https://github.com/prisma/prisma/issues/10512 has been fixed and you can try it out on `@prisma/client@dev`. Thanks for looking into this @cesconix and happy to see future contributions from you. Thanks:

PR #11848, opened 2022-02-16
chore: use swc instead of ts-jest for some packages

Jolg42:
See https://github.com/pnpm/pnpm/issues/2174


package.json
```diff
@@ -85,9 +86,9 @@
     "util": "0.12.4"
   },
   "lint-staged": {
-    "*.{md,yml,yaml,json}": "prettier --write",
```
Jolg42, line 88: Only yaml file is pnpm-lock

packages/sdk/package.json
```diff
@@ -21,7 +21,7 @@
   "scripts": {
     "dev": "DEV=true node -r esbuild-register helpers/build.ts",
     "build": "node -r esbuild-register helpers/build.ts",
-    "prepublishOnly": "pnpm run build && pnpm run test",
+    "prepublishOnly": "pnpm run build",
```
millsp, line 24: Could we remove this too as we already did a build before this?

Jolg42: Out of scope of this PR I would say

millsp: Fair enough :) I'll come with a followup PR

package.json
```diff
@@ -73,6 +73,7 @@
     "prettier": "2.5.1",
     "redis": "3.1.2",
     "redis-lock": "0.1.4",
+    "regenerator-runtime": "^0.13.9",
```
millsp, line 76:
```suggestion
    "regenerator-runtime": "0.13.9",
```

millsp (approved):
Nice cleanup and test boost :rocket: Thank you!

package.json
Jolg42: I'll let renovate do it, because it's going to need a lockfile update

PR #11880, opened 2022-02-17
fix(db pull): check if mongodb reintrospection and error early

### LINKED ISSUE #11185: MongoDB Introspection connects to database even when failing with error message that should not require connection to database at all
Minutes to get to a result that would not even require talking to the database:
image

(Slowness probably caused by https://github.com/prisma/prisma/issues/11340 - not topic of this issue)

Why does this even connect in this case?

Jolg42:
Closes https://github.com/prisma/prisma/issues/11185


packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -707,8 +730,6 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
       Run prisma generate to generate Prisma Client.
       Prisma schema loaded from prisma/no-model.prisma
       Datasource "my_db"
-
-      Introspecting based on datasource defined in prisma/no-model.prisma …
```
janpio, line 711: Huh what causes this change? Why is the `Prisma schema loaded from prisma/no-model.prisma` and `Datasource "my_db"` before still present?

janpio: Shouldn't everything from line 717 not be present?

Jolg42: So the error throws before we print the log https://github.com/prisma/prisma/pull/11880/files#diff-9eb32f67be2f4d93af4eb94473521226af5a1172b168031ef4002b934e15f21bR199

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -551,55 +551,80 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
   })
 
+  test('introspection --force', async () => {
+    ctx.fixture('schema-only-mongodb')
+    const introspect = new DbPull()
+    await introspect.parse(['--force'])
+    expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchSnapshot()
```
janpio, line 558: Should we maybe pull in these snapshots as InlineSnapshots as well to make it more obvious what the full result it?

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -630,19 +655,19 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
   test('introspection --print --composite-type-depth=-1', async () => {
```
janpio, line 655:
```suggestion
  test('introspection --print --composite-type-depth=-1 (no existing models)', async () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -551,55 +551,80 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
...
-                `)
+                                    // *** WARNING ***
+                                    // 
+                                    // The following fields had data stored in multiple types. The most common type was chosen. If loading data with a type that does not match the one in the data model, the client will crash. Please see the issue: https://github.com/prisma/prisma/issues/9654
+                                    // - Model "users", field: "numberOrString1", chosen data type: "Int32"
+                                    // 
+                        `)
   })
 
   test('introspection --print --composite-type-depth=1', async () => {
```
janpio, line 613:
```suggestion
  test('introspection --print --composite-type-depth=1 (no existing models)', async () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -551,55 +551,80 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
...
+      - Model "users", field: "numberOrString1", chosen data type: "Int32"
+      - Type "UsersHobbies", field: "numberOrString2", chosen data type: "Int32"
+      - Type "UsersHobbiesObjects", field: "numberOrString3", chosen data type: "Int32"
+
+      Run prisma generate to generate Prisma Client.
+    `)
+    expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+  })
+
   test('introspection --print', async () => {
```
janpio, line 579:
```suggestion
  test('introspection --print (no existing models)', async () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -551,55 +551,80 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
   })
 
+  test('introspection --force', async () => {
```
janpio, line 554:
```suggestion
  test('introspection --force (existing models)', async () => {
```

janpio: Line 529 could also have the `(no existing models)` in the test name.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -551,55 +551,80 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
...
+                  // 
+                  // The following fields had data stored in multiple types. The most common type was chosen. If loading data with a type that does not match the one in the data model, the client will crash. Please see the issue: https://github.com/prisma/prisma/issues/9654
+                  // - Model "users", field: "numberOrString1", chosen data type: "Int32"
+                  // - Type "UsersHobbies", field: "numberOrString2", chosen data type: "Int32"
+                  // - Type "UsersHobbiesObjects", field: "numberOrString3", chosen data type: "Int32"
+                  // 
+            `)
   })
 
   test('introspection --print --composite-type-depth=0', async () => {
```
janpio, line 597:
```suggestion
  test('introspection --print --composite-type-depth=0 (no existing models)', async () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -551,55 +551,80 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
...
+                                    // *** WARNING ***
+                                    // 
+                                    // The following fields had data stored in multiple types. The most common type was chosen. If loading data with a type that does not match the one in the data model, the client will crash. Please see the issue: https://github.com/prisma/prisma/issues/9654
+                                    // - Model "users", field: "numberOrString1", chosen data type: "Int32"
+                                    // - Type "UsersHobbies", field: "numberOrString2", chosen data type: "Int32"
+                                    // 
+                        `)
   })
 
   test('introspection --force --composite-type-depth=-1', async () => {
```
janpio, line 630:
```suggestion
  test('introspection --force --composite-type-depth=-1 (existing models)', async () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -22,7 +22,8 @@ describe('common/sqlite', () => {
   test('basic introspection', async () => {
     ctx.fixture('introspection/sqlite')
     const introspect = new DbPull()
-    await introspect.parse(['--print'])
+    const result = introspect.parse(['--print'])
+    await expect(result).resolves.toMatchInlineSnapshot(``)
```
janpio, line 26: What does this test? No stdout? No error? What exactly is the return of this?

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -685,38 +846,19 @@ describeIf(process.platform !== 'win32' && !isMacOrWindowsCI)('MongoDB', () => {
...
     ctx.fixture('schema-only-mongodb')
     const introspect = new DbPull()
-    await introspect.parse(['--schema=./prisma/no-model.prisma'])
-    // now re-introspection
-    const result = introspect.parse(['--schema=./prisma/no-model.prisma'])
+    const result = introspect.parse([])
     await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
             Iterating on one schema using re-introspection with db pull is currently not supported with MongoDB provider (Preview).
             You can explicitely ignore and override your current local schema file with prisma db pull --force
             Some information will be lost (relations, comments, mapped fields, @ignore...), follow https://github.com/prisma/prisma/issues/9585 for more info.
```
janpio, line 856: This is still logged though, right? Would it show up in `expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(...)`?

janpio (approved):

packages/migrate/src/__tests__/DbPull.test.ts
janpio: Yeah that was a more general question - as you are repeating this code in every test - where and when is this _supposed_ return anything?

Jolg42: Since we always return an empty string for `.parse()` for `db pull` it will be like this everywhere 
https://github.com/prisma/prisma/blob/75ec7c42ed0825e108bfe98eec6e977ece581a61/packages/migrate/src/commands/DbPull.ts#L317:L317

PR #11897, opened 2022-02-18
ci: fix mongo db seed with --drop

Jolg42:
We got a few flaky runs, were seeding was run twice, probably because the service was reused in Buildkite and seeding executed twice.

and small cleanup renaming


Jolg42:
Looks like failing tests are just flaky

.buildkite/publish/docker-compose.yml
```diff
@@ -60,8 +60,8 @@ services:
       - mariadb
       - mssql
       - mongo
```
millsp, line 62:
```suggestion
      - mongodb
```

millsp (approved):

.buildkite/publish/docker-compose.yml
Jolg42: No, I didn't rename the client docker image, it's still called mongo

PR #11967, opened 2022-02-21
feat: BuildPulse flaky test detection

janpio:
This PR adds [BuildPulse](https://buildpulse.io/) integration to our test GitHub Actions workflow:

> Automatically detect flaky tests. Focus on the top offenders to quickly improve your builds. Save developer time and frustration. Regain trust in your test suite.

That sounds like something we can use right now. 

Currently it adds the reporter only to our no-docker MacOS runs, where I think we have the most flaky tests right now.

To make the BuildPulse integration smoother, I also extracted the `Client` tests from the current `no-docker` job into a `no-docker-client` at the cost of adding 4 jobs per workflow run. I think that is worth it to get a result quicker. (Along the way I resorted our `jest.config.js`s to be the same order of properties)

PS: BuildPulse is currently working on a Windows reporter, as soon as that is available we can also add it for the Windows matrix runs. If Buildpulse turns out to work correctly, we can also add the reporting to our other Linux jobs of course.

Thanks @jasonrudolph for the concierge service with this.


.github/workflows/test.yml
```diff
@@ -735,3 +735,13 @@ jobs:
           files: ./packages/generator-helper/src/__tests__/coverage/clover.xml
           flags: generator-helper,${{ matrix.os }},library,binary
           name: generator-helper-${{ matrix.os }}
+
+      - name: Upload test results to BuildPulse for flaky test detection
+        if: '!cancelled()' # Run this step even when the tests fail. Skip if the workflow is cancelled.
```
jasonrudolph, line 740: :wave: @janpio: If I'm reading the workflow correctly, this job runs on macOS and Windows:

https://github.com/prisma/prisma/blob/89cf96728993b1d06102317170956cf9345f2408/.github/workflows/test.yml#L587-L590

Given that there's [not yet a Windows binary for the BuildPulse test-reporter](https://github.com/Workshop64/buildpulse-action/issues/2#issuecomment-1047229576), how would you feel about adjusting this to be macOS-only for now?

I think the syntax will be something like this:

```suggestion
        if: ${{ runner.os == 'macos-11' && !cancelled() }} # Run this step even when the tests fail. Skip if the workflow is cancelled.
```

janpio: Reasonable suggestion to get it more working for now 👍

.github/workflows/test.yml
```diff
@@ -735,3 +735,13 @@ jobs:
...
+
+      - name: Upload test results to BuildPulse for flaky test detection
+        if: ${{ runner.os == 'macos-11' && !cancelled() }} # Run this step even when the tests fail. Skip if the workflow is cancelled.
+        uses: Workshop64/buildpulse-action@master
+        with:
+          account: 17219288
+          repository: 192925833
+          path: .
+          key: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
+          secret: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
```
jasonrudolph, line 747: Following up on the discussion in https://github.com/Workshop64/buildpulse-action/issues/2#issuecomment-1047229576, this change will use the test-reporter binary for macOS (amd64):

```suggestion
        env:
          BUILDPULSE_ACCESS_KEY_ID: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
          BUILDPULSE_SECRET_ACCESS_KEY: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
        run: |
          curl -fsSL https://get.buildpulse.io/test-reporter-darwin-amd64 > ./buildpulse-test-reporter
          chmod +x ./buildpulse-test-reporter
          ./buildpulse-test-reporter submit junit.xml --account-id 17219288 --repository-id 192925833
```

Note that this also updates the path from `.` to `junit.xml`. By default, jest-junit outputs the report to `junit.xml` in the current working directory [[docs](https://github.com/jest-community/jest-junit/blob/v13.0.0/README.md?plain=1#L60-L61)]. By specifying the exact path of the file (as opposed to specifying a directory), we can avoid spending any time looking through the directory structure for XML files. :-)

janpio: That looks good.

I think we are creating multiple `junit.xml` in different `packages/*` locations though, as the workflow runs tests in different working-directories. Can we make it look in multiple locations at the same time?

jasonrudolph: Oh, thanks for pointing that out. :bow:

In that case, the easiest option is to point to the `packages` directory and have the test-reporter locate the XML files. It'll still be super fast. :-)

```suggestion
        env:
          BUILDPULSE_ACCESS_KEY_ID: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
          BUILDPULSE_SECRET_ACCESS_KEY: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
        run: |
          curl -fsSL https://get.buildpulse.io/test-reporter-darwin-amd64 > ./buildpulse-test-reporter
          chmod +x ./buildpulse-test-reporter
          ./buildpulse-test-reporter submit packages --account-id 17219288 --repository-id 192925833
```

.github/workflows/test.yml
```diff
@@ -735,3 +720,97 @@ jobs:
...
+
+      - uses: codecov/codecov-action@v2
+        if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
+        with:
+          files: ./packages/client/src/__tests__/coverage/clover.xml
+          flags: client,${{ matrix.os }},${{ matrix.queryEngine }}
+          name: client-${{ matrix.os }}-${{ matrix.queryEngine }}
+
+      - name: Upload test results to BuildPulse for flaky test detection
+        if: ${{ !cancelled() }} # Run this step even when the tests fail. Skip if the workflow is cancelled.
```
jasonrudolph, line 809: 👋 @janpio: In our email chat, I mentioned that we should only run this step on branches that are in this repository (as opposed to running it in forks, since forks don't have access to secrets). To set this up, we can use use this check (similar to the [approach used in Homebrew/brew](https://github.com/Homebrew/brew/blob/01506f120abd0d258afeb942b56deb3dd82470a9/.github/workflows/tests.yml#L264)):

```suggestion
        # Only run this step for branches where where we have access to secrets.
        # Run this step even when the tests fail. Skip if the workflow is cancelled.
        if: startsWith(github.repository, 'prisma/') && !cancelled()
```

.github/workflows/test.yml
```diff
@@ -735,3 +720,97 @@ jobs:
           files: ./packages/generator-helper/src/__tests__/coverage/clover.xml
           flags: generator-helper,${{ matrix.os }},library,binary
           name: generator-helper-${{ matrix.os }}
+
+      - name: Upload test results to BuildPulse for flaky test detection
+        if: ${{ !cancelled() }} # Run this step even when the tests fail. Skip if the workflow is cancelled.
```
jasonrudolph, line 725: 👋 @janpio: In our email chat, I mentioned that we should only run this step on branches that are in this repository (as opposed to running it in forks, since forks don't have access to secrets). To set this up, we can use use this check (similar to the [approach used in Homebrew/brew](https://github.com/Homebrew/brew/blob/01506f120abd0d258afeb942b56deb3dd82470a9/.github/workflows/tests.yml#L264)):

```suggestion
        # Only run this step for branches where where we have access to secrets.
        # Run this step even when the tests fail. Skip if the workflow is cancelled.
        if: startsWith(github.repository, 'prisma/') && !cancelled()
```

jasonrudolph (commented):
👋 @janpio: A couple suggestions to make sure that forks don't run into any issues attempting to upload test results. I hope this helps. If you have any questions, just let me know. :bow:

janpio:
The last thing left @jasonrudolph is to get the check for the OS working (before it was always skipped for some reason) until you have the Windows reporter ready (and then we will need to adapt the script to download the correct binary of course). Then I think we would be good for a first merge to get actual test run data.

jasonrudolph:
@janpio: That sounds right to me. We need to teach these two lines to only run for macOS:

https://github.com/prisma/prisma/blob/2e140ecfb043e8313cb7a11d2702f9534849b48b/.github/workflows/test.yml#L727

https://github.com/prisma/prisma/blob/2e140ecfb043e8313cb7a11d2702f9534849b48b/.github/workflows/test.yml#L813

I'll see if I can figure out the right syntax, and I'll let you know what I find out.

.github/workflows/test.yml
```diff
@@ -735,3 +720,101 @@ jobs:
           files: ./packages/generator-helper/src/__tests__/coverage/clover.xml
           flags: generator-helper,${{ matrix.os }},library,binary
           name: generator-helper-${{ matrix.os }}
+
+      - name: Upload test results to BuildPulse for flaky test detection
+        # Only run this step for branches where where we have access to secrets.
+        # Run this step even when the tests fail. Skip if the workflow is cancelled.
+        if: startsWith(github.repository, 'prisma/') && !cancelled()
```
jasonrudolph, line 727:
```suggestion
        # Only run this step for branches where where we have access to secrets.
        # Only run this step on macOS for now.
        # Run this step even when the tests fail. Skip if the workflow is cancelled.
        if: startsWith(github.repository, 'prisma/') && runner.os == 'macOS' && !cancelled()
``````

File: .github/workflows/test.yml [hunk 6]
```diff
@@ -735,3 +720,101 @@ jobs:
...
+        if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
+        with:
+          files: ./packages/client/src/__tests__/coverage/clover.xml
+          flags: client,${{ matrix.os }},${{ matrix.queryEngine }}
+          name: client-${{ matrix.os }}-${{ matrix.queryEngine }}
+
+      - name: Upload test results to BuildPulse for flaky test detection
+        # Only run this step for branches where where we have access to secrets.
+        # Run this step even when the tests fail. Skip if the workflow is cancelled.
+        if: startsWith(github.repository, 'prisma/') && !cancelled()
```
jasonrudolph, line 813:
```suggestion
        # Only run this step for branches where where we have access to secrets.
        # Only run this step on macOS for now.
        # Run this step even when the tests fail. Skip if the workflow is cancelled.
        if: startsWith(github.repository, 'prisma/') && runner.os == 'macOS' && !cancelled()
``````

**Review by jasonrudolph — COMMENTED** [review]
Following up on https://github.com/prisma/prisma/pull/11967#issuecomment-1048103363, I think these suggestions will do the trick. 

It turns out that `runner.os` has the values documented at https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context, so we need to check for `macOS` as the value.

If we were using `matrix.os`, then we'd check for `macos-latest` as the value.

File: packages/migrate/jest.config.js [hunk 7]
```diff
@@ -6,6 +6,7 @@ module.exports = {
   coverageReporters: ['clover'],
   coverageDirectory: 'src/__tests__/coverage',
   collectCoverageFrom: ['src/**/*.ts', '!**/__tests__/**/*'],
+  snapshotSerializers: ['@prisma/sdk/src/utils/jestSnapshotSerializer'],
```
millsp, line 9: Nice!

janpio:
Thanks @jasonrudolph, that worked as intended - in the last commit we have a failure on Windows and one on MacOS - and only the MacOS one triggered the BuildPulse step and successfully reported.

We'll try to review and merge this now and see if we can get BuildPulse to recognize some of these as flaky (vs. currently "always broken" I guess 😛). Then we can also try to add Windows (when you have it working) and Linux via the other builds.

jasonrudolph:
> Thanks @jasonrudolph, that worked as intended

@janpio: That's great! Thanks for taking the time to get things set up. :zap:

millsp (approved):
Nice, thank you! Let's try this new tool out :rocket:

PR #12004, opened 2022-02-23
feat(client): orderBy for composite types

### LINKED ISSUE #2514: Implement order by composites
Similar to relations, we want users to be able to order by composites.
Concretely, this means that we want to be able to order by to-one composite fields and by aggregate for to-many composites, if possible.

Design Doc: https://www.notion.so/prismaio/Composite-OrderBy-f442231e19194f72a10e03079e3d05c7

millsp:
Closes https://github.com/prisma/prisma-engines/issues/2514


Jolg42:
I don't see any issue related to this, could you check if there is one or create it?

packages/client/src/__tests__/integration/happy/composites-mongo/update/list.ts
```diff
@@ -247,7 +247,7 @@ describeIf(!process.env.TEST_SKIP_MONGODB)('update > list', () => {
   /**
    * Simple deleteMany
    */
-  test('deleteMany', async () => {})
+  test.skip('deleteMany', async () => {})
```
Jolg42, line 250: Missing test? Can you create an issue?

packages/client/src/__tests__/integration/happy/composites-mongo/findMany/required.ts
```diff
@@ -90,4 +90,38 @@ describeIf(!process.env.TEST_SKIP_MONGODB)('findMany > required', () => {
...
+                vote: true,
+              },
+            ],
+          },
+          country: France,
+          id: 3aaaaaaaaaaaaaaaaaaaaaaa,
+        },
+      ]
+    `)
+  })
```
Jolg42, line 126: Since there is only one item in upvotes it's hard to say if the orderBy upvotes works here, could you add one more item?

packages/client/src/__tests__/integration/happy/composites-mongo/count/required.ts
```diff
@@ -0,0 +1,59 @@
...
+const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
+
+let PrismaClient, prisma
+
+const id = '9ccccccccccccccccccccccc'
+
+/**
+ * Test count operations on required composite fields
+ */
+describeIf(!process.env.TEST_SKIP_MONGODB)('count > required', () => {
```
Jolg42, line 12:
```suggestion
describeIf(!process.env.TEST_SKIP_MONGODB)('count: required', () => {
```
Nit: A bit weird as it reads as an operator

packages/client/src/__tests__/integration/happy/composites-mongo/findMany/required.ts
Jolg42: I think it applies to many tests here

packages/client/src/__tests__/integration/happy/composites-mongo/update/list.ts
millsp: These will come during this sprint :)

packages/client/src/__tests__/integration/happy/composites-mongo/count/required.ts
millsp: OK, I'll fix that in a followup PR after this is merged.

packages/client/src/__tests__/integration/happy/composites-mongo/findMany/required.ts
millsp: Fixed

packages/client/jest.config.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     '<rootDir>/src/__tests__/integration/happy/exhaustive-schema/generated-dmmf.ts',
     '<rootDir>/src/__tests__/integration/happy/exhaustive-schema-mongo/generated-dmmf.ts',
     '__helpers__/',
+    'utils/',
```
Jolg42, line 20: Is that to ignore this?
`packages/client/src/__tests__/integration/happy/composites-mongo/utils/build-data/commentOptionalPropDataA.ts`

Could you narrow it, since all utils folders everywhere will be ignored (and then if a utils folder has a test it will also get ignored?)

Jolg42 (approved):
Nice!

See comment about Jest config

aqrln (commented):
Retroactive LGTM

PR #12031, opened 2022-02-25
chore(codesee): remove uneeded action steps

Jolg42:
Causes noise like https://github.com/prisma/prisma/pull/12029

### DIFF
--- a/.github/workflows/codesee-arch-diagram.yml
+++ b/.github/workflows/codesee-arch-diagram.yml
@@ -26,35 +26,11 @@ jobs:
         id: detect-languages
         uses: Codesee-io/codesee-detect-languages-action@latest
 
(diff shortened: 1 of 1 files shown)


Jolg42:
Canceled tests, they are not testing the change here and fail because of Microsoft service problem
```
Creating network "docker_default" with the default driver
Pulling mssql (mcr.microsoft.com/mssql/server:2019-latest)...
Get "[mcr.microsoft.com/v2](https://mcr.microsoft.com/v2/)": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

janpio:
I let Codesee know that it would be nice to do this automatically (however that would work) or make it more obvious that this is an option.

PR #12044, opened 2022-02-25
feat(ci): Upload junit.xml from Jest as GH Actions build artifact and improve junit.xml naming

janpio:
Test runs are now called `macos-11/node-16/binary/no-docker/engine-core` or similar so the GH Actions workflow becomes better identifiable in Buildpulse (flaky test detection). Additionally the resulting `junit.xml` files are uploaded as artifacts in the workflow, so it is easier to debug (any analyze) them.

image

cc FYI @jasonrudolph

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -588,9 +588,6 @@ jobs:
         node: [16]
         queryEngine: ['library', 'binary']
 
-    env:
-      JEST_JUNIT_SUITE_NAME: '${{ matrix.os }}/node-${{ matrix.node }}/${{ matrix.queryEngine }}/${{ github.job }}'
-
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


jasonrudolph (commented):
> Test runs are now called `macos-11/node-16/binary/no-docker/engine-core` or similar so the GH Actions workflow becomes better identifiable in Buildpulse (flaky test detection).

Looks good.

I'll probably need to adjust the logic that @BuildPulse uses for displaying test names in the UI to ensure that it's taking full advantage of the rich information that's in the suite name. The JUnit XML reports uploaded by various test frameworks have a great deal of inconsistency in how they make use of the suite name, the class name, and test case name, so BuildPulse tries to intelligently find the most useful info to construct a heading and subheading for each test from the available info. With the changes in this pull request, you'll be providing richer info (which is great!) than most test frameworks produce in their JUnit XML reports. Once this is merged, I'll take a look at how the test cases are getting rendered in the UI and see if there are rendering adjustments I can make that would be useful for you.

> Additionally the resulting `junit.xml` files are uploaded as artifacts in the workflow, so it is easier to debug (any analyze) them.

Great idea! 🌟

Jolg42 (approved):

millsp (approved):

janpio:
That sounsd good @jasonrudolph - if there are better ways to do this, let me know - the env var feels a bit hacky, but if it works 🤷

PR #12048, opened 2022-02-26
fix(resolveBinary): Clarify, comment and fix a small bug

janpio:
- Adds some comment about what is going on
- Adds a few TODOs where it is not clear what is going on
- Removes a dead line of code
- Rewrites some code to make it clearer what is happening (multiple different paths are tried one after another)
- Fix the bug that error message only outputs 2 out of 4 tested paths
- Also add TODO to clean up soon when a pkg change is released

### DIFF
--- a/packages/sdk/src/resolveBinary.ts
+++ b/packages/sdk/src/resolveBinary.ts
@@ -73,7 +73,11 @@ export async function resolveBinary(name: BinaryType, proposedPath?: string): Pr
 
   // Still here? Could not find the engine, so error out.
   throw new Error(
-    `Could not find ${name} binary. Searched in ${prismaPath}, ${prismaPath2}, ${prismaPath3}, ${prismaPath4}`,
+    `Could not find ${name} binary. Searched in:
+- ${prismaPath}
+- ${prismaPath2}
+- ${prismaPath3}
+- ${prismaPath4}`,
   )
 }
 
--- a/packages/sdk/src/resolveBinary.ts
+++ b/packages/sdk/src/resolveBinary.ts
@@ -84,6 +84,8 @@ export async function maybeCopyToTmp(file: string): Promise<string> {
     // in this case, we are in a "pkg" context with a virtual fs
     // to make this work, we need to copy the binary to /tmp and execute it from there
     // TODO Why is this needed? What happens if you do not do it?
+    // TODO Probably to be able to make the file executable?
+    // TODO Go and Python Client (which use pkg) actually provide the binaries _outside_ of the CLI via env vars - so never and up here
     const targetDir = path.join(tempDir, 'prisma-binaries')
     await makeDir(targetDir)
     const target = path.join(targetDir, path.basename(file))
--- a/packages/sdk/src/resolveBinary.ts
+++ b/packages/sdk/src/resolveBinary.ts
@@ -31,7 +31,7 @@ export const engineEnvVarMap = {
 }
 export { BinaryType }
 export async function resolveBinary(name: BinaryType, proposedPath?: string): Promise<string> {
-  // if file exists at propostedPath (and does not start with /snapshot (= pkg), use that one
+  // if file exists at proposedPath (and does not start with `/snapshot/` (= pkg), use that one
   if (proposedPath && !proposedPath.startsWith('/snapshot/') && fs.existsSync(proposedPath)) {
     return proposedPath
   }
--- a/packages/sdk/src/resolveBinary.ts
+++ b/packages/sdk/src/resolveBinary.ts
@@ -31,12 +31,13 @@ export const engineEnvVarMap = {
 }
 export { BinaryType }
 export async function resolveBinary(name: BinaryType, proposedPath?: string): Promise<string> {
+  // if file exists at propostedPath (and does not start with /snapshot (= pkg), use that one
   if (proposedPath && !proposedPath.startsWith('/snapshot/') && fs.existsSync(proposedPath)) {
     return proposedPath
   }
 
+  // If engine path was provided via env var, check and use that one
   const envVar = engineEnvVarMap[name]
-
   if (process.env[envVar]) {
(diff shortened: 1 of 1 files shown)


packages/sdk/src/resolveBinary.ts
```diff
@@ -31,67 +31,69 @@ export const engineEnvVarMap = {
...
 export async function maybeCopyToTmp(file: string): Promise<string> {
-  // in this case, we are in a "pkg" context with a virtual fs
-  // to make this work, we need to copy the binary to /tmp and execute it from there
-
   const dir = eval('__dirname')
+  
   if (dir.startsWith('/snapshot/')) {
+    // in this case, we are in a "pkg" context with a virtual fs
+    // to make this work, we need to copy the binary to /tmp and execute it from there
+    // TODO Why is this needed? What happens if you do not do it?
```
janpio, line 86:
```suggestion
    // TODO Why is this needed? What happens if you do not do it?
    // TODO Probably to be able to make the file executable?
    // TODO Go and Python Client (which use pkg) actually provide the binaries _outside_ of the CLI via env vars - so never and up here
```

packages/sdk/src/resolveBinary.ts
```diff
@@ -31,67 +31,71 @@ export const engineEnvVarMap = {
 }
 export { BinaryType }
 export async function resolveBinary(name: BinaryType, proposedPath?: string): Promise<string> {
+  // if file exists at propostedPath (and does not start with /snapshot (= pkg), use that one
   if (proposedPath && !proposedPath.startsWith('/snapshot/') && fs.existsSync(proposedPath)) {
     return proposedPath
   }
```
Jolg42, line 37:
```suggestion
  // if file exists at proposedPath (and does not start with `/snapshot/` (= pkg), use that one
  if (proposedPath && !proposedPath.startsWith('/snapshot/') && fs.existsSync(proposedPath)) {
    return proposedPath
  }
```

packages/sdk/src/resolveBinary.ts
```diff
@@ -31,67 +31,71 @@ export const engineEnvVarMap = {
...
+  // for pkg (related: https://github.com/vercel/pkg#snapshot-filesystem)
+  const prismaPath2 = path.join(__dirname, '..', binaryName)
+  if (fs.existsSync(prismaPath2)) {
+    return maybeCopyToTmp(prismaPath2)
   }
 
-  prismaPath = path.join(__dirname, '../..', binaryName)
-  if (fs.existsSync(prismaPath)) {
-    return maybeCopyToTmp(prismaPath)
+  // TODO for ??
```
Jolg42, line 62: Looks like we check 2 levels up "for some reason"

packages/sdk/src/resolveBinary.ts
```diff
@@ -31,67 +31,71 @@ export const engineEnvVarMap = {
...
   }
 
+  // Still here? Could not find the engine, so error out.
   throw new Error(
-    `Could not find ${name} binary. Searched in ${path.join(dir, '..', binaryName)} and ${path.join(
-      dir,
-      '../..',
-      binaryName,
-    )}`,
+    `Could not find ${name} binary. Searched in ${prismaPath}, ${prismaPath2}, ${prismaPath3}, ${prismaPath4}`,
```
Jolg42, line 76:
```suggestion
    `Could not find ${name} binary. Searched in:
- ${prismaPath}
- ${prismaPath2}
- ${prismaPath3}
- ${prismaPath4}`,
```

Would maybe make it easier to read? Especially if paths are long?

packages/sdk/src/resolveBinary.ts
janpio: Yeah but what is the reason? :D I have no idea.

packages/sdk/src/resolveBinary.ts
```diff
@@ -31,67 +31,71 @@ export const engineEnvVarMap = {
...
+  const prismaPath3 = path.join(__dirname, '../..', binaryName)
+  if (fs.existsSync(prismaPath3)) {
+    return maybeCopyToTmp(prismaPath3)
   }
 
-  // needed to come from @prisma/client/generator-build to @prisma/client/runtime
-  prismaPath = path.join(__dirname, '../runtime', binaryName)
-  if (fs.existsSync(prismaPath)) {
-    return maybeCopyToTmp(prismaPath)
+  // TODO for ?? / needed to come from @prisma/client/generator-build to @prisma/client/runtime
```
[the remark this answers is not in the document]
janpio: If you are confident, add as suggestion to comment - but still deservs a TODO a think.

packages/sdk/src/resolveBinary.ts
```diff
@@ -31,67 +31,71 @@ export const engineEnvVarMap = {
...
     const targetDir = path.join(tempDir, 'prisma-binaries')
     await makeDir(targetDir)
     const target = path.join(targetDir, path.basename(file))
+    
+    // We have to read and write until https://github.com/zeit/pkg/issues/639 is resolved
     const data = await readFile(file)
     await writeFile(target, data)
-    // We have to read and write until https://github.com/zeit/pkg/issues/639
-    // is resolved
+    // TODO Undo when https://github.com/vercel/pkg/pull/1484 is released
```
Jolg42, line 96: Looks like it was closed with this commit, so resolved?
https://github.com/vercel/pkg/commit/72bc8b36ba3e674f4818ac8d57881cd161c0e4d7

packages/sdk/src/resolveBinary.ts
janpio: Does this keep line breaks like this in the output?

Jolg42 (approved):
Looks good, thanks!
We should look at the todo and figure some of these things out

packages/sdk/src/resolveBinary.ts
Jolg42: Yes it should :)

PR #12054, opened 2022-02-26
temp: Cross-platform BuildPulse GitHub Action

janpio:
cc @jasonrudolph 

As suggested in https://github.com/Workshop64/buildpulse-action/issues/2#issuecomment-1052684667

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,6 +44,7 @@ on:
       - 'graphs/**'
 
 env:
+  HAS_BUILDPULSE_SECRETS: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID != '' && secrets.BUILDPULSE_SECRET_ACCESS_KEY != '' }}
   PRISMA_TELEMETRY_INFORMATION: 'prisma test.yml'
 
 jobs:
@@ -724,7 +725,7 @@ jobs:
       - name: Upload test results to BuildPulse for flaky test detection
         # Only run this step for branches where where we have access to secrets.
         # Run this step even when the tests fail. Skip if the workflow is cancelled.
-        if: startsWith(github.repository, 'prisma/') && !cancelled()
+        if: env.HAS_BUILDPULSE_SECRETS == 'true' && !cancelled()
         uses: Workshop64/buildpulse-action@cross-platform
         with:
           account: 17219288
@@ -732,7 +733,6 @@ jobs:
           path: packages
           key: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
           secret: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
-
 
   #
   # Run Client tests on macOS and Windows. (Rasoning see above)
@@ -811,7 +811,7 @@ jobs:
       - name: Upload test results to BuildPulse for flaky test detection
         # Only run this step for branches where where we have access to secrets.
         # Run this step even when the tests fail. Skip if the workflow is cancelled.
-        if: startsWith(github.repository, 'prisma/') && !cancelled()
+        if: env.HAS_BUILDPULSE_SECRETS == 'true' && !cancelled()
         uses: Workshop64/buildpulse-action@cross-platform
         with:
           account: 17219288
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -723,16 +723,16 @@ jobs:
 
       - name: Upload test results to BuildPulse for flaky test detection
         # Only run this step for branches where where we have access to secrets.
-        # Only run this step on macOS for now.
         # Run this step even when the tests fail. Skip if the workflow is cancelled.
-        if: startsWith(github.repository, 'prisma/') && runner.os == 'macOS' && !cancelled()
-        env:
-          BUILDPULSE_ACCESS_KEY_ID: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
-          BUILDPULSE_SECRET_ACCESS_KEY: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
-        run: |
-          curl -fsSL https://get.buildpulse.io/test-reporter-darwin-amd64 > ./buildpulse-test-reporter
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -723,16 +723,16 @@ jobs:
...
         # Run this step even when the tests fail. Skip if the workflow is cancelled.
-        if: startsWith(github.repository, 'prisma/') && runner.os == 'macOS' && !cancelled()
-        env:
-          BUILDPULSE_ACCESS_KEY_ID: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
-          BUILDPULSE_SECRET_ACCESS_KEY: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
-        run: |
-          curl -fsSL https://get.buildpulse.io/test-reporter-darwin-amd64 > ./buildpulse-test-reporter
-          chmod +x ./buildpulse-test-reporter
-          ./buildpulse-test-reporter submit packages --account-id 17219288 --repository-id 192925833
+        if: startsWith(github.repository, 'prisma/') && && !cancelled()
```
janpio, line 727:
```suggestion
        if: startsWith(github.repository, 'prisma/') && !cancelled()
```
@jasonrudolph

jasonrudolph: Oops! Good catch. 🤦

jasonrudolph:
Thanks for trying this out, @janpio. The good news is that the action is running for macOS and Windows ([screenshot](https://user-images.githubusercontent.com/2988/155863402-d898e35a-6651-4dec-9c37-d368b1850665.png)), but the bad news is that it's failing due to not having access to the secrets. I was expecting this `if` conditional to _prevent_ the step from running since this branch is from a fork:

https://github.com/prisma/prisma/blob/33fd05e7584daee77ca901a864b927882d8d47b8/.github/workflows/test.yml#L727

However, the step is still running.

I'll looking into this failure and let you know what I find out.

.github/workflows/test.yml
```diff
@@ -723,16 +723,16 @@ jobs:
...
         # Run this step even when the tests fail. Skip if the workflow is cancelled.
-        if: startsWith(github.repository, 'prisma/') && runner.os == 'macOS' && !cancelled()
-        env:
-          BUILDPULSE_ACCESS_KEY_ID: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
-          BUILDPULSE_SECRET_ACCESS_KEY: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
-        run: |
-          curl -fsSL https://get.buildpulse.io/test-reporter-darwin-amd64 > ./buildpulse-test-reporter
-          chmod +x ./buildpulse-test-reporter
-          ./buildpulse-test-reporter submit packages --account-id 17219288 --repository-id 192925833
+        if: startsWith(github.repository, 'prisma/') && !cancelled()
```
janpio, line 727: This could/should work maybe:
```suggestion
        if: github.event.pull_request.head.repo.full_name == github.repository && !cancelled()
```

.github/workflows/test.yml
```diff
@@ -810,13 +810,12 @@ jobs:
...
         # Run this step even when the tests fail. Skip if the workflow is cancelled.
-        if: startsWith(github.repository, 'prisma/') && runner.os == 'macOS' && !cancelled()
-        env:
-          BUILDPULSE_ACCESS_KEY_ID: ${{ secrets.BUILDPULSE_ACCESS_KEY_ID }}
-          BUILDPULSE_SECRET_ACCESS_KEY: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
-        run: |
-          curl -fsSL https://get.buildpulse.io/test-reporter-darwin-amd64 > ./buildpulse-test-reporter
-          chmod +x ./buildpulse-test-reporter
-          ./buildpulse-test-reporter submit packages/client --account-id 17219288 --repository-id 192925833
+        if: startsWith(github.repository, 'prisma/') && !cancelled()
```
janpio, line 814: This could/should work maybe:
```suggestion
        if: github.event.pull_request.head.repo.full_name == github.repository && !cancelled()
```

janpio (commented):
Good we caught that! Left 2 suggestions with code that could possibly work here (via https://github.community/t/how-to-detect-a-pull-request-from-a-fork/18363/3)

jasonrudolph:
> Good we caught that! 

Indeed! :-)

> Left 2 suggestions with code that could possibly work here (via https://github.community/t/how-to-detect-a-pull-request-from-a-fork/18363/3)

Thanks. :bow: 

I gave that a try, but I discovered that it always returns false when there's no pull request. So any time there's a push with no pull request, it would skip the BuildPulse upload step. For example, when pushing to a branch before you have a pull request, it returns false and skips the step. I think the same thing would happen for the merge commit that's created when merging a pull request. That _seems_ undesirable to me, so I looked for another option.

Since we want to skip the step if secrets are unavailable, I looked to see if we could do that explicitly (as opposed to trying to check the organization name or the pull request info. At the moment, GitHub Actions doesn't support checking for secrets in a conditional (https://github.com/actions/runner/issues/1138), and it seems like [a lot of folks are running into this issue](https://github.com/actions/runner/issues/520). However, I came across [a solution that works in the meantime](https://github.com/jon-turney/meson/commit/af12c43843a8da6480487f4f42f10821399d0d82), and 8e92919d09d872091cb30ebc9971908da4b00067 adopts that approach. It's admittedly a bit of a hack, but I think it will meet our needs.

Since this branch is from a fork and therefore doesn't have access to the secrets, we can see that the BuildPulse step is skipped [here](https://github.com/prisma/prisma/runs/5347741810?check_suite_focus=true#step:20:50).

If the diff looks good to you, then I think the next step would be to add this code to a branch in this repository, and verify that the BuildPulse steps _do_ run.

@janpio: What do you think?

janpio:
Interesting. Feels like a pretty gaping hole in GH Actions then.

I'll do exactly that, merge this PR into a branch and then open a new PR for that one. Will @ mention you again.

Issue #11526, opened 2022-01-31
InvalidDatasourceError: Datasource URL should use prisma:// protocol


jacobhq:
I'm using prisma in nextjs, and the env urls start with `postgres://`. I'm working in a gitpod container. When I run `next dev` **or** build for production and run `next start`, I receive this error:

```
InvalidDatasourceError: Datasource URL should use prisma:// protocol
```

If I change my urls to start with `prisma://`, I get:

```
InvalidDatasourceError: No valid API key found in the datasource URL
```

Any ideas? Help would be much appreciated! I've listed my schema and versions of prisma below.

```
datasource db {
  provider             = "postgresql"
  url                  = env("DATABASE_URL")
  shadowDatabaseUrl    = env("DATABASE_MIGRATE_URL")
  referentialIntegrity = "prisma"
}

generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["dataproxy", "referentialIntegrity"]
}

model User {
  id           Int            @id @default(autoincrement())
  WishlistItem WishlistItem[]
}

model Product {
  id           Int            @id @default(autoincrement())
  name         String
  price        Int
  salePrice    Int            @default(0)
  imageUrl     String
  flag         String?
  description  String?
  buyUrl       String
  marketplace  String?
  WishlistItem WishlistItem[]
}

model WishlistItem {
  id        Int     @id @default(autoincrement())
  userId    Int
  productId Int
  Product   Product @relation(fields: [productId], references: [id])
  User      User    @relation(fields: [userId], references: [id])
}

```

```json
    "@prisma/client": "^3.8.1",
    "prisma": "3.8.1",
    "next": "12.0.8",
```

```js
export const getServerSideProps = async ({}) => {
  const prisma = new PrismaClient()
  const products = await prisma.product.findMany()
  return { props: { products } }
}
```

janpio:
How are you building your Prisma Client? From the preview features it seems you might be trying to use the Data Proxy, correct?

jacobhq:
Thanks for pointing that out! This is not a bug it's just the fact that I wasn't using the dataproxy. Thanks!

jacobhq:
> **TLDR**: I solved this by removing `dataproxy` from the `previewFeatures` part of my schema.

[...]

PR #12067, opened 2022-02-27
chore: Improve error message when URLs don't start with `prisma://` when using dataproxy

jacobhq:
This helps people with the headache I experienced in #11526.

### DIFF
--- a/packages/engine-core/src/data-proxy/DataProxyEngine.ts
+++ b/packages/engine-core/src/data-proxy/DataProxyEngine.ts
@@ -237,7 +237,7 @@ export class DataProxyEngine extends Engine {
     const { protocol, host, searchParams } = url
 
     if (protocol !== 'prisma:') {
-      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol. If you are not using the Data Proxy, remove the `dataProxy` from the `previewFeatures` from your schema', {
+      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol. If you are not using the Data Proxy, remove the `dataProxy` from the `previewFeatures` in your schema and ensure that `PRISMA_CLIENT_ENGINE_TYPE` environment variable is not set to `dataproxy`.', {
         clientVersion: this.clientVersion,
       })
     }
--- a/packages/engine-core/src/data-proxy/DataProxyEngine.ts
+++ b/packages/engine-core/src/data-proxy/DataProxyEngine.ts
@@ -237,7 +237,7 @@ export class DataProxyEngine extends Engine {
     const { protocol, host, searchParams } = url
 
     if (protocol !== 'prisma:') {
-      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol', {
+      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol. If you are not using the Data Proxy, remove it from your schema', {
         clientVersion: this.clientVersion,
       })
     }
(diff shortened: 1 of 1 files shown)


packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -237,7 +237,7 @@ export class DataProxyEngine extends Engine {
     const { protocol, host, searchParams } = url
 
     if (protocol !== 'prisma:') {
-      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol', {
+      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol. If you are not using the Data Proxy, remove it from your schema', {
         clientVersion: this.clientVersion,
       })
```
Jolg42, line 242:
```suggestion
      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol. If you are not using the Data Proxy, remove the `dataProxy` from the `previewFeatures` from your schema', {
        clientVersion: this.clientVersion,
      })
```

millsp:
Hey @jacobhq, I'm curious to know how you ran into this and why it has caused confusion. The `dataproxy` has to be enabled in two ways, the one is via the preview features, and the second is via `PRISMA_CLIENT_ENGINE_TYPE=dataproxy`. Have you forgotten removing the `PRISMA_CLIENT_ENGINE_TYPE=dataproxy` env var? If that is the case, that might be a more appropriate piece of info to display in that message and given that `dataProxy` is going to move out of preview in the future.

jacobhq:
@millsp I had copied the schema from another project where I was using the dataproxy, into a project where I wasn't, and hadn't realised. I wrote loads more in #11526 [here](https://github.com/prisma/prisma/issues/11526#issuecomment-1053698253)

millsp:
Thanks. Could you please share that schema with me?

jacobhq:
I did not have that env var set. Here is the schema:
```
datasource db {
  provider             = "postgresql"
  url                  = env("DATABASE_URL")
  shadowDatabaseUrl    = env("DATABASE_MIGRATE_URL")
}

generator client {
  previewFeatures = ["dataproxy", "referentialIntegrity"]
}

model User {
  id           Int            @id @default(autoincrement())
  WishlistItem WishlistItem[]
}

model Product {
  id           Int            @id @default(autoincrement())
  name         String
  price        Int
  salePrice    Int            @default(0)
  imageUrl     String
  flag         String?
  description  String?
  buyUrl       String
  marketplace  String?
  WishlistItem WishlistItem[]
}

model WishlistItem {
  id        Int     @id @default(autoincrement())
  userId    Int
  productId Int
  Product   Product @relation(fields: [productId], references: [id])
  User      User    @relation(fields: [userId], references: [id])
}
```

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -237,7 +237,7 @@ export class DataProxyEngine extends Engine {
     const { protocol, host, searchParams } = url
 
     if (protocol !== 'prisma:') {
-      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol', {
+      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol. If you are not using the Data Proxy, remove the `dataProxy` from the `previewFeatures` from your schema', {
```
aqrln, line 240:
```suggestion
      throw new InvalidDatasourceError('Datasource URL should use prisma:// protocol. If you are not using the Data Proxy, remove the `dataProxy` from the `previewFeatures` in your schema and ensure that `PRISMA_CLIENT_ENGINE_TYPE` environment variable is not set to `dataproxy`.', {
```

janpio: We will need to remember to update this later when/if we update to another mechanism to activate the data proxy client.

millsp (approved):

PR #12093, opened 2022-03-01
ci(buildpulse): Restructure junit suite name and junit.xml artifact filename

janpio:
Makes more sense this way - the job and sub step are most important, then only the matrix params.


Jolg42 (approved):
Nice!

PR #12135, opened 2022-03-03
fix(migrate diff): improve help output alignment

do4gr:
Just different alignment for the options.


Jolg42:
Very nice! 
Screen Shot 2022-03-03 at 16 52 11

Now we can do the same in the other commands when we have some time

Jolg42 (approved):

PR #12223, opened 2022-03-09
ci: Windows -RunAsAdmin

Jolg42:
Main currently failing with https://github.com/prisma/prisma/runs/5481573468?check_suite_focus=true
```
Run bash .github/workflows/setup-mysql.sh
+ '[' Windows = Windows ']'
+ pwsh .github/workflows/install-mysql.ps1
Initializing...
Running the installer as administrator is disabled by default, use -RunAsAdmin parameter if you know what you are doing.
Abort.
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/install-mysql.ps1
+++ b/.github/workflows/install-mysql.ps1
@@ -1,4 +1,5 @@
-Invoke-Expression (New-Object System.Net.WebClient).DownloadString('https://get.scoop.sh')
+iwr -useb 'https://get.scoop.sh' -outfile 'scoopinstaller.ps1'
+.\scoopinstaller.ps1 -RunAsAdmin
 
 scoop install mysql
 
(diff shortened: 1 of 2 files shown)


aqrln (approved):
Unexpected breaking change from scoop 😄
Thank you!

Jolg42:
Indeed!

janpio:
It should not really be necessary for Scoop to run as Admin though. That is why they discourage that and make you supply an additional parameter.

aqrln:
@janpio yes, but the user on GitHub Actions is a full Admin with UAC disabled, we don't get to choose that

janpio:
Yeah, turns out that is indeed a change on Scoop side somehow - also affected engines: https://github.com/prisma/prisma-engines/pull/2755

PR #12227, opened 2022-03-09
feat(migrate diff): add --exit-code

### LINKED ISSUE #9707: Detect schema drift helper / `migrate dev --exit-code`
## Problem

I want to add a CI-job to make sure no schema changes have been done without corresponding migration files have been added. Had a near-miss today where we could’ve borked a whole table since the autogenerated migration later contained a drop column.

## Suggested solution

Add an `--exit-code` that would fail `migrate dev` if there are changes (similar to `git diff --exit-code`)

## Alternatives

Another command, i.e. `migrate detect-drift`

## Additional context

I was asked to create this ticket on Slack - https://prisma.slack.com/archives/CBFFY3066/p1633945797065100

### LINKED ISSUE #11566: migrate diff: Provide a reliable way to detect empty diffs / migrations
## Problem

Diff returns the difference between two schemas. While often, you assume that there is a difference, and you are interested in the content, an obvious extension of this functionality is to detect _whether two schemas are the same_.

Analogous to text diffs like `git diff`, `migrated diff` can tell us if two database schemas are in sync. The only way to detect this programmatically at the moment is to expect a specific empty migration message. This solution is not elegant and subject to potential breaking changes.

## Suggested solution

- An `--exit-code` command like flag that would work like git diff's `--exit code` flag

> --exit-code
> Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.

## Alternatives

This is a thought I just had, more research should go into the suggested solution before we rush into implementation.

Jolg42:
Closes https://github.com/prisma/prisma/issues/11566
Closes https://github.com/prisma/prisma/issues/9707

### DIFF
--- a/packages/migrate/src/__tests__/MigrateDiff.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDiff.test.ts
@@ -101,9 +101,25 @@ describe('migrate diff', () => {
     it('should diff --from-url=file:doesnotexists.db --to-empty ', async () => {
       ctx.fixture('schema-only-sqlite')
 
-      const result = MigrateDiff.new().parse(['--preview-feature', '--from-empty', '--to-url=file:doesnotexists.db'])
+      const result = MigrateDiff.new().parse(['--preview-feature', '--from-url=file:doesnotexists.db', '--to-empty'])
       await expect(result).resolves.toMatchInlineSnapshot(``)
       expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(`No difference detected.`)
+    })
+    it('should fail if path does not exist', async () => {
+      ctx.fixture('schema-only-sqlite')
+
+      const result = MigrateDiff.new().parse([
+        '--preview-feature',
+        '--from-url=file:./something/doesnotexists.db',
+        '--to-empty',
+      ])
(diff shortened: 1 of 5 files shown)


do4gr (approved):

packages/migrate/src/__tests__/MigrateDiff.test.ts
```diff
@@ -208,6 +224,75 @@ describe('migrate diff', () => {
...
+        const result = MigrateDiff.new().parse([
+          '--preview-feature',
+          '--from-empty',
+          '--to-url=file:doesnotexists.db',
+          '--script',
+          '--exit-code',
+        ])
+
+        await expect(result).resolves.toMatchInlineSnapshot(``)
+        expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(`-- This is an empty migration.`)
```
tomhoule, line 293: Is this one missing the assertion on the exit code?

tomhoule (approved):

packages/migrate/src/__tests__/MigrateDiff.test.ts
Jolg42: Yes because we don't test normal exit codes directly anywhere, though `resolves` is a success and `rejects` is an error.
So here a success = 0

PR #12255, opened 2022-03-11
fix(client): include promise properties in fluent api own properties

### LINKED ISSUE #12252: `'then' in PrimsaPromise` returns false
### Bug description

Some libraries use `'then' in x` as part of their detection for promise-like/thenable objects.

The way prisma client builds its fluent APIs with proxies causes this to fail.

### How to reproduce

`console.log('then' in prisma.someModel.findUnique({ where: { id: 1 }))`

`console.log(typeof prisma.someModel.findUnique({ where: { id: 1 })).then === 'function')`

This is basically because ownKeys here https://github.com/prisma/prisma/blob/3855438b964b40edeb72a3815c641d97f8c433c4/packages/client/src/runtime/core/model/applyFluent.ts#L106 does not include `then` (or other promise methods)

### Expected behavior

`'then' in promise` should return true.

### Environment & setup

- OS: 
- Database: 
- Node.js version: 

### Prisma Version

```
prisma                  : 3.10.0
@prisma/client          : 3.10.0
Query Engine (Node-API) : libquery-engine 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at ../../node_modules/.pnpm/@prisma+engines@3.10.0-50.73e60b76d394f8d37d8ebd1f8918c79029f0db86/node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at ../../node_modules/.pnpm/@prisma+engines@3.10.0-50.73e60b76d394f8d37d8ebd1f8918c79029f0db86/node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at ../../node_modules/.pnpm/@prisma+engines@3.10.0-50.73e60b76d394f8d37d8ebd1f8918c79029f0db86/node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary           : prisma-fmt 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at ../../node_modules/.pnpm/@prisma+engines@3.10.0-50.73e60b76d394f8d37d8ebd1f8918c79029f0db86/node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash    : 73e60b76d394f8d37d8ebd1f8918c79029f0db86
Studio                  : 0.458.0
```

hayes:
fixes #12252

### DIFF
--- a/packages/client/src/runtime/core/model/applyFluent.ts
+++ b/packages/client/src/runtime/core/model/applyFluent.ts
@@ -119,7 +119,7 @@ export function applyFluent(
         // we allow for chaining more with this recursive call
         return applyFluent(client, ...modelArgs, ...dataArgs)
       },
-      ...defaultProxyHandlers(ownKeys),
+      ...defaultProxyHandlers([...ownKeys, ...Object.getOwnPropertyNames(prismaPromise)]),
     })
   }
 }
--- a/packages/client/src/__tests__/integration/happy/prismaPromise/test.ts
+++ b/packages/client/src/__tests__/integration/happy/prismaPromise/test.ts
@@ -72,6 +72,20 @@ describe('prismaPromise', () => {
     expect(createResult1).toStrictEqual(createResult2)
   })
 
+  test('fluent promises should have promise properties', async () => {
+    const findUniquePromise = prisma.user.findUnique({
+      where: {
+        email: 'email@email.em',
+      },
+    })
+
+    expect('then' in findUniquePromise).toBe(true)
+    expect('finally' in findUniquePromise).toBe(true)
+    expect('catch' in findUniquePromise).toBe(true)
+
+    await findUniquePromise.finally()
+  })
+
   beforeAll(async () => {
     const PrismaClient = await getTestClient()
     prisma = new PrismaClient()


Jolg42:
Thanks for the issue and PR @hayes!
I unblocked the GitHub Actions run and pinged the Client Team about this.
Note that it could take some time until they have capacity to review 🤞🏼

packages/client/src/runtime/core/model/applyFluent.ts
```diff
@@ -119,7 +119,7 @@ export function applyFluent(
         // we allow for chaining more with this recursive call
         return applyFluent(client, ...modelArgs, ...dataArgs)
       },
-      ...defaultProxyHandlers(ownKeys),
+      ...defaultProxyHandlers([...ownKeys, ...Object.getOwnPropertyNames(prismaPromise)]),
```
hayes, line 122: This currently just grabs own properties from prismaPromise which works fine since prismaPromises are created as an object with methods defined inline.  If prismaPromise were converted to a class in the future, properties like `then` would move to the prototype, and this would break.

The way "ownKeys" are used in the proxy is a little weird, and acts basically like flattening the prototype chain so all props are treated as own props. 

I think the technically more correct solution (not worth it in my opinion) would be to walk up the prototype chain and get the set of all ownPropertyNames and ownPropertySymbols.  I'm not sure that actually adds any real world value though, so going to leave it as is, but happy to make this more robust if you want.

Grayson-code (approved):

millsp (approved):
Thanks for your contribution :rocket:

PR #12297, opened 2022-03-12
fix: Clean up usage of terms "binary" and "engine"

janpio:
This PR depends on https://github.com/prisma/engines-wrapper/pull/343 and currently uses a special version of those packages that include those changes.

- All uses of `Binary` have been checked and replaced with `Engine` if they refer to one of the engines. 
- `Binary` was of course kept when it was actually about the `BinaryEngine` or for `BinaryTargets` related things.
- In Generator context, the term `Executable` is used instead (which was always the more correct term here)

### TODO
- [ ] Identify "public" APIs (Generator [e.g. via [internal link](https://www.notion.so/prismaio/Prisma-Generators-a2cdf262207a4e9dbcd0e362dfac8dc0), SDK [current API], etc) and make sure old arguments/parameters still work to not make this too breaking
    - Prisma Client
		- `new PrismaClient({ __internal: { engine: { binaryPath: customBinaryPath }} })` => `enginePath`
		- type `GenerateClientOptions`: `binaryPaths` => `enginePaths` (e.g. used in `generateClient()`)
		- `download()` now takes `engines` instead of `binaries` in type `DownloadOptions`
	- Generator
		- `GeneratorOptions`: `binaryPaths` => `enginePaths`
			- related method `Generator.setBinaryPaths()` is now `Generator.setEnginePaths()`
		- type `GeneratorManifest` now expects an array of `EngineName` instead of `EngineType` as `requiresEngines` (but string values of that is still the same, for now)
		- `GetGeneratorOptions` in `getGenerators.ts` now expects a property `enginePathsOverride` instead of `binaryPathsOverride` (of type `EnginePathsOverride` instead of `BinaryPathsOverride`), now also has a type `EnginePathsOverride` instead of `BinaryPathsOverride` that also expects an array of `EngineName` instead of `EngineType`.
	  	- Type `GetEnginePathsByVersionInput` now also has `enginePathsOverride` instead of `binaryPathsOverride`, which is used in ` getEnginePathsByVersion()` (which used to be ` getBinaryPathsByVersion()`)
	- ...
- [ ] Understand and fix Studio test failures to enable them again
	- Probably https://cs.github.com/prisma/studio-code/blob/38af3c04fe24ee593deeda9663f412fff7348bc6/packages/pcw/src/pcw.ts?q=binary#L196-L205 and https://github.com/prisma/studio-code/blob/5c3567ce5fa27d42b64bee9e2e8deb24628aa1ac/packages/pcw/src/client.ts#L95

### Bonus (to be turned into separate issues)
- [ ] `EngineType` vs `EngineName` weirdness (`queryEngine` vs. `query-engine` and similar, see separate branch)
- [ ] Move all the manual engine logic that is repeated in tests (e.g. get the path of a query engine, where it manually constructs the path and puts `.exe` at the end on Windows) into the actual logic instead
	- [ ] ` ensureTestClientQueryEngine` also smells like a potentially bad idea skipping actual logic
- [ ] Remove all `binaryTargets = ["native"]` from tested schemas (unless this is explicitly what is to be tested)
- [ ] Clean up uses of `prismaPath` to mean the path of a query engine
- [ ] New `EngineType` clashes with `engineType` from `generator` block in schema - Problem? Better name? (maybe `EngineKind`?)
- [ ] Binary Targets related things (both where they mean "just" the "platform" and where they are mixed into a matrix concept of engine types and platform names) are the remaining big weird area


packages/cli/src/Version.ts
```diff
@@ -126,22 +126,22 @@ export class Version implements Command {
...
     if (pathFromEnv && fs.existsSync(pathFromEnv)) {
-      const version = await getVersion(pathFromEnv, binaryName)
+      const version = await getVersion(pathFromEnv, engineType)
       return { version, path: pathFromEnv, fromEnvVar: envVar }
     }
 
-    const binaryPath = await resolveBinary(binaryName)
-    const version = await getVersion(binaryPath, binaryName)
-    return { path: binaryPath, version }
+    const enginePath = await resolveEngine(engineType)
```
janpio, line 142: This is currently a bug as the local function has the same name as the imported function - so this functionality is probably not properly tested.

janpio: That could explain this:
```
function cleanSnapshot(str: string): string {
  return str.replace(/:(.*)/g, ': placeholder')
}
```
This way the path in the version output (and env var) is never properly checked.

packages/sdk/src/get-generators/getGenerators.ts
```diff
@@ -356,25 +357,25 @@ generator gen {
...
+  enginePathsOverride?: EnginePathsOverride
 }
 
 /**
  * Shortcut for getGenerators, if there is only one generator defined. Useful for testing.
  * @param schemaPath path to schema.prisma
  * @param aliases Aliases like `photonjs` -> `node_modules/photonjs/gen.js`
- * @param version Version of the binary, commit hash of https://github.com/prisma/prisma-engine/commits/master
+ * @param version Version of the engine, commit hash of https://github.com/prisma/prisma-engine/commits/master
  * @param printDownloadProgress `boolean` to print download progress or not
```
janpio, line 379: Is this the correct way to document `options`?

PR #12307, opened 2022-03-13
test(version): Replace blanket sanitization of all version output value with more nuanced sanitization

janpio:
The previous sanitization was pretty broad, potentially hiding bugs in the version command output.
Now only specific variable values like paths or version number are sanitized, making the snapshots much more useful.


Grayson-code (approved):
>

Jolg42 (approved):

PR #12322, opened 2022-03-14
test(client): json-filtering-postgres increase test timeout for Win/Mac CI

Jolg42:
Example https://github.com/prisma/prisma/runs/5536215763?check_suite_focus=true#step:9:1852


millsp:
Does increasing the timeout work or is it indefinitely blocked?

Jolg42:
Apparently it works, see results of this PR, passes here at least

millsp (approved):

PR #12337, opened 2022-03-14
chore: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -64,6 +64,13 @@ module.exports = {
         '@typescript-eslint/no-misused-promises': 'off',
         'jest/expect-expect': 'off',
         'no-empty': 'off',
+        'no-restricted-properties': [
+          'error',
+          {
+            property: 'substr',
+            message: 'Deprecated: Use .slice() instead of .substr().',
+          },
+        ],
         'jest/valid-title': 'off',
         '@typescript-eslint/no-unnecessary-type-assertion': 'off',
         // low hanging fruits:
--- a/scripts/only-allow-pnpm.js
+++ b/scripts/only-allow-pnpm.js
@@ -13,8 +13,8 @@ function pmFromUserAgent(userAgent) {
   const pmSpec = userAgent.split(' ')[0]
   const separatorPos = pmSpec.lastIndexOf('/')
   return {
-    name: pmSpec.substr(0, separatorPos),
-    version: pmSpec.substr(separatorPos + 1),
+    name: pmSpec.substring(0, separatorPos),
+    version: pmSpec.slice(separatorPos + 1),
   }
 }
 
--- a/packages/debug/src/common.ts
+++ b/packages/debug/src/common.ts
@@ -182,7 +182,7 @@ export function setup(env) {
       namespaces = split[i].replace(/\*/g, '.*?')
 
       if (namespaces[0] === '-') {
-        createDebug.skips.push(new RegExp('^' + namespaces.substr(1) + '$'))
+        createDebug.skips.push(new RegExp('^' + namespaces.slice(1) + '$'))
       } else {
         createDebug.names.push(new RegExp('^' + namespaces + '$'))
       }


Jolg42:
Hi @CommanderRoot 
Thanks for your first PR here! 💚

Could you check The CLAassistant [comment up](https://github.com/prisma/prisma/pull/12337#issuecomment-1067375348) and sign the CLA?

CommanderRoot:
I did sign the CLA but for some reason the bot doesn't recognizes it. I also tried to click the "recheck" button but that doesn't do anything either.
grafik

scripts/only-allow-pnpm.js
```diff
@@ -13,8 +13,8 @@ function pmFromUserAgent(userAgent) {
   const pmSpec = userAgent.split(' ')[0]
   const separatorPos = pmSpec.lastIndexOf('/')
   return {
-    name: pmSpec.substr(0, separatorPos),
-    version: pmSpec.substr(separatorPos + 1),
+    name: pmSpec.substring(0, separatorPos),
```
aqrln, line 16: I guess I'd probably prefer using `slice` here as well for consistency (or, alternatively, `substring` everywhere – they have the same arguments anyway in all of these cases).

aqrln (approved):
Hi and thank you for your contribution!

I think it's a nice thing to do, yes. Generally this looks good to me, however, is there a specific reason why you used `substring` in one case and `slice` in the others or it just so happened?

scripts/only-allow-pnpm.js
CommanderRoot: `slice` and `substring` are very similar but [differ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) in how they handle the second parameter, especially if it's negative. I used it here to have the same result as `substr`

CommanderRoot:
Based on the comment above I've gone ahead to switch all uses of `substring` to `slice`. This way we only use one of these very similar functions. `slice` is shorter and generally a bit faster than `substring`, that's why I used it.
I've also added an ESLint rule to prevent further usage of `substr` or `substring`

aqrln:
For some reason I can't leave the comment in that thread, so copying it here:

@CommanderRoot 
> slice and substring are very similar but [differ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) in how they handle the second parameter, especially if it's negative. I used it here to have the same result as substr

I see, I didn't notice that `separatorPos` might technically be negative since it comes from `lastIndexOf()`.

aqrln:
Regarding the new changes and the ESLint rule:

@Jolg42 @millsp is this something we want or is it a bit excessive, wdyt?

.eslintrc.js
```diff
@@ -64,6 +64,17 @@ module.exports = {
         '@typescript-eslint/no-misused-promises': 'off',
         'jest/expect-expect': 'off',
         'no-empty': 'off',
+        'no-restricted-properties': [
+          'error',
+          {
+            property: 'substring',
+            message: 'Use .slice instead of .substring.',
```
Jolg42, line 71:
```suggestion
            message: 'Deprecated: use .slice() instead of .substring().',
```

aqrln (approved):

Jolg42 (approved):
Thank you for your first PR here 💚

PR #12348, opened 2022-03-15
test(client/full-test-search-mysql): increase timeout for mac/win

Jolg42:
Example failure https://github.com/prisma/prisma/runs/5550140286?check_suite_focus=true#step:9:1312


Jolg42:
and https://github.com/prisma/prisma/runs/5567300690?check_suite_focus=true

millsp (approved):

PR #12391, opened 2022-03-16
fix(db pull): add code 104 warning

Jolg42:
Related https://github.com/prisma/prisma-engines/pull/2697


millsp (approved):

millsp:
Is it possible to add a test for this?

Jolg42:
We have little tests about this as of today, as we rely on introspection-ci project for these.

Jolg42:
We could focus on adding tests here one day

janpio:
Starting with error codes or warning messages like this here would be an easy way to do that. Can you create an issue @Jolg42?

Jolg42:
@janpio see https://github.com/prisma/prisma/issues/12472

PR #12392, opened 2022-03-16
ci(publish): remove unpkg.com and replace with npm registry

Jolg42:
We don't need unpkg since npm registry can give us that info. So we can remove a 3rd party dependency here.

Was in my git stash for a while 😄

### DIFF
--- a/scripts/setup.ts
+++ b/scripts/setup.ts
@@ -3,8 +3,6 @@ import execa from 'execa'
 import fetch from 'node-fetch'
 import path from 'path'
 
-import { getPackageDependencies, getPackages, getPublishOrder } from './ci/publish'
-
 async function main() {
   // this is for when you want to use it locally
   const buildOnly = process.argv[2] === '--build' // can we remove this?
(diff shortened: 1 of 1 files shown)


millsp (approved):

scripts/setup.ts
```diff
@@ -149,14 +147,14 @@ async function branchExists(dir: string, branch: string): Promise<boolean> {
   return exists
 }
 
-async function getPrismaCommitFromPackageJsonViaUnpkg(npmVersion: string): Promise<{ prisma: string }> {
-  return fetch(`https://unpkg.com/prisma@${npmVersion}/package.json`, {
+async function getPrismaCommitFromPackageJsonViaNpmRegistry(npmVersion: string): Promise<{ prisma: string }> {
+  return fetch(`https://registry.npmjs.org/prisma`, {
```
aqrln, line 151:
```suggestion
  return fetch(`https://registry.npmjs.org/prisma/${npmVersion}`, {
```

scripts/setup.ts
```diff
@@ -149,14 +147,14 @@ async function branchExists(dir: string, branch: string): Promise<boolean> {
...
     headers: {
       accept: 'application/json',
     },
   })
     .then((res) => res.json())
     .then((pkg) => {
-      return pkg.prisma.prismaCommit
+      return pkg.versions[npmVersion].prisma.prismaCommit
```
aqrln, line 158:
```suggestion
      return pkg.prisma.prismaCommit
```

aqrln (commented):
Looks good, but I'd prefer to download just a single package manifest from npm, like it used to do from unpkg. `https://registry.npmjs.org/prisma` without version is already 19.7 MB and growing, which might potentially cause issues or flakiness due to network errors in the future.

Ref: https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#getpackageversion

PR #12416, opened 2022-03-17
fix(client): resolve client output using static paths

njmaeff:
Previously the client resolved the generated directory using the `findSync` method. This method is tricky to extend, so instead, we use a new, more specific function to determine this directory using absolute paths to match instead of patterns.

Closes #10433

PR #12425, opened 2022-03-18
fix(client): join `runtimeDir` and `runtimeName` using `path.join()`

njmaeff:
Previously these paths used a hard-coded forward slash. This change should fix an issue on Windows when `runtimeDir` is a relative path.

### DIFF
--- a/packages/client/src/generation/utils.ts
+++ b/packages/client/src/generation/utils.ts
@@ -329,3 +329,12 @@ export function unique<T>(arr: T[]): T[] {
 
   return result
 }
+
+/**
+ *
+ * @param paths - list of paths in posix or win32 format to join and then
+ *     replace all backslash separators with a forward slash
+ */
+export function makePosixImportPath(...paths: string[]): string {
+  return path.join(...paths.map((segment) => segment.replace(/\\$/, ''))).replace(/\\/g, path.posix.sep)
+}
--- a/packages/client/src/generation/utils.ts
+++ b/packages/client/src/generation/utils.ts
@@ -1,4 +1,5 @@
 import indent from 'indent-string'
+import os from 'os'
 import path from 'path'
 
 import type { DMMFHelper } from '../runtime/dmmf'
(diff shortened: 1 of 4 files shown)


njmaeff:
This should fix the failing test on this [ci run](https://github.com/prisma/prisma/runs/5598121285?check_suite_focus=true#step:9:1575) and PR #12416.

aqrln (changes requested):
Hi @njmaeff and thank you for opening the PR and raising the discussion here!

Using a consistent path separator is a nice thing to do (although for `require` a forward slash is typically used as it works as a path separator on all operating systems), but this change as it is won't solve the issue you are referring to without fixing the actual bug that causes it. It doesn't really matter if forward and backward slashes are both used in the same path on Windows, and it is exactly the backslashes that are causing troubles here, because they are not escaped, so `..\..\..\..\..\..\..\runtime/index` is interpreted as
```
..............
untime/index
```
and with these changes it will even become
```
..............
untimeindex
```

The fix would be to either escape the backslashes (e.g., by using `JSON.stringify` or manually replacing backslashes with double backslashes), or just replace them with forward slashes which are cross-platform.

njmaeff:
@aqrln You're right; thank you for explaining that to me. I added a function to join paths and convert the separator to POSIX. Please let me know your thoughts.

packages/client/src/generation/utils.ts
```diff
@@ -329,3 +329,12 @@ export function unique<T>(arr: T[]): T[] {
...
   return result
 }
+
+/**
+ *
+ * @param paths - list of paths in posix or win32 format to join and then
+ *     replace all backslash separators with a forward slash
+ */
+export function makePosixImportPath(...paths: string[]): string {
+  return path.join(...paths.map((segment) => segment.replace(/\\$/, ''))).replace(/\\/g, path.posix.sep)
```
aqrln, line 339: Not sure how much it's practically important here, but generally on Unix a backslash may be a valid character in a file name (although it would break things when embedded into a string without proper escaping nevertheless).

aqrln: I would probably prefer to replace `path.sep` rather than hardcoding `\\`, but I understand that it may make testing this function more complicated.

njmaeff: I made a few more changes based on your feedback. I added the prefix `'./'` for when the runtime would be in a child directory because `path.join` automatically removes the prefix and `require` will go looking in the `node_modules`. I also only replace the back slash on Windows in case it is a part of the file name on Linux.

aqrln (approved):
Sorry for taking so long to get back to this, it looks good to me!

aqrln:
Had to revert because of test failures on `main`: https://github.com/prisma/prisma/pull/13201

njmaeff:
> Had to revert because of test failures on `main`: #13201

I'm sorry this failed... I would love a chance to fix the issue, should I create a new PR?

aqrln:
No worries! Yes, a new PR sounds good.

PR #12427, opened 2022-03-18
fix(client): use findSync only if no __dirname

### LINKED ISSUE #10512: Prisma client cannot find datamodel path ("schema.prisma")

### LINKED ISSUE #12083: Prisma 3.9.x+ multiple clients being generated, and in the wrong Yarn 1 workspace
### Bug description

In yarn 1 workspaces, sometimes prisma will create the .prisma folder at one or more of the nested workspaces' node_modules.

Migrations and generation produces the client in the root `node_modules/.prisma` but when running inside a workspace, it picks a _different_ prisma client, causing this error

In our case:
```
├── node_modules
│   ├── .prisma # 👈 this is good!
├── web
│   ├── node_modules
├── api
│   ├── node_modules
│   │   ├── .prisma  # 😨 whai?
```
Node version: v16.13.1
Yarn version: 1.23.0-20220130.1630 (set using yarn version classic)

### Reproduction is in this repo here: 
- https://github.com/dac09/prisma-multiple-clients-repro
- To launch on gitpod: https://gitpod.io/#https://github.com/dac09/prisma-multiple-clients-repro

### Steps to reproduce:
1. Check that theres no prisma folder in api/node_modules
```
ls -la api/node_modules
```
2. Go into api and generate prisma client
```
cd api && yarn rw prisma generate
```
3. Notice that prisma has now been generated in api node_modules
```
❯ ls -a api/node_modules
.       ..      .prisma
```

4. And to see the actual problem this causes run:

```
yarn rw prisma migrate dev
# equivalent to
# yarn prisma migrate dev --schema "./api/db/schema.prisma"
```

and then
```
yarn rw prisma db seed
```

And you should see:
```
Invalid `db.post.create()` invocation in
/Users/dac09/Experiments/prisma-repro/scripts/seed.ts:24:41

  21
  22 Promise.all(
  23   posts.map(async (post) => {
→ 24     const newPost = await db.post.create(
  The table `main.Post` does not exist in the current database.
    at Object.request (/Users/dac09/Experiments/prisma-repro/node_modules/@prisma/client/runtime/index.js:39809:15)
    at PrismaClient._request (/Users/dac09/Experiments/prisma-repro/node_modules/@prisma/client/runtime/index.js:40637:18)
    at /Users/dac09/Experiments/prisma-repro/scripts/seed.ts:24:27
    at async Promise.all (index 0) {
  code: 'P2021',
  clientVersion: '3.10.0',
  meta: { table: 'main.Post' }
```

### Expected behavior

Prisma should only generate client at root of the workspace, not in nested workspaces.

### Prisma information

Schema: https://github.com/dac09/prisma-multiple-clients-repro/blob/main/api/db/schema.prisma
Query: https://github.com/dac09/prisma-multiple-clients-repro/blob/main/api/src/services/posts/posts.ts#L6

### Environment & setup

- OS: MacOS, Linux, Windows
- Database: Tested with SQLite, but likely problem with all DBs
- Node.js version: 16.13.1

### Prisma Version

```
prisma                  : 3.10.0
@prisma/client          : 3.10.0
Query Engine (Node-API) : libquery-engine 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt 73e60b76d394f8d37d8ebd1f8918c79029f0db86 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : 73e60b76d394f8d37d8ebd1f8918c79029f0db86
Studio                  : 0.458.0
```

--- 

#### Mirrored Redwood Issue
https://github.com/redwoodjs/redwood/issues/4605

millsp:
Fixes https://github.com/prisma/prisma/issues/12083
Fixes https://github.com/prisma/prisma/issues/10512

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -302,21 +302,6 @@ export async function generateClient({
     await writeFile(packageJsonTargetPath, pkgJson)
   }
 
-  // if (!testMode && process.env.INIT_CWD) {
-  //   const backupPath = path.join(process.env.INIT_CWD, 'node_modules/.prisma/client')
-  //   console.log('backupPath', backupPath)
-  //   console.log('finalOutputDir', finalOutputDir)
-
-  //   // if (finalOutputDir !== backupPath && !generator?.isCustomOutput) {
-  //   //   await copy({
-  //   //     from: finalOutputDir,
-  //   //     to: backupPath,
-  //   //     recursive: true,
-  //   //     parallelJobs: process.platform === 'win32' ? 1 : 20,
-  //   //     overwrite: true,
-  //   //   })
-  //   // }
-  // }
 
   const proxyIndexJsPath = path.join(outputDir, 'index.js')
   const proxyIndexBrowserJsPath = path.join(outputDir, 'index-browser.js')
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -104,6 +104,7 @@ export async function buildClient({
   }
 }
 
+// TODO: explore why we have a special case for excluding pnpm
 async function getDotPrismaDir(outputDir: string): Promise<string> {
   if (outputDir.endsWith('node_modules/@prisma/client')) {
     return path.join(outputDir, '../../.prisma/client')
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -301,18 +301,21 @@ export async function generateClient({
     await writeFile(packageJsonTargetPath, pkgJson)
   }
 
-  if (!testMode && process.env.INIT_CWD) {
-    const backupPath = path.join(process.env.INIT_CWD, 'node_modules/.prisma/client')
-    if (finalOutputDir !== backupPath && !generator?.isCustomOutput) {
-      await copy({
-        from: finalOutputDir,
-        to: backupPath,
-        recursive: true,
-        parallelJobs: process.platform === 'win32' ? 1 : 20,
-        overwrite: true,
-      })
-    }
-  }
+  // if (!testMode && process.env.INIT_CWD) {
+  //   const backupPath = path.join(process.env.INIT_CWD, 'node_modules/.prisma/client')
+  //   console.log('backupPath', backupPath)
+  //   console.log('finalOutputDir', finalOutputDir)
+
+  //   // if (finalOutputDir !== backupPath && !generator?.isCustomOutput) {
+  //   //   await copy({
+  //   //     from: finalOutputDir,
+  //   //     to: backupPath,
+  //   //     recursive: true,
+  //   //     parallelJobs: process.platform === 'win32' ? 1 : 20,
+  //   //     overwrite: true,
+  //   //   })
+  //   // }
+  // }
 
   const proxyIndexJsPath = path.join(outputDir, 'index.js')
   const proxyIndexBrowserJsPath = path.join(outputDir, 'index-browser.js')
--- a/packages/client/src/generation/utils/buildDirname.ts
+++ b/packages/client/src/generation/utils/buildDirname.ts
@@ -23,7 +23,8 @@ export function buildDirname(clientEngineType: ClientEngineType, relativeOutdir:
  * a build step for platforms like Vercel or Netlify. On top of that, the
  * feature is especially useful for Next.js/Webpack users since the client gets
  * moved and copied out of its original spot. It all fails, it falls-back to
- * `__dirname`, which is never available on bundles.
+ * `findSync`, when `__dirname` is not available (eg. bundle, electron) or
+ * nothing has been found around `__dirname`.
  * @param defaultRelativeOutdir
  * @param runtimePath
  * @returns
@@ -35,14 +36,22 @@ function buildDirnameFind(defaultRelativeOutdir: string, runtimePath: string) {
   return `
 const { findSync } = require('${runtimePath}')
 
-// use findSync when __dirname is either not available or not useful
-const useFindSync = typeof __dirname === 'undefined' || __dirname === '/'
+const hasDirname = typeof __dirname !== 'undefined' && __dirname !== '/'
 
-const dirname = useFindSync ? findSync(process.cwd(), [
+// will work in most cases, ie. if the client has not been bundled
+const regularDirname = hasDirname ? path.dirname(findSync(__dirname, [
+    ${JSON.stringify('schema.prisma')},
+], ['f'], ['d'], 1)[0] || '') : undefined
+
+// if the client has been bundled, we need to look for the folder
+const foundDirname = findSync(process.cwd(), [
     ${defaultRelativeOutdir ? `${JSON.stringify(defaultRelativeOutdir)},` : ''}
     ${serverlessRelativeOutdir ? `${JSON.stringify(serverlessRelativeOutdir)},` : ''}
-], ['d'], ['d'], 1)[0] : __dirname`
+], ['d'], ['d'], 1)[0]
+
+const dirname = regularDirname || foundDirname || __dirname`
 }
+// TODO: 👆 all this complexity could fade away if we embed the schema
 
 /**
  * Builds a simple `dirname` for when it is not important to have one.
--- a/packages/client/src/generation/utils/buildDirname.ts
+++ b/packages/client/src/generation/utils/buildDirname.ts
@@ -35,10 +35,13 @@ function buildDirnameFind(defaultRelativeOutdir: string, runtimePath: string) {
   return `
 const { findSync } = require('${runtimePath}')
 
-const dirname = findSync(process.cwd(), [
+// use findSync when __dirname is either not available or not useful
+const useFindSync = typeof __dirname === 'undefined' || __dirname === '/'
+
+const dirname = useFindSync ? findSync(process.cwd(), [
     ${defaultRelativeOutdir ? `${JSON.stringify(defaultRelativeOutdir)},` : ''}
     ${serverlessRelativeOutdir ? `${JSON.stringify(serverlessRelativeOutdir)},` : ''}
-], ['d'], ['d'], 1)[0] || __dirname`
+], ['d'], ['d'], 1)[0] : __dirname`
 }
 
 /**
--- a/packages/client/src/generation/utils/buildDirname.ts
+++ b/packages/client/src/generation/utils/buildDirname.ts
@@ -36,12 +36,16 @@ function buildDirnameFind(defaultRelativeOutdir: string, runtimePath: string) {
   return `
 const { findSync } = require('${runtimePath}')
 
+// some frameworks or bundlers disallow or totally remove __dirname
 const hasDirname = typeof __dirname !== 'undefined' && __dirname !== '/'
 
 // will work in most cases, ie. if the client has not been bundled
-const regularDirname = hasDirname ? path.dirname(findSync(__dirname, [
+const regularFilename = hasDirname ? findSync(__dirname, [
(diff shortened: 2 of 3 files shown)


packages/client/src/generation/generateClient.ts
```diff
@@ -301,18 +301,21 @@ export async function generateClient({
...
-      await copy({
-        from: finalOutputDir,
-        to: backupPath,
-        recursive: true,
-        parallelJobs: process.platform === 'win32' ? 1 : 20,
-        overwrite: true,
-      })
-    }
-  }
+  // if (!testMode && process.env.INIT_CWD) {
```
millsp, line 304: This has been reported by redwood as being incorrect. It causes the generator to copy the generated files to `backupPath`. As far as I understand, `backupPath` will cause to the copy to happen wherever you run `prisma generate`. I don't see how this could be useful since the we already generated the client where it should be in `finalOutputDir`.

(help needed, cc @Jolg42 @aqrln 🙏)

Jolg42: It looks like this code was added to support pnpm, see commit https://github.com/prisma/prisma/commit/2be6b71dfa27202058eeff5274dbaf0aabc31c73

millsp: Thanks, I don't think this is actually needed for `pnpm`. What this does is to copy the contents from the actual final directory to the root of `node_modules`. That cannot be used by `pnpm` and is causing pollution on yarn (and probably others).

So before this PR, you would get an unnecessary generated artifact copied in the root of `node_modules`.
Screenshot from 2022-03-21 11-16-17

Similarly, this is causing worse directory pollution by creating a `node_modules` folder and copying the `.prisma/client` in it, as described in the original issue https://github.com/prisma/prisma/issues/12083

And after this PR, you don't get a polluted `pnpm` `node_modules` or `yarn` workspace.
Screenshot from 2022-03-21 11-17-17

The client will get generated where it belongs, and only wherever it has been hoisted.
Screenshot from 2022-03-21 11-16-17

Happy to set a call up today if needed, I need to merge this soon for a patch :)

millsp: All E2E tests are passing except for one with `pnpm` because I was explicitly checking for the existence of `node_modules/.prisma/client` when I should have check that to be in the hoisted folder(like in the screenshot just above)  :facepalm:

Jolg42: Nice! Then it sounds like we can get rid of this 🧹

Jolg42: Note that there is some related code in `getDotPrismaDir` about `INIT_CWD` and pnpm https://github.com/prisma/prisma/blob/integration/fix/yarn1-workspaces/packages/client/src/generation/generateClient.ts#L111-L126

packages/client/src/generation/utils/buildDirname.ts
```diff
@@ -34,12 +35,23 @@ function buildDirnameFind(defaultRelativeOutdir: string, runtimePath: string) {
...
+const regularDirname = hasDirname && fs.existsSync(path.join(__dirname, 'schema.prisma')) && __dirname
+
+// if the client has been bundled, we need to look for the folders
+const foundDirname = !regularDirname && findSync(process.cwd(), [
     ${defaultRelativeOutdir ? `${JSON.stringify(defaultRelativeOutdir)},` : ''}
     ${serverlessRelativeOutdir ? `${JSON.stringify(serverlessRelativeOutdir)},` : ''}
-], ['d'], ['d'], 1)[0] || __dirname`
+], ['d'], ['d'], 1)[0]
+
+const dirname = regularDirname || foundDirname || __dirname`
```
Jolg42, line 52: How can someone easily debug this, for example looking at logs?
(Should we somehow have debug logs for each const?)

millsp: I can easily add a debug log for `dirname` so that one can see the path that was found.

packages/client/src/generation/generateClient.ts
millsp: Added it as a small item to our deep dive sessions :diving_mask:

packages/client/src/generation/utils/buildDirname.ts
millsp: Ideally, we would have any notion of paths here and bundle the schema on generate. But that's another topic to explore.

millsp: I added this into to the client debug logs.

Jolg42 (approved):
🤞🏼 

Can't wait to see if we can remove all of this complexity with schema embedding

packages/client/src/generation/utils/buildDirname.ts
```diff
@@ -34,12 +35,23 @@ function buildDirnameFind(defaultRelativeOutdir: string, runtimePath: string) {
 
   return `
 const { findSync } = require('${runtimePath}')
+const fs = require('fs')
 
-const dirname = findSync(process.cwd(), [
+// some frameworks or bundlers replace or totally remove __dirname
+const hasDirname = typeof __dirname !== 'undefined' && __dirname !== '/'
```
Jolg42, line 41: Would __dirname on Windows always be different than `'/'` actuallu?

packages/client/src/generation/generateClient.ts
```diff
@@ -301,18 +302,21 @@ export async function generateClient({
...
+  //   // if (finalOutputDir !== backupPath && !generator?.isCustomOutput) {
+  //   //   await copy({
+  //   //     from: finalOutputDir,
+  //   //     to: backupPath,
+  //   //     recursive: true,
+  //   //     parallelJobs: process.platform === 'win32' ? 1 : 20,
+  //   //     overwrite: true,
+  //   //   })
+  //   // }
+  // }
```
aqrln, line 319: Should we just remove dead code or leave it here commented out until we investigate?

packages/client/src/generation/utils/buildDirname.ts
```diff
@@ -34,12 +35,23 @@ function buildDirnameFind(defaultRelativeOutdir: string, runtimePath: string) {
...
+
+// will work in most cases, ie. if the client has not been bundled
+const regularDirname = hasDirname && fs.existsSync(path.join(__dirname, 'schema.prisma')) && __dirname
```
aqrln, line 44: What is the type of this? `boolean | string`?

packages/client/src/generation/utils/buildDirname.ts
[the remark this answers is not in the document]
aqrln: Btw, does it make sense to have `__dirname` as a final fallback here? If `regularDirname` is false, then `__dirname` is either undefined or doesn't have a schema, so it might be more useful to just throw an error at this point.

packages/client/src/generation/utils/buildDirname.ts
aqrln: Looking at the comment, I assume that frameworks and bundlers that replace `__dirname` with `/` might do so regardless of the OS. That said, I'm not sure if it's necessary to check for this. If this path is fake, it will be caught by the `fs.existsSync` check below, otherwise it might be a legit real path (it would be an edge case, but something that might happen in someone's Docker container).

packages/client/src/generation/utils/buildDirname.ts
millsp: `false | string` yes

packages/client/src/generation/utils/buildDirname.ts
millsp: Yes, this is really to catch bundlers like webpack that may replace it with `/` or might not even exist like in `electron`. If it turns out that `/` was indeed an existing path, then `findSync` just below will find it for sure.

packages/client/src/generation/generateClient.ts
millsp: Thanks, I forgot to remove it.

packages/client/src/generation/utils/buildDirname.ts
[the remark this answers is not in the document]
aqrln: Oh, I see. Makes sense.

packages/client/src/generation/generateClient.ts
millsp:
```suggestion
```

aqrln (approved):

Faithfinder:
Is there a build I can install to try it out, or do I wait for... 3.12, I guess?

millsp:
You can try `prisma@dev` & `@prisma/client@dev`

Faithfinder:
Thank you! Just checked and it works with our setup as well (`rush` monorepo with `pnpm` as package manager). 🥳

PR #12488, opened 2022-03-23
feat(client): isSet, updateMany, deleteMany, and filter re-write

millsp:
Integrates https://github.com/prisma/prisma-engines/pull/2797

### DIFF
--- a/packages/client/src/__tests__/integration/happy/composites-mongo/update/list.ts
+++ b/packages/client/src/__tests__/integration/happy/composites-mongo/update/list.ts
@@ -1,4 +1,5 @@
 import { getTestClient } from '../../../../../utils/getTestClient'
+import { commentRequiredListDataB } from '../__helpers__/build-data/commentRequiredListDataB'
 
 const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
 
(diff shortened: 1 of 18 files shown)


packages/client/src/__tests__/integration/happy/composites-mongo/update/list.ts
```diff
@@ -242,12 +242,108 @@ describeIf(!process.env.TEST_SKIP_MONGODB)('update > list', () => {
...
+      }
+    `)
+  })
 
   /**
    * Simple deleteMany
    */
-  test.skip('deleteMany', async () => {})
+  test('deleteMany', async () => {
+    const comment = await prisma.commentRequiredList.update({
```
Jolg42, line 305: test says `deleteMany` but client does `update`

packages/client/src/__tests__/integration/happy/composites-mongo/update/list.ts
```diff
@@ -242,12 +242,108 @@ describeIf(!process.env.TEST_SKIP_MONGODB)('update > list', () => {
...
+                userId: 10,
+                vote: true,
+              },
+            ],
+          },
+          Object {
+            text: Hello World,
+            upvotes: Array [
+              Object {
+                userId: Another Comment,
```
Jolg42, line 289: Nit: it's a bit weird to mix a number and a string here for a userId.

I see the type is `userId String` in the schema

Jolg42 (approved):
Looks good! 

⚠️ don't merge as is because it's using an integration engine version

packages/client/src/__tests__/integration/happy/composites-mongo/update/list.ts
millsp: Nice catch, thanks! These will be caught when we have automated typechecks (coming soon).

PR #12505, opened 2022-03-24
fix(client): internal endpoint setup for BinaryEngine

garrensmith:
This fix allows us to do:

```
   __internal: {
    engine: {
      endpoint: "http://127.0.0.1:4466",
    },
  },
});
```

And connect to a local running binary engine. This will make testing certain things a lot easier.

### DIFF
--- a/packages/engine-core/src/binary/BinaryEngine.ts
+++ b/packages/engine-core/src/binary/BinaryEngine.ts
@@ -519,7 +519,8 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
       }
       if (this.engineEndpoint) {
         try {
-          await pRetry(() => this.connection.get('/'), {
+          this.connection.open(this.engineEndpoint)
+          await pRetry(() => this.connection.get('/status'), {
             retries: 10,
           })
         } catch (e) {


packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -519,7 +519,8 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
       }
       if (this.engineEndpoint) {
         try {
-          await pRetry(() => this.connection.get('/'), {
+          this.connection.open(this.engineEndpoint)
+          await pRetry(() => this.connection.get('/status'), {
```
janpio, line 523: Unrelated: Do we do this waiting for status to work in the normal BinaryEngine flow as well?

janpio (commented):
Optimally we would have a test for this API as well, but probbly not the right time to ask for that 🤷

packages/engine-core/src/binary/BinaryEngine.ts
garrensmith: No, normally on startup we check logs to know the binary engine has started.

janpio: Ah right, because we trigger the start and then can watch the logs, that is easier and equally reliable.

garrensmith:
> Optimally we would have a test for this API as well, but probbly not the right time to ask for that 
Given this is an internal feature only, I'm ok with no test.

Jolg42 (approved):
Looks good! Given that it's internal and only about Binary Engine I also think we can skip the test right now.

Jolg42:
Issue to add a test here https://github.com/prisma/prisma/issues/12507

halfmatthalfcat:
Just ran into this last night testing local engine changes ha. Thank you for fixing.

Issue #12419, opened 2022-03-18
Specify a request timeout for client


endor:
## Problem

When running in a serverless environment, any queries executed by the prisma client should take less time than the timeout imposed by the external environment, so that there can be proper error management instead of the serverless function just timing out.

## Suggested solution

Allow specifying a general request timeout on prisma client.

## Alternatives

Allow specifying a timeout per request.

janpio:
Note that PostgreSQL for example has `statement_timeout`: https://www.postgresql.org/docs/current/runtime-config-client.html#:~:text=statement_timeout%20(integer) This could be set on connection creation to configure each connection to have a set timeout.

millsp:
This could perhaps be experimented on with a simple client extension.

zeevo:
This would really help me as I have some situations where I want to wait in perpetuity for `findMany` calls but timeout for `count` calls.

uhyo:
Hello, let me +1 this and explain another use case.

Recently our database server (PostgreSQL) suddenly stopped responding to any query.
This resulted in existing Prisma connections being hung up indefinitely and subsequent connection pool timeout errors.

Fortunately some of our application code used the `$transaction` API that had default timeout settings. Such code was able to timeout and create a new, working connections.

Global query timeout settings would be beneficial for quicker recovery from this kind of incident. In this case database-level setting like `statement_timeout` is not helpful since DB is in an abnormal state.

mlnj:
The same thing mentioned by `uhyo` happened in production for me. Some of the hung queries were around for  over 6 hours. Had to kill the queries manually when the Prisma client could not create new connections.

Am surprised that this is not a quick setting to apply.

ViktorCollin:
We are seeing similar issues where queries get stuck long after the request that caused the query has timed out

kayflybgd:
I had an issue with query on postgres which was sometimes long to run, and I wanted just to abort after 10 seconds.

I achieved it this way:

```
 const result = await prisma.$transaction(async (tx) => {
      await tx.$executeRawUnsafe('set statement_timeout to 10000;');
      return await tx.$queryRawUnsafe(
        query,
      );
 });

```

NBMSacha:
It's very sursprising that this is still not implemented while even stone-age ORMs like PG have this

MiloWang2048:
[...]

PR #12539, opened 2022-03-26
fix(data-proxy): dont retry gateway timeout errors

endor:
This is towards https://github.com/prisma/prisma/issues/12419. If the data proxy returns a gateway timeout error, the request should not be retried.

### DIFF
--- a/packages/engine-core/src/data-proxy/errors/GatewayTimeoutError.ts
+++ b/packages/engine-core/src/data-proxy/errors/GatewayTimeoutError.ts
@@ -1,0 +1,14 @@
+import type { DataProxyAPIErrorInfo } from './DataProxyAPIError'
+import { DataProxyAPIError } from './DataProxyAPIError'
+import { setRetryable } from './utils/setRetryable'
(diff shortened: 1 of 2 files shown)


millsp (approved):

packages/engine-core/src/data-proxy/errors/GatewayTimeoutError.ts
```diff
@@ -0,0 +1,14 @@
+import type { DataProxyAPIErrorInfo } from './DataProxyAPIError'
+import { DataProxyAPIError } from './DataProxyAPIError'
+import { setRetryable } from './utils/setRetryable'
+
+export interface GatewayTimeoutErrorInfo extends DataProxyAPIErrorInfo {}
+
+export class GatewayTimeoutError extends DataProxyAPIError {
+  public name = 'GatewayTimeoutError'
+  public code = 'P5009'
```
millsp, line 9: Maybe it is already the case, but has this new error code been documented?

PR #12542, opened 2022-03-26
feat(engine-core): allow setting request timeout

### LINKED ISSUE #12419: Specify a request timeout for client

endor:
Closes SCL-28
Closes #12419


packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -111,6 +112,7 @@ export class BinaryEngine extends Engine {
   private lastVersion?: string
   private lastActiveProvider?: ConnectorType
   private activeProvider?: string
+  private requestTimeoutMS?: number
```
matthewmueller, line 115: Is this public facing? Maybe can just call this `timeout`?

millsp (changes requested):
Can we extract the common code into a utility function?

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -19,6 +21,7 @@ import { request } from './utils/request'
 // TODO this is an issue that we cannot share types from the client to other packages
 
 const MAX_RETRIES = 10
+const DEFAULT_REQUEST_TIMEOUT_MS = 30000
```
Jolg42, line 24:
```suggestion
const DEFAULT_REQUEST_TIMEOUT_MS = 30_000
```

Why is the default 30s? I'm curious

Jolg42:
@endor Nice! So this could be used like this?

```
  const db = new PrismaClient({
    __internal: {
      engine: {
        queryTimeout: 10_000,
      },
    }
  })
```

Could you list the current defaults for the queryTimeout for the different engines and the changes with this PR (if any)

What happens if the value of `queryTimeout` is 0 or negative?

Could you add some tests for this?
For Binary and Library engines, I think you could copy this test https://github.com/prisma/prisma/blob/main/packages/client/src/__tests__/integration/happy/client-engine/test.ts

For the DataProxy it would be good to have a test too. We currently test it e2e in https://github.com/prisma/ecosystem-tests/tree/dev/dataproxy but having the test closer to the codebase would be better if we can.

packages/client/src/__tests__/integration/happy/queryTimeout/test.ts
```diff
@@ -0,0 +1,120 @@
+import { ClientEngineType } from '@prisma/sdk'
+import fs from 'fs'
+import os from 'os'
+import path from 'path'
+
+import { generateTestClient } from '../../../../utils/getTestClient'
+
+const buildSchema = (engineType?: string) => `
```
Jolg42, line 8:
```suggestion
const buildSchema = (engineType?: string) => /* Prisma */ `
```
Small trick of our VS Code syntax highlighting

packages/client/src/__tests__/integration/happy/queryTimeout/test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  for (const engineType of engineTypes) {
+    test(
+      `engineType=${engineType} setting a short queryTimeout times out`,
+      buildQueryTimeoutTest({ engineType, queryTimeout: 30, timesOut: true }),
+    )
+
+    test(
+      `engineType=${engineType} setting a long queryTimeout does not time out`,
+      buildQueryTimeoutTest({ engineType, queryTimeout: 30000, timesOut: false }),
```
Jolg42, line 102:
```suggestion
      buildQueryTimeoutTest({ engineType, queryTimeout: 30_000, timesOut: false }),
```

Nit: it's easier to read

packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -428,7 +428,7 @@ You may have to run ${chalk.greenBright('prisma generate')} for your changes to
   }
 
   private async timedRequestInternal(query: string, headers: QueryEngineRequestHeaders) {
-    if (this.queryTimeout) {
+    if (this.queryTimeout !== undefined) {
```
Jolg42, line 431: Why was this needed 🤔 ? Can it be null?

endor: No, but it can be `0`.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
Jolg42: Note from Slack: Magic number, could be any other value that seems reasonable.

packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -6,10 +6,12 @@ import chalk from 'chalk'
 import EventEmitter from 'events'
 import fs from 'fs'
 import path from 'path'
+import { threadId } from 'worker_threads'
```
millsp, line 9: Looks like it was mistakenly added

millsp (approved):
Great job! Looks good to me (+see comment)

packages/engine-core/src/library/LibraryEngine.ts
endor: Fixed.

millsp:
Note: ecosystem-tests are currently failing https://github.com/prisma/ecosystem-tests/actions/runs/2200916963

Jolg42 (changes requested):
We need to check what's not passing in ecosystem-tests before merging

endor:
There will be some delay in getting this resolved due to other priorities, but I'll try to look into it when there's some free time.

Jolg42:
Closing as this is old now.

PR #12560, opened 2022-03-28
test(client): new test client setup

### LINKED ISSUE #11573: Create a test factory for newer and cleaner tests
https://www.notion.so/prismaio/Improving-TypeScript-Tests-b14a42ff4b254e5aa4f821bb57957a0a

millsp:
* Fixed `watch` mode
    * Debounce will now prevent infinite compilation loops (because of intermediary files).
* New entries into `blaze`. But why blaze?
	* It's in magnitudes [faster than ramda, lodash](https://jsben.ch/yAJqf), and even native methods
	* It's built to be type-first, it brings [first class type-safety from ts-toolbelt](https://github.com/millsp/ts-toolbelt)
* Matrix based testing
	* Define your `_matrix.ts` and dynamically run a test suite with all possibilities
	* The matrix is to automatically prepare schemas, and powers multi-provider testing
* Automatic type testing
    * For each generated schema, the test suite is tested against the generated types
    * The type checks will let you know which generated test/schema has failed
* Some of the perks
  * We can now write one test suite and run against many providers
  * Each test suite runs on an isolated and unique database
  * The teardown/setup/generation/types is done automatically
  * No more boiler plate, you can get started in just a few lines of code
  * Generation is as close as it can be to a real-world project

How to run:
`pnpm run test:functional:code && pnpm run test:functional:types`
or just
`pnpm run test:functional`

fixes https://github.com/prisma/prisma/issues/11573


packages/sdk/src/engine-commands/getConfig.ts
```diff
@@ -20,9 +20,9 @@ const unlink = promisify(fs.unlink)
 const MAX_BUFFER = 1_000_000_000
 
 export interface ConfigMetaFormat {
-  datasources: DataSource[] | []
-  generators: GeneratorConfig[] | []
-  warnings: string[] | []
+  datasources: DataSource[]
+  generators: GeneratorConfig[]
+  warnings: string[]
```
Jolg42, line 25: Looks like a revert of a recent change (https://github.com/prisma/prisma/commit/52eed079145ee1ff8f04c8ed775a96da293a6477#diff-318f6444e319ef8973fda14740ee74061b8c384a0c7f06789a8105115aa62ecb)

Any reason to change this?

millsp: It does not look like it's needed anymore? I see the inference issues are gone, but I might be wrong. Let me know :)

Jolg42: About this I think it’s safer to revert and make it `DataSource[] | []`  since it was happens in reality
If you don’t revert this would be valid
`config.datasources[0]`
but could actually be undefined

Jolg42: I discovered that in this commit https://github.com/prisma/prisma/commit/52eed079145ee1ff8f04c8ed775a96da293a6477
changed the type and then adapted to `config.datasources?.[0]`

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/**
+ * Get the generated test suite features, used for client generation.
+ * @param suiteConfig
+ * @returns
+ */
+export function getTestSuitePreviewFeatures(suiteConfig: TestSuiteConfig) {
+  // eslint-disable-next-line prettier/prettier
+  return [...(suiteConfig['#FEATURES']?.split(', ') ?? []), ...(suiteConfig['#EXTRA_FEATURES']?.split(', ') ?? [])]
```
danstarns, line 42: **(non-blocking)** I think this line needs rethinking, its too complicated.

packages/sdk/src/engine-commands/getConfig.ts
millsp: Thanks for the explanation! Fixed. We will try to add `noUncheckedIndexedAccess` globally via the tech debt initiative.

packages/client/tests/functional/_utils/setupQueryEngine.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export async function setupQueryEngine(clientEngineType: ClientEngineType, platform: Platform) {
+  const engineDownloadDir = path.join(__dirname, '..', '..', '..')
+  const queryEngineLibraryPath = path.join(engineDownloadDir, getNodeAPIName(platform, 'fs'))
+  const queryEngineBinaryPath = path.join(
+    engineDownloadDir,
+    `query-engine-${platform}${platform === 'windows' ? '.exe' : ''}`,
+  )
+
+  if (clientEngineType === ClientEngineType.Library && !fs.existsSync(queryEngineLibraryPath)) {
```
danstarns, line 24: **(question)** Could we benefit from using the async fs methods here over synchronous ?

helpers/blaze/range.ts
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
danstarns, line 2: **(non-blocking)** Is this rule, and the few others added in this PR doing anything? `eslint-disable-next-line` but nothing on the next line.

helpers/blaze/select.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  return selected
+}
+
+/**
+ * Select the subset that matches the predicate.
+ *
+ * @param object to be selected
+ * @param pred to select with
+ * @returns
```
danstarns, line 38: **(non-blocking, question)** Are these JS-DOC comments correct or helpful, and does Typescript do this for you? Is this just a convention the codebase follows?

danstarns (commented):
Unsure on this PR as a whole because of my lack of context however added some non blocking observations for you.

helpers/blaze/select.ts
millsp: The TS docs are correct, I'm not sure I get the point about the helpfulness? Maybe you meant that the docs are not clear enough here? In that case we can always elaborate more about what `select` does, yes. No convention, but we aim to have at least minimal docs for all functions (Here it is pretty much WIP so it does not matter so much).

helpers/blaze/range.ts
millsp: Nice catch, will remove.

helpers/blaze/record.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Builds a {@link Record} like TS does.
+ * @param keys
+ * @param values
+ * @returns
+ */
+function record<K extends A.Key, V>(keys: K[], values: V[]): Record<K, V> {
+  return reduce(keys, (acc, key, pos) => ({ ...acc, [key]: values[pos] }), {} as Record<K, V>)
+}
+
+// TODO improve types as values might not be the same length
```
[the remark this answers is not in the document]
millsp: It's a reminder for myself

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
millsp: It's now on two lines so it should be easier to digest.

packages/client/tests/functional/_utils/setupQueryEngine.ts
millsp: Nice catch, will fix.

helpers/blaze/select.ts
danstarns: Cool no problems here, was just curios! Thank you for providing detail.

helpers/blaze/range.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument */
+
+function _range(from: number, to: number, range: number[] = []): number[] {
+  if (from === to) return [...range, from]
+  if (from > to) return _range(from - 1, to, [...range, from])
+  if (from < to) return _range(from + 1, to, [...range, from])
+
+  return range
+}
```
aqrln, line 9: I think it might be problematic to build ranges recursively, on my machine `_range(0, 5248)` already crashes.

helpers/compile/build.ts
```diff
@@ -178,48 +181,53 @@ export async function build(options: BuildOptions[]) {
...
-      console.error(rebuildResult.message)
-    }
+      if (rebuildResult instanceof Error) {
+        console.error(rebuildResult.message)
+      }
 
-    console.log(`${Date.now() - timeBefore}ms`)
-  })
+      console.log(`${Date.now() - timeBefore}ms [${options.name ?? ''}]`)
+    }, 10)
```
aqrln, line 213: I wonder if it makes sense to have a bigger timeout (maybe not)

helpers/compile/build.ts
```diff
@@ -178,48 +181,53 @@ export async function build(options: BuildOptions[]) {
...
-      watchLock = true
-      await build(options)
-    }
-  })
+      // only one watcher will do this task
+      if (watchLock === false) {
+        watchLock = true
+        await build(allOptions)
+      }
+    }, 10)
```
aqrln, line 224: ditto

packages/client/tests/functional/_utils/setupQueryEngine.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param clientEngineType
+ * @param platform
+ */
+export async function setupQueryEngine(clientEngineType: ClientEngineType, platform: Platform) {
+  const engineDownloadDir = path.join(__dirname, '..', '..', '..')
+  const queryEngineLibraryPath = path.join(engineDownloadDir, getNodeAPIName(platform, 'fs'))
+  const queryEngineBinaryPath = path.join(
+    engineDownloadDir,
+    `query-engine-${platform}${platform === 'windows' ? '.exe' : ''}`,
+  )
```
aqrln, line 22: Out of scope of this PR, but: I've seen this repeated in one form or the other in different places in the codebase. I think this should be extracted to a common function similar to `getNodeAPIName`.

helpers/compile/build.ts
millsp: Yep, I was also unsure let's experiment with it and change if needed. For now, that seems to have at least fixed the compilation issues we had in watch mode.

packages/client/tests/functional/_utils/setupQueryEngine.ts
millsp: I'll leave a todo to be tackled via tech debt :+1:

helpers/blaze/range.ts
millsp: Thanks, I replaced it with a loop instead. Should be much faster now.

aqrln (approved):

PR #12066, opened 2022-02-27
fix(engine-core): allow library engine to reestablish connection after initial error

chronotc:
Fixes https://github.com/prisma/prisma/issues/12028

Looking for some help with unit tests ....

PR #12587, opened 2022-03-30
fix(client) Retry start engine on connect

garrensmith:
This fixes the issue where if the client fails to connect to the database
on start up (it throws a P1001) even if the database connection is recovered, the client is
not able to reconnect.


millsp:
https://github.com/prisma/prisma/pull/12066 :smile:

garrensmith:
Closing this for https://github.com/prisma/prisma/pull/12066

PR #12592, opened 2022-03-30
fix(client): count action name

### LINKED ISSUE #12222: Count action ends up being called aggregate when checking in middleware
### Bug description

When you use `.count()` on any model, and try to intercept it in the middleware, the name of the action ends up being `aggregate` instead of `count`

### How to reproduce

I created a middlewere that would override the default `count` and count only entities where `isDeleted: false` like so:

```ts
orm.$use(async (parameters, next) => {
    // Override native `count` and only count non soft deleted entities
    if (parameters.action === 'count') {
        parameters.args = {
            where: {
                isDeleted: false,
            },
        }
    }

    return next(parameters)
})
```

Then I call the `count` like so: 
```ts
const foundSectionsCount = await orm.section.count()
```

What I am expecting to happen is the middleware will be triggered since we are calling `count` and it is checking if the action is named `count`.

Instead, it outputs the `action` as `aggregate`.

Here is the `parameters` log like so:

```
action:  {
  args: { select: { _count: [Object] } },
  dataPath: [],
  runInTransaction: false,
  action: 'aggregate',
  model: 'Section'
}
```

Im assuming `count` is using `aggreate` under the hood but I think the action should persist to be called `count` if we are calling `count`. Especially since the type for `action` specifies `count` exists: 

```ts
export type PrismaAction =
  | 'findUnique'
  | 'findMany'
  | 'findFirst'
  | 'create'
  | 'createMany'
  | 'update'
  | 'updateMany'
  | 'upsert'
  | 'delete'
  | 'deleteMany'
  | 'executeRaw'
  | 'queryRaw'
  | 'aggregate'
  | '**count**'
  | 'runCommandRaw'
```

### Expected behavior

Since I am calling the `.count()` function, I'm expecting the action in the middleware to also be called `count`.

### Prisma information

This is the model im using in  this example

```prisma
model Section {
    name      String
    number    Int?
    createdAt DateTime @default(dbgenerated("timezone('utc'::text, now())")) @map("created_at") @db.Timestamp(6)
    isDeleted Boolean  @default(false) @map("is_deleted")
    id        String   @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
    items     Item[]

    @@map("sections")
}

```

### Environment & setup

- OS: Ubuntu 20.04
- Database: PostgreSQL
- Node.js version: 14.17.6

### Prisma Version

```
^3.10.0
```

millsp:
fixes https://github.com/prisma/prisma/issues/12222

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -18,7 +18,7 @@ import type { PrismaPromise } from './core/request/PrismaPromise'
 import { getLockCountPromise } from './core/transaction/utils/createLockCountPromise'
 import { getCallSite } from './core/utils/getCallSite'
 import { DMMFHelper } from './dmmf'
-import { DMMF } from './dmmf-types'
+import type { DMMF } from './dmmf-types'
 import { getLogLevel } from './getLogLevel'
 import { mergeBy } from './mergeBy'
 import type { EngineMiddleware, Namespace, QueryMiddleware, QueryMiddlewareParams } from './MiddlewareHandler'
@@ -1053,15 +1053,14 @@ new PrismaClient({
         rootField = action
       }
 
-      // TODO: Replace with lookup map for speedup
       let mapping
-      if (model) {
+      if (model !== undefined) {
         mapping = this._dmmf.mappingsMap[model]
-        if (!mapping) {
+        if (mapping === undefined) {
           throw new Error(`Could not find mapping for model ${model}`)
         }
 
-        rootField = mapping[action]
+        rootField = mapping[action === 'count' ? 'aggregate' : action]
       }
 
       if (operation !== 'query' && operation !== 'mutation') {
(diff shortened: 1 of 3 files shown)


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1181,14 +1180,3 @@ function transactionProxy<T>(thing: T, txId: string): T {
     },
   }) as any as T
 }
-
```
millsp, line 1184: Not used anywhere

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1053,15 +1053,14 @@ new PrismaClient({
         rootField = action
       }
 
-      // TODO: Replace with lookup map for speedup
       let mapping
-      if (model) {
+      if (model !== undefined) {
```
Jolg42, line 1057: Can it be null? 🤔

millsp: No, model can either be `string | undefined`. All calls are properly typed here at least 😅

Jolg42 (approved):

Issue #12495, opened 2022-03-23
Implement "CLI engine crash reporting" for Formatter Engine and `getConfig` and `getDmmf` of Query Engine


janpio:
This relies on https://github.com/prisma/prisma/issues/12493 and https://github.com/prisma/prisma/issues/12494 being implemented. (Done)

We want to implement error reporting similar to how we already have it for Migration and Introspection Engine crashes for the Formatter Engine and QueryEngine (`getConfig` and `getDmmf` parts). Via the above mentioned issues the necessary mechanisms in the engines have been implemented, and can now be used here as well.

1. - [x] **Allow triggering the new debug functionality** (implemented via #12493 and #12494) [similar to e.g. in Migrations with an environment variable](https://github.com/prisma/prisma/blob/cc5022f115f4843f4580a31f5bad2cf516a8e32f/packages/migrate/src/MigrateEngine.ts#L277-L280). 
	- (Note that we want to be able to trigger `getConfig` and `getDmmf` crashes separately, so they need a different "trigger" on the CLI side as they are often called together or close to each other.)
	- If one of the issues is not implemented yet, just fake the code so pretend it was already there - and we can add the actual implementation later.
2. - [x] **Make sure errors are correctly thrown** so error handling UI can be triggered to start upload process etc (Internal documentation: https://www.notion.so/prismaio/CLI-Engine-Crash-Reporting-7bad95ea3c5a45f4a5a3250755a324a5#606a4379530c46e1bcd9cd12b107f221)
3. - [x] **Implement CLI Engine Error handling** for all uses of Formatter Engine and QueryEngine (`getConfig` and `getDmmf`)
4. - [x] **Add tests** similar to existing IE and ME error triggering to make sure this keeps working
5. - [x] To be documented at https://www.notion.so/prismaio/CLI-Engine-Crash-Reporting-7bad95ea3c5a45f4a5a3250755a324a5#a44d5d144e144d00920fb2e4d90dfac3 and https://github.com/prisma/prisma/pull/13359

janpio:
Previous example of how to trigger it (might be fixed, but there should be workaround now): https://github.com/prisma/prisma/issues/10206 Also includes a link to https://github.com/prisma/prisma/pull/10205/files that could give ideas for the implementation.

janpio:
Step 1, "Formatter Engine" part was already implemented: https://github.com/prisma/prisma/pull/12569

jkomyno:
[...]

PR #12616, opened 2022-04-01
feat(sdk): Replace prisma-fmt binary with @prisma/prisma-fmt-wasm

tomhoule:
This replaces the binary based implementation of the `prisma format`
command with the wasm module.

- Does not touch or delete any of the download logic, although the
  binary is not used anywhere anymore (as far as I can tell).

### DIFF
--- a/packages/sdk/src/engine-commands/formatSchema.ts
+++ b/packages/sdk/src/engine-commands/formatSchema.ts
@@ -1,13 +1,10 @@
 import Debug from '@prisma/debug'
-import { BinaryType } from '@prisma/fetch-engine'
-import execa from 'execa'
 import fs from 'fs'
+import prismaFmt from '@prisma/prisma-fmt-wasm'
 
 import { resolveBinary } from '../resolveBinary'
 
 const debug = Debug('prisma:formatSchema')
-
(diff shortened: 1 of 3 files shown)


packages/sdk/src/engine-commands/formatSchema.ts
```diff
@@ -19,34 +16,32 @@ export async function formatSchema({ schemaPath, schema }: { schemaPath?: string
     throw new Error(`Parameter schema or schemaPath must be passed.`)
   }
 
-  const prismaFmtPath = await resolveBinary(BinaryType.prismaFmt)
-  const showColors = !process.env.NO_COLOR && process.stdout.isTTY
```
janpio, line 23: Is this still supported?

janpio:
Depends on https://github.com/prisma/prisma/issues/12495 being implemented first to be able to test the panic case, then see if this still works the same afterwards.

packages/sdk/src/engine-commands/formatSchema.ts
tomhoule: This was dead code — reformatting never produces output with terminal colours.

PR #12619, opened 2022-04-01
chore: Revert "chore(deps): update dependency react to v18"

Jolg42:
Reverts prisma/prisma#12578

We need to keep v17 as `peerDependencies`

We could open to >= v17 if there is a need?


millsp (approved):

Jolg42:
Auto merge didn't want to wait 😅 

Should be ok though

Issue #12394, opened 2022-03-17
EXC: Extending the Prisma Client Support


matthewmueller:
## Problem

Unlike many other ORMs the Prisma Client isn't very extendable:
- You can't add custom methods to your models.
- You can't write custom validation.
- You can't exclude certain fields
- You can't group database calls into a logical action (e.g. a signup action may need multiple queries to the database)

To counteract some of these restrictions, we wrote some guides on how to do handle these patterns in your application manually. 

- Custom Validation: https://www.prisma.io/docs/concepts/components/prisma-client/custom-validation
- Computed Fields: https://www.prisma.io/docs/concepts/components/prisma-client/computed-fields
- Excluding Fields: https://www.prisma.io/docs/concepts/components/prisma-client/excluding-fields
- Custom Models: https://www.prisma.io/docs/concepts/components/prisma-client/custom-models

This was a good initial step, but we'd like to start baking some of these patterns into the client itself.

## Tasks

- [ ] EXC: Create a design document discussing the API
- [ ] EXC: Share the design with the community
- [ ] EXC: Finalize design and break down into subsequent tasks

## Reference

- Community Issue: https://github.com/prisma/prisma/issues/7161. There's a ton of good feedback in here.

matthewmueller:
I'd also have a look at the Model and Field sections from the [Feature Matrix](https://www.notion.so/prismaio/Prisma-Client-Feature-Matrix-f559abdd7be54c7ab859db0006f10552) for ideas.

PR #12622, opened 2022-04-01
feat(client): manual transactions

### LINKED ISSUE #12458: Callback-free interactive transactions
## Problem

Some users are asking for a less restricted way of opening an interactive transaction. Instead of having a callback for the transaction, there would be manual calls to `$begin`,  `$commit`, and  `$rollback`.

## Suggested solution

Example of a use-case

```ts
describe('some test', () => {
   let trx: Transaction;

   beforeEach(async () => {
      trx = await prisma.$begin();
      jest.mock('./prisma', () => trx);
   });

   afterEach(async () => {
      await trx.$rollback();
   });

   it('.....', () => {
      // ....
   });
});
```

## Additional context

https://github.com/prisma/prisma/issues/1844#issuecomment-1067348856

sroettering:
Expose API for handling transactions manually.

Closes prisma/prisma#12458

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@ dist/
 !.vscode/launch.json.default
 coverage
 
+.idea
+
 .eslintcache
 
 node_modules
--- a/packages/client/src/__tests__/integration/happy/transaction/manual-transaction.test.ts
+++ b/packages/client/src/__tests__/integration/happy/transaction/manual-transaction.test.ts
@@ -1,0 +1,110 @@
+import crypto from 'crypto'
+import fs from 'fs'
+import path from 'path'
+
+import { TX_ID } from '../../../../runtime/getPrismaClient'
+import { generateTestClient } from '../../../../utils/getTestClient'
+
+beforeAll(() => {
+  fs.copyFileSync(path.join(__dirname, 'dev.db'), path.join(__dirname, 'dev2.db'))
+})
+
+test('manual transaction rollback', async () => {
+  await generateTestClient()
+  const {
+    PrismaClient,
+    Prisma: { prismaVersion },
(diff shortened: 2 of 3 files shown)


sroettering:
This is how simple manual transactions could be done without changing too much. 
I am not sure how nested transactions would and should behave so maybe someone could add that later.
If anything is missing feel free to drop a comment or make changes directly.

matthewmueller:
Hey @sroettering, thanks for this PR! We really appreciate it. 

I don't see us adding this to our core offering because it's too easy to misused this API. If the app throws in the middle of a transaction, we won't be able to trigger a rollback, leading to the transaction to stay open until the timeout.

I do see this being useful in testing environments. What I'd recommend today is creating this as helper functions that you pass into Prisma.

We're also working on a feature we're calling [Extending the Prisma Client](https://github.com/prisma/prisma/issues/12394), where you can add custom methods. We'd like to use this example as a potential plugin and potential use case for that feature.

sroettering:
> Hey @sroettering, thanks for this PR! We really appreciate it.
> 
> I don't see us adding this to our core offering because it's too easy to misused this API. If the app throws in the middle of a transaction, we won't be able to trigger a rollback, leading to the transaction to stay open until the timeout.
> 
> I do see this being useful in testing environments. What I'd recommend today is creating this as helper functions that you pass into Prisma.
> 
> We're also working on a feature we're calling [Extending the Prisma Client](https://github.com/prisma/prisma/issues/12394), where you can add custom methods. We'd like to use this example as a potential plugin and potential use case for that feature.

Ok no problem, this was just a suggestion.

However, if you are able to extend the prisma client with such functionality via plugins or something then your argument with not being able to rollback on errors still kind of holds.

PR #12669, opened 2022-04-05
ci: disable GitHub annotations from linting

Jolg42:
[Internal Slack](https://prisma-company.slack.com/archives/C025RHD851U/p1648733473903529)

Screen Shot 2022-04-05 at 09 13 03

They are from lint, there is even a button on the right side and the full list is in the run https://github.com/prisma/prisma/runs/5824934981?check_suite_focus=true#step:6:14

GitHub only displays the first few

Apparently it’s defined by setup-node action
https://github.com/actions/setup-node/blob/bacd6b4b3ac3127b28a1e1920c23bf1c2cadbb85/src/main.ts#L54-L60

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,6 +93,14 @@ jobs:
           cache: 'pnpm'
           node-version: '12'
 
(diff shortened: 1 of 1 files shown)


millsp (approved):
Thank you very much for the investigation :100:

PR #12660, opened 2022-04-04
chore: remove mongo preview

millsp:
Integrates https://github.com/prisma/prisma/pull/12656
Closes https://github.com/prisma/prisma/issues/12628

PR #12673, opened 2022-04-05
fix(cli): add CockroachDB to the providers list

### LINKED ISSUE #11633: `prisma init` needs CockroachDB in the providers list
### Bug description

  Currently the output of `prisma init` has the following line:

```
2. Set the provider of the datasource block in schema.prisma to match your database: postgresql, mysql, sqlite, sqlserver or mongodb (Preview).
```

This now needs `cockroachdb` added to the list.

### How to reproduce

1. run `npx prisma init`

### Expected behavior

Should include CockroachDB.

### Prisma information

N/A

### Environment & setup

N/A

### Prisma Version

```
3.9.1
```

jkomyno:
fix(cli): add CockroachDB to the providers list

Closes #11633

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -19,6 +19,23 @@ import { isError } from 'util'
 import { printError } from './utils/prompt/utils/print'
 
 export const defaultSchema = (provider: ConnectorType = 'postgresql') => {
+  // add preview flag
+  if (provider === 'cockroachdb') {
+    return `// This is your Prisma schema file,
+    // learn more about it in the docs: https://pris.ly/d/prisma-schema
+    
+    generator client {
+      provider        = "prisma-client-js"
+      previewFeatures = ["${provider}"]
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URL")
+    }
+    `
+  }
+
   return `// This is your Prisma schema file,
 // learn more about it in the docs: https://pris.ly/d/prisma-schema
 
@@ -278,9 +295,7 @@ export class Init implements Command {
             'schema.prisma',
           )} to match your database: ${chalk.green('postgresql')}, ${chalk.green('mysql')}, ${chalk.green(
             'sqlite',
-          )}, ${chalk.green('sqlserver')}, ${chalk.green('mongodb')} or ${chalk.green(
-            'cockroachdb',
-          )} (Preview).`,
+          )}, ${chalk.green('sqlserver')}, ${chalk.green('mongodb')} or ${chalk.green('cockroachdb')} (Preview).`,
         )
       }
 
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -194,9 +194,9 @@ export class Init implements Command {
       url = args['--url']
     } else if (args['--datasource-provider']) {
       const providerLowercase = args['--datasource-provider'].toLowerCase()
-      if (!['postgresql', 'mysql', 'sqlserver', 'sqlite', 'mongodb'].includes(providerLowercase)) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/Init.ts
```diff
@@ -211,9 +211,9 @@ export class Init implements Command {
       url = args['--url']
     } else if (args['--datasource-provider']) {
       const providerLowercase = args['--datasource-provider'].toLowerCase()
-      if (!['postgresql', 'mysql', 'sqlserver', 'sqlite', 'mongodb'].includes(providerLowercase)) {
+      if (!['postgresql', 'mysql', 'sqlserver', 'sqlite', 'mongodb', 'cockroachdb'].includes(providerLowercase)) {
         throw new Error(
-          `Provider "${args['--datasource-provider']}" is invalid or not supported. Try again with "postgresql", "mysql", "sqlite", "sqlserver" or "mongodb".`,
+          `Provider "${args['--datasource-provider']}" is invalid or not supported. Try again with "postgresql", "mysql", "sqlite", "sqlserver", "mongodb", or "cockroachdb".`,
```
jkomyno, line 216: It may be worth creating a small utility that transforms the array of supported datasource providers to a stringified version to append after the "Try again with " string.

janpio: I think that would be an ok optimization to do along the way, on the other hand another stacking PR (towards this one, that will be change to `main` automatically when this here is merged) might be clearer.

jkomyno:
Note: the CI error seems to be triggered by code outside the scope of this PR, as verified by @janpio

packages/cli/src/__tests__/commands/Init.test.ts
```diff
@@ -58,6 +58,26 @@ test('works with provider param - postgresql', async () => {
   `)
 })
 
+test('works with provider param - cockroachdb', async () => {
+  ctx.fixture('init')
+  const result = await ctx.cli('init', '--datasource-provider', 'cockroachdb')
+  expect(stripAnsi(result.stdout)).toMatchSnapshot()
+
+  const schema = fs.readFileSync(join(ctx.tmpDir, 'prisma', 'schema.prisma'), 'utf-8')
+  expect(schema).toMatch(defaultSchema('cockroachdb'))
```
Jolg42, line 67: The default schema needs an extra line for the generator like 
```
```

I can suggest doing it like we did for mongodb and removed here
https://github.com/prisma/prisma/pull/12660/files#diff-3a2fd0c2f24931daa62a694bc42215135099cded6cc9da76ef7c64db4fb6f10fL21-L39

Jolg42 (changes requested):
Looks great! See comment about missing `previewFeatures` for cockroachdb

Jolg42 (approved):
Nice one!

packages/cli/src/Init.ts
```diff
@@ -19,6 +19,23 @@ import { isError } from 'util'
...
+    
+    generator client {
+      provider        = "prisma-client-js"
+      previewFeatures = ["${provider}"]
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URL")
+    }
```
aqrln, line 35: I believe this will result in extra spacing starting from the second line in the generated schema file.

janpio: I think we previously had similar for for MongoDB which was _just_ removed a few days ago. Might also be worth to just quickly go back and check how we did this before.

(Also might be worth to extract the list of providers that should get a preview feature flag to a constant, then we can just keep this around next time - although no provider might be in preview at that time)

jkomyno: I believe @aqrln is right, the `generator` and `datasource` blocks are shifted compared to [how you folks did for MongoDB](https://github.com/prisma/prisma/pull/12660/files#diff-3a2fd0c2f24931daa62a694bc42215135099cded6cc9da76ef7c64db4fb6f10f). New PR incoming to fix it

PR #12696, opened 2022-04-06
fix(cli): remove spacing in cockroachdb preview schema

jkomyno:
Fixes a spacing regression introduced by https://github.com/prisma/prisma/pull/12673.
CC @aqrln @janpio

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -22,18 +22,18 @@ export const defaultSchema = (provider: ConnectorType = 'postgresql') => {
   // add preview flag
   if (provider === 'cockroachdb') {
     return `// This is your Prisma schema file,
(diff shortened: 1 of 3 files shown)


janpio:
Now that we know this was not caught before, can we add a test fo this maybe? Thanks.

Jolg42:
There is already a test though https://github.com/prisma/prisma/blob/57f5129d4c5d5c517597d7f330389467550872ea/packages/cli/src/__tests__/commands/Init.test.ts#L61-L79

We could replace the `expect(schema).toMatch(defaultSchema('cockroachdb'))` by `expect(schema).toMatchSnapshot()` here probably (and in all the tests).

janpio:
Yes, that would have caught this and also makes more sense as a real test 👍

Jolg42 (approved):

aqrln (approved):

PR #12718, opened 2022-04-07
fix(client): add 'findRaw' to PrismaAction for mongodb findRaw middleware queries

roboncode:
TS middleware has error

`This condition will always return 'false' since the types 'PrismaAction' and '"findRaw"' have no overlap.ts(2367)`

Example code:

```typescript
const results = await prisma.user.findRaw({
  filter: { ... }
})
```

```typescript
  if (params.action === 'findRaw') {
    // ...
  }
  return next(params)
})
```

Added action `findRaw` to enumerator.


Jolg42:
Looks like this PR is open since 18 days now.
Someone from client team will need to review this before we can merge @millsp @aqrln

aqrln (approved):

PR #12762, opened 2022-04-11
fix(sdk): Ensure the user is informed when the error report can't be submitted

### LINKED ISSUE #12739: "reportFailedMessage" is never displayed in case a panic error fails to be submitted to the reporting service
Whenever a panic arises and the user decides to report an error, if our error reporting service is not reachable, the Prisma CLI terminates without reporting anything to the user. The root cause is that an exception is silently caught and dropped.
We should thus refactor the panic error reporting mechanism, and write tests that simulate the user interaction.

From [handlePanic.ts](https://github.com/prisma/prisma/blob/c1d4f6ea4a0499759dcc1e43cc6c846ccb8051a8/packages/migrate/src/utils/handlePanic.ts#L51-L66):

```typescript  
  const reportFailedMessage = `${chalk.bold.red('Oops. We could not send the error report.')}`

  if (response.value) {
    let reportId: number | void
    try {
      console.log('Submitting...')
      reportId = await sendPanic(error, cliVersion, engineVersion)
    } catch (error) {
      // this block never executes, as the exeption bubbling out from `sendPanic` is silently dropped
      console.log(reportFailedMessage)
    }
    
    // ...
  }
  ```

### PR DESCRIPTION
**Pull request by jkomyno**
Closes https://github.com/prisma/prisma/issues/12739

- Add tests for `sendPanic`
- Port `handlePanic` from `@prisma/migrate` to `@prisma/sdk`
- Fix typo bugs in existing tests for `handlePanic`, and added additional tests

### CODE REVIEW

**Comment by jkomyno** [comment]
[This commit](https://github.com/prisma/prisma/pull/12762/commits/2607673c9923f57d260bf3749941c31c3b3cdcab) introduces the [`ts-pattern`](https://github.com/gvergnaud/ts-pattern) dependency, which allows performing lazy pattern matching on any TypeScript value à la Rust. In particular, this is used to clarify how the `schema` and `maskedSchema` constants are defined (we could have a similar behavior without this additional dependency if the "if/else" blocks in TypeScript were expressions rather than statements), and how they are depedent on `error.schema` or `error.schemaPath`. Before applying this small refactoring, I have added new schema-related tests in [this commit](https://github.com/prisma/prisma/pull/12762/commits/21cc7cf0ca811ef4fe409ac1d8d94db4efb05aac), ensuring that the same test also pass after applying the refactoring.

I also plan on using `ts-pattern` in [`handlePanic.ts`](https://github.com/prisma/prisma/blob/c1d4f6ea4a0499759dcc1e43cc6c846ccb8051a8/packages/migrate/src/utils/handlePanic.ts#L32-L53), to trigger the error reporting both when the user manually selects the option in the CLI prompt, and when the command that fails is submitted with the `PRISMA_SUBMIT_ERROR_REPORT=1` environment variable (this helps bypassing the manual prompt in the tests). Example:

```typescript
import { match } from 'ts-pattern'

// ...

const shouldSubmitReport: boolean = await match(process.env.PRISMA_SUBMIT_ERROR_REPORT)
.with('1', () => Promise.resolve(true))
.otherwise(async () => {
  const submitErrorPromptResponse = await prompt({
    type: 'select',
    name: 'value',
    message: 'Submit error report',
    initial: 0,
    choices: [
      {
        title: 'Yes',
        value: true,
        description: 'Send error report once',
      },
      {
        title: 'No',
        value: false,
        description: 'Don't send error report',
      },
    ],
  })
  return !!submitErrorPromptResponse.value
})

if (shouldSubmitReport) {
  // call sendPanic, report error message if it fails
}
```

Could you please confirm that this is ok before I proceed further?

packages/sdk/src/sendPanic.ts
```diff
@@ -82,7 +79,7 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
       dbVersion: dbVersion,
     }
 
-    const signedUrl = await createErrorReport(params)
+    const signedUrl = await exports.createErrorReport(params)
```
jkomyno, line 82: `exports.createErrorReport` allows me to mock `createErrorReport` from `jest`, to change its implementation to a function that always throws a specific error `error-report-creation-failed`

packages/sdk/src/sendPanic.ts
```diff
@@ -93,6 +90,7 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
     return id
   } catch (e) {
     debug(e)
+    throw e
```
jkomyno, line 93: This propagates the exception to `handlePanic` as required by https://github.com/prisma/prisma/issues/12739

packages/sdk/src/__tests__/__fixtures__/blog-masked.prisma
```diff
@@ -0,0 +1,37 @@
+datasource db {
+  provider = "sqlite"
+  url = "***"
+}
+
+generator client {
+  provider      = "prisma-client-js"
+  binaryTargets = ["native"]
```
Jolg42, line 8:
```suggestion
```

Should not be needed, is there a reason why it was set?

jkomyno: I have copied [`blog.prisma`](https://github.com/prisma/prisma/blob/main/packages/sdk/src/__tests__/__fixtures__/blog.prisma#L8) and simply replaced the datasource URL with the masked equivalent

Jolg42: I think we could remove on both sides this part since it's not needed, might be more confusing than anything else to ha ve it around

Jolg42:
@jkomyno About the additional env var `PRISMA_SUBMIT_ERROR_REPORT`, I'm not sure it's needed since you could add this line in your test to pass yes for example:

```ts
import prompt from 'prompts'

prompt.inject(['y'])
```

packages/migrate/src/utils/getGithubIssueUrl.ts
```diff
@@ -37,32 +38,33 @@ interface IssueOptions {
...
-          value: false,
-          description: `Don't create a new GitHub issue`,
-        },
-      ],
+  const shouldCreateNewIssue = await match(options.prompt)
+    .with(true, async () => {
+      const createNewIssueResponse = await prompt({
+        type: 'select',
+        name: 'value',
+        message: 'Would you like to create a Github issue?',
```
Jolg42, line 46:
```suggestion
        message: 'Would you like to create a GitHub issue?',
```

Jolg42:
@jkomyno Could you run `pnpm --filter "@prisma/sdk" run test formatSchema -- -u` to update the snapshot here?
Also if you commit the small GitHub nit typo you might need to update some snapshots for that test as well.

Reviewing of the PR in progress, I'll add a "GitHub review" in a bit

packages/migrate/src/__tests__/handlePanic.test.ts
```diff
@@ -13,6 +13,9 @@ import { Migrate } from '../Migrate'
 import { handlePanic } from '../utils/handlePanic'
 import CaptureStdout from './__helpers__/captureStdout'
 
+// follow this issue suggestion: https://github.com/aelbore/esbuild-jest/issues/26#issuecomment-893763840
```
Jolg42, line 16:
```suggestion
// hack to avoid the `TypeError: Cannot redefine property:`
// follow this issue suggestion: https://github.com/aelbore/esbuild-jest/issues/26#issuecomment-893763840
```

jkomyno: Note: I have tried any combination of previous knowledge I had on Jest mocks, plus what I have read in the esbuild-jest issue mentioned there, but somehow I didn't get to mock sendPanic as I wanted too

Jolg42 (commented):
So 2 things:
We need to test in the same context in CI and locally, so an easy way is to add something like that in the test
```
// Disable prompts (Emulate CI env all the time)
process.env.GITHUB_ACTIONS = '1'
```
- Looking at the mocking problem, maybe it just goes away if we cleanup a bit and move “handlePanic” and the test to the sdk package. I’m not sure if there is a good reason to have one part in sdk and one part in migrate package here.

jkomyno:
> So 2 things: We need to test in the same context in CI and locally, so an easy way is to add something like that in the test
> 
> ```
> // Disable prompts (Emulate CI env all the time)
> process.env.GITHUB_ACTIONS = '1'
> ```
> 
> * Looking at the mocking problem, maybe it just goes away if we cleanup a bit and move “handlePanic” and the test to the sdk package. I’m not sure if there is a good reason to have one part in sdk and one part in migrate package here.

Thank you for the feedback. I'll try to move `handlePanic` to "@prisma/sdk", and report here any good reason NOT to do that (should I find any)

jkomyno:
Update about the tests: I have fixed Windows-related issues, integration-tests, and sdk. I'm now focusing on the tests in @prisma/migrate that are currently failing.

packages/sdk/src/__tests__/handlePanic.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      undefined, // schema
+      undefined, // introspectionUrl
+    )
+
+    prompt.inject(['y']) // submit report
+    await handlePanic(rustPanic, cliVersion, engineVersion, command)
+
+    expect(spySendPanic).toHaveBeenCalledTimes(1)
+    expect(spyWouldYouLikeToCreateANewIssue).toHaveBeenCalledTimes(1)
+    expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchSnapshot()
```
janpio, line 132:
```suggestion
    expect(stripAnsi(ctx.mocked['console.log'].mock.calls.join('\n'))).toMatchSnapshot()
```

jkomyno:
# Manual Test

- Run `npx prisma@3.13.0-integration-report-failed-message-is-never-displayed.1 init --datasource-provider sqlite`
- Turn your laptop in flight mode / turn off your WiFi
- Run `FORCE_PANIC_MIGRATION_ENGINE=1 npx prisma@3.13.0-integration-report-failed-message-is-never-displayed.1 db push --skip-generate`
- When prompted to submit an error report, select "Yes"

Before this PR, the user wouldn't be able to see "Oops. We could not send the error report", but now they can.

image

packages/migrate/package.json
```diff
@@ -55,13 +55,12 @@
...
     "mssql": "8.1.0",
-    "new-github-issue-url": "0.2.1",
-    "open": "7.4.2",
     "pg": "8.7.3",
     "pkg-up": "3.1.0",
     "prompts": "2.4.2",
     "strip-ansi": "6.0.1",
-    "strip-indent": "3.0.0"
+    "strip-indent": "3.0.0",
+    "ts-pattern": "^4.0.1"
```
jkomyno, line 63: ts-pattern can probably be removed

pnpm-lock.yaml
```diff
@@ -1584,38 +1592,34 @@ packages:
...
-  /@prisma/studio-pcw/0.459.0_839d6f5e1e290969a8d17594ebaffe12:
+  /@prisma/studio-pcw/0.459.0:
     resolution: {integrity: sha512-V3bvY8XXtRq4dRxtAci3+MAHee0joQU1kVmtjT2RIbFXonnOSo3jkqRMLV2OpNx96oysS5h1u3K86a/EavG2bw==}
     engines: {node: '>= 12'}
     peerDependencies:
       '@prisma/client': '*'
       '@prisma/sdk': '*'
     dependencies:
-      '@prisma/client': link:packages/client
-      '@prisma/sdk': link:packages/sdk
```
Jolg42, line 1595: Could you check which version of pnpm you have, post it here and then try upgrading to latest v6 with 
```
pnpm add -g pnpm
```
And update the lockfile again? I think these will be back after. (Theory to be confirmed)

Jolg42 (approved):
Well done! Nice fix and cleanup here ✨

pnpm-lock.yaml
jkomyno: Currently I have pnpm@6.32.2

jkomyno: `pnpm-lock.yaml` doesn't get updated even when I purge the existing `node_modules` folder.

packages/migrate/src/__tests__/handlePanic.migrate.test.ts
```diff
@@ -221,11 +187,13 @@ describe('handlePanic', () => {
...
-      expect(JSON.stringify(error)).toMatchInlineSnapshot(
-        `{"rustStack":"","area":"LIFT_CLI","schemaPath":"Some Schema Path"}`,
-      )
+    } catch (error) {
+      expect(error).toMatchSnapshot()
+      expect(JSON.parse(JSON.stringify(error))).toMatchObject({
+        area: 'LIFT_CLI',
+        schemaPath,
+      })
+      expect(error.rustStack).toContain('This is the debugPanic artificial panic')
```
janpio, line 196: How can both of this be true?

jkomyno: Because Jest's API names can be misleading, `toMatchObject` actually performs exact match on a subset of properties.

packages/sdk/src/__tests__/__snapshots__/handlePanic.test.ts.snap
```diff
@@ -1,11 +1,13 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`handlePanic ask to submit the panic error in interactive mode 1`] = `
+exports[`handlePanic when sendPanic fails, the user should be alerted by a reportFailedMessage 1`] = `
 "Oops, an unexpected error occured!
-Some error message!
```
janpio, line 5: Not sure if this (outputting the original error I think) is functionality that got lost along the way?

packages/sdk/src/__tests__/getGithubIssueUrl.test.ts
```diff
@@ -9,6 +9,8 @@ describe('getErrorMessageWithLink', () => {
       body: 'This is a body',
     })
 
-    expect(stripAnsi(message)).toMatchInlineSnapshot(`TEST_GITHUB_LINK`)
+    expect(stripAnsi(message)).toMatchInlineSnapshot(
+      `"https://github.com/prisma/prisma/issues/new?body=This+is+a+body&title=This+is+a+title&template=bug_report.md"`,
```
janpio, line 13: What triggered this change?

packages/sdk/src/__tests__/__fixtures__/.gitattributes
```diff
@@ -0,0 +1 @@
+*.prisma text eol=lf
```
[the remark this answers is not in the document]
jkomyno: This is what allows Windows tests not to break when reading the example schema `blog-masked.schema`. I had explained this on call to @Jolg42  earlier, but it's probably best if I jot it down somewhere

packages/sdk/src/__tests__/handlePanic.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+describe('handlePanic', () => {
+  // testing with env https://stackoverflow.com/a/48042799/1345244
+  const OLD_ENV = process.env
+
+  beforeEach(async () => {
+    jest.resetModules() // most important - it clears the cache
+    process.env = { ...OLD_ENV } // make a copy
+    process.cwd = () => testRootDir
+    await mkdir(testRootDir)
```
janpio, line 42: Should not be required: https://github.com/sindresorhus/tempy#tempydirectoryoptions

packages/sdk/src/__tests__/handlePanic.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const ctx = jestContext.new().add(jestConsoleContext()).assemble()
+
+describe('handlePanic', () => {
+  // testing with env https://stackoverflow.com/a/48042799/1345244
+  const OLD_ENV = process.env
+
+  beforeEach(async () => {
+    jest.resetModules() // most important - it clears the cache
+    process.env = { ...OLD_ENV } // make a copy
```
janpio, line 40: I know this comment is only copied from SO, but: Does this really _make_ a copy? Does that not use the "backup" and overwrite the current env values?

packages/sdk/src/__tests__/handlePanic.test.ts
jkomyno: This is straight-out copied from the previously existing [handlePanic.test.ts](https://github.com/prisma/prisma/blob/c1d4f6ea4a0499759dcc1e43cc6c846ccb8051a8/packages/migrate/src/__tests__/handlePanic.test.ts#L56-L61) file. I can "modernize" it as you suggest, though

packages/sdk/src/__tests__/__snapshots__/handlePanic.test.ts.snap
jkomyno: Nice catch, I will investigate this tomorrow

janpio (approved):

packages/migrate/src/__tests__/handlePanic.migrate.test.ts
janpio: Of course 🤷

packages/sdk/src/__tests__/__fixtures__/.gitattributes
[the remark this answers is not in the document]
janpio: I think I have that configured as my default for checkouts via Git on some machines, globally.

packages/sdk/src/__tests__/handlePanic.test.ts
janpio: Not necessary if it effetively works, on the other hand, boy scout rule 😆

packages/sdk/src/__tests__/__fixtures__/.gitattributes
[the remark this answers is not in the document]
jkomyno: That's fine if you have it configured locally, however, this file is primarily needed when checking out the codebase on Windows CI machines. Also, now that I recall our chat with @Jolg42, I should move this .gitattributes file closer to the root directory.

packages/sdk/src/__tests__/__snapshots__/handlePanic.test.ts.snap
jkomyno: Ok, I think I have figured this out: long story short, no functionality got lost, but we mocked a bit too aggressively.

In particular:
- in the `when sendPanic fails, the user should be alerted by a reportFailedMessage` test case, we mock `sendPanic` and we make it fail
-  we then call `expect(stripAnsi(ctx.mocked['console.log'].mock.calls.join('\n'))).toMatchSnapshot()`, which has created the snapshot file above
- since we mock `sendPanic`, this [`debug line here`](https://github.com/prisma/prisma/blob/fix/report-failed-message-is-never-displayed/packages/sdk/src/sendPanic.ts#L92) doesn't get executed

packages/sdk/src/__tests__/getGithubIssueUrl.test.ts
jkomyno: I'm gonna be honest, I don't recall what triggered this change. I'm actually surprised that this test would succeed earlier. Maybe when this specific test was in `@prisma/migrate` it was excluded from running? Or maybe the `getGithubIssueUrl` utility was mocked?

janpio: `TEST_GITHUB_LINK` looks like something we would be the result of a string replacement on the snapshots, we have snapshotserializers somewhere that do this kind of thing. Maybe the URL is covered in the other pacakge, but not here.

Jolg42: Yes here, https://github.com/prisma/prisma/blob/main/packages/sdk/src/utils/jestSnapshotSerializer.js#L43:L45

Though in this case it's better to have the URL I think.

Jolg42 (approved):

packages/sdk/src/__tests__/getGithubIssueUrl.test.ts
jkomyno: So, should I keep my edit as it is now?

packages/sdk/src/__tests__/handlePanic.test.ts
jkomyno: Yes, `process.env = { ...OLD_ENV }` actually makes a shallow copy of the environment variables that were set before the test suite actually ran.  `process.env = OLD_ENV` then restores the original environment variables after each test suite has executed.

packages/sdk/src/__tests__/getGithubIssueUrl.test.ts
Jolg42: I think it's good as it is yes 👍🏼

janpio: As the URL here is also a test URL (I hope), keeping it around is perfectly fine - no additional value in more sanitization here.

packages/sdk/src/__tests__/__snapshots__/handlePanic.test.ts.snap
janpio: As not clear from the message: Was this mocking changed to make sure we actually log the original error message? Or is that tested somewhere else?

PR #12768, opened 2022-04-11
Test NAPI-rs V2 integration

cprieto:


janpio:
Most of the test failures I saw are unrelated and probably caused by this Engine being branched off from an earlier commit, while newer commits were integrated (concrete: https://github.com/prisma/prisma-engines/commit/a2a89a7dfb88dbe9b1a27e8d1a6325b35e1ca41f) and the snapshots updated. Rebasing the Engine branch should fix this (or branching off here from an older commit before that integration)

PR #12770, opened 2022-04-11
chore(migrate): remove sqlite check now performed engine-side

tomhoule:
See https://github.com/prisma/prisma-engines/pull/2836 for the details
on the engines side of things. The consequence of that PR is that we are
now doing that check in engines, and doing it in more cases than the
check removed in this PR.


Jolg42 (approved):
Nice! Less code is definitely better!

PR #12819, opened 2022-04-14
feat(client): expose iTX client type

### LINKED ISSUE #11511: Expose the type that is passed to the interactive $transaction function
## Problem

When using interactive transactions, the type passed to the `$transaction` function is different that the Prisma type:

```ts
prismaClient.$transaction(async (prisma) => {
  // The type of `prisma` here is what this issue refers to
})
```

## Suggested solution

Using the VSCode hover type, I was able to extract this as the type:

```ts
Omit<
   PrismaClient<Prisma.PrismaClientOptions, never, Prisma.RejectOnNotFound | Prisma.RejectPerOperation | undefined>,
   '$connect' | '$disconnect' | '$on' | '$transaction' | '$use'
 >;
```

I think this type should be exposed by the `@prisma/client` package.

aqrln:
Closes #11511


millsp (approved):
Looks good! Let's add a test in the new test setup?

aqrln:
@millsp I wrote down a reminder for myself

PR #12821, opened 2022-04-14
feat(dataproxy): expose error info to users

endor:
Data proxy already sends useful information with error responses. This PR now exposes that information to users.

### DIFF
--- a/packages/engine-core/src/data-proxy/errors/utils/responseToError.ts
+++ b/packages/engine-core/src/data-proxy/errors/utils/responseToError.ts
@@ -16,6 +16,19 @@ export async function responseToError(
 
   const info = { clientVersion, response }
 
+  // Explicitly handle 400 errors which contain known errors
+  if (response.status === 400) {
+    let knownError
+    try {
+      const body = await response.json()
+      knownError = body?.EngineNotStarted?.reason?.KnownEngineStartupError
+    } catch (_) {}
+
+    if (knownError) {
+      throw new BadRequestError(info, knownError.msg, knownError.error_code)
+    }
+  }
+
   if (response.status === 401) {
     throw new UnauthorizedError(info)
   }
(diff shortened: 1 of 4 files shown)


packages/engine-core/src/data-proxy/errors/utils/responseToError.ts
```diff
@@ -40,6 +53,24 @@ export async function responseToError(
...
   if (response.status >= 500) {
+    let body
+    try {
+      body = await response.json()
+    } catch (err) {
+      throw new ServerError(info)
+    }
+
+    if (typeof body?.EngineNotStarted?.reason === 'string') {
+      throw new ServerError(info, body?.EngineNotStarted?.reason)
```
Jolg42, line 64:
```suggestion
      throw new ServerError(info, body.EngineNotStarted.reason)
```
Nit: should work with less `?` here

packages/engine-core/src/data-proxy/errors/utils/responseToError.ts
```diff
@@ -40,6 +53,24 @@ export async function responseToError(
...
+    } else if (typeof body?.EngineNotStarted?.reason === 'object') {
+      const keys = Object.keys(body?.EngineNotStarted?.reason)
+      if (keys.length > 0) {
+        const reason = body?.EngineNotStarted?.reason
```
Jolg42, line 68:
```suggestion
      const keys = Object.keys(body.EngineNotStarted.reason)
      if (keys.length > 0) {
        const reason = body.EngineNotStarted.reason
```
```suggestion
      throw new ServerError(info, body.EngineNotStarted.reason)
```
Nit: should work with less `?` here

Jolg42 (approved):
That looks useful indeed! Nice ✨

janpio:
Is there a way to test this functionality?

PR #12832, opened 2022-04-15
chore(migrate): remove --preview-feature flag from "migrate diff" and "db execute"

### LINKED ISSUE #12679: Remove `--preview-feature` flag for Primitives (`migrate diff` and `db execute`)
... from commands and CLI help output.

Note: Not only the `--preview-feature` needs to be removed, also the "Preview" mentions about these commands in help outputs for example.

Final step of https://github.com/prisma/prisma/issues/12136

jkomyno:
Closes https://github.com/prisma/prisma/issues/12679.

I've also added another test case for each command to make sure that if `--preview-feature` is still used, it triggers a warning.

### DIFF
--- a/packages/migrate/src/commands/DbExecute.ts
+++ b/packages/migrate/src/commands/DbExecute.ts
@@ -7,12 +7,11 @@ import path from 'path'
 
 import { Migrate } from '../Migrate'
 import type { EngineArgs } from '../types'
-import { DbExecuteNeedsPreviewFeatureFlagError } from '../utils/errors'
 
 const helpOptions = format(
   `${chalk.bold('Usage')}
 
-${chalk.dim('$')} prisma db execute --preview-feature [options]
+${chalk.dim('$')} prisma db execute [options]
 
 ${chalk.bold('Options')}
 
@@ -27,7 +26,6 @@ ${chalk.italic('Script input, only 1 must be provided:')}
 
 ${chalk.bold('Flags')}
 
---preview-feature    Run Preview Prisma commands
 --stdin              Use the terminal standard input as the script to be executed`,
 )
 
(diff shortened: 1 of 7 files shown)


packages/migrate/src/commands/MigrateDiff.ts
```diff
@@ -148,8 +146,8 @@ ${chalk.bold('Examples')}
       return this.help()
     }
 
-    if (!args['--preview-feature']) {
-      throw new MigrateDiffNeedsPreviewFeatureFlagError()
+    if (args['--preview-feature']) {
+      console.warn('--preview-feature is deprecated and will be removed in the next major version.')
```
janpio, line 150: I think this can get a nice wording here. Something along the lines that the feature is not a preview feature any more and that the command is not needed any more. I also don't think we would ever really make --preview-feature fail but just remove the warning one day.

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -104,8 +99,8 @@ ${chalk.bold('Examples')}
       return this.help()
     }
 
-    if (!args['--preview-feature']) {
-      throw new DbExecuteNeedsPreviewFeatureFlagError()
+    if (args['--preview-feature']) {
+      console.warn('--preview-feature is deprecated and will be removed in the next major version.')
```
Jolg42, line 103:
```suggestion
      logger.warn(`"prisma db execute" was in Preview and is now Generally Available.
You can now remove the ${chalk.red('--preview-feature')} flag.`)
```
Something like this maybe, also use our `logger.warn()`, useful for Redwood for example so they can disable the warnings.

packages/migrate/src/commands/MigrateDiff.ts
Jolg42:
```suggestion
      logger.warn(`"prisma migrate diff" was in Preview and is now Generally Available.
You can now remove the ${chalk.red('--preview-feature')} flag.`)
```
Something like this maybe, also use our `logger.warn()`, useful for Redwood for example so they can disable the warnings.

Jolg42 (changes requested):
Looks good, but I think there are still these 2 files to check and update:
- https://github.com/prisma/prisma/blob/integration/remove-preview-from-migrate-diff-and-db-execute/packages/migrate/src/commands/MigrateCommand.ts
- https://github.com/prisma/prisma/blob/integration/remove-preview-from-migrate-diff-and-db-execute/packages/migrate/src/commands/DbCommand.ts

jkomyno:
> Looks good, but I think there are still these 2 files to check and update:
> 
> * [https://github.com/prisma/prisma/blob/integration/remove-preview-from-migrate-diff-and-db-execute/packages/migrate/src/commands/MigrateCommand.ts](https://github.com/prisma/prisma/blob/integration/remove-preview-from-migrate-diff-and-db-execute/packages/migrate/src/commands/MigrateCommand.ts?rgh-link-date=2022-04-19T08%3A48%3A33Z)
> * [https://github.com/prisma/prisma/blob/integration/remove-preview-from-migrate-diff-and-db-execute/packages/migrate/src/commands/DbCommand.ts](https://github.com/prisma/prisma/blob/integration/remove-preview-from-migrate-diff-and-db-execute/packages/migrate/src/commands/DbCommand.ts?rgh-link-date=2022-04-19T08%3A48%3A33Z)

Nice catch, I believe I have just fixed that

Jolg42 (approved):
Nice! ✨

jkomyno:
(Re-running the CI because of flaky MongoDB tests)

PR #12835, opened 2022-04-15
fix(cli): `prisma init --datasource-provider=cockroachdb --url ...` now sets `provider` to `cockroach`

### LINKED ISSUE #12360: `prisma init --datasource-provider=cockroachdb --url ...` sets provider to `postgresql`
Note: `--datasource-provider` is not read if `--url`  is passed.

There is an edge case not handled here for cockroachdb urls which will be added with `postgresql` provider (since they start with `postgresql://`)

Using `--datasource-provider=cockroachdb` will work, but with an example url instead of what is supplied via `--url`.

We should probably open that up to allow setting it.

Maybe also document that `--url` leads to it trying to auto detect.

jkomyno:
Closes https://github.com/prisma/prisma/issues/12360.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -119,7 +119,8 @@
     "preinstall": "node scripts/preinstall-entry.js"
   },
   "dependencies": {
-    "@prisma/engines": "3.13.0-9.a10084a836a379babc008c28b143dc1c7e644453"
+    "@prisma/engines": "3.13.0-9.a10084a836a379babc008c28b143dc1c7e644453",
+    "ts-pattern": "^4.0.1"
   },
   "sideEffects": false
 }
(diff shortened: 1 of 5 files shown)


jkomyno:
# Manual Test

- Run `docker-compose -f docker/docker-compose.yml up` to run the `cockroachdb` service (you can comment out the other services)
- Open another terminal
- Run `npx prisma@3.13.0-integration-prisma-init-provider-cockroachdb.1 init --datasource-provider=cockroachdb --url postgres://prisma@localhost:26257/defaultdb`

Before this PR, the generated schema in `prisma/schema.prisma` was

```prisma

generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
}
```

but now it's

```prisma

generator client {
}

datasource db {
  provider = "cockroachdb"
}
```

Jolg42 (approved):
Nice cleanup with ts-pattern ✨

PR #12869, opened 2022-04-18
chore: set endOfLine to auto in prettier configs

YassinEldeeb:
Apparently, the end line in prettier is set to `CR` by default which is supposed to be the end line of old macs(pre-OSX), so better to set it to auto to pick up `CRLF` on Windows and `LF` on Linux/Mac and stop showing these annoying warnings on every line.

Screenshot 2022-04-18 235517

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -3,3 +3,4 @@ trailingComma: all
 singleQuote: true
 semi: false
 printWidth: 120
+endOfLine: auto
--- a/packages/cli/.prettierrc.yml
+++ b/packages/cli/.prettierrc.yml
@@ -3,3 +3,4 @@ trailingComma: all
 singleQuote: true
 semi: false
 printWidth: 120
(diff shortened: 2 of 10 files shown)


YassinEldeeb:
btw, I was curious why there's an identical `.prettierrc.yml` config file in the root and in every package, I guess it's because the Prettier VSC extension can't pick up the prettier config unless It was in the current VSC workspace, so If someone opened his VSC workspace in a particular package, then the VSC extension can search and find the existing config at the root of the package and start working.

wmadden:
Hi @YassinEldeeb, thanks for your contribution! 

Unfortunately, I'm going to decline your PR because we actually want consistent line endings across our files (and we want that line ending to be `\n` not `\r\n`). You're seeing those squiggles presumably because you're on windows and your files have CRLF line endings, which prettier's default setting of `lf` doesn't permit.

PR #12893, opened 2022-04-20
chore(migrate diff): edit help output

Jolg42:
Small changes after [Slack thread](https://prisma-company.slack.com/archives/CGMCB82N8/p1650386592444239)

image


janpio:
- help explanation is still aligned much further right
- optimally all the options would be indented a bit like the output when you run `prisma migrate`
- 2 lines for the to/from is so much better
- suggestion: "uses the datamodel for ..." => "uses the _datamodel_ for ..." (and also for the one below to highlight that difference)
- `--exit-code` I am not sure what the default behavior is and what it is changed to from the explanation

janpio (approved):

Jolg42:
@janpio did some edits, all should be good now 
Screen Shot 2022-04-20 at 12 36 55

aqrln (approved):

PR #12897, opened 2022-04-20
feat(db pull): add spinner in CLI when introspecting

### LINKED ISSUE #12632: MongoDB: Add spinner to Introspection
Usually Introspection is pretty fast. But with MongoDB where we need to look at the documents, it can easily take 30 seconds - where you could get the impression something is stuck:

```
PS C:\Users\Jan\Documents\throwaway\12dev41> npx prisma db pull
Datasource "db"

Introspecting based on datasource defined in prisma\schema.prisma …
Terminate batch job (Y/N)?
^C
PS C:\Users\Jan\Documents\throwaway\12dev41> npx prisma db pull
Datasource "db"

Introspecting based on datasource defined in prisma\schema.prisma …

✔ Introspected 1 model and 7 embedded documents and wrote them into prisma\schema.prisma in 30.86s
      
*** WARNING ***

These fields are not supported by the Prisma Client, because Prisma currently does not support their types.
- Model "listingsAndReviews", field: "bathrooms", original data type: "Decimal128"
- Model "listingsAndReviews", field: "cleaning_fee", original data type: "Decimal128"
- Model "listingsAndReviews", field: "extra_people", original data type: "Decimal128"
- Model "listingsAndReviews", field: "guests_included", original data type: "Decimal128"
- Model "listingsAndReviews", field: "monthly_price", original data type: "Decimal128"
- Model "listingsAndReviews", field: "price", original data type: "Decimal128"
- Model "listingsAndReviews", field: "security_deposit", original data type: "Decimal128"
- Model "listingsAndReviews", field: "weekly_price", original data type: "Decimal128"

```

Here it was me. With a spinner running while only `Introspecting based on datasource defined in prisma\schema.prisma …` is shown, that could easily be avoided.

jkomyno:
Closes https://github.com/prisma/prisma/issues/12632.

- added [`ora`](https://github.com/sindresorhus/ora) spinner library to the dependencies. It's used in a type-friendly wrapper with mock implementations (not to disrupt the jest snapshots)
- added `jestProcessContext` to mock `process.stdout.write` and `process.stderr.write`


janpio (approved):

packages/sdk/src/utils/spinner.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+
+    actualOptions.stream?.write('\n')
+
+    /**
+     * Ships the mocks with the actual implementation in order to have sane jest snapshots.
+     * This is a workaround because jest doesn't support mocking classes.
+     */
+    if (process.env.NODE_ENV === 'test' || process.env.CI) {
```
Jolg42, line 42: 🤔 So Jest here defaults to `NODE_ENV=test` if not already defined.

That also applied to users of the Prisma CLI which might have this env var set.

Maybe we could remove the `(spinner)` parts since it's a bit weird from user perspective to get that?

packages/sdk/package.json
```diff
@@ -70,6 +70,7 @@
     "new-github-issue-url": "0.2.1",
     "node-fetch": "2.6.7",
     "open": "7",
+    "ora": "^5",
```
Jolg42, line 73: Could you replace by the exact version? Renovate will manage the updates after

jkomyno: Sure, I'll stick with the latest "5" version (as version 6 and above require ESM)

packages/sdk/src/utils/spinner.ts
jkomyno: The `(spinner)` text actually appears in the Jest snapshots only, the users will see a rotating thing in the CLI. I thought it could be useful so we can understand from snapshots where the user sees a spinner, vs where the user is left hanging on the terminal. Should I drop it?

Btw, you can test it directly in your terminal by running an introspection on any schema using `prisma@3.13.0-integration-add-spinner-to-introspection.1`.

jkomyno: About the env var, I understand your concerns. Maybe I could drop the factory method altogether, expose a single spinner instance, and hope that we can mock it with jest (even though we've had issues some days ago with mocking functions exposed by other modules)

Jolg42: Note from my DM: Looking at the docs it looks we could simplify since the lib already checks if the context is TTY / CI https://github.com/sindresorhus/ora/tree/v5.4.1#isenabled
```
isEnabled

Type: boolean

Force enable/disable the spinner. If not specified, the spinner will be enabled if the stream is being run inside a TTY context (not spawned or piped) and/or not in a CI environment.

Note that {isEnabled: false} doesn't mean it won't output anything. It just means it won't output the spinner, colors, and other ansi escape codes. It will still log text.
```

packages/sdk/src/utils/spinner.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Methods available to a spinner instance that has already started.
+ */
+export interface SpinnerStarted {
+  success(text?: string): void
+  failure(text?: string): void
+}
+
+/**
+ * Closure that starts a spinner if `enableOutput` is true, and returns a `SpinnerStarted` instance.
+ * Note: the spinner is automatically disabled when `process.env.CI` is defined.
```
Jolg42, line 21:
```suggestion
 * Note: the spinner will only be enabled if the stream is being run inside a TTY context (not spawned or piped) and/or not in a CI environment.
```
From ora README, might be more accurate

packages/sdk/src/utils/spinner.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  return (text: string): SpinnerStarted => {
+    if (!enableOutput) {
+      return {
+        success: () => {},
+        failure: () => {},
+      }
+    }
+
+    actualOptions.stream?.write('\n')
```
Jolg42, line 37: Is that still needed?

jkomyno: I'd leave it as it is, so that the spinner doesn't appear too close to the upward text. But I'm open to get rid of that line if you feel like it :) (though we'll need to update the snapshots)

Jolg42 (approved):

Issue #5030, opened 2021-01-13
Support ES6 modules export


schickling:
It would be great if the Prisma Client JS generator would be compatible with ES6 modules out of the box.

Note: Related to prisma/prisma-client-js#175

pantharshit00:
@Yamboy1 See https://github.com/prisma/prisma/pull/4920

saikatdas0790:
@pantharshit00 Hi, Can we look at this again? This pull request cites NodeJS 10 as the [reason](https://github.com/prisma/prisma/pull/4920#issuecomment-756805461), why this can't be supported. However, release 2.25.0 dropped support for Node 10, so maybe we can support this now?

benbender:
@janpio #7053 is another case where missing native support for es-modules breaks prisma. If you try to use `prisma db seed` in an es-module project (as in `type: module` in the projects package.json, as f.e. in `sveltekit` or `vite`-projects), you'll end up with:

````
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /path-to-my-project/play-prisma-sqlite/prisma/seed.ts
    at new NodeError (node:internal/errors:329:5)
    at Loader.defaultGetFormat [as _getFormat] (node:internal/modules/esm/get_format:71:15)
````

if you follow the documentation for seeding via typescript. Or with:
````
> prisma db seed --preview-feature

Environment variables loaded from .env
Prisma schema loaded from db/schema.prisma
Running seed from "db/seed.js" ...
Result:
internal/modules/cjs/loader.js:1102
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: [redacted]/db/seed.js
require() of ES modules is not supported.
require() of [redacted]/db/seed.js from [redacted]/[eval] is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename seed.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from [redacted]/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1102:13)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:14)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:92:18)
[...]

PR #12907, opened 2022-04-21
fix(client): Reexporting in the way that cjs-module-lexer can detect

tnzk:
# Background

Users of Svelte Kit, Nuxt or other Vite based frameworks haven't been able to import PrismaClient or other exports in the way the documentation suggests[^1]. While there's [several](https://github.com/prisma/prisma/discussions/9027#discussion-3548083) [workarounds](https://github.com/prisma/prisma/issues/5030#issuecomment-970157451) [known](https://www.mikenikles.com/blog/svelte-kit-prisma-a-match-made-in-digital-heaven#commento-comment-name-c13f447ec60eb97a32ae2cb8fe7aaabd760208be22be65d1345f533c5a51f251) in the community, they still have some shortcomings like inability to import Enum values or `Prisma` object for SQL templates nicely.

Those frameworks emit ES modules as their build artifacts, unlike Next that emits CJS bundles. When an ES module tries to `import` named exports from other CJS modules, Node recognizes only the exports that detected in static analysis for specific patterns. [^2]

> For better compatibility with existing usage in the JS ecosystem, Node.js in addition attempts to determine the CommonJS named exports of every imported CommonJS module to provide them as separate ES module exports using a static analysis process.
> 
> The detection of named exports is based on common syntax patterns but does not always correctly detect named exports.

Because of this, current glue code just reexports default export from perspective of Node, while TS compiler somehow recognized the named exports. This causes a weird situation where they would work fine in dev server, but fails with the error message like this once they have been bundled for production, as ES modules with static `import` s.

# Solution

This PR proposes to change the glue code to the form[^3] that Node can recognize all the reexporting of named exports.

[^1]: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/instantiate-prisma-client
[^2]: https://nodejs.org/api/esm.html#commonjs-namespaces
[^3]: https://github.com/nodejs/cjs-module-lexer/tree/1.2.2

I'm relatively new to Prisma, and this is my first PR here. I've confirmed tests below `packages/client` passed, but please tell me if anything need to be merged


janpio:
Amazing 🥇 Thanks for doing this!

We have https://github.com/prisma/ecosystem-tests, where we try to have test projects in all forms and colors where something around Prisma could be going wrong - and seems you mentioned a class of projects that would be worth adding there. Right now they obviously would all be failing, but with your PR here merged they would then succeed - correct? If so, it would be great if you could help us actually create these projects over there. The setup is not the easiest, so maybe just start with an issue where you describe what kind of projects would be missing and then we discuss from there how you can best help us set up a minimal project we can add there.

(I also approved the test run on this PR so you get our CI feedback soon as well)

tnzk:
Sure, ecosystem tests seem interesting! I'll make an issue as a starter and try adding the projects in Svelte Kit if possible.

aqrln (approved):

aqrln:
This looks good to me, thank you!

@janpio is there anything that still needs to be done before merging this (e.g., related to ecosystem tests)? If not, this looks ready to merge to me.

husayt:
this assumes that output is going to be ".prisma". What about overriding `output` in schema, if I set `output  = "/myprisma/client"` then it should use that not `.prisma`

PR #12908, opened 2022-04-21
chore(cli): refactor checkpoint call and sanitize options

Jolg42:
Related https://github.com/prisma/prisma-private/issues/171

### DIFF
--- a/packages/cli/src/__tests__/checkpoint.test.ts
+++ b/packages/cli/src/__tests__/checkpoint.test.ts
@@ -1,12 +1,12 @@
-import { jestConsoleContext, jestContext } from '@prisma/sdk'
+import { jestContext } from '@prisma/sdk'
 
 import { redactCommandArray, SENSITIVE_CLI_OPTIONS, tryToReadDataFromSchema } from '../utils/checkpoint'
 
-const ctx = jestContext.new().add(jestConsoleContext()).assemble()
+const ctx = jestContext.new().assemble()
 
 it('should redact --option [value]', () => {
   for (const option of SENSITIVE_CLI_OPTIONS) {
-    expect(redactCommandArray(['cmd', option, `secret`])).toEqual(['cmd', option, '[redacted]'])
+    expect(redactCommandArray(['cmd', option, 'secret'])).toEqual(['cmd', option, '[redacted]'])
   }
 })
 
@@ -14,6 +14,89 @@ it('should redact --option=[value]', () => {
   for (const option of SENSITIVE_CLI_OPTIONS) {
     expect(redactCommandArray(['cmd', `${option}=secret`])).toEqual(['cmd', `${option}=[redacted]`])
   }
+})
+
+it('should redact a PostgreSQL connection string', () => {
+  expect(redactCommandArray(['init', '--url', '"postgresql://janedoe:mypassword@localhost:5432/mydb?schema=sample"']))
+    .toMatchInlineSnapshot(`
+    Array [
+      init,
+      --url,
+      [redacted],
+    ]
+  `)
+})
+
+it('should redact a MySQL connection string', () => {
+  expect(
+    redactCommandArray([
+      'init',
+      `--url "mysql://janedoe:mypassword@localhost:3306/mydb?connection_limit=5&socket_timeout"`,
+    ]),
+  ).toMatchInlineSnapshot(`
+    Array [
+      init,
+      --url=[redacted],
+    ]
+  `)
+})
+
+it('should redact a MongoDB connection string', () => {
+  expect(
+    redactCommandArray([
+      'init',
+      `--url "mongodb+srv://root:<password>@cluster0.ab1cd.mongodb.net/myDatabase?retryWrites=true&w=majority"`,
+    ]),
+  ).toMatchInlineSnapshot(`
+    Array [
+      init,
+      --url=[redacted],
+    ]
+  `)
+})
+
+it('should redact a SQLite connection string', () => {
+  expect(redactCommandArray(['init', '--url', '"file:./dev.db"'])).toMatchInlineSnapshot(`
+    Array [
+      init,
+      --url,
+      [redacted],
+    ]
+  `)
+})
+
+it('should redact a SQL Server connection string', () => {
+  expect(
+    redactCommandArray([
+      'init',
+      `--url="sqlserver://localhost:1433;initial catalog=sample;user=sa;password=mypassword;"`,
+    ]),
+  ).toMatchInlineSnapshot(`
+    Array [
+      init,
+      --url=[redacted],
+    ]
+  `)
+})
+
+it('should redact a path with for example --schema', () => {
+  expect(redactCommandArray(['cmd', '--schema', '../../../../directory/my_schema.prisma'])).toMatchInlineSnapshot(`
+    Array [
+      cmd,
+      --schema,
+      [redacted],
+    ]
+  `)
+})
+
+it('should redact a name with for example --name', () => {
+  expect(redactCommandArray(['cmd', '--name', '1234_my_name'])).toMatchInlineSnapshot(`
+    Array [
+      cmd,
+      --name,
+      [redacted],
+    ]
+  `)
 })
 
 it('should read data from Prisma schema', async () => {
--- a/packages/cli/src/__tests__/checkpoint.test.ts
+++ b/packages/cli/src/__tests__/checkpoint.test.ts
@@ -1,26 +1,19 @@
 import { jestConsoleContext, jestContext } from '@prisma/sdk'
 
-import { redactCommandArray, tryToReadDataFromSchema } from '../utils/checkpoint'
+import { redactCommandArray, SENSITIVE_CLI_OPTIONS, tryToReadDataFromSchema } from '../utils/checkpoint'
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/__tests__/checkpoint.test.ts
```diff
@@ -0,0 +1,42 @@
...
+})
+
+it('should redact --url=[value]', () => {
+  expect(redactCommandArray(['cmd', '--url=redactme'])).toMatchInlineSnapshot(`
+    Array [
+      cmd,
+      --url=[redacted],
+    ]
+  `)
+})
```
janpio, line 24: Maybe we should add a few more params here?

packages/cli/src/bin.ts
```diff
@@ -164,63 +153,21 @@ async function main(): Promise<number> {
...
-    }
-  } catch (e) {
-    debug(e)
+  /**
+   * Prepare data and run the Checkpoint Client
+   * See function for more info
+   */
+  const checkResult = await runCheckpointClientCheck({
+    commandArray,
+    isPrismaInstalledGlobally,
```
janpio, line 162: Is this only used in checkpoint? Then could also make sense to move inside.

packages/cli/src/bin.ts
```diff
@@ -164,63 +153,21 @@ async function main(): Promise<number> {
...
+    schemaPath: args['--schema'],
+    telemetryInformation: args['--telemetry-information'],
+    version: packageJson.version,
```
janpio, line 165: How did you decide what version is read inside `runCheckpointClientCheck` vs. here? 
Would `packageJson.version` not also make sense to read inside?

packages/cli/src/utils/checkpoint.ts
```diff
@@ -0,0 +1,178 @@
...
+ */
+export async function runCheckpointClientCheck({
+  schemaPath,
+  isPrismaInstalledGlobally,
+  version,
+  commandArray,
+  telemetryInformation,
+}: {
+  schemaPath: string
+  isPrismaInstalledGlobally: 'npm' | 'yarn' | false
```
janpio, line 25: ugh

Jolg42: That's the type, a very weird boolean

packages/cli/src/bin.ts
Jolg42: Since we use the version for the error reporting there as well, it's easier to pass it around and also nicer to not read it from too many places.

packages/cli/src/bin.ts
Jolg42: It's nicer to separate concerns and not put too many things inside one function, here.

packages/cli/src/__tests__/checkpoint.test.ts
Jolg42: Updated the test to test all options

packages/cli/src/utils/checkpoint.ts
janpio: Oh I know. It is just really terrible and bad, but of course works and I know how it got this way.

packages/cli/src/bin.ts
janpio: 🤷

packages/cli/src/bin.ts
janpio: 🦐

janpio (approved):

packages/cli/src/__tests__/checkpoint.test.ts
```diff
@@ -0,0 +1,35 @@
+import { jestConsoleContext, jestContext } from '@prisma/sdk'
+
+import { redactCommandArray, SENSITIVE_CLI_OPTIONS, tryToReadDataFromSchema } from '../utils/checkpoint'
+
+const ctx = jestContext.new().add(jestConsoleContext()).assemble()
```
jkomyno, line 5: You don't really need `jestConsoleContext`, right?
```suggestion
import { jestConsoleContext, jestContext } from '@prisma/sdk'

import { redactCommandArray, SENSITIVE_CLI_OPTIONS, tryToReadDataFromSchema } from '../utils/checkpoint'

const ctx = jestContext.new().add(jestConsoleContext()).assemble()
```

```ts
import { jestContext } from '@prisma/sdk'

import { redactCommandArray, SENSITIVE_CLI_OPTIONS, tryToReadDataFromSchema } from '../utils/checkpoint'

```

packages/cli/src/utils/checkpoint.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+      // else check for partial match and redact the value
+      else if (argIndexPartialMatch !== -1) {
+        commandArray[i] = `${option}=[redacted]`
+      }
+    })
+  }
+
+  return commandArray
+}
```
jkomyno, line 177: This seems to be working, but it might be hard to read/extend this code and understanding the flow.
I'd suggest normalizing the commands first (getting rid of "=", which is a syntactic sugar that is discarded without any additional semantics meaning), and I'd avoid mutating the input array directly.

Consider something like:

```ts
export const redactCommandArray = (commandArray: string[]): string[] => {
  const SENSITIVE_CLI_OPTIONS_SET = new Set(SENSITIVE_CLI_OPTIONS)

  // ['cmd', '--applied', 'secret', '--url=secret'] -> ['cmd', '--applied', 'secret', '--url', 'secret']
  const normalizedCommandArray = commandArray.flatMap((cmd) => cmd.split('='))

  const indexesToRedact = normalizedCommandArray
    .map((cmd, i) => [SENSITIVE_CLI_OPTIONS_SET.has(cmd), i] as const)
    .filter(([isCmdSensitive, _]) => isCmdSensitive)
    .map(([_, i]) => i + 1)

  const REDACTED_TAG = '[redacted]'

  // replace items at `indexesToRedact` with `REDACTED_TAG`
  const redactedCommandArray = [...normalizedCommandArray]
  indexesToRedact.forEach(i => {
    redactedCommandArray[i] = REDACTED_TAG
  })

  return redactedCommandArray
}
```

Should you accept this change, the test case 'should redact --option=[value]' would become essentially identical to 'should redact --option [value]', with a `.toEqual(['cmd', option, '[redacted]'])`.
```suggestion
export const redactCommandArray = (commandArray: string[]) => {
  for (let i = 0; i < commandArray.length; i++) {
    const arg = commandArray[i]
    // redact --option arguments
    SENSITIVE_CLI_OPTIONS.forEach((option: string) => {
      // --url file:./dev.db
      // arg is `--url` and a complete match
      const argIndexCompleteMatch = arg === option
      // --url=file:./dev.db
      // arg value is `--url=file:./dev.db` and a partial match
      const argIndexPartialMatch = arg.indexOf(option)

      // First check for complete match and redact the value
      if (argIndexCompleteMatch) {
        commandArray[i + 1] = '[redacted]'
      }
      // else check for partial match and redact the value
      else if (argIndexPartialMatch !== -1) {
        commandArray[i] = `${option}=[redacted]`
      }
    })
  }

  return commandArray
}
```

packages/cli/src/utils/checkpoint.ts
```diff
@@ -0,0 +1,177 @@
...
+      cli_install_type: isPrismaInstalledGlobally ? 'global' : 'local',
+      // Command with redacted options
+      command: commandAsString,
+      // Internal: Additional information from `--telemetry-information` option or `PRISMA_TELEMETRY_INFORMATION` env var
+      // Default: undefined
+      information: telemetryInformation || process.env.PRISMA_TELEMETRY_INFORMATION,
+      // Absolute CLI path
+      // Note: This won't be sent to the checkpoint server.
+      // TODO: Check if we can remove, probably not needed since cli_path_hash is defined
+      cli_path: process.argv[1],
```
jkomyno, line 67: Is this something that can be checked now, or is it better we wait a future PR?

jkomyno (changes requested):
Overall this seems going to the right direction, that's great! Please check the above comments though

packages/cli/src/__tests__/checkpoint.test.ts
Jolg42: Weird, it shows as a suggestion I cannot just add/commit

packages/cli/src/utils/checkpoint.ts
Jolg42: I think we'll have the opportunity to do it later and merge this now

packages/cli/src/utils/checkpoint.ts
Jolg42: Since the connection strings likely have `=` chars this would mess up the `commandArray` by removing them.
So I don't think we can apply this as is, though I like it 👍🏼 

Examples
```
init --url="postgresql://janedoe:mypassword@localhost:5432/mydb?schema=sample"
init --url="sqlserver://localhost:1433;initial catalog=sample;user=sa;password=mypassword;"
```
https://www.prisma.io/docs/reference/database-reference/connection-urls

jkomyno: Thank you for pointing out this edge case!

jkomyno (approved):

jkomyno (approved):

PR #12910, opened 2022-04-21
fix(db pull): --url no longer overrides the datasource provider

### LINKED ISSUE #11527: Using `--url` for `db pull` ignored/overwrites `provider=cockroachdb`
With this schema:
```prisma
generator client {
}

datasource db {
  provider = "cockroachdb"
  url      = "postgresql://jo_l:11...oud:26257/introspection-ci-5134.aragon-notification-service?sslmode=verify-full"
}
```

this is the output of `prisma db pull --url=postgresql://jo....cloud:26257/introspection-ci-5134.aragon-notification-service?sslmode=verify-full`:
```
λ npx prisma db pull --url=postgresql://jo_l:1...oud:26257/introspection-ci-5134.aragon-notification-service?sslmode=verify-full
Datasource "db": CockroachDB database "introspection-ci-5134.aragon-notification-service", schema "public" at "free-tier...ud:26257"

Introspecting …

✔ Introspected 6 models and wrote them into prisma\schema.prisma in 6.80s

```
(which is great!)

but this the resulting schema:

```
generator client {
}

datasource db {
  provider = "postgresql"
  url      = "postgresql://jo_l:11ekBxMQuUr07kHv@free-tier.gcp-us-central1.cockroachlabs.cloud:26257/introspection-ci-5134.aragon-notification-service?sslmode=verify-full"
}

model eventsources {
  eventsource_id   BigInt          @id
  created_at       DateTime?       @default(now()) @db.Timestamptz(6)
  enabled          Boolean         @default(true)
  contract_address String          @db.VarChar(255)
```

Note how it says `provider = "postgresql"`  there.

I think it would be ok to keep ´provider=cockroachdb` around when already set in the original schema.

(Everything works fine when you just run `db pull` without `--url`)

Related: https://github.com/prisma/prisma/issues/11347

jkomyno:
Closes https://github.com/prisma/prisma/issues/11527.


jkomyno:
We currently have 45 successful checks. Let's see how it goes using the `cockroachdb` service

jkomyno:
In order to fix the tests in this PR, we need:
- a valid `.github/setup-cockroachdb.sh` script, with a related "Setup CockroachDB" CI step for the "no-docker" workflow
- understand how to make sure the same snapshots that work for Github Actions also work for Buildkite

cc @janpio @Jolg42

janpio:
I think you can skip the Cockroach related tests in no-docker - even the Docker configuration for CockroachDB is still very much in flux and will change in the coming does.

What exactly is the problem with the snapshots?

jkomyno:
The problem with snapshots in Buildkite is that the URL with `cockroachdb` as host is printed, whereas in Github Actions the `localhost` URL is printed, but if I change the "COCKROACH" env var in Buildkite to the same used in Github Actions, I get a connection error. Please see the last 2/3 commits for further details

janpio:
Ok, that difference between the docker-compose files for Buildkite and GH Actions always confuses me as well. I do not know how we deal with that usually, but @Jolg42 probably does and can help out here. (If all fails we can sanitize it away in the snapshot)

.buildkite/test/docker-compose.14.yml
```diff
@@ -31,7 +31,7 @@ services:
       - TEST_MONGO_URI=mongodb://root:prisma@mongo:27018/tests?authSource=admin
       - TEST_MONGO_URI_MIGRATE=mongodb://root:prisma@mongodb_migrate:27017/tests-migrate?authSource=admin
       # CockroachDB
-      - TEST_COCKROACH_URI=postgresql://prisma@cockroachdb:26257/
+      - TEST_COCKROACH_URI=postgresql://prisma@localhost:26257/tests
```
Jolg42, line 34:
```suggestion
      - TEST_COCKROACH_URI=postgresql://prisma@cockroachdb:26257/tests
```
Host needs to be `cockroachdb` in Buildkite (name of the docker-compose step)

.buildkite/publish/docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
       - TEST_MONGO_URI=mongodb://root:prisma@mongo:27018/tests?authSource=admin
       - TEST_MONGO_URI_MIGRATE=mongodb://root:prisma@mongodb_migrate:27017/tests-migrate?authSource=admin
       # CockroachDB
-      - TEST_COCKROACH_URI=postgresql://prisma@cockroachdb:26257/
+      - TEST_COCKROACH_URI=postgresql://prisma@localhost:26257/tests
```
Jolg42, line 34:
```suggestion
      - TEST_COCKROACH_URI=postgresql://prisma@cockroachdb:26257/tests
```
Host needs to be `cockroachdb` in Buildkite (name of the docker-compose step)

packages/migrate/src/__tests__/__snapshots__/DbPull.test.ts.snap
```diff
@@ -66,6 +66,121 @@ model jobs {
 }
 
 
+// introspectionSchemaVersion: NonPrisma,
+`;
+
+exports[`cockroachdb basic introspection --url 2`] = `
+datasource db {
+  provider = "postgresql"
+  url      = "postgresql://prisma@localhost:26257/tests"
```
Jolg42, line 75: The value here will be `cockroachdb` instead of `localhost` for Buildkite so we can normalize this here 
https://github.com/prisma/prisma/blob/main/packages/sdk/src/utils/jestSnapshotSerializer.js#L73:L75

jkomyno: Thanks, this was the sanitization part I was missing yesterday!

jkomyno: @Jolg42 still, we need a `.github/setup-cockroachdb.sh` file. I'll try to write one, but I'm not that familiar with cockroachdb

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -491,6 +491,64 @@ describe('postgresql', () => {
...
+    await expect(result).resolves.toMatchInlineSnapshot(``)
+    expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchSnapshot()
+    expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+  })
+
+  // TODO: Windows: fails with
+  // Error: P1012 Introspection failed as your current Prisma schema file is invalid·
+  //     Please fix your current schema manually, use prisma validate to confirm it is valid and then run this command again.
+  //     Or run this command with the --force flag to ignore your current schema and overwrite it. All local modifications will be lost.
```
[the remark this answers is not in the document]
jkomyno: Both. Copy-paste of another test case (just plugging cockroachdb in it) which happens to fail with that error on Windows

packages/migrate/src/commands/DbPull.ts
```diff
@@ -144,22 +146,33 @@ Set composite types introspection depth to 2 levels
...
-    } else if (schemaPath) {
-      schema = fs.readFileSync(schemaPath, 'utf-8')
-    } else {
-      throw new Error('Could not find a `schema.prisma` file')
-    }
+          if (input.url) {
+            const config = await getConfig({
+              datamodel: rawSchema,
+              ignoreEnvVarErrors: true,
+            })
```
Jolg42, line 159: Note: one more place where we use `getConfig()` @janpio

packages/migrate/src/commands/DbPull.ts
```diff
@@ -231,6 +244,9 @@ Then you can run ${chalk.green(getCommandWithExecutor('prisma db pull'))} again.
       } else if (e.code === 'P1012') {
         // Schema Parsing Error
         console.info() // empty line
+
+        // TODO: this error is misleading, as it gets thrown even when the schema is valid, put the protocol of the given
+        // '--url' argument is different than the one written in the schema.prisma file.
```
Jolg42, line 249: Do you have a reproduction? Creating a small issue would be good to track this 📝

packages/migrate/src/commands/DbPull.ts
```diff
@@ -144,22 +146,33 @@ Set composite types introspection depth to 2 levels
...
-    } else {
-      throw new Error('Could not find a `schema.prisma` file')
-    }
+          if (input.url) {
+            const config = await getConfig({
+              datamodel: rawSchema,
+              ignoreEnvVarErrors: true,
+            })
+            const provider = config.datasources[0]?.provider
+            const schema = this.urlToDatasource(input.url, provider) + removeDatasource(rawSchema)
```
Jolg42, line 161:
```suggestion
            const schema = `${this.urlToDatasource(input.url, provider)}${removeDatasource(rawSchema)}`
```
Usually, it's a safer default to go for template string concatenation
It should do the same in the end.

packages/migrate/src/commands/DbPull.ts
```diff
@@ -144,22 +146,33 @@ Set composite types introspection depth to 2 levels
       throw new NoSchemaFoundError()
     }
 
-    let schema: string | null = null
+    const schema = await match({ url, schemaPath })
+      .when(
```
Jolg42, line 150: Could use some small comments about the different cases maybe, so it's easier to read without going to the details?

packages/migrate/src/commands/DbPull.ts
jkomyno: Right, check the new commit!

packages/migrate/src/commands/DbPull.ts
jkomyno: Right, check the new commit!

Jolg42:
a test fails with 
```

  ● cockroachdb › basic introspection --url

    expect(received).resolves.toMatchInlineSnapshot()

    Received promise rejected instead of resolved
    Rejected to value: [Error: [introspection-engine/connectors/sql-introspection-connector/src/version_checker.rs:78:97] called `Result::unwrap()` on an `Err` value: Error("unknown variant `String`, expected one of `SmallInt`, `Integer`, `BigInt`, `Decimal`, `Money`, `Inet`, `Oid`, `Citext`, `Real`, `DoublePrecision`, `VarChar`, `Char`, `Text`, `ByteA`, `Timestamp`, `Timestamptz`, `Date`, `Time`, `Timetz`, `Boolean`, `Bit`, `VarBit`, `Uuid`, `Xml`, `Json`, `JsonB`", line: 0, column: 0)]
```

Might be since the merge of main in this PR

Jolg42 (approved):
Nice PR! Last step would be to fix the test error and then merging time :shipit:

packages/migrate/src/commands/DbPull.ts
jkomyno: Here is how to reproduce:

- Initialize a new schema with e.g. the `mongodb` provider: `npx prisma init --datasource-provider mongodb`
- Ensure the schema is valid: `npx prisma validate`
- Run introspection with a URL having a different datasource protocol: `npx prisma db pull --url postgres://prisma:prisma@localhost:5432/tests`

You should see an error similar to the following:

```text
Datasource "db"

✖ Introspecting

Error: P1012 Introspection failed as your current Prisma schema file is invalid

Please fix your current schema manually, use prisma validate to confirm it is valid and then run this command again.
Or run this command with the --force flag to ignore your current schema and overwrite it. All local modifications will be lost.
```

Jolg42:
Tests should pass after https://github.com/prisma/prisma-engines/pull/2872 is merged

packages/migrate/src/__tests__/__snapshots__/DbPull.test.ts.snap
```diff
@@ -146,6 +181,45 @@ enum Role {
...
+  coinflips Boolean[]
+  User      User?     @relation(fields: [authorId], references: [id])
+}
+
+model User {
+  id    String  @id
+  email String  @unique(map: "User.email")
+  name  String?
+  Post  Post[]
+}
```
janpio, line 215: We should probably make some of the `String` fields different native types, that are different between CockroachDB and PostgreSQL provider.

jkomyno: So, in this case they should be `VarChars`, right? Since we have a postgresql schema

Jolg42 (changes requested):
We discussed that we might want to go through some things here before merging.

packages/migrate/src/commands/DbPull.ts
```diff
@@ -231,6 +256,9 @@ Then you can run ${chalk.green(getCommandWithExecutor('prisma db pull'))} again.
       } else if (e.code === 'P1012') {
         // Schema Parsing Error
         console.info() // empty line
+
+        // TODO: this error is misleading, as it gets thrown even when the schema is valid but the protocol of the given
+        // '--url' argument is different than the one written in the schema.prisma file.
```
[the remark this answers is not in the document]
jkomyno: Here is how to reproduce:

- Initialize a new schema with e.g. the `mongodb` provider: `npx prisma init --datasource-provider mongodb`
- Ensure the schema is valid: `npx prisma validate`
- Run introspection with a URL having a different datasource protocol: `npx prisma db pull --url postgres://prisma:prisma@localhost:5432/tests`

You should see an error similar to the following:

```text
Datasource "db"

✖ Introspecting

Error: P1012 Introspection failed as your current Prisma schema file is invalid

Please fix your current schema manually, use prisma validate to confirm it is valid and then run this command again.
Or run this command with the --force flag to ignore your current schema and overwrite it. All local modifications will be lost.
```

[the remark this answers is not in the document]
janpio: Ok, then my analysis would be that the combination of `provider=mongodb` and the `--url` value lead to validation failing before Introspeciton is run. (Similar how setting an env var with a mismatched `protocol` would lead to a similar problem)

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -58,8 +59,7 @@ describe('common/sqlite', () => {
     expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
   })
 
-  test('basic introspection with invalid --url', async () => {
-    ctx.fixture('introspection/sqlite')
+  test('basic introspection with invalid --url if schema is unspecified', async () => {
```
janpio, line 62: why this change and removal of fixtures on that existing test?

.buildkite/publish/docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
       - TEST_MONGO_URI=mongodb://root:prisma@mongo:27018/tests?authSource=admin
       - TEST_MONGO_URI_MIGRATE=mongodb://root:prisma@mongodb_migrate:27017/tests-migrate?authSource=admin
       # CockroachDB
-      - TEST_COCKROACH_URI=postgresql://prisma@cockroachdb:26257/
+      - TEST_COCKROACH_URI=postgresql://prisma@cockroachdb:26257/tests
```
janpio, line 34: Is this env var already used anywhere in code? 
(If thought there were not CockroachDB tests - why do we have the env var then?)

packages/migrate/src/__tests__/DbPull.test.ts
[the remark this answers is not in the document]
janpio: That indeed sounds like something we need to figure out still.

.buildkite/publish/docker-compose.yml
jkomyno: This PR is the first that introduces cockroachdb tests

packages/migrate/src/__tests__/DbPull.test.ts
jkomyno: Because the schema is automatically inferred if `prisma/prisma.schema` is available in the current location. We wanted to test the case in which the `--url` flag is executed without any knowledge of a prisma schema.

packages/migrate/src/__tests__/DbPull.test.ts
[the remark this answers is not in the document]
jkomyno: Agree, but in another PR imho. The same "Windows fails with error P1012" is excluded from the CI for other databases in other test cases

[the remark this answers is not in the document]
janpio: Ok then please create an issue and optimally link it in the TODO

packages/migrate/src/__tests__/DbPull.test.ts
janpio: And the previously existing test is useless and is not needed any more?

packages/migrate/src/commands/DbPull.ts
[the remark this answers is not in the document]
janpio: Can you open an issue for this TODO as well pls?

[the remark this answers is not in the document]
jkomyno: https://github.com/prisma/prisma/issues/13076

packages/migrate/src/__tests__/DbPull.test.ts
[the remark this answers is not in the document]
jkomyno: Opened https://github.com/prisma/prisma/issues/13077.

packages/migrate/src/__tests__/__snapshots__/DbPull.test.ts.snap
jkomyno: On a closer inspection, the way these models are generated is not dependent on the code shipped with this PR, as both the postgres schema and the cockroach one have no model defined. We are merely introspecting existing tables in the postgres and cockroach images running on Docker. I have just merged `main` in `integration/fix-db-pull-url-overwrites-provider-cockroachdb` to trigger the CI, let's see if anything changes.

jkomyno: The CI succeeds nonetheless, I suggest we open a new issue to verify why the introspection results in "general types" rather than "native types" (if that's unexpected at all), and merge this PR in the meantime.
Do you agree? @janpio @Jolg42

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -491,6 +491,74 @@ describe('postgresql', () => {
...
+  test('basic introspection (no schema) --url', async () => {
+    const introspect = new DbPull()
+    const result = introspect.parse(['--print', '--url', setupParams.connectionString])
+    await expect(result).resolves.toMatchInlineSnapshot(``)
+    expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchSnapshot()
+    expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+  })
+
+  // TODO: Windows: fails with
```
janpio, line 547:
```suggestion
  // TODO: (https://github.com/prisma/prisma/issues/13077) Windows: fails with
```

packages/migrate/src/__tests__/__snapshots__/DbPull.test.ts.snap
```diff
@@ -66,6 +66,160 @@ model jobs {
 }
 
 
+// introspectionSchemaVersion: NonPrisma,
+`;
+
+exports[`cockroachdb basic introspection (no schema) --url 2`] = `
+datasource db {
+  provider = "postgresql"
```
janpio, line 74: Is this expected to be PostgreSQL?

janpio: Yes, as we only use the provider protocol or the existing provider in the schema file.

Jolg42 (approved):
💯 Here we gooo

PR #12984, opened 2022-04-25
fix(client/cli): Env Var Conflict Checking

### LINKED ISSUE #10958: Env var conflict checking only throws if values are the same
### Bug description

The env var checking mechanism used when reading multiple env files reports an error if the value of a var is identical and ignores vars with different values. See:

https://github.com/prisma/prisma/blob/main/packages/sdk/src/utils/tryLoadEnvs.ts#L86

### How to reproduce

The relevant test is here:

https://github.com/prisma/prisma/tree/main/packages/client/src/__tests__/integration/happy/blog-dot-env-both-conflict>

### Expected behavior

It seems like the opposite behaviour is desirable - an error should be shown when vars are different but duplicate names should be permitted if the values are the same.

### Prisma information

N/A

### Environment & setup

- OS: Mac OS
- Database: PostgreSQL
- Node.js version: 16.6.1

### Prisma Version

```
prisma                  : 3.7.0
Query Engine (Node-API) : libquery-engine 8746e055198f517658c08a0c426c7eec87f5a85f (at ../common/prisma/node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 8746e055198f517658c08a0c426c7eec87f5a85f (at ../common/prisma/node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core 8746e055198f517658c08a0c426c7eec87f5a85f (at ../common/prisma/node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt 8746e055198f517658c08a0c426c7eec87f5a85f (at ../common/prisma/node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : 8746e055198f517658c08a0c426c7eec87f5a85f
Studio                  : 0.445.0

```

danstarns:
Removes the error path for conflicting envs and defaults to warn. 

Closes: https://github.com/prisma/prisma/issues/10958


packages/cli/junit.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
danstarns, line 1: This file is autogenerated but not excluded from .gitignore, should it be kept?

Jolg42: @danstarns This is a recent new file, looks like we forgot to add to .gitignore, could you remove it from this PR and add it in .gitignore?

Jolg42: Was introduced in https://github.com/prisma/prisma/pull/12044

danstarns: Sure!

packages/sdk/src/utils/tryLoadEnvs.ts
```diff
@@ -37,14 +37,14 @@ export function tryLoadEnvs(
...
-  opts: { conflictCheck: 'warn' | 'error' | 'none' } = {
+  opts: { conflictCheck: 'warn' | 'none' } = {
     conflictCheck: 'none',
   },
 ): LoadedEnv {
   const rootEnvInfo = loadEnv(rootEnvPath)
   if (opts.conflictCheck !== 'none') {
     // This will throw an error if there are conflicts
-    checkForConflicts(rootEnvInfo, schemaEnvPath, opts.conflictCheck)
+    checkForConflicts(rootEnvInfo, schemaEnvPath)
```
Jolg42, line 47: Here a default value 'none' is assigned to `conflictCheck` but then the value is not passed anymore to `checkForConflicts()`.
Which means we also removed the `conflictCheck: 'none'` behavior?

Jolg42 (commented):
In the issue the user mentions this
>It seems like the opposite behaviour is desirable - an error should be shown when vars are different but duplicate names should be permitted if the values are the same.

In this PR it looks like the change is to remove the thrown error and only warn, which is different from what the user expected. Is that what we want to do?

danstarns:
> In the issue the user mentions this
> 
> > It seems like the opposite behaviour is desirable - an error should be shown when vars are different but duplicate names should be permitted if the values are the same.
> 
> In this PR it looks like the change is to remove the thrown error and only warn, which is different from what the used expected. Is that what we want to do?

Good observation, I wasn't sure if "error should be **shown**" was the same as "error should be **thrown**".  This PR does indeed remove the throwing of the error but [keeps the showing of it. ](https://github.com/danstarns/prisma/blob/fix/conflict-envs/packages/cli/src/__tests__/dotenv-3-conflict.test.ts#L11). I believe that the warning is sufficient, however, realise this could be seen as breaking...

aqrln:
> In the issue the user mentions this
>
> > It seems like the opposite behaviour is desirable - an error should be shown when vars are different but duplicate names should be permitted if the values are the same.
>
> In this PR it looks like the change is to remove the thrown error and only warn, which is different from what the used expected.

That was my understanding too, just flip the check, these changes look far outside the scope of that issue.

danstarns:
My bad I completely miss read this one, here I have flipped the check.

danstarns:
After flipping the switch, and waiting for the suite to run, it looks like we need to differentiate between the conflicts and duplicates, here I've done that!

packages/sdk/src/utils/tryLoadEnvs.ts
```diff
@@ -84,34 +84,46 @@ function checkForConflicts(
...
     const conflicts: string[] = []
+    const duplicates: string[] = []
+
     for (const k in envConfig) {
-      if (parsedRootEnv[k] === envConfig[k]) {
-        conflicts.push(k)
+      const a = parsedRootEnv[k]
+      const b = envConfig[k]
+
+      if (a && b) {
```
aqrln, line 93: Can either of `a` or `b` be an empty string?

packages/sdk/src/utils/tryLoadEnvs.ts
```diff
@@ -84,34 +84,46 @@ function checkForConflicts(
...
-      } else if (type === 'warn') {
-        const message = `Conflict for env var${conflicts.length > 1 ? 's' : ''} ${conflicts
-          .map((c) => chalk.bold(c))
-          .join(', ')} in ${chalk.underline(relativeRootEnvPath)} and ${chalk.underline(relativeEnvPath)}
+        relativeRootEnvPath,
+      )} to consolidate your env vars.\n`
+      throw new Error(message)
+    }
+
+    if (type === 'warn' && duplicates.length) {
```
aqrln, line 120: What's the reason for preserving the old behavior when `type === 'warn'`?

My understanding was that the only thing that needed to change was the condition on line 88 (`if (parsedRootEnv[k] !== envConfig[k]) {` instead of `if (parsedRootEnv[k] === envConfig[k]) {`). Am I missing something?

aqrln: Oh, I missed the [comment](https://github.com/prisma/prisma/pull/12984#issuecomment-1110832049) about this.

I'm starting to have a feeling that it might be a lot more complicated that it seemed, and it would be a good topic for a TS Codebase Deep Dive meeting

millsp:
Ping

PR #13029, opened 2022-04-27
chore: update .gitignore

danstarns:
Adds some exclusion to the gitignore, all these are added when running the test suite. 

Untitled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,7 @@ esm
 reproductions/*
 !reproductions/basic-sqlite
 !reproductions/pnpm-workspace.yaml
+
+dev.db
+junit.xml
+introspection.test.ts.snap
--- a/.gitignore
+++ b/.gitignore
@@ -41,4 +41,3 @@ reproductions/*
 
 dev.db
 junit.xml
-introspection.test.ts.snap


millsp:
Curious to see what changed in your `introspection.test.ts.snap`, looks like it could be a snapshot serializer issue on Windows. Would you mind sharing the diff?

For `dev.db`, maybe we can restrict it to the the `src/__tests__` folder only? How did `dev.db` endup in `sdk` and `client`?

millsp (commented):
Curious to see what changed in your `introspection.test.ts.snap`, looks like it could be a snapshot serializer issue on Windows. Would you mind sharing the diff?

For `dev.db`, maybe we can restrict it to the the `src/__tests__` folder only? How did `dev.db` endup in `sdk` and `client`?

.gitignore
```diff
@@ -38,3 +38,7 @@ esm
 reproductions/*
 !reproductions/basic-sqlite
 !reproductions/pnpm-workspace.yaml
+
+dev.db
+junit.xml
+introspection.test.ts.snap
```
janpio, line 44: That one should not be necessary.

Jolg42, line 44:
```suggestion
```
Indeed, this one should be kept, what's the diff for you here for this file?

danstarns:
I think this has to do with something on windows but I'm not sure exactly what the issue is, possibly line endings. Git thinks there is diff but visibly the changes are not there.

aqrln (approved):

PR #13031, opened 2022-04-27
fix(client): Fix package path in `generate` output

### LINKED ISSUE #7809: `generate` output with `output` has weird package path on Windows
````
λ npx prisma generate

✔ Generated Prisma Client (2.25.0) to .\prisma\foo in 738ms
```
import { PrismaClient } from './prisma\foo'
```
````

Note the `./prisma\foo`

### PR DESCRIPTION
**Pull request by flatplate**
Previously after running generate with a custom output folder on Windows, code snippet that was shown had backslashes in the import statement rather than forward slashes.

This commit fixes that by replacing the backslashes in the path string with forward slashes if the platform is windows.

Fixes https://github.com/prisma/prisma/issues/7809

I am not sure how one would write a test for this, it being a windows specific problem. The reason for the difference is the [`path.relative`](https://nodejs.org/api/path.html#pathrelativefrom-to) call.

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -1,13 +1,15 @@
 /* eslint-disable eslint-comments/disable-enable-pair, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/restrict-template-expressions */
 import { enginesVersion } from '@prisma/engines'
 import { getSchemaPathAndPrint } from '@prisma/migrate'
-import type { Command, Generator } from '@prisma/sdk'
 import {
   arg,
+  Command,
   format,
+  Generator,
   getCommandWithExecutor,
   getGenerators,
   getGeneratorSuccessMessage,
+  getPlatform,
   HelpError,
   highlightTS,
   isError,
@@ -16,10 +18,12 @@ import {
   logger,
   missingGeneratorMessage,
   parseEnvValue,
+  Platform,
 } from '@prisma/sdk'
 import chalk from 'chalk'
 import fs from 'fs'
 import logUpdate from 'log-update'
+import os from 'os'
 import path from 'path'
 import resolvePkg from 'resolve-pkg'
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by janpio** [comment]
We probably already have tests that generate on the different platforms, but we probably normalized the paths aways as they are different on the platforms because of the difference between `/` and `\` - which means that your change will not become visible in any of the snapshots.

A way to test that then would be to have different snapshots per platform which are not normalized. I do not know though if we already have that anywhere or if that would be a completely new concept in our tests.

File: packages/cli/src/Generate.ts [hunk 1]
```diff
@@ -200,9 +203,13 @@ Please run \`prisma generate\` manually.`
       )
       let hint = ''
       if (prismaClientJSGenerator) {
+        const platform = await getPlatform()
```
aqrln, line 206: I think it probably makes more sense as a local variable inside `replacePathSeperatorsIfNecessary` function, not as a parameter.

Also, you don't really need `getPlatform` from `@prisma/sdk` with all of its heavy machinery for your simple use case, just checking if `os.platform() === 'win32'` should be fine, but no harm in using `getPlatform` either of course.

flatplate:
I looked around a little to understand how the testing on different platforms and the normalization work, but unfortunately couldn't find how I could add a non-normalized test.

If you give me a few pointers on adding such tests, I could try that out but it sounds like a bigger decision than I can make 😄

janpio:
You are right, we might not actually have that right now 😞 Hopefully @aqrln or @millsp can point you in the right direction though.

Right now you CI runs are failing because of a dependency update. Rebasing on `main` (or merging from there) should fix it - GH UI should show a nice "Update branch" button now I think that should take care of that.

aqrln:
> A way to test that then would be to have different snapshots per platform which are not normalized. I do not know though if we already have that anywhere or if that would be a completely new concept in our tests.

I'm afraid there's no easy way to do this right now, yes. This might require some big and non-trivial changes in our testing setup.

aqrln (approved):

Jolg42:
@aqrln I think someone should use the tests I provided in https://github.com/prisma/prisma/pull/13153 and include them here, so we can see the change

aqrln:
@Jolg42 I removed the conditionals because the import path must be the same on all platforms, PTAL

Jolg42 (approved):
Looks good! 💯

PR #13051, opened 2022-04-28
fix(migrate): print correct provider name when creating CockroachDB db

aqrln:
I noticed an unused variable in the list of ESLint warnings (`databaseProvider`), and it looked suspicious that the unused variable was somewhere in the middle of the logic.

As far as I understand, it was meant to make sure that we still print "CockroachDB" when the provider is `cockroachdb` but the connection string starts with `postgres://` (?)
Other than this case, `dbType` and `databaseProvider` are equivalent.


Jolg42 (approved):
Indeed, nice catch!

We are planning to add more tests for CockroachDB this cycle, this line is currently untested in a cockroachdb provider context.

aqrln:
CI failures look unrelated but interesting:

image

PR #13064, opened 2022-04-28
chore(client): documented postinstall script during deep-dive

jkomyno:
Documented the `client/script/postinstall.js` file during the TS deep dive.

### DIFF
--- a/packages/client/scripts/postinstall.js
+++ b/packages/client/scripts/postinstall.js
@@ -49,19 +49,34 @@ function findPackageRoot(startPath, limit = 10) {
   return null
 }
 
+/**
+ * The `postinstall` hook of client sets up the ground and env vars for the `prisma generate` command,
+ * and runs it, showing a warning if the schema is not found.
+ * - initializes the ./node_modules/.prisma/client folder with the default index(-browser).js/index.d.ts,
+ *   which define a `PrismaClient` class stub that throws an error if instantiated before the `prisma generate`
+ *   command is successfully executed.
+ * - sets the path of the root of the project (TODO: to verify) to the `process.env.PRISMA_GENERATE_IN_POSTINSTALL`
+ *   variable, or `'true'` if the project root cannot be found.
+ * - runs `prisma generate`, passing through additional information about the command that triggered the generation,
+ *   which is useful for telemetry. It tries to use the local `prisma` package if it is installed, otherwise it
+ *   falls back to the global `prisma` package. If neither options are available, it warns the user to install `prisma` first.
+ */
 async function main() {
   if (process.env.INIT_CWD) {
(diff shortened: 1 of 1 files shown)


millsp (approved):

packages/client/scripts/postinstall.js
```diff
@@ -49,20 +49,35 @@ function findPackageRoot(startPath, limit = 10) {
...
   await ensureEmptyDotPrisma()
 
+  // TODO: consider using the `which` package
   const localPath = getLocalPackagePath()
+
   // Only execute if !localpath
   const installedGlobally = localPath ? undefined : await isInstalledGlobally()
 
-  // this is needed, so that the Generate command does not fail in postinstall
-  process.env.PRISMA_GENERATE_IN_POSTINSTALL = 'true'
```
janpio, line 64: Why can we remove this?

jkomyno: Because it's overridden immediately after. If you expand the file, you should see

```ts
  // this is needed, so that the Generate command does not fail in postinstall
  process.env.PRISMA_GENERATE_IN_POSTINSTALL = 'true'  //  <-- this is what we removed

  // this is needed, so we can find the correct schemas in yarn workspace projects
  const root = findPackageRoot(localPath) //  <-- this doesn't read/write process.env

  process.env.PRISMA_GENERATE_IN_POSTINSTALL = root ? root : 'true'
```

jkomyno: Imho, since `process.env.PRISMA_GENERATE_IN_POSTINSTALL` is written only once (and later read by the `prisma generate` command), it doesn't make much sense to talk about "flow" at all in this case. It's just an atomic declaration

packages/client/scripts/postinstall.js
```diff
@@ -49,20 +49,35 @@ function findPackageRoot(startPath, limit = 10) {
...
+/**
+ * The `postinstall` hook of client sets up the ground and env vars for the `prisma generate` command,
+ * and runs it, showing a warning if the schema is not found.
+ * - initializes the ./node_modules/.prisma/client folder with the default index(-browser).js/index.d.ts,
+ *   which define a `PrismaClient` class stub that throws an error if instantiated before the `prisma generate`
+ *   command is successfully executed.
+ * - sets the path of the root of the project (TODO: to verify) to the `process.env.PRISMA_GENERATE_IN_POSTINSTALL`
+ *   variable, or `'true'` if the project root cannot be found.
+ * - runs `prisma generate`, passing through additional information about the command that triggered the generation,
+ *   which is useful for telemetry. It tries to use the local `prisma` package if it is installed, otherwise it
```
Jolg42, line 61:
```suggestion
 *   which is useful for debugging/telemetry. It tries to use the local `prisma` package if it is installed, otherwise it
```

Jolg42 (approved):

PR #13066, opened 2022-04-28
fix(client): integrate queryRaw improvements

### LINKED ISSUE #12800: RQ: Integrate new QE result for query raw in the client

### LINKED ISSUE #12802: RQ: Make Client send those new type-hinted query parameters
More info here https://github.com/prisma/prisma/issues/12795 (ask @Weakky if more info is needed)

### LINKED ISSUE #13085: queryRaw parameter serializer corrupts user data in nested objects
Currently the raw query parameter serializer in the client unexpectedly replaces the values recursively and leaks the internal representation of values that is only used for top-level values when complex objects are passed to JSON or string fields.

For example, given this schema:

```prisma
model Data {
  id   String
  json Json
}
```

the following query:

```ts
  const value = { a: new Date() }
  await prisma.$queryRaw(Prisma.sql`INSERT INTO "Data" (id, json) VALUES (${value}, ${value})`)
```

will insert this into the database:

```ts
  {
    id: '{"a":{"prisma__type":"date","prisma__value":"2022-04-27T14:24:36.677Z"}}',
    json: {
      a: {
        prisma__type: 'date',
        prisma__value: '2022-04-27T14:24:36.677Z'
      }
    }
  }
```

which is not the expected behavior.

Internal discussion: https://prisma-company.slack.com/archives/C02FZENHD4N/p1651070700445159

This is being fixed as a part of https://github.com/prisma/prisma/issues/12802, this issue is just for reference / for posterity.

aqrln:
Engines PR: https://github.com/prisma/prisma-engines/pull/2847

Closes https://github.com/prisma/prisma/issues/12802
Closes https://github.com/prisma/prisma/issues/12800
Fixes https://github.com/prisma/prisma/issues/13085


packages/client/src/runtime/utils/deserializeRawResults.ts
```diff
@@ -0,0 +1,58 @@
...
+      return BigInt(value as string)
+
+    case 'bytes':
+      return Buffer.from(value as string, 'base64')
+
+    case 'decimal':
+      return new Decimal(value as string)
+
+    case 'datetime':
+      return new Date(value as string)
```
Weakky, line 50: AFAIK, `date` and `time` should also be coerced to JS' `Date`

aqrln: Which format would the value be in (especially in the case of `time`)?

aqrln:
Tests will be failing for now, waiting for the engines version `1c2129f3eaa9bf16c6cd7c14b69b7eb9cc88bbed` to be published.

packages/client/tests/functional/queryRaw-send-type-hints/tests.ts
```diff
@@ -0,0 +1,74 @@
...
+
+// @ts-ignore
+declare let prisma: import('@prisma/client').PrismaClient
+// @ts-ignore
+declare let Prisma: typeof import('@prisma/client').Prisma
+
+setupTestSuiteMatrix((suiteConfig) => {
+  test('Buffer ($queryRaw)', async () => {
+    if (suiteConfig['provider'] === 'mysql') {
+      await prisma.$queryRaw`INSERT INTO \`Entry\` (\`id\`, \`binary\`) VALUES (1, ${Buffer.from('hello')})`
```
danstarns, line 11: **question**: Are you specifically testing the escaped quotes here? I noticed that this one is using: 

```
(\`id\`, \`binary\`)
```

vs the one below: 

```
("id", "binary")
```

aqrln: MySQL and PostgreSQL have different syntax, so I have to choose between backticks and double quotes depending on the provider.

helpers/blaze/map.ts
```diff
@@ -24,7 +24,7 @@ function mapObject<O extends O.Object, I, R>(object: O & O.Record<A.Key, I>, map
 
   const keys = Object.keys(object) as A.Keys<O>[]
   for (let i = 0; i < keys.length; ++i) {
-    mapped[i] = mapper(object[keys[i]], keys[i])
+    mapped[keys[i]] = mapper(object[keys[i]], keys[i])
```
janpio, line 27: unrelated change?

packages/client/tests/functional/jest.config.js
```diff
@@ -25,4 +25,22 @@ module.exports = {
...
   testTimeout: 10000,
   collectCoverage: process.env.CI ? true : false,
+
+  /**
+   * Limit concurrency to 1 worker for now for two reasons.
+   *
+   * First reason:
+   *
+   * Database isolation in setupTestSuiteDbURI() doesn't seem to work properly.
+   * Tests are using to the same database concurrently, which causes constraint
```
janpio, line 35:
```suggestion
   * Tests are using the same database concurrently, which causes constraint
```

packages/client/tests/functional/jest.config.js
```diff
@@ -25,4 +25,22 @@ module.exports = {
...
+   * violation errors and "database does not exist" errors.
+   * TODO: fix this.
```
janpio, line 37: Do we have an issue about this? Looks pretty important to get fixed properly.

packages/client/tests/functional/jest.config.js
```diff
@@ -25,4 +25,22 @@ module.exports = {
...
+   *
+   * Second reason:
+   *
+   * Jest crashes when an assertion fails with an object that contains a BigInt
+   * if more than one worker is used.
+   * See https://github.com/facebook/jest/issues/11617
```
janpio, line 43: Let's +1 this and also leave a comment that we are also affected to raise awareness a bit.

helpers/blaze/map.ts
aqrln: yeah, this function was used in the PR, but not anymore, so this change is out of scope

packages/client/tests/functional/jest.config.js
millsp, line 37: Seems like this is a high priority. Thanks for reporting!

packages/client/tests/functional/jest.config.js
millsp, line 43: Once the db isolation is fixed, then we can limit this config to only a few test suites.

packages/client/src/runtime/utils/serializeRawParameters.ts
```diff
@@ -1,72 +1,113 @@
...
-    for (tmp = new Array(k); k--; ) {
-      tmp[k] = serializeBigInt(data[k])
-    }
-    return tmp
+  if (typeof obj.toJSON === 'function') {
+    return obj.toJSON()
+  }
+
+  // TODO: map from blaze would've been convenient here to map arrays and objects uniformly,
+  // but importing it in this file causes compilation errors in tsd tests.
```
millsp, line 92: Yes, we need to make it a dependency/package at some point.

millsp (commented):
Really great work 👏 The code is really nice and simple and I especially love the care you took to write unit tests as well. Thank you!

packages/client/tests/functional/jest.config.js
aqrln: I created an issue for this: https://github.com/prisma/prisma/issues/13186

packages/client/tests/functional/jest.config.js
aqrln: Yeah, one of the comments in the issue basically suggests to do the same: https://github.com/facebook/jest/issues/11617#issuecomment-1068732414

millsp (approved):

packages/client/src/runtime/utils/serializeRawParameters.ts
mrazauskas: @aqrln Seems like all works with `tsd` after the latest upgrade. At least I can't reproduce the error locally. All worked after bringing back the import how it was done by you in one of commits.

I wasn’t brave enough to open a PR, because my relationship with Docker is bad. I went through setup steps in Contributing Guide few times, but few images do not start so I can’t see if all tests are passing.

In case if the error persist, I would be curious to see how it looks like. Interesting case (;

aqrln: @mrazauskas thanks, that's good to know! You can still open a PR if you'd like to — the tests will run on CI anyway, so it's not critical to run them locally.

As for images that did not start, I assume you might have had a problem with SQL Server if you're on ARM64?

PR #13079, opened 2022-04-29
ci(update-engines-version): Output Prisma-Engines commit hash with link

janpio:
This PR adds an output of the Prisma-Engines commit hash with a link to `prisma-engines` so that it becomes clickable to easier find out which Engines commits (and which change) triggered the PR.

### DIFF
--- a/.github/workflows/update-engines-version.yml
+++ b/.github/workflows/update-engines-version.yml
@@ -43,11 +43,12 @@ jobs:
             echo 'Awesome - proceeding to make the PR'
 
       - name: Extract Engine Commit hash from version
-        uses: bhowell2/github-substring-action@v1
         id: extract-engine-commit-hash
+        uses: actions/github-script@v6
         with:
-          value: "${{ github.event.inputs.version }}"
-          length_from_end: 40
+          script: |
+            const hash = context.payload.inputs.version.split('.').pop()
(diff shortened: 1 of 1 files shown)


.github/workflows/update-engines-version.yml
```diff
@@ -42,6 +42,14 @@ jobs:
             pnpm update -r @prisma/engines@${{ github.event.inputs.version }} @prisma/engines-version@${{ github.event.inputs.version }} @prisma/fetch-engine@${{ github.event.inputs.version }} @prisma/get-platform@${{ github.event.inputs.version }}
             echo 'Awesome - proceeding to make the PR'
 
+      - name: Extract Engine Commit hash from version
+        id: extract-engine-commit-hash
+        uses: actions/github-script@v6
```
Jolg42, line 47: Someone found the all you can script GitHub action 😉

Jolg42 (approved):

.github/workflows/update-engines-version.yml
janpio: Indeed - isn't it amazing!? Much better than doing some weird bash thingie.

Jolg42: YEss!

PR #13087, opened 2022-04-29
ci: Remove unused GH Actions workflows and move scripts into subfolder

janpio:
- Remove 3 GH Actions workflows we never use/look at
- Move scripts to subfolder to separate from actual workflows


.github/workflows/codeql-analysis.yml
```diff
@@ -1,63 +0,0 @@
-name: 'CodeQL: Code Scanning'
```
Jolg42, line 1: We should keep it, it runs on schedule and doesn't disrupt anything

Jolg42 (changes requested):
✨ Would be good to keep `codeql-analysis.yml`

.github/workflows/codeql-analysis.yml
janpio: Recreated, although I am not aware of us ever getting any benefit from it?

Jolg42: It output results here https://github.com/prisma/prisma/security/code-scanning?query=is%3Aopen+branch%3Amain+tool%3ACodeQL

So far nothing that needs to be taken care of, but one day it might find something important

Jolg42 (approved):

PR #13104, opened 2022-04-30
fix: correct "setup" to "set up" when used as a verb

DePasqualeOrg:
"setup" is a noun, and "set up" is a verb. Previously "setup" was used incorrectly as a verb.


aqrln (approved):
Thank you!

janpio:
That rebase action does not seem to work with forks :/ (And generally did not achieve what I was hoping last Friday...)

The "Update branch" button on GH UI might do what it should though @aqrln

aqrln:
Added @zachtil to reviewers as per the conversation with @tanberry and him

PR #13118, opened 2022-05-02
fix(engine-core): tests for terminals with links support

SevInf:
On terminals that support hyperlinks, combination of `terminal-link`,
`stripAscii` and github issue template produces corrupt output, which
does not match saved snaphsot. Since we use URL as a title for link
anyway, the easiest way to fix it and make test pass independent of the
terminal it runs in is simply removing the link ascii escapes and output
link as a plain text - it will still be clickable in most terminal
emulators.

### DIFF
--- a/packages/engine-core/package.json
+++ b/packages/engine-core/package.json
@@ -46,7 +46,6 @@
     "new-github-issue-url": "0.2.1",
     "p-retry": "4.6.2",
     "strip-ansi": "6.0.1",
-    "terminal-link": "2.1.1",
     "undici": "5.0.0"
   },
   "files": [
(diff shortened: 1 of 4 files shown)


aqrln (approved):

SevInf:
Some more details on what it fixes: depending on whether or not terminal, in which tests run supports hyperlinks `terminal-link` outputs one or another escape sequence. For non-interactive terminals, like CI and those that do not support links tests works normally. For those that do, it is broken. Simplified example:

```js
const url = 'https://example.com/+foo'
console.log(stripAnsi(termlinalLink(url, url)))
// outputs ttps://example.com/+foohttps://example.com/+foo,
// where expected output is https://example.com/+foo
```

I assume that `+` sign in url in combination with hyperlink escape sequence causes `stripAnsi` regex to misfire and remove incorrect parts of the string. We could've tried to fix the regex, but in that case the simpliest solution is simply removing `terminal-link`: we output full URL as a title of the link anyway and that will already be clickable in most terminals. Error output will now look identical on all terminals - simply underlined URL and the test will pass independent of the terminal it runs in.

Jolg42 (approved):
✨ Happy 1st PR 🎊

PR #13119, opened 2022-05-02
chore: add docker to codespaces

Jolg42:
Tip from https://github.com/github/feedback/discussions/14542#discussioncomment-2625978


.devcontainer/Dockerfile
```diff
@@ -13,4 +11,4 @@ FROM mcr.microsoft.com/vscode/devcontainers/typescript-node:0-${VARIANT}
 # RUN su node -c "source /usr/local/share/nvm/nvm.sh && nvm install ${EXTRA_NODE_VERSION}"
 
 # To install more global node packages
-RUN su node -c "npm install -g pnpm ts-node"
+RUN su node -c "npm install -g pnpm@6 ts-node"
```
SevInf, line 14: Should we also pin ts-node version, just in case?

Jolg42: We didn't encounter issues with ts-node major so far, maybe we wait and see if it's needed

SevInf (dismissed):

Issue #13092, opened 2022-04-29
Slow Tests


danstarns:
I'm submitting this issue because I believe there to be a problem with the test suite that is really hindering the contributor developer experience, plus there could be a more serious underlying problem that could be causing this. 

I don't believe the specs on my machine to be amazing but nor do I believe them to be low end and I have run test suites, including integrations, of similar size before with no problem. 

I'm using the provided `docker-compose` file in the `docker` dir. - This is using about 2GB of ram. 

When running the tests in the `client` dir for example: 

```
$ cd ./packages/client && pnpm run test
```

I first get a bunch of successful tests and then a warning stating that something has not been awaited: 

```
 PASS  src/__tests__/integration/errors/referentialActions-onDelete-default-foreign-key-error-postgresql/test.ts
 PASS  src/__tests__/integration/happy/client-engine/test.ts (26.568 s)
 PASS  src/__tests__/integration/happy/blog-env-cockroachdb/test.ts
 PASS  src/__tests__/integration/errors/can-not-connect-to-database/test.ts (10.924 s)
 PASS  src/__tests__/integration/errors/int-errors/test.ts
 PASS  src/__tests__/integration/happy/blog/test.ts
 PASS  src/__tests__/integration/errors/referentialActions-onDelete-default-foreign-key-error-sqlite/test.ts
 PASS  src/__tests__/integration/happy/blog-dot-env-both/test.ts
 PASS  src/__tests__/integration/errors/corruption/test.ts
 PASS  src/__tests__/integration/happy/blog-dot-env-root/test.ts
 PASS  src/__tests__/integration/errors/missing-engine-native-binaryTarget/library.test.ts
 PASS  src/__tests__/integration/errors/executeRaw-alter-postgres/test.ts
 PASS  src/__tests__/integration/happy/blog-dot-env-both-conflict/test.ts
 PASS  src/__tests__/integration/happy/blog-dot-env-prisma/test.ts
 PASS  src/__tests__/integration/happy/binary/test.ts
 PASS  src/__tests__/integration/happy/createMany-postgresql/test.ts
 PASS  src/__tests__/integration/happy/custom-engine-binary/test.ts
 PASS  src/__tests__/integration/errors/wrong-native-types-postgres/test.ts
 PASS  src/__tests__/integration/errors/missing-engine/library.test.ts
 PASS  src/__tests__/integration/happy/createMany-mysql/test.ts
 PASS  src/__tests__/integration/happy/browser/test.ts
[...]

PR #13149, opened 2022-05-04
test(client): add integration runner

danstarns:
* Includes and closes: https://github.com/prisma/prisma/pull/13169

## Why

1. https://github.com/prisma/prisma/issues/13092
2. I was finding that the integration tests were running all at once and it was hard to tell jest to treat the integration differently without losing potential on the simple tests.

## What

This PR excludes all of the tests in `./packages/client/tests/integration` from the main test suite and introduces:

`integration-runner.test.ts`

**This new file will** `readDir` both the `happy/` and `errors/` tests and then perform sequential batching of n amount of tests.

 Between each batch, a backoff is added so that we don't completely overwhelm other components such as the database.

## How can I run them?

You **should not need to change anything in your workflow**, however, if you especially want to run the integration test's on their own you can run: 

```
cd ./packages/client && pnpm run test integration
```

## Findings

This shows my recorded stats of the time taken to run the client test suite before, with maxWorkers, and then after this addition. 

#### `cd ./packages/client && pnpm run test`

**Could not run** - Timed out!

```
Time:        560.228 s
Ran all test suites.
 ELIFECYCLE  Test failed. See above for more details.
```

#### `cd ./packages/client && pnpm run test integration`

**Could not run** - Timed out!

```
Time:        297.905 s
Ran all test suites matching /integration/i.
 ELIFECYCLE  Test failed. See above for more details.
```

### Using maxWorkers

* https://github.com/prisma/prisma/pull/13149#pullrequestreview-962094944

These run! However, the issue here is now that **the entire test suite is running at 50% capacity** when in reality we only want to limit the heavy load stuff(the integration tests). 

##### `cd ./packages/client && npx jest --maxWorkers="50%"`

**Pass**

```
Time:        523.493 s, estimated 538 s
Ran all test suites.
```

##### `cd ./packages/client && npx jest integration --maxWorkers="50%"`

**Pass**

```
Time:        330.482 s
Ran all test suites matching /integration/i
```

#### `cd ./packages/client && pnpm run test`

**Pass** - 222.898 seconds faster

```
Time:        300.595 s
Ran all test suites.
```

#### `cd ./packages/client && pnpm run test integration`

**Pass** - 83.634 seconds faster

```
integration/integration-runner
    √ happy (243661 ms)
    √ errors

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        246.848 s
Ran all test suites matching /integration/i.
```

## Regressions

Some things we miss out on in this implementation, this is because we are now spawning Jest using `exec`. This means that the main Jest process doesn't have a full collective view of how many tests are run for example or the coverage of those files tested on inside the spawned process. 

1. Jest Stats - You can still see all in the console but it won't be in one nice neat place like normal.
2. Code Coverage? - IDK yet standby


packages/client/src/__tests__/integration/integration-runner.test.ts
```diff
@@ -0,0 +1,48 @@
+import * as child_process from 'child_process'
+import * as fs from 'fs'
+import * as path from 'path'
+import * as util from 'util'
+
+const exec = util.promisify(child_process.exec)
+
+const MAX = 2
+const BACKOFF_MS = 100
+const excluded = ['exhaustive-schema', 'exhaustive-schema-mongo', 'client-engine', 'ignore']
```
danstarns, line 10: These ones were taking a very long time doing it this way and so I have skipped them for now until I can investigate more.

packages/client/src/__tests__/integration/integration-runner.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+async function spawnJest(options: { dir: string }): Promise<child_process.PromiseWithChild<unknown>> {
+  console.log(`Running: 'pnpm run test ${options.dir}'`)
+  const child = await exec(`pnpm run test ${options.dir}`, { cwd: process.cwd() })
```
danstarns, line 14: Errors will propagate

aqrln (commented):
Have you tried limiting `maxWorkers` in `jest.config.js`? This is not the same thing as what you're doing in this PR (which is also an interesting idea), but it might've solved the issue.

danstarns:
Included and closes: https://github.com/prisma/prisma/pull/13169

danstarns:
> Findings

Yes, and this does help. I have added a section in the PR description **Findings** that shares my stats using **maxWorkers**.

danstarns:
> Have you tried limiting `maxWorkers` in `jest.config.js`? This is not the same thing as what you're doing in this PR (which is also an interesting idea), but it might've solved the issue.

Yes, and this does help. I have added a section in the PR description **Findings** that shares my stats using **maxWorkers**.

packages/client/src/__tests__/integration/integration-runner.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  name: string
+  stderr?: string
+  stdout?: string
+}
+
+async function spawnJest(options: { dir: string; testType: TestType }): Promise<Log> {
+  const name = `${options.testType}/${options.dir}`
+
+  // Errors inside this child will propagate up
+  const child = await exec(`jest  ${name} --config=${__dirname.replace(/\\\\/g, '/')}/jest.config.js --runInBand`, {
```
danstarns, line 25: If this process errors the entire suite will fail fast, we could catch the error and propagate it like with the logs below. WDYT?

packages/client/src/__tests__/integration/integration-runner.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  beforeAll(() => {
+    timeout = setInterval(() => {
+      console.log(`integration/integration-runner still alive: ${logs.length} completed`)
+    }, 5000)
+  })
+
+  afterAll(() => {
+    clearTimeout(timeout)
+
+    const log = logs
```
danstarns, line 77: Logging all at once so that it eases the load on your terminal and that they are all displayed one after the other. We might want to log errors here too...

PR #13159, opened 2022-05-04
ci: Removing /rebase GH Actions

janpio:
as they caused a _lot_ of GH Actions workflow runs that do nothing in 99.99% of cases - but they still consume runners and actually wait quite long: https://github.com/prisma/prisma/actions/workflows/rebase-slash-command.yml
image


Jolg42:
Yes, things that need to wait on busy events, like comments need to be done in a GitHub app, but we already have one (Milestones)

PR #13169, opened 2022-05-04
test/fix(client): await blocking IO

danstarns:
Awaits blocking IO in client/integration tests: 

Related: #13092


packages/client/src/__tests__/integration/happy/client-engine/test.ts
```diff
@@ -54,51 +54,58 @@ function getExpectedEngine(engineType, envVar, envVarValue) {
...
-                name: Alice,
-              },
-            ]
-            `)
-
-          expect(prisma._clientEngineType).toMatch(expectedClientEngine)
-          await prisma.$disconnect()
-        })
+
+  test('engines', async () => {
```
aqrln, line 64: Is there a reason you merged this into a single test?

packages/client/src/__tests__/integration/happy/client-engine/test.ts
```diff
@@ -54,51 +54,58 @@ function getExpectedEngine(engineType, envVar, envVarValue) {
...
+          promises.push(prom)
+        }
       }
     }
-  }
-}
-describe('client engine', () => {
-  buildTests()
+
+    await Promise.all(promises)
```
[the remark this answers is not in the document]
danstarns: Good observation!

janpio:
Potentially stupid question: What is the difference of these to how it was run before?

packages/client/src/__tests__/integration/happy/transaction/test.ts
```diff
@@ -18,8 +18,8 @@ function clean(array: any[]) {
   })
 }
 
-beforeAll(() => {
-  fs.copyFileSync(path.join(__dirname, 'dev.db'), path.join(__dirname, 'dev2.db'))
+beforeAll(async () => {
+  await fs.promises.copyFile(path.join(__dirname, 'dev.db'), path.join(__dirname, 'dev2.db'))
```
aqrln, line 22: Probably doesn't matter in this case. Technically it'll even be kinda slower because of the overhead of creating a new task on the thread pool and then just waiting for it without doing anything else on the main thread (since there shouldn't be anything else running concurrently with the `beforeAll` hook in one Jest worker), but the difference should be negligible.

aqrln:
> it doesn't run tests concurrently in a single worker

Or, rather, it shouldn't. If it somehow does on @danstarns's machine, it may be the reason behind https://github.com/prisma/prisma/issues/13092

Issue #8031, opened 2021-07-02
Test snapshot files have very very long indentation


janpio:
https://github.com/prisma/prisma/blob/30773b7ca34f104b2dc50581cf5e1046cbd27bf1/src/packages/migrate/src/__tests__/DbSeed.test.ts#L78-L79
https://github.com/prisma/prisma/blob/30773b7ca34f104b2dc50581cf5e1046cbd27bf1/src/packages/migrate/src/__tests__/DbSeed.test.ts#L95-L96

Something is wrong there.

Jolg42:
We could check the source code of Jest to get an idea of what's happening.

It's nothing blocking though. A bit annoying to have to scroll.

janpio:
This is also happening in other snapshots: https://github.com/prisma/prisma/pull/13190/files/6b0a6925ef9dfd79467f7446c2a550f81b1de6bf#r865887332

Jolg42:
I tried to find out more about jest inline snapshots diff being always updated while I needed to work on `db pull` tests.

Current behavior: inline snapshots get indented to the right unexpectedly when running with `-u`

After testing, it looks like it only happens when we use `packages/get-platform/src/test-utils/jestSnapshotSerializer.js`
When running jest with `-u` the expected value by Jest and returned value from the serializer don't match, which triggers an update.
If nothing changed then the indentation stays the same.
But if a snapshot was added / modified all the rest gets indented.

Note: this only affects multi-line text inline snapshots

One thing we can do for now for limiting this issue is splitting the massive test file in smaller ones: https://github.com/prisma/prisma/pull/19205
And second, only use `jestSnapshotSerializer` when needed with https://jestjs.io/docs/expect#expectaddsnapshotserializerserializer
```
import serializer from 'my-serializer-module';
expect.addSnapshotSerializer(serializer);
```

Jolg42:
Note: this a regex that I used on VS Code at some point to replace all multi-line snapshots `(toMatchInlineSnapshot\(`)\n([\s\S\n]+?)(`\))$` and replace with `toMatchInlineSnapshot()` so I could try recreating them all from scratch.

jkomyno:
Hi there,
As part of our effort to clean up our GitHub repo, we’re closing this issue to help focus on features and improvements that align with our focus areas and those of our community. 

This issue was originally created for internal tracking purposes, but if you feel this is important to you, feel free to reopen the issue and get like-minded users to upvote as we will prioritize fixing issues based on community interest.

[...]

PR #13190, opened 2022-05-05
fix(client): use correct $-names in error messages

### LINKED ISSUE #9388: Missing `$` from `executeRaw` in error message "Invalid \`prisma.executeRaw()\` invocation:"
Take this example from https://github.com/prisma/prisma/blob/2b6319588bd384a1ca37d0022879ca589d974bd9/packages/client/src/__tests__/integration/happy/interactive-transactions/test.ts#L325

```ts
      await prisma.$executeRaw`INSERT INTO User (id, email) VALUES ("1", "user_1@website.com")`,
      await prisma.$executeRaw`INSERT INTO User (id, email) VALUES ("1", "user_1@website.com")`,
```

It will throw:

```
Invalid \`prisma.executeRaw()\` invocation:


  Raw query failed. Code: \`2067\`. Message: \`UNIQUE constraint failed: User.email\`
`)
```

The $ is required now in front of `executeRaw` (we used to support without the $ at some point)

Here the small fix would be to add the $ like `prisma.$executeRaw()` in the error message.

This probably also applies to `queryRaw`.

SevInf:
In error messages, `$queryRaw` and `$executeRaw` methods used to be
reported without leading $

Fixes #9388

### DIFF
--- a/packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
+++ b/packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
@@ -192,15 +192,15 @@ describe('interactive transactions', () => {
 
     await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
 
-                                    Invalid \`prisma.user.create()\` invocation in
-                                    /client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts:0:0
+                                                Invalid \`prisma.user.create()\` invocation in
+                                                /client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts:0:0
 
-                                      183   },
-                                      184 })
-                                      185 
-                                    → 186 await prisma.user.create(
-                                      Unique constraint failed on the fields: (\`email\`)
-                              `)
+                                                  183   },
+                                                  184 })
+                                                  185 
+                                                → 186 await prisma.user.create(
+                                                  Unique constraint failed on the fields: (\`email\`)
+                                        `)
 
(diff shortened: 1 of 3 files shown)


packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
```diff
@@ -192,15 +192,15 @@ describe('interactive transactions', () => {
 
     await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
 
-                                    Invalid \`prisma.user.create()\` invocation in
-                                    /client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts:0:0
+                                                Invalid \`prisma.user.create()\` invocation in
```
SevInf, line 195: I am not quite sure why those snapshots were updated, tbh. Should it be like that?

janpio: Long standing problem with the snapshots generation: https://github.com/prisma/prisma/issues/8031

We should probably rename and expand that.

janpio (approved):

millsp (approved):

aqrln (approved):

packages/client/src/__tests__/integration/happy/interactive-transactions-postgres/test.ts
aqrln: It's a known issue, yes. You may revert unrelated changes if you want to minimize churn, or you may leave it as is, either way is fine.

PR #13191, opened 2022-05-05
refactor: remove dev.db

danstarns:
Removes the `dev.db` file from source control.

Related: https://github.com/prisma/prisma/pull/13029


aqrln:
> maybe we should not add it to gitignore ?

At least this will prevent us from adding new ones I guess 😄

PR #13198, opened 2022-05-05
test(client): port count test to functional

danstarns:
Replaces the existing integration count test with the new functional style.


packages/client/src/__tests__/integration/happy/count/test.ts
```diff
@@ -2,7 +2,8 @@ import { getTestClient } from '../../../../utils/getTestClient'
 
 let PrismaClient, prisma
 
-describe('count', () => {
+// TODO - Delete me. Replaced in https://github.com/prisma/prisma/pull/13198
+describe.skip('count', () => {
```
aqrln, line 6: Should we maybe remove it in this PR?

packages/client/tests/functional/count/prisma/_schema.ts
```diff
@@ -0,0 +1,32 @@
+export default ({ provider, id, foreignKeyId }) => {
```
SevInf, line 1: You can also `testMatrix.setupSchema` here to get typechecking for parameters

packages/client/tests/functional/count/_matrix.ts
```diff
@@ -0,0 +1,30 @@
+import { defineMatrix } from '../_utils/defineMatrix'
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'sqlite',
+      id: 'Int @id @default(autoincrement())',
+      foreignKeyId: 'Int?',
+      providerFeatures: '',
```
aqrln, line 9: I don't think you use `providerFeatures` in the schema, should we remove them from the matrix?

SevInf (approved):

packages/client/tests/functional/count/tests.ts
```diff
@@ -35,7 +30,7 @@ describe('count', () => {
     const value = await prisma.user.count({
       select: true,
       where: {
-        age: 84,
+        age: 111,
```
millsp, line 33: That's old 🤣

danstarns: 🕸

PR #13215, opened 2022-05-06
fix(client): Correctly serialize Decimal.js in an input

### LINKED ISSUE #13213: Exception when using Decimal.js instance in an input
### Bug description

After investigating #6021, I found out that Prisma crashes if we try to use `Prisma.Decimal` as an input to query:

### How to reproduce

Schema:

```prisma
generator client {
  previewFeatures = []
}

datasource db {
  provider = "sqlite"
  url      = "file:./dev.db"
}

model User {
  id      String   @id @default(uuid())
  decimal Decimal?
}

```

Code:

```ts
await prisma.user.findFirst({ where: { decimal: new Prisma.Decimal(123) } })
```

Crashes with: 

```
/Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:44755
  for (const inputType of arg2.inputTypes) {
                               ^
TypeError: arg2.inputTypes is not iterable
    at valueToArg (/Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:44755:32)
    at /Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:45010:18
    at Array.reduce (<anonymous>)
    at objectToArgs (/Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:44990:28)
    at tryInferArgs (/Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:44889:40)
    at valueToArg (/Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:44756:16)
    at /Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:45010:18
    at Array.reduce (<anonymous>)
    at objectToArgs (/Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:44990:28)
    at tryInferArgs (/Users/serhii/projects/prisma/reproductions/6021-decimal/node_modules/.prisma/client/runtime/index.js:44889:40) {
  clientVersion: '0.0.0'
}
```

### Expected behavior

No exception

### Environment & setup

- OS: MacOS
- Database: any
- Node.js version: 16.15.0

### Prisma Version

```
3.13.0
```

SevInf:
In `isObject` code we have a list of classes which we treat as primitives. There were a couple of issues there:
- `Decimal.js` string was incorrect. I assume it changed after the update and then we just never caught it
- `bigint` string had a typo in it, but more importantly, it is redundant since `typeof` check already takes care of it and it is not possible to construct `BigInt` wrapper object.
- `isObject(null)` returned `null` instead of boolean.

All of these things are fixed and the tests for them added in this PR.
For `query` tests, i choose to move them to their own directory and split by topic. I think it might make sense to re-organize existing `makeDocument` tests in the same way at some point, but let me know if you'd prefer to have them in `select.test.ts` instead.

Fix #13213

### DIFF
--- a/packages/client/src/__tests__/isObject.test.ts
+++ b/packages/client/src/__tests__/isObject.test.ts
@@ -1,0 +1,32 @@
+import Decimal from 'decimal.js'
+
+import { isObject } from '../runtime/utils/isObject'
+
+test('is true for proper object', () => {
+  expect(isObject({})).toBe(true)
+})
+
+test('is false for null', () => {
+  expect(isObject(null)).toBe(false)
+})
+
+test('is false for primitive types', () => {
+  expect(isObject(1)).toBe(false)
+  expect(isObject(undefined)).toBe(false)
+  expect(isObject('hello')).toBe(false)
+  expect(isObject(true)).toBe(false)
+  expect(isObject(BigInt('10'))).toBe(false)
+  expect(isObject(Symbol())).toBe(false)
+})
+
+test('is false for Dates', () => {
+  expect(isObject(new Date('2022-05-06T00:00:00Z'))).toBe(false)
+})
+
+test('is false for Decimal.js instance', () => {
+  expect(isObject(new Decimal('12.34'))).toBe(false)
+})
+
+test('is false for Buffer', () => {
+  expect(isObject(Buffer.from('hello', 'utf8'))).toBe(false)
+})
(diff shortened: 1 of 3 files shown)


aqrln (approved):
LGTM!

millsp:
It would be great to add a functional test :)

millsp:
Awesome, thank you :100:

SevInf:
Tests are added in #13253

PR #13224, opened 2022-05-06
ci: try running integration-tests package tests all at once

Jolg42:
To save GitHub runners, since the slowest part is actually the setup

 integration-tests (library, 12)
succeeded 1 minute ago in 7m 53s 
https://github.com/prisma/prisma/runs/6325770187?check_suite_focus=true

 integration-tests (binary, 12)
succeeded 1 minute ago in 10m 5s 
https://github.com/prisma/prisma/runs/6325770247?check_suite_focus=true

Where before it was taking 6 min just for one run (5min on setup) like
https://github.com/prisma/prisma/runs/6324538464?check_suite_focus=true

So it's going to help with out GitHub Actions queuing!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -266,11 +266,6 @@ jobs:
           SKIP_GIT: true
           GITHUB_CONTEXT: ${{ toJson(github) }}
 
-      # TODO: is this still necessary?
-      # - run: pnpm i sqlite3@5.0.2 --unsafe-perm --reporter=silent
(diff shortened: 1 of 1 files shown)


janpio:
Before: 
image

janpio (approved):

janpio:
The one info that is going away is that we currently have no integration tests for MongoDB (Which the comment indicated at least some way)

millsp (approved):

PR #13253, opened 2022-05-09
fix(client): Allow to use non-bundled Decimal.js versions

### LINKED ISSUE #6021: Prisma throws type mismatch error using Decimal type from from 'decimal.js' package
## Bug description

Prisma generates a namespace for Decimal.js that does not match the type from the import of Decimal.js from npm. This makes it annoying to use when third party code gives you "decimal.js Decimal's" and not "prisma Decimal's".

```
Type 'Decimal[]' is not assignable to type 'string | number | Decimal | Decimal[] | TestCreatefooInput | number[] | string[]'
```

## How to reproduce

```

model Test {
  id  Int       @id @default(autoincrement())
  foo Decimal[]
}
```

```
iimport { PrismaClient } from '@prisma/client'

import {Decimal as PrismaDecimal} from '@prisma/client/runtime'
import {Decimal as DecimalActual} from 'Decimal.js'

const actualDecimals = [new DecimalActual(1), new DecimalActual(2), new DecimalActual(3)]
const prismaDecimals = [new PrismaDecimal(1), new PrismaDecimal(2), new PrismaDecimal(3)]

prisma.test.create({
	data: {
		foo: actualDecimals
	  // ^ Type 'Decimal[]' is not assignable to type 'string | number | Decimal | Decimal[] | TestCreatefooInput | number[] | string[]'
	}
})

prisma.test.create({
	data: {
		foo: prismaDecimals
	}
})

```

## Expected behavior

Doesn't complain about mismatched types

## Prisma information

{
  "@prisma/client": "^2.18.0",
  "decimal.js": "^10.2.1"
}

## Environment & setup

```
prisma               : 2.18.0
@prisma/client       : 2.18.0
Current platform     : darwin
Query Engine         : query-engine da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at ../../.config/yarn/global/node_modules/@prisma/engines/query-engine-darwin)
Migration Engine     : migration-engine-cli da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at ../../.config/yarn/global/node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine : introspection-core da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at ../../.config/yarn/global/node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary        : prisma-fmt da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at ../../.config/yarn/global/node_modules/@prisma/engines/prisma-fmt-darwin)
Studio               : 0.356.0
```

SevInf:
Since we bundle Decimal.js types, if the user has their own copy of the
library, our type will be incompatible with their. Introduce
`DecimalJSLike` interface. which includes public properties of decimal
instances and no methods and use this interface as input type. This
allows us to accept following things:

1. `Prisma.Decimal`
2. User-provided decimal instance, even if it is does not match
   our bundled version of the library.
3. Some compatible libraries, like decimal.js-light (not tested yet)
4. Manuallly constructed object with matching properties

Because of the last item, in some cases we have to do serialization
manually. `Decimal.isDecimal` returns true for items 1-3, but in case it
returns `false` we manually construct the instance from provided
properties.

Also adds a couple of functional tests for decimal.js handling.

Fix #6021

### DIFF
--- a/packages/client/src/runtime/query.ts
+++ b/packages/client/src/runtime/query.ts
@@ -25,6 +25,7 @@ import {
   unionBy,
   wrapWithList,
 } from './utils/common'
+import { isDecimalJsLike, stringifyDecimalJsLike } from './utils/decimalJsLike'
 import { deepExtend } from './utils/deep-extend'
 import { deepGet } from './utils/deep-set'
 import { filterObject } from './utils/filterObject'
@@ -635,8 +636,8 @@ function stringify(value: any, inputType?: DMMF.SchemaArgInputType) {
     return 'null'
   }
 
-  if (Decimal.isDecimal(value)) {
-    return value.toString()
+  if (Decimal.isDecimal(value) || (inputType?.type === 'Decimal' && isDecimalJsLike(value))) {
+    return stringifyDecimalJsLike(value)
   }
 
   if (inputType?.location === 'enumTypes' && typeof value === 'string') {
--- a/packages/client/src/runtime/utils/decimalJsLike.ts
+++ b/packages/client/src/runtime/utils/decimalJsLike.ts
@@ -1,0 +1,36 @@
+import { Decimal } from 'decimal.js'
+/**
+ * Interface for any Decimal.js-like library
+ * Allows us to accept Decimal.js from different
+ * versions and some compatible alternatives
+ */
+export interface DecimalJsLike {
+  d: number[]
+  e: number
+  s: number
+}
+
+export function isDecimalJsLike(value: unknown): value is DecimalJsLike {
+  if (Decimal.isDecimal(value)) {
+    return true
+  }
+  return (
+    value !== null &&
+    typeof value === 'object' &&
+    typeof (value as any).s === 'number' &&
+    typeof (value as any).e === 'number' &&
+    Array.isArray((value as any).d)
+  )
+}
+
+export function stringifyDecimalJsLike(value: DecimalJsLike): string {
+  if (Decimal.isDecimal(value)) {
+    return String(value)
+  }
+
+  const tmpDecimal = new Decimal(0) as DecimalJsLike
+  tmpDecimal.d = value.d
+  tmpDecimal.e = value.e
+  tmpDecimal.s = value.s
+  return String(tmpDecimal)
+}
--- a/packages/client/tests/functional/_utils/idForProvider.ts
+++ b/packages/client/tests/functional/_utils/idForProvider.ts
@@ -1,0 +1,6 @@
+const sqlId = 'Int @id @default(autoincrement())'
+const mongoDbId = 'String @id @default(auto()) @map("_id") @db.ObjectId'
+
+export function idForProvider(provider: string): string {
+  return provider === 'mongodb' ? mongoDbId : sqlId
+}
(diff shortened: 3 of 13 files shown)


packages/client/tests/functional/_utils/idForProvider.ts
```diff
@@ -0,0 +1,6 @@
+const sqlId = 'Int @id @default(autoincrement())'
+const mongoDbId = 'String @id @default(auto()) @map("_id") @db.ObjectId'
+
+export function idForProvider(provider: string): string {
```
SevInf, line 4: Thought this might be a useful helper for the cases, where we don't care about specifics of the id. Which, I assume will be most of the tests. What do you think?

packages/client/tests/functional/_utils/idForProvider.ts
```diff
@@ -0,0 +1,6 @@
+const sqlId = 'Int @id @default(autoincrement())'
```
aqrln, line 1: Maybe 

```suggestion
const sqlId = 'String @id @default(uuid())'
```

so that the type will be the same regardless of the provider

packages/client/src/runtime/query.ts
```diff
@@ -2,6 +2,7 @@ import chalk from 'chalk'
 import Decimal from 'decimal.js'
 import indent from 'indent-string'
 import stripAnsi from 'strip-ansi'
+import { P } from 'ts-toolbelt/out/Object/_api'
```
aqrln, line 5: Doesn't seem to be used

packages/client/src/runtime/utils/decimalJsLike.ts
```diff
@@ -0,0 +1,36 @@
...
+export function stringifyDecimalJsLike(value: DecimalJsLike): string {
+  if (Decimal.isDecimal(value)) {
+    return String(value)
+  }
+
+  const tmpDecimal = new Decimal(0) as DecimalJsLike
+  tmpDecimal.d = value.d
+  tmpDecimal.e = value.e
+  tmpDecimal.s = value.s
+  return String(tmpDecimal)
```
aqrln, line 35: clever trick!

aqrln (approved):
LGTM with one small comment (https://github.com/prisma/prisma/pull/13253#discussion_r869087100).

SevInf:
Thanks @aqrln!
I fixed unused import. I still have no merge rights, so I'd be grateful if somebody could merge it after CI passes.

millsp (commented):
Nice work 👏

jessmorecroft:
Hi, I was going to try to get this working with decimal.js-light but can't see how it would. Those members in the DecimalJsLike interface simply don't exist in the light Decimal.

See here: https://github.com/MikeMcl/decimal.js-light/blob/master/decimal.d.ts

Hopefully I'm missing something?

SimonCockx:
Not solved in `3.15.2`. I get a type error when doing this:
```ts
new PrismaClient().test.findUnique({ where: { id: 0 } }).then((t) => {
    const foo: Decimal = t!.foo;
    //    ^^^
});
```
`Type 'import("[...]/node_modules/@prisma/client/runtime/index").Decimal' is not assignable to type 'import("[...]/node_modules/decimal.js/decimal").default'.
  Types have separate declarations of a private property 'toStringTag'.`

given a model
```prisma
model Test {
  id  Int       @id @default(autoincrement())
  foo Decimal
}
```

Jolg42:
@SimonCockx Hi! Could you [create a separate bug issue here](https://github.com/prisma/prisma/issues/new?assignees=&labels=kind%2Fbug&template=bug_report.yml)? We would need to reproduce this first.

PR #13261, opened 2022-05-10
fix(client): inline schema

### LINKED ISSUE #13052: Inline/bundle the contents of the prisma schema on generate
We could close a whole class of issues where the schema is not found (https://github.com/prisma/prisma/labels/topic%3A%20ENOENT%20schema.prisma) by inlining the contents of the `schema.prisma` directly into the generated `index.js`. Our logic happens to break usually when code is bundled (nextjs, svelte, webpack).

When code is bundled to another location, our lookup paths for the schema do not make sense any longer. We have logic to at best find the original generated folder, which works at best but has not guarantee it could find the correct folder and fails especially on monorepos.

A solution to this is to directly embed the prisma schema, which needs to be passed to the engines. Side note: We might experience the same thing again for sqlite databases since we resolve their normal location relatively to the generated client.

millsp:


trentonsnyder:
Sorry I don't know the best way to contribute to the PR but I'm very excited about it!

I ran some of the failing pipeline tests locally and was able to fix some by changing a line in `binaryEngine.test.ts`
current
```
datamodelPath: path.join(__dirname, './runtime-tests/blog/schema.prisma'),
```
to
```
datamodelPath: '',
```

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -573,10 +573,11 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
         const additionalFlag = this.allowTriggerPanic ? ['--debug'] : []
 
         const flags = [
+          '--datamodel',
+          this.inlineSchema,
```
aqrln, line 577: This may fail with [`E2BIG`](https://man7.org/linux/man-pages/man3/errno.3.html) with large schemas, writing to a temporary file might be a better option (at least if the schema size is more than a few kilobytes).

Jolg42: Indeed

Issue #11674, opened 2022-02-07
Can't save string to Decimals list


ekfn:
### Bug description

According to the types, I can create a record by specifying a list of strings for a field of type `Decimal[]`

```js
prisma.decimalsItem.create({
  data: {
    decimals: ['0.25', '0.15']
  },
})
```

but it throws a `PrismaClientValidationError`

```
Argument decimals: Got invalid value 
[
  '0.25',
  '0.15'
]
on prisma.createOneDecimalsItem. Provided List<String>, expected DecimalsItemInput or Decimal:
type DecimalsItemInput {
  set: List<Decimal>
}
type DecimalsItemInput {
  set: List<Decimal>
}
```

Although I can convert the strings to Prisma.Decimal as a workaround

```js
prisma.decimalsItem.create({
  data: {
    decimals: ['0.25', '0.15'].map(n => new Prisma.Decimal(n))
  },
})
```

I don't want to do that because default precision of `decimals.js` is only 20 significant digits.

### How to reproduce

1. Run 

```
prisma.decimalsItem.create({
  data: {
    decimals: ['0.25', '0.15']
  },
})
```
2. See error

### Expected behavior

No validation error

### Prisma information

```
model DecimalItem {
  id              Int       @id @default(autoincrement())
  decimals        Decimal[] @db.Decimal(16, 8)
}
```

### Environment & setup

- OS: Mac OS, Windows, Debian, CentOS
- Database: PostgreSQL
- Node.js version: v16.13.0

### Prisma Version

```
3.9.1
```

PR #13272, opened 2022-05-10
refactor(client): Refactor scalar arguments validation

SevInf:
After few dives into `query.ts` I believe we could handle scalar arguments validation a little bit better. 

What we are doing right now: for every passed in value we try to guess it's type, based on typeof, schema type, matching certain regexps, etc. From that we build a stringified name of the type and compare with expected name, built from the schema. In a long list of conditionals then we try to see if the types are compatible (you can assign ints to float types, but not vice-versa) as well as correct some incorrect "guesses" from previous step (if a string looks like a date we'd report that it's a date, but then allow to pass dates to string types if we guessed incorrectly).

This has a couple of issues:
1. Since we compare types as strings, lists have to be processed separately from scalars. If we say that `Foo` is compatible with `Bar`, we need to also not forget that `List<Foo>` is compatible with `List<Bar>`, it won't happen automatically. It causes issues like #11674 
2. Need to correct incorrect guesses, like in String-Date example from above.

What I am proposing instead in this PR:

1. Given DMMF schema for types, we build a series of classes for each supported scalar type, as well as the list. 
2. Each class is responsible for checking, if provided value is compatible. We don't try to guess the type from the value. If user passes us a string that looks like a `DateTime`, we won't say it is a `DateTime` unless it's used in a position where `DateTime` is an acceptable input.
3. List type reuses the classes of scalar types, so if we implement changes in particular scalar type logic, lists of those scalars will handle this case as well.

I believe this a clearer and less susceptible to errors approach.  It does not seem to break any existing tests. Plus, comment above `hasCorrectScalarType` said to refactor it :)

This PR is for gathering initial feedback on this from people who know the codebase better than I do: do you think it's viable approach? Does it actually improve things? Am I missing some edge cases?

If we actually decide to ship it, it will need much more tests, including tests for `queryArgsTypecheck` module itself, hence the draft. I also did not remove `getGraphQLType` "guessing" function, since we seem to use it for error messages and it still might be useful.

### DIFF
--- a/packages/client/src/runtime/queryArgsTypecheck.ts
+++ b/packages/client/src/runtime/queryArgsTypecheck.ts
@@ -64,7 +64,7 @@ class DateTimeType implements Type {
       return false
     }
 
-    return RFC_3339_REGEX.test(value)
+    return RFC_3339_REGEX.test(value) && String(new Date(value)) !== 'Invalid Date'
   }
 }
 
--- a/packages/client/src/runtime/queryArgsTypecheck.ts
+++ b/packages/client/src/runtime/queryArgsTypecheck.ts
@@ -1,0 +1,177 @@
+import Decimal from 'decimal.js'
+
+import type { DMMF } from './dmmf-types'
+import { isDecimalJsLike } from './utils/decimalJsLike'
+
+const RFC_3339_REGEX =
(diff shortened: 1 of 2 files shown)


packages/client/src/runtime/queryArgsTypecheck.ts
```diff
@@ -0,0 +1,177 @@
...
+    return new ListType(type)
+  }
+
+  return type
+}
+
+function typeByName(name: string): Type {
+  const TypeConstructor = dmmfTypesMapping[name]
+  if (!TypeConstructor) {
+    throw new TypeError(`Unknown scalar type: ${name}`)
```
SevInf, line 174: Question: do we have a specific error type for the cases of "This should never have happened in real life, but it did and here we are"?

SevInf:
JSON protocol supersedes this

PR #13273, opened 2022-05-10
fix(client): Accept strings inside of decimal lists

### LINKED ISSUE #11674: Can't save string to Decimals list

SevInf:
I believe #13272 would've been a better way to solve it, but if we want quicker and less risky fix, then this PR will work too.

Fix #11674

### DIFF
--- a/packages/client/src/__tests__/query/decimal.test.ts
+++ b/packages/client/src/__tests__/query/decimal.test.ts
@@ -38,6 +38,49 @@ test('allows to pass it decimal instance', () => {
       }
     }
   `)
+  expect(() => document.validate()).not.toThrow()
+})
+
+test('allows to pass it a string', () => {
+  const document = makeDocument({
+    dmmf,
+    rootTypeName: 'query',
+    rootField: 'findManyUser',
+    select: { where: { money: '123456789.12334' } },
+  })
+
+  expect(document.toString()).toMatchInlineSnapshot(`
+    query {
+      findManyUser(where: {
+        money: "123456789.12334"
+      }) {
+        id
+        money
+      }
+    }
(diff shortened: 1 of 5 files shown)


aqrln (approved):
LGTM, but might need rebasing now

PR #13277, opened 2022-05-10
test(client): Case for internal engine port

### LINKED ISSUE #12507: Add test for internal engine port
Note: This is a great starter issue for someone onboarding to learn how we manage the binary engines. 

For internal development and testing we are able to do this:

```
  // @ts-ignore
  __internal: {
    engine: {
      endpoint: "http://127.0.0.1:4466",
    },
  },
});
```

This then gets the Prisma Client to connect to an already running engine binary. The work done at https://github.com/prisma/prisma/pull/12505 fixes this because it stopped working as we had on tests to monitor this functionality. It would be nice to add a test for this so this does not break again.

danstarns:
Specifies that the PrismaClient should use the binary engine and then spawns another PrismaClient utilizing/talking to that already running engine. 

Closes: #12507


packages/client/tests/functional/reuse-binary-engine/tests.ts
```diff
@@ -0,0 +1,70 @@
...
+        },
+      ],
+    })
+
+    let prismaClient2 = undefined
+    let internalURL = ''
+
+    prismaClient1.$on('info', (data) => {
+      if (/Started query engine/.test(data.message)) {
+        const words = data.message.split(' ')
```
garrensmith, line 23: I think this could be simplified into:
```
let port = data.message.split(':').pop() // pops off the last one
let URL = `http://127.0.0.1:${port}`
```
```

File: packages/client/tests/functional/reuse-binary-engine/tests.ts [hunk 2]
```diff
@@ -0,0 +1,70 @@
...
+      }
+    })
+
+    await prismaClient1.$connect()
+
+    await (() => {
+      return new Promise((resolve) => {
+        const interval = setInterval(async () => {
+          if (!internalURL) {
+            await util.promisify(setTimeout)(0)
```
garrensmith, line 38: What does this do?
Shouldn't this just return here if there is no internalUrl?
```
if (!internalURL) return;
```

garrensmith (changes requested):
Nice work on this. I've got two comments. The one might just be my lack of understanding of the test setup

packages/client/tests/functional/reuse-binary-engine/tests.ts
SevInf: I'd actually suggest `URL` class:

```new URL(url).port`

File: packages/client/tests/functional/reuse-binary-engine/tests.ts [hunk 3]
```diff
@@ -0,0 +1,70 @@
+import * as util from 'util'
+
+import { setupTestSuiteMatrix } from '../_utils/setupTestSuiteMatrix'
+
+// https://github.com/prisma/prisma/issues/12507
+setupTestSuiteMatrix(() => {
+  test('should create data using one PrismaClient and read using another', async () => {
+    // @ts-ignore
+    const prismaClient1 = new PrismaClient({
```
aqrln, line 9: I think the type of `prismaClient1` will be `any` even in the generated tests if you do it like this (correct me if I'm wrong @millsp).

A better way would be to add

```ts
// @ts-ignore
declare let PrismaClient: typeof import('@prisma/client').PrismaClient
```

near the top of the file.

---

@millsp is there a way to avoid having to add these declarations manually in all tests? Maybe with a `.d.ts` file?

packages/client/tests/functional/reuse-binary-engine/tests.ts
```diff
@@ -0,0 +1,70 @@
...
+            },
+          })
+
+          clearInterval(interval)
+
+          // @ts-ignore
+          resolve()
+        }, 100)
+      })
+    })()
```
aqrln, line 56: This logic looks overcomplicated to me, I'd prefer something simpler, maybe something along the lines of

```suggestion
    const timeout = util.promisify(setTimeout)

    while (!internalURL) {
      await timeout(100)
    }

    // @ts-ignore
    prismaClient2 = new PrismaClient({
      __internal: {
        engine: {
          endpoint: internalURL,
        },
      },
    })
```

danstarns:
Thanks, @garrensmith & @aqrln for pointing out my complexity with the promise, it was there to wait for the `prismaClient1.$on('info', (data) => {` event to fire so that we could grab the port. 

What I have done is wrap the `info` event in a promise so it's clearer on what we are awaiting plus no need for the timeout.

packages/client/tests/functional/reuse-binary-engine/tests.ts
```diff
@@ -14,15 +14,18 @@ setupTestSuiteMatrix(() => {
...
-            const port = data.message.split(':').pop() // pops off the last one
+            const port = data.message.split(':').pop()
             resolve(`http://127.0.0.1:${port}`)
           }
         })
 
-        prismaClient1.$connect()
+        prismaClient1
+          .$connect()
+          .then(() => {})
```
aqrln, line 27:
```suggestion
```

packages/client/tests/functional/reuse-binary-engine/tests.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const created = await prismaClient1.user.create({ data: { name } })
+    const [found] = await prismaClient2.user.findMany({ where: { name } })
+
+    expect(created).toMatchObject(found)
+
+    await prismaClient1.$disconnect()
+    await prismaClient2.$disconnect()
+
+    jest.clearAllTimers()
```
aqrln, line 53: Is this still necessary?

packages/client/tests/functional/reuse-binary-engine/tests.ts
```diff
@@ -0,0 +1,55 @@
+import { getClientEngineType, ClientEngineType } from '@prisma/sdk'
+
+import testMatrix from './_matrix'
+
+// https://github.com/prisma/prisma/issues/12507
+testMatrix.setupTestSuite(() => {
+  test('should create data using one PrismaClient and read using another', async () => {
+    if (getClientEngineType() === ClientEngineType.Library) {
```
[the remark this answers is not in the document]
aqrln: (`ClientEngineType.DataProxy`)

packages/client/tests/functional/reuse-binary-engine/tests.ts
```diff
@@ -0,0 +1,54 @@
...
+    const name = 'some-random-name'
+
+    const created = await prismaClient1.user.create({ data: { name } })
+    const [found] = await prismaClient2.user.findMany({ where: { name } })
+
+    expect(created).toMatchObject(found)
+
+    await prismaClient1.$disconnect()
+    await prismaClient2.$disconnect()
+
```
aqrln, line 52:
```suggestion
```

packages/client/tests/functional/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   globalSetup: './_utils/globalSetup.js',
   snapshotSerializers: ['@prisma/sdk/src/utils/jestSnapshotSerializer'],
   setupFilesAfterEnv: ['./_utils/setupFilesAfterEnv.ts'],
-  testTimeout: 10000,
+  testTimeout: 60000,
```
aqrln, line 26: Is this an intentional change?

packages/client/tests/functional/reuse-binary-engine/_matrix.ts
```diff
@@ -0,0 +1,12 @@
+import { defineMatrix } from '../_utils/defineMatrix'
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'sqlite',
+      id: 'Int @id @default(autoincrement())',
+      providerFeatures: '',
```
aqrln, line 8:
```suggestion
```

aqrln (approved):

garrensmith (approved):
From my side this is lookig good. Nice one. I think @aqrln has some things that still need to be addressed then I imagine you can merge. Thanks for doing this

packages/client/tests/functional/reuse-binary-engine/tests.ts
danstarns: Not sure how to type this one, I tried this however I get: 

```
Message:
      Argument of type '{ __internal: { engine: { endpoint: string; }; }; }' is not assignable to parameter of type 'Subset<PrismaClientOptions, PrismaClientOptions>'.
      Object literal may only specify known properties, and '__internal' does not exist in type 'Subset<PrismaClientOptions, PrismaClientOptions>'.
    at (tests/functional/reuse-binary-engine/.generated/reuse-binary-engine (sqlite) ()/tests.ts:39:7)

      47 |     for (const checkPath of keys(suiteChecks)) {
      48 |       if (checkPath.includes(path.dirname(suiteFilePath))) {
    > 49 |         assert.fail(suiteChecks[checkPath])
         |                ^
      50 |       }
      51 |     }
```

packages/client/tests/functional/jest.config.js
danstarns: No my bad!

packages/client/tests/functional/reuse-binary-engine/tests.ts
millsp: @danstarns It does not exist on the types, it's on purpose. The best is to annotate with `// @ts-ignore`

millsp: @aqrln I think that could be a good idea to explore along with making the types available while writing the test

millsp: @danstarns `// @ts-ignore` should be applied on `__internal` and removed from the client instantiation and add
```ts
// @ts-ignore
declare let PrismaClient: typeof import('@prisma/client').PrismaClient
```

aqrln (approved):

PR #13296, opened 2022-05-11
fix(engine-core): Use commit hash as an engine version

### LINKED ISSUE #10125: "Query Engine" misreported as `0.1.0` in basic error reports from Prisma Client
Recently the basic error reports we get via the generated GitHub link when something goes wrong in Prisma Client started misreporting the "Query Engine" version as `0.1.0`:

https://github.com/prisma/prisma/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22topic%3A+basic+error+report%22+

https://github.com/prisma/prisma/issues/9933
https://github.com/prisma/prisma/issues/9896
https://github.com/prisma/prisma/issues/9875

That is unfortunate, as this information of course sometimes could be helpful to debug problems.

SevInf:
Actual package version is always 0.1.0, so not very useful.

Fix #10125


danstarns (commented):
Is there a way this could be tested so the issue doesn't get reintroduced again?

SevInf:
@danstarns I'd like to do it, but i don't see good way, unfortunately. There are number of problems with testing this:

- To really be safe against re-introduction we'd have to run against real engine, mock probably won't be helpful, since person re-introducing the bug will just update mock value. 
- This code path is executed only in case of rust panic, which ideally should never happen. So, if we are running against real engine, we'll have to crash it somehow during the test. I suppose engine people could've added a special method which will panic every time it's called specially for testing this kinds of things, but:
- Hash is compiled into the dynamic library on rust side, not read from any external sources. Since we have to be running against real engine, hash will always update with the new engine version. So, we'll have to resort to mocking again

aqrln (approved):

aqrln:
> Hash is compiled into the dynamic library on rust side, not read from any external sources. Since we have to be running against real engine, hash will always update with the new engine version. So, we'll have to resort to mocking again

Alternatively, it should be possible to check the properties of this string instead of comparing it with a fixed value, i.e., check that it looks like a hash using a regex.

janpio:
You could also compare the string returned by the engine with what we expect the Engine version to be via `@prisma/engine-version`.

PR #13303, opened 2022-05-11
chore(deps): update devdependencies (non-major)

millsp:
Takes over the blocked renovate PR. The following tasks are all related and needed:
* Fixes internal types for sql template tags
* Fixes and simplifies overall esbuild build logic
* Tree-shaking via TypeScript paths instead of esm (hack)
  * Tree-shaking via esm to cjs does not work anymore
  * https://github.com/evanw/esbuild/releases/tag/v0.14.4
* Removes all commonjs exports, needed for tree-shaking
* Simplifies our logic for `debug`, extends the original one


packages/sdk/src/sendPanic.ts
```diff
@@ -79,6 +79,7 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
       dbVersion: dbVersion,
     }
 
+    // TODO: What is the exports doing here? A jest thing?
     const signedUrl = await exports.createErrorReport(params)
```
millsp, line 83: I wanted to remove this `exports`, but turns out it is needed for the tests?

millsp: cc @jkomyno

Jolg42:
Looks like it's failing in ecosystem-tests using `3.15.0-integration-renovate.3`

In `dataproxy (cloudflare-workers, dataproxy, ubuntu-latest)`
https://github.com/prisma/ecosystem-tests/runs/6396873000?check_suite_focus=true#step:6:163
```
👀  ./node_modules/@prisma/client/runtime/proxy.js
Module not found: Error: Can't resolve 'child_process' in '/home/runner/work/ecosystem-tests/ecosystem-tests/dataproxy/cloudflare-workers/node_modules/@prisma/client/runtime'
resolve 'child_process' in '/home/runner/work/ecosystem-tests/ecosystem-tests/dataproxy/cloudflare-workers/node_modules/@prisma/client/runtime'
  Parsed request is a module
  using description file: /home/runner/work/ecosystem-tests/ecosystem-tests/dataproxy/cloudflare-workers/node_modules/@prisma/client/package.json (relative path: ./runtime)
    Field 'browser' doesn't contain a valid alias configuration
    resolve as module
      /home/runner/work/ecosystem-tests/ecosystem-tests/dataproxy/cloudflare-workers/node_modules/@prisma/client/runtime/node_modules doesn't exist or is not a directory
```

and
`dataproxy (vercel-edge-functions, dataproxy, ubuntu-latest)`
https://vercel.com/prisma-orm/e2e-vercel-edge-functions/8Z9AbstMm5vjJtmJLxKP5yi9osfH
Screen Shot 2022-05-12 at 08 37 27

Jolg42:
Link to `3.15.0-integration-renovate.4` which just started ⏩ 
https://github.com/prisma/ecosystem-tests/actions/runs/2312033019

packages/cli/helpers/build.ts
```diff
@@ -11,27 +11,6 @@ import { run } from '../../../helpers/compile/run'
...
-const resolveHelperPlugin: esbuild.Plugin = {
-  name: 'resolveHelperPlugin',
-  setup(build) {
-    // for any import of @prisma/sdk, resolve to this one
-    build.onResolve({ filter: /^@prisma\/sdk$/ }, () => {
-      return { path: require.resolve('@prisma/sdk') }
-    })
-  },
-}
-
```
Jolg42, line 33: @millsp How is this not needed anymore?

packages/sdk/tsconfig.build.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.build.json",
   "compilerOptions": {
     "outDir": "dist",
-    "emitDeclarationOnly": true
+    "emitDeclarationOnly": true,
+    "paths": {}
```
Jolg42, line 6: What does that do? Maybe add a comment here, it's allowed in tsconfig

packages/debug/package.json
```diff
@@ -38,8 +38,9 @@
   ],
   "dependencies": {
     "@types/debug": "4.1.7",
+    "debug": "4.3.4",
```
Jolg42, line 41: Note, here we use the `debug` npm package now and not the harcoded version of it (which was deleted in this PR)

packages/client/helpers/build.ts
```diff
@@ -89,6 +84,11 @@ function bundleTypeDefinitions(filename: string, outfile: string) {
       ],
       compiler: {
         tsconfigFilePath: 'tsconfig.build.json',
+        overrideTsconfig: {
+          compilerOptions: {
+            paths: {}, // bug with api extract + paths
```
Jolg42, line 89: So this is a workaround? What's the bug? (is there an issue to link?)

packages/debug/src/__tests__/basic.test.ts
```diff
@@ -1,15 +1,17 @@
...
 
 describe('debug', () => {
-  test('shouldnt log if its not enabled', () => {
+  test('should not log if it is not enabled', () => {
     const debug = Debug('my-namespace')
     const logs: string[] = []
+
     debug.log = (...args) => {
-      logs.push(stripAnsi(args[0]).trimStart())
+      logs.push(stripAnsi(`${args[0]}${args[1]}`).trim())
```
Jolg42, line 12: Why do we need `args[1]` added now?

millsp: So that's the standard way for hooking in a logger. Not sure why we were doing the concatenation ourselves. That's the way the api works, now it works in a more standard way.
```ts
    const debug = Debug('my-namespace')

    debug.log = (...args) => {
      logs.push(stripAnsi(`${args[0]}${args[1]}`).trim())
    }
```

 I can make that it behaves like the previous one if needed, but pretty much internal, so I did this choice.

helpers/compile/build.ts
```diff
@@ -36,51 +34,23 @@ const DEFAULT_BUILD_OPTIONS = {
  * Apply defaults to allow us to build tree-shaken esm
  * @param options the original build options
  */
-const applyEsmDefaults = (options: BuildOptions): BuildOptions => ({
```
Jolg42, line 39: What's the short human version of this change? 😄

packages/client/helpers/build.ts
millsp: I might need to report it, basically, when paths are in use, it bundles entire classes an not just the class definitions.

millsp: That leads to an invalid `d.ts` bundle.

helpers/compile/build.ts
millsp: Basically, we were doing a hack of compiling TypeScript-commonjs-ish code to esm so that it can get tree-shaken, then that code would be transformed back to cjs. It was a hack that worked until [esbuild 0.14.4/5](https://github.com/evanw/esbuild/releases/tag/v0.14.4). I happily got rid of the hack to use TS `paths` instead.

packages/cli/helpers/build.ts
millsp: This is now done automatically via TS `paths`. If you look at the root `tsconfig.build.json`, I added paths for the build step, that info is used by `esbuild` at compile time, so it already knows where to find the `@prisma/sdk`.

packages/sdk/tsconfig.build.json
millsp: We only want to use `paths` for the build step if we are bundling. So all non-bundled packages have an empty one. We don't want `esbuild` to build via the paths, because it is a clear and separate dependency. Additionally, for building types, it also adds friction because TS would include all the types from the paths into `dist` instead of using the ones from `node_modules`.

millsp: I think the tl;dr; is used an empty `path` is you're not bundling.

packages/debug/src/__tests__/basic.test.ts
millsp: `...args` is supposed to be the raw arguments passed to a `debug` call, not a stringified version of it (like we had).

millsp: (probably not used besides the test)

packages/cli/package.json
```diff
@@ -108,7 +108,7 @@
     "strip-ansi": "6.0.1",
     "tempy": "1.0.1",
     "ts-pattern": "4.0.2",
-    "typescript": "4.5.4"
+    "typescript": "4.6.4"
```
SevInf, line 111: Just a thought: have we considered pulling common dev dependencies, like `typescript`, `jest`, `esbuild` and the likes into workspace root so that we don't have to manage their version individually in each package?

packages/migrate/src/utils/byline.ts
```diff
@@ -23,13 +24,13 @@
...
-// convinience API
+// convenience API
 export default function byline(readStream, options?: any) {
-  return module.exports.createStream(readStream, options)
+  return createStream(readStream, options)
 }
 
 // basic API
-module.exports.createStream = function (readStream, options) {
+export function createStream(readStream, options) {
```
SevInf, line 33: Not related to dependencies update per se, just curious, if there a reason why this code is repeated several times in the different packages?

packages/sdk/tsconfig.build.json
SevInf: Do you think it would make sense to have a separate ts preset for non-bundled packages, with those 2 options and this comment, so we don't have to repeat them everywhere?

packages/migrate/src/utils/byline.ts
millsp: :shrug: Things to be cleaned and shoved into the `@prisma/sdk` :trollface: Also why I started our own utility library since many functional utils are copied across projects.

packages/sdk/tsconfig.build.json
millsp: Definitely, we could have at the root `tsconfig.build.bundle.json` and `tsconfig.build.regular.json` that is extended in the package `tsconfig.build.json`.

packages/client/src/byline.ts
```diff
@@ -26,11 +26,11 @@ import util from 'util'
 
 // convenience API
 export default function byline(readStream, options?: any) {
-  return module.exports.createStream(readStream, options)
+  return createStream(readStream, options)
 }
 
 // basic API
-module.exports.createStream = function (readStream, options) {
+export function createStream(readStream, options) {
```
aqrln, line 33: Fun fact: since this is treated as an ES module by esbuild, it just ignored `module.exports` and didn't transform it in any way, so previously `createStream` and `LineStream` were defined globally on `module.exports` of the whole bundle:

```
Welcome to Node.js v18.1.0.
Type ".help" for more information.
> require('./packages/cli/build/index.js')
{
  createStream: [Function (anonymous)],
  LineStream: [Function: LineStream]
}
```

packages/debug/package.json
aqrln: nice!

packages/engine-core/src/tools/byline.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-this-alias */
```
aqrln, line 1: If the content differs, it would be nice to diff `byline.ts` files across the packages to see if they are functionally equivalent.

Hopefully the difference was only in the comments and type annotations, but if there are different versions of the code, we don't really know which one was used, because these modules weren't isolated and would override each other's `createStream` and `LineStream` on the bundle's `module.exports`.

helpers/compile/build.ts
Jolg42: Indeed! Very nice 👍🏼

packages/cli/package.json
Jolg42: Renovate does the update management and pnpm is very efficient, so maybe not a big priority here to do. 
It's also more flexible to specify in each package, if one package cannot be upgraded for some reason for example.

packages/migrate/src/utils/byline.ts
Jolg42: Tech debt topic, something to clean indeed

packages/engine-core/src/tools/byline.ts
Jolg42: They should be the same besides some comments I think

packages/client/src/byline.ts
millsp: 🤣🙈

packages/debug/src/__tests__/basic.test.ts
millsp: Are you ok with that @Jolg42?

aqrln (approved):

Jolg42 (approved):
Niiiice upgrade! 💚 

Thanks for the good work here! 🙌🏼

packages/debug/src/__tests__/basic.test.ts
Jolg42: We checked on a huddle 👍🏼

PR #13333, opened 2022-05-12
feat(client): add forceInlineSchema generator parameter

### LINKED ISSUE #13052: Inline/bundle the contents of the prisma schema on generate

itamar-getloam:
Closes #13052

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -129,7 +129,8 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "3.15.0-3.55eef363a3b2e17691b1f98ed28da8b387f3bdb8"
+    "@prisma/engines-version": "3.15.0-3.55eef363a3b2e17691b1f98ed28da8b387f3bdb8",
+    "tmp": "0.2.1"
   },
   "sideEffects": false
 }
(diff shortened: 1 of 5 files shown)


Jolg42 (changes requested):
Thanks for the PR!

I think @millsp is working on something similar at the moment, so I'm blocking the PR for now, since it's something that @millsp should check before you continue here.

millsp:
Hey @itamar-getloam, we are planning to make this a thing https://github.com/prisma/prisma/pull/13261 https://github.com/prisma/prisma/issues/13052

itamar-getloam:
Appreciate the super quick response! @Jolg42 @millsp
Your solution is definitely the better way to go, replacing all `datamodelPath` with `inlineSchema` is much cleaner than the temp file solution I implemented.

Could you give a rough estimation regarding the time it will take to merge your fixes?

Thanks!

millsp:
Thanks, we appreciate your contribution and interest in Prisma. I can't tell you when it will be merged, probably less than 2 weeks for sure. But I can give you an integration version to try it out in a few days (don't hesitate to ping me again :) )

Issue #11171, opened 2022-01-14
Can't connect to Cloud SQL Postgres using socket


Verhaeg:
### Bug description

According to some other issues, ([#1508](https://github.com/prisma/prisma/issues/1508)) and ([#8638](https://github.com/prisma/prisma/issues/8638)) there was some issue when trying to connect with sockets and it was changed at least for MySQL.

In our case we use CloudSQL Postgres and configuring Cloud Run to connect to this instance using Cloud Proxy where it creates a socket usually in `/cloudsql` folder by creating a new folder inside with the resource link in the format of `<project>:<region>:<instance>` and then a file inside `.s.PGSQL.5432`. 

We set the `DATABASE_URL` following all rules, something in the lines of:
```text
postgresql://<USER>:<PASSWORD>@localhost/<DATABASE>?host=/cloudsql/<CLOUDSQL_INSTANCE_RESOURCE_NAME>&schema=public
```

We tried several combinations but without any success and the error we found was:
```text
Can't reach database server at `/cloudsql/<CLOUDSQL_INSTANCE_RESOURCE_NAME>`:`5432`
```

Looking at the code it seems to be forcing a TLS connection instead of allowing socket like it does for MySQL.

For now the workaround we are working with is (For anyone facing this issue):
1. Enabling Private IP in our CloudSQL instances
2. Created a Serverless VPC Connector (with extra charges =/ )
3. Attaching our Cloud Run service to this VPC connector and using the private IP

### How to reproduce

Configure any socket connection for Postgres and try to connect using this socket.

### Expected behavior

Be able to connect to a Postgres instance through UNIX socket connection.

### Prisma information

prisma:client clientVersion: 2.29.1

### Environment & setup

- Database: Postgres
- Node.js version: 14.17.5

### Prisma Version

```
2.29.1
```

rained23:
having the same issue, but in our production env it is working.

Prisma version 3.6.0

yoshitamagoshi:
I was able to get it to work with the following `DATABASE_URL` format:

`postgresql://<DB_USER>:<DB_PASS>@localhost/<DB_NAME>?host=/cloudsql/<CLOUD_SQL_CONNECTION_NAME>`

Ensure that `CLOUD_SQL_CONNECTION_NAME` is of the format `project_id:region:instance_name` and not just the instance name.

[...]

PR #13347, opened 2022-05-12
fix(client): socket is missing for mariadb client

karupanerura:
I got following error:

```
% # URL is mysql://foo:bar@localhost/siremo?socket=path/to/mysql.sock
% prisma migrate dev --name init
Datasource "db": MySQL database "app" at "localhost:3306"

Error: P1001: Can't reach database server at `localhost`:`3306`

Please make sure your database server is running at `localhost`:`3306`.
```

I noticed that the socket was not passed for `mariadb` and fixed it.
Please consider to merge it.

### DIFF
--- a/packages/client/src/utils/setupMysql.ts
+++ b/packages/client/src/utils/setupMysql.ts
@@ -22,6 +22,7 @@ export async function setupMysql(options: SetupParams): Promise<void> {
     database: credentials.database,
     user: credentials.user,
     password: credentials.password,
+    socketPath: credentials.socket,
     multipleStatements: true,
   })
 


karupanerura:
Additionally, it should be noted that In Google Cloud SQL, the [Cloud SQL Auth Proxy](https://cloud.google.com/sql/docs/mysql/sql-proxy) is commonly used, and a UNIX domain socket is required to use it.
refs. https://cloud.google.com/sql/docs/mysql/connect-run

So, some Google Cloud users may have trouble with this issue. I'm too.

janpio:
The file you edited is only used in testing, and is not really part of Prisma Client or its Query Engine. I don't think this will have any effect on usage of Prisma Client at all unfortunately.

Socket usage should work with Prisma Client and MySQL.

karupanerura:
I see. Sorry, I misunderstood.

I tested again in a different way and found that the problem does not occur when connecting directly to MySQL residing on localhost without using the Cloud SQL Auth Proxy.

I will close this issue for now. Thank you very much.

karupanerura:
I think this problem is similar to #11171

PR #13452, opened 2022-05-23
feat(sdk): add CLI and binary engine versions to parsing/validation errors 

### LINKED ISSUE #11908: Schema Parsing/Validation error should include CLI version
It would be helpful if the default schema validation/parsing output would include the CLI version so users posting the error message for help automatically include the CLI version as well.

Example: https://github.com/prisma/prisma/discussions/11894#discussion-3885241

jkomyno:
Closes https://github.com/prisma/prisma/issues/11908.

Note: I currently expect some snapshot errors.

Example screenshot:

image


packages/sdk/src/engine-commands/errorHelpers.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function addVersionDetailsToErrorMessage(message: string) {
+  const rows = [
+    ['prisma', version],
+    ['Default Engines Hash', enginesVersion],
+  ]
+  return `${message}
+
+Versions:
+${formatTable(rows)}`
```
janpio, line 14:
```suggestion
  return `${message}
Prisma CLI Version: x.y.z`
```

packages/client/src/__tests__/dmmf.test.ts
```diff
@@ -347,23 +347,22 @@ describe('dmmf', () => {
...
+        16 |         AWESOME
+        17 |       }
+           | 
 
-Validation Error Count: 1`
-      if (process.env.PRISMA_CLI_QUERY_ENGINE_TYPE === 'binary') {
-        expect(stripAnsi(e.message)).toEqual(`${snapshot}\n`)
-      } else {
-        expect(stripAnsi(e.message)).toEqual(snapshot)
-      }
```
janpio, line 366: Where did this "logic" go?
(And why was it there in the first place, and maybe why is it not needed any more now?)

janpio (approved):
Generelly looks good - left a few questions about smaller details.

packages/sdk/src/__tests__/engine-commands/getDmmf.test.ts
```diff
@@ -143,28 +143,27 @@ describe('getDMMF', () => {
...
-
-Validation Error Count: 2`
-      if (process.env.PRISMA_CLI_QUERY_ENGINE_TYPE === 'binary') {
-        expect(stripAnsi(e.message)).toEqual(`${snapshot}\n`)
-      } else {
-        expect(stripAnsi(e.message)).toEqual(snapshot)
-      }
+      expect(stripAnsi(e.message)).toMatchInlineSnapshot(`
+        "Get DMMF: Schema parsing
+        error: Error parsing attribute \\"@default\\": The \`autoincrement()\` default value is used on a non-id field even though the datasource does not support this.
```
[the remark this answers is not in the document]
jkomyno: It is, that's automatically generated by jest when using `toMatchInlineSnapshot`. Previously, the binary had an additional newline character that prevented us from using such method directly, and forced us to use a conditional `toEqual`.

packages/sdk/src/engine-commands/getDmmf.ts
```diff
@@ -366,7 +367,7 @@ async function getDmmfBinary(options: GetDMMFOptions): Promise<DMMF.Document> {
           },
         ),
         E.map((errorOutputAsJSON: Record<string, string>) => {
-          const defaultMessage = `${chalk.redBright(errorOutputAsJSON.message)}\n`
```
jkomyno, line 369: > Re: Where did this "logic" go?

This was a leftover newline character from the old `getDmmf` implementation. cc @janpio

packages/client/src/__tests__/dmmf.test.ts
jkomyno: No longer needed as now the output from the binary is the same as the library (at least in terms of error messages). Previously, the binary had an additional newline

packages/sdk/src/engine-commands/getDmmf.ts
janpio: Ok, that connects those dots.

If you are confident the error message still look similar enough, that's fine.

packages/sdk/src/__tests__/engine-commands/getDmmf.test.ts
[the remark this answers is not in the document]
jkomyno: My understanding is that the binary already had an additional newline w.r.t. the library, but that wasn't captured by tests previously, as only the query-engine library was snapshotted. Joel and I knew that there was an additional newline either in the Rust codebase or in the TypeScript one. Since it wasn't a blocker, we merged it with those conditional tests. Now, as I had to slightly touch the getDmmf file again, I have spotted that additional "\n" in the binary and got rid of it.

pantharshit00 (approved):

PR #13453, opened 2022-05-23
migrate: add cockroachdb native types to introspection tests

### LINKED ISSUE #13180: Introspection: Make SQL in tests more advanced so that Introspection results will include native types, not just scalar types
via https://github.com/prisma/prisma/pull/12910#discussion_r858707840

We should adapt the SQL we use to set up the database to test `db pull` to have SQL types that do not introspect only as scalar types, but also get a native type. This will lead to different introspection results in CockroachDB and PostgreSQL at least.

Example files to adapt: 
https://github.com/prisma/prisma/pull/12910/files#diff-8d3fb75dba472fcb2d9df393d7db08f6e4c7af844e621667ab8ebfdcc9debee4
https://github.com/prisma/prisma/pull/12910/files#diff-2796ab4e0bb571192be86351872030ae80768dcf02cf6c52aad15d2f35b0a0e1

jkomyno:
Closes https://github.com/prisma/prisma/issues/13180.

For `cockroach`, I have used the native string types shown [here](https://www.cockroachlabs.com/docs/stable/string.html) and [here](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#cockroachdb).

### DIFF
--- a/packages/migrate/src/__tests__/DbPull.test.ts
+++ b/packages/migrate/src/__tests__/DbPull.test.ts
@@ -492,31 +492,40 @@ describe('postgresql', () => {
 })
 
 describeIf(!process.env.TEST_SKIP_COCKROACHDB)('cockroachdb', () => {
-  const setupParams: SetupParams = {
+  const defaultParams = {
     connectionString: process.env.TEST_COCKROACH_URI || 'postgresql://prisma@localhost:26257/tests',
-    dirname: path.join(__dirname, '..', '__tests__', 'fixtures', 'introspection', 'cockroachdb'),
+  }
+
+  async function testSetup(setupDirname = 'cockroachdb', options = { withFixture: false }) {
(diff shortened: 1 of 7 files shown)


janpio (approved):
Generally this looks great!

If you have the motivation, you could make the SQL a bit smaller to _only_ include columns that are different between PostgreSQL/CockroachDB (or that show some interesting difference in the resulting schema). That would make it easier to discover what is going on I would suspect - right now it took me a while to relaly notice the differences.

jkomyno:
> Generally this looks great!
> 
> If you have the motivation, you could make the SQL a bit smaller to _only_ include columns that are different between PostgreSQL/CockroachDB (or that show some interesting difference in the resulting schema). That would make it easier to discover what is going on I would suspect - right now it took me a while to relaly notice the differences.

You're right, let's simplify things!

Issue #12708, opened 2022-04-07
Node-API Engine is missing crash handling to create GitHub issue link


janpio:
While reproducing https://github.com/prisma/prisma/issues/12704 I noticed that the Engine panic was not handled via creation of a GitHub issue link to report it: https://github.com/prisma/prisma/issues/12704#issuecomment-1091338289

With `library` engine type:
```
PS C:\Users\Jan\Documents\throwaway\12704> node .\script.js
thread 'tokio-runtime-worker' panicked at 'not yet implemented', query-engine\connectors\mongodb-query-connector\src\logger.rs:160:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
PrismaClientUnknownRequestError: 
Invalid `prisma.user.create()` invocation in
C:\Users\Jan\Documents\throwaway\12704\script.js:6:32

  3 const prisma = new PrismaClient()
  4
  5 async function main() {
→ 6   let user = await prisma.user.create(
  not yet implemented
    at Object.request (C:\Users\Jan\Documents\throwaway\12704\node_modules\@prisma\client\runtime\index.js:45582:15)
    at async PrismaClient._request (C:\Users\Jan\Documents\throwaway\12704\node_modules\@prisma\client\runtime\index.js:46405:18)
    at async main (C:\Users\Jan\Documents\throwaway\12704\script.js:6:14) {
  clientVersion: '3.12.0'
}
```

And as it should be with `engineType = binary`:
```
PS C:\Users\Jan\Documents\throwaway\12704> node .\script.js   
PrismaClientUnknownRequestError: 
Invalid `prisma.user.create()` invocation in
C:\Users\Jan\Documents\throwaway\12704\script.js:6:32

  3 const prisma = new PrismaClient()
  4
  5 async function main() {
→ 6   let user = await prisma.user.create(
  PANIC in query-engine\connectors\mongodb-query-connector\src\logger.rs:160:31
not yet implemented

This is a non-recoverable error which probably happens when the Prisma Query Engine has a panic.

[...]

PR #13455, opened 2022-05-23
fix(engine-core): Correctly report panics within LibraryEngine

SevInf:
`LibraryEngine` did not handle panics within GraphQL requests correctly:
unlike `BinaryEngine`, it reported them as
PrismaClientUnknownRequestError and did not add github link.

Most of this commit splits library loading code from `LibraryEngine`
class so that we can test it without involving Rust module.

Ref #12708.


packages/engine-core/src/library/DefaultLibraryLoader.ts
```diff
@@ -0,0 +1,184 @@
...
+export class DefaultLibraryLoader implements LibraryLoader {
+  private config: EngineConfig
+  private libQueryEnginePath: string | null = null
+  private platform: Platform | null = null
+
+  constructor(config: EngineConfig) {
+    this.config = config
+  }
+
+  async loadLibrary(): Promise<Library> {
```
SevInf, line 25: Most of this file is moved from `LibraryEngine` class as-is, including comments and TODOs

aqrln (approved):

PR #13463, opened 2022-05-24
test/refactor(client): add missing helpers

danstarns:
Adds missing helpers in the new functional tests, this ensures we have type checking and auto complete in those files.


packages/client/tests/functional/queryRaw-send-type-hints/prisma/_schema.ts
```diff
@@ -1,18 +1,20 @@
...
+      previewFeatures = [${previewFeatures}]
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URI_${provider}")
+    }
+    
+    model Entry {
+      id     Int @id @default(autoincrement())
```
millsp, line 16: Can this benefit from `idForProvider`?

danstarns: sure!

packages/client/tests/functional/_utils/idForProvider.ts
```diff
@@ -1,6 +1,26 @@
-const sqlId = 'String @id @default(uuid())'
-const mongoDbId = 'String @id @default(auto()) @map("_id") @db.ObjectId'
+export interface IOptions {
```
aqrln, line 1: just `Options` is fine, we don't have a convention to prefix interfaces with `I`

packages/client/tests/functional/_utils/idForProvider.ts
```diff
@@ -1,6 +1,26 @@
...
+  includeDefault: boolean
+}
+
+export function idForProvider(provider: string, options: IOptions = { includeDefault: true }): string {
+  const strs: string[] = ['String @id']
```
aqrln, line 6: doesn't TypeScript infer the correct type automatically here?

packages/client/tests/functional/queryRaw-send-type-hints/tests.ts
```diff
@@ -15,7 +15,7 @@ testMatrix.setupTestSuite((suiteConfig) => {
 
     const record = await prisma.entry.findUnique({
       where: {
-        id: 1,
```
aqrln, line 18: this needs to be updated in the raw SQL query too then (here and in everywhere else in the file)

aqrln: but I'm not really sure if it's worth changing the schema in this test to use `idForProvider` in the first place since it's inherently SQL-only and it's not possible to make this test run with MongoDB

packages/client/tests/functional/queryRaw-send-type-hints/prisma/_schema.ts
aqrln: https://github.com/prisma/prisma/pull/13463#discussion_r883677786

packages/client/tests/functional/queryRaw-send-type-hints/tests.ts
danstarns: sure, I will remove it!

packages/client/tests/functional/queryRaw-send-type-hints/prisma/_schema.ts
```diff
@@ -1,18 +1,20 @@
...
+      previewFeatures = [${previewFeatures}]
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URI_${provider}")
+    }
+    
+    model Entry {
+      id Int @id @default(autoincrement())
```
aqrln, line 16:
```suggestion
      id     Int @id @default(autoincrement())
```

aqrln (approved):

PR #13465, opened 2022-05-24
fix(sdk): restore error reporting after esbuild@0.14.4 broke previous behavior

jkomyno:
I have found a regression introduced by [this PR](https://github.com/prisma/prisma/pull/13303), which installed [esbuild@0.14.4](https://github.com/evanw/esbuild/releases/tag/v0.14.4), which broke the error reporting (but not the issue reporting), The culprit of the issue is that [the workaround](https://github.com/prisma/prisma/pull/12762/files#diff-717d3468803485de569ad254a7eeffdb69aeb1804f07338554f05bbefe463a87R82) used to mock  the `createErrorReport` function no longer works.

Before this PR:

image

After this PR:

image

TL/DR: using `exports.$FUNCTION_NAME` no longer works to expose a function to `jest` mocking library.
A quick and painless way of solving this is simply extracting the involved functions to a separate module (from `sendPanic.ts` to `errorReporting.ts`).

### DIFF
--- a/packages/sdk/src/sendPanic.ts
+++ b/packages/sdk/src/sendPanic.ts
@@ -78,7 +78,6 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
       dbVersion: dbVersion,
     }
 
-    // TODO: What is the exports doing here? A jest thing?
     const signedUrl = await createErrorReport(params)
 
     if (error.schemaPath) {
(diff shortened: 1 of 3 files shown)


packages/sdk/src/sendPanic.ts
```diff
@@ -79,8 +78,7 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
       dbVersion: dbVersion,
     }
 
-    // TODO: What is the exports doing here? A jest thing?
```
janpio, line 82: Was it a jest thing in the end?

jkomyno: More accurately, it was something to expose a function to jest in order to mock it (for testing `sendPanic`)

janpio (approved):

PR #13479, opened 2022-05-25
ci: hopefully fix buildkite mariadb issue

jkomyno:


millsp (approved):

aqrln:
Thank you!!!

Do we have any idea why a newer version doesn't work or fails to start and what needs to be done to unblock using a newer version?

PR #13498, opened 2022-05-26
test(client): Functional Create Many

danstarns:
Moves old int create many tests to the new functional style. **Adds faker** for random data.


packages/client/tests/functional/create-many/_matrix.ts
```diff
@@ -0,0 +1,26 @@
...
+    {
+      provider: 'postgresql',
+      providerFeatures: '',
+    },
+    {
+      provider: 'mongodb',
+      providerFeatures: '',
+    },
+  ],
+  [
```
SevInf, line 18: Do those previews features make any difference for this particular test?

packages/client/tests/functional/create-many/_matrix.ts
```diff
@@ -0,0 +1,26 @@
+import { defineMatrix } from '../_utils/defineMatrix'
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'mysql',
+      providerFeatures: '',
```
SevInf, line 7: I think we can just drop this property from the matrix if it always stays empty

packages/client/tests/functional/create-many/tests.ts
```diff
@@ -0,0 +1,102 @@
+import * as faker from 'faker'
```
SevInf, line 1: IIRC, fresher maintained version of the package is `@faker-js/faker`

SevInf (changes requested):

packages/client/tests/functional/create-many/tests.ts
millsp: (Idea: I found https://github.com/snaplet/copycat to be more maintainable/readable, made by our friends at snaplet).

SevInf: Oh, wow, I actually worked with the guy who made it, and I know that he knows a lot about generating very complex test data deterministically. So yea, changing my suggestion to `copycat` as well.

danstarns: Ive gone with `@faker-js/faker` simply because copycat currently has 9 weekly downloads and are we sure we want to trust a lib with this little usage? It looks as thoe copycat just uses and exports faker anyway!

packages/client/tests/functional/create-many/_matrix.ts
danstarns: No, my bad a copy and paste error!

packages/client/tests/functional/create-many/prisma/_schema.ts
```diff
@@ -0,0 +1,30 @@
...
+      previewFeatures = [${previewFeatures}]
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URI_${provider}")
+    }
+    
+    model User {
+      id ${idForProvider(provider)}
```
millsp, line 17: Small nit here about the indentation.

danstarns: https://github.com/prisma/prisma/pull/13498/commits/ce59f5b9820d8a5459c90115b6c0fff978451972

SevInf (approved):

PR #13507, opened 2022-05-26
test(client): Opt Out Of Providers

danstarns:
This PR adds a helper to enforce that the functional test suite covers all providers listed in the `providers` file. You can opt-out of specific providers using `options.outOut`. This will make it easier to enforce coverage across all providers and allow us to easily see what isn't covered. 

## How to use: 
**The test suite will enforce that you cover all providers listed in the `providers` file. Given you have the following matrix and only want to test mongodb:** 

```ts
import { defineMatrix } from '../_utils/defineMatrix'

export default defineMatrix(() => [
  [
    {
      'provider': 'mongodb',
      'id': 'String @id @default(auto()) @map("_id") @db.ObjectId',
      'providerFeatures': '"mongoDb", ',
      'previewFeatures': '"filterJson"',
    },
  ]
*/
```

In your `test.ts` file you must `optOut` of other providers to only use `mongodb` like so: 

```ts
import { getTestSuiteSchema } from '../_utils/getTestSuiteInfo'

declare let prisma: import('@prisma/client').PrismaClient

testMatrix.setupTestSuite(
  (suiteConfig, suiteMeta) => {

    test('...', () => {
    })

    test('...', () => {
    })

    test('...', () => {
    })
  },
  { 
      optOut:  { // <--------------  👀
        from: ['postgresql', 'mysql', 'sqlite', 'cockroachdb'],
        reason: 'I only want to test MongoDB here'
      }
   } 
)
```

Given this use case, if you do not opt-out of the missing providers it will throw the following error: 

```
 FAIL  tests/functional/_example/tests.ts
  ● Test suite failed to run

    Test: '_example' is missing providers 'sqlite, postgresql, mysql, cockroachdb' optIn using options.optIn

      84 |
      85 |     if (missingProviders.length) {
    > 86 |       throw new Error(
         |             ^
      87 |         `Test: '${suiteMeta.testDirName}' is missing providers '${missingProviders.join(
      88 |           ', ',
      89 |         )}' optIn using options.optIn`,

      at Object.setupTestSuiteMatrix [as setupTestSuite] (_utils/setupTestSuiteMatrix.ts:86:13)
      at Object.setupTestSuite (_example/tests.ts:9:12)
      at TestScheduler.scheduleTests (../../../../node_modules/.pnpm/@jest+core@28.1.0_ts-node@10.7.0/node_modules/@jest/core/build/TestScheduler.js:317:13)
```


packages/client/tests/functional/_utils/providers.ts
```diff
@@ -0,0 +1,9 @@
+export enum Providers {
```
danstarns, line 1: Do we test SQLSERVER and MARIADB?

packages/client/tests/functional/_utils/setupFilesAfterEnv.ts
```diff
@@ -3,5 +3,6 @@ process.env.DATABASE_URI_sqlite = 'file:dev.db'
 process.env.DATABASE_URI_mongodb = process.env.TEST_MONGO_URI
 process.env.DATABASE_URI_postgresql = process.env.TEST_POSTGRES_URI
 process.env.DATABASE_URI_mysql = process.env.TEST_MYSQL_URI
+process.env.DATABASE_URI_cockroachdb = process.env.TEST_COCKROACH_URI
```
danstarns, line 6: SQLSERVER and MARIADB here?

packages/client/tests/functional/_utils/providers.ts
SevInf: Not sure about mariadb, but we have a couple of sql-server-specific tests in old setup.

SevInf:
I am wondering if it will make sense to have some light-weight to opt-in without changing the code while in development. I usually start with a single provider and then expand it to others once suite is more or less ready. True, I could just pass `optIn` manually, but then it'll be easy to just forget there after I am done. Maybe some env variable or cli flag that will skip provider check will make sense? In that case, I can still start with a test for single provider, but then my tests will fail on `CI` without explicit opt-in.

Another consideration: maybe `optOut` would be a more future-proof of similar functionality. Suppose we add new provider `SuperAwesomeDb`. Can we reasonably be sure that tests, that currently opt-in to `postgres` and `sqlite` only don't make sense for new `SuperAwesomeDb`? With explicit opt-out we'd at least get a failing test in this case, and then we can decide if the functionality is broken or the test just does not make any sense for the new provider.  With opt-in the only way to find that out would be to manually look through all of the tests that use `optIn` and check if need to add new provider there.

SevInf:
Thinking about it more, what will you think about requiring the reason for opt-out:

```js
optOut: {
  from: ['mongodb'],
  reason: 'Test makes sense only for SQL databases'
}
```

In that case:
1. We'll have a clearly specified opt-out reason for posteriority
2. It sort of takes care of my first point: if I don't have a good reason to opt out of certain providers and just doing it during development, it will stand out during the review

danstarns:
> Thinking about it more, what will you think about requiring the reason for opt-out:
> 
> ```js
> optOut: {
>   from: ['mongodb'],
>   reason: 'Test makes sense only for SQL databases'
> }
> ```
> 
> In that case:
> 
> 1. We'll have a clearly specified opt-out reason for posteriority
> 2. It sort of takes care of my first point: if I don't have a good reason to opt out of certain providers and just doing it during development, it will stand out during the review

Thanks for the feedback @SevInf I have changed it to use `optOut` over opt-in and the reason message really helps highlight why certain providers arent being tested.

SevInf (approved):

SevInf:
I like how this feature immediately exposed missing connectors in already ported tests

danstarns:
> I like how this feature immediately exposed missing connectors in already ported tests

100%

packages/client/tests/functional/decimal-list/tests.ts
```diff
@@ -3,28 +3,39 @@ import { setupTestSuiteMatrix } from '../_utils/setupTestSuiteMatrix'
...
+      })
     })
-  })
-})
+  },
+  {
+    optOut: {
+      from: ['mongodb', 'mysql', 'sqlite'],
+      reason: `
+        Mongodb connector does not support the Decimal type.
```
millsp, line 36: Nice idea, love it.

millsp (commented):
Nice work and idea, thank you :100:

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -35,11 +43,38 @@ export type TestSuiteMeta = ReturnType<typeof getTestSuiteMeta>
...
+) {
   const originalEnv = process.env
   const suiteMeta = getTestSuiteMeta()
-  const suiteTable = getTestSuiteTable(suiteMeta)
+  const suiteConfig = getTestSuiteConfigs(suiteMeta)
+  const suiteTable = getTestSuiteTable(suiteMeta, suiteConfig)
   const forceInlineSnapshot = process.argv.includes('-u')
+  const suiteConfigProviders = suiteConfig.map(({ provider }) => provider)
+
+  const missingProviders = Object.values(Providers)
```
millsp, line 57: I'd like to extract this into a utility function instead (+use reduce instead of map+filter)

```ts
/**
 * Ensure that we are not forgetting to add a provider to the matrix unless we
 * explicitly opt out during in the {@link setupTestSuiteMatrix} creation.
 * @param suiteConfigs
 * @param suiteMeta
 * @param options
 */
function checkMissingProviders(suiteConfigs: TestSuiteConfig[], suiteMeta: TestSuiteMeta, options: MatrixOptions = {}) {
  const suiteConfigProviders = suiteConfigs.map(({ provider }) => provider)

  const missingProviders = Object.values(Providers).reduce((acc, provider) => {
    if (suiteConfigProviders.includes(provider)) return acc
    if (options.optOut?.from.includes(provider)) return acc

    return [...acc, provider]
  }, [] as Providers[])

  if (missingProviders.length) {
    throw new Error(
      `Test: '${suiteMeta.testDirName}' is missing providers '${missingProviders
        .map((x) => `'${x}'`)
        .join(', ')}' out out using options.optOut`,
    )
  }
}
```

millsp (changes requested):

packages/client/tests/functional/queryRaw-send-type-hints/tests.ts
```diff
@@ -5,70 +5,80 @@ declare let prisma: import('@prisma/client').PrismaClient
...
     })
-
-    expect(record?.binary).toEqual(Buffer.from('hello'))
-  })
-})
+  },
+  {
+    optOut: {
+      from: ['mongodb', 'cockroachdb'],
+      reason: '$queryRaw only works on SQL based providers',
```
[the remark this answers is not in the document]
janpio: Should work fine on CRDB.

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
danstarns: Nice idea, added https://github.com/prisma/prisma/pull/13507/files#diff-96bd19e93b8f591ba66f10ae527c63fd90a115bde817d872d378cfce788a210f

packages/client/tests/functional/queryRaw-send-type-hints/tests.ts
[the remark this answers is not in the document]
janpio: (Note: Although the snippet does not show "outdated", that is because only line 80 was edited. 👍 )

packages/client/tests/functional/metrics-disabled/tests.ts
```diff
@@ -21,8 +21,11 @@ testMatrix.setupTestSuite(
   },
   {
     optOut: {
-      from: ['mongodb'],
-      reason: 'Metrics dont work for mongodb',
+      from: ['mongodb', 'cockroachdb'],
+      reason: `
+        Metrics dont work for mongodb
```
SevInf, line 26: Metrics are provider-independent and should work for both of those. Could you point me to the failed run?

danstarns: Fixed now, mongodb don't expose these metrics: https://github.com/prisma/prisma/pull/13507/files#diff-cfa5347424177cc489589c50fb6c439d2cfdc502a36521f07bce990549b62948R35

packages/client/tests/functional/_example/tests.ts
```diff
@@ -38,15 +39,24 @@ testMatrix.setupTestSuite((suiteConfig, suiteMeta) => {
...
+    test('getTestSuiteSchema', async () => {
+      const schemaString = await getTestSuiteSchema(suiteMeta, suiteConfig)
+
+      expect(schemaString).toContain('generator')
+      expect(schemaString).toContain('datasource')
+      expect(schemaString).toContain('model')
+    })
+  },
+  // Use to optOut to optout from testing selected providers
+  // - otherwize the suite will require all providers to be specified.
```
millsp, line 55:
```suggestion
  // otherwise the suite will require all providers to be specified.
```

millsp (approved):

packages/client/tests/functional/_utils/providers.ts
```diff
@@ -0,0 +1,7 @@
...
```
aqrln, line 1: SQL Server probably still needs to be added

Issue #12356, opened 2022-03-15
`TypeError: at normalizeAndValidateHeaderValue` with data proxy in cloudflare workers/miniflare


clintonwoo:
### Bug description

Client is bugged in my closed source project but also in this open source project I found it so using that as an example.

When using Prisma Data Proxy in Cloudflare Workers with `prisma` 3.10.0 and `miniflare` 2.3.0 (also using `remix` 1.2.3 but probably not related) an error occurs when making any request since the `transactionId` header is passed as undefined in prisma causing a header normalization error (since undefined isn't a valid header).

### Workaround

I've managed to fix the error locally by modifying line 93 col 13582 of `node_modules/@prisma/client/runtime/proxy.js` to be `const headers={...n};if(typeof t=="string"&&r){headers.transactionId= t}return{inTx:typeof t=="number"&&r?!0:void 0,headers}` instead of `return{inTx:typeof t=="number"&&r?!0:void 0,headers:{transactionId:typeof t=="string"&&r?t:void 0,...n}` since in the bugged code the `transactionId` header will be undefined if not present, thus causing the `normalizeAndValidateHeaderValue` error.

### How to reproduce

1. Go to https://github.com/marcomafessolli/remix-prisma-cloudflare-workers/tree/d86bb949016ce35f805ab113f4df89f3e96b3644 (hard link for current point in time of master branch)
2. Git clone that project
3. Follow the steps to run the project (add an .env, `npm i`, `PRISMA_CLIENT_ENGINE_TYPE=dataproxy npx prisma generate` and create your prisma data proxy project if not already using one)
4. Run `npm run dev`
5. Open the root index page for the website (only one page exists).
6. See error

### Expected behavior

```
TypeError:
[dev:*worker]     at normalizeAndValidateHeaderValue (`<PATH>/remix-prisma-cloudflare-workers/node_modules/undici/lib/fetch/headers.js:42:11`)
[dev:*worker]     at HeadersList.append (`<PATH>/remix-prisma-cloudflare-workers/node_modules/undici/lib/fetch/headers.js:98:29`)
[dev:*worker]     at fill (`<PATH>/remix-prisma-cloudflare-workers/node_modules/undici/lib/fetch/headers.js:86:15`)
[dev:*worker]     at new Request (`<PATH>/remix-prisma-cloudflare-workers/node_modules/undici/lib/fetch/request.js:433:9`)
[...]

PR #13512, opened 2022-05-26
chore(client): split the runtime into two [DPGA, 2] 

millsp:
This PR introduces a distinction between `@prisma/client` working on Node.js and it working on edge runtimes (Cloudflare Workers, Vercel Edge, etc...). The new behavior is as follows:
* DataProxy + Node = `@prisma/client`
* DataProxy + Edge = `@prisma/client/edge`

Prev https://github.com/prisma/prisma/pull/13561
Next https://github.com/prisma/prisma/pull/13599

https://github.com/orgs/prisma/projects/70#card-82299269


helpers/compile/plugins/fill-plugin/fillPlugin.ts
```diff
@@ -191,14 +191,14 @@ const fillPlugin = (
       process: {
         inject: path.join(__dirname, 'fillers', 'process.ts'),
       },
-      global: {
```
millsp, line 194: Not 100% this should be removed yet.

millsp: (I'll re-add as needed one I get to writing the ecosystem tests for this)

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
```diff
@@ -50,20 +46,16 @@ function getSelectedEnvVarNames(datasources: InternalDatasource[]) {
...
 
   for (const envVarName of envVarNames) {
-    const loadedEnvVar = loadedEnv?.parsed[envVarName]
-
-    // note that we willingly keep `undefined` values
-    // because we need that later in `declareInlineEnv`
-    selectedEnv.parsed[envVarName] = loadedEnvVar
+    /** note that we willingly keep `undefined` values because
+     * we need that later in {@link declareInjectableEdgeEnv} **/
+    selectedEnv.parsed[envVarName] = undefined
```
danstarns, line 58: This is very odd and I'm still unsure, even with your comment, why we are doing this!

packages/client/helpers/build.ts
```diff
@@ -6,15 +6,13 @@ import { build } from '../../../helpers/compile/build'
...
-const runtimeBuildConfig: BuildOptions = {
+const nodeRuntimeBuildConfig: BuildOptions = {
   name: 'runtime',
   entryPoints: ['src/runtime/index.ts'],
   outfile: 'runtime/index',
   bundle: true,
   define: {
-    'globalThis.NOT_PRISMA_DATA_PROXY': 'true',
-    // that fixes an issue with lz-string umd builds
-    'define.amd': 'false',
```
millsp, line 17: Looks like this issue does not exist anymore, it was probably because of the previous build hacks.

packages/client/helpers/build.ts
```diff
@@ -28,18 +26,16 @@ const browserBuildConfig: BuildOptions = {
 }
 
 // we define the config for proxy
-const proxyBuildConfig: BuildOptions = {
-  name: 'proxy',
+const edgeRuntimeBuildConfig: BuildOptions = {
+  name: 'edge',
   entryPoints: ['src/runtime/index.ts'],
-  outfile: 'runtime/proxy',
+  outfile: 'runtime/edge',
```
millsp, line 32: Renaming the runtime to `edge` for `@prisma/client/edge`. Data Proxy can still be used with `@prisma/client` on Node.

packages/client/scripts/postinstall.js
```diff
@@ -197,30 +194,46 @@ function run(cmd, params, cwd = process.cwd()) {
...
+    const defaultEdgeIndexDtsPath = path.join(dotPrismaClientDir, 'edge.d.ts')
+
+    if (!fs.existsSync(defaultNodeIndexPath)) {
+      await copyFile(path.join(__dirname, 'default-index.js'), defaultNodeIndexPath)
+    }
 
-    if (!fs.existsSync(defaultIndexJsPath)) {
-      await copyFile(path.join(__dirname, 'default-index.js'), defaultIndexJsPath)
+    if (!fs.existsSync(defaultEdgeIndexPath)) {
+      await copyFile(path.join(__dirname, 'default-edge.js'), defaultEdgeIndexPath)
```
millsp, line 217: Copying the empty dummy file to the generation folder.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -11,8 +11,8 @@ import type { GetPrismaClientConfig } from '../../runtime/getPrismaClient'
 import type { InternalDatasource } from '../../runtime/utils/printDatasources'
 import { buildDirname } from '../utils/buildDirname'
 import { buildDMMF } from '../utils/buildDMMF'
+import { buildInjectableEdgeEnv as buildInjectableEdgeEnv } from '../utils/buildInjectableEdgeEnv'
```
millsp, line 14:
```suggestion
import { buildInjectableEdgeEnv } from '../utils/buildInjectableEdgeEnv'
```

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
millsp: I'll update the comment. The answer to your question should be clarified in `declareInjectableEdgeEnv`.

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -46,14 +46,16 @@ export type Request = {
...
-  const headers = requests[0].headers
+  const headers = requests[0].headers ?? {}
 
   // if the tx has a number for an id, then it's a regular batch tx
   const _inTx = typeof txId === 'number' && inTx ? true : undefined
   // if the tx has a string for id, it's an interactive transaction
   const _txId = typeof txId === 'string' && inTx ? txId : undefined
 
-  return { inTx: _inTx, headers: { transactionId: _txId, ...headers } }
+  if (_txId !== undefined) headers.transactionId = _txId
```
millsp, line 56: Fixes an issue with miniflare, remix, and others where `fetch` is injected with undici's implementation for local development. The problem was that `transactionId` was being passed `undefined`, which caused runtime errors.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -42,12 +42,11 @@ const ALTER_RE = /^(\s*alter\s)/i
...
   // eslint-disable-next-line no-var
-  var NOT_PRISMA_DATA_PROXY: true
+  var NOT_EDGE_CLIENT: true
 }
 
-// @ts-ignore esbuild trick to set a default
-// eslint-disable-next-line no-self-assign
-;(globalThis = globalThis).NOT_PRISMA_DATA_PROXY = true
+// used by esbuild for tree-shaking
+globalThis.NOT_EDGE_CLIENT = true
```
millsp, line 49: Not a workaround that we need anymore since `esbuild` upgrade.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -1,6 +1,3 @@
-/// <reference lib="webworker" />
```
millsp, line 1: Removed because it causes global scope pollution.

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -1,23 +1,19 @@
...
-export type RequestOptions = O.Patch<
-  {
-    headers?: { [k: string]: string }
-    body?: string
-  },
-  RequestInit
->
-
-// our implementation handles less
+export type RequestOptions = O.Patch<{ headers?: { [k: string]: string }; body?: string }, RequestInit>
```
millsp, line 9: Reusing types from `node-fetch` instead of `webworkers`.

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -39,8 +35,7 @@ export async function request(url: string, options: RequestOptions = {}): Promis
  */
 function buildHeaders(options: RequestOptions): RequestOptions['headers'] {
   return {
-    // this ensures headers will always be valid
-    ...JSON.parse(JSON.stringify(options.headers)),
```
millsp, line 43: Was used to removed `undefined` values.

packages/client/helpers/build.ts
```diff
@@ -6,15 +6,13 @@ import { build } from '../../../helpers/compile/build'
...
+const nodeRuntimeBuildConfig: BuildOptions = {
   name: 'runtime',
   entryPoints: ['src/runtime/index.ts'],
   outfile: 'runtime/index',
   bundle: true,
   define: {
-    'globalThis.NOT_PRISMA_DATA_PROXY': 'true',
-    // that fixes an issue with lz-string umd builds
-    'define.amd': 'false',
+    NOT_EDGE_CLIENT: 'true',
```
SevInf, line 15: Have been curious regarding this env var ever since I saw it under old name: why not the opposite, `EDGE_CLIENT: true`. Or, in this case, why not `NODE_CLIENT: true`. IMHO, it's much easier to mentally parse without double negatives.

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -80,13 +75,12 @@ function buildResponse(incomingData: Buffer[], response: IncomingMessage): Reque
  * @param options
  * @returns
  */
-function nodeFetch(url: string, options: RequestOptions = {}): Promise<RequestResponse> {
+async function nodeFetch(url: string, options: RequestOptions = {}): Promise<RequestResponse> {
+  const https: typeof Https = await globalThis[['e', 'v', 'a', 'l'].join('')](`import('https')`)
```
millsp, line 79: Used to obfuscate `eval` from bundlers. Vercel Edge Functions fail on build if any eval is detected, but this part of the code is not reachable at runtime, so it is safe. The purpose of this is for `@prisma/client/edge` to work on Node as well, because `fetch` is only available from Node@18.

helpers/compile/plugins/fill-plugin/fillPlugin.ts
```diff
@@ -191,14 +191,14 @@ const fillPlugin = (
...
-        define: '{}',
-      },
-      globalThis: {
-        define: '{}',
-      },
 
       // not needed
+      // global: {
```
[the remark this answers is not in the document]
millsp: Yes, there is [one point here](https://github.com/prisma/prisma/pull/13512/files/e48f255c0c53b5ef3fab6a014653582b9cb6e5cb#r885385013). FYI, most of the polyfills are commented out, but just there so you know, if you ever need it.

packages/client/helpers/build.ts
```diff
@@ -28,18 +26,16 @@ const browserBuildConfig: BuildOptions = {
 }
 
 // we define the config for proxy
```
millsp, line 28:
```suggestion
// we define the config for edge
```

packages/client/helpers/build.ts
millsp: No actual reason. Fixed.

packages/client/scripts/default-edge.d.ts
```diff
@@ -0,0 +1,43 @@
+/**
```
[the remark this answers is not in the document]
SevInf: Would it make sense to use `never` here instead of `any`, so we'll have compile-time error (instead of runtime-only) if we try to use it without generating?

SevInf:
Question: do we actually envision that somebody might want to bundle import of `@prisma/client/edge` in their browser bundle? For example, some isomorphic code, not touching databases but using enums? If so, we might package it slightly differently so bundlers will recognize it:

- make `.edge` a directory
- put `package.json` in it with `main`, `types` and `browsers` fields pointing to corresponding files
- generate the files into that subdirectory

packages/client/scripts/postinstall.js
```diff
@@ -197,30 +194,49 @@ function run(cmd, params, cwd = process.cwd()) {
...
-      await copyFile(path.join(__dirname, 'default-index-browser.js'), defaultIndexBrowserJSPath)
+
+    if (!fs.existsSync(defaultEdgeIndexPath)) {
+      await copyFile(path.join(__dirname, 'default-edge.js'), defaultEdgeIndexPath)
     }
 
-    if (!fs.existsSync(defaultIndexDTSPath)) {
-      await copyFile(path.join(__dirname, 'default-index.d.ts'), defaultIndexDTSPath)
+    if (!fs.existsSync(defaultEdgeIndexDtsPath)) {
+      await copyFile(path.join(__dirname, 'default-index.d.ts'), defaultEdgeIndexDtsPath)
```
aqrln, line 232: Would it be easier to prepare the whole directory template in `scripts` and copy it here instead of building the directory structure file by file?

(Maybe not much easier, since what's happening here is a bit more complicated than just copying, but at least it should be a more scalable approach I think)

aqrln (approved):
Overall looks good to me, thank you!

packages/client/helpers/build.ts
janpio: Should this be committed?

packages/client/helpers/build.ts
```diff
@@ -28,18 +26,16 @@ const browserBuildConfig: BuildOptions = {
...
+const edgeRuntimeBuildConfig: BuildOptions = {
+  name: 'edge',
   entryPoints: ['src/runtime/index.ts'],
-  outfile: 'runtime/proxy',
+  outfile: 'runtime/edge',
   bundle: true,
   minify: true,
   legalComments: 'none',
   define: {
     // that helps us to tree-shake unused things out
```
janpio, line 37: comment still valid although former properties were fully replaced? Or does `NODE_CLIENT` now fulfill this function?

packages/client/scripts/postinstall.js
```diff
@@ -197,30 +194,49 @@ function run(cmd, params, cwd = process.cwd()) {
   })
 }
 
-async function ensureEmptyDotPrisma() {
+/**
+ * Copies our default "throw" files into the default generation folder. These
+ * files are dummy and informative because they just throw an error to let the
+ * user know that they have forgotten to run `prisma generate`.
```
janpio, line 200: - Maybe mention `node_modules/.prisma/client` as the default generateion folder
- Technically this does not have to mean that they forgot, but also that they might not have a schema file yet. Maye reword that it just means they or package installation did not run `prisma generate` successfully yet to create a generated Client in that location

packages/client/scripts/postinstall.js
```diff
@@ -197,30 +194,49 @@ function run(cmd, params, cwd = process.cwd()) {
...
+ */
+async function createDefaultGeneratedThrowFiles() {
   try {
     const dotPrismaClientDir = path.join(__dirname, '../../../.prisma/client')
+    const defaultNodeIndexPath = path.join(dotPrismaClientDir, 'index.js')
+    const defaultBrowserIndexPath = path.join(dotPrismaClientDir, 'index-browser.js')
+    const defaultNodeIndexDtsPath = path.join(dotPrismaClientDir, 'index.d.ts')
```
janpio, line 207:
```suggestion
    const defaultNodeIndexDtsPath = path.join(dotPrismaClientDir, 'index.d.ts')
    const defaultBrowserIndexPath = path.join(dotPrismaClientDir, 'index-browser.js')
```
switch lines two "NodeIndex...Path" are next to each other.

packages/client/src/generation/TSClient/Generatable.ts
```diff
@@ -1,23 +1,18 @@
...
 }
 
-export function JS(gen: Generatable): string | Promise<string> {
-  if (gen.toJS) {
-    return gen.toJS()
-  }
-
-  return ''
+export function JS(gen: Generatable, edge = false): string | Promise<string> {
+  return gen.toJS?.(edge) ?? ''
```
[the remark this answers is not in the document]
janpio: I would honestly not know what this does without looking it up. I like the more boring code from before 🦐

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
```diff
@@ -10,22 +7,21 @@ type LoadedEnv = {
...
- * @param envPaths
  * @returns
  */
-export function buildInlineEnv(dataProxy: boolean | undefined, datasources: InternalDatasource[], envPaths: EnvPaths) {
-  if (dataProxy === true) {
+export function buildInjectableEdgeEnv(edge: boolean | undefined, datasources: InternalDatasource[]) {
+  if (edge === true) {
     const envVarNames = getSelectedEnvVarNames(datasources)
-    const loadedEnv = loadSelectedEnvVars(envPaths, envVarNames)
+    const loadedEnv = getEmptyEnvObjectForVars(envVarNames)
```
janpio, line 22: Did I understand the underlying code (and comments) correctly that this now somethat that makes sure we have fallbacks for all the used env vars - and the fallback is `undefined`? Why can we not just do anything here and still have the same result?

packages/client/src/runtime/RequestHandler.ts
janpio: https://github.com/prisma/prisma/issues/12356 I guess

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -238,13 +237,17 @@ export interface GetPrismaClientConfig {
   inlineSchema?: string
 
   /**
-   * The contents of the env saved into a special object
+   * A special env object jus for the data proxy edge runtime.
```
janpio, line 240:
```suggestion
   * A special env object just for the data proxy edge runtime.
```

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
janpio: Ah this comment here clarified some more: https://github.com/prisma/prisma/pull/13512/files#diff-09784fc0113a735572cff48990efa35434fb6ec6ec8876845ec9cf423642e5fdR240-R242

janpio (commented):
I skimmed the code and asked a few questions and left typo suggestions.

Generally this looks good - I can follow most changed and they are consistent and make sense 👍

packages/client/scripts/default-edge.d.ts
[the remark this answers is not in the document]
millsp: I am reusing the what we currently ship as the default, happy to revisit this sometime soon.

packages/client/src/runtime/RequestHandler.ts
millsp: Correct, I should have done this in the isolated bugs PR.

packages/client/helpers/build.ts
```diff
@@ -115,10 +111,9 @@ function bundleTypeDefinitions(filename: string, outfile: string) {
   }
 }
 
-void build([generatorBuildConfig, runtimeBuildConfig, browserBuildConfig, proxyBuildConfig]).then(() => {
+void build([generatorBuildConfig, nodeRuntimeBuildConfig, browserBuildConfig, edgeRuntimeBuildConfig]).then(() => {
   if (process.env.DEV !== 'true') {
     bundleTypeDefinitions('declaration/client/src/runtime/index', 'runtime/index')
-    bundleTypeDefinitions('declaration/client/src/runtime/index', 'runtime/proxy')
```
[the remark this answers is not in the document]
millsp: We now use the default index in all cases (or point to it), no need to generate a duplicate anymore (`edge` and `index` are the same)

packages/client/helpers/build.ts
millsp: `NOT_PRISMA_DATA_PROXY=false` actually means `NODE_CLIENT-true`, but is easier to parse, as Serhii suggested.

packages/client/src/generation/generateClient.ts
```diff
@@ -200,7 +225,11 @@ export async function generateClient({
   }
 
   await makeDir(finalOutputDir)
+  await makeDir(path.join(finalOutputDir, 'edge'))
```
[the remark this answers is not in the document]
millsp: Thanks, I added this check in the finalization PR.

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
millsp: Will add the correct comments in the finalization PR.

packages/client/scripts/postinstall.js
millsp: I will update this in the finalization PR.

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
millsp: I will update the comment in the finalization PR.

helpers/compile/plugins/fill-plugin/fillPlugin.ts
millsp: I think this needs to be restored, I will make it more resilient in the finalization PR.

PR #13537, opened 2022-05-28
ci: added support for additional cockroachdb tests

### LINKED ISSUE #13268: Add more integration tests for CockroachDB for `db` and `migrate` commands
The prisma engines have good testing of the connector but we would get a better confidence that everything is working as expected with additional integration tests on the CLI side here in prisma/prisma.

Previous issue was about having a minimal testing setup (db pull tests were added)
https://github.com/prisma/prisma/issues/12926

Now we could expand and cover more commands
- [ ] db push (currently, only SQLite is tested: https://github.com/prisma/prisma/issues/13538)
- [ ] migrate reset (currently, only SQLite is tested: https://github.com/prisma/prisma/issues/13539)
- [x] db execute
- [x] migrate dev
- [x] migrate diff

jkomyno:
This PR adds cockroachdb tests for:

- [x] db execute
- [x] migrate dev
- [x] migrate diff 

Closes https://github.com/prisma/prisma/issues/13268


jkomyno:
Note: merging https://github.com/prisma/prisma/pull/13547 is necessary to make this PR's MacOS tests succeed.

packages/migrate/src/__tests__/MigrateResolve.test.ts
```diff
@@ -203,6 +203,7 @@ describeIf(!process.env.TEST_SKIP_COCKROACHDB)('cockroachdb', () => {
           `)
 
     expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(`
+      Environment variables loaded from prisma/.env
```
janpio, line 206: What causes this unrelated change?

packages/migrate/src/__tests__/fixtures/schema-only-cockroachdb/prisma/shadowdb.prisma
```diff
@@ -1,11 +1,7 @@
-generator client {
-  provider = "prisma-client-js"
-}
-
 datasource db {
   provider          = "cockroachdb"
-  url               = env("TEST_COCKROACHDB_SHADOWDB_URI_MIGRATE")
-  shadowDatabaseUrl = env("TEST_COCKROACHDB_SHADOWDB_URI_MIGRATE")
+  url               = env("TEST_COCKROACH_SHADOWDB_URI_MIGRATE")
+  shadowDatabaseUrl = env("TEST_COCKROACH_SHADOWDB_URI_MIGRATE")
```
janpio, line 4: I see no equivalent renaming of an env var definition anywhere - did I miss it?

janpio (approved):
Generally looks good, and tests seem to be passing.

(Some actual TS developer should review besides me before merge though)

Also see questions and comments.

pimeys (approved):
Looks good to me. More tests are better than no tests, especially if they're green. Code quality looks decent to me.

tomhoule (approved):

packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -1133,6 +1134,200 @@ describe('postgresql', () => {
...
+  })
+
+  // eslint-disable-next-line jest/no-identical-title
+  it('schema only', async () => {
+    ctx.fixture('schema-only-cockroachdb')
+
+    const result = MigrateDev.new().parse([])
+    await expect(result).resolves.toMatchInlineSnapshot(``)
+    expect(ctx.mocked['console.log'].mock.calls).toMatchSnapshot()
+    expect(ctx.mocked['console.error'].mock.calls).toMatchSnapshot()
```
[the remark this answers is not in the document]
jkomyno: You're correct. I had simply kept the same test structure as before to minimize the `git diff`

packages/migrate/src/__tests__/MigrateResolve.test.ts
jkomyno: The fact that `packages/migrate/src/__tests__/fixtures/schema-only-cockroachdb/prisma` contains a `.env` file, which is automatically load by the [`loadEnv`](https://github.com/prisma/prisma/blob/c1d4f6ea4a0499759dcc1e43cc6c846ccb8051a8/packages/sdk/src/utils/tryLoadEnvs.ts#L119-L121) function.

packages/migrate/src/__tests__/fixtures/schema-only-cockroachdb/prisma/shadowdb.prisma
janpio: So how did this test ever work? Should it not have failed for a missing env var before? Something doesn't add up, right?

packages/migrate/src/__tests__/MigrateDev.test.ts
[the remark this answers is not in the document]
janpio: The diff of the snapshot file with all empty strings made me worry this was not testing the right thing, but on checking the test file I assumed it _expected_ those empty strings.

packages/migrate/src/__tests__/MigrateResolve.test.ts
janpio: Ah, and before the `schema-only-cockroachdb` test fixture did not - you added that 👍

packages/migrate/src/__tests__/fixtures/schema-only-cockroachdb/prisma/shadowdb.prisma
jkomyno: The answer is pretty easy: `shadowdb.prisma` was never loaded before. In other words, this pre-existing schema was just noise in the test codebase

packages/migrate/src/__tests__/MigrateDev.test.ts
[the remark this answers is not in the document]
janpio: You added new snapshots here. The use the suboptimal way to store the snapshots for the use case (= confirming that something does not return actual data). That is what should be changed.

Updating other, existing snapshot logic in other tests should not become part of this PR of course and if we have the same problem elsewhere we should create an issue and fix that independently.

packages/migrate/src/__tests__/fixtures/schema-only-cockroachdb/prisma/shadowdb.prisma
janpio: Ok, that explains it 👍 Good we fixed that then.

packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -651,7 +651,7 @@ describe('sqlite', () => {
 
     `)
     expect(ctx.mocked['console.log'].mock.calls).toEqual([])
-    expect(ctx.mocked['console.error'].mock.calls).toEqual([])
+    expect(ctx.mocked['console.error'].mock.calls).toMatchSnapshot()
```
janpio: Putting this one into an inline snapshot would also be perfectly fine I guess for the tests, no need to have and look into an additional file. But of course not necessary to merge this PR now.

janpio (approved):

PR #13561, opened 2022-05-30
chore(client, cli): implement --data-proxy [DPGA, 1]

millsp:
This PR replaces the `PRISMA_CLIENT_ENGINE_TYPE=dataproxy` mechanism in favor of `--data-proxy` via the cli.

Next https://github.com/prisma/prisma/pull/13512

https://github.com/orgs/prisma/projects/70#card-82299232

PR #13584, opened 2022-05-31
feat(client): Expose metrics to the client

### LINKED ISSUE #12875: Expose Metrics in Prisma Client
Once the query engine has implemented metrics we need to expose it via the Prisma Client Api

Depends on https://github.com/prisma/prisma/issues/12874
Epic:https://github.com/prisma/client-planning/issues/2

SevInf:
Close #12875

### DIFF
--- a/packages/client/src/runtime/core/metrics/index.ts
+++ b/packages/client/src/runtime/core/metrics/index.ts
@@ -1,0 +1,1 @@
+export { MetricsClient } from './MetricsClient'
--- a/packages/client/src/runtime/core/metrics/index.ts
+++ b/packages/client/src/runtime/core/metrics/index.ts
@@ -1,1 +1,0 @@
-export { MetricsClient } from './MetricsClient'
--- a/packages/client/tests/functional/metrics-disabled/tests.ts
+++ b/packages/client/tests/functional/metrics-disabled/tests.ts
@@ -1,0 +1,19 @@
+import testMatrix from './_matrix'
+import { expectTypeOf } from 'expect-type'
+
+// @ts-ignore this is just for type checks
+declare let prisma: import('@prisma/client').PrismaClient
+
+testMatrix.setupTestSuite(() => {
+  test('attempt to use $metrics a compile-time error', () => {
+    // TODO: this is always an error in the editor. Maybe we need to generate
+    // schema-dependent client types somehow
+    expectTypeOf(prisma).not.toHaveProperty('$metrics')
+  })
+
+  test('attempt to use $metrics a run-time error', () => {
+    expect(() => (prisma as any).$metrics).toThrowErrorMatchingInlineSnapshot(
+      `\`metrics\` preview feature must be enabled in order to access metrics API`,
+    )
+  })
+})
--- a/packages/client/tests/functional/metrics/_matrix.ts
+++ b/packages/client/tests/functional/metrics/_matrix.ts
@@ -6,12 +6,4 @@ export default defineMatrix(() => [
       provider: 'sqlite',
     },
   ],
-  [
-    {
-      engineType: 'library',
-    },
-    {
-      engineType: 'binary',
-    },
-  ],
 ])
--- a/packages/client/tests/functional/metrics/_matrix.ts
+++ b/packages/client/tests/functional/metrics/_matrix.ts
@@ -1,0 +1,17 @@
+import { defineMatrix } from '../_utils/defineMatrix'
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'sqlite',
+    },
+  ],
+  [
+    {
+      engineType: 'library',
+    },
+    {
+      engineType: 'binary',
+    },
+  ],
+])
--- a/packages/client/tests/functional/metrics/tests.ts
+++ b/packages/client/tests/functional/metrics/tests.ts
@@ -1,0 +1,101 @@
+import testMatrix from './_matrix'
+
+// @ts-ignore this is just for type checks
+declare let prisma: import('@prisma/client').PrismaClient
+// @ts-ignore this is just for type checks
+declare let PrismaClient: typeof import('@prisma/client').PrismaClient
+
+testMatrix.setupTestSuite(() => {
+  beforeAll(async () => {
+    const email = 'user@example.com'
+    await prisma.user.create({
+      data: {
+        email,
+      },
+    })
+
+    await prisma.user.findFirst({ where: { email } })
+  })
+
+  test('returns metrics in prometheus format', async () => {
+    const metrics = await prisma.$metrics.prometheus({ globalLabels: { foo: 'bar' } })
+
+    expect(metrics).toContain('query_total_operations')
+    expect(metrics).toContain('query_total_queries')
+    expect(metrics).toContain('pool_active_connections')
+    expect(metrics).toContain('pool_active_connections')
+    expect(metrics).toContain('pool_idle_connections')
+    expect(metrics).toContain('pool_wait_count')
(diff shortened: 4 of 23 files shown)


packages/client/tests/functional/metrics/tests.ts
```diff
@@ -0,0 +1,101 @@
...
+  })
+
+  test('includes global labels in prometheus format', async () => {
+    const metrics = await prisma.$metrics.prometheus({ globalLabels: { label1: 'value1', label2: 'value2' } })
+
+    expect(metrics).toContain('{label1="value1",label2="value2"}')
+  })
+
+  test('returns metrics in json format', async () => {
+    const { counters, gauges, histograms } = await prisma.$metrics.json()
```
SevInf, line 41: So, here, rather than doing snaphots and introducing dependency on engine implementation details, we assert that returned metrics has the data and general properties of the data kind of make sense. This should make test more resilient to engine internal changes, which will cause the changes to the metrics.

packages/client/tests/functional/metrics/tests.ts
```diff
@@ -0,0 +1,101 @@
...
+    for (const gauge of metrics.gauges) {
+      expect(gauge.labels).toEqual({ label1: 'value1', label2: 'value2' })
+    }
+
+    for (const histogram of metrics.histograms) {
+      expect(histogram.labels).toEqual({ label1: 'value1', label2: 'value2' })
+    }
+  })
+
+  describe('mutliple instances', () => {
```
SevInf, line 77: This nested describe is here only for `afterAll`'s sake - we need to ensure that second client gets disconnected even if the test fails.

garrensmith (commented):
This is looking good. Easy code to follow.

packages/client/tests/functional/metrics/tests.ts
```diff
@@ -0,0 +1,101 @@
...
+      await secondClient.user.create({
+        data: {
+          email: 'second-user@example.com',
+        },
+      })
+
+      const metrics1 = await prisma.$metrics.json()
+      const metrics2 = await secondClient.$metrics.json()
+
+      expect(metrics1).not.toEqual(metrics2)
```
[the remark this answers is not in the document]
SevInf: Yes, `toEqual` does deep comparison.

SevInf:
Now it uses updated engines and is ready for review

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -254,4 +255,12 @@ export class DataProxyEngine extends Engine {
 
     return [host, apiKey]
   }
+
+  metrics(options: MetricsOptionsJson): Promise<Metrics>
+  metrics(options: MetricsOptionsPrometheus): Promise<string>
+  metrics(options: EngineMetricsOptions): Promise<Metrics> | Promise<string> {
+    throw new NotImplementedYetError('Metric are not yet supported', {
```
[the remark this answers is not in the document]
millsp: Should we perhaps error at generation time like we do for `interactiveTransactions` with `--data-proxy`?

packages/client/src/runtime/core/metrics/index.ts
```diff
@@ -0,0 +1 @@
+export { MetricsClient } from './MetricsClient'
```
millsp, line 1: Is this file/pattern in case we want to have better/centralized exports?

packages/client/tests/functional/metrics-disabled/tests.ts
```diff
@@ -0,0 +1,19 @@
+import testMatrix from './_matrix'
+import { expectTypeOf } from 'expect-type'
+
+// @ts-ignore this is just for type checks
+declare let prisma: import('@prisma/client').PrismaClient
+
+testMatrix.setupTestSuite(() => {
+  test('attempt to use $metrics a compile-time error', () => {
+    // TODO: this is always an error in the editor. Maybe we need to generate
```
millsp, line 9: Do you mean that we need to have dev-time types/intellisense?

packages/client/tests/functional/metrics/_matrix.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'sqlite',
+    },
+  ],
+  [
+    {
+      engineType: 'library',
```
millsp, line 11: I really like that we could test the engineTypes this way! However, the CI already does that by running all the tests with binary and again with library, and it does it via the `PRISMA_CLIENT_ENGINE_TYPE` env var and via `getClientEngineType` in the code. tl;dr; You don't need to add anything and can remove both engine types.

millsp (commented):
Very nice work 👏

packages/client/tests/functional/metrics-disabled/tests.ts
SevInf: Yea, basically, right now it always highlighted as an error in editors, since `prisma` has `any` type there and any can have anything. It works when you actually run the tests, since there `prisma` is properly typed by then. Having proper types during development will fix this. I'd maybe do it separately as a part of testing project

packages/client/src/runtime/core/metrics/index.ts
SevInf: So, here I want two things:

1. Centralized place for importing anything metrics-related, hence separate folder with `index.ts`
2. Still being to import to open `MetricsClient.ts` file by name, hence `index.ts` doing just a re-export

Admittedly, "anything metrics-related" right now is only `MetricsClient` class and it will quite likely stay that way, so I guess we could remove this file. What would you prefer?

packages/engine-core/src/data-proxy/DataProxyEngine.ts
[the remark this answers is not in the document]
SevInf: Ok, i'll wait until #13561  is merged then

garrensmith:
I've noticed a bug if I use the client in binary mode. If I do:
```
  log: ["info", "error", "query", "warn"],
});
let run = async () => {
  let metrics = await prisma.$metrics.prometheus();
};
```

I get the error:
```
/Users/garren/dev/prisma/reports/ttt/node_modules/@prisma/client/runtime/index.js:43171
    throw new Error("Connection has not been opened");
          ^
Error: Connection has not been opened
    at assertHasPool (/Users/garren/dev/prisma/reports/ttt/node_modules/@prisma/client/runtime/index.js:43171:11)
    at Connection.raw (/Users/garren/dev/prisma/reports/ttt/node_modules/@prisma/client/runtime/index.js:43197:5)
    at Connection.post (/Users/garren/dev/prisma/reports/ttt/node_modules/@prisma/client/runtime/index.js:43215:17)
    at BinaryEngine.metrics (/Users/garren/dev/prisma/reports/ttt/node_modules/@prisma/client/runtime/index.js:43977:44)
    at MetricsClient.prometheus (/Users/garren/dev/prisma/reports/ttt/node_modules/@prisma/client/runtime/index.js:39905:25)
 ```

but if I do
```
  log: ["info", "error", "query", "warn"],
});
let run = async () => {
  await prisma.$connect();
  let metrics = await prisma.$metrics.prometheus();
};
```
Then it works. I'm guessing we need to make sure we open our HTTP client.

SevInf:
@garrensmith fixed

packages/engine-core/src/data-proxy/DataProxyEngine.ts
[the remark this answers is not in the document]
janpio: You are right, this should indeed be handled as early as possible.

millsp (approved):

Issue #11344, opened 2022-01-24
Create enum for error codes


luxaritas:
## Problem

Currently, Prisma [documents](https://www.prisma.io/docs/reference/api-reference/error-reference) the possible errors that the client can raise and possible error codes for each error type. However, if you want to react to specific errors (such as a unique constraint violation), you wind up with something like:
```ts
if (e instanceof Prisma.PrismaClientKnownRequestError) {
  if (e.code === 'P2002') {
    console.log(
      'There is a unique constraint violation, a new user cannot be created with this email'
    )
  }
}
```
The `P2002` is a magic string that needs to be commented or looked up in the documentation to understand it's precise meaning. Additionally, to discover possible error codes you have to look in the documentation, and this is potentially error-prone if Prisma changes how it handles its error codes.

## Suggested solution
It would be great if Prisma would expose an enum, so that instead of using a magic string you could use something like `PrismaErrorCode.UNIQUE_CONSTRAINT_FAILED`

## Alternatives
* Current behavior
* Creating separate types for each specific error - also a good option, but I assume that it was not already implemented this way for a reason? This could also provide better typing for contextual metadata if that's relevant.

halvardssm:
First time contributor here, would a PR be accepted? If so, where should I place this code? 😃

aqrln:
@luxaritas

> Creating separate types for each specific error - also a good option, but I assume that it was not already implemented this way for a reason? This could also provide better typing for contextual metadata if that's relevant.

In fact, the only reason this hasn't been implemented yet is because we didn't have the capacity to, but this is the preferable alternative which is we are currently working on (although I can't share a specific timeline yet). Here's the original issue: https://github.com/prisma/prisma/issues/5040, and here's the recently created epic for implementation: https://github.com/prisma/prisma/issues/12340.

@halvardssm

> First time contributor here, would a PR be accepted?

First of all, thank you for you willingness to contribute, it is really appreciated! That said, unfortunately, I don't think an external PR for this issue specifically would be feasible at this point (although in general PRs are always welcome!).

On the Rust side, [this PR](https://github.com/prisma/prisma-engines/pull/2451) would need to be finished first (it would've been enough to fix the conflicts and merge it to make an enum with error codes possible on the TypeScript side, but it still needs some further work for making full type-safe errors possible, which was the original intention). Then some work needs to happen in the TypeScript client to generate those error types and make use of them at run time.

[...]

PR #13588, opened 2022-05-31
refactor(client): added enum for error codes

florestankorp:
I was watching a NestJS tutorial on YouTube and it used Prisma, but when I saw the error messages as strings I thought it might be a good idea to have an enum those, so I just wanted to check if that's something that could be interesting for the project. If so, I will complete the list tomorrow...

Here is the whole list of error codes by the way:
https://www.prisma.io/docs/reference/api-reference/error-reference


Jolg42:
Note: we have some code like this in multiple places

In private repos
`KNOWN_QUERY_ENGINE_ERROR_CODES`
https://github.com/prisma/data-proxy/blob/165c23b04097f2955d3ade16e7a0529070b90b1c/qe-tools/qe-log-parser/src/lib.rs#L7-L22

`Codes`
https://github.com/prisma/pdp-engines/blob/85b52f723a0667b37b27d25ba637e6e807ac6c3b/lib/prisma/Errors.ts#L3-L11

and in a public one we have something that looks like what you need (⚠️ it's not a public API) 
https://github.com/prisma/reflector/blob/7a2e5ded034d7101b630eebb1f9fe039a3b1e436/src/Errors/index.ts

How do you like the approach in the last one?

aqrln:
Relevant issue: https://github.com/prisma/prisma/issues/11344

aqrln:
Note that even though the alternative we prefer is full type-safe structured errors, an enum with error codes is a nice first step in that direction, which we could start with if there is demand.

As for the approach, I'm generally -1 on hardcoding the list of errors in the client and maintaining it manually. It must be requested from the engine during generation (and https://github.com/prisma/prisma-engines/pull/2451 is a good starting point to make that possible).

garrensmith:
@florestankorp I'm going to close this. we would prefer to implement this in our DMMF first so that it is easier to keep up to date. You can follow along in this ticket https://github.com/prisma/prisma/issues/5040

florestankorp:
> @florestankorp I'm going to close this. we would prefer to implement this in our DMMF first so that it is easier to keep up to date. You can follow along in this ticket #5040

Cheers. Thanks for your message 👍🏽

PR #13599, opened 2022-06-01
chore(client): fix isolated data proxy issues [DPGA, 3]

millsp:
This PR fixes isolated data proxy issues that haven't been fixed by previous PRs (see commit messages). For reviewing this PR, please look at the changes of each individual commit since the issues being fixed are unrelated.

Prev https://github.com/prisma/prisma/pull/13512
Next https://github.com/prisma/prisma/pull/13627

https://github.com/orgs/prisma/projects/70#card-82640155


packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -48,24 +47,6 @@ export class DataProxyEngine extends Engine {
     this.remoteClientVersion = getClientVersion(this.config)
     this.headers = { Authorization: `Bearer ${apiKey}` }
     this.host = host
-
```
millsp, line 51: We can remove this logic, prone to errors and already handled in `requestInternal`

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -18,13 +19,20 @@ declare let fetch: typeof nodeFetch
...
-  } else {
-    return nodeFetch(url, options)
+  try {
+    if (jsRuntimeName === 'browser') {
+      return await fetch(url, options)
+    } else {
+      return await nodeFetch(url, options)
+    }
+  } catch (e) {
+    throw new NetworkError({ clientVersion: options.clientVersion })
```
millsp, line 35: Handle and retry when there are network errors.

packages/engine-core/src/data-proxy/errors/NetworkError.ts
```diff
@@ -0,0 +1,14 @@
...
+import { setRetryable } from './utils/setRetryable'
+
+export interface NetworkErrorInfo extends DataProxyErrorInfo {}
+
+export class NetworkError extends DataProxyError {
+  public name = 'NetworkError'
+  public code = 'P5010'
+
+  constructor(info: NetworkErrorInfo) {
+    super('Cannot fetch data from service', setRetryable(info, true))
```
millsp, line 12: This error is retryable, so hopefully users will never notice unless there is downtime.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -96,7 +96,7 @@ export class TSClient implements Generatable {
     const relativeOutdir = path.relative(process.cwd(), outputDir)
 
     const code = `${commonCodeJS({ ...this.options, browser: false })}
-${buildRequirePath(dataProxy)}
+${buildRequirePath(edge)}
```
millsp, line 98: On Node.js we need to import `path` because we do env var loading/checking in `buildWarnEnvConflicts`, which uses `path`.

packages/client/src/generation/utils/buildDirname.ts
```diff
@@ -2,13 +2,13 @@ import path from 'path'
...
- * @param dataProxy
+ * @param edge
  * @param relativeOutdir
  * @param runtimeDir
  * @returns
  */
-export function buildDirname(dataProxy: boolean | undefined, relativeOutdir: string, runtimeDir: string) {
-  if (dataProxy === true) {
+export function buildDirname(edge: boolean | undefined, relativeOutdir: string, runtimeDir: string) {
+  if (edge === true) {
```
millsp, line 11: So now, only on `edge`, we will build an empty dirname since `edge` cannot load `.env` files.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -59,13 +59,12 @@ export class TSClient implements Generatable {
...
       sqliteDatasourceOverrides,
       outputDir,
-      schemaDir,
+      schemaPath,
       runtimeDir,
       runtimeName,
       datasources,
       dataProxy,
     } = this.options
-    const schemaPath = path.join(schemaDir, 'schema.prisma')
```
millsp, line 68: Because the schema path is hardcoded here, and because we read it here in order to inline it, this would lead to schema not being found if it sits in a custom path. Now instead, we use the value from the CLI.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -83,7 +82,7 @@ export class TSClient implements Generatable {
       generator,
       relativeEnvPaths,
       sqliteDatasourceOverrides,
-      relativePath: path.relative(outputDir, schemaDir),
+      relativePath: path.relative(outputDir, path.dirname(schemaPath)),
```
millsp, line 85: I have ditched the `schemaDir` as it had little value, using `path.dirname(schemaPath)` is fine I believe.

packages/client/src/generation/generateClient.ts
```diff
@@ -316,9 +313,9 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
     }
   }
 
-  const datamodelTargetPath = path.join(finalOutputDir, 'schema.prisma')
-  if (datamodelPath !== datamodelTargetPath) {
-    await copyFile(datamodelPath, datamodelTargetPath)
+  const schemaTargetPath = path.join(finalOutputDir, 'schema.prisma')
```
millsp, line 316: Slowly eradicating "datamodel" 😄

packages/client/jest.config.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
   preset: 'ts-jest',
+  globals: {
+    'ts-jest': {
+      tsconfig: 'tsconfig.json',
+      isolatedModules: true,
```
millsp, line 6: The reason behind a few hours being wasted... (ts-jest hanging in CI) Surprisingly it worked fine for a long time, until I made some small type changes.

packages/cli/src/Generate.ts
```diff
@@ -138,7 +138,7 @@ ${chalk.bold('Examples')}
         printDownloadProgress: !watchMode,
         version: enginesVersion,
         cliVersion: pkg.version,
-        dataProxy: args['--data-proxy'],
+        dataProxy: !!args['--data-proxy'],
```
millsp, line 141: All the changes here are about not passing `undefined` values anymore (after ts-jest fixes).

packages/engine-core/src/data-proxy/utils/request.ts
SevInf, line 35: Would it make sense to:
1. Preserve original error message somehow?
2. Ensure that `e` is really related to network and not some other JS exception.

packages/client/jest.config.js
SevInf: Weird, shouldn't that already be set? `tsconfig.json` inherits the config from the root repo that already enables `isolatedModules`.

millsp: `isolatedModules` is misleading because what it does is to disable type checking.
https://kulshekhar.github.io/ts-jest/docs/getting-started/options/#options
https://github.com/kulshekhar/ts-jest/issues/3507

Side note:
Btw, since the codebase was migrated to use `@swc/jest`, we also lost all test type checking.
(I'll add this to the tech debt list so that we can fix this asap)

aqrln (approved):

PR #13605, opened 2022-06-01
chore(test): Add scaffolding for new functional tests

SevInf:
Based on yeoman (I know, basically a pre-historic tech in JS world) and a couple of templates, generates boilerplate for a
new functional tests. It's not much, but could help with the setup a tiny bit.

Looks like this:

Kapture 2022-06-01 at 15 09 15

### DIFF
--- a/packages/client/helpers/generator-test/index.ts
+++ b/packages/client/helpers/generator-test/index.ts
@@ -1,0 +1,43 @@
+import path from 'path'
+import Generator from 'yeoman-generator'
+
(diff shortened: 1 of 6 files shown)


packages/client/helpers/generator-test/index.ts
```diff
@@ -0,0 +1,44 @@
+'use strict'
```
aqrln, line 1: It should be emitted by TypeScript automatically since we don't have `"alwaysStrict": false"` in `tsconfig.json`

aqrln (approved):

danstarns (approved):

PR #13616, opened 2022-06-02
fix(client): Better type inference for boolean in select/include

### LINKED ISSUE #13571: Using boolean variables within select/include statements breaks type inference
### Bug description

As detailed [on Twitter](https://twitter.com/t3dotgg/status/1531361225241530368), using variables within a select or include object seems to _always infer existence_

```ts
export const getUserExample = async () => {
  const testBoolean = Math.random() > 0.5;

  const testUser = await prisma.user.findFirst({
    select: {
      id: testBoolean,
    },
  });

  // Typescript thinks this always exists, even though we may not be selecting ID
  const testUserId = testUser!.id;
};
```

### How to reproduce

```ts
export const getUserExample = async () => {
  const testBoolean = Math.random() > 0.5;

  const testUser = await prisma.user.findFirst({
    select: {
      id: testBoolean,
    },
  });

  // Typescript thinks this always exists, even though we may not be selecting ID
  const testUserId = testUser!.id;
};
```

Weirder yet, putting the two conditions under an 'if' (or in here a ternary) results in losing the type entirely

```ts
export const getUserTernaryExample = async () => {
  const testBoolean = Math.random() > 0.5;

  const testUser = testBoolean
    ? await prisma.user.findFirst({
        select: {
          id: true,
        },
      })
    : await prisma.user.findFirst({
        select: {
          id: false,
        },
      });

  // Typescript thinks id NEVER exists, the `false`y select prevents the truthy one from working
  const testUserId = testUser!.id;
};
```

Funny enough, this works if every condition returns *at least* one thing that is different from others
```typescript

export const getUserTernaryWtf = async () => {
  const testBoolean = Math.random() > 0.5;

  const testUser = testBoolean
    ? await prisma.user.findFirst({
        select: {
          id: true,
          name: false,
        },
      })
    : await prisma.user.findFirst({
        select: {
          id: false,
          name: true,
        },
      });

  // CORRECT type error
  // ERR: Property 'id' does not exist on type '{ id: string; } | { name: string; }'.
  const testUserId = testUser!.id;
};
```

### Expected behavior

```typescript
{id: string | undefined}
```

### Prisma information

Not needed

### Environment & setup

Not needed

### Prisma Version

```
3.10.0
```

PS-Soundwave:
Closes #13571

Still needs tests adjusted if the changes are accepted.


packages/client/tests/functional/issue-13571/tests.ts
```diff
@@ -0,0 +1,62 @@
...
+            id,
+          },
+          select: {
+            id: true, // we need to select atleast one field
+            bool: false,
+          },
+        })
+
+    // This is for typechecks
+    const casted: Boolean | undefined = testUser.bool
```
danstarns, line 60: This is currently failing with Typescript error: 

```
Message:
      Property 'bool' does not exist on type '{ id: string; }'.
    at (tests/functional/issue-13571/.generated/issue-13571 (provider=cockroachdb)/tests.ts:60:50)
```

packages/client/tests/functional/issue-13571/tests.ts
```diff
@@ -0,0 +1,62 @@
...
+        id,
+      },
+      select: {
+        id: true, // we need to select atleast one field
+        bool: testBoolean,
+      },
+    })
+
+    // This is for typechecks
+    const casted: Boolean | undefined = testUser.bool
```
danstarns, line 27: This casing works.

danstarns (commented):
@PS-Soundwave Thanks for your contribution. I have pushed snapshot updates and added a test suite so that we can assert that the types are correct. You can run this test by running: 

```sh
$ cd ./packages/client
$ pnpm run test:functional:code issue-13571 && pnpm run test:functional:types issue-13571
```

Please confirm the intended outcome using the tests, I am not sure if your fix is fixing the issue.

packages/client/tests/functional/issue-13571/tests.ts
```diff
@@ -0,0 +1,62 @@
...
+        id,
+      },
+      select: {
+        id: true, // we need to select atleast one field
+        bool: testBoolean,
+      },
+    })
+
+    // This is for typechecks - this will be asserted in the test:functional:types test suite
+    const casted: Boolean | undefined = testUser!.bool
```
danstarns, line 27: This casing works.

PR #13627, opened 2022-06-02
chore(cli, client): check that `prisma://` cannot be used with unsupported commands [DPGA, 4]

millsp:
In this PR we forbid attempts to use the data proxy via CLI commands that do not support it. The engine does throw an error as well but that we will only use in the generated Prisma Client. In the long term, the CLI commands should support `prisma://` and we can remove this code.

Affected CLI commands:
- studio
- db drop
- db execute
- db pull
- db push
- migrate deploy
- migrate dev
- migrate diff (some parameters only)
- migrate reset
- migrate resolve
- migrate status

Prev https://github.com/prisma/prisma/pull/13599
Next https://github.com/prisma/prisma/pull/13629

https://github.com/orgs/prisma/projects/70#card-82300438


packages/migrate/src/commands/DbExecute.ts
```diff
@@ -100,6 +100,8 @@ ${chalk.bold('Examples')}
       return this.help(args.message)
     }
 
+    await checkUnsupportedDataProxy('db execute', args, !args['--url'])
```
millsp, line 103: Only implicitly load the schema if `args['--url']` is not passed (see logic in the code).

packages/cli/src/Doctor.ts
```diff
@@ -64,6 +65,8 @@ ${chalk.bold('Examples')}
       return this.help(args.message)
     }
 
+    await checkUnsupportedDataProxy('doctor', args, true)
```
millsp, line 68: `doctor` => command name
`args` => command flags
`true` => implicitly load `schema.prisma`

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,76 @@
...
+ * @param command the cli command (eg. db push)
+ * @param args the cli command arguments
+ * @param implicitSchema if this command implicitly loads a schema
+ */
+async function _checkUnsupportedDataProxy(command: string, args: Args, implicitSchema: boolean) {
+  const argList = Object.entries(args)
+
+  // when the schema can be implicit, we use its default location
+  if (implicitSchema === true) {
+    args['--schema'] = (await getSchemaPath(args['--schema'])) ?? undefined
```
millsp, line 45: Similar logic to what we do in the cli when we need to get the location.

packages/migrate/src/commands/MigrateDiff.ts
```diff
@@ -144,6 +153,8 @@ ${chalk.bold('Examples')}
       return this.help(args.message)
     }
 
+    await checkUnsupportedDataProxy('migrate diff', args, false)
```
millsp, line 156: Migrate diff does not implicitly load the schema at it's default location (it needs --from and --to always), so we apply this to our logic too.

packages/migrate/src/commands/DbPull.ts
```diff
@@ -1,6 +1,7 @@
-import type { Command, IntrospectionSchemaVersion, IntrospectionWarnings } from '@prisma/sdk'
 import {
   arg,
+  checkUnsupportedDataProxy,
+  Command,
```
janpio, line 4: What is that change from `import type {Command ...` to `import { Command ...` that we have in multiple places? Relevant?

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,76 @@
...
+ * These are the cli args that we check the data proxy for. If in use
+ */
+const checkedArgs = {
+  '--url': true,
+  '--to-url': true,
+  '--from-url': true,
+  '--shadow-database-url': true,
+  '--schema': true,
+  '--from-schema-datamodel': true,
+  '--to-schema-datamodel': true,
```
janpio, line 17:
```suggestion
  // Directly contain connection string
  '--url': true,
  '--to-url': true,
  '--from-url': true,
  '--shadow-database-url': true,
  // Contain path to schema file, containing connection string (directly or via env var)
  '--schema': true,
  '--from-schema-datamodel': true,
  '--to-schema-datamodel': true,
```

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type Args = O.Optional<O.Update<typeof checkedArgs, any, string>>
+
+/**
+ * Get the message to display when a command is forbidden with a data proxy flag
+ * @param command the cli command (eg. db push)
+ * @returns
+ */
+export const forbiddenCmdWithDataProxyFlagMessage = (command: string) => `
+${chalk.green(`prisma ${command}`)} command is not yet available with ${chalk.green('prisma://')}.
```
janpio, line 28: This was my suggestion from the issue. Not sure if better.
```suggestion
Using the Data Proxy (connection URL starting with protocol ${chalk.green('prisma://')}) is not supported for this CLI command ${chalk.green(`prisma ${command}`)} yet. Please use a direct connection to your database for now.
```
I definitely like how it spells out "Data Proxy" and "connection URL" and "protocol". I don't expect our users to really understand that `prisma://` means = Data Proxy connection string (yet).

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,76 @@
...
+
+More information in our documentation:
+${link('https://pris.ly/d/data-proxy')}
```
janpio, line 31: We should probably have an anchor/section on that page that explicitly documents the current workaround (I think we advise people to have a different env var and then override `DATABASE_URL` before running `migrate` etc.

Easiest way to prepare for that is to create a new pris.ly link that we can then change to that headline/anchor when it exists.

Suggestions for naming? `/d/data-proxy-???`?

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
janpio: Does that mean `getSchemaPath` is some logic like 

1. default `prisma/schema.prisma`, 
2. otherwise what is in `--schema`, 
3. otherwise what is defined in `package.json` as `prisma.schema`"`

or something like that, making sure that the same schema file is used that the CLI command would use?

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,76 @@
...
+  const argList = Object.entries(args)
+
+  // when the schema can be implicit, we use its default location
+  if (implicitSchema === true) {
+    args['--schema'] = (await getSchemaPath(args['--schema'])) ?? undefined
+  }
+
+  for (const [argName, argValue] of argList) {
+    // for all the args that represent an url ensure data proxy isn't used
+    if (argName.includes('url') && argValue.includes('prisma://')) {
```
janpio, line 50: lol nice, I did not even realize those ones all contain `url` explicitly 👍

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,76 @@
...
+      console.error(forbiddenCmdWithDataProxyFlagMessage(command))
+      process.exit(1)
+    }
+
+    // for all the args that represent a schema path ensure data proxy isn't used
+    if (argName.includes('schema')) {
+      loadEnvFile(argValue, false)
```
janpio, line 57: Do we really load env files this way for the `migrate diff` command as well?

packages/migrate/src/commands/DbExecute.ts
janpio: If I understand this correctly ("Only try to load a schema file if no explicit `--url` is given" - correct?), I think the same logic should apply to `db pull` then. It also has an (undocumented) param `--url` that can be used to override the connection string from the schema file.

janpio (approved):
Chosen logic is elegant and makes sense to me. Usage in CLI command logic is consistent and looks fine.

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
millsp: Yes, we do https://github.com/prisma/prisma/blob/5aadcaf52c3832ad52d7f4749edc698217aa8db5/packages/migrate/src/commands/MigrateDiff.ts/#L200

janpio: Nice. That is also a good sanity check then if we got all the relevant params of `diff` - and it seems we did 👍

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,76 @@
...
+${link('https://pris.ly/d/data-proxy')}
+`
+
+/**
+ * Check that the data proxy cannot be used through the given args
+ * @param command the cli command (eg. db push)
+ * @param args the cli command arguments
+ * @param implicitSchema if this command implicitly loads a schema
+ */
+async function _checkUnsupportedDataProxy(command: string, args: Args, implicitSchema: boolean) {
```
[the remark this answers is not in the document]
millsp: Normally env loading (`dotenv`) uses the available env var over `.env` contents, but I can double check that.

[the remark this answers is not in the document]
janpio: No need, that is how it works indeed. Setting the env var in the terminal command makes sure the other env var with the `prisma://` value is overwritten - so this is safe.

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
millsp: /d/data-proxy-limitations with a page with all the current limitations + workarounds?

packages/migrate/src/commands/DbPull.ts
millsp: It's the linting + type imports that stopped working (not a big deal but something to attend).

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
janpio: That could work. `data-proxy-cli` could be a less judgy alternative 😆

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
millsp: Yes, definitely. CLI commands use both `getSchemaPathAndPrint` and `getSchemaPath`.
1. If --schema had a value, check if it exists and get its absolute path
2. If empty, then look in these locations 
	  ```ts
	  (await getSchemaPathFromPackageJson(opts.cwd)) ??
	  (await getRelativeSchemaPath(opts.cwd)) ??
	  (await resolveYarnSchema(opts.cwd))
	  ```

packages/cli/src/Doctor.ts
millsp: Implicit refers to the behavior of some of the cli commands. Some of them, will load the schema at it's default location, and some others won't, and some others will only do it conditionally. I did not want to alter or touch this logic, so it is reflected by a boolean instead.

packages/migrate/src/commands/DbExecute.ts
millsp: It looks like it is not the case anymore https://github.com/prisma/prisma/blob/5aadcaf52c3832ad52d7f4749edc698217aa8db5/packages/migrate/src/commands/DbPull.ts/#L167-L168, unless `.when` runs in reverse (but I doubt that). (Also arguably harder to read than what it was before). I might have overlooked something here...

janpio: I do not even pretend that I can still read that code 🤷 

I go from my experience, which is that `--url` works on `db pull` both with no schema present (then you get a schema file with that URL as datasource and the introspection result afterwards) and with a schema present (then the datasource.url value in there is overwritten with what is provided to the CLI).

I think the command describes my understanding of the behavior: https://github.com/prisma/prisma/blob/5aadcaf52c3832ad52d7f4749edc698217aa8db5/packages/migrate/src/commands/DbPull.ts#L155-L162

jkomyno:
Three main things aren't really convincing me:
- we're hiding `process.exit(1)` in an internal function, imho this should be explicitly called for each command to give our devs visibility about the execution flow when they're reading a Prisma command's code
- every possible error in `_checkUnsupportedDataProxy` is suppressed, including potential panics in `getConfig`
- no tests are provided

millsp:
> we're hiding process.exit(1) in an internal function, imho this should be explicitly called for each command to give our devs visibility about the execution flow when they're reading a Prisma command's code
I don't understand, how would you error then if you have to check that the thing in use has `prisma://`?

> every possible error in _checkUnsupportedDataProxy is suppressed, including potential panics in getConfig
That's voluntary, I made the choice to only error for `prisma://`, such errors should be handled by the cli (and probably will). I especially don't want to alter the behavior of the current cli implementation.

> no tests are provided
Correct, I'm working on adding those in the follow up PR

tomhoule (approved):
looks good to me :+1:

packages/migrate/src/commands/DbExecute.ts
janpio: @jkomyno recently rewrote that code, so should be able to confirm.

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
millsp: `data-proxy-cli` will be.

janpio: I will set up the redirect on pris.ly 👍

janpio: Done: https://pris.ly/d/data-proxy-cli

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (argName.includes('url') && argValue.includes('prisma://')) {
+      console.error(forbiddenCmdWithDataProxyFlagMessage(command))
+      process.exit(1)
+    }
+
+    // for all the args that represent a schema path ensure data proxy isn't used
+    if (argName.includes('schema')) {
+      loadEnvFile(argValue, false)
+
+      const datamodel = await fs.promises.readFile(argValue, 'utf-8')
```
aqrln, line 64: This didn't age well 😂

image

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -0,0 +1,82 @@
...
+        process.exit(1)
+      }
+    }
+  }
+}
+
+export async function checkUnsupportedDataProxy(command: string, args: Args, implicitSchema: boolean) {
+  try {
+    await _checkUnsupportedDataProxy(command, args, implicitSchema)
+  } catch (_) {}
```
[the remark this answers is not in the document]
aqrln: Although I'd probably prefer to at least log the error here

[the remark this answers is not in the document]
aqrln: But idk, maybe all the cases where an error may occur here are already covered with better error messages later in all commands

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
millsp: 🤣

millsp: https://www.notion.so/prismaio/Technical-Debt-TS-ORM-c199f16035304e2db13793f176098c24?p=7b86980cf7c5424fb86ef6e74e25787c

packages/sdk/src/cli/checkUnsupportedDataProxy.ts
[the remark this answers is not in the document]
millsp: Yes, it either goes again through getConfig, or the engine. I don't want to alter the behavior in any way, this must stay purely additional.

aqrln (approved):

PR #13656, opened 2022-06-06
fix(client): allow prisma.empty to $executeRaw & $queryRaw

### LINKED ISSUE #11233: Cannot pass Prisma.empty to $executeRaw function
### Bug description

We have a function which generates raw queries based on its input. then the generated query will be passed to `$executeRaw` to run it. in some conditions, the function may return `Prisma.empty`. In this case Prisma throws a PrismaClientValidationError 

we had no problems in version 2.30.3, This happened after upgrading "prisma" & "@prisma/client" version  to 3.8.1 

### How to reproduce

`await db.$executeRaw(Prisma.empty);`

throws this error: 
```
    PrismaClientValidationError: `$executeRaw` is a tag function, please use it like the following:
    const result = await prisma.$executeRaw`UPDATE User SET cool = ${true} WHERE email = ${'user@email.com'};`
    
    Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access#executeraw
```

### Expected behavior

to run it without any errors. preferably don't issue any queries if the input is `Prisma.empty`

### Prisma information

N/A

### Environment & setup

- OS: Debian
- Database: PostgreSQL
- Node.js version: 14.17.6

### Prisma Version

```
10.1.0
```

danstarns:
Closes: https://github.com/prisma/prisma/issues/11233

Having issues passing an empty sql stmt to the db and so return null when its empty.


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -680,7 +680,12 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
      * @returns
      */
     $executeRaw(query: TemplateStringsArray | sqlTemplateTag.Sql, ...values: any[]) {
-      return createPrismaPromise((txId, lock, otelCtx) => {
+      return createPrismaPromise(async (txId, lock, otelCtx): Promise<unknown> => {
+        const isInstanceOfSQL = query.constructor.name === 'Sql'
+        if (isInstanceOfSQL && (query as { sql: string }).sql === '') {
+          return null
+        }
```
aqrln, line 687: Should we maybe make `isInstanceOfSQL` a type guard?

Something like

```ts
function isSql(value: unknown): value is sqlTemplateTag.Sql {
  return value.constructor.name === 'Sql'
}
```

and then, here

```ts
if (isSql(query) && query.sql === '') {
  return null
}
```

No strong opinion though, if you don't find it more readable, then  it's also fine as it is.

aqrln (commented):
Looks good overall, just a couple of comments

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -680,7 +680,12 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
+
         if ((query as TemplateStringsArray).raw || (query as sqlTemplateTag.Sql).sql) {
```
millsp, line 689:
```suggestion
        if ((query as TemplateStringsArray).raw !== undefined || (query as sqlTemplateTag.Sql).sql !== undefined) {
```

This should be enough.

millsp (changes requested):
The same problem applies to `$queryRaw`, I believe.

packages/client/src/runtime/getPrismaClient.ts
danstarns: Sure I added it!

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -680,8 +681,12 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
      */
     $executeRaw(query: TemplateStringsArray | sqlTemplateTag.Sql, ...values: any[]) {
-      return createPrismaPromise((txId, lock, otelCtx) => {
-        if ((query as TemplateStringsArray).raw || (query as sqlTemplateTag.Sql).sql) {
+      return createPrismaPromise(async (txId, lock, otelCtx): Promise<unknown> => {
+        if (isSql(query) && query.sql === '') {
+          return null
+        }
+
+        if ((query as TemplateStringsArray).raw !== undefined || (query as sqlTemplateTag.Sql).sql !== undefined) {
```
aqrln, line 689: I wonder if the previous (the newly added) check is now necessary or just adding `!== undefined` here is enough to fix the bug.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -856,7 +861,11 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
      * @returns
      */
     $queryRaw(query: TemplateStringsArray | sqlTemplateTag.Sql, ...values: any[]) {
-      return createPrismaPromise((txId, lock, otelCtx) => {
+      return createPrismaPromise(async (txId, lock, otelCtx): Promise<unknown> => {
+        if (isSql(query) && query.sql === '') {
+          return null
+        }
+
         if ((query as TemplateStringsArray).raw || (query as sqlTemplateTag.Sql).sql) {
```
aqrln, line 869: This should also probably be changed the same way

```suggestion
        if ((query as TemplateStringsArray).raw !== undefined || (query as sqlTemplateTag.Sql).sql !== undefined) {
```

packages/client/src/runtime/getPrismaClient.ts
millsp: I think the check can now be removed

danstarns: This check needs to be here because we cannot send an empty string to the DB. So the newly added check returns null if the string is empty.

```
Invalid `prisma.$executeRaw()` invocation:


      Raw query failed. Code: `1065`. Message: `Query was empty`

      48980 |       message = this.sanitizeMessage(message);
      48981 |       if (e.code) {
    > 48982 |         throw new PrismaClientKnownRequestError(message, e.code, this.client._clientVersion, e.meta);
            |               ^
      48983 |       } else if (e.isPanic) {
      48984 |         throw new PrismaClientRustPanicError(message, this.client._clientVersion);
      48985 |       } else if (e instanceof PrismaClientUnknownRequestError) {

      at RequestHandler.request (../../runtime/index.js:48982:15)
      at Proxy._request (../../runtime/index.js:49888:18)
      at Object.<anonymous> (empty-to-raw/tests.ts:11:22)
```

millsp (changes requested):
Let's fix https://github.com/prisma/prisma/pull/13656/files#r902878957 and merge 🚀

packages/client/src/runtime/getPrismaClient.ts
danstarns: Sure it could be seen as breaking! I have reverted the changes I made and instead now assert the error could be thrown or the `0` or the `[]`. My reason for adding the null return is because it's deterministic, easier to reason about/document plus the issue expected 'to run it without any errors. preferably don't issue any queries if the input is Prisma.empty'.

packages/client/tests/functional/empty-to-raw/tests.ts
```diff
@@ -0,0 +1,90 @@
...
+        case 'sqlserver':
+          expect(result).toEqual([])
+          break
+
+        case 'mysql':
+          expect((result as Error).message).toContain('Raw query failed. Code: `1065`. Message: `Query was empty`')
+          break
+
+        default:
+          throw new Error('invalid prodiver')
```
millsp, line 80:
```suggestion
          throw new Error('invalid provider')
```

danstarns: good catch!

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -693,8 +693,8 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
      * @returns
      */
     $executeRaw(query: TemplateStringsArray | sqlTemplateTag.Sql, ...values: any[]) {
-      return createPrismaPromise((txId, lock, otelCtx) => {
-        if ((query as TemplateStringsArray).raw || (query as sqlTemplateTag.Sql).sql) {
+      return createPrismaPromise(async (txId, lock, otelCtx): Promise<unknown> => {
```
danstarns, line 696:
```suggestion
      return createPrismaPromise(async (txId, lock, otelCtx) => {
```

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -693,8 +693,8 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
      * @returns
      */
     $executeRaw(query: TemplateStringsArray | sqlTemplateTag.Sql, ...values: any[]) {
-      return createPrismaPromise((txId, lock, otelCtx) => {
-        if ((query as TemplateStringsArray).raw || (query as sqlTemplateTag.Sql).sql) {
+      return createPrismaPromise(async (txId, lock, otelCtx) => {
```
danstarns, line 696:
```suggestion
      return createPrismaPromise((txId, lock, otelCtx) => {
```

packages/client/tests/functional/empty-to-raw/tests.ts
```diff
@@ -0,0 +1,90 @@
...
+          expect(result).toEqual([])
+          break
+
+        case 'cockroachdb':
+          expect(result).toEqual([])
+          break
+
+        case 'sqlserver':
+          expect(result).toEqual([])
+          break
```
Jolg42, line 73:
```suggestion
        case 'postgresql':
        case 'cockroachdb':
        case 'sqlserver':
          expect(result).toEqual([])
          break
```
Can be simplified

packages/client/tests/functional/empty-to-raw/tests.ts
```diff
@@ -0,0 +1,90 @@
...
+          expect(result).toEqual(0)
+          break
+
+        case 'cockroachdb':
+          expect(result).toEqual(0)
+          break
+
+        case 'sqlserver':
+          expect(result).toEqual(0)
+          break
```
Jolg42, line 36:
```suggestion
        case 'postgresql':
        case 'cockroachdb':
        case 'sqlserver':
          expect(result).toEqual(0)
          break
```

millsp (approved):

PR #13659, opened 2022-06-06
chore(client): reset the generated folder before generate [DPGA] (bonus)

millsp:
We don't actually copy the engine files anymore on generate (`--data-proxy`), but as it turns out, they can be there anyways. Why? When you start fresh on your project, the postinstall script will trigger an automatic call to the standard `prisma generate`. If you use `prisma generate --data-proxy` right after, the Query Engine was already copied to the generated folder... since we only overwrite files.


packages/client/helpers/build.ts
```diff
@@ -73,6 +73,14 @@ const generatorBuildConfig: BuildOptions = {
   bundle: true,
 }
 
+// we define the config for postinstall
+const postinstallBuildConfig: BuildOptions = {
+  name: 'postinstall',
+  entryPoints: ['src/scripts/postinstall.ts'],
```
millsp, line 79: Postinstall is now a ts file because we want to bundle it.

packages/client/src/generation/defaults/defaultBrowserIndex.ts
```diff
@@ -0,0 +1,13 @@
+export const defaultBrowserIndex = `class PrismaClient {
```
millsp, line 1: These aren't files that we copy anymore, but bundle-able contents (used by the generator and postinstall).

packages/client/src/generation/generateClient.ts
```diff
@@ -212,6 +216,7 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
     throw new DenylistError(message)
   }
 
+  await deleteOutputDir(finalOutputDir)
```
millsp, line 219: We safely clear the previous generation folder.

packages/client/src/scripts/postinstall.ts
```diff
@@ -1,36 +1,34 @@
-// @ts-check
-const childProcess = require('child_process')
-const { promisify } = require('util')
-const fs = require('fs')
-const path = require('path')
-const c = require('./colors')
+import Debug from '@prisma/debug'
```
millsp, line 1: This file logic is almost intact, it's mostly been converted to TS so that we can bundle it.

millsp:
Replaced by https://github.com/prisma/prisma/pull/13702

PR #13666, opened 2022-06-07
test: Port interactive transaction tests to new setup

SevInf:
Close prisma/client-planning#23


packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
```diff
@@ -0,0 +1,121 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`interactive-transactions (mongodb) () batching rollback 1`] = `
```
SevInf, line 3: All these snapshots have different output per provider, so I had to convert inline snapshots into external ones.

packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -1,15 +1,20 @@
...
+// @ts-ignore this is just for type checks
+declare let PrismaClient: typeof import('@prisma/client').PrismaClient
+
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
 const testIf = (condition: boolean) => (condition ? test : test.skip)
 
-let PrismaClient, prisma
+testMatrix.setupTestSuite(({ provider }) => {
+  beforeEach(async () => {
+    await prisma.user.deleteMany()
```
danstarns, line 15: After each test, we drop the DB https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts#L62, and foreach test we setup a new db, does this need to be done?

packages/client/tests/functional/_utils/setupFilesAfterEnv.ts
```diff
@@ -3,5 +3,7 @@ process.env.DATABASE_URI_sqlite = 'file:dev.db'
 process.env.DATABASE_URI_mongodb = process.env.TEST_MONGO_URI
 process.env.DATABASE_URI_postgresql = process.env.TEST_POSTGRES_URI
 process.env.DATABASE_URI_mysql = process.env.TEST_MYSQL_URI
+process.env.DATABASE_URI_cockroachdb = process.env.TEST_COCKROACH_URI
```
danstarns, line 6: Nice work on adding these, I noticed however you arent testing either, can we add coverage for both of these providers in this test?

packages/client/tests/functional/interactive-transactions/tests.ts
danstarns: Would using unique values help, such as using something like faker?

SevInf: I think it might, let me try that

danstarns: Thanks for checking!

SevInf:
Going to draft it until all CI issues are fixed

SevInf:
Needs prisma/client-planning#65 to proceed

packages/client/tests/functional/interactive-transactions/prisma/_schema.ts
```diff
@@ -0,0 +1,38 @@
...
+    provider = "${provider}"
+    url      = env("DATABASE_URI_${provider}")
+  }
+  
+
+  // / User model comment
+  model User {
+    id    ${id}
+    email String  @unique
+    // / name comment
```
millsp, line 21: I think we can clean these comments every time you see them. These don't work in this form, anyways.

millsp (approved):

PR #13682, opened 2022-06-07
test(client): higher timeout on exhaustive tests

danstarns:
Increases timeout on long-running exhaustive tests.

~~My observation is that these two tests **bearly** ever pass on both of the `no-docker` test suites, here I skip them based on an environment variable that is passed in when executing those tests. I've done this because I believe these tests to be slowing down the feedback loop and wasting resources.~~

Example error

```
Summary of all failing tests
FAIL src/__tests__/integration/happy/exhaustive-schema-mongo/dmmf-types.test.ts (41.611 s)
  ● dmmf-types
    thrown: "Exceeded timeout of 40000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
       8 |  * Makes sure, that the actual dmmf value and types are in match
       9 |  */
    > 10 | test('dmmf-types', async () => {
         | ^
      11 |   const datamodel = fs.readFileSync(path.join(__dirname, 'schema.prisma'), 'utf-8')
      12 |   const dmmf = await getDMMF({
      13 |     datamodel,
      at Object.<anonymous> (src/__tests__/integration/happy/exhaustive-schema-mongo/dmmf-types.test.ts:10:1)
FAIL src/__tests__/integration/happy/exhaustive-schema/dmmf-types.test.ts (41.594 s)
  ● dmmf-types
    thrown: "Exceeded timeout of 40000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
       8 |  * Makes sure, that the actual dmmf value and types are in match
       9 |  */
    > 10 | test('dmmf-types', async () => {
         | ^
      11 |   const datamodel = fs.readFileSync(path.join(__dirname, 'schema.prisma'), 'utf-8')
      12 |   const dmmf = await getDMMF({
      13 |     datamodel,
      at Object.<anonymous> (src/__tests__/integration/happy/exhaustive-schema/dmmf-types.test.ts:10:1)
Test Suites: 2 failed, 43 skipped, 147 passed, 149 of 192 total
Tests:       2 failed, 258 skipped, 523 passed, 783 total
Snapshots:   434 passed, 434 total
Time:        659.213 s
Ran all test suites.
```

Here is a random select set of workflows where this error occurs: 
1. https://github.com/prisma/prisma/runs/6780773681?check_suite_focus=true
2. https://github.com/prisma/prisma/runs/6763730976?check_suite_focus=true
3. https://github.com/prisma/prisma/runs/6780764561?check_suite_focus=true
4. https://github.com/prisma/prisma/runs/6758802530?check_suite_focus=true

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -595,7 +595,6 @@ jobs:
           echo "TEST_MONGO_URI_MIGRATE=mongodb://localhost:27017/tests-migrate" >> $GITHUB_ENV
           echo "PRISMA_CLI_QUERY_ENGINE_TYPE=${{ matrix.queryEngine }}" >> $GITHUB_ENV
           echo "PRISMA_CLIENT_ENGINE_TYPE=${{ matrix.queryEngine }}" >> $GITHUB_ENV
-          echo "TEST_SKIP_EXHAUSTIVE=true" >> $GITHUB_ENV
 
       - name: Set up PostgreSQL
         run: bash .github/workflows/scripts/setup-postgres.sh
@@ -767,7 +766,6 @@ jobs:
           echo "TEST_MONGO_URI_MIGRATE=mongodb://localhost:27017/tests-migrate" >> $GITHUB_ENV
           echo "PRISMA_CLI_QUERY_ENGINE_TYPE=${{ matrix.queryEngine }}" >> $GITHUB_ENV
           echo "PRISMA_CLIENT_ENGINE_TYPE=${{ matrix.queryEngine }}" >> $GITHUB_ENV
-          echo "TEST_SKIP_EXHAUSTIVE=true" >> $GITHUB_ENV
 
       - name: Set up PostgreSQL
         run: bash .github/workflows/scripts/setup-postgres.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,669 +74,669 @@ jobs:
     steps:
       - uses: fkirc/skip-duplicate-actions@v4.0.0
 
-  # #
-  # # Linting
-  # #
-  # lint:
-  #   timeout-minutes: 7
-  #   runs-on: ubuntu-latest
-
-  #   steps:
-  #     - uses: actions/checkout@v3
-
-  #     - uses: pnpm/action-setup@v2.2.2
-  #       with:
-  #         version: 6
-
-  #     - uses: actions/setup-node@v3
-  #       with:
-  #         cache: 'pnpm'
-  #         node-version: '12'
-
-  #     # https://github.com/actions/toolkit/blob/master/docs/commands.md#problem-matchers
-  #     # Matchers are added in setup-node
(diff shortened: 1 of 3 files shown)


danstarns:
Never mind, this won't work because of obsolete snapshots. Opened an issue here: 

https://github.com/facebook/jest/issues/12921

janpio:
Are the failing because something loops, or just because something takes a long time?

danstarns:
> exhaustive

Something takes a long time, the exhaustive tests are testing a very large schema, and timing out, in this env, when generating and JSON stringifying the DMMF.

danstarns:
> Are the failing because something loops, or just because something takes a long time?

Something takes a long time, the exhaustive tests are testing a very large schema, and timing out, in this env, when generating and JSON stringifying the DMMF.

Jolg42:
@danstarns Why not "just" increasing the timeout?

Because we want to run these tests on macos and windows since they are "exhaustive" maybe?

It looks like they are slow with a good machine (16s) but on a bad machine (shared macos on GitHub Actions) it apparently can go over our 40s timeout.

So If we put a 80s timeout here we should be good? (or less)

danstarns:
> @danstarns Why not "just" increasing the timeout?
> 
> Because we want to run these tests on macos and windows since they are "exhaustive" maybe?
> 
> It looks like they are slow with a good machine (16s) but on a bad machine (shared macos on GitHub Actions) it apparently can go over our 40s timeout.
> 
> So If we put a 80s timeout here we should be good? (or less)

This is a much better idea thanks! I was just button bashing because of being frustrated with how many times they failed.

Jolg42 (approved):

Jolg42 (changes requested):
Could you do that for the no-docker only? with something like this at the top of the file it should work (copy pasted from another test, maybe we can have a helper one day)

```js
const isMacOrWindowsCI = Boolean(process.env.CI) && ['darwin', 'win32'].includes(process.platform)
if (isMacOrWindowsCI) {
  jest.setTimeout(80_000)
}
```

danstarns:
> ```js
> const isMacOrWindowsCI = Boolean(process.env.CI) && ['darwin', 'win32'].includes(process.platform)
> if (isMacOrWindowsCI) {
>   jest.setTimeout(80_000)
> }
> ```

Sure, I'll add this to the top of the file. I went to write the helper:

```ts
export function testCITimeout({
  platforms,
  timeout,
  defaultTimeout,
}: {
  platforms: NodeJS.Platform[]
  timeout: number
  defaultTimeout?: number
}) {
  if (!process.env.CI) {
    return
  }

  const isIncluded = platforms.includes(process.platform)

  if (isIncluded) {
    jest.setTimeout(timeout)
  } else if (defaultTimeout) {
    jest.setTimeout(defaultTimeout)
  }
}
```

Then realized that it needs to be used in all packages! So will just copy and paste it and come back to the helper another time.

Jolg42 (approved):

danstarns:
related: https://github.com/prisma/prisma/pull/13572

PR #13688, opened 2022-06-08
chore: Update CODEOWNERS with client-team-ts

SevInf:
We want to try out automatic reviews assignments for our teams. For
that we need to switch code owners from individuals to a team.


Jolg42:
Note @SevInf it currently errors with 
```
Unknown owner on line 12: make sure the team @prisma/client-team-ts exists and has write access to the repository

/packages/engine-core/       @prisma/client-team-ts
[...]
```

SevInf:
@Jolg42 now should be good.

PR #13702, opened 2022-06-08
fix(client): reset generated folder before generate

millsp:
When generating with  `--data-proxy` we don't actually copy the engine files anymore on generate, but as it turns out, they can be there anyways. Why? When you start fresh on your project, the postinstall script will trigger an automatic call to the standard `prisma generate`. If you use `prisma generate --data-proxy` right after, the Query Engine was already copied to the generated folder... because we only overwrite files.


packages/cli/helpers/build.ts
```diff
@@ -20,10 +20,8 @@ const cliLifecyclePlugin: esbuild.Plugin = {
     // we only do this for the first one of the builds
     if (build.initialOptions?.format === 'esm') return
 
-    build.onStart(async () => {
-      // provide a copy of the client for studio to work
-      await run('node -r esbuild-register ./helpers/copy-prisma-client.ts')
-    })
+    // provide a copy of the client for studio to work
+    build.onStart(copyPrismaClient)
```
millsp, line 24: This simple change results in much faster builds for the cli.

packages/client/src/generation/defaults/defaultBrowserIndex.ts
```diff
@@ -0,0 +1,13 @@
...
+        \`@prisma/client did not initialize yet. Please run "prisma generate" and try to import it again.
+  In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues/new\`,
+      )
+    }
+  }
+  
+  module.exports = {
+    PrismaClient,
+  }
+`
```
millsp, line 13: Instead of having files that we copy around, we have file contents that we write to disk.

packages/client/src/generation/generateClient.ts
```diff
@@ -110,6 +112,8 @@ export async function buildClient({
   fileMap['index.js'] = await JS(nodeTsClient, false)
   fileMap['index.d.ts'] = await TS(nodeTsClient)
   fileMap['index-browser.js'] = await BrowserJS(nodeTsClient)
+  fileMap['edge.js'] = defaultEdgeIndex
+  fileMap['edge.d.ts'] = defaultIndexDts
```
millsp, line 116: Re-use the default file contents to write to disk.

packages/client/src/generation/generateClient.ts
```diff
@@ -439,3 +444,16 @@ async function getGenerationDirs({ testMode, runtimeDirs, generator, outputDir }
...
+ * Attempts to delete the output directory.
+ * @param finalOutputDir
+ */
+async function deleteOutputDir(finalOutputDir: string) {
+  try {
+    // we want to make sure that if we delete, we delete the right directory
+    if (require(`${finalOutputDir}/package.json`).name === '.prisma/client') {
+      await fs.promises.rmdir(finalOutputDir, { recursive: true })
+    }
+  } catch {}
```
millsp, line 458: We want to make sure that we don't erase wrong things by accident. For this we look that we can find `.prisma/client` in a `package.json` in the output dir, and we only delete the files that we know about/that we generated (TODO).

packages/client/src/scripts/postinstall.ts
```diff
@@ -0,0 +1,393 @@
+import Debug from '@prisma/debug'
```
millsp, line 1: In there we are now taking advantage of TypeScript, and we compile the postinstall script with esbuild. The diff show better in the original commit [`ef49571` (#13702)](https://github.com/prisma/prisma/pull/13702/commits/ef49571235807986558560e025b401813554bc64)

PR #13705, opened 2022-06-08
chore: drop Node.js 12 and modernize the toolchain

aqrln:
- Stop testing on Node.js 12, start testing on Node.js 18
  (i.e., test on Node.js 14, 16, and 18).

*Before the changes:*

| Workflow                     | Node.js Version    |
| ---------------------------- | ------------------ |
| Buildkite (Tests):           | 14                 |
| Buildkite (Release):         | 12                 |
| GitHub Actions (Tests):      | 12, 16             |
| GitHub Actions (Other Jobs): | 12                 |

*After the changes:*

| Workflow                     | Node.js Version    |
| ---------------------------- | ------------------ |
| Buildkite (Tests):           | 14                 |
| Buildkite (Release):         | 14                 |
| GitHub Actions (Tests):      | 16, 18             |
| GitHub Actions (Other Jobs): | 16                 |

- Update the example commands in CONTRIBUTING.md:

  - Update the nvm installation command.

  - Update the Node.js installation command to install Node.js 16 since
    the text recommends using Active LTS and Node.js 14 is not Active
    LTS anymore (it's already Maintenance LTS and on track to becoming
    EOL in 2023).

- Update the VSCode Dev Container to use the current Active LTS, update
  the comments there.

- Prepare to un-pin ts-node in a follow-up PR.
  Update ts-node in `packages/{client,debug}` from 28.0.3 to 28.0.4
  because `ts-node@28.0.3` has a peer dependency on `@types/jest@27`
  and new pnpm now errors on incorrect peer dependency instead of just
  printing a warning.

Ref: https://github.com/prisma/client-planning/issues/19

### DIFF
--- a/.buildkite/publish/docker-compose.yml
+++ b/.buildkite/publish/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.7'
 services:
   app:
-    image: node:12-buster
+    image: node:14-buster
     volumes:
       - ../../:/app
     working_dir: /app
--- a/.buildkite/publish/run.sh
+++ b/.buildkite/publish/run.sh
@@ -3,7 +3,7 @@
 set -ex
 
 # Install pnpm
-npm i --silent -g pnpm --unsafe-perm # TODO: is this unsafe-perm needed?
+npm i --silent -g pnpm@6 --unsafe-perm # TODO: is this unsafe-perm needed?
 
 # Install packages
 pnpm i
--- a/.buildkite/publish/run.sh
+++ b/.buildkite/publish/run.sh
@@ -3,7 +3,7 @@
 set -ex
 
 # Install pnpm
-npm i --silent -g pnpm@6 --unsafe-perm # TODO: is this unsafe-perm needed?
+npm i --silent -g pnpm --unsafe-perm # TODO: is this unsafe-perm needed?
 
 # Install packages
 pnpm i
--- a/.buildkite/test/run.sh
+++ b/.buildkite/test/run.sh
@@ -16,7 +16,7 @@ if [ "$BUILDKITE_PARALLEL_JOB" = "1" ]; then
 fi
 
 # Install pnpm
-npm i --silent -g pnpm@6 --unsafe-perm # TODO: is this unsafe-perm needed?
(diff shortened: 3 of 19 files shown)


SevInf:
Should we somehow enforce minimum `pnpm` major version, maybe in `scripts/only-allow-pnpm`? Right now pnpm 6 and 7 produce different versions of lockfiles, mutually warn about it and update them.

Jolg42 (changes requested):
@SevInf Good point we should update the engines in the root package.json, currently
```
  "engines": {
    "node": ">=12.6",
    "pnpm": ">=6.14.1"
  },
```

millsp (changes requested):
I think we should update this if package some managers just emit a warning on bad node engine versions:
https://github.dev/prisma/prisma/blob/0d6028db17995bfdc6111f23e065bd1829556450/packages/cli/scripts/preinstall.js#L45-L48

aqrln:
I get the same linter errors locally as on Buildkite. What's weird is why they don't happen on GitHub Actions, and didn't happen on Buildkite when running ESLint with Node.js 12.

Jolg42:
@aqrln probably depends on Node.js version, should be easy to remove `async` from the function definition I think

aqrln:
> probably depends on Node.js version

That was my first thought to, but it would be weird for the rule to work on Node.js 14 (Buildkite) and Node.js 18 (locally) but not on Node.js 16 (GH Actions). But it's easy to check if that's the case by switching the version locally.

aqrln:
Same errors with Node.js 16 and even Node.js 12 locally :/

Something is wrong with our CI workflows for linting.

Jolg42:
@aqrln but we run the same command `pnpm run lint` for both so let's ignore that for now I guess

aqrln:
@Jolg42 yeah it's definitely out of scope here, but something to investigate

Jolg42 (approved):

Jolg42:
Do you know if someone is already planning to update the docs like https://www.prisma.io/docs/reference/system-requirements?

Maybe open a PR there

millsp (approved):

Jolg42:
@aqrln could be that the problems in test failures are from pnpm v7, maybe separate this into another PR

PR #13706, opened 2022-06-08
fix(engine-core): Fix race conditon during panic in BinaryEngine

### LINKED ISSUE #13621: Triggering a panic makes the client unusable if the exception is handled in user code
See this comment: https://github.com/prisma/prisma/issues/10586#issuecomment-1018996620

I can reproduce that the client becomes unusable after the panic if I catch the error in user code outside Prisma (which will happen in most scenarios; for example, when the error is caught by the http framework in the request handler to return 500).

Expected behavior: query engine is restarted after the panic on subsequent requests and the client is usable.
Actual behavior: all subsequent calls to the Prisma Client re-throw the same error and the application needs to be restarted manually.

SevInf:
Rough sequence of events:
1. We start the engine. `startPromise` will now contain the promise of
`startInternal` and will stay non-null until child process exits.
2. We make a request and receive a panic. We store the error, child
   process either dies by itself or gets killed by TS code in
   `handlePanic` eventually
3. We make a second request. It supposed to restart the server. However,
   on we don't really wait until child process finished at step 2 and
   reset of `startPromise` and happens only in `exit` handler of the
   process. So, in case engine process have not died yet by the time of
   the second request, it won't restart. Request will fail and we
   re-throw the same panic we had on the step 2, since reset of
   `lastPanic` happens only after successful restart.

`BinaryEngine` class already has a way to wait until process finishes
before restart: it waits on `stopPromise` if it is set. Before it was
only set after explicit `stop` call. Now, we also set it on panic, which
fixes race condition.

Fix #13621

### DIFF
--- a/packages/engine-core/src/binary/BinaryEngine.ts
+++ b/packages/engine-core/src/binary/BinaryEngine.ts
@@ -4,7 +4,7 @@ import type { ConnectorType, GeneratorConfig } from '@prisma/generator-helper'
 import type { Platform } from '@prisma/get-platform'
 import { getPlatform, platforms } from '@prisma/get-platform'
 import chalk from 'chalk'
-import type { ChildProcessByStdio } from 'child_process'
+import type { ChildProcess, ChildProcessByStdio } from 'child_process'
 import { spawn } from 'child_process'
 import EventEmitter from 'events'
 import execa from 'execa'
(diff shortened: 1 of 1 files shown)


SevInf:
I'd like to add a test for this, but reliably reproducing it requires quite a specific set of unfortunate circumstances and I am not sure it is possible to do it in test environment without extensive `BinaryEngine` refactoring

aqrln (approved):

PR #13712, opened 2022-06-08
test: Allow to conditionally skip type tests in new setup

SevInf:
Syntax: `@ts-error-if: condition` where condition can be any JS boolean
expression. It may also use any properties from suite config. During the
tests it will then turn into `// @ts-expect-error` commend if condition
evaluates to truthy value and stay the as is otherwise.

Fix prisma/client-planning#65

### DIFF
--- a/packages/client/tests/functional/_example/tests.ts
+++ b/packages/client/tests/functional/_example/tests.ts
@@ -50,6 +50,11 @@ testMatrix.setupTestSuite(
       expect(schemaString).toContain('datasource')
       expect(schemaString).toContain('model')
     })
+
+    testIf(suiteConfig.provider !== 'mongodb')('conditional @ts-test-if', async () => {
+      // @ts-test-if: provider !== 'mongodb'
+      await prisma.$queryRaw`SELECT 1;`
+    })
   },
   // Use `optOut` to opt out from testing the default selected providers
   // otherwise the suite will require all providers to be specified.
--- a/packages/client/tests/functional/_utils/setupFilesAfterEnv.ts
+++ b/packages/client/tests/functional/_utils/setupFilesAfterEnv.ts
@@ -5,4 +5,7 @@ process.env.DATABASE_URI_postgresql = process.env.TEST_POSTGRES_URI
 process.env.DATABASE_URI_mysql = process.env.TEST_MYSQL_URI
 process.env.DATABASE_URI_cockroachdb = process.env.TEST_COCKROACH_URI
 
+globalThis.testIf = (condition: boolean) => (condition ? test : test.skip)
+globalThis.describeIf = (condition: boolean) => (condition ? describe : describe.skip)
+
 export {}
--- a/packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
+++ b/packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
@@ -1,6 +1,7 @@
 import crypto from 'crypto'
 import fs from 'fs-extra'
 import path from 'path'
+import { Script } from 'vm'
 
 import { DbDrop } from '../../../../migrate/src/commands/DbDrop'
 import { DbPush } from '../../../../migrate/src/commands/DbPush'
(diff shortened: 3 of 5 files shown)


packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -18,21 +19,37 @@ export async function setupTestSuiteFiles(suiteMeta: TestSuiteMeta, suiteConfig:
...
 
-async function copyWithImportsAdjust(from: string, to: string): Promise<void> {
+async function copyPreprocessed(from: string, to: string, suiteConfig: TestSuiteConfig): Promise<void> {
   // we adjust the relative paths to work from the generated folder
   const contents = await fs.readFile(from, 'utf8')
-  const newContents = contents.replace(/'..\//g, "'../../../")
+  const newContents = contents
+    .replace(/'..\//g, "'../../../")
+    .replace(/\/\/\s*@ts-error-if:(.+)/g, (match, condition) => {
+      if (evaluateMagicComment(condition, suiteConfig)) {
```
millsp, line 37: I love this sorcery 🤯

packages/client/tests/functional/_example/tests.ts
```diff
@@ -6,6 +6,7 @@ import testMatrix from './_matrix'
 // @ts-ignore this is just for type checks
 declare let prisma: import('@prisma/client').PrismaClient
 
+const testIf = (condition: boolean) => (condition ? test : test.skip)
```
millsp, line 9: I know this is a short one, but can we make it a util too?

packages/client/tests/functional/_example/tests.ts
```diff
@@ -50,6 +51,11 @@ testMatrix.setupTestSuite(
       expect(schemaString).toContain('datasource')
       expect(schemaString).toContain('model')
     })
+
+    testIf(suiteConfig.provider !== 'mongodb')('conditional @ts-expect-error', async () => {
+      // @ts-error-if: provider === 'mongodb'
```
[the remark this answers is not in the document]
SevInf: That what i started with, but unfortuntaly TS will treat it the same way as `ts-expect-error`, does not matter the `-if` at the end. I open to some other suggestions though

millsp (approved):
That's really going to help us having even less tests duplicates. Thanks!

Related: If checked if TypeScript would support block ts comments. [Not there yet](https://github.com/microsoft/TypeScript/issues/19573).
But it would have been really nice to have this done automatically on `testIf` 😁

Looks great overall, and I love the idea (+ see comments)

packages/client/tests/functional/_example/tests.ts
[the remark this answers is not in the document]
millsp: I have a few ideas:
* @ts-skip-if
* @ts-accept-error-if
* @ts-no-error-if
* @ts-test-if

I prefer `@ts-test-if` because it's 1-to-1 with the `testIf` condition.
So we could copy paste the condition just above, without negation.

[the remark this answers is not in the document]
SevInf: Agree, will update to `@ts-test-if` then

packages/client/tests/functional/typescript/tests.ts
```diff
@@ -35,9 +35,12 @@ function getAllTestSuiteTypeChecks(fileNames: string[]) {
...
 
 describe('typescript', () => {
-  const suitePaths = glob.sync('./**/.generated/**/*.ts', {
-    ignore: ['./**/.generated/**/*.d.ts', './**/.generated/**/_schema.ts'],
-  })
+  const suitePaths = glob
+    .sync('./**/.generated/**/*.ts', {
+      ignore: ['./**/.generated/**/*.d.ts', './**/.generated/**/_schema.ts'],
+    })
+    .concat([path.resolve(__dirname, '..', 'globals.d.ts')])
```
millsp, line 42: Why does it need type checking?

packages/client/tests/functional/_utils/setupFilesAfterEnv.ts
```diff
@@ -5,4 +5,7 @@ process.env.DATABASE_URI_postgresql = process.env.TEST_POSTGRES_URI
 process.env.DATABASE_URI_mysql = process.env.TEST_MYSQL_URI
 process.env.DATABASE_URI_cockroachdb = process.env.TEST_COCKROACH_URI
 
+globalThis.testIf = (condition: boolean) => (condition ? test : test.skip)
```
millsp, line 8: Neat.

packages/client/tests/functional/typescript/tests.ts
SevInf: It does not really need typechecking, but since nobody really directly imports it it has no be included in the source set so global definitions for `testIf` and `describeIf` will be discovered.

millsp: Makes sense. Can you please add a short comment above concat about this?

PR #13727, opened 2022-06-09
chore(tests): Generate `optOut` block during scaffolding

SevInf:


packages/client/helpers/generator-test/templates/tests.ts.tpl
```diff
@@ -5,6 +5,15 @@ declare let prisma: import('@prisma/client').PrismaClient
 
 testMatrix.setupTestSuite((suiteConfig, suiteMeta) => {
   test('example', () => {
-      expect.fail('No test defined')
+    throw new Error('No test defined')
```
SevInf, line 8: `expect.fail` is no longer a thing in fresh jest

danstarns (approved):

PR #13731, opened 2022-06-09
chore(client): debug unpkg.com in _getClientVersion()

Jolg42:
To debug the following in dev / patch-dev / integration versions

https://prisma-company.slack.com/archives/C02C34KRHBQ/p1654775627840349

https://github.com/prisma/ecosystem-tests/runs/6810795540?check_suite_focus=true#step:6:340
```
$ jest
FAIL ./index.test.ts
  use data proxy
    ✕ fetch response (104 ms)
  ● use data proxy › fetch response
    SyntaxError: Unexpected token F in JSON at position 0
        at JSON.parse (<anonymous>)
      at Object.json (node_modules/@prisma/client/runtime/index.js:44246:22)
      at _getClientVersion (node_modules/@prisma/client/runtime/index.js:44291:23)
      at getClientVersion (node_modules/@prisma/client/runtime/index.js:44299:20)
      at DataProxyEngine.url (node_modules/@prisma/client/runtime/index.js:44347:36)
      at DataProxyEngine.requestInternal (node_modules/@prisma/client/runtime/index.js:44395:29)
      at DataProxyEngine.requestBatch (node_modules/@prisma/client/runtime/index.js:44388:29)
```
Code at https://github.com/prisma/ecosystem-tests/blob/dev/dataproxy/nodejs/src/index.ts#L12:L12
```

const prisma = new PrismaClient({})

export function getUsers() {
  return prisma.$transaction([
    prisma.user.findFirst(),
    prisma.user.findMany()
  ])
}
```

### DIFF
--- a/packages/engine-core/src/data-proxy/utils/getClientVersion.ts
+++ b/packages/engine-core/src/data-proxy/utils/getClientVersion.ts
@@ -35,8 +35,19 @@ async function _getClientVersion(config: EngineConfig) {
     // we resolve with the closest or previous version published on npm
     const pkgURL = prismaPkgURL(`<=${major}.${minor}.${patch}`)
     const res = await request(pkgURL, { clientVersion })
-
-    return (await res.json())['version'] as string
+    
+    const bodyAsText = await res.text()
+    debug('body fetched from unpkg.com: ', bodyAsText)
+    
+    let bodyAsJson
+    try {
+      bodyAsJson = JSON.parse(bodyAsText)
+    } catch (e) {
+      console.error('JSON.parse error: body fetched from unpkg.com: ', bodyAsText)
+      throw e
+    }
+    
+    return bodyAsJson['version'] as string
   }
 
   // nothing matched, meaning that the provided version is invalid
--- a/packages/engine-core/src/data-proxy/utils/getClientVersion.ts
+++ b/packages/engine-core/src/data-proxy/utils/getClientVersion.ts
@@ -35,10 +35,10 @@ async function _getClientVersion(config: EngineConfig) {
     // we resolve with the closest or previous version published on npm
     const pkgURL = prismaPkgURL(`<=${major}.${minor}.${patch}`)
     const res = await request(pkgURL, { clientVersion })
-    
+
     const bodyAsText = await res.text()
     debug('body fetched from unpkg.com: ', bodyAsText)
-    
+
     let bodyAsJson
     try {
       bodyAsJson = JSON.parse(bodyAsText)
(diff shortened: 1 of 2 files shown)


jkomyno (approved):

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -70,6 +73,7 @@ function buildOptions(options: RequestOptions): Https.RequestOptions {
  */
 function buildResponse(incomingData: Buffer[], response: IncomingMessage): RequestResponse {
   return {
+    text: () => Buffer.concat(incomingData).toString(),
     json: () => JSON.parse(Buffer.concat(incomingData).toString()),
```
aqrln, line 77: Should we extract the common expression into a variable? Probably doesn't matter much in practice, but still feels wrong to perform these operations twice.

Jolg42: Not sure, it's very useful here. Here we need to have a function as it's only done "on demand" and it's very minimal.

aqrln: Ah, sorry, I didn't notice these were functions.

aqrln (approved):

packages/engine-core/src/data-proxy/utils/request.ts
aqrln: I was reading too fast I guess 😅

packages/engine-core/src/data-proxy/utils/getClientVersion.ts
```diff
@@ -36,7 +36,18 @@ async function _getClientVersion(config: EngineConfig) {
     const pkgURL = prismaPkgURL(`<=${major}.${minor}.${patch}`)
     const res = await request(pkgURL, { clientVersion })
 
-    return (await res.json())['version'] as string
+    const bodyAsText = res.text()
```
aqrln, line 39: One more question: shouldn't this be 
```suggestion
    const bodyAsText = await res.text()
```
?

Jolg42: @aqrln Not because it's not node-fetch here, we are stealing types from it 😄 but in our implementation we don't need to await `text()`.

aqrln (approved):

Jolg42:
I'm thinking that this should go into an integration branch before we can merge

Jolg42:
Created `integration/debug-unpkg.com`

Release in progress https://buildkite.com/prisma/release-prisma-typescript/builds/6948

Jolg42:
In the end @aqrln we need, to await because we switch to global "fetch" for edge runtimes and there it will fail as it's required.

Issue #13717, opened 2022-06-08
Error: Get config: Unable to establish a connection to query-engine-node-api library


PiecePaperCode:
```sh
$/prisma/starter$ npx prisma generate
Prisma schema loaded from prisma/schema.prisma
Error: Get config: Unable to establish a connection to query-engine-node-api library

Prisma CLI Version : 3.15.1-dev.1
```

**How to Reproduce**
OS: 22.04 Ubuntu
Node: 16.15.1
Follow these instructions. https://www.prisma.io/docs/getting-started/quickstart#1-download-the-starter-project
Switch to latest possible release: 3.15.1-dev.1 as per today.

**What have i tried so far**
Switching databases sqlite, postgres
Switching node from 16 to 17.
using 3.13 3.14 gives me this error
```
/snap/core20/current/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /lib/x86_64-linux-gnu/libcrypto.so.3)
```
in ubuntu i could not figure out how to install GLIBC_2.33
tying with npm and yarn

janpio:
Error message is from our new error reporting code from @jkomyno: https://github.com/prisma/prisma/blob/a490f5a82baeccf985c2cc30476f45748ed0fcad/packages/sdk/src/engine-commands/queryEngineCommons.ts#L80

janpio:
To confirm:
Using 3.14.0 you are getting a different error message around GLIBC?

irfanabdnst:
I got this error when using 3.14.0 and 3.13.0
`libssl.so.1.1: cannot open shared object file: No such file or directory`

PiecePaperCode:
i confirm using version 3.13.x 3.14.x gives this error 

`/snap/core20/current/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /lib/x86_64-linux-gnu/libcrypto.so.3)

`

when using 3.15 gives the Config Error described above.

in bouth cases making prisma work on ubuntu 22.04 is not possible.

janpio:
@irfanabdnst Please create a new issue providing all the info the bug issue template asks for. Thanks.

pantharshit00:
Looks like you are using snap to either install Node.js or openssl or maybe even glibC. Since snaps are sandboxes we can't dynamically link to required runtime libraries and so it will throw out that error. So please try installing these libs using `apt`:
```
sudo apt install openssl openssl-dev libc6 
```

[...]

PR #13736, opened 2022-06-09
feat(sdk): improved error reporting by adding detail to getConfig and getDmmf error

jkomyno:
Improve error reporting occurring in https://github.com/prisma/prisma/issues/13717.

Screenshot 2022-06-09 at 16 51 14


packages/sdk/src/engine-commands/queryEngineCommons.ts
```diff
@@ -77,7 +77,7 @@ export function loadNodeAPILibrary(queryEnginePath: string) {
       () => load<NodeAPILibraryTypes.Library>(queryEnginePath),
       (e) => ({
         type: 'connection-error' as const,
-        reason: 'Unable to establish a connection to query-engine-node-api library',
+        reason: `Unable to establish a connection to query-engine-node-api library. Have you tried installing 'openssl'?`,
```
janpio, line 80: I don't think this is usually caused by missing OpenSSL - that is just one of many possible reasons.
(If we want to react to OpenSSL missing explicitly, we should look at the detailed error message only and trigger by that and append something to the error string)

packages/sdk/src/engine-commands/queryEngineCommons.ts
```diff
@@ -75,11 +76,24 @@ export function loadNodeAPILibrary(queryEnginePath: string) {
...
-        error: e as Error,
-      }),
+      (e) => {
+        const error = e as Error
+        const defaultErrorMessage = `Unable to establish a connection to query-engine-node-api library.`
+        const proposedErrorFixMessage = match(error.message)
+          .when(
+            (errMessage) => errMessage.includes('openssl'),
+            () => {
+              return ` Have you tried installing 'openssl'?`
```
janpio, line 86: I fear that is not a super helpful error message (The snap thing for example would not be really covered by this msg, wouldn't it?). 

Maybe we should ahve a troubleshooting entry in our docs and a pris.ly redirect pointing there that we can link to here instead?

packages/sdk/src/engine-commands/queryEngineCommons.ts
```diff
@@ -75,11 +76,24 @@ export function loadNodeAPILibrary(queryEnginePath: string) {
...
-        type: 'connection-error' as const,
-        reason: 'Unable to establish a connection to query-engine-node-api library',
-        error: e as Error,
-      }),
+      (e) => {
+        const error = e as Error
+        const defaultErrorMessage = `Unable to establish a connection to query-engine-node-api library.`
+        const proposedErrorFixMessage = match(error.message)
+          .when(
+            (errMessage) => errMessage.includes('openssl'),
```
janpio, line 84: A test for this, even mocked, would be good I guess.

packages/sdk/src/engine-commands/queryEngineCommons.ts
jkomyno: We can talk about this later with Lucy. I like the idea

janpio: If you want to merge this without further changes on that message, extract this suggestion into an issue pls.
Maybe still ask @pantharshit00 for a potentially better description of what there is to do right now to fix this in a general (non snap specific) way.

packages/sdk/src/engine-commands/queryEngineCommons.ts
janpio: Nice work!

janpio (changes requested):
some snapshot updates or test differentiation based on engine type required it seems

packages/sdk/src/engine-commands/getConfig.ts
```diff
@@ -37,9 +37,44 @@ export type GetConfigOptions = {
...
+
+type GetConfigErrorInit = {
+  // e.g., `Schema parsing - Error while interacting with query-engine-node-api library`
+  reason: string
+
+  // e.g., Error validating model "public": The model name \`public\` is invalid.
+  message: string
+} & (
+  | {
+      readonly _tag: 'parsed'
```
Jolg42, line 49:
```suggestion
      // parsed as JSON
      readonly _tag: 'parsed'
```

packages/sdk/src/engine-commands/getConfig.ts
```diff
@@ -37,9 +37,44 @@ export type GetConfigOptions = {
...
+
+      // e.g., `P1012`
+      errorCode?: string
+    }
+  | {
+      readonly _tag: 'unparsed'
```
Jolg42, line 55:
```suggestion
      // text
      readonly _tag: 'unparsed'
```

packages/sdk/src/engine-commands/getDmmf.ts
```diff
@@ -39,9 +39,43 @@ export type GetDMMFOptions = {
...
+  message: string
+} & (
+  | {
+      readonly _tag: 'parsed'
+
+      // e.g., `P1012`
+      errorCode?: string
+    }
+  | {
+      readonly _tag: 'unparsed'
```
Jolg42, line 56:
```suggestion
      // text
      readonly _tag: 'unparsed'
```

packages/sdk/src/engine-commands/getDmmf.ts
```diff
@@ -39,9 +39,43 @@ export type GetDMMFOptions = {
...
 
+type GetDmmfErrorInit = {
+  // e.g., `Schema parsing - Error while interacting with query-engine-node-api library`
+  reason: string
+
+  // e.g., Error validating model "public": The model name \`public\` is invalid.
+  message: string
+} & (
+  | {
+      readonly _tag: 'parsed'
```
Jolg42, line 50:
```suggestion
      // JSON
      readonly _tag: 'parsed'
```

Jolg42:
Looks good but I can see tests are failing in `sdk` with
```
FAIL src/__tests__/engine-commands/queryEngineCommons.test.ts
  loadNodeAPILibrary
    ✕ error path (121 ms)
    ✕ error path, openssl (122 ms)
    ✕ happy path (116 ms)
```

and in `client` with
```
     - Get DMMF: Schema parsing - Error while interacting with query-engine-node-api library
    + Get DMMF: Schema parsing - Error while interacting with query-engine binary
```
I guess it's because there is only one snapshot for library and engine and we need to do something here

packages/sdk/src/engine-commands/queryEngineCommons.ts
jkomyno: I had sent a Slack message to @pantharshit00 asking him to take a look here, I'll ping him back

packages/client/src/__tests__/dmmf.test.ts
```diff
@@ -324,7 +326,55 @@ describe('dmmf', () => {
       }
     `)
   })
+})
+
+describeIf(process.env.PRISMA_CLI_QUERY_ENGINE_TYPE === 'library')('dmmf library', () => {
```
Jolg42, line 331: Maybe instead of duplicating here and skipping for binary in `packages/client/src/__tests__/generation/generator.test.ts`

We could try to normalize / search and replace `query-engine-node-api library` to something like `query-engine-NORMALIZED` or so
With a if 
```
if (process.env.PRISMA_CLI_QUERY_ENGINE_TYPE === 'binary') {
 expect(stripAnsi(e.message).toContain("query-engine binary"))
} else {
expect(stripAnsi(e.message).toContain("query-engine-node-api library"))
}
```

What do you think?

packages/sdk/src/engine-commands/queryEngineCommons.ts
janpio: Harshit is out sick today. 

The wording here is pretty broad in my opinion, and as far as I know it being not installed was never the problem. It was mostly about it not being installed correctly or via the correct method. Maybe we can make it more neutral as "The error message indicates this is a problem with your OpenSSL installation" or something like that.

packages/client/scripts/get-packed-client.js
```diff
@@ -9,4 +9,6 @@ async function main() {
   console.log(`Saving packed client to ${target}`)
 }
 
-main()
+main().catch((e) => {
+  throw e
+})
```
jkomyno, line 14: After merging `main` back to this branch, `eslint` was throwing an error due to `@typescript-eslint/no-floating-promises`:

> Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator

Alternatively, this can be rewritten as

```suggestion
void main()
```

packages/internals/src/engine-commands/queryEngineCommons.ts
```diff
@@ -75,11 +76,24 @@ export function loadNodeAPILibrary(queryEnginePath: string) {
...
-        error: e as Error,
-      }),
+      (e) => {
+        const error = e as Error
+        const defaultErrorMessage = `Unable to establish a connection to query-engine-node-api library.`
+        const proposedErrorFixMessage = match(error.message)
+          .when(
+            (errMessage) => errMessage.includes('openssl'),
+            () => {
+              return ` Have you tried installing 'openssl'?`
```
Jolg42, line 86: Seems that the previous comments still apply here
https://github.com/prisma/prisma/pull/13736#discussion_r901492117

packages/internals/src/index.ts
```diff
@@ -70,6 +70,7 @@ export { parseBinaryTargetsEnvValue, parseEnvValue } from './utils/parseEnvValue
 export { pick } from './utils/pick'
 export { platformRegex } from './utils/platformRegex'
 export { printConfigWarnings } from './utils/printConfigWarnings'
+export { serializeQueryEngineName } from './utils/serializeQueryEngineName'
```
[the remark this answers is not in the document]
Jolg42: Probably better to move it there to keep things together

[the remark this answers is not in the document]
jkomyno: Actually, I'd advise against this, as it would not allow us to observe the right output when explicitly snapshot testing on "binary" and "library" (we would just see `query-engine-NORMALIZED` everywhere, and at that point either we trust the output code 100%, or we might have typo regressions in the future in `getDmmf` and `getConfig`)

packages/internals/src/engine-commands/queryEngineCommons.ts
Jolg42: This is the last blocking part before merging

packages/internals/src/__tests__/engine-commands/queryEngineCommons.test.ts
```diff
@@ -0,0 +1,68 @@
+import { BinaryType } from '@prisma/fetch-engine'
+import * as E from 'fp-ts/Either'
+
+import { loadNodeAPILibrary } from '../../engine-commands/queryEngineCommons'
+import { resolveBinary } from '../../resolveBinary'
+import * as loadUtils from '../../utils/load'
+
+const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
+
+describeIf(process.env.PRISMA_CLI_QUERY_ENGINE_TYPE === 'library')('loadNodeAPILibrary', () => {
```
Jolg42, line 10: Would be better to run it by default even if the env var is not set 
```suggestion
describeIf(process.env.PRISMA_CLI_QUERY_ENGINE_TYPE !== 'binary')('loadNodeAPILibrary', () => {
```

Jolg42 (approved):
Nice one with the spy/mock tests!

See last comment about a test before merging 🚢

packages/internals/src/engine-commands/queryEngineCommons.ts
jkomyno: It now says `It seems there is a problem with your OpenSSL installation!`

janpio:
Is there an issue this can close so this can appear in the issue list in our release notes?

jkomyno:
> Is there an issue this can close so this can appear in the issue list in our release notes?

This PR closes https://github.com/prisma/prisma/issues/14006

PR #13743, opened 2022-06-10
docs: Add docs for client functional tests

SevInf:


Jolg42 (approved):
This looks great! 

Note: I didn't try it yet but looks valid

TESTING.md
```diff
@@ -76,8 +76,9 @@ In the `prisma/prisma` repository we have a few places where you can write tests
   - Tests for `prisma studio`, `prisma version`, `prisma format`, `prisma generate`, `prisma doctor`, loading .env files, testing the built cli
 - **`client`**
   - `src/__tests__/*.test.ts` - Unit tests
-  - `src/__tests__/integration/happy/**` - Integration tests for the happy path
-  - `src/__tests__/integration/errors/**` - Integration tests for error cases
+  - `test/functional` - New functional tests setup
+  - `src/__tests__/integration/happy/**` - Legacy integration tests for the happy path. Please, write functional test instead.
+  - `src/__tests__/integration/errors/**` - Legacy integration tests for error cases. Please write functional test instead.
```
janpio, line 81:
```suggestion
  - `src/__tests__/integration/happy/**` - Legacy integration tests for the happy path. Please, write functional tests instead.
  - `src/__tests__/integration/errors/**` - Legacy integration tests for error cases. Please write functional tests instead.
```

TESTING.md
```diff
@@ -116,6 +117,148 @@ The `integration/happy/minimal` test is always a good start if you just want to
 
 In case you want to test the actually generated client, have a look at the `integration/happy/blog` test as an example.
 
+## Functional tests for the client
+
+Functional tests in the client package are testing that all aspects of client and query engine work correctly. They strive to be as close as possible to the way client will be used in real project: they generate an actual client, talk to real databases, perform the type checks and generally test the client through its public API.
```
janpio, line 122:
```suggestion
Functional tests in the client package are testing that all aspects of client and query engine work correctly. They strive to be as close as possible to the way client will be used in real project: they generate an actual client, talk to a real database, perform the type checks and generally test the client through its public API.
```

TESTING.md
```diff
@@ -116,6 +117,148 @@ The `integration/happy/minimal` test is always a good start if you just want to
...
+
+- test matrix `_matrix.ts`
+- schema template `prisma/_schema.ts`
+- test suite `tests.ts`
+
+#### Test matrix
+
+`_matrix.ts` file defines parameters for generating test suites. It can have as many parameters as necessary, but at minimum, it should define at least one provider
+
+This matrix defines 2 test suites: one for SQLite provider and one for MongoDB, while also providing a corresponding ID string for each provider:
```
janpio, line 146:
```suggestion
This example matrix defines 2 test suites: one for the SQLite provider and one for MongoDB, while also providing a corresponding ID string for each provider:
```

TESTING.md
```diff
@@ -116,6 +117,148 @@ The `integration/happy/minimal` test is always a good start if you just want to
...
+    },
+    {
+      provider: 'mongodb',
+      id: 'String @id @default(auto()) @map("_id") @db.ObjectId',
+    },
+  ],
+])
+```
+
+If matrix has multiple dimensions, test suites will be generated for all permutations of the parameters. For example. following matrix:
```
janpio, line 165:
```suggestion
If the matrix has multiple dimensions, test suites will be generated for all permutations of the parameters. For example, the following matrix:
```

TESTING.md
```diff
@@ -116,6 +117,148 @@ The `integration/happy/minimal` test is always a good start if you just want to
...
+    
+    model User {
+      id ${idForProvider(provider)}
+    }
+  `
+})
+```
+
+`setupSchema` callback receives all parameters from the matrix for a particular test suite as an argument.
+`idForProvider` is a helper function which returns a correct primary key definition for each of the supported providers.
```
janpio, line 225: so it requires both `provider` and `id` to be present in the matrix as in the example above, right?
so `id` is kind of a magic name up there with meaning, and not just a random string?

are there others than `id`?

TESTING.md
```diff
@@ -116,6 +117,148 @@ The `integration/happy/minimal` test is always a good start if you just want to
...
+      // if you are skipping tests for certain providers, you
+      // have to list them here and specify the reason
+      from: ['mongodb'],
+      reason: 'The test is for SQL databases only',
+    },
+  },
+)
+```
+
+This tests will run for every permutation of the parameters from the matrix. Each suite will start with a clean already set up database, generated and initialized client, available via `prisma` global. After suite is finished, database will be dropped and client instance will disconnect automatically.
```
janpio, line 254:
```suggestion
This test will run for every permutation of the parameters from the matrix. Each suite will start with a clean already set up database, generated and initialized client, available via `prisma` global. After suite is finished, database will be dropped and client instance will disconnect automatically.
```

janpio (approved):
Yay this is already great, I have the feeling I know what to do now 👍

TESTING.md
SevInf: No, it actually does not need the `id` in the matrix, it's just a bunch of "if provider is this, that return that hardcoded string" statements under the hood. I've added `id` to the matrix example  to demonstrate that you can have more than just `provider` there, but I see now how these two examples together make a wrong impression. I'll come up with something else for the matrix definition part.

janpio: Good I actually asked this question then 👍

janpio: Ahhhh, so the manual solution got replaced by a hardcoded utility. Makes so much sense now. Thanks for connecting these dots for me.

TESTING.md
```diff
@@ -147,6 +148,170 @@ To change the default Rust artifacts' type used under the hood, you can set the
...
+```
+
+This test will run for every permutation of the parameters from the matrix. Currently used combination is available via `suiteConfig` parameter. Each suite will start with a clean already set up database, generated and initialized client, available via `prisma` global. After suite is finished, database will be dropped and client instance will disconnect automatically.
+
+### Running functional tests
+
+- `pnpm test:functional:code` generates and runs the test suites, defined by test matrix. It does no typechecking, but prepares all necessary files for it.
+- `pnpm test:functional:types` runs typechecking on all the suites, generated by `pnpm test:functional:code` command. If it reports any errors, you might want to examine generated test suite under `tests/functional/<your test name>/.generated` directory to get a better diagnostic.
+- `pnpm test:functional` will run tests and perform type checks.
+
```
danstarns, line 299:
```suggestion

If you would like to run a test on its own: append the folder name after the pnpm script, for example running the `_example` test: 

`$ pnpm run test:functional:code _example`

```

danstarns (approved):

PR #13748, opened 2022-06-10
test(client): Improvements to functional tests DX

SevInf:
Adds custom wrapper around jest for tests which allow to do the
following:

- properly updates both inline and external snapshots when -u flag is
  passed. This is done in 2 passes: first one runs only the first suite
  for the matrix (otherwise, inline snapshots fail), second one runs all
  but the first one to properly update remaining external snapshots.
- allows to run tests only for selected providers with `-p` switch.

Fix prisma/client-planning#64


packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -44,22 +44,20 @@ function setupTestSuiteMatrix(
...
-  ;(forceInlineSnapshot ? [suiteTable[0]] : suiteTable).forEach((suiteEntry) => {
-    const [suiteName, suiteConfig] = suiteEntry
-
+  for (const { name, suiteConfig, skip } of testPlan) {
+    const describeFn = skip ? describe.skip : describe
     // we don't run tests for some providers that we want to skip on the CI
-    if (suiteConfig['provider']?.toLowerCase() === 'mongodb' && process.env.TEST_SKIP_MONGODB) return
-    if (suiteConfig['provider']?.toLowerCase() === 'sqlserver' && process.env.TEST_SKIP_MSSQL) return
-    if (suiteConfig['provider']?.toLowerCase() === 'cockroachdb' && process.env.TEST_SKIP_COCKROACHDB) return
+    if (suiteConfig['provider']?.toLowerCase() === 'mongodb' && process.env.TEST_SKIP_MONGODB) continue
```
SevInf, line 56: Was not sure if moving that to `getTestSuitePlan` will make sense: here it will not even register the test suite, where test plan's skip option will still create the suite, just wraps it with `describe.skip` (otherwise jest will remove skipped snapshots). @millsp do you know if using `describe.skip` instead of bailing out will make anything worse?

SevInf: Moved this code to the `getTestSuitePlan`

danstarns (approved):
Reviewed face to face!

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -0,0 +1,83 @@
...
+    includedProviders: process.env.ONLY_TEST_PROVIDERS?.split(','),
+    excludedProviders: getExcludedProviders(),
+    updateSnapshots: process.env.UPDATE_SNAPSHOTS as 'inline' | 'external' | undefined,
+  }
+}
+
+const excludeEnvToProviderMap = {
+  TEST_SKIP_MONGODB: 'mongodb',
+  TEST_SKIP_MSSQL: 'sqlserver',
+  TEST_SKIP_COCKROACHDB: 'cockroachdb',
```
danstarns, line 73:
```suggestion
  TEST_SKIP_COCKROACHDB: 'cockroachdb',
  TEST_SKIP_POSTGRESQL: 'postgresql',
  TEST_SKIP_SQLITE: 'sqlite'
```

danstarns (approved):

PR #13749, opened 2022-06-10
ci: reduce macos jobs, combine all tests

Jolg42:
Client test takes ~10min and setup 5min but also means a lot of queuing is happening since the max concurrent macos runners is 5

Example https://github.com/prisma/prisma/runs/6798858585?check_suite_focus=true


Jolg42:
I put client to test right after sdk so it should be as good as before for feedback loop

aqrln (approved):

PR #13753, opened 2022-06-10
ci: add dockerhub login to ci

jkomyno:


.github/workflows/test.yml
```diff
@@ -86,6 +86,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
```
Jolg42, line 94: Not needed in lint
```suggestion
```

.github/workflows/test.yml
```diff
@@ -444,6 +480,12 @@ jobs:
...
```
Jolg42, line 488:
```suggestion
```

Not needed in cli-commands

.github/workflows/test.yml
```diff
@@ -499,6 +541,12 @@ jobs:
...
```
Jolg42, line 549:
```suggestion
```
Not needed in others

Jolg42 (changes requested):
I added suggestions to remove the login where it's not needed (where we do not run docker-compose)

pimeys (approved):
seems to work.

Jolg42 (approved):

.github/workflows/test.yml
```diff
@@ -132,6 +132,12 @@ jobs:
...
```
janpio, line 140: Here and lines 255-259 it would make sense to put it directly before the `docker-compose` call to make it easier to correlate and see hte connection between the two.

PR #13783, opened 2022-06-13
fix(client): use objects for DbNull, JsonNull, AnyNull

### LINKED ISSUE #9243: Avoid using magic string values for JsonNull/DbNull
## Problem

Since Prisma 3.0, the input for filtering and updating nullable JSON fields no longer accepts `null`, but instead the `Prisma.JsonNull` and `Prisma.DbNull` values to better distinguish between the two types of nulls.

I can see that this solves a real problem, but the current implementation has a significant drawback: `JsonNull` and `DbNull` are implemented as magic string values. This means their types overlap with regular string data that can be stored in a JSON field. Some problems this could cause:
- Filtering/storing `Prisma.DbNull` on a non-nullable JSON field will type-check, but produce a JSON string value of `"DbNull"` instead of an actual `NULL`, which may lead to confusing bugs
- User-provided strings that get stored as a top-level value in a JSON field need to be sanitized to make sure they're not `"JsonNull"` or `"DbNull"`, or users might be able to inject either kind of null value and cause an unexpected state

## Suggested solution

A possible alternative approach would be making JsonNull and DbNull `Symbol`s or some kind of class instance, so that they don't overlap with the `JsonValue` type

aqrln:
Closes https://github.com/prisma/client-planning/issues/18
Fixes https://github.com/prisma/prisma/issues/9243

Make `DbNull`, `JsonNull` and `AnyNull` special objects.

**Breaking change:** when passing `"DbNull"`, `"JsonNull"` and `"AnyNull"` strings to JSON fields, they will now always be interpreted as strings (while previously they could be interpreted either as strings or as special values).

* If you use literal `"DbNull"`, `"JsonNull"` and `"AnyNull"` strings, please update to use `Prisma.DbNull`, `Prisma.JsonNull` and `Prisma.AnyNull`. If you are already using the named constants (which should be the case for most people since that's how it's documented they should be used), no action is required on your side.
* If you now get a type error when passing `Prisma.DbNull` as the value of a JSON field, this indicates a bug in your code which wasn't caught by our types previously. The field you are trying to store `DbNull` in is not nullable in the schema, and a literal `"DbNull"` string was stored in the database instead of `NULL`.
* If you had to sanitize user input to not be equal to `"DbNull"` or `"JsonNull"` before using it as a value of a JSON column, this is no longer needed, it is safe to pass it as is. Most people would use `String` and not `Json` columns in the cases which would have been affected by this, but I'm mentioning it anyway because it was raised as one of the concerns in the original issue.

---

**First iteration notes**

This implementation is fairly generic on the runtime side, but I had to dumb down the generation side compared to what I had initially.

My first approach was to make it possible to mark any enum as symbol enum, generate a registry of all used symbols in `Prisma.Symbols` namespace on-the-fly as such enums are encountered when we traverse the DMMF and then reference them in each specific enum from that registry.

It's important that we can't just declare the symbols as `unique symbol` in each specific enum in isolation because that would make, e.g., `JsonNull` from `JsonNullValueInput` and `NullableJsonNullValueInput` incompatible between each other on the type level, while we want to preserve the behavior of string literal enums that makes enum variants with identical names interchangeable across different enums — otherwise it won't be possible to use the global helpers in the `Prisma` namespace like `Prisma.JsonNull`.

However, that didn't play well with those helpers in a different aspect: since they are hardcoded and always present, they would need `Prisma.Symbols.JsonNull`, `Prisma.Symbols.DbNull` and `Prisma.Symbols.AnyNull` to always be present as well, which, as it turned out, is not always the case, since they would only be generated if JSON fields are used in the schema.

Thus I decided to declare the top-level helpers as unique symbols and reference them directly from the `Prisma` namespace in the generated enums.

If we need any new symbol enums in the future and we don't want to pollute the `Prisma` namespace with their variants, it will still be possible to re-introduce the symbols namespace and generate the top-level helpers conditionally (possibly even defined either as a reference to a symbol from the registry or as a `unique symbol` stub for compatibility, so that all global symbols are always present). However, for now it seems like an overkill.

**Second iteration notes**

`DbNull`, `JsonNull` and `AnyNull` are now special objects instead of symbols. This allows for much better type errors when these values are misused.

Before:

```
Type 'unique symbol' is not assignable to type 'unique symbol | InputJsonValue'.
```

After:

```
Type 'DbNull' is not assignable to type 'JsonNull | InputJsonValue'.
```


packages/client/src/runtime/utils/common.ts
```diff
@@ -400,6 +415,11 @@ export function isGroupByOutputName(type: string): boolean {
 
 export function isSchemaEnum(type: any): type is DMMF.SchemaEnum {
   return (
-    typeof type === 'object' && type.name && typeof type.name === 'string' && type.values && Array.isArray(type.values)
+    typeof type === 'object' &&
+    type !== null &&
+    type.name &&
+    typeof type.name === 'string' &&
+    type.values &&
+    Array.isArray(type.values)
```
SevInf, line 423:
```suggestion
    typeof type === 'object' &&
    type !== null &&
    typeof type.name === 'string' &&
    Array.isArray(type.values)
```

I know it's an old code that just got reformatted, but I've noticed that couple of checks are already covered by other conditions.

packages/client/src/runtime/utils/common.ts
```diff
@@ -179,6 +177,23 @@ export function getGraphQLType(value: any, potentialType?: string | DMMF.SchemaE
...
 
+export function isValidEnumValue(value: any, inputType?: DMMF.SchemaArgInputType) {
+  const enumType = inputType?.type
+
+  if (!isSchemaEnum(enumType)) {
+    return false
+  }
+
+  if (inputType?.namespace === 'prisma' && symbolEnums.includes(enumType.name)) {
+    if (typeof value !== 'symbol' || !value.description) {
```
SevInf, line 188:
```suggestion
    if (typeof value !== 'symbol' || !Symbol.keyFor(value)) {
```

Very minor thing, but it will also exclude local symbols that happen to have same description.

packages/client/src/generation/TSClient/common.ts
```diff
@@ -66,9 +66,9 @@ Prisma.validator = () => (val) => val
 /**
  * Shorthand utilities for JSON filtering
  */
-Prisma.DbNull = 'DbNull'
-Prisma.JsonNull = 'JsonNull'
-Prisma.AnyNull = 'AnyNull'
+Prisma.DbNull = Symbol.for('DbNull')
```
SevInf, line 69: Question: do we really need global symbols here or we could get away with local ones (`Symbol('DbNull')`)?

SevInf (approved):
I left couple of questions/suggestions, but LGTM overall.

packages/client/src/__tests__/json.test.ts
```diff
@@ -89,6 +89,23 @@ describeIf(process.platform !== 'win32')('json', () => {
...
+      },
+      rootTypeName: 'query',
+      rootField: 'findManyOptionalUser',
+    })
+    document.validate(undefined, false, 'user', 'colorless')
+    expect(String(document)).toMatchSnapshot()
+  })
+
+  test('should not consider "JsonNull" string an enum value', () => {
     const document = makeDocument({
```
aqrln, line 109: The diff, though technically correct, is logically misleading somewhat.

What happened here is I changed `'JsonNull'` to `Symbol('JsonNull')` in the existing test (`should be able filter json null`) and added a new test (`should not consider "JsonNull" string an enum value`) that just so happens to be identical to the old one but has a new meaning.

packages/client/src/runtime/utils/common.ts
```diff
@@ -84,6 +85,7 @@ export const JSTypeToGraphQLType = {
   string: 'String',
   boolean: 'Boolean',
   object: 'Json',
+  symbol: 'Symbol',
```
millsp, line 88: Is this a valid type and can this be ever reached? Or does it have another purpose?

millsp (approved):
Looks great, nice work 👏

packages/client/src/runtime/utils/common.ts
aqrln: It's reachable when a symbol that is not a valid enum value is passed (maybe even just some random JavaScript symbol). We need something here so that `getGraphQLType` doesn't return `undefined`. The exact name of the type isn't super important, it will basically only be used in the validation error message.

aqrln: ah no, it's actually also used in one check in `hasCorrectScalarType`

aqrln: So, now it's not really used in this PR anymore, but it's still nice to have for better validation errors if an unexpected JS symbol is passed to whatever field.

Should I keep it here, or move it to a separate PR (with a test for that optimally)?

packages/client/src/runtime/object-enums.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+const secret = Symbol()
+
+/**
+ * Base class for unique values of object-valued enums.
+ */
+export abstract class ObjectEnumValue {
+  #representation: string
+
+  constructor(arg?: symbol) {
```
millsp, line 18: What are the benefits of using a secret vs a boolean? I have the feeling have I missed something here...

packages/client/src/runtime/utils/common.ts
millsp: I think it's ok to keep indeed.

packages/client/tests/functional/json-null-types/tests.ts
```diff
@@ -0,0 +1,119 @@
...
+                   79   await expect(
+                →  80     prisma.requiredJsonField.create({
+                            data: {
+                              json: new Prisma.NullTypes.JsonNull()
+                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+                            }
+                          })
+
+                Argument json: Provided value new Prisma.NullTypes.JsonNull() of type JsonNull on prisma.createOneRequiredJsonField is not a enumTypes.
+                → Possible values: JsonNullValueInput.JsonNull
```
millsp, line 102: Should it be `Prisma.JsonNull` here?

millsp (approved):

SevInf (approved):

packages/client/tests/functional/json-null-types/tests.ts
aqrln: Well, good question.

On the one hand, it would've been nice UX, but, on the other hand, it might be pretty complicated to change and inconsistent with validation errors that involve other enums (they are always printed as `Enum.Variant`).

FWIW, using `Prisma.JsonNullValueInput.JsonNull` is totally valid and identical to just `Prisma.JsonNull`.

packages/client/src/runtime/object-enums.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Module-private symbol used to distinguish between instances of
+ * `ObjectEnumValue` created inside and outside this module.
+ */
+const secret = Symbol()
+
+/**
+ * Base class for unique values of object-valued enums.
+ */
+export abstract class ObjectEnumValue {
+  #representation: string
```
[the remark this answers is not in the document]
aqrln: I actually checked this, and it was compiled to a `WeakMap` unfortunately.
We really need to bump the target in TypeScript and esbuild — I believe private properties were supported natively even in Node.js 12.

packages/client/src/runtime/object-enums.ts
aqrln: With a boolean it would still be possible to ignore the types and do `new Prisma.NullTypes.JsonNull(true)` in the application code, with a symbol we can't do that even in Prisma code outside this module

packages/client/tests/functional/json-null-types/tests.ts
aqrln: So, I took a look, and fixed a bug on the line above while I was there (should have said `... is not a JsonNullValueInput`, but due to a typo it printed `bestFittingType.location` instead of `bestFittingType.type` and said `... is not a enumTypes` — now fixed for all enums).

It looks like it shouldn't be hard to add a special case there, but I would prefer the output to be consistent for all kinds of enums:
* either `JsonNullValueInput.JsonNull`, `SortOrder.asc` etc (like it is now)
* or `Prisma.JsonNull`, `'asc'` etc (we can change it to be like this)

What do you think? Or should we discuss and possibly change this later?

millsp: Thanks for fixing the bug! Let's do that later and merge this. Feel free to open an improvement/DX issue, but I don't think it's very urgent right now.

PR #13788, opened 2022-06-13
fix: skip docker login if no docker secrets are found

### LINKED ISSUE #13775: Fix CI support in prisma forks
Our CI currently relies on logging in to Dockerhub via the `docker/login-action@v2` action, in order to be granted higher `docker pull` rate limits.
To do so, we set up the `DOCKER_USERNAME` and `DOCKER_PASSWORD` secrets in the [prisma/prisma](https://github.com/prisma/prisma) repository. However, Github secrets in a repo **DO NOT** propagate when a user forks such repository.

## Solutions

- We could try to log in via `docker/login-action@v2`  only if both the `DOCKER_USERNAME` and `DOCKER_PASSWORD` secrets are set up (preferred)
- We could scope the action so that it only triggers in the [prisma/prisma](https://github.com/prisma/prisma) repository

danstarns:
Closes: https://github.com/prisma/prisma/issues/13775

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -134,6 +134,10 @@ jobs:
 
       - name: Login to Docker Hub
         uses: docker/login-action@v2
+        env:
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+        if: "${{ env.DOCKERHUB_USERNAME != '' && env.DOCKERHUB_TOKEN != '' }}"
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -134,6 +134,10 @@ jobs:
 
       - name: Login to Docker Hub
         uses: docker/login-action@v2
+        env:
```
danstarns, line 137: https://stackoverflow.com/questions/70249519/how-to-check-if-a-secret-variable-is-empty-in-if-conditional-github-actions

Jolg42: It's weird but 🤷🏼 if that works ✔️

Jolg42 (approved):

jkomyno (approved):
LGTM

PR #13806, opened 2022-06-14
test(client): port restart

danstarns:
Ports the restart test into functional. 

Closes: https://github.com/prisma/client-planning/issues/61


packages/client/tests/functional/restart/tests.ts
```diff
@@ -0,0 +1,50 @@
...
+    const client = new PrismaClient()
+    await client.$connect()
+
+    const username = faker.internet.userName()
+
+    await client.user.create({ data: { username } })
+
+    // @ts-ignore kill the binary child process
+    client._engine.child.kill()
+    await sleep(1000)
```
SevInf, line 33: Do we have any idea why we sleep here and in the loop below? Is it to make sure that process died? If so, maybe we should more explicitly wait for it: for example, use [this package](https://github.com/sindresorhus/process-exists)

danstarns: I could not use this because of jest and babel issues importing a module, however, I have simplified the logic and added an await process.exit function.

packages/client/tests/functional/restart/tests.ts
```diff
@@ -0,0 +1,57 @@
...
+    // @ts-ignore internal
+    const waiter2 = waitForChildExit(client._engine.child as ChildProcess)
+    // @ts-ignore internal
+    client._engine.child.kill()
+    await waiter2
+
+    const [found2] = await client.user.findMany({ where: { username } })
+    expect(found2).toBeTruthy()
+
+    client.$disconnect().catch(() => undefined)
```
SevInf, line 55: It's better to move this to `afterAll` or `afterEach`: otherwise, if any of the assertions fail cleanup won't happen.

SevInf (changes requested):

packages/client/tests/functional/restart/tests.ts
danstarns: Good idea! Added.

SevInf (approved):

PR #13814, opened 2022-06-14
test(client): port binary test

danstarns:
Closes: https://github.com/prisma/client-planning/issues/53


aqrln:
> Engine download is handled by test setup itself in our case, not hitting the code path end users will hit.

I'm not sure when exactly download happens, but isn't this also true for the original test? It's just that there's no external test setup code that manages this, but test itself calls `generateTestClient`.

aqrln:
> This is extensively tested via ecosystem-tests

Does it make sense to have this test at all then? Especially since it essentially tests the test setup and not the code that users run.

millsp:
If this is the only thing this test is doing (copying and checking it's existence), I'd say that we can just use ecosystem-tests.

PR #13835, opened 2022-06-15
fix(client, cli): patch for data proxy fixes

millsp:
Related https://github.com/prisma/docs/pull/3274

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -138,7 +138,7 @@ ${chalk.bold('Examples')}
         printDownloadProgress: !watchMode,
         version: enginesVersion,
         cliVersion: pkg.version,
-        dataProxy: !!args['--data-proxy'],
+        dataProxy: !!args['--data-proxy'] || !!process.env.PRISMA_CLI_GENERATE_DATAPROXY,
       })
 
       if (!generators || generators.length === 0) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/Generate.ts
```diff
@@ -138,7 +138,7 @@ ${chalk.bold('Examples')}
         printDownloadProgress: !watchMode,
         version: enginesVersion,
         cliVersion: pkg.version,
-        dataProxy: !!args['--data-proxy'],
+        dataProxy: !!args['--data-proxy'] || !!process.env.PRISMA_CLI_GENERATE_DATAPROXY,
```
janpio, line 141: Are we testing `--data-proxy` here in any way? Can we add one for `PRISMA_CLI_GENERATE_DATAPROXY` then as well? (Fine if not, then we need to via ecosystem-tests)

janpio (approved):

packages/cli/src/Generate.ts
millsp: Outstanding task for post GA, only tested via ecotests right now.

millsp: (...but `--data-proxy` is tested via `getGenerators`)

janpio: So in theory we could replicate that test with `--data-proxy` but with an env var?

janpio:
Note: Needs to be merged back to `main` later.

Issue #2698, opened 2020-06-09
Special error message for default DATABASE_URL


janpio:
After using `prisma init` the default `DATABASE_URL` looks like this:

```
DATABASE_URL="postgresql://johndoe:randompassword@localhost:5432/mydb?schema=public"
```

This is almost never what we want people to use (unless in the edge case of edge cases where a user really creates these credentials and db name to match this), so it would make sense to have a special error message that tells the user about it and suggests appropriate steps.

thebiglabasky:
Would almost put that as an alternative improvement to make `prisma init` default to SQLite honestly.

sean-brydon:
Hey :) Thinking of picking this nice easy one up as my first issue. Is this still be worked on ?

If so ... what approach would you like to go with this, making the default SQLite or throwing a warning if the DB gets a connection with this default string?

janpio:
Warning on default connection string for all different `datasource-provider` values of `init`.

theWolf1198:
@janpio, is this issue available to work on?

janpio:
There is an open PR at https://github.com/prisma/prisma/pull/13836, but it has not been progressed or merged - so yes.

ArgR4N:
Hey @theWolf1198, are you on it? I´d be glad to try it if not.

theWolf1198:
@ArgR4N, please go ahead and work on it. Currently, I am not working on this.

Issue #26, opened 2019-06-15
CLI generated `undefined` when you enter and delete a field in database credentials flow.


pantharshit00:
To Reproduce:
1. Go the credentials entering flow
2. Add a field
3. remove the field.

timsuchanek:
Fixed in the latest version.

tvvignesh:
@timsuchanek This occurs even now when you try to press the delete button (backspace works though) - Using prisma2@2.0.0-preview-3.1 - Screenshot below.

![undefined](https://user-images.githubusercontent.com/1165845/62069196-59cbbf00-b255-11e9-9562-855a424cbd8f.png)

schickling:
This will be fixed as part of #278

divyenduz:
I am still able to reproduce parts of this on Mac, on a Mac

`delete` key acts like backspace on windows/linux systems
- Not reproducible here anymore

`fn` + `delete` key acts like delete on windows/linux systems
- It is reproducible here

timsuchanek:
We disabled the `Delete` key for now.

PR #13836, opened 2022-06-15
feat: warn user if database url is set on default

codams:
Hi, I gave a try to [#26](https://github.com/prisma/prisma/issues/2698)
It's my first PR so do not hesitate to give me hints 😄 

Also should I rebase or do you squash when PR are merged ?

### DIFF
--- a/packages/sdk/src/get-generators/getGenerators.ts
+++ b/packages/sdk/src/get-generators/getGenerators.ts
@@ -11,11 +11,13 @@ import fs from 'fs'
 import pMap from 'p-map'
 import path from 'path'
 
+import { defaultEnv, defaultURL } from '../../../cli/src/Init'
 import { getConfig, getDMMF } from '..'
 import { Generator } from '../Generator'
 import type { GeneratorPaths } from '../predefinedGeneratorResolvers'
 import { predefinedGeneratorResolvers } from '../predefinedGeneratorResolvers'
 import { resolveOutput } from '../resolveOutput'
+import { defaultDatasource } from '../utils/defaultDatasource'
 import { extractPreviewFeatures } from '../utils/extractPreviewFeatures'
 import { mapPreviewFeatures } from '../utils/mapPreviewFeatures'
 import { missingDatasource } from '../utils/missingDatasource'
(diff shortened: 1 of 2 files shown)


packages/sdk/src/get-generators/getGenerators.ts
```diff
@@ -11,11 +11,13 @@ import fs from 'fs'
 import pMap from 'p-map'
 import path from 'path'
 
+import { defaultURL } from '../../../cli/src/Init'
```
Jolg42, line 14: Cannot import like this, fails with
```
Error: . build: packages/sdk build: ../cli/src/Init.ts(13,8): error TS2307: Cannot find module '@prisma/sdk' or its corresponding type declarations.
```
We need to find an alternative here

Jolg42:
>It's my first PR so do not hesitate to give me hints 😄
Thanks for your first PR here! This looks interesting, I think the message might need some tweaks (we will add a suggestion later).

>Also should I rebase or do you squash when PR are merged ?
No, only if there is a commit

We are rather busy with other things at the moment, I think we can get back to you next week 😅

packages/sdk/src/get-generators/getGenerators.ts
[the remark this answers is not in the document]
janpio: Yes, definitely! We use that all over the place :D

codams:
Hey back ! 
Since I need to make an exception for the sqlite case and the import fails, should I rewrite the defaultURL function ? I'm not sure how to import from Init since the relative import fails in the tests 😵‍💫

packages/sdk/src/get-generators/getGenerators.ts
Jolg42: Or in `defaultDatasource.ts` that you just created actually

packages/sdk/src/get-generators/getGenerators.ts
```diff
@@ -120,6 +122,13 @@ export async function getGenerators(options: GetGeneratorOptions): Promise<Gener
     throw new Error(missingDatasource)
   }
 
+  config.datasources.forEach((datasource) => {
+    // If the value is one of the defaults, let's print a warning to the user
+    if (parseEnvValue(datasource.url) === defaultURL(datasource.provider)) {
```
Jolg42, line 127:
```suggestion
    // If the value is one of the defaults, let's print a warning to the user
    // we skip SQLite since the default is `file:./dev.db` and will work "out-of-the-box".
    if (datasource.provider !== 'sqlite' && parseEnvValue(datasource.url) === defaultURL(datasource.provider)) {
```
We should skip SQLite as it will work (it's just a file name, no username or password to set)

wmadden:
Hi @codams, 

It looks like there haven't been any updates since [the last review](https://github.com/prisma/prisma/pull/13836#pullrequestreview-1041990720). I'm closing this PR for now as part of our [ongoing efforts to reduce the number of open issues and PRs](https://pris.ly/orm-manifesto) in the repo.

If you feel like this is a mistake, feel free to reopen the PR and address the comments :)

PR #13839, opened 2022-06-15
test: Port `composites-mongo` test to new setup

SevInf:
First of all, there 2 commits with changes to the functional tests runner:

1. Allow to have multiple test files within single test folder, including subfolders. All tests within the folder will use the same schema and matrix, but will create separate directories in `.generated`. Original `composites-mongo` tests were split up in different files per operation and I thought it will be nice if we could preserve this structure instead of either mashing them together into single file or creating one functional test directory per operation and copy-pasting schema everywhere.
2. Make all matrix parameters a part of test suite name and a folder in `.generated`. 

Now, regarding the porting of the tests itself: original suite had 3 sets of tests for each operation: required 1:1 composite, optional 1:1 composite, list of composites. First 2 had a lot of identical tests between them, so I converted them to a single functional test, which controls optional/required type via matrix. There are still some differences to how those 2 handle `null`, `unset` and `update` operators and there we conditionally adjust them with `testIf` and `@ts-test-if`.  Lists on the other hand have different expected output for most of the cases, so I moved them into their own suite,.

Fix prisma/client-planning#28


packages/client/tests/functional/composites-mongo-list/count.ts
```diff
@@ -0,0 +1,50 @@
...
+import { setupTestSuite } from './_matrix'
+
+// @ts-ignore this is just for type checks
+declare let prisma: import('@prisma/client').PrismaClient
+
+const id = faker.database.mongodbObjectId()
+
+setupTestSuite(() => {
+  beforeAll(async () => {
+    await pRetry(
```
SevInf, line 13: @Jolg42 do you think we still need those retries for `create` considering that each suite starts will clean DB and we use randomized ids so we never have to clean the database between different tests?

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -18,27 +18,24 @@ export async function setupTestSuiteFiles(suiteMeta: TestSuiteMeta, suiteConfig:
   const suiteFolder = getTestSuiteFolderPath(suiteMeta, suiteConfig)
 
   // we copy the minimum amount of files needed for the test suite
-  await fs.copy(path.join(suiteMeta.testDir, 'prisma'), path.join(suiteFolder, 'prisma'))
-  await copyPreprocessed(
-    path.join(suiteMeta.testDir, suiteMeta.testFileName),
-    path.join(suiteFolder, suiteMeta.testFileName),
-    suiteConfig,
-  )
-  await copyPreprocessed(path.join(suiteMeta.testDir, '_matrix.ts'), path.join(suiteFolder, '_matrix.ts'), suiteConfig)
```
SevInf, line 27: We don't this anymore since we also adjust current directory imports to point to original files.

packages/client/tests/functional/composites-mongo-list/count.ts
millsp: I think the database isolation does not work correctly at the moment, and for that reason, the amount of workers is set to `1` in the jest config. This also means it is safe to remove the retry logic, but it also means that we need to fix the isolation issues.

packages/client/tests/functional/composites-mongo-object/_matrix.ts
```diff
@@ -0,0 +1,30 @@
...
+      provider: 'mongodb',
+    },
+  ],
+  [
+    {
+      contentProperty: 'required',
+    },
+
+    {
+      contentProperty: 'optional',
```
millsp, line 15: Very nice to see how you condensed these tests :100: Definitely what I had in mind when I wrote them 😁

packages/client/tests/functional/composites-mongo-object/findMany.ts
```diff
@@ -1,50 +1,36 @@
...
+import { setupTestSuite } from './_matrix'
+import { commentDataA, commentDataB } from './_testData'
 
-const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
+// @ts-ignore this is just for type checks
+declare let prisma: import('@prisma/client').PrismaClient
 
-let PrismaClient, prisma
+// can't use faker ids here because the order between generated ids is arbitrary
+const id1 = '8aaaaaaaaaaaaaaaaaaaaaaa'
```
millsp, line 11: That's when `@snaplet/copycat` would be useful :P

packages/client/tests/functional/composites-mongo-object/_matrix.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const testMatrix = defineMatrix(() => [
+  [
+    {
+      provider: 'mongodb',
+    },
+  ],
+  [
+    {
+      contentProperty: 'required',
```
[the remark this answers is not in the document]
millsp: Ah nevermind, I overlooked this one and it's already done. Awesome.

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -18,27 +18,24 @@ export async function setupTestSuiteFiles(suiteMeta: TestSuiteMeta, suiteConfig:
...
-  await fs.copy(path.join(suiteMeta.testDir, 'package.json'), path.join(suiteFolder, 'package.json')).catch(() => {})
+  await fs.copy(path.join(suiteMeta.testRoot, 'prisma'), path.join(suiteFolder, 'prisma'))
+  await fs.mkdir(path.join(suiteFolder, suiteMeta.rootRelativeTestDir), { recursive: true })
+  await copyPreprocessed(suiteMeta.testPath, path.join(suiteFolder, suiteMeta.rootRelativeTestPath), suiteConfig)
 }
 
 async function copyPreprocessed(from: string, to: string, suiteConfig: TestSuiteConfig): Promise<void> {
```
millsp, line 26: Can you please add a quick tsdoc here too?

millsp (approved):

packages/client/tests/functional/composites-mongo-list/_matrix.ts
```diff
@@ -0,0 +1,21 @@
+import { defineMatrix } from '../_utils/defineMatrix'
```
millsp, line 1: I think we can rename the folders to `composites-list` and `composites-object` as composites might come for more providers in the future.

packages/client/tests/functional/composites-mongo-list/count.ts
SevInf: Thanks everyone, I'll keep retries for now and re-check it once we fix DB isolation.

packages/client/tests/functional/composites-mongo-object/findMany.ts
SevInf: My thoughts as well :)

packages/client/tests/functional/composites-object/count.ts
```diff
@@ -0,0 +1,52 @@
...
+import { setupTestSuite } from './_matrix'
+
+// @ts-ignore this is just for type checks
+declare let prisma: import('@prisma/client').PrismaClient
+
+const id = faker.database.mongodbObjectId()
+
+setupTestSuite(() => {
+  beforeAll(async () => {
+    await pRetry(
```
danstarns, line 13: WDYT about removing the retry logic? Do you know why we are doing it?

packages/client/tests/functional/composites-object/findMany.ts
```diff
@@ -104,7 +85,9 @@ describeIf(!process.env.TEST_SKIP_MONGODB)('findMany > required', () => {
       },
     })
 
-    expect(comment).toMatchInlineSnapshot(`
+    expect(comment).toMatchInlineSnapshot(
```
danstarns, line 88: WDYT about using `expect(result).toMatchObject({})` over snapshots for the tests in this PR? This way we can use truly random data.

danstarns (approved):

packages/client/tests/functional/composites-object/count.ts
millsp: I believe we should be able to remove it while we are running tests in band. As soon as `workers: 1` is removed, we will need to make sure that there is proper database isolation, otherwise the problem will come back.

packages/client/tests/functional/composites-object/findMany.ts
SevInf: Not really sure about it in this case, to be honest. It will be mostly the same snapshot, but written and updated manually instead.

PR #13840, opened 2022-06-15
test(client): Enable functional tests on windows and macos

SevInf:
Fix prisma/client-planning#67

- [x] Set `lower_case_table_names = 1` for mysql on Unix
- [ ] Figure out why error snapshots do not have file name on Windows
- [ ] Fix sqlite transaction issues on Win/OSX


.github/workflows/test.yml
```diff
@@ -699,6 +698,17 @@ jobs:
           flags: client,${{ matrix.os }},${{ matrix.queryEngine }}
           name: client-${{ matrix.os }}-${{ matrix.queryEngine }}
 
+      - name: Test packages/client/functional
+        if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
+        run: pnpm run test:functional -- --jest-args='--testTimeout=40000'
```
aqrln, line 703: I think it should be
```suggestion
        run: pnpm run test:functional --jest-args='--testTimeout=40000'
```
with pnpm 7

packages/client/tests/functional/findFirstOrThrow/tests.ts
```diff
@@ -83,22 +83,16 @@ testMatrix.setupTestSuite((suiteConfig, suiteMeta) => {
...
+
+    try {
+      await prisma.user.findFirstOrThrow({
+        where: { email: existingEmail },
+        // @ts-expect-error passing not supported option on purpose
+        rejectOnNotFound: false,
+      })
+    } catch (error) {
+      expect(error.message).toContain(`'rejectOnNotFound' option is not supported`)
+    }
```
aqrln, line 96:
```suggestion
    const record = prisma.user.findFirstOrThrow({
      where: { email: existingEmail },
      // @ts-expect-error passing not supported option on purpose
      rejectOnNotFound: false,
    })

    await expect(record).rejects.toThrowError(/'rejectOnNotFound' option is not supported/)
```

packages/client/tests/functional/findUniqueOrThrow/tests.ts
```diff
@@ -83,22 +83,16 @@ testMatrix.setupTestSuite((suiteConfig, suiteMeta) => {
...
+
+    try {
+      await prisma.user.findUniqueOrThrow({
+        where: { email: existingEmail },
+        // @ts-expect-error passing not supported option on purpose
+        rejectOnNotFound: false,
+      })
+    } catch (error) {
+      expect(error.message).toContain(`'rejectOnNotFound' option is not supported`)
+    }
```
aqrln, line 96:
```suggestion
    const record = prisma.user.findUniqueOrThrow({
      where: { email: existingEmail },
      // @ts-expect-error passing not supported option on purpose
      rejectOnNotFound: false,
    })

    await expect(record).rejects.toThrowError(/'rejectOnNotFound' option is not supported/)
```

packages/client/tests/functional/findFirstOrThrow/tests.ts
aqrln: Actually, looking at the docs, I think a regexp is not even needed here, it should match a substring and not the whole message if just a string is passed

SevInf:
Closing in favor of #14281, since quite a few things about test setup have changed since then and a lot of the changes in this PR are not needed.

PR #13879, opened 2022-06-17
test: cover missing tests

danstarns:
Some tests were not being executed, this pr includes those tests and then updates the snapshot of one test that has not been updated in a while. 

**SDK:** 

Before: 
176 total

After: 
178 total

**CLI:** 

Before: 
80 total

After: 
81 total


aqrln (approved):
Thank you!

PR #13920, opened 2022-06-20
feat(client): Add `findUniqueOrThrow` and `findFirstOrThrow` methods

### LINKED ISSUE #12159: Consider replacing `rejectOnNotFound`
## Problem

**1. Globally**

```ts
  rejectOnNotFound: true,
})
```

```ts
  rejectOnNotFound: {
    findFirst: {
    },
    findUnique: {
    },
  },
})
```

**2. Locally**

```ts
  where: {
    id: 42,
  },
  rejectOnNotFound: true
})
```

## Suggested solution

## Alternatives

## Additional context

SevInf:
Replacement for `rejectOnNotFound`. Introduces new gategory of model
actions: client-only actions, that are not defined on DMMF, but exist
solely on the client and work by wrapping the requests to the existing
DMMF-defined action and adding logic before or after it's performed.

Why do wrap request function specifically, and not, say, orginal methods
themselves. Two reasons:
- To be 1:1 replacement to `rejectOnNotFound`, new methods should return
  `PrismaPromise` and do not start the query until awaited
- Fluent API needs to work with new methods the same way it did with
  originals

Deprecate `rejectOnNotFound`. Deprecation warinings will be shown in
editors (via jsdoc comments) as well as logged to console once per
model/action combination

Close #12159


packages/client/src/generation/TSClient/Input.ts
```diff
@@ -19,7 +19,7 @@ export class InputField implements Generatable {
 
     const optionalStr = field.isRequired ? '' : '?'
     const deprecated = field.deprecation
-      ? `@deprecated since ${field.deprecation.sinceVersion} because ${field.deprecation.reason}\n`
+      ? `@deprecated since ${field.deprecation.sinceVersion}: ${field.deprecation.reason}\n`
```
SevInf, line 22: This way it look slightly nicer for any kind of text inside of a `reason`

packages/client/src/runtime/utils/printStack.ts
```diff
@@ -119,7 +119,8 @@ function parseStack({
         params.callsiteStr = ':'
       } else {
         // Why even all this effort? Because if a user calls the client instance "db", we want to be able to also say "db.user.findMany"
-        const prismaClientRegex = /(\S+(create|createMany|updateMany|deleteMany|update|delete|findMany|findUnique)\()/
+        const prismaClientRegex =
+          /(\S+(create|createMany|updateMany|deleteMany|update|delete|findMany|findUnique|findFirst|findUniqueOrThrow|findFirstOrThrow)\()/
```
SevInf, line 123: Wondering if we can build this regex dynamically from `allClientModelActions` array. Some of the actions are omitted here and i am not sure if it's intentional. 

On a tangent, this whole module might be a topic for deep dive, i feel

danstarns: I agree that this needs to be dynamically generated as this is unmaintainable to keep on updating!

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1196,3 +1210,25 @@ function transactionProxy<T>(thing: T, txId: string): T {
...
+  findUnique: 'findUniqueOrThrow',
+  findFirst: 'findFirstOrThrow',
+}
+
+function warnAboutRejectOnNotFound(
+  rejectOnNotFound: RejectOnNotFound,
+  model: string | undefined,
+  action: string,
+): void {
+  if (rejectOnNotFound) {
```
danstarns, line 1224: Do we need this if?

packages/client/src/generation/TSClient/helpers.ts
```diff
@@ -61,7 +62,9 @@ export function getArgs(modelName: string, actionName: DMMF.ModelAction) {
     actionName === DMMF.ModelAction.findMany ||
     actionName === DMMF.ModelAction.findFirst ||
     actionName === DMMF.ModelAction.deleteMany ||
-    actionName === DMMF.ModelAction.createMany
+    actionName === DMMF.ModelAction.createMany ||
+    actionName === 'findFirstOrThrow' ||
```
danstarns, line 66: Is it worth adding a type for this and the one below like the ones above?

packages/client/src/generation/utils.ts
```diff
@@ -138,6 +140,12 @@ export function getModelArgName(modelName: string, action?: DMMF.ModelAction): s
       return `${modelName}FindRawArgs`
     case DMMF.ModelAction.aggregateRaw:
       return `${modelName}AggregateRawArgs`
+    case 'findFirstOrThrow':
+      return `${modelName}FindFirstOrThrowArgs`
```
millsp, line 144: Can we reuse the args from the real findFirst here instead of generating the same types again?

packages/client/src/runtime/clientActions.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function getDmmfActionName(name: ClientModelAction): DMMF.ModelAction {
+  if (isClientOnlyAction(name)) {
+    return clientOnlyActions[name].wrappedAction
+  }
+  return name
+}
+
+export function isClientOnlyAction(action: string): action is ClientOnlyModelAction {
+  return Object.prototype.hasOwnProperty.call(clientOnlyActions, action)
```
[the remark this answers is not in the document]
SevInf: That won't be direct equivalent though. `isClientOnlyAction('toString')` will be `true` if we use `in`.  How about I move all prototype shenanigans into a helper function in internals, so we can do something like `hasOwn(clientOnlyActions, action)` here?

packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -77,6 +90,17 @@ function getOwnKeys(client: Client, dmmfModelName: string) {
...
 
 // tells if a given `action` is valid & available for a `model`
-function isValidActionName(client: Client, dmmfModelName: string, action: string): action is Action {
+function isValidActionName(
+  client: Client,
+  dmmfModelName: string,
+  action: string,
+): action is Action | ClientOnlyModelAction {
+  if (isClientOnlyAction(action)) {
+    return isValidActionName(client, dmmfModelName, clientOnlyActions[action].wrappedAction)
```
millsp, line 99: Is this a recursive call?

packages/client/src/generation/utils.ts
SevInf: The problem here is that `rejectOnNotFound` right now is included into `FindFirstArgs`. I agree that generating huge type just because of one field is not very optimal. The cleanest way to fix it, in my opinion, would be:

- Define base args type, say, `FindFirstBaseArgs` which will include all arguments minus `rejectOnNotFound`
- Define `FindFirstArgs` as `FindFirstBaseArgs & { rejectOnNotFound }`
- Define `FindFirstArgs` as a type alias for `FindFirstBaseArgs`.

Last option will leave us some wiggle room in case in future we'd want to have some arguments specific for `OrThrow` variants.  If you think that's premature, we can ditch base type altogether and simple define `FindFirstArgs` as `FindFirstOrThrowArgs & { rejectOnNotFound }`. WDYT?

packages/client/src/generation/TSClient/helpers.ts
danstarns: Thanks for the explanation!

packages/client/src/generation/utils.ts
millsp: I'm happy with both, though the first proposal feels the cleanest.

aqrln:
Overall it looks very good to me.

I agree with Pierre that it would be nice to avoid duplicating types.

packages/client/src/generation/utils.ts
aqrln: One thing to consider is in the future we'll be dropping `rejectOnNotFound` completely, which will make both types identical.

It might be important to keep this in mind when choosing between the approaches.

packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -31,35 +39,40 @@ export function applyModel(client: Client, dmmfModelName: string) {
 
   // we construct a proxy that acts as the model interface
   return new Proxy(baseObject, {
-    get(target, prop: string): F.Return<ModelAction> | undefined {
+    get(target, prop: string, receiver): F.Return<ModelAction> | undefined {
```
millsp, line 42:
```suggestion
    get(target, prop: string): F.Return<ModelAction> | undefined {
```

packages/client/src/generation/utils.ts
SevInf: @aqrln @millsp Updated, implemented first approach. When we will be removing `rejectOnNotFound`, either both args types could be an alias for the same type (in case we want distinct names for consitency) or we'll skip `OrThrow` generation completely and just re-use them args generated by base variant.

packages/client/src/runtime/core/model/applyClientOnlyWrapper.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+function applyOrThrowWrapper(jsModelName: string, requestCallback: RequestCallback): RequestCallback {
+  return async (requestParams) => {
+    if ('rejectOnNotFound' in requestParams.args) {
+      const { stack } = printStack({
+        originalMethod: requestParams.clientMethod,
+        callsite: requestParams.callsite,
+      })
+      throw new TypeError(`${stack}\n'rejectOnNotFound' option is not supported`)
```
millsp, line 29: Shoudl this be a PrismaClientValidationError ?

packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -35,31 +43,36 @@ export function applyModel(client: Client, dmmfModelName: string) {
       // only allow actions that are valid and available for this model
       if (prop in target || typeof prop === 'symbol') return target[prop]
       if (!isValidActionName(client, dmmfModelName, prop)) return undefined
+      const dmmfActionName = getDmmfActionName(prop as ClientModelAction)
 
+      let requestFn = (params: InternalRequestParams) => client._request(params)
```
millsp, line 48: Would it make sense to have a one-liner here to return `requestFn` instead of asserting that it never happens?
```suggestion
      const requestFn = applyClientOnlyWrapper(prop, jsModelName, (p) => client._request(p))
```

millsp (approved):

aqrln (approved):

aqrln (approved):

PR #13926, opened 2022-06-21
chore(init): change wording for existing .gitignore warning

### LINKED ISSUE #13925: Poor grammar and confusing language in Prisma CLI
### Bug description

The particular message I'm talking about happens when you run `prisma init` in a project with a .gitignore already in it. It says:

```
warn You already have a .gitignore. Don't forget to exclude .env to not commit any secret.
```
The language in `not commit any secret` is poor in grammar, and the usage of the word `exclude` is confusing.

### How to reproduce

1. mkdir prismaissue
2. cd prismaissue
3. touch .gitignore
4. npx prisma init

### Expected behavior

I would suggest changing the prompt to:
```
warn You already have a .gitignore. Don't forget to include .env in it to not commit any private information.
```

### Environment & setup

- OS: Windows
- Database: N/A
- Node.js version: 14.17.6

### Prisma Version

```
prisma                  : 3.15.2
@prisma/client          : 3.15.2
Query Engine (Node-API) : libquery-engine 461d6a05159055555eb7dfb337c9fb271cbd4d7e (at node_modules\@prisma\engines\query_engine-windows.dll.node)
Migration Engine        : migration-engine-cli 461d6a05159055555eb7dfb337c9fb271cbd4d7e (at node_modules\@prisma\engines\migration-engine-windows.exe)
Introspection Engine    : introspection-core 461d6a05159055555eb7dfb337c9fb271cbd4d7e (at node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary           : prisma-fmt 461d6a05159055555eb7dfb337c9fb271cbd4d7e (at node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash    : 461d6a05159055555eb7dfb337c9fb271cbd4d7e
Studio                  : 0.462.0

```

givensuman:
style(cli): Fix typo in existing .gitignore warning for `prisma init` package

Closes #13925

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -262,7 +262,7 @@ export class Init implements Command {
         warnings.push(
           `${chalk.yellow(
             'warn',
-          )} You already have a .gitignore. Don't forget to exclude .env to not commit any secret.`,
+          )} You already have a .gitignore. Don't forget to include .env in it to not commit any private information.`,
         )
       } else {
         console.error('Failed to write .gitignore file, reason: ', e)
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -262,7 +262,7 @@ export class Init implements Command {
         warnings.push(
           `${chalk.yellow(
             'warn',
-          )} You already have a .gitignore file. Don't forget to add `.env` in it to not commit any private information.`,
+          )} You already have a .gitignore file. Don't forget to add \`.env\` in it to not commit any private information.`,
         )
       } else {
         console.error('Failed to write .gitignore file, reason: ', e)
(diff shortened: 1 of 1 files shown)


packages/cli/src/Init.ts
```diff
@@ -262,7 +262,7 @@ export class Init implements Command {
         warnings.push(
           `${chalk.yellow(
             'warn',
-          )} You already have a .gitignore. Don't forget to exclude .env to not commit any secret.`,
+          )} You already have a .gitignore. Don't forget to include .env in it to not commit any private information.`,
```
Jolg42, line 265: One can argue that you have to "include" and exclusion rule, so "Don't forget to exclude .env to not commit any secret." here means add a line to your `.gitignore` file.

I'm not sure this change is better than before 🤔

givensuman: I understand the use of the word exclude, and a developer reading it would probably go through the same line of thinking as you have. I would also argue that the version with the least ambiguity is better, however, and the message's intent should be understood intuitively, which is to say with no thinking at all.

Jolg42 (approved):
Thanks!

PR #13875, opened 2022-06-17
fix(client): optional bool dataProxy

danstarns:
The [benchmark](https://github.com/prisma/prisma/blob/main/.github/workflows/benchmark.yml) runner that never seems to pass because of this typescript error: 

```
Run pnpm run bench
> dev-env@ bench /home/runner/work/prisma/prisma
> ts-node scripts/bench.ts | tee output.txt
yarn run v1.22.19
warning dev-env@: The engine "pnpm" appears to be invalid.
$ /home/runner/work/prisma/prisma/node_modules/.bin/ts-node ./packages/client/src/__tests__/benchmarks/huge-schema/huge-schema.bench.ts
/home/runner/work/prisma/prisma/node_modules/.pnpm/ts-node@10.8.1_4b6f1a3cbcae[7](https://github.com/prisma/prisma/runs/6934483652?check_suite_focus=true#step:8:8)475db077e32af0f29eb/node_modules/ts-node/src/index.ts:[8](https://github.com/prisma/prisma/runs/6934483652?check_suite_focus=true#step:8:9)43
    return new TSError(diagnosticText, diagnosticCodes, diagnostics);
           ^
TSError: ⨯ Unable to compile TypeScript:
Error: packages/client/src/utils/getTestClient.ts(53,[9](https://github.com/prisma/prisma/runs/6934483652?check_suite_focus=true#step:8:10)): error TS2741: Property 'dataProxy' is missing in type '{ document: DMMF.Document; generator: GeneratorConfig | undefined; dirname: string; relativePath: string; clientVersion: string; engineVersion: string; relativeEnvPaths: EnvPaths; datasourceNames: any[]; activeProvider: ConnectorType; }' but required in type 'GetPrismaClientConfig'.
    at createTSError (/home/runner/work/prisma/prisma/node_modules/.pnpm/ts-node@[10](https://github.com/prisma/prisma/runs/6934483652?check_suite_focus=true#step:8:11).8.1_4b6f1a3cbcae7475db077e32af0f29eb/node_modules/ts-node/src/index.ts:843:12)
    at reportTSError (/home/runner/work/prisma/prisma/node_modules/.pnpm/ts-node@10.8.1_4b6f1a3cbcae7475db077e32af0f29eb/node_modules/ts-node/src/index.ts:847:19)
[...]

PR #13928, opened 2022-06-21
fix(client): ts-error in benchmark

danstarns:
Related: https://github.com/prisma/prisma/pull/13875

Failing benchmark test: 

```
SError: ⨯ Unable to compile TypeScript:
Error: packages/client/src/generation/TSClient/Model.ts(334,22): error TS2345: Argument of type '(gen: Generatable, edge?: boolean) => string | Promise<string>' is not assignable to parameter of type '(value: Generatable, index: number, array: Generatable[]) => string | Promise<string>'.
  Types of parameters 'edge' and 'index' are incompatible.
    Type 'number' is not assignable to type 'boolean | undefined'.
```

This is because we were passing all the parameters from `Array.map` into the `TS` function when it just accepts one - a `Generatable`.

### DIFF
--- a/packages/client/src/generation/TSClient/Model.ts
+++ b/packages/client/src/generation/TSClient/Model.ts
@@ -331,7 +331,7 @@ ${new PayloadType(this.outputType!, !this.dmmf.typeMap[model.name]).toTS()}
 ${new ModelDelegate(this.outputType!, this.dmmf, this.generator).toTS()}
 
 // Custom InputTypes
-${this.argsTypes.map(TS).join('\n')}
+${this.argsTypes.map((gen) => TS(gen)).join('\n')}
 `
   }
 }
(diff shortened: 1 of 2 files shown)


aqrln (approved):

packages/client/src/generation/TSClient/Model.ts
```diff
@@ -331,7 +331,7 @@ ${new PayloadType(this.outputType!, !this.dmmf.typeMap[model.name]).toTS()}
 ${new ModelDelegate(this.outputType!, this.dmmf, this.generator).toTS()}
 
 // Custom InputTypes
-${this.argsTypes.map(TS).join('\n')}
+${this.argsTypes.map((gen) => TS(gen)).join('\n')}
```
aqrln, line 334: I think you also need to change this in `Count.ts`, it has the same problem.

aqrln: see https://prisma-company.slack.com/archives/C032H34TACR/p1655733194404859?thread_ts=1655727931.637109&cid=C032H34TACR

danstarns: sure

PR #13973, opened 2022-06-23
chore: avoid crash in prisma version

jkomyno:


packages/internals/src/engine-commands/getBinaryVersion.ts
```diff
@@ -1,18 +1,14 @@
...
 
 import { resolveBinary } from '../resolveBinary'
 import { load } from '../utils/load'
 
-const debug = Debug('prisma:getVersion')
-
-const MAX_BUFFER = 1_000_000_000
-
-export async function getVersion(enginePath?: string, binaryName?: BinaryType): Promise<string> {
+export async function getBinaryVersion(enginePath?: string, binaryName?: BinaryType): Promise<string> {
```
janpio, line 11: You are not actually returning a binary version here, but an engine version (binary = executable files, engines = things that represent prisma-engines, including binaries and libraries).
(We know that all the other vars etc are misnamed, but maybe we should prefer not to add more functions to that pile to clean up.)

I suggest `getEngineVersion` instead.

PR #13811, opened 2022-06-14
perf(client): stop exporting dmmf.schema

aqrln:
Closes https://github.com/prisma/client-planning/issues/17

PR #13974, opened 2022-06-23
fix(client): prevent ts-toolbelt from being needed in the runtime

millsp:
Follows https://github.com/prisma/prisma/pull/13811
See https://prisma-company.slack.com/archives/CRGDU19K6/p1655969139124359

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -16,7 +16,6 @@ import { AsyncResource } from 'async_hooks'
 import fs from 'fs'
 import path from 'path'
 import * as sqlTemplateTag from 'sql-template-tag'
-import { O } from 'ts-toolbelt'
 
 import { getPrismaClientDMMF } from '../generation/getDMMF'
 import type { InlineDatasources } from '../generation/utils/buildInlineDatasources'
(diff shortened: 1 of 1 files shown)


packages/client/package.json
```diff
@@ -137,7 +137,8 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "3.16.0-43.81a0ee489e5f8c8ce98440028ea9050092430503"
+    "@prisma/engines-version": "3.16.0-43.81a0ee489e5f8c8ce98440028ea9050092430503",
+    "ts-toolbelt": "9.6.0"
```
Jolg42, line 141: Why bundling doesn't work? (Until now we bundled all deps)

millsp: Bundling currently produces broken types. It's also a huge library, and bundling would include a lot of unused types. It would be very clumsy to copy the sources of eg. `O.Optional`.

millsp: The other alternative is to not correctly type for now and remove the dependency.

Jolg42: So only for that line here?
https://github.com/prisma/prisma/pull/13811/files#diff-09784fc0113a735572cff48990efa35434fb6ec6ec8876845ec9cf423642e5fdR214

millsp: Correct

millsp: Replaced with `Omit` as per your suggestion 👍

SevInf (approved):

aqrln (approved):

PR #13976, opened 2022-06-23
test: Try to parallelize tests on builtkite

SevInf:


Jolg42 (changes requested):
@SevInf 
Note that this might have a huge impact on CI cost as it multiplies the setup time (setup * job) and it's using huge AWS machines behind.
Beside cost, it would also cause agent queuing time affecting the prisma-engines repo and might be more trouble than improvement.

So I'm not sure if it's a good idea to multiply too much the number of jobs here, since we can already do that for GitHub Actions, for cheap (or use buildjet runners like currently done for client tests [which takes 14min all in one job](https://github.com/prisma/prisma/runs/7020021916?check_suite_focus=true))

So just to be safe 🔴 (let's discuss)

SevInf:
Decided we try to parallelize test within the single job instead

PR #13981, opened 2022-06-23
(experimental): TCK tests

danstarns:
Technology Compatibility Kit (TCK) tests. Designed for transparency in our queries, and could help with debugging e2e issues. In short, you perform a query, and then you get given the logs of that query, you should in the test assert those logs using a snapshot so it's very easy to see what's going on under the hood.

This is experimental.

```js
 test('simple findMany', async () => {
  // Give the 'client' a query
  const query = () => client.prisma?.user.findMany({ where: { id: 'some-random-id' } })

  const response = await client.query(query)

  // Assert how many DB queries are performed
  expect(response.logs.length).toEqual(1)

  // Assert what the query looks like
  expect(response.logs[0].query).toMatchInlineSnapshot(
    `SELECT "public"."User"."id", "public"."User"."username" FROM "public"."User" WHERE "public"."User"."id" = $1 OFFSET $2`,
  )

  // Assert what the params looks like
  expect(response.logs[0].params).toMatchInlineSnapshot(`["some-random-id",0]`)
})
```

Just to keep it really simple this suite only tests PostgreSQL with simple CRUD operations, the matrix-like nature could be harder to implement with something like this, however, maybe it’s something that could fit into the functional suite?

**How to run?**

```
$ cd ./packages/tck-tests && pnpm run test
```

**How to update snapshots?**

```
$ cd ./packages/tck-tests && pnpm run test -- -u
```


packages/tck-tests/src/tests/create.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    provider: 'postgresql',
+    uri: process.env.TEST_POSTGRES_URI as string,
+  })
+
+  beforeAll(async () => {
+    await client.generate()
+  })
+
+  test('simple create', async () => {
+    const query = () => client.prisma?.user.create({ data: { id: 'some-random-id', username: 'some-random-username' } })
```
aqrln, line 31: What do you think about passing `prisma` as the argument of the `client.query` callback so that it is self-contained and doesn't need to depend on the same instance of `TestClient` via a closure?

aqrln: You can also do the null check inside `client.query` itself if it's necessary, so that the API will look like this basically:

```ts
    const query = (prisma) => prisma.user.create({ data: { id: 'some-random-id', username: 'some-random-username' } })

    const response = await client.query(query)
```

PR #13912, opened 2022-06-20
chore(deps): update engines to 3.16.0-28.p4-scalar-list-defaults-b10480c0399eac3960d9b844bb6f42a838338d31

aqrln:
Ref: https://github.com/prisma/prisma-engines/pull/2946

Issue #13321, opened 2022-05-12
Implement scalar list defaults proposal in query engine and client


tomhoule:
https://www.notion.so/prismaio/Proposal-Scalar-List-Defaults-94d0891a3355475999f10b270e2a311a

TL;DR we will allow `@default(...)` attributes on scalar list fields in PSL. The behaviour should match that of regular scalar defaults.

Models like the following are currently not valid, but they will become valid on connectors that support scalar lists (mongodb, cockroachdb, postgres):

```prisma
model Pizza {
  id       BigInt   @id @default(autoincrement())
  toppings String[] @default(["cheese", "tomato sauce", "pineapple"])
}
```

On first approximation, the foreseeable client work to make this work is:

- Make the query engine insert these defaults when no value for the field is provided for the user, for consistency with other scalar defaults.
- Determine if this needs to be reflected in the client API. The PSL proposal says "The client will not need to change since it was already possible to not specify values for scalar lists or provide undefined when writing client requests.".

tomhoule:
https://github.com/prisma/prisma-engines/pull/2946 was merged — feel free to close this if you consider the feature done from the client / QE side.

PR #13984, opened 2022-06-23
fix(client): support list defaults in DMMF types

aqrln:
Ref: https://github.com/prisma/prisma/issues/13321
Ref: https://github.com/prisma/prisma-engines/pull/2946
Ref: https://github.com/prisma/prisma/pull/13912
Ref: https://github.com/prisma/prisma/pull/13934

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -244,7 +244,8 @@ exports.Prisma.DScalarFieldEnum = makeEnum({
   byteA: 'byteA',
   xml: 'xml',
   json: 'json',
-  jsonb: 'jsonb'
+  jsonb: 'jsonb',
+  list: 'list'
 });
 
 exports.Prisma.EScalarFieldEnum = makeEnum({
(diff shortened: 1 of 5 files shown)


SevInf (approved):

millsp (approved):
Can you please create an issue so that we can add tests for this later if we don't have the time to do it now?

aqrln:
@millsp 

> Can you please create an issue so that we can add tests for this later if we don't have the time to do it now?

Sure! Do you see anything else that needs to be tested on the client side here btw (except the `exhaustive-schema` test updated in this PR)? I think what's probably missing is an introspection test in `packages/migrate`, but that would be an issue for the Schema Team then.

millsp:
I think that we should have basic queries on lists like filtering, and setting values. That would be similar to what we've done in mongo composites. See the new input types:
* https://github.com/prisma/prisma/pull/13984/files#diff-5ad6f4abad4b644d8dd981d5de9d33c198f960f78733482a7021c420174c4d7dR20291-R20295
* https://github.com/prisma/prisma/pull/13984/files#diff-5ad6f4abad4b644d8dd981d5de9d33c198f960f78733482a7021c420174c4d7dR19788-R19794

aqrln:
I took a quick look, and

> * https://github.com/prisma/prisma/pull/13984/files#diff-5ad6f4abad4b644d8dd981d5de9d33c198f960f78733482a7021c420174c4d7dR20291-R20295

I don't see any tests for this.

> * https://github.com/prisma/prisma/pull/13984/files#diff-5ad6f4abad4b644d8dd981d5de9d33c198f960f78733482a7021c420174c4d7dR19788-R19794

This is tested in `packages/client/src/__tests__/integration/happy/scalar-list`

aqrln:
https://github.com/prisma/prisma/issues/13996

millsp:
Thank you :100:

PR #14032, opened 2022-06-28
test(client): port prisma promise

danstarns:
Coverage for more methods than just create. 

Closes: https://github.com/prisma/client-planning/issues/34


packages/client/tests/functional/prisma-promise/tests.ts
```diff
@@ -0,0 +1,314 @@
...
+      const email = faker.internet.email()
+
+      return prisma.user.create({
+        data: {
+          email,
+        },
+      })
+    }
+
+    test('repeated calls to .then', async () => {
```
millsp, line 19: Can we extract the copy/pasted `test` cases into a separate function?

millsp (dismissed):
Would you mind adding missing cases for `aggregate`, `count`, etc...?
Also, would you mind adding test cases for methods staring with `$`?

SevInf:
For the future, could we note that cases are missing and save their implementation for the future project? For this project we said we are focusing on porting only, without filling in the gaps.

SevInf (approved):

PR #13737, opened 2022-06-09
fix(internals): avoid crash in prisma version, rename 'getVersion' -> 'getEngineVersion'

jkomyno:
Closes https://github.com/prisma/prisma/issues/14014.

PR #14053, opened 2022-06-29
fix: prisma --version (#13737)

### LINKED ISSUE #14014: `prisma --version` crashes if `openssl` isn't properly installed
Compare `prisma --version` on a system where `openssl` is properly installed, with the same command on another system where `openssl` isn't available.

## With proper `openssl` installation

`prisma --version`

image

## Without `openssl` support

`prisma --version`

image

## Proposed Solution

Regardless of the errors that might arise while reading the versions and hashes of every internal package and engine `prisma` depends on, we should still try to display a table of all the versions we retrieved to `stdout`, and the list of possible errors we accumulated into `stderr`. If at least one error arises, let the process terminate with exit code `1`, but only after displaying a version table as complete as possible.

jkomyno:
Essentially, PR [#13737](https://github.com/prisma/prisma/pull/13737) without breaking ecosystem-tests.

Closes https://github.com/prisma/prisma/issues/14014.


packages/internals/src/engine-commands/getEnginesMetaInfo.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * An engine path read from the environment is valid only if it exists on disk.
+ * @param pathFromEnv engine path read from process.env
+ */
+function isPathFromEnvValid(pathFromEnv: string | undefined): pathFromEnv is string {
+  // note: The following wouldn't infer correctly:
+  // return Boolean(pathFromEnv) && fs.existsSync(pathFromEnv)
+  return !!pathFromEnv && fs.existsSync(pathFromEnv)
```
jkomyno, line 134:
```suggestion
  return !!pathFromEnv && fs.existsSync(pathFromEnv)
```

packages/internals/src/engine-commands/getEnginesMetaInfo.ts
```diff
@@ -0,0 +1,168 @@
...
+  return [versionMessage, errors] as const
+}
+
+/**
+ * An engine path read from the environment is valid only if it exists on disk.
+ * @param pathFromEnv engine path read from process.env
+ */
+function isPathFromEnvValid(pathFromEnv: string | undefined): pathFromEnv is string {
+  return !!pathFromEnv && fs.existsSync(pathFromEnv)
+}
```
jkomyno, line 133: Note: The following wouldn't infer types correctly (typescript prefers using `!!` to `Boolean(-)`):

```ts
return Boolean(pathFromEnv) && fs.existsSync(pathFromEnv)
```

jkomyno:
Most likely, an invocation of [these lines](https://github.com/prisma/prisma/blob/a2efd22592013e5031734179f2fd494cb61cb0e2/packages/internals/src/engine-commands/getEngineVersion.ts#L17-L23) from inside `pkg` caused a segmentation fault:

```ts
export async function getEngineVersion(enginePath?: string, binaryName?: BinaryType): Promise<string> {
  if (!binaryName) {
    binaryName = getCliQueryEngineBinaryType()
  }
  // >>> from here:  
  enginePath = await resolveBinary(binaryName, enginePath)
  if (binaryName === BinaryType.libqueryEngine) {
    await isNodeAPISupported()

    const QE = load<NodeAPILibraryTypes.Library>(enginePath)
    return `libquery-engine ${QE.version().commit}`
  } else {
    // ...
  }
}
```

The previous code in https://github.com/prisma/prisma/pull/13737 assumed that the engine version could be read in a logical `try / catch` expression, resulting in either a string, or an error. Apparently, if `enginePath` is `undefined`, the execution of the above lines fails badly, causing the `require` defined inside `load` to read memory it's not supposed to read, thus causing a segmentation fault.

packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
```diff
@@ -38,6 +40,18 @@ Studio                  : STUDIO_VERSION
...
+Query Engine (Binary) : query-engine ENGINE_VERSION (at sanitized_path/query-engine-TEST_PLATFORM, resolved by PRISMA_QUERY_ENGINE_BINARY)
+Migration Engine      : migration-engine-cli ENGINE_VERSION (at sanitized_path/migration-engine-TEST_PLATFORM, resolved by PRISMA_MIGRATION_ENGINE_BINARY)
+Introspection Engine  : introspection-core ENGINE_VERSION (at sanitized_path/introspection-engine-TEST_PLATFORM, resolved by PRISMA_INTROSPECTION_ENGINE_BINARY)
+Format Binary         : prisma-fmt ENGINE_VERSION (at sanitized_path/prisma-fmt-TEST_PLATFORM, resolved by PRISMA_FMT_BINARY)
+Default Engines Hash  : ENGINE_VERSION
+Studio                : STUDIO_VERSION
+`;
+
+exports[`version version with custom binaries 2`] = `
+prisma                : 0.0.0
```
Jolg42, line 54: This is weird, only one snapshot is expected for me here 🤔 
(as there is only one expect `expect(cleanSnapshot(data.stdout, version)).toMatchSnapshot()` for the `version with custom binaries` test)

Jolg42: Maybe one more snapshot update round is needed, seems that some tests are failing around Version tests

Jolg42 (approved):

PR #14074, opened 2022-06-30
fix(*): use process.env value or default for RUST_LOG RUST_BACKTRACE

### LINKED ISSUE #13931: migrate-cli: do not override RUST_LOG from the environment
We want to be able to control engines logging independently of CLI logging. Logs are not used for any business logic at present, so there is seemingly no valid reason to override RUST_LOG.

Jolg42:
closes https://github.com/prisma/prisma/issues/13931


tomhoule (approved):
lgtm!

packages/migrate/src/MigrateEngine.ts
```diff
@@ -192,10 +192,14 @@ export class MigrateEngine {
...
-            ...rest,
-            SERVER_ROOT: this.projectDir,
+            // The following environment variables can be overridden by the user.
             RUST_LOG: 'info',
             RUST_BACKTRACE: '1',
+            // Take env values from process.env (willl override values set before)
+            ...processEnv,
+            // Set SERVER_ROOT
+            // TODO why is this needed?
+            SERVER_ROOT: this.projectDir,
```
Jolg42, line 202:
```suggestion
```

Jolg42: And removed ✅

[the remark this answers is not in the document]
Jolg42: And removed ✅

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -511,6 +508,9 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
       ...this.env, // user-provided env vars
       ...process.env,
       ...env,
+      // use value from process.env or use default
+      RUST_BACKTRACE: process.env.RUST_BACKTRACE ?? '1',
+      RUST_LOG: process.env.RUST_LOG ?? 'info',
```
SevInf, line 513:
```suggestion
      RUST_BACKTRACE: '1',
      RUST_LOG: 'info',
      ...this.env, // user-provided env vars
      ...process.env,
      ...env,
```

PR #14075, opened 2022-06-30
feat(client): Export NotFoundError in Prisma namespace

### LINKED ISSUE #12159: Consider replacing `rejectOnNotFound`

lodi-g:
The `NotFoundError` used by `findXOrThrow` is not available in the Prisma namespace, which makes it hard to perform error verification. 

This PR adds the necessary code to export the `NotFoundError`, both the type declaration and the runtime code.

Closes #12159

### DIFF
--- a/packages/client/src/generation/TSClient/common.ts
+++ b/packages/client/src/generation/TSClient/common.ts
@@ -26,6 +26,7 @@ const {
   PrismaClientRustPanicError,
   PrismaClientInitializationError,
   PrismaClientValidationError,
+  NotFoundError,
   decompressFromBase64,
   getPrismaClient,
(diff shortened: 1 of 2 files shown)


lodi-g:
Closing because I commited with the wrong email address. I will reopen soon.

PR #14076, opened 2022-06-30
feat(client): Export NotFoundError in Prisma namespace

### LINKED ISSUE #12159: Consider replacing `rejectOnNotFound`

lodi-g:
The `NotFoundError` used by `findXOrThrow` is not available in the Prisma namespace, which makes it hard to perform error verification. 

This PR adds the necessary code to export the `NotFoundError`, both the type declaration and the runtime code.

Ref: https://github.com/prisma/prisma/issues/12159#issuecomment-1167274606

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -43,6 +43,10 @@ Prisma.PrismaClientValidationError = () => {
   throw new Error(\`PrismaClientValidationError is unable to be run in the browser.
 In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues\`,
 )}
+Prisma.NotFoundError = () => {
+  throw new Error(\`NotFoundError is unable to be run in the browser.
+In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues\`,
+)}
 Prisma.Decimal = Decimal
 
 /**
@@ -868,6 +872,7 @@ export namespace Prisma {
   export import PrismaClientRustPanicError = runtime.PrismaClientRustPanicError
   export import PrismaClientInitializationError = runtime.PrismaClientInitializationError
   export import PrismaClientValidationError = runtime.PrismaClientValidationError
+  export import NotFoundError = runtime.NotFoundError
 
   /**
    * Re-export of sql-template-tag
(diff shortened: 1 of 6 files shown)


aqrln (approved):
Thank you!

aqrln (changes requested):
On the second look, I think it will fail at run time because it's still missing in the JS bundle — you have only added it to the TS typings and the browser bundle stub, as far as I can tell.

lodi-g:
Hi @aqrln,

I have tested the runtime by copy/pasting the basic-sqlite reproduction and replacing the script.ts with the following code:


```ts
import { PrismaClient, Prisma } from '.prisma/client'

async function main() {

  try {
    const user = await prisma.user.findUniqueOrThrow({
      where: { email: 'a@a.com ' },
    })

    console.log({ user })
  } catch (e: unknown) {
    if (e instanceof Prisma.NotFoundError) {
      console.error('The error is an instance of Prisma.NotFoundError')
    } else {
      console.error('Unknown error')
    }
  }

  console.log('------------')

  try {
    throw new Prisma.NotFoundError('message')
  } catch (e: unknown) {
    console.error(e)
  }
}

void main().catch((e) => {
  console.log(e.message)
  process.exit(1)
})
```

I have tested it using the following commands:
```sh
❯ rm -rf node_modules && pnpm i && pnpx prisma generate && pnpx ts-node index.ts
Scope: all 2 workspace projects
..                                       | Progress: resolved 16, reused 16, downloaded 0, added 0, done

dependencies:
+ @prisma/client 0.0.0 <- ../../packages/client

devDependencies:
+ @types/node 17.0.21
+ prisma 0.0.0 <- ../../packages/cli
+ ts-node 10.8.1
+ typescript 4.7.4

✔ Generated Prisma Client (0.0.0 | library) to ./node_modules/.prisma/client in 152ms
```
import { PrismaClient } from './node_modules/.prisma/client'
```
The error is an instance of Prisma.NotFoundError
------------
NotFoundError: message
    at main (/private/tmp/prisma/reproductions/not-found-error/index.ts:23:11)
```

In my eyes, it seems to work, but I may be missing something?


Thanks

aqrln:
@lodi-g sorry, I haven't had time to have another look yet. I'll get back to you tomorrow.

In the meantime, would you mind adding small tests for this in `packages/client/tests/functional/findFirstOrThrow/tests.ts` and `packages/client/tests/functional/findUniqueOrThrow/tests.ts`? No worries if you don't want to or don't have time for it, I can do it myself later.

And thank you once again for your contribution, we'll get it merged soon!

aqrln (approved):

solomonhawk:
Great timing, was just reaching for `Prisma.NotFoundError` today. Is this slated to be released in 4.1.0 (around the end of July)?

aqrln:
@solomonhawk yes, it's going to be released in `4.1.0` on July 19.

It's also available in `dev` versions starting with `4.1.0-dev.38` if you want to test it, but keep in mind that using pre-release versions is not recommended in production and there are no stability guarantees about them.

PR #13769, opened 2022-06-12
chore: bring `engines-wrapper` packages back & pnpm 7

millsp:
With this PR, we are bringing the [engine download logic packages](https://github.com/prisma/engines-wrapper/) back into the monorepo. This move will improve maintainability and consistency across packages. In this PR, we can already see the benefit of that. It enabled tree-shaking and the runtime size dropped from `2.6MB` to `1MB`. As a bonus, I am enabling `@swc/jest` with bundling, which means that we can now update all our outdated `cjs` deps to `esm`. We might also need to upgrade to `pnpm@7` because of [a bug in replacing](https://github.com/pnpm/pnpm/pull/4627) `workspace:*` on publish (to be tested via `integration`).

-----

I had to downgrade Windows tests to run on Node 14. The memory leaks are [too high on Node 16+](https://github.com/prisma/prisma/runs/7028259912?check_suite_focus=true). The extra memory leaks are caused by `@swc/jest`, which pushed it just over the tipping point on Windows. I think it will only be a matter of time before we start seeing memory issues on Linux too (as we add more and more tests), and it might make sense to plan work on the solution found in https://www.notion.so/prismaio/Memory-leaks-in-Jest-954dab67a40a465aa9418b5011254d64. I believe the priority of getting ESM upgrades is higher than testing windows on Node 16 (given that we also support Node 14).

-----

🟢 [ecosystem-tests](https://github.com/prisma/ecosystem-tests/actions/runs/2583417690)

PR #14121, opened 2022-07-04
ci: test fetch-engine & get-platform & engines packages

### LINKED ISSUE #141: Add CodeCov
Similar to `prisma/prisma`: https://github.com/prisma/prisma/search?q=codecov

Jolg42:
Looks like we missed that during review of https://github.com/prisma/prisma/pull/13769

Closes https://github.com/prisma/engines-wrapper/issues/141


Jolg42:
Note: It was not tested on Windows before and it fails with
https://github.com/prisma/prisma/runs/7179867723?check_suite_focus=true#step:25:793
```
  ● download › basic download all current binaries
    expect(received).toContain(expected) // indexOf
    Expected substring: "ca8f3da9f74523f1c4b16dc99b81f058c07a8413"
    Received string:    "query-engine da41d2bb3406da22087b849f0e911199ba4fbf11"
      93 |
      94 |     // Check that all binaries git hash are the same
    > 95 |     expect(await getVersion(queryEnginePath)).toContain(CURRENT_BINARIES_HASH)
         |                                               ^
      96 |     expect(await getVersion(introspectionEnginePath)).toContain(CURRENT_BINARIES_HASH)
      97 |     expect(await getVersion(migrationEnginePath)).toContain(CURRENT_BINARIES_HASH)
      98 |     expect(await getVersion(prismafmtPath)).toContain(CURRENT_BINARIES_HASH)
      at Object.<anonymous> (src/__tests__/download.test.ts:95:47)
```

packages/fetch-engine/src/__tests__/node-api.test.ts
```diff
@@ -49,64 +50,64 @@ describe('download', () => {
...
-  Object {
-    "name": "query_engine-windows.dll.node",
-    "size": "X",
-  },
-]
-`)
+      Array [
+        Object {
+          "name": ".gitkeep",
+          "size": "X",
```
janpio, line 56: Is this size correct?

packages/fetch-engine/src/__tests__/node-api.test.ts
```diff
@@ -10,16 +10,17 @@ const CURRENT_BINARIES_HASH = enginesVersion
 
 jest.setTimeout(120_000)
 
-describe('download', () => {
+const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
+const isNodeAPI = process.env.PRISMA_CLI_QUERY_ENGINE_TYPE !== 'binary'
+
+describeIf(isNodeAPI)('download', () => {
```
janpio, line 16: Is this really NodeAPI specific or does this just guarantee that this is only ran once?

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -10,19 +10,24 @@ import { checkVersionCommand, download, getBinaryName, getVersion } from '../dow
...
 
 describe('download', () => {
   beforeEach(async () => {
     // completely clean up the cache and keep nothing
     await cleanupCache(0)
-    await del(__dirname + '/**/*engine*')
-    await del(__dirname + '/**/prisma-fmt*')
+    // Make sure to not mix forward and backward slashes in the path
+    // or del glob pattern would not work on Windows
+    const dirname = process.platform === 'win32' ? __dirname.split(path.sep).join('/') : __dirname
```
janpio, line 25: Huh how did this even work before?

Jolg42: Was not running on Windows before, these changes are needed for the windows runner

packages/fetch-engine/src/__tests__/node-api.test.ts
Jolg42: We put the X instead of the size here, because we don't check for the size but for all the artifacts in this snapshot of current engine

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -83,112 +59,298 @@ describe('download', () => {
...
-
-    expect(await checkVersionCommand(targetPath)).toBe(true)
-  })
-
-  test('handle non-existent binary target', async () => {
-    await expect(
-      download({
-        binaries: {
-          'query-engine': __dirname,
+    const files = getFiles(baseDir).map((f) => ({ ...f, size: 'X' }))
```
aqrln, line 71: Maybe just 
```suggestion
    const files = getFiles(baseDir).map((f) => f.name)
```
to make the inline snapshot shorter?

Just an idea, no strong preference either way.

aqrln (approved):

PR #14125, opened 2022-07-04
chore(internals): remove maxBuffer for getEngineVersion

Jolg42:
Since the default is already very high: 100MB https://github.com/sindresorhus/execa#maxbuffer


aqrln:
Checked that it's also 100 MB in the version we use: https://github.com/sindresorhus/execa/tree/v5.1.1#maxbuffer ✅

aqrln (approved):

PR #14126, opened 2022-07-04
chore(internals): rename getVersion to getVersionFromPackageJson

Jolg42:
More explicit name to not be confused by getVersion which was now renamed to getEnginesVersion

Note: I could not find any usage of it?
We could actually remove it, but I would prefer if someone verifies that in case I missed something.


jkomyno (approved):
Let's go

Issue #13232, opened 2022-05-06
Prisma memory leaks on Multitenant applications


TheVoid-0:
### Bug description

I believe this is a known issue ([#5050](https://github.com/prisma/prisma/issues/5050)), however I'm creating this issue to see if someone can help me  with any workaround.

This is what happens to ALL our applications currently using prisma: 
![image](https://user-images.githubusercontent.com/55706765/167197775-d7e29fca-0098-4bf6-b24d-2d660d4f2c1a.png)
In the image you can see the APIs are restarting once it reaches 600Mbs of memory, and without the memory cap they go on forever until the containers crash.

What have I tried?

I tried reading the prisma client code to see what could be the cause, only thing I found is an engines array on `LibraryEngine.ts` file that gets pushed everytime a new client is instantiated, considering a new client is instantiated each request on a multitenant application, that could clearly cause a memory leak, however I created a build slicing that array every time the client stops the engine to see what happens
```ts
// stop method from LibraryEngine
if (this.libraryStarted) {
      // eslint-disable-next-line no-async-promise-executor
      this.libraryStoppingPromise = new Promise(async (resolve, reject) => {
        try {
          await new Promise((r) => setTimeout(r, 5))
          debug('library stopping')
          await this.engine?.disconnect()
          this.libraryStarted = false
          this.libraryStoppingPromise = undefined
          // Next 3 lines were added;
          const engineIdx = engines.indexOf(this)
          engines.splice(engineIdx, 1)
          console.log(engines)

          debug('library stopped')
          resolve()
        } catch (err) {
          reject(err)
        }
      })
      return this.libraryStoppingPromise
    }
```
Well that clears the engines array, but it didn't solve the issue, even with this modified code the application still keeps draining memory. Wich means I couldn't understand all the code logic, the issue linked mentioned using only one query engine, but that would lock out the multitenancy, because prisma depends on the query engine to connect to each database, right? So a tenant application NEEDS several query engines to exist, but they need to be garbage collected

What can be done to avoid such memory leaks? Is there any workarounds even if it means creating slight modifications inside the prisma code, we could run a custom client built locally for now.

Any help on this is really apreciated!

### How to reproduce

1- Create a loop
2- Instantiate a PrismaClient inside the loop
3- Initialize (start the inner query engine) inside the loop
4- wait until your node process crashes you PC :)

### Expected behavior

[...]

PR #14174, opened 2022-07-06
fix: Allow engine to be GCed correctly

### LINKED ISSUE #8989: Memory leaks in Jest when running tests serially with `nApi` enabled
#### Background
Currently trying to migrate to the library engine for an existing project (>1kloc prisma schema). After enabling the `nApi` flag, I've noticed that Jest will periodically crash while running test suites serially (hundreds of tests spread across almost 30 test suites).

After taking a closer look using jest's `--logHeapUsage` flag, I've found that memory usage goes up by 100MB+ per test suite. About thirds into a test run, jest would be eating up over 2GB of memory and crash soon after.

#### Limitations
Unfortunately, I've found no successful mechanism that would allow running tests in parallel in an isolated environment when using Prisma. I've tried setting up a test environment that creates temporary schemas (since Prisma doesn't seem to allow the usage of `pg_temp`) and applies migrations for each suite, but haven't achieved desirable results.

#### The problem
Instantiating a library engine leads to memory leaks when using jest ([barebones example](https://github.com/driimus/napi-jest/blob/master/tests/logged.test.ts)), which is noticeable when running tests using the `--runInBand` flag. The issue also gets picked up when using `--detectLeaks`.

I've also tested a version of the library engine with logging disabled ([repo](https://github.com/driimus/prisma-engines)) and did not see the issue: neither on a simple instantiation([barebones example](https://github.com/driimus/napi-jest/blob/master/tests/unlogged.test.ts)), nor when using it in a generated prisma client (by manually replacing the path in `node_modules/@prisma/client/runtime/index.js`).

#### Minimal reproduction repo
https://github.com/driimus/prisma-leaks - see the action runs for [logs](https://github.com/driimus/prisma-leaks/runs/3454148207?check_suite_focus=true)

Steps:
1. Enable the `nApi` feature
2. Run some Jest test suites using the `--runInBand` or `-w 1` flag and monitor memory usage (e.g. by also using the `--logHeapUsage` flag)
3. Note the memory usage going up.
4. When running large test suites, it is noted that the runner may crash:
```
<--- Last few GCs --->

[594:0x5d8b870]   327150 ms: Scavenge (reduce) 1912.3 (2075.9) -> 1912.0 (2076.4) MB, 3.8 / 0.0 ms  (average mu = 0.394, current mu = 0.400) allocation failure 
[594:0x5d8b870]   327155 ms: Scavenge (reduce) 1912.7 (2076.4) -> 1912.3 (2076.4) MB, 3.0 / 0.0 ms  (average mu = 0.394, current mu = 0.400) allocation failure 
[594:0x5d8b870]   327162 ms: Scavenge (reduce) 1913.0 (2076.4) -> 1912.6 (2076.9) MB, 3.0 / 0.0 ms  (average mu = 0.394, current mu = 0.400) allocation failure 


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0xb02cd0 node::Abort() [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 2: 0xa1812d node::FatalError(char const*, char const*) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 3: 0xceb72e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 4: 0xcebaa7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 5: 0xeb5485  [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 6: 0xeb5f74  [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 7: 0xec43e7 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 8: 0xec779c v8::internal::Heap::AllocateRawWithRetryOrFailSlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
 9: 0xe89d25 v8::internal::Factory::AllocateRaw(int, v8::internal::AllocationType, v8::internal::AllocationAlignment) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
10: 0xe82934 v8::internal::FactoryBase<v8::internal::Factory>::AllocateRawWithImmortalMap(int, v8::internal::AllocationType, v8::internal::Map, v8::internal::AllocationAlignment) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
11: 0xe84630 v8::internal::FactoryBase<v8::internal::Factory>::NewRawOneByteString(int, v8::internal::AllocationType) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
12: 0x110da42 v8::internal::String::SlowFlatten(v8::internal::Isolate*, v8::internal::Handle<v8::internal::ConsString>, v8::internal::AllocationType) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
13: 0x1097e77 v8::internal::JSRegExp::Initialize(v8::internal::Handle<v8::internal::JSRegExp>, v8::internal::Handle<v8::internal::String>, v8::base::Flags<v8::internal::JSRegExp::Flag, int>, unsigned int) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
14: 0x10987ff v8::internal::JSRegExp::Initialize(v8::internal::Handle<v8::internal::JSRegExp>, v8::internal::Handle<v8::internal::String>, v8::internal::Handle<v8::internal::String>) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
15: 0x120b798 v8::internal::Runtime_RegExpInitializeAndCompile(int, unsigned long*, v8::internal::Isolate*) [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
16: 0x15cddf9  [/home/driimus/.nvm/versions/node/v16.7.0/bin/node]
Aborted
```

### Expected behavior

_No response_

### Prisma information

Schema: https://github.com/driimus/prisma-leaks/blob/main/prisma/schema.prisma

### Environment & setup

- OS: `macOS`, `debian`
- Database: `PostgreSQL`
- Node.js version: `v16.7.0` , `LTS`

### Prisma Version

```
prisma                : 2.30.0
@prisma/client        : 2.30.0
Current platform      : debian-openssl-1.1.x
Query Engine (Binary) : query-engine 60b19f4a1de4fe95741da371b4c44a92f4d1adcb (at node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x)
Migration Engine      : migration-engine-cli 60b19f4a1de4fe95741da371b4c44a92f4d1adcb (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine  : introspection-core 60b19f4a1de4fe95741da371b4c44a92f4d1adcb (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary         : prisma-fmt 60b19f4a1de4fe95741da371b4c44a92f4d1adcb (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash  : 60b19f4a1de4fe95741da371b4c44a92f4d1adcb
Studio                : 0.422.0
Preview Features      : nApi
```

SevInf:
In combination with napi-rs fixes this significantly improves jest
memory situation. Fix consist of 3 parts:

- Remove global engines array. For instance counter we use simple number
  variable. For exit hooks: we don't retain the engine unless it has
  `beforeExitHook`.
- Logging callback uses weak reference to engine class.
- In test setup: intorduce helper function for creating extra instances
  of test client, that ensures that this instance would be correctly
  cleaned up at the end of the test.

Fix #8989, #13232

### DIFF
--- a/packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
+++ b/packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
@@ -11,48 +11,48 @@ Invalid \`prisma.$executeRaw()\` invocation:
 exports[`interactive-transactions (cockroachdb) () batching rollback 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:234:19
+/client/tests/functional/interactive-transactions/tests.ts:237:19
 
-  231  */
-  232 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
-  233   const result = prisma.$transaction([
-→ 234     prisma.user.create(
+  234  */
+  235 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
+  236   const result = prisma.$transaction([
+→ 237     prisma.user.create(
   Unique constraint failed on the fields: (\`email\`)
 `;
 
 exports[`interactive-transactions (cockroachdb) () rollback query 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:168:25
+/client/tests/functional/interactive-transactions/tests.ts:171:25
 
-  165   },
-  166 })
-  167 
-→ 168 await prisma.user.create(
+  168   },
+  169 })
+  170 
+→ 171 await prisma.user.create(
   Unique constraint failed on the fields: (\`email\`)
 `;
 
 exports[`interactive-transactions (mongodb) () batching rollback 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:234:19
+/client/tests/functional/interactive-transactions/tests.ts:237:19
 
-  231  */
-  232 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
-  233   const result = prisma.$transaction([
-→ 234     prisma.user.create(
+  234  */
+  235 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
+  236   const result = prisma.$transaction([
+→ 237     prisma.user.create(
   Unique constraint failed on the constraint: \`User_email_key\`
 `;
 
 exports[`interactive-transactions (mongodb) () rollback query 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:168:25
+/client/tests/functional/interactive-transactions/tests.ts:171:25
 
-  165   },
-  166 })
-  167 
-→ 168 await prisma.user.create(
+  168   },
+  169 })
+  170 
+→ 171 await prisma.user.create(
   Unique constraint failed on the constraint: \`User_email_key\`
 `;
 
@@ -67,24 +67,24 @@ Invalid \`prisma.$executeRaw()\` invocation:
 exports[`interactive-transactions (mysql) () batching rollback 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:234:19
+/client/tests/functional/interactive-transactions/tests.ts:237:19
 
-  231  */
-  232 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
-  233   const result = prisma.$transaction([
-→ 234     prisma.user.create(
+  234  */
+  235 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
+  236   const result = prisma.$transaction([
+→ 237     prisma.user.create(
   Unique constraint failed on the constraint: \`User_email_key\`
 `;
 
 exports[`interactive-transactions (mysql) () rollback query 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:168:25
+/client/tests/functional/interactive-transactions/tests.ts:171:25
 
-  165   },
-  166 })
-  167 
-→ 168 await prisma.user.create(
+  168   },
+  169 })
+  170 
+→ 171 await prisma.user.create(
   Unique constraint failed on the constraint: \`User_email_key\`
 `;
 
@@ -99,24 +99,24 @@ Invalid \`prisma.$executeRaw()\` invocation:
 exports[`interactive-transactions (postgresql) () batching rollback 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:234:19
+/client/tests/functional/interactive-transactions/tests.ts:237:19
 
-  231  */
-  232 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
-  233   const result = prisma.$transaction([
-→ 234     prisma.user.create(
+  234  */
+  235 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
+  236   const result = prisma.$transaction([
+→ 237     prisma.user.create(
   Unique constraint failed on the fields: (\`email\`)
 `;
 
 exports[`interactive-transactions (postgresql) () rollback query 1`] = `
 
 Invalid \`prisma.user.create()\` invocation in
-/client/tests/functional/interactive-transactions/tests.ts:168:25
+/client/tests/functional/interactive-transactions/tests.ts:171:25
 
-  165   },
-  166 })
-  167 
-→ 168 await prisma.user.create(
+  168   },
+  169 })
+  170 
+→ 171 await prisma.user.create(
   Unique constraint failed on the fields: (\`email\`)
 `;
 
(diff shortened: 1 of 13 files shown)


packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -191,6 +197,11 @@ You may have to run ${chalk.greenBright('prisma generate')} for your changes to
         this.QueryEngineConstructor = this.library.QueryEngine
       }
       try {
+        // Using strong reference to `this` inside of log callback will prevent
+        // this instance from being GCed while native engine is alive. At the same time,
+        // `this.engine` field will prevent native instance from being GCed. Using weak ref helps
+        // to avoid this cycle
+        const weakThis = new WeakRef(this)
```
SevInf, line 204: Alternative to `WeakRef`: explicitly set `this.engine`  to undefined in `stop` and re-create instance in `start`. This would be more consistent with `BinaryEngine`, where we kill engine process on `stop`, but I was not sure if we concisely diverge here.

packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
```diff
@@ -11,48 +11,48 @@ Invalid \`prisma.$executeRaw()\` invocation:
...
+/client/tests/functional/interactive-transactions/tests.ts:237:19
 
-  231  */
-  232 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
-  233   const result = prisma.$transaction([
-→ 234     prisma.user.create(
+  234  */
+  235 testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
+  236   const result = prisma.$transaction([
+→ 237     prisma.user.create(
```
jkomyno, line 19: Snapshots like this one, with paths to the source code and code lines should be sanitized, as they're subject to frequent changes.

packages/client/tests/functional/invalid-env/tests.ts
```diff
@@ -1,7 +1,11 @@
 import testMatrix from './_matrix'
 
 // @ts-ignore this is just for type checks
-declare let prisma: import('@prisma/client').PrismaClient
+type PrismaClient = import('@prisma/client').PrismaClient
+// @ts-ignore this is just for type checks
+declare const newPrismaClient: NewPrismaClient<typeof PrismaClient>
+// @ts-ignore
+declare let Prisma: typeof import('@prisma/client').Prisma
```
jkomyno, line 8: Maybe we could add a comment about what these variables are for?

packages/engine-core/src/library/ExitHooks.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  install() {
+    if (this.areHooksInstalled) {
+      return
+    }
+
+    this.installHook('beforeExit')
+    this.installHook('exit')
+    this.installHook('SIGINT', true)
+    this.installHook('SIGUSR2', true)
```
jkomyno, line 21: TODO: check whether support for the `SIGUSR2` signal can be dropped

packages/engine-core/src/library/ExitHooks.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+    return this.idToListenerMap.get(id)
+  }
+
+  private installHook(event: string, shouldExit = false) {
+    process.once(event, async (code) => {
+      debug(`exit event received: ${event}`)
+      for (const listener of this.idToListenerMap.values()) {
+        await listener()
+      }
```
jkomyno, line 56: `BeforeExitListener` has type `() => unknown`, I think we should explicitly type it as `() => Promise<unknown>` before `await`-ing on `listener()`. Unless both sync and async listeners are supported

jkomyno (changes requested):

danstarns (approved):

packages/client/tests/functional/invalid-env/tests.ts
aqrln: That needs to be documented once somewhere, duplicating the comment in each test file is not very practical.

aqrln: On the other hand, we do often duplicate "this is just for type checks" comments from the example test together with the `// @ts-ignore` directive. Perhaps we could replace them with a brief but useful description of each variable?

aqrln: (that would be out of scope of this PR though, IMO)

packages/engine-core/src/library/ExitHooks.ts
aqrln: By the way I don't think it's really possible to run async code in `process.on('exit', ...)` hook. `SIGINT` and `SIGTERM` hooks will remove the default handlers so Node.js will only terminate when we explicitly call `process.exit` after setting them, with `'exit'` Node.js will terminate as soon as all listeners finish (and this listener actually finishes before the first `await` as far as its caller is concerned since it doesn't expect an async function).

aqrln (approved):
Amazing work!

packages/client/tests/functional/invalid-env/tests.ts
SevInf: I think what we could do is instead of providing them as globals, provide them as a first argument to `setupTestSuite`, for example:

```ts

// @ts-ignore
testMatrix.setupTestSuite<typeof import('@prisma/client')>((generated, suiteConfig, suiteMeta) => {
// generated.Prisma, generated.client and generated.newPrismaClient are available here
})
```

And `@ts-ignore` can go away once we start generating client for development as well.
I agree that's out of scope of this PR, created a separate issue for that: #14322

packages/engine-core/src/library/ExitHooks.ts
SevInf: I guess not until prisma 5? Since `beforeExit` event is a part of [public API](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/connection-management#exit-hooks)

Jolg42: If we want to remove it in next major, let's mark it deprecated everywhere, then?

packages/engine-core/src/library/ExitHooks.ts
jkomyno: You're right, `exit` doesn't support asynchronous handlers, as the event loop would've been emptied and abandoned already. If we manually set a listener on `beforeExit` (which we do), we can invoke asynchronous handlers instead (they would be queued to the empty event loop, processed by Node.js, and eventually the `exit` event would be emited)

jkomyno: And since `BeforeExitListener` refers to `process.once('beforeExit', fn)`, where `fn: BeforeExitListener`, I guess the right type would be something like:

```typescript
type BeforeExitListener<T = unknown>: () => (Promise<T> | T)
```

SevInf: I don't think we actually need return value for this, so the proper type I beleive is `() => void | Promise<void>` and that's what I changed it to

packages/engine-core/src/library/ExitHooks.ts
SevInf: Will do a proposal on deprecating it in 4.2.0 and removing in 5.0

jkomyno (approved):
Approved as the required changes have been postponed to another PR. LGTM

PR #14193, opened 2022-07-07
feat(client): add context to middleware

tinezmatias:
# Summary

This is my first pr to contribute, if I am missing something let me know.
It is the continuation of the work of a previous pr, it is a feature to add a context that can be used in middleware.
A very common and discussed use case in the community is the multi tenat architecture.

[Original Pr](https://github.com/prisma/prisma/pull/9759)

# How to use it

```JS
prisma.$setContext({ someValue: 'something' })
```

```JS
prisma.$use((params, next, ctx) => {
  // Do something with the ctx
  console.log(ctx)
  return next(params)
})
```

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -689,6 +689,11 @@ export class PrismaClient<
    */
   $use(cb: Prisma.Middleware): void
 
+  /**
+   * Add a set context handler
+   */
+   $setContext(arg: Prisma.QueryMiddlewareContext): any;
+
 /**
    * Allows the running of a sequence of read/write operations that are guaranteed to either succeed or fail as a whole.
    * @example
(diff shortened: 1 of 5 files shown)


packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -1485,8 +1490,14 @@ export namespace Prisma {
   export type Middleware<T = any> = (
     params: MiddlewareParams,
     next: (params: MiddlewareParams) => Promise<T>,
+    context: QueryMiddlewareContext
   ) => Promise<T>
 
+  /**
+   * Context type
+   */
+  export type QueryMiddlewareContext = Record<string, string | number | boolean>;
```
rijk, line 1499: Would like to have `string[]` as well, e.g. for user roles (could also do that as a `number` with bitwise operators, but why not).

packages/client/src/__tests__/integration/happy/exhaustive-schema/__snapshots__/test.ts.snap
```diff
@@ -664,6 +664,11 @@ export class PrismaClient<
    */
   $use(cb: Prisma.Middleware): void
 
+  /**
+   * Add a set context handler
+   */
+   $setContext(arg: Prisma.QueryMiddlewareContext): any;
```
rijk, line 670: Suggestion for API: `$context` instead of `$setContext`

PR #14211, opened 2022-07-08
fix: Always report package size in MB

SevInf:
Benchmark action does not understand different units

### DIFF
--- a/packages/client/src/__tests__/benchmarks/huge-schema/huge-schema.bench.ts
+++ b/packages/client/src/__tests__/benchmarks/huge-schema/huge-schema.bench.ts
@@ -2,6 +2,7 @@
 
 import Benchmark from 'benchmark'
 import execa from 'execa'
+import fs from 'fs'
 import path from 'path'
 
 import { compileFile } from '../../../utils/compileFile'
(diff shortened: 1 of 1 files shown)


aqrln (approved):

packages/client/src/__tests__/benchmarks/huge-schema/huge-schema.bench.ts
```diff
@@ -68,25 +69,9 @@ suite
...
+function getSize(targetPath: string): void {
+  const size = fs.statSync(targetPath).size() / 1024 / 1024
   console.log(
-    `${targetPath.replace('./node_modules/', '').replace('./', '')} size x ${pkgSize.size} ${
-      pkgSize.unit
-    }B ±0.00% (1 runs sampled)`,
+    `${targetPath.replace('./node_modules/', '').replace('./', '')} size x ${size} MB ±0.00% (1 runs sampled)`,
   )
-
-  return pkgSize
```
janpio, line 91: no return necessary any more?

aqrln: @janpio https://github.com/prisma/prisma/pull/14214

PR #14250, opened 2022-07-11
feat: add deno support for data proxy client

kt3k:
This PR adds `--deno` option to `prisma generate` command.

If the option is specified, the generate command additionally generates:

- `<output_dir>/deno.js`
- `<output_dir>/deno.ts`
- `<output_dir>/deno.d.ts`
- `<output_dir>/runtime/deno.js`

and it outputs the example snippet like:

```ts
import { PrismaClient } from './generated/client/deno.ts'
```

`<output_dir>/deno.ts` can work with Deno CLI and Deno Deploy ([deno deploy example](https://github.com/kt3k/prisma-deno-deploy-example)).
It also passes the type check of Deno (`deno check <output_dir>/deno.ts` passes), and the code completion of Model fields is available with Deno's language server (`deno lsp`)

(Note: `--deno` option suppose `output` is always specified in `generator client {}` because Deno doesn't resolve modules under `node_modules`. So I think the default output path (`node_modules/.prisma/client`) doesn't make sense for Deno)

I'm planning to test this change in https://github.com/prisma/ecosystem-tests/pull/2928

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -51,6 +51,7 @@ ${chalk.bold('Options')}
     -h, --help   Display this help message
       --schema   Custom path to your Prisma schema
   --data-proxy   Enable the Data Proxy in the Prisma Client
+        --deno   Enable Deno support in the Prisma Client
        --watch   Watch the Prisma schema and rerun after a change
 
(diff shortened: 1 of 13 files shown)


kt3k:
Closing in favor of #14386

PR #14265, opened 2022-07-12
refactor: Refactor `printStack` function

SevInf:
1. Add unit tests
2. Rename function to `createErrorMessageWithContext`, it describes what
it is doing a little bit better
3. Split context code snippet manipulation into `SourceFileSlice` class
4. Pull in arguments printing into the function and make function return
   string
5. Append error message to the function result within the function.
6. Split function into one that prepares template parameters and one
that prints them so that the structure of the message will be a little
bit more clear
7. Remove a bunch of unused options
8. Make `showColors: false` actually skip colors
9. Build regexp for matching a code dynamicall from the list of all
   known actions
10. Make `CallSite` a class and shift responsibility of determining
specific file location to it

There are some whitespace differences in some of the snapshot. I believe there 2 reasons for that:
- code for padding line number column had off by 1 error, it seems
- the way we display message was inconsistent for different places.


packages/client/src/__tests__/integration/errors/incorrect-column-type/test.ts
```diff
@@ -20,8 +20,9 @@ test('incorrect-column-type', async () => {
             12   ],
             13 })
             14 
-          → 15 await expect(prisma.user.findMany()).rejects.toThrowErrorMatchingInlineSnapshot(
-            Error converting field "name" of expected non-nullable type "String", found incompatible value of "123".
+          → 15 await expect(prisma.user.findMany()).rejects.toThrowErrorMatchingInlineSnapshot(\`
```
janpio, line 23: That additional backtick is intended here?

packages/client/src/__tests__/integration/errors/incorrect-column-type/test.ts
```diff
@@ -20,8 +20,9 @@ test('incorrect-column-type', async () => {
...
+
```
janpio, line 24: Intended the additional empty line?

packages/client/src/runtime/utils/CallSite.ts
```diff
@@ -0,0 +1,64 @@
...
+      return (
+        t.file &&
+        t.file !== '<anonymous>' && // Ignore as we can not read an <anonymous> file
+        !t.file.includes('@prisma') &&
+        !t.file.includes('getPrismaClient') &&
+        !t.file.startsWith('internal/') && // We don't want internal nodejs files
+        !t.methodName.includes('new ') &&
+        !t.methodName.includes('getCallSite') &&
+        !t.methodName.includes('Proxy.') &&
+        t.methodName.split('.').length < 4
```
janpio, line 44: This could use a few more comments (if we actually understand why)

janpio (commented):
Already looks a _lot_ better, and not suprised you fixed a few bugs along the way.

packages/client/src/__tests__/integration/errors/incorrect-column-type/test.ts
SevInf: I think there was another off by 1 error in original code. If there is no `{` in the line, it cut of last character, where the intent was, I think, was to keep the entire line unchanged.
But on the other hand, this snippet is wrong in both "before" and "after" snapshots. It should've ended at `await expect(prisma.user.findMany(`, I'll see if I can quickly fix it.

packages/client/src/__tests__/createErrorMessageWithContext.test.ts
```diff
@@ -0,0 +1,381 @@
...
+      callsite: mockCallsite('/project/some-file.js', 1, 1),
+      message: 'What a terrible failure!',
+    }),
+  ).toMatchInlineSnapshot(`
+
+    Invalid \`prisma.model.findFirst()\` invocation:
+
+
+    What a terrible failure!
+  `)
```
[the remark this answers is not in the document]
SevInf: I think it could happen if either code is bundlled and source map is incorrect, or users doing something weird re-assignments along the lines of:

```ts
let somethingDifferent = prisma.user.findFirst()
somethingDifferent()
```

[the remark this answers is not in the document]
janpio: Do we want to print somethign like `(Invocation location could not be found)` instead? Or even make users create an issue? Or just remove the `:` maybe and one of the two empty lines?

Or actually just leave it as it is? :D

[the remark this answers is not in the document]
SevInf: Sorry, this PR got lost for quite some time. Not sure if you remember the context of this conversation but will reply anyway. I am leaning towards leaving it as is for following reasons:

- "Invocation line could not be found" might be confused for an actual error message
- There is hardly anything actionable for us: if it is a bundling/source map issue, then it is caused by a bundler and we can not do anything about it. If it is weird re-assignment from example above - everything works as expected, I can't imagine we will ever be able to detect that. So, suggesting to open github issue might also not make more sense.
- Error message and the method name is still present and it actuality, raw stack will be present as well. So, even though message might not look as nice, we are at least not using any information present in original exception.

packages/client/src/__tests__/integration/errors/incorrect-column-type/test.ts
SevInf: Update: "quickly = 1.5 month" 🤣

packages/client/src/runtime/utils/CallSite.ts
SevInf: I could not trigger about a half of those back then, but at the same time, not 100% confident we could remove them.

packages/client/src/__tests__/integration/errors/incorrect-column-type/test.ts
janpio: Don't want to know what "soon" would mean then :D

packages/client/src/runtime/utils/CallSite.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+  getLocation(): LocationInFile | null {
+    const stack = this._error.stack
+    if (!stack) {
+      return null
+    }
+    const stackFrames = stackTraceParser.parse(stack)
+    // TODO: more resilient logic to check that it's not relative to cwd
+    const frame = stackFrames.find((t) => {
+      // Here we are trying to find the location in the users code which caused the error
```
janpio, line 34: nit:
```suggestion
      // Here we are trying to find the location in the user's code which caused the error
```

janpio (approved):
Clear improvement 👍

PR #14281, opened 2022-07-13
test: Enable functional tests on mac and windows

SevInf:
Enables functional tests on mac and windows and fixes few discrepances
between them and Windows CI.

- MySQL has different `lowercase_table_names` settings on different
  platforms which some times shows up in error message, when print DB
  error verbatim. Easiest way to work around it is forcing linux to convert
  table names to lowercase.

- `swc` sometimes produces incorrect source maps. In our case, that
  happens only on Windows. Fixed by temprary switching `ts-jest` on
  Windows only.

- Some iTx tests produce different error message depending on timing,
  which didn't really show up before. Fixed by adjusting the tests to
  assert only on consistent porition of those messages.

### DIFF
--- a/.buildkite/publish/docker-compose.yml
+++ b/.buildkite/publish/docker-compose.yml
@@ -103,7 +103,7 @@ services:
 
   mysql:
     image: mysql:8.0
-    command: --default-authentication-plugin=mysql_native_password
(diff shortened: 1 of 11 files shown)


danstarns (approved):
Nice work! This issue has been such a pain, can't believe this fixes it!

PR #14301, opened 2022-07-14
fix(fetch-engine): Prevent macOS from killing node on engine update

### LINKED ISSUE #14058: macOS 12 sometimes kills Node.js process when loading the QE library
When there's a fresh release that not a lot of people have run on their machines yet, macOS 12+ on Apple Silicon often kills the Node.js process when trying to load the library. This happens with dev builds all the time, and recently we started getting reports from people encountering it with public releases.

Logs in `/Applications/Utilities/Console.app` show it was killed because of "Invalid code signature".

When this happens, one can either:

* Sign the Query Engine with their own Apple Developer ID — this immediately fixes the issue locally.

  ```sh
  find . -name libquery_engine-darwin-arm64.dylib.node | xargs codesign -f -s '<YOUR NAME AS IN APPLE DEVELOPER ID CERT>'
  ```

  (for example, `find . -name libquery_engine-darwin-arm64.dylib.node | xargs codesign -f -s 'Oleksii Orlenko'`)

* Wait until the Query Engine hash is whitelisted by macOS Gatekeeper and the issue magically resolves itself. This usually takes up to a day.

* Try using the binary engine instead of the library. macOS is usually not as strict towards the standalone binary as it is towards a dynamic shared library loaded by the Node.js process. Perhaps also because Node.js binary itself is signed?

Ostensibly, this happens because we don't sign our binaries, and should be fixed by signing them with a valid Apple Developer ID.

---

Confirmed cases:

* https://github.com/prisma/prisma/issues/14057
* https://prisma-company.slack.com/archives/C016KUHB1R6/p1655233721657349
* https://prisma-company.slack.com/archives/CUXLS0Z6K/p1637144570062100?thread_ts=1637143821.061000&cid=CUXLS0Z6K

Suspected cases:

* https://www.reddit.com/r/node/comments/vj4atm/prisma_314_315_node_crash/

---

Some "keywords" to make this issue googleable and maybe gather more cases:

```
zsh: killed     npx prisma generate
```

```
zsh: killed     ts-node index.ts
```

```
zsh: killed     npm start
```

SevInf:
Fix #14058

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -1,4 +1,4 @@
-import { BinaryType } from '@prisma/fetch-engine'
+import { BinaryType, overwriteFile } from '@prisma/fetch-engine'
 import type { BinaryPaths, DataSource, DMMF, GeneratorConfig } from '@prisma/generator-helper'
 import type { Platform } from '@prisma/internals'
 import { ClientEngineType, getClientEngineType, getEngineVersion } from '@prisma/internals'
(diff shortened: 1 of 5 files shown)


garrensmith (approved):
This works for me. I tested it locally.

PR #14354, opened 2022-07-18
fix(get-platform): separate type export definition

Jolg42:
Luan found out that it currently fails in studio-code with https://github.com/prisma/studio-code/runs/7385714105?check_suite_focus=true#step:5:35
```
Error: node_modules/@prisma/get-platform/dist/index.d.ts(4,15): error TS1005: ',' expected.

4 export { type Platform, platforms } from './platforms';
```

### DIFF
--- a/packages/get-platform/src/index.ts
+++ b/packages/get-platform/src/index.ts
@@ -1,4 +1,5 @@
 export { getNodeAPIName } from './getNodeAPIName'
 export { getos, getPlatform } from './getPlatform'
 export { isNodeAPISupported } from './isNodeAPISupported'
-export { type Platform, platforms } from './platforms'
+export type { Platform } from './platforms'
+export { platforms } from './platforms'


Jolg42:
Seems this is only supported in TypeScript 4.5
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names
https://github.com/microsoft/TypeScript/pull/44619

Jolg42:
From @luanvdw: studio will upgrade `typescript` dependency, so we can close

luanvdw:
> Luan found out that it currently fails in studio-code with

This was isolated to the `pcw` package within Studio. As @Jolg42 pointed out with it working TypeScript `4.5` Updating the TS version within the `pcw` of Studio resolved the issue and the package built without errors.

Here is the [PR](https://github.com/prisma/studio-code/pull/412) containing this effort.

Issue #13893, opened 2022-06-19
 ERR_PNPM_BAD_PACKAGE_JSON : Invalid name: ".prisma/client"


alexbatis:
### Bug description

After upgrading from prisma 3.14.0 to 3.15.2 with pnpm, I'm unable to install new packages. When trying to install a new package, I get the error : ` ERR_PNPM_BAD_PACKAGE_JSON  /myproject/node_modules/@prisma/client/package.json: Invalid name: ".prisma/client"`

Taking a look at `node_modules/@prisma/client/package.json` it shows the following:
version 3.14.0: 
```
{
    "name": "@prisma/client",
    "version": "3.14.0",
    "description": "Prisma Client is an auto-generated, type-safe and modern JavaScript/TypeScript ORM for Node.js that's tailored to your data. Supports MySQL, PostgreSQL, MariaDB, SQLite databases.",
    "keywords": [
      "orm",
      "prisma2",
      "prisma",
      "client",
      "query",
      "database",
      "sql",
      "postgres",
      "postgresql",
      "mysql",
      "sqlite",
      "mariadb",
      "mssql",
      "typescript",
      "query-builder"
    ],
    "main": "index.js",
    "browser": "index-browser.js",
    "types": "index.d.ts",
    "license": "Apache-2.0",
    "engines": {
      "node": ">=12.6"
    },
    "homepage": "https://www.prisma.io",
    "repository": {
      "type": "git",
      "url": "https://github.com/prisma/prisma.git",
      "directory": "packages/client"
    },
    "author": "Tim Suchanek <suchanek@prisma.io>",
    "maintainers": [
      "Joël Galeran <galeran@prisma.io>",
      "Pierre-Antoine Mills <mills@prisma.io>",
      "Alexey Orlenko <orlenko@prisma.io>"
    ],
    "bugs": "https://github.com/prisma/prisma/issues",
    "files": [
      "README.md",
      "runtime",
      "scripts",
      "generator-build",
      "index.js",
      "index.d.ts",
      "index-browser.js"
    ],
    "devDependencies": {
      "@jest/test-sequencer": "28.1.0",
      "@microsoft/api-extractor": "7.19.3",
      "@opentelemetry/api": "1.0.3",
      "@prisma/debug": "3.14.0",
      "@prisma/engine-core": "3.14.0",
      "@prisma/engines": "3.14.0-36.2b0c12756921c891fec4f68d9444e18c7d5d4a6a",
[...]

PR #14361, opened 2022-07-18
fix(client): error on attempt to generate a package in main client directory

SevInf:
Overriding this directory might lead to all kinds of problems. This
never supposed to work in the first place and it's already broken as of
4.0.0, we just improving error message with a fix suggestion.

Ref #13893

### DIFF
--- a/packages/client/src/__tests__/generation/generator.test.ts
+++ b/packages/client/src/__tests__/generation/generator.test.ts
@@ -40,6 +40,7 @@ describe('generator', () => {
       baseDir: __dirname,
       printDownloadProgress: false,
       skipDownload: true,
+      dataProxy: false,
     })
 
     const manifest = omit<any, any>(generator.manifest, ['version']) as any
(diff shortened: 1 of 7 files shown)


packages/generator-helper/src/generatorHandler.ts
```diff
@@ -24,8 +24,10 @@ export function generatorHandler(handler: Handler): void {
           jsonrpc: '2.0',
           error: {
             code: -32000,
-            message: e.stack || e.message,
-            data: null,
+            message: e.message,
```
SevInf, line 27: This is needed for separating stack-trace and the error message like with normal js errors. Previous code put stacktrace into the message itself.

packages/client/src/__tests__/generation/generator.test.ts
```diff
@@ -181,6 +184,29 @@ describe('generator', () => {
...
+      Generating client into /client/src/__tests__/generation/__fixture__/@prisma/client is not allowed.
+      This package is used by \`prisma generate\` and overwriting its content is dangerous.
+
+      Suggestion:
+      In /client/src/__tests__/generation/main-package-override.prisma replace:
+
+      8 output   = "./__fixture__/@prisma/client"
+      with
+      8 output   = "./__fixture__/.prisma/client"
+
```
aqrln, line 206: Let's also add something along the lines of

> Note:
> You won't need to change your `import` statements. Importing from `@prisma/client` will forward to `.prisma/client`.

to the message.

aqrln (approved):
Great work! LGTM with one suggestion.

PR #14375, opened 2022-07-19
refactor(client): remove unused

danstarns:
Removes two substantial bits of code that are not used anywhere and were getting in the way each time I searched the codebase.


packages/client/src/generation/TSClient/Count.ts
```diff
@@ -3,25 +3,12 @@ import indent from 'indent-string'
 
 import type { DMMFHelper } from '../../runtime/dmmf'
 import { DMMF } from '../../runtime/dmmf-types'
-import {
-  getAggregateArgsName,
```
SevInf, line 7: Do you know if any of those functions can be removed as well?

SevInf (approved):
image

The best kind of changeset. Great job!

packages/client/src/generation/TSClient/Count.ts
danstarns: All used, mostly in `/TsClient/Model.ts`.

PR #14376, opened 2022-07-19
test(client): add orderby test

danstarns:
Tests order by null.

### DIFF
--- a/packages/client/tests/functional/order-by-null/tests.ts
+++ b/packages/client/tests/functional/order-by-null/tests.ts
@@ -1,0 +1,160 @@
+import testMatrix from './_matrix'
+
+// @ts-ignore this is just for type checks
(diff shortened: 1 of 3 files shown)


packages/client/tests/functional/order-by-null/tests.ts
```diff
@@ -0,0 +1,160 @@
+import testMatrix from './_matrix'
+
+// @ts-ignore this is just for type checks
+declare let prisma: import('@prisma/client').PrismaClient
+
+testMatrix.setupTestSuite(
+  () => {
+    beforeAll(async () => {
+      await prisma.user.create({
```
SevInf, line 9: `createMany`?

danstarns: It don't work for SQLite so easier just to do this!

SevInf (approved):

PR #14383, opened 2022-07-19
test: Port logging test to new setup

SevInf:


packages/client/tests/functional/logging-mongo/tests.ts
```diff
@@ -0,0 +1,73 @@
...
+            timestamp: expect.any(Date),
+          },
+        ],
+      ])
+    })
+  },
+  {
+    optOut: {
+      from: ['sqlite', 'postgresql', 'mysql', 'cockroachdb', 'sqlserver'],
+      reason: 'SQL databases tested in logging suite',
```
danstarns, line 70: There isn't much difference between the two tests, wondering if we could merge all the providers into one and not have the separate mongo logging, wdyt?

SevInf: I've started with a single suite, but branched mongo ones into separate one. There are much more differences then might seem from a first glance:

- mongo does not start connection pool and has no message about it
- `target` field for all events has different value for mongo,
- `duration` field is missing
- "begin transaction" and "end transaction" messages are missing.

Before the spit, every assert looked like "if it's mongo check if it's one thing, if it isn't check the different one"

packages/client/tests/functional/logging/tests.ts
```diff
@@ -0,0 +1,102 @@
...
+        [
+          {
+            duration: expect.any(Number),
+            params: expect.any(String),
+            query: expect.any(String),
+            target: 'quaint::connector::metrics',
+            timestamp: expect.any(Date),
+          },
+        ],
+
```
danstarns, line 85:
```suggestion
```

packages/client/tests/functional/logging/tests.ts
```diff
@@ -0,0 +1,102 @@
...
+        [
+          {
+            duration: expect.any(Number),
+            params: expect.any(String),
+            query: transactionBegin,
+            target: 'quaint::connector::metrics',
+            timestamp: expect.any(Date),
+          },
+        ],
+
```
danstarns, line 75:
```suggestion
```

danstarns: I think these spaces will cause a linting error.

packages/client/tests/functional/logging-mongo/tests.ts
```diff
@@ -0,0 +1,73 @@
...
+      const onQuery = jest.fn()
+
+      prisma.$on('query', onQuery)
+
+      await prisma.user.findMany()
+
+      expect(onQuery).toHaveBeenCalledWith({
+        params: expect.any(String),
+        query: expect.any(String),
+        target: 'mongodb_query_connector::query',
```
danstarns, line 32:
```suggestion
        target: expect.any(String),
```
Is it worth doing any for this? They could change it engine side and it will force us to update tests.

danstarns (approved):

danstarns:
Ref: https://github.com/prisma/client-planning/issues/30

millsp:
Ping

SevInf:
Done as a part of PDP Tracing

PR #14384, opened 2022-07-19
fix: Negative zero warning in dependencies

### LINKED ISSUE #14188: Warning during build: equals-negative-zero
### Bug description

Hi,
I get this warning on a regular base. Everything works but i don't think it is intended.
```
 ▲ [WARNING] Comparison with -0 using the "===" operator will also match 0 [equals-negative-zero]
 
     prisma/dist/runtime/index.js:17308:29:
       17308 │             } else if (x === -0) {
             ╵                              ~~
 
   Floating-point equality is defined such that 0 and -0 are equal, so "x === -0" returns true for both 0 and -0. You need to use "Object.is(x, -0)" instead to test for -0.
```

### How to reproduce

Here is a minimal schema:
```
generator client {
  binaryTargets = ["native", "darwin", "debian-openssl-1.1.x", "linux-musl", "windows"]
  output        = "./dist"
}

datasource db {
  provider = "mysql"
  url      = env("APP_MAIN_DATABASE_URL")
}

model customerPictures {
  id                    BigInt    @id @default(autoincrement())
}
```
I am using prisma with nestjs.
Once i build i get following message:

```
▲ [WARNING] Comparison with -0 using the "===" operator will also match 0 [equals-negative-zero]

    prisma/dist/runtime/index.js:17308:29:
      17308 │             } else if (x === -0) {
            ╵                              ~~

  Floating-point equality is defined such that 0 and -0 are equal, so "x === -0" returns true for
  both 0 and -0. You need to use "Object.is(x, -0)" instead to test for -0.
```
Here the relevant generated code snipped from prsima/dist/runtime/index.js

```
      function createNumberConversion(bitLength, typeOpts) {
        if (!typeOpts.unsigned) {
          --bitLength;
        }
        const lowerBound = typeOpts.unsigned ? 0 : -Math.pow(2, bitLength);
        const upperBound = Math.pow(2, bitLength) - 1;
        const moduloVal = typeOpts.moduloBitLength ? Math.pow(2, typeOpts.moduloBitLength) : Math.pow(2, bitLength);
        const moduloBound = typeOpts.moduloBitLength ? Math.pow(2, typeOpts.moduloBitLength - 1) : Math.pow(2, bitLength - 1);
        return function(V, opts2) {
          if (!opts2)
            opts2 = {};
          let x = +V;
          if (opts2.enforceRange) {
            if (!Number.isFinite(x)) {
              throw new TypeError("Argument is not a finite number");
            }
            x = sign2(x) * Math.floor(Math.abs(x));
            if (x < lowerBound || x > upperBound) {
              throw new TypeError("Argument is not in byte range");
            }
            return x;
          }
          if (!isNaN(x) && opts2.clamp) {
            x = evenRound(x);
            if (x < lowerBound)
              x = lowerBound;
            if (x > upperBound)
              x = upperBound;
            return x;
          }
          if (!Number.isFinite(x) || x === 0) {
            return 0;
          }
          x = sign2(x) * Math.floor(Math.abs(x));
          x = x % moduloVal;
          if (!typeOpts.unsigned && x >= moduloBound) {
            return x - moduloVal;
          } else if (typeOpts.unsigned) {
            if (x < 0) {
              x += moduloVal;
            } else if (x === -0) {                         // <------------- Here it is
              return 0;
            }
          }
          return x;
        };
```

### Expected behavior

No message

### Prisma information

```prisma
generator client {
  binaryTargets = ["native", "darwin", "debian-openssl-1.1.x", "linux-musl", "windows"]
  output        = "./dist"
}

datasource db {
  provider = "mysql"
  url      = env("APP_MAIN_DATABASE_URL")
}

model customerPictures {
  id                    BigInt    @id @default(autoincrement())
}
```

### Environment & setup

- OS: Ubuntu
- Database: MariaDB
- Node.js version: v16.15.0

### Prisma Version

```
prisma                  : 3.14.0
@prisma/client          : 4.0.0
Query Engine (Node-API) : libquery-engine 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at ../../.nvm/versions/node/v16.15.0/lib/node_modules/prisma/node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Migration Engine        : migration-engine-cli 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at ../../.nvm/versions/node/v16.15.0/lib/node_modules/prisma/node_modules/@prisma/engines/migration-engine-debian-openssl-3.0.x)
Introspection Engine    : introspection-core 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at ../../.nvm/versions/node/v16.15.0/lib/node_modules/prisma/node_modules/@prisma/engines/introspection-engine-debian-openssl-3.0.x)
Format Binary           : prisma-fmt 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at ../../.nvm/versions/node/v16.15.0/lib/node_modules/prisma/node_modules/@prisma/engines/prisma-fmt-debian-openssl-3.0.x)
Default Engines Hash    : 2b0c12756921c891fec4f68d9444e18c7d5d4a6a
Studio                  : 0.460.0
```

SevInf:
Appears in fetch-engines->node-fetch->whatwg-url->webidl-conversions
package. Ideally, we'd like to be able to update node-fetch@3, but it's
currently blocked. `node-fetch@2` uses whatwg-url for polyfilling `URL`
object if it is not present. Since we support node 14+, we never need
this polyfill and we can patch out node-fetch->whatwg-url dependency.

Fix #14188

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,9 @@
       "allowedVersions": {
         "@octokit/core": ">=3"
       }
+    },
+    "patchedDependencies": {
+      "node-fetch@2.6.7": "patches/node-fetch@2.6.7.patch"
     }
   },
   "lint-staged": {
--- a/patches/node-fetch@2.6.7.patch
+++ b/patches/node-fetch@2.6.7.patch
@@ -1,0 +1,102 @@
+diff --git a/lib/index.es.js b/lib/index.es.js
+index 4852f7cbac354d002a601212332187487deb3afa..b334d377a01bf38afcbb1d608a267cad1c58286e 100644
+--- a/lib/index.es.js
++++ b/lib/index.es.js
+@@ -3,7 +3,6 @@ process.emitWarning("The .es.js file is deprecated. Use .mjs instead.");
+ import Stream from 'stream';
+ import http from 'http';
+ import Url from 'url';
+-import whatwgUrl from 'whatwg-url';
+ import https from 'https';
+ import zlib from 'zlib';
+ 
+@@ -1138,7 +1137,7 @@ Object.defineProperty(Response.prototype, Symbol.toStringTag, {
+ });
+ 
+ const INTERNALS$2 = Symbol('Request internals');
+-const URL = Url.URL || whatwgUrl.URL;
++const URL = Url.URL;
+ 
+ // fix an issue where "format", "parse" aren't a named export for node <10
+ const parse_url = Url.parse;
+@@ -1401,7 +1400,7 @@ AbortError.prototype = Object.create(Error.prototype);
+ AbortError.prototype.constructor = AbortError;
+ AbortError.prototype.name = 'AbortError';
+ 
+-const URL$1 = Url.URL || whatwgUrl.URL;
++const URL$1 = Url.URL;
+ 
+ // fix an issue where "PassThrough", "resolve" aren't a named export for node <10
+ const PassThrough$1 = Stream.PassThrough;
+diff --git a/lib/index.js b/lib/index.js
+index e5b04f107f765db04265c7635d3ca1d86b146fb8..c444206f23a5ed50a3301378eb01eba80938c187 100644
+--- a/lib/index.js
++++ b/lib/index.js
+@@ -7,7 +7,6 @@ function _interopDefault (ex) { return (ex && (typeof ex === 'object') && 'defau
+ var Stream = _interopDefault(require('stream'));
+ var http = _interopDefault(require('http'));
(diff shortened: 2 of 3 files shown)


Jolg42 (approved):
Nice one! Thank you 💚

Jolg42:
I guess, @SevInf that node-fetch@3 is blocked because it's an ESM-only module?

PR #14386, opened 2022-07-19
feat: add deno support, alt approach

kt3k:
This is an alternative approach to #14250

This PR updates `prisma generate` command. When `previewFeatures = ["denoDeploy"]` is specified in the schema, then generate command outputs the client compatible with Deno and Deno Deploy.

---
Summary of changes (all changes only happen when `previewFeatures = ["denoDeploy"]` is specified):
- Updates the generated client scripts to fit to deno
- Generates `runtime/edge-esm.js` which is used by deno
- Shows the error message when `output` is not specified in `generator client`
- Shows the example which is suitable for deno (`import ... from './client/edge.ts'` instead of `import ... from './client/edge'`)


packages/client/src/generation/TSClient/common.ts
```diff
@@ -18,6 +22,24 @@ const {
...
+  NotFoundError,
+  objectEnumValues,
+  PrismaClientInitializationError,
+  PrismaClientKnownRequestError,
+  PrismaClientRustPanicError,
+  PrismaClientUnknownRequestError,
+  PrismaClientValidationError,
+  raw,
+  sqltag,
+} from "${runtimeDir}/${runtimeName}.js"
```
kt3k, line 42: Changed this to ES Module syntax and also added `.js` extension here. This is necessary for Deno, and this also looks ok for cloudflare workers because webpack can resolve this.

packages/client/src/generation/TSClient/common.ts
```diff
@@ -93,15 +115,41 @@ In case this error is unexpected for you, please report it in https://github.com
 }
 
 export const commonCodeTS = ({ runtimeDir, runtimeName, clientVersion, engineVersion }: TSClientOptions) => ({
-  tsWithoutNamespace: () => `import * as runtime from '${runtimeDir}/${runtimeName}';
+  tsWithoutNamespace: (edge = false) => `${
+    // Mock out the runtime types for edge client
+    edge
+      ? `declare namespace runtime {
```
kt3k, line 121: Note

Uses mock namespace for `runtime` in `edge.d.ts`. The original type file (`runtime/index.d.ts`) has `@types/node` dependency and that doesn't work for Deno (I believe that's also something inconvenient for cloudflare workers as they don't have node APIs. So I think using mock here makes sense for both cf workers and deno)

While `runtime` types are mock now, `PrismaClient` and user defined data models still have full typings. So I believe this change doesn't cause any major negative impact to the developer experience.

kt3k:
I believe the CI failures are caused by the flakiness of test cases. I found similar errors in main's CI

- The error in `@prisma/client` `tests/functional/missing-env/tests.ts` [1](https://github.com/prisma/prisma/runs/7409546156?check_suite_focus=true) [2](https://github.com/prisma/prisma/runs/7407524066?check_suite_focus=true)
- The error in `@prisma/generator-helper` `src/__tests__/generatorHandler.test.ts` [3](https://github.com/prisma/prisma/runs/7357069269?check_suite_focus=true)

packages/cli/src/Generate.ts
```diff
@@ -103,6 +104,7 @@ ${chalk.bold('Examples')}
       '--watch': Boolean,
       '--schema': String,
       '--data-proxy': Boolean,
+      '--deno': Boolean,
```
matthewmueller, line 107: Is there anyway to make Deno work without additional configuration? Mentioned before, but I'm hesitant to introduce another flavor of generated client.

matthewmueller: **Update:** just saw your comment in Slack

> I created another flavor (deno.js) because Deno Deploy doesn’t resolve CommonJS module (`require`). Does it make sense to change `edge.js` to use ES Modules instead of CommonJS? If so, I’ll try to make `edge.js` work both on Deno and other edge workers.
> BTW what are the supported workers of edge client? I think it supports cloudflare workers, but is there any other environment it should support?

We're currently targeting Node, Cloudflare and Vercel. Once we have this working, we can add Deno to our ecosystem tests: https://github.com/prisma/ecosystem-tests/tree/dev/dataproxy

kt3k: Ok. I'll remove '--deno' option as well. (What '--deno' option currently does is some validations about generator inputs, but that can be covered by the documentation)

> We're currently targeting Node, Cloudflare and Vercel. Once we have this working, we can add Deno to our ecosystem tests: https://github.com/prisma/ecosystem-tests/tree/dev/dataproxy

OK. I'll check this branch with those environments + deno.

bartlomieju: @kt3k FYI `require` should now work on Deploy, let's check if we can make Prisma work using it

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -129,14 +129,12 @@ ${buildWarnEnvConflicts(edge, runtimeDir, runtimeName)}
...
 exports.PrismaClient = PrismaClient
 Object.assign(exports, Prisma)
+${this.options.runtimeName === 'edge' ? 'export { exports as default, Prisma, PrismaClient }' : ''}
 ${buildNFTAnnotations(dataProxy, engineType, platforms, relativeOutdir)}
 `
     return code
   }
   public toTS(edge = false): string {
-    // edge exports the same ts definitions as the index
-    if (edge === true) return `export * from './index'`
```
millsp, line 138: We can work on not exposing node types and importing `util` which is a side-effect of bundling the `sql-template-tag` library. We could easily write a wrapper around it to make sure these type imports don't leak into the final bundle.

packages/client/helpers/build.ts
```diff
@@ -37,6 +37,7 @@ const edgeRuntimeBuildConfig: BuildOptions = {
   outfile: 'runtime/edge',
   bundle: true,
   minify: true,
+  format: 'esm',
```
millsp, line 40: This could be considered a breaking change, as it might have unintended side effects. I suggest that we would add `edge` to become a folder, with both `cjs` and `esm` files that have respective entrypoints in a `package.json`.

packages/client/src/generation/TSClient/common.ts
millsp: (https://github.com/prisma/prisma/pull/14386/files#r930359993)

packages/client/src/generation/TSClient/common.ts
millsp: (https://github.com/prisma/prisma/pull/14386/files#r930355689)

packages/client/src/generation/TSClient/common.ts
```diff
@@ -328,8 +375,10 @@ type IsObject<T extends any> = T extends Array<any>
 ? False
 : T extends Date
 ? False
-: T extends Buffer
-? False
+${
+  // Edge runtime doesn't have Buffer type
+  edge ? '' : `: T extends Buffer\n? False`
```
millsp, line 380: Nice catch, thanks.

packages/client/src/generation/generateClient.ts
```diff
@@ -125,6 +125,16 @@ export async function buildClient({
   if (dataProxy === true) {
     fileMap['edge.js'] = await JS(edgeTsClient, true)
     fileMap['edge.d.ts'] = await TS(edgeTsClient, true)
+
+    // Generates shims for deno
+    fileMap['deno.ts'] = `import "https://deno.land/std@0.148.0/dotenv/load.ts";
```
millsp, line 130: How does this work? Does it mean you need to import the generated client like `import ... from '.prisma/client/deno.ts'` ?
If that is the case, we can simplify things by not touching `edge` related stuff and have a dedicated deno entry point.

millsp: (Related: I'm also thinking that one day we might have support for N-API (Query Engine) *and* Deno Deploy (Data Proxy), so it might make sense to plan have different entry points from the start).

kt3k: > How does this work? Does it mean you need to import the generated client like import ... from '.prisma/client/deno.ts' ?

Yes.

> If that is the case, we can simplify things by not touching edge related stuff and have a dedicated deno entry point.

My previous PR does that and didn't change anything in edge.js (See https://github.com/prisma/prisma/pull/14250). I got feedback from @matthewmueller that suggests unifying edge.js and deno.ts. So I modified edge.js and made it work on both Deno and Cloudflare (I admit it's a little bit trickier solution than the previous one). I'm open to both directions.

packages/client/src/generation/TSClient/TSClient.ts
kt3k: OK. I'll try to make `runtime/index.d.ts` not requiring Node.js types

When I tried that last week, I was able to get rid of `util` dependency, but `api-extractor` sill added `/// <reference types="node" />` directive at the top of the bundled file (even when there's no node.js type reference in the output), and that part errored in Deno. Do you think it's ok to replace that `/// <reference ...>` part after processing it via `api-extractor`?

kt3k: If we change the type of `QueryEngineConfig['env']` (in `engine-core/src/common/types/QueryEngine.ts`) from `NodeJS.ProcessEnv | Record<string, string>` to `Record<string, string | undefined>`, then node.js type dependency disappears. Do you think it's an acceptable change?

millsp: Sure, totally acceptable.

packages/client/src/generation/generateClient.ts
millsp: If we want to keep a single entry point for `edge`, we could as well do this dynamically the the generated client. Check that `Deno` exists and then polyfill the `process` with `Deno.env` if that is the case.

millsp: Can `import "https://deno.land/std@0.148.0/dotenv/load.ts"` be made dynamic with top-level await?

kt3k:
@millsp Updated the PR. This branch now uses `previewFeatures = ["denoDeploy"]` and doesn't affect existing prisma clients. PTAL

Summary of changes (all changes only happen when `previewFeatures = ["denoDeploy"]` is specified):
- Updates the generated client scripts (`edge.js` and `runtime/index.d.ts`) to fit to deno
- Generates deno specific scripts (`edge.ts` (thin wrapper), `deno-polyfill.js` (small polyfill))
- Generates `runtime/edge-esm.js` which is used by deno
- Shows the error message when `output` is not specified in `generator client`
- Shows the example which is suitable for deno (`import ... from './client/edge.ts'` instead of `import ... from './client/edge'`)

kt3k:
BTW I currently can't build main branch because of `missingDependencies ["is-ci"]` error in `pnpm run build` of `packages/internals`. ( I created this diff from b39cf852fc00a25b59400e2aeba0ca90f868ed39 )

packages/client/src/generation/TSClient/common.ts
millsp: This has been changed in the latest release, we now use `Uint8Array` so we can be runtime agnostic, this can now be removed.

packages/client/src/generation/TSClient/common.ts
```diff
@@ -331,8 +360,12 @@ type IsObject<T extends any> = T extends Array<any>
 ? False
 : T extends Date
 ? False
-: T extends Buffer
-? False
+${
```
millsp, line 363: This can now be removed, we use `Uint8Array` instead of `Buffer`.

packages/cli/src/Generate.ts
```diff
@@ -228,15 +237,15 @@ ${breakingChangesMessage}`
 This might lead to unexpected behavior.
 Please make sure they have the same version.`
             : ''
-
-        hint = `You can now start using Prisma Client in your code. Reference: ${link('https://pris.ly/d/client')}
+        if (!isDenoDeploy) {
```
millsp, line 240: I think this could be a problem once the preview feature fades away. Is there are reliable way to detect if this is being run inside a Deno project?

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -131,7 +131,11 @@ ${buildWarnEnvConflicts(edge, runtimeDir, runtimeName)}
 ${buildDebugInitialization(edge)}
 const PrismaClient = getPrismaClient(config)
 exports.PrismaClient = PrismaClient
-Object.assign(exports, Prisma)
+Object.assign(exports, Prisma)${
+      this.options.generator?.previewFeatures.includes('denoDeploy')
```
millsp, line 135: I think this could be a problem once the preview fades away. I think we mentioned we could work around that by generating special deno entrypoint to be imported like `.prisma/client/deno/edge.ts`

packages/client/src/generation/generateClient.ts
```diff
@@ -135,6 +135,16 @@ export async function buildClient({
     fileMap['edge.d.ts'] = await TS(edgeTsClient, true)
   }
 
+  if (generator?.previewFeatures.includes('denoDeploy')) {
+    // We use modified version of edge.js in denoDeploy mode.
+    fileMap['edge.js'] = await JS(edgeTsClient, true)
```
millsp, line 140: You could pass another boolean here to signify that this is a Deno+Edge generation. Something like
```suggestion
    fileMap['edge.js'] = await JS(edgeTsClient, true, true)
```

packages/client/src/generation/generateClient.ts
```diff
@@ -135,6 +135,16 @@ export async function buildClient({
...
+    fileMap['edge.ts'] = `import 'https://deno.land/std@0.152.0/dotenv/load.ts'
```
millsp, line 141: Assuming that we don't add a Deno folder and keep it this way. How would we name the entry point for the regular Deno runtime? Would it be `index.ts`?

packages/client/src/generation/TSClient/TSClient.ts
millsp: (Solution in comments below)

millsp: https://github.com/prisma/prisma/pull/14386/files#r960846120

packages/client/src/generation/TSClient/common.ts
kt3k: Thanks! Updated

packages/client/src/generation/generateClient.ts
kt3k: Oops, I forgot we talked about `/deno/` directory. I'll move deno specific files there.

packages/cli/src/Generate.ts
kt3k: I think there could be a few options.

- If there's deno.json(c) in the root of the project, then it's deno project
- If there's no `package.json` or `node_modules`, then it's deno project
- If prisma CLI is executed in Deno's [npm compatibility mode](https://deno.com/blog/v1.25#experimental-npm-support), then it's deno project

None of the above are 100% reliable, but we can document what triggers the deno mode.

kt3k:
@Jolg42 I was able to fix the issue when I ran `rm -rf packages/internals/dist` before running `pnpm run build`

millsp:
Oh, I see.  Our dependency check includes all the files from the package. We should at least exclude `dist` folders.

packages/client/src/generation/generateClient.ts
kt3k: Is it better to add the 3rd param to `JS()`? That causes us to pass the option `denoDeploy: boolean` around in many places.

How about adding `denoDeploy?: boolean` option to `TSClientOptions` interface?

kt3k:
I now see an error in my integration test setting (like `ENOENT: no such file or directory, open '/path/to/prisma/reproductions/basic-sqlite/prisma/generated/client/deno/index.d.ts'`. I'll address that tomorrow

packages/client/src/generation/generateClient.ts
millsp: Yes, that's actually better.

millsp:
I pinged you on slack about the deno folder, so that we can have a quick conversation about it.

kt3k:
> I now see an error in my integration test setting (like ENOENT: no such file or directory, open '/path/to/prisma/reproductions/basic-sqlite/prisma/generated/client/deno/index.d.ts'. I'll address that tomorrow

Addressed this issue. The generated client can make request to data proxy in my settings. It also can pass the type checking by deno's type checker (type completions of prisma client, models, etc are available).

Now the `prisma generate` creates directory `/deno/`, and deno specific files are created in it. The main entrypoint for deno deploy client is `${output}/deno/deploy.ts` for now. I'm open to change it back to the root if we prefer it. What do you think?

packages/client/src/generation/TSClient/common.ts
```diff
@@ -9,10 +9,31 @@ export const commonCodeJS = ({
...
+  getPrismaClient,
+  sqltag,
+  empty,
+  join,
+  raw,
+  Decimal,
+  Debug,
+  objectEnumValues,
+  makeStrictEnum
+} from '${runtimeDir}/edge-esm.js'`
```
millsp, line 35: Can this be removed if we add a `runtimeDirs.deno`?

millsp: Ah right, I missed the import keyword

janpio:
(Sorry, I tried to resolve the conflicts and had messed up - so had to force push the previous state again. No changes.)

packages/client/src/generation/generateClient.ts
```diff
@@ -135,6 +135,25 @@ export async function buildClient({
...
+      ...tsClientOptions,
+      dataProxy: true, // edge only works w/ data proxy
+      runtimeName: 'index.d.ts',
+      runtimeDir: '../' + runtimeDirs.edge,
+      denoDeploy: true,
+    })
+
+    fileMap['deno/edge.js'] = await JS(denoTsClient, true)
+    fileMap['deno/index.d.ts'] = await TS(denoTsClient)
+    fileMap['deno/edge.ts'] = `import 'https://deno.land/std@0.152.0/dotenv/load.ts'
```
millsp, line 150: I this is loading `.env` files, then this is not needed. Edge clients don't support this:
https://www.prisma.io/docs/data-platform/data-proxy#edge-runtimes

packages/client/src/generation/generateClient.ts
```diff
@@ -135,6 +135,25 @@ export async function buildClient({
...
+import './polyfill.js'
+// @deno-types="./index.d.ts"
+export * from './edge.js'`
+    fileMap['deno/polyfill.js'] = 'globalThis.process = { env: Deno.env.toObject() }; globalThis.global = globalThis'
```
millsp, line 154: I'm wondering if this would also work?
```suggestion
    fileMap['deno/index.js'] = await JS(denoTsClient, true)
    fileMap['deno/edge.d.ts'] = await TS(denoTsClient)
    fileMap['deno/edge.js'] = `import 'https://deno.land/std@0.152.0/dotenv/load.ts'
import './polyfill.js'
export * from './index.js'`
    fileMap['deno/polyfill.js'] = 'globalThis.process = { env: Deno.env.toObject() }; globalThis.global = globalThis'
```

packages/client/src/generation/TSClient/common.ts
millsp, line 35: Could we do this?
```suggestion
} from '${runtimeDir}/${runtimeName}'`
```

packages/client/src/generation/generateClient.ts
kt3k: Renaming `deno/edge.js` -> `deno/index.js` and `deno/index.d.ts` -> `deno/edge.d.ts` is fine, but renaming `deno/edge.ts` to `deno/edge.js` is problematic. Deno doesn't automatically pick up types from `deno/edge.d.ts` for `deno/edge.js`. If we do the above change, the users don't get the type completion anymore. The entrypoint for the users needs to be `.ts` to make types available

millsp: Thanks, this sentence is a good summary for me.

> Deno doesn't automatically pick up types from deno/edge.d.ts

packages/cli/src/Generate.ts
```diff
@@ -206,6 +206,15 @@ Please run \`prisma generate\` manually.`
       )
       let hint = ''
       if (prismaClientJSGenerator) {
+        const generator = prismaClientJSGenerator.options?.generator
+        const isDenoDeploy = generator?.previewFeatures.includes('denoDeploy') && !!globalThis.Deno
+        if (isDenoDeploy && !generator?.isCustomOutput) {
+          throw new Error(`Can't find output dir for generator ${chalk.bold(
+            generator?.name,
+          )} with provider ${chalk.bold(generator?.provider.value)}.
+When you enable denoDeploy preview feature, you need to define \`output\` in client generator section of schema.prisma file.`)
```
janpio, line 215:
```suggestion
        const isDenoPreview = generator?.previewFeatures.includes('denoDeploy') && !!globalThis.Deno
        if (isDenoPreview && !generator?.isCustomOutput) {
          throw new Error(`Can't find output dir for generator ${chalk.bold(
            generator?.name,
          )} with provider ${chalk.bold(generator?.provider.value)}.
When you enable the 'deno' preview feature, you need to define \`output\` in the 'generator' section of your 'schema.prisma' file.`)
```

packages/cli/src/Generate.ts
janpio, line 240:
```suggestion
        if (!isDenoPreview) {
```

packages/cli/src/Generate.ts
```diff
@@ -246,8 +255,18 @@ ${chalk.dim('```')}
...
+          }${breakingChangesStr}${versionsWarning}`
+        } else {
+          hint = `You can now start using Prisma Client in your code. Reference: ${link('https://pris.ly/d/client')}
+${chalk.dim('```')}
+${highlightTS(`\
+import { PrismaClient } from '${importPath}/deno/edge.ts'`)}
+${chalk.dim('```')}
+
+You will need a Prisma Data Proxy connection string. See documentation: ${link('https://pris.ly/d/data-proxy')}
+${breakingChangesStr}${versionsWarning}`
```
janpio, line 268: I don't like that deno previw is in the `else`. Can we rewrite this that this also reacts to `isDenoPreview` @millsp?

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -66,6 +67,7 @@ export class TSClient implements Generatable {
       runtimeName,
       datasources,
       dataProxy,
+      denoDeploy,
```
janpio, line 70:
```suggestion
      deno,
```

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -132,7 +134,7 @@ ${buildWarnEnvConflicts(edge, runtimeDir, runtimeName)}
 ${buildDebugInitialization(edge)}
 const PrismaClient = getPrismaClient(config)
 exports.PrismaClient = PrismaClient
-Object.assign(exports, Prisma)
+Object.assign(exports, Prisma)${denoDeploy ? '\nexport { exports as default, Prisma, PrismaClient }' : ''}
```
janpio, line 137:
```suggestion
Object.assign(exports, Prisma)${deno ? '\nexport { exports as default, Prisma, PrismaClient }' : ''}
```

packages/client/src/generation/TSClient/common.ts
```diff
@@ -9,10 +9,31 @@ export const commonCodeJS = ({
   browser,
   clientVersion,
   engineVersion,
-}: TSClientOptions): string => `
+  denoDeploy,
+}: TSClientOptions): string => `${denoDeploy ? 'const exports = {}' : ''}
```
janpio, line 13:
```suggestion
  deno,
}: TSClientOptions): string => `${deno ? 'const exports = {}' : ''}
```

packages/client/src/generation/generateClient.ts
```diff
@@ -135,6 +135,25 @@ export async function buildClient({
     fileMap['edge.d.ts'] = await TS(edgeTsClient, true)
   }
 
+  if (generator?.previewFeatures.includes('denoDeploy') && !!globalThis.Deno) {
```
janpio, line 138:
```suggestion
  if (generator?.previewFeatures.includes('deno') && !!globalThis.Deno) {
```

packages/client/src/generation/generateClient.ts
```diff
@@ -222,6 +241,9 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
 
   await makeDir(finalOutputDir)
   await makeDir(path.join(outputDir, 'runtime'))
+  if (generator?.previewFeatures.includes('denoDeploy') && !!globalThis.Deno) {
```
janpio, line 244:
```suggestion
  if (generator?.previewFeatures.includes('deno') && !!globalThis.Deno) {
```

janpio:
New engine to use: `[4.5.0-38.71723d0f37369be033e71bd7cd314749d8348f9f](https://npmjs.com/package/@prisma/engines-version/v/4.5.0-38.71723d0f37369be033e71bd7cd314749d8348f9f)`

millsp:
Work continued here https://github.com/prisma/prisma/pull/15281

millsp:
We've merged this work via https://github.com/prisma/prisma/pull/15281. Kudos to @kt3k, we're super thankful for your contribution and good work! :partying_face: :tada:

kt3k:
@millsp So glad to see this landed! Thank you for your reviews and guidances!

PR #14424, opened 2022-07-21
feat(client): adapt generate success message for Data Proxy

aqrln:
````

✔ Generated Prisma Client (0.0.0 | dataproxy) to ./node_modules/.prisma/client in 85ms
```
```

To use Prisma Client in edge runtimes like Cloudflare Workers or Vercel Edge Functions, import it like this:
```
import { PrismaClient } from '@prisma/client/edge'
```

You will need a Prisma Data Proxy connection string. See documentation: https://pris.ly/d/data-proxy
````

Closes https://github.com/prisma/client-planning/issues/73

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -243,6 +243,7 @@ ${chalk.dim('```')}
 ${highlightTS(`\
 import { PrismaClient } from '${importPath}/edge'`)}
 ${chalk.dim('```')}
+
 You will need a Prisma Data Proxy connection string. See documentation: ${link('https://pris.ly/d/data-proxy')}
 `
             : ''
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -238,12 +238,12 @@ ${chalk.dim('```')}${
           prismaClientJSGenerator.options?.dataProxy
             ? `
 
-To use Prisma Client in edge environments like Cloudflare Workers or Vercel Edge Functions, import it like this:
+To use Prisma Client in edge runtimes like Cloudflare Workers or Vercel Edge Functions, import it like this:
 ${chalk.dim('```')}
 ${highlightTS(`\
 import { PrismaClient } from '${importPath}/edge'`)}
 ${chalk.dim('```')}
-Data Proxy documentation: ${link('https://www.prisma.io/docs/data-platform/data-proxy')}
+You will need a Prisma Data Proxy connection string. See documentation: ${link('https://pris.ly/d/data-proxy')}
 `
             : ''
         }${breakingChangesStr}${versionsWarning}`
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -234,8 +234,21 @@ ${chalk.dim('```')}
 ${highlightTS(`\
 import { PrismaClient } from '${importPath}'
 const prisma = new PrismaClient()`)}
-${chalk.dim('```')}${breakingChangesStr}${versionsWarning}`
+${chalk.dim('```')}${
+          prismaClientJSGenerator.options?.dataProxy
+            ? `
+
+To use Prisma Client in edge environments like Cloudflare Workers or Vercel Edge Functions, import it like this:
+${chalk.dim('```')}
+${highlightTS(`\
+import { PrismaClient } from '${importPath}/edge'`)}
+${chalk.dim('```')}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/cli/src/Generate.ts [hunk 1]
```diff
@@ -234,8 +234,21 @@ ${chalk.dim('```')}
...
+${chalk.dim('```')}${
+          prismaClientJSGenerator.options?.dataProxy
+            ? `
+
+To use Prisma Client in edge environments like Cloudflare Workers or Vercel Edge Functions, import it like this:
+${chalk.dim('```')}
+${highlightTS(`\
+import { PrismaClient } from '${importPath}/edge'`)}
+${chalk.dim('```')}
+Data Proxy documentation: ${link('https://www.prisma.io/docs/data-platform/data-proxy')}
```
Jolg42, line 246:
```suggestion
Data Proxy documentation: ${link('https://pris.ly/d/data-proxy')}
```

packages/cli/src/Generate.ts
```diff
@@ -234,8 +234,21 @@ ${chalk.dim('```')}
 ${highlightTS(`\
 import { PrismaClient } from '${importPath}'
 const prisma = new PrismaClient()`)}
-${chalk.dim('```')}${breakingChangesStr}${versionsWarning}`
+${chalk.dim('```')}${
+          prismaClientJSGenerator.options?.dataProxy
+            ? `
+
+To use Prisma Client in edge environments like Cloudflare Workers or Vercel Edge Functions, import it like this:
```
Jolg42, line 241:
```suggestion
To use Prisma Client in Edge runtimes like Cloudflare Workers or Vercel Edge Functions, import it like this:
```
Maybe?

aqrln: I don't think "Edge" should be capitalized here, it's not a proper name or brand, "edge computing" is a generic term. Same as we don't capitalize, e.g., "cloud". No preference on "environments" vs "runtimes", but our docs use "runtimes" too, so I'm gonna change it for consistency as well 👍

packages/cli/src/Generate.ts
aqrln: > https://pris.ly/d/data-proxy

Oh, nice, it works now, thank you! When I tried this, it redirected me to https://www.prisma.io/docs/concepts/data-proxy instead of https://www.prisma.io/docs/data-platform/data-proxy and I got 404. I don't know if I did something wrong and tried a slightly different URL, or if it was a glitch/bug in the docs.

Jolg42 (approved):
Looks good to me!

It reminds me that one day we could do something related to https://github.com/prisma/prisma/issues/3390

packages/cli/src/Generate.ts
```diff
@@ -234,8 +234,21 @@ ${chalk.dim('```')}
...
+${chalk.dim('```')}${
+          prismaClientJSGenerator.options?.dataProxy
+            ? `
+
+To use Prisma Client in edge runtimes like Cloudflare Workers or Vercel Edge Functions, import it like this:
+${chalk.dim('```')}
+${highlightTS(`\
+import { PrismaClient } from '${importPath}/edge'`)}
+${chalk.dim('```')}
+You will need a Prisma Data Proxy connection string. See documentation: ${link('https://pris.ly/d/data-proxy')}
```
janpio, line 246: This applies to the standard import as well when generated via `--data-proxy`, right?

aqrln: I believe so, yes. Should we add an empty line before the sentence to separate it from the edge import and make it a paragraph on its own?

Jolg42: @aqrln This  `prismaClientJSGenerator.options?.dataProxy ? 'your added' text` makes it look like it will only be added if `dataProxy` is truthy. So only adding the new text if true?

aqrln: @Jolg42 yes, correct. We only print this text when generating the pair of Data Proxy clients with `prisma generate --data-proxy`, the message is unchanged for the regular client.

millsp (approved):

packages/cli/src/Generate.ts
aqrln:
```suggestion

You will need a Prisma Data Proxy connection string. See documentation: ${link('https://pris.ly/d/data-proxy')}
```

janpio:
Optimally the empty line change should also have required a test update to adapt the equivalent test that makes sure that this output indeed works as intended. We either do not have a test for `--data-proxy` or not tests for this output at all. We might want to change that.

PR #14495, opened 2022-07-26
test: Port `chaining` test to new setup

SevInf:
Rename it too `fluent-api` as well as that's we call this style
elsewhere.

Move fluent-api cases for for `findFirstOrThrow` and
`findUnqiueOrThrow` into the ported suite.

Fix prisma/client-planning#32


packages/client/tests/functional/fluent-api/prisma/_schema.ts
```diff
@@ -0,0 +1,74 @@
+import { idForProvider } from '../../_utils/idForProvider'
+import testMatrix from '../_matrix'
+
+export default testMatrix.setupSchema(({ provider }) => {
+  const id = idForProvider(provider)
+  return /* Prisma */ `
+    generator client {
```
danstarns, line 7: Wondering if this entire schema is used in the tests or if we could cut it down to only what's used?

danstarns (approved):

packages/client/tests/functional/fluent-api/prisma/_schema.ts
SevInf: I think it is, actually. Those tests are all about querying deeply nested relationships and we need all the models here.

PR #14498, opened 2022-07-26
feat(client): Allow to set iTx isolation level

SevInf:
Uses `TransactionIsolationLevel` enum from DMMF and adds an option for setting isolation level to `$transaction`. 
This enum is generated as a "strict" enum in JS code. The difference with normal enums: it will throw on attempt to access undefined value. This is needed for the following case:

Suppose user writes their code in JS without any typechecking and they write following snippet:

```ts 
await prisma.$transaction(
  async (tx) => {
    await tx.user.create({ data: { email: 'user@example.com' } })
  },
  { isolationLevel: Prisma.TransactionIsolationLevel.RepeatableRead },
)
```

If `RepeatableRead` level is not supported by connector, corresponding value won’t exist on the enum and `Prisma.TransactionIsolationLevel.RepeatableRead` will be `undefined`. Without compile-time checking, code above will then be interpreted as `isolationLevel: undefined`, which is valid and will work, but it’s not what user really wanted to achieve. And it’s quite likely that the error won’t be noticed for a long time.  We can’t really throw an error on `isolationLevel: undefined` - this will be inconsistent with the rest of our APIs, including other iTx options.

Generating `TransactionIsolationLevel` as a strict enum helps to avoid this issue and throw a runtime error, semantically matching compile-time one.

Notes on tests: due to the complexity, we don't test on the effects of setting different isolation levels. We simply check, that transaction goes through if supported isolation level is set or we get an exception if unsupported one is picked.


packages/client/package.json
```diff
@@ -140,7 +140,7 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "4.2.0-10.a7370859a29d1da1fb77313857c210926d22396a"
+    "@prisma/engines-version": "4.2.0-2.query-engine-itx-isolation-17c351246f7ac88bb1914dfadfc973e2765366c4"
```
Jolg42, line 143: Note before merging: use an engines version from main and not an integration one

garrensmith (commented):
This looks good to me. But better typescript engineers need to give this a proper review. Nice job

packages/client/src/runtime/strictEnum.ts
```diff
@@ -0,0 +1,12 @@
+export const strictEnumNames = ['TransactionIsolationLevel']
+
+export function makeStrictEnum<T extends Record<PropertyKey, string | number>>(definition: T): T {
```
millsp, line 3: Could you please add a quick description and why it is useful in some cases?

packages/client/src/__tests__/strictEnum.test.ts
```diff
@@ -0,0 +1,35 @@
+import { makeStrictEnum } from '../runtime/strictEnum'
```
millsp, line 1: What do you think about keeping the spec/unit tests near their parent file instead of a separate folder?
VSCode has a neat feature to collapse those into a single entry in the file browser panel.
```
    "explorer.fileNesting.patterns": {
        "*.js": "${capture}.js.map, ${capture}.min.js, ${capture}.d.ts",
        "*.jsx": "${capture}.js",
        "*.tsx": "${capture}.ts",
        "tsconfig.json": "tsconfig.*.json",
        "package.json": "package-lock.json, yarn.lock",
        "*.ts": "${capture}.spec.ts" // <= collapsing spec files
    },
```

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -90,31 +91,46 @@ Prisma.NullTypes = {
 // https://github.com/microsoft/TypeScript/issues/3192#issuecomment-261720275
 function makeEnum(x) { return x; }
 
-exports.Prisma.PostScalarFieldEnum = makeEnum({
+exports.Prisma.AScalarFieldEnum = makeEnum({
```
millsp, line 94: This feels hard to review and error-prone, is it because of changes of in ordering in the dmmf?

millsp: If so, I wonder if it could help to "sort" the dmmf before hand.

SevInf: That's actually exactly what engine started doing in this PR - previously, order of enums in DMMF was consistent, albeit arbitrary, now `TransactionIsolationLevel` got added at arbitrary point, which was not consistent between the runs. To fix this, engine started sorting enums by name alphabetically. So, this diff is bigger, but we should be good for the future.

packages/client/src/__tests__/strictEnum.test.ts
SevInf: Nice, will do just that.

millsp (approved):
Nice job :100:

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -33,10 +33,16 @@ function interactiveTransactionDefinition(this: PrismaClientClass) {
...
+
+  if (this.dmmf.hasEnumInNamespace('TransactionIsolationLevel', 'prisma')) {
+    txOptions.push('isolationLevel?: Prisma.TransactionIsolationLevel')
+  }
 
   return `
-  $transaction<R>(fn: (prisma: Prisma.TransactionClient) => Promise<R>, options?: ${txOptions}): Promise<R>;`
+  $transaction<R>(fn: (prisma: Prisma.TransactionClient) => Promise<R>, options?: {${txOptions.join(
+    ', ',
+  )}}): Promise<R>;`
```
danstarns, line 45:
```suggestion
const options = `{${txOptions.join(", ")}}`;
  return `
    $transaction<R>(fn: (prisma: Prisma.TransactionClient) => Promise<R>, options?: ${options}): Promise<R>;
    `;
```

What do you think about defining the options above so its easier to read?

danstarns (approved):

packages/client/package.json
danstarns: +1

PR #14501, opened 2022-07-26
fix(engines): allow engine overwrite on postinstall

millsp:
Fixes local development engines that would not get overwritten/updated if the `dist` folder already existed (eg. if you updated the `engines-version`). Issue comes from recent changes for porting `engines-wrapper` packages back to the monorepo.


danstarns (approved):

Jolg42 (approved):
Thank you for the quick fix!

PR #14510, opened 2022-07-27
fix: Memory leak in tracing test setup

SevInf:
Global context manager was set up for every test without subsequent
cleanup, which caused each test to leak around 100 MN of memory. Fix
consists of 2 parts:

1. Do tracing setup only for tracing test since this is the only test
   that needs it so far.

2. Call `context.disable()` after the test is done.


danstarns (approved):
Nice work, works for me! Not sure what issues there were with: 

```
// Tracing to happen top level because many different instances of OTEL will conflict(using global vars)
```

But its not happening anymore!

SevInf:
No that comment is right, issues do actually still happpen if we create exporter in `beforeAll` block within `setupTestSuite` callback, but moving that `beforeAll` outside of `setupTestSuite` prevents them.

PR #14512, opened 2022-07-27
fix(internals): improve error messages removing false positives and internal details

### LINKED ISSUE #14104: OpenSSL error message appearing while using query-engine has false positives
Currently, we print the error `It seems there is a problem with your OpenSSL installation!` even when the actual problem isn't related to `openssl`, but rather to executing an engine that either:
- hasn't been compiled to support `musl` (rather than `libc`), or
- hasn't been compiled to support the `aarch64` architecture on Linux (e.g., on a Docker container), or
- a combination thereof.

See [this comment](https://github.com/prisma/prisma/issues/14043#issuecomment-1172487886) by @janpio.

## Example 1 (wrong, false positive)

```
#21 2.546 Error: Get Config: Unable to establish a connection to query-engine-node-api library. It seems there is a problem with your OpenSSL installation!
#21 2.546 Details: Unable to require(`/app/node_modules/.pnpm/prisma@4.0.0/node_modules/prisma/libquery_engine-linux-arm64-openssl-1.1.x.so.node`)
#21 2.546  Error relocating /app/node_modules/.pnpm/prisma@4.0.0/node_modules/prisma/libquery_engine-linux-arm64-openssl-1.1.x.so.node: __res_init: symbol not found
```

## Example 2 (right)

```
Error: Get Config: Unable to establish a connection to query-engine-node-api library. It seems there is a problem with your OpenSSL installation!
Details: Unable to require(`/home/florian/zac-management/node_modules/.pnpm/prisma@4.0.0/node_modules/prisma/libquery_engine-debian-openssl-1.1.x.so.node`)
 libssl.so.1.1: cannot open shared object file: No such file or directory
```

## Proposed solution

- Look for the `libssl` string in [here](https://github.com/prisma/prisma/blob/8e90f8a1ca12f9b596f92bae811d9d92266aab7a/packages/internals/src/engine-commands/queryEngineCommons.ts#L82-L88), rather than looking for `openssl`
- As a fallback, if we find the `Unable to require` string, we might emit an appropriate suggestion, like `It seems that the current architecture $architecture is not supported, or that libc is missing.` or something of the sort.

### LINKED ISSUE #14588: Calling `dmmf` raises "Schema parsing - Error while interacting with query-engine-node-api library" misleading error message when there is a schema validation error.
For instance, trying to use an invalid referential action prints the following error:

```text
Get DMMF: Schema parsing - Error while interacting with query-engine-node-api library
    Error code: P1012
    error: Error validating: Invalid referential action: `SetDefault`. Allowed values: (`Cascade`, `Restrict`, `NoAction`, `SetNull`)
```

t means that calling `NodeAPIQueryEngineLibrary.dmmf(datamodel)` had thrown an error which was successfully JSON-parsed. This is coherent with what was happening before the getConfig/getDmmf refactoring (https://github.com/prisma/prisma/blob/3.11.0/packages/sdk/src/engine-commands/getDmmf.ts#L58-L64).

In practice, such an error doesn't necessarily imply a schema parsing issue, so we should be able to disambiguate actual "Schema Parsing" errors vs these "Invalid Request" errors.

See internal discussion [here](https://www.notion.so/prismaio/Phase-1-Report-on-findings-f21c7bb079c5414296286973fdcd62c2#1b1afacfff5543979fe9bed951fa5a2b).

jkomyno:
Close https://github.com/prisma/prisma/issues/14104.
Close https://github.com/prisma/prisma/issues/14588.

- Errors like `Get DMMF: Error while trying to read datamodel path` now are `Error while trying to read datamodel path`
- Errors like `Get Config: Unable to establish a connection to query-engine-node-api library.` now are `Error (query-engine-node-api library)`
- Errors like `Get DMMF: Schema parsing - Error while interacting with query-engine-node-api library` now are `Schema validation error - Error (query-engine-node-api library)`


packages/internals/src/engine-commands/queryEngineCommons.ts
```diff
@@ -122,3 +138,7 @@ export const createDebugErrorType =
   ({ type, reason, error }: { type: string; reason: string; error: Error }) => {
     debug(`error of type "${type}" in ${fnName}:\n`, { reason, error })
   }
+
+export function createSchemaValidationError(reason: string) {
+  return `${chalk.redBright.bold('Schema validation error')} - ${reason}`
+}
```
jkomyno, line 144: If we want to change the `Schema validation error - ...` text label, we just need to modify this definition

packages/internals/src/__tests__/engine-commands/queryEngineCommons.test.ts
```diff
@@ -31,7 +33,9 @@ describeIf(process.env.PRISMA_CLI_QUERY_ENGINE_TYPE !== 'binary')('loadNodeAPILi
   })
 
   it('error path, openssl', async () => {
-    const spyLoadTag = 'error-load, something something openssl installation'
+    const spyLoadTag = `Unable to require(\`/home/florian/zac-management/node_modules/.pnpm/prisma@4.0.0/node_modules/prisma/libquery_engine-debian-openssl-1.1.x.so.node\`)
```
Jolg42, line 36:
```suggestion
    const spyLoadTag = `Unable to require(\`/home/user/my-project/node_modules/.pnpm/prisma@4.0.0/node_modules/prisma/libquery_engine-debian-openssl-1.1.x.so.node\`)
```

packages/internals/src/__tests__/engine-commands/queryEngineCommons.test.ts
```diff
@@ -54,6 +58,32 @@ describeIf(process.env.PRISMA_CLI_QUERY_ENGINE_TYPE !== 'binary')('loadNodeAPILi
     }
   })
 
+  it('error path, architecture or libc', async () => {
+    const spyLoadTag = `Unable to require(\`/app/node_modules/.pnpm/prisma@4.0.0/node_modules/prisma/libquery_engine-linux-arm64-openssl-1.1.x.so.node\`)
+Error relocating /app/node_modules/.pnpm/prisma@4.0.0/node_modules/prisma/libquery_engine-linux-arm64-openssl-1.1.x.so.node: __res_init: symbol not found`
+
+    const spyLoad = jest.spyOn(loadUtils, 'load').mockImplementation((id: string) => {
+      throw new Error(spyLoadTag)
+    })
```
Jolg42, line 67: I think this will cause problems like in https://github.com/prisma/schema-team/issues/360 which means we can never upgrade the `@swc/core` dependency of the internals package...

We really need to find a solution (didn't find any) or test differently

packages/internals/src/engine-commands/queryEngineCommons.ts
```diff
@@ -80,12 +82,26 @@ export function loadNodeAPILibrary(queryEnginePath: string) {
...
               return ` It seems there is a problem with your OpenSSL installation!`
             },
           )
+
+          // handle incompatible arch or c library error
+          .when(
+            (errMessage) => errMessage.includes('Unable to require'),
+            () => {
+              const architecture = os.arch()
+              return ` It seems that the current architecture ${chalk.redBright(
```
Jolg42, line 98:
```suggestion
              return `It seems that the current architecture ${chalk.redBright(
```
nit

packages/internals/src/engine-commands/getConfig.ts
```diff
@@ -60,8 +61,6 @@ type GetConfigErrorInit = {
 
 export class GetConfigError extends Error {
   constructor(params: GetConfigErrorInit) {
-    const headline = chalk.redBright.bold('Get Config: ')
-
```
Jolg42, line 64: I don't remember that we decided to remove this part? 🤔

packages/internals/src/engine-commands/getDmmf.ts
```diff
@@ -61,8 +62,6 @@ type GetDmmfErrorInit = {
 
 export class GetDmmfError extends Error {
   constructor(params: GetDmmfErrorInit) {
-    const headline = chalk.redBright.bold('Get DMMF: ')
-
```
Jolg42, line 65: I don't remember that we decided to remove this part? 🤔

packages/internals/src/engine-commands/getConfig.ts
janpio: I kind of agree that this is not helpful for users, it is _meta_ information about what happens behind the scenes. We should probably move it somewhere else (less prominent, end of line or similar) as context for us to understand what component triggered a warning or error.

packages/internals/src/engine-commands/getDmmf.ts
janpio: https://github.com/prisma/prisma/pull/14512/files#r954200035

packages/internals/src/engine-commands/queryEngineCommons.ts
```diff
@@ -80,12 +82,26 @@ export function loadNodeAPILibrary(queryEnginePath: string) {
...
+                architecture,
```
janpio, line 99: What does this apply to?

janpio (approved):
Generally I do not like the new location of the getConfig/getDmmf information a lot, but it seems my preferred location (which details if this was thrown via query engine binary or engine at the end of the first line) is not actually available in all errors, so it would not be easy to just move it there - so let's just go ahead with this and observe the errors a bit over time and take another shot at this in the future.

PR #14516, opened 2022-07-27
chore(typo): prisma.schema -> schema.prisma

Jolg42:
I randomly searched for that brain :exploding_head: that is prisma.schema because it's wrong, as the official file extension is .prisma and therefore schema.prisma it should be :smile:

### DIFF
--- a/packages/cli/src/__tests__/commands/CLI.test.ts
+++ b/packages/cli/src/__tests__/commands/CLI.test.ts
@@ -95,7 +95,7 @@ it('introspect should include deprecation warning', async () => {
 
   await expect(result).rejects.toMatchInlineSnapshot(`
           Could not find a schema.prisma file that is required for this command.
-          You can either provide it with --schema, set it as \`prisma.schema\` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location
+          You can either provide it with --schema, set it as \`schema.prisma\` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location
         `)
   expect(ctx.mocked['console.log'].mock.calls).toHaveLength(0)
   expect(ctx.mocked['console.info'].mock.calls).toHaveLength(0)
--- a/packages/cli/src/__tests__/commands/CLI.test.ts
+++ b/packages/cli/src/__tests__/commands/CLI.test.ts
@@ -95,7 +95,7 @@ it('introspect should include deprecation warning', async () => {
 
   await expect(result).rejects.toMatchInlineSnapshot(`
           Could not find a schema.prisma file that is required for this command.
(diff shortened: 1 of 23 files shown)


packages/cli/src/__tests__/commands/CLI.test.ts
```diff
@@ -95,7 +95,7 @@ it('introspect should include deprecation warning', async () => {
 
   await expect(result).rejects.toMatchInlineSnapshot(`
           Could not find a schema.prisma file that is required for this command.
-          You can either provide it with --schema, set it as \`prisma.schema\` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location
+          You can either provide it with --schema, set it as \`schema.prisma\` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location
```
SevInf, line 98: I think in this case `prisma.schema` was correct since it's talking about `pacakge.json` key, not the file.

packages/cli/src/__tests__/commands/Doctor.test.ts
```diff
@@ -24,7 +24,7 @@ it('should fail when Prisma schema is missing', async () => {
   const result = Doctor.new().parse([])
   await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
           Could not find a schema.prisma file that is required for this command.
-          You can either provide it with --schema, set it as \`prisma.schema\` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location
+          You can either provide it with --schema, set it as \`schema.prisma\` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location
```
SevInf, line 27: And here as well, `schema.prisma` looks correct

packages/internals/src/cli/getSchema.ts
```diff
@@ -40,7 +40,7 @@ export async function getSchemaPathInternal(
     return customSchemaPath
   }
 
-  // 2. Try the package.json `prisma.schema` custom path
+  // 2. Try the package.json `schema.prisma` custom path
```
SevInf, line 43: Again, `package.json` key, not the path

SevInf (changes requested):
A good chunk of messages talks about `prisma.schema` key in `package.json`, not the file path, could you double check?

packages/cli/src/__tests__/commands/CLI.test.ts
Jolg42: Ouch, good point!

Jolg42:
@SevInf Thanks I fixed (reverted) the ones about the package.json

SevInf (approved):

PR #14522, opened 2022-07-27
test: Port fulltext search tests to new setup

SevInf:
Close prisma/client-planning#36


SevInf:
Note: I hat to change the way we generate file paths for generated client in this setup. Previously, when we had this matrix:

```ts
[
  [
     { provider: 'sqlite', foo: 'abc' },
     { provider: 'postgresql', foo: 'def' }
  ],
  [
    { bar: '123', baz: '321' },
    { bar: '456', baz: '321 }
  ]
]
```

every single parameter was included in the path name, for example `provider=sqlite, foo=abc, bar=123, baz=321`.
I think with a long enough parameter list, like in this test, we hit character limit on the path in Windows. So, to fix this, only first key-value pair from each dimension is included. Example above will now be `provider=sqlite, bar=123`. This should be unique enough so clients, generated by different tests won't clash, but at the same time, consistently short enough so we don't have to worry about number of parameters we use in the matrix.

packages/client/tests/functional/fulltext-search/tests.ts
```diff
@@ -0,0 +1,158 @@
...
+          name: {
+            search: noResultsQuery,
+          },
+        },
+      })
+
+      expect(result).toEqual([])
+    })
+
+    testIf(process.platform !== 'win32')('bad query', async () => {
```
danstarns, line 80: Do we know why this is not run on windows?

danstarns (approved):

packages/client/tests/functional/fulltext-search/tests.ts
SevInf: I think it runs on windows, but error message is very different from linux/mac and it is impossible to update snapshot without windows machine.

PR #14527, opened 2022-07-28
feat(client): Tracing Connect and Disconnect

danstarns:
1. Add's tracing for the connect and disconnect methods. 

2. Refactors the connect and disconnect methods to use async-await. 

3. Engine core now requires the `runInSpan` and `runInActive` functions and so now all the tracing functions are exported from engine-core, **I cannot export the tracing functions from internals because this introduces a circular dependency**. 

Related: https://github.com/prisma/prisma-engines/pull/3087


packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -320,63 +322,87 @@ You may have to run ${chalk.greenBright('prisma generate')} for your changes to
...
-      return this.libraryStoppingPromise
+    if (!this.libraryStarted) {
+      return
     }
+
+    const tracingConfig = getTracingConfig(this)
+
+    const stopFn = async () => {
+      // TODO - What is doing, why '5'?
+      await new Promise((r) => setTimeout(r, 5))
```
millsp, line 387: I think this was to avoid a race condition in the event loop, probably something isn't awaited properly somewhere.

millsp: Could be because of the use of `async` in `Promise`, which is an anti-pattern. So we can maybe just remove this line.

danstarns: I have removed it! 🎉

danstarns: Adding this back because it makes this test fail: https://github.com/prisma/prisma/blob/main/packages/client/src/__tests__/integration/happy/disconnect-race/test.ts

millsp (approved):

PR #14537, opened 2022-07-28
fix(client): enable debug logging in edge and fix process.env access

### LINKED ISSUE #12681: DataProxy Client logging is different from normal logging
DataProxy Client:
```
C:\Users\Jan\Documents\throwaway\3.12.0_simple>npx ts-node scripts.ts
prisma:query query {
  findManyFoo {
    id
    name
  }
}
prisma:info Schema (re)uploaded (hash: 11e52bd7ecad95f9f3809fd9d8fec1533840f048c3c245ce8b703d8ff19d2fb3)
prisma:info Calling https://aws-us-east-1.prisma-data.com/3.12.0/11e52bd7ecad95f9f3809fd9d8fec1533840f048c3c245ce8b703d8ff19d2fb3/graphql (n=0)
    id
    name
  }
}
prisma:info Calling https://aws-us-east-1.prisma-data.com/3.12.0/11e52bd7ecad95f9f3809fd9d8fec1533840f048c3c245ce8b703d8ff19d2fb3/graphql (n=0)
foo proxy [ { id: '624c5afcef6abe4179942612', name: 'foo' } ]
```

Normal `library` Client:
```
  prisma:client:libraryEngine Searching for Query Engine Library in C:\Users\Jan\Documents\throwaway\3.12.0_simple\.prisma\client +53ms
  prisma:client:libraryEngine Searching for Query Engine Library in C:\Users\Jan\Documents\throwaway\3.12.0_simple\prisma\library +0ms
  prisma:client:libraryEngine loadEngine using C:\Users\Jan\Documents\throwaway\3.12.0_simple\prisma\library\query_engine-windows.dll.node +1ms
  prisma:client Prisma Client call: +920ms
  prisma:client prisma.foo.findMany({}) +2ms
  prisma:client Generated request: +0ms
  prisma:client query {
  prisma:client   findManyFoo {
  prisma:client     id
  prisma:client     name
  prisma:client   }
  prisma:client }
  prisma:client  +1ms
  prisma:client:libraryEngine sending request, this.libraryStarted: false +871ms
  prisma:client:libraryEngine library starting +1ms
  prisma:client:libraryEngine library started +43ms
prisma:query db.Foo.aggregate([ { $project: { _id: 1, name: 1, }, }, ])
foo library [ { id: '624c5afcef6abe4179942612', name: 'foo' } ]
  prisma:client:libraryEngine library stopping +675ms
  prisma:client:libraryEngine library stopped +2ms
  prisma:client:libraryEngine hookProcess received: beforeExit +1ms
  prisma:client:libraryEngine runBeforeExit +1ms
  prisma:client:libraryEngine hookProcess received: exit +1ms
```

While it is expected that the DataProxy client can not log the query that is executed right now, there is no reason why e.g. "Prisma Client call" should not also be logged the same way it is with the normal Client.

Related: https://github.com/prisma/prisma/issues/12677

### LINKED ISSUE #13771: Prisma Edge Client + CF Worker `process not defined`
### Bug description

Getting `process not found error` when deploying the prisma Edge client to Cloudflare pages. Isn the Edge client supposed to prevent this error from happening? I am only using the client in SvelteKit Endpoints https://kit.svelte.dev/docs/routing#endpoints

NOTE: I am using dot-env package to run the local dev script in package.json as prisma cannot find my .env DATABASE_URL var without this command (Maybe this is related to the issue?)

Err Log: 

` {
        text: 'Uncaught ReferenceError: process is not defined\n' +
          '  at line 8755 in node_modules/.prisma/client/edge.js\n' +
          '  at line 16 in __require2\n' +
          '  at line 8768 in node_modules/@prisma/client/edge.js\n' +
          '  at line 16 in __require2\n' +
          '  at line 8774\n' +
          '  at line 8789\n' +
          ' [code: 10021]'
      }
`

### How to reproduce

1. Set up Prisma data-proxy in SvelteKit following this instructions https://www.prisma.io/docs/concepts/data-platform/data-proxy
2. Deploy to CF pages and see that it builds successfully, but when visiting a page it will have the process not found error

You should be using the Edge Client and --data-proxy option with npx prisma generate. And your database_url should start wit prisma://

### Expected behavior

I'm not executing code that is meant for the server on the client so I dont understand why this error is happening

### Prisma information

I have a demo of the issue here https://github.com/vidiabtc/sveltekit-test

### Environment & setup

- OS: MacOS-->
- Database: MySQL
- Node.js version: v16.7.0

### Prisma Version

```
		"prisma": "^3.15.1",
		"@prisma/client": "^3.15.1",

```

### LINKED ISSUE #14536: Edge client crashes when enabling debug logs in constructor
When enabling the debug logs in the constructor:

```ts
  __internal: {
    debug: true
  }
})
```

The edge Data Proxy client crashes with:

```
TypeError: v.stderr.write is not a function
```

aqrln:
- Forward `process.env.DEBUG` or `global.DEBUG` from the entry point of
  the edge client to the runtime bundle as `process.env.DEBUG`. This
  makes it possible to enable the debug logs during development using an
  environment variable.

- Patch the debug logger to make it work in the edge environment and use
  `console.debug` or `console.log` instead of `process.stderr.write`.

- Use uniform way of accessing the environment variables in the entry
  point both for `DATABASE_URL` and `DEBUG` and fix incorrect
  unconditional access to the `process` object which might not be
  present.

Fixes https://github.com/prisma/prisma/issues/12681
Fixes https://github.com/prisma/prisma/issues/14536
Fixes https://github.com/prisma/prisma/issues/13771

### DIFF
--- a/packages/client/src/runtime/index.ts
+++ b/packages/client/src/runtime/index.ts
@@ -18,6 +18,7 @@ export type { DecimalJsLike } from './utils/decimalJsLike'
 export { findSync } from './utils/find'
 export { NotFoundError } from './utils/rejectOnNotFound'
 export { warnEnvConflicts } from './warnEnvConflicts'
+export { Debug } from '@prisma/debug'
 export {
   Engine,
   PrismaClientInitializationError,
--- a/packages/client/src/generation/TSClient/TSClient.ts
+++ b/packages/client/src/generation/TSClient/TSClient.ts
@@ -9,6 +9,7 @@ import { DMMFHelper } from '../../runtime/dmmf'
 import type { DMMF } from '../../runtime/dmmf-types'
 import type { GetPrismaClientConfig } from '../../runtime/getPrismaClient'
 import type { InternalDatasource } from '../../runtime/utils/printDatasources'
+import { buildDebugInitialization } from '../utils/buildDebugInitialization'
 import { buildDirname } from '../utils/buildDirname'
 import { buildDMMF } from '../utils/buildDMMF'
 import { buildInjectableEdgeEnv } from '../utils/buildInjectableEdgeEnv'
@@ -126,6 +127,7 @@ ${await buildInlineSchema(dataProxy, schemaPath)}
 ${buildInlineDatasource(dataProxy, datasources)}
 ${buildInjectableEdgeEnv(edge, datasources)}
 ${buildWarnEnvConflicts(edge, runtimeDir, runtimeName)}
+${buildDebugInitialization(edge)}
 const PrismaClient = getPrismaClient(config)
 exports.PrismaClient = PrismaClient
 Object.assign(exports, Prisma)
--- a/packages/client/src/generation/TSClient/common.ts
+++ b/packages/client/src/generation/TSClient/common.ts
@@ -35,6 +35,7 @@ const {
   join,
   raw,
   Decimal,
+  Debug,
   objectEnumValues,
   makeStrictEnum
 } = require('${runtimeDir}/${runtimeName}')
--- a/packages/client/src/generation/utils/buildDebugInitialization.ts
+++ b/packages/client/src/generation/utils/buildDebugInitialization.ts
@@ -1,0 +1,26 @@
+import { getRuntimeEdgeEnvVar } from './buildInjectableEdgeEnv'
+
+/**
+ * Builds the code to initialize the `debug` package.
+ *
+ * The code running in the Edge Client entry point has access to `process.env`
+ * if it's defined, but the code in the runtime bundle doesn't. Furthermore, in
+ * some environments `DEBUG` may be defined as a global variable rather than
+ * available in `process.env`. The entry point fetches the value of `DEBUG` and
+ * passes in to the `debug` package.
+ *
+ * @param edge Whether the edge runtime is used
+ */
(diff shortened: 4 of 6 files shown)


packages/client/src/runtime/index.ts
```diff
@@ -17,6 +17,7 @@ export type { DecimalJsLike } from './utils/decimalJsLike'
 export { findSync } from './utils/find'
 export { NotFoundError } from './utils/rejectOnNotFound'
 export { warnEnvConflicts } from './warnEnvConflicts'
+export { default as debug } from '@prisma/debug'
```
millsp, line 20: I think this also works
```suggestion
export { Debug } from '@prisma/debug'
```

millsp (approved):
Nice work, thanks for fixing this!

millsp:
I think that for [this fix](https://github.com/prisma/prisma/issues/13771), there will need to be some changes here as well
https://github.com/prisma/prisma/blob/main/packages/engine-core/src/data-proxy/DataProxyEngine.ts/#L42

packages/client/src/runtime/index.ts
aqrln: Done!

PR #14541, opened 2022-07-28
chore: transpile esm in tests

millsp:
With this PR, we let `@swc/jest` always transpile everything in `node_modules` for the tests since we don't suffer from memory issues any longer, so now we can afford to enable this. I have updated two deps to esm as proof for the reviewers.

Turns out this this is just the first step in the right direction for jest to be able to run tests. While that can work in the tests, it does not mean we will produce a valid package. In fact esbuild produces invalid javascript (currently emits `require(<esm>)`). So we need to switch the codebase to ESM first.

### DIFF
--- a/helpers/jest/forceTranspile.js
+++ b/helpers/jest/forceTranspile.js
@@ -1,8 +1,0 @@
-/**
- * To force jest to transpile only certain modules #ESM.
- */
-function forceTranspile(names = ['uuid']) {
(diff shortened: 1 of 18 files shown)


kt3k:
FYI prettier team does the similar thing https://github.com/prettier/prettier/pull/12217

They convert ESM-only modules to CJS and commit those files under `/vendor/` directory

millsp:
Thanks for sharing!

Issue #13355, opened 2022-05-13
Add support for rich text output for some terminals on Windows


jkomyno:
Currently, rich text output with fancy characters is not displayed on Windows.

Here are some examples:

```ts
// from: packages/migrate/src/commands/MigrateDev.ts
console.info(`\n${process.platform === 'win32' ? '' : '🌱  '}The seed command has been executed.\n`)
```

```ts
// from: packages/cli/src/CLI.ts
process.platform === 'win32' ? '' : chalk.bold.green('◭  ')
```

which excludes rich text like emojis for all terminal users. Technically, terminals like [Windows Terminal](https://apps.microsoft.com/store/detail/9N0DX20HK701) and (probably) Git Bash support rich text output as well, so it'd make sense to support them, rather than excluding all rich text output on Windows altogether.

## Proposed solution

- figure out which major terminal apps for Windows support rich text output, and how to detect programmatically whether the `cli` is running on them
- create an internal utility to centralize the boolean check needed to evaluate whether to display rich text or not. We could call this function `canDisplayRichText()`
- apply  `canDisplayRichText()` in places where `process.platform === 'win32'` was used while printing some rich text output

jkomyno:
Hi there,
As part of our effort to clean up our GitHub repo, we’re closing this issue to help focus on features and improvements that align with our focus areas and those of our community. 

This issue was originally created for internal tracking purposes, but if you feel this is important to you, feel free to reopen the issue and get like-minded users to upvote as we will prioritize fixing issues based on community interest.

For more insight into our priorities and how we’re thinking about the future of Prisma, please see our blog post: https://www.prisma.io/blog/prisma-orm-manifesto.

Thank you for your understanding and support!

PR #14551, opened 2022-07-29
feat(migrate): add ability to test against existing `MongoDB` databases

### LINKED ISSUE #13776: Add jest snapshots for improved `db push` output in MongoDB
Reference: https://github.com/prisma/prisma/pull/13700

We want to add these tests after `@prisma/sdk` gets renamed to `@prisma/internals`.
This issue also implying creating a `setupMongo` / `tearDownMongo` helper for the snapshot tests.

jkomyno:
Closes https://github.com/prisma/prisma/issues/13776.


jkomyno:
Currently investigating test cases failing when run in sequence, but succeeding when run singularly.

packages/migrate/src/__tests__/DbPush.test.ts
```diff
@@ -250,15 +255,104 @@ describeIf(process.platform !== 'win32')('push', () => {
...
+      Use the --force-reset flag to drop the database before push like prisma db push --force-reset
+      All data will be lost.
+              
+    `)
     expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
   })
 })
+
+describeIf(process.platform !== 'win32' && !process.env.TEST_SKIP_MONGODB)('push existing-db with mongodb', () => {
```
janpio, line 271: Do we have an issue that tracks that these tests are excluded for Windows completely because of Emoji? We should find a better solution for that :D

pnpm-lock.yaml
```diff
@@ -1442,7 +1444,7 @@ packages:
     engines: {node: ^12.13.0 || ^14.15.0 || ^16.10.0 || >=17.0.0}
     dependencies:
       '@jest/types': 28.1.3
-      '@types/node': 18.6.3
+      '@types/node': 17.0.45
```
janpio, line 1447: huh?

janpio (approved):

janpio (commented):
(PR title update recommended before merge)

pnpm-lock.yaml
jkomyno: This is coming from a `git merge main`. Thanks for spotting it, I'll revert to `@types/node@18.6.3` (even though we should depend on `@types/node@14.17.0` imho, as that's the oldest Node.js version we support)

packages/migrate/src/__tests__/DbPush.test.ts
jkomyno: Here's a related issue: https://github.com/prisma/prisma/issues/13355

Jolg42:
Note: I updated TESTING.md to add the `TEST_MONGO_URI_MIGRATE_EXISTING_DB` env var

Issue #7618, opened 2021-06-11
Cannot upgrade some dependencies to their latest version because they are ESM only 


millsp:
### Bug description

Some modules cannot be upgraded because they use ESM syntax. In turn, our tests fail.

### How to reproduce

For example, upgrade https://github.com/prisma/prisma/pull/7613#event-4877764798 results in 
https://buildkite.com/prisma/prisma2-test/builds/7766#f0603d89-a746-43a6-871b-42bd5b778f13/103-1249

### Expected behavior

We should be able to run the tests since we are able to compile ES modules.

millsp:
`NODE_OPTIONS='--experimental-vm-modules' jest`
`testEnvironment: 'jest-environment-node'`

Jolg42:
Note: it's blocking us upgrading our dependencies.

matthewmueller:
_Internal Note:_ Moved from Ready to be Worked on to Unread to reflect that it's blocking the upgrade

aqrln:
Some things to consider:

* If we make the tests pure ESM, we'll need to make them work with CommonJS modules including the generated Prisma Client (`const require = createRequire(import.meta.url)` etc). Unless there are any pitfalls, it should be doable, just quite a bit of churn.

* If we want to keep the tests CommonJS, we'll need to transpile the dependencies to CommonJS modules. One possible pitfall here is that some of the updated dependencies use the `node:` prefix for built-in module names (e.g., `import fs from "node:fs"`). If those were to be transpiled to `require("node:fs")` etc, it won't work on Node.js 12, so those prefixes will need to be removed as well. It's also something that will need to be taken care of when bundling.

jkomyno:
Hi there,
As part of our effort to clean up our GitHub repo, we’re closing this issue to help focus on features and improvements that align with our focus areas and those of our community. 

This issue was originally created for internal tracking purposes, but if you feel this is important to you, feel free to reopen the issue and get like-minded users to upvote as we will prioritize fixing issues based on community interest.

For more insight into our priorities and how we’re thinking about the future of Prisma, please see our blog post: https://www.prisma.io/blog/prisma-orm-manifesto.

Thank you for your understanding and support!

PR #14589, opened 2022-08-01
refactor: remove "node" type dependency from client runtime

kt3k:
This PR removes 'node.js' type dependency from `runtime/index.d.ts`

There were 2 node.js type dependencies in client runtime.

- `sql-template-tag@4.0.0` depends on Node.js `util`, which was removed in v4.1.0
- `QueryEngineConfig["env"]` and `GetConfigOptions["env"]` use `NodeJS.ProcessEnv` type.

This PR removes the above 2 deps by upgrading `sql-template-tag` and replacing `NodeJS.ProcessEnv` with `Record<string, string | undefined>`

---
*motivations*

This is a preliminary work for Deno support. Deno doesn't have 'node' types. This makes `runtime/index.d.ts` usable as is in deno.
I believe this change is also desirable for other edge runtimes as they don't provide node.js modules by default


Jolg42:
I'll let @aqrln review, note from the past
We mentioned in https://github.com/prisma/prisma/pull/13823 that we want to `We want to upgrade directly to 5.x later.`, I'm not sure why. Maybe it's fine to do the upgrade to `4.1.0` here.

kt3k:
`sql-template-tag` dropped `object` type from `RawValue` type (see https://github.com/blakeembrey/sql-template-tag/commit/d1fb92ec5653a5680f0bf9abfbecd4518b56bef4). `RawValue` type is used as parameter of `sql` tagged template literal. That causes type errors in the usages like the below:

```ts
Prisma.sql`INSERT INTO "Entry" ("id", "binary") VALUES ('4', ${Buffer.from('hello')})`
```
We need to change this, for example, like the below to avoid type error.
```ts
Prisma.sql`INSERT INTO "Entry" ("id", "binary") VALUES ('4', ${'hello'})`
```

Is this change acceptable? (If not, I'll try `sql-template-tag@5.0.3`, which allow `unknown` type for `RawValue` type.)

kt3k:
> Prisma.sql`INSERT INTO "Entry" ("id", "binary") VALUES ('4', ${Buffer.from('hello')})`

It looks like these Buffer values actually need to be Buffer type. I'll revert the test changes and try to upgrade `sql-template-tag` to v5.0.3.

aqrln:
> I'll let @aqrln review, note from the past
We mentioned in https://github.com/prisma/prisma/pull/13823 that We want to upgrade directly to 5.x later., I'm not sure why. Maybe it's fine to do the upgrade to 4.1.0 here.

Sorry for not responding, I was off this Monday and Tuesday.

From the comments and commits it looks like @kt3k has already figured this out (thank you!), but just to confirm and provide additional context: `sql-template-tag@4.1.0` introduced an accidental breaking change that affected us, when we noticed that (which took quite a while because our renovate pipeline was stalled), the author immediately responded to us and released a fix in 5.0.3. However we couldn't upgrade to 5.0.3 because of https://github.com/prisma/prisma/issues/7618.

Here's the relevant PR: https://github.com/prisma/prisma/pull/11506

Adding it to `forceTranspile` like in https://github.com/prisma/prisma/pull/14589/commits/bcc492724c020257dd91ceb5175710034d95f859 might do the trick and make it work, if my understanding is correct (@millsp knows more about it).

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -54,6 +54,7 @@ import { getRejectOnNotFound } from './utils/rejectOnNotFound'
 import { serializeRawParameters } from './utils/serializeRawParameters'
 import { validatePrismaClientOptions } from './utils/validatePrismaClientOptions'
 
+const sqltag = sqlTemplateTag.default
```
aqrln, line 57: Maybe `import sqltag, { Sql } from 'sql-template-tag'` instead on line 28 and replace `sqlTemplateTag.Sql` with just `Sql` in the file

packages/client/src/utils/getTestClient.ts
```diff
@@ -50,7 +49,7 @@ export async function getTestClient(schemaDir?: string, printWarnings?: boolean)
   const outputDir = absSchemaDir
   const relativeEnvPaths = getEnvPaths(schemaPath, { cwd: absSchemaDir })
   const activeProvider = config.datasources[0].activeProvider
-  const options: GetPrismaClientConfig = {
+  const options = {
```
aqrln, line 52: Was the type annotation removed just because it wasn't necessary or because it was causing some problems?

kt3k: I changed the reference to runtime module from `../runtime/` (`client/src/runtime`) to `../../runtime/` (`client/runtime`) in the above lines, and `client/runtime` doesn't export `GetPrismaClientConfig`. So I removed `GetPrismaClientConfig`. (also it doesn't look absolutely necessary)

I changed the reference to runtime like this because if we import from `client/src/runtime/` it throws require error at `sql-template-tag` because of ESM issue. The built version of runtime is in CommonJS, so it can avoid that issue. I'm not 100% sure this approach is acceptable or not. What do you think?

packages/client/src/runtime/getPrismaClient.ts
kt3k: Good idea! Updated.

packages/client/src/utils/getTestClient.ts
kt3k: Ah, this change seems causing the mismatch of query engine version in CI

I now see these errors..
```
Query engine library for current platform "debian-openssl-1.1.x" could not be found.
    You incorrectly pinned it to debian-openssl-1.1.x

    This probably happens, because you built Prisma Client on a different platform.
    (Prisma Client looked in "/home/runner/work/prisma/prisma/packages/client/runtime/libquery_engine-debian-openssl-1.1.x.so.node")
```

kt3k:
@aqrln Thank you for the feedbacks!

> However we couldn't upgrade to v5 because of https://github.com/prisma/prisma/issues/7618.

This problem seems very complex to me. I've been trying a few workarounds for this, but I still see the errors..

> Adding it to forceTranspile like in https://github.com/prisma/prisma/commit/bcc492724c020257dd91ceb5175710034d95f859 might do the trick and make it work, if my understanding is correct, but I'm not sure (@millsp knows more about it).

This seems to solve the issue for the case when sql-template-tag is directly imported in jest environment.

But there are other cases where test case invokes another node.js process (e.g. `packages/client/src/__tests__/integration/happy/signals/test.ts`), and that process imports sql-template-tag, which still causes error while importing it.

To avoid that one, I did [this change](https://github.com/prisma/prisma/pull/14589/commits/92277d445893b89a6ba3f5dc77fff33f4e5db0f3), and now I see another error about version mismatch of query engines..

kt3k:
In what path does the query engine exist in CI?

The error message says it couldn't find it in the following paths, for example:
```
/home/ubuntu/actions-runner/_work/prisma/prisma/.prisma/client
/home/ubuntu/actions-runner/_work/prisma/prisma/packages/client/runtime
/home/ubuntu/actions-runner/_work/prisma/prisma/packages/client
/home/ubuntu/actions-runner/_work/prisma/prisma/packages/client/src/__tests__/integration/happy/insensitive-postgresql-feature-flag
/home/ubuntu/actions-runner/_work/prisma/prisma/packages/client/src/__tests__/integration/happy/insensitive-postgresql-feature-flag
/tmp/prisma-engines
/home/ubuntu/actions-runner/_work/prisma/prisma/packages/client/src/__tests__/integration/happy/insensitive-postgresql-feature-flag
```
But on my local machine, the query engine exists at `packages/client/libquery_engine-darwin.dylib.node`. So it feels strange that it couldn't find it from `packages/client`..

millsp:
> In what path does the query engine exist in CI?

The engine is copied in two locations, directly into `@prisma/engines` on postinstall, which has also the side-effect to put a copy into the cache (if possible in eg. `/home/.cache/prisma/...`. Then there is a third location for our new `packages/client/tests/functional` (which is an effort to port tests from `__tests__/integration`), in this case the engine is also copied to the root of `packages/client`. Which tests are failing?

kt3k:
Thanks for the explanation!

> Which tests are failing?

Some examples of failing tests are:

- `packages/client/src/__tests__/integration/happy/rejectNotFound/test.ts`
- `packages/integration-tests/src/__tests__/integration/sqlite/runtime.test.ts`
- `packages/integration-tests/src/__tests__/integration/postgresql/runtime.test.ts`
- `packages/integration-tests/src/__tests__/integration/mariadb/runtime.test.ts`

kt3k:
Having looked at details of current test failures, the number of failures are overwhelming. I think it's not a good idea to to do workaround for `sql-template-tag@5.0.3`. Especially this commit [92277d4](https://github.com/prisma/prisma/pull/14589/commits/92277d445893b89a6ba3f5dc77fff33f4e5db0f3) looks bad

I'd rather try type casting with `sql-template-tag@4.1.0`, as the problem of it is only about typing

aqrln:
> I'd rather try type casting with sql-template-tag@4.1.0, as the problem of it is only about typing

The problem is we're re-exporting the template tag from that library. We could work around it for ourselves with type casting, but our users' code would still be broken with that version.

kt3k:
@millsp @aqrln I think this PR is now ready for review. PTAL

packages/client/src/utils/getTestClient.ts
millsp: It could have been a viable option to call `generateInFolder` instead, that might have solved the issue. I feel like using the runtime as it is shipped (like you proposed) is a viable option. The tests in `__tests__` are currently being upgraded to a cleaner and more standard setup, so it is only a matter of time until we remove them. Additionally, swapping `getPrismaClient` to `generateInFolder` should be transparent.

packages/client/src/runtime/utils/sql-template-tag.ts
```diff
@@ -0,0 +1,16 @@
+// Wrapper script to work around the type bug of sql-template-tag@4.1.0
+
+import { empty, join as joinOriginal, raw, Sql, sqltag as sqltagOriginal } from 'sql-template-tag'
+
+declare function joinFunc(values: RawValue[], separator?: string): Sql
+declare function sqltagFunc(strings: ReadonlyArray<string>, ...values: RawValue[]): Sql
+
+const join = joinOriginal as typeof joinFunc
+const sqltag = sqltagOriginal as typeof sqltagFunc
```
millsp, line 9: This looks good although I prefer the alternative with `generateInFolder`, if you have time and if it actually works out that easily. This seems to be viable, but of course it adds that thin layer between us and the library, which I'd like to avoid if possible. If not, `__tests__` are being phased out, and we can add a TODO as a comment here that when we don't depend on `getTestClient` anymore, we can basically upgrade to v5 since we bundle to CJS.

packages/client/src/utils/getTestClient.ts
millsp: (see https://github.com/prisma/prisma/pull/14589#discussion_r947135925)

packages/client/src/runtime/utils/sql-template-tag.ts
kt3k: Ok, I'll experiment with `generateInFolder` for a bit.

packages/integration-tests/src/__tests__/__helpers__/integrationTest.ts
```diff
@@ -239,7 +239,13 @@ export function runtimeIntegrationTest<Client>(input: Input<Client>) {
       const { ctx, state } = await setupScenario(kind, input, scenario)
       states[scenario.name] = state
 
-      const PrismaClient = await getTestClient(ctx.fs.cwd())
+      await generateInFolder({
```
kt3k, line 242: Changed this `getTestClient` call to `generateInFolder`. This has currently 2 problems:

- The test fails with OOM error (if I reduces the number of scenarios, then they pass.)
- The assertions about Decimal don't pass because the Decimal constructors don't match

packages/client/src/runtime/utils/sql-template-tag.ts
millsp: If it is too tricky to get the integration-tests to pass, just leave it as is and ping me.

packages/integration-tests/src/__tests__/__helpers__/integrationTest.ts
millsp: I see, let's keep your original work, I hoped this would not get too tricky. Thanks for trying!

millsp: (these tests are also legacy tests that we will eventually get rid of)

kt3k: Ok, I'll move this back to `sql-template-tag@4.1.0 + type modification` version

packages/client/src/runtime/utils/sql-template-tag.ts
```diff
@@ -0,0 +1,22 @@
+// Wrapper script for modifying the type of sql-template-tag@4.1.0
+// TODO: We should remove this file when we upgrade sql-template-tag to 5.0.3 or above.
+// Note: The reason why we don't upgrade it now is that it publishes with ESM syntax.
+// We can upgrade it when we get rid of the use of `getTestClient` in testing.
+// See the below for more context:
```
millsp, line 5: Awesome, thank you.

aqrln (approved):
Nice work, thank you!

millsp:
Hey @kt3k, I have taken some time to upgrade to `sql-template-tag@5` via this [PR](https://github.com/prisma/prisma/pull/14885). I'm waiting for reviews on it, yours is also welcome :)

kt3k:
@millsp Great! Thank you!

I'll update this PR soon.

kt3k:
Updated. Now this PR only has `NodeJS.ProcessEnv | Record<string, string>` -> `Record<string, string | undefined>` update and the test of the runtime declaration file.

aqrln (approved):
LGTM!

millsp (approved):
Thank you for this PR and your efforts to get Prisma working with Deno :sauropod: :rocket:

PR #14600, opened 2022-08-02
fix(client): Exempt a bunch of properties from strict enum checks

SevInf:
Various JS constructs use special methods/properties. Strict enums
should not throw on attempts to get those properites to avoid breaking
said constructs.
See https://prisma-company.slack.com/archives/CRGDU19K6/p1659431633254319

### DIFF
--- a/packages/client/src/runtime/strictEnum.ts
+++ b/packages/client/src/runtime/strictEnum.ts
@@ -1,6 +1,17 @@
 export const strictEnumNames = ['TransactionIsolationLevel']
 
 /**
+ * List of properties that won't throw exception on access and return undefined instead
+ */
+const allowList = new Set([
+  'toJSON', // used by JSON.stringify
+  'asymmetricMatch', // used by Jest
+  Symbol.iterator, // used by various JS constructs/methods
+  Symbol.toStringTag, // Used by .toString()
+  Symbol.isConcatSpreadable, // Used by Array#concat,
+  Symbol.toPrimitive, // Used when getting converted to primitive values
+])
+/**
  * Generates more strict variant of an enum which, unlike regular enum,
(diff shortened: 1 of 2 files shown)


danstarns (approved):
LGTM, curious as to what failed/what particular issue this is fixing.

packages/client/src/runtime/strictEnum.ts
```diff
@@ -1,5 +1,16 @@
 export const strictEnumNames = ['TransactionIsolationLevel']
 
+/**
+ * List of properties that won't throw exception on access and return undefined instead
+ */
+const allowList = new Set([
+  'toJSON', // used by JSON.stringify
+  'asymmetricMatch', // used by Jest
```
danstarns, line 8: Bit odd we have to change the src code for testing.

danstarns (approved):

SevInf:
@danstarns https://github.com/prisma/ecosystem-tests/runs/7582980876?check_suite_focus=true#step:7:141 is what prompted it.
But generally, all tests added in this PR will fail without the fix.

packages/client/src/runtime/strictEnum.ts
SevInf: Jest will try to get a property every time you do `toEqual(StrictEnum)` or `.toBe(StrictEnum)`. Not only in our tests, but in users tests as well.
Normally I'd agree with you, but at the same time, i think Jest is popular enough to warrant a fix on our side in this case.

danstarns:
> @danstarns https://github.com/prisma/ecosystem-tests/runs/7582980876?check_suite_focus=true#step:7:141 is what prompted it. But generally, all tests added in this PR will fail without the fix.

Thanks! LGTM.

PR #14637, opened 2022-08-03
test(client): Lock engine directory during engine download

SevInf:
Multiple jest workers can run in parallel and when engine is not
downloaded yet, it is possible those downloads interfere with each other
(for example, one worker might delete engine file which was about to be
used by other worker). This than may cause test setup to hang and never
finish setup properly.

Fixed by locking directory for the duration of "check if engine exists,
download if not" routine which ensures that there is always only one
download running at the same time at the engine will be downloaded at
most once no matter how many workers we have.


Jolg42 (commented):
Looks good to me, and it looks like it has little impact on test time?

SevInf:
It actually looks like tests got a little bit faster, since other workers now won't start engine download. We might revisit poll timeouts for the future if locking ever causes any performance issues, library is pretty flexible in that regard.

millsp (approved):

PR #14638, opened 2022-08-03
feat(client): Tracing Fixes Integration

danstarns:
Related: https://github.com/prisma/prisma-engines/pull/3087


packages/client/tests/functional/tracing/tests.ts
```diff
@@ -20,17 +21,42 @@ type Tree = {
...
   const childrenSpans = spans.filter((span) => span.parentSpanId === tree.span.spanContext().spanId)
   if (childrenSpans.length) {
     tree.children = childrenSpans.map((span) => buildTree({ span }, spans))
   } else {
     tree.children = []
   }
 
-  return tree
+  // Remove unused keys for easier debugging
+  const simpleTree = JSON.stringify(
```
millsp, line 32: Is it still needed?

millsp (approved):

packages/client/tests/functional/tracing/tests.ts
danstarns: Its handy to keep so that when someone console logs the tree it's really small - I will keep it.

millsp:
@danstarns it seems the reason for the failure is that the links were removed during merging of engine PR?

PR #14641, opened 2022-08-03
 fix(migrate): enable --force on prisma migrate dev

### LINKED ISSUE #14610: --force or --auto-accept flag for prisma migrate dev --create-only
## Problem

My team is trying to automate migration generation for `prisma migrate dev --create-only` and our script bombs when warnings that we would normally approve locally arise. For instance:

```
⚠️  Warnings for the current datasource:

  • A unique constraint covering the columns `[name]` on the table `Customer` will be added. If there are existing duplicate values, this will fail.

✖ Are you sure you want create this migration? … 
```

We would like to be able to auto approve these creations with a flag so we don't have to simulate an interactive shell in our deployment process. We have a PR review process in place as a hedge against these issues and would like to circumvent the check at times.

## Suggested solution

`--auto-aprove` flag that responds `y` to all warnings prompts

## Alternatives

Simulating an interactive shell which would be costly for a simple fix from a development perspective

## Additional context

Thanks in advance!

Joe-Kollin:
This adds the `-f` and `--force` flags for `prisma migrate dev`

closes https://github.com/prisma/prisma/issues/14610


Jolg42 (changes requested):
Blocking this PR, so we don't merge it inadvertently.

First, @Joe-Kollin Could you provide us some feedback about what do you expect this should do?
Which problem does this solves for you? (I could not find a linked issue or a description about this here.)

Joe-Kollin:
@Jolg42 I would like to use a Github Actions workflow to generate Prisma migrations in a PR to be merged against a repo we have that centrally contains all of our migrations for dev/qa/prod. I would like there to be a `--force` option added to the `prisma migrate dev` context so that user input (y/n) against warnings will default to a yes/be unchecked. Presently, the command fails due to it being a non-interactive prompt. I opened an issue on it as well and have included a link below. 

https://github.com/prisma/prisma/issues/14610

Joe-Kollin:
@Jolg42 @seth @jasonrudolph  any word on this?

jsgv:
Is it possible to get this merged soon? This is also something I need.

justinTM:
hi @Jolg42 @jkomyno

this has sat for over a year now with no response. not only is disabling non-interactive environments a curious (i.e., frustrating) decision for a command with ostensibly no user interaction, but the lack of communication from the maintainers for over a year makes this even more of a pain. at least an explanation in the documentation would be nice.

many developers are moving to a CI/CD paradigm where requires their developers to remember their migration workflow manually for each commit is not reliable nor efficient.

thanks in advance for your reply

mattfysh:
Added a comment to https://github.com/prisma/prisma/issues/14610 with a workaround, or you could use a package patcher (e.g. `patch-package` or `pnpm patch`) to add this change to your own repo. Not sure why the radio silence from the prisma team, this either isn't a priority or they don't want to support any issues that could begin to occur if this PR were merged

SevInf:
Hey everyone.
I am sorry, but after internal discussion we decided to close this PR. While we are not strictly against this feature and would like to provide the support for your use-cases, we are not sure this is the right way to do it. At the very least, we all concluded that better safety mechanism is needed to prevent miss-use and accidental data loss. We are not sure yet what it is: maybe it's just `--accept-data-loss` flag in addition to `--force`, maybe it's something more complicated that would ensure `--force` flag will fail if ran against production databases. Let's continue the discussion in #14610.

Meanwhile, [workaround](https://github.com/prisma/prisma/issues/14610#issuecomment-1992892233) by programmatically using `Migrate` would work for CI usecases (thanks for posting it @mattfysh).

PR #14645, opened 2022-08-03
fix(engines): ensure that library engine is the correct one

millsp:
Fixing situations where we can potentially endup with using the wrong engine locally without knowing it
(Not critical for users, but that can cause severe confusion and a waste of time for us while developing).
This situation happens when you switch engine-version or have an engine already copied in your tests.


danstarns (approved):
Looks good to me, we always have version `0.0.0` in development.

packages/fetch-engine/src/download.ts
```diff
@@ -291,19 +291,8 @@ async function binaryNeedsToBeDownloaded(
     }
   }
 
-  // If there is no cache and the file doesn't exist, we for sure need to download it
-  if (!targetExists) {
```
millsp, line 295: This would cause the following issue:
- If you have an engine file already downloaded
- This check will be skipped and we continue below

packages/fetch-engine/src/download.ts
```diff
@@ -291,19 +291,8 @@ async function binaryNeedsToBeDownloaded(
...
-    debug(`file ${job.targetFilePath} does not exist and must be downloaded`)
-    return true
-  }
-
-  // 3. If same platform, always check --version
-  if (job.binaryTarget === nativePlatform && job.binaryName !== BinaryType.libqueryEngine) {
```
millsp, line 301: - Then we land here, and we mostly use `library` engine
- That also means that we skip this and continue below

packages/fetch-engine/src/download.ts
```diff
@@ -291,19 +291,8 @@ async function binaryNeedsToBeDownloaded(
...
-    const works = await checkVersionCommand(job.targetFilePath)
-    return !works
-  } // TODO: this is probably not useful anymore
-
-  return false
```
millsp, line 306: We return `false`
- If we already have a file
- If it is also a `libraryengine`

That's a problem because it means we use whatever engine is available if one exists.
Fortunately, this only affects us during local development and not actual users.

packages/fetch-engine/src/download.ts
```diff
@@ -291,19 +291,8 @@ async function binaryNeedsToBeDownloaded(
...
+  // if it's not cached, it needs to be downloaded
+  return true
```
millsp, line 295: > That's a problem because it means we use whatever engine is available if one exists.
> Fortunately, this only affects us during local development and not actual users.

So the solution is to just download if we haven't cached it. If it is cached, it's served from there.

packages/fetch-engine/src/download.ts
millsp: It causes us to have wrong engines if we switch `engine-version`. It affects the whole monorepo as well as our tests (locally only).

packages/engines/scripts/postinstall.js
```diff
@@ -1,11 +1,10 @@
-const fs = require('fs')
 const path = require('path')
 
 const postInstallScriptPath = path.join(__dirname, '..', 'dist', 'scripts', 'postinstall.js')
 
 // that's when we develop in the monorepo, `dist` does not exist yet
 // so we compile postinstall script and trigger it immediately after
-if (fs.existsSync(postInstallScriptPath) === false) {
+if (require('../package.json').version === '0.0.0') {
```
millsp, line 7: This will ensure that this script can be re-run after `engines-version` updates (they are bundled into the script).

packages/client/tests/functional/_utils/setupQueryEngine.ts
```diff
@@ -12,31 +9,29 @@ import lockfile from 'proper-lockfile'
  * normally the downloading of the required engine is done in `getGenerators`. As the test
  * clients bypass this we need to ensure the correct engine is present.
  * @param clientEngineType
- * @param platform
  */
-export async function setupQueryEngine(clientEngineType: ClientEngineType, platform: Platform) {
+export async function setupQueryEngine(clientEngineType: ClientEngineType) {
```
SevInf, line 13: If I read this correctly, after this change every test suite will always download an engine during `beforeAll`? If so, we might bring flakiness back that way: lock lasts only for duration of `setupQueryEngine`. So, previously, when engine did not exist, first test suite acquired the lock, downloaded the engine and continued. All subsequent workers will now immediately realize that there is already an engine file and skip the download. After this change, following situation is possible:
- Worker A acquires lock, downloads engine, releases the lock and proceeds with the tests
- Worker B acquires the lock, starts download, in process removes the engine file, which is still used by Worker A.

I could see a couple of solutions for this:
- keep existence check, but make it more sofisticated, i.e.  check checksum of the existing engine file or compare existing and desired engine
- move engine setup out of jest completely and do it in `helpers/functional-tests` before running any tests instead.
- keep simple existence check, but remove all downloaded engines in `bump-engines` script.

SevInf (changes requested):

.npmrc
```diff
@@ -0,0 +1 @@
+shell-emulator=true
```
Jolg42, line 1: Why do we need this now?

Is it to "support cross platform shell commands", in which case did you find it useful? I'm curious 👀

packages/client/tests/functional/_utils/setupQueryEngine.ts
millsp: Thanks @SevInf, I have applied the suggested changes to use the version as an invalidation method.

.npmrc
millsp: Yes, I found it "nice" for windows users but totally unrelated, I can remove it if needed.

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -452,16 +461,19 @@ It took ${timeInMsToDownloadAll}ms to execute download() for all binaryTargets.`
       ]
     `)
 
-    expect(await getVersion(queryEnginePath)).toMatchInlineSnapshot(
+    expect(await getVersion(queryEnginePath, BinaryType.queryEngine)).toMatchInlineSnapshot(
       `"query-engine da41d2bb3406da22087b849f0e911199ba4fbf11"`,
     )
-    expect(await getVersion(introspectionEnginePath)).toMatchInlineSnapshot(
+    expect(await getVersion(libqueryEnginePath, BinaryType.libqueryEngine)).toMatchInlineSnapshot(
```
millsp, line 467: We now test for `libqueryEngine` too.

packages/fetch-engine/src/download.ts
```diff
@@ -100,10 +100,7 @@ export async function download(options: DownloadOptions): Promise<BinaryPaths> {
   // creates a matrix of binaries x binary targets
   const binaryJobs = flatMap(Object.entries(opts.binaries), ([binaryName, targetFolder]: [string, string]) =>
     opts.binaryTargets.map((binaryTarget) => {
-      const fileName =
-        binaryName === BinaryType.libqueryEngine
-          ? getNodeAPIName(binaryTarget, 'fs')
-          : getBinaryName(binaryName, binaryTarget)
+      const fileName = getBinaryName(binaryName, binaryTarget)
```
millsp, line 103: See changes in getBinaryName.

packages/fetch-engine/src/download.ts
```diff
@@ -294,37 +291,44 @@ async function binaryNeedsToBeDownloaded(
...
-  if (job.binaryTarget === nativePlatform && job.binaryName !== BinaryType.libqueryEngine) {
-    const works = await checkVersionCommand(job.targetFilePath)
-    return !works
-  } // TODO: this is probably not useful anymore
+  if (job.binaryTarget === nativePlatform) {
+    const currentVersion = await getVersion(job.targetFilePath, job.binaryName)
 
-  return false
-}
+    if (currentVersion?.includes(version) === false) {
```
millsp, line 302: We check that the current binary/file/engine is aligned with the version we want.

packages/fetch-engine/src/__tests__/download.test.ts
millsp: Not happening as it does not work on windows and mac.

packages/client/tests/functional/_utils/setupQueryEngine.ts
```diff
@@ -1,46 +1,22 @@
...
   const engineDownloadDir = path.join(__dirname, '..', '..', '..')
 
-  // Multiple jest workers can run this function at the same time
-  // we need to ensure that no 2 downloads run in parallel. In order to achieve
-  // that, we lock engine directory and do not unlock it again until download is done
-  const releaseLock = await lockfile.lock(engineDownloadDir, {
-    lockfilePath: path.join(engineDownloadDir, 'dir.lock'),
-    retries: {
-      forever: true,
+  await download({
```
millsp, line 13: We download both at the start of the test setup in jest, so once only.

packages/fetch-engine/src/download.ts
```diff
@@ -294,37 +291,44 @@ async function binaryNeedsToBeDownloaded(
...
-  if (job.binaryTarget === nativePlatform && job.binaryName !== BinaryType.libqueryEngine) {
-    const works = await checkVersionCommand(job.targetFilePath)
-    return !works
-  } // TODO: this is probably not useful anymore
+  if (job.binaryTarget === nativePlatform) {
+    const currentVersion = await getVersion(job.targetFilePath, job.binaryName)
 
-  return false
-}
+    if (currentVersion?.includes(version) !== true) {
```
millsp, line 302: Better equivalent of the `checkVersionCommand`

millsp: Invalidates if the commit of the engines don't match.

SevInf (approved):
Thanks a lot, this bug caused a lot of confusion.

packages/engines/scripts/postinstall.js
Jolg42: @millsp  This line could fail I think, we could add some try catch here to avoid any issue on the users side?

packages/fetch-engine/src/download.ts
```diff
@@ -294,37 +291,44 @@ async function binaryNeedsToBeDownloaded(
   // If there is no cache and the file doesn't exist, we for sure need to download it
   if (!targetExists) {
     debug(`file ${job.targetFilePath} does not exist and must be downloaded`)
+
     return true
   }
 
   // 3. If same platform, always check --version
```
Jolg42, line 298:
```suggestion
  // 3. If same platform, check --version and compare to expected version
  // If version don't match, force download
```
Maybe something like that?

aqrln (approved):

packages/client/tests/functional/_utils/setupQueryEngine.ts
SevInf: Once per worker, isn't it? If so, we still need a lock so multiple workers won't start downloading at the same time

SevInf (changes requested):

packages/client/tests/functional/_utils/setupQueryEngine.ts
millsp: Are you sure it is per worker? https://jestjs.io/docs/configuration#globalsetup-string

> This option allows the use of a custom global setup module, which must export a function (it can be sync or async). The function will be triggered once before all test suites and it will receive two arguments

millsp: The setup phase of the query engine is now happening in the `_utils/globalSetup.js` file

aqrln: Hmm, this paragraph also hints at this likely being run in the main process but not in workers:

> Any global variables that are defined through globalSetup can only be read in globalTeardown. You cannot retrieve globals defined here in your test suites.

SevInf: Yea, we figured it out in private chat: `globalSetup` part is correct, but `await` for global set here in `beforeAll` is not. That await confused me, so I assumed it runs for every worker.

millsp: So yes it is triggered once, but globals cannot be exported from there as Serhii pointed out, which means `await (global as any).queryEngineSetup` is actually undefined :see_no_evil: . So I will need to make the global setup async instead.

SevInf: Looks good now, i think dependency on `proper-lockfile' and `@types/proper-lockfile` can be removed now.

packages/engines/scripts/postinstall.js
millsp: Done.

packages/client/tests/functional/_utils/setupQueryEngine.ts
millsp: Done.

SevInf (approved):

PR #14655, opened 2022-08-04
fix(client): bundle types from @prisma/debug and @opentelemetry/api

aqrln:
Fix failing ecosystem tests by bundling the types.

Ref: https://github.com/prisma/prisma/pull/14537
Ref: https://github.com/prisma/prisma/pull/14533

### DIFF
--- a/packages/client/helpers/build.ts
+++ b/packages/client/helpers/build.ts
@@ -90,6 +90,8 @@ function bundleTypeDefinitions(filename: string, outfile: string) {
         '@prisma/internals',
         '@prisma/engine-core',
         '@prisma/generator-helper',
+        '@prisma/debug',
+        '@opentelemetry/api',
       ],
       compiler: {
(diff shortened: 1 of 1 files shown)


packages/client/helpers/build.ts
```diff
@@ -90,6 +90,8 @@ function bundleTypeDefinitions(filename: string, outfile: string) {
         '@prisma/internals',
         '@prisma/engine-core',
         '@prisma/generator-helper',
+        '@prisma/debug',
+        '@opentelemetry/api',
```
aqrln, line 94: This is definitely expected for `@prisma/debug` and I'm sure adding it here it correct, but I'm not so sure about the `@opentelemetry/api`.

@millsp is this the right way to fix this?

millsp (approved):

millsp:
Bug still exists after merging because the produced types aren't bundled correctly.

PR #14671, opened 2022-08-05
fix(debug): use own typings for Debug and Debugger

aqrln:
- Reduce API surface and use better types.
- Use modules instead of namespaces. Types in the `debug` namespace
  didn't get bundled by API Extractor, which broke the runtime .d.ts
  bundle.

Ref: https://github.com/prisma/prisma/pull/14537
Ref: https://github.com/prisma/prisma/pull/14655

### DIFF
--- a/packages/debug/src/index.ts
+++ b/packages/debug/src/index.ts
@@ -1,5 +1,7 @@
 import debug from 'debug'
 
+import { Debug, Debugger } from './types'
+
 const MAX_LOGS = 100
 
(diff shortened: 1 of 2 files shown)


jkomyno (approved):

millsp (approved):

aqrln:
ecosystem-tests: https://github.com/prisma/ecosystem-tests/actions/runs/2803275047

PR #14693, opened 2022-08-08
test: Introduce memory tests

### LINKED ISSUE #14298: Test engine for memory leaks absence
We had a memory leak in napi module in the past which was quite tricky to fix and easy to accidentally bring back. To avoid repeating it, we should have a test that ensure that `QueryEngine` instances and all it's context get's correctly garbage collected.

## Approach

1. Use [napi-rs leak reproduction](https://github.com/SevInf/napi-rs-repro/blob/master/script-leaking.js) as a base
2. Replace `leakingFunc` call with `new QueryEngine()` call
3. Instead of printing memory usage to the console, store the value somewhere
4. At the end, compute standard deviation for the memory usage throughout whole run and assert that it's below the threshold. 

## Where should this test live?

Ideally - in the engines repo. However, it has to be JS test and engines have no infrastructure for running node.js in CI at the moment. So, for the beginning we could implement such test in client repo, which already has all necessary infrastructure. Trade-off this approach: we'd find out about the memory problem only after new engine version is pull requested to the client repo.

SevInf:
Separate test suite, not using jest.
Each test is launched in a child process to avoid polluting the heap
by test setup code itself as much as possible. It runs provdied scenario
for specified amounts of iterations, forcing garbage collection and measuring
heap usage in the process. After child process is finsihed, main process then
determines if there was a memory leak or not:

- First, it fits gathered data into a straight line. This step is very
  useful for filtering out the outliers/occasional spikes.
- From the fitted line, it looks at the slope and compares it with the
  threshold. If the slope is above the threshold, memory leak is
  reported.

It also plots a graph of memory usage over time to help with debugging.

TODO:

- [x] fix ci
- [x] binary engine is leaking (skipping tests for binary engine for now, will fix in separate issue)
- [x] docs

Fix #14298


packages/client/tests/memory/_utils/buildMemoryTest.ts
```diff
@@ -0,0 +1,16 @@
+import esbuild from 'esbuild'
+
+import { MemoryTestDir } from './MemoryTestDir'
+
+/**
+ * Bundles memory test file and all its dependencies with esbuild
+ * @param testDir
+ */
+export async function buildMemoryTest(testDir: MemoryTestDir): Promise<void> {
```
SevInf, line 9: We are not using `esbuild-register` (or any other on-import ways of transpiling the code) in this suite to avoid polluting the heap with artifacts of their work.

packages/client/tests/memory/_utils/generateMemoryUsageReport.ts
```diff
@@ -0,0 +1,75 @@
+import fs from 'fs/promises'
+
+import { MemoryTestDir } from './MemoryTestDir'
+
+export async function generateMemoryUsageReport(testDir: MemoryTestDir, usageOverTime: number[]) {
+  await fs.writeFile(testDir.reportPath, getHtmlTemplate(testDir, usageOverTime))
+}
+
+function getHtmlTemplate(testDir: MemoryTestDir, usageOverTime: number[]) {
```
SevInf, line 9: I tried couple of ASCII charting libraries for terminal, but the once I tried either did not work that well with this dataset (for example, not trying to fit the graph into the available space) or haven't been updated for ages, so I opted for html report instead.

.github/workflows/test.yml
```diff
@@ -744,7 +744,7 @@ jobs:
 
       - name: Test packages/client
         if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
-        run: pnpm run test:functional:code && pnpm run test-notypes --testTimeout=40000
+        run: pnpm run test:functional:code && pnpm run test:memory && pnpm run test-notypes --testTimeout=40000
```
SevInf, line 747: Should we run it on buildkite as well? What are our rules on running stuff in GH actions only vs GH actions + Buildkite?

packages/client/package.json
```diff
@@ -85,7 +86,7 @@
     "@types/jest": "28.1.6",
     "@types/js-levenshtein": "1.1.1",
     "@types/mssql": "8.0.3",
-    "@types/node": "12.20.55",
+    "@types/node": "14.17.34",
```
SevInf, line 89: I wanted to have `fs/promises` definition. What do we usually pin `@types/node` to? Lowest supported version or latest supported version?

packages/client/helpers/memory-tests.ts
```diff
@@ -0,0 +1,17 @@
+import { ClientEngineType, getClientEngineType } from '@prisma/internals'
+
+import { runAllMemoryTests } from '../tests/memory/_utils/runAllMemoryTests'
+
+if (getClientEngineType() === ClientEngineType.Binary) {
+  console.log('Memory tests are skipped for binary engine at the moment')
```
SevInf, line 6: Binary engine leaks, so I skipped it for now. I think it will make sense to fix it in a separate issue, but it's up for discussion.

packages/client/tests/memory/_utils/runMemoryTest.ts
```diff
@@ -0,0 +1,135 @@
...
+// import { byline } from '@prisma/internals'
+import { linearRegression } from 'simple-statistics'
+
+import { buildMemoryTest } from './buildMemoryTest'
+import { dropMemoryTestDatabase, setupMemoryTestDatabase } from './database'
+import { generateMemoryTestClient } from './generateMemoryTestClient'
+import { generateMemoryUsageReport } from './generateMemoryUsageReport'
+import { MemoryTestDir } from './MemoryTestDir'
+
+const GROWTH_RATE_THRESHOLD_IN_BYTES = 200
```
danstarns, line 14: What made you pick 200?

danstarns:
Is there a way we could fold this work directly into the functional test suite so that we have memory tests for all our existing tests? It looks to me that we need to create a whole new test to test for memory leeks.

packages/client/package.json
jkomyno: Lowest supported version, so `^14.17.0` should be fine

SevInf:
>Is there a way we could fold this work directly into the functional test suite so that we have memory tests for all our existing tests? It looks to me that we need to create a whole new test to test for memory leeks.

I don't think there is a way to do it without introducing a ton of overhead and skewing all measurements. Memory test workers strive to be as minimal as possible and hardly allocate any memory on top of what client does. If we'd run them with jest, then transpilation, sandboxing and  a lot of jest internal stuff will happen during the test, which makes memory measurements much harder.
Also, to accurately measure memory usage over time we'd need to run the same code in a loop repeatedly. I don't think we want this for every one of functional tests. And I don't think we'd need a whole lot of memory tests to be honest. Just the basic test from here will probably be enough to catch most of the leaks we've encountered so far.

packages/client/tests/memory/_utils/runMemoryTest.ts
SevInf: Just a bunch of runs with leaking vs non-leaking client. Although, I think, I've picked up the threshold before I've added v8 flags for disabling caches and optimizations, now memory consumption is much stable from iteration to iteration and this threshold can be lowered.

packages/client/tests/memory/_utils/generateMemoryUsageReport.ts
```diff
@@ -0,0 +1,75 @@
...
+  return `
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
millsp: Could be cool to have it here https://prisma.github.io/prisma/dev/bench/

packages/client/tests/memory/_utils/runMemoryTest.ts
SevInf: Lowered it to 50, since after latest changes growth rate stays close to 0.

.github/workflows/test.yml
```diff
@@ -167,7 +167,7 @@ jobs:
           SKIP_GIT: true
           GITHUB_CONTEXT: ${{ toJson(github) }}
 
-      - run: pnpm run test-notypes && pnpm run test:functional
+      - run: pnpm run test-notypes && pnpm run test:functional && pnpm run test:memory
```
millsp, line 170: If we plan on adding many tests, would it make sense to extract it into its own job?

millsp: Related: It would be nice if we could plug in existing tests automatically somehow, but that might require much more work (eg. injecting `describe`, `test`, etc with esbuild to get a jest-like "runnable" suite).

packages/client/tests/memory/query/test.ts
```diff
@@ -0,0 +1,11 @@
+import { createMemoryTest } from '../_utils/createMemoryTest'
+
+//@ts-ignore
+type PrismaModule = typeof import('./.generated/node_modules/@prisma/client')
+
+void createMemoryTest<PrismaModule>(async ({ PrismaClient }) => {
+  const client = new PrismaClient()
+  await client.$connect()
+  await client.user.findMany({})
```
millsp, line 9: Do you have an idea in mind of all the possible things we should be testing? If we don't have time to write all these, we could open tasks in order to achieve it over time.

.github/workflows/test.yml
millsp: The historical rule has been that everything that runs on GH also runs on BK. So I'd say yes by default, and since BK is our publish pipeline it also runs the tests before the release. Probably not ideal to maintain both? Let's discuss that some day.

.github/workflows/test.yml
SevInf: It might be, they are not the fastest, since they are running with optimizing compiler disabled.

packages/client/tests/memory/query/test.ts
SevInf: Generally, I think it might make sense to test the scenarios where we now client is doing something really differently from existing cases. For example, I don't think it makes sense to test different queries, as they more or less boil down to the same request to the engine. These 3 are first scenarios that came to mind, not sure what we can add right now, but if anybody has any ideas I'd be happy to implement them now — writing these types of tests does not take that much time.

packages/client/package.json
SevInf: AFAIK, patch version in `@types/node` does not correspond to node.js version (node 14.17.34 does not even exist) and denotes type-specific fixes, so I guess it would be ok to leave `14.17.34`

packages/client/tests/memory/_utils/generateMemoryUsageReport.ts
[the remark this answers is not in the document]
SevInf: Done, see bot comments below

SevInf:
@danstarns @millsp I think this is ready for re-review. Difference from the last time we spoke:

1. Tests now have `prepare` and `clenup` functions, which can be used for executing a code before and after test start looping. This allows us to avoid the need for `$connect` and `$disconnect` in every iteration for `query` and `transactions` test, which helps to isolate leaks happening within those calls.
2. We now create one report per suite (previous version had one report per test) and publish it to github pages. You can see the results in bot comments above.

.github/workflows/test.yml
```diff
@@ -196,6 +196,72 @@ jobs:
...
+
+    needs: detect_jobs_to_run
+    if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        queryEngine: ['library'] # TODO: binary engine is leaking at the moment
+        os: [buildjet-4vcpu-ubuntu-2004]
+        node: [16, 18]
```
Jolg42, line 214:
```suggestion
        node: [14, 16, 18]
```

millsp (changes requested):
Really nice work. I'm excited to have this, and it looks like this will be the base for our performance testing if we can start utilizing metrics, tracing, and other measurements (cpu load, etc...). Thank you! I am fundamentally approving this PR and your approach, but requesting some changes just as a reminder from the discussions:
* See if we can drop the threshold by a few megabytes
* Add a test with a huge schema (or exhaustive  schema? :trollface:)
* Use github action for bot comments instead of script (if suitable)
* Display engine type in the comment (for later when we have binary engine)

janpio:
We use a GH Action here to comment things onto PRs: https://github.com/prisma/docs/blob/main/.github/workflows/list-changed-pages.yml

SevInf:
Few notes on comments:
- I switched to github action instead of script
- We will have one comment per failing job. So, 3 comments in worst case scenario for now.  Without overly complicating the pipeline, the choice was between one comment per failing job with auto-update after each run vs one comment per run always, i've picked less noisy option in my opinion.
- Comments will be posted only if memory test failed during pull request. If the same job later succeeds, the comment will be updated to reflect this, however, if no memory tests failed at all we will not be posting "Success" comments.

SevInf:
@millsp @Jolg42 
I believe this could use another look.
- Comments are now posted by GH action, see notes about that above
- Failure threshold is now 10 bytes/iteration
- Added a test with a larger `openfootball` schema.

TESTING.md
```diff
@@ -318,6 +319,56 @@ If JS expression, following `@ts-test-if:` evaluates to truthy value, `@ts-expec
 inserted in its place during the generation. All parameters from the test matrix can be used within that
 expression.
 
+## Memory tests
+
+This suite tests client for memory leaks. It works by repeatedly running test code in a loop
+and monitoring V8 heap usage after every iteration. If it then detects that memory usage growth
+with a rat, higher then threshold it will report memory leak and file the test.
```
Jolg42, line 326:
```suggestion
with a rate, higher then threshold it will report memory leak and file the test.
```
Nit

Jolg42: What does this means by the way `and file the test`

SevInf: It means I can't type, it should've been `fail` :D

TESTING.md
```diff
@@ -318,6 +319,56 @@ If JS expression, following `@ts-test-if:` evaluates to truthy value, `@ts-expec
 inserted in its place during the generation. All parameters from the test matrix can be used within that
 expression.
 
+## Memory tests
+
+This suite tests client for memory leaks. It works by repeatedly running test code in a loop
+and monitoring V8 heap usage after every iteration. If it then detects that memory usage growth
+with a rate, higher then threshold it will report a memory leak and file the test.
```
Jolg42, line 326: `and file the test.` what does this means?

Jolg42: I saw your comment now
>It means I can't type, it should've been fail :D

Jolg42:
```suggestion
with a rate, higher then threshold it will report a memory leak and fail the test.
```

packages/client/tests/memory/_utils/runMemoryTest.ts
```diff
@@ -0,0 +1,138 @@
+import chalk from 'chalk'
+import childProcess from 'child_process'
+import { once } from 'events'
+import fs from 'fs/promises'
+// import { byline } from '@prisma/internals'
```
Jolg42, line 5:
```suggestion
```

packages/internals/src/index.ts
```diff
@@ -77,6 +77,7 @@ export { createSpinner } from './utils/spinner'
 export type { Position } from './utils/trimBlocksFromSchema'
 export { trimBlocksFromSchema, trimNewLine } from './utils/trimBlocksFromSchema'
 export { tryLoadEnvs } from './utils/tryLoadEnvs'
+export { default as byline } from './utils/byline'
```
Jolg42, line 80: This looks unrelated and not needed?
```suggestion
```

Jolg42 (approved):
Nice one! This definitely increases the confidence about not having memory leaks, well done 🙌🏼

TESTING.md
```diff
@@ -318,6 +319,56 @@ If JS expression, following `@ts-test-if:` evaluates to truthy value, `@ts-expec
 inserted in its place during the generation. All parameters from the test matrix can be used within that
 expression.
 
+## Memory tests
+
+This suite tests client for memory leaks. It works by repeatedly running test code in a loop
+and monitoring V8 heap usage after every iteration. If it then detects that memory usage growth
+with a rate, higher then threshold it will report a memory leak and fail the test.
```
millsp, line 326:
```suggestion
and monitoring V8 heap usage after every iteration. If it detects that memory usage grows
with a rate higher than the threshold, it will report a memory leak and fail the test.
```

millsp (approved):

PR #14701, opened 2022-08-08
fix(CLI): update pkg annotations

Jolg42:
It fixes the errors with pkg v5, see ecosystem-test PR 
https://github.com/prisma/ecosystem-tests/pull/2899/


aqrln (approved):
It would be nice if we could generate this automatically from the list of supported platforms in the future.

PR #14712, opened 2022-08-09
fix(client): Ensure `getDMMF` gets called only once

### LINKED ISSUE #14695: Schema size affects runtime speed
### Bug description

After migration from 3.15.2 to 4.0.0, my production server has started running out of memory during syncs between 2 databases which occur every 45 seconds. The provided client code is a simplified version of the production code.

Test code is provided below. It builds 500 upsert queries and executes them via `$transaction`

### Prisma 4.1.1

With 201 tables in schema:
```
Runtime: 48.124s
Runtime: 1.610s
Runtime: 1.578s
```

With 1 table in schema:
```
Runtime: 2.181s
Runtime: 1.509s
Runtime: 1.539s
```

### Prisma 3.15.2

With 201 tables in schema:
```
Runtime: 1.691s
Runtime: 1.508s
Runtime: 1.482s
```

With 1 table in schema:
```
Runtime: 1.776s
Runtime: 1.599s
Runtime: 1.479s
```

By the way, this also happens with prisma.table.create. I have not tried other cases.

### How to reproduce

1. Use the provided client code with provided prisma schema. 
2. Try with prisma 4.1.1. 
3. Check runtimes and RAM/CPU usages with 1 table in schema and with 200 tables which are commented out. 

There is a difference in runtime with schema size.

### Expected behavior

1. The first run shouldn't be so slow and use that much RAM. It wasn't this slow in 3.15.2.
2. Number of tables inside schema shouldn't affect the query speed. 

### Prisma information

```ts
import { v4 } from 'uuid'

export async function go() {
    const queries = []

    for (let i = 0; i < 500; i++) {
        const element = prisma.client.upsert({
            where: {
                id: i,
            },
            create: {
                username: v4(),
            },
            update: {
                username: v4(),
            },
        })
        queries.push(element)
    }

    console.time('Runtime')
    await prisma.$transaction(queries)
    console.timeEnd('Runtime')
}

async function main() {
    await go()
    await go()
    await go()
}

main()
```

Find the prisma schema at [pastebin](https://pastebin.com/raw/1wXsJGJH)

### Environment & setup

- OS: Mac OS, Amazon Linux AMI
- Database: PostgreSQL
- Node.js version: v16.15.0

### Prisma Version

```
prisma                  : 4.1.1
@prisma/client          : 4.1.1
Query Engine (Node-API) : libquery-engine 8d8414deb360336e4698a65aa45a1fbaf1ce13d8 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 8d8414deb360336e4698a65aa45a1fbaf1ce13d8 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core 8d8414deb360336e4698a65aa45a1fbaf1ce13d8 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt 8d8414deb360336e4698a65aa45a1fbaf1ce13d8 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : 8d8414deb360336e4698a65aa45a1fbaf1ce13d8
Studio                  : 0.469.0
```

SevInf:
Problem: when we have a freshly instantiated client (haven't done
any queries yet) and we execute multiple queries at the same event loop
tick, each one of them will start `getDMMF` request, which could be
quite slow.

Fixed by caching the results promise first time `getDMMF` is called and
returning the same promise for subsequent calls.

Fix #14695


packages/internals/src/utils/callOnce.ts
```diff
@@ -0,0 +1,32 @@
...
+          }),
+        }
+        return cachedValue.value
+      case 'in-progress':
+        return cachedValue.value
+      case 'complete':
+        return Promise.resolve(cachedValue.value)
+    }
+  }
+}
```
millsp, line 32: Suggesting a smaller version that seems equivalent in terms of performance results. 
```suggestion
export function callOnce<P extends any[], R>(fn: (...args: P) => R) {
  let result: { value: R } | undefined

  return (...args: P) => {
    return result?.value ?? (result = { value: fn(...args) }).value
  }
}
```

millsp:
I read this comment
https://github.com/prisma/prisma/issues/14695#issuecomment-1209226061

```
Runtime: 1.863s
Runtime: 1.756s
Runtime: 1.756s
```

However, I get
```
Runtime: 4.381s
Runtime: 320.656ms
Runtime: 261.909ms
```

(With the user repro and full schema)

SevInf:
Sorry, forgot to push. Can you check now? The difference is that `DMMFHelper` creation is also moved into `callOnce`

millsp:
Awesome, thanks for the quick fix!

millsp (approved):
I'll leave it up to you if you think that the shorter callOnce is suitable or not. Great work. Thanks!

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1215,13 +1214,14 @@ new PrismaClient({
       })
     }
 
-    private async _getDmmf(params: Pick<InternalRequestParams, 'clientMethod' | 'callsite'>) {
+    private _getDmmf = callOnce(async (params: Pick<InternalRequestParams, 'clientMethod' | 'callsite'>) => {
       try {
-        return await this._engine.getDmmf()
+        const dmmf = await this._engine.getDmmf()
+        return new DMMFHelper(getPrismaClientDMMF(dmmf))
```
danstarns, line 1220:
```suggestion
        if(this._dmmf) { return this_dmmf }
        const dmmfString = await this._engine.getDmmf();
        this._dmmf = new DMMFHelper(getPrismaClientDMMF(dmmfString));
        return this_dmmf
```

Would this be simpler, just set and cache a property on the class?

SevInf: Code calling code already doing this. The problem is, there is nothing cached while the promise is in flight, so if 500 queries happen in the same tick like in original issue `getDMMF` will be called 500 time

danstarns: I see, I do believe in the codebase we already have a solution for this and that's returning the cached promise see: 

https://github.com/prisma/prisma/blob/main/packages/engine-core/src/library/LibraryEngine.ts#L321

But I do like your helper so maybe we could utilize it in another PR there too.

danstarns (approved):

packages/client/src/runtime/getPrismaClient.ts
SevInf: Yea, I went with helper exactly because I saw we are doing it in several places already and we might want to pull common code out and reuse it.

packages/internals/src/utils/callOnce.ts
SevInf: Thanks for suggestion, I went with slightly different variant based on it

packages/internals/src/utils/callOnce.ts
```diff
@@ -0,0 +1,15 @@
...
+ * that `fn` will be called only once:
+ *
+ * - If the first call is not finished yet, returns the promise to the same result
+ * - If the first call is finished, returns the result of this call
+ * @param fn
+ * @returns
+ */
+export function callOnce<Args extends unknown[], R>(fn: AsyncFn<Args, R>): AsyncFn<Args, R> {
+  let result: Promise<R> | undefined
+  return (...args) => (result ??= fn(...args))
```
millsp, line 14: Oh wow this is so cool. I keep on forgetting this!

aqrln (approved):

PR #14735, opened 2022-08-10
fix(publish): Correctly get latest published version when there is only release

SevInf:
`npm view version` not always returns the array, it might return a
single string if there was only one version published


Jolg42 (approved):
Thanks for the fix! (we checked and this errors on npm v6 in Buildkite and npm v8 has the same behavior locally)

SevInf:
Looks like "Detect jobs to run" script is stuck again. Could somebody with admin rights merge this PR?

janpio:
We should fix that script...

PR #14738, opened 2022-08-10
feat: add request ids to all messages for DataProxyAPIErrors

mavilein:
Inside the Scale Team we are currently working on adding request IDs to all responses that we return to the Prisma Client. We would like to surface those request IDs in all error messages related to the Data Proxy. This will allow PDP customers to reach out to us with this ID in case of errors so that we can check our logs for this given request.
The behavior only kicks in if the request id header is returned by the data plane so it is essentially feature flagged.

### DIFF
--- a/packages/engine-core/src/data-proxy/errors/DataProxyAPIError.ts
+++ b/packages/engine-core/src/data-proxy/errors/DataProxyAPIError.ts
@@ -15,11 +15,10 @@ export abstract class DataProxyAPIError extends DataProxyError {
     this.response = info.response
 
     // add request id to response message if it is present in the response header
-    const requestId = this.response.headers?.get('PDP-Request-Id');
-    if(requestId){
-      const messageSuffix = `The request id is: ${requestId}`;
-      this.message = this.message + ' ' + messageSuffix;
+    const requestId = this.response.headers?.['PDP-Request-Id']
+    if (requestId) {
+      const messageSuffix = `(The request id was: ${requestId})`
+      this.message = this.message + ' ' + messageSuffix
     }
-    
   }
 }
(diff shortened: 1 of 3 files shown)


packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -8,8 +8,10 @@ import { getJSRuntimeName } from './getJSRuntimeName'
 
 // our implementation handles less
 export type RequestOptions = O.Patch<{ headers?: { [k: string]: string }; body?: string }, RequestInit>
+
+type HeadersType = NodeJS.Dict<string | string[]>
```
millsp, line 12:
```suggestion
type Headers = Record<string | string[] | undefined>
```

packages/engine-core/src/data-proxy/errors/DataProxyAPIError.ts
```diff
@@ -13,5 +13,12 @@ export abstract class DataProxyAPIError extends DataProxyError {
     super(message, info)
 
     this.response = info.response
+
+    // add request id to response message if it is present in the response header
+    const requestId = this.response.headers?.['Prisma-Request-Id']
+    if (requestId) {
+      const messageSuffix = `(The request id was: ${requestId})`
```
millsp, line 20: Would it be better to have a property with the requestId, or do we want it to be always logged?

mavilein: We want it always to be logged so that users can easily provide those to support. But we can add this property additionally if you like to?

millsp: That makes sense. Thanks, I think we're good then.

millsp (approved):

PR #14764, opened 2022-08-11
docs: update npm README

nikolasburk:
The current [npm README](https://www.npmjs.com/package/prisma) for the `prisma` package uses the README inside the [`cli`](https://github.com/prisma/prisma/tree/main/packages/cli) package.

Since npm is a place where people might also discover and learn about Prisma initially, I think we can improve the page a little bit and gear it more towards newcomers rather than "describing the Prisma CLI" which the page currently does. My suggestion is to make the npm README


Jolg42 (approved):
Looks good, thanks @nikolasburk!

Issue #8413, opened 2021-07-23
Ability to filter count in "Count Relation Feature"


sandbox-apps:
## Problem
Currently count relation feature just accepts `true` or `false` values and count all the related records.
Sometimes we just need to count some of the records depends on the query.

For example

```ts
this.prisma.groupExamAttempt.findMany({
  where: {
    groupId: param.groupId,
    examId: param.examId,
  },
  include: {
    _count: {
      select: {
        answers: true,
      },
    },
  },
});
```

This will just passively count all the records there.

## Suggested solution
To filter maybe we can add a where clause

```ts
this.prisma.groupExamAttempt.findMany({
  where: {
    groupId: param.groupId,
    examId: param.examId,
  },
  include: {
    _count: {
      select: {
        answers: {
          where: {
            isCorrect: true,
          },
        },
      },
    },
  },
});
```

Maybe if applicable, It can also be based on the deeper relationship filter. (IF ONLY THIS IS DOABLE)

sandbox-apps:
Another use case:
Instead of
```ts
return this.prisma.group.create({
      data: {
        name: dto.name,
        shortname: dto.shortname,
        description: dto.description,
      },
      include: {
        _count: {
          select: {
            groupContents: true,
            groupMembers: true,
          },
        },
      },
    });
```

We can do

```ts
return this.prisma.group.create({
      data: {
        name: dto.name,
        shortname: dto.shortname,
        description: dto.description,
      },
      include: {
        _count: {
          select: {
            groupContents: {
              where: {
                createdAt: "within 28 days here...",
              },
            },
            groupMembers: {
              where: {
                createdAt: "within 28 days here...",
              },
            },
          },
        },
      },
    });
```

So we can have a flexible counts

sandbox-apps:
Another example would be: 
```ts
return this.prisma.product.findMany({
  data: {
    // ...
  },
  include: {
    _count: {
      as: "stats",
      select: {
        ratings: [
          {
            as: "fiveStars",
            where: { rateValue: 5 },
          },
          {
            as: "fourStars",
            where: { rateValue: 4 },
          },
          {
            as: "threeStars",
            where: { rateValue: 3 },
          },
          {
            as: "twoStars",
            where: { rateValue: 2 },
          },
          {
            as: "oneStars",
            where: { rateValue: 1 },
          },
        ],
      },
    },
  },
});
```

The json output would be
```ts
{
  "stats": {
[...]

PR #14810, opened 2022-08-15
feat(client): integrate filter by count feature

SevInf:
Ref #8413

### DIFF
--- a/packages/client/src/generation/TSClient/Args.ts
+++ b/packages/client/src/generation/TSClient/Args.ts
@@ -142,6 +142,7 @@ export class MinimalArgsType implements Generatable {
     protected readonly args: DMMF.SchemaArg[],
     protected readonly type: DMMF.OutputType,
     protected readonly action?: DMMF.ModelAction,
+    protected readonly generatedTypeName = getModelArgName(type.name, action),
   ) {}
   public toTS(): string {
     const { action, args } = this
@@ -151,13 +152,11 @@ export class MinimalArgsType implements Generatable {
       arg.comment = getArgFieldJSDoc(this.type, action, arg)
     }
 
-    const typeName = getModelArgName(name, action)
-
     return `
 /**
  * ${name} ${action ? action : 'without action'}
  */
-export type ${typeName} = {
+export type ${this.generatedTypeName} = {
 ${indent(
   args
     .map((arg) => {
--- a/packages/client/src/generation/TSClient/Count.ts
+++ b/packages/client/src/generation/TSClient/Count.ts
@@ -3,8 +3,8 @@ import indent from 'indent-string'
 
 import type { DMMFHelper } from '../../runtime/dmmf'
 import { DMMF } from '../../runtime/dmmf-types'
-import { getFieldArgName, getSelectName } from '../utils'
-import { ArgsType } from './Args'
+import { capitalize, getFieldArgName, getSelectName } from '../utils'
+import { ArgsType, MinimalArgsType } from './Args'
 import { TAB_SIZE } from './constants'
 import type { Generatable } from './Generatable'
 import { TS } from './Generatable'
@@ -22,12 +22,19 @@ export class Count implements Generatable {
 
     argsTypes.push(new ArgsType([], this.type))
 
+    for (const field of this.type.fields) {
+      if (field.args.length > 0) {
+        argsTypes.push(
+          new MinimalArgsType(field.args, this.type, undefined, getCountArgsType(this.type.name, field.name)),
+        )
+      }
+    }
+
     return argsTypes
   }
   public toTS(): string {
     const { type } = this
     const { name } = type
-
     const outputType = new OutputType(this.dmmf, this.type)
 
     return `
@@ -40,9 +47,21 @@ ${outputType.toTS()}
 export type ${getSelectName(name)} = {
 ${indent(
   type.fields
-    .map(
-      (f) => `${f.name}?: boolean` + (f.outputType.location === 'outputObjectTypes' ? ` | ${getFieldArgName(f)}` : ''),
-    )
+    .map((field) => {
+      const types = ['boolean']
+      if (field.outputType.location === 'outputObjectTypes') {
+        types.push(getFieldArgName(field))
+      }
+
+      // TODO: what should happen if both args and output types are present?
+      // Right new, they both will be part of the union, but is it correct?
+
+      if (field.args.length > 0) {
+        types.push(getCountArgsType(name, field.name))
+      }
+
+      return `${field.name}?: ${types.join(' | ')}`
+    })
     .join('\n'),
   TAB_SIZE,
 )}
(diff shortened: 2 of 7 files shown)


packages/client/src/generation/TSClient/Args.ts
```diff
@@ -144,10 +144,10 @@ export type ${modelArgName} = ${baseTypeName}
 
 export class MinimalArgsType implements Generatable {
   constructor(
-    protected readonly args: DMMF.SchemaArg[],
-    protected readonly type: DMMF.OutputType,
-    protected readonly action?: DMMF.ModelAction,
-    protected readonly collector?: ExportCollector,
+    private readonly args: DMMF.SchemaArg[],
```
SevInf, line 147: See no reasons for them to be `protected`, nobody extends this class.

packages/client/src/generation/TSClient/Args.ts
```diff
@@ -144,10 +144,10 @@ export type ${modelArgName} = ${baseTypeName}
 
 export class MinimalArgsType implements Generatable {
   constructor(
-    protected readonly args: DMMF.SchemaArg[],
-    protected readonly type: DMMF.OutputType,
-    protected readonly action?: DMMF.ModelAction,
-    protected readonly collector?: ExportCollector,
```
SevInf, line 150: `ExportsCollector` is not used.

packages/client/src/generation/TSClient/Count.ts
```diff
@@ -42,9 +49,21 @@ ${outputType.toTS()}
...
-    .map(
-      (f) => `${f.name}?: boolean` + (f.outputType.location === 'outputObjectTypes' ? ` | ${getFieldArgName(f)}` : ''),
-    )
+    .map((field) => {
+      const types = ['boolean']
+      if (field.outputType.location === 'outputObjectTypes') {
+        types.push(getFieldArgName(field))
+      }
+
+      // TODO: what should happen if both args and output types are present?
```
SevInf, line 58: I don't think that happens now, but it could happen in the future: DMMF definition allows it.

packages/client/src/runtime/query.ts
```diff
@@ -833,16 +833,7 @@ export function selectionToFields(
...
-      typeof value !== 'boolean' &&
-      field.outputType.location === 'scalar' &&
-      field.name !== 'executeRaw' &&
-      field.name !== 'queryRaw' &&
-      field.name !== 'runCommandRaw' &&
-      outputType.name !== 'Query' &&
-      !name.startsWith('aggregate') &&
-      field.name !== 'count' // TODO: Find a cleaner solution
-    ) {
+    if (field.outputType.location === 'scalar' && field.args.length === 0 && typeof value !== 'boolean') {
```
SevInf, line 836: I believe `args.length === 0` covers all removed cases, please, correct me if  I am wrong.

packages/engines/package.json
```diff
@@ -7,7 +7,7 @@
   "author": "Tim Suchanek <suchanek@prisma.io>",
   "devDependencies": {
     "@prisma/debug": "workspace:*",
-    "@prisma/engines-version": "4.3.0-13.6263eb57ecfe61a0fa05d847f7a34f3eba9fba5e",
+    "@prisma/engines-version": "4.3.0-14.feat-count-filter-2272ab800a0e3b9d4fb0c09374f0ec9fa5b7b56b",
```
SevInf, line 10: Note: update to upstream version before merging

packages/client/src/generation/TSClient/Args.ts
danstarns: Is it used elsewhere? We might be able to remove it completely. I did a search and it seems we append symbols to a collector in a few places `.addSymbol()` but we never call the `getSymbols` method.

danstarns (approved):
LGTM - some comments but nothing blocking.

packages/client/tests/functional/filter-count-relations/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const { id: authorId } = await prisma.user.create({
+      data: {
+        email,
+        balance: 70,
+        groups: { connect: { id: groupId } },
+      },
+    })
+
+    await prisma.post.create({ data: { published: true, upvotes: 10, authorId } })
```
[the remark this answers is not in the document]
SevInf: sqlite :(

packages/client/tests/functional/filter-count-relations/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+    await prisma.post.create({ data: { published: true, upvotes: 150, authorId } })
+    await prisma.post.create({ data: { published: false, upvotes: 120, authorId } })
+    await prisma.post.create({ data: { published: true, upvotes: 15, authorId } })
+  })
+
+  test('without condition', async () => {
+    const user = await prisma.user.findFirst({ where: { email }, select: { _count: { select: { posts: true } } } })
```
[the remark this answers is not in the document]
SevInf: `prettier` undoes every formatting attempt. I could wrap this block in `// @prettier-ignore` comments and do formatting myself, but not quite sure we should start doing it in general.

packages/client/tests/functional/filter-count-relations/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+        users: {
+          select: {
+            _count: { select: { posts: { where: { published: true } } } },
+          },
+        },
+      },
+    })
+
+    expect(group?.users).toContainEqual({ _count: { posts: 3 } })
+    expect(group?.users).toContainEqual({ _count: { posts: 0 } })
```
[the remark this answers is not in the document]
danstarns: Oh, I get it! Yeah, its a bit tricky with the ordering, I guess another way would be `.find(WHERE _count === 3)` and then assert that the find returns something.

packages/client/tests/functional/filter-count-relations/tests.ts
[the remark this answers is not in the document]
danstarns: :(

packages/client/tests/functional/filter-count-relations/tests.ts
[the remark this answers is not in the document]
danstarns: Maybe it's just our line count rules, this line seems quite long for me. Thanks for trying.

packages/client/tests/functional/filter-count-relations/tests.ts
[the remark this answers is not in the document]
SevInf: Actually, I think if we use `cuid` instead of `uuid` in `idForProvider` we'd get stable ordering, let me try that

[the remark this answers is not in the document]
SevInf: `cuid` worked, now we can rely on automatic ids always being consecutive. Rewrote this test to do full assert.

packages/client/src/generation/TSClient/Args.ts
SevInf: Went ahead and removed it completely in a [separate PR](https://github.com/prisma/prisma/pull/14853)

packages/client/src/generation/TSClient/Args.ts
seth: I don't have an opinion on protected vs private here but would like to see this sort of change be its own commit with a clear commit message.

PR #14819, opened 2022-08-15
fix(migrate): --create-only flag patch

bandinopla:
### Why?
currently, if you run `prisma migrate dev --create-only` and the DB schema is not equal to the `scema.prisma` the command will ask for a reset, and it will reapply all the migrations before analyzing the new changes required. 

With this modification, when a developer uses the **--create-only** flag, prisma migrate will skip the DB resetting and applying any migration and jump ahead to finding the diferences between the current schema and the DB's schema...

### motivation?
In the case a developer is working with an existing database and trying to start using prisma migrate **for the first time**, they will most likely be working with a local database already in place in their local environment populated with data, and calling `prisma migrate dev --create-only` as it is right now, it will ask for a reset deleting everything in the DB. _It should allow for a review of the SQL migration file before doing anything._

### useful? when?
To allow the development of a baseline migration quickly:
1. `prisma db pull` on an existing db to create the schema
2. `prisma migrate dev --create-only` to create the first "base" migration
3. `prisma migrate resolve --applied XXX` where XXX is the name of the migration created above

Also... let's be semantic here... **"create only"** should not produce a "reset" nor "apply" anything...


janpio (changes requested):
Making explicit that this is a BREAKING CHANGE. Can not just be merged without prior discussion and handling.

bandinopla:
> Making explicit that this is a BREAKING CHANGE. Can not just be merged without prior discussion and handling.

what whould be the breaking portion of my fix? what breaks? im curious...

janpio:
Conflicts happen when the branch you are trying to merge into changes, leading GitHub to have trouble to figure out what code goes where.

My comment re "breaking change" was not because of that, but because the behavior of the CLI changes. We and our users could rely on that behavior in CLI setups and so on, so something like this could always only be released via a major release.

Jolg42:
Indeed, changing this behavior will need to happen in a major release.

We are aware that this is a misleading flag ([we have a label for it](https://github.com/prisma/prisma/labels/topic%3A%20prisma%20migrate%20dev%20--create-only)).
It's something we want to improve in a major version one day, but before we'll need to redesign it.

So your feedback is interesting for us, how you use Migrate in your workflow, what you expect, what's missing...
If you see other problems with Migrate, you can create an issue, we'll be happy to check that out!

wmadden:
Hi @bandinopla,

Thank you for your contribution and for proposing the --create-only flag modification. After careful consideration, we’ve decided not to implement this change as proposed. We plan to prevent prisma migrate dev from attempting to reset the database and decompose the migrate dev command into smaller operations, allowing users to achieve the same result without adding another flag and without making the dev command's behavior even more conditional. 

Therefore, we’re closing this PR. We appreciate your understanding and look forward to your future contributions.

PR #14833, opened 2022-08-16
chore(temp): benchmark fmt wasm no build copy

jkomyno:


jkomyno:
This PR doesn't make sense, as it fails without the build copy.

PR #14867, opened 2022-08-17
test(migrate): better/cleaner process.exit mock

Jolg42:
Removes 
```
          // For snapshot test, because exit() is mocked
          return ``
```
And makes the expectations clearer -> should reject when `process.exit(1)`


packages/migrate/src/__tests__/DbDrop.test.ts
```diff
@@ -106,15 +106,14 @@ No such file or directory (os error 2)
 
   it('should be cancelled (prompt)', async () => {
     ctx.fixture('reset')
-    const mockExit = jest.spyOn(process, 'exit').mockImplementation()
+    const mockExit = jest.spyOn(process, 'exit').mockImplementation((number) => {
+      throw new Error('process.exit: ' + number)
+    })
```
jkomyno, line 111: I'd move all these duplicated `mockExit` definitions elsewhere, defining them only once.
There’s this `jestProcessContext` defined [here](https://github.com/prisma/prisma/blob/27c038fa2b61161fb9164f826f0ef1b574dff445/packages/internals/src/utils/jestContext.ts#L142-L167). Currently it mocks `process.stdout.write` and `process.stderr.write`. I’d suggest using an analogous approach for mocking `process.exit` in this PR, but we can postpone this to a future PR as discussed in private.

jkomyno (approved):
Nice PR overall, with a comment to address in a future PR. Let's 🚢 this

PR #14881, opened 2022-08-18
test(client): Strip @ts-ignore when generating tests

SevInf:
Internal discussion: https://prisma-company.slack.com/archives/C02FZENHD4N/p1660821454931419

Also, fixes all the type errors

### DIFF
--- a/packages/client/helpers/generator-test/templates/tests.ts.tpl
+++ b/packages/client/helpers/generator-test/templates/tests.ts.tpl
@@ -1,7 +1,8 @@
 import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from '@prisma/client'
 
-// @ts-ignore this is just for type checks
-declare let prisma: import('@prisma/client').PrismaClient
+declare let prisma: PrismaClient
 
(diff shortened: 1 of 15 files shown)


packages/client/helpers/generator-test/templates/tests.ts.tpl
```diff
@@ -1,7 +1,8 @@
 import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from '@prisma/client'
 
-// @ts-ignore this is just for type checks
-declare let prisma: import('@prisma/client').PrismaClient
+declare let prisma: PrismaClient
```
aqrln, line 5: Is this change now required or is it just a code style change to reduce the number of `// @ts-ignore` directives?

aqrln (approved):

packages/client/helpers/generator-test/templates/tests.ts.tpl
SevInf: It is necessary if we want to use `NewPrismaClient ` type. For some reason I could not get to accept neither `import('@prisma/client').PrismaClient` nor `typeof import('@prisma/client').PrismaClient`

millsp (approved):

PR #14882, opened 2022-08-18
chore: avoid circular type dependencies

jkomyno:
- [x] move `Library` type from @prisma/engine-core to @prisma/internals
- [x] move `plusX` utility dependency in `@prisma/internals` from `@prisma/engine-core` to `@prisma/fetch-engine`


packages/fetch-engine/src/util.ts
```diff
@@ -88,3 +88,17 @@ async function removeFileIfExists(filePath: string) {
     }
   }
 }
+
+export function plusX(file: fs.PathLike): void {
+  const debug = Debug('plusX')
```
aqrln, line 93:
```suggestion
  const debug = Debug('prisma:fetch-engine:plusX')
```

packages/fetch-engine/src/util.ts
```diff
@@ -33,6 +31,8 @@ export async function getRootCacheDir(): Promise<string | null> {
 }
 
 export async function getCacheDir(channel: string, version: string, platform: string): Promise<string | null> {
+  const debug = Debug('prisma:cache-dir')
```
janpio, line 34:
```suggestion
  const debug = Debug('prisma:fetch-engine:cache-dir')
```

millsp (changes requested):
Let's remove the dependency out of the `package.json` now :)

packages/internals/src/engine-commands/Library.ts
```diff
@@ -0,0 +1,35 @@
...
+  env: NodeJS.ProcessEnv | Record<string, string>
+}
+
+export type ConfigMetaFormat = {
+  datasources: DataSource[]
+  generators: GeneratorConfig[]
+  warnings: string[]
+}
+
+export type Library = {
```
millsp, line 21: I think this is now duplicated in `engine-core`, right? Could you please import this in `engine-core` in `type/Library` to deduplicate it?

jkomyno:
> Let's remove the dependency out of the `package.json` now :)

Let's recall we still have this, @millsp:

```typescript
// https://github.com/prisma/prisma/blob/main/packages/internals/src/get-generators/getGenerators.ts#L2

import { fixBinaryTargets, getOriginalBinaryTargetsValue, printGeneratorConfig } from '@prisma/engine-core'

// ...
```

millsp:
> Let's recall we still have this, @millsp:

Nice catch, thanks. What should we do with this then? Do you think it belongs in `internals`?

jkomyno:
> > Let's recall we still have this, @millsp:
> 
> Nice catch, thanks. What should we do with this then? Do you think it belongs in `internals`?

Hard to move things considering the current state of things. 

Currently, `fixBinaryTargets` and `printGeneratorConfig` is used in a duplicated `getFixedGenerator()` method defined twice in `@prisma/engine-core`:

```typescript
  // TODO Fixed as in "not broken" or fixed as in "written down"? If any of these, why and how and where?
  private getFixedGenerator(): string {
    const fixedGenerator = {
      ...this.config.generator!,
      binaryTargets: fixBinaryTargets(this.config.generator!.binaryTargets, this.platform!),
    }

    return printGeneratorConfig(fixedGenerator)
  }
```

This `getFixedGenerator()` is used once in [`BinaryEngine.ts`](https://github.com/prisma/prisma/blob/f6d1a6591ea56747f31ec099dce16a2b0b338b5e/packages/engine-core/src/binary/BinaryEngine.ts#L417-L427) and once in [`DefaultLibraryLoader.ts`](https://github.com/prisma/prisma/blob/9d038c873bc66a71d9ea74708b39614f7b1050ee/packages/engine-core/src/library/DefaultLibraryLoader.ts#L108-L119) in a similar fashion, which are again in `@prisma/engine-core`.

jkomyno:
> ```ts
> // https://github.com/prisma/prisma/blob/main/packages/internals/src/get-generators/getGenerators.ts#L2
> 
> import { fixBinaryTargets, getOriginalBinaryTargetsValue, printGeneratorConfig } from '@prisma/engine-core'
> ```

Current problem about this line:
- we initially thought we could potentially move `get-generators` to `@prisma/client`, then we realized `@prisma/migrate` uses `get-generators` as well, and as `@prisma/cli` depends on both `@prisma/client` and `@prisma/migrate`, we would simply move the circular dependency problem without solving it.

packages/internals/src/engine-commands/Library.ts
jkomyno: Before doing anything like this, we need to stop having `@prisma/internals` depend on `@prisma/engine-core`

millsp:
Sounds reasonable, `client generation` isn't a migration concern, after all. What this the benefit of passing `tryRunGenerate` as an argument to migration commands vs doing it it in the cli after just executing the migration command?

jkomyno:
> Sounds reasonable. Client generation isn't a migration concern, after all. What this the benefit of passing `tryRunGenerate` as an argument to migration commands vs doing it it in the cli after just executing the migration command?

There are `migrate` commands that invoke `tryRunGenerate` before the `$Command:parse()` method has completed its execution. One such example is `prisma migrate dev`, as seen in [here](https://github.com/prisma/prisma/blob/4e7bc7b32b5a8bf9fa0148b4762a72b24e2a7fd6/packages/migrate/src/commands/MigrateDev.ts#L308-L312).

jkomyno:
As of [this commit](https://github.com/prisma/prisma/pull/14882/commits/8f7295ba4993ef86257fb8c81ba7b867d078a262):
- `get-generators` is now in `@prisma/client/generator-build`
- `@prisma/internals` no longer depends on `@prisma/engine-core`
- `@prisma/migrate` no longer depends on `get-generators`

Please take this commit as an evaluation of whether it made sense to move all these files around to achieve the scope of avoiding circular dependencies, and to highlight any file that shouldn't have been moved (if any).
Note:
- `@prisma/studio` is looking for `get-generators` in `@prisma/internals`, thus building the `cli` won't work
- I haven't moved around tests yet, that'll come later

PR #14884, opened 2022-08-18
chore(internals): error on generate for proxy + tracing

millsp:
As discussed during the deep dive, we are adding a check failing when `tracing` and `dataProxy` feature flags are used together.


packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -55,12 +55,6 @@ export class DataProxyEngine extends Engine {
     this.host = host
 
     debug('host', this.host)
-
-    if (this.config.previewFeatures?.includes('tracing')) {
```
janpio, line 59: maybe replace with a mention of packages/internals/src/get-generators/utils/check-feature-flags/checkFeatureFlags.ts, as this seems to be a natural location to do this manually instead?

Jolg42 (approved):

PR #14889, opened 2022-08-18
test(integration): unexpected citext case sensitivity for joins

mechanical-turk:
Hi prisma team,

I'm opening this PR to show you with a test case that prisma's WHERE-IN mechanic for relational joins appear to be not respecting case-insensitivity defaults of `CITEXT` columns. Here's a breakdown of the test case:

1) Create 2 tables. One of the tables has a foreign key to the other, where the key is a CITEXT on both tables.
2) Insert to both tables
3) Run a findUnique with an include to the other and see that the join was successful
4) Update the column that acts as a foreign key with an uppercase version
5) Run the same findUnique query again and expect to get a successful join
6) But see that the join was not successful

# Conclusion

Prisma's internal join logic treats CITEXT as case sensitive by default, even though it shouldn't. How can I help you guys fix this? I'm happy to send a followup commit if you can point me in the right direction

### DIFF
--- a/packages/integration-tests/src/__tests__/integration/postgresql/__scenarios.ts
+++ b/packages/integration-tests/src/__tests__/integration/postgresql/__scenarios.ts
@@ -4,6 +4,70 @@ import type { Input } from '../../__helpers__/integrationTest'
 
 export const scenarios = [
   {
+    name: 'citext joins should respect case insensitivity by default',
+    up: `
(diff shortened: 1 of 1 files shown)


matthewmueller:
Thanks @mechanical-turk, we've confirmed this bug and created an issue for this bug: https://github.com/prisma/prisma/issues/14935

Jolg42:
Closing this as there is an issue for it.

PR #14910, opened 2022-08-20
fix(client): Prisma Client always receive engine spans

### LINKED ISSUE #14842: Prisma Client always received engine spans even when not tracing
It looks like the engine always emits spans regardless of whether we send it a traceparent for an operation or not. I think this could cause performance issues if we always emit spans even for users that are not actually using otel?

On the client side, we receive a `parent_span_id` of `000000000000` when we land in this situation. This came up while testing sampling for the documentation section. An easy way to reproduce this is to use a sampler but set to sample nothing, then see how engine spans are still being emitted. Might be useful to add it as a regression test:

```ts
  const provider = new BasicTracerProvider({
    sampler: new TraceIdRatioBasedSampler(0),
    resource: new Resource({
      // we can define some metadata about the trace resource
      [SemanticResourceAttributes.SERVICE_NAME]: "basic-service",
      [SemanticResourceAttributes.SERVICE_VERSION]: "1.0.0",
    }),
  });
```

_Originally posted by @millsp in https://github.com/prisma/client-planning/issues/71#issuecomment-1208552297_

Potential Solution:
Stop passing `undefined` as `traceId`

danstarns:
Closes: https://github.com/prisma/prisma/issues/14842


packages/client/src/runtime/RequestHandler.ts
```diff
@@ -149,7 +150,7 @@ export class RequestHandler {
             document,
             runInTransaction,
           },
-          (params) => this.dataloader.request(params),
+          (params) => this.dataloader.request({ ...params, tracingConfig: this.client._tracingConfig }),
```
millsp, line 153: I think this area of the code is barely tested, and unused (see all the missing args). I think we should totally remove it at some point, it's not part of the public API.

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -81,9 +82,9 @@ export class RequestHandler {
     this.hooks = hooks
     this.dataloader = new DataLoader({
       batchLoader: (requests) => {
-        const info = getRequestInfo(requests[0])
+        const info = getRequestInfo({ ...requests[0], tracingConfig: client._tracingConfig })
```
millsp, line 85: The tracing config should already be on the request object
```suggestion
        const info = getRequestInfo(requests[0])
```

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -81,9 +82,9 @@ export class RequestHandler {
...
         const queries = requests.map((r) => String(r.document))
-        const traceparent = getTraceParent(requests[0].otelParentCtx)
+        const traceparent = getTraceParent({ context: requests[0].otelParentCtx, tracingConfig: client._tracingConfig })
```
millsp, line 87:
```suggestion
        const context = requests[0].otelParentCtx
        const traceparent = getTraceParent({ ...requests[0], context })
```

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -92,7 +93,7 @@ export class RequestHandler {
         return this.client._engine.requestBatch(queries, info.headers, info.inTx)
       },
       singleLoader: (request) => {
-        const info = getRequestInfo(request)
+        const info = getRequestInfo({ ...request, tracingConfig: client._tracingConfig })
```
millsp, line 96:
```suggestion
        const info = getRequestInfo(request)
```

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -52,13 +52,14 @@ export type Request = {
   headers?: Record<string, string>
   otelParentCtx?: Context
   otelChildCtx?: Context
+  tracingConfig: TracingConfig
```
millsp, line 55:
```suggestion
  tracingConfig?: TracingConfig
```

packages/engine-core/src/tracing/getTraceParent.ts
```diff
@@ -1,16 +1,26 @@
...
  * context if no parameter is provided.
  * @param context an otel context
  */
-export function getTraceParent(context?: Context) {
+export function getTraceParent({
+  context,
+  tracingConfig,
+}: {
+  context?: Context
+  tracingConfig: TracingConfig
```
millsp, line 15:
```suggestion
  tracingConfig?: TracingConfig
```

millsp:
We should remove this now
https://github.com/prisma/prisma/blob/e3d8d6af3f49b6a3ea3addefca31626f5d673a25/packages/engine-core/src/tracing/createSpan.ts/#L14-L15

packages/client/src/__tests__/batching.test.ts
```diff
@@ -1,15 +1,23 @@
+import { TracingConfig } from '@prisma/engine-core'
+
 import { blog } from '../fixtures/blog'
 import { getDMMF } from '../generation/getDMMF'
 import { DMMFClass, makeDocument } from '../runtime'
 import { RequestHandler } from '../runtime/RequestHandler'
 
 describe('batching', () => {
```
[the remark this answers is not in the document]
danstarns: It's required so there is no workaround for it. Without it will fail with 'cannot read .enabeld' of undefined.

danstarns:
> We should remove this now https://github.com/prisma/prisma/blob/e3d8d6af3f49b6a3ea3addefca31626f5d673a25/packages/engine-core/src/tracing/createSpan.ts/#L14-L15

traceFlags is a required type on `SpanContext` what should it be, always Sampled?

packages/client/src/runtime/RequestHandler.ts
danstarns: This is tested here: https://github.com/prisma/prisma/blob/main/packages/client/src/__tests__/integration/happy/middlewares/test.ts#L54

danstarns:
Related: https://github.com/prisma/prisma-engines/pull/3143

millsp:
It was a workaround because engines would always send the spans. Now that the engines aren't sending the span if we tell it that an operation is not sampled (via `00-00-00-00`), we should be able to remove it. In other words, yes, sampling should always be true there and the test you added for non-sampling should succeed.

packages/client/src/runtime/RequestHandler.ts
millsp: I guess I wanted it more for being standard with the other props, and if one day we unit test this (if not already) then we would not need to pass any config.

danstarns:
> It was a workaround because engines would always send the spans. Now that the engines aren't sending the span if we tell it that an operation is not sampled (via `00-00-00-00`), then this isn't needed. In other words, yes, sampling should always be true there and the test you added for non-sampling should succeed.

Cool thanks for the explanation, it's been removed.

packages/client/src/runtime/RequestHandler.ts
danstarns: Good point, I did not think about this. Have removed it!

packages/client/src/__tests__/integration/happy/logging/library.ts
```diff
@@ -104,34 +104,34 @@ test('interactive transactions logging - library', async () => {
...
+          query: BEGIN,
+          target: quaint::connector::metrics,
+          timestamp: 1970-01-01T00:00:00.000Z,
+        },
+      ],
+      Array [
+        Object {
+          duration: 0,
+          params: [0],
+          query: SELECT "public"."User"."id" FROM "public"."User" WHERE 1=1 OFFSET $1 /* traceparent=00-00-00-00 */,
```
millsp, line 121: I think this needs to only appear if we are actually tracing, wdyt? Let's take a team decision for what we want to do.

millsp (approved):
Nice work! I approve this PR and your approach but please double check my comment just above related to the engines. Thanks!

packages/client/src/__tests__/integration/happy/logging/library.ts
garrensmith: I think adding it in here is acceptable. The final `00` means this shouldn't be sampled. From my understanding of tracing it seems ok to pass a trace-id and span id to other components even if it is not being sampled.

PR #14925, opened 2022-08-21
fix(client): use `globalThis` to get env vars in Cloudflare Workers

### LINKED ISSUE #14924: Environment variables not available when using Wrangler 2
### Bug description

When using the Data Proxy feature on the Cloudflare Workers runtime I'm getting an error caused by what I think is inability of Prisma edge client to access env variables set by Wrangler.
What's important to note is that I'm using Wrangler 2 (what you get by default if you run `npm i wrangler`). It's different than what's used in the [ecosystem-tests](https://github.com/prisma/ecosystem-tests/tree/dev/dataproxy/cloudflare-workers) where the legacy Wrangler 1 is used.

```
Your worker has access to the following bindings:
- Vars:
  - CF_DATA_PROXY_URL: "..."

...

'Uncaught Error: InvalidDatasourceError: Datasource "db" references an environment
  variable "CF_DATA_PROXY_URL" that is not set\n'
```

### How to reproduce

1. `git clone git@github.com:prisma/ecosystem-tests.git`
2. `cd ecosystem-tests/dataproxy/cloudflare-workers`
3. `npm i `
4. `npm i -D wrangler`
5. `npm uninstall @cloudflare/wrangler`
6. `source ./prepare.sh`
7. Remove `usage_model = ''` line from `wrangler.toml`
8. `npx prisma generate --data-proxy`
9. `npx wrangler dev src/index.ts`

### Expected behavior

There's no error.

### Prisma information

As in the [example](https://github.com/prisma/ecosystem-tests/tree/dev/dataproxy/cloudflare-workers).

### Environment & setup

- OS: macOS Monterey 12.1
- Database: PostgreSQL
- Node.js version: 16.13.2

### Prisma Version

Experienced it in my project on `4.2.1`. Reproduced also on `4.3.0-dev.48`.

drzamich:
Fixes #14924

Use `globalThis` instead of `global` to access global variables in Cloudflare Workers. 

`globalThis` is a runtime-agnostic reference and will always work, while `global` is not defined in Web Workers.

### DIFF
--- a/packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
+++ b/packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
@@ -71,7 +71,7 @@ function getSelectedEnvVarNames(datasources: InternalDatasource[]) {
  */
 export function getRuntimeEdgeEnvVar(envVarName: string) {
   // for cloudflare workers, an env var is a global js variable
-  const cfwEnv = `typeof global !== 'undefined' && global['${envVarName}']`
+  const cfwEnv = `typeof globalThis !== 'undefined' && globalThis['${envVarName}']`
   // for vercel edge functions, it's injected statically at build
   const nodeOrVercelEnv = `typeof process !== 'undefined' && process.env && process.env.${envVarName}`
 


aqrln (approved):
Thank you!

PR #14951, opened 2022-08-23
fix(migrate dev): exit(1) if seeding errors

### LINKED ISSUE #14862: CLI: `migrate dev` should return a non-successful exit code (1) when there is an error during seeding
### Bug description

From https://github.com/prisma/prisma/issues/12349

We have a different behavior for the error exit code between `migrate dev` and `migrate reset` if there is an error during seeding: dev will exit 0, while reset 1 

### How to reproduce

- add the following to the package.json
```
  "prisma": {
    "seed": "node ./prisma/seed.js"
  },
```
- Create the `./prisma/seed.js` file containing:
```js
throw new Error("Error from seeding")
```
- create a `./prisma/schema.prisma` containing the following
```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
  url      = "file:dev.db"
}

model Post {
}
```
- run `npx prisma migrate dev`

### Expected behavior

`npx prisma migrate dev` should exit with an exit code = 1.

### Prisma Version

```
4.3.0-dev.27
```

Jolg42:
Closes https://github.com/prisma/prisma/issues/14862

### DIFF
--- a/packages/migrate/src/__tests__/MigrateDev.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDev.test.ts
@@ -826,6 +826,9 @@ describe('sqlite', () => {
   it('one broken seed.js file', async () => {
     ctx.fixture('seed-sqlite-js')
     fs.write('prisma/seed.js', 'BROKEN_CODE_SHOULD_ERROR;')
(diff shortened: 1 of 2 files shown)


jkomyno (approved):
Nice!

Issue #12349, opened 2022-03-15
[CLI] Review exit code behaviour of CLI commands


do4gr:
Migrate status still returns a 0 exit code even though a clearly defined error is returned (P1000 failed to connect to db in my case). We should change the code to non-zero.

We should also review the behaviour of the other CLI commands.

Jolg42:
Related issues so far
https://github.com/prisma/prisma/issues/9786
https://github.com/prisma/prisma/issues/10265

AndrewSouthpaw:
Can confirm I just bumped into this issue as well. Here's my context:

I have a CLI I'm building with `oclif`, to add additional help for my team in interacting with the DB. We have our local DB, but we also have a test DB. In my `db:migrate` CLI command, I first want to migrate the local dev DB, and then reset the test DB. However, if the person cancels the local dev DB migration (e.g. they're prompted to reset and they decline), I don't want to proceed. I was expecting a cancelled reset to return an exit code of `1`, but it's `0`.

In case it's helpful to hear about use cases! I'm not sure what other workarounds would be present here.

Jolg42:
Thank you @AndrewSouthpaw for your detailed feedback 👍🏼

shayneczyzewski:
Hi gang! I will +1 this initiative. I work at Wasp, where we are building a simple DSL for developing full-stack web apps with less code. We use Prisma in our stack and really like it! :) We also use the Prisma CLI under the covers in our Wasp CLI. One thing we want to do next is to check the status of our users' DB when we generate the web app and warn them if they should migrate.

Some commands, like `prisma migrate diff` have an `--exit-code` flag, which is super helpful when programmatically invoking it. Other commands, like `prisma status`, do not offer this (nor does it even return 1 on error), which means we have to parse stdout output, making it both error-prone and fragile. (Aside: I would use diff to check if migrations needed to be applied, but I would rather not have to deal with the shadow database stuff.)

I'd love to see `--exit-code` make its way to other commands, (selfishly) starting with `prisma status`. But even just getting success/error (0/1) working for all commands would be a huge step in the right direction for programmatic invocation! :D

Thanks, and keep up the great work!

Jolg42:
Tested with `4.3.0-dev.27`

| Command | Success | Error | Notes |
| --- | --- | --- |  --- |
| `format` | 0 | 1 (e.g. schema validation error) |
| `generate` | 0 | 1 (e.g. schema validation error) |
[...]

PR #14979, opened 2022-08-24
fix(cli): return exit code 130 (SIGINT) for no answer/canceling a prompt

Jolg42:
We use the https://github.com/terkelg/prompts library
It doesn't do any handling about exit codes.

Adding a cancel exit code makes it possible to build things on top of the CLI:
https://github.com/prisma/prisma/issues/12349#issuecomment-1204657971

Inquirer.js (used by Heroku CLI) handles cancellation with 130 (SIGINT)

This PR changes are
- return SIGINT = `exit(130)`
	- for all cancelled prompts
	- if a SIGINT event is catched by the CLI
- `handlePanic` now returns `exit(1)` instead of `exit(0)` to signal that there was an error (even though the handling might be successful, the end result is that something crashed and a report was maybe sent.)

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -153,7 +153,7 @@ async function main(): Promise<number> {
   const result = await cli.parse(commandArray)
   // Did it error?
   if (result instanceof HelpError) {
-    console.error(result)
+    console.error(result.message)
     // TODO: We could do like Bash (and other)
     // = return an exit status of 2 to indicate incorrect usage like invalid options or missing arguments.
     // https://tldp.org/LDP/abs/html/exitcodes.html
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -153,10 +153,11 @@ async function main(): Promise<number> {
   const result = await cli.parse(commandArray)
   // Did it error?
   if (result instanceof HelpError) {
-    console.error(result.message)
-    // Like Bash
-    // return an exit status of 2 to indicate incorrect usage like invalid options or missing arguments.
-    return 2
+    console.error(result)
+    // TODO: We could do like Bash (and other)
+    // = return an exit status of 2 to indicate incorrect usage like invalid options or missing arguments.
+    // https://tldp.org/LDP/abs/html/exitcodes.html
+    return 1
   } else if (isError(result)) {
     console.error(result)
     return 1
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -58,7 +58,7 @@ process.on('unhandledRejection', (e) => {
   debug(e)
 })
 // Listen to Ctr + C and exit
-process.on('SIGINT', () => {
+process.once('SIGINT', () => {
   process.exit(130)
 })
 
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -57,6 +57,10 @@ process.on('uncaughtException', (e) => {
 process.on('unhandledRejection', (e) => {
   debug(e)
 })
+// Listen to Ctr + C and exit
+process.on('SIGINT', () => {
+  process.exit(130)
+})
 
 if (process.argv.length > 1 && process.argv[1].endsWith('prisma2')) {
   console.log(
@@ -144,16 +148,21 @@ async function main(): Promise<number> {
       'telemetry',
     ],
   )
-  // parse the arguments
-  const result = await cli.parse(commandArray)
 
+  // Execute the command
+  const result = await cli.parse(commandArray)
+  // Did it error?
   if (result instanceof HelpError) {
     console.error(result.message)
-    return 1
+    // Like Bash
+    // return an exit status of 2 to indicate incorrect usage like invalid options or missing arguments.
+    return 2
   } else if (isError(result)) {
     console.error(result)
     return 1
   }
+
+  // Success
   console.log(result)
 
   /**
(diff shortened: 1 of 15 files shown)


packages/cli/src/bin.ts
```diff
@@ -144,16 +148,22 @@ async function main(): Promise<number> {
...
   )
-  // parse the arguments
-  const result = await cli.parse(commandArray)
 
+  // Execute the command
+  const result = await cli.parse(commandArray)
+  // Did it error?
   if (result instanceof HelpError) {
-    console.error(result.message)
+    console.error(result)
```
Jolg42, line 156:
```suggestion
    console.error(result.message)
```

packages/internals/src/utils/getGithubIssueUrl.ts
```diff
@@ -82,17 +82,16 @@ export async function wouldYouLikeToCreateANewIssue(options: IssueOptions) {
...
 const issueTemplate = (platform: string, options: IssueOptions) => {
   return stripAnsi(`
-Hi Prisma Team! Prisma Migrate just crashed. ${
-    options.reportId
-      ? `This is the report:
-  Report Id: ${options.reportId}`
-      : ''
-  }
+Hi Prisma Team! The following command just crashed.
+${options.reportId ? `The report Id is: ${options.reportId}` : ''}
```
Jolg42, line 94: Note: random improvement

Jolg42: We have many cases where we say `Prisma Migrate just crashed` but the command is not `migrate x`.

packages/migrate/src/commands/MigrateDev.ts
```diff
@@ -249,8 +252,10 @@ ${chalk.bold('Examples')}
       const getMigrationNameResult = await getMigrationName(args['--name'])
 
       if (getMigrationNameResult.userCancelled) {
+        console.info(getMigrationNameResult.userCancelled)
```
Jolg42, line 255:
```suggestion
```

packages/migrate/src/commands/MigrateDev.ts
```diff
@@ -250,7 +253,8 @@ ${chalk.bold('Examples')}
 
       if (getMigrationNameResult.userCancelled) {
         migrate.stop()
```
Jolg42, line 255:
```suggestion
	console.log(getMigrationNameResult.userCancelled)
        migrate.stop()
```

packages/migrate/src/commands/MigrateDev.ts
```diff
@@ -249,8 +252,10 @@ ${chalk.bold('Examples')}
       const getMigrationNameResult = await getMigrationName(args['--name'])
 
       if (getMigrationNameResult.userCancelled) {
+	console.log(getMigrationNameResult.userCancelled)
         migrate.stop()
```
Jolg42, line 256:
```suggestion
        console.log(getMigrationNameResult.userCancelled)
        migrate.stop()
```

packages/internals/src/__tests__/handlePanic.test.ts
```diff
@@ -38,6 +38,8 @@ describe('handlePanic', () => {
   beforeEach(async () => {
     jest.resetModules() // most important - it clears the cache
     process.env = { ...OLD_ENV } // make a copy
+    // Simulate CI environment
+    process.env.GITHUB_ACTIONS = 'true'
```
jkomyno, line 42:
```suggestion
    process.env = { ...OLD_ENV, GITHUB_ACTIONS: 'true' } // make a copy and simulate CI environment
```

packages/cli/src/bin.ts
```diff
@@ -144,16 +148,22 @@ async function main(): Promise<number> {
...
-  const result = await cli.parse(commandArray)
 
+  // Execute the command
+  const result = await cli.parse(commandArray)
+  // Did it error?
   if (result instanceof HelpError) {
     console.error(result.message)
+    // TODO: We could do like Bash (and other)
+    // = return an exit status of 2 to indicate incorrect usage like invalid options or missing arguments.
+    // https://tldp.org/LDP/abs/html/exitcodes.html
```
jkomyno, line 159: Is there anything stopping us from returning `2`, then?

packages/cli/src/bin.ts
```diff
@@ -57,6 +57,10 @@ process.on('uncaughtException', (e) => {
 process.on('unhandledRejection', (e) => {
   debug(e)
 })
+// Listen to Ctr + C and exit
+process.on('SIGINT', () => {
+  process.exit(130)
+})
```
jkomyno, line 63: Let's be explicit

```suggestion
// Listen to Ctr + C and exit
process.once('SIGINT', () => {
  process.exit(130)
})
```

jkomyno (approved):
Approved with comments. In general, I believe a oneliner `const exitDueToCancel = () => { process.exit(130) }` function definition would have probably been more explicit, rather than copying the `// Return SIGINT exit code to signal that the process was cancelled.` comment several times, what would you say?

packages/cli/src/bin.ts
Jolg42: I think we can do that separately with tests, and merge the current PR asap.

packages/migrate/src/bin.ts
```diff
@@ -26,6 +26,10 @@ process.on('uncaughtException', (e) => {
 process.on('unhandledRejection', (e, promise) => {
   console.log(String(e), String(promise))
 })
+// Listen to Ctr + C and exit
+process.on('SIGINT', () => {
```
Jolg42, line 30:
```suggestion
process.once('SIGINT', () => {
```

PR #14982, opened 2022-08-24
feat(client): Integrate field references

### LINKED ISSUE #5048: Allow WHERE conditions to compare columns in same table
## Problem
Prisma should allow `WHERE` conditions to compare columns in same table

## Suggested solution
Expand `WHERE` filters to include columns of the same type from the same table

Simple example model
```
model Shipment {
  id               String @id @default(cuid())
  quantityShipped  Int
  quantityReceived Int
}
```

Raw SQL:
```
SELECT * FROM Shipment WHERE quantityReceived < quantityShipped;
```

Currently for example model:
```
type IntFilter = {
  equals?: number | null
  not?: number | IntFilter | null
  in?: Enumerable<number> | null
  notIn?: Enumerable<number> | null
  lt?: number | null
  lte?: number | null
  gt?: number | null
  gte?: number | null
}
```

Could alternately be:
```
type IntFilter = {
  equals?: number | ShipmentIntColumnInput | null
  not?: number | IntFilter | ShipmentIntColumnInput | null
  in?: Enumerable<number> | Enumerable<ShipmentIntColumnInput> | null
  notIn?: Enumerable<number> | Enumerable<ShipmentIntColumnInput> | null
  lt?: number | ShipmentIntColumnInput | null
  lte?: number | ShipmentIntColumnInput | null
  gt?: number | ShipmentIntColumnInput | null
  gte?: number | ShipmentIntColumnInput | null
}
```

## Alternatives
Currently only raw queries work

Thanks!

### LINKED ISSUE #14664: [MongoDB] Insensitive filters don't exclude undefineds
## Overview

Consider the following datamode 👇l,

```prisma
model User {
  id Int @id @map("_id")
  name? String
}
```

the following dataset 👇,

```js
[
  { id: 1, name: "toto" },
  { id: 2 },
]
```

And the following query 👇:

```ts
await prisma.user.findMany({ where: { name: { lt: "z", mode: insensitive } } })
```

It should return:
```js
[
  { id: 1, name: "toto" },
]
```

But it returns:
```js
[
  { id: 1, name: "toto" },
  { id: 2 },
]
```

SevInf:
Integrates engine changes from prisma/prisma-engines#3088.

Generates new kinds of types from DMMF: field references. Those
types are generic and parametrized with 2 parameters:
- Model name
- Type name
Those 2 generic types allow us to have compile-time error in case
incorrect/incompatible column reference is used.

As a result of this change, some of the input types have also became
generic. Logic for deciding if type needs generic `$PrismaModel`
paramter is following:
- if `meta.source` property is set, type never requires generic
  parameter. This new property was added to DMMF to indicate that the
  type is derived from specific model. If that's the case, we know which
  model field references use for sure and don't need generic parameters
- Otherwise, if input type has any fields that accept field references,
  it requires generic paramters. This logic also applies to nested
  object types: if object type needs generic parameter, all objects
  including the field of this type will need one as well, unless
  `meta.source` is known.

In runtime, `model.fields` implemented as a proxy, returning
wrapper objects for references. For runtime validation, we also need to
know model name, this particular field refers to.

Fixes #5048


SevInf:
Note: tests are failing because engines are not yet bumbed here, waiting for test fix and publish

seth:
Dropping by with a ❤️ for this commit message and PR summary. I like how you've described the change at a high level.

packages/client/src/generation/utils.ts
```diff
@@ -344,3 +349,73 @@ export function unique<T>(arr: T[]): T[] {
...
+      // if the type is recursive, let other fields make a decision
+      continue
+    }
+    visited.add(name)
+
+    if (current.location === 'fieldRefTypes') {
+      return true
+    }
+
+    if (current.location === 'inputObjectTypes' && typeof current.type === 'object') {
```
danstarns, line 378:
```suggestion
    if (current.location === 'inputObjectTypes') {
```

Would it always be an object?

packages/client/src/generation/utils.ts
```diff
@@ -344,3 +349,73 @@ export function unique<T>(arr: T[]): T[] {
...
+
+/**
+ * Determines if arg types need generic <$PrismaModel> argument added.
+ * Essentially, performs breadth-first search for any fieldRefTypes that
+ * do not have corresponding `meta.source` defined.
+ *
+ * @param type
+ * @returns
+ */
+export function needsGenericModelArg(type: DMMF.SchemaArgInputType) {
```
danstarns, line 361: Maybe i don't see the larger picture here, but could this be simplified down into something along the lines of: 

```ts
export function needsGenericModelArg(type: DMMF.SchemaArgInputType) {
  if (type.location === 'fieldRefTypes') {
    return true
  }

  const hasFields =
    type.location === 'inputObjectTypes' && typeof type.type === 'object' && !type.type.meta?.source

  if (hasFields) {
    return type.type.fields.some((type) => needsGenericModelArg(type))
  }

  return false
}
```

?

packages/generator-helper/src/dmmf.ts
```diff
@@ -116,6 +115,10 @@ export namespace DMMF {
       model?: SchemaEnum[]
       prisma: SchemaEnum[]
     }
+    fieldRefTypes: {
+      // model?: FieldRefType[]
```
danstarns, line 119: 👀

packages/client/src/generation/utils.ts
SevInf: This is very similar to what I've started with, but it has a problem: could be recursive, for example `TypeA` has field of `TypeB` and type has a field of `TypeA`. In that case, this version will never terminate. We need to keep track of types we've already seen to avoid this. Current version is more or less standard approach to breadth- or depth-first search in the graph.

packages/client/src/generation/utils.ts
SevInf: Thought so as well, but it some cases it is actually not: for example, [here](https://github.com/prisma/prisma/blob/2d36554ac7a537d082a405ac1d722d48efe86147/packages/client/src/generation/TSClient/Args.ts#L57) we create an input type manually and put a string there. There are couple of other similar places in generation code. Ideally, it would be nice to get rid of such cases and rely purely on `location` but I don't think it's feasible to do it in the scope of this task.

packages/client/src/generation/utils.ts
danstarns: > This is very similar to what I've started with, but it has a problem: types could be recursive, for example `TypeA` has field of `TypeB` and type has a field of `TypeA`. In that case, this version will never terminate. We need to keep track of types we've already seen to avoid this. Current version is more or less standard approach to breadth- or depth-first search in the graph.

Ok, I understand and this is why you have the 'visited' map. Nice work!

packages/client/src/generation/utils.ts
```diff
@@ -344,3 +349,73 @@ export function unique<T>(arr: T[]): T[] {
...
+        continue
+      }
+
+      if (inputType.meta?.source) {
+        // if source is defined, we know model for sure and do not need generic argument
+        continue
+      }
+
+      for (const field of inputType.fields) {
+        toVisit.push(...field.inputTypes)
```
millsp, line 390: Interesting way of achieving recursion with a loop. This might not be obvious at first, would you mind adding a comment?

packages/client/src/generation/TSClient/Input.ts
```diff
@@ -141,13 +155,20 @@ ${indent(
...
   TAB_SIZE,
 )}
 }`
     return `
-export type ${type.name} = ${body}`
+export type ${this.getTypeName()} = ${body}`
+  }
+
+  private getTypeName() {
+    if (inputTypeNeedsGenericModelArg(this.type)) {
```
millsp, line 169: This method looks exclusive to field references, should we rename it to make it more explixit?

packages/generator-helper/src/dmmf.ts
```diff
@@ -116,6 +115,10 @@ export namespace DMMF {
       model?: SchemaEnum[]
       prisma: SchemaEnum[]
     }
+    fieldRefTypes: {
```
millsp, line 118: It could have been easier to understand perhaps if it was called inputRefTypes (I think you suggested this @weakky)

packages/client/src/generation/utils.ts
```diff
@@ -344,3 +349,73 @@ export function unique<T>(arr: T[]): T[] {
...
+}
+
+function getTypeName(type: DMMF.SchemaArgInputType) {
+  if (type.location === 'scalar' || type.location === 'fieldRefTypes') {
+    return type.type as string
+  }
+  return (type.type as DMMF.InputType).name
+}
+
+export function getRefAllowedTypeName(type: DMMF.OutputTypeRef) {
```
millsp, line 409: How does it work with enums? I think they must be prefixed with `Prisma.` or did I miss something?

millsp: Or is it not necessary because we are in the right namespace already?

packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -34,14 +35,20 @@ const aggregateProps = ['aggregate', 'count', 'groupBy'] as const
 export function applyModel(client: Client, dmmfModelName: string) {
   // we use the javascript model name for display purposes
   const jsModelName = dmmfToJSModelName(dmmfModelName)
+  const model = client._baseDmmf.modelMap[dmmfModelName]
+  const fieldsProxyEnabled = client._engineConfig.previewFeatures?.includes('fieldReference')
   const ownKeys = getOwnKeys(client, dmmfModelName)
   const baseObject = {} // <-- user mutations go in there
+  let fieldsProxy: FieldProxy | null = null
```
millsp, line 42:
```suggestion
  let fieldsProxy: FieldProxy | undefined
```
I try to always avoid null, is there a specific reason in this case?

packages/client/src/runtime/FieldRef.ts
```diff
@@ -0,0 +1,28 @@
...
+  public readonly isList: boolean
+
+  constructor(modelName: Model, name: string, fieldType: FieldType, isList: boolean) {
+    this.modelName = modelName
+    this.name = name
+    this.typeName = fieldType
+    this.isList = isList
+  }
+
+  _toGraphQLInputType() {
```
millsp, line 24: Neat.

packages/client/src/runtime/FieldRef.ts
```diff
@@ -0,0 +1,28 @@
...
+ * A reference to a specific field of a specific model
+ */
+export interface FieldRef<Model, FieldType> {
+  readonly modelName: Model
+  readonly name: string
+  readonly typeName: FieldType
+  readonly isList: boolean
+}
+
+export class FieldRefImpl<Model, FieldType> implements FieldRef<Model, FieldType> {
```
millsp, line 11: I think this belongs in core/model because anything outside core is considered "dirty", which isn't the case here :D.

packages/client/src/runtime/query.ts
```diff
@@ -610,6 +615,10 @@ function stringify(value: any, inputType?: DMMF.SchemaArgInputType) {
     return JSON.stringify(value.toString('base64'))
   }
 
+  if (value instanceof FieldRefImpl) {
+    return `{ _ref: ${JSON.stringify(value.name)}}`
```
millsp, line 619: Do we have a form of validation at runtime or is it type-level only?

packages/client/tests/functional/field-reference/json/tests.ts
```diff
@@ -0,0 +1,131 @@
...
+      expect(products).toEqual([expect.objectContaining({ title: 'potato' })])
+    })
+  },
+  {
+    optOut: {
+      from: ['sqlserver', 'sqlite'],
+      reason: 'JSON column type is not supported',
+    },
+  },
+)
```
millsp, line 131: It would be nice to have a test case where it fails at type/runtime?

packages/generator-helper/src/dmmf.ts
```diff
@@ -116,6 +115,10 @@ export namespace DMMF {
...
+      // model?: FieldRefType[]
+      prisma?: FieldRefType[]
```
millsp, line 120: Question out of curiosity: Is this treated as an InputObjectType in the query.ts.

millsp (approved):
I really appreciate the care you took to write the tests. I left you some comments but nothing stood out as being problematic. Great work.

packages/client/tests/functional/field-reference/json/tests.ts
```diff
@@ -0,0 +1,131 @@
...
+      })
+
+      const products = await prisma.product.findMany({
+        where: { properties1: { equals: prisma.product.fields.properties2 } },
+      })
+
+      expect(products).toEqual([expect.objectContaining({ title: 'Potato' })])
+    })
+
+    test('does not conflict with {_ref: "something"} json value', async () => {
```
danstarns, line 46: What does this mean, and whats the reason for this edge case?

packages/client/tests/functional/field-reference/json/tests.ts
```diff
@@ -0,0 +1,131 @@
...
+            properties1: { tags: ['red', 'tasty'] },
+            properties2: 'not potato',
+          },
+        ],
+      })
+      const products = await prisma.product.findMany({
+        where: {
+          properties1: {
+            // @ts-test-if: provider === 'postgresql' || provider === 'cockroachdb'
+            path: ['tags'],
```
danstarns, line 116: It would be good to see a deeper path here, maybe add a test case for a few levels deep like: 

```
path: ['tags', 'foo', 'bar']
```

millsp:
I'm surprised there's an error in the runtime `d.ts` bundle eg:

```
    export enum ModelAction {
        findUnique = "findUnique",
        findFirst = "findFirst",
        findMany = "findMany",
        create = "create",
        createMany = "createMany",
        update = "update",
        updateMany = "updateMany",
        upsert = "upsert",
        delete = "delete",
        deleteMany = "deleteMany",
        groupBy = "groupBy",
        count = "count",
        aggregate = "aggregate",
        findRaw = "findRaw",
        aggregateRaw = "aggregateRaw"
    }
        {}; // <------ here
}
```

Something for you to take a look on Monday @SevInf

packages/client/src/generation/utils.ts
SevInf: Yes, we are in `Prisma` namespace already.

packages/client/src/runtime/query.ts
```diff
@@ -1504,9 +1539,13 @@ function scalarToArg(key: string, value: any, arg: DMMF.SchemaArg, inputType: DM
 function objectToArgs(
   initialObj: any,
   inputType: DMMF.InputType,
+  context: MakeDocumentContext,
   possibilities?: DMMF.SchemaArgInputType[],
   outputType?: DMMF.OutputType,
 ): Args {
+  if (inputType.meta?.source) {
+    context = { modelName: inputType.meta.source }
```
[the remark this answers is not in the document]
SevInf: Model name is actually stored on the reference. We are passing it down from through the call stack just so we can be sure that the reference used references the correct model:

```ts
prisma.user.findMany({
   select: {
      posts: { where: {
         
         /** 
         * We know that `someField` referes to `NotPosts` model, but we still
         * need to know what kinds of models `gt` accepts to perform the validation
         * /
         updated: { gt: prisma.notPosts.fields.someField  }
      }}
   }
})
```

packages/client/src/runtime/query.ts
SevInf: It's both run- and compile-time.

packages/client/tests/functional/field-reference/json/tests.ts
SevInf: When converting to GraphQL, field reference turns into `{ _ref: "columnName" }` object.

So this query:

```ts
where: { field: { gt: prisma.model.fields.someField }}
```

becomes this:

```query
where: { field: { gt: { _ref: 'someField' }}}
```

In case of JSON column, we need to ensure that if user explicitly passes `{ _ref: 'something' }` as an input to json filter it is treated as a raw json value and not as a field reference.

packages/client/tests/functional/field-reference/json/tests.ts
SevInf: Added one above

packages/client/src/generation/utils.ts
SevInf: Added a comment at the start of the function

packages/client/tests/functional/field-reference/json/tests.ts
SevInf: Done

packages/client/src/runtime/FieldRef.ts
SevInf: Moved the file

aqrln (approved):
Belated LGTM, great work! 🚀

PR #14986, opened 2022-08-24
feat(client): Multi Schema

danstarns:
Adds CRUD coverage across relationships using Multi Schema

Integrates: https://github.com/prisma/prisma-engines/pull/3139


SevInf:
Would you think a test case where we have modles with the same table name (with `@@map`) in different schemas will make sense? Just to make sure nothing explodes in this case.

danstarns:
> Would you think a test case where we have modles with the same table name (with `@@map`) in different schemas will make sense? Just to make sure nothing explodes in this case.

See change: https://github.com/prisma/prisma/pull/14986/commits/31c1b538882b06209d846818e3b6f8547ea7dd7b

packages/client/package.json
```diff
@@ -142,7 +142,7 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "4.3.0-26.6200af94deb7856d96ef4d83298f645ee24b9d3f"
+    "@prisma/engines-version": "4.3.0-25.qe-multi-schema-eb81cea0c7c7c3624102ebb6bcba12b8b8a602cc"
```
Jolg42, line 145: Will be on main in a bit with 
https://github.com/prisma/prisma/pull/15001

packages/client/tests/functional/multi-schema/_matrix.ts
```diff
@@ -0,0 +1,18 @@
...
+      provider: 'postgresql',
+    },
+
+    {
+      provider: 'sqlserver',
+    },
+  ],
+  [
+    {
+      mapTable: 'true',
```
millsp, line 15: Doesn't this mean that `mapTable` is always true? Perhaps you could also make it a boolean instead of a string.

danstarns: Using the matrix here so we run the suite first without this item and then second with it, this is a typing issue will update the types to allow booleans too.

packages/client/tests/functional/multi-schema/tests.ts
```diff
@@ -0,0 +1,131 @@
...
+            },
+          },
+        },
+        select: {
+          email: true,
+          posts: true,
+        },
+      })
+
+      expect(found).toHaveLength(0)
```
Jolg42, line 122:
```suggestion
      expect(await prisma.post.findMany({
        where: {
          title: newTitle
        }
      })).toHaveLength(0)

      expect(await prisma.user.findMany({
        where: {
          email: newEmail
        }
      })).toHaveLength(0)
```

Would give a better confidence that both are indeed deleted, the current query will be empty even if only the user was deleted

Jolg42 (approved):
Looks good, see comment for better delete check

packages/client/tests/functional/multi-schema/tests.ts
danstarns: Nice idea.

PR #14996, opened 2022-08-25
test: try reduce flaky tests in CI

Jolg42:
Trying to get rid of most flaky tests recently.
Increasing timeout
for `invalid-executable` use `jest.retryTimes(3)`

### DIFF
--- a/packages/fetch-engine/src/__tests__/download.test.ts
+++ b/packages/fetch-engine/src/__tests__/download.test.ts
@@ -13,8 +13,10 @@ const CURRENT_ENGINES_HASH = enginesVersion
 console.debug({ CURRENT_ENGINES_HASH })
 const FIXED_ENGINES_HASH = 'da41d2bb3406da22087b849f0e911199ba4fbf11'
 const dirname = process.platform === 'win32' ? __dirname.split(path.sep).join('/') : __dirname
+const isMacOrWindowsCI = Boolean(process.env.CI) && ['darwin', 'win32'].includes(process.platform)
 
-jest.setTimeout(200_000)
+// Network can be slow, especially for macOS in CI.
+jest.setTimeout(300_000)
 
 describe('download', () => {
   beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -14,7 +14,8 @@ console.debug({ CURRENT_ENGINES_HASH })
 const FIXED_ENGINES_HASH = 'da41d2bb3406da22087b849f0e911199ba4fbf11'
 const dirname = process.platform === 'win32' ? __dirname.split(path.sep).join('/') : __dirname
 
-jest.setTimeout(200_000)
+// Network can be slow, especially for macOS in CI.
+jest.setTimeout(300_000)
```
jkomyno, line 18: Should we increase the timeout programmatically in some target systems, like `darwin`, then?

Jolg42: It can be slow regardless of the system so a global increase is ok. Also we are not benchmarking our CDN, just having a "reasonable limit"

Jolg42: This is the download everything case, which is a lot and totally not what a user would experience.

packages/generator-helper/src/__tests__/generatorHandler.test.ts
```diff
@@ -73,6 +73,12 @@ describe('generatorHandler', () => {
 
   // TODO: Windows: this test fails with ENOENT even though the .cmd file is there and can be run manually.
   testIf(process.platform !== 'win32')('parsing error', async () => {
+    // It fails often and randomly on GitHub Actions with
+    // Received promise resolved instead of rejected
+    // Resolved to value: undefined
+    // So we retry it with the hope that it a retry would help
+    jest.retryTimes(3)
```
jkomyno, line 80: This might be ok for the time being, but I don't think solves the flakiness problem at its source. I'm a bit skeptical of the `GeneratorProcess` definition and its use of `cross-spawn`, which might use some improvements in the future

jkomyno (approved):
Let's merge it now, and keep an eye on the comments for a future PR (after the correctness sprint)

PR #15000, opened 2022-08-25
test(client): Sanitize error snapshots

### LINKED ISSUE #14321: Sanitize error snapshots in new test setup
Our error snapshots contain file paths and line numbers of the offending code. This causes the need to update snapshots every time we add or remove a line below, which causes additional noise during code reviews. We should sanitize line/column information to make snapshots more resilient. For example, we could sanitize them to look like: 

```
Invalid \`prisma.user.create()\` invocation in
/client/tests/functional/interactive-transactions/tests.ts:XX:XX
  XX  */
  XX testIf(getClientEngineType() === ClientEngineType.Library)('batching rollback', async () => {
  XX   const result = prisma.$transaction([
→ XX    prisma.user.create(
```

SevInf:
Introduce custom jest matchers that sanitize output of
`printError` function. I opted for custom matcher instead of
adding this functionality to default snapshot serializer because there
might be some cases, where we actually want to see the line numbers in
the snapshots (#14265 is one example). This won't be possible if line
numbers are always sanitized.

Close #14321

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -107,6 +107,7 @@
     "is-regexp": "2.1.0",
     "jest": "28.1.3",
     "jest-junit": "14.0.0",
+    "jest-snapshot": "28.1.3",
     "js-levenshtein": "1.1.6",
     "klona": "2.0.5",
     "lz-string": "1.4.4",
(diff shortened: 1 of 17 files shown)


danstarns (commented):
Should this wait until your referred PR has been merged so we know we have coverage for the line numbers?

SevInf:
> Should this wait until your referred PR has been merged so we know we have coverage for the line numbers?

That's a good idea, i'll try to find time for reviving that one soon.

danstarns (approved):
Now that you merged [the refactor of the print stack function](https://github.com/prisma/prisma/pull/14265), with its unit tests, the changes here are fine to merge. Nice work - this is soo much more manageable.

PR #15008, opened 2022-08-25
fix(internals, migrate): split isCi into isInteractive

Jolg42:
Note: We still want to throw errors if standard input is not a TTY (= non interactive) or if `isCI() === true`
If not interactive: we cannot prompt, prompt will not work
if running in CI: the prompt will hang the process until a timeout kicks in

Also, in the past we used `GITHUB_ACTIONS` env var in our tests to simulate CI because we evaluated this lazily at runtime.
I moved the logic we used to be evaluated lazily at runtime, it should be fast enough and makes it more flexible for testing.

Follows https://github.com/prisma/prisma/pull/14649


packages/internals/src/utils/isCi.ts
```diff
@@ -1,5 +1,52 @@
...
+// Returns true if the current environment is a CI environment.
 export const isCi = (): boolean => {
-  return !process.stdin.isTTY || isCiLib || Boolean(process.env.GITHUB_ACTIONS)
+  const env = process.env
+
+  // From https://github.com/watson/ci-info/blob/44e98cebcdf4403f162195fbcf90b1f69fc6e047/index.js#L54-L61
+  // Evaluating at runtime makes it possible to change the values in our tests
+  // This list is probably not exhaustive though `process.env.CI` should be enough
+  //  but since we were using this utility in the past, we want to keep the same behavior
+  return !!(
```
aqrln, line 9: Do we need all of these? Maybe let's just leave what GitHub Actions and Buildkite use?

aqrln: Actually I think we can just check for `env.CI`:

* https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
* https://buildkite.com/docs/pipelines/environment-variables#CI

Jolg42: The detection is user facing so that users don't get their CI run hanging forever on our interactive prompts.
These are from `ci-info` package which was what the previous logic was using, so there is no expected change after.

packages/internals/src/utils/__tests__/isInteractive.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  })
+  afterAll(() => {
+    process.env = { ...originalEnv }
+  })
+
+  describe('in non TTY environment', () => {
+    const mockedValue = { isTTY: false }
+    test('isInteractive should be false', () => {
+      // @ts-ignore
+      expect(isInteractive({ stream: mockedValue })).toBe(false)
```
jkomyno, line 17: Do we really want to avoid typing the `isInteractive()` argument?

packages/internals/src/utils/canPrompt.ts
```diff
@@ -0,0 +1,18 @@
...
+export const canPrompt = (): boolean => {
+  // Note: We use prompts.inject() for testing prompts in our CI
+  if (Boolean((prompt as any)._injected?.length) === true) {
+    return true
+  } else if (isInteractive() && !isCi()) {
+    return true
+  } else {
+    return false
+  }
+}
```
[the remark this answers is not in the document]
jkomyno: Nit, I don't find branched boolean returns useful in general

jkomyno (approved):
Nice PR, approved with comments

packages/internals/src/utils/__tests__/isInteractive.test.ts
Jolg42: What do you mean?

Jolg42: Ah the `@ts-ignore`

Jolg42: fixed

PR #15010, opened 2022-08-25
test(client): run functional tests in Data Proxy simulator

aqrln:
It is now possible to use `pnpm test:functional --data-proxy` or `pnpm test:functional --data-proxy --edge-client` to run the functional tests in Mini-Proxy — the local Data Proxy simulator specifically made for use in tests (internal: [repo](https://github.com/prisma/mini-proxy/)).

It is similar to how, e.g., Cloudflare has Miniflare, but is an internal tool for now.

Although it doesn't share any code or architecture with the actual Data Proxy and is significantly simpler implementation of the Data Proxy API, the expectation is it should be a reasonable approximation for use in Prisma Client tests and allow us to have more confidence in the `DataProxyEngine` code and faster feedback loop. However, it is still essential to have `ecosystem-tests` running with the real Data Proxy.

Note: `bash` and `openssl` need to be installed and available in `PATH` for this to work on Windows.

Closes https://github.com/prisma/client-planning/issues/80

---

Use this to make some diffs in tests manageable to read:


packages/client/src/__tests__/integration/happy/blog-env-postgresql/test.ts
```diff
@@ -19,6 +21,17 @@ test('Blog fixture: Postgres', async () => {
...
+    const { stdout } = await execa('mini-proxy', [
+      'connection-string',
+      '-u',
+      originalConnectionString,
+      '-e',
+      'TEST_POSTGRES_URI',
+    ])
+    originalConnectionString = stdout.trim()
+  }
+
```
aqrln, line 34: Not necessary anymore

```suggestion
```

packages/client/src/utils/getTestClient.ts
```diff
@@ -60,7 +60,7 @@ export async function getTestClient(schemaDir?: string, printWarnings?: boolean)
     relativeEnvPaths,
     datasourceNames: config.datasources.map((d) => d.name),
     activeProvider,
-    dataProxy: false,
+    dataProxy: !!process.env.DATA_PROXY,
```
aqrln, line 63: Likely also not necessary anymore, to re-check

aqrln: although doesn't hurt to have it anyway to be able to run old tests manually

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -20,12 +21,18 @@ type SuitePlanContext = {
...
 
-export function getTestSuitePlan(suiteMeta: TestSuiteMeta, suiteConfig: NamedTestSuiteConfig[]): TestPlanEntry[] {
+export function getTestSuitePlan(
+  suiteMeta: TestSuiteMeta,
+  suiteConfig: NamedTestSuiteConfig[],
+  options?: MatrixOptions,
+): TestPlanEntry[] {
   const context = buildPlanContext()
 
+  const shouldSkipAll = !!process.env.DATA_PROXY && !!options?.skipDataProxy
```
aqrln, line 31: Document new option

.eslintignore
```diff
@@ -14,9 +14,10 @@ packages/client/generator-build/
 packages/client/declaration/
 packages/client/runtime/
 packages/client/src/__tests__/types/
+packages/client/helpers/mini-proxy/
```
aqrln, line 17: To be removed when the package is published

.gitignore
```diff
@@ -44,4 +44,6 @@ reproductions/*
 
 dev.db
 junit.xml
-/output.txt
\ No newline at end of file
+/output.txt
+
+/packages/client/helpers/certs
```
aqrln, line 49: To be removed when the package is published

.prettierignore
```diff
@@ -0,0 +1 @@
+/packages/client/helpers/mini-proxy
```
aqrln, line 1: To be removed when the package is published

packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
```diff
@@ -61,30 +61,30 @@ exports[`interactive-transactions (provider=mysql) batching raw rollback 1`] = `
 Invalid \`prisma.$executeRaw()\` invocation:
 
 
-  Raw query failed. Code: \`1062\`. Message: \`Duplicate entry '1' for key 'user.PRIMARY'\`
+  Raw query failed. Code: \`1062\`. Message: \`Duplicate entry '1' for key 'PRIMARY'\`
```
aqrln, line 64: Unrelated change in the snapshot, not sure why it was different on my machine.

```suggestion
  Raw query failed. Code: \`1062\`. Message: \`Duplicate entry '1' for key 'user.PRIMARY'\`
```

packages/client/tests/functional/_utils/getTestSuitePlan.ts
aqrln: also consider making it

```ts
skipDataProxy: {
  reason: string
}
```

packages/client/helpers/mini-proxy/generate-certs.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
aqrln, line 1: this will be in a separate npm package

packages/client/helpers/mini-proxy/index.d.ts
```diff
@@ -0,0 +1,54 @@
+/// <reference types="node" />
```
aqrln, line 1: build artifact, this will be in a separate npm package

packages/client/src/utils/getTestClient.ts
Jolg42, line 63:
```suggestion
    dataProxy: Boolean(process.env.DATA_PROXY),
```
If it's still needed, using `Boolean()` is an easier one to read

packages/client/tests/functional/_utils/getTestSuitePlan.ts
Jolg42, line 31:
```suggestion
  const shouldSkipAll = Boolean(process.env.DATA_PROXY) && Boolean(options?.skipDataProxy)
```

Jolg42:
Note:
- packages/client/helpers/mini-proxy/index.js
- packages/client/helpers/mini-proxy/mini-proxy.js
Seems to be build artifacts and could be removed from git?

Jolg42 (commented):
Could you run the tests here too?
https://github.com/prisma/prisma/blob/main/.buildkite/publish/run.sh#L24-L32
https://github.com/prisma/prisma/blob/main/.buildkite/test/run.sh#L40-L45

aqrln:
@Jolg42 they are build artifacts, yes. I left comments about them above. They will be removed once `mini-proxy` is on npm.

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -12,45 +17,102 @@ const args = arg(
     '--no-types': Boolean,
     '--types-only': Boolean,
     '--provider': [String],
+    '--data-proxy': Boolean,
+    '--no-mini-proxy': Boolean,
+    '--mini-proxy-debug': Boolean,
```
Jolg42, line 22: Maybe add a small comment about what the flags are doing?

aqrln:
@Jolg42 

> Could you run the tests here too?
> https://github.com/prisma/prisma/blob/main/.buildkite/publish/run.sh#L24-L32
> https://github.com/prisma/prisma/blob/main/.buildkite/test/run.sh#L40-L45

Good point about the publish pipeline.

I didn't run it in the test pipeline on Buildkite intentionally and added Node.js 14 to the matrix on GitHub Actions instead to avoid running it twice (once with `PRISMA_CLIENT_ENGINE_TYPE=binary` and once with `PRISMA_CLIENT_ENGINE_TYPE=library`, which would be irrelevant and only waste CPU time). But it still needs to be run in the publish pipeline I think.

Jolg42:
Discussed offline:
We can ignore adding it to Buildkite, the only benefit would be that we run Node.js 14 on Buildkite but this will be tested on Node.js 14 on GitHub actions.

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -148,19 +149,39 @@ export async function dropTestSuiteDatabase(
...
 }
 
+export type DatasourceInfo = {
+  envVarName: string
+  databaseUrl: string
+  dataProxyUrl?: string
+}
+
 /**
  * Generate a random string to be used as a test suite db url.
```
aqrln, line 159: update comment

.github/workflows/test.yml
```diff
@@ -196,6 +196,71 @@ jobs:
           files: ./packages/client/src/__tests__/coverage/clover.xml
           flags: client,${{ matrix.os }},${{ matrix.queryEngine }}
           name: client-${{ matrix.os }}-${{ matrix.queryEngine }}
+
+  #
+  # CLIENT (functional tests with mini-proxy)
+  #
+  client-dataproxy:
```
millsp, line 203: I think the Data Proxy should be exclusively tested with the binary engine, I think this is what it is deployed with?

aqrln: Yes, that's basically the reason I didn't include it in the Buildkite pipeline, to avoid running the Data Proxy tests twice.

(Although even if we did run it with `PRISMA_CLIENT_ENGINE_TYPE=library`, it wouldn't have any effect, mini-proxy always uses the binary engine)

millsp: > always uses the binary engine

Oh, I see. Perfect. Thanks!

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -50,13 +59,17 @@ async function main(): Promise<number | void> {
       NODE_EXTRA_CA_CERTS: miniProxy.defaultCertificatesConfig.caCert,
     })
 
-    if (!args['--no-mini-proxy']) {
-      const qePath = await getBinaryForMiniProxy()
+    if (args['--edge-client']) {
+      jestCli = jestCli.withEnv({
+        DATA_PROXY_EDGE_CLIENT: 'true',
```
Jolg42, line 64: Maybe rename to `TEST_DATA_PROXY_EDGE_CLIENT`, because it looks like a user facing env var at the moment

aqrln: done

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -68,12 +82,18 @@ async function copyPreprocessed(from: string, to: string, suiteConfig: Record<st
...
-function evaluateMagicComment(conditionFromComment: string, suiteConfig: Record<string, string>): boolean {
+function evaluateMagicComment(
+  conditionFromComment: string,
+  suiteConfig: Record<string, string>,
+  clientMeta: ClientMeta,
+): boolean {
   const script = new Script(conditionFromComment)
   const value = script.runInNewContext({
     ...suiteConfig,
+    clientMeta,
```
SevInf, line 96: Is it used anywhere?

SevInf (approved):
Great job! Amazing that we got not only data proxy testing from it, but edge client as well.

packages/client/tests/functional/methods/findFirstOrThrow/tests.ts
```diff
@@ -42,7 +42,7 @@ testMatrix.setupTestSuite((suiteConfig, suiteMeta) => {
     expect(record).toBeNull()
   })
 
-  test('works with interactive transactions', async () => {
+  testIf(!process.env.DATA_PROXY)('works with interactive transactions', async () => {
```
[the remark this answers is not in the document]
aqrln: Yeah, I think it might, but it also might be a bit complicated.

It used to be that way before https://github.com/prisma/prisma/pull/13561, and was changed for reasons, as far as I remember. I feel like it did introduce some inconsistencies in the codebase with the need to have special case handling in some cases, like this: https://github.com/prisma/prisma/blob/d4e397ef893ce3e530adcbf5ea0d9e7bb1e101e2/packages/client/src/runtime/getPrismaClient.ts#L469
but at the same time, without looking at the diff in that PR, I believe it probably should have fixed some other inconsistencies.

The problem here is the ambiguity of the notion of an "engine type": there is _no_ engine of type "Data Proxy" as in the actual Query Engine flavor, but there _is_ a client-side engine abstraction of type `DataProxyEngine`, which is one of the three implementations of the `Engine` abstract class that we have. Wherever these two notions are conflated, there will be some inconsistencies in one or the other way. Thus these should be two different enums and the code might need some refactoring.

Let's discuss this together with @millsp, it will be a good topic for the Deep Dive meeting.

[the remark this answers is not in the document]
aqrln: (FWIW you are also commenting on an outdated diff here, right now it looks like `testIf(!clientMeta.dataProxy)`. It doesn't fundamentally change anything, but just pointing it out because judging from the other comment it looks like you have reviewed different files from different commits — perhaps something went wrong in GitHub UI)

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
aqrln: It isn't. I thought it would be useful for doing something like `// @ts-test-if: clientMeta.runtime !== 'edge'`, but now I see it won't — the types are always the same (and should be). I'll remove this.

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -8,49 +12,133 @@ const allProviders = new Set(Object.values(Providers))
...
+        codeTestCli.withArgs(['--']).withArgs(args['_']).run()
+      }
+
+      if (!args['--no-types']) {
+        jestCli.withArgs(['--', 'typescript']).run()
+      }
+    }
+  } catch (error) {
+    if (error.exitCode) {
+      // If it's execa error, exit without logging: we
```
danstarns, line 112: Would it be handy to log the error here just in case something gets swallowed without us knowing? I can imagine it causing some headaches.

danstarns (commented):
Wondering if having an 'in memory' mocked mini proxy using something like: https://github.com/httptoolkit/mockttp would be beneficial - Could it remove some setup overhead? Though it would be further from the 'real thing'. Keen to hear your thoughts.

.gitignore
```diff
@@ -45,4 +45,4 @@ reproductions/*
 
 dev.db
 junit.xml
-/output.txt
\ No newline at end of file
+/output.txt
```
danstarns, line 48:
```suggestion
/output.txt
server.key
server.csr
server.crt
root_ca.srl
root_ca.key
root_ca.crt
```
Screenshot 2022-08-31 at 09 01 29

Let's exclude these, I am getting all these diffs when running the test suite locally

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -8,49 +12,133 @@ const allProviders = new Set(Object.values(Providers))
...
+    // Don't start the Mini-Proxy server, expect it to be started externally
+    // and listening on the default port.
+    '--no-mini-proxy-server': Boolean,
+    // Don't override NODE_EXTRA_CA_CERTS. Useful if a custom mini-proxy
+    // server you start is not the one in node_modules. You then need to run
+    // `eval $(mini-proxy env)` in your shell before starting the tests.
+    '--no-mini-proxy-default-ca': Boolean,
+    // Enable debug logs in the bundled Mini-Proxy server
+    '--mini-proxy-debug': Boolean,
     '-p': '--provider',
```
danstarns, line 36: Just a style thing, could we put this one next to the `--provider` on line 22? They are tightly related but now a fair few lines apart.

.github/workflows/test.yml
```diff
@@ -196,6 +196,83 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+
+    needs: detect_jobs_to_run
+    if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [buildjet-4vcpu-ubuntu-2004]
+        node: [14, 16, 18]
```
danstarns, line 214:
```suggestion
        node: [14, 16, 18]
        type: ['--data-proxy', '--data-proxy --edge-client']
```

Could we leverage a matrix here? It looks like it's only the command-like arguments that are changing... see other suggestions.

.github/workflows/test.yml
```diff
@@ -196,6 +196,83 @@ jobs:
...
+          cache: 'pnpm'
+          node-version: ${{ matrix.node }}
+
+      - run: bash .github/workflows/scripts/setup.sh
+        env:
+          CI: true
+          SKIP_GIT: true
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+
+      - run: pnpm run test:functional --data-proxy
```
danstarns, line 246:
```suggestion
      - run: pnpm run test:functional ${{ matrix.type }}
```

.github/workflows/test.yml
```diff
@@ -196,6 +196,83 @@ jobs:
...
+          TEST_FUNCTIONAL_MSSQL_URI: 'sqlserver://localhost:1433;database=PRISMA_DB_NAME;user=SA;password=Pr1sm4_Pr1sm4;trustServerCertificate=true;'
+          TEST_FUNCTIONAL_MONGO_URI: 'mongodb://root:prisma@localhost:27018/PRISMA_DB_NAME?authSource=admin'
+          TEST_FUNCTIONAL_COCKROACH_URI: 'postgresql://prisma@localhost:26257/PRISMA_DB_NAME'
+
+      - uses: codecov/codecov-action@v3
+        with:
+          files: ./packages/client/src/__tests__/coverage/clover.xml
+          flags: client,${{ matrix.os }},dataproxy
+          name: client-${{ matrix.os }}-dataproxy
+
```
danstarns, line 275:
```suggestion
```

.github/workflows/test.yml
danstarns: I guess with this it may be harder to debug either one or the other.

danstarns (approved):
**My comments are not blocking** -  Working fine, nice work.

.gitignore
aqrln: This doesn't sound right, they shouldn't be there.

aqrln: Mini-proxy used to live in `helpers` before it was published on npm on Friday, but I can see no way these files could end up there now. Could you check if they still get generated there if you remove them?

aqrln: Also try

```
git fetch
git reset --hard origin/data-proxy-tests
```

before doing that since the branch got force pushed quite a few times

.github/workflows/test.yml
aqrln: This is a nice idea!

aqrln: Thank you!

packages/client/helpers/functional-test/run-tests.ts
aqrln: Good point

packages/client/helpers/functional-test/run-tests.ts
aqrln: I thought so too when I first encountered this code, but in fact it wouldn't really be handy and would only introduce extra noise when any test fails without adding any new information.

Basically there are two possibilities:

1. The error is an actual error with useful information. In that case the check here will fail and the error will be re-thrown below and fully logged onto the screen.

2. The error is artificially raised by `execa` because of a non-zero exit code that the Jest process has terminated with. In that case, the error carries no useful information except the specific exit code. All of the context in the form of `stderr` output has already been printed onto the screen. The exit code doesn't really need to be printed (Jest wouldn't do this when running it without our wrapper, and it will be available in the shell regardless if you need to see it), but we have to make sure to exit with this code (which we do).

It would make sense to print it, though, if we, let's say, always exited with exit code 1 whenever Jest fails and not forwarded the code back to the shell.

.gitignore
danstarns: No problems now, was my branch out of sync.

.github/workflows/test.yml
aqrln: Ah, actually I think I know why I didn't do this (I didn't think of leveraging the matrix but I did think about making these separate jobs manually): that might create a bit too many CI jobs and exhaust the number of runners. Let's leave it for now, and split it later if we see that a single job takes too long.

PR #15014, opened 2022-08-25
chore: engine overrides

millsp:
With this PR, we are enabling faster iterations and easier local development with the engines.

1. Modify `prisma/fetch-engine/package.json`, choose one:

```diff
  "enginesOverride": {
+    "branch": "feat/column-comparison",
+    "folder": "/home/john/dev/prisma/prisma-engines/target/release"
  }
```
* `branch`: to clone the prisma branch and build it
* `folder`: to re-use your own local engine build

2. `pnpm i`

3. Done. The changes are effective across the whole monorepo.

Notes:
This only works locally for now. If you publish an integration version of this, it will not use the override. This is so that we cannot accidentally ship an overridden engine. We also wouldn't want to build the engines on CI (eg. `ecosystem-tests`).


Jolg42:
@millsp could you document this in https://github.com/prisma/prisma/blob/main/TESTING.md?

packages/engines/src/scripts/localinstall.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    // we make sure that we always have the latest changes from the branch
+    await execa('git', ['pull', 'origin', branch], {
+      cwd: enginesRepoDir,
+      stdio: 'inherit',
+    })
+
+    // we use cargo to build all the engines for the branch we checked out
+    await execa('cargo', ['build', '--release'], {
```
Jolg42, line 50: Quickly mention that Rust needs to installed in TESTING.md

Jolg42 (approved):
Looks good, didn't try it yet, but very curious to try that soon!

packages/fetch-engine/package.json
```diff
@@ -12,6 +12,7 @@
     "directory": "packages/fetch-engine"
   },
   "bugs": "https://github.com/prisma/prisma/issues",
+  "enginesOverride": {},
```
danstarns, line 15: Wondering if its best to have this config in a file like: 

```
engines-override.json
```

This is because I see us(well me at least) accidentally commenting changes to this.

millsp: We could make it an env var as well? That way you could not accidentally commit it. But I think the nice part of this is that you could run this on our CI.

Issue #7186, opened 2021-05-20
Migrations commands crash when run on empty schema.prisma


janpio:
- `npm init -y`
- `npm install prisma@dev`
- Then:

```
C:\Users\Jan\Documents\throwaway\envRepro  (envRepro@1.0.0)
λ npx prisma db push
Prisma schema loaded from schema.prisma
Error: Get config thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', query-engine\query-engine\src\opt.rs:148:21
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

C:\Users\Jan\Documents\throwaway\envRepro  (envRepro@1.0.0)
λ npx prisma migrate dev
Prisma schema loaded from schema.prisma
Error: Get config thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', query-engine\query-engine\src\opt.rs:148:21
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

vs. handled error in Introspection:

```
C:\Users\Jan\Documents\throwaway\envRepro  (envRepro@1.0.0)
λ npx prisma introspect
Prisma schema loaded from schema.prisma

Introspecting based on datasource defined in schema.prisma …
Error: There is no datasource in the schema.
```

tomhoule:
Pushing this to team/client because the crash is in the query engine (most likely getConfig).

tomhoule:
We took this anyway since it's introspection-relevant and we know how to fix it.

matthewmueller:
Thanks! I've moved this along.

tomhoule:
We currently don't have bandwidth for that, assigning the client team in addition to us, since it's a crash in the QE, so the issue doesn't get lost.

janpio:
Error message is now different in 3.6.0:

```
C:\Users\Jan\Documents\throwaway\7186  (7186@1.0.0)
λ npx prisma db push
Prisma schema loaded from schema.prisma
Error: Cannot read property 'url' of undefined

C:\Users\Jan\Documents\throwaway\7186  (7186@1.0.0)
λ npx prisma migrate dev
Prisma schema loaded from schema.prisma
Error: Cannot read property 'url' of undefined
```

jkomyno:
# Current behavior

## prisma validate

[...]

PR #15059, opened 2022-08-29
test(cli): empty-schema -> incomplete-schemas

jkomyno:
Brings progress to https://github.com/prisma/prisma/issues/7186.

- [x] `datasource-block-url-env-unset` tests: the schema has a `datasource db` block with an undefined env url 
- [x] `datasource-block-no-url` tests: the schema has a `datasource db` block with no url field
- [x] `empty-schemas` tests: the schema is completely empty, no models, no blocks
- [x] error messages are normalized across library/binary engines


packages/cli/src/__tests__/fixtures/incomplete-schemas/empty-schema/prisma/schema.prisma
```diff
@@ -0,0 +1 @@
+// the schema is completely empty, no models, no blocks
```
Jolg42, line 1: :trollface: it has a comment, so not really "empty" 😄

Jolg42 (approved):
Looks good.

It's not easy to see the differences between the commands (needs to read the full file and take notes)
Maybe having some const would help for the ones that repeat 🤔 (instead of snapshot)

jkomyno:
> Looks good.
> 
> It's not easy to see the differences between the commands (needs to read the full file and take notes) Maybe having some const would help for the ones that repeat 🤔 (instead of snapshot)

I was thinking something among those lines as well, but I was thinking of putting a summary in the original issue directly. Maybe it's clearer if I do both things

Jolg42 (approved):

PR #15088, opened 2022-08-30
test: use mysql in docker-compose.arm64.yml

aqrln:
The official MySQL image has now been supporting aarch64 in the `8.0`
tag for a while (and even before that via the `-oracle` flavor), so
there's no need to override it with MariaDB in
`docker-compose.arm64.yml` anymore.


jkomyno (approved):
Nice, thank you!

PR #15115, opened 2022-08-31
test(client): unskip 2 tests using the isolated postgresql / mysql uris

Jolg42:


millsp:
I think it could be worth porting this to the functional tests (+ all dbs are isolated by default there).

Jolg42:
@millsp I think someone from Client Team could finish that then

millsp:
I am OK merging this for now, because everyone is fully booked atm

Jolg42:
👍🏼 I'll check later, maybe the easy option would be to disable them for the `no-docker` runs

millsp (approved):

PR #15122, opened 2022-08-31
fix(client): performance regression in generate

### LINKED ISSUE #15109: Prisma 4.3.0 takes 100x more time to generate types
### Bug description

V. 4.3.0
187598994-598fb1a8-ab77-4328-8cad-d986b8f75804

V.4.2.1
Screen Shot 2565-08-31 at 12 27 22

### How to reproduce

- Generate Prisma type

### Expected behavior

- Generate Prisma type time should be the same

### Environment & setup

- OS: Mac OS, Ubuntu
- Database: PostgreSQL
- Node.js version: 16

### Prisma Version

```
4.3.0
```

SevInf:
In #14982, we introduced generic input types and a function for
determining if particular input type needs generic argument. That
function traversed whole type graph starting with given types in search
for field references. It did not do any caching, so if particular type
was encountered multiple types in different relations, whole traversal
happened each time.

Fixed by replacing that helper function with a class with a built-in
cache for results. Now, each type would be checked at most once.

Fix #15109


capaj:
Wow that was very fast

packages/client/src/generation/GenericsArgsInfo.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Determines if arg types need generic <$PrismaModel> argument added.
+   * Essentially, performs breadth-first search for any fieldRefTypes that
+   * do not have corresponding `meta.source` defined.
+   *
+   * @param type
+   * @returns
+   */
+  needsGenericModelArg(topLevelType: DMMF.SchemaArgInputType): boolean {
```
millsp, line 22: Could it have simplified things to have a real recursion of a memoized function?

SevInf: Tried that before and not really: I think I severely underestimated amount of DMMF types, created by relations. With recursive version, benchmark from this PR runs out of stack space already with around 30 models.

SevInf: What I do think will make it easier to follow is separating iterative traversal code and actual logic for determining the need for generic. Then, with leaving out a couple of edge-case, this function could be simplified to something along the lines of:

```js
let currentAnswer = false
const typesStack =  []
traverse(startingType, {
   enterInputType(type) {
      currentAnswer = false
   },
   
   enterArg(arg) {
       if (arg.location === 'fieldRefTypes') {
           currentAnswer = true
       }
   }),
   
   leaveInputType(type) {
      this._cache.set(type, currentAnswer)
   }
})
```

However, I've decided not to spend the time on such helper for now, as pull request is already quite big, generic `traverse` helper will take more time and overall complexity of PR won't really decrease. I do think it will make sense to do it eventually, but I'd suggest we take time for that after patch release.

millsp (approved):

packages/client/src/generation/GenericsArgsInfo.ts
millsp: Thanks for the explanation.

Jolg42 (commented):
Note: current benchmarks output
https://github.com/prisma/prisma/runs/8130142882?check_suite_focus=true#step:8:20
```
client generation 100 models with relations x 0.50 ops/sec ±4.82% (7 runs sampled)
```

@SevInf based on this https://github.com/prisma/prisma/blob/8110f2ceb008169fcaa6eef99850290448b61808/.github/workflows/benchmark.yml#L80-L81
It would only alert us for a "200%" threshold?

SevInf:
@Jolg42 yes. It is compares it with a previous run on `main`. So in that case we'll start getting alerts from a drop to 25 ops/second.

PR #15142, opened 2022-09-01
test(setup): copy the types into the test suite root

millsp:
### Short summary
1. With this PR, you will benefit from having types while writing tests
2. Just prefix `@prisma/client` type imports with `./node_modules/`

### More details

With this PR, we improve our own DX for writing tests more comfortably. When you work in a test suite, you will now get type hints for the Prisma Client. To do this, we simply copy the first types that are generated and put the into a `node_modules` at the root of your test suite. With this, you will be able to write tests faster and with more confidence. See the modified `_example` to learn how to use this.

To ensure that you benefit from the most sound behavior, you need to use `--provider=<provider>`. That is not a hard requirement, however, as the emitted types are almost always the same across providers. And you need to run the test suite at least once to get them updated.

### Notes
The first types to end up in the `node_modules` are the ones from the first generated client for a given matrix item (ie. the first test).

### DIFF
--- a/packages/client/tests/functional/_example/tests.ts
+++ b/packages/client/tests/functional/_example/tests.ts
@@ -1,10 +1,14 @@
 import * as path from 'path'
 
 import { getTestSuiteSchema } from '../_utils/getTestSuiteInfo'
+import { NewPrismaClient } from '../_utils/types'
 import testMatrix from './_matrix'
+// @ts-ignore
+import type { Prisma as PrismaNamespace, PrismaClient } from './node_modules/@prisma/client'
 
-// @ts-ignore this is just for type checks
-declare let prisma: import('./node_modules/@prisma/client').PrismaClient
+declare let prisma: PrismaClient
+declare let Prisma: typeof PrismaNamespace
+declare let newPrismaClient: NewPrismaClient<typeof PrismaClient>
 
 testMatrix.setupTestSuite(
   (suiteConfig, suiteMeta) => {
--- a/packages/client/tests/functional/_example/tests.ts
+++ b/packages/client/tests/functional/_example/tests.ts
@@ -4,7 +4,7 @@ import { getTestSuiteSchema } from '../_utils/getTestSuiteInfo'
 import testMatrix from './_matrix'
 
 // @ts-ignore this is just for type checks
-declare let prisma: import('@prisma/client').PrismaClient
+declare let prisma: import('./node_modules/@prisma/client').PrismaClient
 
 testMatrix.setupTestSuite(
   (suiteConfig, suiteMeta) => {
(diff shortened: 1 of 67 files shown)


millsp:
Just played a bit with it, I think I need to add a lockfile for when multiple test suites files are running concurrently.

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -78,6 +81,18 @@ function setupTestSuiteMatrix(
...
 
+      // for better dx, copy the first generated client into the test suite root
+      beforeAll(async () => {
+        const rootNodeModuleFolderPath = path.join(suiteMeta.testRoot, 'node_modules')
+
+        const suiteFolderPath = getTestSuiteFolderPath(suiteMeta, suiteConfig)
+        const suiteNodeModuleFolderPath = path.join(suiteFolderPath, 'node_modules')
+
+        await fs.copy(suiteNodeModuleFolderPath, rootNodeModuleFolderPath, { recursive: true }).catch((e) => {
+          if (e.code !== 'EEXIST') throw e
```
SevInf, line 92: In which case `EEXIST` will happen with `recursive`? I thought it will always override existing files in this case.

packages/client/tests/functional/_example/tests.ts
```diff
@@ -4,7 +4,7 @@ import { getTestSuiteSchema } from '../_utils/getTestSuiteInfo'
 import testMatrix from './_matrix'
 
 // @ts-ignore this is just for type checks
-declare let prisma: import('@prisma/client').PrismaClient
+declare let prisma: import('./node_modules/@prisma/client').PrismaClient
```
SevInf, line 7: Could you also update template for new tests in `helpers/generator-test`?

SevInf (approved):
Did my own my complicated attempt [here](https://github.com/prisma/prisma/commit/8ad44f632cc4e01126f60141b9a509569d9a0bd6), with a goal removing `@ts-ignores` and running it on watch but did not fiinish it before correctness adventures started.

Looking at your PR now - it is so much simpler, that I no longer see that removing `@ts-ignore` is really worth all that complexity I was introducing. Amazing job. I left one comment and one question, but nothing critical.

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
millsp: Yep, you're right. Wrong assumption from my side. Thanks. My initial intention was that if the `node_modules` folder exists, then it would fail and not try to overwrite. Kinda like having a lockfile, but with a folder so that only one jest job does the copy into `node_modules`. I think I can achieve that with mkdir instead.

millsp: Fixed via `mkdir` to use `node_modules` as a locking mechanism.

millsp:
Just a heads up that I updated all the tests to use this from now on.

Issue #8612, opened 2021-08-06
`updateMany()` causes lost-updates


boiledfroginthewell:
### Bug description

`updateMany()` returns incorrect updated counts when multiple processes or asynchronous functions run simultaneously.

The SQL log of `updateMany()` shows it emits `SELECT` and then `UPDATE`. Since prisma uses the default isolation level of DBMS, those queries are NON-REPEATEBLE READ in many DBMS and can cause lost-updates. 
I believe `SELECT ... FOR UPDATE` or single query `UPDATE .... WHERE ...` is required.

### How to reproduce

<details>
<summary>bookSeat.js</summary>

```js
const {PrismaClient} = require('@prisma/client')

const client = new PrismaClient({
  // log: ["query"]
})


async function bookSeat(userId) {
  const movieName = 'Hidden Figures'

  // Find the first available seat
  // availableSeat.version might be 0
  const availableSeat = await client.seat.findFirst({
    where: {
      // Movie: {
      //  name: movieName,
      // },
      claimedBy: null,
    },
    orderBy: [{id: "asc"}]
  })

  if (!availableSeat) {
    throw new Error(`Oh no! ${movieName} is all booked.`)
  }

  // Only mark the seat as claimed if the availableSeat.version
  // matches the version we're updating. Additionally, increment the
  // version when we perform this update so all other clients trying
  // to book this same seat will have an outdated version.
  const seats = await client.seat.updateMany({
    data: {
      claimedBy: userId,
      version: {
        increment: 1,
      },
    },
    where: {
      id: availableSeat.id,
      version: availableSeat.version, // This version field is the key; only claim seat if in-memory version matches database version, indicating that the field has not been updated
    },
  })

  if (seats.count === 0) {
    throw new Error(`That seat is already booked! Please try again.`)
  }

  return seats.count
}


async function demonstrateLostUpdate() {
  if (process.argv[2] === "createData") {
    await client.seat.deleteMany()
    for (let i = 0; i < 1000; i++) {
      await client.seat.create({
        data: {
          id: i,
          version: 0,
          movieId: 1,
          claimedBy: null,
        }
      })
    }
    process.exit()
  }


  const userId = process.argv[2]

  let updatedCount = 0
  for (let i = 0; i < 1000; i++) {
    try {
      updatedCount += await bookSeat(userId)
    } catch {
      // ignore lock failure
    }
  }


  // Detect lost-updates
  const actualCount = await client.seat.count({
    where: {
      claimedBy: userId
[...]

PR #15147, opened 2022-09-02
test(client): regression test for OCC #8612

millsp:
How to run this:
* Clone this repo
* `cd prisma && pnpm i && pnpm run build`
* Make sure that you have these [env vars set](https://github.com/prisma/prisma/blob/main/TESTING.md)
* `cd packages/client && pnpm run test:functional --provider postgresql row-select-update`
* Look at the added test in `row-select-update` and the comments for the methodology


packages/client/tests/functional/row-select-update/prisma/_schema.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URI_${provider}")
+    }
+    
+    model Resource {
+      id       ${idForProvider(provider)}
+      occStamp Int @default(0) @unique
```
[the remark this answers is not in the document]
millsp: This is to easily expose the behavior on single `update`s, which require unique keys in the where condition. This works well because we are only doing occ on one item.

millsp:
Will be merged once fix is working

packages/client/tests/functional/optimistic-concurrency-control/tests.ts
```diff
@@ -80,7 +80,8 @@ testMatrix.setupTestSuite(({ provider }) => {
     expect(totalCount).toBe(1)
   })
 
-  test('upsert', async () => {
+  // issue with mysql
```
janpio, line 83:
```suggestion
  // issue with mysql: https://github.com/prisma/prisma/issues/15470
```

millsp: Thanks!

packages/client/tests/functional/_utils/setupFilesAfterEnv.ts
```diff
@@ -37,4 +34,33 @@ function sanitizeLineNumbers(message: string) {
   return stripAnsi(message).replace(/^(\s*→?\s+)\d+/gm, '$1XX')
 }
 
+/**
+ * Utility that busts test flakiness by running the same test multiple times.
+ * This is especially useful for tests that fail locally but not on CI.
+ * @param n Number of times to repeat the test
+ * @returns
+ */
+function testRepeat(n: number) {
```
SevInf, line 43: Doesn't seem to be used.
Also, what's the difference with `jest.retries(n)`?

janpio: https://prisma-company.slack.com/archives/C03V1UE750D/p1663711087498559?thread_ts=1663709917.516639&cid=C03V1UE750D

> `testRepeat` repeats a test `n` times and fails if any of the repetitions have failed. It's the opposite of `jest.retryTimes` which will succeed if any has passed.

SevInf (approved):

Issue #9372, opened 2021-09-21
Maximum call stack size exceeded when creating huge amount of related rows 


Nickersoft:
### Bug description

Hey Prisma folks!

I'm currently writing a seed script for seeding an online dictionary app, and attempting to insert about 320,000 entry rows for a single Dictionary entity. The code I'm using to achieve this is: 

https://github.com/TheOpenDictionary/odict-web/blob/4658354830399b76c3c488a007433b77a2d139a4/prisma/seed.ts#L33-L80

However, attempting to seed the dictionary results in: 

```
RangeError: Maximum call stack size exceeded
    at flatten (/Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:34611:33)
    at flatMap (/Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:34615:10)
    at Arg.collectErrors (/Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:35855:22)
    at valueToArg (/Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:36181:47)
    at /Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:36431:17
    at Array.reduce (<anonymous>)
    at objectToArgs (/Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:36408:28)
    at tryInferArgs (/Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:36310:40)
    at valueToArg (/Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:36180:16)
    at /Users/tnickerson/base/odict-web/node_modules/@prisma/client/runtime/index.js:36431:17 {
  clientVersion: '3.1.1'
}
```

Which, honestly isn't surprising given the number of rows, but I'm curious if the only alternative here is to run 320,000 promises inserting one row each. I feel like Prisma should attempt to optimize for large transactions such as this under the hood. I know there is been some discussion around this, but I'm curious as to what the best approach for this case currently is. 

### How to reproduce

1. Clone [this repo](https://github.com/TheOpenDictionary/odict-web)
2. Run `yarn`
3. Run `yarn db:up` to spin up a local database (requires Docker)
4. Run `yarn prisma db seed` to attempt to seed

### Expected behavior

Seeding should succeed successfully

### Prisma information

Prisma schema is available in the OSS repo above

[...]

PR #15153, opened 2022-09-02
test: Reproduce #9372

SevInf:
Seems like it is caused by nested create. `createMany` works normally for the same amount of data.


SevInf:
Superseded by #15279

PR #15157, opened 2022-09-02
fix(client): Fix some of the precision issues with small decimals

### LINKED ISSUE #5925: Precision issue when persisting Decimal.js objects
## Bug description

Persisting `new Prisma.Decimal('8.7')` to db results in `8.699999999999999` being saved to db. Same happens with `new Decimal('8.7')` and `8.7` (plain js number).

Same issue happens with a few other numbers:
```
8.8 => 8.800000000000001
9.2 => 9.199999999999999
```

## How to reproduce

Reproduce with repo https://github.com/eemelipa/prisma-test/tree/prisma-decimals branch "prisma-decimals".

Run
```
npm run db:up
npm install
npm run migrate:dev
npm run start
```

See error logs in console and printed object

## Expected behavior

Persisting any of the following

```
    price1: 8.7,
    price2: new Decimal('8.7'),
    price3: new Prisma.Decimal('8.7'),
    price4: '8.7',
```

should result in `8.7` saved to db without precision issues. Currently only passing the value as a string (price4) works as expected.

## Prisma information

See https://github.com/eemelipa/prisma-test/tree/prisma-decimals

## Environment & setup

- OS: Mac OS
- Database: PostgreSQL (at least)
- Node.js version: v14.15.4 
- Prisma version:

```
prisma               : 2.17.0
@prisma/client       : 2.17.0
Current platform     : darwin
Query Engine         : query-engine 3c463ebd78b1d21d8fdacdd27899e280cf686223 (at node_modules/@prisma/engines/query-engine-darwin)
Migration Engine     : migration-engine-cli 3c463ebd78b1d21d8fdacdd27899e280cf686223 (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine : introspection-core 3c463ebd78b1d21d8fdacdd27899e280cf686223 (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary        : prisma-fmt 3c463ebd78b1d21d8fdacdd27899e280cf686223 (at node_modules/@prisma/engines/prisma-fmt-darwin)
Studio               : 0.353.0
```

## Debug Logs

[debug-logs.txt](https://github.com/prisma/prisma/files/6064283/debug-logs.txt)

SevInf:
One of the reasons for decimal precision loss is serializing and parsing
them as floats when sending over to the engine. Serializing them as
strings fixes those kinds of issues.

Note that it does not fixes all cases of precision loss yet: there is
another, separate issue, where precision will be lost after certain
amount of digits. This PR does not fixes those.

Fix #5925

### DIFF
--- a/packages/client/src/runtime/utils/decimalJsLike.ts
+++ b/packages/client/src/runtime/utils/decimalJsLike.ts
@@ -25,12 +25,12 @@ export function isDecimalJsLike(value: unknown): value is DecimalJsLike {
 
 export function stringifyDecimalJsLike(value: DecimalJsLike): string {
   if (Decimal.isDecimal(value)) {
-    return String(value)
+    return JSON.stringify(String(value))
   }
 
-  const tmpDecimal = new Decimal(0) as DecimalJsLike
-  tmpDecimal.d = value.d
-  tmpDecimal.e = value.e
-  tmpDecimal.s = value.s
-  return String(tmpDecimal)
+  const tmpDecimal = new Decimal(0)
+  ;(tmpDecimal as DecimalJsLike).d = value.d
+  ;(tmpDecimal as DecimalJsLike).e = value.e
+  ;(tmpDecimal as DecimalJsLike).s = value.s
+  return JSON.stringify(String(tmpDecimal))
 }
--- a/packages/client/tests/functional/decimal/tests.ts
+++ b/packages/client/tests/functional/decimal/tests.ts
@@ -8,7 +8,7 @@ declare let prisma: PrismaClient
 declare let Prisma: typeof PrismaNamespace
 
 testMatrix.setupTestSuite(
-  () => {
+  ({ provider }) => {
     describe('possible inputs', () => {
       beforeAll(async () => {
         await prisma.user.create({
@@ -83,7 +83,8 @@ testMatrix.setupTestSuite(
       })
 
       // https://github.com/prisma/prisma/issues/5925
-      test.failing('preserves precision when writing shorter numbers to to db', async () => {
+      // fails on sqlite, because sqlite decimal is actually a float
+      testIf(provider !== 'sqlite')('preserves precision when writing shorter numbers to to db', async () => {
         await prisma.user.create({
           data: { money: new Prisma.Decimal('8.7') },
         })
--- a/packages/client/src/__tests__/query/decimal.test.ts
+++ b/packages/client/src/__tests__/query/decimal.test.ts
@@ -31,7 +31,7 @@ test('allows to pass it decimal instance', () => {
   expect(document.toString()).toMatchInlineSnapshot(`
     query {
       findManyUser(where: {
-        money: 123456789.12334
+        money: "123456789.12334"
       }) {
         id
         money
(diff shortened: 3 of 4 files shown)


millsp (approved):

packages/client/src/runtime/utils/decimalJsLike.ts
```diff
@@ -25,12 +25,12 @@ export function isDecimalJsLike(value: unknown): value is DecimalJsLike {
...
+    return JSON.stringify(String(value))
   }
 
-  const tmpDecimal = new Decimal(0) as DecimalJsLike
-  tmpDecimal.d = value.d
-  tmpDecimal.e = value.e
-  tmpDecimal.s = value.s
-  return String(tmpDecimal)
+  const tmpDecimal = new Decimal(0)
+  ;(tmpDecimal as DecimalJsLike).d = value.d
```
millsp, line 32: What is the reason for this change?

SevInf: Previously, cast was done right after tmp instance creation. This is needed because `d`, `p`, `q` are not writable properties.
However, `DecimalJSLike` has no `toFixed` method defined and adding it to the signature now theoretically would be a breaking change. So instead, we keep `tmpDecimal` type to `Decimal` and cast to `DecimalJSLike` when we need writable property.

aqrln (approved):

packages/client/src/runtime/utils/decimalJsLike.ts
millsp: Thanks for the thorough explanation 👍

packages/client/tests/functional/decimal/tests.ts
```diff
@@ -83,7 +83,8 @@ testMatrix.setupTestSuite(
       })
 
       // https://github.com/prisma/prisma/issues/5925
-      test.failing('preserves precision when writing shorter numbers to to db', async () => {
+      // fails on sqlite, because sqlite decimal is actually a float
```
janpio, line 86: Does this point to a bigger problem for SQLite? Is just accepting this (and not running the test) ok? Should we somehow deal with Decimals and SQLite differently?

SevInf: I don't think we should do anything differently for SQLIte, this is mostly on SQLIte itself and not on us. Basically, their `NUMERIC` type [always ends up being either int or float](https://www.sqlite.org/datatype3.html#:~:text=A%20column%20with%20NUMERIC%20affinity%20may%20contain%20values%20using%20all%20five%20storage%20classes.%20When%20text%20data%20is%20inserted%20into%20a%20NUMERIC%20column%2C%20the%20storage%20class%20of%20the%20text%20is%20converted%20to%20INTEGER%20or%20REAL%20(in%20order%20of%20preference)) and there is no real arbitrary precision decimal types (`DECIMAL` is just an alias for `NUMERIC`), just [an optional extension](https://sqlite.org/floatingpoint.html#the_decimal_c_extension) for working with  strings as if they were decimals. So, imho, it is very similar situations to enums - maybe we could tell `migrate` to store them as strings, but then we won't be able to re-introspect them.

With that in mind, I think it's ok to just skip the test, as we are not dealing with decimals any worth then underlying DB does.

Issue #9678, opened 2021-10-08
Unique constraint failed during `$transaction([deleteMany, createMany])`


eannam:
### Bug description

I have a prisma transaction which first runs a `deleteMany` operations followed by a `createMany` operation.

However, occasionally the data in the `createMany` operation violates a unique constraint (a combination of two fields).

The application code should ensure that uniqueness is guaranteed in the data being passed and logging the actual data confirms there are no duplicates on the fields that the error indicates.

Is it possible that the `deleteMany` operation is not completing fully before the `createMany` begins? As there would be overlap across the data being removed and the data being created (relates to refreshing a user's feed).

My understanding is that each operation in a transaction is fully awaited before beginning the next but this is the only thing I can think of.

### Expected behavior

That the `deleteMany` operation completes fully before `createMany` begins and there is no unique constraint violation.

### Environment & setup

- OS: Ubuntu 18.04.5 LTS
- Database: PostgreSQL
- Node.js version: 12.22.6

### Prisma Version

```
prisma                  : 3.0.2
@prisma/client          : 3.0.2
Current platform        : debian-openssl-1.1.x
Query Engine (Node-API) : libquery-engine 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : 2452cc6313d52b8b9a96999ac0e974d0aedf88db
Studio                  : 0.423.0
Preview Features        : fullTextSearch
```

janpio:
Can you share your code? Do you `await` both function calls?

eannam:
Following up here, I replaced the transaction with two individually awaited calls to the DB, i.e.

```    await prisma.playlistStack.deleteMany({ where: { playlistId } });
    await prisma.playlistStack.createMany({ data: playlistStackData });```

[...]

PR #15162, opened 2022-09-02
test(client): reproduction for #9678

millsp:


packages/client/tests/functional/concurrent-transaction-unique/tests.ts
```diff
@@ -0,0 +1,45 @@
...
+        prisma.resource.createMany({ data }),
+      ])
+    }
+
+    for (let i = 0; i < 50; i++) {
+      await fn()
+    }
+  })
+
+  testIf(provider !== 'sqlite' && provider !== 'mysql' && provider !== 'sqlserver').failing(
```
janpio, line 32: Why exclude SQlite, MySQL and SQL Server?

packages/client/tests/functional/concurrent-transaction-unique/tests.ts
```diff
@@ -0,0 +1,45 @@
...
+    const fn = async () => {
+      await prisma.$transaction([
+        prisma.resource.deleteMany({ where: { name: 'name' } }),
+        prisma.resource.createMany({ data }),
+      ])
+    }
+
+    for (let i = 0; i < 50; i++) {
+      await fn()
+    }
```
janpio, line 29: This is meant to test that one can run this a lot of times serially without it failing?

millsp: Yes, the user said it also fails that way, which was suprising but I tested it anyways.

packages/client/tests/functional/concurrent-transaction-unique/tests.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+testMatrix.setupTestSuite(({ provider }) => {
+  testIf(provider !== 'sqlite')('serialized deleteMany/createMany', async () => {
+    const fn = async () => {
+      await prisma.$transaction([
+        prisma.resource.deleteMany({ where: { name: 'name' } }),
+        prisma.resource.createMany({ data }),
+      ])
+    }
+
```
janpio, line 26:
```suggestion
    // Test that running many times serially never causes this to fail (as reported by one user in the issue)
```

packages/client/tests/functional/concurrent-transaction-unique/tests.ts
janpio: https://github.com/prisma/prisma/pull/15162/files#r980403641

packages/client/tests/functional/concurrent-transaction-unique/tests.ts
janpio: Reply via Slack: Because there it does not fail as described (and tested here).

janpio: I assume we do not have a nice way to test that it fails on Mongo, POstgres and Cockroach but _succeeds_ for the others?

millsp: No, I couldn't come up with something better

packages/client/tests/functional/concurrent-transaction-unique/tests.ts
```diff
@@ -0,0 +1,42 @@
...
+  testIf(provider !== 'sqlite')('concurrent deleteMany/createMany', async () => {
+    const fn = async () => {
+      await prisma.$transaction([
+        prisma.resource.deleteMany({ where: { name: 'name' } }),
+        prisma.resource.createMany({ data }),
+      ])
+    }
+
+    await Promise.all([fn(), fn(), fn(), fn(), fn()])
+  })
```
janpio, line 41: - Why does this currently not fail for MySQL and SQL Server?
- CockroachDB and MongoDB are failing with a "nice" error message instead of the Unique constraint failure: `Transaction failed due to a write conflict or a deadlock. Please retry your transaction`

janpio:
I split off this PR: https://github.com/prisma/prisma/pull/15547

millsp:
Completed via https://github.com/prisma/prisma/pull/15494

PR #15196, opened 2022-09-05
fix(client): throw values as errors in itx

### LINKED ISSUE #12663: Interactive Transactions: fatal when throwing string instead of Error() since 3.10.0
### Bug description

works in 3.9.2, fatals as of 3.10.0:

```
await prisma.$transaction((prisma) => { throw "abort transaction"; }) // throwing string, not new Error()
```

error since 3.10.0: `TypeError: Cannot create property 'clientVersion' on string "abort transaction"`

### How to reproduce

^- code above, this one seems straightforward but just let me know if reproduction repo required

### Expected behavior

able to throw an expression of type as an error, as accepted by Typescript compiler
(yes I do typically follow best practice re: `Error`s 😉 ; this broke a unit test that was rolling back a transaction where there wasn't need for a full `Error`)

### Prisma information

N/A

### Environment & setup

- OS: Mac OS
- Database: Postgres
- Node.js version: 14

### Prisma Version

worked in versions up to + including `3.9.2`, fails in `3.10.0` and `3.11.1`

millsp:
closes https://github.com/prisma/prisma/issues/12663

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -998,7 +998,6 @@ new PrismaClient({
         // it went bad, then we rollback the transaction
         await this._engine.transaction('rollback', headers, info).catch(() => {})
 
-        e.clientVersion = this._clientVersion
         throw e // silent rollback, throw original error
       }
 
--- a/packages/client/tests/functional/interactive-transactions/tests.ts
+++ b/packages/client/tests/functional/interactive-transactions/tests.ts
@@ -60,6 +60,8 @@ testMatrix.setupTestSuite(({ provider }) => {
 
     await expect(result).rejects.toMatchObject({
       message: expect.stringContaining('Transaction API error: Transaction already closed'),
+      code: 'P2028',
+      clientVersion: '0.0.0',
     })
 
     expect(await prisma.user.findMany()).toHaveLength(0)
@@ -127,7 +129,7 @@ testMatrix.setupTestSuite(({ provider }) => {
       throw 'you better rollback now'
     })
 
-    await expect(result).rejects.toThrow(`you better rollback now`)
+    await expect(result).rejects.toMatchInlineSnapshot(`you better rollback now`)
 
     const users = await prisma.user.findMany()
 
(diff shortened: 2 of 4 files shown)


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -998,7 +998,6 @@ new PrismaClient({
         // it went bad, then we rollback the transaction
         await this._engine.transaction('rollback', headers, info).catch(() => {})
 
-        e.clientVersion = this._clientVersion
```
aqrln, line 1001: Should we do something like

```ts
if (typeof e === 'object' && e !== null) {
}
```

or was it never intended to be set here in the first place?

packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -113,6 +115,27 @@ testMatrix.setupTestSuite(({ provider }) => {
...
+      await prisma.user.create({
+        data: {
+          email: 'user_1@website.com',
+        },
+      })
+
+      throw 'you better rollback now'
+    })
+
+    await expect(result).rejects.toMatchInlineSnapshot(`you better rollback now`)
```
SevInf, line 132:
```suggestion
    await expect(result).rejects.toBe(`you better rollback now`)
```

So we could be sure we caught a thrown string and not, for example, `Error` object with the same message.

packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -196,19 +219,21 @@ testMatrix.setupTestSuite(({ provider }) => {
...
     })
 
-    1
     await expect(result).rejects.toMatchObject({
       message: expect.stringContaining('Transaction API error: Transaction already closed'),
+      code: 'P2028',
+      clientVersion: '0.0.0',
     })
+
     await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
```
SevInf, line 228:
```suggestion
    await expect(result).rejects.toMatchPrismaErrorInlineSnapshot(`
```

My bad, missed this in my sanitation PR, but it will help with the diff.

packages/client/src/runtime/getPrismaClient.ts
aqrln: (Looking at the new tests — apparently never intended to be set here; if the error was meant to have this field, it would have had it already set before re-throwing, if my understanding is correct. Marking the comment as resolved.)

aqrln (approved):
LGTM pending @SevInf's comments.

Issue #9248, opened 2021-09-14
Validation error when inserting data of type `List<String | DateTime>` into field of type `List<String>`


zhyuhan:
### Bug description

I have a column of type `String[]`, and when I try to insert an array that contains a mix of regular strings and an ISO 8601 formatted datetime string, e.g. `["hello", "2021-09-14T00:00:00.000Z"]`, an error is thrown.

```
PS C:\Users\zhyuhan\Development\prisma-string-datetime> npx prisma db seed
Environment variables loaded from .env
Running seed command `ts-node prisma/seed.ts` ...
prisma:info Starting a postgresql pool with 13 connections.
prisma:query BEGIN
prisma:query SELECT "public"."Post"."id" FROM "public"."Post" WHERE "public"."Post"."id" = $1 OFFSET $2
prisma:query SELECT "public"."Post"."id" FROM "public"."Post" WHERE "public"."Post"."id" = $1
prisma:query SELECT "public"."Post"."id", "public"."Post"."words" FROM "public"."Post" WHERE "public"."Post"."id" = $1 LIMIT $2 OFFSET $3
prisma:query COMMIT
Successfully created post #1 with words hello
prisma:query BEGIN
prisma:query SELECT "public"."Post"."id" FROM "public"."Post" WHERE "public"."Post"."id" = $1 OFFSET $2
prisma:query SELECT "public"."Post"."id" FROM "public"."Post" WHERE "public"."Post"."id" = $1
prisma:query SELECT "public"."Post"."id", "public"."Post"."words" FROM "public"."Post" WHERE "public"."Post"."id" = $1 LIMIT $2 OFFSET $3
prisma:query COMMIT
Successfully created post #2 with words 2021-09-14T00:00:00.000Z
prisma:query BEGIN
prisma:query SELECT "public"."Post"."id" FROM "public"."Post" WHERE "public"."Post"."id" = $1 OFFSET $2
prisma:query SELECT "public"."Post"."id" FROM "public"."Post" WHERE "public"."Post"."id" = $1
prisma:query SELECT "public"."Post"."id", "public"."Post"."words" FROM "public"."Post" WHERE "public"."Post"."id" = $1 LIMIT $2 OFFSET $3
prisma:query COMMIT
Successfully created post #3 with words 2021-09-14T00:00:00.000Z,2021-09-20T12:00:00.000Z

An error occured while running the seed command:
PrismaClientValidationError: 
Invalid `prisma.post.upsert()` invocation:

{
[...]

PR #15210, opened 2022-09-06
test(client): add reproduction for gh-9248

aqrln:
Add reproduction for validation logic bug with lists containing strings
that look like potential dates or UUIDs.

Ref: https://github.com/prisma/prisma/issues/9248


aqrln:
Merged as a part of https://github.com/prisma/prisma/pull/15286

Issue #12003, opened 2022-02-23
orderBy using self-referential relation in a nested relation fails with "table name \"TableName\" specified more than once"


adaddeo:
### Bug description

A `findMany` query that attempts to orderBy a self-referntial relation nested inside another relation in a ConnectorError with message:
```
table name \"TableName\" specified more than once
```

Here's an example generated query:
```
'SELECT "public"."FulfillmentOrder"."orderId", "public"."FulfillmentOrder"."preSaleId", "public"."FulfillmentOrder"."preSaleOrderId", "public"."FulfillmentOrder"."distributorFee", "public"."FulfillmentOrder"."platformFee", "public"."FulfillmentOrder"."platformFeeOverride", "public"."FulfillmentOrder"."missingPlatformFeeConfiguration", "public"."FulfillmentOrder"."createdAt", "public"."FulfillmentOrder"."updatedAt" FROM "public"."FulfillmentOrder" LEFT JOIN "public"."Order" AS "orderby_0_Order" ON ("public"."FulfillmentOrder"."orderId" = "orderby_0_Order"."id") LEFT JOIN "public"."Order" AS "orderby_0_Order" ON ("orderby_0_Order"."producerPurchaseOrderId" = "orderby_0_Order"."id") WHERE ("public"."FulfillmentOrder"."orderId") IN (SELECT "t0"."orderId" FROM "public"."FulfillmentOrder" AS "t0" INNER JOIN "public"."Order" AS "j0" ON ("j0"."id") = ("t0"."orderId") WHERE ("j0"."type" = $1 AND "j0"."cancelledAt" IS NULL AND "j0"."deliveredAt" IS NULL AND "j0"."placedOn" >= $2 AND ("j0"."id") IN (SELECT "t1"."id" FROM "public"."Order" AS "t1" INNER JOIN "public"."Order" AS "j1" ON ("j1"."id") = ("t1"."producerPurchaseOrderId") WHERE (("j1"."id") IN (SELECT "t2"."id" FROM "public"."Order" AS "t2" INNER JOIN "public"."TrackingInformation" AS "j2" ON ("j2"."orderId") = ("t2"."id") WHERE ("j2"."deliveredAt" IS NOT NULL AND "t2"."id" IS NOT NULL)) AND "t1"."id" IS NOT NULL)) AND "t0"."orderId" IS NOT NULL)) ORDER BY "orderby_0_Order"."placedOn" ASC LIMIT $3 OFFSET $4',
```

[...]

PR #15215, opened 2022-09-06
test(client): order by self #12003

millsp:


packages/client/tests/functional/_utils/idForProvider.ts
```diff
@@ -25,3 +25,13 @@ export function idForProvider(provider: string, options: Options = { includeDefa
 
   return strs.join(' ')
 }
+
+export function idForProviderType(provider: string) {
```
danstarns, line 29: Can we use this function in the `idForProvider` function in the same file?

danstarns (approved):

Issue #13097, opened 2022-04-29
groupBy crashes when using enums


zomars:
### Bug description

Using `groupBy` with a enum array causes an infinite loop while developing a Next.js app:

```
prisma:query SELECT "public"."App"."categories" FROM "public"."App" WHERE 1=1 GROUP BY "public"."App"."categories" OFFSET $1
error - Error:
Invalid `prisma.app.groupBy()` invocation:

 Inconsistent column data: Could not convert value ["payment"] to type `Enum("AppCategories")`.
```

### How to reproduce

```js
const result = await prisma.app.groupBy({
    by: ["categories"],
});
```

### Expected behavior

To no crash

### Prisma information

Use this schema:

```
enum AppCategories {
  calendar
  messaging
  payment
  other
}

model App {
  slug       String          @id @unique
  categories AppCategories[]
  createdAt  DateTime        @default(now())
  updatedAt  DateTime        @updatedAt
}
```

### Environment & setup

- OS: mac os
- Database: PostgreSQL
- Node.js version: v14.17.6

### Prisma Version

```
prisma                  : 3.13.0
@prisma/client          : 3.13.0
Current platform        : darwin
Query Engine (Node-API) : libquery-engine efdf9b1183dddfd4258cd181a72125755215ab7b (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli efdf9b1183dddfd4258cd181a72125755215ab7b (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core efdf9b1183dddfd4258cd181a72125755215ab7b (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary           : prisma-fmt efdf9b1183dddfd4258cd181a72125755215ab7b (at node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash    : efdf9b1183dddfd4258cd181a72125755215ab7b
Studio                  : 0.459.0
```

zomars:
Can confirm that this is still happening on `3.14.0`

millsp:
Internal conversation https://prisma-company.slack.com/archives/C040N0UADF0/p1662503764623529

millsp:
Added test at https://github.com/prisma/prisma/pull/15217

PR #15217, opened 2022-09-06
test(client): group by enum #13097

millsp:


packages/client/tests/functional/issues/13097-group-by-enum/tests.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Regression test for #13097
+ * groupBy on enum fields
+ */
+testMatrix.setupTestSuite(
+  ({ provider }) => {
+    beforeAll(async () => {
+      if (provider !== 'mysql') {
+        await prisma.resource.create({
```
danstarns, line 15: Wonder if it's worth adding 2 or 3 resource's here, here we only test against one resource. wdyt?

danstarns (approved):

packages/client/tests/functional/issues/13097-group-by-enum/tests.ts
millsp: I think we're good since the bug originated from the query plan.

PR #15253, opened 2022-09-09
test(client): Add property-based test for decimal precision

SevInf:
If you are not familiar with property-based testing, [fast-check docs](https://github.com/dubzzz/fast-check/blob/main/packages/fast-check/documentation/HandsOnPropertyBased.md) contain a nice intro.

Basic idea is: instead of coming up with hardcoded examples of input and
expected output we generate large amount of random inputs and assert,
that certain properties of the output are always stay true, regardless
of the input.

This a good match for testing decimal precision issues: using fast-check
library, we generate a large amount of random decimals and assert that
after writing them we always get back the same value without precision
loss.

Another change happening in this PR: it is now possible to exclude
certain combinations from test matrix (see docs). This was necessary
because different providers have different limits for precision and
scale of decimals.

Outcome of the test: all decimal precision issues are fixed. Previously
failing test was me misunderstanding default precision, not something
actually broken.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -268,6 +268,39 @@ Will generate following test suites:
 - `{ provider: 'postgresql', providerFeatures: '' }`
 - `{ provider: 'postgresql', providerFeatures: 'improvedQueryRaw' }`
 
+You can also optionally exclude certain combinations from matrix by using second argument of `defineMatrix` function:
+
+```ts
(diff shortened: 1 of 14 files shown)


aqrln (approved):
💯

millsp (approved):

PR #15268, opened 2022-09-11
fix(client): nixos aarch is also linux-nixos

clansty:


aqrln (approved):
I believe this should be fine for NixOS specifically since we don't provide the pre-compiled binaries the same way as for the officially supported platforms. Otherwise we'd need to introduce a new platform here, something like `linux-nixos-arm64`.

(The reason I'm writing it here is for future reference, because we still have an issue with musl on aarch64, and the fix for it will not be similar).

aqrln:
@Clansty thank you for your contribution! Could you please sign the CLA so we could merge your changes? (https://github.com/prisma/prisma/pull/15268#issuecomment-1242943034)

Clansty:
CLA signed.

And NixOS will fetch the correct binary or build locally, so `linux-nixos` is enough because it always points the current platform.

Another problem is that in NixOS package building environment there's no `/etc/os-release` so the release detection not work. And when I build package using prisma and invoke `prisma generate`, it cannot detect platform correctly. I resolve it by patching the script and let it return `linux-NixOS` when in `nix-build` environment. I think it's better to add an environment variable to specify the platform or use a better way to detect NixOS.

PR #15281, opened 2022-09-12
feat(client): preview feature deno

millsp:
Integration version of https://github.com/prisma/prisma/pull/14386


janpio:
(Sorry, I tried to resolve the conflicts and had messed up - so had to force push the previous state again. No changes.)

packages/cli/src/Generate.ts
```diff
@@ -246,8 +255,18 @@ ${chalk.dim('```')}
...
+        } else {
+          hint = `You can now start using Prisma Client in your code. Reference: ${link('https://pris.ly/d/client')}
+${chalk.dim('```')}
+${highlightTS(`\
+import { PrismaClient } from '${importPath}/deno/edge.ts'`)}
+${chalk.dim('```')}
+
+You will need a Prisma Data Proxy connection string. See documentation: ${link('https://pris.ly/d/data-proxy')}
+${breakingChangesStr}${versionsWarning}`
+        }
```
janpio, line 269: This feels a bit meh that deno is the `else` clause. Can we not react to `isDeno` here as well?

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -66,6 +67,7 @@ export class TSClient implements Generatable {
       runtimeName,
       datasources,
       dataProxy,
+      deno,
```
SevInf, line 70: Question: how does `--deno` and `--data-proxy` interact with each other? Does `--deno` imply `--data-proxy`? Or are they incompatible? If latter, will we have an error if they are both specified?

SevInf (approved):
LGTM with a small question

packages/client/src/generation/TSClient/TSClient.ts
janpio: `--deno` does not exist, only `deno` in the list of preview features in the schema.

SevInf: Got it, thanks

millsp: The Deno CLI is detected, on top of that. So 3 requirements at the moment

PR #15286, opened 2022-09-12
fix(client): string/uuid/date mixed inputs #9248

millsp:
With this PR, we apply our type input checks to input lists in a more generic way.

Fixes https://github.com/prisma/prisma/issues/9248

PR #15315, opened 2022-09-14
test: remove obsolete comments

aqrln:
Follow up to https://github.com/prisma/prisma/pull/15286: remove TODO comments since the bug is now fixed.


millsp (approved):
Nice catch. Thank you

PR #15323, opened 2022-09-14
fix(client): transaction for stored queries #11740

### LINKED ISSUE #11740: Failed transactions trigger multipleResolves
### Bug description

We have prisma transactions running sequentially which fail and the try-catch logs the error.

```
try {
  await prisma.$transaction([multipleSequential]));
 } catch (error) {
 }
```

But as we listen to the event multipleResolves on node:
`process.on('multipleResolves', (type, promise, reason) => {
  console.error(type, promise, reason);
  setImmediate(() => process.exit(1));
});`

We also log for each transaction:
```
Multiple resolves. Type: reject Promise: Promise {
<rejected> PrismaClientKnownRequestError: 
Invalid `prisma.prismaNode.createMany()` invocation:
Unique constraint failed on the fields:
```

Is this a bug or why these are multiple times rejected?

### How to reproduce

Listen to https://nodejs.org/api/process.html#event-multipleresolves when transactions fail.

### Expected behavior

Expect no logging on multipleResolves but only the log inside try-catch for the transaction.

### Environment & setup

- OS: Fedora
- Database: PostgreSQL
- Node.js version: v16.13.2

### Prisma Version

also 3.9.2

millsp:
Fixes my comment in #11740 and validates that the OP is not reproducible anymore.

### Problem

It initially started as an investigation of the issue just above then discovered that I could not reproduce the issue. I found something worse instead. The process was exiting silently and without leaving any error or clue that it is Prisma causing it, we would otherwise (maybe) have had more error reports on this. The problem occurs when you have more than one query that is the same PrismaPromise instance.

After getting a reproduction to work, the problem boils down to PrismaPromises locking indefinitely. The internal mechanism causing this is rather odd to be explained. In short, we have a way to support middlewares with batched transactions that requires to “lock” any execution of the batch (ie. request emission) until all middlewares have finished. While this is reproducible even without middlewares, it is the mechanism causing the locking. Because we cache PrismaPromises so they cannot be executed twice (just like real promises), they don’t signal the “unlock” properly, and causes an infinite wait on that lock, which causes the Node.js process to exit silently.

### Solution

Don’t cache the PrismaPromise for batched transactions (`.requestTransaction`), allowing all the locks to be released.


packages/client/tests/functional/transaction-stored-query/tests.ts
```diff
@@ -0,0 +1,50 @@
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+/**
+ * Regression test for #11740 & comment
```
aqrln, line 8: Should it be in `tests/functional/issues` together with other regression tests?

packages/client/tests/functional/transaction-stored-query/tests.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const query = prisma.resource.create({ data: { email: 'john@prisma.io' } })
+
+    const result = prisma.$transaction([query, query])
+
+    await expect(result).rejects.toMatchPrismaErrorSnapshot()
+
+    expect(mockMultipleResolve).toHaveBeenCalledTimes(0)
+  })
+})
```
aqrln, line 50:
```suggestion
}, {
  skipDataProxy: {
    runtimes: ['edge'],
    reason: 'Skipped because of the error snapshots on edge',
  }
})
```

packages/client/tests/functional/transaction-stored-query/tests.ts
millsp: I didn't know regressions should go in there, but I can move them there, thanks for noticing. Did all of us actually put the reproductions in there during the correctness initiative?

aqrln: That's a good question, not sure about that. I personally don't have a strong opinion about it, but I thought everyone else was putting them there 😅

I think @danstarns added this folder even before the correctness initiative, if I recall correctly.

millsp: Moved into that folder, I think keeping the folder name is useful for browsing but I prepended the issue number.

packages/client/tests/functional/issues/11740-transaction-stored-query/tests.ts
```diff
@@ -0,0 +1,58 @@
...
+    testIf(process.env.PRISMA_CLIENT_ENGINE_TYPE !== 'binary')('no multiple resolves should happen', async () => {
+      const mockMultipleResolve = jest.fn()
+
+      process.on('multipleResolves', mockMultipleResolve)
+
+      const query = prisma.resource.create({ data: { email: 'john@prisma.io' } })
+
+      const result = prisma.$transaction([query, query])
+
+      await expect(result).rejects.toMatchPrismaErrorSnapshot()
```
SevInf, line 47: Do we need a snapshot here? Or just checking that promise rejects would be enough, since it is not main point of the test.

SevInf (approved):
LGTM, with few tiny comments

PR #15326, opened 2022-09-14
chore: fix the local engine install

millsp:
Using engine for local dev might not be copied correctly because two scripts are racing against each other.

### DIFF
--- a/packages/engines/scripts/postinstall.js
+++ b/packages/engines/scripts/postinstall.js
@@ -12,9 +12,13 @@ try {
 
     execa.sync('node', ['-r', 'esbuild-register', buildScriptPath], {
       env: { DEV: true },
+      stdio: 'inherit',
     })
 
-    require(localInstallScriptPath) // may install engines overrides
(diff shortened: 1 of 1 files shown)


packages/engines/scripts/postinstall.js
```diff
@@ -12,9 +12,13 @@ try {
...
     execa.sync('node', ['-r', 'esbuild-register', buildScriptPath], {
       env: { DEV: true },
+      stdio: 'inherit',
     })
 
-    require(localInstallScriptPath) // may install engines overrides
+    // if enabled, it will install engine overrides into the cache dir
+    execa.sync('node', [localInstallScriptPath], {
+      stdio: 'inherit',
+    })
```
jkomyno, line 21: Nice catch!

> When using the detached option to start a long-running process, the process will not stay running in the background after the parent exits unless it is provided with a stdio configuration that is not connected to the parent. If the parent's stdio is inherited, the child will remain attached to the controlling terminal.

jkomyno (approved):

Issue #14001, opened 2022-06-24
invalid character in $let in $lookup pipeline when using cursor and order by query


adamwimmj:
### Bug description

Our database consists of pascal case field names, therefore when using the cursor query options, and order by on a particular field (that starts with a capital letter), prisma builds up an aggregate query and uses the pascal case field name for a $let veriable in the pipeline. According to [https://www.mongodb.com/docs/v4.4/reference/aggregation-variables/](https://www.mongodb.com/docs/v4.4/reference/aggregation-variables/) let variables cannot start with uppercase letters which I think is my problem.

I've created an [example project](https://bitbucket.org/awimmjsystems/prisma/src/master/) that seeds the database with pascal case field names to demonstrate the error.

Here is all the output without running the example project:

When using this schema:

```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mongodb"
  url      = env("DATABASE_URL")
}

model order {
  id      String @id @default(auto()) @map("_id") @db.ObjectId
  OrderBy Int @unique()
}
```
And running this query:

```
await prisma.order.findMany({
    cursor: { OrderBy: 1 },
    orderBy: { OrderBy: "asc" }
   });
```

This error is thrown (debug output):

```
prisma:client  Generated request: +0ms
  prisma:client  query {
  findManyorder(
    cursor: {
      OrderBy: 1
    }
    orderBy: [
      {
        OrderBy: asc
      }
    ]
  ) {
    id
    OrderBy
  }
}
 +0ms
  prisma:client:libraryEngine  sending request, this.libraryStarted: true +13ms
  prisma:client:request_handler  PrismaClientUnknownRequestError: Error occurred during query execution:
ConnectorError(ConnectorError { user_facing_error: None, kind: RawDatabaseError { code: "unknown", message: "Command failed (Location16867): 'OrderBy' starts with an invalid character for a user variable name)" } })
    at prismaGraphQLToJSError (/Users/a/dev/tests/prisma/node_modules/@prisma/client/runtime/index.js:43011:10)
    at LibraryEngine.request (/Users/a/dev/tests/prisma/node_modules/@prisma/client/runtime/index.js:44865:17)
    at async RequestHandler.request (/Users/a/dev/tests/prisma/node_modules/@prisma/client/runtime/index.js:48992:24)
    at async PrismaClient._request (/Users/a/dev/tests/prisma/node_modules/@prisma/client/runtime/index.js:49919:18)
[...]

PR #15327, opened 2022-09-14
test(client): mongo query conflict #14001

millsp:


danstarns:
Related: https://github.com/prisma/prisma/issues/14001

Related: https://github.com/prisma/prisma-engines/pull/3185

packages/client/tests/functional/issues/14001-mongo-order-by-conflict/tests.ts
```diff
@@ -0,0 +1,69 @@
...
+      await expect(result).resolves.toMatchInlineSnapshot(`null`)
+    })
+
+    test('findMany', async () => {
+      const result = prisma.resource.findMany({
+        where: { OrderBy: { gt: 0 } },
+        orderBy: { OrderBy: 'asc' },
+        cursor: { OrderBy: 1 },
+        distinct: ['OrderBy'],
+      })
```
janpio, line 30: The chose nquery here is different than the original reproduction from the user: https://bitbucket.org/awimmjsystems/prisma/src/de00769d30bb07c78f797b2dfe6e1d4adaeef982/index.js#lines-8:11 Intentional or confirmed irrelevant?

millsp: I confirmed that it was indeed failing first with this test, then made sure it does not fail on other levels and across other kinds of queries just to be more thorough.

janpio:
I confirmed manually that this fails with engines version `4.4.0-18.dc7a16204a5f0f2d962929cb297b3f4b3766d689` (which is 1 before the fix got merged afaik).

packages/client/tests/functional/issues/14001-mongo-order-by-conflict/tests.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      await expect(result).resolves.toMatchInlineSnapshot(`Array []`)
```
danstarns, line 32: Wondering if we should be asserting that, (1) the query doesn't crash, (2) that the data returned is correct...

Thinking we should add some dummy data in the beforeAll hook and assert the response of these queries.

danstarns (approved):

packages/client/tests/functional/issues/14001-mongo-order-by-conflict/tests.ts
millsp: Yes, I think having dummy data does not hurt, I'll add.

millsp: Turns out it uncovered something surprising. I will merge PR as-is, and report the crash in a new issue.

millsp: I got confused, this is the one that I tried to add dummy data to and it crashed https://github.com/prisma/prisma/pull/15215 I reported the issue already

PR #15362, opened 2022-09-16
feat(client): Implement isolation levels for batched transactions

SevInf:
Engines part: prisma/prisma-engines#3199

Batch `$transaction` call will now receive optional options
objects with `isolationLevel` property, allowing to pick any level,
supporting by current transaction. Internally, refactors a bunch of
things about the way transaction info is stored in `PrismaPromise`,
`Request` and `InternalRequest` types.

- All transaction-related properties are now grouped under `transaction`
  key. Distiction between iTx and batch transaction is done explcitly
  via discriminated union type.
- `runInTransaction` property is removed. Empty `transaction` field now
  means that request is not running in transaction.

  For testing we are checking that engine correctly generates corresponding SQL.
  Properly testing different isolation level effects would be very
  difficult, especially in case of batch transactions, where we can not
  execute any code between the queries.

Ref #9678


packages/client/package.json
```diff
@@ -147,7 +147,7 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "4.4.0-12.ef027724572e12df6f7ce8c487bd5579b008cd71"
+    "@prisma/engines-version": "4.4.0-28.feat-batch-isolation-level-6a098eca831b9d3cf008d71133236e5a7782d523"
```
SevInf, line 150: Note: remove before merging

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -30,11 +31,10 @@ export type RequestParams = {
   clientMethod: string
   callsite?: CallSite
   rejectOnNotFound?: RejectOnNotFound
-  runInTransaction?: boolean
```
SevInf, line 33: In all instances it was set, it was set to `Boolean(transactionId)` so this looks redundant to me. I refactored all the usages so they look at `transaction` field instead.

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -149,7 +151,7 @@ export class RequestHandler {
         const result = await engineHook(
           {
             document,
-            runInTransaction,
+            runInTransaction: Boolean(transaction),
```
SevInf, line 154: This is a part of public middleware API, so here `runInTransaction` stays, it just gets derived from `transaction` field.

packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -61,13 +61,27 @@ export function applyModel(client: Client, dmmfModelName: string) {
...
-          const data = { args: userArgs, dataPath: [] } // data and its dataPath for nested results
-          const action = { action: dmmfActionName, model: dmmfModelName } // action name and its related model
-          const method = { clientMethod: `${jsModelName}.${prop}`, jsModelName } // method name for display only
-          const tx = { runInTransaction: !!txId, transactionId: txId, lock } // transaction information
-          const trace = { callsite: callSite } // stack trace
-          const params = { ...data, ...action, ...method, ...tx, ...trace }
+        return createPrismaPromise((transaction, lock) => {
+          const params: InternalRequestParams = {
+            // data and its dataPath for nested results
+            args: userArgs,
```
SevInf, line 67: TS could not show type errors within multiple spreads below. Changed that to single object and errors now correctly show up

SevInf:
Tests are failing due to #15360

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -9,12 +9,14 @@ import {
   EngineEventType,
   getTraceParent,
   getTracingConfig,
+  IsolationLevel,
```
millsp, line 12: This looks unused?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -9,12 +9,14 @@ import {
...
   LibraryEngine,
   Options,
   runInChildSpan,
   SpanOptions,
   TracingConfig,
 } from '@prisma/engine-core'
+import { BatchTransactionOptions } from '@prisma/engine-core/dist/common/Engine'
```
millsp, line 19: Can you export it via an entry point instead?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1016,7 +1017,7 @@ new PrismaClient({
       if (typeof input === 'function' && this._hasPreviewFlag('interactiveTransactions')) {
         callback = () => this._transactionWithCallback({ callback: input, options })
       } else {
-        callback = () => this._transactionWithArray(input)
+        callback = () => this._transactionWithArray(input, options)
```
millsp, line 1020: Small nit, can we make this an object like the call above?

packages/client/src/runtime/RequestHandler.ts
millsp: I don't really know if we made this api public, by looking at the generated types

packages/client/src/runtime/core/model/applyModel.ts
millsp: Nice, thanks

packages/client/src/runtime/core/request/createPrismaPromise.ts
```diff
@@ -26,24 +26,32 @@ export function createPrismaPromise(
 
   return {
     then(onFulfilled, onRejected, txId?: string) {
```
millsp, line 28: I think it might be worth changing the `transactionProxy` to pass a `transaction` object here to stay consistent with `requestTransaction`.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -129,14 +136,21 @@ export class DataProxyEngine extends Engine {
...
   }
 
-  async requestBatch<T>(queries: string[], headers: Record<string, string>, isTransaction = false, attempt = 0) {
+  async requestBatch<T>(
+    queries: string[],
+    headers: Record<string, string>,
+    transaction?: BatchTransactionOptions,
+    attempt = 0,
+  ) {
+    const isTransaction = Boolean(transaction)
```
millsp, line 145: This looks like it can be moved directly into the object

packages/client/src/runtime/getPrismaClient.ts
SevInf: Oops, auto-import from VS Code misfire.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
SevInf: It's also used in a log message below

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -48,29 +48,30 @@ export type HandleErrorParams = {
...
 
 function getRequestInfo(request: Request) {
-  const txId = request.transactionId
-  const inTx = request.runInTransaction
+  const transaction = request.transaction
   const headers = request.headers ?? {}
   const traceparent = getTraceParent({ tracingConfig: request.tracingConfig })
 
-  // if the tx has a number for an id, then it's a regular batch tx
-  const _inTx = typeof txId === 'number' && inTx ? true : undefined
```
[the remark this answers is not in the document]
SevInf: Ah, I now see it's purpose. Lack of failing tests indeed convinced me that is safe to remove. Let me check on `main` if that functionality actually works and get back to this PR.

packages/client/src/runtime/RequestHandler.ts
SevInf: Ah, i see, this is a different type of middleware

packages/client/src/runtime/RequestHandler.ts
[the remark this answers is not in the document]
SevInf: I fixed this in the latest commit.
Rather then restoring `runInTransaction` everywhere, I opted for checking it's value after middlewares have run and unsetting `transction` object on request if it is `false`.  Relevant part [is here](https://github.com/prisma/prisma/pull/15362/files#diff-09784fc0113a735572cff48990efa35434fb6ec6ec8876845ec9cf423642e5fdR1089-R1101).

I still think it is cleaner if the rest of the code can rely only on `transaction` property for determining if request is done in transaction and handling `runInTransaction` this way achieves it.

SevInf:
@millsp I beleive it can now go for additional round of reviews, while I am working on engines

packages/client/src/runtime/core/request/PrismaPromise.ts
```diff
@@ -8,31 +26,34 @@ export interface PrismaPromise<A> extends Promise<A> {
...
-  finally(onfinally?: (() => void) | undefined | null, txId?: string): Promise<A>
+  finally(onfinally?: (() => void) | undefined | null, transaction?: InteractiveTransactionOptions): Promise<A>
 
   /**
    * Called when executing a batch of regular tx
-   * @param txId for regular tx ids
+   * @param transaction transaction options for regular tx
    */
-  requestTransaction?(txId: number, lock?: PromiseLike<void>): PromiseLike<unknown>
+  requestBatchTransaction?(transaction: BatchTransactionOptions, lock?: PromiseLike<void>): PromiseLike<unknown>
```
SevInf, line 58: Took the liberty to rename this method to be more specific about which of the 2 kinds of transactions it requests. It is not a part of public API, but let me know if you still consider this rename dangerous, I'll revert.

millsp: I've been wanting to even remove this or hide it but some people rely on it so we decided not to do so, although it's not documented as public API :hear_no_evil:

SevInf: Ok, reverting the change then

millsp (approved):
Looks good to go! Thanks for adding the missing tests

PR #15415, opened 2022-09-20
test(client): Regression test for gh-15176

### LINKED ISSUE #15176: only one @updatedAt is filled by the QE although multiple are accepted in the schema
We should either allow and document only one use of @updatedAt per model or update all of them if the row is touched instead of only the first one. 

```
// Valid datamodel
model TestModel {
      id                    Int       @id
      test                  Int?
      updatedAt_w_default   DateTime  @default(now()) @updatedAt
      updatedAt_wo_default  DateTime? @updatedAt
      createdAt             DateTime  @default(now())
}
```

```
field.rs
// code only surfacing first updatedAt per model
 pub fn updated_at(&self) -> &Option<ScalarFieldRef> {
        self.updated_at.get_or_init(|| {
            self.scalar_weak()
                .iter()
                .map(|sf| sf.upgrade().unwrap())
                .find(|sf| sf.is_updated_at)
        })
    }

```

danstarns:
Closes: https://github.com/prisma/prisma/issues/15176

Related: https://github.com/prisma/prisma-engines/pull/3168


packages/client/tests/functional/issues/15176/tests.ts
```diff
@@ -0,0 +1,42 @@
...
+    const id = provider === 'mongodb' ? faker.database.mongodbObjectId() : faker.random.alpha(10)
+
+    const created = await prisma.TestModel.create({
+      data: {
+        id,
+      },
+    })
+
+    expect(created.updatedAt_w_default).toBeTruthy()
+    expect(created.updatedAt_wo_default).toBeFalsy()
```
do4gr, line 23: This should be truthy I think

millsp (approved):

packages/client/tests/functional/issues/15176/prisma/_schema.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URI_${provider}")
+    }
+    
+    model TestModel {
+      id                    ${idForProvider(provider)}
+      test                  Boolean?
```
millsp, line 17:
```suggestion
      bool                  Boolean?
```

packages/client/tests/functional/issues/15176/tests.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    expect(getTime(created.updatedAt_w_default)).toEqual(getTime(created.createdAt))
+    expect(getTime(created.updatedAt_wo_default!)).toEqual(getTime(created.createdAt))
+
+    const updated = await prisma.testModel.update({
+      where: {
+        id,
+      },
+      data: {
+        test: false,
```
danstarns, line 35:
```suggestion
        bool: false,
```

Issue #14543, opened 2022-07-29
mongodb cannot use $runCommand Raw in transaction


Lizhooh:
### Bug description

mongodb cannot use $runCommand Raw in transaction

```js
Invalid `prisma.$runCommandRaw()` invocation:


  The current database provider doesn't support a feature that the query used: Raw queries
2022-07-29 09:52:21 error: Error: 
Invalid `prisma.$runCommandRaw()` invocation:


  The current database provider doesn't support a feature that the query used: Raw queries

```

### How to reproduce

demo: 

```js 
await prisma.$transaction(async $prisma => {
   $prisma.$runCommandRaw({
        "update": "component",
        "updates": [
            {
              "q": { "_id": "EDMBuss0V4" },
              "u": { "$set": { "alias": "abc" } },
              "multi": false,
              "upsert": false
            }
        ]
    })
})
```

### Environment & setup

- OS: Windows 
- Database: MongoDB 
- Node.js version: v14.20.0 

### Prisma Version

```
prisma                  : 4.0.0
@prisma/client          : Not found
Current platform        : debian-openssl-1.1.x
Query Engine (Node-API) : libquery-engine da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : da41d2bb3406da22087b849f0e911199ba4fbf11
Studio                  : 0.465.0
```

Jolg42:
Can reproduce in our internal dev version `4.4.0-dev.27`
```
/Users/j42/Repros/14543/node_modules/@prisma/client/runtime/index.js:30850
      throw new PrismaClientKnownRequestError(message, error2.code, this.client._clientVersion, error2.meta);
            ^
PrismaClientKnownRequestError: 
Invalid `prisma.$runCommandRaw()` invocation:


[...]

Issue #13405, opened 2022-05-18
[mongoDB] `findRaw` does not work within an `interactiveTransaction`


paulrostorp:
### Bug description

MongoDB's `findRaw` method works fine when running normally, but fails within an `interactiveTransaction`: 

```
Invalid `prisma.project.findRaw()` invocation:


  The current database provider doesn't support a feature that the query used: Raw queries
```

### How to reproduce

```js
await prisma.$transaction(async txn => {
  const projects = await txn.project.findRaw({
    filter: {
      _id: {
        $in: ["foo", "bar"],
      },
    },
  });
});
```

### Prisma information

```prisma
generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["interactiveTransactions"]
}

datasource db {
  provider = "mongodb"
  url      = env("MONGODB_URL")
}

model Project {
  id            String   @id @map("_id") @db.String
  name          String

  @@map("projects")
}

```

### Environment & setup

- OS: `Mac OS m1 chip`
- Database: `MongoDb`
- Node.js version: `v16.13.2`

### Prisma Version

```
prisma                  : 3.14.0
@prisma/client          : 3.14.0
Current platform        : darwin-arm64
Query Engine (Node-API) : libquery-engine 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt 2b0c12756921c891fec4f68d9444e18c7d5d4a6a (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : 2b0c12756921c891fec4f68d9444e18c7d5d4a6a
Studio                  : 0.460.0
Preview Features        : interactiveTransactions
```

Also tested on `3.13.0`

do4gr:
Related: https://github.com/prisma/prisma/issues/14543

Jolg42:
Can reproduce in our internal dev version 4.4.0-dev.27

PR #15442, opened 2022-09-21
test(client): add regression tests for mongo raw iTX issue

### LINKED ISSUE #15532: Add tests for mongodb's `findRaw`, `aggregateRaw` and `runCommandRaw` in sequential transactions
Creating this issue as a reminder that we are missing functional tests for `findRaw`, `aggregateRaw` and `runCommandRaw` in sequential transactions, we currently have no guarantees that it's working as expected.

+Reminder: Check if this is tested on the engines

aqrln:
Add regression tests for using `findRaw`, `aggregateRaw` and `runCommandRaw` inside interactive transactions with MongoDB.

Ref: https://github.com/prisma/prisma/issues/13405
Ref: https://github.com/prisma/prisma/issues/14543
Ref: https://github.com/prisma/prisma-engines/pull/3206
Closes: https://github.com/prisma/prisma/issues/15532


packages/client/package.json
```diff
@@ -147,7 +147,7 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "4.4.0-33.94146edaafca9703aa6ad050adf1b1fd2dff438e"
+    "@prisma/engines-version": "4.4.0-39.integration-query-raw-itx-mongo-a52d215fb2b3cba53b195bbf56b09fc615ddad3d"
```
janpio, line 150: Reset before merging

millsp (approved):
It looks good! Related question: do you know if we test these on sequential transactions?

millsp:
> It looks good! Related question: do you know if we test these on sequential transactions?

Answer seems to be no, I'll create an issue for it.

aqrln:
> Answer seems to be no, I'll create an issue for it.

I was just going to post an answer to this thread — the answer was no, so I went ahead and added them here, I'll push it in a bit

aqrln:
@millsp done, PTAL

aqrln:
Actually this could've been moved into the the same regression test suite as, thinking about it, sequential transactions were in fact broken too — it's the same piece of logic in the engine where the issue was that is responsible for both. But there's also a benefit in keeping it separate since we get Data Proxy testing for it while the iTX test is skipped.

PR #15457, opened 2022-09-22
fix(client): Don't execute normal batches in transaction

SevInf:
During isolation level works, i accidentally forced normal batches to be executed in transaction as well. This commit fixes it.

### DIFF
--- a/packages/client/src/runtime/RequestHandler.ts
+++ b/packages/client/src/runtime/RequestHandler.ts
@@ -92,9 +92,7 @@ export class RequestHandler {
         // TODO: pass the child information to QE for it to issue links to queries
         // const links = requests.map((r) => trace.getSpanContext(r.otelChildCtx!))
 
-        return this.client._engine.requestBatch(queries, info.headers, {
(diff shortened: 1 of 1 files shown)


millsp (commented):
Would it be possible to add a test case for this via the logs or something else?

SevInf:
>Would it be possible to add a test case for this via the logs or something else?

Logs are really nice idea. I tried to come up with a test case, but could not quite figure out how, will add test in a moment.

SevInf:
@millsp Tried to write a test for this and it didn't work out. And also, probably it wasn't actually broken. Here is what happens:

1. We batch only `findUnique`s on the same model
2. Engine can turn those batched `findUnique` into single SQL query
3. Since it is reduced to a single query, engine never starts transaction, no matter what client sends

I still think it might be valuable to merge this PR: at least it makes it clear in the client code that `transaction` is not happening, without relying on engine magic.

millsp (approved):
Agreed, I'm also surprised. Approving this.

PR #15481, opened 2022-09-23
feat(client): support marking input types as readonly

### LINKED ISSUE #5025: Usage of Readonly types instead of mutable in TS typings
Currently all arrays in input positions use `Array` type and it leads to a lot of issues, such as incompatibility with Facebook `DataLoader` library. DataLoader expect keys to be a ReadonlyArray, while Prisma expects `{where: { id: { in: keys } }}` to accept mutable Array. I think it's safe to change types of arrays in input positions to `ReadonlyArray` as well (i believe Prisma doesn't mutate them under the hood, so it's safe to do that), while keeping types in output positions as `Array` (we shouldn't enforce users to keep result array immutable, so we can't use `ReadonlyArray` here).
If I understand correctly, currently we use only one generic for all input types: `export declare type Enumerable<T> = T | Array<T>;`, so changing it to `export declare type Enumerable<T> = T | ReadonlyArray<T>;
` should be enough.

It would be nice to have immutable types for objects as well, like 
```typescript
export declare type UserCreateInput = Readonly<{
    id?: string | null;
    name: string;
}>;
```
instead of 
```typescript
export declare type UserCreateInput = {
    id?: string | null;
    name: string;
};
```
because of the same reasons as described above. If Prisma doesn't mutate this object, it's better to mark it as immutable, so user will be able to pass both mutable/immutable objects, not only mutable ones like right now

uhyo:
Fixes #5025.

This PR adds a `readonly-input` option which, if set to `true`, marks all objects and arrays in input types as `readonly` in the generated TypeScript client.

I wanted developers in our team to always construct search queries in an immutable way. This feature is necessary for development in that style.
Looking at the discussion in #5025, the mutable way semms also in demand, so I implemented this feature as an option.

Hope this can be accepted. Thank you.

## Example

```prisma
generator client {
  provider       = "prisma-client-js"
  readonly-input = true
}

model User {
  id        String   @id

  profile Profile?
  Post    Post[]
}
```

This will generate, for example:

```ts
  export type UserWhereInput = {
    readonly AND?: ReadonlyEnumerable<UserWhereInput>
    readonly OR?: ReadonlyEnumerable<UserWhereInput>
    readonly NOT?: ReadonlyEnumerable<UserWhereInput>
    readonly id?: StringFilter | string
    readonly createdAt?: DateTimeFilter | Date | string
    readonly profile?: XOR<ProfileRelationFilter, ProfileWhereInput> | null
    readonly Post?: PostListRelationFilter
  }
```

### DIFF
--- a/packages/client/src/__tests__/types/readonly-input/schema.prisma
+++ b/packages/client/src/__tests__/types/readonly-input/schema.prisma
@@ -19,6 +19,8 @@ model User {
 model Profile {
   id String @id
 
+  name String
+
   user   User   @relation(fields: [userId], references: [id])
   userId String @unique
 }
--- a/packages/client/src/__tests__/types/readonly-input/schema.prisma
+++ b/packages/client/src/__tests__/types/readonly-input/schema.prisma
@@ -1,0 +1,33 @@
+generator client {
+  provider       = "prisma-client-js"
+  readonly-input = true
+}
+
+datasource db {
+  provider = "sqlite"
+  url      = env("DATABASE_URL")
+}
+
+model User {
+  id        String   @id
+  createdAt DateTime @default(now())
+
+  profile Profile?
+  Post    Post[]
+}
+
+model Profile {
+  id String @id
+
+  user   User   @relation(fields: [userId], references: [id])
+  userId String @unique
+}
+
+model Post {
+  id        String   @id
+  createdAt DateTime @default(now())
+
+  content String
+  user    User   @relation(fields: [userId], references: [id])
+  userId  String
+}
--- a/packages/client/src/__tests__/types/readonly-input/test.ts
+++ b/packages/client/src/__tests__/types/readonly-input/test.ts
@@ -10,6 +10,13 @@ async function main() {
   // @ts-expect-error where should be readonly
   where.OR = []
 
+  // @ts-expect-error arrays should also be readonly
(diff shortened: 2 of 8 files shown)


uhyo:
@SevInf Sorry for interrupting you, but could you take a look or assign someone who can review please?

janpio:
@SevInf is on vacation. But I can tell you that we can not just accept a PR adding a new datasource property unfortunately without further internal discussion. Make sure your PR description has all the information we need to evaluate the feature and the user case for it - thanks.

hokaccha:
@janpio @uhyo What is the status of this PR? I truly desire this feature.

uhyo:
@janpio @SevInf Hi, as a random reminder I rebased this pull request to the latest main. I hope you could take time to determine whether this is worth a new generator option. Thanks 🙂

FGoessler:
Hi @uhyo 👋 
Thank you very much for this contribution! Unfortunately it currently does not fully align with our direction. Adding additional generator options like this significantly increases the permutations of our API surface area. Hence it complicates testing and long term maintenance.

That being said we have it on our radar to better support readonly input types. In particular to tackle #5025. But likely this will incur somewhat breaking changes so it is something we postpone at least until the next major release.

If you have input regarding this please head over to #5025 and leave a comment there.

Issue #13388, opened 2022-05-16
`Error: The provided database string is invalid. Unable to parse URL. in database URL.` 


jvolker:
### Bug description

I'm trying to follow the getting started guide for [MongoDB](https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/mongodb-node-mongodb). Running the most basic example with very minimal credentials throws the following error:

```
  prisma:tryLoadEnv Environment variables loaded from /Users/username/myRepo/.env +0ms
  prisma:tryLoadEnv Environment variables loaded from /Users/username/myRepo/.env +1ms
  prisma:client dirname /Users/username/myRepo/node_modules/.prisma/client +0ms
  prisma:client relativePath ../../../prisma +0ms
  prisma:client cwd /Users/username/myRepo/prisma +0ms
  prisma:client clientVersion: 3.14.0 +1ms
  prisma:client clientEngineType: library +0ms
  prisma:client:libraryEngine internalSetup +0ms
  prisma:client:libraryEngine Searching for Query Engine Library in /Users/username/myRepo/node_modules/.prisma/client +1ms
  prisma:client:libraryEngine loadEngine using /Users/username/myRepo/node_modules/.prisma/client/libquery_engine-darwin-arm64.dylib.node +0ms
  prisma:client:libraryEngine library starting +3ms
(node:57335) UnhandledPromiseRejectionWarning: Error: The provided database string is invalid. Unable to parse URL. in database URL. Please refer to the documentation in https://www.prisma.io/docs/reference/database-reference/connection-urls for constructing a correct connection string. In some cases, certain characters must be escaped. Please check the string for any illegal characters.
    at /Users/username/myRepo/node_modules/@prisma/client/runtime/index.js:41273:21
(Use `node --trace-warnings ...` to show where the warning was created)
(node:57335) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:57335) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

[...]

Issue #11883, opened 2022-02-17
`The provided database string is invalid. Unable to parse URL. in database URL.` on invalid (?) connection string


janpio:
Something in this connection string seems to be weird, but the error message also does not make sense:

Connection string:
```
mongodb://localhost:C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==@localhost:10255/e2e-tests?ssl=true
```

Error message:
```
D:\a\e2e-tests\e2e-tests\databases-windows\mongodb-azure-cosmosdb-emulator\script.ts:7:21

   4 
   5 // A `main` function so that we can use async/await
   6 async function main() {
→  7   await client.user.deleteMany(
  The provided database string is invalid. Unable to parse URL. in database URL. Please refer to the documentation in [prisma.io/docs/reference/database-reference/connection-urls](https://www.prisma.io/docs/reference/database-reference/connection-urls) for constructing a correct connection string. In some cases, certain characters must be escaped. Please check the string for any illegal characters.
    at cb (D:\a\e2e-tests\e2e-tests\databases-windows\mongodb-azure-cosmosdb-emulator\node_modules\@prisma\client\runtime\index.js:39809:17) {
  clientVersion: '3.10.0-dev.54',
  errorCode: undefined
}
```

janpio:
Same grammar is present in other error messages around connection string: https://github.com/prisma/prisma/issues/11684

matthewmueller:
Looks like the slashes in the password are the culprit. I'm tempted to close this since it's not a valid URL and the error message says that.

Looks like it's base64 encoded which can produce invalid URLs: https://en.wikipedia.org/wiki/Base64#The_URL_applications

Where did you get this URL?

janpio:
It is the default connection string of Azure CosmosDB Emulator MongoDB API. It also needs to be escaped in the Mongo Node Client, so us needing that is fine I guess.

But that means we definitely need to update this message here which the error message links to: https://www.prisma.io/docs/reference/database-reference/connection-urls#special-characters

Also there is something weird going on in the error message: `Unable to parse URL. in database URL. `. That sentence looks broken to me.

janpio:
https://github.com/prisma/docs/issues/3002

stevemu:
for me, I had to define the "authentication database" to make it work, so:

this didn't work (in .env):
```
[...]

PR #15486, opened 2022-09-23
test(client): add test for MongoDB connection string error messages

### LINKED ISSUE #15548: Binary engine: `$disconnect` never returns if engine failed to start
See https://github.com/prisma/prisma/pull/15486#issuecomment-1258564198

Calling `$disconnect` with the binary engine when the engine failed to start causes the client to hang waiting for an event that never happens.

These lines:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L832-L834
create a promise that is supposed to be resolved inside the `this.child.on('exit', ...)` event handler:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L668
after killing the child process a line after:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L836
and then wait until that promise resolves:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L839-L841

However, that event never happens and the promise is never resolved, because the child has long exited before that method was called (maybe even before the exit even handler was set).

aqrln:
TypeScript tests for https://github.com/prisma/prisma-engines/pull/3216.

The tests should fail now, they will pass once the engines are updated to [`bf2452642b8ba07806c16e139c239b99b39ff7e2`](https://github.com/prisma/prisma-engines/commit/bf2452642b8ba07806c16e139c239b99b39ff7e2) or later and this PR is rebased.

Ref: https://github.com/prisma/prisma/issues/13388
Ref: https://github.com/prisma/prisma/issues/11883
Fixes: https://github.com/prisma/prisma/issues/15548

### DIFF
--- a/packages/client/tests/functional/issues/11883-13388-mongodb-url-errors/_matrix.ts
+++ b/packages/client/tests/functional/issues/11883-13388-mongodb-url-errors/_matrix.ts
@@ -1,0 +1,9 @@
+import { defineMatrix } from '../../_utils/defineMatrix'
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'mongodb',
+    },
+  ],
+])
--- a/packages/client/tests/functional/issues/11883-13388-mongodb-url-errors/prisma/_schema.ts
+++ b/packages/client/tests/functional/issues/11883-13388-mongodb-url-errors/prisma/_schema.ts
@@ -1,0 +1,19 @@
+import { idForProvider } from '../../../_utils/idForProvider'
+import testMatrix from '../_matrix'
+
+export default testMatrix.setupSchema(({ provider }) => {
+  return /* Prisma */ `
+    generator client {
+      provider = "prisma-client-js"
+    }
+
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URI_${provider}")
+    }
+
+    model Resource {
+      id      ${idForProvider(provider)}
+    }
+  `
+})
--- a/packages/client/tests/functional/issues/11883-13388-mongodb-url-errors/tests.ts
+++ b/packages/client/tests/functional/issues/11883-13388-mongodb-url-errors/tests.ts
@@ -44,5 +44,18 @@ testMatrix.setupTestSuite(
       from: ['cockroachdb', 'mysql', 'postgresql', 'sqlite', 'sqlserver'],
       reason: 'Test for MongoDB-specific errors',
     },
+    skipDataProxy: {
+      runtimes: ['node', 'edge'],
+      reason: `
+        The test is not relevant for the Data Proxy. It is not possible to
+        create a project with invalid connection string via PDP UI. If an
+        invalid connection string somehow ended up saved for the project due to
+        a bug in PDP, the behavior is unspecified. We could adapt this to be a
+        contrived test tailor-made for mini-proxy (by importing mini-proxy API
+        and programmatically generating mini-proxy connection strings pointing
+        at broken URLs), but it would not reflect real use cases and would not
+        bring much value.
+      `,
+    },
   },
 )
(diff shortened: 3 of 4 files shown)


janpio:
Right now normal tests fail with 1 timeout, data proxy with this:
```
FAIL issues/11883-13388-mongodb-url-errors/tests.ts
  ● issues.11883-13388-mongodb-url-errors (provider=mongodb) › unescaped slashes in password, causes the rest to be interpreted as database name

    InvalidDatasourceError: Could not parse URL of the datasource

      26650 |       url = new URL(dataProxyURL);
      26651 |     } catch (e) {
    > 26652 |       throw new InvalidDatasourceError("Could not parse URL of the datasource", {
            |             ^
      26653 |         clientVersion: this.clientVersion
      26654 |       });
      26655 |     }

      at DataProxyEngine.extractHostAndApiKey (../../runtime/index.js:26652:13)
      at new extractHostAndApiKey (../../runtime/index.js:26508:33)
      at PrismaClient.getEngine (../../runtime/index.js:31391:16)
      at new getEngine (../../runtime/index.js:31365:29)
      at globalThis.newPrismaClient (_utils/setupTestSuiteMatrix.ts:81:26)
      at Object.newPrismaClient (issues/11883-13388-mongodb-url-errors/tests.ts:12:22)

  ● issues.11883-13388-mongodb-url-errors (provider=mongodb) › mongodb+srv used together with a port

    InvalidDatasourceError: Datasource URL must use prisma:// protocol when --data-proxy is used

      26656 |     const { protocol, host, searchParams } = url;
      26657 |     if (protocol !== "prisma:") {
    > 26658 |       throw new InvalidDatasourceError("Datasource URL must use prisma:// protocol when --data-proxy is used", {
            |             ^
      26659 |         clientVersion: this.clientVersion
      26660 |       });
      26661 |     }

      at DataProxyEngine.extractHostAndApiKey (../../runtime/index.js:26658:13)
      at new extractHostAndApiKey (../../runtime/index.js:26508:33)
      at PrismaClient.getEngine (../../runtime/index.js:31391:16)
      at new getEngine (../../runtime/index.js:31365:29)
      at globalThis.newPrismaClient (_utils/setupTestSuiteMatrix.ts:81:26)
      at Object.newPrismaClient (issues/11883-13388-mongodb-url-errors/tests.ts:27:22)
```

millsp (approved):

millsp:
It looks like there is a timeout on one of the tests for the binary engine only.

aqrln:
Wow, it looks like we have found a bug in the `BinaryEngine`. Strange that it has never been caught by any tests before.
Apparently, calling `$disconnect` with the binary engine when the engine failed to start causes the client to hang waiting for an event that never happens.

These lines:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L832-L834
create a promise that is supposed to be resolved inside the `this.child.on('exit', ...)` event handler:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L668
after killing the child process a line after:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L836
and then wait until that promise resolves:
https://github.com/prisma/prisma/blob/f9d4ae0654887554eb212c26171390a3cfad9efb/packages/engine-core/src/binary/BinaryEngine.ts#L839-L841

However, that even never happens and the promise is never resolved, because the child has long exited before that method was called (maybe even more than that — before the exit even handler is set — I didn't check if that's the case because it doesn't really matter).

I pushed a commit that fixes the issue for me locally, let's see what happens on CI and if other tests are happy with the change.

PR #15520, opened 2022-09-26
test(client): Remove flaky test

danstarns:
This test is not really needed, seeming we are asserting that a runtime error is thrown when using the contains filter. 

It's flaky and is failing: 

here: https://github.com/prisma/prisma/actions/runs/3126991100/jobs/5073123688

here: https://github.com/prisma/prisma/actions/runs/3126253148/jobs/5073021273

and other places...

### DIFF
--- a/packages/client/tests/functional/issues/9007/tests.ts
+++ b/packages/client/tests/functional/issues/9007/tests.ts
@@ -1,16 +1,12 @@
-import { expectTypeOf } from 'expect-type'
-import fs from 'fs'
-import path from 'path'
-
 import testMatrix from './_matrix'
 // @ts-ignore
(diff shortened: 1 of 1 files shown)


packages/client/tests/functional/issues/9007/tests.ts
```diff
@@ -21,10 +19,6 @@ testMatrix.setupTestSuite(
         }),
       ).rejects.toThrowError()
     })
-
-    test('should not generate the contains field on the where type', () => {
-      expectTypeOf<Prisma.UuidFilter>().not.toHaveProperty('contains')
-    })
```
janpio, line 27: How can this be flaky?

PR #15524, opened 2022-09-26
chore: fix pipeline

millsp:


packages/client/tests/functional/issues/9007/tests.ts
```diff
@@ -24,17 +22,7 @@ testMatrix.setupTestSuite(
       ).rejects.toThrowError()
     })
 
-    test('should not generate the contains field on the where type', async () => {
-      const generatedTypeScriptSrc = await fs.promises.readFile(
```
millsp, line 28: This part of the test was added before. What it does:
It checks that a generated type does not have a `contains` property.

packages/client/tests/functional/issues/9007/tests.ts
```diff
@@ -24,17 +22,7 @@ testMatrix.setupTestSuite(
...
-        'utf-8',
-      )
-
-      const [uuidFilter] = generatedTypeScriptSrc.split('export type UuidFilter = {')[1].split('}')
-
-      const hasContains = uuidFilter.includes('contains')
-
-      expect(hasContains).toEqual(false)
+    test('should not generate the contains field on the where type', () => {
       expectTypeOf<Prisma.UuidFilter>().not.toHaveProperty('contains')
```
millsp, line 26: This check does exactly the same but on the type-level, so not reading the file, but using the type system. That's why I am removing the code above.

.buildkite/publish/docker-compose.yml
```diff
@@ -126,7 +126,7 @@ services:
       - '3307:3306'
 
   mariadb:
-    image: mariadb:10.9.3
+    image: mariadb:10.7.3
```
millsp, line 129: MariaDB revert. I will be work looking at it soon to understand why it failed on publish, but not in tests.

Jolg42: Just checked and the update only bumped the publish pipeline version and not the test pipeline version 🤷🏼 
https://github.com/prisma/prisma/pull/15425/files

Jolg42 (approved):
thanks!

Issue #14954, opened 2022-08-23
Calling `findUnique` concurrently with a `DateTime` column causes it to return `null`


stayradiated:
### Bug description

Seems to be the same issue as in https://github.com/prisma/prisma/issues/4438 and https://github.com/prisma/prisma/issues/5941.

There is an issue with Prisma which causes `findUnique()` to return null values, even though the row definitely exists.

- Only happens with `findUnique`
- Only when multiple `findUnique` calls are made concurrently and Prisma runs them in a batched query.
- Only seems to happen when a `DateTime` column is used.

### How to reproduce

I have created a standalone repository with reproduction code available at: https://github.com/stayradiated/prisma-issue-findUnique-returns-null

### Expected behavior

The `findUnique` method should have the same behaviour whether run by itself or concurrently.

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

model ModelWithUniqueDate {
  id   Int      @id @default(autoincrement())
  date DateTime @unique @db.Date
}
```

```javascript
const prisma = new PrismaClient();

await prisma.modelWithUniqueDate.createMany({
  data: [
    { date: "2011-01-01T00:00:00Z" },
    { date: "2022-02-02T00:00:00Z" }
  ],
});

const rowList = await Promise.all([
  prisma.modelWithUniqueDate.findUnique({
    where: { date: "2011-01-01T00:00:00Z" },
  }),
  prisma.modelWithUniqueDate.findUnique({
    where: { date: "2022-02-02T00:00:00Z" },
  })
]);

console.log(rowList) // prints [null, null]
```

### Environment & setup

- OS: Ubuntu (also reproduced on macOS)
- Database: PostgreSQL
- Node.js version: v16.17.0

### Prisma Version

```
Environment variables loaded from .env
prisma                  : 4.2.1
@prisma/client          : 4.2.1
Current platform        : debian-openssl-1.1.x
Query Engine (Node-API) : libquery-engine 2920a97877e12e055c1333079b8d19cee7f33826 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 2920a97877e12e055c1333079b8d19cee7f33826 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
[...]

Issue #5952, opened 2021-03-03
N+1 Issue with `Decimal` data type and combining queries (batching)


mikeburgh:
## Bug description

When using Decimal for the data type, Prisma correctly combines the queries, however they are not returned correctly (un combined?) and graphQL throws an error (if set to notNull) or displays no data

## How to reproduce

The following repo has a reproducible example setup as a bare minimum with GraphQL and SQLite DB. 
Also provides notes on changing the schema to Int and re running to show it working correctly.

Clone https://github.com/mikeburgh/prisma-nexus and follow instructions in readme

## Expected behavior

Correct results to be returned once queries are combined

## Prisma information

```graphql
generator client {
  provider        = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
  url      = "file:./dev.db"
}

model Parent {
  id      Decimal @id 
  childId Decimal 
  child   Child   @relation(fields: [childId], references: [id])
}

model Child {
  id     Decimal  @id 
  Parent Parent[]
}
```

See repo above for full details

## Environment & setup

- OS: Ubuntu
- Database: SQLite and Sql Server
- Node.js version: v14.15.4
- Prisma version:

```
prisma               : 2.18.0
@prisma/client       : 2.18.0
Current platform     : debian-openssl-1.1.x
Query Engine         : query-engine da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x)
Migration Engine     : migration-engine-cli da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine : introspection-core da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary        : prisma-fmt da6fafb57b24e0b61ca20960c64e2d41f9e8cff1 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Studio               : 0.356.0
```

tarazena:
[...]

PR #15526, opened 2022-09-26
test(client): add regression tests for #14954 #5952

millsp:


packages/client/tests/functional/issues/14954-date-batch/tests.ts
```diff
@@ -0,0 +1,112 @@
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(() => {
+  beforeAll(async () => {
+    await prisma.resource.create({
+      data: { date: '2011-01-01T00:00:00Z' },
```
danstarns, line 10:
```suggestion
      data: { date: date1 },
```

packages/client/tests/functional/issues/14954-date-batch/tests.ts
```diff
@@ -0,0 +1,112 @@
...
+    })
+    await prisma.resource.create({
+      data: { date: '2022-02-02T00:00:00Z' },
```
danstarns, line 13:
```suggestion
      data: { date: date2 },
```

packages/client/tests/functional/issues/14954-date-batch/tests.ts
```diff
@@ -0,0 +1,112 @@
...
+      Array [
+        Object {
+          date: 2011-01-01T00:00:00.000Z,
+        },
+        Object {
+          date: 2022-02-02T00:00:00.000Z,
+        },
+      ]
+    `)
+  })
```
danstarns, line 39:
```suggestion
    expect(result).toMatchObject([{ date: date1 }, { date: date2 }]);
  })
```

packages/client/tests/functional/issues/14954-date-batch/tests.ts
```diff
@@ -0,0 +1,112 @@
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(() => {
```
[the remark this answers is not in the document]
danstarns: I would even suggest that these dates are always generated so we know we are testing a random date each time: 

```js
const date1 = new Date().toString()
const date2 = new Date().toString();
```

danstarns (approved):
Looks good to me, **my nonblocking suggestions** are about defining the test variables at the top and then reusing them throughout the suite.

packages/client/tests/functional/issues/14954-date-batch/tests.ts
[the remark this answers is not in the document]
millsp: const date1 = new Date().toISOString()

I remember I could not use that one, because that actually worked. The queries only failed when the date had the `.000Z` omitted.

PR #15527, opened 2022-09-26
refactor(client): Remove unused date arg

danstarns:
Removes some unused code in this PR: https://github.com/prisma/prisma/pull/15415


packages/client/tests/functional/issues/15176/tests.ts
```diff
@@ -8,9 +8,7 @@ declare let prisma: PrismaClient
 
 // https://github.com/prisma/prisma/issues/15176
 testMatrix.setupTestSuite(({ provider }) => {
-  const getTime = (dt: string | Date): number => {
-    return dt instanceof Date ? dt.getTime() : new Date(dt).getTime()
-  }
+  const getTime = (dt: Date): number => dt.getTime()
```
aqrln, line 11: I think at this point we could remove this function altogether and just call the method directly

aqrln (approved):

packages/client/tests/functional/issues/15176/tests.ts
danstarns: I thought this but then found it to be ugly and brittle calling .getTime() everywhere.

Issue #13913, opened 2022-06-20
Postgres integer column silently overflows to bad value


jajarzombek:
### Bug description

On inserting record into table with integer column a value that overflows postgres max integer value of 2147483647 no error message is returned and record with bad value is silently stored. 
E.g. `2265000000` is saved as `-2029967296`

### How to reproduce

1. Create table with int column
2. Insert record via Prisma with value that exceeds max postgres integer
3. Wrong value in table is stored

### Expected behavior

An error should be thrown

### Prisma information

```
generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["improvedQueryRaw"]
}

datasource db {
  provider = "postgresql"
}
```

### Prisma Version

```
3.15
```

pimeys:
There was some prior work to fix this here: https://github.com/prisma/prisma/issues/10586

But, the default postgres type for ints is integer, and that's a 32-bit value. We should improve the validation here.

PR #15545, opened 2022-09-26
test(client): int overflow #13913

millsp:
Regression test for https://github.com/prisma/prisma/issues/13913

### DIFF
--- a/packages/client/tests/functional/issues/13913-integer-overflow/tests.ts
+++ b/packages/client/tests/functional/issues/13913-integer-overflow/tests.ts
@@ -12,7 +12,13 @@ testMatrix.setupTestSuite(() => {
       },
     })
 
-    await expect(promise).rejects.toMatchPrismaErrorInlineSnapshot(`
+    await expect(promise).rejects.toMatchObject({
+      message: expect.stringContaining('Failed to validate the query'),
+      code: 'P2009',
+    })
+
+    if (!process.env.DATA_PROXY) {
+      await expect(promise).rejects.toMatchPrismaErrorInlineSnapshot(`
 
       Invalid \`prisma.resource.create()\` invocation in
       /client/tests/functional/issues/13913-integer-overflow/tests.ts:0:0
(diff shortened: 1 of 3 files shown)


packages/client/tests/functional/issues/13913-integer-overflow/tests.ts
```diff
@@ -0,0 +1,27 @@
...
+    await expect(promise).rejects.toMatchPrismaErrorInlineSnapshot(`
+
+      Invalid \`prisma.resource.create()\` invocation in
+      /client/tests/functional/issues/13913-integer-overflow/tests.ts:0:0
+
+        XX 
+        XX testMatrix.setupTestSuite(() => {
+        XX   test('int overflow', async () => {
+      → XX     const promise = prisma.resource.create(
+      Failed to validate the query: \`Unable to match input value to any allowed input type for the field. Parse errors: [Query parsing/validation error at \`Mutation.createOneResource.data.ResourceCreateInput.number\`: Unable to fit integer value '2265000000' into a 32-bit signed integer for field 'number'., Query parsing/validation error at \`Mutation.createOneResource.data.ResourceUncheckedCreateInput.number\`: Unable to fit integer value '2265000000' into a 32-bit signed integer for field 'number'.]\` at \`Mutation.createOneResource.data\`
```
janpio, line 24: This looks like a "bad", unhandled error message that should get a proper error code and name if we keep doing this.

millsp: The error looks ok, it also has a `P2023` error code. The message isn't super user-friendly though `Mutation.createOneResource.data.ResourceUncheckedCreateInput.number`, this is an implementation detail probably leaking into the message.

janpio: Are we just skipping the error code part in our snapshots?

millsp: Yes, I assume this is because the `code` property is not enumerable. We should fix that.

janpio: :+1: Def confused me.

danstarns (approved):

Issue #15470, opened 2022-09-22
In some cases, concurrent `upsert` cause data not to be written for `mysql`


millsp:
### Bug description

The following example is shows how in some cases, concurrent `upsert` cause data not to be written for `mysql`. I am not able to reproduce this locally, but I can get almost-consistent failures on CI.

[Internal discussion](https://prisma-company.slack.com/archives/C03V1UE750D/p1663709917516639)

> So the test for OCC is for upsert, which is roughly laid out the same as update: we increment but only if the occStamp matches the where clause.
> This test does this five times and concurrently, and with OCC working your occStamp wont be 5, but 1 since the first query locks the resource.

### How to reproduce

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mysql"
  url      = env("DATABASE_URI_mysql")
}

model Resource {
  id       String @id @default(cuid())
  occStamp Int @default(0) @unique
}
```

```ts
  testIf(provider !== 'mysql')('upsert', async () => {
    const fn = async () => {
      const resource = (await prisma.resource.findFirst())!

      expect(resource).toMatchObject({ occStamp: 0 })

      await prisma.resource.upsert({
        where: { occStamp: resource.occStamp },
        update: { occStamp: { increment: 1 } },
        create: {},
      })
    }

    await Promise.allSettled([fn(), fn(), fn(), fn(), fn()])

    expect(await prisma.resource.findFirst()).toMatchObject({ occStamp: 1 })
  })
```

### Expected behavior

It should either work and yield `{ occStamp: 1 }` or error, but not give back `{ occStamp: 0 }`

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
// Add your code using Prisma Client
```

### Environment & setup

- OS: Linux, Windows, MacOS
- Database: MySQL
- Node.js version: 14,16

### Prisma Version

```
4.3.0 (dev)
```

PR #15550, opened 2022-09-27
test(client): add occ test with nested operation

millsp:
Adds a missing case for nested operations on relations.

### DIFF
--- a/packages/client/tests/functional/optimistic-concurrency-control/tests.ts
+++ b/packages/client/tests/functional/optimistic-concurrency-control/tests.ts
@@ -98,4 +98,30 @@ testMatrix.setupTestSuite(({ provider }) => {
 
     expect(await prisma.resource.findFirst()).toMatchObject({ occStamp: 1 })
   })
+
+  test('update with upsert relation', async () => {
+    const fn = async () => {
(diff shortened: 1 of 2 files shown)


janpio (approved):

packages/client/tests/functional/optimistic-concurrency-control/tests.ts
```diff
@@ -98,4 +98,30 @@ testMatrix.setupTestSuite(({ provider }) => {
 
     expect(await prisma.resource.findFirst()).toMatchObject({ occStamp: 1 })
   })
+
+  test('update with upsert relation', async () => {
```
danstarns, line 102: Does this have a related issue we could refer to? 
```suggestion
  // Related https://github.com/prisma/prisma/issues/foo-bar?
  test('update with upsert relation', async () => {
```

danstarns (approved):

packages/client/tests/functional/optimistic-concurrency-control/tests.ts
danstarns: Found it https://github.com/prisma/prisma/issues/15470

Issue #9326, opened 2021-09-20
Raw query failed when trying to insert batch of data


PedramRZM:
### Bug description

I have a `rawInsert` function which is in charge of executing a raw query to insert multiple rows in a table. Function header looks like this: 

`export default async <T extends Record<string, any>>(
  table: string,
  fields: DistributiveKeyof<T>[],
  values: T[],
  opts?: { onConflict?: { target: (keyof T)[]; action: 'update' } },
): Promise<number>`

When I try to load test my API  using [k6](https://k6.io) this function throws an error. This error happens when my batch size gets more than ~5000 rows.
Error:
```
Raw query failed. Code: `N/A`. Message: `N/A`
```

### How to reproduce

1. generate a query like below and execute it using `executeRaw`. make sure you insert more than 5000 rows in this query.
`INSERT INTO table_name (col1, col2, col3, col4, col5, col6)
VALUES
    ($1, $2, $3, $4, $5, $6),
    ($7, $8, $9, $10, $11, $12),
    ...`
2. run a load test over this query using [k6](https://k6.io). iterating through this also reproduces this error.
3. run the exact same query using `pg` driver and everything is working just fine.

### Expected behavior

To have a clear error message and insert all of my batch data rows into table.

### Prisma information

```
generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["selectRelationCount"]
}

datasource db {
  provider = "postgresql"
}
```

### Environment & setup

- OS: Mac OS
- Database: PostgreSQL
- Node.js version :v14.17.6

### Prisma Version

```
prisma                  : 3.0.2
@prisma/client          : Not found
Current platform        : darwin
Query Engine (Node-API) : libquery-engine 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at ../../.npm/_npx/52649/lib/node_modules/prisma/node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at ../../.npm/_npx/52649/lib/node_modules/prisma/node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at ../../.npm/_npx/52649/lib/node_modules/prisma/node_modules/@prisma/engines/introspection-engine-darwin)
[...]

Issue #8832, opened 2021-08-19
Misleading "Can't reach database server" error when more than `32766` variables are bound inside an `IN` clause in Postgres


some-user123:
### Bug description

After upgrading to 2.29.1 my app is not starting properly anymore. I tested it with 2.29.0 --> same error, versions below (e.g. 2.28.0 and 2.27.0 are working fine).

I'm using Prisma in a NestJS application running in a Docker container with an Postgres DB running in a separate container.

I receive this error:
```
/app/node_modules/@prisma/client/runtime/index.js:35943
          throw new import_engine_core3.PrismaClientKnownRequestError(message, e.code, this.prisma._clientVersion, e.meta);
                ^

PrismaClientKnownRequestError3 [PrismaClientKnownRequestError]:
Invalid `prisma.XXX.findMany()` invocation:


  Can't reach database server at `db`:`5432`

Please make sure your database server is running at `db`:`5432`.
    at cb (/app/node_modules/@prisma/client/runtime/index.js:35943:17)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async XXX.XXX (/app/dist/XXX/XXX.service.js:41:28) {
  code: 'P1001',
  clientVersion: '2.29.0',
  meta: { database_host: 'db', database_port: 5432 }
}
```

Before I receive this error message `prisma migrate deploy` runs successfully and two DB query are executed successfully. Therefore I highly doubt the error message, that the DB is not reachable. (Plus the error does not occur with the same DB and earlier versions of prisma.)

Executing the same app with `DEBUG="*"` (removed lines from express and NestJS for brevity):
```
yarn run v1.22.5
$ /app/node_modules/.bin/prisma migrate deploy
2021-08-19T20:01:12.805Z prisma:loadEnv project root found at /app/package.json
2021-08-19T20:01:12.863Z prisma:tryLoadEnv Environment variables not found at null
2021-08-19T20:01:12.864Z prisma:tryLoadEnv Environment variables not found at undefined
2021-08-19T20:01:12.864Z prisma:tryLoadEnv No Environment variables loaded
2021-08-19T20:01:12.867Z prisma:engines using Node API: false
2021-08-19T20:01:12.867Z prisma:engines binaries to download query-engine, migration-engine, introspection-engine, prisma-fmt
Prisma schema loaded from prisma/schema.prisma
[...]

PR #15556, opened 2022-09-27
test(postgres): add regression test for #8832 and #9326

jkomyno:
This PR wraps the following two integration PRs:
- https://github.com/prisma/prisma/pull/15522
- https://github.com/prisma/prisma/pull/15490

### DIFF
--- a/packages/client/tests/functional/issues/8832/_matrix.ts
+++ b/packages/client/tests/functional/issues/8832/_matrix.ts
@@ -1,0 +1,9 @@
+import { defineMatrix } from '../../_utils/defineMatrix'
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'postgresql',
+    },
+  ],
+])
--- a/packages/client/tests/functional/issues/8832/prisma/_schema.ts
+++ b/packages/client/tests/functional/issues/8832/prisma/_schema.ts
@@ -1,0 +1,33 @@
+import { idForProvider } from '../../../_utils/idForProvider'
+import testMatrix from '../_matrix'
+
+export default testMatrix.setupSchema(({ provider }) => {
+  return /* Prisma */ `
+    generator client {
+      provider = "prisma-client-js"
+    }
+    
+    datasource db {
+      provider = "${provider}"
+      url      = env("DATABASE_URI_${provider}")
+    }
+    
+    model Post {
+      id   Int           @id
+      tags TagsOnPosts[]
+      @@map("post")
+    }
+    model Tag {
+      id    Int           @id
+      posts TagsOnPosts[]
+      @@map("tag")
+    }
+    model TagsOnPosts {
+      postId Int
+      tagId  Int
+      post   Post @relation(fields: [postId], references: [id], onDelete: Cascade)
+      tag    Tag  @relation(fields: [tagId], references: [id], onDelete: Cascade)
+      @@id([postId, tagId])
+    }
+  `
+})
--- a/packages/client/tests/functional/issues/8832/prisma/_schema.ts
+++ b/packages/client/tests/functional/issues/8832/prisma/_schema.ts
@@ -17,11 +17,13 @@ export default testMatrix.setupSchema(({ provider }) => {
       tags TagsOnPosts[]
       @@map("post")
     }
+
     model Tag {
       id    Int           @id
       posts TagsOnPosts[]
       @@map("tag")
     }
+
     model TagsOnPosts {
       postId Int
       tagId  Int
--- a/packages/client/tests/functional/issues/8832/tests.ts
+++ b/packages/client/tests/functional/issues/8832/tests.ts
@@ -46,7 +46,7 @@ testMatrix.setupTestSuite(
         await clean()
       }, 10_000)
 
-      test('should succeed when "in" has 32766 or less ids', async () => {
+      test('should succeed when "in" has 32766 ids', async () => {
         const n = 32766
         const ids = await createTags(n)
         const tags = await getTagsFindManyIn(ids)
@@ -54,7 +54,7 @@ testMatrix.setupTestSuite(
         expect(tags.length).toBe(n)
       })
 
-      test('should succeed when "include" involves 32766 or less records', async () => {
+      test('should succeed when "include" involves 32766 records', async () => {
         const n = 32766
         await createTags(n)
         const tags = await getTagsFindManyInclude()
@@ -62,7 +62,7 @@ testMatrix.setupTestSuite(
         expect(tags.length).toBe(n)
       })
 
-      test('should succeed when "in" has 32767 or less ids', async () => {
+      test('should succeed when "in" has 32767 ids', async () => {
         const n = 32767
         const ids = await createTags(n)
         const tags = await getTagsFindManyIn(ids)
@@ -70,7 +70,7 @@ testMatrix.setupTestSuite(
         expect(tags.length).toBe(n)
       })
 
-      test('should succeed when "include" involves 32767 or less records', async () => {
+      test('should succeed when "include" involves 32767 records', async () => {
         const n = 32767
         await createTags(n)
         const tags = await getTagsFindManyInclude()
@@ -78,7 +78,7 @@ testMatrix.setupTestSuite(
         expect(tags.length).toBe(n)
       })
 
-      test('should succeed when "in" has 32768 or less ids', async () => {
+      test('should succeed when "in" has 32768 ids', async () => {
         const n = 32768
         const ids = await createTags(n)
         const tags = await getTagsFindManyIn(ids)
@@ -86,7 +86,7 @@ testMatrix.setupTestSuite(
         expect(tags.length).toBe(n)
       })
 
-      test('should succeed when "include" involves 32768 or less records', async () => {
+      test('should succeed when "include" involves 32768 records', async () => {
         const n = 32768
         await createTags(n)
         const tags = await getTagsFindManyInclude()
@@ -94,7 +94,7 @@ testMatrix.setupTestSuite(
         expect(tags.length).toBe(n)
       })
 
-      test('should succeed when "in" has 65536 or less ids', async () => {
+      test('should succeed when "in" has 65536 ids', async () => {
         const n = 65536
         const ids = await createTags(n)
         const tags = await getTagsFindManyIn(ids)
@@ -102,7 +102,7 @@ testMatrix.setupTestSuite(
         expect(tags.length).toBe(n)
       })
 
-      test('should succeed when "include" involves 65536 or less records', async () => {
+      test('should succeed when "include" involves 65536 records', async () => {
         const n = 65536
         await createTags(n)
         const tags = await getTagsFindManyInclude()
@@ -146,7 +146,7 @@ testMatrix.setupTestSuite(
         }
       })
 
-      test('should fail with `value too large to transmit` when "in" has 32766 or less ids and a "take" filter', async () => {
+      test('should fail with `value too large to transmit` when "in" has 32766 ids and a "take" filter', async () => {
         expect.assertions(3)
         const n = 32766
         const ids = await createTags(n)
@@ -170,7 +170,7 @@ testMatrix.setupTestSuite(
         }
       })
 
-      test('should succeed when "in" has 32766 or less ids and a "skip" filter', async () => {
+      test('should succeed when "in" has 32766 ids and a "skip" filter', async () => {
         const n = 32766
         const ids = await createTags(n)
 
--- a/packages/client/tests/functional/issues/8832/tests.ts
+++ b/packages/client/tests/functional/issues/8832/tests.ts
@@ -109,7 +109,13 @@ testMatrix.setupTestSuite(
 
         expect(tags.length).toBe(n)
       })
+    })
 
+    /**
+     * Highlight scenarios of the query chunking logic behavior depending on how the bind variables are counted
+     * in the query filters.
+     */
+    describe('unhandled filters', () => {
       test('should fail with `value too large to transmit` when "in" is repeated at least twice and "n" is $QUERY_BATCH_SIZE', async () => {
         expect.assertions(3)
         const n = 32766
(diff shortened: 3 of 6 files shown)


packages/client/tests/functional/issues/8832/prisma/_schema.ts
```diff
@@ -0,0 +1,33 @@
...
+      posts TagsOnPosts[]
+      @@map("tag")
+    }
+    model TagsOnPosts {
+      postId Int
+      tagId  Int
+      post   Post @relation(fields: [postId], references: [id], onDelete: Cascade)
+      tag    Tag  @relation(fields: [tagId], references: [id], onDelete: Cascade)
+      @@id([postId, tagId])
+    }
```
janpio, line 31:
```suggestion
    model Post {
      id   Int           @id
      tags TagsOnPosts[]
      @@map("post")
    }

    model Tag {
      id    Int           @id
      posts TagsOnPosts[]
      @@map("tag")
    }

    model TagsOnPosts {
      postId Int
      tagId  Int
      post   Post @relation(fields: [postId], references: [id], onDelete: Cascade)
      tag    Tag  @relation(fields: [tagId], references: [id], onDelete: Cascade)
      @@id([postId, tagId])
    }
```

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,192 @@
...
+
+      test('should succeed when "include" involves 65536 or less records', async () => {
+        const n = 65536
+        await createTags(n)
+        const tags = await getTagsFindManyInclude()
+
+        expect(tags.length).toBe(n)
+      })
+
+      test('should fail with `value too large to transmit` when "in" is repeated at least twice and "n" is $QUERY_BATCH_SIZE', async () => {
```
janpio, line 113: This and following test cases (except the last one) are the ones we did not fix yet, correct?
So when we do fix that, these tests here would also need to pass?
Do we have an issue for that work yet? 

I think it would make sense to group this into its own "describe" and link to the issue for context.

packages/client/tests/functional/issues/9326/prisma/_schema.ts
```diff
@@ -0,0 +1,33 @@
...
+      posts TagsOnPosts[]
+      @@map("tag")
+    }
+    model TagsOnPosts {
+      postId Int
+      tagId  Int
+      post   Post @relation(fields: [postId], references: [id], onDelete: Cascade)
+      tag    Tag  @relation(fields: [tagId], references: [id], onDelete: Cascade)
+      @@id([postId, tagId])
+    }
```
janpio, line 31:
```suggestion
    model Post {
      id   Int           @id
      tags TagsOnPosts[]
      @@map("post")
    }

    model Tag {
      id    Int           @id
      posts TagsOnPosts[]
      @@map("tag")
    }

    model TagsOnPosts {
      postId Int
      tagId  Int
      post   Post @relation(fields: [postId], references: [id], onDelete: Cascade)
      tag    Tag  @relation(fields: [tagId], references: [id], onDelete: Cascade)
      @@id([postId, tagId])
    }
```

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,192 @@
...
+      const cleanPrismaPromises = [prisma.tagsOnPosts.deleteMany(), prisma.post.deleteMany(), prisma.tag.deleteMany()]
+      await prisma.$transaction(cleanPrismaPromises)
+    }
+
+    describe('issue #8832', () => {
+      beforeEach(async () => {
+        await clean()
+      }, 10_000)
+
+      test('should succeed when "in" has 32766 or less ids', async () => {
```
janpio, line 49: why "or less" here? Isn't the text exactly that number?

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,192 @@
...
+
+      test('should succeed when "include" involves 32767 or less records', async () => {
+        const n = 32767
+        await createTags(n)
+        const tags = await getTagsFindManyInclude()
+
+        expect(tags.length).toBe(n)
+      })
+
+      test('should succeed when "in" has 32768 or less ids', async () => {
```
janpio, line 81: similar as above (and for all tests):
why "or less" here? Isn't the text exactly that number?

packages/client/tests/functional/issues/9326/tests.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      test('should succeed when the number of params is 32767 or less', async () => {
+        const n = 32767
+        const ids = await createTags(n)
+        const tags = await getTagsParams(ids)
+
+        expect(tags.length).toBe(n)
+      })
+
+      test('should fail with `value too large to transmit` when the number of params is 32768+', async () => {
```
janpio, line 51: re "less" and "+/more": Doesn't the test only test exactly that one case from the issue?
Otherwise we needed another test for a "less" or "more" number.

packages/client/tests/functional/issues/9326/tests.ts
```diff
@@ -0,0 +1,81 @@
...
+          expect(error.meta).toMatchObject({
+            database_error: 'too many bind variables in prepared statement, expected maximum of 32767, received 32768',
+          })
+        }
+      })
+    })
+  },
+  {
+    optOut: {
+      from: ['mysql', 'sqlserver', 'cockroachdb', 'sqlite', 'mongodb'],
```
janpio, line 73: Do we have tests to make sure this does not happen with the other databases as well?
Would it make sense to add those as a follow up to make sure our chunking and error message logic works - via a future PR? THen an issue would be good.

packages/client/tests/functional/issues/9326/tests.ts
```diff
@@ -0,0 +1,81 @@
...
+      reason: 'we have only captured this issue with postgres',
+    },
+    skipDataProxy: {
+      runtimes: ['node', 'edge'],
+      reason: "Mini-Proxy can't currently handle big queries, un-skip when it starts using QE server instead of CLI",
```
janpio, line 78: Do we have an issue we can link for this?

jkomyno: https://github.com/prisma/mini-proxy/issues/30

packages/client/tests/functional/issues/8832/tests.ts
jkomyno: The `less` conveys information on the behavior in the range `[0..32766]` ids in the `in` filter clause.
I get that this might be confusing in the test name. I'll get rid of it

packages/client/tests/functional/issues/9326/tests.ts
jkomyno: Created: https://github.com/prisma/prisma-private/issues/192

packages/client/tests/functional/issues/8832/tests.ts
janpio: Ok, then I understood correctly 👍

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,202 @@
...
+    async function createTags(length: number): Promise<number[]> {
+      const ids = Array.from({ length }, (_, i) => i + 1)
+      const data = ids.map((id) => ({ id }))
+      await prisma.tag.createMany({
+        data,
+      })
+      return ids
+    }
+
+    async function getTagsFindManyIn(ids: number[]): Promise<unknown[]> {
```
aqrln, line 19:
```suggestion
    async function getTagsFindManyIn(ids: number[]): Promise<Tag[]> {
```

and import `Tag` above

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,202 @@
...
+        where: {
+          id: {
+            in: ids,
+          },
+        },
+      })
+      return tags
+    }
+
+    async function getTagsFindManyInclude(): Promise<unknown[]> {
```
aqrln, line 30: ditto

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,202 @@
...
+                    id: { in: ids },
+                  },
+                  {
+                    id: { in: ids },
+                  },
+                ],
+              },
+            })
+          } catch (e) {
+            const error = e as Error & { code: number; meta?: unknown }
```
aqrln, line 137: maybe
```suggestion
            const error = e as Prisma.PrismaClientKnownRequestError
```
(and import `Prisma` above)

but no strong opinion one way or the other

packages/client/tests/functional/issues/8832/_matrix.ts
```diff
@@ -0,0 +1,9 @@
+import { defineMatrix } from '../../_utils/defineMatrix'
+
+export default defineMatrix(() => [
+  [
+    {
+      provider: 'postgresql',
```
Jolg42, line 6: Why do we only test for PostgreSQL here?

This comment https://github.com/prisma/prisma/issues/8832#issuecomment-1237850099 mentions that the limit is different on other databases, maybe it would be good to test these as well?

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,202 @@
...
+        expect(tags.length).toBe(n)
+      })
+
+      test('should succeed when "include" involves 32768 records', async () => {
+        const n = 32768
+        await createTags(n)
+        const tags = await getTagsFindManyInclude()
+
+        expect(tags.length).toBe(n)
+      })
```
Jolg42, line 95: Why do we need all these tests (32766, 32767, 32768)?
Can we only keep the last 2 with n = 32768 maybe?

packages/client/tests/functional/issues/8832/_matrix.ts
janpio: Regression was only identified for PostgreSQL. We created an issue to test others as well (it is mentioned somewhere else here)

packages/client/tests/functional/issues/8832/tests.ts
Jolg42: Or remove them all and only keep the ones below where n = 65536?

packages/client/tests/functional/issues/8832/_matrix.ts
janpio: Seems issue was only created, not mentioned here at all: https://github.com/prisma/prisma-private/issues/192

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      /**
+       * Highlight scenarios of the query chunking logic behavior depending on how the bind variables are counted
+       * in the query filters.
+       */
+      describe('unhandled filters', () => {
+        test('should fail with `value too large to transmit` when "in" is repeated at least twice and "n" is $QUERY_BATCH_SIZE', async () => {
+          expect.assertions(3)
+          const n = 32766
```
Jolg42, line 120: Since `QUERY_BATCH_SIZE` is not defined (can't find it), how do we know that `n === QUERY_BATCH_SIZE` here?

janpio: I _think_ what Alberto means here is that n is the same as the default chunk size for PostgreSQL (which can be modifed via `QUERY_BATCH_SIZE`).

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,202 @@
...
+            )
+            expect(error.code).toBe('P2035')
+            expect(error.meta).toMatchObject({
+              database_error:
+                'too many bind variables in prepared statement, expected maximum of 32767, received 65533',
+            })
+          }
+        })
+
+        test('should fail with `value too large to transmit` when "in" has 32766 ids and a "take" filter', async () => {
```
Jolg42, line 149: About "should fail with `value too large to transmit`" (applicable to all occurences)
I don't see `value too large to transmit` in the expects 🤔 
Is it part of the error message?

janpio: I think that is actually outdated - the query never reaches tokio-postgres that would throw that error message, and instead we have something in quaint that catches it and responds with the error message that is expected below. We should probably update the description here.

packages/client/tests/functional/issues/8832/tests.ts
jkomyno: `$QUERY_BATCH_SIZE` indeed indicates the environment variable `QUERY_BATCH_SIZE`, which influences the default values in [here](https://github.com/prisma/prisma-engines/blob/main/query-engine/connectors/sql-query-connector/src/sql_info.rs).

packages/client/tests/functional/issues/8832/tests.ts
jkomyno: Because previously the ones with `32767` and `32768` were failing. We shouldn't test a single `n` because there's also a query batching logic kicking in around those numbers (at a certain point, it used to work with `32766` and `32768` records, but not with `32767` records).

packages/client/tests/functional/issues/8832/tests.ts
jkomyno: Thanks, the message is indeed outdated. Fixing it

packages/client/tests/functional/issues/8832/tests.ts
janpio: `QUERY_BATCH_SIZE` is not used in this test.

jkomyno: Updated test name as:

> should fail with `Assertion violation` when "in" is repeated at least twice and "n" is the default value of $QUERY_BATCH_SIZE

Jolg42 (approved):
Nice!

jkomyno:
Unrelated flaky tests aside, we can safely merge. Monitoring the `main` CI later.

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        expect(tags.length).toBe(n)
+      })
+
+      /**
+       * Highlight scenarios of the query chunking logic behavior depending on how the bind variables are counted
+       * in the query filters.
+       */
+      describe('unhandled filters', () => {
+        test('should fail with `Assertion violation` when "in" is repeated at least twice and "n" is the default value of $QUERY_BATCH_SIZE', async () => {
```
janpio, line 118: This is not actually the _intended_ behavior, it just _is_ currently because our chunking logic is not able to handle that, right?

PR #15557, opened 2022-09-27
test: try Jest's --onlyChanged for functional tests

Jolg42:
Only for GitHub Actions

This checks if all the files modified (or added) by a PR are located inside `packages/client/tests/functional`
If true we only run these tests, thanks to Jest's `--onlyChanged`
else we run like before (all client tests)


Jolg42:
Merging now and trying it with https://github.com/prisma/prisma/pull/15563

Jolg42:
Only worked locally but not in GitHub Actions 😢

PR #15558, opened 2022-09-27
test: Reenable SQL Server for P2034 tests (issue #9678)

janpio:


packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -961,9 +961,6 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
       return { data, elapsed } as any
     } catch (e: any) {
       logger('req - e', e)
-      if (e instanceof PrismaClientKnownRequestError) {
```
millsp, line 964: I'm trying to understand how moving this into `handleRequestError` fixes the issue

SevInf: This used to be a part of `request` method and was not happening for `requestBatch`. So, in case of batch request: 
1. Write conflict error happens. Known error is created and thrown.
2. "ROLLBACK with no BEGIN" error is logged on the engine side, but it's ignored and not propagated to the client by design.
3. This error gets caught in `catch` part of `requestBatch`.
4. `handleRequestError` is called. It sees that there is "ROLLBACK with no BEGIN"  in the logs, creates unknown error from that message, which replaces known error.

Moving this check into `handleRequestError` forces it to happen for both `request` and `requestBatch`. So, if we already have a known error, we will never replace it with unknown one.

millsp: Ah that makes a lot more sense now. Thanks

millsp (approved):

PR #15563, opened 2022-09-27
chore: replace `clean` transaction with sequence of promises, increased `beforeEach` timeout

jkomyno:
Attempt to unlock CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -187,7 +187,7 @@ jobs:
 
       # If we detected (in detect_jobs_to_run) that only functional test files changed
       # We only run the tests that changed (or were added)
-      - run: pnpm run test:functional --changedSince=main
+      - run: pnpm run test:functional --onlyChanged
         if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-clientOnlyChangedFunctionalTests-') }}
         working-directory: packages/client
       # If not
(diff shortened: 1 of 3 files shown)


janpio:
(Now better no merges from `main` any more as that disables these tests)

Jolg42:
Note: the test is skipped on main since https://github.com/prisma/prisma/pull/15565 because it was blocking the release.
After the release: We need to continue this PR here to make the test more robust and unskip it.

.github/workflows/test.yml
```diff
@@ -187,7 +187,7 @@ jobs:
 
       # If we detected (in detect_jobs_to_run) that only functional test files changed
       # We only run the tests that changed (or were added)
-      - run: pnpm run test:functional --onlyChanged
+      - run: pnpm run test:functional --changedSince=main
```
Jolg42, line 190:
```suggestion
      - run: pnpm run test:functional --onlyChanged
```

Jolg42:
Sorry for the mess here, everything is back to normal.

Note I created a PR to checl that only unskip the test to see the current status and it fails in Buildkite https://github.com/prisma/prisma/pull/15575

jkomyno:
Let’s consider that most likely increasing timeouts in those tests is just a temporary patch for a broader issue elsewhere: too many concurrent Prisma clients are competing for resources, resulting in transactions (e.g., the one used for cleaning the db tables) taking ages

Jolg42 (approved):
Did a rerun and it works! let's merge ✅

PR #15575, opened 2022-09-28
test(client): unskip 8832

Jolg42:
Currently fails (see logs)


Jolg42:
Will be "fixed" in https://github.com/prisma/prisma/pull/15563

PR #15677, opened 2022-10-05
feat(client): foundation for extension implementation

millsp:
This is a base code, on which we will base our future client extensions implementation. Implements basic functionality for `$extends` function and some basic types for extension.


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -303,38 +300,10 @@ const actionOperationMap = {
   aggregateRaw: 'query',
 }
 
-const TX_ID = Symbol.for('prisma.client.transaction.id')
-
-// TODO improve all these types, need a common place to share them between type
-// gen and this. This will be relevant for type gen tech debt refactor
-export interface Client {
```
SevInf, line 310: Nice

packages/client/src/runtime/core/request/createRequestProxy.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (proxyOptions?.forbiddenProps?.includes(prop as string)) return undefined
+
+      // we override and handle every function call within the proxy
+      if (typeof target[prop] === 'function') {
+        return (...args: unknown[]) => {
+          // we hijack promise calls to pass txId to prisma promises
+          if (prop === 'then') return target[prop](args[0], args[1], proxyOptions)
+          if (prop === 'catch') return target[prop](args[0], proxyOptions)
+          if (prop === 'finally') return target[prop](args[0], proxyOptions)
+          if (prop === '$extends') return target[prop](args[0], proxyOptions)
```
SevInf, line 30: Shouldn't `$extends` be among forbidden properties? Or am I misunderstanding something? Like, you should not be able to do:

```ts
prisma.$transaction((tx) => {
   tx.extends(someExtension)
})
```

packages/client/src/runtime/core/extension/$extends.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ *
+ * @param this
+ */
+export function $extends(this: Client, input: any, proxyOptions: RequestProxyOptions = {}) {
+  const forkedProxyOptions: Undefinable<Required<RequestProxyOptions>> = {
+    forbiddenProps: proxyOptions.forbiddenProps ? [...proxyOptions.forbiddenProps] : undefined,
+    transaction: proxyOptions.transaction ? { ...proxyOptions.transaction } : undefined,
+    extensions: proxyOptions.extensions ? [...proxyOptions.extensions, input] : undefined,
```
SevInf, line 66:
```suggestion
    extensions: (proxyOptions.extensions ?? []).concat(input)
```

Shouldn't we always add `input` to extensions, even they were undefined initially?

packages/client/src/runtime/core/extension/$extends.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  return createRequestProxy(this, forkedProxyOptions)
```
SevInf, line 69: I am probably missing something, but why wouldn't something along this lines work?

```ts
class PrismaClient {
   private _extensions = []
   ....
   $extends(ext) {
         return Object.create(this, {
            _extensions: {
                get: () => this._extensions.concat(ext)
            }
         })
   }
}
```

Using request proxy for this looks like ф ыдшпре overkill for me, but again, I might miss something here.

packages/client/src/runtime/core/request/RequestPromise.ts
```diff
@@ -0,0 +1,66 @@
...
+   * @param onfinally same as regular promises
+   * @param options execution options
+   */
+  finally(onfinally?: (() => void) | undefined | null, options?: RequestPromiseOptions): Promise<A>
+
+  /**
+   * Called when executing a batch of regular tx
+   * @param options execution options
+   */
+  requestTransaction?(options?: RequestPromiseOptions): PromiseLike<unknown>
```
SevInf, line 65: Would `requestTransaction` call make sense for `{ kind: 'itx'}`? Original code assumed it won't and made `requestTransaction` explicitly accept only batch options.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -506,7 +477,7 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
       return 'PrismaClient'
     }
 
-    private getEngine(): Engine {
+    getEngine(): Engine {
```
SevInf, line 480: I am curios, why `private` removal?

SevInf (commented):
Mostly comments/questions for now.

packages/client/src/runtime/core/request/createRequestPromise.ts
```diff
@@ -0,0 +1,55 @@
...
+      const promise = _callback(options, false)
+
+      // we want to have support for nested promises
+      if (promise.requestTransaction) {
+        return promise.requestTransaction(options)
+      }
+
+      return promise
+    },
+    [Symbol.toStringTag]: batchable ? 'PrismaPromise' : 'RequestPromise',
```
millsp, line 53: Thanks might not be necessary, after all.

packages/client/src/runtime/getPrismaClient.ts
millsp: That's for being able to export the types automatically via `infer`, typescript complained about it. I don't think it had much value anyways.

packages/client/src/runtime/core/request/createRequestProxy.ts
millsp: I think that should work, why should you not be able to do that?

packages/client/src/runtime/core/extension/$extends.ts
millsp: That looks simpler indeed, I'll change to use that instead.

packages/client/src/runtime/core/request/createRequestProxy.ts
SevInf: Nothing really wrong, it's just seems weird to me. Like, itx ideally live for couple of dozens milliseconds, it seems weird to add extension that would be active just for that duration.

millsp: True. On the other hand, extensions should generally have minimal instantiation overhead. I think extending queries could be useful there, and perhaps results too. For sure, model and client would not make much sense there.

millsp: I've restricted it. Well see if users ever ask for it.

SevInf (approved):

PR #15684, opened 2022-10-06
fix(client): remove retries for all errors except network connection/socket errors

### LINKED ISSUE #12862: Some errors are obfuscated by interactive transactions when using `binary` engine
### Bug description

Runtime errors such as `CHECK` constraint failures are obfuscated by interactive transactions when `engineType  = "binary"`, making it impossible to appropriately handle rejected transactions.

Only tested this with `postgres` so far. 

Note that the `library` engine does not seem to have the same issue. CI output comparing both engine types: https://github.com/driimus/prisma-interactive-tx-errors/actions/runs/2184082955

### How to reproduce

1. Clone https://github.com/driimus/prisma-interactive-tx-errors
2. Run the tests locally using a `binary` engine
3. Notice that the test using an interactive transaction fails due to a nondescript error.

### Expected behavior

The original error / error message should be surfaced, similar to the `library` engine's behavior.

Expected:

```
ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(Error { kind: Db, cause: Some(DbError { severity: \"ERROR\", parsed_severity: Some(Error), code: SqlState(\"23514\"), message: \"new row for relation \\\"Post\\\" violates check constraint \\\"Post_viewCount_check\\\"\", detail: Some(\"Failing row contains (1, 2022-04-18 12:42:07.239, 2022-04-18 12:42:07.239, Minima nulla ea., null, f, -1, 1).\"), hint: None, position: None, where_: None, schema: Some(\"public\"), table: Some(\"Post\"), column: None, datatype: None, constraint: Some(\"Post_viewCount_check\"), file: Some(\"execMain.c\"), line: Some(2003), routine: Some(\"ExecConstraints\") }) }) })"
```

Received:

```
 ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(Error { kind: Db, cause: Some(DbError { severity: \"ERROR\", parsed_severity: Some(Error), code: SqlState(\"25P02\"), message: \"current transaction is aborted, commands ignored until end of transaction block\", detail: None, hint: None, position: None, where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some(\"postgres.c\"), line: Some(1718), routine: Some(\"exec_bind_message\") }) }) })"
```

### Prisma information

[Minimal reproduction repo](https://github.com/driimus/prisma-interactive-tx-errors)

```schema.prisma
generator client {
  engineType      = "binary"
}

datasource db {
  provider = "postgres"
  url      = "postgresql://postgres:example@localhost"
}

model User {
  email String  @unique
  name  String?
  posts Post[]
}

model Post {
  updatedAt DateTime @updatedAt
  title     String
  content   String?
  viewCount Int      @default(0)
  author    User?    @relation(fields: [authorId], references: [id])
  authorId  Int?
}
```

### Environment & setup

- OS: `Ubuntu 20.04.4 LTS`
- Database: `PostgreSQL`
- Node.js version: `16.14.2`

### Prisma Version

```
prisma                  : 3.12.0
@prisma/client          : 3.12.0
Query Engine (Node-API) : libquery-engine 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/.pnpm/@prisma+engines@3.12.0-37.22b822189f46ef0dc5c5b503368d1bee01213980/node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/.pnpm/@prisma+engines@3.12.0-37.22b822189f46ef0dc5c5b503368d1bee01213980/node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/.pnpm/@prisma+engines@3.12.0-37.22b822189f46ef0dc5c5b503368d1bee01213980/node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/.pnpm/@prisma+engines@3.12.0-37.22b822189f46ef0dc5c5b503368d1bee01213980/node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : 22b822189f46ef0dc5c5b503368d1bee01213980
Studio                  : 0.459.0
```

danstarns:
The binary engine had retry logic that was causing inconsistent behavior for users. 

See https://prisma-company.slack.com/archives/CUXLS0Z6K/p1664962929905829

- **Removes retries for all write operations using the BinaryEngine.** 
- **Removes retires for all batch operations --if a child request contains a write--.** 

This PR closes https://github.com/prisma/prisma/issues/12862 and contains tests to assert that.


garrensmith:
@danstarns this isn't correct. From our internal discussion we want to remove retries for all errors except network connection/socket errors

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -1110,66 +1130,70 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
...
-      // https://github.com/nodejs/undici/blob/2.x/lib/core/errors.js
-      error.code === 'ECONNRESET' ||
-      error.code === 'ECONNREFUSED' ||
-      error.code === 'UND_ERR_CLOSED' ||
-      error.code === 'UND_ERR_SOCKET' ||
-      error.code === 'UND_ERR_DESTROYED' ||
-      error.code === 'UND_ERR_ABORTED' ||
-      error.message.toLowerCase().includes('client is destroyed') ||
-      error.message.toLowerCase().includes('other side closed') ||
-      error.message.toLowerCase().includes('the client is closed')
```
aqrln, line 1140: I think these three (now removed) also refer to some kind of network errors, but yeah, might be safer not to match them to avoid false positives.

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -92,16 +93,33 @@ export class RequestHandler {
         // TODO: pass the child information to QE for it to issue links to queries
         // const links = requests.map((r) => trace.getSpanContext(r.otelChildCtx!))
 
+        const writeWords = ['create', 'update', 'delete']
+        const requestContainsWrite = requests.some((r) =>
+          writeWords.some((writeWord) => {
+            r.clientMethod.includes(writeWord)
```
millsp, line 99: Is there a missing `return` here?

danstarns: Good catch!

packages/engine-core/src/binary/BinaryEngine.ts
danstarns: We now check only the codes, this felt like a double-check for no reason

https://github.com/nodejs/undici/blob/2.x/lib/core/errors.js

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -92,13 +93,25 @@ export class RequestHandler {
         // TODO: pass the child information to QE for it to issue links to queries
         // const links = requests.map((r) => trace.getSpanContext(r.otelChildCtx!))
 
-        return this.client._engine.requestBatch(queries, info.headers, info.batchTransaction)
+        const writeWords = ['create', 'update', 'delete']
+        const requestContainsWrite = requests.some((r) =>
+          writeWords.some((writeWord) => {
+            r.clientMethod.includes(writeWord)
```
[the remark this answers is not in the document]
danstarns: `r.type` don't exist, do you mean something else?

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -960,23 +970,33 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
     } catch (e: any) {
       logger('req - e', e)
 
-      await this.handleRequestError(e, numTry <= MAX_REQUEST_RETRIES)
+      const { error, shouldRetry } = await this.handleRequestError(e)
+
+      const isWrite = ['create', 'update', 'delete'].some((method) => clientMethod.includes(method))
```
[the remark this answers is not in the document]
danstarns: I have made a shared util for the iswrite check.

packages/client/src/runtime/RequestHandler.ts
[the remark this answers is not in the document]
aqrln: ah, sorry, `r.document.type`

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -931,12 +932,19 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
...
-    query: string,
-    headers: QueryEngineRequestHeaders = {},
-    _transaction?: InteractiveTransactionOptions<undefined>,
+  async request<T>({
+    query,
+    headers = {},
     numTry = 1,
-  ): Promise<QueryEngineResult<T>> {
+    clientMethod,
+    _transaction,
```
aqrln, line 940:
```suggestion
```

aqrln: it was an unused parameter

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -931,12 +932,19 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
...
+  }: {
+    query: string
+    headers: QueryEngineRequestHeaders
+    numTry: number
+    clientMethod: string
+    _transaction?: InteractiveTransactionOptions<undefined>
```
aqrln, line 946:
```suggestion
    transaction?: InteractiveTransactionOptions<undefined>
```

or we could even drop it here

packages/engine-core/src/binary/BinaryEngine.ts
aqrln, line 940:
```suggestion
    transaction,
```

packages/engine-core/src/binary/BinaryEngine.ts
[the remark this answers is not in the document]
aqrln: Just moving this to a function is an improvement, but it still happens on two different levels of abstraction inconsistently. I'll fix this.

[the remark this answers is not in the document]
aqrln: Also I think it would be safer to flip the condition and make it `isRead` because if a new write operation is added we will probably forget to add it here at first. I'll do that too if I won't be able to use the document type instead.

packages/client/src/runtime/RequestHandler.ts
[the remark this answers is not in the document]
aqrln: fixed

aqrln (approved):

PR #15692, opened 2022-10-06
test(client): Flaky test higher timeout

danstarns:
Nothing much to see here, just a test that is timing out a lot. 🐑


aqrln:
In this test the Query Engine should fail to start, so the test should finish rather quickly. If it fails due to timeout, then increasing the timeout won't probably help as the test likely hangs infinitely.

danstarns:
Closed in favour of https://github.com/prisma/prisma/pull/16075

PR #15694, opened 2022-10-06
feat(client): enable ITX support for DataProxy

danstarns:
Adds DataProxy support for interactive transactions. 

**mini-proxy pr:** https://github.com/prisma/mini-proxy/pull/36/files
**internal doc:** https://www.notion.so/prismaio/iTX-between-DP-and-the-Client-91c133e5b60d43dbbfa409fde2b430b3

How to run the tests?

1. Clone the branch and follow the instructions [here](https://github.com/prisma/mini-proxy/tree/master/docs) to 'spin up' the mini-proxy. 

2. Run the interactive transaction tests in this branch with: 

```
cd ./packages/client && pnpm run test:functional --data-proxy --no-mini-proxy-server --no-mini-proxy-default-ca interactive-transactions
```


packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -11,222 +11,221 @@ declare let newPrismaClient: NewPrismaClient<typeof PrismaClient>
 
 const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
 
-testMatrix.setupTestSuite(
```
danstarns, line 14: Changes in this file are to do with the fact that I removed the: 

```
skipDataProxy: {
      runtimes: ['node', 'edge'],
      reason: 'Interactive transactions are not supported with Data Proxy yet',
    },
```

And the indent has gone.

janpio: (Hiding whitespace changes in GitHub PR file view hides this pretty well)

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -219,11 +227,73 @@ export class DataProxyEngine extends Engine {
...
+   * @param action START, COMMIT, or ROLLBACK
+   * @param headers headers for tracing
+   * @param options to change the default timeouts
+   * @param info transaction information for the QE
+   */
+  async transaction(action: 'start', headers: Tx.TransactionHeaders, options?: Tx.Options): Promise<Tx.Info>
+  async transaction(action: 'commit', headers: Tx.TransactionHeaders, info: Tx.Info): Promise<undefined>
+  async transaction(action: 'rollback', headers: Tx.TransactionHeaders, info: Tx.Info): Promise<undefined>
+  async transaction(action: any, headers: Tx.TransactionHeaders, arg?: any) {
+    await this.start()
```
millsp, line 241: I think we can remove this line

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -219,11 +227,73 @@ export class DataProxyEngine extends Engine {
...
+
+    const errorHandler = async (response: RequestResponse) => {
+      if (!response.ok) {
+        const json = await response.json()
+
+        if (json) {
+          throw new Error(`Error in transaction start: ${JSON.stringify(json)}`)
```
millsp, line 248: Can errors only happen in `start`? Or can they also happen in `commit`, etc..

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -219,11 +227,73 @@ export class DataProxyEngine extends Engine {
...
+
+      const endpoint = json['data-proxy'].endpoint as string
+
+      const id = endpoint.split('/').pop() as string
+
+      const data: Tx.Info = {
+        id,
+      }
+
+      return data as Tx.Info
```
aqrln, line 283: This ignores the hostname, which will be different for the transaction compared to the main hostname. This will work with mini-proxy right now but will not work with the real Data Proxy.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -219,11 +227,73 @@ export class DataProxyEngine extends Engine {
...
+    await this.start()
+
+    const errorHandler = async (response: RequestResponse) => {
+      if (!response.ok) {
+        const json = await response.json()
+
+        if (json) {
+          throw new Error(`Error in transaction start: ${JSON.stringify(json)}`)
+        } else {
+          throw new Error(`Error in transaction start: ${response.text()}`)
```
millsp, line 250: Ideally, there would be an error code and be documented like the rest of the errors. There a documented on the Data Proxy page in our docs.

millsp: In which cases can a transaction not start?

aqrln: I would expect errors for transaction start being pretty similar to query request, so the code should use `responseToError` helper and handle them pretty much the same way (including retries and uploading the schema). As for commit and rollback, we should definitely use the same mechanism for error handling and convert user-facing errors to instances of standard Data Proxy errors before throwing them, but maybe we don't have to handle `SchemaMissingError` for those two.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -30,6 +31,18 @@ const P = Promise.resolve()
 
 const debug = Debug('prisma:client:dataproxyEngine')
 
+// TODO: temporary hack; we need to store url on the client itx proxy instead and pass it to the engine similar to id
+// (probably some opaque `engineMeta: unknown` field and downcast it here to avoid leaking implementation details)
+const txBaseUrls = new Map<string, string>()
```
aqrln, line 36: todo: what the comment says

aqrln:
New PR for the integration branch: https://github.com/prisma/prisma/pull/16005

PR #15728, opened 2022-10-10
test(referential-action-set-null): add SQL tests for 1:1, 1:n relations with SET NULL referential actions

jkomyno:
To run tests:

- `cd package/client`
- `pnpm test:functional referential-action-set-null/tests_1-to-1.ts`
- `pnpm test:functional referential-action-set-null/tests_1-to-n.ts`

These tests expect the following environment variables (which are already used by the test codebase):
- `TEST_FUNCTIONAL_MYSQL_URI`
- `TEST_FUNCTIONAL_POSTGRES_URI`
- `TEST_FUNCTIONAL_COCKROACH_URI` (unnecessary when `TEST_SKIP_COCKROACHDB` is defined)
- `TEST_MSSQL_URI` (unnecessary when `TEST_SKIP_MSSQL` is defined)

## Outcome

- For both 1:1 and 1:n relations:
  - `postgres` doesn't validate DDL statements that set a `SET NULL` referential action on a foreign key constraint whose referenced column (or one of the referenced columns) is `NOT NULL`, all other database throw a validation error in this case.
  - `postgres` fails at runtime with a not-null constraint violation error when `UPDATE` /  `DELETE` statements trigger the `SET NULL` referential action on a column defined as `NOT NULL`
- `sqlserver` fails at runtime on 1:1 NULL relations because `NULL` conflicts with `UNIQUE` indexes in that particular database

## TODOs
- [x] fix `cockroach` and `postgres` in `insert + update + delete` tests
- [x] fix `Migration engine error:
    db error: ERROR: database “PRISMA_DB_NAME” is being accessed by other users`
- [x] Skip miniproxy
- [ ] Fix buildkite CI 
  - [ ] on `postgres`: `error: database "PRISMA_DB_NAME" does not exist`
  - [ ] on `sqlserver`: `Login failed for user 'sa'.`
- [ ] (LOW PRIORITY, can be done after we go GA with referential integrity) define `TEST_FUNCTIONAL_MARIADB_URI` in CI as `"mysql://root:root@localhost:4306/PRISMA_DB_NAME"`

### DIFF
--- a/packages/client/tests/functional/referential-action-set-null/tests_1-to-1.ts
+++ b/packages/client/tests/functional/referential-action-set-null/tests_1-to-1.ts
@@ -1,8 +1,9 @@
-import type { ProviderFlavor } from './_matrix'
 import testMatrix from './_matrix'
+import { getDatabaseURL } from './_utils/getDatabaseURL'
 import * as mssql from './_utils/mssql'
 import * as mysql from './_utils/mysql'
 import * as postgres from './_utils/postgres'
+import { setup } from './_utils/setup'
 import type { SetupParams } from './_utils/types'
 
 // @ts-ignore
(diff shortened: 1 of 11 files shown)


packages/client/tests/functional/referential-action-set-null/tests_1-to-1.ts
```diff
@@ -0,0 +1,405 @@
...
+// @ts-ignore
+const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
+
+// @ts-ignore
+const testIf = (condition: boolean) => (condition ? test : test.skip)
+
+function getDatabaseURL(providerFlavor: ProviderFlavor) {
+  switch (providerFlavor) {
+    case 'mssql':
+      return 'mssql://SA:Pr1sm4_Pr1sm4@localhost:1433/PRISMA_DB_NAME'
```
Jolg42, line 18: We could use from the GitHub Actions `TEST_MSSQL_URI: mssql://SA:Pr1sm4_Pr1sm4@localhost:1433/master`

PR #15729, opened 2022-10-10
chore: Use swc on windows and remove ts-jest

SevInf:
We introduced fallback for ts-jest back in #14281 because of source maps
issues in swc. As for latest update, corresponding issue is closed, we
can remove the fallback and use swc on all OSes.


SevInf:
Does not seems fixed, unfortinately.

PR #15781, opened 2022-10-13
test(client): pretty names for relationMode tests

### LINKED ISSUE #15765: relationMode: Change test naming
Currently we name the tests like this:
```
  relationMode.tests_1-to-n (provider=cockroachdb,id=String @id,relationMode=prisma,onUpdate=NoAction,onDelete=NoAction, previewFeatures=referentialIntegrity)
```

As parsing and understanding these is essential to be able to really _get_ the full test suite and its results, I would prefer this order:
```
  relationMode.tests_1-to-n (relationMode=prisma,provider=cockroachdb,onUpdate=NoAction,onDelete=NoAction, previewFeatures=referentialIntegrity,id=String @id)
```

Then the actually relevant information to understand what is going on is listed first (1 test run is always of the same relationMode, then we "matrix" over the providers, and then the different onUpdate/onDelete values).

This has to be changed in `packages/client/tests/functional/_utils/relationMode/computeMatrix.ts` (where we create the object that is stringified for this output), or in `packages/client/tests/functional/_utils/getTestSuiteInfo.ts` where we can replace the automated stringify with a conscious order in the string.

Jolg42:
Closes https://github.com/prisma/prisma/issues/15765

### DIFF
--- a/packages/client/tests/functional/_utils/getTestSuiteInfo.ts
+++ b/packages/client/tests/functional/_utils/getTestSuiteInfo.ts
@@ -121,14 +121,13 @@ export function getTestSuiteConfigs(suiteMeta: TestSuiteMeta): NamedTestSuiteCon
 function getTestSuiteParametersString(configs: Record<string, string>[]) {
   return configs
     .map((config) => {
-      // Note if we use the JSON.stringy name for all tests
-      // Some tests will error with `ENAMETOOLONG: name too long` as this is used for the directory name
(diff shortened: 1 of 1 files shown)


janpio:
`previewFeatures=referentialIntegrity` is added with a space before at the end of the string, but that is irrelevant for what I wanted to achieve 👍

janpio (approved):

PR #15787, opened 2022-10-14
feat(client): extended where

millsp:
Integration PR for https://github.com/prisma/prisma-engines/pull/3281

Tests will pass once the engine has been merged


packages/client/src/generation/TSClient/common.ts
```diff
@@ -416,6 +416,16 @@ type _Record<K extends keyof any, T> = {
   [P in K]: T;
 };
 
+// cause typescript not to expand types and preserve names
+type NoExpand<T> = T extends unknown ? T : never;
```
SevInf, line 420: For my own understanding, when `T extends unknown` would not be true?

packages/client/src/generation/TSClient/common.ts
```diff
@@ -416,6 +416,16 @@ type _Record<K extends keyof any, T> = {
...
+
+// this type assumes the passed object is entirely optional
+type AtLeast<O extends object, K extends string> = NoExpand<
```
SevInf, line 423: What would you think about pulling this type into separate file and writing a couple of test for it using `expectTypeOf` ?

packages/client/tests/functional/extended-where/findFirstOrThrow.ts
```diff
@@ -0,0 +1,44 @@
...
+  })
+
+  test('findFirstOrThrow with cursor 1 unique (PK)', async () => {
+    const data = await prisma.post.findFirstOrThrow({
+      cursor: {
+        id: vars.postId2,
+      },
+    })
+
+    expect(data?.id).toBe(vars.postId2)
```
SevInf, line 21:
```suggestion
    expect(data.id).toBe(vars.postId2)
```

`data` is non-nullabale in this case

packages/client/tests/functional/extended-where/findFirstOrThrow.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  test('findFirstOrThrow with cursor 2 uniques (PK & non-PK)', async () => {
+    const data = await prisma.post.findFirstOrThrow({
+      cursor: {
+        id: vars.postId2,
+        title: 'Hello World 2',
+      },
+    })
+
+    expect(data?.id).toBe(vars.postId2)
```
SevInf, line 32:
```suggestion
    expect(data.id).toBe(vars.postId2)
```

packages/client/tests/functional/extended-where/findFirstOrThrow.ts
```diff
@@ -0,0 +1,44 @@
...
+  })
+
+  test('findFirstOrThrow with cursor 1 unique (non-PK)', async () => {
+    const data = await prisma.post.findFirstOrThrow({
+      cursor: {
+        title: 'Hello World 2',
+      },
+    })
+
+    expect(data?.id).toBe(vars.postId2)
```
SevInf, line 42:
```suggestion
    expect(data.id).toBe(vars.postId2)
```

packages/client/tests/functional/extended-where/findUnique.ts
```diff
@@ -0,0 +1,38 @@
...
+declare let prisma: PrismaClient
+
+let vars: Awaited<ReturnType<typeof setup>>
+testMatrix.setupTestSuite(() => {
+  beforeAll(async () => {
+    vars = await setup(prisma)
+  })
+
+  test('findUnique with where 1 unique (PK)', async () => {
+    await prisma.user.findUnique({
```
SevInf, line 15: Shouldn't we assert something here?

SevInf (approved):
Couple of comments/questions, nothing blocking

packages/client/src/generation/TSClient/common.ts
millsp: It's always true

packages/client/tests/functional/extended-where/findUnique.ts
millsp: Yes, I think I should assert the results

packages/client/src/generation/TSClient/common.ts
millsp: It's completely artificial. TypeScript has a way (I cannot recall how they call this) to preserve type names and not expand them when wrapped inside more complex generic types. So here, i am improving DX by utilizing that behavior at our advantage. See with and without:

Without:
Screenshot from 2022-10-17 06-00-57

With:
Screenshot from 2022-10-17 06-00-38

SevInf: Thanks for the explanations!

millsp: Auto completion stays the same within the object.

PR #15797, opened 2022-10-14
[DRAFT] use keep alive in nodeFetch used by DataProxyEngine

mavilein:
It could be that we establish a new TCP connection everytime we make a request to the Data Proxy. This small branch is a quick way to build a package that we can use to validate this.


Jolg42:
Closing this as stale, we can reopen if needed

PR #15801, opened 2022-10-14
feat(client): Runtime for model extensions

SevInf:
Core new idea here: we introduce new way of building our proxies from
building blocks. For example:

```
createCompositeProxy(target, [
  addProperty('foo', () => 1),
  forwardTo(someObject),
  modelLayer(client, modelName)
])
```

This will build a proxy around around `target`, that will add property
`foo` on top, all properties from `someObject` and all model methods
for `modelName`. Not only it allows us to reuse some of the code,
previously duplicated between different proxies, it will also makes
implementation of model and client extensions pretty straightforward:
each extension is just another layer of the composite proxy.

### DIFF
--- a/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
+++ b/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
@@ -89,6 +89,6 @@ function mapKeysToLayers(layers: CompositeProxyLayer[]) {
 }
 
 function getOwnKeys(target: object, layerKeys: (string | symbol)[]) {
-  const set = new Set([...Object.keys(target), ...layerKeys])
+return [...new Set([...Object.keys(target), ...layerKeys])]
   return Array.from(set)
 }
--- a/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
+++ b/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
@@ -89,6 +89,5 @@ function mapKeysToLayers(layers: CompositeProxyLayer[]) {
 }
 
 function getOwnKeys(target: object, layerKeys: (string | symbol)[]) {
-return [...new Set([...Object.keys(target), ...layerKeys])]
-  return Array.from(set)
+  return [...new Set([...Object.keys(target), ...layerKeys])]
 }
--- a/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
+++ b/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
@@ -14,7 +14,7 @@ export interface CompositeProxyLayer<KeyType extends string | symbol = string | 
   getPropertyValue(key: KeyType): unknown
 
   /**
-   * Gets a descriptor for given property. If not implemented or undefined is returned, { enumerable: true, writeable: true, configurable: true}
+   * Gets a descriptor for given property. If not implemented or undefined is returned, { enumerable: true, writeable: true, configurable: true} is defaulted
    * is used
    * @param key
    */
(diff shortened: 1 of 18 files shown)


packages/client/tests/functional/extensions/enabled/model.ts
```diff
@@ -0,0 +1,211 @@
...
+            firstMethod,
+          },
+        },
+      })
+      .$extends({
+        model: {
+          user: {
+            // TODO: remove any once types are generated
+            secondMethod(this: any) {
+              this.firstMethod()
```
millsp, line 201: I think this needs discussion as it is not currently permitted by the proposal.

packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Returns a value for a property for a given key (one of the keys, returned
+   * from `getKeys`)
+   * @param key
+   */
+  getPropertyValue(key: KeyType): unknown
+
+  /**
+   * Gets a descriptor for given property. If not implemented or undefined is returned, { enumerable: true, writeable: true, configurable: true}
```
millsp, line 17:
```suggestion
   * Gets a descriptor for given property. If not implemented or undefined is returned, { enumerable: true, writeable: true, configurable: true} is defaulted
```

packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
```diff
@@ -0,0 +1,94 @@
...
+    const keys = layer.getKeys()
+    for (const key of keys) {
+      keysToLayerMap.set(key, layer)
+    }
+  }
+  return keysToLayerMap
+}
+
+function getOwnKeys(target: object, layerKeys: (string | symbol)[]) {
+  const set = new Set([...Object.keys(target), ...layerKeys])
```
millsp, line 92:
```suggestion
return [...new Set([...Object.keys(target), ...layerKeys])]
```

packages/client/src/runtime/core/compositeProxy/forwardTo.ts
```diff
@@ -0,0 +1,20 @@
+import type { CompositeProxyLayer } from './createCompositeProxy'
+
+/**
+ * Composite proxy layer that forwards all reads
+ * to provided object
+ *
+ * @param object
+ * @returns
+ */
+export function forwardTo(object: object): CompositeProxyLayer {
```
millsp, line 10: I would prefer a different name like objectToCompositeProxyLayer. Much longer put more descriptive imo.

packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -93,35 +110,38 @@ export function applyModel(client: Client, dmmfModelName: string) {
...
-  action: string,
-): action is Action | ClientOnlyModelAction {
-  if (isClientOnlyAction(action)) {
-    return isValidActionName(client, dmmfModelName, clientOnlyActions[action].wrappedAction)
-  }
-  return getOwnKeys(client, dmmfModelName).includes(action)
+  actionKeys.push('count')
+
+  const clientOnlyActionKeys = Object.keys(clientOnlyActions).filter((actionKey) =>
+    actionKeys.includes(clientOnlyActions[actionKey].wrappedAction),
```
millsp, line 130: What is the reason for this change?

millsp (approved):
Looks great and thorough. Let's sync over the next steps tomorrow.

packages/client/src/runtime/core/model/applyModel.ts
SevInf: It think it was a mistake during inital `*OrThrow` method implementation: they were never exposed as enumerable properties. Since now layer  does not gets called if it does not declare the specific key, error became more obvious.

packages/client/src/runtime/core/compositeProxy/forwardTo.ts
SevInf: Renamed to `addObjectProperties`: more descriptive than `forwardTo`, but also consistent with `addProperty` layer. Let me know if you still prefer different name.

millsp: That's better, thanks.

Issue #15583, opened 2022-09-28
Multi-schema support for `db pull`


tomhoule:
Within the `multiSchema` preview feature, `db pull` should by default introspect all schemas in the database.

Jolg42:
We need some TS tests in prisma/prisma repo.

Jolg42:
(working on it now)

pimeys:
Missing: MySQL.

Jolg42:
We'tracking the progress for MySQL for in https://github.com/prisma/prisma/issues/16943

PR #15832, opened 2022-10-17
test(db pull): multiSchema

Jolg42:
Related https://github.com/prisma/prisma/issues/15583

TODO:
- [x] same tests for SQL Server
- [x] sanitize foreign key names in snapshots


packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -21,6 +21,22 @@ const ctx = jestContext.new().add(jestConsoleContext()).add(jestProcessContext()
...
+  const regexp = new RegExp(/map: "PK__(.*)__(.*)"/)
+  const schemaRowsSanitized = schemaRows.map((row) => {
+    const match = regexp.exec(row)
+    if (match) {
+      row = row.replace(match[2], 'RANDOM_ID_SANITIZED')
+    }
+    return row
+  })
+  return schemaRowsSanitized.join('\n')
+}
```
jkomyno, line 38: I believe `match[2]` is the table's field name that appears in the primary key constraint name, but an explanatory comment would be useful here.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -485,11 +501,300 @@ describe('postgresql', () => {
...
+  })
+})
+
+describeIf(!process.env.TEST_SKIP_MSSQL)('sqlserver-multi-schema', () => {
+  if (process.env.CI) {
+    // to avoid timeouts on macOS
+    jest.setTimeout(80_000)
+  } else {
+    jest.setTimeout(20_000)
+  }
```
jkomyno, line 519: How about setting a single `jest.setTimeout(80_000)`?

jkomyno (changes requested):
The PR seems mostly fine, but please check the review comments.
I'd like another person with more context on multi-schema to give a quick glance at the snapshots and double-check everything's ok.

packages/migrate/src/__tests__/DbPull.test.ts
Jolg42: It's a copy paste from the other sql server describe and it works so I don't think we need to change it really

packages/migrate/src/__tests__/DbPull.test.ts
Jolg42: added

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -21,6 +21,25 @@ const ctx = jestContext.new().add(jestConsoleContext()).add(jestProcessContext()
 // To avoid the loading spinner locally
 process.env.CI = 'true'
 
+// describeIf is making eslint not happy about the names
+/* eslint-disable jest/no-identical-title */
+
+// We want to remove unique IDs to have stable snapshots
+// Example:
+// `PK__User__3213E83E450CDF1F` will be changed to `PK__User__RANDOM_ID_SANITIZED`
```
tomhoule, line 29: The way we deal with that in engines tests is to make the names explicit in our SQL migrations. I've had many bad experiences with test setups that post-process the output.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -151,20 +170,20 @@ describe('common/sqlite', () => {
...
+                  These models were enriched with \`@@map\` information taken from the previous Prisma schema.
+                  - Model "AwesomeUser"
+                  - Model "AwesomeNewPost"
+                  - Model "AwesomeProfile"
 
-      Run prisma generate to generate Prisma Client.
+                  Run prisma generate to generate Prisma Client.
 
-    `)
+            `)
```
tomhoule, line 186: What caused this diff? I see only new test code.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -485,11 +504,300 @@ describe('postgresql', () => {
...
+
+describeIf(!process.env.TEST_SKIP_MSSQL)('sqlserver-multi-schema', () => {
+  if (process.env.CI) {
+    // to avoid timeouts on macOS
+    jest.setTimeout(80_000)
+  } else {
+    jest.setTimeout(20_000)
+  }
+
+  const connectionString = process.env.TEST_MSSQL_URI || 'mssql://SA:Pr1sm4_Pr1sm4@localhost:1433/master'
```
tomhoule, line 524: Is the second branch supposed to persist on main, or just something for local testing?

tomhoule (approved):

packages/migrate/src/__tests__/DbPull.test.ts
Jolg42: Unfortunately, Jest snapshots do that, we do not know how to avoid this at the moment 😢

packages/migrate/src/__tests__/DbPull.test.ts
Jolg42: I think we should get rid of the fallbacks, some are still around but not needed since the local setup relies on `.envrc`

Jolg42: We could do that in a separate PR

PR #15840, opened 2022-10-17
refactor: fix some type safety issues

aqrln:
Closes: https://github.com/prisma/team-orm/issues/105

### DIFF
--- a/packages/client/src/runtime/query.ts
+++ b/packages/client/src/runtime/query.ts
@@ -321,7 +321,7 @@ ${fieldErrors.map((e) => this.printFieldError(e, missingItems, errorFormat === '
     return undefined
   }
 
-  protected printArgError = ({ error, path, id }: ArgError, hasMissingItems: boolean, minimal: boolean) => {
+  protected printArgError = ({ error, path }: ArgError, hasMissingItems: boolean, minimal: boolean) => {
     if (error.type === 'invalidName') {
       let str = `Unknown arg ${chalk.redBright(`\`${error.providedName}\``)} in ${chalk.bold(
         path.join('.'),
--- a/packages/client/src/runtime/query.ts
+++ b/packages/client/src/runtime/query.ts
@@ -774,7 +774,9 @@ ${indent(value.toString(), 2)}
   }
 }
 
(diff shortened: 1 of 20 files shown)


packages/client/src/runtime/query.ts
```diff
@@ -767,7 +765,9 @@ ${indent(value.toString(), 2)}
   }
 }
 
-export type ArgValue = string | boolean | number | undefined | Args | string[] | boolean[] | number[] | Args[] | null
+export type ArgValueScalar = string | boolean | number | Args
+
+export type ArgValue = ArgValueScalar | ArgValueScalar[] | undefined | null
```
aqrln, line 770: It's not obvious what this influences from the diff as the new type isn't used explicitly, but it fixed a compiler warning on line 684:

```ts
681    this.hasError =
682      Boolean(error) ||
683      (value instanceof Args ? value.hasInvalidArg : false) ||
684 ->   (Array.isArray(value) && value.some((v) => (v instanceof Args ? v.hasInvalidArg : false)))
                                              ^
                                         here |
```

TypeScript was able to infer the correct type of `value` after the `isArray` check, but it gave up on `v`.

Initially I had extracted `ArgValueScalar` to annotate `v` explicitly with it, but just extracting it and thus simplifying `ArgValue` turned out to be already enough for TypeScript to infer the correct type here.

aqrln, line 770: It's not obvious what this influences from the diff as the new type (`ArgValueScalar`) isn't used explicitly, but it fixed a compiler warning on line 684:

```ts
681    this.hasError =
682      Boolean(error) ||
683      (value instanceof Args ? value.hasInvalidArg : false) ||
684 ->   (Array.isArray(value) && value.some((v) => (v instanceof Args ? v.hasInvalidArg : false)))
                                              ^
                                         here |
```

TypeScript was able to infer the correct type of `value` after the `isArray` check, but it gave up on `v`.

Initially I had extracted `ArgValueScalar` to annotate `v` explicitly with it, but just extracting it and thus simplifying `ArgValue` turned out to be already enough for TypeScript to infer the correct type here.

PR #15843, opened 2022-10-17
chore(migrate): delete dead code in introspection engine

tomhoule:
Noticed while working on https://github.com/prisma/prisma/pull/15834


tomhoule:
We can delete the engines code when this is merged.

PR #15860, opened 2022-10-18
chore: Reduce flakiness of the tests

SevInf:


packages/client/tests/functional/_utils/setupFilesAfterEnv.ts
```diff
@@ -1,8 +1,11 @@
+import { faker } from '@faker-js/faker'
 import { toMatchInlineSnapshot, toMatchSnapshot } from 'jest-snapshot'
 import stripAnsi from 'strip-ansi'
 
 process.env.PRISMA_HIDE_PREVIEW_FLAG_WARNINGS = 'true'
 
+faker.seed([process.pid, Date.now()])
```
jkomyno, line 7: I'm curious, what's this for?

SevInf: I had a hypothesis that different process have different jest workers might accidentally get synchornized random generators (faker uses it's own implementation rather that `Math.random`) and that's why we occasionally get unique constraint violations for the test that creates a single record. So, this was an attempt to explicitly make process id a part of the seed of the PRNG. I don't think that was the real issue though, judging by the last run, so it will probably be reverted.

packages/client/tests/functional/_utils/waitFor.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Waits until provided callback stops throwing.
+ * If it does not happen within MAX_WAIT, throws last thrown error
+ *
+ * Simplified version of testing-library's waitFor that works without DOM
+ * https://testing-library.com/docs/dom-testing-library/api-async/#waitfor
+ * @param cb
+ */
+export async function waitFor(cb: () => void | Promise<void>) {
```
aqrln, line 13: oooh this will be super useful

aqrln (approved):

garrensmith (approved):

PR #15884, opened 2022-10-19
test(generator-handler): Attempt to improve flaky test

SevInf:
`generatorHandler` waits for 200ms and if generator have not failed in
this time considers it a success. This is results in quite flaky
behaviour on CI. For now, attempting to overcome it by introducing the
executable that fails faster. Ideally, the error handling logic of
`generatorHandler` needs changing: it should explicictly confirm
succesful startup.


SevInf:
Test in question was fixed in #15956

PR #15897, opened 2022-10-20
fix(client): middleware args consistency

### LINKED ISSUE #15644: Args are out of sync with the action type when creating prisma middeware
@vuki656 @millsp I'm curious how you are handling other types of aggregate queries in your middleware, as I'm sure you would also want to exclude soft deleted entries there as well. Why did you think that count needs it's own special case? I think this is actually a bad change that was made because the inputs in the parameters of the middleware function are only valid for an aggregate action `{"select":{"_count":{"select":{"_all":true}}}}`. If you try to reconstruct the prisma action from the model, argument, and parameter information, for example to try and implement row level security, it is now broken 

```
//this creates an invalid query because the args in param.args are not valid for params.action=count
//it produces this instead {"args":{"select":{"_count":{"select":{"_count":{"select":{"_all":true}}}}}},"dataPath":[],"runInTransaction":true,"action":"count","model":"User"}
const prismaAction = prisma[modelName][params.action](params.args);

const [, results] = await prisma.$transaction([
  prisma.$executeRawUnsafe(`SET LOCAL request.orgId = '${org}'`),
  prismaAction
]);
```

The only workaround is to set override this back to aggregate
```
if (params.action == "count") params.action="aggregate"
//this works now
const prismaAction = prisma[modelName][params.action](params.args);

const [, results] = await prisma.$transaction([
  prisma.$executeRawUnsafe(`SET LOCAL request.orgId = '${org}'`),
  prismaAction
]);
```

_Originally posted by @joe-giunti-kiefa in https://github.com/prisma/prisma/issues/12222#issuecomment-1265706520_

millsp:
Closes #15644

With this PR, we are fixing consistency issues in the shape of the arguments passed to middlewares. 

To put it shortly, if you use an aggregate query like count, groupBy, or aggregate, the args you get in middleware are not exactly the ones you pass.

The reason for that is that we’ve added some “sugar” on top of the TS interfaces for aggregates to improve the DX, in turn it means we are transforming a nicer input into something that is understood by the Query Engine. That’s what ends up being passed to the middleware.

The solution is to pass the initial user input to the middlewares, not the one we have pre-transformed for the Query Engine (that should always be a hidden detail).


packages/client/src/runtime/core/model/aggregates/groupBy.ts
```diff
@@ -8,15 +7,15 @@ import { desugarUserArgs as desugarUserArgsAggregate } from './aggregate'
...
  * handling for the `by` clause. The goal here is to desugar it into something
  * that is understood by the QE.
- * @param userArgs to transform
+ * @param args to transform
  * @returns
  */
-function desugarUserArgs(userArgs: UserArgs) {
-  const _userArgs = desugarUserArgsAggregate(userArgs)
+function desugarUserArgs(args?: UserArgs) {
+  const _userArgs = desugarUserArgsAggregate(args ?? {})
```
millsp, line 14: Nit: I think I can revert this small name change

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return next(params)
+      })
+
+      await prisma.resource.create({ data: {} })
+    })
+
+    test('middleware with delete', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 106:
```suggestion
    test('middleware with delete', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return next(params)
+      })
+
+      await prisma.resource.findUniqueOrThrow({ where: { id: '0' } }).catch(() => {})
+    })
+
+    test('middleware with create', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 94:
```suggestion
    test('middleware with create', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return next(params)
+      })
+
+      await prisma.resource.findFirstOrThrow({ take: 1 }).catch(() => {})
+    })
+
+    test('middleware with findUniqueOrThrow', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 82:
```suggestion
    test('middleware with findUniqueOrThrow', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+
+        return next(params)
+      })
+
+      await prisma.resource.findUnique({ where: { id: '0' } })
+    })
+
+    test('middleware with findFirstOrThrow', async () => {
+      prisma.$use((params, next) => {
+        if (params.action === ('findFirstOrThrow' as string)) {
```
danstarns, line 71:
```suggestion
    test('middleware with findFirstOrThrow', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
        if (params.action === ('findFirstOrThrow' as string)) {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return next(params)
+      })
+
+      await prisma.resource.findFirst({ take: 1 })
+    })
+
+    test('middleware with findUnique', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 58:
```suggestion
    test('middleware with findUnique', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return next(params)
+      })
+
+      await prisma.resource.groupBy({ by: ['id'], orderBy: {} })
+    })
+
+    test('middleware with findFirst', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 46:
```suggestion
    test('middleware with findFirst', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return next(params)
+      })
+
+      await prisma.resource.aggregate({ skip: 1, _count: true })
+    })
+
+    test('middleware with groupBy', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 34:
```suggestion
    test('middleware with groupBy', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return next(params)
+      })
+
+      await prisma.resource.count({ skip: 1 })
+    })
+
+    test('middleware with aggregate', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 22:
```suggestion
    test('middleware with aggregate', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(
+  (suiteConfig, suiteMeta) => {
+    test('middleware with count', async () => {
+      prisma.$use((params, next) => {
```
danstarns, line 10:
```suggestion
    test('middleware with count', async () => {
      expect.assertions(1)
      prisma.$use((params, next) => {
```

danstarns: Let's use `expect.assertions` to ensure that the assertion happens in the callback.

danstarns (approved):

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
```diff
@@ -0,0 +1,183 @@
+import testMatrix from './_matrix'
```
danstarns, line 1: For these tests, you may be able to use https://jestjs.io/docs/api#testeachtablename-fn-timeout

I think it would look something like: 


```ts
test.each(['create', 'read'])('middlewear with %s', (method) => {
  prisma.$use((params, next) => {
    if (params.action === method) {
      expect(params.args).toStrictEqual({ data: {} })
    }

    return next(params)
  })

  await prisma.resource[method]({ data: {} })
})
```

packages/client/src/runtime/core/model/aggregates/groupBy.ts
```diff
@@ -8,15 +7,15 @@ import { desugarUserArgs as desugarUserArgsAggregate } from './aggregate'
  * It reuses the logic from the `.aggregate` shorthand and adds additional
  * handling for the `by` clause. The goal here is to desugar it into something
  * that is understood by the QE.
- * @param userArgs to transform
+ * @param args to transform
  * @returns
  */
-function desugarUserArgs(userArgs: UserArgs) {
-  const _userArgs = desugarUserArgsAggregate(userArgs)
+function desugarUserArgs(args?: UserArgs) {
```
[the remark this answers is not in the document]
millsp: Thanks, changed it in the relevant places

packages/client/tests/functional/issues/15644-middleware-arg-types/tests.ts
millsp: Thanks I'll keep that in mind for next time!

PR #15909, opened 2022-10-20
refactor(client): Simplify conditional in BinaryEngine.ts

miguelff:
Evaluation of `(a || b) && a = a` provided that b does not have side effects. Thus the conditional expression replaced is equivalent to the one that existed before

Stumbled across this by coincidence.


SevInf (approved):

SevInf:
Thank you!

PR #15962, opened 2022-10-24
test(client): Working on `mongo-urls` flaky test

SevInf:


.github/workflows/test.yml
```diff
@@ -206,7 +206,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        queryEngine: ['library', 'binary']
+        queryEngine: ['binary']
```
jkomyno, line 209: Isn't the removal of `library` here a bit scary?

aqrln: the test only fails on `binary`, `library` is temporarily removed to make iterating on code and debugging easier

PR #15989, opened 2022-10-25
chore: types for watch mode

millsp:
While our internal `pnpm run watch` is rather useful and fast, one of its limitations is that it does not generate types the types necessary for the `runtime`. The current workaround is to generate the types before hand. One downside is that this takes a bit of time. The other downside is that we cannot rely on runtime types to be up-to-date since they have to be bundled. For local development it would be great if that could be seamless.

What does this solve? I always work in locally linked projects and repros, and *I am planning to move many of the types generated via strings* into a *real* ts file. I don't wan't to rebuild/wait for bundling those, changes must be instant, and updated types must be instantly picked up by ts in my linked repros.

In essence, this will reduce the feedback loop when developing type utilities while also providing an avenue for moving types stored in strings to real ts files that will benefit from ts support.


package.json
```diff
@@ -55,6 +57,7 @@
     "eslint-plugin-simple-import-sort": "8.0.0",
     "eventemitter3": "4.0.7",
     "execa": "5.1.1",
+    "fs-extra": "10.1.0",
```
SevInf, line 60: Why do we need `fs-extra` here? The only usage I see is `writeFileSync`, which is also available in builtin module.

helpers/compile/plugins/tscPlugin.ts
```diff
@@ -1,23 +1,101 @@
...
+        await run(`tsc --project ${options.tsconfig} --paths null`)
+      }
+
+      // we bundle types if we also bundle the entry point and types exists
+      if (options.bundle === true && options.entryPoints?.[0].endsWith('.ts')) {
+        const tsconfig = require(`${process.cwd()}/${options.tsconfig}`) // tsconfig
+        const typeOutDir = tsconfig?.compilerOptions?.outDir ?? '.' // type out dir
+        const entryPoint = options.entryPoints?.[0].slice(0, -3) // drop ".ts"
+        const typeOutPath = `${typeOutDir}/${entryPoint.slice(4)}` // drop "src/"
+        const bundlePath = options.outfile!.slice(0, -3) // drop ".ts"
```
SevInf, line 87:
```suggestion
        const typeOutDir = tsconfig?.compilerOptions?.outDir ?? '.' // type out dir
        const entryPoint = options.entryPoints?.[0].replace(/\.ts$/, '')
        const typeOutPath = `${typeOutDir}/${entryPoint.replace(/^src\//, '')}` // drop "src/"
        const bundlePath = options.outfile!.replace(/\.ts$/, '')
```

Do not insist on this change, but in my opinion that makes intention a little bit clearer and protects us in case any of the parts we replace change. Feel free to ignore if you disagree.

SevInf (approved):
LGTM with few optional comments

helpers/compile/plugins/tscPlugin.ts
millsp: I think that's better indeed, and removes the need for comments.

package.json
millsp: Just a habit of always using fs-extra.

PR #16005, opened 2022-10-26
feat(client): enable ITX support for DataProxy

### LINKED ISSUE #16177: Support interactive transactions in the Data Proxy
## Problem

At the moment, it is not possible to use Interactive Transaction when using the Data Proxy.

## Suggested solution

Add Interactive Transaction support to the Data Proxy.

aqrln:
New PR that supersedes https://github.com/prisma/prisma/pull/15694 because GitHub doesn't allow to change the branch for an existing PR, and we need integration releases for new commits here.

**Original PR by @danstarns**: https://github.com/prisma/prisma/pull/15694
**Mini-Proxy PR**: https://github.com/prisma/mini-proxy/pull/36
**RFC (internal)**:  https://www.notion.so/prismaio/iTX-between-DP-and-the-Client-91c133e5b60d43dbbfa409fde2b430b3
**Tests for Data Proxy**: https://github.com/prisma/itx-pdp-manual-tests

Closes https://github.com/prisma/client-planning/issues/92
Closes https://github.com/prisma/prisma/issues/16177


packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -153,76 +167,119 @@ export class DataProxyEngine extends Engine {
...
+        const data = await response.json()
 
-      if (e instanceof SchemaMissingError) {
-        await this.uploadSchema()
-        throw new ForcedRetryError({
-          clientVersion: this.clientVersion,
-          cause: e,
-        })
-      }
+        // TODO: headers contain `x-elapsed` and it needs to be returned
```
aqrln, line 205: this is out of scope of this pr and not related to itx — this is just something i noticed

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -130,18 +140,22 @@ export class DataProxyEngine extends Engine {
...
-  request<T>(query: string, headers: Record<string, string>, attempt = 0) {
+  request<T>(
+    query: string,
+    headers: QueryEngineRequestHeaders = {},
+    transaction?: InteractiveTransactionOptions<DataProxyTxInfoPayload>,
+  ): Promise<QueryEngineResult<T>> {
     this.logEmitter.emit('query', { query })
 
-    return this.requestInternal<T>({ query, variables: {} }, headers, attempt)
+    // TODO: `elapsed`?
```
aqrln, line 150: out of scope of this pr — just a todo for future

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -153,76 +167,119 @@ export class DataProxyEngine extends Engine {
       isolationLevel: transaction?.isolationLevel,
     }
 
-    const { batchResult } = await this.requestInternal<T>(body, headers, attempt)
+    const { batchResult } = await this.requestInternal<T, true>(body, headers)
 
+    // TODO: add elapsed to each result similar to BinaryEngine
+    // also check that the error handling is correct for batch
```
aqrln, line 173: also a general comment about `DataProxyEngine`, not related to itx

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -70,6 +70,7 @@ function getRequestInfo(request: Request) {
 
   return {
     batchTransaction: transaction?.kind === 'batch' ? transaction : undefined,
+    interactiveTransaction: transaction?.kind === 'itx' ? transaction : undefined,
```
SevInf, line 73: If we need both now, why don't just pass `transaction` as it is and differentiate by `kind` field? That was my original intention with this type.

packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -11,221 +11,221 @@ declare let newPrismaClient: NewPrismaClient<typeof PrismaClient>
...
 
-        await delay(6000)
-      })
-
-      await expect(result).rejects.toMatchObject({
-        message: expect.stringContaining('Transaction API error: Transaction already closed'),
-        code: 'P2028',
-        clientVersion: '0.0.0',
-      })
+        await new Promise((res) => setTimeout(res, 600))
```
SevInf, line 82:
```suggestion
        await delay(600)
```

packages/engine-core/src/common/utils/runtimeHeadersToHttpHeaders.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    if (runtimeHeaderKey === 'transactionId') {
+      httpHeaderKey = 'X-transaction-id'
+    }
+
+    // if header key isn't changed, a copy happens
+    acc[httpHeaderKey] = headers[runtimeHeaderKey]
+
+    return acc
+  }, {} as Record<string, string | undefined>)
```
SevInf, line 20:
```suggestion
  if (headers.transactionId) {
      const { transactionId, ...httpHeaders } = headers
      httpHeaders['X-transaction-id'] = transactionId
      return httpHeaders
  }
  return headers
```
No need to traverse whole object for one rename.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -153,76 +167,119 @@ export class DataProxyEngine extends Engine {
...
+    return this.withRetry({
+      actionGerund: 'querying',
+      callback: async ({ logHttpCall }) => {
+        const url = itx ? `${itx.payload.endpoint}/graphql` : await this.url('graphql')
+
+        logHttpCall(url)
+
+        const response = await request(url, {
+          method: 'POST',
+          headers: { ...runtimeHeadersToHttpHeaders(headers), ...this.headers },
```
SevInf, line 191:
```suggestion
          headers: runtimeHeadersToHttpHeaders(headers)
```

I think `runtimeHeadersToHttpHeaders` preserves all original headers, except `transactionId`?

packages/engine-core/src/common/utils/runtimeHeadersToHttpHeaders.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Takes runtime data headers and turns it into QE HTTP headers
+ * @param headers to transform
+ * @returns
+ */
+export function runtimeHeadersToHttpHeaders(headers: QueryEngineRequestHeaders): Record<string, string | undefined> {
+  return Object.keys(headers).reduce((acc, runtimeHeaderKey) => {
+    let httpHeaderKey = runtimeHeaderKey
+
+    if (runtimeHeaderKey === 'transactionId') {
+      httpHeaderKey = 'X-transaction-id'
```
danstarns, line 13: Do we still need this header?

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -318,4 +375,51 @@ export class DataProxyEngine extends Engine {
       clientVersion: this.clientVersion,
     })
   }
+
+  private async withRetry<T>(args: {
+    callback: (api: { logHttpCall: (url: string) => void }) => Promise<T>
+    actionGerund: string
+  }): Promise<T> {
+    for (let attempt = 0; ; attempt++) {
```
danstarns, line 383: Will the for loop with await be completely blocking for all other operations in the runtime?

danstarns (approved):

SevInf (approved):
Couple of comments, suggestions, but nothing really blocking. Great job @aqrln and @danstarns!

packages/engine-core/src/data-proxy/DataProxyEngine.ts
aqrln: `headers` comes from the function argument and isn't a superset of `this.headers` (which is basically just the `authorization` header), so we need both.

Technically we shouldn't even need `runtimeHeadersToHttpHeaders(headers)` here anymore with the recent Data Proxy changes and could just make it `headers: this.headers` and move `runtimeHeadersToHttpHeaders` back to `BinaryEngine` as a private function, but let's better do that after the release probably. There's no harm in sending those headers, but we haven't tested not sending them yet and I'd avoid doing that change last minute. And they would have to come back when we support tracing with Data Proxy anyway (although we could filter out `transactionId` specifically).

packages/engine-core/src/data-proxy/DataProxyEngine.ts
aqrln: No, it won't, it's basically like a coroutine

packages/engine-core/src/common/utils/runtimeHeadersToHttpHeaders.ts
aqrln: For Data Proxy — not really, only for binary engine (this function is still used by both, it was actually extracted from `BinaryEngine.ts`). There's no harm in sending it for Data Proxy, but optimally we'd want to get rid of it because it should be ignored by the server. We'll change this after the release.

PR #16019, opened 2022-10-27
feat(client): client extensions types

millsp:


packages/client/src/generation/TSClient/Input.ts
```diff
@@ -1,4 +1,5 @@
 import indent from 'indent-string'
+import { cpuUsage } from 'process'
```
SevInf, line 2: leftover from some testing?

packages/client/src/generation/TSClient/Input.ts
```diff
@@ -178,7 +183,7 @@ export type ${this.getTypeName()} = ${wrapWithAtLeast(body, type)}`
     if (this.genericsInfo.inputTypeNeedsGenericModelArg(this.type)) {
       return `${this.type.name}<$PrismaModel = never>`
     }
-    return this.type.name
+    return `${this.type.name}<$PrismaModel = never>`
```
SevInf, line 186: I think now `$PrismaModel=never` would be added in all cases, regardless of the check above? Is it intended?

packages/client/src/generation/TSClient/Input.ts
```diff
@@ -56,6 +57,10 @@ function stringifyInputType(
     return 'null'
   }
 
+  if (typeof type === 'string' && (type.endsWith('Select') || type?.endsWith('Include'))) {
```
SevInf, line 60: Question: if there possibility of collision here? Will we be safe if one of the model names ends with `Input`?

SevInf: Also, do `Select` types pass through `Input` class in general? From `Model.ts` it seems like they don't

millsp: > Question: if there possibility of collision here? Will we be safe if one of the model names ends with Input?

I'm not sure exactly, I would say no since it has to end with `Select/Include`. What concerns you here?

millsp: > Also, do Select types pass through Input class in general? From Model.ts it seems like they don't

Through InputField, yes

millsp:
PR reviewed together with Serhii.

packages/client/src/generation/TSClient/Input.ts
SevInf: Sorry, I meant to say "what if one of the model names ends with `Select`"

Issue #6927, opened 2021-05-04
Client tests for Vitess, with `relationMode=prisma`


janpio:
With Vitess being tested in `prisma-engines`, we can now also run our MySQL like (MySQL, MariaDB) tests in this repository on a `vttestserver` Docker container (or 2 actually, one for MySQL 57 and one for 80) to make sure everything works as expected here.

- Docker Compose: https://github.com/prisma/prisma-engines/blob/6d7dfe95742ea6d5f9988d8fc9d81ef9be2ebb49/docker-compose.yml#L147-L167

Jolg42:
Note: The [hacks.patch](https://github.com/planetscale/vitess-framework-testing/blob/master/frameworks/javascript/prisma/hacks.patch?rgh-link-date=2021-05-04T17%3A53%3A15Z) file is now 404.
[From Slack](https://prisma-company.slack.com/archives/C01CUJDJ7UN/p1660663890521279): It was last modified in June of 2021, and looked like this
```patch
diff --git a/src/packages/sdk/src/convertCredentials.ts b/src/packages/sdk/src/convertCredentials.ts
index 3bcc844f..45a0b11b 100644
--- a/src/packages/sdk/src/convertCredentials.ts
+++ b/src/packages/sdk/src/convertCredentials.ts
@@ -140,7 +140,7 @@ export function uriToCredentials(
     database,
     schema: schema || defaultSchema,
     uri: connectionString,
-    ssl: Boolean(uri.searchParams.get('sslmode')),
+    ssl: (uri.searchParams.get('sslmode') == 'true'),
     socket: socket || undefined,
     extraFields,
   }
diff --git a/src/packages/integration-tests/src/__tests__/integration/mariadb/__database.ts b/src/packages/integration-tests/src/__tests__/integration/mariadb/__database.ts
index f143c928..3edcbac1 100644
--- a/src/packages/integration-tests/src/__tests__/integration/mariadb/__database.ts
+++ b/src/packages/integration-tests/src/__tests__/integration/mariadb/__database.ts
@@ -15,15 +15,22 @@ export const database = {
       port: credentials.port,
       user: credentials.user,
       password: credentials.password,
+      ssl: credentials.ssl,
+      allowPublicKeyRetrieval: true,
       multipleStatements: true,
     })
   },
   beforeEach: async (db, sqlScenario, ctx) => {
-    const sqlUp = `
-    DROP DATABASE IF EXISTS ${ctx.id};
[...]

Issue #10829, opened 2021-12-22
Error: Unique constraint failed on the (not available)


janpio:
### Bug description

PlanetScale/Vitess does not provide the column name. But additionally to that not showing, we are also missing the additional `fields` or `constraint` string that is present in the error, helping the user to figure out what might be wrong.

### How to reproduce

Cause unique constraint error on PlanetScale/Vitess

### Expected behavior

More helpful error message.

### Environment & setup

- OS: 
- Database: 
- Node.js version: 

### Prisma Version

3.7.0

pantharshit00:
I can confirm with `planetscale`.

jonahallibone:
+1 can also confirm

rcornacchia:
Also experiencing this

yujong-lee:
Also experienceing this with pscale.

dyeoman2:
+1

Saleh-Albrahim:
Also experienceing this with pscale.

dpnick:
Anyone found a solution to this using pscale ?

dyeoman2:
> Anyone found a solution to this using pscale ?

After checking fields one at a time, I found out that pscale wanted the createdAt and updatedAt dates provided in my create request even though I had them set to automatically populate in my Prisma schema. Hope this narrows things down for you.

janpio:
This issue here is about the error **message**, which can **not provide the actual column name** as it is not provided by PlanetScale/Vitess, as described in the original issue. There is no real solution than to fix Vitess or implement some workaround that gets the column name from the original query.

If you have any code that causes the `Unique constraint failed ...` error message although it is unexpected, this issue here is not the correct one. Please open a new issue and describe your problem. (E.g. your problem @dyeoman2 would be a good new bug report so we can look into how this can happen)

ansh:
Also experiencing this with Planetscale

dovydasvg:
Same problem with Planetscale

janpio:
See my previous message @Ansh and @dovydasvg: https://github.com/prisma/prisma/issues/10829#issuecomment-1065842833

poutingemoji:
+1

vincentdchn:
+1

bjcgg:
Yeah, I have the same issue as well.

ElyasAsmad:
I am currently having the same issue while using the update request.

My workaround for this issue:
I used prisma.$executeRaw and everything works as expected.

rutikwankhade:
[...]

PR #16051, opened 2022-10-28
test(client): add `vitess` flavor in relationMode test suite

jkomyno:
This PR:
- adds support for `vitess-8` (using the Docker image `vitess/vttestserver:mysql80`)
- introduces the concept of `ProviderFlavor`, a superset of `Providers` that allows distinguishing slightly different variants of providers. E.g., the `vitess-8` flavor uses `mysql` as the `provider` in the Prisma schema, but its url is mapped to the `TEST_FUNCTIONAL_VITESS_8_URI` env variable rather than `TEST_FUNCTIONAL_MYSQL_URI`.

**Note**: I have noticed Vitess is particularly slow. Maybe there's something sketchy on my local machine and it's faster on the CI.

Related https://github.com/prisma/prisma/issues/6927.


packages/client/tests/functional/relationMode/tests_1-to-1.ts
```diff
@@ -601,6 +603,7 @@ testMatrix.setupTestSuite(
                           [Providers.MYSQL]: 'Unique constraint failed on the constraint: `ProfileOneToOne_userId_key`',
                           [Providers.SQLSERVER]: 'Unique constraint failed on the constraint: `dbo.ProfileOneToOne`',
                           [Providers.SQLITE]: 'Unique constraint failed on the fields: (`userId`)',
+                          [ProviderFlavors.VITESS_8]: 'Unique constraint failed on the (not available)',
```
jkomyno, line 606: Btw, notice that Prisma error messages on Vitess are somewhat lacking compared to the other providers.

.buildkite/publish/docker-compose.yml
```diff
@@ -125,6 +127,20 @@ services:
     ports:
       - '3307:3306'
 
+  vitess-8:
+    image: vitess/vttestserver:mysql80@sha256:9412e3d51bde38e09c3039090b5c68808e299579f12c79178a4ec316f7831889
```
Jolg42, line 131: Note: it's the same version we test in prisma-engines

packages/client/tests/functional/relationMode/tests_1-to-1.ts
janpio: https://github.com/prisma/prisma/issues/10829

packages/client/tests/functional/_utils/relationMode/computeMatrix.ts
```diff
@@ -42,9 +46,10 @@ export function computeMatrix({ relationMode, providersDenyList }: ComputeMatrix
   //
   // We skip these combinations in the matrix (= filtering them out)
 
-  const referentialActionsDenylistByProvider = {
+  const referentialActionsDenylistByProviderFlavor = {
     foreignKeys: {
       [Providers.SQLSERVER]: ['Restrict'],
+      [ProviderFlavors.VITESS_8]: referentialActionsBase, // no action is executed for Vitess & relationMode="foreignKeys"
```
Jolg42, line 52: It's confusing because it can be read as `NO ACTION` which is an action 😄 
```suggestion
      [ProviderFlavors.VITESS_8]: referentialActionsBase, // skip all actions for Vitess & relationMode="foreignKeys"
```

Jolg42: But Why do we need to skip all?

Jolg42: I improved the comment

Jolg42 (approved):
Looks good!

Note: only the `relationMode` tests are also tested on vitess.

janpio:
> Note: I have noticed Vitess is particularly slow. Maybe there's something sketchy on my local machine and it's faster on the CI.

What is the impact on CI/test run time?

jkomyno:
> > Note: I have noticed Vitess is particularly slow. Maybe there's something sketchy on my local machine and it's faster on the CI.
> 
> What is the impact on CI/test run time?

You can see that `CI / client-relationMode (prisma, ubuntu-latest, 16)` got canceled after 35m. I will investigate ways of reducing the time needed for that.

docker/docker-compose.yml
```diff
@@ -38,14 +38,21 @@ services:
...
       PORT: '33804' # unused in testing, but required by vttestserver
-      KEYSPACES: 'keyspace' # unused in testing, but required by vttestserver
+      KEYSPACES: 'unsharded' # unused in testing, but required by vttestserver
       NUM_SHARDS: '1' # unused in testing, but required by vttestserver
-
       MYSQL_BIND_HOST: '0.0.0.0'
       FOREIGN_KEY_MODE: 'disallow'
       VT_DIALECT: 'mysql80'
+      PLANNER_VERSION: 'gen4fallback'
+      TOPOLOGY_FLAGS: '-shutdown_grace_period 2'
```
Jolg42, line 55: What are these settings?

jkomyno: - `PLANNER_VERSION`: it uses the most recent planner for Vitess, allowing a fallback to the previous planner version in case the newest fails (see the [docs](https://vitess.io/docs/15.0/get-started/vttestserver-docker-image/#environment-variables))
- `TOPOLOGY_FLAGS`: added `-shutdown_grace_period 2` as suggested in [this comment](https://github.com/vitessio/vitess/issues/9936#issuecomment-1074498933), using an environment variable shown in [here](https://vitess.io/docs/13.0/reference/programs/vttablet/#managed-mysql) (which, as far as I can tell, is overridable)

.github/workflows/test.yml
```diff
@@ -253,7 +258,7 @@ jobs:
         run: |
           echo "PRISMA_CLIENT_ENGINE_TYPE=${{ matrix.queryEngine }}" >> $GITHUB_ENV
 
-      - run: docker-compose -f docker/docker-compose.yml up --detach postgres postgres_isolated mysql mysql_isolated mssql mongo cockroachdb
+      - run: docker-compose -f docker/docker-compose.yml up --detach postgres postgres_isolated mysql mysql_isolated mssql mongo cockroachdb vitess-8
```
janpio, line 261: Is this needed?

.github/workflows/test.yml
```diff
@@ -310,7 +315,7 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-      - run: docker-compose -f docker/docker-compose.yml up --detach postgres mysql mssql mongo cockroachdb
+      - run: docker-compose -f docker/docker-compose.yml up --detach postgres mysql mssql mongo cockroachdb vitess-8
```
janpio, line 318: Is this needed?

docker/docker-compose.yml
Jolg42: If these don't change anything (does it make the test faster or anything?), maybe we could remove them. Going with the defaults feels better, I think.

.github/workflows/test.yml
Jolg42: I don't think it's needed now because we have no tests in client functional tests that test on vitess. But as soon as we add one test, it will be needed.

.github/workflows/test.yml
```diff
@@ -1033,7 +1040,7 @@ jobs:
           CI: true
           SKIP_GIT: true
           GITHUB_CONTEXT: ${{ toJson(github) }}
-          NODE_OPTIONS: '--max-old-space-size=8096'
+          NODE_OPTIONS: '--max-old-space-size=10240' # allow Node.js to allocate at most 10gb of heap
```
Jolg42, line 1043: This is about the `CLIENT (memory tests)` job, so, it looks unrelated to the current PR. What is it solving?

jkomyno: Random flakiness on Windows that was happening on this PR (although not caused by this PR)

docker/docker-compose.yml
jkomyno: It was the latest attempt to make Vitess faster. Removing them now, let's see whether the test suite stays within the new timeout boundaries (60 mins)

.github/workflows/test.yml
jkomyno: So @Jolg42, is the `relationMode` functional test suite excluded from the rest of the client functional tests unless `--relation-mode-tests-only` is specified? Or is it run regardless, but with the implicit `relationMode = "foreignKeys"` setting?

Jolg42: `relationMode` tests are always ignored, unless `--relation-mode-tests-only` is passed, which we only do in the client relationMode job on GitHub Actions

.github/workflows/test.yml
jkomyno: No, removed.

.github/workflows/test.yml
jkomyno: Removed.

.buildkite/test/docker-compose.14.yml
```diff
@@ -64,6 +65,7 @@ services:
       - mongodb_migrate
       - mongodb_migrate_seed
       - cockroachdb
+      - vitess-8
```
Jolg42, line 68:
```suggestion
# TODO: uncomment when we test vitess in client functional test suite
# Currently we only test `relationMode` tests in GitHub Actions
#      - vitess-8
```

.buildkite/publish/docker-compose.yml
```diff
@@ -74,6 +75,7 @@ services:
       - mongodb_migrate
       - mongodb_migrate_seed
       - cockroachdb
+      - vitess-8
```
Jolg42, line 78:
```suggestion
# TODO: uncomment when we test vitess in client functional test suite
# Currently we only test `relationMode` tests in GitHub Actions
#      - vitess-8
```

PR #16079, opened 2022-10-31
test(client): native/atomic upserts

danstarns:
Integrates: https://github.com/prisma/prisma-engines/pull/3257


danstarns:
Currently failing due to https://github.com/prisma/prisma-engines/pull/3257#issuecomment-1297699932

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
```diff
@@ -0,0 +1,241 @@
...
+      await client.user.upsert({
+        where: {
+          name,
+        },
+        create: {
+          name,
+        },
+        update: {
+          name,
+          // Because this is a nested mutation
```
janpio, line 66: Is this the only possible nested mutation? Should we maybe test "all" of them?

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
```diff
@@ -0,0 +1,241 @@
...
+        create: {
+          // Because the the where 'name' is 'equal' to the create 'name'
+          name,
+        },
+        update: {
+          name: name + '1',
+        },
+      })
+      doesSQLContainOnConflict = (await sqlPromise).includes('ON CONFLICT')
+      expect(doesSQLContainOnConflict).toEqual(true)
```
janpio, line 198: Not directly related to the two lines I comment on here:

Do we have a simple way to add a snapshot of the full SQL query being executed for all these? I think it would be valuable to surface the full query here. (Of is our `faker` usage really dynamic and hence this would need to be sanitized?)

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
```diff
@@ -0,0 +1,241 @@
...
+    })
+  },
+  {
+    optOut: {
+      from: ['mongodb', 'mysql', 'sqlserver'],
+      reason: 'Other providers do not support native INSERT ... ON CONFLICT SET .. WHERE',
```
janpio, line 237: Do we have normal tests for these other providers that just make sure that `upsert` works as it should?
Might it be worth adding test functionality here as well that tests the actual "update (if exists already) or insert (if not)" functionality of these upserts?

packages/client/tests/functional/methods/upsert/native-atomic/prisma/_schema.ts
```diff
@@ -0,0 +1,28 @@
...
+      id             ${idForProvider(provider)}
+      name           String @unique
+      posts          Post[]
+    }
+
+    model Post {
+      id              ${idForProvider(provider)}
+      title           String @unique
+      author          User?     @relation(fields: [authorId], references: [id])
+      authorId        String
```
garrensmith, line 25: Could you add tests for compound uniques and ids

```
@@unique([field1, field2])
@@id([id1, id2](
```

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
danstarns: I decided not to snapshot the query because it's different for each provider, and so you then need to switch the query or dump it into a separate file 'miles' away from the test itself. I think if we want to be snapshotting the entire SQL we should look more into https://github.com/prisma/prisma/pull/13981 .

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
danstarns: Added simple upsert test's

SevInf (approved):

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
garrensmith: I've added a few more different operations.

Issue #16064, opened 2022-10-31
Prisma Client: logging not done in transactions


volkerrichert:
### Bug description

I'l logging my queries using the `prisma.$on`-method. But no logs will be written within an (interactive!) transaction.

### How to reproduce

- setup a simple prisma client
- register `$on()` to log queries
- create a query / update within this transaction

### Expected behavior

Logging should be the same.

### Prisma information
```prisma
generator client {
  provider = "prisma-client-js"
  previewFeatures = ["interactiveTransactions", "fieldReference"]
}

datasource db {
  provider = "mysql"
  url      = env("DATABASE_URL")
}

model Device {
  imei         String @id @db.VarChar(15)
  eId          String? @db.Char(32)
  ban          String @db.VarChar(20)

  brand        String @db.VarChar(60)
  model        String @db.VarChar(60)

  owner        String? @db.VarChar(60)
  email        String? @db.VarChar(60)

  subscriber   Subscriber? @relation(fields: [serialNumber], references: [serialNumber])
  serialNumber String?     @unique @db.VarChar(25)
  serialNumberUnchecked String? @db.VarChar(25)

  source    deviceSource
  createdAt DateTime     @default(now())
  changedAt DateTime     @updatedAt
  @@unique([ban, serialNumberUnchecked])
}

model Ban {
  id            String   @id @default(cuid()) @db.Char(32)
  toBeSynced    Int?
  synced        Int?
  syncStartedAt DateTime?
  subscribers   Subscriber[]

  createdAt  DateTime @default(now())
  changedAt  DateTime @updatedAt
}

model Subscriber {
  banId                  String           @db.VarChar(20) @default("")
  serialNumber           String           @unique @db.VarChar(20)
  subscriberNumber       String           @db.VarChar(20)
  productType            String           @db.VarChar(20)
  subscriberStatus       subscriberStatus
  simType                simType
  eSimStatus             eSimStatus
  pushStatus             pushStatus       @default(UNKNOWN)
  tariffName             String           @db.VarChar(60)

  device                 Device?
  ban                    Ban              @relation(fields: [banId], references: [id], onDelete: Cascade)

  createdAt              DateTime         @default(now())
  changedAt              DateTime         @updatedAt

  @@id([banId, serialNumber])
  @@index([banId, subscriberNumber])
}

model BulkPush {
  id         String         @default(cuid()) @db.Char(32) @id
  ban        String         @db.VarChar(20)

[...]

PR #16080, opened 2022-10-31
test(client): Logging queries inside a interactive transaction

danstarns:
Related: https://github.com/prisma/prisma/issues/16064


packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -762,6 +763,53 @@ testMatrix.setupTestSuite(
...
+        })
+
+        await tx.user.findFirst({
+          where: {
+            id: user.id,
+          },
+        })
+      })
+
+      const queries = await queriesPromise
```
danstarns, line 808: This currently times out because no query events are emitted thus causing the promise to never resolve.

packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -762,6 +763,53 @@ testMatrix.setupTestSuite(
...
+            entries.push(query)
+
+            if (query.includes('COMMIT')) {
+              resolve(entries)
+            }
+          })
+        })
+      })()
+
+      await prisma.$transaction(async (tx) => {
```
SevInf, line 794:
```suggestion
      await client.$transaction(async (tx) => {
```

garrensmith: I've simplified this test in a new commit

aqrln:
I'm currently going through the old GH notifications that I missed, do I understand it correctly that the referenced issue was not actually broken, we just want to add more test coverage here and convert some old tests to the new test setup? Does anything need to happen here except rebasing, reviewing and merging?

danstarns:
> I'm currently going through the old GH notifications that I missed, do I understand it correctly that the referenced issue was not actually broken, we just want to add more test coverage here and convert some old tests to the new test setup? Does anything need to happen here except rebasing, reviewing and merging?


Yes - the referenced issue was not actually broken, we just want to add more test coverage here and convert some old tests to the new test setup.

packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -156,79 +111,196 @@ testMatrix.setupTestSuite(({ provider }, _suiteMeta, clientMeta) => {
             email: 'user_2@website.com',
           },
         })
+
+        return prisma.user.findMany()
```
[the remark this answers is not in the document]
aqrln: Ah, I see, it looks like a merge/rebase gone wrong.

packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -338,434 +410,471 @@ testMatrix.setupTestSuite(({ provider }, _suiteMeta, clientMeta) => {
...
-    test('invalid value', async () => {
-      // @ts-test-if: provider === 'mongodb'
-      const result = prisma.$transaction(
-        async (tx) => {
-          await tx.user.create({ data: { email: 'user@example.com' } })
-        },
-        {
-          // @ts-test-if: provider !== 'mongodb'
-          isolationLevel: 'NotAValidLevel',
+          expect(await didLogCommit).toEqual(true)
```
aqrln, line 875: It would be nice to check what exactly is logged, is there a specific reason why this test is more minimal than the original one?

PR #16098, opened 2022-11-02
test(migrate): add test for baselining

Jolg42:
- Covered in docs: https://www.prisma.io/docs/guides/database/developing-with-prisma-migrate/add-prisma-migrate-to-a-project
- We have a dev and prod database
- And we want to baseline the production database with a baseline migration
- For that, we use `db pull`, `migrate dev`, `migrate resolve` and `migrate deploy` commands

[[Internal Slack](https://prisma-company.slack.com/archives/CGMCB82N8/p1667378484640799)]


packages/migrate/src/__tests__/Baseline.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      Applying migration \`20201231000000_\`
+
+      The following migration(s) have been applied:
+
+      migrations/
+        └─ 20201231000000_/
+          └─ migration.sql
+
+      Your database is now in sync with your schema.
+    `)
```
jkomyno, line 83: The migration name should be sanitized.

packages/migrate/src/__tests__/Baseline.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      Prisma schema loaded from prisma/schema.prisma
+      Datasource "my_db": SQLite database "dev.db" at "file:./dev.db"
+    `)
+    ctx.mocked['console.info'].mockReset()
+
+    // migrate dev --create-only
+    prompt.inject(['y'])
+    const migrateDevCreateOnly = MigrateDev.new().parse(['--create-only'])
+    await expect(migrateDevCreateOnly).resolves.toMatchInlineSnapshot(`
+      Prisma Migrate created the following migration without applying it 20201231000000_
```
jkomyno, line 43: The migration name should be sanitized.

packages/migrate/src/__tests__/Baseline.test.ts
```diff
@@ -0,0 +1,115 @@
...
+// And we want to baseline the production database with a baseline migration
+// For that, we use `db pull`, `migrate dev`, `migrate resolve` and `migrate deploy` commands
+describe('Baselining', () => {
+  it('SQLite: should succeed', async () => {
+    ctx.fixture('baseline-sqlite')
+    fs.remove('prisma/migrations')
+    fs.copy('prisma/dev.db', 'prisma/prod.db')
+
+    // Start with the dev database
+    process.env.DATABASE_URL = 'file:./dev.db'
```
jkomyno, line 28: We should always restore the `process.env` global object as we have found it. Maybe use something like

```typescript
  // backup env vars
  const OLD_ENV = { ...process.env }

  afterEach(() => {
    // reset env vars to backup state
    process.env = { ...OLD_ENV }
  })
```

as used elsewhere in the test codebase?

jkomyno (changes requested):
Tests are good, but some additional effort is required before merging imho, see the comments :)

packages/migrate/src/__tests__/Baseline.test.ts
```diff
@@ -0,0 +1,115 @@
...
+import { MigrateDeploy } from '../commands/MigrateDeploy'
+import { MigrateDev } from '../commands/MigrateDev'
+import { MigrateResolve } from '../commands/MigrateResolve'
+
+const ctx = jestContext.new().add(jestConsoleContext()).assemble()
+
+// Disable prompts
+process.env.GITHUB_ACTIONS = '1'
+// Disable generate
+process.env.PRISMA_MIGRATE_SKIP_GENERATE = '1'
```
tomhoule, line 15: Is this isolated to a single test? global variables are always scary

packages/migrate/src/__tests__/Baseline.test.ts
tomhoule: I'm wondering whether it wouldn't make even more sense to spawn a CLI as a separate process for each test — mocking and mutating the environment are both going to cause headaches, doubly so as soon as concurrency is involved.

tomhoule: But that's probably out of scope for this PR.

packages/migrate/src/__tests__/Baseline.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(`
+      Prisma schema loaded from prisma/schema.prisma
+      Datasource "my_db": SQLite database "dev.db" at "file:./dev.db"
+    `)
+    ctx.mocked['console.info'].mockReset()
+
+    // migrate dev --create-only
+    prompt.inject(['y'])
+    const migrateDevCreateOnly = MigrateDev.new().parse(['--create-only'])
+    await expect(migrateDevCreateOnly).resolves.toMatchInlineSnapshot(`
```
tomhoule, line 42: what isresolves` here?

packages/migrate/src/__tests__/Baseline.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      Datasource "my_db": SQLite database "prod.db" at "file:./prod.db"
+
+      1 migration found in prisma/migrations
+
+
+    `)
+
+    expect(ctx.mocked['console.log'].mock.calls).toEqual([])
+    expect(ctx.mocked['console.error'].mock.calls).toEqual([])
+  })
```
tomhoule, line 114: If we are only testing what the commands would print, I don't know if we print anything, but shouldn't there be a message about no migrations having been run?

packages/migrate/src/__tests__/Baseline.test.ts
tomhoule: I don't think this is in scope for this PR.

tomhoule: If we want to make a change there, it should go through an issue first. My bias would be to go in the opposite direction and take the user input literally, without tranforming the string in the CLI.

packages/migrate/src/__tests__/Baseline.test.ts
Jolg42: It is already

packages/migrate/src/__tests__/Baseline.test.ts
Jolg42: It's a bit useless because there is only one test though

packages/migrate/src/__tests__/Baseline.test.ts
Jolg42: Should be only for this test file yes

packages/migrate/src/__tests__/Baseline.test.ts
Jolg42: It means the `Promise` was resolved (success)

packages/migrate/src/__tests__/Baseline.test.ts
Jolg42: So the message you are looking for is line 102 `No pending migrations to apply.`
If we switched to mocking `stdout` we would get all the prints in one place.

tomhoule: :+1:

PR #16131, opened 2022-11-04
test(*): remove connection fallbacks as we centralize this on `.envrc`

Jolg42:
See https://github.com/prisma/prisma/blob/main/TESTING.md

Related https://www.notion.so/prismaio/Remove-connection-string-fallbacks-in-tests-ca7dc657a5cc41d7aeeae7f1cab9fa61


Jolg42:
We might have more to do, but I would say it can go in another PR, this was already too painful 😢

packages/migrate/src/__tests__/MigrateDiff.test.ts
```diff
@@ -550,11 +546,10 @@ describe('migrate diff', () => {
...
   describeIf(!process.env.TEST_SKIP_MSSQL)('sqlserver', () => {
-    const jdbcConnectionString = (
-      process.env.TEST_MSSQL_JDBC_URI_MIGRATE ||
-      'sqlserver://mssql:1433;database=tests-migrate;user=SA;password=Pr1sm4_Pr1sm4;trustServerCertificate=true;'
-    ).replace('tests-migrate', 'tests-migrate-diff')
-
+    if (!process.env.TEST_SKIP_MSSQL && !process.env.TEST_MSSQL_JDBC_URI_MIGRATE) {
+      throw new Error('You must set a value for process.env.TEST_MSSQL_JDBC_URI_MIGRATE. See TESTING.md')
+    }
+    const jdbcConnectionString = process.env.TEST_MSSQL_JDBC_URI_MIGRATE?.replace('tests-migrate', 'tests-migrate-diff')
```
jkomyno, line 552: Would be sufficient to get rid of the `force not-undefined` `!` symbols used later in this file?

```suggestion
    const jdbcConnectionString = process.env.TEST_MSSQL_JDBC_URI_MIGRATE.replace('tests-migrate', 'tests-migrate-diff')
```

Jolg42: This one is needed here because it would error with
```
'process.env.TEST_MSSQL_JDBC_URI_MIGRATE' is possibly 'undefined'.ts(18048)
```

Jolg42: But in reality it can't

jkomyno: What's the point of checking `process.env.TEST_SKIP_MSSQL` again? It's already included in the `describeIf` in the line above, line 548. If you remove `process.env.TEST_SKIP_MSSQL` from `if (!process.env.TEST_SKIP_MSSQL && !process.env.TEST_MSSQL_JDBC_URI_MIGRATE) {`, you can drop both the `?` and the `!` symbols in the rest of the scoped test.

Likewise for the CockroachDB test

Jolg42: Discovered during this PR that `describeIf()` only skips the test run but the code is still checked so it errors if it is removed 🙈

Issue #14933, opened 2022-08-22
findFirstOrThrow / findUniqueOrThrow errors are not reported via `error` event


Dam-Buty:
### Bug description

When `findFirstOrThrow` or `findUniqueOrThrow` can't find a model instance, an error is thrown. This error is not currently catchable as it doesn't trigger a Prisma `error` event, nor is it thrown by `next` inside of middlewares.

This is the issue which was described by @bbenezech in [this comment](https://github.com/prisma/prisma/issues/14209#issuecomment-1180183113) and confirmed to be a bug by @janpio .

For us the immediate effect is that the error is not logged by our custom logger, which results in 5 or 6 error lines in Datadog (triggering error alerting).

### How to reproduce

1. Setup a `prisma.$on("error", ...)` event hook
2. Setup a `prisma.$use(...)` middleware
3. Use the Prisma client to make a `findFirstOrThrow` query that is guaranteed to fail
4. The error is thrown and logged directly in console without going through the error event hook, or being thrown inside the middleware

### Expected behavior

I would expect one of those two behaviours (or both) :

* The `error` event hook is called
* The error is thrown by `next` in the middleware so i can catch it and treat it accordingly

### Prisma information

This happens for all models but should be reproducible with this minimal schema :

```prisma
datasource db {
  provider = "postgresql"
  url      = env("DB_URL")
}

generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["interactiveTransactions", "orderByNulls"]
}

model User {
  id                 Int                       @id @default(autoincrement())
  email              String                    @unique
}
```

Client queries : 

```typescript
  const prisma = new PrismaClient({
    datasources: {
      db: {
        url: config.database.url,
      },
    },
    log: [
      { level: "query", emit: "event" },
      { level: "info", emit: "event" },
      { level: "warn", emit: "event" },
      { level: "error", emit: "event" },
    ],
  });


  prisma.$on("info", (event) => {
    log.info(`[prisma] ${event.message}`, {
      target: event.target,
    });
  });

  prisma.$on("warn", (event) => {
    log.warn(`[prisma] ${event.message}`, {
      target: event.target,
    });
  });

  prisma.$on("error", (event) => {
    // This code does not appear to be called when throwing
    log.error(`[prisma] ${event.message}`, {
      target: event.target,
    });
  });

  prisma.$use(async (params, next) => {
    console.log(1, params);
[...]

PR #16162, opened 2022-11-07
feat: use engine implementation of findFirstOrThrow / findUniqueOrThrow

### LINKED ISSUE #10642: When rejectOnNotFound is used, chaining deeper into a related table still throws if it doesn't find anything
### Bug description

`rejectOnNotFound` should not cause chained queries to fail if they don't find an entity.
It is best understood by looking at the reproduction example:

### How to reproduce

Have a code like this:
```ts
import { PrismaClient, Prisma } from '@prisma/client'

const prismaClient = new PrismaClient()

; (async () => {
        const existingUser = await prismaClient.user.findUnique({
            where: {
                id: 1
            },
        })
        console.log('~ existingUser', existingUser)
        const user = prismaClient.user.findUnique({
            where: {
                id: 1
            },
            rejectOnNotFound: true
        })
        console.log( await user.profilePicture())
})()
```

Running this code using `ts-node pr.ts` prints out:
```
~ existingUser {
  id: 1,
  createdAt: 2021-12-10T11:24:30.096Z,
  email: 'john@example.com',
  name: null,
  role: 'USER',
  profilePictureId: null
}
[NotFoundError: No User found] { clientVersion: '3.6.0' }
```

### Expected behavior

Expected output for me is this:
```
~ existingUser {
  id: 1,
  createdAt: 2021-12-10T11:24:30.096Z,
  email: 'john@example.com',
  name: null,
  role: 'USER',
  profilePictureId: null
}
null
```
Without throwing any errors.

### Prisma information
```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
  url      = "file:./bugRepro.db"
}

model User {
  email     String   @unique
  name      String?

  role             String @default("USER")
  profilePictureId Int?
  profilePicture   Image? @relation(fields: [profilePictureId], references: [id])
}

model Image {
  updatedAt DateTime @updatedAt
  isHeader  Boolean  @default(false)
  url       String
  User      User[]
}

```
### Environment & setup

- OS: ubuntu
- Database: any
- Node.js version: 17

### Prisma Version

```

```

miguelff:
Fixes https://github.com/prisma/prisma/issues/10642
Closes https://github.com/prisma/prisma/pull/15887

Client changes to use [the engine implementation of `findFirstOrThrow` / `findUniqueOrThrow`](https://github.com/prisma/prisma-engines/pull/3356) rather than the emulated client actions added in https://github.com/prisma/prisma/pull/13920.

Once this is merged, I will tackle the following: 
- [x] https://github.com/prisma/prisma/issues/14933


packages/client/src/runtime/RequestHandler.ts
```diff
@@ -189,6 +188,11 @@ export class RequestHandler {
 
   handleRequestError({ error, clientMethod, callsite }: HandleErrorParams): never {
     debug(error)
+    // TODO: This is a workaround to keep backwards compatibility with clients
+    // consuming NotFoundError
+    if (error instanceof NotFoundError) {
```
millsp, line 193: Will this be made a PrismaClientKnownRequestError in the future?

millsp: Found the answer here https://github.com/prisma/prisma/pull/16162/files#diff-99659f0376b81639e19aee7c9cf5dee20ca6f2a4d68cdf549b2439940466d5f3R14

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -2695,6 +2717,63 @@ export namespace Prisma {
   }
       
 
+  /**
+   * Post findFirstOrThrow
+   */
+  export type PostFindFirstOrThrowArgs = {
```
millsp, line 2723: I think these args a really the same as `PostFindFirstArgs`, could we maybe not emit these or is it too difficult?

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -4033,18 +4179,6 @@ export namespace Prisma {
   }
 
 
-  /**
-   * User: findUniqueOrThrow
-   */
-  export type UserFindUniqueOrThrowArgs = UserFindUniqueArgsBase
```
millsp, line 4039: Question: Are `UserFindUniqueArgsBase` still being generated? If so could we remove them?

packages/client/src/runtime/core/model/applyOrThrowErrorAdapter.ts
```diff
@@ -0,0 +1,59 @@
...
+import { createErrorMessageWithContext } from '../../utils/createErrorMessageWithContext'
+import { NotFoundError } from '../../utils/rejectOnNotFound'
+
+type RequestCallback = (requestParams: InternalRequestParams) => Promise<unknown>
+
+/**
+ * `findUniqueOrThrow` and `findFirstOrThrow` are implemented in the engine, but before that,
+ * they used to be implemented on the client side.
+ *
+ * If the above mentioned methods returned no result, and error of type NotFoundError
```
millsp, line 15:
```suggestion
 * If the above mentioned methods returned no result, an error of type NotFoundError
```

packages/client/src/runtime/core/model/applyOrThrowErrorAdapter.ts
```diff
@@ -0,0 +1,59 @@
...
+ * was thrown. This error is client-side and in order to not break existing code relying
+ * on this error, we need to wrap the request callback in a function that catches the
+ * error thrown by the server and rethrow it as a client side error.
```
millsp, line 18:
```suggestion
 * error thrown by the Query Engine and rethrow it as a client side error.
```

millsp (commented):
I think the PR is in a good state, good work on this non-trivial implementation! :tada: I have left some comments/questions to see if we can improve things a bit. Will approve once resolved :)

packages/client/src/runtime/RequestHandler.ts
miguelff: Yes, that's the intention. For the moment the NotFoundError became deprecated, it inherited from `PrismaClientKnownRequestError` thus getting a code and the rest of its properties, and once we reach the next major version it should be replaced.

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
miguelff: Good call, I did the code changes in https://github.com/prisma/prisma/pull/16162/commits/1b806115f3619bbfc6cedf5e961afe4e704d17ce and pushed the updated snapshots in https://github.com/prisma/prisma/pull/16162/commits/b6decd849ccf0a02778283813ecf10968d661f65

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
miguelff: They are being generated and used for [extending the type to deprecate `rejectOnNotFound`](https://github.com/prisma/prisma/blob/b6decd849ccf0a02778283813ecf10968d661f65/packages/client/src/generation/TSClient/Args.ts#L111-L123)

I'm not completely sure how to monomorphise the base type and remove the child definition, as it's the only one. I'm going to dig a little bit and see what I find.

miguelff: Now tests are breaking because `ModelFind(First|Unique)Args` are duplicated. They  are generated once from the original type, and then a second time add the rejectOnNotFound option. 

We can do two things in here:
1. We could generate only one type, having the optional `rejectOnNotFound` property for both `Find(First|Unique)`  `Find(First|Unique)OrThrow` 
2. We could revert the changes and leave the aliased type

My understanding is that 1 will introduce an unexpected change. Before this, findUnique or findFirst, wouldn't allow `rejectOnNotFound` anymore.  It's only OrThrow variants the ones that do, by doing 1, we will be promoting to all the input types the rejectOnNotFound property. -- Is that maybe a bug? in any case it doesn't make sense to take back a backward compatibility that once was broken --

miguelff: Given the above, I think the best course of action as of now is to do 2. and revert the changes and leave the aliased type. And I did that in https://github.com/prisma/prisma/pull/16162/commits/3f9eedb0934050886d9403149a5c2cf7ebc11a04 and https://github.com/prisma/prisma/pull/16162/commits/c55b6e74f71f6293bfa77c78db021bc335fe2512

I will remove the revert and reverted commits before merge.

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -2191,6 +2191,22 @@ export namespace Prisma {
       args: SelectSubset<T, PostFindUniqueArgs>
     ): HasReject<GlobalRejectSettings, LocalRejectSettings, 'findUnique', 'Post'> extends True ? Prisma__PostClient<PostGetPayload<T>> : Prisma__PostClient<PostGetPayload<T> | null, null>
 
+    /**
+     * Find one Post that matches the filter or throw
```
SevInf, line 2195: Might be a good idea to update jsdoc to refer to the error code instead of deprecated class. Not critical though, at least not until class gets removed.

SevInf (commented):
One (optional) comment in addition to what Pierre said.
Great job! This certainly wasn't a trivial removal.

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
miguelff: 👍🏾 Addressed in [64b0f70](https://github.com/prisma/prisma/pull/16162/commits/64b0f7057b3c03929a5c13bc74588381ec8f9a7d)

SevInf (approved):

PR #16163, opened 2022-11-07
docs: update TESTING.md to clarify how to update snapshots

miguelff:
I faced a bit of friction in understanding how to update a snapshot in the scope of https://github.com/prisma/prisma/pull/16160. 

In particular, I found that:
1. `pnpm run test` from the root required `BUILDKITE_BRANCH` env var to be set
1.  When running ` pnpm run test getDmmf -- -u` to update getDmmf snapshots, tests ran against the new snapshot but the snapshot was not updated. 

This patch clarifies a bit the docs accordingly, although the first issue (required env var) is not addressed, as I think that would require a different code fix.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -63,6 +63,12 @@ To run tests requiring a database, start the test databases using Docker, see [D
 
 ## Jest tips
 
+1. Update your dependencies:
+
+   ```sh
+   pnpm i
+   ```
+
 1. We use the [Jest test framework](https://jestjs.io/). Its CLI is powerful and removes the need for npm scripts mostly. For most cases this is what you need to know:
 
    Note: the following command `pnpm run test` can be used inside the packages folders like `packages/client`. In the base folder you can only run `pnpm run test` without extra arguments.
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -63,6 +63,12 @@ To run tests requiring a database, start the test databases using Docker, see [D
 
 ## Jest tips
 
+1. Update your dependencies:
+
+   ```sh
+   pnpm i
+   ```
+
```
Jolg42, line 71: This doesn't belong under `Jest tips` I think

Maybe we move it to the top under a new subtitle like `Install dependencies`?

TESTING.md
```diff
@@ -71,10 +77,10 @@ To run tests requiring a database, start the test databases using Docker, see [D
    pnpm run test <fileNamePattern> -t <testNamePattern>
    ```
 
-   and to update snapshots use the -u option like this (the `--` are required, anything after the dashes will be passed to Jest):
+   and to update snapshots use the -u uption
```
Jolg42, line 80:
```suggestion
   and to update snapshots use the -u option
```

TESTING.md
miguelff: I think someone looking to run unit tests will arrive at this section for specific help. It's kind of a precondition to update your dependencies to have your tests under a stable fixture. If this was in a different place, the reader will have to know where and have read the whole documentation. Having this step in here allows for the section to be self-sufficient.

TESTING.md
miguelff: Good catch! thank you.

TESTING.md
miguelff: Having said that, feel free to go ahead and change it to a better place if you have a stronger opinion. I defer it to you.

TESTING.md
```diff
@@ -63,6 +63,12 @@ To run tests requiring a database, start the test databases using Docker, see [D
 
 ## Jest tips
 
+1. Update your dependencies:
```
Jolg42, line 66:
```suggestion
1. Install/update the dependencies, run the following command in the root directory:

```

Jolg42 (approved):

Issue #13076, opened 2022-04-29
Introspection from different datasource protocol leads to misleading error message


jkomyno:
Here is how to reproduce:

- Initialize a new schema with e.g. the `mongodb` provider: `npx prisma init --datasource-provider mongodb`
- Ensure the schema is valid: `npx prisma validate`
- Run introspection with a URL having a different datasource protocol: `npx prisma db pull --url postgres://prisma:prisma@localhost:5432/tests`

You should see an error similar to the following:

```text
Prisma schema loaded from prisma/schema.prisma
Environment variables loaded from .env
Datasource "db"

✖ Introspecting

Error: P1012 Introspection failed as your current Prisma schema file is invalid

Please fix your current schema manually, use prisma validate to confirm it is valid and then run this command again.
Or run this command with the --force flag to ignore your current schema and overwrite it. All local modifications will be lost.
```

However, the schema automatically generated in `prisma/schema.prisma` is perfectly valid, as checked earlier with `npx prisma validate`.

## Proposed Solution

Prisma should check whether the protocol of the `-url` argument of the `dp pull` command is compatible with the one written in the datasource block of the schema, if any. If this doesn't hold, show an error message among the lines of: "Error: Introspection failed as the connection string given by `--url` is not compatible with the datasource stored in the schema file".

jkomyno:
We’re closing this issue to help focus on features and improvements that align with our focus areas and those of our community. 

This issue was originally created for internal tracking purposes, but if you feel this is important to you, feel free to reopen the issue and get like-minded users to upvote as we will prioritize fixing issues based on community interest.

For more insight into our priorities and how we’re thinking about the future of Prisma, please see our blog post: https://www.prisma.io/blog/prisma-orm-manifesto.

Thank you for your understanding and support!

PR #16200, opened 2022-11-09
feat: improve `prisma db pull` error message

### LINKED ISSUE #16176: (re)-Introspection errors have a misleading error message
## Reproduction

- Given the following `docker-compose.yml` definition,

```yaml
version: '3.7'

services:
  postgres:
    image: postgres:15
    restart: always
    command: postgres -c 'max_connections=1000'
    environment:
      POSTGRES_PASSWORD: "prisma"
      POSTGRES_HOST_AUTH_METHOD: "md5"
      POSTGRES_INITDB_ARGS: "--auth-host=md5"
    ports:
      - "5432:5432"
```

run a `postgres` instance via

```sh
docker-compose up postgres
```

- Create the following `schema.prisma`:

```prisma
generator client {
  previewFeatures = ["referentialIntegrity"]
}

datasource db {
  provider     = "postgres"
  url          = env("DATABASE_URL")
}

model A {
  id Int @id
  bs B[]
}

model B {
  id  Int @id
  aId Int
  a   A   @relation(fields: [aId], references: [id], onUpdate: Cascade, onDelete: NoAction)
}
```

- Set up the `DATABASE_URL` environment variable via

```sh
export TEST_POSTGRES_URI="postgresql://postgres:prisma@localhost:5432/tests"
```

- Push it to the database via `pnpx prisma@4.4.0 db push`, which uses the latest Prisma version that still supported an emulated `NoAction` referential action on `postgres`

- Re-introspect the database via `pnpx prisma@4.5.0 db pull`. This results in the following error, without any further detail:

> Error: P1012 Introspection failed as your current Prisma schema file is invalid
>
> Please fix your current schema manually, use prisma validate to confirm it is valid and then run this command again.
Or run this command with the --force flag to ignore your current schema and overwrite it. All local modifications will be lost.

## Solutions

We should tell the user to first run `prisma validate` to understand why the schema is not valid, then fix the current schema depending on what `prisma validate` suggests, and run the same command again to confirm the validity of the schema. (We might even suggest to use `prisma format` to both validate and format the Prisma schema)

The new error message could be something like:

> Error: P1012 Introspection failed as your current Prisma schema file is invalid
>
> Please run `prisma validate` to understand why the Prisma schema is not valid, modify the schema as suggested, and run `prisma validate` again to confirm the validity of the schema; you may then run this command again.
Alternatively, you can run this command with the `--force` flag to ignore your current schema and overwrite it. All local modifications will be lost.

jkomyno:
Closes https://github.com/prisma/prisma/issues/16176 with the new message:

> Please fix your current schema manually (using either prisma validate or the Prisma VSCode extension to understand what's broken and confirm you fixed it), and then run this command again.

Quick notes before the review:
- updating the `prisma db pull` snapshot tests resulted in `DbPull.test.ts` having an unexpected space diff. Still, this shouldn't result in any visible difference for the users.
- recall that this error also appears when it shouldn't (https://github.com/prisma/prisma/issues/13076), I've added a few TODOs for clarification


packages/migrate/src/commands/DbPull.ts
```diff
@@ -264,11 +270,13 @@ Then you can run ${chalk.green(getCommandWithExecutor('prisma db pull'))} again.
 
         // TODO: this error is misleading, as it gets thrown even when the schema is valid but the protocol of the given
         // '--url' argument is different than the one written in the schema.prisma file.
+        // We should throw another error earlier in case the URL protocol is not compatible with the schema provider.
```
jkomyno, line 273: I had spotted this error handling bug over 6 months ago. I have left a couple of other TODOs in the file, as we should be able to throw an error before this code block in case the protocol of the `--url` input is not compatible with the `provider` attribute in the `prisma.schema` file.

packages/migrate/src/commands/DbPull.ts
```diff
@@ -163,41 +163,47 @@ Set composite types introspection depth to 2 levels
      *
      * If neither these variables were set, we'd have already thrown a `NoSchemaFoundError`.
      */
-    const schema = await match({ url, schemaPath })
+    const { config, schema } = await match({ url, schemaPath })
```
jkomyno, line 166: Centralized `getConfig` usage here, while preserving the pre-existing behavior. This came handy while inspecting `input.url` protocol validity (see TODOs below), and reduces redundant calls on the same arguments.

packages/migrate/src/commands/DbPull.ts
```diff
@@ -163,41 +163,47 @@ Set composite types introspection depth to 2 levels
...
+            ignoreEnvVarErrors: true,
+          })
 
           if (input.url) {
-            const config = await getConfig({
-              datamodel: rawSchema,
-              ignoreEnvVarErrors: true,
-            })
+            // TODO: ensure that the `input.url` protocol is compatible with the schema provider
+            // to avoid ambiguities in case of schema parsing errors (error code P1012)
```
jkomyno, line 178: This would require https://github.com/prisma/prisma/issues/16201 to be solved before proceeding

packages/generator-helper/src/types.ts
```diff
@@ -57,9 +57,16 @@ export type ConnectorType =
...
   | 'cockroachdb'
 
+  // TODO: this should be removed in favor of `'sqlserver'`, as per `getConfig({ ... }).datasources[0]?.provider` from a schema with `provider = "sqlserver"`
+  // 'jdbc:sqlserver' has been removed in https://github.com/prisma/prisma-engines/pull/2830
+  | 'jdbc:sqlserver'
+
+// TODO: this is also a valid provider as returned by `getConfig({ ... }).datasources[0]?.provider` from a schema with `provider = "postgres"`,
+// but adding it would currently break the TypeScript compilation.
+// | 'postgres'
+
```
jkomyno, line 69: This type definition could be simplified when https://github.com/prisma/prisma/issues/16201 is solved.

Jolg42:
Closed in favor of https://github.com/prisma/prisma/pull/16956

PR #16240, opened 2022-11-11
fix(client): Correctly report error location in batch transaction

### LINKED ISSUE #14373: The error is incorrectly pointed out when using the $transaction method
### Bug description

When multiple update queries are set to the $transaction method and an error occurs in one of them, it is always indicated that the first query in the array of arguments has failed.

### How to reproduce

https://github.com/aya-eiya/prisma-tx-error

schema.prisma

```schema.prisma
datasource db {
  provider = "sqlite"
  url      = "file:./dev.db"
}

generator client {
  provider = "prisma-client-js"
}

model User {
  email     String   @unique
  memo      String?
}

```

main.js

```javascript
const { PrismaClient } = require('@prisma/client')

async function main() {
  await prisma.user.deleteMany({})
  await prisma.$transaction([
    prisma.user.updateMany({
      where: {
        email: {
          contains: '@corp.com',
        },
      },
      data: {
        memo: 'corp',
      },
    }),
    prisma.user.update({
      where: {
        id: 1,
      },
      data: {
        memo: 'id is 1',
      },
    }),
  ])
}

main()
```

This script will get

```
./src/main.js:7:17

   4 const prisma = new PrismaClient()
   5 await prisma.user.deleteMany({})
   6 await prisma.$transaction([
→  7   prisma.user.updateMany(
  An operation failed because it depends on one or more records that were required but not found. Record to update not found.
    at RequestHandler.handleRequestError (/Users/aya-eiya/Work/prisma/node_modules/@prisma/client/runtime/index.js:49670:13)
    at RequestHandler.request (/Users/aya-eiya/Work/prisma/node_modules/@prisma/client/runtime/index.js:49652:12)
    at async PrismaClient._request (/Users/aya-eiya/Work/prisma/node_modules/@prisma/client/runtime/index.js:50572:18)
    at async main (/Users/aya-eiya/Work/prisma/src/main.js:6:3)
(node:45257) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:45257) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

> `An operation failed because it depends on one or more records that were required but not found. Record to update not found.
`

is never happen on `prisma.user.updateMany`. the actual error line is `prisma.user.update`.

this makes investigating difficult and causes confusion. so I think this must be a bug of $trancation.

### Expected behavior

It must point out an actual error line.

### Environment & setup

node: v14.17.0
OS: macOS 12.4 (21F79)
Kernel: Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:29 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T8101

### Prisma Version

```
"@prisma/client": "^4.0.0",
"prisma": "^4.0.0"
```

### LINKED ISSUE #15433: wrong error message/header on multiple table upsert using transaction
### Bug description

Hi Prisma team,
So, I have 3 tables that need to upsert at once using [transaction](https://www.prisma.io/docs/concepts/components/prisma-client/transactions#sequential-prisma-client-operations), and I need to log the exception when something wrong on execution.

I tried to make the execution failed by putting some invalid data of first table. and it worked well, I got the error message & the error detail correctly, something like `Error: Invalid `prisma.table_1.upsert()` invocation: --detail error message that related to table 1--`.

on second attempt, I put invalid data on the second table, and it's not worked as expected, the error message/header still referring to first table while the error detail relates to second table.  `Error: Invalid `prisma.table_1.upsert()` invocation: --detail error message that related to table 2--`.

it also happen when something is wrong on the 3rd table, the error header still refering to 1st table.

### Expected behavior

expected behavior: error message/header should refer to the table where the error comes.

`Error: Invalid `prisma.table_1.upsert()` invocation: --detail error message that related to table 1--`.
`Error: Invalid `prisma.table_2.upsert()` invocation: --detail error message that related to table 2--`.
`Error: Invalid `prisma.table_3.upsert()` invocation: --detail error message that related to table 3--`.

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
```

### Environment & setup

- OS: windows, ubuntu
- Database: SQL Server
- Node.js version: v16.13.2

### Prisma Version

```
"prisma": "^4.0.0"
```

SevInf:
Engine PR: https://github.com/prisma/prisma-engines/pull/3384

Uses newly added `batch_request_idx` property of an errors to identify
and correctly report error location within a batch.

Fix #15433
Fix #14373


packages/client/package.json
```diff
@@ -149,7 +149,7 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "4.7.0-8.379f9d33a885f66fbe10a30a7b6c14bc56f3de3b"
+    "@prisma/engines-version": "4.7.0-9.feat-error-req-idx-26e00ab332793084ea0280df387b126fb5e17eba"
```
SevInf, line 152: TODO: revert before merging

aqrln (approved):

packages/client/tests/functional/issues/14373-batch-tx-error/tests.ts
```diff
@@ -0,0 +1,24 @@
...
+        where: {
+          email: 'notthere@example.com',
+        },
+        data: {
+          memo: 'id is 1',
+        },
+      }),
+    ])
+
+    await expect(result).rejects.toThrowError('Invalid `prisma.user.update()` invocation')
```
millsp, line 22: Question: Does the error point at the line number of the error?

packages/client/src/runtime/utils/waitForBatch.ts
```diff
@@ -0,0 +1,36 @@
+import { hasBatchIndex } from '@prisma/engine-core'
+
+/**
+ * Waits for result of batch $transaction and picks the best possible error to report if any
+ * of the request fails. Best error is determined as follows:
+ *
+ * - if engine have reported and index of failed request in the batch, the best error is the one
+ * who's index matches that
+ * - otherwise, first error is used (like Promise.all)
```
millsp, line 9: Do you know if it would be possible to widen this feature to `Promise.all` (if not already)? If I understood correctly, it uses the same batching API, so do you think we could use the DataLoader/Engine code to implement it for both?

SevInf:
@millsp @aqrln I'd ask for another re-review here. I had to rewrite `waitForBatch` function to be closer to how `Promise.all` behaves and reject early on all non-batch errors. This is necessary, because if one of the batch requests fails early, with, for example, validation error, batch will not be executed at all and remaining promises won't settle. `Promise.all` rejected batch promise correctly in this case, older `waitForBatch`, based on `Promise.allSetlled` did not. Now it fixed, with unfortunate downside of a more complicated code.

packages/client/tests/functional/issues/14373-batch-tx-error/tests.ts
SevInf: Yes. Did not include full snapshot here because it differes for data proxy for some reason, but you can check interactive-transactions snapshots here - they now also point to correct query.

packages/client/tests/functional/issues/14373-batch-tx-error/_matrix.ts
```diff
@@ -0,0 +1,24 @@
...
+      provider: 'mongodb',
+    },
+    {
+      provider: 'cockroachdb',
+    },
+    {
+      provider: 'sqlserver',
+    },
+  ],
+])
```
danstarns, line 24:
```suggestion
import { defineMatrix } from '../../_utils/defineMatrix'
import { allProviders } from '../../_utils/providers'

export default defineMatrix(() => [allProviders])
```

SevInf: Nice, somehow missed that this is a thing now!

danstarns (approved):

aqrln (approved):
Amazing work!

millsp (approved):

PR #16276, opened 2022-11-15
test(client): Test directory names

danstarns:
The issues dir contained a lot of tests where the directory was just a number, this was cumbersome to debug because it was difficult to know the context of what failed. This PR adds a somewhat more helpful name to the end of the directory name and a link to the missing issue.


aqrln (approved):

Jolg42:
Could be that special characters like `()` or `$` are the problem here

PR #16287, opened 2022-11-15
fix(qe): fix error logging

### LINKED ISSUE #16226: Bug(qe): errors don't get logged
Supersedes https://github.com/prisma/prisma/issues/14933

### Background

Prisma can be configured to [log events to stdout](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging#log-to-stdout) or [emit logging events](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging#event-based-logging). Those events can be categorized into [four levels ](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#log-levels) which do not have a hierarchical relationship with each other. (A logging level does not include events from any other level)

One of these levels is `error`, but the [reference](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#log-levels) doesn't specify what is eligible to be logged using this level. 

Given that the reference has a [section on errors](https://www.prisma.io/docs/reference/api-reference/error-reference) users might have the reasonable expectation that, when any of those happen, an event is [logged to the console](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging#log-to-stdout) or [emitted](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging#event-based-logging) for subscription depending on the configuration. 

This is manifested in past issues like https://github.com/prisma/prisma/issues/14933, but given the reproduction scenarios below, the problem is not restricted to `findFirstOrThrow` and `findUniqueOrThrow` `NotFound` errors, but **to all the errors happening in Prisma**

### Reproduction

I created two reproduction scenarios, using the [libray engine](https://www.prisma.io/docs/reference/api-reference/environment-variables-reference#prisma_query_engine_binary) and the [binary engine](https://www.prisma.io/docs/reference/api-reference/environment-variables-reference#prisma_query_engine_binary) (As they exercise two different code paths for dispatching requests coming from the prisma client)

Both resulted on the error not being logged.

<details>
<summary>Premise: the engine was tricked to through a panic when dispatching a query</summary>

```diff
diff --git a/query-engine/core/src/interpreter/query_interpreters/read.rs b/query-engine/core/src/interpreter/query_interpreters/read.rs
index 6916f78fcc..878f876c5e 100644
--- a/query-engine/core/src/interpreter/query_interpreters/read.rs
+++ b/query-engine/core/src/interpreter/query_interpreters/read.rs
@@ -16,6 +16,10 @@ pub fn execute<'conn>(
     trace_id: Option<String>,
 ) -> BoxFuture<'conn, InterpretationResult<QueryResult>> {
     let fut = async move {
+        if 1 < 2 {
+            panic!("Damn that's a panic");
+        }
+
         match query {
             ReadQuery::RecordQuery(q) => read_one(tx, q, trace_id).await,
             ReadQuery::ManyRecordsQuery(q) => read_many(tx, q, trace_id).await,

```
</details>
<details>
<summary>Scenario 1 - Binary engine</summary>

`index.ts`
```typescript
import { PrismaClient, Prisma } from '.prisma/client'
​
async function main() {
    log: [
      {
        emit: 'event',
        level: 'error',
      },    
    ],
    __internal: {
      engine: {
        endpoint: "http://127.0.0.1:57581",
      },
    },
  } as any);
​
  (prisma as PrismaClient<Prisma.PrismaClientOptions, 'error'>).$on("error", (e) => {
    console.log("WORKS WELL")
    process.exit(1)
  })
​
  const email = `user.${Date.now()}@prisma.io`
​
  const res = await prisma.user.findFirst({
    where: {
      email,
    }
  })
​
  console.log(res)
}
​
void main().catch((e) => {
  console.log("NOOOO!")
  console.log(e)
  process.exit(1)
})
```

`prisma.schema`
```schema
generator client {
  engineType      = "binary"
  previewFeatures = []
}
datasource db {
  provider = "postgresql"
  url      = "postgresql://postgres:prisma@localhost:5432"
}
model User {
  id    String @id @default(cuid())
  email String @unique
  posts Post[]
}
model Post {
  id       String @id @default(cuid())
  title    String
  author   User   @relation(fields: [authorId], references: [id])
  authorId String
}
```
Result:
Screenshot 2022-11-10 at 16 19 09
</details>

<details>
<summary>Scenario 2 - Library engine</summary>

`index.ts`
```
import { PrismaClient, Prisma } from '.prisma/client'

async function main() {
    log: [
      {
        emit: 'event',
        level: 'error',
      },
    ],
  });

  prisma.$on("error", (e) => {
    console.log("WORKS WELL")
    process.exit(1)
  })

  const email = `user.${Date.now()}@prisma.io`

  const res = await prisma.user.findFirst({
    where: {
      email,
    }
  })

  console.log(res)
}

void main().catch((e) => {
  console.log("NOOOO!")
  console.log(e)
  process.exit(1)
})
```

`prisma.schema`
```schema
generator client {
  previewFeatures = []
}
datasource db {
  provider = "postgresql"
  url      = "postgresql://postgres:prisma@localhost:5432"
}
model User {
  id    String @id @default(cuid())
  email String @unique
  posts Post[]
}
model Post {
  id       String @id @default(cuid())
  title    String
  author   User   @relation(fields: [authorId], references: [id])
  authorId String
}
```

Result
Screenshot 2022-11-10 at 15 41 33
</details>

### Expected behavior

The specifics have to be determined. There's no previous record in the codebase to determine the correct behavior. And the [reference](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#log-levels)  of what constitutes an error is vague enough as to require a specification of how this should behave.

My proposal is to log with the error level what is said to constitute an error in the [error reference](https://www.prisma.io/docs/reference/api-reference/error-reference) thus following the principle of least astonishment. But I'd like to count with @janpio's take on how this should behave before jumping into implementation.

### Definition of done

- [x] Expected behavior clarified
- [x] Implementation and tests ensuring expected behavior
- [ ] Docs on the [log levels](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#log-levels) clarify the expected behavior

miguelff:
Fixes #16226 

This patch ensures that errors resulting in a failed query generate an event if the error logger is enabled.  

This aligns with the principle of least astonishment: The user expects errors to be logged, and due to [the lack of a formal definition of what constitutes an error](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#log-levels), the most sensible approach in my opinion (backed by bug reports like https://github.com/prisma/prisma/issues/14933) is to log any condition leading to an error in the processing of the query, which is, in turn, any of [the four well-known classes of errors](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#log-levels).

This is done now [in the RequestHandler](https://github.com/prisma/prisma/pull/16287/files#diff-41021953e5dbe1050efb578ed85e14befe861bce256f4c312fac4bef801e3731R188-R203) thus capturing any possible error condition happening during query processing. Before, every different engine was doing this differently, and errors were not properly captured.

A smoke test suite proves that `NotFoundError` [errors in `findXOrThrow` are now properly emitted](https://github.com/prisma/prisma/issues/14933). I consider this enough, given that the code that handles the error is the same for every exception thrown during query execution.


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -339,12 +340,17 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
     _dataProxy: boolean
     _extensions: Extension[]
+    _logEmitter: EventEmitter
 
     constructor(optionsArg?: PrismaClientOptions) {
       if (optionsArg) {
         validatePrismaClientOptions(optionsArg, config.datasourceNames)
       }
 
+      this._logEmitter = new EventEmitter()
```
miguelff, line 350: Different engines have their own instance of the EventEmitter, that for some reason is not a singleton. I thought it made more sense to inject the dependency, as the event pub/sub should be the same across the client codebase.

Then, wherever else it is used, it's injected.

packages/engine-core/src/common/errors/PrismaClientRustError.ts
```diff
@@ -1,28 +1,27 @@
 import type { PrismaClientRustErrorArgs } from './types/PrismaClientRustErrorArgs'
-import { getBacktraceFromLog, getBacktraceFromRustError } from './utils/log'
```
miguelff, line 2: Changes in this file, need to be reviewed together with changes in [log.ts](https://github.com/prisma/prisma/pull/16287/files#diff-7cdf39e07157969396116c74b51ad1ab01705f3d2691f9e1a6aa60ba1007846e) and [some changes in BinaryEngine.ts] which were aimed at simplifying error processing logic during engine initialization, which remains functionally untouched, but a bit easier to understand. 

To refactor that part of BinaryEngine, I only made the `PrismaClientRustError` treat an error uniformly from an ErrorLog trace than from an actual Panic, and for that, I hid specific methods that dealt with `RustLog` and `RustError` and make equivalent methods that dealt with the union type exported.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -342,6 +343,10 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
         validatePrismaClientOptions(optionsArg, config.datasourceNames)
       }
 
+      const logEmitter = new EventEmitter().on('error', () => {
+        // to prevent unhandled error events
```
danstarns, line 347: Should we at least log or DEBUG these?

packages/client/tests/functional/query-error-logging/_matrix.ts
```diff
@@ -0,0 +1,24 @@
...
+      provider: 'mongodb',
+    },
+    {
+      provider: 'cockroachdb',
+    },
+    {
+      provider: 'sqlserver',
+    },
+  ],
+])
```
danstarns, line 24:
```suggestion
import { defineMatrix } from '../_utils/defineMatrix'
import { allProviders } from '../_utils/providers'

export default defineMatrix(() => [allProviders])
```

miguelff:
Taking a look at the windows failures. Please don't let this hold you from a review.

packages/client/src/runtime/getPrismaClient.ts
miguelff: Hey @danstarns I was about to a debug trace in here like suggested, but then ended up writing a clarification comment and not doing it. Let me see if it makes sense.

From https://github.com/prisma/prisma/pull/16287/commits/a13c40cf1d4a44c2bc11b9b5039cd4a678b9b432

```typescript
      const logEmitter = new EventEmitter().on('error', (e) => {
        // this is a no-op to prevent unhandled error events
        //
        // If the user enabled error logging this would never be executed. If the user did not
        // enabled error logging, this would be executed, and a trace for the error would be logged
        // in debug mode, which is like going in the opposite direction than what the user wanted by
        // not enabling error logging in the first place.
      })
```

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -342,6 +343,15 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
         validatePrismaClientOptions(optionsArg, config.datasourceNames)
       }
 
+      const logEmitter = new EventEmitter().on('error', (e) => {
```
millsp, line 346: In `eventemitter3`

> We do not throw an error when you emit an error event and nobody is listening.

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -236,36 +234,40 @@ You may have to run ${chalk.greenBright('prisma generate')} for your changes to
...
+      this.handlePanic()
     }
   }
 
+  private emitError(err: any) {
+    const message = err.message ?? err
+    this.logEmitter.emit('error', {
+      message: message,
+      timestamp: new Date(),
+      target: 'binary engine lifecycle management',
```
millsp, line 267: Is this intentional?

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -403,7 +397,9 @@ export class DataProxyEngine extends Engine {
           }
         }
 
-        this.logEmitter.emit('warn', { message: 'This request can be retried' })
+        this.logEmitter.emit('warn', {
```
millsp, line 400: I don't think sentence was documented but let's check if that was the case

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -236,36 +234,40 @@ You may have to run ${chalk.greenBright('prisma generate')} for your changes to
...
+      clientVersion: this.clientVersion!,
+      error: err,
+    })
+    this.emitError(wrappedError)
+    if (wrappedError.isPanic()) {
+      this.handlePanic()
     }
   }
 
+  private emitError(err: any) {
```
millsp, line 262: (Warning: Potentially stupid question)
Why are we emitting errors here still vs always emitting in `RequestHandler`, is it for a specific set of errors that only happen in BinaryEngine? https://github.com/prisma/prisma/pull/16287/files#r1029921786

packages/client/src/runtime/getPrismaClient.ts
miguelff: Agreed that's one of the reasons why we should use eventemitter3 altogether and stop depending on Node's implementation.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
miguelff: Good call: [It was not documented](https://www.google.com/search?q=%22can+be+retried%22+site%3Aprisma.io&ei=3tx9Y8T4DrWnkdUPyq2A8AM&ved=0ahUKEwjE37ib9MP7AhW1U6QEHcoWAD4Q4dUDCA8&uact=5&oq=%22can+be+retried%22+site%3Aprisma.io&gs_lcp=Cgxnd3Mtd2l6LXNlcnAQA0oECEEYAUoECEYYAFCvCFj0OWCXPGgCcAB4AoAB2AGIAY0MkgEGMTIuMy4xmAEAoAEBwAEB&sclient=gws-wiz-serp)

packages/engine-core/src/binary/BinaryEngine.ts
miguelff: I have simplified the binary engine error workflow extensively by relying on the type system, in particular removing the union type `RustError`, consolidating the states: `lastPanic`, `lastError`, `lastErrorlog` into `lastError: PrismaClientRustError` which allowed me to remove a lot of conditionals: and finally getting rid of the error emission, in `setError` (now it just sets the error, handling a possible panic), and making it explicit in the places where the state of the server is managed. 

https://github.com/prisma/prisma/pull/16287/commits/00a497ed5ad82deebbcdcb09dd633bfb0c00d7ce

I will still need to do some manual testing on different error conditions. I would love to be able to automate things like I with https://github.com/Shopify/toxiproxy to introduce latency or network errors as part of the test suite, but I think that goes way beyond the scope of that issue.

miguelff:
Did the following:
- Tampered the binary engine:

```diff
diff --git a/query-engine/request-handlers/src/graphql/handler.rs b/query-engine/request-handlers/src/graphql/handler.rs
index 39cb2e08e1..15690c5b47 100644
--- a/query-engine/request-handlers/src/graphql/handler.rs
+++ b/query-engine/request-handlers/src/graphql/handler.rs
@@ -26,7 +26,6 @@ impl<'a> GraphQlHandler<'a> {
 
     pub async fn handle(&self, body: GraphQlBody, tx_id: Option<TxId>, trace_id: Option<String>) -> PrismaResponse {
         tracing::debug!("Incoming GraphQL query: {:?}", &body);
-
         match body.into_doc() {
             Ok(QueryDocument::Single(query)) => self.handle_single(query, tx_id, trace_id).await,
             Ok(QueryDocument::Multi(batch)) => match batch.compact(self.query_schema) {
@@ -44,7 +43,7 @@ impl<'a> GraphQlHandler<'a> {
 
     async fn handle_single(&self, query: Operation, tx_id: Option<TxId>, trace_id: Option<String>) -> PrismaResponse {
         use user_facing_errors::Error;
-
+        dbg!("Running single {}", query.clone());
         let gql_response = match AssertUnwindSafe(self.handle_graphql(query, tx_id, trace_id))
             .catch_unwind()
             .await
@@ -201,6 +200,9 @@ impl<'a> GraphQlHandler<'a> {
         tx_id: Option<TxId>,
         trace_id: Option<String>,
     ) -> query_core::Result<ResponseData> {
+        if 2 > 1 {
+            panic!("Forced")
+        }
         self.executor
             .execute(tx_id, query_doc, self.query_schema.clone(), trace_id)
             .await

```

- Ran Prisma against the binary engine. The connection is reset before the actual panic is processed asynchronously. In this case, the error is correctly received by the client

- Ran Prisma against the library engine. The behavior is correct.

I think this is ready to be merged.

packages/engine-core/src/binary/BinaryEngine.ts
miguelff: I made some changes based on https://github.com/prisma/prisma/pull/16287#discussion_r1029925661. This is done differently now.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -382,10 +380,6 @@ export class DataProxyEngine extends Engine {
       try {
         return await args.callback({ logHttpCall })
       } catch (e) {
-        this.logEmitter.emit('error', {
```
aqrln, line 385: I think we would lose the logging for non-retryable errors which are not request errors if we remove it (e.g. Data Proxy HTTP 400 errors), wouldn't we? Or is that not a problem since they'll be thrown anyway?

aqrln (commented):
One comment/question, but other than that, looks very good to me! Thank you for the fix and for the nice refactoring!

packages/engine-core/src/data-proxy/DataProxyEngine.ts
miguelff: They will be thrown anyway, yes. Shouldn't be a problem.

aqrln (approved):

PR #16302, opened 2022-11-16
revert: "test(client): Test directory names"

danstarns:
Reverts prisma/prisma#16276


Jolg42 (approved):
Reverted because "no-docker" client test on Windows are failing since, see https://github.com/prisma/prisma/pull/16276#issuecomment-1317125040

Thanks @danstarns !

PR #16343, opened 2022-11-18
feat(client): result extensions runtime

SevInf:
Core of implemenation consists of 3 functions:

- `getAllComputedFields` is responsible for resolving all depenencies
  and naming conflicts between computed fields and producing a list of
  computed that could potentially be applied to specific model.
- `applyResultExtensions` function applies computed field to single part
  of the query result. Does not traverse relationship.
- `visitQueryResult` traverses the response and calls provided
  callback for each part of the response, corresponding to particular
  model and allows to modify it.

By combining all 3 of them, we are able to resolve and apply all
computed fields to any query response recursively, following all the
relationships.

- [ ] Evaluate performance overhead of computed fields


packages/client/src/runtime/RequestHandler.ts
```diff
@@ -263,6 +277,26 @@ export class RequestHandler {
     return unpack({ document, data, path: getPath })
   }
 
+  applyResultExtensions({ result, modelName, args, extensions }: ApplyExtensionsParams) {
```
millsp, line 280: Do you think it makes sense to keep these together in `core/extensions`? I'm planing to park `applyQueryExtensions` there

millsp (approved):

PR #16408, opened 2022-11-23
feat(client): client extensions query

millsp:


packages/client/src/runtime/core/extensions/$extends.ts
```diff
@@ -11,7 +11,7 @@ export type RequiredArgs = ResultArgs & ModelArgs & ClientArgs & QueryOptions
 
 type ResultArgs = {
   result: {
-    [modelName: string]: ResultModelArgs
+    [ModelName in string]: ResultModelArgs
```
SevInf, line 14: What's the difference?

SevInf (approved):

packages/client/tests/functional/extensions/enabled/query.ts
```diff
@@ -0,0 +1,832 @@
...
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+let prisma: PrismaClient<{ log: [{ emit: 'event'; level: 'query' }] }>
+declare const newPrismaClient: NewPrismaClient<typeof PrismaClient>
+
+const randomId1 = randomBytes(12).toString('hex')
+const randomId2 = randomBytes(12).toString('hex')
+const randomId3 = randomBytes(12).toString('hex')
+
+jest.retryTimes(3)
```
SevInf, line 18: I see below you are using `waitFor`. Do we still need `retryTimes` in that case?

PR #16424, opened 2022-11-23
fix(client): Fixed warnAboutRejectOnNotFound message

maxsynnott:
The following example:

```ts
const prisma = new PrismaClient({ rejectOnNotFound: true })
await prisma.user.create({ data: { email: `user@example.com` } })
```

currently produces the following warning:

```prisma:warn `rejectOnNotFound` option is deprecated and will be removed in Prisma 5. Please use `prisma.comment.undefined` method instead```

As you can see there is a bug in the second sentence.
When the action is something other than `findUnique` or `findFirst`, there is no `replacementAction` which results in instructing the user to use `undefined` method instead.
This fix will only log the second sentence if there is a replacement action therefore avoiding this buggy message.

The same example above will now produce the following warning:

```prisma:warn `rejectOnNotFound` option is deprecated and will be removed in Prisma 5```


Jolg42:
Hi @maxsynnott 

Thanks for your first PR here! We're quite busy with tomorrow's release, I pinged the Client Team to review this. Looks good to me 💚

maxsynnott:
> Hi @maxsynnott
> 
> Thanks for your first PR here! We're quite busy with tomorrow's release, I pinged the Client Team to review this. Looks good to me 💚

Sounds good 👌

maxsynnott:
@Jolg42 
Just a note that this issue is still present in 4.8.0

aqrln:
Thank you for your contribution @maxsynnott and sorry for taking long to respond!

This is definitely an incremental improvement, and I wouldn't be strictly opposed to merging this, however to me it feels like it only hides the problem rather than solves it. One of the main purposes of this message is to tell users what exactly to use instead of `rejectOnNotFound`, and this will still remain broken even after these changes. Arguably, the new message might even be worse in that regard: previously, it would imply that some alternative exists (even though it couldn't tell what the alternative is exactly), and now it might look like the functionality will be completely gone in Prisma 5.

I think the alternatives are:

* Print a hardcoded generic message instead of removing the second sentence, something along the lines of "Please use `findUniqueOrThrow` or `findFirstOrThrow` method instead"
* Defer the warning until `findFirst` or `findUnique` is called for the first time
* Maybe a combination of the previous two (i.e., print a generic warning first, probably even in the `PrismaClient` constructor and not during the first request, and then print a specific suggestion first time either `findFirst` or `findUnique` is called)

WDYT?

/cc @SevInf @miguelff

maxsynnott:
I've implemented the first of your suggestions @aqrln .

Sorry for the late response. Busy times here. 

I think warning on the constructor and every following `findFirst` and `findUnique` (where `rejectOnNotFound` is not overrided) is probably the best solution.

Honestly though I don't have the time at the moment to implement that haha. 

If anyone else would like to I think it would be a great addition. 

I still think the current changes in this PR would be beneficial in the meantime to remove the confusing `undefined` deprecated messages.

maxsynnott:
@aqrln 
Just a note that this issue is still present in 4.9.0

Jolg42:
I'll close this now as we'll actually remove `rejectOnNotFound` in the next weeks, see https://github.com/prisma/prisma/issues/19315

PR #16452, opened 2022-11-25
feat(cli): add --generator option to generate cmd

### LINKED ISSUE #10527: Specify the generator to use for sharing schema files in different environments
### What I want to do
prisma's schema structure is simple, clear, and brilliant!
I am planning to use prisma schema in several programming languages.

For example, let's assume the following case:
Nodejs will handle the web server functions, and python will handle the once-a-day batch aggregation from the database.
I want to use `prisma-client-js` to connect from node and `prisma-client-py` to connect from python.
I believe that the benefit gained is quite large if one schema can be shared by multiple environments and programming languages.
(For example, it could be a shared language for exchanging data such as `protocol-buffer`.)

### Current issues
There are three items living together in the `schema.prisma` file: `datasource`, `generator`, and `model`.
Of these, only `model` is an important part of the data handling schema.
The `datasource` part is useful for generating connection methods, but the `generator` part is more dependent on each environment.
For example, if you want to share a schema between the above two languages, a simple description would look like this.
```prisma
generator client-js {
  provider = "prisma-client-js"
}

generator client-py {
  provider = "prisma-client-py"
}
```

However, this will not work as expected, because one of the two `generator` will output an error if both are not installed.

## Suggested solution
I need a way to select and execute only what I need in the arguments.
For example :
```
prisma generate --schema ./schema/schema.prisma --generator client-js
```

It would also be great if you could specify multiple generators.
Alternatively, it may be better to have the generator part specified by each user as an add-on.

## Alternatives.
There is currently a way to force a workaround using environment variables.
```prisma
generator client {
  provider = env(PRISMA_CLIENT)
}
```

I have created a sample that uses this method.
https://github.com/arika0093/prisma-multi-language

However, the drawback is that there is a lot of work to specify, and it is a bit of a hack.

## Additional context

Creating a schema first, and then automatically creating the fetch code and database from there, will cause a paradigm shift in system development.
I think it would be pretty cool to have a system where only the `schema.prisma` is managed in git and each project can refer to it from there.
What are your thoughts on this? I'd appreciate your input.

yukukotani:
Closes #10527

Add `--generator` option to `prisma generate` command to specify which generator will be used.

### Example

schema:
```prisma
datasource db {
  provider = "sqlite"
  url      = "file:./dev.db"
}

generator client_1 {
  output          = "../node_modules/.prisma/client-1"
}

generator client_2 {
  output          = "../node_modules/.prisma/client-2"
}

generator client_3 {
  output          = "../node_modules/.prisma/client-3"
}

model User {
  id   Int    @id
  name String
}
```

`prisma generate --generator client_1 --generator client_2` generates `client-1` directory and `client-2` and won't generates `client-3` directory.


packages/internals/src/get-generators/getGenerators.ts
```diff
@@ -495,3 +497,18 @@ In case you want to fix this, you can provide ${chalk.greenBright(
...
+function filterGenerators(generators: GeneratorConfig[], generatorNames: string[]) {
+  if (generatorNames.length < 1) {
+    return generators
+  }
+
+  const filtered = generators.filter((generator) => generatorNames.includes(generator.name))
+
+  if (filtered.length !== generatorNames.length) {
+    const missing = generatorNames.find((name) => filtered.find((generator) => generator.name === name) == null)
+    throw new Error(`Generator ${chalk.bold(missing)} is not found.`)
```
yukukotani, line 510: Maybe this message can be made more accessible to understand but I have no idea

MadcowD:
@yukukotani Excited for this to land!

Jolg42:
@yukukotani Thanks for this PR, it looks quite good to me! 
We would need a review from someone from the Prisma Client Team before we can merge.
Note that it might take a few days, we are quite busy at the moment, so if there is no activity here for the next few days you know why 💚

packages/internals/src/get-generators/getGenerators.ts
Jolg42, line 510: Can missing be more than 1 generator here actually?
Maybe we could have something like
"The specified generator(s) `${chalk.bold(generatorNames.join(', ')}` do not exist in your Prisma schema."

yukukotani: @Jolg42 Sounds good! 386ef65

yukukotani:
@Jolg42 No problem, I'm not in a hurry

packages/internals/src/__tests__/getGenerators/getGenerators.test.ts
```diff
@@ -381,8 +381,7 @@ describe('getGenerators', () => {
   })
 
   it('basic - binaryTargets as env var - darwin, windows, debian', async () => {
-    process.env.BINARY_TARGETS_ENV_VAR_TEST =
```
janpio, line 384: Why the removal of these?

packages/internals/src/get-generators/getGenerators.ts
```diff
@@ -495,3 +497,18 @@ In case you want to fix this, you can provide ${chalk.greenBright(
...
+function filterGenerators(generators: GeneratorConfig[], generatorNames: string[]) {
+  if (generatorNames.length < 1) {
+    return generators
+  }
+
+  const filtered = generators.filter((generator) => generatorNames.includes(generator.name))
+
+  if (filtered.length !== generatorNames.length) {
+    const missings = generatorNames.filter((name) => filtered.find((generator) => generator.name === name) == null)
+    throw new Error(`The specified generator(s) ${chalk.bold(missings.join(', '))} do not exist in your Prisma schema`)
```
janpio, line 510: This error message is only possible when `--generator` is used, right?
Then it could specifically mention "The generator `foo` specified via `--generator` does not exist in your Prisma schema" for example.

Additionally, differentiating between the singular and plural case by counting `missings` would be 👨‍🍳 💋 here.

janpio (commented):
Looking good already.

packages/internals/src/__tests__/getGenerators/getGenerators.test.ts
yukukotani: oops maybe I made mistake on rebase. fixed.

packages/internals/src/get-generators/getGenerators.ts
yukukotani: @janpio You're right. [e59dc48](https://github.com/prisma/prisma/pull/16452/commits/e59dc48d5aa03896d7d78deb3fe561872ade837a)

packages/internals/src/__tests__/getGenerators/getGenerators.test.ts
```diff
@@ -800,4 +832,30 @@ describe('getGenerators', () => {
...
+    }
+
+    await expect(
+      getGenerators({
+        schemaPath: path.join(__dirname, 'multiple-generators-schema.prisma'),
+        dataProxy: false,
+        providerAliases: aliases,
+        generatorNames: ['predefined-generator-1', 'invalid_generator'],
+      }),
+    ).rejects.toThrow('do not exist')
```
aqrln, line 859: Could we match the whole error message here?

If I understand correctly, only `invalid_generator` does not exist, so the error should have bee in singular, shouldn't it?

aqrln (commented):
Looks very good to me overall, just one remark/question.

Thank you!

packages/internals/src/__tests__/getGenerators/getGenerators.test.ts
yukukotani: Good catch, the test was wrong. Fixed in 4d150a3 :bow:

packages/internals/src/__tests__/getGenerators/__snapshots__/getGenerators.test.ts.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`getGenerators fail if generator not found 1`] = `"The generator [1minvalid_generator[22m specified via [1m--generator[22m does not exist in your Prisma schema"`;
```
Jolg42, line 3:
```suggestion
exports[`getGenerators fail if generator not found 1`] = `"The generator invalid_generator specified via --generator does not exist in your Prisma schema"`;
```

Jolg42: Because test fails with https://github.com/prisma/prisma/actions/runs/3742671182/jobs/6353907985#step:10:1002
```
    Snapshot: "The generator invalid_generator specified via --generator does not exist in your Prisma schema"
    Received: "The generator invalid_generator specified via --generator does not exist in your Prisma schema"
```

aqrln (approved):
Looks very good to me, thank you for the contribution and sorry for the delay in reviewing! I went ahead and changed the failing test to strip escape sequences explicitly, which should fix the failure on Buildkite. An alternative would've been to set the `NO_COLOR` env var and use `expect(...).toThrowErrorMatchingInlineSnapshot()` in all tests in the file, but I think it's be more appropriate to repeat the same pattern in the new test and then refactor the whole file separately if necessary.

aqrln:
Vitess failures are unrelated, and the same tests currently fail on main. I'm going to go ahead and merge this.

janpio:
Can you please create a docs issue @aqrln? Thanks.

eh-93:
So this would allow generating a python and js prisma client from the same prisma schema?

PR #16491, opened 2022-11-28
fix(validate,format): disable warnings when PRISMA_DISABLE_WARNINGS is truthy

Jolg42:
Closes https://github.com/prisma/schema-team/issues/373

Also
- turn warn into a lazily evaluated env var for tests


jkomyno (approved):
Approved as discussed privately

danstarns (approved):

PR #16496, opened 2022-11-28
perf(client): Cache extensions computations

SevInf:
Replaces extensions array with a linked list of extensions
(`MergedExtensionsList` class). This class is also responsible for resolving
various computations we do when applying extensions:

- List of all client extension methods with naming conflicts resolved
- List of all model extensions (per model) with naming conflicts
  resolved
- List of all computed fields to apply to specific model (with all
  conflicts resolved)
- List of all query callbacks to call for each model/action combination.

There are number of performance advantages here:
- We never recompute what haven't changed: if user adds `client`
  extensions, we won't recompute all `result` depenencies. If user adds
  computed field to `Account` model, we won't recompute computed fields
  lists for all other models.
- We cache all computations: provided extension list didn't change we
  won't be computing the list of callbacks for `query` extensions or
  list or resoling computed fields for every request
- When we do recompute stuff, we usually do it in incremental steps. For
  example, if user has a fork with 10 extensions and adds 11th, we only
  apply this extension to previously computed result and won't traverse
  whole list of extensions.
- Because of the last point, it generally produces simpler code. See
  `getComputedFields` diff or all of the `apply*Extensions` methods.


millsp (approved):

.github/workflows/test.yml
```diff
@@ -1039,7 +1039,7 @@ jobs:
           GITHUB_CONTEXT: ${{ toJson(github) }}
           # allow Node.js to allocate at most 14GB of heap on macOS and 7GB on Windows
```
Jolg42, line 1040:
```suggestion
          # Allow Node.js to allocate at most a certain amount heap on macOS and a different amount on Windows
          # Note that this value should be lower than the total amount of RAM available
```

PR #16501, opened 2022-11-28
feat: add validation warnings to other cmds

jkomyno:
This PR adds support for validation warnings in:
- [x] `prisma generate` (it already displayed warnings via `printConfigWarnings`, which we should consider deprecated)
- [x] `prisma db push`
- [x] `prisma db pull`

Related tests have also been updated to reflect that warnings can be shown alongside errors, and are hidden when `process.env.PRISMA_DISABLE_WARNINGS` is truthy.

This PR contributes to https://github.com/prisma/team-orm/issues/90.

## Screenshots

- `prisma generate`
Screenshot 2022-11-29 at 13 10 39

- `prisma db pull`
Screenshot 2022-11-29 at 13 09 41

- `prisma db push`
Screenshot 2022-11-29 at 13 11 18


akingundogdu (approved):

Druue (approved):

janpio (approved):

packages/internals/src/get-generators/getGenerators.ts
```diff
@@ -120,8 +119,6 @@ export async function getGenerators(options: GetGeneratorOptions): Promise<Gener
     throw new Error(missingDatasource)
   }
 
-  printConfigWarnings(config.warnings)
-
```
Jolg42, line 124: Are the config warnings the same warnings as our lint warnings?
If so, we could check where it's used and remove it later

Jolg42 (changes requested):
~~I'm still testing this but I think we should block this.~~

~~Currently, it breaks `npx db pull --print` output~~
Screenshot 2022-11-29 at 13 48 43

Jolg42:
`db pull` is now quite confusing 
Screenshot 2022-11-29 at 13 47 15

Jolg42:
`prisma generate` has 1 warning and 1 info about the same thing 
Screenshot 2022-11-29 at 13 54 10

It looks like this info should be removed (also it looks like there should have been 3 info because that's actually 3 deprecated preview features in the schema)

jkomyno:
Responding to @Jolg42's comments.

- `prisma generate`: It looks like that `The preview flag "${flag}" is not needed anymore` is coming from [`getDmmf`](https://github.com/prisma/prisma/blob/3077f505d317e3b3f1d2b20682f6d0273f5c59c4/packages/internals/src/engine-commands/getDmmf.ts). There is no reason for this warning logic to be defined in TypeScript, I can also tell it's outdated w.r.t. the deprecated feature flags defined in Rust (for example, it doesn't take the deprecated `referentialIntegrity` preview feature into account). 

- `prisma db pull`: I think that's fine, and that @janpio had already stated that the chance of a user having both validation warnings and introspection warnings at once was so slim that we don't need to optimise the output messages for this corner case (at least, not now)

- `prisma db pull --print`: That's a valid concern, thanks for bringing it up. A test for `--print` on a schema with warnings is needed

janpio:
(A simple improvement to `db pull` could be to rename the `WARNING` to `Re-Introspection Information` or something like that. Those are not really _warnings_ anyway.)

jkomyno:
**TODO**:
- [ ] remove outdate preview feature warnings in `BinaryEngine.ts` and `getDmmf.ts`
- [ ] adjust output in `prisma db pull`, so that the validation warnings appear immediately before the introspection warnings
- [ ] add new tests for warnings in `prisma db pull --print`

PR #16515, opened 2022-11-29
feat(client): Prisma.defineExtension

millsp:


packages/client/src/runtime/core/extensions/$extends.ts
```diff
@@ -78,25 +78,25 @@ type QueryOptions = {
...
+
+  // re-apply models to the extend client: they always capture specific instance
+  // of the client and without re-application they would not see new extensions
   const oldClient = unapplyModelsAndClientExtensions(this)
-  const newExtensions =
-    typeof extension === 'function' ? this._extensions.append(extension(this)) : this._extensions.append(extension)
   const newClient = Object.create(oldClient, {
     _extensions: {
-      get: () => newExtensions,
+      get: () => this._extensions.append(extension),
```
SevInf, line 97: This will re-do `append` every time property is accessed. I think what we want is `value: this._extensions.append(extension)`

PR #16542, opened 2022-11-30
fix(client): Force `result` extensions to log correct properties

### LINKED ISSUE #16554: `console.log` outputs incorrect value when `result` extensions are used
This code

```ts

  result: {
    user: {
      fullName: {
        needs: { firstName: true, lastName: true },
        compute(user) {
          return `${user.firstName} ${user.lastName}`
        }
      }
    }
  }
})


async function main() {
  await prisma.user.deleteMany()
  await prisma.user.create({
    data: {
      firstName: 'Boaty',
      lastName: 'McBoatface'
    }
  })

  const data = await prisma.user.findFirst({
    select: {
      fullName: true
    }
  })

  console.log(data)

}

main()
  .then(async () => {
    await prisma.$disconnect()
  })
  .catch(async (e) => {
    console.error(e)
    await prisma.$disconnect()
    process.exit(1)
  })
```

Outputs `{firstName: 'Boaty', lastName: 'McBoatface' }` which is incorrect. Expected result is and object with only `fullName` property.
However, this:

```ts
 console.log((data as any).firstName)
  console.log((data as any).lastName)
  console.log(data?.fullName)
```

Outputs:

```
undefined
undefined
Boaty McBoatface
```

So, the returned value is correct, just `console.log` displays it incorrectly.

SevInf:
Node.js `util.inspect` and console functions deliberately avoid
triggering any of the proxy traps. If result extensions are used, that
forces us to show completely incorrect result - it will show the target
which would not match what user is actually getting there.

Fix #16554

### DIFF
--- a/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.test.ts
+++ b/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.test.ts
@@ -1,3 +1,5 @@
+import util from 'util'
+
 import { createCompositeProxy } from './createCompositeProxy'
 
 test('forwards properties to the target', () => {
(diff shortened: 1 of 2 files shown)


aqrln (approved):
nice!

millsp (approved):

PR #16560, opened 2022-11-30
ci(vitess): faster vitess tests in functional tests

Jolg42:
Before
Screenshot 2022-11-30 at 19 35 20

After
Screenshot 2022-11-30 at 19 44 52

All tests for a given PR now take 18min to complete, before ~38min ⏩

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -32,7 +32,6 @@ services:
 
   # Planetscale
   # From https://github.com/prisma/prisma-engines/blob/976a00ae3c30ab9507fa742986c9c6f5327ba10f/docker-compose.yml
-
   vitess-8:
     image: vitess/vttestserver:mysql80@sha256:1f8c98f716649ac095bae78ead88f522eaf236e9e03baa770b6690d61cbbe29a
     restart: always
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -50,32 +49,6 @@ services:
       NUM_SHARDS: '1' # unused in testing, but required by vttestserver
       MYSQL_BIND_HOST: '0.0.0.0'
       FOREIGN_KEY_MODE: 'disallow'
-      VT_DIALECT: 'mysql80'
```
Jolg42, line 53: I didn't find anything related to this param in vitess docs or GitHub so I assume it is safe to remove

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -208,7 +208,16 @@ export function setupTestSuiteDbURI(suiteConfig: Record<string, string>, clientM
       return { envVarName, newURI }
     })
 
-  const databaseUrl = newURI.replace(DB_NAME_VAR, dbId)
+  let databaseUrl = newURI
+  // Vitess takes about 1 minute to create a database the first time
+  // So we can reuse the same database for all tests
```
janpio, line 213: Why do we usually _not_ do that?

docker/docker-compose.yml
janpio: I think we originally copied it from https://github.com/planetscale/vitess-framework-testing where we got our first docker config for Vitess

jkomyno: I believe I added it a few weeks back. It explicitly specifies that we consider Vitess as a flavor of `mysql:8+`. I believe it's the default option btw

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
Jolg42: Because we need full isolation for tests usually.
For Vitess we can break isolation without problem now but when we run more tests on vitess we would need to adapt this, maybe we could pass `options?.databaseName,` and if provided use that instead of a cuid.
Which would make it flexible to control isolation where needed.

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -115,7 +115,17 @@ export async function setupTestSuiteDatabase(
...
+    const providerFlavor = suiteConfig['providerFlavor'] as ProviderFlavor | undefined
+    // `--force-reset` is great but only using it where necessary makes the tests faster
+    // Since we have full isolation of tests / database,
+    // we do not need to force reset
+    // But we currently break isolation for Vitess (for faster tests),
+    // so it's good to force reset in this case
+    if (providerFlavor === ProviderFlavors.VITESS_8) {
+      dbpushParams.push('--force-reset')
+    }
+    await DbPush.new().parse(dbpushParams)
```
Jolg42, line 128: Hard to say without more benchmark data, but it might be ~1 minute saved

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
SevInf: Note that there are several jest workers running in parallel.  So, if one of the workers resets the and the other one currently runs tests against it, we might have a problem. We might either: 
- ensure each workers talk to it's own DB. I think easiest way to do so would be to include `pid` in db name.
- limit worker count to 1 - but that might negate all of performance gains

Jolg42: We only have tests for vitess in `relationMode` directory and they run separately from other functional tests.
We should change this later if we add vitess tests to the functional tests probably.
But for now we have a clear gain of at least 10-15min where the relationMode=prisma tests would be the last action running.
So we win 10-15 waiting time for each PR

SevInf (approved):

Jolg42:
Note from discussion with @SevInf 

But you are not limiting max workers?
No but for relationMode test files we run them one by one on GitHub action, so it's limited to 1 worker in the end.

PR #16563, opened 2022-12-01
fix(client): query extensions type and runtime issues

### LINKED ISSUE #16551: Prisma Client Extensions: `args` is `undefined`
### Bug description

Hello Prisma!

I've just tried out the new preview feature for Prisma Client Extensions and I'm getting an issue with `args` being undefined. I've been referencing the sample [here](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/query#add-or-modify-a-custom-query-for-a-specific-operation-in-a-specific-model).

For my experimentation, I'm trying to create a Prisma Client instance that scopes down queries to records belonging to a specific user or subset of users.

In this case, I'm hard-coding a `customerId` to work with.

```ts
const customerId = "cl9ztaj7a0000pko377xgkq0u";

const nonAdminPrisma = new PrismaClient().$extends({
  query: {
    invoice: {
      async findMany({ model, operation, args, query }) {
        console.log("args", args); // undefined

        args.where = { customer: { id: customerId }, ...args.where };

        const invoices = await query(args);
        return query(invoices);
      },
    },
  },
});
```

I then try to call this client as follows:

```ts
const invoices = await nonAdminPrisma.invoice.findMany();
```

This throws with the following:

```bash
args undefined
(node:12311) UnhandledPromiseRejectionWarning: PrismaClientExtensionError: Error caused by an extension: Cannot read property 'where' of undefined
    at /Users/ryanchenkie/Desktop/code/prisma-examples/node_modules/@prisma/client/runtime/index.js:27989:63
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at PrismaClient._request (/Users/ryanchenkie/Desktop/code/prisma-examples/node_modules/@prisma/client/runtime/index.js:35900:16)
```

I've tried using both a specific model (`invoice`) and also `$allModels` and I get the same behavior. 

Thanks!

### How to reproduce

1. Use the `clientExtensions` preview feature.
2. Set up a new client using `$extends` and use the `query` block.
3. Try to access something like `where` on `args`
4. Find `args` is undefined

### Expected behavior

_No response_

### Prisma information

```prisma
generator client {
}

datasource db {
  provider = "postgresql"
}

model Customer {
  id             String    @id @default(cuid())
  name           String
  email          String    @unique
  districtNumber Int       @default(0)
  invoices       Invoice[]
  createdAt      DateTime  @default(now())
  updatedAt      DateTime  @updatedAt
  address        String?
  active         Boolean   @default(true)
}

model Invoice {
  id         String   @id
  amount     Int
  dueDate    DateTime
  customer   Customer @relation(fields: [customerId], references: [id])
  customerId String
  createdAt  DateTime @default(now())
  updatedAt  DateTime
}

```

```ts
const customerId = "cl9ztaj7a0000pko377xgkq0u";

const nonAdminPrisma = new PrismaClient().$extends({
  query: {
    invoice: {
      async findMany({ model, operation, args, query }) {
        console.log("args", args); // undefined

        args.where = { customer: { id: customerId }, ...args.where };

        const invoices = await query(args);
        return query(invoices);
      },
    },
  },
});

async function main() {
  const invoices = await nonAdminPrisma.invoice.findMany();

  console.log("the invoices", invoices);
}
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: 14.9.3

### Prisma Version

```
4.7.0
```

millsp:


packages/client/src/runtime/core/extensions/applyQueryExtensions.ts
```diff
@@ -30,7 +30,7 @@ function iterateAndCallQueryCallbacks(
     return queryCbs[i]({
       model: params.model,
       operation: params.action,
-      args: klona(params.args),
+      args: klona(params.args ?? {}),
```
millsp, line 33: The actual fix

packages/client/tests/functional/extensions/enabled/query.ts
```diff
@@ -862,6 +862,47 @@ testMatrix.setupTestSuite(
...
+    })
+
+    test('passing incorrect argument errors', () => {
+      prisma.$extends({
+        query: {
+          user: {
+            async findFirst({ args, query, operation, model }) {
+              const user = await query(args)
+              // @ts-expect-error
+              return query(user)
```
millsp, line 900: A type fix to prevent mistakes. `query` previously did not typecheck inputs (just `object`), now has full safety.

SevInf (approved):

Jolg42:
Let's merge and ignore the failures, https://github.com/prisma/prisma/pull/16573 was the problem

PR #16573, opened 2022-12-01
ci(gh-tests): replace ubuntu-latest by ubuntu-20.04

Jolg42:
We have tests failing because of the upgrade, see
https://github.com/prisma/prisma/pull/16572


Jolg42:
Other example
https://github.com/prisma/prisma/actions/runs/3589130747/jobs/6045483734#step:8:1124

PR #16574, opened 2022-12-01
fix: client handling of batched queries when some fail, and some not

miguelff:
Addresses https://github.com/prisma/prisma-engines/pull/3458#issuecomment-1332626419

Depends on merging and updating engine hashes for  https://github.com/prisma/prisma-engines/pull/3458 as I need to write some unit tests to prevent regressions. 

Before the release of native findXOrThrow, there were no instances of batching where commonly some promises failed while others succeeded. And the code handling that scenario was broken. This PR attempts to fix it.

### DIFF
--- a/packages/client/tests/functional/find-unique-or-throw-batching/tests.ts
+++ b/packages/client/tests/functional/find-unique-or-throw-batching/tests.ts
@@ -6,27 +6,39 @@ import type { PrismaClient } from './node_modules/@prisma/client'
 
 class NotFoundError extends Error {}
 declare let prisma: PrismaClient
-
-const id = faker.random.alphaNumeric(11)
 const missing = faker.random.alphaNumeric(11)
 
 testMatrix.setupTestSuite((suiteConfig, suiteMeta, clientMeta) => {
   beforeEach(async () => {
     await prisma.user.deleteMany()
-    await prisma.user.create({
-      data: {
-        id,
-      },
-    })
   })
 
   testIf(!clientMeta.dataProxy)('batched errors are serialized properly', async () => {
+    const id = await prisma.user
+      .create({
+        data: {},
+      })
+      .then((u) => u.id)
+
+    const id2 = await prisma.user
+      .create({
+        data: {},
+      })
+      .then((u) => u.id)
+
     const found = prisma.user.findUniqueOrThrow({ where: { id } })
-    const foundToo = prisma.user.findUnique({ where: { id } })
-    const notFound = prisma.user.findUniqueOrThrow({ where: { id: missing } })
-    const result = await Promise.allSettled([found, foundToo, notFound])
+    const foundToo = prisma.user.findUniqueOrThrow({ where: { id: id2 } })
+    const result = await Promise.allSettled([found, foundToo])
     expect(result).toEqual([
       { status: 'fulfilled', value: { id: id } },
+      { status: 'fulfilled', value: { id: id2 } },
+    ])
+
+    await prisma.user.delete({ where: { id: id2 } })
+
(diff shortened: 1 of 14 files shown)


packages/client/tests/functional/find-unique-or-throw-batching/tests.ts
```diff
@@ -0,0 +1,43 @@
...
+        { status: 'fulfilled', value: { id: id } },
+        { status: 'fulfilled', value: { id: id } },
+        { reason: new NotFoundError('No User found'), status: 'rejected' },
+      ])
+    })
+  },
+  {
+    optOut: {
+      from: ['postgresql', 'mysql', 'mongodb', 'cockroachdb', 'sqlserver'],
+      reason:
```
SevInf, line 39: Normally we still run them on all providers, unless there is some reason while it **can't ** run on all of them.

packages/client/tests/functional/find-unique-or-throw-batching/tests.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    test('batched errors are serialized properly', async () => {
+      const found = prisma.user.findUniqueOrThrow({ where: { id } })
+      const foundToo = prisma.user.findUnique({ where: { id } })
+      const notFound = prisma.user.findUniqueOrThrow({ where: { id: missing } })
+      const result = await Promise.allSettled([found, foundToo, notFound])
+      expect(result).toEqual([
+        { status: 'fulfilled', value: { id: id } },
+        { status: 'fulfilled', value: { id: id } },
+        { reason: new NotFoundError('No User found'), status: 'rejected' },
```
SevInf, line 32:
```suggestion
        { reason: exepct.objectMatching({ code: 'P2025' }) },
```

miguelff: Unfortunately, this doesn't work directly (even while replacing the typo), and I cannot find any other example at first glance, where that construct is used

miguelff:
The regression affected binary and library engines. For data proxy, there's a different problem that manifests when executing the same test (The test exercises two different problems)

The problem in particular (after debugging the data proxy) is that the engine adapter is not able to discern errors coming from individual queries in a batch, and as a consequence, doesn't reject a promise that errored.


If we do this:
```ts
    const found = prisma.user.findUniqueOrThrow({ where: { id } })
    const notFound = prisma.user.findUniqueOrThrow({ where: { id: missing } })
    const result = await Promise.allSettled([found, notFound])
```
We get:

```js
    [
      { status: 'fulfilled', value: { id: id } },
      { reason: new NotFoundError('No User found'), status: 'rejected' },
    ]
```
while with the data proxy, we get.

```js
    [
      { status: 'fulfilled', value: { id: 'xqwhv5rgkt2' } },
      { status: 'fulfilled', value: undefined }
    ]
```

I leave to @aqrln the investigation and address of the problem described here which. To reproduce it, is enough to unskip the test provided in this patch.

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -392,8 +392,8 @@ ${searchedLocations
         ) {
           errorText += `
 You already added the platform${this.generator.binaryTargets.length > 1 ? 's' : ''} ${this.generator.binaryTargets
-              .map((t) => `"${chalk.bold(t.value)}"`)
-              .join(', ')} to the "${chalk.underline('generator')}" block
+            .map((t) => `"${chalk.bold(t.value)}"`)
+            .join(', ')} to the "${chalk.underline('generator')}" block
```
miguelff, line 396: Irrelevant: formatting

aqrln (approved):

PR #16655, opened 2022-12-06
ci: Fix a couple of memory leaks in test setups

### LINKED ISSUE #16594: Investigate memory issues on Windows
Our CI pipelines fail with out of memory issues since recently. It looks like there might be a memory leak again. We need to take a look into it and fix the problem.

SevInf:
This fixes 2 memory leaks in our test setup:

1. Jest passes every file through transform and caches the results
on disk and in memory. That included every generated client as well, so
they were kept in memory from the time they were first loaded until
process finishes, even though they are needed only for 1 test suite.
Fixed by using `transformIgnorePatterns` options in jest config.

2. `debug` library keeps a list of last 100 logs that would be used for
panic error messages later. If error object gets into this list, it will
keep whole jest sandbox alive until process finishes. Fixed by clearing
logs on `$disconnect`

Fix #16594

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -1,5 +1,5 @@
 import { Context, context } from '@opentelemetry/api'
-import Debug from '@prisma/debug'
+import Debug, { clearLogs } from '@prisma/debug'
 import {
   BatchTransactionOptions,
   BinaryEngine,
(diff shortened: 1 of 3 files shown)


packages/client/tests/functional/jest.config.js
```diff
@@ -1,10 +1,18 @@
...
+const path = require('path')
+
+const runtimeDir = path.dirname(require.resolve('../../runtime'))
+const packagesDir = path.resolve('..', '..', '..')
 
 module.exports = () => {
   const configCommon = {
     testMatch: ['**/*.ts', '!(**/*.d.ts)', '!(**/_utils/**)', '!(**/_*.ts)', '!(**/.generated/**)'],
-    transformIgnorePatterns: [],
+    transformIgnorePatterns: [
```
Jolg42, line 11: Can you add a comment about this?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -591,6 +591,7 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
         e.clientVersion = this._clientVersion
         throw e
       } finally {
+        clearLogs()
```
Jolg42, line 594: Can you add a comment, why it's needed?

Jolg42:
Looks good to me, let's make it obvious with code comments why these new lines are important 👍🏼

millsp (approved):

aqrln (approved):

Jolg42 (approved):

Issue #12921, opened 2022-04-21
next.js ENOENT trying to open schema.prisma


SimonBerens:
### Bug description

In a monorepo, when exporting `PrismaClient` from one package and consuming it in a next.js api route, I get 
```
error - Error: ENOENT: no such file or directory, open 'C:\Users\sbere\Desktop\Projects\prisma-pnpm-workspace-test\apps\frontend\.next\server\pages\api\schema.prisma'
```

### How to reproduce

1. Git clone https://github.com/SimonBerens/prisma-pnpm-workspace-test
2. run
```bash
pnpm install
cd packages/prisma
pnpx prisma generate
cd ../../apps/frontend
pnpm run dev
```
3. send a GET request to `localhost:3000/api/hello` 

### Expected behavior

The api runs as expected

### Prisma information

```  
prisma:tryLoadEnv Environment variables not found at null +0ms
  prisma:tryLoadEnv Environment variables not found at undefined +0ms
  prisma:tryLoadEnv No Environment variables loaded +1ms
  prisma:tryLoadEnv Environment variables not found at null +1ms
  prisma:tryLoadEnv Environment variables not found at undefined +0ms
  prisma:tryLoadEnv No Environment variables loaded +0ms
  prisma:client dirname C:\Users\sbere\Desktop\Projects\prisma-pnpm-workspace-test\apps\frontend\.next\server\pages\api +0ms
  prisma:client relativePath ..\..\..\..\..\..\packages\prisma +0ms
  prisma:client cwd C:\Users\sbere\Desktop\Projects\prisma-pnpm-workspace-test\packages\prisma +0ms
  prisma:client clientVersion: 3.12.0 +2ms
  prisma:client clientEngineType: library +0ms
```

### Environment & setup

- OS: Windows 11
- Database: PostgreSQL
- Node.js version: v16.14.2

### Prisma Version

```
prisma                  : 3.12.0
@prisma/client          : 3.12.0
Current platform        : windows
Query Engine (Node-API) : libquery-engine 22b822189f46ef0dc5c5b503368d1bee01213980 (at ..\..\node_modules\.pnpm\@prisma+engines@3.12.0-37.22b822189f46ef0dc5c5b503368d1bee01213980\node_modules\@prisma\engines\query_engine-windows.dll.node)
[...]

Issue #12823, opened 2022-04-14
NextJS try to open "schema.prisma" at wrong path.


GrassHopper42:
### Bug description

I'm using Lerna and Yarn workspace for monorepo.

```
root
└ packages
  └ prisma      (@my-app/primsa)
    └ schema.prisma    //generated
  └ server
  └ web       (@my-app/web)
└ prisma
  └ schema.prisma
```

I instantiating new client in web package just for next-auth. There is another instantiated client in server package.

```typescript
...
import { PrismaClient } from "@my-app/prisma";

const prisma = new PrimsaClient();

export default NextAuth({
  adapter: PrismaAdapter(prisma),
...
```

If I run dev and try to signin, server throw this error
```bash
warn  - ../prisma/runtime/index.js
Module not found: Can't resolve 'encoding' in 'root/packages/prisma/runtime'
error - Error: ENOENT: no such file or directory, open 'root/packages/web/.next/server/pages/api/auth/schema.prisma'
```

I found similar issue, but it fixed in `@3.12.0`.
And I try again after delete yarn cache, but nothing to change.

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
  output   = "../packages/prisma"
}

datasource db {
  provider = "mongodb"
  url      = env("DATABASE_URL")
}

model Account {
  id                String   @id @default(auto()) @map("_id") @db.ObjectId
  createdAt         DateTime @default(now())
  updatedAt         DateTime @updatedAt
  user              User     @relation(fields: [userId], references: [id])
  userId            String   @db.ObjectId
  type              String
  provider          String
  providerAccountId String
  refresh_token     String?  @db.String
  access_token      String
  expires_at        Int?
  token_type        String?
  scope             String?
  id_token          String?  @db.String
  session_state     String?

  @@unique([provider, providerAccountId])
}

model Session {
  id           String   @id @default(auto()) @map("_id") @db.ObjectId
  sessionToken String   @unique
  user         User     @relation(fields: [userId], references: [id])
  userId       String   @db.ObjectId
  expires      DateTime
}

model VerificationToken {
  id         String   @id @default(auto()) @map("_id") @db.ObjectId
  identifier String
  token      String   @unique
  expires    DateTime

  @@unique([identifier, token])
}

model User {
  id            String    @id @default(auto()) @map("_id") @db.ObjectId
  createdAt     DateTime  @default(now())
  updatedAt     DateTime  @updatedAt
  email         String    @unique
[...]

PR #16684, opened 2022-12-08
fix: improve interaction between nextjs, yarn 3 and custom output paths

Schmavery:
I've been trying to use prisma with yarn 3 which doesn't have a standard node_modules. I used a [custom output path](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client#using-a-custom-output-path) to create the client in my ./src folder instead as suggested in a couple of related issues.

This seemed to work but then I got error at runtime similar to issues like #12823, #9435 and #12921 . I did some digging and realized that the client is failing to find the `schema.prisma` at runtime due to the findSync search being depth first. This makes it quite dependent on the order of the contents of the root directory which might explain why some people seem to "randomly" run into the problem in several different configurations?

In my case, it was picking up the `./.yarn/unplugged/prisma-npm-<version>-<hash>/prisma-client` folder before the `./src/prisma/prisma-client`, the first of which didn't have the prisma-schema. One option would have been to hardcode a check that ignores the ".yarn" folder, but it seems in general that it would be less surprising behavior to pick the match closest to the root of the search.

I have been using this as a patch locally but wanted to share in case it is useful to anyone else. I tried to match the existing style where possible but feel free to tweak it.


aqrln:
Hi @Schmavery and thank you for the PR! As we're pretty busy at the moment, I apologize if there's any delay in reviewing, but we'll try to review it as soon as possible. Meanwhile, would you mind fixing the build errors? Thank you again!

```
Error: packages/client build: src/runtime/utils/find.ts(94,15): error TS2339: Property 'item' does not exist on type '{ item: Dirent; dir: string; } | undefined'.
Error: packages/client build: src/runtime/utils/find.ts(94,21): error TS2339: Property 'dir' does not exist on type '{ item: Dirent; dir: string; } | undefined'.
Error: packages/client build: src/runtime/utils/find.ts(125,25): error TS2345: Argument of type '"d" | "f" | "l" | undefined' is not assignable to parameter of type '"d" | "l"'.
```

Jolg42:
Note: we probably need to look at this, not only from a yarn 3 point of view (I mean see how it affects, package managers, lik pnpm, and workspaces for example)

@Schmavery we will need tests for this yarn 3 issue that motivated your issue.
Did you provide a minimal reproduction somewhere? It would be great to have one for your case here.

Schmavery:
Hey all, sorry for the build error, will fix those.

Was just working on a minimal repro and realized there was more going on than I expected. Basically the fact that I'm using nextjs seems to be breaking the normal __dirname-based resolution (otherwise the findSync code never even gets called and everything seems to work fine even with yarn 3 😄 ).

Will share the actual repro soon, thanks for taking a look so quickly!

Schmavery:
Here is the repro for now:
https://github.com/Schmavery/prisma-findsync-repro

Will look into setting up the proper prisma contribution environment and fixing the errors when I get a chance.

Schmavery:
@aqrln I fixed the type error and now most of the checks appear to be passing. I admit I'm not sure what the actual errors are in the 2 failing tests. Happy to try and fix but I could use another pointer when you get some time.

janpio:
(Restarted both failing test runs for the case this was flakey)

packages/client/src/runtime/utils/find.ts
```diff
@@ -68,36 +71,45 @@ export function findSync(
...
     // we check that the root is a directory
     if (direntToType(statSync(realRoot)) !== 'd') {
       return found
     }
 
-    // we list the items in the current root
-    const items = readdirSync(root, { withFileTypes: true })
+    // we list the items in the search root
+    const items = readdirSync(realRoot, { withFileTypes: true })
+    const stack = [...items.map((i) => ({ item: i, dir: root }))]
```
aqrln, line 81: nit: technically it's a queue, not a stack, a stack would lead to depth-first search

Schmavery: hilarious, fixed

PR #16721, opened 2022-12-10
test(client, cli): local ecosystem tests

millsp:
Provide a way to write ecosystem tests locally easily and with a fast feedback loop. It is at prototype stage, which works but probably isn't bullet proof yet, for that reason tests are not yet hooked into CI. They can be run manually via `pnpm run test:ecosystem` (in `client`) for now. I aim to use this to write many of the reproductions/tests we need for fixing serverless/lifecycle issues.


packages/client/tests/ecosystem/_example/_steps.ts
```diff
@@ -0,0 +1,20 @@
...
+  test: async () => {
+    await $`ts-node src/index.ts`
+    await $`pnpm exec prisma -v`
+    await $`pnpm exec jest`
+  },
+  finish: async () => {
+    await $`echo "done"`
+  },
+  // keep: true,
+})
```
millsp, line 20: Define the test execution steps, similar to what we have in ecosystem tests.

packages/client/tests/ecosystem/_example/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@vercel/ncc": "0.36.0",
+    "@prisma/client": "../../prisma-client-0.0.0.tgz",
+    "prisma": "../../prisma-0.0.0.tgz"
```
millsp, line 14: Client and CLI are linked from a local module tarball that has been packaged but not published.

packages/client/tests/ecosystem/_utils/runContainers.ts
```diff
@@ -0,0 +1,21 @@
+import { $ } from 'zx'
+
+async function main() {
+  await $`cd ${__dirname}/../../../../cli && sed -i 's/__bundleDependencies/bundleDependencies/g' package.json`
```
millsp, line 4: Activate dependency bundling so that engines package is bundled with the cli.

packages/client/tests/ecosystem/_utils/runContainers.ts
```diff
@@ -0,0 +1,21 @@
...
+  await $`cd ${__dirname}/../../../../cli && ECOSYSTEM=true pnpm pack --pack-destination ${__dirname}/../`
+  await $`cd ${__dirname}/../../../../cli && sed -i 's/bundleDependencies/__bundleDependencies/g' package.json`
+  await $`cd ${__dirname}/../../../../client && ECOSYSTEM=true pnpm pack --pack-destination ${__dirname}/../`
+  await $`docker compose -f ${__dirname}/docker-compose.yml build >/dev/null 2>&1`
+  await $`docker compose -f ${__dirname}/docker-compose.yml down >/dev/null 2>&1`
+  await $`docker compose -f ${__dirname}/docker-compose.yml up >/dev/null 2>&1`
+
+  const findErrors = await $`find "$(pwd)" -not -name .logs.0.txt -name .logs.*.txt`.quiet()
```
millsp, line 12: Running tests always have logs that are persisted. In general successful jobs are named with `.logs.0.txt`, the number in the middle is the exit code of the job.

packages/client/tests/ecosystem/_utils/standard.cmd.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+(
```
millsp, line 3: This script copies files (tarballs, tests, utils) into the container from the shared volume and prepares the test, puts it into `/test` while also logging its output into a file.

packages/client/tests/ecosystem/_utils/standard.dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node:16
+ARG NAME
+ENV TEST_NAME=$NAME
+
+RUN npm i -g pnpm zx typescript jest ts-node esbuild esbuild-register @swc/jest
```
millsp, line 5: This image always comes with these utilities.

helpers/compile/plugins/tscPlugin.ts
```diff
@@ -78,6 +78,13 @@ export const tscPlugin: (emitTypes?: boolean) => esbuild.Plugin = (emitTypes?: b
         await run(`tsc --project ${options.tsconfig} --paths null`)
       }
 
+      // for ecosystem tests, skip the bundling, just copy the types to go faster
```
millsp, line 81: Slightly different from reality, ideally I'd like to have a quick mode for local dev like this and maybe a mode that is slower and does the full type bundling. They should be equivalent, but we never now.

helpers/compile/build.ts
```diff
@@ -123,7 +123,7 @@ async function dependencyCheck(options: BuildOptions) {
   // we need to bundle everything to do the analysis
   const buildPromise = esbuild.build({
     entryPoints: glob.sync('**/*.{j,t}s', {
-      ignore: ['./src/__tests__/**/*'],
+      ignore: ['./src/__tests__/**/*', './tests/ecosystem/**/*'],
```
millsp, line 126: We don't check dependencies in ecosystem tests because tests are isolated from the build.

helpers/compile/build.ts
```diff
@@ -144,6 +144,8 @@ async function dependencyCheck(options: BuildOptions) {
  * @param options
  */
 export async function build(options: BuildOptions[]) {
+  if (process.env.ECOSYSTEM === 'true') return
```
millsp, line 147: When we trigger `pnpm pack` we actually don't want to build again to go faster. We re-use what has been already build; and also implies you ran `pnpm run watch/dev/build` before hand. In the future, I think I'll enable a full build + packaging mode to be much closer to reality, this could be done on CI.

packages/cli/helpers/copy-prisma-client.ts
```diff
@@ -9,10 +10,11 @@ const clientPath = path.dirname(require.resolve('@prisma/client'))
 const clientPkg = require('@prisma/client/package.json')
 
 // we compute the paths of the files that would get npm published
-const clientFiles = (clientPkg.files ?? []).map((file: string) => path.join(clientPath, file))
+const clientFiles = fg.sync(clientPkg.files, { cwd: clientPath, dot: true, onlyFiles: false })
```
millsp, line 13: Small improvement for future proofing, using a glob library to understand patterns in `files`. Using tooling from npm would be even better, but it is in magnitudes slower compared to this approach. Ideally, it would be great if we could remove this somehow.

packages/client/tests/ecosystem/_example/package.json
millsp, line 14: Local package is packed and made available in the docker containers.

packages/client/tests/ecosystem/_utils/run.ts
```diff
@@ -0,0 +1,73 @@
...
+  const restoreOriginal = async () => {
+    await $`cd ${tmpDir} && cp cli.package.json ${cliPkgJsonPath}`.quiet()
+    await $`cd ${tmpDir} && cp client.package.json ${clientPkgJsonPath}`.quiet()
+    await $`cd ${tmpDir} && cp client.runtime.d.ts ${clientRuntimeDtsPath}`.quiet()
+  }
+
+  // if process is killed by hand, ensure that package.json is restored
+  process.on('SIGINT', () => restoreOriginal().then(() => process.exit(0)))
+
+  // use bundleDependencies to directly include deps like @prisma/engines
```
millsp, line 33: Because we cannot publish to npm, we tell the package manager to `bundleDependencies`
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bundledependencies

packages/client/tests/ecosystem/_utils/run.ts
```diff
@@ -0,0 +1,73 @@
...
+  cliPkgJson.bundleDependencies = Object.keys(cliPkgJson.dependencies)
+  clientPkgJson.bundleDependencies = Object.keys(clientPkgJson.dependencies)
+
+  // write the modified package.json to overwrite the original package.json
+  await fs.writeFile(cliPkgJsonPath, JSON.stringify(cliPkgJson, null, 2))
+  await fs.writeFile(clientPkgJsonPath, JSON.stringify(clientPkgJson, null, 2))
+  // this is to avoid bundling types and locally link directly to the sources
+  await fs.writeFile(clientRuntimeDtsPath, `export * from '/client/src/runtime/index'`)
+  // TODO: have a fast mode (this one) and a full mode with type bundling
```
millsp, line 42: As mentioned before, not having a full build is nice to quickly test (ie. skipping type bundling here), but in a v2 we may want to enable a full build to test the package as it is published. Type bundling should produce compatible types, but could as well fail and we wouldn't notice that in ecosytem tests. So that's a future thing to enable.

packages/client/tests/ecosystem/_utils/run.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  console.log('🐳 Starting tests in docker')
+  // tarball was created, ready to send it to docker and begin ecosystem tests
+  await $`docker compose -f ${__dirname}/docker-compose.yml build`.quiet()
+  await $`docker compose -f ${__dirname}/docker-compose.yml down`.quiet()
+  await $`docker compose -f ${__dirname}/docker-compose.yml up`.quiet()
+
+  // let the tests run and gather a list of logs for containers that have failed
+  const findErrors = await $`find "$(pwd)" -not -name .logs.0.txt -name .logs.*.txt`.quiet()
```
millsp, line 64: It will find the log files that are not named `.logs.0.txt`, where the number is the exit code. Basically, it will allow you to jump to log failures.

packages/client/tests/ecosystem/jest.config.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
```
millsp, line 1: Base jest config for all tests.

packages/client/tests/ecosystem/_utils/standard.dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM node:16
```
millsp, line 1: Base docker image for all tests.

packages/client/tests/ecosystem/_utils/standard.cmd.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
millsp, line 1: Script that prepares and starts the test.

helpers/compile/build.ts
Jolg42, line 147:
```suggestion
  // When we trigger pnpm pack we actually don't want to build again to go faster.
  // We re-use what has been already build; and also implies you ran pnpm run watch/dev/build before hand.
  // In the future, I think I'll enable a full build + packaging mode to be much closer to reality, this could be done on CI.
  if (process.env.ECOSYSTEM === 'true') return
```

packages/cli/helpers/copy-prisma-client.ts
Jolg42, line 13:
```suggestion
// this uses a glob library to understand patterns in files.
// Using tooling from npm would be even better, but it is in magnitudes slower compared to this approach.
// Ideally, it would be great if we could remove this somehow.
const clientFiles = fg.sync(clientPkg.files, { cwd: clientPath, dot: true, onlyFiles: false })
```

packages/client/tests/ecosystem/_utils/run.ts
```diff
@@ -0,0 +1,73 @@
...
+  await $`cd ${tmpDir} && cp ${cliPkgJsonPath} cli.package.json`.quiet()
+  await $`cd ${tmpDir} && cp ${clientPkgJsonPath} client.package.json`.quiet()
+  await $`cd ${tmpDir} && cp ${clientRuntimeDtsPath} client.runtime.d.ts`.quiet()
+
+  // we provide a function that can revert modified package.json back
+  const restoreOriginal = async () => {
+    await $`cd ${tmpDir} && cp cli.package.json ${cliPkgJsonPath}`.quiet()
+    await $`cd ${tmpDir} && cp client.package.json ${clientPkgJsonPath}`.quiet()
+    await $`cd ${tmpDir} && cp client.runtime.d.ts ${clientRuntimeDtsPath}`.quiet()
+  }
```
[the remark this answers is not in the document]
Jolg42: like
```
$.cwd = somepath
// now cwd is somepath for next commands
```

Jolg42 (approved):
This looks very interesting! I'm curious to try it later 👀

Jolg42:
Should we merge this one day @millsp ?

PR #16748, opened 2022-12-12
feat(client): DataProxy query logs

danstarns:
Tracked in https://github.com/prisma/client-planning/issues/158
Sibling to https://github.com/prisma/prisma-engines/pull/3618

### DIFF
--- a/packages/client/tests/functional/logging/tests.ts
+++ b/packages/client/tests/functional/logging/tests.ts
@@ -7,176 +7,229 @@ import type { Prisma, PrismaClient } from './node_modules/@prisma/client'
 
 declare let newPrismaClient: NewPrismaClient<typeof PrismaClient>
 
-testMatrix.setupTestSuite(
-  (suiteConfig, _suiteMeta, clientMeta) => {
-    let client: PrismaClient<Prisma.PrismaClientOptions, 'query'>
+testMatrix.setupTestSuite((suiteConfig, _suiteMeta, clientMeta) => {
+  let client: PrismaClient<Prisma.PrismaClientOptions, 'query'>
 
-    test('should log queries on a method call', async () => {
-      client = newPrismaClient({
-        log: [
-          {
-            emit: 'event',
-            level: 'query',
-          },
-        ],
-      })
-
-      const queryLogPromise = new Promise<Prisma.QueryEvent>((resolve) => {
-        client.$on('query', (data) => {
-          if ('query' in data) {
-            resolve(data)
-          }
-        })
-      })
-
-      await client.user.findMany()
-
-      const queryLogEvents = await queryLogPromise
-      expect(queryLogEvents).toHaveProperty('query')
-      expect(queryLogEvents).toHaveProperty('duration')
-      expect(queryLogEvents).toHaveProperty('timestamp')
-
-      if (suiteConfig.provider === 'mongodb') {
-        expect(queryLogEvents.query).toContain('db.User.aggregate')
-      } else {
-        expect(queryLogEvents.query).toContain('SELECT')
-      }
-
-      if (!clientMeta.dataProxy) {
-        expect(queryLogEvents).toHaveProperty('params')
-        expect(queryLogEvents).toHaveProperty('target')
-      }
+  test('should log queries on a method call', async () => {
+    client = newPrismaClient({
+      log: [
+        {
+          emit: 'event',
+          level: 'query',
+        },
+      ],
     })
 
-    test('should log queries inside a ITX', async () => {
-      client = newPrismaClient({
-        log: [
-          {
-            emit: 'event',
-            level: 'query',
-          },
-        ],
+    const queryLogPromise = new Promise<Prisma.QueryEvent>((resolve) => {
+      client.$on('query', (data) => {
+        if ('query' in data) {
+          resolve(data)
+        }
       })
+    })
 
-      const queryLogs = new Promise<Prisma.QueryEvent[]>((resolve) => {
-        const logs: Prisma.QueryEvent[] = []
+    await client.user.findMany()
 
-        client.$on('query', (data) => {
-          if ('query' in data) {
-            logs.push(data)
+    const queryLogEvents = await queryLogPromise
+    expect(queryLogEvents).toHaveProperty('query')
+    expect(queryLogEvents).toHaveProperty('duration')
+    expect(queryLogEvents).toHaveProperty('timestamp')
 
-            if (suiteConfig.provider === 'mongodb' && logs.length === 3) {
-              resolve(logs)
-            }
+    if (suiteConfig.provider === 'mongodb') {
+      expect(queryLogEvents.query).toContain('db.User.aggregate')
+    } else {
+      expect(queryLogEvents.query).toContain('SELECT')
+    }
 
-            if (logs.length === 5) {
-              resolve(logs)
-            }
+    if (!clientMeta.dataProxy) {
+      expect(queryLogEvents).toHaveProperty('params')
+      expect(queryLogEvents).toHaveProperty('target')
+    }
+  })
+
+  test('should log queries inside a ITX', async () => {
+    client = newPrismaClient({
+      log: [
+        {
+          emit: 'event',
+          level: 'query',
+        },
+      ],
+    })
+
+    const queryLogs = new Promise<Prisma.QueryEvent[]>((resolve) => {
+      const logs: Prisma.QueryEvent[] = []
+
+      client.$on('query', (data) => {
+        if ('query' in data) {
+          logs.push(data)
+
+          if (suiteConfig.provider === 'mongodb' && logs.length === 3) {
+            resolve(logs)
           }
-        })
+
+          if ((data.query as string).includes('COMMIT')) {
+            resolve(logs)
+          }
+        }
+      })
+    })
+
+    await client.$transaction(async (tx) => {
+      const id = suiteConfig.provider === 'mongodb' ? faker.database.mongodbObjectId() : faker.random.numeric()
+
+      await tx.user.create({
+        data: {
+          id,
+        },
       })
 
-      await client.$transaction(async (tx) => {
-        const id = suiteConfig.provider === 'mongodb' ? faker.database.mongodbObjectId() : faker.random.numeric()
+      return tx.user.findMany({
+        where: {
+          id,
+        },
+      })
+    })
 
-        await tx.user.create({
-          data: {
-            id,
-          },
-        })
+    const logs = await queryLogs
 
-        return tx.user.findMany({
+    if (suiteConfig.provider === 'mongodb') {
+      expect(logs).toHaveLength(3)
+
+      expect(logs[0].query).toContain('User.insertOne')
+      expect(logs[1].query).toContain('User.aggregate')
+      expect(logs[2].query).toContain('User.aggregate')
+    } else {
+      expect(logs).toHaveLength(5)
+
+      expect(logs[0].query).toContain('BEGIN')
+      expect(logs[1].query).toContain('INSERT')
+      expect(logs[2].query).toContain('SELECT')
+      expect(logs[3].query).toContain('SELECT')
+      expect(logs[4].query).toContain('COMMIT')
+    }
+  })
+
+  test('should log batched queries inside a ITX', async () => {
+    client = newPrismaClient({
+      log: [
+        {
+          emit: 'event',
+          level: 'query',
+        },
+      ],
+    })
+
+    const queryLogs = new Promise<Prisma.QueryEvent[]>((resolve) => {
+      const logs: Prisma.QueryEvent[] = []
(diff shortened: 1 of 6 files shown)


packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -72,7 +77,7 @@ export class DataProxyEngine extends Engine {
 
     const [host, apiKey] = this.extractHostAndApiKey()
     this.remoteClientVersion = P.then(() => getClientVersion(this.config))
-    this.headers = { Authorization: `Bearer ${apiKey}` }
+    this.headers = { Authorization: `Bearer ${apiKey}`, 'prisma-capture-logs': true }
```
aqrln, line 80:
```suggestion
    this.headers = { Authorization: `Bearer ${apiKey}`, 'prisma-capture-logs': 'true' }
```

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -102,7 +102,10 @@ async function nodeFetch(url: string, options: RequestOptions = {}): Promise<Req
     // we execute the https request and build a fetch response out of it
     const request = https.request(url, httpsOptions, (response) => {
       // eslint-disable-next-line prettier/prettier
```
aqrln, line 104:
```suggestion
      // eslint-disable-next-line prettier/prettier
```

since you'd already reformatted it

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -56,7 +61,7 @@ export class DataProxyEngine extends Engine {
 
   private clientVersion: string
   readonly remoteClientVersion: Promise<string>
-  readonly headers: { Authorization: string }
+  readonly headers: { Authorization: string; 'prisma-capture-logs': boolean }
```
aqrln, line 64:
```suggestion
  readonly headers: { Authorization: string; 'prisma-capture-logs': string }
```

headers are always strings, they can't be booleans (although maybe it would get implicitly converted it, but I'd still prefer to use correct types)

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -210,6 +217,13 @@ export class DataProxyEngine extends Engine {
 
         const data = await response.json()
 
+        console.log('DATA PROXY RESPONSE', data)
+
+        const logs = data.logs as LogResponse[]
+        logs.forEach((log) => {
+          if (log.Kind === 'Query') this.logEmitter.emit('query', log.log)
```
aqrln, line 224: This isn't what the engine currently returns, is it?

packages/engine-core/src/data-proxy/utils/request.ts
```diff
@@ -7,7 +7,7 @@ import { RequestError } from '../errors/NetworkError'
 import { getJSRuntimeName } from './getJSRuntimeName'
 
 // our implementation handles less
-export type RequestOptions = O.Patch<{ headers?: { [k: string]: string }; body?: string }, RequestInit>
+export type RequestOptions = O.Patch<{ headers?: { [k: string]: any }; body?: string }, RequestInit>
```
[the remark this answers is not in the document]
aqrln:
```suggestion
export type RequestOptions = O.Patch<{ headers?: { [k: string]: string }; body?: string }, RequestInit>
```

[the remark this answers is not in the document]
danstarns: Fixed, thanks

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -46,6 +47,10 @@ type DataProxyTxInfoPayload = {
 
 type DataProxyTxInfo = Tx.Info<DataProxyTxInfoPayload>
 
+type LogResponse =
+  | { Kind: 'Query'; log: { timestamp: string; query: string; params: string; duration: number; target: string } }
```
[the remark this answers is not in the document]
danstarns: No reason, I lowercased it now.

danstarns:
Closes: https://github.com/prisma/client-planning/issues/144

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -72,7 +82,7 @@ export class DataProxyEngine extends Engine {
 
     const [host, apiKey] = this.extractHostAndApiKey()
     this.remoteClientVersion = P.then(() => getClientVersion(this.config))
-    this.headers = { Authorization: `Bearer ${apiKey}` }
+    this.headers = { Authorization: `Bearer ${apiKey}`, 'X-capture-traces': 'true' }
```
aqrln, line 85: This should not be sent unconditionally, but only when query logs are enabled in the client

packages/client/tests/functional/logging/tests.ts
```diff
@@ -7,177 +7,168 @@ import type { Prisma, PrismaClient } from './node_modules/@prisma/client'
...
-            emit: 'event',
-            level: 'query',
-          },
-        ],
-      })
+testMatrix.setupTestSuite((suiteConfig, _suiteMeta, clientMeta) => {
+  let client: PrismaClient<Prisma.PrismaClientOptions, 'query'>
+
+  test('should log queries on a method call', async () => {
+    client = newPrismaClient({
```
miguelff, line 14: I observed this test failed against latest changes in [prisma-engines ](https://github.com/prisma/prisma-engines/pull/3505) when it shouldn't.

The response to the /transaction/tx_id/commit endpoint during the test was:

```json
{
   "extensions":{
      "logs":[
         {
            "span_id":"5ad656c7781a5886",
            "name":"COMMIT",
            "level":"query",
            "timestamp":[
               1674129157,
               441746000
            ],
            "attributes":{
               "duration_ms":2.564
            }
         }
      ],
      "traces":[
         
      ]
   }
   ```

**Comment by miguelff** [comment]
I added a test for non-interactive transaction logging. I think we should merge this.

File: packages/engine-core/src/data-proxy/DataProxyEngine.ts [hunk 9]
```diff
@@ -51,6 +52,94 @@ type RequestInternalOptions = {
...
+    apiKey,
+    tracingConfig,
+    logLevel,
+    logQueries,
+  }: {
+    apiKey: string
+    tracingConfig: TracingConfig
+    logLevel: EngineConfig['logLevel']
+    logQueries: boolean | undefined
+    engine: DataProxyEngine
```
aqrln, line 95: doesn't seem to be used here

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -76,13 +165,25 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
-    this.remoteClientVersion = P.then(() => getClientVersion(this.config))
-    this.headers = { Authorization: `Bearer ${apiKey}` }
     this.host = host
 
+    this.headerBuilder = new DataProxyHeaderBuilder({
+      apiKey,
+      tracingConfig: getTracingConfig(this.config.previewFeatures || []),
+      logLevel: config.logLevel,
+      logQueries: config.logQueries,
+      engine: this,
```
aqrln, line 175: only required by the type but doesn't seem to be used

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -91,6 +192,48 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
+
+    if (extensions?.logs?.length) {
+      extensions.logs.forEach((log) => {
+        switch (log.level) {
+          case 'debug':
+          case 'error':
+          case 'trace':
+          case 'warn':
+          case 'info':
+            // TODO these are propgated into the response.errors key
```
aqrln, line 206: nit: typo
```suggestion
            // TODO these are propagated into the response.errors key
```

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -91,6 +192,48 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
+            break
+          case 'query': {
+            let dbQuery =
+              typeof log.attributes?.query === 'string' && log.attributes?.query ? log.attributes.query : log.name
```
aqrln, line 210:
```suggestion
              typeof log.attributes?.query === 'string' ? log.attributes.query : log.name
```

I think the second condition is already implied by the first one

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -91,6 +192,48 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
+              const [query] = dbQuery.split('/* traceparent')
+              dbQuery = query
+            }
+
+            this.logEmitter.emit('query', {
+              query: dbQuery,
+              timestamp: log.timestamp,
+              duration: log.attributes.duration_ms,
+              // params: log.params - Missing
+              // target: log.target - Missing
```
aqrln, line 225: They are both `string` in the user-facing `QueryEvent` type, should we at least pass empty strings here instead?

packages/engine-core/src/data-proxy/DataProxyEngine.ts
miguelff: Yes, it is, it means attributes is not undefined. Addressing it.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
miguelff: I think they are present now, gonna address.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -91,6 +190,48 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
+        switch (log.level) {
+          case 'debug':
+          case 'error':
+          case 'trace':
+          case 'warn':
+          case 'info':
+            // TODO these are propagated into the response.errors key
+            break
+          case 'query': {
+            let dbQuery = typeof log.attributes.query === 'string' && log.attributes.query ? log.attributes.query : ''
```
aqrln, line 207: I believe we should now be able to drop the second clause in the `&&` this time (last time I probably wasn't really correct since I missed it could've been an empty string and there was a different value in the else branch):
```suggestion
            let dbQuery = typeof log.attributes.query === 'string ? log.attributes.query : ''
```

`log.attributes.query` is truthy if it's not null, undefined or an empty string
However, it is already checked that it is not null or undefined by checking `typeof log.attributes.query === 'string'`.

Thus the whole expression is equivalent to `typeof log.attributes.query === 'string' && log.attributes.query !== '' ? log.attributes.query : ''`, which is equivalent to `typeof log.attributes.query === 'string' ? log.attributes.query : ''`

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -91,6 +190,48 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
+              // Example query: 'SELECT /* traceparent=00-123-0-01 */'
+              const [query] = dbQuery.split('/* traceparent')
+              dbQuery = query
+            }
+
+            this.logEmitter.emit('query', {
+              query: dbQuery,
+              timestamp: log.timestamp,
+              duration: log.attributes.duration_ms,
+              params: log.attributes.params,
```
aqrln, line 222: Here `log.attributes.params` is an array, but `params` is a `string` in the user-facing `QueryEvent` type. What do library and binary engines emit as `params` in their log events? Do we need to change this to
```suggestion
              params: JSON.stringify(log.attributes.params),
```
here or is the type wrong for other engine types too?

miguelff: I think is fine either way, to me, knowing that params is an array, it could be treated as such. However, the information is there fore being rendered to the user. I will default to consistency and make it a string.

aqrln: > I think is fine either way, to me, knowing that params is an array, it could be treated as such.

Yeah, I mean, an array is definitely better, but we also can't break the existing API, so we'd have to change it to an array in Prisma 5 for all `Engine` class implementations if the other engines adhere to the contract there. It's a different question though if the type is plainly wrong and the library and binary engines return an array there too (we already had a precedent of having the types completely wrong for some fields for the library engine there in the past — the internal API boundary here is not type safe at all and we also probably lack tests that would check the actual types).

Making this a string is probably the safest option for now.

aqrln (approved):

PR #16766, opened 2022-12-13
feat(migrate-engine): start moving out of introspection-engine

jkomyno:
Closes https://github.com/prisma/schema-team/issues/375

**TODO**s after `prisma@4.9.0`:
- [x] https://github.com/prisma/prisma-private/issues/203


packages/internals/src/sendPanic.ts
```diff
@@ -31,23 +43,11 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
...
-        sqlDump = await engine.getDatabaseDescription(schemaOrUrl)
-        dbVersion = await engine.getDatabaseVersion(schemaOrUrl)
-        engine.stop()
-      } catch (e) {
-        debug(e)
-        if (engine && engine.isRunning) {
-          engine.stop()
-        }
-      }
+    if (error.area === ErrorArea.LIFT_CLI && schemaOrUrl) {
```
jkomyno, line 49: The `MigrateEngine` may throw Rust panics with the `LIFT_CLI` error area, rather than the previous `INTROSPECTION_CLI` error area thrown by the `IntrospectionEngine`

packages/internals/src/sendPanic.ts
```diff
@@ -20,7 +19,20 @@ const debug = Debug('prisma:sendPanic')
...
+
+  // retrieve the database version for the given schema or url, without throwing any error
+  getDatabaseVersionSafe: (schemaOrUrl: string) => Promise<string | undefined>
+}
+export async function sendPanic({
+  error,
+  cliVersion,
+  enginesVersion,
+  getDatabaseVersionSafe,
+}: SendPanic): Promise<number> {
```
jkomyno, line 35: Turning a list of arguments to an object with named properties for ease of use. It'd be easy passing the arguments in the wrong sequence instead.

jkomyno:
Integration test errors should be fixed by https://github.com/prisma/prisma-engines/pull/3495.

packages/cli/src/Doctor.ts
```diff
@@ -87,13 +86,14 @@ ${chalk.bold('Examples')}
       throw new Error(`${canConnect.code}: ${canConnect.message}`)
     }
 
-    const engine = new IntrospectionEngine({
-      cwd: path.dirname(schemaPath),
+    const engine = new MigrateEngine({
+      projectDir: path.dirname(schemaPath),
```
janpio, line 90: For later: Would `cwd` maybe be a better name for `projectDir` as well?

packages/internals/src/sendPanic.ts
```diff
@@ -20,7 +19,20 @@ const debug = Debug('prisma:sendPanic')
 // cleanup the temporary files even when an uncaught exception occurs
 tmp.setGracefulCleanup()
 
-export async function sendPanic(error: RustPanic, cliVersion: string, engineVersion: string): Promise<number> {
+type SendPanic = {
+  error: RustPanic
+  cliVersion: string
+  enginesVersion: string
```
janpio, line 25: I think the old name of `engineVersion` (singular, not plural) was chosen because it is the version of _one_ engine. In theory all engines in the CLI can use different versions, and this highlights that this is specifically the version from that one engine file.

packages/internals/src/sendPanic.ts
```diff
@@ -31,23 +43,11 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
 
     const maskedSchema: string | undefined = schema ? maskSchema(schema) : undefined
 
-    let sqlDump: string | undefined
```
janpio, line 34: Not used before anyway so dead?

packages/internals/src/sendPanic.ts
janpio: No chance of this now triggered for ME errors as well? (probably undertested area)

packages/internals/src/sendPanic.ts
janpio: Ah, I see below.

packages/internals/src/sendPanic.ts
janpio: Ah, I see this is only to get the database version in a safe way - not more.

packages/internals/src/sendPanic.ts
```diff
@@ -31,23 +43,11 @@ export async function sendPanic(error: RustPanic, cliVersion: string, engineVers
...
     let dbVersion: string | undefined
     // For a SQLite datasource like `url = "file:dev.db"` only schema will be defined
     const schemaOrUrl = schema || error.introspectionUrl
-    if (error.area === ErrorArea.INTROSPECTION_CLI && schemaOrUrl) {
-      let engine: undefined | IntrospectionEngine
-      try {
-        engine = new IntrospectionEngine()
-        sqlDump = await engine.getDatabaseDescription(schemaOrUrl)
```
janpio, line 42: Are we sure that was not used for something and e.g. submitted with the error report?

packages/migrate/src/MigrateEngine.ts
```diff
@@ -135,11 +142,29 @@ export class MigrateEngine {
...
+    return this.runCommand(this.getRPCPayload('getDatabaseDescription', { schema }))
+  }
+
   /**
    * Get the database version for error reporting.
    */
-  public getDatabaseVersion(): Promise<string> {
-    return this.runCommand(this.getRPCPayload('getDatabaseVersion', undefined))
+  public getDatabaseVersion({ schema }: EngineArgs.GetDatabaseVersionParams): Promise<string> {
+    return this.runCommand(this.getRPCPayload('getDatabaseVersion', { schema }))
```
janpio, line 153: Why does this now also take/need the schema?

packages/migrate/src/MigrateEngine.ts
```diff
@@ -135,11 +142,29 @@ export class MigrateEngine {
...
+  }
+
+  /**
+   * Given a Prisma schema, introspect the database definitions and update the schema with the results.
+   * `compositeTypeDepth` is optional, and only required for MongoDB.
+   * TODO: change `compositeTypeDepth` to -1 once https://github.com/prisma/prisma-engines/pull/3487 is merged.
```
janpio, line 159: PR https://github.com/prisma/prisma-engines/pull/3487 got dismissed

packages/migrate/src/MigrateEngine.ts
```diff
@@ -296,6 +324,8 @@ export class MigrateEngine {
           },
         })
 
+        this.isRunning = true
```
janpio, line 327: Now way that the code after this still makes it fail in some way that `isRunning=true` would be inaccurate?

packages/migrate/src/utils/getDatabaseVersionSafe.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { MigrateEngine } from '../MigrateEngine'
+
+const debug = Debug('prisma:cli')
+
+/**
+ * Retrieve the database version from the given schema or url.
+ * This function never throws, and was introduced to prevent circular dependencies in `@prisma/internals`.
+ */
+export async function getDatabaseVersionSafe(schemaOrUrl: string): Promise<string | undefined> {
```
janpio, line 11: When reading the code, this name threw me off, but I guess it makes sense to describe what it is doing.

packages/migrate/src/utils/getDatabaseVersionSafe.ts
```diff
@@ -0,0 +1,28 @@
...
+  let engine: MigrateEngine | undefined
+  let dbVersion: string | undefined
+  try {
+    engine = new MigrateEngine({
+      projectDir: process.cwd(),
+    })
+    dbVersion = await engine.getDatabaseVersion({ schema: schemaOrUrl })
```
janpio, line 18: Does this always start up a new Engine to make this query, or would it reuse an existing one if it already exists?

packages/migrate/src/MigrateEngine.ts
jkomyno: I had done that and reverted it in https://github.com/prisma/prisma/pull/16766#:~:text=%E2%80%A6-,392a6d2,-jkomyno%20added%20this, as apparently that PR alone doesn't fix this type mismatch problem in JSON-RPC between JS and Rust

packages/migrate/src/MigrateEngine.ts
jkomyno: This `isRunning` properties is implemented mirroring the same property in the IntrospectionEngine, and is updated in the same way. The idea is to first get the MigrateEngine on par with the features exposed by the IntrospectionEngine we need to make use of, and later verify possible improvements that could be done

packages/migrate/src/MigrateEngine.ts
jkomyno: The `getDatabaseVersion` RPC method had always required a schema. We didn't notice that the `getDatabaseVersion` recently added to MigrateEngine was missing an input argument because that method wasn't used in tests

packages/internals/src/sendPanic.ts
jkomyno: I had asked both Joel and Tom before removing `getDatabaseDescription` (which isn't implemented for MigrateEngine and isn't used anywhere else). The rationale is that we haven't made use of it in a long time (Tom never used it, Matthias did once or twice), and if we can remove legacy non-user-facing stuff, that's "always the best option"

packages/cli/src/Doctor.ts
jkomyno: That's a potential choice, yeah

packages/internals/src/sendPanic.ts
janpio: This referred to the resulting `sqlDump` - if we should not be using that to report to the error report. Might very well be that this system is actually set up to accept that and might benefit from doing so.

packages/internals/src/sendPanic.ts
jkomyno: I'm glad we discussed this then! I'm creating a private issue about it

Jolg42:
Note from Alberto: we need Tom to unblock the progress on this PR on engines side.

packages/migrate/src/commands/DbPull.ts
```diff
@@ -252,9 +252,9 @@ Some information will be lost (relations, comments, mapped fields, @ignore...),
       debug(`Introspection Schema Version: ${introspectionResult.version}`)
     } catch (e: any) {
       introspectionSpinner.failure()
-      if (e.code === 'P4001') {
+      if (e.code === 'P1003') {
         if (introspectionSchema.trim() === '') {
-          throw new Error(`\n${chalk.red.bold('P4001 ')}${chalk.red('The introspected database was empty:')} ${
+          throw new Error(`\n${chalk.red.bold('P1003 ')}${chalk.red('The introspected database was empty:')} ${
```
Jolg42, line 257: I think that this shows a breaking change in our public API here before we had the error code `P4001` and now it disappears (changed to P1003)
We need to find another way and keep the old codes or go for a major version bump.

jkomyno: Answered here: https://prisma-company.slack.com/archives/C04HZ2EQS65/p1673446997564369?thread_ts=1673440456.031899&cid=C04HZ2EQS65

Jolg42: TODO:
We now we have 2 cases
- Database is empty
- Database does not exist
We need to check if we have tests for both in introspection

jkomyno: I've added human-friendly error support for both `P4001` and `P1003`. Please check https://github.com/prisma/prisma/pull/16766/commits/9c820ae3348de82991defb068750189426bbfaa5

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -778,7 +814,7 @@ describe('postgresql-multi-schema', () => {
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
     expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
     expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
-  })
+  }, 50_000)
```
Jolg42, line 817: Why is this skipped now?

pnpm-lock.yaml
```diff
@@ -229,7 +229,7 @@ importers:
       '@prisma/debug': workspace:*
       '@prisma/engine-core': workspace:*
       '@prisma/engines': workspace:*
-      '@prisma/engines-version': 4.9.0-32.f307669e852dde0c751a36420b11f932fc3cf268
+      '@prisma/engines-version': 4.9.0-33.bb177fac1c08b8561c02b04d9ad51a31981cd8a3
```
Jolg42, line 232: Diff in the pnpm lock is unexpected, did you install this engine before it landed in main?

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -412,6 +414,8 @@ describe('common/sqlite', () => {
     expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
   })
 
+  // failing, it rejects with P1012 (schema is invalid) rather than succeeding.
+  // The error tells us to try --force, which we do already. Perhaps --force isn't passed properly to Rust?
   it('should succeed when schema is invalid and using --force', async () => {
     ctx.fixture('introspect')
```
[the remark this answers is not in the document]
jkomyno: As of yesterday evening, not anymore. We can get rid of it

pnpm-lock.yaml
jkomyno: Yes, I had tried it out before it landed in main to see whether all tests passed. I can revert https://github.com/prisma/prisma/pull/16766/commits/fcbd3bcf9cf70db9125b12ea70b59739d2fc0243 and merge `main` back to this branch

packages/migrate/src/utils/test-handlePanic.ts
```diff
@@ -11,9 +11,16 @@ async function main() {
   )
 
   const packageJsonVersion = '0.0.0'
-  const engineVersion = '734ab53bd8e2cadf18b8b71cb53bf2d2bed46517'
+  const enginesVersion = '734ab53bd8e2cadf18b8b71cb53bf2d2bed46517'
+  const getDatabaseVersionSafe = () => Promise.resolve(undefined)
```
Jolg42, line 15: Could you use the one from `getDatabaseVersionSafe.ts` ?

jkomyno: I see that the same test in `packages/cli` uses the literal `prismaEngineVersionHash` (https://github.com/prisma/prisma/blob/dba8c89d305d91b7cdd6cac6fea49f5192d39af3/packages/cli/src/utils/test-handlePanic.ts#L8), I may just use the same in `packages/migrate`

packages/migrate/src/types.ts
```diff
@@ -137,6 +137,197 @@ export namespace EngineArgs {
...
+  }
+  interface IntrospectionWarningsCustomPrimaryKeyNamesReintro extends IntrospectionWarning {
+    code: 18
+    affected: AffectedModel[]
+  }
+  interface IntrospectionWarningsRelationsReintro extends IntrospectionWarning {
+    code: 19
+    affected: AffectedModel[]
+  }
+
```
Jolg42, line 309: It's missing
```
interface IntrospectionWarningsTopLevelItemNameIsADupe extends IntrospectionWarning {
  code: 20
  affected: AffectedTopLevel[]
}
```

Can you add the recent changes from this commit?
https://github.com/prisma/prisma/commit/c2d32eea77e963778ba0b98cc3557ad0de62a5b7#diff-a3ca29a928d7e7c324513f5239c93d68b07cbd190c678587efc3eb2a93c32d82

Jolg42: Note: It's the only change since this PR was opened

packages/migrate/src/__tests__/DbPull.test.ts
jkomyno: Heck, this was unintentional, I had CMD+F'd the "skip" string in https://github.com/prisma/prisma/pull/16766/files without realizing `DbPull.test.ts` was compressed due to the diff being too large

packages/migrate/src/utils/test-handlePanic.ts
Jolg42: Actually this file is not even used anywhere, it's a small script that someone can run locally (not part of tests)

So anything works I guess, but let's still remove `const getDatabaseVersionSafe = () => Promise.resolve(undefined)` so we don't forget later

pnpm-lock.yaml
Jolg42: It's probably fine in the end but you will get a conflict in this PR once a new engine version lands

packages/migrate/src/__tests__/DbPull.test.ts
jkomyno: Thankfully the test passes anyway

packages/migrate/src/utils/test-handlePanic.ts
jkomyno: Why would we remove `const getDatabaseVersionSafe = () => Promise.resolve(undefined)`? It's purposedly a mocked function

pnpm-lock.yaml
jkomyno: I have already done the steps described in my previous message with no conflicts whatsoever :)

packages/migrate/src/types.ts
jkomyno: Done

packages/cli/schema.prisma
```diff
@@ -0,0 +1,14 @@
+datasource db {
```
Jolg42, line 1: This schema file at the root of the CLI package is unexpected to me.
Could you remove it? (Is it needed for anything?)

packages/migrate/src/__tests__/introspection/introspection.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      version: NonPrisma,
+      warnings: [],
+    }
+  `)
+
+  // TODO: uncomment once https://github.com/prisma/prisma-private/issues/203 is closed.
+  // const dbVersion = await engine.getDatabaseVersion({ schema })
+  // expect(dbVersion.length > 0).toBe(true)
+
+  engine.stop()
```
Jolg42, line 54: Still needed?

packages/integration-tests/src/__tests__/__helpers__/integrationTest.ts
```diff
@@ -310,7 +307,12 @@ async function setupScenario(kind: string, input: Input, scenario: Scenario) {
     ${datasourceBlock}
   `
 
-  const introspectionResult = await engine.introspect(schemaBase)
+  const engine = new MigrateEngine({
+    projectDir: process.cwd(),
+  })
+  const introspectionResult = await engine.introspect({ schema: schemaBase })
+  engine.stop()
```
Jolg42, line 314: Still needed?

packages/internals/src/__tests__/introspection/introspection.test.ts
```diff
@@ -70,6 +64,5 @@ test('introspection basic', async () => {
 
   const json = JSON.parse(description)
   expect(typeof json).toBe('object')
-
   engine.stop()
```
Jolg42, line 67: Still needed?

packages/cli/src/__tests__/artificial-panic.test.ts
```diff
@@ -48,8 +56,7 @@ describe('artificial-panic introspection', () => {
 
       `),
         expect(e).toMatchObject({
-          area: 'INTROSPECTION_CLI',
-          schemaPath: undefined,
+          area: 'LIFT_CLI',
```
janpio, line 59: If now introspection errors also report `LIFT_CLI`, do we still have enough information in the error reporting tool to quickly place where the error comes from?

packages/internals/src/__tests__/introspection/introspection.test.ts
Jolg42: I think it is actually needed here

packages/cli/src/__tests__/artificial-panic.test.ts
jkomyno: Yes, once https://github.com/prisma/prisma-private/issues/203 is closed (`prisma@4.10.0`)

packages/internals/src/IntrospectionEngine.ts
```diff
@@ -269,6 +269,10 @@ export class IntrospectionEngine {
   public getDatabaseVersion(schema: string): Promise<string> {
     return this.runCommand(this.getRPCPayload('getDatabaseVersion', { schema }))
   }
+
+  /**
+   * @deprecated Use `MigrateEngine.introspect()` instead
```
Jolg42, line 274:
```suggestion
   * @deprecated Use `MigrateEngine.introspect()` instead from `@prisma/migrate` package
```

packages/integration-tests/src/__tests__/__helpers__/integrationTest.ts
janpio, line 314: Is this really necessary here? Doesn't `introspect()` take care of that itself?

packages/integration-tests/src/__tests__/integration/mariadb/__database.ts
```diff
@@ -25,6 +25,7 @@ export const database = {
     CREATE DATABASE ${ctx.id};
     USE ${ctx.id};`
     await db.query(sqlUp + sqlScenario)
+    await db.end()
```
janpio, line 28: Why is this needed?

packages/integration-tests/src/__tests__/integration/mysql/__database.ts
```diff
@@ -24,6 +24,7 @@ export const database = {
     CREATE DATABASE ${ctx.id};
     USE ${ctx.id};`
     await db.query(sqlUp + sqlScenario)
+    await db.end()
```
janpio, line 27: Why is this needed?

packages/integration-tests/src/__tests__/integration/postgresql/__database.ts
```diff
@@ -19,6 +19,7 @@ export const database = {
     create schema ${ctx.id};
     set search_path to ${ctx.id};`
     await db.query(sqlUp + sqlScenario)
+    await db.end()
```
janpio, line 22: Why is this needed?

packages/migrate/src/MigrateEngine.ts
```diff
@@ -43,6 +43,13 @@ export class MigrateEngine {
   private lastError: MigrateEngineLogLine['fields'] | null = null
   private initPromise?: Promise<void>
   private enabledPreviewFeatures?: string[]
+
+  // `latestSchema` is set to the latest schema that was used in `introspect()`
+  private latestSchema?: string
```
janpio, line 48: Is this only used for Introspection? Then it might make sense to include that in the name maybe. `latestIntrospectedSchema` for example.

packages/migrate/src/MigrateEngine.ts
```diff
@@ -135,11 +142,42 @@ export class MigrateEngine {
...
 
+  public getDatabaseDescription(schema: string): Promise<string> {
+    return this.runCommand(this.getRPCPayload('getDatabaseDescription', { schema }))
+  }
+
   /**
    * Get the database version for error reporting.
+   * - TODO: this command currently requires the engine to be initialized with a schema path.
+   *   The IntrospectionEngine supported reading a database version from an inline schema or URL instead,
+   *   which if more flexible and fundamental for the error reporting use case. We should add that here too.
```
janpio, line 153: Can we already create an issue for this?

packages/migrate/src/MigrateEngine.ts
```diff
@@ -458,6 +503,8 @@ export class MigrateEngine {
 
 /** The full message with context we return to the user in case of engine panic. */
 function serializePanic(log: string): string {
+  // TODO: do we still want the "Please create an issue ..." part? That's supposedly
+  // already handled in cli/bin.ts when a panic occurs.
```
janpio, line 507: Issue?

packages/internals/src/__tests__/handlePanic.test.ts
```diff
@@ -119,14 +134,20 @@ describe('handlePanic', () => {
       'test-message',
       rustStackTrace,
       'test-request',
-      ErrorArea.INTROSPECTION_CLI, // area
+      ErrorArea.LIFT_CLI, // area
```
Jolg42, line 137: Not sure what we want to do here, it's still about introspection but the code leaves in the migration engine.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -169,9 +169,11 @@ describe('common/sqlite', () => {
     const introspect = new DbPull()
     const result = introspect.parse(['--print', '--url', 'postgresql://root:prisma@/prisma'])
     await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
-            Error parsing connection string: empty host in database URL
```
janpio, line 172: Nice improvement that this is a proper error now!

packages/internals/src/__tests__/handlePanic.test.ts
jkomyno: The `RustPanic` in `MigrateEngine` are statically defined as `LIFT_CLI`. Although it's technically introspection-related, this tells us the engine that was used under the hood. Besides, this `ErrorArea` label is of internal-use only

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -169,9 +169,11 @@ describe('common/sqlite', () => {
...
+      P1013
 
-          `)
+      The provided database string is invalid. empty host in database URL. Please refer to the documentation in https://www.prisma.io/docs/reference/database-reference/connection-urls for constructing a correct connection string. In some cases, certain characters must be escaped. Please check the string for any illegal characters.
```
janpio, line 174: This message is not optimal, but I guess unrelated to this PR and also not super important.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -427,19 +429,51 @@ describe('common/sqlite', () => {
...
 
-                                                            ✔ Introspected 3 models and wrote them into prisma/invalid.prisma in XXXms
-                                                                  
-                                                            Run prisma generate to generate Prisma Client.
+                                                                        ✔ Introspected 3 models and wrote them into prisma/invalid.prisma in XXXms
+                                                                              
+                                                                        Run prisma generate to generate Prisma Client.
 
-                                        `)
+                                                `)
```
janpio, line 438: OT: One day I hope we will figure out what adds these useless spaces all the time and can fix it.

packages/migrate/src/__tests__/DbPull.test.ts
jkomyno: Yep, this is unrelated to this PR

packages/migrate/src/__tests__/DbPull.test.ts
jkomyno: I agree. Using jest with the `-u` flag causes this, from time to time. Not sure why exactly it happens, though

packages/migrate/src/__tests__/introspection/introspection.test.ts
janpio: Is this still needed?

packages/integration-tests/src/__tests__/__helpers__/integrationTest.ts
jkomyno: I think @Jolg42 ninjed you, I have fixed this already while you were writing this :)

packages/integration-tests/src/__tests__/integration/mariadb/__database.ts
jkomyno: I'll be honest, in retrospect it doesn't seem useful

packages/integration-tests/src/__tests__/__helpers__/integrationTest.ts
janpio: I can live with that.

packages/cli/src/__tests__/artificial-panic.test.ts
janpio: How does that influence it?

janpio (approved):

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -427,19 +429,51 @@ describe('common/sqlite', () => {
...
+
+  // replace database name, e.g., 'tests-migrate', with 'unknown-database'
+  const connectionString = defaultConnectionString.split('/').slice(0, -1).join('/') + '/unknown-database'
+
+  test('basic introspection --url', async () => {
+    const introspect = new DbPull()
+    const result = introspect.parse(['--print', '--url', connectionString])
+    await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
+
+      P1003 The introspected database does not exist: postgres://prisma:prisma@localhost:5432/unknown-database
```
Jolg42, line 457: Unrelated to this PR I guess: I'm not a fan of printing the user and password to the logs 😨 
I'll create an issue for this

Jolg42: It's only for --url, still undocumented so it's not a blocker

jkomyno: I think we already handle secretation elsewhere. Perhaps this should be a function defined just once in the Rust engines and exported via Wasm. Can you please open an issue for this?

packages/integration-tests/src/__tests__/integration/mariadb/__database.ts
jkomyno: `git blame` tells me that @Jolg42 added it

packages/migrate/src/MigrateEngine.ts
jkomyno: Done

packages/integration-tests/src/__tests__/integration/mariadb/__database.ts
Jolg42: Yes it's a nice improvement we already did that for SQL Server only I think, found out when we had the exhaustion of connections on PostrgeSQL

packages/migrate/src/__tests__/DbPush.test.ts
```diff
@@ -190,7 +190,7 @@ describe('push', () => {
       Push cancelled.
     `)
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
-    expect(mockExit).toBeCalledWith(130)
+    expect(mockExit).toHaveBeenCalledWith(130)
```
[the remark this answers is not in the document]
jkomyno: They are behaviorally the same, but the new one was suggested by the Jest extension I'm using in VSCode, which leads me to believe it's the most common "expect" method for that task. It's also compatible with the standard `jasmine` nomenclature, adopted by all other test frameworks. I didn't really notice this change as it was an automated one

packages/migrate/src/__tests__/MigrateDiff.test.ts
```diff
@@ -30,35 +30,35 @@ describe('migrate diff', () => {
       ctx.fixture('empty')
 
       const result = MigrateDiff.new().parse([])
-      await expect(result).rejects.toThrowError()
+      await expect(result).rejects.toThrow()
```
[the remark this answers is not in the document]
jkomyno: Same as above

packages/migrate/src/MigrateEngine.ts
jkomyno: There's already an issue linked in the comments

Jolg42 (approved):
Nice! Let's do this 🚀

packages/internals/src/__tests__/handlePanic.test.ts
Jolg42: `INTROSPECTION_CLI` would be better to keep for the meaning of the error was actually triggered from the "introspection part of the CLI"
But since it would be complicated and would not provide any benefit, we can go for `LIFT_CLI`
Technically, we could also remove completely the `ErrorArea` one day (but we need to handle that in the error reporting backend)

packages/cli/src/__tests__/artificial-panic.test.ts
jkomyno: Besides the actual database version (which is WIP) and the error area (which doesn't matter), this PR will preserve every panic information as before

packages/migrate/src/MigrateEngine.ts
jkomyno, line 507:
```suggestion
  // TODO: https://github.com/prisma/prisma/issues/17268
```

packages/migrate/src/MigrateEngine.ts
janpio: That seemed to be a separate TODO, so I expected a different issue. 
(The other issue is also not really about that TODO, but a requesite for that TODO)

packages/cli/src/__tests__/artificial-panic.test.ts
janpio: > and the error area (which doesn't matter)

There is literally a filter by error area in the error report tool.

packages/migrate/src/__tests__/DbPull.test.ts
Jolg42: https://github.com/prisma/prisma/issues/17286

janpio:
Let's create a public `prisma/prisma` issue for this that will show up in our release notes so people can learn about this.

Jolg42:
Note: first appeared in `4.9.0-dev.55`

PR #16798, opened 2022-12-14
perf(client): Use dmmf from a running engine

SevInf:
To get dmmf, engine needs to perform validation and internal model
construction. If we are re-using processed schema from running engines,
those steps could be skipped and we can get directly to schema <-> dmmf
conversion, saving some time.

Needs: https://github.com/prisma/prisma-engines/pull/3491
Close prisma/client-planning#157


SevInf:
Small performance improvement on init time for large schemas:
image

aqrln (approved):

millsp (approved):

Issue #16553, opened 2022-11-30
Support OpenSSL 3.0 for Alpine Linux


carlocorradini:
`OpenSSL 3.0` is currently the default `OpenSSL` version in `Alpine Linux 3.17` (see [CHANGELOG](https://alpinelinux.org/posts/Alpine-3.17.0-released.html)).
For `Alpine Linux`, `Prisma` only supports `OpenSSL 1.X` at this time.
As a result, the following error is generated:
```sh
Error: Unable to establish a connection to query-engine-node-api library. It seems there is a problem with your OpenSSL installation!
Details: Unable to require(`OMITTED/node_modules/prisma/libquery_engine-linux-musl.so.node`)
 Error loading shared library libssl.so.1.1: No such file or directory (needed by OMITTED/node_modules/prisma/libquery_engine-linux-musl.so.node)
[Context: getDmmf]
```
## Suggested solution

Add `OpenSSL 3.0` support to `Alpine Linux`

## Workaround
Install `openssl1.1-compat`:
```sh
apk add openssl1.1-compat
```
Note that this is only a temporary fix until the `Prisma` team adds support for OpenSSL 3.0.

rene-demonsters:
Running into the same issue here. Our gitlab-ci started failing because the docker:latest base image got updated to Alpine 3.17.

rene-demonsters:
I probably need to swap out `docker:latest` for `docker:20.10.21-dind-alpine3.16` or `docker:20.10.21-alpine3.16`. Currently trying the former. Will update this message once I've figured it out but this might help people already in the mean time.

Tags from: https://hub.docker.com/_/docker/tags?page=1&name=alpine3.16

carlocorradini:
Workaround, install `openssl1.1-compat`
```sh
apk add openssl1.1-compat
```

claudiowilson:
Running into this as well!

CooperBills:
We're hitting this as well. The official docker node image just bumped all their alpine images to Alpine v3.17 a few hours ago, which bundles OpenSSL 3.0 instead of 1.1

solarisn:
> We're hitting this as well. The official docker node image just bumped all their alpine images to Alpine v3.17 a few hours ago, which bundles OpenSSL 3.0 instead of 1.1

Same issue here. This broke our CI today and took some time to track down. Now that the Alpine version of the official docker node image uses Alpine v3.17 (as of today) I think a lot of people are going to start running into this issue.

[...]

PR #16831, opened 2022-12-15
feat(get-platform): add ssl detection support for Alpine

jkomyno:
On Linux Alpine, we now look at either `/lib/libssl.so.3` or `/lib/libssl.so.1.1` to understand which openssl version is installed in the system.

Related:
https://github.com/prisma/prisma-engines/pull/3519
https://github.com/prisma/prisma/issues/16553


packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -79,6 +80,7 @@ describe('download', () => {
         "introspection-engine-linux-arm64-openssl-1.1.x",
         "introspection-engine-linux-arm64-openssl-3.0.x",
         "introspection-engine-linux-musl",
+        "introspection-engine-linux-musl-openssl-3.0.x.so.node",
```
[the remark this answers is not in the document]
janpio:
```suggestion
```

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -104,6 +107,7 @@ describe('download', () => {
         "migration-engine-linux-arm64-openssl-1.1.x",
         "migration-engine-linux-arm64-openssl-3.0.x",
         "migration-engine-linux-musl",
+        "migration-engine-linux-musl-openssl-3.0.x.so.node",
```
janpio, line 110:
```suggestion
```

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -117,6 +121,7 @@ describe('download', () => {
         "prisma-fmt-linux-arm64-openssl-1.1.x",
         "prisma-fmt-linux-arm64-openssl-3.0.x",
         "prisma-fmt-linux-musl",
+        "prisma-fmt-linux-musl-openssl-3.0.x.so.node",
```
janpio, line 124:
```suggestion
```

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -130,6 +135,7 @@ describe('download', () => {
         "query-engine-linux-arm64-openssl-1.1.x",
         "query-engine-linux-arm64-openssl-3.0.x",
         "query-engine-linux-musl",
+        "query-engine-linux-musl-openssl-3.0.x.so.node",
```
janpio, line 138: I think this should not exist as well
```suggestion
```

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -282,6 +293,10 @@ It took ${timeInMsToDownloadAll}ms to execute download() for all binaryTargets.`
           "name": "libquery_engine-linux-musl.so.node",
           "size": 40477248,
         },
+        {
+          "name": "libquery_engine-linux-musl-openssl-3.0.x.so.nod",
```
janpio, line 297:
```suggestion
          "name": "libquery_engine-linux-musl-openssl-3.0.x.so.node",
```

packages/fetch-engine/src/__tests__/download.test.ts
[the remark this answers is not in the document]
aqrln:
```suggestion
        "introspection-engine-linux-musl-openssl-3.0.x",
```

packages/fetch-engine/src/__tests__/download.test.ts
aqrln:
```suggestion
        "migration-engine-linux-musl-openssl-3.0.x",
```

packages/get-platform/src/__tests__/parseOpenSSLVersion.test.ts
```diff
@@ -1,28 +1,55 @@
...
+    {
+      name: 'libssl 1.1',
+      content: `/lib/libssl.so.1.1`,
+      expect: '1.1.x',
+    },
+    {
+      name: 'libssl 3.0',
+      content: `/lib/libssl.so.3`,
+      expect: '3.0.x',
+    },
```
aqrln, line 46: Maybe also add test cases for the full version (just three numbers and also three numbers + letter). For example rhel has both `/lib64/libssl.so.1.1` (as a symlink) and `/lib64/libssl.so.1.1.1g` (as the real file), would be nice to know that we can parse both.

packages/fetch-engine/src/__tests__/download.test.ts
janpio:
```suggestion
        "prisma-fmt-linux-musl-openssl-3.0.x",
```

packages/get-platform/src/getPlatform.ts
```diff
@@ -187,7 +209,18 @@ export async function getPlatform(): Promise<Platform> {
...
 
   if (platform === 'linux' && distro === 'musl') {
-    return 'linux-musl'
+    const base = 'linux-musl'
+    if (!libssl) {
+      return base
+    }
+
+    if (isLibssl1x(libssl)) {
+      // Alpine 3.16 or inferior linked with OpenSSL 1.1
```
aqrln, line 218:
```suggestion
      // Alpine 3.16 or below linked with OpenSSL 1.1
```

packages/get-platform/src/getPlatform.ts
```diff
@@ -187,7 +209,18 @@ export async function getPlatform(): Promise<Platform> {
...
+      return base
+    } else {
+      // Alpine 3.17 or superior linked with OpenSSL 3.0
```
aqrln, line 221:
```suggestion
      // Alpine 3.17 or above linked with OpenSSL 3.0
```

packages/get-platform/src/getPlatform.ts
```diff
@@ -206,5 +239,36 @@ export async function getPlatform(): Promise<Platform> {
   }
 
   // use the debian build with OpenSSL 1.1 as a last resort
+  // TODO: perhaps we should default to 'debian-openssl-3.0.x'
```
aqrln, line 242: btw, if we still had the real statically linked musl build, like before introduction of Node-API library, as opposed to just calling alpine-specific build "musl", we could use it as the default — it would work on absolutely any linux distribution (at the expense of larger binaries)

aqrln: we should look into bringing back statically linked musl later and see if we could make it work with Node-API, otherwise we can't support distroless images

packages/get-platform/src/getPlatform.ts
```diff
@@ -206,5 +239,36 @@ export async function getPlatform(): Promise<Platform> {
...
+async function discardError<T>(runPromise: () => Promise<T>): Promise<T | undefined> {
+  try {
+    return await runPromise()
+  } catch (e) {
+    return undefined
+  }
+}
+
+/**
+ * Given a list of system commands, returns the first successful command's stdout.
```
aqrln, line 259: I'd clarify that "first" means first in the order in the input array, not the one that completed first.

janpio (commented):
Generally looks good ✅ , see suggestions though so tests can succeed later when engine files are available.

Jolg42:
I merged main since I merged https://github.com/prisma/prisma/pull/16857 just before

packages/get-platform/src/getPlatform.ts
Jolg42, line 259:
```suggestion
 * Given a list of system commands, it runs them in the given order and returns the first successful command's stdout.
```

Jolg42 (approved):
Looks great! 💯
Note: The tests are failing, but they should pass after updating the test snapshots (see suggestions)

Jolg42 (commented):
In download.test.ts the `FIXED_ENGINES_HASH` value needs to be changed for the test to pass
I can suggest using `c9e863f2d8de6fa0c4bcd609df078ea2dde3c2b2`
from https://github.com/prisma/prisma/pull/16857

aqrln (approved):

PR #16833, opened 2022-12-15
feat(get-platform): add ssl detection support for Alpine

jkomyno:
Integration PR for https://github.com/prisma/prisma/pull/16831.


packages/get-platform/src/getPlatform.ts
```diff
@@ -102,46 +106,47 @@ export function parseOpenSSLVersion(input: string): string | undefined {
...
+    .with({ distro: 'musl' }, () => {
+      /* Linux Alpine */
+      return getFirstSuccessfulExec(['ls -l /lib/libssl.so.3', 'ls -l /lib/libssl.so.1.1'])
+    })
+    .otherwise(() => {
+      return getFirstSuccessfulExec(['ls -l /lib64 | grep ssl', 'ls -l /usr/lib64 | grep ssl'])
+    })
+
+  if (libsslVersion) {
+    const match = /libssl\.so\.(\d+\.\d+)\.\d+/.exec(libsslVersion)
```
aqrln, line 130: I think this regex won't match the filename on Alpine

jkomyno: I was just looking into that, running things locally

packages/get-platform/src/__tests__/parseOpenSSLVersion.test.ts
```diff
@@ -1,28 +1,55 @@
...
+      expect: '3.0.x',
+    },
+  ]
+
+  for (const t of tests) {
+    test(t.name, () => {
+      const actual = parseLibSSLVersion(t.content)
+      expect(actual).toBe(t.expect)
+    })
+  }
```
aqrln, line 54: fwiw Jest has a built-in helper for this, I think it here it would be something like this (didn't check if it actually works):

```suggestion
  test.each(tests)('$name', (t) => {
    const actual = parseLibSSLVersion(t.content)
    expect(actual).toBe(t.expect)
  });
```

no strong opinion though, either is fine with me

PR #16844, opened 2022-12-15
feat(client): Integrate extended where for optional 1:1 relationships

SevInf:
Problem with integration of this PR before: we never generated argument
types for nested field. For m:n relationships, we had a hack which just
reused FindManyArgs. For 1:1 we can't use anything similar: neither
`findFirst` nor `findUnique` types pass.
In this PR we would start geterating such args types. Names of these
type is derived from model and field name, i.e. if `User` model has
`account` field with arguments, corresponding type would be named
`UserAccountArgs`. It applies to both 1:1 and 1:m relationships (we
won't be reusing `FindMany` type anymore).

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -2598,7 +2598,7 @@ export namespace Prisma {
   }
 
   /**
-   * Post: findUnique
+   * Post findUnique
    */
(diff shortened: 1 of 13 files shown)


packages/client/package.json
```diff
@@ -149,7 +149,7 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "4.8.0-51.cb66a7dac27b38cbc5dc34b27cf389f4be35d5e2"
+    "@prisma/engines-version": "4.8.0-49.feat-to-one-read-where-filter-bdc11e554ded89befc229418635bf7e058d6ca9a"
```
SevInf, line 152: TODO: revert before merging

millsp (approved):

PR #16871, opened 2022-12-16
feat(get-platform): sanitise OpenSSL 3+

jkomyno:
This PR always returns `n.0.x` for any `n.x.x` version, where `n` is a major version with `n >= 3`.

OpenSSL 3 adopted [a new versioning policy](https://www.openssl.org/policies/general/versioning-policy.html) which is basically semver, and will not break API and ABI until OpenSSL 4, but our engines naming scheme still follows the old OpenSSL versioning scheme, where the second and not the first number indicated breaking changes.

### DIFF
--- a/packages/get-platform/src/__tests__/parseOpenSSLVersion.test.ts
+++ b/packages/get-platform/src/__tests__/parseOpenSSLVersion.test.ts
@@ -17,6 +17,11 @@ describe('parseOpenSSLVersion', () => {
       content: `OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)`,
       expect: '3.0.x',
     },
+    {
+      name: 'openssl 3.1',
+      content: `OpenSSL 3.1.0 sometimes in 2023`,
+      expect: '3.0.x',
+    },
   ]
 
   test.each(tests)('$name', (t) => {
@@ -50,6 +55,11 @@ describe('parseLibSSLVersion', () => {
     {
       name: 'libssl 3.0',
       content: `/lib/libssl.so.3`,
+      expect: '3.0.x',
+    },
+    {
+      name: 'libssl 3.1',
+      content: `/lib/libssl.so.3.1`,
       expect: '3.0.x',
     },
   ]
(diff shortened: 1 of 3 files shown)


packages/get-platform/src/platforms.ts
```diff
@@ -10,7 +10,7 @@ export type Platform =
   | 'rhel-openssl-3.0.x'
   | 'linux-arm64-openssl-1.1.x'
   | 'linux-arm64-openssl-1.0.x'
-  | 'linux-arm64-openssl-3.0.x'
+  | 'linux-arm64-openssl-3.x.x'
```
aqrln, line 13: I don't think we're renaming the engines just yet

jkomyno, line 13:
```suggestion
  | 'linux-arm64-openssl-3.0.x'
```

packages/get-platform/src/platforms.ts
```diff
@@ -36,7 +36,7 @@ export const platforms: Array<Platform> = [
   'rhel-openssl-3.0.x',
   'linux-arm64-openssl-1.1.x',
   'linux-arm64-openssl-1.0.x',
-  'linux-arm64-openssl-3.0.x',
+  'linux-arm64-openssl-3.x.x',
```
jkomyno, line 39:
```suggestion
  'linux-arm64-openssl-3.0.x',
```

packages/get-platform/src/__tests__/parseOpenSSLVersion.test.ts
```diff
@@ -52,6 +57,11 @@ describe('parseLibSSLVersion', () => {
       content: `/lib/libssl.so.3`,
       expect: '3.0.x',
     },
+    {
+      name: 'libssl 3.1',
+      content: `/lib/libssl.so.3.1`,
+      expect: '3.0.x',
```
Jolg42, line 63:
```suggestion
      expect: '3.0.x',
```
Technically `3.x.x` is better maybe but it doesn't change anything in the logic? (except that we would need to change the built engine names too on the CDN) 
So maybe not necessary to do overall?

Jolg42 (approved):
Looks good to me, but I probably miss some details so better if @aqrln can review too

packages/get-platform/src/__tests__/parseOpenSSLVersion.test.ts
jkomyno: We're keeping `3.0.x` for now, which we have already tested, and revisit this decision for [Prisma 5](https://www.notion.so/prismaio/Standardise-compilation-target-names-platforms-on-Linux-a4a4c3e91d254f5ca10fcd976ea5a7da).

Issue #11807, opened 2022-02-14
Output version of generated Prisma Client in `prisma -v`


janpio:
Currently you can output `prisma -v` and think you are all good, but you actually still have an outdated old Prisma Client generated with an older version that will fail, which you will only possibly notice in the Client error message. This would be easier to spot if `prisma -v` would also output the version (and location!) of the currently generated Client - and possibly even highlight an outdated version (or one that does not match your `@prisma/client` in your project).

jkomyno:
Hi there,
As part of our effort to clean up our GitHub repo, we’re closing this issue to help focus on features and improvements that align with our focus areas and those of our community. 

This issue was originally created for internal tracking purposes, but if you feel this is important to you, feel free to reopen the issue and get like-minded users to upvote as we will prioritize fixing issues based on community interest.

For more insight into our priorities and how we’re thinking about the future of Prisma, please see our blog post: https://www.prisma.io/blog/prisma-orm-manifesto.

Thank you for your understanding and support!

PR #16874, opened 2022-12-16
feat(cli): Output version of generated Prisma Client in prisma -v

DavidHancu:
# Scope
This Pull Request represents the base implementation for #11807 and adds the version of all generated Prisma Clients to the `prisma -v` command, which will be expanded in a nicer format later on.
# Example Output
```prisma                  : 0.0.0
@prisma/client          : 0.0.0
Query Engine (Node-API) : libquery-engine c9e863f2d8de6fa0c4bcd609df078ea2dde3c2b2 (at ..\..\packages\engines\query_engine-windows.dll.node)
Migration Engine        : migration-engine-cli c9e863f2d8de6fa0c4bcd609df078ea2dde3c2b2 (at ..\..\packages\engines\migration-engine-windows.exe)
Introspection Engine    : introspection-core c9e863f2d8de6fa0c4bcd609df078ea2dde3c2b2 (at ..\..\packages\engines\introspection-engine-windows.exe)
Format Binary           : prisma-fmt c9e863f2d8de6fa0c4bcd609df078ea2dde3c2b2 (at ..\..\packages\engines\prisma-fmt-windows.exe)
Format Wasm             : @prisma/prisma-fmt-wasm 4.8.0-48.c9e863f2d8de6fa0c4bcd609df078ea2dde3c2b2
Default Engines Hash    : c9e863f2d8de6fa0c4bcd609df078ea2dde3c2b2
Generator client        : 0.0.0 (at ..\node_modules\.prisma\client)
Generator client1       : 0.0.0 (at ..\.prisma\client)
Generator client2       : 0.0.0 (at ..\another\prisma\client)
```


Jolg42:
After taking a second look, we are actually missing tests here in https://github.com/prisma/prisma/blob/main/packages/cli/src/__tests__/commands/Version.test.ts

We should test `prisma-client-js` generator with
- no custom output
- custom output

And in 2 different situations
- the generated directory exist
- the generated directory doesn't exist

Note: we might need tests using different package managers potentially, but at the moment in the codebase we do not have a setup for this. So we could start with some manual tests and see how it goes with npm  / yarn 1 / pnpm at least.

I'm wondering if adding the file modified date would make sense in the output 🤔

DavidHancu:
> We should test `prisma-client-js generator` with
> - no custom output
> - custom output

I've already tested this manually and implemented a `filterDuplicates` function that ensures that all generators added have different output paths. 

> And in 2 different situations
> - the generated directory exist
> - the generated directory doesn't exist

If the directory doesn't exist, the `getGeneratedClientVersion` function will return null which will exclude this generator from the final output.

If necessary, I will write the tests required for this pull request, but I will need some indication on what kind of tests need to be created and which cases should be checked.

janpio:
What about adding a `-v` output to the end of the Ecosystem Tests runs of https://github.com/prisma/ecosystem-tests/tree/dev/packagers? That would be a good visual confirmation (potentially with no assertion, but better than nothing), @Jolg42? Then we "just" need an integration version of this PR and it would run automatically.

DavidHancu:
> What about adding a `-v` output to the end of the Ecosystem Tests runs of https://github.com/prisma/ecosystem-tests/tree/dev/packagers? That would be a good visual confirmation (potentially with no assertion, but better than nothing), @Jolg42? Then we "just" need an integration version of this PR and it would run automatically.

As suggested, I've submitted a Pull Request implementing `prisma -v` across all packagers: https://github.com/prisma/ecosystem-tests/pull/3348.

packages/cli/src/Version.ts
```diff
@@ -143,6 +166,39 @@ export class Version implements Command {
...
+    }[]
+  > {
+    if (!schemaPath) {
+      return []
+    }
+
+    const datamodel = await getSchema()
+    const config = await getConfig({
+      datamodel,
+      ignoreEnvVarErrors: false,
```
janpio, line 182:
```suggestion
      ignoreEnvVarErrors: true,
```

janpio:
The ecosystem-tests for the packagers that @DavidHancu added worked fine, but some unrelated Docker tests now failed. It turned out there was a `getConfig` call that now validated env vars like `DATABASE_URL` used in the schema, and everywhere we call `prisma -v` in ecosystem-tests this seems to be the case - except in the new Docker tests. So bug identified, and fix suggested above.

Not sure this would have been caught via the suggested tests - but now that the complexity of this is clearer, we definitely need tests with some expectations that need to be met - and that is much better implemented here than in ecosystem-tests.

What might not be obvious to @DavidHancu is how to add a "generate a Client" step to these existing tests. Should this call out to the CLI itself, or do it some other way?

packages/cli/src/Version.ts
```diff
@@ -90,17 +93,38 @@ export class Version implements Command {
     })
 
     const prismaClientVersion = await getInstalledPrismaClientVersion()
-
-    const rows = [
+    const schemaPath = await getSchemaPath()
```
jkomyno, line 96: Note, `getSchemaPath`'s return type changed recently.

```suggestion
    const { schemaPath } = await getSchemaPath()
```

packages/cli/src/Version.ts
```diff
@@ -143,6 +166,39 @@ export class Version implements Command {
     return []
   }
 
+  private async getPrismaClientJSGenerators(schemaPath: string | null): Promise<
```
jkomyno, line 169: This function should take the `{ schemas }` output of `getSchemaPath` as input, rather than calling `getSchema()` internally.

wmadden:
Hi @DavidHancu,

Thank you for your contribution and the effort you’ve put into this pull request. As part of our recent initiative to streamline our backlog and focus on active developments, as outlined in our [Prisma ORM Manifesto](https://www.prisma.io/blog/prisma-orm-manifesto), we’re closing inactive pull requests. Given that this PR has been inactive for an extended period, we’re going to close it for now. If you plan to revisit this enhancement in the future, please feel free to reopen the PR or submit a new one. We genuinely appreciate your efforts and look forward to any future contributions you might have.

PR #16883, opened 2022-12-18
feat(client): nullable relation where

LonJonn:
I'm opening this draft PR in hopes of improving nullable relation types.

Currently, querying for where a relation exists doesn't return the correct TS type and the field is still nullable.

e.g.
``` ts
/**
 * Imagine we have a job board app and users can assign themselves to a job.
 * We want to write a query to get all users with an active job.
 *
 * If we are querying where `activeJob` exists, it should not be nullable in the resulting type.
 */
const usersWithJob = await prisma.user.findMany({
  include: { activeJob: true },
  where: {
    activeJob: {},
  },
})

console.log(usersWithJob[0].activeJob.description)
//                          ^ but `activeJob` is still `Job | null`
```

This PR is a very hacky POC and isn't intended to be a proper implementation (I don't know how to write TS that well haha...) but hopefully it helps illustrate my problem and provides some general guidance.

I've included a repro project that demonstrates a basic example. It doesn't handle `AND` or `OR` statements and since I'm not too familiar with the Prisma repo, I'm sure there are other things I've missed as well.

Let me know if you need any more info or if there's anything else I can do to help. Would love to help get this merged in eventually in any way I can!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ esm
 reproductions/*
 !reproductions/basic-sqlite
 !reproductions/tracing
+!reproductions/nullable-relation-where
 !reproductions/pnpm-workspace.yaml
 
 dev.db
--- a/packages/client/src/generation/TSClient/Payload.ts
+++ b/packages/client/src/generation/TSClient/Payload.ts
@@ -30,7 +30,7 @@ export class PayloadType implements Generatable {
       isModel && this.findMany ? ` | ${getModelArgName(name, DMMF.ModelAction.findMany)}${ifExtensions('', '')}` : ''
 
     return `\
-export type ${getPayloadName(name)}<S extends boolean | null | undefined | ${argsName}${ifExtensions(
+export type ${getPayloadName(name)}<S extends boolean | null | undefined | Pick<${findManyArg}, "include" | "select" | "where">${ifExtensions(
       `, ExtArgs extends runtime.Types.Extensions.Args = runtime.Types.Extensions.DefaultArgs, _${name} = runtime.Types.Extensions.GetResultPayload<${name}, ExtArgs['result']['${lowerCase(
         name,
       )}']>`,
@@ -87,7 +87,7 @@ ${indent(
       (f) =>
         `P extends '${f.name}' ? ${this.wrapType(
           f,
-          `${getPayloadName((f.outputType.type as DMMF.OutputType).name)}<S['${projection}'][P]${ifExtensions(
+          `${getPayloadName((f.outputType.type as DMMF.OutputType).name)}<S['${projection}'][P] & { where: Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> }${ifExtensions(
             ', ExtArgs',
             '',
           )}>`,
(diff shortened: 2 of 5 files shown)


LonJonn:
I've updated the PR to use a better type that accounts for nested `AND` and `OR` queries.

The `Extract` is necessary since `AND` and `OR` may resolve element types that don't include P as a property.
e.g. below should still resolve in `child` being non-nullable.
```ts
{
  where: {
    AND: [
      {
        child: {}
      },
      {
        someOtherProperty: true
      },
    ]
  }
}
```

It looks a bit hectic at the moment since the "where" type needs to passed into the nested `_GetPayload` to support nested relation queries e.g. `result.child.nestedChild` should also be non-nullable.
```ts
{
  where: {
    child: {
      nestedChild: {}
  }
}
```
, but I'm not sure how else it could be done.
 
 Here's is an example diff of the generated output:
```diff
  export type UserGetPayload<S extends boolean | null | undefined | Pick< | UserFindManyArgs, "include" | "select" | "where">> =
    S extends { select: any, include: any } ? 'Please either choose `select` or `include`' :
    S extends true ? User :
    S extends undefined ? never :
    S extends { include: any } & (UserArgs | UserFindManyArgs)
    ? User  & {
    [P in TruthyKeys<S['include']>]:
-       P extends 'activeJob' ? JobGetPayload<S['include'][P]> | null :
-       P extends '_count' ? UserCountOutputTypeGetPayload<S['include'][P]> :  never
+       P extends 'activeJob' ? JobGetPayload<S['include'][P] & { where: Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> }> | (Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> extends never ? null : never) :
+       P extends '_count' ? UserCountOutputTypeGetPayload<S['include'][P] & { where: Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> }> :  never
  } 
    : S extends { select: any } & (UserArgs | UserFindManyArgs)
      ? {
    [P in TruthyKeys<S['select']>]:
-       P extends 'activeJob' ? JobGetPayload<S['select'][P]> | null :
-       P extends '_count' ? UserCountOutputTypeGetPayload<S['select'][P]> :  P extends keyof User ? User[P] : never
+       P extends 'activeJob' ? JobGetPayload<S['select'][P] & { where: Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> }> | (Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> extends never ? null : never) :
+       P extends '_count' ? UserCountOutputTypeGetPayload<S['select'][P] & { where: Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> }> :  P extends keyof User ? User[P] : never
  } 
      : User
```

The type can be read as:
```ts
// If `where[P]`, `where.AND[number][P]` and `where.OR[number][P]` all don't satisfy `{}`, then the user isn't querying for `P` so return `null`
type _ = Extract<S['where'][P] | (S['where']['AND'] & S['where']['OR'])[number][P], {}> extends never ? null : never
```

LonJonn:
While this change has been working fine on my local machine as a patch, I still think it can be improved further, but I'm not sure how to do it myself.

Keen to get feedback.

PR #16923, opened 2022-12-20
feat: remove prisma-fmt binary

### LINKED ISSUE #16924: Remove `prisma-fmt` formatter binary
We've been using the Wasm formatter for several releases already. It's time we stop downloading the `prisma-fmt` binary.
We should remove support for `prisma-fmt` for the same release that will get rid of `introspection-engine`.

jkomyno:
DO NOT MERGE UNTIL `prisma@4.10.0` is ready.

This PR stops downloading the `prisma-fmt` binary, and removes every old reference of it in the codebase.

Closes https://github.com/prisma/prisma/issues/16924.


packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
```diff
@@ -20,7 +19,6 @@ Current platform        : TEST_PLATFORM
 Query Engine (Node-API) : libquery-engine ENGINE_VERSION (at sanitized_path/libquery_engine-TEST_PLATFORM.LIBRARY_TYPE.node, resolved by PRISMA_QUERY_ENGINE_LIBRARY)
 Migration Engine        : migration-engine-cli ENGINE_VERSION (at sanitized_path/migration-engine-TEST_PLATFORM, resolved by PRISMA_MIGRATION_ENGINE_BINARY)
 Introspection Engine    : introspection-core ENGINE_VERSION (at sanitized_path/introspection-engine-TEST_PLATFORM, resolved by PRISMA_INTROSPECTION_ENGINE_BINARY)
-Format Binary           : prisma-fmt ENGINE_VERSION (at sanitized_path/prisma-fmt-TEST_PLATFORM, resolved by PRISMA_FMT_BINARY)
```
Jolg42, line 23: Random thought after reading the code here is that it's now impossible to specify an env var for the format binary (because removed)
So we also need to remove that from the docs after
- https://www.prisma.io/docs/concepts/components/prisma-engines#using-custom-engine-libraries-or-binaries
- https://www.prisma.io/docs/reference/api-reference/environment-variables-reference#prisma_fmt_binary

Jolg42 (approved):
Nice cleanup! That will make the CLI a bit smaller 🎊

packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
janpio: That is actually a change in functionality that we probably need to look at in a bigger context - the same will be true if we move `getDmmf` to the wasm module in CLI :/

jkomyno: Resolving this as we have started this discussion on [Slack](https://prisma-company.slack.com/archives/C02A1646TBR/p1671553918011699) already.

janpio: We need to get to an agreed solution to that discussion though before we merge it.

Jolg42:
Note: we want to merge this PR and https://github.com/prisma/prisma/pull/17301 together in 4.10.0

PR #16964, opened 2022-12-21
feat(get-platform): improve ssl detection on Debian

### LINKED ISSUE #16962: OS Support: improve SSL detection on `debian` distros
When downloading the Prisma engines, we're currently resolving the `openssl` version with:
- `ls -l /lib64 | grep ssl`
- `ls -l /usr/lib64 | grep ssl`
- `openssl version -v`

However, Debian, Ubuntu, and similar distros define the `libssl.so.*` library in a specific path:

```console
"/usr/lib/$(uname -m)-linux-gnu/libssl.so.*"
```

We should prefer extracting the `openssl` version from that path.

jkomyno:
Closes https://github.com/prisma/prisma/issues/16962.


jkomyno:
Closing in favor of https://github.com/prisma/prisma/pull/16965

PR #16965, opened 2022-12-21
feat(get-platform): improve ssl detection on Debian

jkomyno:
Integration PR for https://github.com/prisma/prisma/pull/16964.


packages/get-platform/src/getPlatform.ts
```diff
@@ -12,6 +12,7 @@ const exec = promisify(cp.exec)
 
 // https://www.geeksforgeeks.org/node-js-process-arch-property/
 export type Arch = 'x32' | 'x64' | 'arm' | 'arm64' | 's390' | 's390x' | 'mipsel' | 'ia32' | 'mips' | 'ppc' | 'ppc64'
+export type UnameArch = 'i386' | 'x86_64' | 'arm64' | 'arm7l'
```
aqrln, line 15:
```suggestion
export type UnameArch = 'i386' | 'x86_64' | 'arm64' | 'arm7l' | 'aarch64'
```

`arm64` is on macOS (which we might not even care about here), on Linux `uname -m` will report `aarch64`

aqrln: There's also quite a bunch of other variations only of these architectures that this type doesn't cover, not even counting completely different architectures covered by the `Arch` type. Do we need to enumerate them (i.e. are you planning to match on this type anywhere), or could we just use `string` instead?

packages/get-platform/src/getPlatform.ts
```diff
@@ -289,6 +309,15 @@ function getFirstSuccessfulExec(commands: string[]) {
...
 }
 
+/**
+ * Returns the architecture of a system from the output of `uname -m` (whose format is different than `process.arch`).
+ * This function never throws.
+ */
+async function getArchFromUname(): Promise<UnameArch | undefined> {
+  const arch = await getFirstSuccessfulExec(['uname -m'])
+  return arch?.trim() as UnameArch
+}
```
jkomyno, line 319:
```suggestion
async function getArchFromUname(): Promise<string | undefined> {
  const arch = await getFirstSuccessfulExec(['uname -m'])
  return arch?.trim()
}
```

packages/get-platform/src/getPlatform.ts
jkomyno, line 15:
```suggestion
```

aqrln (approved):

jkomyno:
Closes https://github.com/prisma/prisma/issues/16962.

Jolg42:
Note: ecosystem-tests on main are failing since this PR was merged and released in 4.9.0-dev.9 

https://github.com/prisma/ecosystem-tests/actions/runs/3760098323/jobs/6390461782#step:7:192
with
```
Error: Pattern matching error: no pattern matches value {"arch":"x64","distro":"rhel"}
```

PR #17044, opened 2022-12-28
fix: Allow result extensions to shadow fields

### LINKED ISSUE #16643: `result` extension that shadows a field of the same name causes `Maximum call stack exceeded` at runtime
### Bug description

When a `result` extension adds a computed field that shadows a field of the same name, AND the extension field `needs` the shadowed field, the following error occurs:

```
RangeError: Maximum call stack size exceeded
    at Array.flatMap (<anonymous>)
    at /Users/stephen/Projects/prisma-client-extensions-shadowing/node_modules/@prisma/client/runtime/index.js:31121:48
    at Cache.getOrCreate (/Users/stephen/Projects/prisma-client-extensions-shadowing/node_modules/@prisma/client/runtime/index.js:31090:19)
    at resolveNeeds (/Users/stephen/Projects/prisma-client-extensions-shadowing/node_modules/@prisma/client/runtime/index.js:31119:18)
    at Array.flatMap (<anonymous>)
    at /Users/stephen/Projects/prisma-client-extensions-shadowing/node_modules/@prisma/client/runtime/index.js:31121:48
    at Cache.getOrCreate (/Users/stephen/Projects/prisma-client-extensions-shadowing/node_modules/@prisma/client/runtime/index.js:31090:19)
    at resolveNeeds (/Users/stephen/Projects/prisma-client-extensions-shadowing/node_modules/@prisma/client/runtime/index.js:31119:18)
    at Array.flatMap (<anonymous>)
    at /Users/stephen/Projects/prisma-client-extensions-shadowing/node_modules/@prisma/client/runtime/index.js:31121:48 {
  clientVersion: '4.7.1'
}
```

Here is where the issue actually happens:

https://github.com/prisma/prisma/blob/main/packages/client/src/runtime/core/extensions/resultUtils.ts#L58

(`resolveNeeds` doesn't check for cycles in the dependencies)

### How to reproduce

Repository with a minimal reproduction and instructions in the README: https://github.com/sbking/prisma-client-extensions-shadowing

### Expected behavior

A computed field should be able to shadow a field of the same name, and use the original value to compute the computed value.

This works at the type level - the extended client uses the data type returned by the computed field, not the type of the original field.

The TypeScript types also seem suggest that you can chain extensions that shadow the same computed field. This would be intuitive, but the docs suggest that the last extension always wins in the case of conflicts. See the inferred inlay hints in this screenshot:

Screen Shot 2022-12-05 at 5 43 04 PM

### Prisma information

```prisma
generator client {
}

datasource db {
  provider = "postgresql"
}

model Post {
  id        String   @id @default(cuid())
  title     String
}
```

Here is the extension code:

```ts
  name: "i18n",
  result: {
    post: {
      createdAt: {
        needs: { createdAt: true },
        compute(post) {
          return formatDistanceToNow(post.createdAt, { locale: de });
        },
      },
    },
  },
});

async function main() {
  const posts = await prisma.post.findMany({
    select: {
      title: true,
      createdAt: true,
    },
    take: 5,
  });

  for (const post of posts) {
    console.info(`- ${post.title} (${post.createdAt})`);
  }
}
```

And here is where the error occurs:

```ts
async function main() {
  const posts = await prisma.post.findMany({
    select: {
      title: true,
      createdAt: true,
    },
    take: 5,
  });

  for (const post of posts) {
    console.info(`- ${post.title} (${post.createdAt})`);
  }
}
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: v18.10.0

### Prisma Version

```
4.7.1
```

SevInf:
Was broken because of the 2 things:

1. Dependency resolution did not account for circular dependencies
2. When building a query, all computed fields in selection were ignored,
   even if there is an actual DB field of the same name.

Fix #16643

### DIFF
--- a/packages/client/tests/functional/extensions/enabled/result.ts
+++ b/packages/client/tests/functional/extensions/enabled/result.ts
@@ -224,6 +224,37 @@ testMatrix.setupTestSuite(() => {
     expect(user?.firstName).toBe('JOHN')
   })
 
+  test('shadowing dependency multiple times', async () => {
+    const xprisma = prisma
+      .$extends({
+        result: {
+          user: {
+            firstName: {
+              needs: { firstName: true },
+              compute(user) {
+                return user.firstName.toUpperCase()
+              },
+            },
+          },
+        },
+      })
+      .$extends({
+        result: {
+          user: {
+            firstName: {
+              needs: { firstName: true },
+              compute(user) {
+                return `${user.firstName}!!!`
+              },
+            },
+          },
+        },
+      })
+
+    const user = await xprisma.user.findFirst()
+    expect(user?.firstName).toBe('JOHN!!!')
+  })
+
   test('empty extension does nothing', async () => {
     const xprisma = prismaWithExtension()
       .$extends({
--- a/packages/client/tests/functional/extensions/enabled/result.ts
+++ b/packages/client/tests/functional/extensions/enabled/result.ts
@@ -206,6 +206,24 @@ testMatrix.setupTestSuite(() => {
     expect(user?.loudName).toBe('JOHN SMITH')
   })
 
+  test('shadowing dependency', async () => {
+    const xprisma = prisma.$extends({
+      result: {
+        user: {
+          firstName: {
+            needs: { firstName: true },
+            compute(user) {
+              return user.firstName.toUpperCase()
+            },
+          },
+        },
+      },
+    })
+
+    const user = await xprisma.user.findFirst()
+    expect(user?.firstName).toBe('JOHN')
+  })
+
   test('empty extension does nothing', async () => {
     const xprisma = prismaWithExtension()
       .$extends({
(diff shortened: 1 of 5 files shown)


packages/client/tests/functional/extensions/enabled/result.ts
```diff
@@ -206,6 +206,24 @@ testMatrix.setupTestSuite(() => {
     expect(user?.loudName).toBe('JOHN SMITH')
   })
 
+  test('shadowing dependency', async () => {
```
millsp, line 209: Would it be useful to add a test that shadows a shadowed field like in the linked issue?

SevInf: This was a really good suggestion since the functionality was actually broken. Now it's fixed and PR is ready for another re-review.

millsp (approved):

PR #17113, opened 2023-01-03
tests: Rename uuid.test.ts to uuid-coercion.test.ts

janpio:
Test gave false impression of testing uuid, but really tests something related to coercion from uuid to strong.


Jolg42:
Looks like it's about making sure that what goes into `String[]` doesn't blow up during validation because of a uuid, so actually testing the runtime validation for `String[]`

janpio:
Replaced by https://github.com/prisma/prisma/pull/17969

PR #17126, opened 2023-01-04
feat: openssl investigation follow-up

### LINKED ISSUE #16962: OS Support: improve SSL detection on `debian` distros

jkomyno:
Draft investigation inspired by [this Slack thread](https://prisma-company.slack.com/archives/C04DG9JL6TV/p1672770455111859).

This adds:
- debug (`prisma:get-platform`) logging when resolving which `binaryTarget` Prisma should download
- uses `ls -r` and `ldconfig -p | ... | grep ssl | sort -r` to retrieve the newest available versions of `openssl` first
- RHEL-specific libssl paths to default to

This also closes https://github.com/prisma/prisma/issues/16962, as we now try both `/usr/lib/` and `/lib` baseline paths when looking for `libssl.so*`


packages/get-platform/src/getPlatform.ts
```diff
@@ -151,60 +156,75 @@ type GetOpenSSLVersionParams = {
  * On Linux, returns the OpenSSL version excluding the patch version, e.g. "1.1.x".
  * Reading the version from the libssl.so file is more reliable than reading it from the openssl binary.
  * This function never throws.
+ *
+ * TODO: we should probably validate the output of this function, and ensure it's contained in `supportedLibSSLVersions`
```
Jolg42, line 160: Is that a TODO for now or later?

packages/get-platform/src/getPlatform.ts
```diff
@@ -151,60 +156,75 @@ type GetOpenSSLVersionParams = {
...
+    'ls -r /lib64 | grep ssl',
+    'ls -r /usr/lib64 | grep ssl',
   ])
 
-  if (libsslVersion) {
-    const matchedVersion = parseLibSSLVersion(libsslVersion)
-    if (matchedVersion) {
-      return matchedVersion
+  if (libsslFilename) {
+    debug(`Found libssl.so file using "ldconfig" and other generic paths: ${libsslFilenameFromSpecificPath}`)
```
Jolg42, line 206:
```suggestion
    debug(`Found libssl.so file using "ldconfig" or other generic paths: ${libsslFilenameFromSpecificPath}`)
```
nit

packages/get-platform/src/getPlatform.ts
jkomyno: For https://github.com/prisma/prisma/issues/17154

Jolg42 (approved):
This looks good! And nice to have more debug logs for this!

PR #17132, opened 2023-01-04
chore(*): switch make-dir to fs-extra + remove @types/ws & rimraf

Jolg42:
@types/ws and rimraf were unused in the modified package.json in this PR

https://www.npmjs.com/package/fs-extra is maintained and reduces the number of dependencies

https://www.npmjs.com/package/make-dir is not maintained anymore, last publish was 3 years ago


jkomyno (approved):
Removing code feels nice, doesn't it? ☺️

PR #17145, opened 2023-01-04
refactor: Introduce typesafe builders for TS typedefs

SevInf:
Instead of building strings, introduces set of helpers that allow us to
build definitions in a typesafe manner. For example:

```ts
ts.objectType()
    .add(ts.property("foo", ts.namedType("MyType")))
```

will generate `{ foo: MyType }`.

Advantages compared to template strings:
- Type safe: if our TS compilation passes, they should always produce valid
  TS code.
- Can be reused, mixed, combined into more high-level helpers without
  losing sturctural information.
- Allows to easily split complex type building into stages
- Takes care of formatting and indentation management and ensures
  consistent formatting in generated code.

I also refactored a couple of generator parts to showcase usage (see
`PrismaClient` and `Input`).

The goal is to be able to use the helpers for new code and
opportunitistically refactor old one.


packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -679,9 +679,9 @@ export class PrismaClient<
    * 
    * Read more in our [docs](https://www.prisma.io/docs/concepts/components/prisma-client/transactions).
    */
-  $transaction<P extends PrismaPromise<any>[]>(arg: [...P]): Promise<UnwrapTuple<P>>;
+  $transaction<P extends PrismaPromise<any>[]>(arg: [...P]): Promise<UnwrapTuple<P>>
```
SevInf, line 682: Note: diff int this file is mostly 3 things:

- Consistent formatting of doc comments
- Consistency regarding semicolon usage
- Consistency regarding `Foo[]` vs `Array<Foo>` type annotation.

danstarns: Wondering if we could run the generated typescript through a linter after generating... maybe not for this PR but something like this would make the style & formatting of the generated types much more consistent.

danstarns (approved):
Nice work, this will really improve code readability. Wondering if, at another time, we should do something like this for GraphQL queries too.

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
SevInf: Ideally, if we slowly migrate generator to this style in generators, our formatting would be consistent.

Issue #16232, opened 2022-11-10
Error: Unknown binaryTarget linux-arm64-openssl-undefined and no custom engine files were provided                               


janpio:
>         @janpio  I've recreated it with the following minimal setup.
> 
> Dockerfile
> 
> ```
> ARG FUNCTION_DIR="/var/task"
> FROM node:16-slim
> 
> # Include global arg in this stage of the build
> ARG FUNCTION_DIR
> 
> WORKDIR ${FUNCTION_DIR}
> COPY . ${FUNCTION_DIR}
> 
> RUN yarn add prisma
> RUN yarn prisma generate
> ```
> 
> prisma.schema
> ```
> // This is your Prisma schema file,
> // learn more about it in the docs: https://pris.ly/d/prisma-schema
> 
> generator client {
>   provider      = "prisma-client-js"
>   binaryTargets = ["rhel-openssl-1.0.x", "linux-arm64-openssl-1.0.x"]
> }
> 
> datasource db {
>   provider = "mongodb"
>   url      = env("DATABASE_URL")
> }
> 
> model User {
>    objectID  String   @id @default(auto()) @map("_id") @db.ObjectId
> }
> ```
> 
> docker-compose.yml
> ```
> version: "3.9"
> services:
>   api:
>     build: .
>     volumes:
>       - ./:/var/task
> ```
> 
> When running `docker compose build` you should get this error
> 
> ```
> Error: Unknown binaryTarget linux-arm64-openssl-undefined and no custom engine files were provided                               
> #0 1.446 error Command failed with exit code 1.
> ```

_Originally posted by @jcampbell05 in https://github.com/prisma/prisma/issues/861#issuecomment-1310206111_

jcampbell05:
So this error is unrelated to ARM and is caused by a lack of an installed SSL library. Some slim Docker images no longer come with a SSL library installed by default in order to keep the image as small as possible. This is why switching to a full sized image resolves this issue.

The fix is to simply install a SSL library.

Unfortunately the error from Prisma doesn't provide any context that this is the issue. So I would suggest the following change:

- Change the name of the Linux based package to indicate the number in the name is supposed to represent the version of the locally installed SSL library i.e "linux-rhel-ssl-1.0.0". Wasn't clear that this version was related to that.
- When a linux library has been selected (Either manually or automatically via "native") then if a SSL library isn't installed rather than return "undefined" as the version number. Throw an error indicating that they lack a SSL library.

[...]

PR #17151, opened 2023-01-05
fix: "linux-arm64-openssl-undefined" is now "linux-arm64-openssl-1.1.x"

jkomyno:
Contributes https://github.com/prisma/prisma/issues/16232

`libssl` can be undefined when the following blocks are run:

```typescript
if (platform === 'linux' && arch === 'arm64') {
  // 64 bit ARM
  return `linux-arm64-openssl-${libssl}` as Platform
}

if (platform === 'linux' && arch === 'arm') {
  // 32 bit ARM
  return `linux-arm-openssl-${libssl}` as Platform
}
```

which could happen e.g. in `node:lts-buster-slim` / `node:16-slim`, as Debian Buster doesn't come with any `libssl` installed by default. This fix at least tells the uses which `binaryTarget` Prisma tried to download, but Prisma will likely still fail at runtime if `libssl` is actually missing from the system.


packages/get-platform/src/getPlatform.ts
```diff
@@ -282,7 +287,7 @@ export async function getPlatform(): Promise<Platform> {
 
   // use the debian build with OpenSSL 1.1 as a last resort
   // TODO: perhaps we should default to 'debian-openssl-3.0.x'
-  return 'debian-openssl-1.1.x'
+  return `debian-openssl-${defaultLibssl}`
```
Jolg42, line 290: Maybe this makes the change a bit clearer for the untouched code?
```diff
  // if just OpenSSL is known, fallback to debian with a specific libssl version
  if (osResult.libssl) {
-    return ('debian-openssl-' + osResult.libssl) as Platform
+    return ('debian-openssl-' + osResult.libssl) as Platform
  }
```

Jolg42 (approved):
Looks good 👍🏼 

(See comment before merging)

packages/get-platform/src/getPlatform.ts
jkomyno: I've followed a similar advice to make the default more explicit, see https://github.com/prisma/prisma/pull/17151/commits/6b9dfb64b7bc632a174ab2b0ada71f70c2132b1f

janpio (changes requested):
Internal discussion: https://prisma-company.slack.com/archives/C04DG9JL6TV/p1672933294609189?thread_ts=1672926697.271059&cid=C04DG9JL6TV

jkomyno:
Note: the code of this PR was included in https://github.com/prisma/prisma/pull/17334, which got merged for `prisma@4.9.0`.

Issue #11730, opened 2022-02-10
Reduce Prisma client's bundle size


vladinator1000:
## Problem
Hi folks, I've been using Prisma with Cloudflare Workers and I noticed Prisma takes up 40% of my size allowance. Sometimes my script fails to upload because it's so big. The client alone takes up 400kB.

<img width="823" alt="Code_d3cZDDYTT7" src="https://user-images.githubusercontent.com/1226564/153333682-a8c21e4a-60f1-4bda-bb38-5465f5668f24.png">

## Solution
Reduce / cull dependencies in favor of smaller libraries / in-house versions of code or use tree-shaking when building the client.

vladinator1000:
Here's how I'm dealing with this temporarily.

```ts
async function getPrismaClient() {
// Had to import prisma dynamically because it's too big, otherwise Cloudflare would errored like this:
// Script startup timed out. This could be due to script exceeding size limits or expensive code in the global scope.
     const { PrismaClient } = await import('@prisma/client/edge')
     return new PrismaClient()
 }
```

pencilcheck:
<img width="1260" alt="Screenshot 2023-01-19 at 4 26 39 AM" src="https://user-images.githubusercontent.com/116532/213442707-200167db-ec62-41fd-bb9f-ac3e167712a6.png">
For me it is huge, not sure why...

janpio:
The issue here was specifically about the sice of `@prisma/client/edge` @pencilcheck, I think your screenshot shows `@prisma/client` itself, probably including the Query Engine (so not the edge Client) - and then a size of 17MB is actually expected. If you think this is a problem, please open a new discussion or issue. Thanks.

SevInf:
Hi everyone!
Enabling [json protocol](https://github.com/prisma/prisma/issues/18095) will reduce size of generated client (the gain is bigger the bigger your schema is). Version 4.14 further reduces the size of both client and runtime.  Please, let us know if that works for you.

millsp:
Note for later: We are using some polyfills for `Buffer` and `util` and we certainly don't use their full API, and they cannot be tree-shaken due to the old style they have been written with. We could port only the portions of interest to save around 20kB.

PR #17156, opened 2023-01-05
feat(client): split runtime by the engine and minify it

SevInf:
Builds separate runtime files for library, binary and data proxy
engines. Generated client will then use corresponding runtime.

This decreases the size of library engine runtime by half and data proxy
by 2/3.

Close prisma/client-planning#202

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -69,6 +69,7 @@ const ALTER_RE = /^(\s*alter\s)/i
 declare global {
   // eslint-disable-next-line no-var
   var NODE_CLIENT: true
+  const TARGET_ENGINE_TYPE: 'binary' | 'library' | 'data-proxy' | 'all'
 }
 
 // used by esbuild for tree-shaking
@@ -76,6 +77,13 @@ typeof globalThis === 'object' ? (globalThis.NODE_CLIENT = true) : 0
 
 function isReadonlyArray(arg: any): arg is ReadonlyArray<any> {
   return Array.isArray(arg)
+}
+
+if (typeof TARGET_ENGINE_TYPE !== 'undefined' && TARGET_ENGINE_TYPE === 'all') {
+  console.warn('imports from "@prisma/client/runtime" are deprecated.')
+  console.warn(
+    'Use "@prisma/client/runtime/library",  "@prisma/client/runtime/data-proxy" or  "@prisma/client/runtime/binary"',
+  )
 }
 
 // TODO also check/disallow for CREATE, DROP
@@ -454,12 +462,18 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
     }
 
     getEngine(): Engine {
-      if (this._dataProxy === true) {
+      if (this._dataProxy === true && (TARGET_ENGINE_TYPE === 'data-proxy' || TARGET_ENGINE_TYPE === 'all')) {
         return new DataProxyEngine(this._engineConfig)
-      } else if (this._clientEngineType === ClientEngineType.Library) {
-        return NODE_CLIENT && new LibraryEngine(this._engineConfig)
-      } else if (this._clientEngineType === ClientEngineType.Binary) {
-        return NODE_CLIENT && new BinaryEngine(this._engineConfig)
+      } else if (
+        this._clientEngineType === ClientEngineType.Library &&
+        (TARGET_ENGINE_TYPE === 'library' || TARGET_ENGINE_TYPE === 'all')
+      ) {
+        return new LibraryEngine(this._engineConfig)
+      } else if (
+        this._clientEngineType === ClientEngineType.Binary &&
+        (TARGET_ENGINE_TYPE === 'binary' || TARGET_ENGINE_TYPE === 'all')
+      ) {
+        return new BinaryEngine(this._engineConfig)
       }
 
       throw new PrismaClientValidationError('Invalid client engine type, please use `library` or `binary`')
--- a/packages/client/src/runtime/utils/CallSite.ts
+++ b/packages/client/src/runtime/utils/CallSite.ts
@@ -40,6 +40,8 @@ class EnabledCallSite implements CallSite {
         !t.file.endsWith('/runtime/binary.js') && // Bundled runtimes
         !t.file.endsWith('/runtime/library.js') &&
         !t.file.endsWith('/runtime/data-proxy.js') &&
+        !t.file.endsWith('/runtime/edge.js') &&
+        !t.file.endsWith('/runtime/edge-esm.js') &&
         !t.file.startsWith('internal/') && // We don't want internal nodejs files
         !t.methodName.includes('new ') && // "new CallSite" call and maybe other constructors
         !t.methodName.includes('getCallSite') && // getCallSite function from this module
--- a/packages/client/src/runtime/utils/CallSite.ts
+++ b/packages/client/src/runtime/utils/CallSite.ts
@@ -1,3 +1,4 @@
+import path from 'path'
 import * as stackTraceParser from 'stacktrace-parser'
 
 import { ErrorFormat } from '../getPrismaClient'
@@ -32,17 +33,22 @@ class EnabledCallSite implements CallSite {
     // TODO: more resilient logic to check that it's not relative to cwd
     const frame = stackFrames.find((t) => {
       // Here we are trying to find the location in the user's code which caused the error
+      if (!t.file) {
+        return false
+      }
+
+      // convert windows path to posix path
+      const posixFile = t.file.split(path.sep).join('/')
       return (
-        t.file &&
-        t.file !== '<anonymous>' && // Ignore as we can not read an <anonymous> file
-        !t.file.includes('@prisma') && // Internal, unbundled code
-        !t.file.includes('/packages/client/src/runtime/') && // Runtime sources when source maps are used
-        !t.file.endsWith('/runtime/binary.js') && // Bundled runtimes
-        !t.file.endsWith('/runtime/library.js') &&
-        !t.file.endsWith('/runtime/data-proxy.js') &&
-        !t.file.endsWith('/runtime/edge.js') &&
-        !t.file.endsWith('/runtime/edge-esm.js') &&
-        !t.file.startsWith('internal/') && // We don't want internal nodejs files
+        posixFile !== '<anonymous>' && // Ignore as we can not read an <anonymous> file
+        !posixFile.includes('@prisma') && // Internal, unbundled code
+        !posixFile.includes('/packages/client/src/runtime/') && // Runtime sources when source maps are used
+        !posixFile.endsWith('/runtime/binary.js') && // Bundled runtimes
+        !posixFile.endsWith('/runtime/library.js') &&
+        !posixFile.endsWith('/runtime/data-proxy.js') &&
+        !posixFile.endsWith('/runtime/edge.js') &&
+        !posixFile.endsWith('/runtime/edge-esm.js') &&
+        !posixFile.startsWith('internal/') && // We don't want internal nodejs files
         !t.methodName.includes('new ') && // "new CallSite" call and maybe other constructors
         !t.methodName.includes('getCallSite') && // getCallSite function from this module
         !t.methodName.includes('Proxy.') && // Model proxies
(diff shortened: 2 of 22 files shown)


SevInf:
Results:
| Runtime       | Size, bytes | Difference, bytes |
|---------------|-------------|-------------------|
| all(baseline) | 1,305,137   | 0                 |
| library       | 612,378     | -692,759          |
| data-proxy    | 463,806     | -841,331          |
| binary        | 1,253,838   | -51,299           |

Note: the size difference will be visible amount of code loaded/bundled. Size of `@prisma/client` package on disk will increase by 2.3 MB because it will now contain 3 additional versions of the runtime.

PR continues to build version of runtime with all engines bundled for backward compatibility: I've seen `import {something} 
 from '@prisma/client/runtime'` popping up in user's reproduction from time to time. I am not sure it ever supposed to be public API - if we decide if it is not, we could stop building this version, then npm size of the package will increase only by `~1 MB`. 
 
 We could also stop using undici in binary engine - then we might produce roughly the same size package.

Jolg42:
`'@prisma/client/runtime'` is quite popular apparently (and used by popular projects)
See search 
https://cs.github.com/?scopeName=All+repos&scope=&q=%40prisma%2Fclient%2Fruntime

Random collection of results
```
import { DMMF } from '@prisma/client/runtime'
import type { DMMF } from '@prisma/client/runtime';
import { DMMFClass } from '@prisma/client/runtime';
import { PrismaClientKnownRequestError } from "@prisma/client/runtime";
import {
  PrismaClientKnownRequestError,
  PrismaClientValidationError,
} from "@prisma/client/runtime";
export {Decimal} from '@prisma/client/runtime'
const { raw } = require("@prisma/client/runtime");
```

SevInf:
Latest commit enables minifcation of the runtime during build. That gives us following numbers:

```
| Runtime    | Size |
|------------|------|
| all        | 612K |
| library    | 244K |
| data-proxy | 216K |
| binary     | 584K |
```

Error messages will need to be fixed for it to be mergeable.

SevInf:
This is ready for another review. Summary of what happened since the last time it was updated:
- we now generate source maps for minified runtime. `jest` picks them out automatically, reproduction folder scripts got adjusted to do so as well. We'll now have the stacktraces that point to original sources in our dev setup.
- we don't copy source maps into generated client, since it provides little value to the end users in my opinion (they are mostly useful for us), but blows up the size of the client. It is possible to get them copied across with env variable.
- import from `@prisma/client/runtime` will produce a warning, telling users to pick one of the specific runtimes. 
- bunch of fancy errors fixes: some of them were needed to adjust for minification, some of them were fixes for the stuff that is broken on `main` as well.

Jolg42:
I think this issue https://github.com/prisma/prisma/issues/11730 might be a good place to "report" the improvements? (add a comment with the decrease in size) 
What do you think @SevInf ?

packages/client/tests/functional/jest.config.js
```diff
@@ -2,7 +2,7 @@
 const os = require('os')
 const path = require('path')
 
-const runtimeDir = path.dirname(require.resolve('../../runtime'))
+const runtimeDir = path.dirname(require.resolve('../../runtime/library'))
```
Jolg42, line 5: Wondering if this is important: In CI we also run all the tests with these env vars
```
    PRISMA_CLI_QUERY_ENGINE_TYPE: binary
    PRISMA_CLIENT_ENGINE_TYPE: binary
```

So it feels wrong that the `runtimeDir` is set to `library` in the case CI runs with `binary`?

SevInf:
>I think this issue https://github.com/prisma/prisma/issues/11730 might be a good place to "report" the improvements? (add a comment with the decrease in size)
>What do you think @SevInf ?

Edge client is mostly unaffected by this unfortinately — it was altready minified and excluded all engines except for data proxy.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -63,6 +63,7 @@ const ALTER_RE = /^(\s*alter\s)/i
 declare global {
   // eslint-disable-next-line no-var
   var NODE_CLIENT: true
+  const TARGET_ENGINE_TYPE: 'binary' | 'library' | 'data-proxy' | 'all'
```
Jolg42, line 66: Since these would leak as global vars (right?), maybe it would be better to prefix them with `PRISMA_`

Jolg42:
@SevInf Maybe double check where `@prisma/client/runtime` is mentioned in the codebase? https://cs.github.com/?q=%40prisma%2Fclient%2Fruntime%20repo%3Aprisma%2Fprisma&scopeName=prisma&scope=org%3Aprisma

Also, it looks like it is used in 
- `pdp-control-plane` [see here](https://cs.github.com/?q=%40prisma%2Fclient%2Fruntime%20repo%3Aprisma%2Fpdp-control-plane&scopeName=prisma&scope=org%3Aprisma) maybe ping Jason/Team CDN
- `reflector` [see here](https://cs.github.com/?q=%40prisma%2Fclient%2Fruntime%20repo%3Aprisma%2Freflector&scopeName=prisma&scope=org%3Aprisma) ping Jason/Team CDN about this change?
- `accelerate` [see here](https://cs.github.com/?q=%40prisma%2Fclient%2Fruntime%20repo%3Aprisma%2Fprisma-extension-accelerate&scopeName=prisma&scope=org%3Aprisma) ping Team Real-time about this?

packages/client/src/runtime/getPrismaClient.ts
millsp: I think these are always stripped by the compiler when value is provided, that is the case for `NODE_CLIENT` for sure.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -505,12 +513,18 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
     }
 
     getEngine(): Engine {
-      if (this._dataProxy === true) {
+      if (this._dataProxy === true && (TARGET_ENGINE_TYPE === 'data-proxy' || TARGET_ENGINE_TYPE === 'all')) {
```
millsp, line 516: Is esbuild able to tree-shake this now? When I first introduced it, it really had to be `VAR && value` to be able to remove via value injection when it is `false`.

packages/client/src/runtime/utils/CallSite.ts
```diff
@@ -32,18 +33,29 @@ class EnabledCallSite implements CallSite {
     // TODO: more resilient logic to check that it's not relative to cwd
     const frame = stackFrames.find((t) => {
       // Here we are trying to find the location in the user's code which caused the error
+      if (!t.file) {
+        return false
+      }
+
+      // convert windows path to posix path
+      const posixFile = t.file.split(path.sep).join('/')
```
millsp, line 41: Does this explain why we had weird stack traces on windows only with swc?

packages/client/src/runtime/getPrismaClient.ts
millsp: So I guess it just matters in the tests?

packages/client/tests/functional/methods/count/tests.ts
```diff
@@ -98,43 +98,45 @@ testMatrix.setupTestSuite(() => {
...
+        name: true,
+        // @ts-expect-error
+        posts: true,
+      },
+    })
+
+    await expect(err).rejects.toMatchPrismaErrorInlineSnapshot(`
+
+      Invalid \`prisma.user.count()\` invocation in
+      /client/tests/functional/methods/count/tests.ts:0:0
```
millsp, line 116: Nice

packages/client/tests/functional/jest.config.js
millsp: Right, I also wonder if this has any impact on `getTestSuiteClient.ts` since we pass runtime paths there, maybe not.

reproductions/basic-sqlite/package.json
```diff
@@ -5,9 +5,9 @@
   "main": "index.ts",
   "scripts": {
     "dbpush": "prisma db push --skip-generate",
-    "generate": "prisma generate",
-    "test": "ts-node index.ts",
-    "debug": "node -r ts-node/register --inspect-brk index.ts"
+    "generate": "PRISMA_COPY_RUNTIME_SOURCEMAPS=1 prisma generate",
```
millsp, line 8: I assume this works out of the box in functional tests because we link directly to the runtime file?

millsp (commented):
Do you think it would make sense to have a test in the functional tests that can give a signal (eg. a property passed down) that can give us confidence that the correct runtime is in use, for example, we could retrieve it's name in the generated client?

reproductions/basic-sqlite/package.json
SevInf: Yes, and source map is always besides it. Also, `jest` uses `source-map-support` package by default, so there is no need to use `--enable-source-maps` flag.

packages/client/src/runtime/utils/CallSite.ts
SevInf: No, they are couple of lines off there, and we did not filter based on path separator before this PR.

packages/client/src/runtime/getPrismaClient.ts
SevInf: Yes, probably they've improved a bit since then. Also, minification removes more dead code then just tree-shaking.

SevInf:
Added a test

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -78,6 +79,13 @@ function isReadonlyArray(arg: any): arg is ReadonlyArray<any> {
   return Array.isArray(arg)
 }
 
+if (typeof TARGET_ENGINE_TYPE !== 'undefined' && TARGET_ENGINE_TYPE === 'all') {
+  console.warn('imports from "@prisma/client/runtime" are deprecated.')
+  console.warn(
```
millsp, line 84: How does that work for default-index.d.ts?

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -59,6 +60,7 @@ function setupTestSuiteMatrix(
   })
 
   for (const { name, suiteConfig, skip } of testPlan) {
+    const generatedFolder = getTestSuiteFolderPath(suiteMeta, suiteConfig)
```
millsp, line 63: Could this be called directly in the test instead?

millsp (approved):

packages/client/src/runtime/getPrismaClient.ts
SevInf: Ah, forgot about those. I guess right now it will produce a warning and we have no good way to pick up right runtime before client is generated. Do you think it would make sense to remove a warning for now then?

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
SevInf: `suteConfig` we have here is not quite the same test suite gets: this one includes the name of the test and all matrix parameters, and we path only matrix parameters to the callback. Unfortunately, name is needed for this function.

packages/client/src/runtime/getPrismaClient.ts
millsp: Slightly related: on the type emission, how does it work? Are we emitting double the amount of types (1 for each runtime)? (runtime, edge, binary, library)?

SevInf: We are emitting single `.d.ts` file for default runtime, then specific ones just genereate `export * from './index'` definition.

millsp: Nice thanks

packages/client/src/runtime/getPrismaClient.ts
SevInf: Yes, that's correct, those constant never actually exist in compiled bundles.

sighash256:
how can I use this features? my bundle size is more than ~350kb. I'm also using `previewFeatures = ["jsonProtocol"]`.

SevInf:
@sighash256 this was merged into prisma 4.9.0 and available by default since then.

PR #17198, opened 2023-01-09
chore(instrumentation): Update OpenTelemetry dependencies to remove deprecation warning

unflxw:
Updating `@opentelemetry/instrumentation` to `0.34.0` removes `@opentelemetry/api-metrics` as a dependency, therefore removing the deprecation warning about it when running `npm install`.

The deprecation warning says to use `@opentelemetry/api` version `1.3.0` or greater, so use that instead.
https://www.npmjs.com/package/@opentelemetry/api-metrics
Screenshot 2023-01-09 at 13 00 11


Jolg42:
It would make sense to also update the version in the other packages (engine-core & client)
https://github.com/search?q=repo%3Aprisma%2Fprisma+OpenTelemetry+language%3AJSON&type=code&l=JSON

unflxw:
@Jolg42 Thanks for the quick review! Done and done. In the process I ended up updating other dependencies as to avoid peer dependency issues.

Jolg42 (approved):
Thank you, @unflxw, for your first PR here 💚 
This looks great to me, I just asked @danstarns to review as well

danstarns (approved):
Nice work, thanks for this. Looks good to me and all tests passing! Cheers

unflxw:
Thanks y'all @danstarns @Jolg42! And thanks for all the quality work on Prisma and the OpenTelemetry instrumentation! It's great to see a library where OpenTelemetry support is built-in instead of monkey-patched on top.

PR #17200, opened 2023-01-09
chore(*): remove cuid and unused dependencies

Jolg42:
cuid is deprecated and we don't need a dependency just for generating one random id for our tests.
The rest was purely unused 🧹


packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -204,7 +204,7 @@ export type DatasourceInfo = {
 export function setupTestSuiteDbURI(suiteConfig: Record<string, string>, clientMeta: ClientMeta): DatasourceInfo {
   const provider = suiteConfig['provider'] as Providers
   const providerFlavor = suiteConfig['providerFlavor'] as ProviderFlavor | undefined
-  const dbId = cuid()
+  const dbId = faker.random.alphaNumeric(25)
```
SevInf, line 207: Can we add a timstamp to id as well, just to be sure? `faker` data is not unique

Jolg42: Looks like `fzzao-60483-1673272740041` on my local macOS machine

SevInf (approved):

Issue #6485, opened 2021-04-09
Override database connection string for Migrate (and other commands)


janpio:
## Problem

When using a connection pooler like PgBouncer, you need to use a non pooled connection URL for running Migrations. That is inconvenient, as Prisma Client and Migrate use the same datasource definition (with the same ENV var defining the connection string). So if you want to use the pooled connection for Prisma Client, running a Prisma Migrate command by default fails (see e.g. https://github.com/prisma/prisma/issues/6480).

The same is true when you want to use a different user for your Prisma Client that does not have permissions to CREATE, DROP or ALTER tables for example. As Prisma Client and Migrate use the same datasource definition (with the same ENV var defining the connection string), this is not trivial to get right and can lead to quite some confusion on the user side.

## Suggested solution

We could have a dedicated ENV var that Migrate uses to override whatever is defined in the datasource. Naming could be `PRISMA_MIGRATIONS_URL` or similar. Then we could tell people to set that to the un-pooled connection string when using a pooled connection for Prisma Client.

## Alternatives

- Set a different ENV var value only when running Migrate

## Additional context

Internal discussion: https://prisma-company.slack.com/archives/CL0AJEVSA/p1617976543036600

dac09:
This might come in very useful in other situations too, e.g. preventing accidental migrations when running builds on CI for pull requests

janpio:
Added another use case besides PgBouncer to the issue description.

dac09:
Hi @janpio - just wanted to follow up on this again :) Any chance this could get prioritised? Would make migrations a lot easier.

The workaround I was thinking of for now, is to override the database url in the migrate command only like:

in my CI
```
DATABASE_URL=$MIGRATE_DB_URL yarn prisma migrate prod
```

chrisj-back2work:
For those reporting this issue, can you say what your deployment environment looks like?

I've been running migrates in my kubernetes cluster via a pgBouncer connection with no issues.

I wonder if I've just been lucky, or if our environments are different enough to affect the operation.

MentalGear:
I agree, adding an env var for migration sounds simple enough.

bingomanatee:
You could simply have a different host in the db definition as in 
```
datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
  migration_url      = env("MIGRATION_DATABASE_URL")
}
```

Erbros:
[...]

PR #17210, opened 2023-01-09
feat: directUrl support

eviefp:
Add support for `directUrl`.

This is a follow-up for https://github.com/prisma/prisma-engines/pull/3547 and https://github.com/prisma/prisma-engines/pull/3551

I followed the usages of `getConfig`, which is where these changes sprung up from. I would love it if someone would do the same to validate I did not miss anything.

Also, I am unsure on how to test this. Happy to be given some direction here.

Part of https://github.com/prisma/prisma/issues/6485


Jolg42:
@eviefp Could you sign the CLA from https://github.com/prisma/prisma/pull/17210#issuecomment-1375942082?

Also, it looks like we are missing tests for the error message 👀

Jolg42:
We could add a test, for
- https://github.com/prisma/prisma/blob/main/packages/migrate/src/__tests__/ensureDatabaseExists.test.ts
- https://github.com/prisma/prisma/blob/main/packages/migrate/src/__tests__/MigrateDeploy.test.ts

Not sure if it makes sense to test all commands using this for this error message, maybe one is enough.

eviefp:
Added a test for reset, would that be enough?

packages/internals/src/engine-commands/getConfig.ts
```diff
@@ -83,6 +83,12 @@ ${detailsHeader} ${message}`
   }
 }
 
+export function getDirectUrl(ds: DataSource): EnvValue {
```
janpio, line 86: Nit:

This confused me a bit: Why is the method that 99% returns the `url` and only sometimes the `directUrl` called `getDirectUrl`?

But I think I understand now: Either `url` or `directUrl` are the _direct URL_ we want to use. Is that your motivation for this naming?

As "direct URL" and "`directUrl`" are the same term, I think I might name the function `getEffectiveUrl` or something like that.

Jolg42: A rename would be nice to avoid confusion, can be done later if needed

Jolg42 (approved):
Looks good! Let's merge this :shipit: 

About the test failure, a snapshot update is needed or maybe you forgot to add the dev.db file to git?

(see comments)

packages/internals/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -62,8 +62,9 @@ async function checkUnsupportedDataProxyMessage(command: string, args: Args, imp
 
       const datamodel = await fs.promises.readFile(argValue, 'utf-8')
       const config = await getConfig({ datamodel, ignoreEnvVarErrors: true })
-      const urlFromValue = config.datasources[0]?.url.value
-      const urlEnvVarName = config.datasources[0]?.url.fromEnvVar
+      const url = getEffectiveUrl(config.datasources[0])
+      const urlFromValue = url.value
+      const urlEnvVarName = url.fromEnvVar
```
jkomyno, line 67: Nit:

```suggestion
      const { value: urlFromValue, fromEnvVar: urlFromEnvVarName } = getEffectiveUrl(config.datasources[0])
```

janpio (approved):

packages/internals/src/cli/checkUnsupportedDataProxy.ts
Jolg42: since it's only a nit Let's merge without this change, once tests are passing

Jolg42:
Sorry for the merge conflicts.... I can check that tomorrow!

janpio:
```
Error: packages/migrate build: src/utils/ensureDatabaseExists.ts(159,31): error TS2304: Cannot find name 'activeDatasource'.
Error: packages/migrate build: src/utils/ensureDatabaseExists.ts(162,49): error TS2345: Argument of type 'string | null' is not assignable to parameter of type 'string'.
packages/migrate build:   Type 'null' is not assignable to type 'string'.
```

packages/migrate/src/utils/ensureDatabaseExists.ts
```diff
@@ -154,9 +156,10 @@ export async function ensureDatabaseExists(action: MigrateAction, schemaPath?: s
   }
 
   const schemaDir = (await getSchemaDir(schemaPath))!
+  const url = getEffectiveUrl(activeDatasource).value
```
janpio, line 159:
```suggestion
  const url = getEffectiveUrl(firstDatasource).value
```

packages/migrate/src/utils/ensureDatabaseExists.ts
```diff
@@ -154,9 +156,10 @@ export async function ensureDatabaseExists(action: MigrateAction, schemaPath?: s
   }
 
   const schemaDir = (await getSchemaDir(schemaPath))!
+  const url = getEffectiveUrl(firstDatasource).value
 
-  // url.value exists because `ignoreEnvVarErrors: false` would have thrown an error if not
-  const canConnect = await canConnectToDatabase(firstDatasource.url.value!, schemaDir)
+  // url exists because `ignoreEnvVarErrors: false` would have thrown an error if not
+  const canConnect = await canConnectToDatabase(url, schemaDir)
```
janpio, line 162:
```suggestion
  const canConnect = await canConnectToDatabase(url!, schemaDir)
```

PR #17225, opened 2023-01-10
feat(internals): Warn when 2 questions marks are used in connection s…

edtervit:
Warn when 2 questions marks (instead of question mark and ampersand) are used in connection strings

Closes #16677

My first ever open source PR, be gentle 

Will throw error for following connection Url: "mysql://root:root@localhost:3306/prisma-test?test=true?test2=false"

### DIFF
--- a/packages/internals/src/migrateEngineCommands.ts
+++ b/packages/internals/src/migrateEngineCommands.ts
@@ -2,6 +2,7 @@ import { BinaryType } from '@prisma/fetch-engine'
 import execa from 'execa'
 import fs from 'fs'
 import path from 'path'
+import * as NodeURL from 'url'
 import { promisify } from 'util'
 
 import { getSchemaDir } from './cli/getSchema'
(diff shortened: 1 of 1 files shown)


Jolg42 (changes requested):
Hi! Thanks @edtervit for your first PR here 🙌🏼 

Don't be scared by the red status, I just want to make it obvious that this should not be merged.

I think this might be better done in the engines (in Rust), so it's done once. This PR only contains a change that adds a check in a particular place.

Probably in one of these, but I'm unsure what would be best.
https://github.com/prisma/prisma-engines/blob/main/psl/psl-core/src/validate/datasource_loader.rs
https://github.com/search?q=repo%3Aprisma%2Fprisma-engines%20validate_url&type=code

I can ask a colleague if you are motivated to continue. Sounds good?

edtervit:
> Hi! Thanks @edtervit for your first PR here 🙌🏼
> 
> Don't be scared by the red status, I just want to make it obvious that this should not be merged.
> 
> I think this might be better done in the engines (in Rust), so it's done once. This PR only contains a change that adds a check in a particular place.
> 
> Probably in one of these, but I'm unsure what would be best. https://github.com/prisma/prisma-engines/blob/main/psl/psl-core/src/validate/datasource_loader.rs https://github.com/search?q=repo%3Aprisma%2Fprisma-engines%20validate_url&type=code
> 
> I can ask a colleague if you are motivated to continue. Sounds good?

Hi mate,

I appreciate the kind feedback but I've never touched rust before so is an issue probably best left to somebody else.

Jolg42:
Understandable! Sorry 🙏🏼 for the issue being unclear about this, I posted that additional info in the original issue.

PR #17239, opened 2023-01-11
refactor: Select and Include type generation

SevInf:
Split type generation from `Model.ts` and use ts-builders for
generation. Adds several helpers, necessary for this refactoring:
`typeDeclaration`, `export` and `keyType`.


packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -1736,9 +1736,7 @@ export namespace Prisma {
   }
 
 
-  export type EmbedInclude = {
-
-  } 
+  export type EmbedInclude = {}
```
SevInf, line 1739: Whitespace changes only

packages/client/src/generation/ts-builders/KeyType.ts
```diff
@@ -0,0 +1,34 @@
...
+import { BasicBuilder } from './BasicBuilder'
+import { FunctionType } from './FunctionType'
+import { ObjectType } from './ObjectType'
+import { UnionType } from './UnionType'
+import { Writer } from './Writer'
+
+export class KeyType implements BasicBuilder {
+  constructor(public baseType: AnyTypeBuilder, public key: string) {}
+  write(writer: Writer): void {
+    const needsBraces =
```
aqrln, line 11: Should this be something like `hasBraces` or `needsParens` instead?

Also, would it make sense to define this as a method in `BasicBuilder` interface and let each type tell if it needs to be wrapped when indexing, or is not worth it and not going to be useful elsewhere?

packages/client/src/generation/ts-builders/TypeDeclaration.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  expect(stringify(decl)).toMatchInlineSnapshot(`
+    /**
+     * Type for stuff
+     */
+    type B = A
+  `)
+})
+
+test('with generic parameters', () => {
+  const decl = typeDeclaration('B', A).addGenericParameter(genericParameter('T'))
```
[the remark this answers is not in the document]
SevInf: Added.

packages/client/src/generation/ts-builders/KeyType.ts
SevInf: Thanks, did that in the next commit.

aqrln (approved):
💯

Issue #14042, opened 2022-06-28
OpenSSL version detection


Maxime-J:

fakel:
Got a similar problem, my host provider is using CloudLinux but prisma detects the OS as debian instead of RHEL and the version of OpenSSL to 1.1.x instead of 1.0.x

Maxime-J:
Hi @jkomyno , it looks good!

I tested the get-platform which has been merged into main, but there might be one issue.
In that block:
```
const libsslFilename: string | undefined = await getFirstSuccessfulExec([
  'ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl | sort',
  'ls /lib64 | grep ssl',
  'ls /usr/lib64 | grep ssl',
])
```
First set of commands returns
```
sed: -e expression #1, char 7: unknown option to `s'
```
So on a setup with undefined distro, it still goes to the `openssl version -v` fallback

Not familiar with sed though, so I can't debug straight away.

jkomyno:
@Maxime-J you're right, as per [this thread](https://stackoverflow.com/a/9366940/6174476), it should be something like:

```typescript
/**
 * The `ldconfig -p` returns the dynamic linker cache paths, where libssl.so files are likely to be included.
 * Each line looks like this:
 * 	libssl.so (libc6,hard-float) => /usr/lib/arm-linux-gnueabihf/libssl.so.1.1
 * But we're only interested in the filename, so we use sed to remove everything before the `=>` separator,
 * and then we remove the path and keep only the filename.
 * The second sed commands uses `|` as a separator because the paths may contain `/`, which would result in the
 * `unknown option to 's'` error (see https://stackoverflow.com/a/9366940/6174476) - which would silently
 * fail with error code 0.
 */
'ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep ssl | sort',
```

Thanks for helping us discovering this.

jkomyno:
@Maxime-J I believe this issue is fixed with today's release. May you please try using `prisma@4.9.0` and confirm this issue is resolved? Thank you!

Maxime-J:
@jkomyno I confirm, issue is resolved.
Thanks!

jkomyno:
Thanks for your answer @Maxime-J! Closing this as it's resolved.

mattgraphlan:
[...]

PR #17240, opened 2023-01-11
Revert "feat: openssl investigation follow-up"

jkomyno:
Reverts prisma/prisma#17126

PR #17241, opened 2023-01-11
feat: prefer older libssl versions rather than the newest ones (partial revert of #17126)

### LINKED ISSUE #17264: #17126 broke the `platform-serverless-vercel/vercel-cli` ecosystem tests
- [x] we should understand why that's the case
- [x] revert the PR and gradually add back the `openssl` improvements we intended, by monitoring the ecosystem-tests

jkomyno:
Integration PR (which deprecates https://github.com/prisma/prisma/pull/17240) that reverts the "pick up the most recent libssl.so version" feature from prisma/prisma#17126.

Closes https://github.com/prisma/prisma/issues/17264.

## Vercel issues

#17240 broke ecosystem-tests for Vercel serverless.
As it turns out, Vercel uses a different environment for building and running serverless functions, and these environments have different `libssl.so` system dependencies installed.

The Vercel **build** environment has the following `libssl.so` versions installed, as can be seen via `ls -r /lib64 | grep libssl.so`:

```sh
libssl.so.1.1.1g
libssl.so.1.1
libssl.so.1.0.2k
libssl.so.10
libssl.so
```

which resolves to the `openssl-1.1.x` binaryTarget.

The Vercel **runtime**, on the other hand, has only:

```sh
libssl.so.1.0.2k
libssl.so.10
```

which resolves to `openssl-1.0.x` binaryTarget.

Prisma Client would then fail because it cannot find `libquery-openssl-1.1.x` at runtime.

Takeaway: for the time being we shouldn't use the most recent versions of `libssl.so`, rather the oldest ones. It's ugly :(


janpio:
Ecosystem tests run that show that this indeed fixes the failing vercel ecosystem tests: https://github.com/prisma/ecosystem-tests/actions/runs/3892460754 (The remaining deno failure is known and expected)

packages/get-platform/src/getPlatform.ts
```diff
@@ -156,34 +156,33 @@ type GetOpenSSLVersionParams = {
...
-    .with({ distro: 'rhel' }, async () => {
-      /* Linux Red Hat, OpenSuse etc */
-      debug('Trying platform-specific paths for "rhel"')
-      return Promise.resolve(['/lib64', '/usr/lib64'])
-    })
+    // .with({ distro: 'rhel' }, () => {
+    //   /* Linux Red Hat, OpenSuse etc */
+    //   debug('Trying platform-specific paths for "rhel"')
+    //   return ['/lib64', '/usr/lib64']
+    // })
```
jkomyno, line 178:
```suggestion
    .with({ distro: 'rhel' }, () => {
      /* Linux Red Hat, OpenSuse etc */
      debug('Trying platform-specific paths for "rhel"')
      return ['/lib64', '/usr/lib64']
    })
```

aqrln:
> The Vercel **build** environment has the following `libssl.so` versions installed, as can be seen via `ls -r /lib64 | grep libssl.so`:
> 
> ```sh
> libssl.so.1.1.1g
> libssl.so.1.1
> libssl.so.1.0.2k
> libssl.so.10
> libssl.so
> ```
> 
> which resolves to the `openssl-1.0.x` binaryTarget.
> 
> The Vercel **runtime**, on the other hand, has only:
> 
> ```sh
> libssl.so.1.0.2k
> libssl.so.10
> ```
> 
> which resolves to `openssl-1.1.x` binaryTarget.

Should this be the other way around, i.e. 1.1 in the first case and 1.0 in the second case? (I assume it's a typo, but wanted to confirm just in case.)

jkomyno:
- ecosystems-tests run: https://github.com/prisma/ecosystem-tests/actions/runs/3903136873/jobs/6667102884
- `docker/alpine:3.17` is failing merely because it’s expecting `openssl-3.0.x` rather than `openssl-1.1.x` (both versions of libssl are installed in the system, and we’re selecting the oldest one)
- many `platforms-serverless-vercel` tests that were failing before. now pass
- however, `vercel-cli` still fails because we detect that `Postinstall hook was NOT ADDED`, in which case we explicitly exit with error code `1`.

jkomyno:
Incidentally, this PR may close https://github.com/prisma/prisma/issues/14042. See [this comment](https://github.com/prisma/prisma/issues/14042#issuecomment-1380692643).

jkomyno:
## TODOs (if this PR is accepted as is)

- [ ] update `EXPECTED_PRISMA_TARGET_PLATFORM` for `docker/alpine:3.17` images to expect `openssl-1.1.x` rather than `openssl-3.0.x`
- [ ] update `platform-serverless gcp/firebase` to expect `openssl-1.0.x` rather than `openssl-1.1.x`

packages/get-platform/src/getPlatform.ts
```diff
@@ -156,34 +156,33 @@ type GetOpenSSLVersionParams = {
  * On Linux, returns the OpenSSL version excluding the patch version, e.g. "1.1.x".
  * Reading the version from the libssl.so file is more reliable than reading it from the openssl binary.
  * This function never throws.
- *
- * TODO: we should probably validate the output of this function, and ensure it's contained in `supportedLibSSLVersions`
  */
 export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetOSResult['libssl'] | undefined> {
-  const libsslSpecificPaths = await match(args)
-    .with({ distro: 'musl' }, async () => {
+  const archFromUname = await getArchFromUname()
```
aqrln, line 161: Can we move it to where it's used so we don't call `uname` on platforms where we don't need it?

packages/get-platform/src/getPlatform.ts
```diff
@@ -156,34 +156,33 @@ type GetOpenSSLVersionParams = {
...
       /* Linux Red Hat, OpenSuse etc */
       debug('Trying platform-specific paths for "rhel"')
-      return Promise.resolve(['/lib64', '/usr/lib64'])
+      return ['/lib64', '/usr/lib64']
     })
     .otherwise(({ distro, arch }) => {
       /* Other Linux distros, we don't do anything specific and fall back to the next blocks */
       debug(`Don't know any platform-specific paths for "${distro}" on ${arch}`)
-      return Promise.resolve(undefined)
+      return undefined
```
aqrln, line 182: Would it be better to return an empty array here and remove the default value below on line 185?

packages/get-platform/src/getPlatform.ts
```diff
@@ -197,9 +196,9 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
 
   debug('Falling back to "ldconfig" and other generic paths')
   const libsslFilename: string | undefined = await getFirstSuccessfulExec([
-    'ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl | sort -r',
-    'ls -r /lib64 | grep ssl',
-    'ls -r /usr/lib64 | grep ssl',
+    'ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl',
```
aqrln, line 199: maybe

```suggestion
    'ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl | sort',
```

to be safe

aqrln: or is it guaranteed to be lexicographically ordered already?

jkomyno: Hi, I also had second thoughts on this. It's probably better to keep `| sort` and be sure with it, as the paths returned by `ldconfig -p` may have wildly different basedirs

packages/get-platform/src/getPlatform.ts
jkomyno: Yes, we can do that

packages/get-platform/src/getPlatform.ts
jkomyno: I'd rather not: this was something we did before, but made it cumbersome dealing with all the non-`debian` pattern branches: they had useless `async` declarations and needed `Promise.resolve` before returning the result. Besides, The function that computes `uname` is guaranteed to not raise errors by design, is cheap to call and lies in a non-performance-critical block of Prisma

Jolg42:
Recent ecosystem-test run https://github.com/prisma/ecosystem-tests/actions/runs/3910377877

aqrln (approved):

PR #16868, opened 2022-12-16
fix(*): change datasource printing for multiSchema + cleanup

tomhoule:
closes https://github.com/prisma/prisma/issues/10043
closes https://github.com/prisma/prisma/issues/16635

Better error for https://github.com/prisma/prisma/issues/14732

PR #17034, opened 2022-12-27
feat(migrate): add `--schemas` param to `prisma db pull`

jkomyno:
Contributes to https://github.com/prisma/prisma/issues/16455.

Can be used like 
```
db pull --url=... --schemas=base
db pull --url=... --schemas=base,something,else
```

It requires the multiSchema preview feature, the check is done in engines

PR #17252, opened 2023-01-11
fix(dbPull): Normalize postgres to postgresql

janpio:
Solves problem that was introduced in #16868 when `postgres` was used as the provider instead of `postgresql`.

Definitely a test missing for this though.

### DIFF
--- a/packages/migrate/src/commands/DbPull.ts
+++ b/packages/migrate/src/commands/DbPull.ts
@@ -180,7 +180,7 @@ Set composite types introspection depth to 2 levels
           const firstDatasource = config.datasources[0] ? config.datasources[0] : undefined
 
           if (input.url) {
-            const providerFromSchema = firstDatasource?.provider
+            let providerFromSchema = firstDatasource?.provider
             // protocolToConnectorType ensures that the protocol from `input.url` is valid or throws
             // TODO: better error handling with better error message
             // Related https://github.com/prisma/prisma/issues/14732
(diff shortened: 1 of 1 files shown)


packages/migrate/src/commands/DbPull.ts
```diff
@@ -187,6 +187,12 @@ Set composite types introspection depth to 2 levels
             const providerFromUrl = protocolToConnectorType(`${input.url.split(':')[0]}:`)
             const schema = `${this.urlToDatasource(input.url, providerFromSchema)}\n${removeDatasource(rawSchema)}`
 
+            // Normalize provider from schema for PostgreSQL
+            // For PostgreSQL the protocol can also only be `postgres` which will not match the provider from the schema (which is normalized somewhere already)
+            if(providerFromSchema === 'postgres') {
```
jkomyno, line 192:
```suggestion
            if (providerFromSchema === 'postgres') {
```

jkomyno: The above `providerFromSchema` should become a `let` declaration

janpio: Thanks, the dangers of "coding" in Github UI :D

janpio:
*sigh types, I give up. @Jolg42 can take it from here easily I think.

Jolg42:
Done in https://github.com/prisma/prisma/pull/17034

Issue #17300, opened 2023-01-13
Remove `introspection-engine` binary from CLI


jkomyno:
Now that we've stopped using the `introspection-engine` in favor of `migration-engine`, it's time we stop downloading the `introspection-engine` binary.
We should remove support for `introspection-engine` for the same release that will get rid of `prisma-fmt`.

jkomyno:
Closed via https://github.com/prisma/prisma/pull/17301.

PR #17301, opened 2023-01-13
feat: stop downloading introspection-engine

jkomyno:
Contributes to https://github.com/prisma/prisma/issues/17300.

TODO: remove references of `introspection-engine` in the docs.


packages/internals/src/index.ts
```diff
@@ -36,7 +36,6 @@ export { getGenerator, getGenerators } from './get-generators/getGenerators'
 export { getPackedPackage } from './getPackedPackage'
 export { highlightDatamodel, highlightSql, highlightTS } from './highlight/highlight'
 export type { IntrospectionSchemaVersion, IntrospectionWarnings } from './IntrospectionEngine'
```
jkomyno, line 38: We can freely remove these exports, right?

Jolg42: We need to ping the team that consumes the IntrospectionEngine but yes we should remove this along with the code later

Jolg42: Happened in https://prisma-company.slack.com/archives/C01UR43BV6W/p1674224135195389

packages/engines/src/index.ts
```diff
@@ -59,28 +58,6 @@ export { enginesVersion } from '@prisma/engines-version'
...
-path.join(__dirname, '../introspection-engine-debian-openssl-3.0.x')
-
-path.join(__dirname, '../query-engine-rhel-openssl-1.0.x')
-path.join(__dirname, '../introspection-engine-rhel-openssl-1.0.x')
-
-path.join(__dirname, '../query-engine-rhel-openssl-1.1.x')
-path.join(__dirname, '../introspection-engine-rhel-openssl-1.1.x')
-
-path.join(__dirname, '../query-engine-rhel-openssl-3.0.x')
-path.join(__dirname, '../introspection-engine-rhel-openssl-3.0.x')
```
Jolg42, line 83: Not sure why the `query-engine` lines were removed here?
Maybe not needed 🤔 , though we could keep it safe and keep them and figure this out another time?

jkomyno: Thanks for spotting this, I've recovered the mentions to the query-engine

packages/engines/src/index.ts
```diff
@@ -59,28 +58,13 @@ export { enginesVersion } from '@prisma/engines-version'
  */
 
 path.join(__dirname, '../query-engine-darwin')
```
janpio, line 60: OT, but does it make sense to list the non Linux ones here? Isn't this just for deployment platforms that make sense, so mostly Linux?

jkomyno: You may be right. According to `git blame`, these `path.join`s were added to address https://github.com/vercel/nft/issues/104, and Vercel has Linux only machines afaik

janpio: This might be a bigger discussion we should bring to Pierre and others in a new issue, or Slack thread.

Jolg42 (approved):
Nice! ✨

packages/engines/src/index.ts
jkomyno: @janpio @millsp I've created this issue: https://github.com/prisma/prisma-private/issues/207

PR #17312, opened 2023-01-13
feat(fetch-engine): add PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING env property

### LINKED ISSUE #12230: Allow skipping sha checksum
## Problem
Installing Prisma in an Enterprise CI (Jenkins), I encounter the issue that downloading the sha values (checkSum!?) are forbidden by company policy. That is due to a policy were by every download has do go throgh our enterprise reposetory.

## Suggested solution
Add --skipChecksum, that would prevent downloading the sha value.

## Logs
```
> npm --ignore-scripts i
> npx prisma db pull
Error: request to https://binaries.prisma.sh/all_commits/73e60b76d394f8d37d8ebd1f8918c79029f0db86/debian-openssl-1.1.x/migration-engine.gz.sha256 failed, reason: connect ECONNREFUSED 108.156.107.28:443
```

## Alternatives
1) I suppose a [mirror](https://www.prisma.io/docs/concepts/components/prisma-engines#hosting-engines) could work.
But since we are only using Prisma for tests, I haven't got to try and mirror the binaries / checkSum

2) commit all node_modules/*prisma, 
I tried that, people wasn't happy (enterprise monorepo, so it added 330Mb to all).

## Additional context
To be clear, I know the error is for pulling the sha, but I couldn't verify if the binaries were loaded,
Is there a way to confirm it?
[Open Question](https://github.com/prisma/prisma/discussions/12195)
[Requested here as well](https://github.com/prisma/prisma/issues/5105)

spudly:
Fixes #12230


packages/fetch-engine/src/downloadZip.ts
```diff
@@ -17,28 +17,35 @@ const del = promisify(rimraf)
...
+    if (!response.ok) {
+      throw new Error(`Failed to fetch sha256 checksum at ${checksumUrl}. ${response.status} ${response.statusText}`)
+    }
+    const body = await response.text()
+    // We get a string like this:
+    // "3c82ee6cd9fedaec18a5e7cd3fc41f8c6b3dd32575dc13443d96aab4bd018411  query-engine.gz\n"
+    // So we split it by whitespace and just get the hash, as that's what we're interested in
+    const [checksum] = body.split(/\s+/)
+    if (!/^[a-f0-9]{64}$/gi.test(checksum)) {
+      throw new Error(`Unable to parse checksum from response body for ${checksumUrl}`)
```
Jolg42, line 39:
```suggestion
      throw new Error(`Unable to parse checksum from ${checksumUrl} - response body: ${body}`)
```
Would probably help debug if this happens

packages/fetch-engine/src/downloadZip.ts
```diff
@@ -97,12 +104,12 @@ export async function downloadZip(
           const hash = await hashPromise
           const zippedHash = await zippedHashPromise
 
-          if (zippedHash !== zippedSha256) {
-            throw new Error(`sha256 of ${url} (zipped) should be ${zippedSha256} but is ${zippedHash}`)
+          if (zippedSha256 != null && zippedSha256 !== zippedHash) {
+            return reject(new Error(`sha256 of ${url} (zipped) should be ${zippedSha256} but is ${zippedHash}`))
```
Jolg42, line 108:
```suggestion
          if (zippedSha256 !== null && zippedSha256 !== zippedHash) {
            return reject(new Error(`sha256 checksum of ${url} (zipped) should be ${zippedSha256} but is ${zippedHash}`))
```

packages/fetch-engine/src/downloadZip.ts
```diff
@@ -17,28 +17,35 @@ const del = promisify(rimraf)
...
+    if (!/^[a-f0-9]{64}$/gi.test(checksum)) {
+      throw new Error(`Unable to parse checksum from response body for ${checksumUrl}`)
+    }
+    return checksum
+  } catch (error) {
+    if (process.env.PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING) {
+      debug(error)
+      return null
+    }
+    throw error
```
Jolg42, line 47: Actually, this catch would catch these 2 obvious cases after reading the code:
- Can't download the checksum (e.g. Error on the network or 404)
- Download succeeded 
	- but the file content is invalid, we expected a checksum, and it is something else, (e.g. empty)

packages/fetch-engine/src/downloadZip.ts
```diff
@@ -17,28 +17,35 @@ const del = promisify(rimraf)
...
+    // "3c82ee6cd9fedaec18a5e7cd3fc41f8c6b3dd32575dc13443d96aab4bd018411  query-engine.gz\n"
+    // So we split it by whitespace and just get the hash, as that's what we're interested in
+    const [checksum] = body.split(/\s+/)
+    if (!/^[a-f0-9]{64}$/gi.test(checksum)) {
+      throw new Error(`Unable to parse checksum from response body for ${checksumUrl}`)
+    }
+    return checksum
+  } catch (error) {
+    if (process.env.PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING) {
+      debug(error)
```
Jolg42, line 44:
```suggestion
      debug(`fetchChecksum() failed and was ignored as the PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING environment variable is truthy.\nError: ${error}`)

```

Jolg42:
Note: I want to try this locally, but so far it looks amazing!

packages/fetch-engine/src/downloadZip.ts
spudly: correct

spudly:
All builds are passing now. Anything else I can do to move this along? (not trying to be pushy - just want to make sure I'm not the bottleneck)

SirMatGuy:
Hi! Any updates on when the fix will be published?
Thank you very much 💖💖💖

spudly:
Really hoping to see some action taken on this soon. Anything I can do to move it along?

spudly:
Can I get some insight into why this ticket has stalled? I've got some work stuff that depends on it and I'm coming up to some deadlines so I'll need to take other actions if this doesn't get merged and released soon.

I understand if things are crazy and you just don't have time for it; I'd just like a heads up so I can make other plans if needed.

spudly:
@Jolg42 @jkomyno @danstarns, can any of you review this?

janpio:
No they can't. Dan does not work here any more, Joel is on his sabbatical and Alberto is busy with other things right now.

This is still on our list, but not at the top as we are all working on other things right now. Sorry, you'll have to wait a bit longer for us to get to this.

spudly:
Thanks for the update!

packages/fetch-engine/src/downloadZip.ts
```diff
@@ -48,7 +56,7 @@ export async function downloadZip(
 ): Promise<DownloadResult> {
   const tmpDir = tempy.directory()
   const partial = path.join(tmpDir, 'partial')
-  const { sha256, zippedSha256 } = await fetchSha256(url)
+  const [zippedSha256, sha256] = await Promise.all([fetchChecksum(url), fetchChecksum(url.slice(0, url.length - 3))])
```
glenselle, line 59: These keys had their order reversed. Would it make sense to stick `sha256` back as the first item in the array and adjust the right-hand side accordingly?

glenselle:
Our project is also needing this ability due to the security environment in which we deploy. The code looks good. Thanks for doing this @spudly ! Hoping the maintainers get to this soon. We have considered dropping Prisma if we cannot get our deploys working without doing outbound network requests during a `prisma generate`.

erikmellum:
Also in need of this feature. Let us know if there is anything we can do to help.

packages/fetch-engine/src/downloadZip.ts
Jolg42: I don't think it matters here, this code is correct, I don't like the `url.slice` so much but that was there before 😅

packages/fetch-engine/src/downloadZip.ts
```diff
@@ -28,7 +28,11 @@ async function fetchChecksum(url: string): Promise<string | null> {
       agent: getProxyAgent(url) as any,
     })
     if (!response.ok) {
-      throw new Error(`Failed to fetch sha256 checksum at ${checksumUrl}. ${response.status} ${response.statusText}`)
+      let errorMessage = `Failed to fetch sha256 checksum at ${checksumUrl}. ${response.status} ${response.statusText}`
+      if (!process.env.PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING) {
+        errorMessage += `\n\nIf you need to ignore this error (e.g. in an offline environment), set the PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING environment variable to a truthy value.\nExample: PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING=1`
+      }
+      throw new Error(errorMessage)
```
Jolg42, line 35: Note: I thought it would be a good idea to suggest this in the error message to make it discoverable when using Prisma offline.

Jolg42: Note: We might change the logic in the future and make this obsolete but in the meantime this would help.

spudly:
Super excited to see something happening with this PR. Thanks @Jolg42 !

janpio:
```    
env.PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING=1
      √ if checksum downloads and matches, does not throw (407 ms)
      × if checksum downloads but doesn't match, throws (625 ms)
      × if checksum download fails, logs warning but does not throw (626 ms)

  ● download › env.PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING=1 › if checksum downloads but doesn't match, throws

    EPERM: operation not permitted, unlink 'D:\a\prisma\prisma\packages\fetch-engine\src\__tests__\all\query_engine-windows.dll.node'



  ● download › env.PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING=1 › if checksum download fails, logs warning but does not throw

    EPERM: operation not permitted, unlink 'D:\a\prisma\prisma\packages\fetch-engine\src\__tests__\all\query_engine-windows.dll.node'
```
Windows does not like to delete/unlink files that it considers "in use" (via a running process usually).

Jolg42 (approved):

thw0rted:
Thanks @Jolg42 !  As an end user, how can I tell when this has landed in a new release on NPM?

Jolg42:
@thw0rted It will be part of the official `4.16.0` release planned, Tuesday 20th.
It's now possible to try it in our `4.16.0-dev.63` dev version, which is not recommended to use in production.

If you try it, we're curious to hear about any feedback on how it worked out for you.

PR #17315, opened 2023-01-14
fix: change type `generator.config.output` to never as it is a dedicated field.

arthurfiorette:
During some tests at arthurfiorette/prisma-json-types-generator#7 I discovered that the output configuration field never is defined into `generator.config` and it is the correct behavior, but some kind of warning would be great.

Take the below example:

```prisma
generator A {
  name = "A"
  output = "specified"
  opt = "test"
}

generator B {
  name = "B"
  opt = "test"
}
```

```ts
// generator A & B source file

generatorHandler({
  async onGenerate(options) {
    console.log(options.generator)
  }
})

outputOfA = {
  // ...
  output: {
    value: '~/prisma/specified',
    fromEnvVar: null
  },
  config: {
    // note that output is not here.
    opt: 'test'
  }
};

outputOfB = {
  // ...
  output: {
    value: '~/prisma/specified',
    fromEnvVar: null
  },
  config: {
    opt: 'test'
  }
};
```

This PR just add a warn for those who try to access output from config.


arthurfiorette:
Hey @Jolg42, @SevInf & @aqrln, any updates here??

SevInf:
Hey @arthurfiorette.
I apologize for the huge delay. The change makes sense to me, but I think we'll need to go a step further -  `output` is not only the property that can not exist in `config`: `provider`, `binaryTargets` and `previewFeatures` are also on this list ([link](https://github.com/prisma/prisma-engines/blob/ccf1968def472ca29cdd82b022b4a6a57fa4f87a/psl/psl-core/src/validate/generator_loader.rs#L21)).

Would you mind adding them to the PR as well? Totally ok if not, I'll do it myself and merge the PR by the end of the week then. 
Once more, apologies for delay.

packages/generator-helper/src/types.ts
```diff
@@ -28,14 +28,22 @@ export namespace JsonRPC {
...
+    /** `output` is a reserved name and will only be available directly at `generator.output` */
+    output?: never
+    /** `provider` is a reserved name and will only be available directly at `generator.provider` */
+    provider?: never
+    /** `binaryTargets` is a reserved name and will only be available directly at `generator.binaryTargets` */
+    binaryTargets?: never
+    /** `previewFeatures` is a reserved name and will only be available directly at `generator.previewFeatures` */
+    previewFeatures?: never
+    [key: string]: string | string[]
+  }
```
SevInf, line 46:
```suggestion
  config: {
    /** `output` is a reserved name and will only be available directly at `generator.output` */
    output?: never
    /** `provider` is a reserved name and will only be available directly at `generator.provider` */
    provider?: never
    /** `binaryTargets` is a reserved name and will only be available directly at `generator.binaryTargets` */
    binaryTargets?: never
    /** `previewFeatures` is a reserved name and will only be available directly at `generator.previewFeatures` */
    previewFeatures?: never
  } & {
    [key: string]: string | string[]
  }
```

i think we need to separate index signature and fixed proeprties to make TS happy.

SevInf:
Hey @arthurfiorette.
I am sorry again, but [this test failure](https://github.com/prisma/prisma/actions/runs/8332019715/job/22800238746?pr=17315#step:5:3871) looks related to the change. Could you take a look? The code for the test is [here](https://github.com/prisma/prisma/tree/main/packages/client/tests/e2e/generator-config-types).

SevInf:
Nevermind, fixed it myself, ready to merge after CI is green

PR #17322, opened 2023-01-15
fix(get-platform): sed unknown option to "s"

### LINKED ISSUE #17321: When `distro` is undefined, the output of `ldconfig` is ignored because of the `sed: -e expression #1, char 7: unknown option to `s'` silent error
I tested the get-platform which has been merged into main, but there might be one issue.
In that block:
```
const libsslFilename: string | undefined = await getFirstSuccessfulExec([
  'ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl | sort',
  'ls /lib64 | grep ssl',
  'ls /usr/lib64 | grep ssl',
])
```
First set of commands returns
```
sed: -e expression #1, char 7: unknown option to `s'
```
So on a setup with undefined distro, it still goes to the `openssl version -v` fallback

Not familiar with sed though, so I can't debug straight away.

_Originally posted by @Maxime-J in https://github.com/prisma/prisma/issues/14042#issuecomment-1382603550_

jkomyno:
Closes https://github.com/prisma/prisma/issues/17321.

### DIFF
--- a/packages/get-platform/src/getPlatform.ts
+++ b/packages/get-platform/src/getPlatform.ts
@@ -200,7 +200,21 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
 
   debug('Falling back to "ldconfig" and other generic paths')
   const libsslFilename: string | undefined = await getFirstSuccessfulExec([
-    'ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl | sort',
+    /**
+     * The `ldconfig -p` returns the dynamic linker cache paths, where libssl.so files are likely to be included.
+     * Each line looks like this:
+     * 	libssl.so (libc6,hard-float) => /usr/lib/arm-linux-gnueabihf/libssl.so.1.1
+     * But we're only interested in the filename, so we use sed to remove everything before the `=>` separator,
+     * and then we remove the path and keep only the filename.
+     * The second sed commands uses `|` as a separator because the paths may contain `/`, which would result in the
+     * `unknown option to 's'` error (see https://stackoverflow.com/a/9366940/6174476) - which would silently
+     * fail with error code 0.
+     */
+    'ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep ssl | sort',
+
+    /**
+     * Fall back to the rhel-specific paths (although "distro" isn't detected as rhel) when the "ldconfig" command fails.
+     */
     'ls /lib64 | grep ssl',
     'ls /usr/lib64 | grep ssl',
   ])


packages/get-platform/src/getPlatform.ts
```diff
@@ -200,7 +200,21 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
...
+     * The `ldconfig -p` returns the dynamic linker cache paths, where libssl.so files are likely to be included.
+     * Each line looks like this:
+     * 	libssl.so (libc6,hard-float) => /usr/lib/arm-linux-gnueabihf/libssl.so.1.1
+     * But we're only interested in the filename, so we use sed to remove everything before the `=>` separator,
+     * and then we remove the path and keep only the filename.
+     * The second sed commands uses `|` as a separator because the paths may contain `/`, which would result in the
+     * `unknown option to 's'` error (see https://stackoverflow.com/a/9366940/6174476) - which would silently
+     * fail with error code 0.
+     */
+    'ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep ssl | sort',
```
jkomyno, line 213: Here's how to test this:
- Open a debian armv7l shell with `docker run --platform="linux/arm" --rm -ti --entrypoint=/bin/sh node:lts-slim`
- Run the original command: `ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl | sort`
- You should see the message "sed: -e expression #1, char 7: unknown option to `s"
- Run `echo $?` to check the status code, which should be `0`
- Now run the updated command: `ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep ssl | sort`
- The result should be `libssl.so.1.1`

Screenshot 2023-01-15 at 11 51 49

aqrln (approved):

packages/get-platform/src/getPlatform.ts
```diff
@@ -200,7 +200,21 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
...
-    'ldconfig -p | sed "s/.*=>s*//" | sed "s/.*///" | grep ssl | sort',
+    /**
+     * The `ldconfig -p` returns the dynamic linker cache paths, where libssl.so files are likely to be included.
+     * Each line looks like this:
+     * 	libssl.so (libc6,hard-float) => /usr/lib/arm-linux-gnueabihf/libssl.so.1.1
+     * But we're only interested in the filename, so we use sed to remove everything before the `=>` separator,
+     * and then we remove the path and keep only the filename.
+     * The second sed commands uses `|` as a separator because the paths may contain `/`, which would result in the
+     * `unknown option to 's'` error (see https://stackoverflow.com/a/9366940/6174476) - which would silently
+     * fail with error code 0.
```
janpio, line 211: Nice explanation!

Issue #17154, opened 2023-01-05
Fail Prisma installation / warn hard when needed system libraries are missing


janpio:
https://github.com/prisma/prisma/issues/16970 suggest to add a better warning message when needed system libraries are missing. I think we should fail or warn during Prisma installation already,  instead/additionally of failing when Prisma Client tries to execute and then shows a targetted error message.

This would apply for OpenSSL, libz and similar.

jkomyno:
From Slack:

we should print warnings in those scenarios where we know Prisma will likely fail at runtime:
when the distro is Alpine (“musl”) but the arch is not `x86_64`, we should warn users that we don’t support that environment. We should recommend using `docker buildx build --platform linux/amd64` for their Alpine-based Dockerfile instead. We can detect if they’re running on Docker via `ls -la /.dockerenv`, or we can simply assume that each Alpine user is running Alpine on Docker
when `libssl` is undefined in `getPlatform()`, we should warn users that we didn’t find any libssl version installed, so we defaulted to the openssl-1.1.x binaryTargets, but Prisma may not work unless they manually install openssl 1.1.x. If they’re on Debian, we could also tell them that if they’re using Debian Buster or `node:16-slim` or `node:18-buster-slim` Prisma won’t work unless they install `openssl 1.1.x`. We could point them to this Github comment

jkomyno:
Trying to detect the presence of `libz` at build time may be tricky and unreliable. Let's proceed with detecting `libssl` and have `distro / arch`-specific warnings and errors instead.

jkomyno:
Closing this via https://github.com/prisma/prisma/pull/17334.

Issue #8478, opened 2021-07-28
Precompiled engine files for `aarch64-unknown-linux-musl` target (Alpine Linux on ARM, e.g. M1)


pantharshit00:
## Problem

Right now you can't use Prisma on Docker alpine on M1 mac as it requires `aarch64-unknown-linux-musl` binaries. As M1 will see more adoption, people will run into this frequently. 

## Suggested solution
Provide precompiled binaries for `aarch64-unknown-linux-musl` target like we have precompiled binaries for `x86-linux-musl`

## Alternatives
1. Ask user to compile the binary themselves
2. Document use `node:lts` instead of `node:lts-alpine` which is not ideal.

## Additional context
https://github.com/prisma/prisma/issues/7755

Naming as `unknown` is intended like that, see more information internally: https://prisma-company.slack.com/archives/C016KUHB1R6/p1627469574008300?thread_ts=1626873669.001400&cid=C016KUHB1R6

Brooooooklyn:
I'm maintaining Docker images for cross compile: https://github.com/orgs/napi-rs/packages/container/package/napi-rs%2Fnodejs-rust.
The `ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-alpine` is for compile `aarch64-unknown-linux-musl` target, see https://github.com/napi-rs/package-template/blob/main/.github/workflows/CI.yaml#L79 for example.

safareli:
This could be "solved" by using `docker build --platform linux/arm64 ` [see this article](
https://web.archive.org/web/20210527154633/https://blog.jaimyn.dev/how-to-build-multi-architecture-docker-images-on-an-m1-mac/)

jensmeindertsma:
What can I do to help get support for this implemented?

huikaihoo:
Any plan or timeline to support this?

artursudnik:
v3.7.0 has been released, but it seems that this issue is not fixed, is it?

janpio:
The issue is not closed, hence the problem is not solved and stays open here for tracking.

artursudnik:
I see. Thanks. it has been added to the 3.7.0 milestone this is why I am asking.

Josh-a-e:
Yeah, this has become an issue for our engineers using M1s - is there any indication as to whether this item is any sort of priority?

Axxxx0n:
what can we do to help?

muqshots:
Any update on this?

jensmeindertsma:
[...]

PR #17334, opened 2023-01-16
feat(get-platform): fail / warn hard when Prisma can't detect OS details at build time

### LINKED ISSUE #6211: Detect absense of openssl and inform the user about it
## Problem

Right now we don't detect presence of openssl on the CLI but it is dynamically linked dependency for the engines on linux. This can lead to undefined behaviour like CLI trying to download the binaries with no idea about openssl version leading to platform detection like `linux-arm-openssl-undefined` or spawning errors when node tries to spawn query-engine.  

## Suggested solution

Detect openssl presence on the CLI on linux and inform the user about openssl requirement. 

## Alternatives

Statically link/Vendor OpenSSL. This might be actually a requirement for upcoming napi preview. 

## Additional context

I hit this when trying to prisma studio on m1 mac docker which was using the node alpine image.

jkomyno:
Contributes to https://github.com/prisma/prisma/issues/17154.
Closes https://github.com/prisma/prisma/issues/6211.

- Raise informative error when running `prisma` on Linux Alpine on an architecture different than `amd64` (`x86_64`)
- Warn with an informative message when `prisma` can't detect the `libssl` version to use
- Warn with an informative message when `prisma` can't detect the Linux `distro` (less useful, but nice to have)

This PR deprecates https://github.com/prisma/prisma/pull/17151.


packages/get-platform/src/getPlatform.ts
```diff
@@ -313,22 +348,22 @@ export async function getPlatform(): Promise<Platform> {
...
   // if just the distro is known, fallback to latest OpenSSL 1.1
   if (distro) {
-    return (distro + '-openssl-1.1.x') as Platform
+    return `${distro}-openssl-${defaultLibssl}` as Platform
   }
 
   // use the debian build with OpenSSL 1.1 as a last resort
   // TODO: perhaps we should default to 'debian-openssl-3.0.x'
-  return 'debian-openssl-1.1.x'
+  return `${defaultDistro}-openssl-${defaultLibssl}`
```
aqrln, line 366: 1. Unless I'm missing something, these four cases (lines 349-366) can this be simplified to

```ts
return `${distro ?? defaultDistory}-openssl-${libssl ?? defaultLibssl}` as Platform
```

2. Line 350 is the last one where we check that `platform === 'linux'`, but we still assume it afterwards. I feel like we are missing a warning similar to other ones added in this PR that we didn't recognise the operating system and assumed Linux as the default fallback.

packages/get-platform/src/getPlatform.ts
```diff
@@ -255,7 +264,33 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
...
+
+    console.warn(`Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-${defaultLibssl}".
+${additionalMessage}`)
+  }
+
+  // sometimes we fail to detect the distro in use, so we default to debian
+  const defaultDistro = 'debian' as const
+  if (platform === 'linux' && distro === undefined) {
+    console.warn(
+      `Prisma failed to detect the Linux distro in use, and may not work as expected. Defaulting to "${defaultDistro}".`,
```
aqrln, line 291: I think this is a very nice warning, but I can also see how annoying it would be to me if I still used Arch for my Linux system 😄

Can `getPlatform` be called multiple times during the lifetime of a Prisma client? Could we make sure these warnings are only printed once? We have a `warnOnce` utility for that in `@prisma/internals` which you could use or re-implement.

It would generally be preferable to use our existing warnings infrastructure in `@prisma/internals` as that would avoid code duplication, provide consistent styling and respect the `PRISMA_DISABLE_WARNING` env var, but I'm not sure without checking if we can do that without circular dependencies between packages.

packages/get-platform/src/getPlatform.ts
```diff
@@ -255,7 +264,33 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
...
+
+  // sometimes we fail to detect the libssl version to use, so we default to 1.1.x
+  const defaultLibssl = '1.1.x' as const
+  if (platform === 'linux' && libssl === undefined) {
+    /**
+     * Ask the user to install openssl manually, and provide some additional instructions based on the detected Linux distro.
+     */
+    const additionalMessage = match({ distro })
+      .with({ distro: 'debian' }, () => {
+        return "Please manually install OpenSSL via `apt-get update -y && apt-get install -y openssl` and try installing Prisma again. If you're running Prisma on Docker, you may also try to replace your base image with `node:lts-slim`, which already ships with OpenSSL installed."
```
aqrln, line 277: Non-blocking, but something to consider.

Optimally we'd only tell the user to install the `libssl` library and not the whole OpenSSL distribution including the CLI which we don't need. However we don't know if `apt-get install -y libssl1.1` or `apt-get install -y libssl3` should be used on the current system. Do you have any ideas?

Maybe we could say something along the lines of "please run `apt-get install -y libssl1.1` or `apt-get install -y libssl3` depending on your system, or `apt-get install -y openssl` if you are not sure"?

packages/get-platform/src/getPlatform.ts
jkomyno: `getPlatform` is already memoized in `BinaryEngine.ts`. In `LibraryEngine.ts`, it's not, but it's only called once in case of errors

packages/get-platform/src/getPlatform.ts
jkomyno: 1. The whole cascade of `if`s in `getPlatform()` could be refactored / simplified, but I wouldn't risk doing that right now

2. Would you print a warning like `Prisma detected unsupported OS "${platform}, and may not work as expected. Defaulting to "linux".` after the latest `platform === 'linux'` check?

packages/get-platform/src/getPlatform.ts
jkomyno: I'd keep the error message as simple as possible: users that care about minimal Docker images wouldn't use Debian in the first place imho. I think it's fine to ask them to install `openssl`, which automatically downloads a version of `libssl` compatible with the system, rather than us having to figure out whether they should opt for `libssl1.1` or other versions

packages/get-platform/src/getPlatform.ts
aqrln: 2. Basically yes, I'd only slightly reword it: `Prisma detected unknown platform "${platform}", and may not work as expected. Defaulting to "linux".`. Known unsupported OSes should be covered by checks early in the function. I wonder if it should even be a hard error and not a warning.

packages/get-platform/src/getPlatform.ts
aqrln: > getPlatform is already memoized in BinaryEngine.ts. In LibraryEngine.ts, it's not, but it's only called once in case of errors

I'd still not rely on this fact as it could easily be changed accidentally.

packages/get-platform/src/getPlatform.ts
jkomyno: If we made this an error, and it worked with "linux" as a default, it'd be a breaking change in a non-major release

packages/get-platform/src/getPlatform.ts
jkomyno: I would treat this problem separately only if users complain about too many warnings. Our focus is on popular systems also used on Docker, like Alpine and Debian

packages/get-platform/src/getPlatform.ts
```diff
@@ -55,9 +55,18 @@ export async function getos(): Promise<GetOSResult> {
 
   const distro = await resolveDistro()
 
+  if (distro === 'musl' && arch !== 'x64') {
+    // TODO: evaluate adding a pris.ly link to this error message
```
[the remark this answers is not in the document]
janpio: Agree, not needed right now:
```suggestion
```

packages/get-platform/src/getPlatform.ts
```diff
@@ -55,9 +55,18 @@ export async function getos(): Promise<GetOSResult> {
...
+    throw new Error(
+      `Prisma only supports Linux Alpine on the amd64 (x86_64) system architecture. If you're running Prisma on Docker, please use Docker Buildx to simulate the amd64 architecture on your device as explained by this comment: https://github.com/prisma/prisma/issues/8478#issuecomment-1355209706`,
```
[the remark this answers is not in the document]
janpio: Can we remove the concrete bit and just put this in the linked comment maybe?

packages/get-platform/src/getPlatform.ts
```diff
@@ -255,7 +264,33 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
 }
 
 export async function getPlatform(): Promise<Platform> {
-  const { platform, libssl, distro, arch } = await getos()
+  const { platform, distro, arch, libssl } = await getos()
```
janpio, line 267: Is this functionally identical to the old one?

packages/get-platform/src/getPlatform.ts
janpio: Optimally the error/warning message just says what the problem is (libssl could not be detected on the system) and how to fix it (install it) and then link to a docs page with more information (via pris.ly link), that explains the different approaches and so on.

But we can totally do that in a follow up.

packages/get-platform/src/getPlatform.ts
janpio: I am fine with potentially shipping to many warnings for now.

packages/get-platform/src/getPlatform.ts
```diff
@@ -313,22 +348,22 @@ export async function getPlatform(): Promise<Platform> {
...
   }
 
   // if just the distro is known, fallback to latest OpenSSL 1.1
   if (distro) {
-    return (distro + '-openssl-1.1.x') as Platform
+    return `${distro}-openssl-${defaultLibssl}` as Platform
   }
 
   // use the debian build with OpenSSL 1.1 as a last resort
   // TODO: perhaps we should default to 'debian-openssl-3.0.x'
```
janpio, line 365: When do we add `as Platform` and when do we not?

packages/get-platform/src/getPlatform.ts
janpio: Definitely not an error, warning would be fine.

I am ok with current state. Feel free to turn comments into comments or an issue to tackle later.

janpio (approved):

packages/get-platform/src/getPlatform.ts
aqrln: yes

packages/get-platform/src/getPlatform.ts
aqrln: We have to cast it to `Platform` when the inferred type of the expression is `string`. We could make it more type safe by making `distro` a union type of known distros instead of a string, and make the TS compiler check that the built platform name is actually included in the `Platform` type and force us to handle unknown distros specially.

aqrln: I think it's a bit out of scope of this PR though

aqrln (approved):

packages/get-platform/src/getPlatform.ts
aqrln: I'm totally fine with doing it in a follow-up PR, but I'd still prefer to have the consistent warnings with styling and functionality that the logger from `@prisma/internals` provides if possible

packages/get-platform/src/getPlatform.ts
janpio: Yeah, I was just wondering based on some of them changing from old to new.

packages/get-platform/src/getPlatform.ts
aqrln: I'm totally fine with doing it in a follow-up PR, but I'd still prefer to have the consistent warnings with styling and functionality that the logger from `@prisma/internals` provides if it's possible and doesn't introduce circular dependencies.

jkomyno:
Client Dataproxy tests are failing due to flakiness.

packages/get-platform/src/getPlatform.ts
jkomyno: I've copied the functionality needed from `@prisma/internals` - unfortunately it was either that or circular dependencies.

Jolg42:
@jkomyno
```
packages/get-platform build: missingDependencies ["chalk"]
```
Looks like you'll need to add chalk to get-platform

Jolg42 (approved):
Nice ✨ !

Note: add missing dep before merging

packages/get-platform/src/getPlatform.ts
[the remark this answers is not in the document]
jkomyno: I've updated the comment in https://github.com/prisma/prisma/issues/8478#issuecomment-1355209706, and merging this PR as is

kewei5zhang:
Hey @jkomyno, [the error message](https://github.com/prisma/prisma/pull/17334/files#diff-9ad51dcd7849f6fbcc1297541f4aa1584a42192e5cab5e5c9a8bbed95f7fdba6R59) introduced here basically sentenced all walkaround following [this thread](https://github.com/prisma/prisma/issues/8478#issuecomment-1059559520) to run Prisma on alpine arm64 to death. I believe the error message will be helpful, but maybe we shouldn't stop people from running Prisma on alpine arm64 if they managed to do so?

PR #17346, opened 2023-01-16
chore: revive binary size benchmark

Weakky:
## Overview

Attempts to revive the binary size benchmark

### DIFF
--- a/packages/client/src/__tests__/benchmarks/huge-schema/huge-schema.bench.ts
+++ b/packages/client/src/__tests__/benchmarks/huge-schema/huge-schema.bench.ts
@@ -7,6 +7,7 @@ import path from 'path'
 
 import { compileFile } from '../../../utils/compileFile'
 import { generateTestClient } from '../../../utils/getTestClient'
+import { renderSchema } from './builder'
 
 const suite = new Benchmark.Suite('typescript')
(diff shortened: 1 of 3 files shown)


packages/client/src/__tests__/benchmarks/huge-schema/huge-schema.bench.ts
```diff
@@ -7,6 +7,7 @@ import path from 'path'
 
 import { compileFile } from '../../../utils/compileFile'
 import { generateTestClient } from '../../../utils/getTestClient'
+import { renderSchema } from './builder'
 
 const suite = new Benchmark.Suite('typescript')
 // @ts-ignore
```
millsp, line 13: Could we put this into a factory function and have a bench file for binary and library?

Weakky: The problem is that the whole benchmark is a hack already. We're utilizing a benchmark library to compute file size information. It's not designed to do that. It's designed to compute function performance in ms.

Meaning, if we split it into multiple benchmarks, we'll have to come up with a random benchmark of a different schema (using the binary), and then, on completion, we'll be able to additionally compute the size of the binary.

PR #17347, opened 2023-01-16
fix(client): missing relations for includes in client extensions

### LINKED ISSUE #17349: Deeply nested objects aren't typed when clientExtensions is enabled
### Bug description

Originally, I ran into the `RangeError: Maximum call stack size exceeded` error when trying to build my application. Using prisma version `4.9.0-integration-feat-client-ext-generic-type-api.13` as mentioned in #16600 resolved that issue. However, I am now running into an issue where deeply nested objects aren't included in my object types. 

### How to reproduce

Using the schema and code listed below, run pnpm run build. The following error is logged to the console

```
Type error: Property 'item' does not exist on type '{ id: string;  name: string; plaidItemId: string; }'.
```

### Expected behavior

_No response_

### Prisma information

```prisma
model PlaidInstitution {
  id                          String      @id
  name                        String
  logoUrl                     String?     @map("logo_url")
  items                       PlaidItem[]
  @@map("plaid_institutions")
}

model PlaidItem {
  id                          String      @id
  institutionId               String      @map("institution_id")
disabledAt                  DateTime?   @map("disabled_at")
  institution     PlaidInstitution        @relation(fields: [institutionId], references: [id], onDelete: Restrict)

  accounts                    PlaidAccount[]

  @@map("plaid_items")
  @@index([institutionId])
}

model PlaidAccount {
  id                          String      @id
  plaidItemId                 String      @map("plaid_item_id")
  name                        String
  destinations                Destination[] 
  item                        PlaidItem   @relation(fields: [plaidItemId], references: [id], onDelete: Cascade)
  @@map("plaid_accounts")
  @@index([plaidItemId])
}

model Destination {
  id                          String      @id @default(uuid())
  integration                 String
  name                        String
  accounts                    PlaidAccount[]
  
  @@map("destinations")
}
```

```ts
const destination = await db.destination.findFirst({ 
    include: { 
      accounts: { where: { item: { disabledAt: null } }, include: { item: { include: { institution: true }} }},
    }
  });
```

### Environment & setup

- OS: macOS
- Database: PlanetScale
- Node.js version: v16.17.0

### Prisma Version

```
10.1.0
```

millsp:


packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -34,19 +34,19 @@ type GetFindResult<P extends Payload, A> =
   ? {
       [K in keyof S as S[K] extends false | undefined | null ? never : K]:
         S[K] extends true
-        ? P extends { objects: { [k in K]: (infer O extends Payload)[] } }
```
millsp, line 37: Because we need to support many ts versions, this is invalid in older versions.

danstarns (commented):
Is this related to an issue and is it possible to add a regression for it?

danstarns (approved):

PR #17352, opened 2023-01-16
feat(github): /create-local-pr command

DavidHancu:
# Description

`/create-local-pr` is a command that supercharges Pull Request operations across Prisma repositories.

# Usage

This command can only be used by people with `write` permissions for the repository where the command has been used. This command works *exclusively* in Pull Requests. To use it, post a comment with the following content:
```/create-local-pr```

This will create a local branch for the Pull Request and a draft Pull Request for said branch, while posting a comment with a link to the draft Pull Request in the original Issue.

# Current State

Currently, this Pull Request only implements the `/create-local-pr` command and is meant to be a showcase of the full workflow. @janpio's comment:

> we'll have to review this anyway to see how it works etc before we can think about merging etc.

`/create-local-pr` - Implemented.
`/sync-local-pr` - Not Implemented, will come in a later Pull Request.

# Considerations

- This workflow uses a personal access token from a secret named `PAT`. This has to be changed so it matches the correct secret naming from this repository. My suggestion was to use @prisma-bot for this.
- This workflow uses `local_` as the prefix for local branches, which doesn't trigger integration tests. The prefix is configurable and needs to be discussed before merging this Pull Request.

# Testing

To facilitate the testing of this workflow, I've created a test Pull Request on a new repository.
https://github.com/DavidHancu/local-pr-test/pull/1

A list of all tests ran can be found here:
https://github.com/DavidHancu/local-pr-test

Please keep in mind that the https://github.com/DavidHancu/local-pr-test repository uses my own PAT, so the messages will appear posted by me.

### CODE REVIEW

File: .github/workflows/local-pr.yaml [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
+    needs: [pr_commented]
+    steps:
+      # Check for write permission
+      - name: Check write permission
+        id: check
+        uses: actions-cool/check-user-permission@v2
+        with:
+          require: write
+        env:
+          GITHUB_TOKEN: ${{ secrets.pat }}
```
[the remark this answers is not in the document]
DavidHancu: Yes.

.github/workflows/local-pr.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+  check_user_permission:
+    runs-on: ubuntu-latest
+    name: Check permissions
+    needs: [pr_commented]
+    steps:
+      # Check for write permission
+      - name: Check write permission
+        id: check
+        uses: actions-cool/check-user-permission@v2
```
janpio, line 25: We probably want to review this for a) what it does with the token and b) if it really checks for write permissions properly on the repo for the supplied (or currently default) user.

.github/workflows/local-pr.yaml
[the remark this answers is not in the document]
DavidHancu: Under the hood, it uses `context.actor`, which is the user that triggered this workflow to run. (in this case, it is the user that left a comment)

.github/workflows/local-pr.yaml
```diff
@@ -0,0 +1,171 @@
...
+run-name: "Local PR requested for #${{ github.event.issue.number }} by @${{ github.actor }}"
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  pr_commented:
+    name: Identify comment source
+    if: ${{ github.event.issue.pull_request }} || ${{ github.event.issue.state == 'open' }}
```
janpio, line 11: Is `github.event.issue.state == 'open'` the same as "source = PR"?
Or should this be `&&` instead of `||`?

DavidHancu: This should be `&&` instead.

.github/workflows/local-pr.yaml
DavidHancu: It is not required to supply a PAT for this action.

DavidHancu:
I've prepared a repository that can be used to test this action. I've added the link to a test Pull Request in the original message.

.github/workflows/local-pr.yaml
DavidHancu: I've removed the PAT from the permission check and added a test to the testing repository for correct permission checking.

wmadden:
Hi @DavidHancu,

Thank you for your contribution and the effort you’ve put into this PR. After reviewing it, we’re not quite sure what the intended use case for the /create-local-pr command is or how it fits into our current workflows. Since there hasn’t been any recent activity on this PR and we don’t have a clear path forward for it, we’re going to close it for now.

If you’d like to provide more context on the motivation behind this feature, feel free to open a discussion or submit a new PR with additional details. We really appreciate your time and effort!

PR #17353, opened 2023-01-16
test(client): Views

danstarns:
Closes: https://github.com/prisma/client-planning/issues/213

Related: 

* https://github.com/prisma/docs/pull/4364
* https://github.com/prisma/prisma-engines/pull/3581


packages/client/tests/functional/views/tests.ts
```diff
@@ -0,0 +1,94 @@
...
+    test('should query a view with where', async () => {
+      const user = await prisma.userInfo.findMany({
+        where: {
+          email: fakeUser.email,
+        },
+      })
+
+      expect(user).toBeDefined()
+
+      expect(user.id).toEqual(fakeUser.id)
```
janpio, line 47:
```suggestion
      const users = await prisma.userInfo.findMany({
        where: {
          email: fakeUser.email,
        },
      })

      expect(users[0]).toBeDefined()

      expect(users[0].id).toEqual(fakeUser.id)
```

packages/client/tests/functional/views/tests.ts
```diff
@@ -0,0 +1,94 @@
...
+            },
+            { $unwind: '$bio' },
+          ])
+        `
+      } else {
+        return `
+          CREATE VIEW "UserInfo" 
+          AS SELECT u.id, email, name, p.bio
+          FROM "User" u
+          LEFT JOIN "Profile" p ON u.id = p."userId"
```
janpio, line 89: This might need to be adapted for MySQL, see Slack.

packages/client/tests/functional/views/tests.ts
```diff
@@ -0,0 +1,94 @@
...
+            {
+              $project: {
+                _id: 1,
+                email: 1,
+                name: 1,
+                bio: '$ProfileData.bio',
+              },
+            },
+            { $unwind: '$bio' },
+          ])
```
janpio, line 82: Not sure if this runs in the correct database, see Slack.
Running it manually in the correct database makes the tests pass.
(Note: I ran `db push` once before creating the view - no idea if this makes a difference in MongoDB)

packages/client/tests/functional/views/tests.ts
janpio: @Weakky is doing something similar in the Engines tests: https://github.com/prisma/prisma-engines/pull/3581/files#diff-22acb9f8d5a6ffc5bc3110715c576543886376b5e1b3c929fa2059701a2f327dR141-R154

packages/client/tests/functional/views/tests.ts
danstarns: Works with `runRawCommand`.

packages/client/tests/functional/views/tests.ts
```diff
@@ -0,0 +1,108 @@
...
+        data: {
+          ...fakeUser,
+          profile: {
+            create: fakeProfile,
+          },
+        },
+      })
+
+      if (suiteConfig.provider === 'mongodb') {
+        // alterstament dont work with mongodb because
```
janpio, line 33:
```suggestion
        // alterStatement doesn't work with mongodb because
```

janpio (approved):

pimeys (approved):
Looks good to me. In the future we might want to test mutations too, but I think that can also just happen in the query engine tests.

PR #17375, opened 2023-01-17
chore: revive binary size benchmark

Weakky:
## Overview

Attempts to revive the binary size benchmark.


Jolg42:
Is this still needed @Weakky ?

PR #17394, opened 2023-01-18
feat: use `grep libssl` rather than `grep ssl`

jkomyno:
As discovered in [here](https://github.com/prisma/ecosystem-tests/pull/3385#issue-1537911837):
- `ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep ssl | sort` may return files containing `ssl` that confuse our `libssl` detection logic. E.g., we may pick up files like `libevent_openssl-2.1.so` or `libgnutls-openssl.so.27`, which would cause `libsslVersion` to be `undefined` and `openssl version -v` to be used as a fallback (which may fail as well, when the `openssl` binary isn't installed in the system)
- `ls ... | grep ssl` suffers from similar problems

This PR replaces `grep ssl` with `grep libssl` in `getPlatform.ts`.

### DIFF
--- a/packages/get-platform/src/getPlatform.ts
+++ b/packages/get-platform/src/getPlatform.ts
@@ -219,13 +219,13 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
      * `unknown option to 's'` error (see https://stackoverflow.com/a/9366940/6174476) - which would silently
      * fail with error code 0.
      */
-    'ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep ssl | sort',
+    'ldconfig -p | sed "s/.*=>s*//" | sed "s|.*/||" | grep libssl | sort',
 
     /**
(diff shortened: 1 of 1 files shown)


jkomyno:
Integration PR: https://github.com/prisma/prisma/pull/17395

jkomyno:
Ecosystem tests is green: https://github.com/prisma/ecosystem-tests/actions/runs/3949306583

(except `codesandbox` failing due to timeouts)

Jolg42 (approved):
I checked some logs from the ecosystem-tests run and indeed it looks good 🙌🏼

Jolg42:
Note we were already looking for libssl in
- https://github.com/prisma/prisma/blob/29fb2a8e5d9b358bd2b79e3fcbb9aa0e3297beaa/packages/get-platform/src/getPlatform.ts#L137
- https://github.com/prisma/prisma/blob/29fb2a8e5d9b358bd2b79e3fcbb9aa0e3297beaa/packages/get-platform/src/getPlatform.ts#L198

PR #17429, opened 2023-01-19
refactor: Remove `beforeRequest` hook and its friends

SevInf:
`beforeRequest` hook was an internal API that got used by the studio 2 years ago. It was never made public and eventually got replaced by middlewares. It is one of the 2 places that could potentially expose GraphQL query and would need changing for JSON protocol.

While removing it I also stumbled upon a couple of generated types that were 1. Not exposed to the public. 2. Did not correspond to current runtime types. So I removed them as well.

[Internal Slack thread](https://prisma-company.slack.com/archives/C04KV4HMEE5/p1674149575396939)


millsp (approved):
Thanks, this was itching me for a long time 😄

Jolg42 (approved):
🧹🧽✨

PR #17443, opened 2023-01-20
feat: unify engine checks between `@prisma/engine-core` and `@prisma/internals`

jkomyno:
We currently handle prisma engines loading errors both in `@prisma/engine-core`, used mostly by the Client, and in `@prisma/internals`, used by every Prisma package. This results in error messages involving the Prisma engines that are inconsistent depending on whether somebody is using the Prisma Client API or the CLI.

## Notes

- `@prisma/internals` imports `@prisma/engine-core`
- Currently, the logic with most "prisma engines loading error" handling (where more precise errors should be added) resides in `@prisma/internals`, in `engine-commands/queryEngineCommons.ts`
- `getConfig` and `getDmmf` are mostly used from the definitions in `@prisma/internals`, but are also exposed by `LibraryEngine` in `@prisma/engine-core`
- `@prisma/engine-core` apparently uses `getConfig` from the `query-engine` (although it has been removed from `libquery`), rather than the `wasm` version defined in `@prisma/prisma-fmt-wasm`


packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -93,6 +93,8 @@ export class LibraryEngine extends Engine {
     exitHooks.setListener(this, listener)
   }
 
+  // TODO: is there any example of someone calling this constructor without a custom library loader?
```
jkomyno, line 96: Perhaps only in [`LibraryEngine.test.ts`](https://github.com/prisma/prisma/blob/e50510f05d7a4df6ee4b28e76753463fa3759347/packages/engine-core/src/__tests__/LibraryEngine.test.ts)

PR #17444, opened 2023-01-20
refactor: Remove engine middleware

SevInf:
Was never a public API and as far as we are aware, it is not used
internally or externally. It exposes GraphQL query to the users directly
and would need to be adjusted for protocol work. Since we are pretty
confident that it's not used, we are opting for removal instead.

[Internal Slack](https://prisma-company.slack.com/archives/C04KV4HMEE5/p1674209080035079?thread_ts=1674202779.830179&cid=C04KV4HMEE5)

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -41,8 +41,8 @@ import { BaseDMMFHelper, DMMFHelper } from './dmmf'
 import type { DMMF } from './dmmf-types'
 import { getLogLevel } from './getLogLevel'
 import { mergeBy } from './mergeBy'
-import type { EngineMiddleware, Namespace, QueryMiddleware, QueryMiddlewareParams } from './MiddlewareHandler'
-import { Middlewares } from './MiddlewareHandler'
+import type { QueryMiddleware, QueryMiddlewareParams } from './MiddlewareHandler'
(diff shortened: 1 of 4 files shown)


millsp (approved):
Types were never exposed so I also think this is safe to remove :+1:

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -493,20 +493,8 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
-        this._middlewares.query.use(arg0 as QueryMiddleware)
-      } else if (arg0 === 'all') {
-        this._middlewares.query.use(arg1 as QueryMiddleware)
-      } else if (arg0 === 'engine') {
-        this._middlewares.engine.use(arg1 as EngineMiddleware)
-      } else {
-        throw new Error(`Invalid middleware ${arg0}`)
-      }
+    $use<T>(middleware: QueryMiddleware<T>) {
+      this._middlewares.use(middleware as QueryMiddleware<unknown>)
```
Jolg42, line 497: Quick read: just wondering if adding `<unknown>` changes something for our users?

Jolg42 (approved):

packages/client/src/runtime/getPrismaClient.ts
SevInf: I don't think so, users use generated types. Will change it in a sec

PR #17475, opened 2023-01-23
refactor: Abstraction over engine protocol

SevInf:
Pulls everything GraphQL-related into
`GraphQLProtocolEncoder`/`GraphQLMessage` classes. Intorudces generic
interfaces around them, so protocol could be later swapped for JSON in
runtime without the need to change `RequestHander`/`PrismaClient` code
that much.

Batching unit test, that was heavily dependent on raw GraphQL input got
converted to functional test that uses only public API.

Close prisma/client-planning#243


packages/client/src/runtime/core/protocol/common.ts
```diff
@@ -0,0 +1,27 @@
...
+  isWrite(): boolean
+  getBatchId(): string | undefined
+  toDebugString(): string
+  toEngineQuery(): EngineQuery
+  deserializeResponse(data: unknown, dataPath: string[]): unknown
+}
+
+export type CreateMessageOptions = {
+  action: Action
+  model?: string
```
millsp, line 18: For consistency, could it be `modelName`, or rename `modelName` to model?

packages/client/src/runtime/core/protocol/graphql.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    const field = this.dmmf?.rootFieldMap[rootField!]
+
+    if (field === undefined) {
+      throw new Error(
+        `Could not find rootField ${rootField} for action ${action} for model ${model} on rootType ${operation}`,
+      )
+    }
+
+    // TODO: figure out if can be done in PrismaClient iteself
```
millsp, line 65: Is the model name enough to make this work for both implementations?

packages/client/tests/functional/batching/tests.ts
```diff
@@ -0,0 +1,75 @@
...
+      })
+    })
+
+    test('does not batch different models', async () => {
+      await Promise.all([
+        prisma.user.findUnique({ where: { id: id1 } }),
+        prisma.post.findUnique({ where: { id: id2 } }),
+      ])
+
+      // binary engine can retry requests sometimes, that's why we
```
millsp, line 48: That's unexpected for me in the context of tests, but maybe I'm missing something. Do you know why it needs to retry requests?

packages/engine-core/src/common/Engine.ts
```diff
@@ -25,16 +25,20 @@ export type BatchTransactionOptions = {
 
 export type InteractiveTransactionOptions<Payload> = Transaction.Info<Payload>
 
-export type RequestOptions<InteractiveTransactionPayload> = {
+export type GraphQLQuery = {
```
millsp, line 28: Depending on what you do next, maybe a discriminant property will bee needed here

packages/client/src/runtime/core/protocol/graphql.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  toEngineQuery(): EngineQuery {
+    return { query: String(this.document) }
+  }
+
+  deserializeResponse(data: unknown, dataPath: string[]): unknown {
+    const rootField = this.getRootField()
+
+    // TODO: figure out what it is
```
millsp, line 125: This basically "lifts" the value of a result at a given path. And `select` and `include` are removed here for the fluent API. This is old code that I wanted to remove when I refactored the API to use proxies but never got to it.

packages/client/src/runtime/core/protocol/graphql.ts
millsp: I think this needs to be resolved before you merge?

packages/client/src/runtime/core/protocol/common.ts
SevInf: Done

packages/client/src/runtime/core/protocol/graphql.ts
SevInf: Oh. actually, I already did this even before I opened PR, just forgot to remove this comment.

packages/engine-core/src/common/Engine.ts
SevInf: Haven't decided on that yet. But if it will be necessary, I'll add it along with JSON protocol.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1177,19 +1116,17 @@ new PrismaClient({
...
-        debug(query + '\n')
+        debug(message.toDebugString() + '\n')
       }
 
       await lock /** @see {@link getLockCountPromise} */
 
       return this._fetcher.request({
-        document,
+        protocolMessage: message,
+        modelName: model,
```
millsp, line 1126: Not sure if the rename makes sense for the request handler too

millsp (approved):
👏

packages/client/tests/functional/batching/tests.ts
millsp: It would be interesting to know what happens in the query engine/db some day

packages/client/src/runtime/getPrismaClient.ts
SevInf: There, `model` name comes from middleware parameters, so I am not quite sure we can rename it that easily

PR #17476, opened 2023-01-23
feat(get-platform): use original distro and distro family in warning messages

### LINKED ISSUE #17480: internal: use original distro and distro family in warning messages
Currently, in `@prisma/get-platform`, we're only keeping track of `distro`, which is the target Linux """distro""" that would be used by the binary targets downloaded by Prisma on initialisation (e.g, `debian`, which would be used with Debian, Ubuntu, Linux Mint, and as a default whenever we can't detect the Linux distro).

Additional information, like the original distro (read from the `ID` field of `/etc/os-release`) and the distro family (e.g., the most representative entry of the `ID_LIKE` field of `/etc/os-release`, which can be used to suggest running commands with a specific package manager).

jkomyno:
Closes https://github.com/prisma/prisma/issues/17480.

This PR:
- [x] renames `distro` into `targetDistro`
- [x] for each Linux distro, keep also track of `originalDistro` (e.g., `ubuntu`, `arch`) and `familyDistro` (e.g., `debian`), which are used to provide more precise warning and log messages

```typescript
export type DistroInfo = {
  /**
   * The original distro is the Linux distro name detected via its release file.
   * E.g., on Arch Linux, the original distro is `arch`. On Linux Alpine, the original distro is `alpine`.
   */
  originalDistro?: string

  /**
   * The family distro is the Linux distro name that is used to determine Linux flavors based on the same base distro, and likely using the same package manager.
   * E.g., both Ubuntu and Debian belong to the `debian` family of distros, and thus rely on the same package manager (`apt`).
   */
  familyDistro?: string

  /**
   * The target distro is the Linux distro associated with the Prisma Engines.
   * E.g., on Arch Linux, Debian, and Ubuntu, the target distro is `debian`. On Linux Alpine, the target distro is `musl`.
   */
  targetDistro?: 'rhel' | 'debian' | 'musl' | 'arm' | 'nixos' | 'freebsd11' | 'freebsd12' | 'freebsd13'
}
```


packages/get-platform/src/getPlatform.ts
```diff
@@ -289,10 +390,11 @@ ${additionalMessage}`,
...
   const defaultDistro = 'debian' as const
-  if (platform === 'linux' && distro === undefined) {
+  if (platform === 'linux' && targetDistro === undefined) {
     warnOnce(
       'distro:undefined',
-      `Prisma failed to detect the Linux distro in use, and may not work as expected. Defaulting to "${defaultDistro}".`,
+      `Prisma doesn't know which engines to download for the Linux distro "${originalDistro}". Falling back to Prisma engines built for "${defaultDistro}". While these engines are likely to work on your system, Prisma may not work as expected.
+Please run \`prisma generate\` and \`prisma db push\` to chat that Prisma is working correctly on your system.`,
     )
   }
```
jkomyno, line 399: Both `originalDistro` and `targetDistro` are used in this warning message, which also now clarifies that Prisma will likely still work. The warning is now "softer".

packages/get-platform/src/getPlatform.ts
```diff
@@ -264,16 +365,16 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
...
     /**
-     * Ask the user to install openssl manually, and provide some additional instructions based on the detected Linux distro.
+     * Ask the user to install openssl manually, and provide some additional instructions based on the detected Linux distro family.
      */
-    const additionalMessage = match({ distro })
-      .with({ distro: 'debian' }, () => {
+    const additionalMessage = match({ familyDistro })
+      .with({ familyDistro: 'debian' }, () => {
         return "Please manually install OpenSSL via `apt-get update -y && apt-get install -y openssl` and try installing Prisma again. If you're running Prisma on Docker, you may also try to replace your base image with `node:lts-slim`, which already ships with OpenSSL installed."
       })
```
jkomyno, line 379: `familyDistro` is used in this warning so that we can tell the user which package manager to use when installing e.g. `openssl`

packages/get-platform/src/getPlatform.ts
```diff
@@ -54,64 +70,149 @@ export async function getos(): Promise<GetOSResult> {
...
+  try {
+    const osReleaseInput = await readFile(osReleaseFile, { encoding: 'utf-8' })
+    return parseDistro(osReleaseInput)
+  } catch (_) {
+    return {
+      targetDistro: undefined,
+      familyDistro: undefined,
+      originalDistro: undefined,
+    }
   }
```
jkomyno, line 216: No reason to check `/etc/alpine-release` separately, which only returns the version ID of Alpine, similarly to `/etc/lsb-release` on Centos.
Even in old Alpine version (e.g., the one returned by `node:12-alpine`), `/etc/os-release` exists, and it contains `ID=alpine`.

packages/get-platform/src/__tests__/parseDistro.test.ts
```diff
@@ -119,16 +179,29 @@ PRETTY_NAME="Linux Mint 18.2"
...
-    test(t.name, () => {
-      const actual = parseDistro(t.content)
-      expect(actual).toBe(t.expect)
-    })
-  }
+  test.each(tests)('$name', (t) => {
+    const actual = parseDistro(t.content)
+    expect(actual).toMatchObject(t.expect)
+  })
 })
```
jkomyno, line 207: Copied this test structure from `getSSLVersion.test.ts`

packages/get-platform/src/__tests__/parseDistro.test.ts
```diff
@@ -1,35 +1,48 @@
...
-  {
-    name: 'simple with quotes',
-    content: `
+      `,
+      expect: {
+        targetDistro: 'debian',
+        familyDistro: 'debian',
+        originalDistro: 'debian',
+      },
+    },
```
jkomyno, line 15: The tests are essentially the same one a before, with an extended returned value and additional test cases (e.g., `unknown` as an `orginalDistro`)

packages/get-platform/src/__tests__/parseDistro.test.ts
```diff
@@ -93,24 +133,44 @@ PLATFORM_ID="platform:f31"
...
 LOGO=archlinux
-  `,
-    expect: undefined, // this will default to debian later.
-  },
-  {
-    name: 'linux mint',
-    content: `
+      `,
+      expect: {
+        targetDistro: 'debian',
```
millsp, line 166: Is this correct?

packages/get-platform/src/getPlatform.ts
```diff
@@ -289,10 +390,11 @@ ${additionalMessage}`,
 
   // sometimes we fail to detect the distro in use, so we default to debian
   const defaultDistro = 'debian' as const
-  if (platform === 'linux' && distro === undefined) {
+  if (platform === 'linux' && targetDistro === undefined) {
     warnOnce(
       'distro:undefined',
-      `Prisma failed to detect the Linux distro in use, and may not work as expected. Defaulting to "${defaultDistro}".`,
+      `Prisma doesn't know which engines to download for the Linux distro "${originalDistro}". Falling back to Prisma engines built for "${defaultDistro}". While these engines are likely to work on your system, Prisma may not work as expected.
+Please run \`prisma generate\` and \`prisma db push\` to chat that Prisma is working correctly on your system.`,
```
janpio, line 397: "chat"?

janpio: These commands are really bad ideas to just run as well. Especially `db push` can completely mess with a database.

This will be very common output for some users (all on not explicitly supported Linux distros).
We should make it shorter, not even longer and more complex.

packages/get-platform/src/getPlatform.ts
janpio: How about we make it a _lot_ shorter and jsut add a pris.ly link to some documentation that explains what this means, and how to test if things work?

This will be very common output for some users (all on not explicitly supported Linux distros). We should make it shorter, not even longer and more complex.

packages/get-platform/src/getPlatform.ts
jkomyno: Whelp, "chat" was supposed to be "check"

packages/get-platform/src/__tests__/parseDistro.test.ts
jkomyno: Yes, working as intended

packages/get-platform/src/getPlatform.ts
```diff
@@ -18,16 +17,33 @@ const supportedLibSSLVersions = ['1.0.x', '1.1.x', '3.0.x'] as const
...
+export type DistroInfo = {
+  /**
+   * The original distro is the Linux distro name detected via its release file.
+   * E.g., on Arch Linux, the original distro is `arch`. On Linux Alpine, the original distro is `alpine`.
+   */
+  originalDistro?: string
+
+  /**
+   * The family distro is the Linux distro name that is used to determine Linux flavors based on the same base distro, and likely using the same package manager.
+   * E.g., both Ubuntu and Debian belong to the `debian` family of distros, and thus rely on the same package manager (`apt`).
```
millsp, line 29: This statement isn't true for arch linux I think it uses pacman and not apt by default?

packages/get-platform/src/getPlatform.ts
```diff
@@ -18,16 +17,33 @@ const supportedLibSSLVersions = ['1.0.x', '1.1.x', '3.0.x'] as const
...
   platform: NodeJS.Platform
   arch: Arch
-  distro?: 'rhel' | 'debian' | 'musl' | 'arm' | 'nixos' | 'freebsd11' | 'freebsd12' | 'freebsd13'
-
   /**
    * Starting from version 3.0, OpenSSL is basically adopting semver, and will be API and ABI compatible within a major version.
    */
   libssl?: typeof supportedLibSSLVersions[number]
-}
+} & Partial<DistroInfo>
```
millsp, line 46: DistroInfo is already partial, so I think you can remove this

millsp: Also, I could suggest making a `distro: DistroInfo` entry, but not a hard requirement. The you would do `distro.family` instead of `familyDistro`.

packages/get-platform/src/getPlatform.ts
```diff
@@ -54,64 +70,149 @@ export async function getos(): Promise<GetOSResult> {
...
+          targetDistro: 'debian',
+          familyDistro: 'arch',
+          originalDistro,
+        } as const
+      }
 
-  if (idLike.includes('debian') || idLike.includes('ubuntu') || id === 'debian') {
-    return 'debian'
-  }
+      if (idLike.includes('centos') || idLike.includes('fedora') || idLike.includes('rhel')) {
```
millsp, line 183: It looks like this case is handled above, do you think we can remove this?

packages/get-platform/src/getPlatform.ts
```diff
@@ -54,64 +70,149 @@ export async function getos(): Promise<GetOSResult> {
...
+          targetDistro: 'debian',
+          familyDistro: 'debian',
+          originalDistro,
+        } as const
+      }
 
-  if (idLike.includes('centos') || idLike.includes('fedora') || idLike.includes('rhel') || id === 'fedora') {
-    return 'rhel'
-  }
+      if (id === 'arch' || idLike.includes('arch')) {
```
millsp, line 175: Can we make this part of the new matching logic? And why do we need to check `idLike` instead of doing a broad matching?

packages/get-platform/src/getPlatform.ts
jkomyno: Above we're only handling the values after `ID=` in `/etc/os-release`.
The `idLike` variables refers to the values after `ID_LIKE=` (if any), so they are different cases that we should not remove.

packages/get-platform/src/getPlatform.ts
millsp: I agree to have a link to the docs as it can be a nice place to store a specific how-to/tutorial (eg. building your own engines, environment variables, etc...).

packages/get-platform/src/getPlatform.ts
```diff
@@ -54,64 +70,149 @@ export async function getos(): Promise<GetOSResult> {
...
+        return {
+          targetDistro: 'rhel',
+          familyDistro: 'rhel',
+          originalDistro,
+        } as const
+      }
+
+      /* Generic distro info fallback */
+      return {
+        targetDistro: undefined,
```
millsp, line 196: Personal preference: I would prefer that 'debian' is explicitly the default here because it's not obvious reading the code. Totally optional, as it is not the point of your PR.

jkomyno: The default (`debian`) is explicitly spelled out in `getPlatform` [here](https://github.com/prisma/prisma/blob/e697f0b099c59b0e313ee15aa404fe9730718691/packages/get-platform/src/getPlatform.ts#L394-L402), where also custom warnings are defined when `targetDistro === undefined`

packages/get-platform/src/getPlatform.ts
jkomyno: Done

packages/get-platform/src/getPlatform.ts
jkomyno: I'm reverting the change to the warning's wording. We can change that in a separate PR, so that we can at least merge the code for the rest of this PR

packages/get-platform/src/getPlatform.ts
jkomyno: That'd be a nice to have, but as you say, no need to do it until it's needed. Let's merge this PR first

millsp: `Partial` isn't needed though

packages/get-platform/src/getPlatform.ts
```diff
@@ -54,64 +70,152 @@ export async function getos(): Promise<GetOSResult> {
...
+   * Red Hat      => ID=rhel, ID_LIKE=fedora                       => targetDistro=rhel, familyDistro=rhel
+   * Centos       => ID=centos, ID_LIKE=rhel fedora                => targetDistro=rhel, familyDistro=rhel
+   * Alma Linux   => ID="almalinux", ID_LIKE="rhel centos fedora"  => targetDistro=rhel, familyDistro=rhel
+   * Fedora       => ID=fedora                                     => targetDistro=rhel, familyDistro=rhel
+   */
 
-  if (idLike.includes('debian') || idLike.includes('ubuntu') || id === 'debian') {
-    return 'debian'
-  }
+  const distroInfo = match({ id, idLike })
```
millsp, line 117: I like this, while being very long it is very easy to read and understand

millsp (approved):
👏

packages/get-platform/src/getPlatform.ts
jkomyno: Right, removed.

PR #17517, opened 2023-01-24
feat: remove deprecated "getConfig" wrappers for libquery and query-engine

jkomyno:
Random find of outdated code while on Zoom call with @millsp.
This removal would have happened during this sprint anyway, as part of the Wasm-ification project.

### DIFF
--- a/packages/internals/src/engine-commands/getConfig.ts
+++ b/packages/internals/src/engine-commands/getConfig.ts
@@ -1,29 +1,16 @@
 import Debug from '@prisma/debug'
-import { getCliQueryEngineBinaryType } from '@prisma/engines'
-import { BinaryType } from '@prisma/fetch-engine'
 import type { DataSource, EnvValue, GeneratorConfig } from '@prisma/generator-helper'
 import chalk from 'chalk'
-import execa from 'execa'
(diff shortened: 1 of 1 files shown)


packages/internals/src/engine-commands/getConfig.ts
```diff
@@ -1,30 +1,17 @@
...
+import { ErrorArea, isWasmPanic, RustPanic } from '../panic'
 import { prismaFmt } from '../wasm'
 import { addVersionDetailsToErrorMessage } from './errorHelpers'
-import {
-  createDebugErrorType,
-  createSchemaValidationError,
-  loadNodeAPILibrary,
-  preliminaryBinaryPipeline,
-  preliminaryNodeAPIPipeline,
-  unlinkTempDatamodelPath,
```
millsp, line 21: Are these utils still used anywhere?

jkomyno: Still used in `getDmmf.ts`, will probably be gone before the sprint is over (as `getDmmf` will be ported to Wasm)

Jolg42 (approved):
🧹

PR #17524, opened 2023-01-25
fix(client): extensions prisma promise wrapping

millsp:
closes https://github.com/prisma/client-planning/issues/264

This PR basically fixes a bug with Prisma Client Extensions that when a `model` extension component returns a `PrismaPromise`, it will actually trigger the `PrismaPromise` by checking if it's `PromiseLike`. While we could have just detected that, another part of the task was to make that the errors originating in either a `Promise` or a `PrismaPromise` are treated in the same manner (ie. re-wrapping the error to prepend the extension name into it). To make this simpler and to also future-proof it, I have unified the way transaction information is passed down - which also means we could get rid of `requestTransaction` in the future while also making the rewrapping logic trivial.

### DIFF
--- a/packages/client/src/runtime/core/extensions/wrapExtensionCallback.ts
+++ b/packages/client/src/runtime/core/extensions/wrapExtensionCallback.ts
@@ -1,6 +1,6 @@
 import { isPromiseLike, mapObjectValues } from '@prisma/internals'
 
-import { omit } from '../../../../../../helpers/blaze/omit'
+import { omit } from '../../../omit'
 import { createPrismaPromise } from '../request/createPrismaPromise'
 import { isPrismaPromise } from '../request/isPrismaPromise'
 
@@ -46,7 +46,6 @@ export function wrapExtensionCallback<ResultT, ThisT, Args extends unknown[]>(
       const result = fn.apply(this, args)
 
       if (isPrismaPromise(result)) {
-        // eslint-disable-next-line @typescript-eslint/unbound-method
         const promise = createPrismaPromise((transaction) => {
           return result.catch((error) => Promise.reject(new PrismaClientExtensionError(name, error)), transaction)
         })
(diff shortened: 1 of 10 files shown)


packages/client/src/runtime/core/extensions/wrapExtensionCallback.ts
```diff
@@ -40,7 +44,15 @@ export function wrapExtensionCallback<ResultT, ThisT, Args extends unknown[]>(
...
     try {
       const result = fn.apply(this, args)
-      if (isPromiseLike(result)) {
+
+      if (isPrismaPromise(result)) {
+        const promise = createPrismaPromise((transaction) => {
+          return result.catch((error) => Promise.reject(new PrismaClientExtensionError(name, error)), transaction)
+        })
+
+        // we keep all other fields stored in the original promise
```
SevInf, line 53: What are those fields? From our `PrismaPromise` definition, only `toStringTag` is there, which would also be present on a new promise. There is also `TX_ID`, but that is added through proxy and won't be detected by `Object.assign`, i think.

millsp: They can be literally field that isn't controlled by us. Some context: Accelerate needs to add `withAccelerateMeta` method on our PrismaPromise to execute an operation with some extra arguments. So it would be like:
```ts
await prisma.user.findMany({
        where: {id: 1454545},
        cache: true,
}).withAccelerateMeta()
```

SevInf (approved):

packages/client/src/runtime/core/extensions/wrapExtensionCallback.ts
SevInf: Got it. Maybe worth adding a test for it, if we don't have one already.

PR #17531, opened 2023-01-25
test(migrate, db): add tests for `directUrl`

Jolg42:
Related: https://github.com/prisma/prisma-engines/pull/3607

Also did a cleanup of our isolation setup for SQL databases.


packages/migrate/src/__tests__/DbExecute.test.ts
```diff
@@ -635,34 +657,43 @@ COMMIT;`,
...
-              ThisisnotSQLitshouldfail
-              ^
+        db error: ERROR: at or near "thisisnotsqlitshouldfail": syntax error
+        DETAIL: source SQL:
+        ThisisnotSQLitshouldfail
+        ^
 
 
-            `)
+      `)
```
jkomyno, line 666: (At some point we should understand how to tell jest to not add new spaces when they're useless)

packages/migrate/src/__tests__/DbPush.test.ts
```diff
@@ -339,9 +349,31 @@ describe('postgresql', () => {
...
+
+      Your database is now in sync with your Prisma schema. Done in XXXms
+    `)
+    expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+  })
+
+  it('should work if url is prisma:// and directUrl defined', async () => {
+    ctx.fixture('schema-only-data-proxy')
+
+    prompt.inject(['n'])
```
jkomyno, line 364: What's this "No" manual input for?

jkomyno (approved):
This seems ok, LGTM

packages/migrate/src/__tests__/DbExecute.test.ts
Jolg42: Here it's fewer spaces, which is nice and rare 😅

packages/migrate/src/__tests__/DbPush.test.ts
Jolg42: To avoid actually doing a reset. The test still shows that `directUrl` is used, if it wasn't it would have failed.

packages/migrate/src/__tests__/fixtures/schema-only-data-proxy/with-directUrl-env.prisma
```diff
@@ -0,0 +1,11 @@
+generator client {
+    provider = "prisma-client-js"
+}
+
+datasource db {
+    provider  = "postgresql"
+    url       = "this should just fail"
```
janpio, line 7: This invalid value is not how a user would use the Data Proxy, they would have `prisma://aws-us-east-1.prisma-data.com/?api_key=MY_API_KEY` in there which currently makes validation fail :(

PR #17533, opened 2023-01-25
feat: add support for Wasm'd getDmmf

jkomyno:
Waiting for
- [x] https://github.com/prisma/prisma-engines/pull/3382
- [x] https://github.com/prisma/prisma-engines/pull/3650

This PR's client tests will fail until https://github.com/prisma/prisma/pull/17660 is merged.


packages/internals/src/engine-commands/getDmmf.ts
```diff
@@ -99,6 +110,109 @@ export async function getDMMF(options: GetDMMFOptions): Promise<DMMF.Document> {
...
+        if (process.env.FORCE_PANIC_QUERY_ENGINE_GET_DMMF) {
+          debug('Triggering a Rust panic...')
+          prismaFmt.debug_panic()
+        }
+
+        const params = JSON.stringify({
+          prismaSchema: options.datamodel,
+        })
+
+        const data = prismaFmt.get_dmmf(params)
```
janpio, line 129: Time to rename `prismaFmt`, at least this variable here already maybe?

jkomyno: I'd gladly do that in another PR, where I'm also going to remove code that's going to be made redundant by this PR.
Does `prismaWasm` sound good? Should we rename the package name as well (from `@prisma/prisma-fmt-wasm` to `@prisma/psl-utils-wasm`)?

janpio: I think we had a plan to rename it to `prisma-schema-wasm` or something like that, right @tomhoule?

tomhoule: yes there's an issue for it I think. That should be done separately.

janpio: (Ignore my comment then, `prismaFmt` still refers to a package called `prisma-fmt-wasm` - so this is correct for now.

packages/internals/src/engine-commands/queryEngineCommons.ts
```diff
@@ -139,6 +139,68 @@ export const createDebugErrorType =
...
+            reason,
+          }
+        })
+
+      return { _tag: 'parsed' as const, message: defaultMessage, ...getConfigErrorInit }
+    }),
+    E.getOrElseW(identity),
+  )
+  return actualError
+}
```
jkomyno, line 206: This block was pretty much copied and renamed from `getConfig.ts`

packages/cli/src/__tests__/artificial-panic.test.ts
```diff
@@ -131,14 +131,14 @@ describe('artificial-panic get-config', () => {
   })
 })
 
-describeIf(process.env.PRISMA_CLI_QUERY_ENGINE_TYPE == 'library')('artificial-panic library', () => {
+describe('artificial-panic get-dmmf', () => {
```
Jolg42, line 134:
```suggestion
describe('artificial-panic get-dmmf (wasm)', () => {
```
maybe? just to make it clear

Jolg42: Or maybe just a comment at the top of the file

packages/internals/src/engine-commands/getDmmf.ts
```diff
@@ -82,395 +49,94 @@ ${detailsHeader} ${message}`
...
        */
-      if (isExecaErrorCausedByRustPanic(e.error)) {
+      if (isWasmPanic(e.error)) {
+        const wasmError = e.error
         const panic = new RustPanic(
-          /* message */ e.error.shortMessage,
-          /* rustStack */ e.error.stderr,
-          /* request */ 'query-engine get-dmmf', // TODO: understand which type it expects
-          ErrorArea.QUERY_ENGINE_BINARY_CLI,
-          /* schemaPath */ options.datamodelPath ?? tempDatamodelPath,
```
Jolg42, line 394: `datamodelPath` seems to be gone but is still part of the options.
I'm unsure if we can remove it without breaking something, something to check maybe?

Jolg42 (commented):
Note:
So technically we don’t have any direct tests against the Query Engine getDmmf anymore after that.
Meaning Client tests will test the Query Engine getDmmf indirectly.

packages/internals/src/engine-commands/getDmmf.ts
jkomyno: We can check that in another PR

Jolg42: we still need to understand if someone is using only datamodelPath , there it will break
quick search says it will break redwoodjs for example
https://github.com/search?q=repo%3Aredwoodjs%2Fredwood%20getDmmf%20datamodelPath&type=code

I don’t think we want to break redwoodjs (potentially others)
Something to consider, there are a few options but an easy one is to keep the API compatible?

jkomyno: Fixed

Jolg42 (approved):
Nice!
```
PASS src/__tests__/engine-commands/getDmmf.test.ts (19.01 s)
  getDMMF
    errors
      ✓ model with autoincrement should fail if sqlite (8 ms)
      ✓ model with autoincrement should fail if mysql (2 ms)
      ✓ fails when reading a datamodel path that doesn't exist (2 ms)
      ✓ panics when the given datamodel isnt' a string (98 ms)
      ✓ validation errors (2 ms)
    success
      ✓ if a datamodel is provided, succeeds even when a non-existing datamodel path is given (4 ms)
      ✓ simple model, no datasource (13 ms)
      ✓ simple model, sqlite (13 ms)
      ✓ simple model, postgresql (25 ms)
      ✓ simple model, mysql (10 ms)
      ✓ simple model, sql server (10 ms)
      ✓ simple model, mongodb (17 ms)
      ✓ @@map model (17 ms)
      ✓ @@unique model (39 ms)
      ✓ @@id model (15 ms)
      ✓ chinook introspected schema (38 ms)
      ✓ big schema read via datamodel path (17610 ms)
```

PR #17356, opened 2023-01-16
feat(client): custom fetch for extensions

millsp:
Mainly fixes https://github.com/prisma/client-planning/issues/211
Closes https://github.com/prisma/client-planning/issues/224
Opportunistically fixes https://github.com/prisma/prisma/issues/17349
Opportunistically fixes https://github.com/prisma/prisma/issues/17563
Opportunistically fixes https://github.com/prisma/client-planning/issues/221

PR #17540, opened 2023-01-25
fix(client): Correctly handle batches inside of itx for data-proxy

SevInf:
When several requests within a single itx were batched, DataProxyEngine
executed the batch outside of itx. Therefore, if transaction failed,
batch did not roll back correctly.

Part of the reason this bug existed, in my opinion, is `Engine`
interface - from how it designed, it appears the only transaction
supported was batch transaction. That's not true - batch can and will be
executed as a part of itx, but signaling that to the engine is rather
tricky - `transaction` property, that describes batch transactions
should be `null`, but `headers` property should contain `transactionId`.
To make it more clear following refactoring has been taken:
- for batches, `transaction` option now describes both itx and batch
  transactions. Request should be executed only if it is not set.
- `headers` option is removed from public `Engine` interface.
  `traceparent`, which is the only remaining header option left after
  transaction option refactoring is moved to
  `RequestOptions`/`BatchRequestOptions`. 
  
 Update: For Accelerate functionality, ability to set custom headers for data proxy via extension is resotred.

Fix prisma/client-planning#267


packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -133,16 +130,25 @@ export class DataProxyEngine extends Engine {
...
 
+    const headers: Record<string, string> = {}
+    if (traceparent) {
+      headers.traceparent = traceparent
+    }
+
+    if (transaction) {
+      headers['X-transaction-id'] = transaction.id
+    }
+
```
miguelff, line 144: Right, no undefined headers being set 👌🏾

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -165,8 +165,7 @@ export type InternalRequestParams = {
    */
   jsModelName?: string
   callsite?: CallSite
-  /** Headers metadata that will be passed to the Engine */
-  headers?: Record<string, string> // TODO what is this
```
millsp, line 169: This is used by the Accelerate team, we'll need to merge other work to provide an alternative

SevInf: Restored the property and functionality for now

SevInf: One note: once #17356 is merged, we could remove it from our side, right?

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -28,10 +37,12 @@ export type RequestParams = {
   modelName?: string
   protocolMessage: ProtocolMessage
   dataPath: string[]
+  debugPanic?: 'fatal' | 'non-fatal'
```
millsp, line 40: This is only implemented in binary engine and not in the two others (which are the main ones), can we maybe remove?

packages/client/src/runtime/getPrismaClient.ts
millsp: Yes

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -64,21 +76,12 @@ type ApplyExtensionsParams = {
 }
 
 function getRequestInfo(request: Request) {
```
millsp, line 78: This is less useful now, maybe we can also remove?

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -96,28 +99,30 @@ export class RequestHandler {
...
+          transaction: getTransactionOptions(info.transaction),
           containsWrite,
+          customDataProxyHeaders: info.customDataProxyHeaders,
         })
       },
       singleLoader: (request) => {
         const info = getRequestInfo(request)
-        const interactiveTransaction = info.transaction?.kind === 'itx' ? info.transaction : undefined
+        const interactiveTransaction =
+          info.transaction?.kind === 'itx' ? getItxTransactionOptions(info.transaction) : undefined
```
millsp, line 118: I think a batch with a single query is also possible, might be useful to set isolation on a `$queryRaw`

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -96,28 +99,30 @@ export class RequestHandler {
...
+          traceparent,
+          transaction: getTransactionOptions(info.transaction),
           containsWrite,
+          customDataProxyHeaders: info.customDataProxyHeaders,
         })
       },
       singleLoader: (request) => {
         const info = getRequestInfo(request)
-        const interactiveTransaction = info.transaction?.kind === 'itx' ? info.transaction : undefined
+        const interactiveTransaction =
```
millsp, line 117: I just noticed that this may be wrong. It is possible for a batch transaction to have a single query. Thay feels edge-casey but it may be useful to set an isolation level of a raw query, for instance. So my suggestion is to just pass the transaction options here as well.

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -895,16 +891,25 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
...
+    }
 
-    // TODO: we don't need the transactionId "runtime header" anymore, we can use the txInfo object here
-    this.currentRequestPromise = this.connection.post(
-      '/',
-      stringifyQuery(query.query),
-      runtimeHeadersToHttpHeaders(headers),
-    )
+    if (transaction) {
+      headers['X-transaction-id'] = transaction.id
```
millsp, line 903: I think this is currently sending a batch id as a header, is it correct?

packages/engine-core/src/common/Engine.ts
```diff
@@ -23,7 +19,17 @@ export type BatchTransactionOptions = {
   isolationLevel?: Transaction.IsolationLevel
 }
 
-export type InteractiveTransactionOptions<Payload> = Transaction.Info<Payload>
+export type TransactionOptions<InteractiveTransactionPayload> =
```
millsp, line 22: Do you think it would be possible to re-use `PrismaPromiseBatchTransaction` and `PrismaPromiseInteractiveTransaction`?

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -944,18 +949,24 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
...
-    const request: QueryEngineBatchRequest = {
-      batch: queries.map(({ query }) => ({ query, variables: {} })),
-      transaction: Boolean(transaction),
-      isolationLevel: transaction?.isolationLevel,
+    const headers: Record<string, string> = {}
+    if (traceparent) {
+      headers.traceparent = traceparent
     }
 
+    const itxId = getInteractiveTransactionId(transaction)
```
millsp, line 961:
```suggestion
    const itx = transaction?.kind === 'itx' ? transaction.options : undefined
```
(Optional, for consistency with dataproxy)

packages/engine-core/src/common/types/Transaction.ts
```diff
@@ -30,6 +30,10 @@ export type Info<Payload = unknown> = {
   payload: Payload
 }
 
+export type TransactionStartParameters = {
```
millsp, line 33: This looks unused

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -133,16 +130,28 @@ export class DataProxyEngine extends Engine {
...
+  ) {
     this.logEmitter.emit('query', { query })
 
+    const headers: Record<string, string> = { ...customDataProxyHeaders }
+    if (traceparent) {
+      headers.traceparent = traceparent
+    }
+
+    if (transaction) {
+      headers['X-transaction-id'] = transaction.id
```
millsp, line 145: I think this can possibly send a batch id, is it intentional?

packages/engine-core/src/data-proxy/DataProxyEngine.ts
```diff
@@ -151,13 +160,19 @@ export class DataProxyEngine extends Engine {
...
-      transaction: isTransaction,
-      isolationLevel: transaction?.isolationLevel,
+    const headers: Record<string, string> = { ...customDataProxyHeaders }
+    if (traceparent) {
+      headers.traceparent = traceparent
+    }
+
+    const itx = transaction?.kind === 'itx' ? transaction.options : undefined
+    if (itx) {
+      headers['X-transaction-id'] = itx.id
```
millsp, line 170: Do you think it would be possible to factor/deduplicate this shared logic into `requestInternal`?

millsp: Looks similar to https://github.com/prisma/prisma/pull/17540/files#diff-b4fc1a3eee6a1f0140a901585106e8b74dfc77aff2f46bb68e5cd502450dabf5R139

packages/engine-core/src/data-proxy/DataProxyEngine.ts
SevInf: No, singular request receives only iTx options

SevInf: Also, batch ids are not propagated to `engine-core` - `transaction: { kind: 'batch}` has no id field.

packages/engine-core/src/binary/BinaryEngine.ts
SevInf: No, batch ids are not propagated to `engine-core`. I'll rename paramter to make it clearer.

packages/engine-core/src/common/Engine.ts
SevInf: I was thinking about it, but in the end decided to not do it - batch id does not have any meaning to the engine, it's purely client invention, so it felt wrong adding it to the engine types, but we absolutely need it in the client.

packages/engine-core/src/binary/BinaryEngine.ts
SevInf: it's now named `interactiveTransaction`.

packages/client/src/runtime/RequestHandler.ts
SevInf: Nicely spot, i think that was broken since forever. Fix is actually more complicated than what you suggesting - only `requestBatch` has transaction options, not singular `requests`. So, if it's a single request with batch tx options set, it has to go through `requestBatch` endpoint. This what I did in separate pr #17556 since it is a separate issue.

packages/engine-core/src/data-proxy/DataProxyEngine.ts
SevInf: Done

millsp (approved):

PR #17577, opened 2023-01-27
ci(test): Run test "At every 15th minute past every hour from 3 through 5."

janpio:
Explanation of the crontab syntax: https://crontab.guru/#*/15_3-5_*_*_*

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,9 @@ on:
       - '*.bench.ts'
       - 'scripts/ci/publish.ts'
       - 'graphs/**'
-
+  schedule:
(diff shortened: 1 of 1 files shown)


Jolg42:
Why do we need this @janpio ?

millsp (approved):

janpio:
Pierre has the theory that BuildPulse, the tool we use to find flaky tests would really benefit from running the same commit a few times to see which tests sometimes pass and sometimes don't - and a few nightly runs seem like a good approach to see if that is true. No one else will need to wait for these.

(Although now writing this, I wonder if this will trigger on all branches or only `main`... We'll know tomorrow morning I guess :yolo:)

PR #17579, opened 2023-01-27
fix(db pull): delete removed introspection warnings

Jolg42:
We found out during our meeting that the engines had less codes than the CLI.
https://github.com/prisma/prisma-engines/blob/main/introspection-engine/connectors/sql-introspection-connector/src/warnings.rs

I investigated with `git log -S 'code: 12'` and found the commits in engines (see notes in code).

and here is the cleanup ✨

### DIFF
--- a/packages/migrate/src/commands/DbPull.ts
+++ b/packages/migrate/src/commands/DbPull.ts
@@ -456,23 +456,9 @@ ${`Run ${chalk.green(getCommandWithExecutor('prisma generate'))} to generate Pri
             .join('\n')
         } else if (warning.code === 4) {
           message += warning.affected.map((it) => `- Enum "${it.enm}", value: "${it.value}"`).join('\n')
-        } else if (
-          warning.code === 5 ||
-          warning.code === 6 ||
-          warning.code === 8 ||
-          warning.code === 11 ||
-          warning.code === 12 ||
-          warning.code === 13 ||
-          warning.code === 16
-        ) {
+        } else if (warning.code === 5 || warning.code === 6 || warning.code === 8) {
           message += warning.affected.map((it) => `- Model "${it.model}", field: "${it.field}"`).join('\n')
-        } else if (
-          warning.code === 7 ||
-          warning.code === 14 ||
-          warning.code === 15 ||
-          warning.code === 18 ||
-          warning.code === 19
-        ) {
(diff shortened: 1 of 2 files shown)


Druue (approved):
LGTM!

packages/migrate/src/commands/DbPull.ts
```diff
@@ -456,23 +456,9 @@ ${`Run ${chalk.green(getCommandWithExecutor('prisma generate'))} to generate Pri
...
+        } else if (warning.code === 5 || warning.code === 6 || warning.code === 8) {
           message += warning.affected.map((it) => `- Model "${it.model}", field: "${it.field}"`).join('\n')
-        } else if (
-          warning.code === 7 ||
-          warning.code === 14 ||
-          warning.code === 15 ||
-          warning.code === 18 ||
-          warning.code === 19
-        ) {
+        } else if (warning.code === 7 || warning.code === 14 || warning.code === 18 || warning.code === 19) {
```
jkomyno, line 461:
```suggestion
        } else if [7, 14, 18, 19].includes(warning.code) {
```

packages/migrate/src/commands/DbPull.ts
```diff
@@ -456,23 +456,9 @@ ${`Run ${chalk.green(getCommandWithExecutor('prisma generate'))} to generate Pri
...
-        } else if (
-          warning.code === 5 ||
-          warning.code === 6 ||
-          warning.code === 8 ||
-          warning.code === 11 ||
-          warning.code === 12 ||
-          warning.code === 13 ||
-          warning.code === 16
-        ) {
+        } else if (warning.code === 5 || warning.code === 6 || warning.code === 8) {
```
jkomyno, line 459:
```suggestion
        } else if ([5, 6, 8].includes(warning.code))  {
```

jkomyno (approved):
LGTM, but please consider the suggestions to avoid long lines of repeated `warning.code === n`

packages/migrate/src/commands/DbPull.ts
Jolg42: Unfortunately, that would mean that the code is not type safe anymore with this suggestion

Jolg42: See before
Screenshot 2023-01-27 at 17 15 07
and with your suggestion
Screenshot 2023-01-27 at 17 14 45

PR #17590, opened 2023-01-27
fix(client): extensions types for i/tx and add tests

### LINKED ISSUE #16582: Client in interactive transaction is not extended
### Bug description

Prisma client in interactive transaction callback is not extended .

### How to reproduce

The following example causes TypeScript compilation error.

```ts

const client = new PrismaClient();

const prisma = client.$extends({
  model: {
    post: {
      async findBy(id: string) {
        return await client.post.findUniqueOrThrow({ where: { id } });
      },
    },
  },
});

prisma.post.findBy("post001");

// @ts-expect-error Property 'findBy' does not exist...
prisma.$transaction(txClient => txClient.post.findBy("post001"));
```

### Expected behavior

No compilation error.

And model / client inner transaction should be extended .

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
```

### Environment & setup

- OS: macOS
- Database: 
- Node.js version: v16.16.0
- TypeScript: 4.9.3

### Prisma Version

```
prisma                  : 4.7.0
@prisma/client          : 4.7.0
Query Engine (Node-API) : libquery-engine 39190b250ebc338586e25e6da45e5e783bc8a635 (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli 39190b250ebc338586e25e6da45e5e783bc8a635 (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core 39190b250ebc338586e25e6da45e5e783bc8a635 (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary           : prisma-fmt 39190b250ebc338586e25e6da45e5e783bc8a635 (at node_modules/@prisma/engines/prisma-fmt-darwin)
Format Wasm             : @prisma/prisma-fmt-wasm 4.7.0-74.39190b250ebc338586e25e6da45e5e783bc8a635
Default Engines Hash    : 39190b250ebc338586e25e6da45e5e783bc8a635
Studio                  : 0.477.0

```

millsp:
Closes https://github.com/prisma/prisma/issues/16582

### DIFF
--- a/packages/client/src/generation/TSClient/PrismaClient.ts
+++ b/packages/client/src/generation/TSClient/PrismaClient.ts
@@ -259,7 +259,22 @@ function interactiveTransactionDefinition(this: PrismaClientClass) {
   const returnType = ts.promise(ts.namedType('R'))
   const callbackType = ts
     .functionType()
-    .addParameter(ts.parameter('prisma', ts.namedType('this')))
+    .addParameter(
+      ts.parameter(
+        'prisma',
+        ts
+          .namedType('Omit')
+          .addGenericArgument(ts.namedType('this'))
+          .addGenericArgument(
+            ts
+              .unionType(ts.stringLiteral('$connect'))
+              .addVariant(ts.stringLiteral('$disconnect'))
+              .addVariant(ts.stringLiteral('$on'))
+              .addVariant(ts.stringLiteral('$transaction'))
+              .addVariant(ts.stringLiteral('$use')),
+          ),
+      ),
+    )
     .setReturnType(returnType)
 
   const method = ts
--- a/packages/client/src/generation/TSClient/PrismaClient.ts
+++ b/packages/client/src/generation/TSClient/PrismaClient.ts
@@ -259,7 +259,7 @@ function interactiveTransactionDefinition(this: PrismaClientClass) {
   const returnType = ts.promise(ts.namedType('R'))
   const callbackType = ts
     .functionType()
-    .addParameter(ts.parameter('prisma', ts.namedType('Prisma.TransactionClient')))
+    .addParameter(ts.parameter('prisma', ts.namedType('this')))
     .setReturnType(returnType)
 
   const method = ts
--- a/packages/client/tests/functional/extensions/enabled/itx.ts
+++ b/packages/client/tests/functional/extensions/enabled/itx.ts
@@ -84,17 +84,7 @@ testMatrix.setupTestSuite(() => {
       expectTypeOf(userA?.fullName).toEqualTypeOf<string>()
     })
 
-    await expect(result).rejects.toMatchPrismaErrorInlineSnapshot(`
-
-      Invalid \`tx.user.create()\` invocation in
-      /client/tests/functional/extensions/enabled/itx.ts:0:0
-
-        XX   },
-        XX })
-        XX 
-      → XX const userB = await tx.user.create(
-      Unique constraint failed on the fields: (\`email\`)
-    `)
+    await expect(result).rejects.toMatchPrismaErrorSnapshot()
 
     const users = await prisma.user.findMany({ where: { email: 'jane@smith.com' } })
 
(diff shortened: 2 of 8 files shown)


packages/client/tests/functional/extensions/enabled/prisma/_schema.ts
```diff
@@ -51,16 +19,12 @@ export default testMatrix.setupSchema(({ provider }) => {
...
   }
 
   model Post {
     id ${idForProvider(provider)}
     user User @relation(fields: [userId], references: [id])
     userId ${foreignKeyForProvider(provider)}
   }
-
-  // the following models aren't used in the tests, but are here to test the performance of the extension
-  ${createModel(20, provider)}
```
millsp, line 64: Not needed anymore, just makes things slower. This is tested in `prisma/packages/client/tests/e2e/issues/16600-tsc-crash-extensions` with 78 models and nested relations.

packages/client/tests/functional/extensions/enabled/itx.ts
```diff
@@ -0,0 +1,212 @@
...
+
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+const email = faker.internet.email()
+
+testMatrix.setupTestSuite(() => {
```
millsp, line 12: This PR is mostly a type change so I thought it would be nice to also test itx at runtime via extensions for some basic sanity checks.

packages/client/tests/functional/extensions/enabled/tx.ts
```diff
@@ -0,0 +1,184 @@
...
+
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+const email = faker.internet.email()
+
+testMatrix.setupTestSuite(() => {
```
millsp, line 12: Similarly, that wasn't strictly required, I added tests for batch transactions

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -259,7 +259,22 @@ function interactiveTransactionDefinition(this: PrismaClientClass) {
...
-    .addParameter(ts.parameter('prisma', ts.namedType('Prisma.TransactionClient')))
+    .addParameter(
+      ts.parameter(
+        'prisma',
+        ts
+          .namedType('Omit')
+          .addGenericArgument(ts.namedType('this'))
+          .addGenericArgument(
+            ts
+              .unionType(ts.stringLiteral('$connect'))
```
SevInf, line 270: I wonder if we should unite this list with runtime one from `getPrismaClient` so those two will always be in sync.

millsp: Good idea, will do

SevInf (approved):

packages/client/src/generation/TSClient/PrismaClient.ts
millsp: Doing so causes issues with Jest 

```ts
    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /home/millsp/Work/prisma/node_modules/.pnpm/sql-template-tag@5.0.3/node_modules/sql-template-tag/dist/index.js:4
    export class Sql {
    ^^^^^^
```

PR #17596, opened 2023-01-28
ci: continue when Ana06/get-changed-files step fails

janpio:
Turns out this action in this step can not real with scheduled runs:

```
Run Ana06/get-changed-files@v2.2.0
  with:
    format: json
    token: ***
    filter: *
  env:
    HAS_BUILDPULSE_SECRETS: true
    PRISMA_TELEMETRY_INFORMATION: prisma test.yml
    PRISMA_HIDE_UPDATE_MESSAGE: true
Error: This action only supports pull requests and pushes, schedule events are not supported. Please submit an issue on this action's GitHub repo if you believe this in correct.
Base commit: undefined
Head commit: undefined
Error: The base and head commits are missing from the payload for this schedule event. Please submit an issue on this action's GitHub repo.
Error: Not Found
```

via https://github.com/prisma/prisma/actions/runs/4029879690/jobs/6928157785

This change should make the next step to run anyway, which would then decide to run _all_ the tests, as intended.

Docs on used property: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
         uses: Ana06/get-changed-files@v2.2.0 # it's a fork of jitterbit/get-changed-files@v1 which works better with pull requests
         with:
           format: 'json'
-        continue-on-error: true
+        continue-on-error: true # This allow this job to fail (e.g. for scheduled runs), and then the `detect` step below will default to `all`
       - id: detect
         run: ./.github/workflows/scripts/detect-jobs-to-run.js <<<'${{ steps.files.outputs.all }}'
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,7 @@ jobs:
         uses: Ana06/get-changed-files@v2.2.0 # it's a fork of jitterbit/get-changed-files@v1 which works better with pull requests
         with:
           format: 'json'
+        continue-on-error: true
       - id: detect
         run: ./.github/workflows/scripts/detect-jobs-to-run.js <<<'${{ steps.files.outputs.all }}'
 


.github/workflows/test.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         uses: Ana06/get-changed-files@v2.2.0 # it's a fork of jitterbit/get-changed-files@v1 which works better with pull requests
         with:
           format: 'json'
+        continue-on-error: true
```
Jolg42, line 67: This sounds interesting, but the conditions below will still be falsy, example
```
contains(needs.detect_jobs_to_run.outputs.jobs, '-all-')
```
So let's merge https://github.com/prisma/prisma/pull/17598

janpio:
Alternative solution from Pierre: https://github.com/prisma/prisma/pull/17598

.github/workflows/test.yml
```diff
@@ -64,6 +64,7 @@ jobs:
...
       - id: detect
         run: ./.github/workflows/scripts/detect-jobs-to-run.js <<<'${{ steps.files.outputs.all }}'
```
Jolg42, line 69: Let's try this, though 100% confident this will work so let's watch how it goes

It relies on the fallback to "all" on https://github.com/prisma/prisma/blob/main/.github/workflows/scripts/detect-jobs-to-run.js

janpio: Oh yes, I should have added that as a comment. Also not sure how this will exactly be displayed etc :/ 

Pierre has a solution that is most probably cleaner, but requires a lot more diff: https://github.com/prisma/prisma/pull/17598

Jolg42 (approved):
If this works, I think it's a good general improvement as well

.github/workflows/test.yml
janpio, line 67:
```suggestion
        continue-on-error: true # This allow this job to fail (e.g. for scheduled runs), and then the `detect` step below will default to `all`
```

Jolg42:
This didn't work because we didn't test it 😄 

I did a commit to fix in https://github.com/prisma/prisma/commit/3fde8fb4aaf2f6b011d5db17e3ed0c7a1f1546d5

janpio:
@Jolg42 That's the problem of individual actions - hard to test :/

Jolg42:
Note: this should work now https://github.com/prisma/prisma/commit/3e695e75aa27059dd957383526b590466f24bad3

PR #17598, opened 2023-01-28
fix(ci): test all if schedule

millsp:
Attempt to fix this https://github.com/prisma/prisma/actions/runs/4030195923/jobs/6928702630


janpio:
Variant with fewer edits: https://github.com/prisma/prisma/pull/17596
Not sure which one is better.

Jolg42:
Note: we merged the variant for now

PR #17642, opened 2023-01-31
fix(client): Support partial selections when boolean type is unknown

RhysSullivan:
Currently the prisma client breaks typesafety when passing in booleans to a select query

image

In the above screenshot, it's possible for name to not be selected if the parameter is false, but the type generated indicates that this will be defined. This is a bug.

This is because caused due to the use of TruthyKeys, which evaluates an unknown boolean type to be the equivalent of true when this is actually an error.

In order to fix this with full Typesafety, the proposed solution in this PR is to first generate the full payload wrapped with a partial, and then perform an intersection on the keys that are actually truthy. The result of this is the following image:

image

As we can see from this image, the name field is being correctly marked as possibly undefined, allowing for typesafe access of it.

**Object spreading is a work around currently**

This screenshot generates the Prisma types correctly on the existing codebase

image

A use case of something like this would be for conditionally fetching data based on permissions. i.e, optionally including an email field to return from an api

The PR is more of a proof of concept to show that this is a possibility and to raise awareness of this issue, the actual implementation may differ if anyone wants to continue from this point

### DIFF
--- a/packages/client/src/generation/TSClient/Payload.ts
+++ b/packages/client/src/generation/TSClient/Payload.ts
@@ -14,7 +14,7 @@ export class PayloadType implements Generatable {
     protected readonly type: OutputType,
     protected readonly dmmf: DMMFHelper,
     protected readonly findMany = true,
-  ) {}
+  ) { }
 
   public toTS(): string {
     const { type } = this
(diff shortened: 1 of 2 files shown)


janpio:
(Can you maybe split out the problem into an issue that this PR will fix? That would make it much easier for us to also list this in release notes and so on later.)

PR #17652, opened 2023-01-31
fix(client): ensure runtime type declaration have no dependencies

SevInf:
We've accidentyaly exposed NodeFetch types to the runtime, that
intoduced dependency on node-fetch package. That breaks runtime types if
the module is not installed. Fixed by redeclaring node-fetch types
instead of reusing them. Also, adds a test to check it won't happen
again.


millsp (approved):

Jolg42:
It made it to `4.10.0-dev.54` which passes 🟢 in ecosystem-tests ✅ 
Nice!

PR #17658, opened 2023-01-31
refactor: remove time from `format` output

Jolg42:
- It's not really useful (it's always fast!)
- It needs extra handling in tests
- sometimes in CI it can take more than 1 sec and fail our test that expects only <1s
	- example https://github.com/prisma/prisma/actions/runs/4056063093/jobs/6980062291#step:15:1208 ``` Expected pattern: /^Formatted (.*) in \d+ms ��$/ Received string:  "Formatted /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/8a1a99abd0e38f21d12a31d73e8b7822/schema.prisma in 1.26s ��" ``` Let's remove this!

### DIFF
--- a/packages/cli/src/Format.ts
+++ b/packages/cli/src/Format.ts
@@ -1,4 +1,4 @@
-import { arg, Command, format, formatms, formatSchema, getDMMF, HelpError } from '@prisma/internals'
+import { arg, Command, format, formatSchema, getDMMF, HelpError } from '@prisma/internals'
 import { getSchemaPathAndPrint } from '@prisma/migrate'
 import chalk from 'chalk'
 import fs from 'fs'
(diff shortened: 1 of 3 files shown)


packages/cli/src/Format.ts
```diff
@@ -65,9 +64,8 @@ Or specify a Prisma schema path
     }
 
     fs.writeFileSync(schemaPath, output)
-    const after = Date.now()
 
-    return `Formatted ${chalk.underline(schemaPath)} in ${formatms(after - before)} 🚀`
```
millsp, line 70: Was this useful to users in any way?

jkomyno: Let's not remove this.

Jolg42: It was not useful for us or users

Jolg42: Can you give a reason at least @jkomyno ?

jkomyno: It's a nice esthetical touch for the users with no drawbacks whatsoever, and it helps conveying the idea that Prisma utilities are fast. Removing it with no reason would be a mistake imho

Jolg42:
Closing as stale 😞

PR #17660, opened 2023-01-31
test(client): use getDMMF from @prisma/internals in client tests

jkomyno:
The client has historically used `getDMMF` both from `@prisma/internals` and from `@prisma/engine-core`.
The client also used to re-export `getDMMF` from `@prisma/internals` (which is compiled to Wasm in https://github.com/prisma/prisma/pull/17533). As verified with @millsp and @Jolg42, this isn't used in Github projects or by the PDP team.

You can verify that the changes in this PR allow this other PR to succeed on the CI: https://github.com/prisma/prisma/pull/17662


packages/client/src/utils/generateInFolder.ts
```diff
@@ -123,6 +118,12 @@ export async function generateInFolder({
           },
         }
 
+  // TODO: use engine.getDmmf()
+  const dmmf = await getDMMF({
+    datamodel,
+    previewFeatures,
+  })
```
jkomyno, line 125: Note: apparently, this TODO isn't even needed

Jolg42 (approved):
Nice!

millsp (approved):

PR #17661, opened 2023-01-31
feat: move snapshot utilities to `@prisma/get-platform`, test `getPlatform` output

jkomyno:
This PR:
- [x] moves `jestContext` and its related utilities to `@prisma/get-platform`
- [x] adds snapshot testing to `getPlatform()` 
- [x] deprecates https://github.com/prisma/prisma/pull/17588 (let's think about introducing a new package in another sprint)


packages/get-platform/src/__tests__/getSSLVersion.test.ts
```diff
@@ -5,8 +5,9 @@ const describeIf = (condition: boolean) => (condition ? describe : describe.skip
 describeIf(process.platform === 'linux')('getSSLVersion', () => {
   it('should not return an error', async () => {
     const arch = 'x64'
-    await getSSLVersion({ targetDistro: 'debian', arch })
-    await getSSLVersion({ targetDistro: 'musl', arch })
-    await getSSLVersion({ targetDistro: 'rhel', arch })
+    const archFromUname = 'x86_64'
+    await getSSLVersion({ targetDistro: 'debian', arch, archFromUname })
+    await getSSLVersion({ targetDistro: 'musl', arch, archFromUname })
+    await getSSLVersion({ targetDistro: 'rhel', arch, archFromUname })
```
jkomyno, line 11: The test was passing, but it's not type-checked, so we had forgot to pass one argument

packages/get-platform/src/getPlatform.ts
```diff
@@ -37,15 +37,28 @@ export type DistroInfo = {
...
+  | {
+      platform: Omit<NodeJS.Platform, 'linux'>
+      arch: Arch
+      targetDistro?: DistroInfo['targetDistro']
+      familyDistro?: never
+      originalDistro?: never
+      archFromUname?: never
+      libssl?: never
+    }
+  | GetOsResultLinux
```
jkomyno, line 60: This is now a discriminated union on the `platform` property.
This was necessary to move the `alpine:unsupported-arch` into `getPlatformInternal` in a type-safe fashion.

packages/get-platform/src/test-utils/jestSnapshotSerializer.js
```diff
@@ -0,0 +1,152 @@
...
+const replaceAll = require('replace-string') // sindre's replaceAll polyfill
+const stripAnsi = require('strip-ansi')
+const { platformRegex } = require('./platformRegex')
+
+// Pipe utility
+const pipe =
+  (...fns) =>
+  (x) =>
+    fns.reduce((v, f) => f(v), x)
+
```
jkomyno, line 11: We've moved this file from @prisma/internals

packages/get-platform/src/test-utils/platformRegex.ts
```diff
@@ -0,0 +1,20 @@
+import escapeString from 'escape-string-regexp'
+
+import { platforms } from '../platforms'
+
```
jkomyno, line 4: We've moved this file from @prisma/internals

packages/get-platform/src/__tests__/getPlatform.test.ts
```diff
@@ -0,0 +1,140 @@
...
+          archFromUname: 'x86_64',
+          familyDistro: 'debian',
+          originalDistro: 'debian',
+          targetDistro: 'debian',
+        }),
+      ).toBe('debian-openssl-1.1.x')
+      expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
```
jkomyno, line 27: We now have warning/log/error snapshot testing for `getPlatform`, like Jan wanted

packages/get-platform/jest.config.js
```diff
@@ -4,9 +4,22 @@ module.exports = {
...
   collectCoverage: process.env.CI ? true : false,
   coverageReporters: ['clover'],
   coverageDirectory: 'src/__tests__/coverage',
   collectCoverageFrom: ['src/**/*.ts', '!**/__tests__/**/*'],
-  testMatch: ['**/src/__tests__/**/*.test.ts'],
+  snapshotSerializers: ['./src/test-utils/jestSnapshotSerializer'],
+  reporters: [
+    'default',
+    [
+      'jest-junit',
```
Jolg42, line 16: You'll need to add `jest-junit` as a dev dependency (like done in the other packages)

packages/get-platform/package.json
```diff
@@ -24,7 +24,13 @@
   "dependencies": {
     "@prisma/debug": "workspace:*",
     "chalk": "4.1.2",
+    "execa": "5.1.1",
+    "escape-string-regexp": "4.0.0",
+    "fs-jetpack": "5.1.0",
+    "replace-string": "3.1.0",
+    "strip-ansi": "6.0.1",
     "terminal-link": "2.1.1",
+    "tempy": "1.0.1",
```
Jolg42, line 33: If these new dependencies are only for tests, then we can move them to `devDependencies`

packages/get-platform/src/__tests__/getPlatform.test.ts
```diff
@@ -0,0 +1,140 @@
...
+          arch: 'x64',
+          archFromUname: 'x86_64',
+          familyDistro: 'debian',
+          originalDistro: 'ubuntu',
+          targetDistro: 'debian',
+        }),
+      ).toBe('debian-openssl-1.1.x')
+      expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`
+        prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".
```
Jolg42, line 80: `Defaulting to "openssl-1.1.x".` should actually be `debian-openssl-1.1.x` here?

packages/get-platform/src/__tests__/getPlatform.test.ts
```diff
@@ -0,0 +1,140 @@
...
+          libssl: undefined,
+          arch: 'x64',
+          archFromUname: 'x86_64',
+          familyDistro: undefined,
+          originalDistro: 'unknown',
+          targetDistro: undefined,
+        }),
+      ).toBe('debian-openssl-1.1.x')
+      expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
```
Jolg42, line 136: Warn should also say " prisma:warn Prisma doesn't know which engines to download for the Linux distro "unknown". Falling back to Prisma engines built "debian"." here?

packages/get-platform/src/__tests__/getPlatform.test.ts
jkomyno: We're only talking about the default value for `libssl`, so no, `openssl-1.1.x` is the right thing to display in this case

packages/get-platform/src/__tests__/getPlatform.test.ts
jkomyno: That's possible, let me double-check that

packages/get-platform/package.json
jkomyno: Right call. Let me fix that.

packages/get-platform/src/__tests__/getPlatform.test.ts
Jolg42: This looks confusing as it has the same formatting as our "binaryTarget", not sure if we can do something but also maybe not the most important now.

Jolg42 (approved):
🧪

Jolg42:
Note: this was accidentally merged with tests failing, new PR coming soon to fix the tests

PR #17678, opened 2023-02-01
feat(get-platform): memoize `getPlatform` for repeated invocations

jkomyno:
- [x] remove `warnOnce`, as it caused indeterministic troubles with test snapshots
- [x] make sure warnings all called once by simply memoizing the result of `getPlatform`


Jolg42 (approved):
Much better!

PR #17694, opened 2023-02-02
fix: read header lower-case

marcoieni:
Header names are lower-cased. See [docs](https://nodejs.org/api/http.html#messageheaders):
image

I also run a manual test:
216300787-d27fa2b6-6950-4521-91a2-2bf337949168_png__2342×846_

### DIFF
--- a/packages/engine-core/src/__tests__/errors.test.ts
+++ b/packages/engine-core/src/__tests__/errors.test.ts
@@ -12,7 +12,7 @@ const response = (body: string, code?: number, requestId?: string): RequestRespo
   ok: false,
   status: code || 400,
   headers: {
-    'Prisma-Request-Id': requestId,
+    'prisma-request-id': requestId,
   },
 })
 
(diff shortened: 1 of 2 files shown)


packages/engine-core/src/__tests__/errors.test.ts
```diff
@@ -168,7 +168,7 @@ describe('getErrorMessageWithLink', () => {
 
       If you want the Prisma team to look into it, please open the link above 🙏
       To increase the chance of success, please post your schema and a snippet of
-      how you used Prisma Client in the issue. 
+      how you used Prisma Client in the issue.
```
MarcoIeni, line 171:
```suggestion
      how you used Prisma Client in the issue. 
```

SevInf (approved):

MarcoIeni:
Some jobs failed. Is the CI a bit flaky? Or do you think it's a problem with my code?

SevInf:
@MarcoIeni that test is unfortinately known to be flaky, your changes look good

MarcoIeni:
Now it's all green!

PR #17708, opened 2023-02-02
test(client): unskip vitess test on buildkite

Jolg42:


Jolg42:
Still fails

Jolg42:
We don't use buildkite here anymore

PR #17753, opened 2023-02-06
integration: add precise engine core error messages when loading `libquery`

jkomyno:
Integration PR for https://github.com/prisma/prisma/pull/17762


packages/engine-core/src/common/errors/utils/handleEngineLoadingErrors.ts
```diff
@@ -0,0 +1,76 @@
...
+  const potentialReasonMessage = match({ message: error.message, code: error.code })
+    .with({ code: 'ENOENT' }, () => `File does not exist.`)
+    .when(
+      ({ message }) => hasLinkingProblem && message.includes('libz'),
+      () => {
+        return `${systemLibraryNotFound('libz')}. Please install it and try again.`
+      },
+    )
+    .when(
+      ({ message }) => hasLinkingProblem && message.includes('libcc_s'),
```
aqrln, line 32:
```suggestion
      ({ message }) => hasLinkingProblem && message.includes('libgcc_s'),
```

packages/engine-core/src/common/errors/utils/handleEngineLoadingErrors.ts
```diff
@@ -0,0 +1,76 @@
...
+      () => {
+        return `${systemLibraryNotFound('libcc_s')}. Please install it and try again.`
```
aqrln, line 34:
```suggestion
        return `${systemLibraryNotFound('libgcc_s')}. Please install it and try again.`
```

packages/get-platform/src/__tests__/getPlatform.test.ts
```diff
@@ -1,4 +1,5 @@
 import stripAnsi from 'strip-ansi'
+import { C } from 'ts-toolbelt'
```
aqrln, line 2: unused import

packages/get-platform/src/getPlatform.ts
```diff
@@ -375,21 +378,39 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
...
+
+function isPlatformWithOSResultDefined(args: Partial<PlatformWithOSResult>): args is PlatformWithOSResult {
+  return args.binaryTarget !== undefined
+}
+
+/**
+ * Get the binary target and other system information (e.g., the libssl version to look for) for the current platform.
+ */
+export async function getPlatformWithOSResult(): Promise<PlatformWithOSResult> {
+  const { memoized, ...rest } = await getPlatformMemoized()
```
aqrln, line 399: Doesn't eslint complain about an unused variable here? I vaguely remember that once when I wanted to do something like this I basically did something analogous to `const { memoized: _, ...rest } = await getPlatformMemoized()` and added `_` to the exceptions of the `no-unused-vars` rule in eslint config, but then I had to revert it because someone else found it unintuitive during the code review and asked to use `delete` instead. I think it's a neat pattern though 😄

packages/engine-core/src/common/errors/utils/handleEngineLoadingErrors.ts
jkomyno: Nice catch

packages/get-platform/src/__tests__/getPlatform.test.ts
jkomyno: How the heck did this end up here? 😅
Thanks for spotting this

packages/get-platform/src/getPlatform.ts
jkomyno: There's no reason to use `delete`, that would really be counterproductive in terms of ergonomics.
Let's go for `const { memoized: _, ...rest }`

PR #17826, opened 2023-02-09
ci: enable buildpulse for client tests

millsp:
TODO: As a bonus I'd like to move repetitive config to a separate file

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -158,6 +158,8 @@ jobs:
       TEST_FUNCTIONAL_MONGO_URI: 'mongodb://root:prisma@localhost:27018/PRISMA_DB_NAME?authSource=admin'
       TEST_FUNCTIONAL_COCKROACH_URI: 'postgresql://prisma@localhost:26257/PRISMA_DB_NAME'
       NODE_OPTIONS: '--max-old-space-size=8096'
+      JEST_JUNIT_SUITE_NAME: '${{ github.job }}/client/functional/${{ matrix.os }}/node-${{ matrix.node }}/${{ matrix.queryEngine }}'
+      JEST_JUNIT_UNIQUE_OUTPUT_NAME: true
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -158,6 +158,8 @@ jobs:
       TEST_FUNCTIONAL_MONGO_URI: 'mongodb://root:prisma@localhost:27018/PRISMA_DB_NAME?authSource=admin'
       TEST_FUNCTIONAL_COCKROACH_URI: 'postgresql://prisma@localhost:26257/PRISMA_DB_NAME'
       NODE_OPTIONS: '--max-old-space-size=8096'
+      JEST_JUNIT_SUITE_NAME: '${{ github.job }}/client/functional/${{ matrix.os }}/node-${{ matrix.node }}/${{ matrix.queryEngine }}'
+      JEST_JUNIT_UNIQUE_OUTPUT_NAME: true
```
millsp, line 162: Ensures that multiple runs do not overwrite each other's `junit.xml` outputs.

packages/client/tests/functional/jest.config.js
```diff
@@ -18,25 +18,28 @@ module.exports = () => {
...
   }
 
+  if (process.env['JEST_JUNIT_DISABLE'] !== 'true') {
+    configCommon.reporters.push([
+      'jest-junit',
+      {
+        addFileAttribute: 'true',
+        ancestorSeparator: ' › ',
+        classNameTemplate: (vars) => {
+          return vars.classname.replace(/ \(provider=.*?\)/g, '')
```
millsp, line 36: Remove matrix information from test name

.github/workflows/test.yml
```diff
@@ -1330,4 +1421,4 @@ jobs:
         if: always()
         with:
           name: ${{ github.job }}_${{ matrix.os }}_node-${{ matrix.node }}_${{ matrix.queryEngine }}_junit.xml
-          path: packages/*/junit.xml
+          path: packages/*/junit*.xml
```
janpio, line 1424: Why the additional `*`? Does `JEST_JUNIT_UNIQUE_OUTPUT_NAME: true` influence this?

janpio (approved):

.github/workflows/test.yml
millsp: Yes, this is to make sure we match the file that now contains an id

janpio: Then this should actually work - seems Buildpulse only uses the string inside to match the test, not the filename as I thought. Nice!

PR #17830, opened 2023-02-09
fix(client): prevent `LibraryEngine` constructor to throw on missing URLs

### LINKED ISSUE #17797: errorCode: 'P1012' PrismaClientInitializationError: error: Environment variable not found: DATABASE_URL.
### Bug description

After upgrading from 4.9.0 to 4.10.0 I am no longer able to run tests without Prisma wanting a `DATABASE_URL` environment variable it seems.

Previously to my understanding, I would generate the client and then that could be utilized, but it seems now somewhere it's trying to access the `DATABASE_URL` environment variable. The tests should be able to utilize the typing without needing to connect to a live database.

### How to reproduce

Tried to update to 4.10.0 from 4.9.0 and run tests, and they're failing with the above error.

### Expected behavior

Should be able to run tests without  a `DATABASE_URL` environment variable.

### Prisma information

Not sure that I can share those currently. Not sure if they're related.

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: v16.15.1

### Prisma Version

```
prisma                  : 4.10.0
@prisma/client          : 4.10.0
Query Engine (Node-API) : libquery-engine ca7fcef713137fa11029d519a9780db130cca91d (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli ca7fcef713137fa11029d519a9780db130cca91d (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Format Wasm             : @prisma/prisma-fmt-wasm 4.10.0-84.ca7fcef713137fa11029d519a9780db130cca91d
Default Engines Hash    : ca7fcef713137fa11029d519a9780db130cca91d
```

millsp:
closes https://github.com/prisma/prisma/issues/17797

This fix prevents the `LibraryEngine` constructor from throwing an error when the URL env var is not provided. This was the previous behavior, and was caught by silently discarding the errors on engine instantiation/getConfig. Instead, we now use a constructor parameter `ignoreEnvVarErrors` to achieve the same outcome. `BinaryEngine` engine was not affected, and `DataProxyEngine` already behaved in that same problematic way, I have not touched that in this PR (yet).

Problem: When that setting is enabled, the Query Engine does not process datasource overrides correctly anymore. Because of this some tests are failing, where we have both the env var and the datasource override.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -158,6 +158,8 @@ jobs:
       TEST_FUNCTIONAL_MONGO_URI: 'mongodb://root:prisma@localhost:27018/PRISMA_DB_NAME?authSource=admin'
       TEST_FUNCTIONAL_COCKROACH_URI: 'postgresql://prisma@localhost:26257/PRISMA_DB_NAME'
       NODE_OPTIONS: '--max-old-space-size=8096'
+      JEST_JUNIT_SUITE_NAME: '${{ github.job }}/client/functional/${{ matrix.os }}/node-${{ matrix.node }}/${{ matrix.queryEngine }}'
+      JEST_JUNIT_UNIQUE_OUTPUT_NAME: true
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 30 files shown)


millsp:
I was intending to remove the revert from https://github.com/prisma/prisma/pull/17838 but was super surprised to find that it was already gone.

aqrln (approved):
lgtm!

aqrln:
> I was intending to remove the revert from https://github.com/prisma/prisma/pull/17838 but was super surprised to find that it was already gone.

Now that I think about it, that would be strange, otherwise the regression test would've been failing failing on `main`, or the regression test has been removed as well and needs to be re-added in this PR. Let me check that.

aqrln:
Yes, I checked and it looks like all the `getConfig` stuff is still on `main`, so the revert hasn't been reverted yet. It's fine to do that either here or in a follow-up PR, no strong opinion.

millsp:
I forgot to pull to my branch, that's all I'll say 🤭

millsp:
The tests is here https://github.com/prisma/prisma/blob/dccba735a12c9cc8b62b3efdd10708d59d8a1e1e/packages/client/tests/functional/17797-no-env-error-init/tests.ts/#L8-L26

PR #17867, opened 2023-02-10
feat(migrate): print ci status

jackall3n:
Print CI status at the beginning of migrate commands, including which variable was detected. Given there's a large list of variables, this helps users investigate which exact variable has caused prisma to detect a non-interactive/CI environment.


janpio:
Is there an issue to describe the problem you are trying to solve here? That also might give us an idea if an additional log line is the right approach here - or if there might be something less impactful.

jackall3n:
> Is there an issue to describe the problem you are trying to solve here? That also might give us an idea if an additional log line is the right approach here - or if there might be something less impactful.

@janpio yes, so I personally had an issue introducing `prisma` into a project that had `BUILD_NUMBER=0` set for another package. This is one of the many variables `prisma` uses to determine if it's a CI environment which stopped me from being able to use the interactive mode to name migrations, but also to handle destructive migrations during `prisma migrate dev`. To resolve I had to go into the `node_modules/.bin/prisma` file and modify the code to work out why it thought the environment was non-interactive. This is intended to explain that instantly. Or, perhaps there's another more appropriate command that prints statuses out?

janpio:
Where did it actually prevent you from using it because of the detected non interactivity? That could be a good spot to output what made it think so - instead of all the time.

jackall3n:
@janpio the first place is while creating the migrations, if it detects the terminal is non-interactive it doesn't allow you to enter a name, so it's '<timestamp>_'. I figured I must've pressed Enter twice too fast or something. 

The second place is during MigrateDev when the database needs you to confirm potentially destructive changes to the database. e.g. Dropping columns, updating enums, and out-of-sync migration files that will end up clearing the db. 

The other areas are essentially [anywhere that uses `canPrompt`](https://github.com/prisma/prisma/search?q=canPrompt).

Originally I added this into the MigrateDev specifically in the area that I had to deal with. It felt a little thrown in, however, I can move it back. I decided to move to a status only during commands where it might be important (like the ones I've mentioned) as I thought it might be cleaner and more helpful to others who are seeing issues in those commands

janpio:
I think my wish would be that this only is output, when some functionality is changed _because_ of it. So when some prompt is skipped, you get an output why that was the case. Then it is only really visible to the people where it matters, and where it might help them understand what is going on - and I think that would already have helped you understand your problem. Agree?

I think codewise that would mean moving the call of `printCiStatus()` before or after the `canPrompt` call maybe? (I am not a developer, so please correct me if that would not work)

Jolg42:
I skipped reading everything in detail here 😅 
Suggestion: maybe it's something that could be interesting to have in `DEBUG=prisma*` logs?

Jolg42:
@jackall3n Just seeing the closing notifications.

There is this new recent command now `prisma debug`
See docs:
https://www.prisma.io/docs/orm/reference/prisma-cli-reference#debug

It doesn't say which env var triggered the CI detection, but I thought it was interesting to mention anyway.

PR #17875, opened 2023-02-10
feat(cli): Introduce multiple flags for init

### LINKED ISSUE #17737: Prisma init: allow setting generator provider value
## Problem

When you run `prisma init`, the generator value is always `prisma-client-js`. This will break if the user is using a different generator.

## Suggested solution

Add a CLI flag `--generator` or `--generator-value` to set the value programmatically.

This would even allow custom clients such as the Go client to override this for the user, so that running `prisma init` would just work for the end-user.

## Alternatives

Alternatively, custom generators could add some hack to manually replace the provider value after prisma init is executed. This is not a good solution tho.

## Additional context

https://github.com/prisma/prisma/blob/cc519a8ceefda50c842fa13400218e508373affe/packages/cli/src/Init.ts#L28

I'm happy to send a PR to add this feature if accepted!

DavidHancu:
This PR introduces these additional flags to `prisma init`:

- `--generator-provider` - Closes https://github.com/prisma/prisma/issues/17737
- `--preview-feature` - Allows the user to set preview features for the generated schema
- `--output` - Allows the user to change the generator output

It also introduces the following help entries:

``` 
Set up a new Prisma project and specify prisma-client-go as the generator provider to use
prisma init --generator-provider prisma-client-go

Set up a new Prisma project and specify x and y as the preview features to use
prisma init --preview-feature x --preview-feature y

Set up a new Prisma project and specify ./generated-client as the output path to use
prisma init --output ./generated-client
```

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -20,13 +20,22 @@ import { isError } from 'util'
 
 import { printError } from './utils/prompt/utils/print'
 
-export const defaultSchema = (provider: ConnectorType = 'postgresql') => {
+export const defaultSchema = (
+  provider: ConnectorType = 'postgresql',
+  generatorProvider: string = defaultGeneratorProvider,
+  previewFeatures: string[] = defaultPreviewFeatures,
+  output: string = defaultOutput,
+) => {
   return `// This is your Prisma schema file,
 // learn more about it in the docs: https://pris.ly/d/prisma-schema
 
 generator client {
-  provider = "prisma-client-js"
-}
+  provider = "${generatorProvider}"
+${
+  previewFeatures.length > 0
+    ? `  previewFeatures = [${previewFeatures.map((feature) => `"${feature}"`).join(', ')}]\n`
+    : ''
+}${output != defaultOutput ? `  output = "${output}"\n` : ''}}
 
 datasource db {
   provider = "${provider}"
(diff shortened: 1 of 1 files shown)


steebchen (approved):
Thanks for this, it would be great to have for the custom clients such as the Go client!

I'd probably prefer `--preview-features` and just adding it `,` separated, e.g. `--preview-features foo,bar` but that's just my 2 cents.

DavidHancu:
@steebchen 

> I'd probably prefer `--preview-features` and just adding it `,` separated, e.g. `--preview-features foo,bar` but that's just my 2 cents.

This behavior was actually requested by @janpio:

> @DavidHancu 
Should the `previewFeatures` option use a comma separated list or something else?
white_check_mark

> @janpio
I think for the --generator we chose that you can supply the same one multiple times
Again, optimally find an example how we do it somwhere else

steebchen:
Alright fair enough :D

janpio:
Prior art for the multiple CLI arguments: https://github.com/prisma/prisma/pull/16452
Slightly different as the end result here will actually be a list of items instead of multiple blocks, but as the preview features are additionally wrapped in quotes it also does not translate directly to the end result.

steebchen:
Please see #21314

Jolg42:
Closing this since https://github.com/prisma/prisma/pull/21314 was now merged

Jolg42:
Thanks to both of you @DavidHancu @steebchen ! 🙌🏼

PR #17883, opened 2023-02-11
ci: discard provider information in junit files

millsp:
Follow up on https://github.com/prisma/prisma/pull/17826 as flaky tests reports are being split by provider instead of being grouped.

### DIFF
--- a/packages/client/tests/functional/jest.config.js
+++ b/packages/client/tests/functional/jest.config.js
@@ -33,7 +33,10 @@ module.exports = () => {
         addFileAttribute: 'true',
         ancestorSeparator: ' › ',
         classNameTemplate: (vars) => {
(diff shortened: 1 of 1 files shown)


packages/client/tests/functional/jest.config.js
```diff
@@ -33,7 +33,10 @@ module.exports = () => {
         addFileAttribute: 'true',
         ancestorSeparator: ' › ',
         classNameTemplate: (vars) => {
-          return vars.classname.replace(/ \(provider=.*?\)/g, '')
+          return vars.classname
```
millsp, line 36: These replaced values prevent BuildPulse to group flaky tests by name, so we remove that info, which will be available in the errors that BuildPulse has collected.

SevInf (approved):

PR #17911, opened 2023-02-13
feat(client): JSON protocol

SevInf:
Client implementation of JSON protocol.
Sibling to prisma/prisma-engines#3624.

Built on top of `ProtocolEncoder` abstraction, introduced earlier. `JsonProtocolEncoder` would take care of encoding the message, produce `JsonProtocolMessage` instance, that will take care of batching, deserialization and debug output. 

Validation is not included, therefore, corresponding tests in both legacy and functional suite are skipped if JSON protocol is active. 

If JSON protocol is active (either through env var or preview feature), schema DMMF is not requested from the engine during the runtime (node client) and is not embedded during the generation (edge client).

For review: most of the implementation is in `core/runtime/protocol/json` directory and `engines-core` package, rest is mostly tests skipping.


packages/client/src/runtime/core/extensions/visitQueryResult.test.ts
```diff
@@ -1,43 +1,6 @@
-import { DMMF } from '@prisma/generator-helper'
```
SevInf, line 1: Moved to `utils/datamodel-builder`

packages/client/src/runtime/utils/decimalJsLike.ts
```diff
@@ -8,6 +8,7 @@ export interface DecimalJsLike {
   d: number[]
   e: number
   s: number
+  toFixed(): string
```
SevInf, line 11: This PR changes decimal-js-like definition to add `toFixed` method. I don't think it is a breaking change, since this interface is hack to accept `Decimal.JS` objects of different versions and they all contain this method.

packages/client/src/__tests__/integration/errors/client-version-error/test.ts
```diff
@@ -1,6 +1,9 @@
+import { getQueryEngineProtocol } from '@prisma/internals'
+
 import { getTestClient } from '../../../../utils/getTestClient'
 
-test('client-version-error', async () => {
+const testIf = (condition: boolean) => (condition ? test : test.skip)
+testIf(getQueryEngineProtocol() !== 'json')('client-version-error', async () => {
```
janpio, line 6: Why skip those for `json`?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -390,8 +399,9 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
         }
 
         this._baseDmmf = new BaseDMMFHelper(config.document)
+        const engineProtocol = getQueryEngineProtocol(config.generator)
 
-        if (this._dataProxy) {
+        if (this._dataProxy && engineProtocol === 'graphql') {
```
janpio, line 404: Why only for `graphql` here?

SevInf: This is the code for loading embedded DMMF schema for edge client, we don't need (and don't have) it if json protocol is used.

packages/client/src/runtime/utils/datmodelBuilder.ts
```diff
@@ -0,0 +1,55 @@
+import { DMMF } from '@prisma/generator-helper'
+
+import { BaseDMMFHelper } from '../dmmf'
+
+export function field(kind: DMMF.FieldKind, name: string, type: string, extra?: Partial<DMMF.Field>): DMMF.Field {
```
millsp, line 5: Is this a runtime util or a test util?

packages/client/src/runtime/utils/decimalJsLike.ts
millsp: I agree this is fine

packages/client/tests/functional/composites/object/update.ts
```diff
@@ -505,7 +508,7 @@ setupTestSuite(({ contentProperty }) => {
         }
       `,
       )
-    } else {
+    } else if (getQueryEngineProtocol() === 'graphql') {
```
millsp, line 511: It might be worth switching to snapshots so that we can have both errors. Not a strong opinion, I am also fine with manual work.

packages/client/tests/functional/tracing/tests.ts
```diff
@@ -768,61 +768,64 @@ testMatrix.setupTestSuite(({ provider }, suiteMeta, clientMeta) => {
     })
 
     // Different order of traces between binary and library
-    testIf(getClientEngineType() === ClientEngineType.Library)('should trace the implicit $connect call', async () => {
-      const email = faker.internet.email()
+    testIf(getClientEngineType() === ClientEngineType.Library && getQueryEngineProtocol() !== 'json')(
```
millsp, line 771: How can we unskip this test in the future?

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -909,8 +910,8 @@ You very likely have the wrong "binaryTarget" defined in the schema.prisma file.
       headers['X-transaction-id'] = interactiveTransaction.id
     }
 
-    this.currentRequestPromise = this.connection.post('/', stringifyQuery(query.query), headers)
-    this.lastQuery = query.query
+    this.currentRequestPromise = this.connection.post('/', JSON.stringify(query), headers)
+    // this.lastQuery = query.query
```
millsp, line 914: Why is this not needed anymore?  I never took the time to understand its purpose (explanations welcome :D)

packages/internals/src/client/getQueryEngineProtocol.ts
```diff
@@ -0,0 +1,17 @@
...
+  const fromEnv = process.env.PRISMA_ENGINE_PROTOCOL
+  if (fromEnv === 'json' || fromEnv == 'graphql') {
+    return fromEnv
+  }
+
+  if (fromEnv !== undefined) {
+    throw new Error(`Invalid PRISMA_ENGINE_PROTOCOL env variable value. Expected 'graphql' or 'json', got '${fromEnv}'`)
+  }
+
+  if (generatorConfig?.previewFeatures?.includes('jsonProtocol')) {
```
millsp, line 13: Do we have a simple functional test testing this path?

millsp (approved):
Amazing work, I especially liked the care you took to write unit tests.
I left some comments, most of which are optional/questions.
(Review/presentation happened on call)

packages/internals/src/client/getQueryEngineProtocol.ts
SevInf: I don't think it is possible with a current setup, unfortunately - they way it is implemented in the engine is `PRISMA_ENGINE_PROTOCOL` env var always trumps preview feature. So, we won't be able to test it on CI because it is either `PRISMA_ENGINE_PROTOCOL` is always set. I can add a unit test though.

packages/client/tests/functional/tracing/tests.ts
SevInf: I think in this case we can just remove the snapshot - test asserts on the order of specific spans anyway.

packages/client/tests/functional/composites/object/update.ts
SevInf: Since we are going to do validation really soon (if stars align correctly, already in 4.11 but most likely in 4.12) I've decided to skip them for now and unskip/come up with other way of testing it when we'll work on the messages.

packages/client/src/runtime/utils/datmodelBuilder.ts
SevInf: Test util. I was not sure where to put it. Same `runtime/utils` folders contains `getTestClient` and friends as well, but we can create a new `testUtils` folder somewhere and move them there.

packages/engine-core/src/binary/BinaryEngine.ts
SevInf: Oops, i think it was not intentional, let me see.

packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -467,9 +469,9 @@ You may have to run ${chalk.greenBright('prisma generate')} for your changes to
     { traceparent, interactiveTransaction }: RequestOptions<undefined>,
   ): Promise<{ data: T; elapsed: number }> {
     debug(`sending request, this.libraryStarted: ${this.libraryStarted}`)
-    const request: QueryEngineRequest = { query: query.query, variables: {} }
+    // const request: QueryEngineRequest = { query: query.query, variables: {} }
```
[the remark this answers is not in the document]
SevInf: Removed.

packages/engine-core/src/binary/BinaryEngine.ts
SevInf: Restored that part. As far as I can tell, it is used only in case of panic for appending `Engine Query` section to automatic github issue.  At the same time, i think, sanitiation function is buggy have been sanitising it down to `{ 
"X": {} } for as long as it existed (2 years, according to git blame). So, maybe we don't need this part at all?

packages/client/tests/functional/tracing/tests.ts
SevInf: Did just that, now test is unskipped (for binary engine as well)

packages/engine-core/src/binary/BinaryEngine.ts
janpio: Optimally we would have a test that caught this...

PR #17921, opened 2023-02-14
test(client): next.js and not found issues

millsp:
The first goal of this PR is to add tests for Next.js and not found issues. Tests are failing and it's normal until I enable the workaround. In order to support running these tests fast enough, I migrated away from docker compose to just use docker containers for running the tests (+ other improvements).


packages/client/tests/e2e/_utils/run.ts
```diff
@@ -84,20 +91,63 @@ async function main() {
 
   console.log('🐳 Starting tests in docker')
   // tarball was created, ready to send it to docker and begin e2e tests
-  const testNames = args._.join(' ').replace(/\//g, '-')
-  await $`docker compose -f ${__dirname}/docker-compose.yml down --remove-orphans`
-  await $`docker compose -f ${__dirname}/docker-compose.yml build ${testNames}`
-  await $`docker compose -f ${__dirname}/docker-compose.yml up ${testNames}`
+  const testStepFiles = await glob('../**/_steps.ts', { cwd: __dirname })
```
millsp, line 94: Removing docker compose to run tests in favor of just running docker containers, as it is faster and also allows to run in band while also easily get exit codes.

packages/client/tests/e2e/_utils/standard.cmd.sh
```diff
@@ -1,19 +1,35 @@
-#!/bin/sh
+#!/bin/bash
 
+# Database Env Vars
+export PRISMA_DB_NAME=$(cat /dev/urandom | LC_ALL=C tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
```
millsp, line 4:
```suggestion
export PRISMA_DB_NAME=$(cat /dev/urandom | tr -dc 'a-z0-9' | head -c 8)
```

packages/client/tests/e2e/_utils/standard.cmd.sh
```diff
@@ -1,19 +1,35 @@
-#!/bin/sh
+#!/bin/bash
 
+# Database Env Vars
+export PRISMA_DB_NAME=$(cat /dev/urandom | tr -dc 'a-z0-9' | head -c 8)
+export TEST_E2E_POSTGRES_URI="postgres://prisma:prisma@172.17.0.1:5432/${PRISMA_DB_NAME}"
```
millsp, line 5: `172.17.0.1` is the default docker bridge address, to easily access the dbs exposed on localhost

millsp:
@SocketSecurity ignore vercel@28.13.0
@SocketSecurity ignore esbuild@0.14.47
@SocketSecurity ignore esbuild@0.15.13
@SocketSecurity ignore esbuild@0.17.5
@SocketSecurity ignore next@13.1.4

packages/client/tests/e2e/nextjs-schema-not-found/10_monorepo-serverComponents-customOutput-noReExport/_steps.ts
```diff
@@ -0,0 +1,28 @@
...
+  setup: async () => {
+    await $`pnpm install`
+    cd('packages/service')
+    await $`pnpm exec prisma db push --force-reset`
+    await $`pnpm exec next build`
+  },
+  test: async () => {
+    await $`rm -fr .next/standalone/node_modules/next`
+    nextJsProcess = $`node .next/standalone/server.js`
+    await $`sleep 5`
```
SevInf, line 16: Will it be possible to something more reliable? Like, for example, checking that the `3000` ports response to requests?

packages/client/tests/e2e/_utils/docker-compose-clean.yml
```diff
@@ -2,14 +2,25 @@ version: '3.7'
 
 services:
   # not an actual service, just a helper to clean up the workspace
-  clean:
+  pre-clean:
```
SevInf, line 5: Question: why are we running it through docker? As far as I can tell, it cleans stuff only in mounts, so we could probably just run same commands on the host. Or am I missing something?

packages/client/tests/e2e/_utils/run.ts
```diff
@@ -84,20 +91,63 @@ async function main() {
...
+  let e2eTestNames = testStepFiles.map((p) => path.relative('..', path.dirname(p)))
+
+  if (args._.length > 0) {
+    e2eTestNames = e2eTestNames.filter((p) => args._.some((a) => p.includes(a)))
+  }
+
+  const dockerVolumes = [
+    `${path.resolve(__dirname, '..')}:/e2e`,
+    `${path.resolve(__dirname, '..', '..', '..')}:/client`,
+    `${path.resolve(__dirname, '..', '..', '..', '..', '..')}:/repo`,
```
SevInf, line 104: Nit: I would've found it a little bit easier to understand/remember what directories it refers to if we did something like this:

```ts
const monorepoRoot = path.resolve(__dirname, '..', '..', '..', '..', '..');
const dockerVolumes = [
   `${monorepoRoot}:/repo`
   `${path.join(monorepoRoot, 'packages', 'client')}:/client`
   `${path.join(monorepoRoot, 'packages', 'client', 'tests', 'e2e')}:/e2e`
]
```

packages/client/tests/e2e/_utils/run.ts
```diff
@@ -84,20 +91,63 @@ async function main() {
...
+    `${path.resolve(__dirname, '..', '.cache')}:/root/.cache`,
```
SevInf, line 105: Not for now, put idea for later. We could pre-build a docker base image with pre-populated cache, so all installs would be faster, not only e2e.

packages/client/tests/e2e/_utils/standard.cmd.sh
```diff
@@ -1,19 +1,35 @@
-#!/bin/sh
+#!/bin/bash
```
SevInf, line 1:
```suggestion
#!/bin/bash
set -eio pipefail
```

I prefer to have this in most of bash scripts for safety & less painful debugging. [Explanation](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/).

Would also allow to skip explicit `exit` at the end of this script, as pipe and individual command errors would be autoamtically proopogated.

packages/client/tests/e2e/_utils/run.ts
```diff
@@ -84,20 +91,63 @@ async function main() {
...
+  }
+
+  const failedJobResults = jobResults.filter((r) => r.exitCode !== 0)
+  const passedJobResults = jobResults.filter((r) => r.exitCode === 0)
+
+  if (args['--verbose'] === true) {
+    for (const result of failedJobResults) {
+      console.log(`🛑 ${result.name} failed with exit code`, result.exitCode)
+      await $`cat ${path.resolve(__dirname, '..', result.name, 'LOGS.txt')}`
+      await sleep(50)
```
SevInf, line 138: Could you add a comment here about why this sleep is necessary?

SevInf (approved):

packages/client/tests/e2e/_utils/docker-compose-clean.yml
millsp: On Linux docker, the created files are created with `root` user. I don't think mac users have this inconvenience, though. I could change the user in the container so that files are created with the correct user, but that would bring other issues. So I went for an universal approach.

packages/client/tests/e2e/_utils/run.ts
millsp: That's a good point, I'll change.

packages/client/tests/e2e/_utils/run.ts
millsp: It was a strange behavior on CI, that I do not fully understand which would prevent proper log output. I think it has something to do with zx, but didn't get the time to debug it. I'll leave a comment.

packages/client/tests/e2e/nextjs-schema-not-found/10_monorepo-serverComponents-customOutput-noReExport/_steps.ts
millsp: Yes I think that is a good idea, it will also shorten the delay most likely. I think something like is is possible with `curl`.

packages/client/tests/e2e/_utils/docker-compose-clean.yml
millsp: Updated the comment

millsp:
I added all your review suggestions and made an improvement to be able to share code across tests.

PR #17950, opened 2023-02-16
feat: capture reason for Wasm panics in CLI

jkomyno:
TypeScript implementation of https://github.com/prisma/prisma-private/issues/212.
Companion of the Rust PR https://github.com/prisma/prisma-engines/pull/3710 (see its description for a thorough example).

This PR fixes the Wasm limitation of not transmitting the reason and code location of a Panic by explicitly sending this panic info to a globally available JS object, `globalThis.WASM_PANIC_REGISTRY`.

### DIFF
--- a/packages/internals/src/WasmPanicRegistry.ts
+++ b/packages/internals/src/WasmPanicRegistry.ts
@@ -1,0 +1,12 @@
+export class WasmPanicRegistry {
+  #message = ''
+
+  get() {
+    return `${this.#message}`
+  }
+
+  // Don't use this method directly, it's only used by the Wasm panic hook in @prisma/prisma-fmt-wasm.
+  private set_message(value: string) {
+    this.#message = `RuntimeError: ${value}`
+  }
+}
--- a/packages/internals/src/WasmPanicRegistry.ts
+++ b/packages/internals/src/WasmPanicRegistry.ts
@@ -1,12 +1,12 @@
 export class WasmPanicRegistry {
-  #message = ''
+  private message = ''
 
   get() {
-    return `${this.#message}`
+    return `${this.message}`
   }
 
   // Don't use this method directly, it's only used by the Wasm panic hook in @prisma/prisma-fmt-wasm.
   private set_message(value: string) {
-    this.#message = `RuntimeError: ${value}`
+    this.message = `RuntimeError: ${value}`
   }
 }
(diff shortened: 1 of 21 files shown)


packages/internals/typings/global.d.ts
```diff
@@ -0,0 +1,6 @@
+import { WasmPanicRegistry } from '../src/WasmPanicRegistry'
+
+declare global {
+  /// Global registry for Wasm panics.
+  var WASM_PANIC_REGISTRY: WasmPanicRegistry
```
jkomyno, line 5:
```suggestion
  // eslint-disable-next-line no-var
  var WASM_PANIC_REGISTRY: WasmPanicRegistry
```

packages/internals/src/WasmPanicRegistry.ts
```diff
@@ -0,0 +1,12 @@
+export class WasmPanicRegistry {
+  #message = ''
```
aqrln, line 2: It may not be currently safe to use private fields, depending on whether they end up in typings: https://prisma-company.slack.com/archives/CUXLS0Z6K/p1675164034311419. Could be okay here though. Better to run ecosystem-tests before merging to make sure.

packages/internals/src/wasm.ts
```diff
@@ -7,3 +8,10 @@ const { dependencies } = require('../package.json')
 export { prismaFmt }
 // e.g. 4.3.0-18.a39215673171b87177b86233206a5d65f2558857
 export const prismaFmtVersion: string = dependencies['@prisma/prisma-fmt-wasm']
+
+/**
+ * Set up a global registry for Wasm panics.
+ * This allows us to retrieve the panic message from the Wasm panic hook,
+ * which is not possible otherwise.
+ */
+globalThis.WASM_PANIC_REGISTRY = new WasmPanicRegistry()
```
aqrln, line 17: I'm not very comfortable with this name as it's too generic and could collide with something with an incompatible API if end users or maintainers of another library take a similar approach in solving the same problem. How about `PRISMA_WASM_PANIC_REGISTRY`?

Since this would need to be changed on the Rust side as well, it's totally okay to do that in follow-up PRs if we want to merge this one ASAP to unbreak `main`, but it's important that we do it before the release.

jkomyno: Agree, let's do this before the release, but after this PR.

packages/internals/src/WasmPanicRegistry.ts
jkomyno: I've simply removed the `#` prefix and replaced it with `private`.

aqrln (approved):

packages/internals/src/wasm.ts
jkomyno: Follow-up PR for Rust: https://github.com/prisma/prisma-engines/pull/3722

PR #17972, opened 2023-02-17
feat: Add client startup benchmarking script

SevInf:
Improved version of https://github.com/prisma/schema-size-bench

For now, is not hooked into CI and has to be called manually. 
Usage and capabilities are documented in README. Allows to benchmark any recent enough prisma version and compare the versions. Separate PRs to the engine and client will add more tracing spans, necessary for performance inventory.


packages/client-benchmark/src/generateClient.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  for (let i = 0; i < numRelations; i++) {
+    addRelation(randomElement(models), randomElement(models), `rel${i}`)
+  }
+
+  return [getHeader(features)].concat(models.map((m) => m.stringify())).join('\n')
+}
+
+function randomElement(array) {
+  return array[Math.floor(Math.random() * array.length)]
```
aqrln, line 90: I think the last element will never be chosen using this formula

aqrln (commented):
I’ll continue reviewing later, amazing work 💯

packages/client-benchmark/src/measurments.ts
```diff
@@ -0,0 +1,77 @@
+import execa from 'execa'
```
aqrln, line 1: typo in the file name: `mv src/measur{,e}ments.ts`

packages/client-benchmark/src/generateClient.ts
SevInf: No, i think this is correct:

1. `Math.random()` is always in [0;1) range.
2. `Math.random() * array.length` therefore is in in [0;array.length) range
3. `Math.floor(Math.random() * array.length)` is then always in [0; array.length - 1] (inclusive) range.

aqrln: ah yes, you are right, sorry

packages/client-benchmark/src/measurementsScripts/otel.ts
```diff
@@ -0,0 +1,48 @@
+import { ExportResult, ExportResultCode, hrTimeToMicroseconds } from '@opentelemetry/core'
+import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http'
```
janpio, line 2:
```suggestion
// import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http'
```

packages/client-benchmark/README.md
```diff
@@ -0,0 +1,83 @@
...
+**Note**: `tracing` preview feature is always enabled as the benchmark itself relies on it.
+
+It is also possible to combine `--features` and `--baseline-features` with `--prisma-version` and `--baseline-version` flags.
+
+## Usage with AWS Lambda
+
+Benchmark script can also be used to measure performance on AWS lambda. This requires some additional configurations:
+
+1. Get AWS Credentials as described [here](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html). The role you are using should
+   be allows to create, delete and invoke lambda functions.
```
millsp, line 79:
```suggestion
   be allowed to create, delete and invoke lambda functions.
```

packages/client-benchmark/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "arg": "5.0.2",
+    "aws-crt": "^1.15.14",
+    "csv-stringify": "^6.2.4",
+    "execa": "5.1.1",
+    "fast-glob": "3.2.12"
+  },
+  "scripts": {
+    "dev": "DEV=true node -r esbuild-register helpers/build.ts",
+    "build": "node -r esbuild-register helpers/build.ts",
+    "measure": "node dist/measureCli.js",
```
millsp, line 35: Would it make sense to delay the GC for as much as we can here to avoid spikes in the data?

packages/client-benchmark/src/targets/local.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  async afterPnpmInstall(workbenchPath: string): Promise<void> {
+    await fs.copyFile(path.join(__dirname, 'measureLocal.mjs'), path.join(workbenchPath, 'measureLocal.mjs'))
+  }
+
+  async afterClientGeneration() {}
+
+  async measure(workbenchPath: string): Promise<Record<string, number>> {
+    const { stdout } = await execa('node', [path.join(workbenchPath, 'measureLocal.mjs')], {
```
millsp, line 19: Would it make sense to delay the GC for as much as we can here to avoid spikes in the data?

Jolg42:
@SevInf Is this still needed? Asking because this PR is old.

millsp:
(I still use it from time to time to validate some changes I make)

wmadden:
Hi @SevInf, thanks for your work here. I'm closing this PR for now because it's been a long time since it had any attention but we may return to it later and incorporate it into our automated benchmarks.

PR #17979, opened 2023-02-18
ci: only trigger buildpulse steps on schedule

millsp:
Since we are way over the quotas, BuildPulse is currently discarding a lot of the received test results, which defeats the purpose of tracking flaky tests. Since tracking flaky tests on PRs isn't strictly necessary, and since we have a scheduled workflow every day just for that, we can disable BuildPulse on PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -223,7 +223,7 @@ jobs:
           secret: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
 
       - uses: actions/upload-artifact@v3
-        if: ${{ github.event_name == 'schedule' }}
+        if: github.event_name == 'schedule'
         with:
           name: ${{ github.job }}_${{ matrix.os }}_node-${{ matrix.node }}_dataproxy_junit.xml
           path: packages/*/junit*.xml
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -223,7 +223,7 @@ jobs:
           secret: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
 
       - uses: actions/upload-artifact@v3
-        if: always()
+        if: ${{ github.event_name == 'schedule' }}
```
janpio, line 226: Are you sure this is correct? The other `if` aboe is without the `{{ .... }}` and I can never remember the difference and meaning.

millsp: https://docs.github.com/en/actions/learn-github-actions/expressions

millsp: If I read correctly, the one [above](https://github.com/prisma/prisma/pull/17979/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R216) looks like it would be evaluated as a string?

janpio: The docs confuse me even more, each time: 
image

millsp: Ok, that settles the debate: both work

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression. For more information about if conditionals, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)."

millsp: Best not to have mentioned it, ever :D

millsp: Fixed it.

janpio (approved):

jkomyno (approved):

PR #17999, opened 2023-02-20
feat(get-platform): test specific libssl paths parsing

### LINKED ISSUE #18001: (internal): add unit tests for path-specific libssl version parsing
In an effort to identify the libssl version present in a system - required for downloading the appropriate query-engine compilation target - we run a pipeline of steps:

1. `libssl-specific-path`: we look for `libssl.so*` files in known system paths, which vary depending on the system configuration
2. `ldconfig`: if the above failed, we look at the `LD_CONFIG` cache for references of `libssl` libraries
3. `openssl-binary`: if the above failed, we look at the result of `openssl version -v`
4. if the above failed, we give up and we fall back to the default `libssl-1.1.x` version.

The `libssl-specific-path` strategy would be the easiest to test in a reproducible fashion via `jestContext`'s fixtures, but that'd require some refactoring in `@prisma/get-platform`.

jkomyno:
Closes https://github.com/prisma/prisma/issues/18001.

This PR extracts the `libsslSpecificPaths` generation from `getSSLVersion` to the newly created function `computeLibSSLSpecificPaths`.

This not only simplified the definition of `getSSLVersion`, but also enabled unit testing the runtime execution of `libsslSpecificCommands` (which runs `ls -v "libssl.so.0*" ${path} | grep libssl.so | grep -v "libssl.so.0"`) in different scenarios.

Moreover, a couple old non-impactful typos / redundancies were identified and fixed.

### DIFF
--- a/packages/get-platform/src/getPlatform.ts
+++ b/packages/get-platform/src/getPlatform.ts
@@ -89,8 +89,8 @@ export async function getos(): Promise<GetOSResult> {
 
   const distroInfo = await resolveDistro()
   const archFromUname = await getArchFromUname()
-
-  const libssl = await getSSLVersion({ arch, archFromUname, familyDistro: distroInfo.familyDistro })
+  const libsslSpecificPaths = computeLibSSLSpecificPaths({ arch, archFromUname, familyDistro: distroInfo.familyDistro })
+  const { libssl } = await getSSLVersion(libsslSpecificPaths)
 
   return {
     platform: 'linux',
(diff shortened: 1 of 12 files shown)


packages/get-platform/src/getPlatform.ts
```diff
@@ -359,38 +371,29 @@ export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetO
...
-
-  const openSSLVersion: string | undefined = await getFirstSuccessfulExec(['openssl version -v'])
-
-  if (openSSLVersion) {
-    const matchedVersion = parseOpenSSLVersion(openSSLVersion)
-    if (matchedVersion) {
-      return matchedVersion
+      return { libssl: openSSLVersion, strategy: 'openssl-binary' }
     }
   }
```
jkomyno, line 392: This block here was duplicated. Perhaps something that escaped our attention during a merge conflict resolution.

aqrln (approved):

PR #18000, opened 2023-02-20
feat(cli): Views Introspection Warnings

### LINKED ISSUE #17578: `db pull`: add new codes for introspection warnings for views in the CLI
New codes were added after https://github.com/prisma/prisma-engines/commit/61863c7b1810c243ede42ae8ed40dc849b38c193

See https://github.com/prisma/prisma-engines/blob/main/introspection-engine/connectors/sql-introspection-connector/src/warnings/generators.rs#L405-L453

Codes 21, 22, 23, 24, 25, 26

They need to be added in
https://github.com/prisma/prisma/blob/main/packages/migrate/src/types.ts
and handled in
https://github.com/prisma/prisma/blob/main/packages/migrate/src/commands/DbPull.ts

Druue:
Added new warnings for views introspection:
- 21: IntrospectionWarningsUnsupportedTypesInViews
- 22: IntrospectionWarningEnrichedWithMapOnFieldInViews
- 23: IntrospectionWarningEnrichedWithMapOnView
- 24: IntrospectionWarningViewsWithoutIdentifier
- 25: IntrospectionWarningEnrichedWithCustomPrimaryKeyNamesInViews
- 26: IntrospectionWarningsFieldsWithEmptyNamesInViews

closes #17578


Druue:
hm, we seem to lose type information here when using includes. but also I don't see an easy way to just add the annotation manually. Do we want to update the imports to add that annotation? or should I manually check equality on all of the codes in this case
image

Druue:
## Warning Code 21
```sql
create table reservations(
  id int primary key,
  room varchar not null,
  dates daterange not null 
);

create view res as select id, room, dates from reservations;
```

add to your `schema.prisma` -- otherwise the following warning message will be obfuscated by a warning that the view has no PK
```prisma
view res {
  id    Int @id
}
```

then running db pull will yield:
```prisma
view res {
  id    Int                       @id
  room  String?                   @db.VarChar
  dates Unsupported("daterange")?
}
```

and the following output in the cli
image

Druue:
## Warning Code 22
```sql
CREATE VIEW "A" AS SELECT 1 AS id;
```

running db pull will yield:
```prisma
/// The underlying view does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
view A {
  id Int?

  @@ignore
}
```

and the following output in the cli
image

Druue:
## Warning Code 23
```sql
CREATE VIEW "if" AS SELECT 1 AS id;
```

running db pull will yield:
```prisma
/// This view has been renamed to 'Renamedif' during introspection, because the original name 'if' is reserved.
/// The underlying view does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
view Renamedif {
  id Int @id

  @@map("if")
  @@ignore
}
```

and the following output in the cli -- note this cli output is obfuscated by the lack of id attribute so one would first need to manually add the following to their schema to get this output.
```prisma
view Renamedif {
  id Int @id

  @@map("if")
}
```
image

Druue:
## Warning Code 24
```sql
CREATE TABLE "User" (
  id SERIAL PRIMARY KEY,
  first_name VARCHAR(255) NOT NULL,
  last_name VARCHAR(255) NULL
);

CREATE VIEW "Schwuser" AS
  SELECT id, first_name, last_name FROM "User";
```

running db pull will yield:
```prisma
model User {
  id         Int     @id @default(autoincrement())
  first_name String  @db.VarChar(255)
  last_name  String? @db.VarChar(255)
}

/// The underlying view does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
view Schwuser {
  id         Int?
  first_name String? @db.VarChar(255)
  last_name  String? @db.VarChar(255)

  @@ignore
}
```

and the following output in the cli
image

Druue:
## Warning Code 25
```sql
CREATE VIEW "B" AS SELECT 1 AS a, 2 AS b;
```

running db pull will yield:
```prisma
/// The underlying view does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
view B {
  a Int?
  b Int?

  @@ignore
}
```

and the following output in the cli -- note this cli output is obfuscated by the lack of id attribute so one would first need to manually add the following to their schema to get this output.
```prisma
view B {
  a Int
  b Int

  @@id([a, b], name: "kekw")
}
```
image

Druue:
## Warning Code 26
```sql
CREATE VIEW "A" AS SELECT 2 AS foo, 1 AS "1";
```

running db pull will yield:
```prisma
/// The underlying view does not contain a valid unique identifier and can therefore currently not be handled by the Prisma Client.
view C {
  foo Int?

  /// This field was commented out because of an invalid name. Please provide a valid one that matches [a-zA-Z][a-zA-Z0-9_]*
  // 1 Int? @map("1")
  @@ignore
}
```

and the following output in the cli
image

Druue (commented):
Note: only code 24 and 26 show up during _introspection_
the rest (barring 22) only show up during _re-introspection_, this is currently as expected per internal discussion with @pimeys

Druue:
22 is now tracked by test :)
image

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+    const setupParams = computeSetupParams(warningCode)
+
+    beforeAll(async () => {
+      await tearDownPostgres(setupParams!).catch((e) => {
+        console.error(e)
+      })
+    })
+
+    beforeEach(async () => {
+      await setupPostgres(setupParams!).catch((e) => {
```
SevInf, line 516: Are we sure we want to silence the error and proceed if setup fails? Chances that any of the tests will succeed in this case are pretty slim, imho.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+    }
+    return setupParams
+  }
+
+  describe('postgresql views 21', () => {
+    const warningCode = 21
+    const setupParams = computeSetupParams(warningCode)
+
+    beforeAll(async () => {
+      await tearDownPostgres(setupParams!).catch((e) => {
```
SevInf, line 510: I don't think we'll need this teardown here, as it should also happen in `afterEach` for all the tests.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
 })
 
+describe('postgresql views re-introspection warnings', () => {
+  const connectionString = process.env.TEST_POSTGRES_URI_MIGRATE!.replace('tests-migrate', 'tests-migrate-db-pull')
+
+  function computeSetupParams(warningCode: EngineArgs.ViewWarningCodes) {
+    const setupParams: SetupParams = {
+      connectionString,
+      // Note: at this location there is a setup.sql file
+      // which will be executed a SQL file so the database is not empty
```
janpio, line 491:
```suggestion
      // which will be executed so the database is not empty
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
   })
 })
 
+describe('postgresql views re-introspection warnings', () => {
+  const connectionString = process.env.TEST_POSTGRES_URI_MIGRATE!.replace('tests-migrate', 'tests-migrate-db-pull')
+
+  function computeSetupParams(warningCode: EngineArgs.ViewWarningCodes) {
+    const setupParams: SetupParams = {
+      connectionString,
+      // Note: at this location there is a setup.sql file
```
janpio, line 490:
```suggestion
      // Note: dirname points to a location with a setup.sql file
```

packages/migrate/src/__tests__/DbPull.test.ts
janpio, line 491:
```suggestion
      // Note: dirname points to a location with a setup.sql file
      // which will be executed to prepare the database with the correct tables, views etc.
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+      await setupPostgres(setupParams!).catch((e) => {
+        console.error(e)
+      })
+      // Back to original env vars
+      process.env = { ...originalEnv }
+      // Update env var because it's the one that is used in the schemas tested
+      process.env.TEST_POSTGRES_URI_MIGRATE = connectionString
+    })
+
+    afterEach(async () => {
```
SevInf, line 525: Feel free to ignore the suggestion if you disagree, but i think this setup is repeated for all of the tests? In that case, it could also be split into helpers:

```ts
function setupPostgressForWarining(warningCode: EngineArgs.ViewWarningCodes) {
  const setupParams = computeSetupParams(warningCode)
  
  beforeEach(async () => {
      await setupPostgres(setupParams!).catch((e) => {
        console.error(e)
      })
      // Back to original env vars
      process.env = { ...originalEnv }
      // Update env var because it's the one that is used in the schemas tested
      process.env.TEST_POSTGRES_URI_MIGRATE = connectionString
    })

    afterEach(async () => {
      // Back to original env vars
      process.env = { ...originalEnv }
      await tearDownPostgres(setupParams!).catch((e) => {
        console.error(e)
      })
    })
}
```

packages/migrate/src/__tests__/DbPull.test.ts
SevInf, line 516:
```suggestion
      await setupPostgres(setupParams).catch((e) => {
```

I don't think `setupParams` are nullable.

packages/migrate/src/types.ts
```diff
@@ -322,6 +370,14 @@ export namespace EngineArgs {
     affected: AffectedModelOrCompositeTypeAndField[]
   }
 
+  export type ViewWarningCodes =
```
SevInf, line 373: Unless I am missing something, this type is used only for test, so I would've moved it there.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+        'fixtures',
+        're-introspection',
+        'postgresql',
+        `views-warning-${warningCode}`,
+      ),
+    }
+    return setupParams
+  }
+
+  describe('postgresql views 21', () => {
```
janpio, line 505:
```suggestion
  describe('postgresql views 21 - unsupported types', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                // - View "res", Field: "dates", Type: "daterange"
+                                                                                                // 
+                                                                        `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  // TODO :: can't generate the correct code
+  describe('postgresql views 22', () => {
```
janpio, line 582:
```suggestion
  describe('postgresql views 22 - @map', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+        // These fields were enriched with \`@map\` information taken from the previous Prisma schema.
+        // - View "A", Field: "id",
+        // 
+      `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 23', () => {
```
janpio, line 651:
```suggestion
  describe('postgresql views 23 - @@map', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                // These views were enriched with \`@@map\` information taken from the previous Prisma schema.
+                                                // - View "Renamedif"
+                                                // 
+                                    `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 24', () => {
```
janpio, line 722:
```suggestion
  describe('postgresql views 24 - no unique identifier', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                        // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                        // - View "Schwuser"
+                                        // 
+                              `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 25', () => {
```
janpio, line 798:
```suggestion
  describe('postgresql views 25 - @@id name', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                // These views were enriched with custom compound id names taken from the previous Prisma schema.
+                                // - View "B"
+                                // 
+                        `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 26', () => {
```
janpio, line 870:
```suggestion
  describe('postgresql views 26 - invalid field name', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
   })
 })
 
+describe('postgresql views re-introspection warnings', () => {
```
janpio, line 484: It's unexpected/surprising that these come before plain Introspection to me.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+        'fixtures',
+        'introspection',
+        'postgresql',
+        `views-warning-${warningCode}`,
+      ),
+    }
+    return setupParams
+  }
+
+  describe('postgresql views 21', () => {
```
janpio, line 969:
```suggestion
  describe('postgresql views 21 - unsupported type (+ no unique identifier, because view)', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                                        // - View "res", Field: "dates", Type: "daterange"
+                                                                                                                                        // 
+                                                                                                      `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  // TODO :: can't generate the correct code
+  describe('postgresql views 22', () => {
```
janpio, line 1052:
```suggestion
  describe('postgresql views 22 - no unique identifier', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                                // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                                // - View "A"
+                                                                                                                                // 
+                                                                                                `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 23', () => {
```
janpio, line 1124:
```suggestion
  describe('postgresql views 23 - automated rename (+ no unique identifier, because view)', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                        // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                        // - View "Renamedif"
+                                                                                                                        // 
+                                                                                          `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 24', () => {
```
janpio, line 1198: What happens here that is specific to code 24?

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                // - View "Schwuser"
+                                                                                                                // 
+                                                                                    `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 25', () => {
```
janpio, line 1274: What happens here that is specific to code 25?

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                // - View "B"
+                                                                                                                // 
+                                                                                    `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  describe('postgresql views 26', () => {
```
janpio, line 1347:
```suggestion
  describe('postgresql views 26 - invalid field name (+ no unique identifier, because view)', () => {
```

packages/migrate/src/__tests__/fixtures/introspection/postgresql/views-warning-26/setup.sql
```diff
@@ -0,0 +1,5 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: I think this should already be resolved, right @pimeys?

packages/migrate/src/__tests__/fixtures/re-introspection/postgresql/views-warning-22/setup.sql
```diff
@@ -0,0 +1,6 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: see other comment

packages/migrate/src/__tests__/fixtures/re-introspection/postgresql/views-warning-23/setup.sql
```diff
@@ -0,0 +1,6 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: see other comment

packages/migrate/src/__tests__/fixtures/re-introspection/postgresql/views-warning-25/setup.sql
```diff
@@ -0,0 +1,6 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: see other comment

packages/migrate/src/__tests__/fixtures/re-introspection/postgresql/views-warning-26/setup.sql
```diff
@@ -0,0 +1,6 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: see other comment

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                // These fields are not supported by the Prisma Client, because Prisma currently does not support their types.
+                                                                                                // - View "res", Field: "dates", Type: "daterange"
+                                                                                                // 
+                                                                        `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  // TODO :: can't generate the correct code
```
janpio, line 581: ??

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,947 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                                        // These fields are not supported by the Prisma Client, because Prisma currently does not support their types.
+                                                                                                                                        // - View "res", Field: "dates", Type: "daterange"
+                                                                                                                                        // 
+                                                                                                      `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
+  // TODO :: can't generate the correct code
```
janpio, line 1051: ??

packages/migrate/src/__tests__/fixtures/introspection/postgresql/views-warning-22/setup.sql
```diff
@@ -0,0 +1,5 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: see other comment

packages/migrate/src/__tests__/fixtures/introspection/postgresql/views-warning-23/setup.sql
```diff
@@ -0,0 +1,5 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: see other comment

packages/migrate/src/__tests__/fixtures/introspection/postgresql/views-warning-25/setup.sql
```diff
@@ -0,0 +1,5 @@
+-- TODO: remove tmp table once we no longer receive error P4001
```
janpio, line 1: see other comment

janpio (approved):
Looks good!

Left suggestion on naming the different tests to explain what the codes actually mean.

packages/migrate/src/__tests__/DbPull.test.ts
Druue: This is what @jkomyno recommended when we paired on this 😅 Didn't think to ask about it at the time

packages/migrate/src/__tests__/DbPull.test.ts
Druue: Ah, this can be removed now 😅

packages/migrate/src/__tests__/DbPull.test.ts
janpio: Maybe doesn't matter - it just took me a bit of scrolling to understand that the first big block is Re-Introspection (with existing schema file), and the second one is normal Introspection.

packages/migrate/src/__tests__/DbPull.test.ts
Druue: Currently nothing, the original intent was to figure out what was happening with the cli warnings only emitting code 24 for introspection but given that it's intended for spam reasons (and in alignment with models), this could be removable now. Same goes for the introspecting 24 test

janpio: I do not understand what "spam reasons" means. 24 and 25 have to do _something_ special that the others do not, or it does not make sense that they exist.

packages/migrate/src/__tests__/fixtures/introspection/postgresql/views-warning-26/setup.sql
Druue: Yep, seems to be!

packages/migrate/src/__tests__/DbPull.test.ts
Druue: Yep, absolutely agree!

packages/migrate/src/__tests__/DbPull.test.ts
Druue: As far as I can tell, they aren't, not sure what the reasoning was for this

packages/migrate/src/__tests__/DbPull.test.ts
Druue: Ah, that we don't emit too many warnings at once basically :)

packages/migrate/src/types.ts
Druue: The reasoning for this was that the underlying types _aren't_ exported but I'm not sure if that's something that we can just update or if we should leave as is

SevInf: Ah, ok then, I think in that case it's fine to keep it where it is.

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
+    afterEach(async () => {
+      // Back to original env vars
+      process.env = { ...originalEnv }
+      await tearDownPostgres(setupParams).catch((e) => {
+        console.error(e)
+      })
+    })
+  }
+
+  describe('postgresql views 21 - unsupported types', () => {
```
janpio, line 527:
```suggestion
  describe('postgresql views 21 - unsupported types (both view and model [code 3])', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
```
janpio, line 852:
```suggestion
  describe('postgresql views 21 - unsupported types (both view and model [code 3])', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                                                        // 
+                                                                                                                                                        // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                                                        // - View "A"
+                                                                                                                                                        // 
+                                                                                                                  `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
```
janpio, line 952:
```suggestion
  // code 22 does not make sense in Introspection, as it requires a schema to pick up the prior @map
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
+
+  describe('postgresql views 23 - view with @@map', () => {
```
janpio, line 954:
```suggestion
  describe('postgresql views 23 - rename view with @@map', () => {
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                                                        // 
+                                                                                                                                                        // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                                                        // - View "A"
+                                                                                                                                                        // 
+                                                                                                                  `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
+
```
janpio, line 953:
```suggestion

// 23 is only re-introspection for keeping @@map from previous schema file

```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                                                // 
+                                                                                                                                                // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                                                // - View "Renamedif"
+                                                                                                                                                // 
+                                                                                                            `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
```
janpio, line 998: Remove this temporarily while @pimeys actually adds a new code for this. 
```suggestion
```

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
+
+  describe('postgresql views 24 - no unique identifier', () => {
```
janpio, line 1000: TODO: Also test with multiple views to see that it does a list instead of 2 entries

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,669 @@ describe('postgresql - missing database', () => {
...
+                                                                                                                                        // *** WARNING ***
+                                                                                                                                        // 
+                                                                                                                                        // The following views were ignored as they do not have a valid unique identifier or id. This is currently not supported by the Prisma Client.
+                                                                                                                                        // - View "B"
+                                                                                                                                        // 
+                                                                                                      `)
+      expect(ctx.mocked['process.stdout.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+      expect(ctx.mocked['process.stderr.write'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    })
+  })
```
janpio, line 1095:
```suggestion

// 25 is re-introspection only to keep @@id names from previous schema 

```

packages/migrate/src/__tests__/DbPull.test.ts
Druue: Hm, I'm less familiar as to what alternatives we have / what would be more idiomatic/helpful here

SevInf: I'd just drop that `.catch`:
- if `setupPostgress` throws exception, we would still see it.
- if we could not setup postgress because of exception, running the rest of the tests is likely pointless - they won't find the database they'd expect.

Druue: Would you recommend the same thing then for the `afterEach` `tearDown`?

SevInf (approved):

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -480,6 +481,601 @@ describe('postgresql - missing database', () => {
...
+        'fixtures',
+        're-introspection',
+        'postgresql',
+        `views-warning-${warningCode}`,
+      ),
+    }
+    return setupParams
+  }
+
+  function setupPostgressForWarning(warningCode: EngineArgs.ViewWarningCodes) {
```
janpio, line 505: Too late, but `Postress` 😆

PR #18006, opened 2023-02-20
feat(client): Add internal tracing spans for performance testing

SevInf:
Allows us to have internal spans, similar to the ones we have in the
engine. Internal spans are not shown unless PRISMA_SHOW_ALL_TRACES=true
env var is set. This matches engine behavoir.

Adds a bunch of internal spans:
- operations around dmmf (getting, parsing, processing).
- schema upload time for data proxy.

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -1114,8 +1114,14 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
 
     _getDmmf = callOnce(async (params: Pick<InternalRequestParams, 'clientMethod' | 'callsite'>) => {
       try {
-        const dmmf = await this._engine.getDmmf()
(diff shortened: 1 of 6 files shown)


packages/engine-core/src/common/Engine.ts
```diff
@@ -63,7 +63,7 @@ export abstract class Engine<InteractiveTransactionPayload = unknown> {
   abstract on(event: EngineEventType, listener: (args?: any) => any): void
   abstract start(): Promise<void>
   abstract stop(): Promise<void>
-  abstract getDmmf(): Promise<DMMF.Document>
+  abstract getDmmf(traceparent?: string): Promise<DMMF.Document>
```
millsp, line 66: Could also retrieve it directly in the engine instead of passing it?

millsp (approved):

packages/engine-core/src/common/Engine.ts
SevInf: Done!

PR #18010, opened 2023-02-20
fix(client): improve schema not found errors

millsp:
This PR improves error handling for schemas not being found. It will point Next.js users to a documentation page where they can learn why this is happening and how they can fix it (via the workaround). The tests have been updated to check that the error message appears as expected, and also accommodate a common error testing logic for both server and non-sever components (see comments in `_shared/test.ts`).

Needs https://github.com/prisma/prisma/pull/17921
Needs https://github.com/prisma/docs/pull/4521
https://github.com/prisma/prisma/pull/18010

### DIFF
--- a/packages/client/tests/e2e/nextjs-schema-not-found/10_monorepo-serverComponents-customOutput-noReExport/_steps.ts
+++ b/packages/client/tests/e2e/nextjs-schema-not-found/10_monorepo-serverComponents-customOutput-noReExport/_steps.ts
@@ -8,7 +8,6 @@ void executeSteps({
     await $`pnpm install`
     cd('packages/service')
     await $`pnpm exec prisma db push --force-reset`
-    await $`pnpm exec next build`
   },
   test: async () => {
     await testServerComponents()
--- a/packages/client/tests/e2e/nextjs-schema-not-found/_shared/test.ts
+++ b/packages/client/tests/e2e/nextjs-schema-not-found/_shared/test.ts
@@ -5,6 +5,7 @@ import { $ } from 'zx'
  * @param endpoint The endpoint to test
  */
 async function test(endpoint: string) {
+  await $`pnpm exec next build`
   await $`rm -fr .next/standalone/node_modules/next`
   const nextJsProcess = $`node .next/standalone/server.js`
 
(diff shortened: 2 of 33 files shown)


SevInf (approved):

packages/client/tests/e2e/nextjs-schema-not-found/10_monorepo-serverComponents-customOutput-noReExport/_steps.ts
```diff
@@ -8,7 +8,6 @@ void executeSteps({
     await $`pnpm install`
     cd('packages/service')
     await $`pnpm exec prisma db push --force-reset`
-    await $`pnpm exec next build`
```
janpio, line 11: Why could all these be removed?

janpio: Found it in `packages/client/tests/e2e/nextjs-schema-not-found/_shared/test.ts` 👍

packages/client/tests/e2e/nextjs-schema-not-found/_shared/test.ts
```diff
@@ -1,30 +1,70 @@
...
  */
-async function test(endpoint: string) {
-  await $`rm -fr .next/standalone/node_modules/next`
-  const nextJsProcess = $`node .next/standalone/server.js`
+async function test(endpoint: string, serverComponents: boolean) {
+  console.log(`Testing ${endpoint} with WORKAROUND=${process.env.WORKAROUND}`)
 
+  // prepare and start the next.js server
+  const nextJsBuild = await $`pnpm exec next build`.nothrow()
+  await $`rm -fr .next/standalone/node_modules/next`.nothrow()
```
janpio, line 17: This is a workaround for an existing Next.js bug, right? We should link there to explicitly document that here.

packages/engine-core/src/library/LibraryEngine.ts
```diff
@@ -100,7 +100,21 @@ export class LibraryEngine extends Engine<undefined> {
   constructor(config: EngineConfig, loader: LibraryLoader = new DefaultLibraryLoader(config)) {
     super()
 
-    this.datamodel = fs.readFileSync(config.datamodelPath, 'utf-8')
+    try {
+      // we try to handle the case where the datamodel is not found
+      this.datamodel = fs.readFileSync(config.datamodelPath, 'utf-8')
+    } catch (e) {
+      if ((e.stack as string).match(/\/\.next|\/next@|\/next\//)) {
```
janpio, line 107: Nice one.

packages/client/tests/e2e/nextjs-schema-not-found/_shared/test.ts
millsp: I guess maybe I should try to start it this way https://nextjs.org/docs/api-reference/cli#production

Issue #17900, opened 2023-02-12
Don't print platform detection warnings when custom binaries are provided


aqrln:
Currently the output on NixOS (e.g., `nix-shell -p nodePackages.prisma --run "prisma --version"`) is too noisy:

```
prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".
Please manually install OpenSSL and try installing Prisma again.
Warning Precompiled engine files are not available for nixos.
prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".
Please manually install OpenSSL and try installing Prisma again.
prisma                  : 4.10.1
@prisma/client          : Not found
Current platform        : linux-nixos
Query Engine (Node-API) : libquery-engine  (at ../../../nix/store/sfbq6v19zqjakvwj9w8vz97wsqcv7q61-prisma-engines-4.10.1/lib/libquery_engine.node, resolved by PRISMA_QUERY_ENGINE_LIBRARY)
Migration Engine        : migration-engine-cli  (at ../../../nix/store/sfbq6v19zqjakvwj9w8vz97wsqcv7q61-prisma-engines-4.10.1/bin/migration-engine, resolved by PRISMA_MIGRATION_ENGINE_BINARY)
Format Wasm             : @prisma/prisma-fmt-wasm 4.10.1-1.80b351cc7c06d352abe81be19b8a89e9c6b7c110
Default Engines Hash    : ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5
Studio                  : 0.481.0
```

Also, an identical warning about OpenSSL is printed twice.

When custom binaries are provided, these warnings become less useful at best or misleading at worst, so it's best to skip them.

VanCoding:
I'm having the same issue. The warning indeed makes no sense when using custom binaries.

Gerschtli:
Is someone working on this issue? Otherwise, I would have a look :)

Gerschtli:
Proposed fix to this in this PR https://github.com/prisma/prisma/pull/20138 :)

aqrln:
Was going to reopen since providing the env vars doesn't skip the OpenSSL warning yet — but that's something that would be better requested in a separate issue (especially now that we technically do have another way to to skip the OpenSSL warning).

Jolg42:
[...]

PR #18012, opened 2023-02-21
feat: search for openssl in OPENSSL_LIB_DIR if provided

hlolli:
This is my first PR here and I approved the prisma CLA.

The reason for this change is that I want to use the native openssl from my unsupported linux distro.
As a compilation environment variable, OPENSSL_LIB_DIR is used when provided in prisma-engines.
It would be very nice if this could be supported on runtime as well. I don't know if there's already a way
to pass in custom openssl lib-dir location (or direct location of the shared library). If there's a way, then
obviously I'll take this PR no further :)

### DIFF
--- a/packages/get-platform/src/getPlatform.ts
+++ b/packages/get-platform/src/getPlatform.ts
@@ -299,27 +299,34 @@ type GetOpenSSLVersionParams = {
  * This function never throws.
  */
 export async function getSSLVersion(args: GetOpenSSLVersionParams): Promise<GetOsResultLinux['libssl'] | undefined> {
-  const libsslSpecificPaths = match(args)
-    .with({ familyDistro: 'musl' }, () => {
-      /* Linux Alpine */
-      debug('Trying platform-specific paths for "alpine"')
-      return ['/lib']
-    })
-    .with({ familyDistro: 'debian' }, ({ archFromUname }) => {
-      /* Linux Debian, Ubuntu, etc */
-      debug('Trying platform-specific paths for "debian" (and "ubuntu")')
-      return [`/usr/lib/${archFromUname}-linux-gnu`, `/lib/${archFromUname}-linux-gnu`]
-    })
-    .with({ familyDistro: 'rhel' }, () => {
-      /* Linux Red Hat, OpenSuse etc */
-      debug('Trying platform-specific paths for "rhel"')
-      return ['/lib64', '/usr/lib64']
-    })
-    .otherwise(({ familyDistro, arch, archFromUname }) => {
-      /* Other Linux distros, we don't do anything specific and fall back to the next blocks */
-      debug(`Don't know any platform-specific paths for "${familyDistro}" on ${arch} (${archFromUname})`)
-      return []
-    })
+  let libsslSpecificPaths: string[]
+
+  if (process.env.OPENSSL_LIB_DIR) {
+    debug(`Found 'OPENSSL_LIB_DIR' path specified in env-vars, ignoring platform specific lib-dir paths...`)
+    libsslSpecificPaths = [process.env.OPENSSL_LIB_DIR]
+  } else {
+    libsslSpecificPaths = match(args)
+      .with({ familyDistro: 'musl' }, () => {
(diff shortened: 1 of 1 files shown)


hlolli:
It seems this line here, tough unrelated, prevents this from succeeding

```
 (path) => `ls -v "libssl.so.0*" ${path} | grep libssl.so | ${excludeLibssl0x}`,
```

I tried both on osx and linux and

```
ls -v "libssl.so.0*" ${path}
```

isn't something that's a valid ls command?

also piping grep into grep without xargs also doesn't work, so I wonder if this part of the function is helpful for what it's intended to do?

jkomyno:
Hi @hlolli, thanks for your PR! Before proceeding further, may you please tell us which Linux distro you're trying to support, which libssl version you have installed, and which CPU architecture it uses?
Moreover, could you please print a screenshot of what would be the content of your folder matching this new env var `OPENSSL_LIB_DIR`?

Thanks :)

hlolli:
hi @jkomyno 

I want to add that prisma still works, but I believe it uses the debian node pre-compiled binary, actually, I'm surprised it works.

The contents of openssl when I use v3 (which I currently do since prisma support it) is

```
engines-3/  libcrypto.so@  libcrypto.so.3*  libssl.so@  libssl.so.3*  ossl-modules/
```

The location of this folder is different depending if I'm playing steam (then it's /lib64), or building locally, then it's /home/hlolli/result/lib/ or globally /nix/store/openssl/lib

Personally I use osx, we deploy using nixos on our CI machine

```
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         36 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  8
  On-line CPU(s) list:   0-7
Vendor ID:               GenuineIntel
  Model name:            Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz
  
  $ uname -a
Linux kudos 5.15.93 #1-NixOS SMP Thu Feb 9 10:26:48 UTC 2023 x86_64 GNU/Linux
```

we deploy to raw docker (non os base-image) on AWS-ECS cluster, I don't have an easy way to ssh into those, but the results from our ci system is 1:1 reporducible.

I wonder if this whole check could be skipped if native binaries are used, because they themselves dynamically link against openssl?



Something like https://www.npmjs.com/package/elfinfo could be used to track down the location if it's necessary. And would be more reliable imo than hardcoded distro-specific library paths.

aqrln:
> Something like https://www.npmjs.com/package/elfinfo could be used to track down the location if it's necessary. And would be more reliable imo than hardcoded distro-specific library paths.

Outside of nix world, normally the full paths to the libraries are not specified in runpath, so it wouldn't be possible to rely on that on other distros. Besides, we'd need to have the binary the read the runpath in the first place, while these checks exist to know which binary to download.

However, these checks are basically meaningless when custom binaries are used, and the printed warnings could even be misleading. I opened an issue about NixOS previously: https://github.com/prisma/prisma/issues/17900. I wonder if we could even skip platform detection completely and not just skip printing warnings as that issue suggests.

Gerschtli:
What is the status of this PR? I would love to get rid of this warning and provide the correct path to openssl via nix directly.

Gerschtli:
Because this PR is quite old and some things changed in prisma, I created another PR https://github.com/prisma/prisma/pull/20138 :)

aqrln:
Thank you for your PR @hlolli and sorry we couldn't bring it to the finish line. We have now merged https://github.com/prisma/prisma/pull/20381 which is a more generic alternative to this idea, see https://github.com/prisma/prisma/pull/20138#issuecomment-1640379179 for the motivation and for the reasoning about why we decided against a specific env var just for OpenSSL.

Also see https://github.com/prisma/prisma/pull/20138#pullrequestreview-1551812556 for the overview of the current state of NixOS user experience (kudos to @Gerschtli for the patches!) and possible future work.

I'll go ahead and close this PR, please let me know if there's still something relevant that wasn't fixed in other way yet.

PR #18077, opened 2023-02-24
feat(client): add support for raw queries in extensions

### LINKED ISSUE #18027: PCE: Support query extensions for raw queries
Prisma Client extensions don't support extending raw queries. `queryRaw`, `executeRaw`, `queryRawUnsafe`, `executeRawUnsafe`, or `$runCommandRaw` currently don't work at runtime and at type-level.

The following API expands the capabilities of the `query` components to enable that:

```ts
  query: {
    $queryRaw({ args, query, operation }) {
      return query(args)
    },
    $executeRaw({ args, query, operation }) {
      return query(args)
    },
    $queryRawUnsafe({ args, query, operation }) {
      return query(args)
    },
    $executeRawUnsafe({ args, query, operation }) {
      return query(args)
    },
  },
})
```

millsp:
closes https://github.com/prisma/prisma/issues/18027


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -688,7 +583,7 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
      */
     $executeRawUnsafe(query: string, ...values: RawValue[]) {
       return createPrismaPromise((transaction) => {
-        return this.$executeRawInternal(transaction, query, ...values)
+        return this.$executeRawInternal(transaction, '$executeRaw', [query, ...values])
```
millsp, line 586:
```suggestion
        return this.$executeRawInternal(transaction, '$executeRawUnsafe', [query, ...values])
```
And add a test

aqrln, line 586: should it be `$executeRawUnsafe`?

aqrln:
```suggestion
        return this.$executeRawInternal(transaction, '$executeRawUnsafe', [query, ...values])
```

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -557,100 +535,17 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
       return this._request({
-        args,
-        clientMethod: '$executeRaw',
-        dataPath: [],
         action: 'executeRaw',
-        callsite: getCallSite(this._errorFormat),
+        args,
         transaction,
+        clientMethod,
+        argsMapper: rawQueryArgsMapper(this, clientMethod),
```
aqrln, line 546: nice, feels cleaner to have it in the `argsMapper`

packages/client/src/runtime/core/model/UserArgs.ts
```diff
@@ -1,8 +1,4 @@
 /**
  * Input that flows from the user into the Client.
  */
-export type UserArgs = {
-  [K in string]: UserArgsProp | UserArgsProp[]
-}
-
-type UserArgsProp = UserArgs | string | number | boolean | bigint | null | undefined
+export type UserArgs = any
```
aqrln, line 4: We can now also change the `args` type in `QueryMiddlewareParams` (where `InternalRequestParams` inherits it from) to explicit `UserArgs` that `argsMapper` expects. The previous too narrow type would somewhat conflict with how args were used in `runCommandRaw` method, but now might be a good idea to use an explicit type even if it's aliased to `any` so if we have a good reason to narrow it down, we can do it in one place and see the effects of doing so.

millsp: I tried hard to keep it type safe but failed because of variance issues, and as such `unknown` isn't a good fit either. I left a comment about it.

millsp: I applied your suggestion, and also moved `UserArgs` to another folder.

packages/client/src/runtime/getPrismaClient.ts
millsp: Fixed and added a test

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -95,8 +95,49 @@ function clientExtensionsQueryDefinition(this: PrismaClientClass) {
...
+
+    return `$${n}`
+  })
+
+  const argsResultMap = {
+    $executeRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
+    $queryRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
+    $executeRawUnsafe: { args: '[query: string, ...values: any[]]', result: 'any' },
+    $queryRawUnsafe: { args: '[query: string, ...values: any[]]', result: 'any' },
+    $runCommandRaw: { args: 'Prisma.InputJsonObject', result: 'Prisma.JsonObject' },
```
aqrln, line 114: I wonder if `Prisma.InputJsonObject` isn't too narrow here. I believe things like dates and buffers should be allowed there but aren't allowed by `InputJsonObject`?

aqrln: the type matches the API without extensions, so out of scope
https://github.com/prisma/prisma/blob/1e0e86795c1d73e5d109c44addb289a411d72e15/packages/client/src/generation/TSClient/PrismaClient.ts/#L439

aqrln (approved):

PR #18081, opened 2023-02-24
fix(client): Exclude source maps from publishing

SevInf:
They are not used by default, will not be usefull to the most users and
increase `@prisma/client` size unnecessary. In the future, we'd build a
separate package for them so we'll have an option to use them when we
need it, but for now, we'd just exclude them from npm package.


aqrln (approved):

aqrln:
Some tests are failing with:

```
    ENOENT: no such file or directory, lstat '/home/ubuntu/actions-runner/_work/prisma/prisma/packages/client/!runtime/*.map'

      56 |   for (const file of [...pkgFiles]) {
      57 |     const dest = import_path.default.join(tmpDir, import_path.default.basename(file));
    > 58 |     (0, import_fs_extra.copySync)(file, dest, { recursive: true, overwrite: true });
         |                                ^
      59 |   }
      60 |   return import_path.default.join(tmpDir);
      61 | }
```

looks like `getPackedPackage` from `@prisma/internals` needs to be adjusted to support or ignore negative patterns.

millsp:
I recently wrote a similar logic here https://github.com/prisma/prisma/blob/ccbb0bb6046ef474e2654da8421a885d3f825779/packages/cli/helpers/copy-prisma-client.ts/#L1-L24

PR #18118, opened 2023-02-27
fix(client): Make json protocol work with data proxy

SevInf:
1. If JSON protocol is used, request was not serialized correctly and
   did not match either of 2 exepected bodies
2. Data proxy tests were not running on CI for JSON protocol, fixes that
   as well.

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -24,6 +24,7 @@ import {
   getClientEngineType,
   getQueryEngineProtocol,
   logger,
+  QueryEngineProtocol,
   tryLoadEnvs,
   warnOnce,
 } from '@prisma/internals'
@@ -245,6 +246,13 @@ export interface GetPrismaClientConfig {
   injectableEdgeEnv?: LoadedEnv
 
   /**
+   * Engine protocol to use within edge runtime. Passed
+   * through config because edge client can not read env variables
+   * @remarks only used for the purpose of data proxy
+   */
+  edgeClientProtocol?: QueryEngineProtocol
+
+  /**
    * The contents of the datasource url saved in a string.
    * This can either be an env var name or connection string.
    * It is needed by the client to connect to the Data Proxy.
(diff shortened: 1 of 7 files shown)


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -369,7 +377,10 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
         }
 
         this._baseDmmf = new BaseDMMFHelper(config.document)
-        const engineProtocol = getQueryEngineProtocol(config.generator)
+        const engineProtocol = NODE_CLIENT
```
millsp, line 380: Could we pass `config` to `getQueryEngineProtocol` and move this logic in there along with the other logic?

millsp: (In theory, NODE_CLIENT should be able to work there too, I think)

SevInf: `getQueryEngineProtocol` lives in `@prisma/internals` and is not build with `NODE_CLIENT` defined.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -244,6 +245,13 @@ export interface GetPrismaClientConfig {
    */
   injectableEdgeEnv?: LoadedEnv
 
+  /**
+   * Engine protocol to use within edge runtime. Passed
+   * through config because edge client can not read env variables
+   * @remarks only used for the purpose of data proxy
+   */
+  edgeClientProtocol?: QueryEngineProtocol
```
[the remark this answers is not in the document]
SevInf: It was not exactly silent, it was just that we were not running data proxy/edge client tests on JSON protocol. As soon we enabled that, this started failing pretty loudly.

packages/client/src/runtime/getPrismaClient.ts
SevInf: `getQueryEngineProtocol` lives in `@prisma/internals`. Wouldn't using `NODE_CLIENT` there mean we'd have to build a separate edge version of internals?

millsp: Somehow I thought that esbuild would inject across the whole bundle, but that does not work.

millsp (approved):
✨

packages/client/src/runtime/getPrismaClient.ts
millsp: Otherwise, I guess you could pass the `NODE_CLIENT` value as a param

PR #18121, opened 2023-02-27
fix(migrate): add stacktrace to `migration-engine` panics

jkomyno:
This PR:
- adds an actual stacktrace to the Rust panics 
- prepends an identifier to the stacktrace of the two RustPanics possibly thrown in `MigrateEngine.ts`, to tell them apart:
  - `[EXIT_PANIC]` for when a panic occurs on the `exit` event of a child process
  - `[RESPONSE_ERROR_PANIC]` for when a panic occurs after a response is received from the engine

Example of a """stacktrace""" before:

```
Starting migration engine RPC server
The migration was not applied because it triggered warnings and the force flag was not passed
```

Example of a stacktrace after:

```
[EXIT_PANIC]
Starting migration engine RPC server
The migration was not applied because it triggered warnings and the force flag was not passed
   0: backtrace::capture::Backtrace::new
   1: migration_engine::set_panic_hook::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic_handler::{{closure}}
   4: std::sys_common::backtrace::__rust_end_short_backtrace
   5: _rust_begin_unwind
   6: core::panicking::panic_fmt
   7: <migration_core::state::EngineState as migration_core::api::GenericApi>::debug_panic::{{closure}}
   8: <core::pin::Pin<P> as core::future::future::Future>::poll
   9: <futures_util::future::future::Then<Fut1,Fut2,F> as core::future::future::Future>::poll
  10: json_rpc_stdio::handle_stdin_next_line::{{closure}}::{{closure}}
  11: tokio::runtime::task::raw::poll
  12: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  13: tokio::runtime::task::raw::poll
  14: tokio::runtime::task::UnownedTask<S>::run
  15: std::sys_common::backtrace::__rust_begin_short_backtrace
  16: core::ops::function::FnOnce::call_once{{vtable.shim}}
  17: std::sys::unix::thread::Thread::new::thread_start
  18: __pthread_deallocate
```

**NOTE**: the stacktrace isn't publicly visible to the user.


packages/migrate/src/__tests__/handlePanic.migrate.test.ts
```diff
@@ -176,13 +176,16 @@ describe('handlePanic migrate', () => {
...
+      const error = e as RustPanic
+
       expect(error).toMatchSnapshot()
       expect(JSON.parse(JSON.stringify(error))).toMatchObject({
         area: 'LIFT_CLI',
         schemaPath,
       })
-      expect(error.rustStack).toContain('This is the debugPanic artificial panic')
+      expect(error.message).toContain('This is the debugPanic artificial panic')
+      expect(error.rustStack).toBeTruthy()
```
janpio, line 188: Can we snapshot this as well?

jkomyno: That would yield different results on every OS, I can perhaps snapshot a subset containing `std::panicking` or something like that

packages/migrate/src/MigrateEngine.ts
```diff
@@ -351,9 +351,10 @@ export class MigrateEngine {
             this.rejectAll(err)
             reject(err)
           }
-          const engineMessage = this.lastError?.message || this.messages.join('\n')
+          const fallbackMessage = this.messages.join('\n')
+          const engineMessage = this.lastError?.message || fallbackMessage
```
millsp, line 355:
```suggestion
          const stackTrace = this.lastError?.message || fallbackMessage
```

packages/migrate/src/MigrateEngine.ts
```diff
@@ -351,9 +351,10 @@ export class MigrateEngine {
...
           const handlePanic = () => {
-            const stackTrace = this.messages.join('\n')
+            const stackTrace = this.lastError?.backtrace || fallbackMessage
```
millsp, line 357:
```suggestion
```

packages/migrate/src/__tests__/handlePanic.migrate.test.ts
```diff
@@ -176,13 +168,20 @@ describe('handlePanic migrate', () => {
...
       expect(JSON.parse(JSON.stringify(error))).toMatchObject({
         area: 'LIFT_CLI',
         schemaPath,
       })
-      expect(error.rustStack).toContain('This is the debugPanic artificial panic')
+      expect(error.message).toContain('This is the debugPanic artificial panic')
+
+      const isWindows = ['win32'].includes(process.platform)
+      if (!isWindows) {
+        expect(error.rustStack).toContain('std::panicking::')
```
millsp, line 183: Why is windows excluded from the test?

packages/migrate/src/MigrateEngine.ts
millsp, line 357: Does the backtrace include the fallback message?

Do you have an example of output before and after this change, removing `fallbackMessage` is some cases and giving priority to `backtrace`?

packages/migrate/src/__tests__/handlePanic.migrate.test.ts
jkomyno: Because Rust panics on Windows are something like

```
<unknown>
<unknown>
<unknown>
...
```

packages/migrate/src/MigrateEngine.ts
jkomyno: `backtrace` contains the actual stacktrace produced by the Rust panics. Example:

```text
           0: backtrace::capture::Backtrace::new
           1: migration_engine::set_panic_hook::{{closure}}
           2: std::panicking::rust_panic_with_hook
           3: std::panicking::begin_panic_handler::{{closure}}
           4: std::sys_common::backtrace::__rust_end_short_backtrace
           5: _rust_begin_unwind
           6: core::panicking::panic_fmt
           7: <migration_core::state::EngineState as migration_core::api::GenericApi>::debug_panic::{{closure}}
           8: <core::pin::Pin<P> as core::future::future::Future>::poll
           9: <futures_util::future::future::Then<Fut1,Fut2,F> as core::future::future::Future>::poll
          10: json_rpc_stdio::handle_stdin_next_line::{{closure}}::{{closure}}
          11: tokio::runtime::task::raw::poll
          12: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
          13: tokio::runtime::task::raw::poll
          14: tokio::runtime::task::UnownedTask<S>::run
          15: std::sys_common::backtrace::__rust_begin_short_backtrace
          16: core::ops::function::FnOnce::call_once{{vtable.shim}}
          17: std::sys::unix::thread::Thread::new::thread_start
          18: __pthread_deallocate
```

packages/migrate/src/__tests__/handlePanic.migrate.test.ts
millsp: Does that mean this is going to show up for windows users or are we using the fallback message just for windows users?

packages/migrate/src/MigrateEngine.ts
jkomyno: In most scenarios found in our error reporting dashboard, `fallbackMessage` is an empty string, but I've found an example where it contains a couple of logs from Rust:

```
Starting migration engine RPC server
The migration was not applied because it triggered warnings and the force flag was not passed.
[/Users/runner/.cargo/git/checkouts/quaint-9f01e008b9a89c14/1550f6c/src/connector/result_set/result_row.rs:59:64] index out of bounds: the len is 1 but the index is 1
```

millsp: Do you think it makes sense then to concatenate the two so we have at least the same previous output? Or does that not matter at all in this case?

jkomyno: Considering this discovery, a better alternative would be including both `this.messages.join('\n')` and `this.lastError?.backtrace` in `RustPanic.stacktrace`.

packages/migrate/src/__tests__/handlePanic.migrate.test.ts
millsp: And is there a no way to have rust stacktraces on windows?

jkomyno: Rust stacktraces are known problems on Windows unfortunately. With the "fallbackMessage" (notice that this part of the code changed throughout the night), it would look something like this:

```
[EXIT_PANIC]
${fallbackMessage}
   <unknown>
  <unknown>
  <unknown>
  ...
  <unknown>
```

As specified in the PR description, this is not user-facing, it's just useful for our error reporting system.

millsp: > As specified in the PR description, this is not user-facing, it's just useful for our error reporting system.
That seems to have been added after I started my review :P Good to know!

millsp (approved):

PR #18128, opened 2023-02-28
test(client): do not resolve dataproxy url in tests

millsp:
A previous PR removed network calls to resolve the data-proxy version in functional tests, but did not work in edge clients because `process.env` isn't accessible there (since it's polyfilled). Instead I switched to rely on the generated client stored engine version which is `0000000000000000000000000000000000000000` only in functional in tests. This should fix the remaining flakyness.


millsp:
(Merge after release)

Issue #18148, opened 2023-03-01
jsonProtocol doesn't query enum fields by default


n3oney:
### Bug description

When querying a table that has an enum field with the jsonProtocol feature enabled, the data won't include the enum field by default.

### How to reproduce

1. Enable the `jsonProtocol` preview feature
2. Query some table `await prisma.db.tableName.findMany();`

### Expected behavior

The returned data will be the same as without the jsonProtocol feature

With jsonProtocol
```js
const products = await prisma.db.productInfo.findMany();

/*
{
  "id": "01002ac5-0098-443f-84ac-3294b3169aae",
  "name": "Product one",
  "description": "Product number one",
}
*/
```

Without jsonProtocol
```js
const products = await prisma.db.productInfo.findMany();

/*
{
  "id": "01002ac5-0098-443f-84ac-3294b3169aae",
  "name": "Product one",
  "status": "Enabled",
  "description": "Product number one",
}
*/
```

Adding a `select` to the query and explicitly telling it to select that field seems to fix it.

With jsonProtocol
```js
const products = await prisma.db.productInfo.findMany({
      select: {
        status: true,
        id: true,
        name: true,
        description: true
      }
});


/*
{
  "id": "01002ac5-0098-443f-84ac-3294b3169aae",
  "name": "Product one",
  "status": "Enabled",
  "description": "Product number one",
}
*/
```

### Prisma information

```prisma
generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["jsonProtocol"]
}

datasource db {
  provider = "cockroachdb"
  url      = env("DATABASE_URL")
}

model ProductInfo {
  id                String             @id(map: "primary") @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  status            Status
  name              String
  description       String

  @@map("product_info")
}

enum Status {
  Enabled
  Disabled
  Hidden

  @@map("status")
}
```

### Environment & setup

- OS: Arch Linux
- Database: CockroachDB, with Data Proxy
- Node.js version: v19.7.0

### Prisma Version

```
4.11.0
```

revmischa:
Confirmed this is happening for me after upgrading from 4.10.1 to 4.11.0

revmischa:
Thanks for fixing this quickly! I would really love to be able to try out the rest of 4.11 so if there is any way you can release a patch version with this fix that would be quite nice

janpio:
[...]

PR #18163, opened 2023-03-01
test(client): Test default selection contents

SevInf:
Should help to prevent issues like #18148

### DIFF
--- a/packages/client/tests/functional/default-selection/tests.ts
+++ b/packages/client/tests/functional/default-selection/tests.ts
@@ -1,0 +1,74 @@
+import { expectTypeOf } from 'expect-type'
+
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { Prisma as PrismaNamespace, PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(({ provider }) => {
+  beforeAll(async () => {
+    const input: Partial<PrismaNamespace.ModelCreateInput> = {
+      value: 'Foo',
+      relation: {
+        create: {},
+      },
+    }
+
+    if (provider !== 'sqlite' && provider !== 'sqlserver') {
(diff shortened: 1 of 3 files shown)


packages/client/tests/functional/default-selection/tests.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    if (provider !== 'sqlite' && provider !== 'sqlserver') {
+      // @ts-test-if: provider !== 'sqlite' && provider !== 'sqlserver'
+      input.enum = 'A'
+    }
+
+    if (provider === 'postgresql' || provider === 'cockroachdb' || provider === 'mongodb') {
+      // @ts-test-if: provider === 'postgresql' || provider === 'cockroachdb' || provider === 'mongodb'
+      input.list = ['Hello', 'world']
+      // @ts-test-if: provider === 'postgresql' || provider === 'cockroachdb' || provider === 'mongodb'
```
aqrln, line 26: would it be worth it to extract this to a variable (e.g. `supportsScalarLists`)?

aqrln (approved):

packages/client/tests/functional/default-selection/tests.ts
jkomyno: I agree with @aqrln. Let's turn line 23 into something like `if (supportsScalarsList.includes(['postgresql', 'cockroachdb', 'mongodb'])) { /* */ }`

SevInf: Variables won't work for `@ts-test-if` - it is evaluated before test is run using  matrix only. So they'll still be duplicated for type tests. With that in mind, would you still prefer variable at least for runtime conditions?

aqrln: > With that in mind, would you still prefer variable at least for runtime conditions?

Either is fine with me if the runtime condition is duplicated too, but I'd probably prefer it the way it currently is then.

PR #18258, opened 2023-03-08
feat(cli): Export types without following @map mutations

### LINKED ISSUE #16908: Export prisma's types without following `@map` mutations
## Problem

In some cases, we have to use the `queryRaw(sqlRequest)` function to use an advanced sql feature for instance. When we use this function, the returned type is `any` because prisma can't know what we are actually doing. 

One solution to keep typesafe is to cast the returned object with a prisma's type like this:
```ts
 const users = await prisma.queryRaw<User[]>('SELECT * from users')
```
The `users` variable will be well typed.

But what if we have a `@map` in our schema.prisma? For example we have this line :
```prisma
phoneNumber String? @map("phone_number")
```

We could not use this type because the query will return a `phone_number` field instead of `phoneNumber`

## Suggested solution

That's why it would be great if we could import from Prisma a type which don't follow `@map` of our schema.prisma

## Alternatives

One solution to this problem is to create our own type with right fields renamed:
```ts
type Rename<T, K extends keyof T, N extends string> = Pick<T, Exclude<keyof T, K>> & { [P in N]: T[K] }
type UserWithoutMap = Rename<User, 'phoneNumber', 'phone_number'
 ```

I would be happy to contribute to this project, so let me know if it sounds a good idea 🚀

Kalo0m:
close https://github.com/prisma/prisma/issues/16908 

I wanted to contribute to prisma's codebase because I love your work and here is my first contribution! Let me know if I have to change something. I did not find where to write the test so tell me if I have to!

Moreover, I have used the property dbName there `this.useDbTypes && field.dbName ? field.dbName : field.name` but the typescript field is `dbNames`. I guess there is an error in the Field interface because everywhere else it is `dbName`.

https://github.com/prisma/prisma/blob/main/packages/generator-helper/src/dmmf.ts#:~:text=export%20interface%20Field%20%7B


wmadden:
Hi @Kalo0m,

Thank you for your contribution and the effort you’ve put into this pull request. After careful consideration, we’ve decided not to proceed with this change, as it could lead to inconsistencies between database field names and application-level types, potentially compromising type safety. As noted by @janpio in [Issue #16908](https://github.com/prisma/prisma/issues/16908), this approach might not align with Prisma’s design principles. Therefore, we’re closing this PR. We appreciate your understanding and encourage you to participate in future discussions or contributions that align with our project’s direction.

PR #18305, opened 2023-03-13
feat(codspeed.io): Try out benchmarking cloud service

janpio:


aqrln (approved):

packages/client/package.json
```diff
@@ -149,6 +149,7 @@
     }
   },
   "dependencies": {
+    "@codspeed/benchmark.js-plugin": "^1.0.1",
```
aqrln, line 152: This must be a `devDependency`

aqrln (changes requested):

packages/client/package.json
```diff
@@ -149,6 +149,7 @@
     }
   },
   "dependencies": {
+    "@codspeed/benchmark.js-plugin": "1.0.1",
```
aqrln, line 152: somehow still in `dependencies` although the commit message says it was moved

janpio: This is what it looks like when I mess up :D

aqrln (approved):

aqrln:
Let's merge and try this!

PR #18318, opened 2023-03-14
ci: test Cirun.io

janpio:


aktech:
Idle runners are available now @janpio https://docs.cirun.io/reference/idle-runners

PR #18372, opened 2023-03-17
fix(tests): fix flaky "generatorHandler > nonexistent executable"

aqrln:
The test sometimes fails on CI with:

```
Error: expect(received).rejects.toThrow()

Received promise resolved instead of rejected
Resolved to value: undefined
```

There are two sources of flakiness in this test:

1. The `setTimeout` timer that waits for `initWaitTime` after the
   process spawns to see if it will exit with an error shortly after
   starting must only be set after the process actually
   starts (`child.on('spawn')` event), otherwise there's a chance that
   the timer fires before the process even finishes spawning.

2. We don't spawn the executable directly, but wrapped in a shell. On
   macOS CI (or locally under big load with `detect_flake -c 'pnpm
   --filter @prisma/generator-helper test generatorHandler.test.ts --
   -t nonexistent' -t 100 -r 2 -p -e`), sometimes it takes more than
   200 ms for the shell to initialize, parse the command and try
   executing the non-existing binary, so the command doesn't fail in
   `initWaitTime` and `GeneratorProcess` considers it successfully
   started. This commit changes `initWaitTime` in this test to 2 seconds.

With these changes, there are no errors for me locally when spawning
200 parallel instances of the test with 5 runs per thread, but without
them at least 3% of the runs fail with just 100 threads. It is still
possible that it would fail sometimes as any tests that rely on timers
are inherently flaky under load, but the probability of this should be
now much lower.

This commit also fixes some TypeScript errors along the way (we don't
type-check the old tests on CI, so they accumulate type errors
eventually).

### DIFF
--- a/packages/generator-helper/src/GeneratorProcess.ts
+++ b/packages/generator-helper/src/GeneratorProcess.ts
@@ -112,18 +112,24 @@ export class GeneratorProcess {
             this.handleResponse(data)
           }
         })
-        // wait initWaitTime for the binary to fail
-        // TODO: this a really flaky way to detect startup failure
-        // In future, generator should explicitly send a notification when it finishes
-        // initialization and we should wait until we got that notification. We can't introduce
-        // it now since it would be a breaking change.
-        setTimeout(() => {
-          if (this.exitCode && this.exitCode > 0) {
-            reject(new Error(`Generator at ${this.executablePath} could not start:\n\n${this.stderrLogs}`))
-          } else {
-            resolve()
-          }
-        }, this.initWaitTime)
+
+        this.child.on('spawn', () => {
+          // Wait initWaitTime for the binary to report an error and exit with non-zero exit code before considering it
+          // successfully started.
+          // TODO: this is not a reliable way to detect a startup error as the initialization could take longer than
+          // initWaitTime (200 ms by default), and this also hurts the generation performance since it always waits even
+          // if the generator succesfully initialized in less than initWaitTime.  The proper solution would be to make
+          // the generator explicitly send a notification when it is ready, and we should wait until we get that
+          // notification. Requiring that would be a breaking change, however we could start by introducing an optional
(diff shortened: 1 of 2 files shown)


packages/generator-helper/src/GeneratorProcess.ts
```diff
@@ -112,18 +112,24 @@ export class GeneratorProcess {
...
+          // Wait initWaitTime for the binary to report an error and exit with non-zero exit code before considering it
+          // successfully started.
+          // TODO: this is not a reliable way to detect a startup error as the initialization could take longer than
+          // initWaitTime (200 ms by default), and this also hurts the generation performance since it always waits even
+          // if the generator succesfully initialized in less than initWaitTime.  The proper solution would be to make
+          // the generator explicitly send a notification when it is ready, and we should wait until we get that
+          // notification. Requiring that would be a breaking change, however we could start by introducing an optional
+          // notification that would stop the waiting timer as a performance optimization.
+          setTimeout(() => {
+            if (this.exitCode && this.exitCode > 0) {
```
millsp, line 126: Could this logic be moved to the `this.child.on('exit',` block and void the need to use `this.initWaitTime` as well as `setTimeout`?

aqrln: Good question.

If we move this logic to `this.child.on('exit')` we'll still need something like `setTimeout(resolve, this.initWaitTime)` in the `spawn` event handler. Otherwise the promise would have already been resolved at the point `exit` handler would be called. That said, it may be cleaner in terms of control flow, and it will allow to get rid of `this.exitCode` since it doesn't seem to be used for anything else.

However, generally it would be ideal to get rid of this logic completely if it's possible. The generator process may fail and exit with error at any point, not just during initialisation, and we need to handle that properly anyway (I'm not sure we do though), so perhaps there might not be much value in treating errors that may happen during the generator startup specially?

millsp (approved):

PR #18419, opened 2023-03-22
feat: TS support to views fs I/O

### LINKED ISSUE #18366: Introspection of the view definition statement
The next step for view introspection, we render the SQL to a file.

The implementation is the first phase of this document:

https://www.notion.so/prismaio/Changes-to-View-Introspection-and-Definition-1c3248b401a84c43bc4e2e1998d5d728

jkomyno:
Closes https://github.com/prisma/prisma/issues/18366.

- see internal [Notion](https://www.notion.so/prismaio/Changes-to-View-Introspection-and-Definition-1c3248b401a84c43bc4e2e1998d5d728)
- see internal [Slack](https://prisma-company.slack.com/archives/C02FNFLDUS3/p1679486661858169)

The engines CI pipeline is currently stuck because of a regression in the engines, which is being fixed in this [Slack thread](https://prisma-company.slack.com/archives/C02FZENHD4N/p1679327741809109).
Please use the `PRISMA_MIGRATION_ENGINE_BINARY` env var giving it an absolute path to the `migration-engine` executable in your system.


packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -481,6 +481,84 @@ describe('postgresql - missing database', () => {
...
+    await tearDownPostgres(setupParams).catch((e) => {
+      console.error(e)
+    })
+  })
+
+  test('introspection creates view definition files', async () => {
+    ctx.fixture(path.join(fixturePath))
+
+    const introspect = new DbPull()
+    const result = introspect.parse([])
```
jkomyno, line 517: Other things to check in this test:
- log/warn/error snapshots

Other scenarios to check, using this test as a template:
- additional files in the `views` folder that are not relevant for the introspected view definitions should be deleted after `prisma db pull`
- same tests with non-standard prisma paths (the `views` folder is defined exactly one level prior to the `prisma.schema` file in the directory tree)
- any other useful expectation that comes to your mind

packages/migrate/src/MigrateEngine.ts
```diff
@@ -170,9 +178,16 @@ export class MigrateEngine {
...
         this.getRPCPayload('introspect', { schema, force, compositeTypeDepth, schemas }),
       )
+      const { views } = introspectResult
+
+      if (views) {
+        const schemaPath = this.schemaPath ?? path.join(process.cwd(), 'prisma')
+        await handleViewsIO({ views, schemaPath })
+      }
+
       return introspectResult
```
jkomyno, line 191: This is the entrypoint of the I/O fs functionality.

- If `views` is undefined → the `views` preview feature is disabled, and we don't get any view definition from the engines.
- If `views` is defined, but empty → we silently delete everything within the `views` folder
- If `views` is defined -> we write files corresponding to the given views definitions, and delete the rest within the `views` folder

janpio: I think `views` is always defined now, but is `null` when the preview feature is disabled or there are not views found.
The logic here you added should still work with that as expected.

packages/migrate/src/types.ts
```diff
@@ -149,10 +152,12 @@ export namespace EngineArgs {
     compositeTypeDepth?: number
     schemas?: string[]
   }
+
   export interface IntrospectResult {
     datamodel: string
     warnings: IntrospectionWarnings[]
     version: IntrospectionSchemaVersion
+    views?: IntrospectionViewDefinition[]
```
janpio, line 160: I think this will always be included now, as (in the words of Julius) "the protocol doesn't let us to have optional keys, the key has to be there".

packages/internals/src/views/handleViewsIO.ts
```diff
@@ -0,0 +1,117 @@
...
+
+type HandleViewsIOParams = {
+  views: IntrospectionViewDefinition[]
+  schemaPath: string
+}
+
+/**
+ * For any given view definitions, the CLI must either create or update the corresponding view definition files
+ * in the file system, in `${path.dirname(schemaPath)}/views/{viewDbSchema}/{viewName}.sql`.
+ * If some other files or folders exist within the `views` directory, the CLI must remove them.
```
janpio, line 36: Why delete other files that are not conflicting with what we want to create?

Druue: Yeah, this feels a little strange to me too

packages/migrate/src/__tests__/DbPull.test.ts
Druue: > same tests with non-standard prisma paths (the views folder is defined exactly one level prior to the prisma.schema file in the directory tree)

I'm honestly not sure whether this is trying to say that the views folder _is_ defined as such, or that that would be a non-standard path and an example to test 🤔

Druue: this is the intended standard, no? so then testing when `/views/` exists on the same dir level as `/prisma/`
image

janpio:
A new engines version got integrated, so this can probably be rebased and then run tests here instead just locally: https://github.com/prisma/prisma/pull/18420#event-8819719986

janpio:
(I clicked the "Update branch" button, that might already be enough. The engine that includes should already include Julius' changes)

packages/internals/src/views/handleViewsIO.ts
```diff
@@ -0,0 +1,117 @@
...
+  const viewsDir = path.join(prismaDir, 'views')
+
+  // collect the newest view definitions
+  const viewEntries = views.map(({ schema, ...rest }) => {
+    const viewDir = path.join(viewsDir, schema)
+    return [viewDir, rest] as const
+  })
+
+  // collect the paths to the view directories (identified by their db schema name)
+  const viewPathsToWrite = viewEntries.map(([viewDir, _]) => viewDir)
```
janpio, line 50: What does this really do? Is this comment correct?

packages/internals/src/views/handleViewsIO.ts
janpio: (This would be lines 79 to 89)

packages/migrate/src/__tests__/DbPull.test.ts
janpio: The `views` folder should be next to the schema file (which by default is `prisma/schema.prisma` - but can also be `schema.prisma` or any other filename on any path if customized via https://www.prisma.io/docs/concepts/components/prisma-schema#prisma-schema-file-location

janpio: Some ideas for better and more extensive tests:
- Test some/all variants for schema file location from https://www.prisma.io/docs/concepts/components/prisma-schema#prisma-schema-file-location
- Snapshot the full Introspection result like in the basic introspection test so we have a full text representation of the returned values in the test file
- Test without preview feature to confirm nothing changes in the files
- also test other databases (all SQL should be supported) with one trivial case at least that should create a file

packages/internals/src/views/handleViewsIO.ts
pimeys: We do not know what views used to exist in the database during introspection, so we return back a list of views what are there now. The file is deleted from the folder if it's not part of the views what we return.

packages/migrate/src/MigrateEngine.ts
pimeys: We change this to be what @jkomyno described in this PR:

https://github.com/prisma/prisma-engines/pull/3814

janpio: I don't think so @pimeys. Didn't we change it to:
- No preview feature = `null`
- Preview feature but no views = `[]`
- Preview feature and views found = `[...]`

pimeys: oh yeah, I don't think this is ever undefined...

packages/migrate/src/__tests__/DbPull.test.ts
```diff
@@ -481,6 +481,231 @@ describe('postgresql - missing database', () => {
...
+          work,
+        ]
+      `)
+
+      const tree = await ctx.fs.findAsync({ directories: false, files: true, recursive: true, matching: 'views/**/*' })
+      expect(tree).toMatchInlineSnapshot(`
+        [
+          views/public/simpleuser.sql,
+          views/work/workers.sql,
+        ]
```
janpio, line 634: These two fails on Windows. How can we not run this test on Windows or skips these snapshots?

packages/migrate/src/MigrateEngine.ts
jkomyno: Confirmed, `views` is never undefined, but can be `null`. The logic doesn't change, but I'll update the type signatures to make it more explicit

packages/internals/src/views/handleViewsIO.ts
jkomyno: Agree with Julius. Closing this as "solved" as it's working like documented in Notion

packages/internals/src/utils/fs-functional.ts
```diff
@@ -0,0 +1,61 @@
...
+export const getFoldersInDir =
+  (dir: string): T.Task<string[]> =>
+  () =>
+    globby(path.posix.join(dir, '**'), { onlyFiles: false, onlyDirectories: true })
+
+// glob patterns can only contain forward-slashes, so we use path.posix.join()
+export const getFilesInDir =
+  (dir: string): T.Task<string[]> =>
+  () =>
+    globby(path.join(dir, '**'), { onlyFiles: true, onlyDirectories: false })
```
aqrln, line 47:
```suggestion
    globby(path.posix.join(dir, '**'), { onlyFiles: true, onlyDirectories: false })
```

packages/internals/src/utils/fs-functional.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Closure that creates a tagged system error for a given error callback.
+ * @param type the tag type for the error, e.g. 'fs-create-dir'
+ * @param meta any additional metadata about what caused the error
+ */
+function createTaggedSystemError<Tag extends string, Meta extends Record<string, unknown>>(type: Tag, meta: Meta) {
+  return (e: Error | unknown /* untyped error */) =>
```
aqrln, line 55: could be just

```suggestion
  return (e: unknown /* untyped error */) =>
```

packages/internals/src/views/handleViewsIO.ts
```diff
@@ -0,0 +1,117 @@
...
+          const viewDirsToRemove = directoriesInViewsDir.filter((dir) => !viewPathsToWrite.includes(dir))
+          return TE.traverseArray(removeDir)(viewDirsToRemove)
+        }),
+      ),
+    ),
+
+    // remove any other files in the views directory beyond the ones just created, concurrently, collapsing the possible errors
+    TE.chainW(() =>
+      pipe(
+        getFilesInDir(viewsDir),
```
aqrln, line 82: Should it maybe only remove stray `.sql` files? Doesn't sound right to remove `README.md` that someone put it or something like that.

packages/migrate/src/__tests__/DbPull.test.ts
aqrln: I don't think we should skip them since it's an actual bug on windows. The reason is probably because of the broken glob pattern on windows (see the review comment below).

packages/internals/src/utils/fs-functional.ts
```diff
@@ -0,0 +1,68 @@
...
+    TE.tryCatch(() => fs.promises.rmdir(dir, { recursive: true }), createTaggedSystemError('fs-remove-dir', { dir })),
+  )
+
+export const removeFile = (filePath: string) =>
+  pipe(TE.tryCatch(() => fs.promises.unlink(filePath), createTaggedSystemError('fs-remove-file', { filePath })))
+
+/**
+ * Removes all backslashes from a possibly Windows path string, which is necessary for globby to work on Windows.
+ */
+const normalizePossiblyWindowsDir = (dir: string) => dir.replace(/\\/g, '/')
```
aqrln, line 40: A backslash is a valid filename character on Linux and macOS, so this will break in that edge case. I'd replace `path.sep` instead of `\\` here, but keeping an explicit backslash is also an option if you wrap it into `process.platform === 'win32'` check.

jkomyno:
- Buildkite tests pass (also on Windows)
- Github Actions fails on CI (we can overcome this temporarily with a `describeIf` block)

aqrln (approved):

PR #18421, opened 2023-03-22
feat(client): error rendering for JSON protocol

SevInf:
- feat(client): Error rendering for validation errors done by client
- MissingRequiredArgument error
- Ability to underline argument values
- Start integrating engine changes


packages/client/tests/functional/query-validation/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+
+                        Invalid \`prisma.user.findUnique()\` invocation in
+                        /client/tests/functional/query-validation/tests.ts:0:0
+
+                          XX 
+                          XX test('missing required argument', async () => {
+                          XX   // @ts-expect-error
+                        → XX   const result = prisma.user.findUnique({
+                                + where: {
+                                +   where: UserWhereUniqueInput
```
millsp, line 151: Is this error correct?

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -320,3 +334,38 @@ function getItxTransactionOptions<PayloadType>(
...
+  if (error.kind === 'Union') {
+    return {
+      kind: 'Union',
+      errors: error.errors.map(convertValidationError),
+    }
+  }
+
+  if (Array.isArray(error['selectionPath'])) {
+    const selectionPath = [...error['selectionPath']]
+    selectionPath.shift()
```
millsp, line 362:
```suggestion
    const [, ...selectionPath] = error['selectionPath']
```

packages/client/src/generation/ts-builders/Writer.ts
```diff
@@ -1,16 +1,18 @@
 import type { BasicBuilder } from './BasicBuilder'
 
-const INDENT_SIZE = 2
-export class Writer {
+export const INDENT_SIZE = 2
+export class Writer<ContextType = undefined> {
```
millsp, line 4: It would be great to document this, it looks really simple and useful.

packages/client/src/runtime/core/errorRendering/ScalarValue.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  override getPrintWidth(): number {
+    return this.text.length
+  }
+
+  override write(writer: ErrorWriter): void {
+    const string = new FormattedString(this.text)
+    if (this.hasError) {
+      string.underline().setColor(writer.context.chalk.redBright)
```
millsp, line 17: Does the color need to come from the context? Or could we more simply pass the chalk color directly?

packages/client/src/runtime/core/errorRendering/ObjectValue.ts
```diff
@@ -0,0 +1,195 @@
...
+      const nextField = field.value.getField(segment)
+      if (!nextField) {
+        return undefined
+      }
+      field = nextField
+    }
+    return field
+  }
+
+  getDeepFieldValue(path: string[]): Value | undefined {
```
millsp, line 69:
```suggestion
  getDeepFieldValue(path: string[]) {
```

packages/engine-core/src/common/types/EngineValidationError.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+export type UnknownArgumentError = {
+  kind: 'UnknownArgument'
+  selectionPath: string[]
+  argumentPath: string[]
+  arguments: ArgumentDescription[]
+}
+
+export type UnknownInputFieldError = {
```
millsp, line 38: It would be great to have a short description for each of these.

packages/client/src/runtime/core/types/ValidationError.ts
```diff
@@ -1,25 +1,13 @@
...
+export type IncludeAndSelectError = {
+  kind: 'IncludeAndSelect'
+  selectionPath: string[]
 }
 
-type OutputTypeDescriptionField = {
-  name: string
-  typeName: string
-  isRelation: boolean
+export type IncludeOnScalarError = {
```
millsp, line 8: It would be great to have a short description for each of these.

millsp: Ideally a bit of docs for some of the fields that aren't obvious.

packages/client/src/runtime/core/protocol/json/JsonProtocolEncoder.ts
```diff
@@ -1,16 +1,17 @@
 import { EngineBatchQuery, EngineQuery, JsonFieldSelection, JsonQuery, JsonQueryAction } from '@prisma/engine-core'
```
millsp, line 1:
```suggestion
import { EngineBatchQuery, JsonQuery, JsonQueryAction } from '@prisma/engine-core'
```

packages/client/tests/functional/query-validation/tests.ts
```diff
@@ -0,0 +1,394 @@
...
+                    ?   id?: String,
+                    ?   email?: String
+                      }
+                    })
+
+            Argument where of type UserWhereUniqueInput needs at least one argument. Available options are listed in green.
+        `)
+  })
+
+  testIf(previewFeatures === '"extendedWhereUnique"')('missing on of the specific required fields', async () => {
```
millsp, line 341:
```suggestion
  testIf(previewFeatures === '"extendedWhereUnique"')('missing one of the specific required fields', async () => {
```

packages/client/tests/functional/query-validation/tests.ts
```diff
@@ -0,0 +1,394 @@
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(({ previewFeatures }, suiteMeta) => {
+  test('include and select are used at the same point', async () => {
```
millsp, line 8:
```suggestion
  test('include and select are used at the same time', async () => {
```

packages/client/tests/functional/query-validation/tests.ts
```diff
@@ -0,0 +1,394 @@
...
+                     select: {
+                   ?   id?: true,
+                   ?   email?: true,
+                   ?   name?: true,
+                   ?   createdAt?: true,
+                   ?   published?: true
+                     }
+                   })
+
+            The \`select\` statement for type User must not be empty. Available options are listed in green.
```
millsp, line 115: It would be nice to have backticks here (and also any other possibly misisng backtick).

packages/client/tests/functional/query-validation/tests.ts
```diff
@@ -0,0 +1,394 @@
...
+              ?   NOT?: UserWhereInput | UserWhereInput[],
+              ?   id?: StringFilter | String,
+              ?   email?: StringFilter | String,
+              ?   name?: StringFilter | String,
+              ?   createdAt?: DateTimeFilter | DateTime,
+              ?   published?: BoolFilter | Boolean
+                }
+              })
+
+      Unknown argument notAValidField. Available options are listed in green.
```
millsp, line 179: Backticks here too?

packages/client/src/runtime/core/protocol/json/serialize.ts
```diff
@@ -72,13 +99,17 @@ function serializeSelectionSet(
   context: SerializeContext,
 ): JsonSelectionSet {
   if (select && include) {
-    throw new Error('select and include are used at the same time')
+    context.throwValidationError({ kind: 'IncludeAndSelect', selectionPath: context.getSelectionPath() })
```
millsp, line 102: I wondered if this could be done in the engines rather?

packages/client/src/runtime/core/errorRendering/applyValidationError.test.ts
```diff
@@ -0,0 +1,1495 @@
...
+          user: {
+            where: {
+      ?       id?: String,
+      ?       email?: String
+            }
+          }
+        }
+      }
+
+      Argument where of type UserWhereUniqueInput needs at least one argument. Available options are listed in green.
```
millsp, line 1301: It's hard to see the colors in the snapshot. Could we replace the ansi codes with eg `<green>...</green>`?

packages/client/src/runtime/core/errorRendering/applyValidationError.ts
```diff
@@ -0,0 +1,494 @@
...
+  })
+}
+
+function applyEmptySelectionError(error: EmptySelectionError, argsTree: ArgumentsRenderingTree) {
+  const outputType = error.outputType
+  const selection = argsTree.arguments.getDeepSelectionParent(error.selectionPath)?.value
+  const isEmpty = selection?.isEmpty() ?? false
+
+  if (selection) {
+    selection.removeAllFields()
```
millsp, line 128: Why must the fields be removed here?

packages/client/src/runtime/core/protocol/json/serialize.ts
SevInf: This and `IncludeOnScalar` can not - `include` and `select` exist only on the client, on protocol/engine level it is just a single selection, so engine can't tell that both `include` and `select` were used. These 2 are the only validation rules left on the client though.

packages/client/src/runtime/core/errorRendering/ScalarValue.ts
SevInf: So, `chalk` object in context can be non-default one, with a different configuration. In particular, if `errorFormat` is anything other than `pretty` it will be a disabled instance which won't add any colors to the strings. This allows us to easily support both colored and colorless output. [Reference code](https://github.com/prisma/prisma/pull/18421/files#diff-2f65b0b20e447d3e2bd7627410b745ee3eb24f93508608fbe40740a322554c74R33-R36)

packages/client/src/runtime/core/errorRendering/applyValidationError.ts
SevInf: So, this a case where we have fields in `select`, but all of them are `false`, so it results in an empty selection which is still not valid. In that case, we have several options on how to render suggestions:

1. Display only those fields that do not already exist on selection. Example rendering:
   
   ```ts
   {
     select: {
        id: false
        posts: false,
    ?   name?: true 
    ?   email?: true 
     }
   }
   ```

   There are couple of possible problems here. First, we are assuming that user needs to add new field to the selection, rather than change one of the existing ones to `true`. Second, we might end up in a situation where all fields are already used in selection and we have nothing left to suggest.

2. Completely ignore users input and suggest all the fields. Example rendering:

    ```ts
   {
     select: {
    ?   id?: true
    ?   posts?: true,
    ?   name?: true 
    ?   email?: true 
     }
   }
   ```

   In that case, we are suggesting that user can either change one of the existing fields to `true` or add a new one that would be `true`. And this is the approach I've taken here. In order to do it, we first remove all the fields user specified here and then add them (+ other fields from suggestion) as suggestion.

packages/client/src/runtime/core/errorRendering/applyValidationError.test.ts
SevInf: Added snapshot serializer that does just that. Downside is, it is harder to judge the alignment of different lines because of added extra content. Let me know which one would you prefer.

millsp: I guess we could have a test that is colorless and one that is with color?

packages/client/src/runtime/core/errorRendering/applyValidationError.ts
millsp: Nice, thanks for the thorough explanation.

millsp: I would have loved that in the comments.

millsp (approved):
image

packages/client/src/runtime/core/errorRendering/applyValidationError.ts
SevInf: Added.

PR #18437, opened 2023-03-23
fix(client): throw error on caching platforms

### LINKED ISSUE #6634: Netlify deploy does not pick up changes to schema file
1. Create a repository with a minimal, mysql Prisma project
2. Create a model in the schema and migrate the DB to contain the schema (via `db push` for example)
3. Create a minimal serverless function `netlify/functions/invocation-count.js` that reads the empty table for example
4. Test everything locally via `netlify dev`
2. Create a new site in Netlify that connects to this repo
3. Set a MySQL `DATABASE_URL` env var and deploy
4. Test that the deployed function works as expected
5. Update the Netlify env var to be a PostgreSQL database (that also has the table via `db push`)
5. Update the `schema.prisma` file to use a `postgresql` provider
6. Watch Netlify deploy the updated project
7. Try the function again

I got an error telling me that the env var should have a different protocol, still `mysql://`.

Somehow it seems that the Client did not generated again when I only changed the schema file in the repository. Deplying again to Netlify with clearing the cache avoided the problem.

Reason probably is that `prisma generate` by default is only run via `npm install`, and as cache is used on a normal deploy that is not necessary. But for Prisma it kinda is, for the schema changes to becomes active in a new generated Client.

### LINKED ISSUE #7291: Vercel: Schema only change does not invalidate build cache
Switching `provider` from `postgresql` to `mysql` does not cause Prisma Client to be regenerated.

Deploy project with this first, observe error:
```
datasource db {
  provider = "postgresql"
  url      = "mysql://root:@127.0.0.1:3306/e2e-tests"
}
```

Switch to this and keep seeing the _same_ error:
```
datasource db {
  provider = "mysql"
  url      = "mysql://root:@127.0.0.1:3306/e2e-tests"
}
```

### LINKED ISSUE #7818: Serverless deployments: Just making a schema level change to a repository does not invalidate `node_modules` cache that contains generated Client
This is more a theory than a bug report, but this could explain why sometimes deployments to Netlify and other deployment providers do not actually deploy a new schema.

Example: https://github.com/netlify/build-image/blob/d51b4da41df1087448804e27626a32d83f50380f/run-build-functions.sh#L184-L197 `npm install` is only run if the sha of `package.json` is different, which it does not have to be if you only updated your schema. No `npm install` means not post install hook, means no `prisma generate`.

We should get a better overview here by designing a few test setups, and testing them at a few of these potentially affected providers to see how they handle that.

millsp:
Detects when Prisma Client is in an environment that is known to cause caching issues and throws an error for the user to fix.

**Reminder of the problem:**
On the first publication to eg. Vercel, the client will auto-generate and it will get cached. Then we change the schema, and on the second publication, the client will not auto-generate. Because @prisma/client is cached already, the auto-generate won't trigger again. We then have a caching problem, and [queries or validation fail](https://www.notion.so/prismaio/Research-Vercel-Netlify-caching-behavior-4ca3e950d7bb43e081a1a2f5a5149518?pvs=4#6fea87b0ed8f4769b1ac80e2d2037730) because of the de-sync.

**Relevant research:**
https://github.com/millsp/vercel-prisma-cache-issue-repro
https://www.notion.so/prismaio/Research-Vercel-Netlify-caching-behavior-4ca3e950d7bb43e081a1a2f5a5149518
https://prisma-company.slack.com/archives/C04D5C9D1FG/p1679440563960249

**Issues:**
Fixes https://github.com/prisma/prisma-private/issues/157
Fixes https://github.com/prisma/prisma/issues/6634
Fixes https://github.com/prisma/prisma/issues/7291
Fixes https://github.com/prisma/prisma/issues/7818
Closes https://github.com/prisma/client-planning/issues/296

**Related PRs:**
https://github.com/prisma/ecosystem-tests/pull/3454


packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -98,6 +100,8 @@ export class TSClient implements Generatable {
       datasourceNames: datasources.map((d) => d.name),
       activeProvider: this.options.activeProvider,
       dataProxy: this.options.dataProxy,
+      postinstall: this.options.postinstall,
+      ciName: ciInfo.name ?? undefined, // a
```
millsp, line 104:
```suggestion
      ciName: ciInfo.name ?? undefined,
```

packages/client/tests/e2e/platform-caching-error/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "private": true,
+  "version": "0.0.0",
+  "main": "index.js",
+  "scripts": {},
+  "dependencies": {
+    "@prisma/client": "../prisma-client-0.0.0.tgz"
+  },
+  "devDependencies": {
+    "@types/node": "16.18.11 ",
```
aqrln, line 10:
```suggestion
    "@types/node": "16.18.11",
```

packages/client/src/runtime/core/init/checkPlatformCaching.ts
```diff
@@ -0,0 +1,39 @@
...
+const debug = Debug('prisma:client')
+
+/**
+ * Known platforms that have caching issues. Updating this list will also update
+ * the error message and the link to the docs, so add docs/links as needed.
+ */
+const cachingPlatforms = {
+  Vercel: 'vercel',
+  'Netlify CI': 'netlify',
+} as const
```
janpio, line 15: I think they key here is the `ci-info` identifier of that platform? 
If so, we should document that so it is clear how this can be expanded if needed.

packages/client/src/runtime/core/init/checkPlatformCaching.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  // if client was not generated manually
+  if (postinstall !== true) return
+
+  // and we generated on one a caching CI
+  if (ciName && ciName in cachingPlatforms) {
+    throw new PrismaClientInitializationError(
+      `We detected that your project setup might lead to outdated Prisma Client being used.
+Please make sure to run the \`prisma generate\` command during your build process.
+Learn how: https://pris.ly/d/${cachingPlatforms[ciName]}-build`,
```
janpio, line 35: A suggestion that takes the angle of explaining a bit and connecting it to the platform (by name):
```suggestion
      `We've detected that you are building your project on ${ciName}, which applies caching to dependencies. This leads to Prisma's automated \`generate\` command potentially not being executed, and your code using an outdated Prisma Client as result. Please make sure to run the \`prisma generate\` command during your build process.
For more information, please visit this link: https://pris.ly/d/${cachingPlatforms[ciName]}-build`,
```
Not sure if better.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -265,6 +266,22 @@ export interface GetPrismaClientConfig {
    * @remarks only used for the purpose of data proxy
    */
   inlineSchemaHash?: string
+
+  /**
+   * A marker to indicate that the client was not generated via `prisma
+   * generate` but was triggered via the `postinstall` script instead.
+   * @remarks used to error for Vercel/Netlify for schema caching issues
```
janpio, line 273:
```suggestion
   * A marker to indicate that the client was not generated via `prisma
   * generate` directly, but was via a `generate` triggered via the `postinstall` 
   * script instead.
   * @remarks used to error for Vercel/Netlify for schema caching issues
```

janpio (approved):

packages/client/src/runtime/core/init/checkPlatformCaching.test.ts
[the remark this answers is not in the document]
millsp: Nothing, I will remove it as I tested it via an e2e test.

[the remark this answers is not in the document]
millsp: I forgot to add a unit test, now done

PR #18454, opened 2023-03-23
refactor: remove dead code

aqrln:
Remove unused code that's not called anywhere.

### DIFF
--- a/packages/internals/src/cli/getSchema.ts
+++ b/packages/internals/src/cli/getSchema.ts
@@ -246,7 +246,6 @@ export async function getSchemaDir(schemaPathFromArgs?: string): Promise<string 
   return path.dirname(schemaPath)
 }
 
-// TODO: This should probably return string | null to stay consistent with the other functions
 export async function getSchema(schemaPathFromArgs?: string): Promise<string> {
   const schemaPath = await getSchemaPath(schemaPathFromArgs)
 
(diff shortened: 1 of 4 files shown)


packages/internals/src/cli/getSchema.ts
```diff
@@ -246,7 +246,6 @@ export async function getSchemaDir(schemaPathFromArgs?: string): Promise<string
   return path.dirname(schemaPath)
 }
 
-// TODO: This should probably return string | null to stay consistent with the other functions
```
aqrln, line 249: This comment was removed because we looked into it with @millsp and @SevInf and concluded that it makes more sense as it currently is.

millsp (approved):

packages/internals/src/cli/getSchema.ts
```diff
@@ -361,41 +360,6 @@ function getRelativeSchemaPathSync(cwd: string): string | null {
...
-  const schemaPath = getSchemaPathSync(schemaPathFromArgs)
-  if (schemaPath) {
-    return path.dirname(schemaPath)
-  }
-
-  return null
-}
-
-// TODO: This should probably return string | null to stay consistent with the other functions
-export function getSchemaSync(schemaPathFromArgs?: string): string {
```
Jolg42, line 381: Note that this was used by third party generators / projects.
Since the async version exists, it might be fine in the end.

PR #18457, opened 2023-03-23
refactor(internals): use copyFile in resolveBinary

aqrln:
Both the referenced issue and the pull request were resolved long ago,
so `copyFile` should work in pkg.


janpio:
```
[Error: ENOENT: no such file or directory, copyfile '/snapshot/node_modules/prisma/node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node' -> '/tmp/prisma-binaries/libquery_engine-debian-openssl-1.1.x.so.node'] {
```

Jolg42:
pkg PR fix says it's only available since pkg@5.6.0 and we are testing using pkg@4.5.1 in https://github.com/prisma/ecosystem-tests/actions/runs/4503937481/jobs/7927807134#step:5:114

janpio:
... which we are doing, because upgrading to pkg@5 makes those tests fail completely: https://github.com/prisma/ecosystem-tests/pull/2899

Jolg42 (approved):

millsp (approved):

PR #18459, opened 2023-03-23
refactor(cli): resolve TODO comment in Version.test.ts

aqrln:
This came up during a deep dive meeting with @millsp and @SevInf.

PR marked as draft because we were not sure.

### DIFF
--- a/packages/cli/src/__tests__/commands/Version.test.ts
+++ b/packages/cli/src/__tests__/commands/Version.test.ts
@@ -112,16 +112,14 @@ function cleanSnapshot(str: string, versionOverride?: string): string {
   //                                                                                    ^^^^^^^^^^^^^^^^^^^
   str = str.replace(/\(at (.*engines)(\/|\\)/g, '(at sanitized_path/')
 
-  // TODO: replace '[a-z0-9]{40}' with 'ENGINE_VERSION'.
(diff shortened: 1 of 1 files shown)


packages/cli/src/__tests__/commands/Version.test.ts
```diff
@@ -112,17 +112,15 @@ function cleanSnapshot(str: string, versionOverride?: string): string {
   //                                                                                    ^^^^^^^^^^^^^^^^^^^
   str = str.replace(/\(at (.*engines)(\/|\\)/g, '(at sanitized_path/')
 
-  // TODO: replace '[a-z0-9]{40}' with 'ENGINE_VERSION'.
-  // Currently, the engine version of @prisma/prisma-fmt-wasm isn't necessarily the same as the enginesVersion
```
millsp, line 116: cc @jkomyno do you know why they aren't necessarily the same versions?

millsp:
Do you remember was the outcome of our conversation our this?

packages/cli/src/__tests__/commands/Version.test.ts
janpio: Technically one could use a different one if there are not relevant changes that affect prisma-fmt/schema-wasm maybe, but I don't think our setup is designed to ever actually do that. They _should_ be in sync. (cc @millsp)

PR #18463, opened 2023-03-23
chore(bench): fix benchmark GH action on main

aqrln:
Currently the "Store benchmark result" step of the benchmark pipeline
fails with `Error: No benchmark result was found in
/home/runner/work/prisma/prisma/output.txt.`. The reason is the second
benchmark run overwrites the `output.txt` file from the first run but
doesn't contain the raw Benchmark.js output. This commit fixes that.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -71,7 +71,7 @@ jobs:
       - name: Run benchmarks for Codspeed
         uses: CodSpeedHQ/action@v1
         with:
-          run: pnpm run bench
+          run: pnpm run bench-stdout-only
           token: ${{ secrets.CODSPEED_TOKEN }}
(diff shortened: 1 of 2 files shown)


SevInf (approved):

.github/workflows/benchmark.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       - name: Run benchmarks for Codspeed
         uses: CodSpeedHQ/action@v1
         with:
-          run: pnpm run bench
+          run: pnpm run bench-stdout-only
           token: ${{ secrets.CODSPEED_TOKEN }}
 
       - name: Store benchmark result
```
janpio, line 77: Move this before Codspeed step (line 71)?

PR #18464, opened 2023-03-23
fix(ci): pin pnpm version because of bug

millsp:
Temporary fix for upstream issue https://github.com/pnpm/pnpm/issues/6272


aqrln (approved):

zkochan:
I have downgraded latest to 7.30.0, so it shouldn't be an issue now.

millsp:
Thanks!

PR #18487, opened 2023-03-24
feat: allow tracing preview feature with data proxy

aqrln:
- Allow `tracing` preview feature when generating Data Proxy client.
- Replace the test that checked that `tracing` was not allowed with a similar test for `metrics` preview feature that wasn't tested.

Closes https://github.com/prisma/client-planning/issues/302


millsp (approved):
Good catch for the missing test!

PR #18511, opened 2023-03-27
chore(client): Unskip some of the json protocol validation tests

SevInf:


packages/client/src/__tests__/integration/errors/int-errors/test.ts
```diff
@@ -55,7 +55,7 @@ describe('int-errors', () => {
     }
   })
 
-  testIf(getQueryEngineProtocol() !== 'json')('overflow-int', async () => {
+  test('overflow-int', async () => {
```
aqrln, line 58: Is it intentional that the previous test is still skipped?

aqrln (approved):

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
```diff
@@ -188,34 +188,15 @@ testMatrix.setupTestSuite(
...
-              name,
-            },
-            update: {
-              name,
-            },
-          }),
-        ).rejects.toThrow('Argument where of type UserWhereUniqueInput needs exactly one argument')
-
-        const checker = new UpsertChecker(client)
+    test('should only use ON CONFLICT when there is only 1 unique field in the where clause', async () => {
```
aqrln, line 191: `getQueryEngineProtocol` import can probably be removed unless I've missed it being used elsewhere in the file

PR #18571, opened 2023-03-30
fix(cli): Exclude source maps from bundled client

SevInf:
Client, bundled in cli still shipped with source maps. Fixed by using
`npm-packlist`, so that contents of bundled client will always be
identical to `@prisma/client` pacakge.

### DIFF
--- a/packages/cli/helpers/copy-prisma-client.ts
+++ b/packages/cli/helpers/copy-prisma-client.ts
@@ -1,23 +1,25 @@
-import fg from 'fast-glob'
 import { copySync } from 'fs-extra'
+import packlist from 'npm-packlist'
 import path from 'path'
 
-// that's where we want to copy the local client for @prisma/studio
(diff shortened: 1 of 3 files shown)


packages/cli/helpers/copy-prisma-client.ts
```diff
@@ -1,23 +1,25 @@
-import fg from 'fast-glob'
 import { copySync } from 'fs-extra'
+import packlist from 'npm-packlist'
```
millsp, line 2: Is `npm-packlist` as performant (for wach mode) as the previous solution? I remeber trying to use npm's arborist and it was pretty slow. So I think as long as this dep does not use that, it should be fine.

SevInf: I see no impact on build/ci times

millsp (approved):

PR #18572, opened 2023-03-30
feat: Support readonly arrays

### LINKED ISSUE #15815: TypeScript, `readonly` arrays combined with `where` and `in`
## Problem

I'm working a lot with `readonly` properties on a lot of my types, for immutability, because I think that makes the code better. I've just started messing around with the `in` filter, and it doesn't seem to work well with `readonly` arrays. For example:

```typescript
        const emailAddresses: readonly string[] = ["test@test.se", "hello@test.se"];
        await client.user.findMany({
            where: {
                emailAddress: {
                    in: emailAddresses, // Type 'readonly string[]' is not assignable to type 'Enumerable<string> | undefined'. The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'. ts(2322)
                },
            },
        });
```

## Suggested solution

I don't see any downsides of allowing `readonly` arrays with the `in` filter, and I can't see any technical constraints against it either (but I'm no TypeScript guru).

## Alternatives

None.

## Additional context

None.

LinusU:
fixes prisma/prisma#15815

-----

I've tested this by simply modifying my generated Prisma client file. The following code now works:

```ts
async function findProductsByRestaurant (restaurantIds: readonly string[]): Promise<Product[]> {

  // The following row previously complained that `readonly string[]` cannot be passed to `Enumerable<string>`
  const rows = await postgres.product.findMany({ where: { restaurantId: { in: restaurantIds } } })


  return restaurantIds.map(id => rows.filter(row => row.restaurantId === id))
}
```

### DIFF
--- a/packages/client/src/generation/TSClient/common.ts
+++ b/packages/client/src/generation/TSClient/common.ts
@@ -322,7 +322,7 @@ type Prisma__Pick<T, K extends keyof T> = {
 };
 
 
-export type Enumerable<T> = T | Array<T>;
+export type Enumerable<T> = T | ReadonlyArray<T>;
 
 export type RequiredKeys<T> = {
   [K in keyof T]-?: {} extends Prisma__Pick<T, K> ? never : K
@@ -379,7 +379,7 @@ type XOR<T, U> =
 /**
  * Is T a Record?
  */
-type IsObject<T extends any> = T extends Array<any>
+type IsObject<T extends any> = T extends ReadonlyArray<any>
 ? False
 : T extends Date
 ? False
@@ -395,7 +395,7 @@ type IsObject<T extends any> = T extends Array<any>
 /**
  * If it's T[], return T
  */
-export type UnEnumerate<T extends unknown> = T extends Array<infer U> ? U : T
+export type UnEnumerate<T extends unknown> = T extends ReadonlyArray<infer U> ? U : T
 
 /**
  * From ts-toolbelt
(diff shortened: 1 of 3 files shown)


packages/client/src/generation/TSClient/common.ts
```diff
@@ -322,7 +322,7 @@ type Prisma__Pick<T, K extends keyof T> = {
 };
 
 
-export type Enumerable<T> = T | Array<T>;
+export type Enumerable<T> = T | ReadonlyArray<T>;
```
SevInf, line 325: I beleive that would unfortinately be a breaking change. Consider following pattern:

```ts
const input: UserWhereInput = { AND: [] };
input.AND.push(condition1)
input.AND.push(condition2)
```

This is the code that works for current version, but will break after PR is merged. Ideally, we should find a way to accept readonly arrays as input, while still allowing to build query input dynamically.

SevInf (changes requested):
Hi @LinusU !
Thank you for the contribution. There are some changes that needs to be made before it can be accepted. Additionally, couple of tests are failing on CI after this change. Please, let us know when you've addressed the feedback and fixed the tests.

packages/client/src/generation/TSClient/common.ts
Akxe: It would not, but you would have to change the line to `const input = { AND: [] } satisfies UserWhereInput;` which is a safer solution anyway

Akxe:
I would like to point out that this PR would fix #20946 and #5025. Also since the keyword `satisfies` this is no longer a deal breaker for other users. It is a slight breaking change, but the fix is to instead of type annotation use the `satisfies` keyword and if users are informed properly in the changelog the fix should be very easy for anyone.

PR #18584, opened 2023-03-31
chore: Remove unused variables and enable no-unused-vars rule

SevInf:
Enables `@typesctipt-eslint/no-unused-vars` and fixes all violations.
There are a couple of exceptions to the rule:

1. Spreads are ignored, making it possible to omit a property with
   object spread operator, i.e. this code valid:

   ```ts
   const { propToOmit, ...rest } = someObject
   ```
2. In some cases, we want to have a function argument that won't be used
   in the function body. Main use-case for this is TS inference tests.
   In that case, argument may be prefixed with `_` and the rule won't
   trigger.


.eslintrc.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
     '@typescript-eslint/explicit-module-boundary-types': 'off',
     '@typescript-eslint/ban-ts-comment': 'off',
     '@typescript-eslint/no-empty-function': 'off',
+    '@typescript-eslint/no-unused-vars': ['error', { ignoreRestSiblings: true, argsIgnorePattern: '^_' }],
```
janpio, line 42: What does this do? Worth a comment?

packages/migrate/src/utils/getDatabaseVersionSafe.ts
```diff
@@ -1,14 +1,15 @@
-import Debug from '@prisma/debug'
+// import Debug from '@prisma/debug'
 
-import { MigrateEngine } from '../MigrateEngine'
+// import { MigrateEngine } from '../MigrateEngine'
 
-const debug = Debug('prisma:cli')
+// const debug = Debug('prisma:cli')
```
janpio, line 5: Intentional?

packages/client/src/__tests__/integration/errors/executeRaw-alter-postgres/test.ts
```diff
@@ -11,7 +11,6 @@ beforeAll(async () => {
   process.env.TEST_POSTGRES_URI += '-execute-raw-alter'
   await tearDownPostgres(process.env.TEST_POSTGRES_URI!)
   await migrateDb({
-    connectionString: process.env.TEST_POSTGRES_URI!,
```
millsp, line 14: Is this not necessary?

SevInf: `connectionString` was not used within `migrateDb` so I removed it everywhere.

millsp (approved):

helpers/blaze/flatten.ts
```diff
@@ -16,7 +16,7 @@ function wrap(item: unknown) {
  * @param list
  * @returns
  */
-function flatten<L extends L.List, I>(list: L & L.List<I>): Flatten<L> {
+function flatten<T extends L.List, I>(list: T & L.List<I>): Flatten<T> {
```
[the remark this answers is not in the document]
SevInf: I think there is a bug in eslint rule and decided to rename the variables to work around it. What happens is: `L` is the name of the namespace from `ts-toolbelt`, as well as generic parameter name in the functions. Rule incorrectly assumes that in `L extends L.List`, namespace-`L`  is shadowed by generic-`L` by the time we get to `extends` and both `L` in that statement refer to generic argument, therefore, namespace-`L` is not used. Actual `TS` does not work that way. I could've disabled the rule for these files, but renaming the parameters also worked.

packages/client/src/__tests__/integration/errors/connection-limit-mysql/test.ts
```diff
@@ -36,11 +36,9 @@ describeIf(process.platform === 'linux')('connection-limit-mysql', () => {
       })
       clients.push(client)
     }
-    let count = 0
     try {
       for (const client of clients) {
         await client.$connect()
-        count++
```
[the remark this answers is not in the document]
SevInf: Not sure, there was no assertions on `count` anywhere. But sounds probable.

packages/migrate/src/utils/getDatabaseVersionSafe.ts
SevInf: Yes, code in this module is commented and TODO says to uncomment it when https://github.com/prisma/prisma-private/issues/203 is fixed. Imports were still there and were marked as unused by the rule.

.eslintrc.js
SevInf: Functions arguments specifically. Added comments.

PR #18592, opened 2023-03-31
docs(cli): add --generator option to the help hint

maxmartynov:
The CLI option `--generator` was added in https://github.com/prisma/prisma/pull/16452 but the CLI's `--help` hint wasn't updated.

Link to the docs https://www.prisma.io/docs/reference/api-reference/command-reference#arguments-1


packages/cli/src/Generate.ts
```diff
@@ -50,6 +50,7 @@ ${chalk.bold('Options')}
       --schema   Custom path to your Prisma schema
   --data-proxy   Enable the Data Proxy in the Prisma Client
        --watch   Watch the Prisma schema and rerun after a change
+   --generator   Specifies which generator to use to generate assets (may be provided multiple times)
```
aqrln, line 53: Here it's probably best to shorten the description a bit more. How about something like this?

```suggestion
   --generator   Generator to use (may be provided multiple times)
```

aqrln (commented):
Nice, thanks for catching this!

PR #18638, opened 2023-04-05
perf(get-platform): Optimize some of the platform detection routines

SevInf:
1. Detect node support using `process.arch` rather than calling `getos`,
   that will re-do openssl search
2. Look for openssl lib using readdir rather than launching child
   process and quit on the first found library, rather then continuing
   executing commands.

Together, those changes improve loading peroformance by 10-20ms on orm
comparison benchmarks.


aqrln (approved):

janpio:
Could we maybe add a benchmark for this code?
Anything that we realize being performance relevant would be nice to track that way.

PR #18647, opened 2023-04-06
fix(client): support TypeScript exactOptionalPropertyTypes

### LINKED ISSUE #10894: Generated client is not compatible with `exactOptionalPropertyTypes`
### Bug description

`exactOptionalPropertyTypes` has TypeScript apply stricter rules around the handling of optional properties - specifically when enabled optional properties are considered to only be allowed to be missing rather than explicitly `undefined`. e.g.

```
interface Options {
  myProp?: string;
}

declare const doSomething: (options: Options) => void;

doSomething({}); // fine
doSomething({ myProp: '' }); // fine
doSomething({ myProp: undefined }); // error if `exactOptionalPropertyTypes` is enabled
```

https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes

If `undefined` is allowed, it needs to be explicitly defined as part of the type:

```
interface Options {
  myProp?: string | undefined;
}

declare const doSomething: (options: Options) => void;

doSomething({ myProp: undefined }); // fine 🎉 
```

Currently it seems like the generated Prisma client is not correctly typed when this flag is enabled based on this code:

```
const fetchProblems = async (
  { id }: Branch,
  ids?: number[]
): Promise<Problem[]> => {
  const auditables = await db.auditable.findMany({
    where: { branch: { id }, id: { in: ids } },
    select: { problems: true }
  });

  return auditables.flatMap(a => a.problems);
};
```

This code appears to work fine at runtime but errors if I have `exactOptionalPropertyTypes` enabled due to the type being:

```
/**
 * Auditable findMany
 */
export type AuditableFindManyArgs = {
  /**
   * Filter, which Auditables to fetch.
   */
  where?: AuditableWhereInput;
};

export type AuditableWhereInput = {
  id?: IntFilter | number;
};

export type IntFilter = {
  equals?: number;
  in?: Enumerable<number>; // <- this needs to be `| undefined`
  notIn?: Enumerable<number>;
  lt?: number;
  lte?: number;
  gt?: number;
  gte?: number;
  not?: NestedIntFilter | number;
};
```

(I've omitted surrounding types for brevity)

I don't know if every optional property in the client can accept an explicit `undefined`, but couldn't find any previous issues about `exactOptionalPropertyTypes` so am opening this as a general one. Overall I expect resolving this will need to be an incremental effort.

I'm happy to help out with this too :)

### How to reproduce

1. Setup a Prisma TypeScript project
2. Enable `exactOptionalPropertyTypes` in `tsconfig.json`
3. Generate Prisma client
4. Attempt to pass explicit `undefined` values to the client

### Expected behavior

No compiler errors

### Prisma information

```
datasource db {
  provider = "sqlite"
}

generator client {
  // WebStorm only properly indexes packages that are defined as actual dependencies
  // in the projects package.json & lock, so we output the client within the
  // @prisma/client package so that the types can be properly resolved
  output   = "../node_modules/@prisma/client/generated"
  provider = "prisma-client-js"
}

model Branch {
  id         Int         @id @default(autoincrement())
  createdAt  DateTime    @default(now())
  updatedAt  DateTime    @updatedAt
  name       String
  auditables Auditable[]
}

model Auditable {
  id        Int       @id @default(autoincrement())
  createdAt DateTime  @default(now())
  updatedAt DateTime  @updatedAt
  branch    Branch    @relation(fields: [branchId], references: [id])
  branchId  Int
  problems  Problem[]
}

model Problem {
  id          Int       @id @default(autoincrement())
  createdAt   DateTime  @default(now())
  updatedAt   DateTime  @updatedAt
  auditable   Auditable @relation(fields: [auditableId], references: [id])
  auditableId Int
}
```

### Environment & setup

- OS: Ubuntu 18.04 (via WSLv1, Windows 10)
- Database: SQLite
- Node.js version: v16.13.1

### Prisma Version

```
```

aniravi24:
Closes #10894

As of TS 4.4, TypeScript [has the ability](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) to treat an optional property differently from undefined. As mentioned in the issue description, Prisma's types need to explicitly mention undefined to work with this TS feature - otherwise, you can't pass undefined to any optional args or object properties even when it's actually allowed at runtime.

This change should be fully backward-compatible, and the client test suite as well as the integration test suite seem to run fine locally.

This is the first time I've made a PR to Prisma or read the source code, so I may have gotten something wrong or missed something important.

The primary change is having the writer append `| undefined` if the argument or object property is optional. As far as I'm aware, this should always be valid, and any existing Prisma code should be unaffected by this.

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -678,7 +678,7 @@ export class PrismaClient<
    */
   $transaction<P extends Prisma.PrismaPromise<any>[]>(arg: [...P]): Promise<UnwrapTuple<P>>
 
-  $transaction<R>(fn: (prisma: Omit<this, "$connect" | "$disconnect" | "$on" | "$transaction" | "$use">) => Promise<R>, options?: { maxWait?: number, timeout?: number }): Promise<R>
+  $transaction<R>(fn: (prisma: Omit<this, "$connect" | "$disconnect" | "$on" | "$transaction" | "$use">) => Promise<R>, options?: { maxWait?: number | undefined, timeout?: number | undefined } | undefined): Promise<R>
 
   /**
    * Executes a raw MongoDB command and returns the result of it.
@@ -1515,7 +1515,7 @@ export namespace Prisma {
     /**
      * Select specific fields to fetch from the UserCountOutputType
      */
-    select?: UserCountOutputTypeSelect | null
+    select?: UserCountOutputTypeSelect | null | undefined
   }
 
 
@@ -1558,7 +1558,7 @@ export namespace Prisma {
     /**
      * Select specific fields to fetch from the EmbedHolderCountOutputType
      */
-    select?: EmbedHolderCountOutputTypeSelect | null
+    select?: EmbedHolderCountOutputTypeSelect | null | undefined
   }
 
 
@@ -1601,7 +1601,7 @@ export namespace Prisma {
     /**
      * Select specific fields to fetch from the MCountOutputType
      */
-    select?: MCountOutputTypeSelect | null
+    select?: MCountOutputTypeSelect | null | undefined
   }
 
 
@@ -1644,7 +1644,7 @@ export namespace Prisma {
     /**
      * Select specific fields to fetch from the NCountOutputType
      */
-    select?: NCountOutputTypeSelect | null
+    select?: NCountOutputTypeSelect | null | undefined
   }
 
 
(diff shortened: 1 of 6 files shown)


aniravi24:
I forgot to include the scope in the commit (i.e. `fix(client)`). I can rebase and force push if needed, let me know.

dimatakoy:
any updates on this?

aqrln:
Thanks a lot for the PR @aniravi24. Our current thinking is that allowing explicit `undefined` to skip parts of the query was a mistake and a footgun, which we saw many times causing users to accidentally lose their data. The direction we are moving to is the opposite of what this PR proposes, and, in fact, we recommend using `exactOptionalPropertyTypes` to prevent this pattern statically at type level. To forbid this at run time, we have a preview feature called `strictUndefinedChecks` that throws a validation error when it detects explicit `undefined` values. This will be the default behaviour in a future major version. As a safer replacement, we have the `Prisma.skip` marker to conditionally skip parts of queries.

I'm going to close this PR because it doesn't align with our current thinking and design goals, but we appreciate the enthusiasm and the work you've done, and hope to see you contribute in the future.

PR #18654, opened 2023-04-06
feat(cli): intro warning (29) support for null-sorted index

Druue:
fixes: https://github.com/prisma/schema-team/issues/383

~~This will need to be updated once https://github.com/prisma/prisma-engines/pull/3853 is in here~~


packages/migrate/src/commands/DbPull.ts
```diff
@@ -482,6 +482,8 @@ ${`Run ${chalk.green(getCommandWithExecutor('prisma generate'))} to generate Pri
           message += warning.affected.map((it) => `- View "${it.view}"`).join('\n')
         } else if (warning.code === 27) {
           message += warning.affected.map((it) => `- Model "${it.model}"`).join('\n')
+        } else if (warning.code === 29) {
+          message += warning.affected.map((it) => `- Index "${it.indexName}", Column: "${it.columnName}"`)
```
aqrln, line 486: To me it feels like we don't need a colon after "Column" here because there's no colon after "Index", but no strong opinion, I'm fine with it either way.

aqrln (approved):

packages/migrate/src/commands/DbPull.ts
Druue: Ah, yeah, you're right. There's a few other places we use it but it's not consistent so I'll just remove it here

janpio (approved):

PR #18657, opened 2023-04-06
feat(LibraryEngine): load library synchronously

SevInf:
Ensures that library starts and finishes loading in constructor,
allowing it to take advantage of higher CPU allocation for init code on
serverless platforms.


SevInf:
I don't think we are going to pursue this anymore - synchronous loading comes with a bunch of disadvantages and there are other ways to ensure library is completely loaded during `init` phase.

PR #18694, opened 2023-04-10
fix(generator-helper): Fix DMMF field dbName name and type

luxaritas:
Cross-referencing with https://github.com/prisma/prisma-engines/blob/main/query-engine/dmmf/src/serialization_ast/datamodel_ast.rs, it should be singular not plural and is indeed a single string, not an array.


aqrln:
Thank you for catching this @luxaritas! You are totally right here.

This field is currently used in `packages/cli/src/Doctor.ts`, could you please update it there as well? This wasn't caught by TypeScript because of the index signature in the `Field` interface. I can also confirm that `prisma doctor` currently ignores added or removed `@map` annotations in the schema, which is fixed by making it use the correct `dbName` field instead of the non-existent `dbNames`. I'll open an issue for that.

luxaritas:
Done. If you want me to add a Jest test for that, I'll need a bit of guidance - I don't see where the CLI fixtures are located

aqrln:
@luxaritas the CLI fixtures are located in `packages/cli/src/__tests__/fixtures` and the few existing tests for this command are in `packages/cli/src/__tests__/commands/Doctor.test.ts`, however it's not really required to add a test for this unless you want to. `prisma doctor` is an undocumented legacy command that will probably be removed in future major releases.

packages/cli/src/Doctor.ts
```diff
@@ -107,7 +107,7 @@ ${bold('Examples')}
       remoteModel: remoteModels[localModel.dbName ?? localModel.name],
     }))
 
-    const getFieldName = (f: DMMF.Field) => (f.dbNames && f.dbNames.length > 0 ? f.dbNames[0] : f.name)
+    const getFieldName = (f: DMMF.Field) => (f.dbName ?? f.name)
```
aqrln, line 110:
```suggestion
    const getFieldName = (f: DMMF.Field) => f.dbName ?? f.name
```

Prettier removes the parens here, not sure why we don't catch this on CI.

luxaritas: Made the tweaks in GH web since they were so small, and was indeed expecting CI to catch something like that. :)

luxaritas:
> prisma doctor is an undocumented legacy command that will probably be removed in future major releases

Sounds good - seems reasonable to leave it as-is

aqrln (approved):
LGTM, thanks for the contribution!

Restarted one flaky test (server couldn't bind to a port which was already in use), and let's merge.

PR #18728, opened 2023-04-13
ci: improve benchmarks stability

art049:
This stabilizes the benchmark results obtained with CodSpeed by making v8 engine optimizations a bit more predictable and consistent across runs.

It could also be worth it to purge the existing results since the execution is a bit different now. Let me know if it's okay to clean them(this feature should come in the UI as well soon).

cc @janpio


scripts/bench.ts
```diff
@@ -20,7 +20,8 @@ async function main() {
 
 async function run(benchmarks: string[]) {
   for (const location of benchmarks) {
-    await execa.command(`node -r esbuild-register ${location}`, {
+    const v8Flags = process.env.CODSPEED_V8_FLAGS ?? '' // Flags defined while running with CodSpeed
```
aqrln, line 23: From what I see in https://github.com/CodSpeedHQ/action/blob/4cb151de0c852c38201e37b3ed77299b3e27b35f/src/run.ts#L95, this disables TurboFan (`--no-opt`). Would a better alternative be making a certain number of warm-up runs, and then measuring the time it takes to run the optimised benchmark code? Or was it tried and turned out to still not be stable enough?

art049: Yes, I gave warmup runs a try but it ends up not being extremely consistent across runs. Sometimes the optimization was done, in others no optimization happened and also sometimes optimization was done during the measurement.
I also tried to force the optimization with native v8 calls (with `%OptimizeFunctionOnNextCall` for example) but it wasn't stable either, especially in async contexts.

From those experiments, the measurement without optimization is the most stable one and still reflects the code performance. But we'll work on it since having both the unoptimized and the optimized performance results would be interesting, for example, to highlight code segments that can't be optimized.

aqrln: > Yes, I gave warmup runs a try but it ends up not being extremely consistent across runs.

Can you reproduce that with other JavaScript code or is it a problem with our benchmarks specifically?

> Any reason why disabling is a bad idea @aqrln?
> We are not trying to measure the actual time the code takes, but to get a stable state that we can observe over time.

The problem is that optimized and unoptimized code may generally behave very differently, so I wouldn't trust the comparisons made with unoptimized code to generalize to the normal optimized code too. While intuitively it does sound right that we should be able to assume that the same change should be either faster or slower regardless of optimizations (at least as long as we say just "faster/slower" and not "how many times faster/slower"), I'm not really sure if that's generally the case or not even when there are no deopts. And, naturally, when something hinders optimizations, the performance characteristics will change quite a lot.

For example, consider inline caches. Here's a quick example I made that would show 2x performance improvement without optimizations (assuming "first" and "second" are old and new behavior respectively) when it is actually more than 20x performance regression: https://github.com/aqrln/benchmark-ic-demo/blob/main/bench.js

```
~/prisma/benchmarks-opt $ node bench.js 
first x 2,961,402 ops/sec ±1.67% (98 runs sampled)
second x 121,662 ops/sec ±2.01% (95 runs sampled)
first is 24.34 times faster than second

~/prisma/benchmarks-opt $ node --no-opt bench.js
first x 13,338 ops/sec ±0.26% (100 runs sampled)
second x 28,526 ops/sec ±0.28% (98 runs sampled)
second is 2.14 times faster than first
```

janpio: Maybe we should add this variation suggested by @art049 as a second run then?
Would give us a very stable one, and one that is currently flaky until we figure out a way to make that more stable a well.

art049: > For example, consider inline caches. Here's a quick example I made that would show 2x performance improvement without optimizations (assuming "first" and "second" are old and new behavior respectively) when it is actually more than 20x performance regression: https://github.com/aqrln/benchmark-ic-demo/blob/main/bench.js

Thanks for this example @aqrln. Super interesting edge case! 

I'm still working on having more stable  results with optimisation enabled.

If you want, we can try with optimization enabled although the results wont be as stable for now. 
It should give a 5-10% variability in the results. Anyway, it will be more reliable than it initially was.

Would it be interesting on your side to have both warm and cold run comparisons ?

aqrln: > I'm still working on having more stable results with optimisation enabled.
> If you want, we can try with optimization enabled although the results wont be as stable for now.
> It should give a 5-10% variability in the results. Anyway, it will be more reliable than it initially was.

That sounds great! 5–10% variability doesn't sound too bad, I think we could live with this already. I think it might be more fruitful at this point to improve the reliability further using statistical methods (i.e. perform some kind of significance testing like t-test with the null hypothesis that there's no difference in performance between two versions, and only report difference if the p-value is small enough). On a related note, since you'll probably need to upload and store the whole distribution of the run times and not just the mean to make it possible, it would be really nice to be able to see a histogram for any benchmark and commit in UI too.

> Would it be interesting on your side to have both warm and cold run comparisons ?

Good question. I believe the usual approach is to only consider the warm runs when benchmarking JIT-compiled languages but I can see how we could be interested in cold runs specifically, for example when benchmarking initialization code. I wonder if this should be something for the benchmark code to specify on a case-by-case basis.

Jolg42 (approved):

aqrln (approved):
Let's merge this for now to reduce noise in the PRs and make the benchmarks stable. Afterwards we'll have to make this work with TurboFan enabled so we can trust the results.

PR #18759, opened 2023-04-15
fix(client): add console error log for caching on top of error

millsp:
Improves current logic by not only throwing an error, but also `console.error` the error message before throwing it. This will make it  easier to read in Vercel or Netlify logs when it occurs.

### DIFF
--- a/packages/client/src/runtime/core/init/checkPlatformCaching.test.ts
+++ b/packages/client/src/runtime/core/init/checkPlatformCaching.test.ts
@@ -7,17 +7,17 @@ test('generated via postinstall on vercel', () => {
 
   expect(() => checkPlatformCaching({ postinstall: true, ciName: 'Vercel', clientVersion: '0.0.0' }))
     .toThrowErrorMatchingInlineSnapshot(`
-    We have detected that you've built your project on Vercel, which caches dependencies.
+    Prisma has detected that this project was built on Vercel, which caches dependencies.
     This leads to an outdated Prisma Client because Prisma's auto-generation isn't triggered.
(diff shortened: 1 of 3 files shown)


janpio (approved):

packages/client/src/runtime/core/init/checkPlatformCaching.test.ts
```diff
@@ -1,31 +1,87 @@
...
+    ]
+  `)
 })
 
 test('generated on vercel', () => {
+  const consoleMock = jest.spyOn(console, 'error').mockImplementation()
+
   expect(() => checkPlatformCaching({ postinstall: false, ciName: 'Vercel', clientVersion: '0.0.0' })).not.toThrow()
+
+  expect(consoleMock.mock.calls[0]).toMatchInlineSnapshot(`
```
aqrln, line 29: Why do we print an error in this case?

janpio: It ensures that the _error_ is shown in a readable format in Netlify and Vercel logs UI instead of being hidden in the general "crash" output or whatever the user has configured their app to do.

aqrln: Yes, but my question is about this test where `postinstall` is `false` and we check that no error is thrown.

aqrln: Ah, this makes sense now. Thank you!

aqrln (approved):

Issue #12691, opened 2022-04-06
Remove the implicit "native" binaryTarget


DusanDimitric:
### Bug description

I want to leave out my "native" binaryTarget, and build the client explicitly for the targets I declard in `schema.prisma`.

This is a problem because when multiple people (with different Operating Systems) are working on the same codebase will generate different clients - but they should be exactly the same.

Doc links:
https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#specify-custom-binarytargets-to-ensure-compatibility-with-the-os

### How to reproduce

For example, in this case:

```
generator client {
  provider      = "prisma-client-js"
  binaryTargets = ["rhel-openssl-1.0.x"]
}
```

I want the client built ONLY for `"rhel-openssl-1.0.x"` and not for my `"native"` target as well.

The `"native"` target is added implicitly for some reason.

### Prisma information

`schema.prisma`:
```
generator client {
  provider      = "prisma-client-js"
  binaryTargets = ["rhel-openssl-1.0.x"]
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

model User {
  uuid      String   @id @db.Uuid
  createdAt DateTime @default(now())
}
```

### Environment & setup

- OS: Mac/Ubuntu/Windows
- Database: PostgreSQL (not relevant)
- Node.js version: v16.13.2

### Prisma Version

package.json:
```
    ...
    "@prisma/client": "^3.12.0",
    "prisma": "^3.12.0",
    ...
```

janpio:
This would indeed be the logical behavior, based on how `native` is documented: https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#specify-the-prisma-client-js-generator-with-the-default-output-previewfeatures-enginetype-and-binarytargets

DusanDimitric:
The thing is - when I specify:
```
binaryTargets = ["debian-openssl-1.1.x"]
```
I want only the `"debian-openssl-1.1.x"` client to be created.

I am on MacOS and my Colleague on Debian - when we generate Prisma Clients we get different output - it should be the same with the configuration above, right?

janpio:
Yes, that is what I confirmed (or tried to at least) with my comment above 😆 We'll look into this. 
Changing this of course would not be that easy, as quite a few setups out there of course could rely on this current behavior - accidentally or by design.

dejanvasic85:
[...]

PR #18806, opened 2023-04-18
feat(cli): add `--binary-target` option to `prisma generate`

### LINKED ISSUE #15267: Allow binaryTargets as a CLI option
## Problem

Having binaryTargets as an option only in the schema file means that we need to specify multiple targets for all our environments. e.g. Locally I may be using "darwin" but I want to deploy "rhel-openssl-1.0.x" only. 

At the moment, this is causing us headaches when we want to deploy Nextjs to Vercel which uses serverless environment and size is limited. 

## Suggested solution

CLI option:
```npx prisma genearate --binaryTargets "rhel-openssl-1.0.x"```

## Alternatives

In a github workflow action, we are having to change the contents of the schema file before running the `generate` method which would remove the targets like "darwin" that we need locally.

## Additional context

https://github.com/prisma/prisma/issues/12691#issuecomment-1242762626

### PR DESCRIPTION
**Pull request by aqrln**
Closes https://github.com/prisma/prisma/issues/15267

### DIFF
--- a/packages/cli/src/__tests__/commands/Generate.test.ts
+++ b/packages/cli/src/__tests__/commands/Generate.test.ts
@@ -1,11 +1,21 @@
 import path from 'node:path'
 
 import { defaultTestConfig } from '@prisma/config'
-import { BaseContext, jestConsoleContext, jestContext } from '@prisma/get-platform'
+import {
+  BaseContext,
+  BinaryTarget,
+  getBinaryTargetForCurrentPlatform,
+  getNodeAPIName,
+  jestConsoleContext,
+  jestContext,
+} from '@prisma/get-platform'
 import { ClientEngineType, getClientEngineType } from '@prisma/internals'
+import { match } from 'ts-pattern'
 
 import { Generate } from '../../Generate'
 import { promotions, renderPromotion } from '../../utils/handlePromotions'
+
+const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
 
 const ctx = jestContext.new().add(jestConsoleContext()).assemble()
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/cli/src/__tests__/commands/Generate.test.ts [hunk 1]
```diff
@@ -191,6 +192,55 @@ describe('--schema from parent directory', () => {
...
+      .with(ClientEngineType.Library, () => getNodeAPIName(platform, 'fs'))
+      .with(
+        ClientEngineType.Binary,
+        () =>
+          // TODO: this is duplicated throughout the codebase, factor it out
+          `query-engine-${platform}${platform === 'windows' ? '.exe' : ''}`,
+      )
+      .exhaustive()
+
+  it('must not implicitly include "native"', async () => {
```
aqrln, line 207: This makes the CLI option a non-breaking solution to https://github.com/prisma/prisma/issues/12691, suggested by @dejanvasic85 in the comments.

aqrln: Actually after checking I don't even think what's written in https://github.com/prisma/prisma/issues/12691 is true (at least anymore): we don't implicitly add `native` to `binaryTargets` in schema either but print a warning instead. `native` is only added implicitly when no binary targets are specified in the schema.

packages/internals/src/get-generators/getGenerators.ts
```diff
@@ -124,6 +127,15 @@ export async function getGenerators(options: GetGeneratorOptions): Promise<Gener
     throw new Error(missingDatasource)
   }
 
+  if (binaryTargetsOverride !== undefined) {
+    for (const generator of config.generators) {
+      generator.binaryTargets = binaryTargetsOverride.map((target) => ({
```
aqrln, line 132: This makes the CLI print the warning like `Warning: Your current platform `darwin-arm64` is not included in your generator's `binaryTargets` configuration ["rhel-openssl-1.0.x"]` which must be skipped.

aqrln: (converting to draft for now because of this)

Jolg42: Maybe check how that changes once https://github.com/prisma/prisma/pull/19330 is merged?

janpio (changes requested):
Needs `chalk` fixes, see comment

wmadden:
Closing this for now @aqrln, feel free to reopen

aqrln:
Yeah, there was a bit more work involved than just a rebase. I don't think we care too much about it to spend time given that query compiler and WASM make this obsolete.

PR #18817, opened 2023-04-19
ci: remove flakiness (1st pass)

millsp:
Fix `jest` transpilation pattern which caused the CI to go OOM, slow down, and eventually timeout.
I tried with doing four runs and tests did not timeout, except for one unrelated download issue.


packages/client/tests/functional/jest.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = () => {
     // That includes all generated clients as well. So, unless we ignore them, they'd be kept in memory until test process
     // is finished, even though they are needed for 1 test only
     transformIgnorePatterns: [
-      '.*[\\.]generated[\\/]@prisma[\\/]client[\\/].*',
```
millsp, line 17: Looks like the `.*` didn't work, removing it achieves the same result anyways.

PR #18828, opened 2023-04-19
perf(client): Remove `mappings` from base dmmf

SevInf:
In runtime,`mappings` part was used only for validating supported actions for each model. Engine already does the validation for this and it does not seem like it is needed in **runtime** for any thrird-party projects so it looks like it is safe to remove. That improves `require` time a little bit, mostly on larger schemas:

image

Fix prisma/client-planning#320

### DIFF
--- a/packages/client/src/runtime/core/model/applyModel.ts
+++ b/packages/client/src/runtime/core/model/applyModel.ts
@@ -69,7 +69,7 @@ function modelMetaLayer(dmmfModelName: string): CompositeProxyLayer {
 function modelActionsLayer(client: Client, dmmfModelName: string): CompositeProxyLayer<string> {
   // we use the javascript model name for display purposes
   const jsModelName = dmmfToJSModelName(dmmfModelName)
-  const ownKeys = getOwnKeys(client, dmmfModelName)
+  const ownKeys = Object.keys(DMMF.ModelAction).concat('count')
 
   return {
     getKeys() {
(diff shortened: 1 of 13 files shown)


packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -69,7 +69,7 @@ function modelMetaLayer(dmmfModelName: string): CompositeProxyLayer {
 function modelActionsLayer(client: Client, dmmfModelName: string): CompositeProxyLayer<string> {
   // we use the javascript model name for display purposes
   const jsModelName = dmmfToJSModelName(dmmfModelName)
-  const ownKeys = getOwnKeys(client, dmmfModelName)
+  const ownKeys = Object.keys(DMMF.ModelAction).concat('count')
```
millsp, line 72: What would happen if I try to call `aggregateRaw` on `postgresql`?

SevInf: This error:

```
PrismaClientUnknownRequestError:
Invalid `prisma.user.aggregateRaw()` invocation in
/Users/serhii/projects/prisma/reproductions/workbench/index.ts:7:21

  4
  5 async function main() {
  6   // @ts-ignore
→ 7   await prisma.user.aggregateRaw(
Operation 'aggregateRaw' for model 'User' does not match any query.
```

millsp: I don't think it's an issue, but more out of curiosity. If you think like this can/should be improved, feel free to do it :)

SevInf: Ah, right, mine was engine-side for JSON. I'll adjust GraphQL error to be more comprehensible.

SevInf: Now for graphql it is:

```
PrismaClientValidationError:
Invalid `prisma.user.aggregateRaw()` invocation in
/Users/serhii/projects/prisma/reproductions/workbench/index.ts:7:21

  4
  5 async function main() {
  6   // @ts-ignore
→ 7   await prisma.user.aggregateRaw(
Model `User` does not support `aggregateRaw` action.

```

millsp: Nice, thanks. Could you please add a functional test for these two?

millsp (approved):

PR #18834, opened 2023-04-19
ci: fix isolated test flakiness

millsp:
Some old tests don't have proper database isolation on GitHub CI as they should, but they had it properly setup on BuildKite. I fixed the URL of the DB to be the isolated one. Ideally, we would port these to the new setup to benefit from type-checks and db isolation, so this is for their time being. Another alternative would be to to rub them in band, but that takes more time.

https://www.notion.so/prismaio/Findings-on-flaky-tests-de5b979a8441402784ae6eedd270f9dd?pvs=4#dc5a2b33836f4c26b332457fb9cae772

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,9 +42,9 @@ on:
       - '*.bench.ts'
       - 'scripts/ci/publish.ts'
       - 'graphs/**'
-#   schedule:
-#     # https://crontab.guru - “At every 15th minute past every hour from 3 through 5.”
-#     - cron: '*/15 3-5 * * *'
+  #   schedule:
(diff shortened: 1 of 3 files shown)


jkomyno:
Note: this PR currently fails with snapshot drifts such as:

```
    Expected substring: "Error querying the database: db error: FATAL: sorry, too many clients already"
    Received string:    "Can't reach database server at `postgres_isolated`:`5432`·
    Please make sure your database server is running at `postgres_isolated`:`5432`."
```

millsp:
I think this is the wrong fix, will dig a bit more into it. Initially this PR's change was to run in bad, which led me to believe there wasn't proper isolation. I think I need to dig a bit more.

PR #18861, opened 2023-04-21
perf(client): Pre-index DMMF datamodel during generation

SevInf:
In the constructor, we spend some time indexing all models and types
by their name, which on a larger schemas has an impact on startup
performance. This PR replaces base-dmmf with a set of 3 pre-indexed
maps.

Win mostly becomes visible for large schemas:

image

For backward compatibility, `Prisma.dmmf` export will be lazily computed
using runtime datamodel if requested.

Close prisma/client-planning#321


packages/client/src/generation/utils/buildDMMF.ts
```diff
@@ -1,52 +1,48 @@
...
 }
 
 /**
- * Build declaration for regular DMMF exports
- * @param dmmf
+ * Given DMMF models, computes runtime datamodel from it and embeds
+ * it into generated client. Creates lazy `Prisma.dmmf` property for backward
+ * compatibility, which will dynamically compute DMMF.Datamodel from runtime
+ * datamodel when accessed.
+ * Note: Prisma client itself never uses `Prisma.dmmf` and uses runtime datamodel
```
millsp, line 34: Do you think this can be added to Prisma 5 candidates?

packages/client/src/runtime/core/runtimeDataModel.ts
```diff
@@ -0,0 +1,51 @@
...
+function buildMapForRuntime<T extends { name: string }>(list: T[]): Record<string, Omit<T, 'name'>> {
+  const result: Record<string, Omit<T, 'name'>> = {}
+  for (const { name, ...rest } of list) {
+    result[name] = rest
+  }
+  return result
+}
+
+export function defineDmmfProperty(target: object, runtimeDataModel: RuntimeDataModel) {
+  const dmmfLazy = lazyProperty(() => runtimeDataModelToBaseDmmf(runtimeDataModel))
```
millsp, line 32: Do we snapshot this somewhere in a test? If not would you mind checking before and after this PR and adding a test?

packages/client/src/runtime/utils/datmodelBuilder.ts
millsp, line 1: File is called `datmodelBuilder` instead of `dataModelBuilder`

packages/client/src/runtime/utils/datmodelBuilder.ts
```diff
@@ -37,12 +37,10 @@ export function model(name: string, fields: DMMF.Field[]): DMMF.Model {
...
 }
 
-export function baseDmmf({ models }: { models: DMMF.Model[] }) {
-  return new DMMFDatamodelHelper({
-    datamodel: {
-      models,
-      enums: [],
-      types: [],
-    },
+export function runtimeDataModel({ models }: { models: DMMF.Model[] }): RuntimeDataModel {
```
millsp, line 40: Is this just for tests? If so could it be moved out of the runtime folder utils?

packages/generator-helper/src/dmmf.ts
```diff
@@ -58,7 +58,7 @@ export namespace DMMF {
     uniqueIndexes: uniqueIndex[]
     documentation?: string
     primaryKey: PrimaryKey | null
-    [key: string]: any // safe net for additional new props // TODO: remove this and the others, not safe
```
millsp, line 61: 😄

millsp (approved):

packages/client/src/runtime/core/runtimeDataModel.ts
SevInf: Added a test (with the snapshot done on `main`)

packages/client/src/runtime/utils/datmodelBuilder.ts
SevInf: moved to `src/testUtils`

PR #18874, opened 2023-04-21
fix(client): retry on engine start failure (binary)

millsp:
Under certain circumstances, it is possible that the Binary Engine will not be able to start correctly. The theoretical case is one that causes flakiness by a conflict of port reservation for the Binary Engine's http server between multiple processes/clients.

In other words, this hacky [code](https://github.com/prisma/prisma/blob/f43bc46b9e37a1a4cca1ba9305504cca713a8667/packages/engine-core/src/binary/BinaryEngine.ts/#L550-L562) starts a server in order to get the next available free port to give the QE for its http server. The problem with that is that there is a small window between the moment the port is found and is still free and the moment it is given to the engine and borrowed by its http server. In that window, there is a chance that another client will find the same port and start the engine before, or be using the hack that checks the free ports (which borrows the port for that simple check).

Adding to that, our n°1 flaky test starts hundreds of clients at once, likely to exacerbate the issues explained in the previous paragraph. So this PR proposes to simply retry on immediate initialization errors. Ideally the engine would decide on the port when it isn't specified and would return that port to the client. Maybe not worth the effort though. Ran 350 times locally without flakyness.

Example error (exit code 255)
https://github.com/prisma/prisma/actions/runs/4760387369/jobs/8461995272?pr=18859#step:11:6486

Example error (exit code 1)
https://github.com/prisma/prisma/actions/runs/4711467409/jobs/8355860114#step:12:1580
This error indicates that something went wrong wrt to spawning the binary, also note how the error contains no stderr logs. Since we don't have any context about this error, I propose to let the retry happen as well.


packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -453,10 +453,24 @@ ${chalk.dim("In case we're mistaken, please report this to us 🙏.")}`)
...
+        await this.internalStart()
+      } catch (e) {
+        if (e.retryable === true && retries.times > 0) {
+          retries.times--
+          await retryInternalStart()
+        }
+
+        throw e
+      }
+    }
```
Jolg42, line 468: Is it missing a `retryable = false` somewhere here?

millsp: Is that what you meant to ask?

millsp: Does that answer your question?

millsp:
(Notice other flaky tests on mongo)

packages/engine-core/src/binary/BinaryEngine.ts
Jolg42: Looks good!

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -447,10 +447,24 @@ ${dim("In case we're mistaken, please report this to us 🙏.")}`)
       await this.stopPromise
     }
 
+    const retries = { times: 10 }
```
Jolg42, line 450: Just curious, why 10?

Jolg42 (approved):

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -447,10 +447,24 @@ ${dim("In case we're mistaken, please report this to us 🙏.")}`)
...
+    const retryInternalStart = async () => {
```
Jolg42, line 451:
```suggestion
    // Retries were added in https://github.com/prisma/prisma/pull/18874#issue-1679183725 to avoid these 2 flaky issues:
    // 1. Query engine exited with code 1
    // 2. Query engine exited with code 255
    // PANIC in /root/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.23/src/server/server.rs:79:13
    // error binding to 127.0.0.1:36437: error creating server listener: Address already in use (os error 98)
    //
    const retryInternalStart = async () => {
```

packages/engine-core/src/binary/BinaryEngine.ts
Jolg42, line 450:
```suggestion
    // Retries were added in https://github.com/prisma/prisma/pull/18874#issue-1679183725 to avoid these 2 flaky issues:
    // 1. Query engine exited with code 1
    // 2. Query engine exited with code 255
    // PANIC in /root/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.23/src/server/server.rs:79:13
    // error binding to 127.0.0.1:36437: error creating server listener: Address already in use (os error 98)
    //
    const retries = { times: 10 }
```

millsp: I ran it 350+ times a few times locally, that's what it took to remove the flakiness. Terrible logic, but is mostly for tests.

Jolg42: Yes, we don't expect this to be as bad in production indeed.

PR #18897, opened 2023-04-24
fix(client): make generated client portable between operating systems

### LINKED ISSUE #15325: Allow sharing files generated by `prisma generate` from Windows to Mac/Linux
### Bug Description

EDIT: After investigating more, I discovered this is more easily reproducible if you run prisma generate from Windows. 

Running `prisma generate` on windows produces a few paths on the generated code, which would cause prisma to throw if using the generated code on a different machine. So, basically, if i generate files on a windows machine (say my personal machine), and try to use it on my work machine (a Mac), this would fail, because the paths generated are windows-specific.

### How to reproduce

1. Generate the client from one environment (tried it with windows)
2. Try to use generated files from other environment (Mac/Linux) - ideally, call it from within a file in the `src`folder while the schema itself is on root.
3. Prisma will throw a very cryptic error, saying `Error: ENOENT: no such file or directory, open 'schema.prisma'` - this is because prisma fails to resolve the path based on the "common paths" that it defines during codegen here:
```ts
const foundDirname = !regularDirname && findSync(process.cwd(), [
    "src\\generated\\prisma-client",
    "generated\\prisma-client",
], ['d'], ['d'], 1)[0]
```

This does not work, obviously - and, to be clear I'm not claiming it _should_, i just had a lot of trouble figuring out this error, although the only reason this doesn't not work, doesn't seem to be due to a very OS specific reason, it seems to be because when we call prisma generate on a windows environment, windows will use `\\` on most of the paths, on `generated/prisma-client/index.js`, If I fix the paths to use a single `/` rather than the double `\\`, it seems to work fine.

### Expected behavior

If possible, the ability to share generated files between environments, if not, perhaps a less cryptic error message would help. Something along the lines of `Unable to locate schema file, did you run prisma generate?`.

### Prisma information

```prisma
generator client {
  output        = "./generated/prisma-client"
  binaryTargets = ["native", "linux-musl", "darwin"]
}

datasource db {
  provider = "mongodb"
  url      = env("MONGO_URI")
}

model Post {
  id          String   @id @default(auto()) @map("_id") @db.ObjectId
  text       String
}
```

In `src/test.ts`
```ts
import { PrismaClient } from "../generated/prisma-client";
const prisma = new PrismaClient({ log: ["query"]  });
```

### Environment & setup
- OS: Windows (to run prisma generate) / Mac (to try to run the files generated on windows)
- Database: Mongo
- Node.js version: 16.13.2

### Prisma version
```
prisma                  : 4.3.1
@prisma/client          : 4.3.1
Query Engine (Node-API) : libquery-engine c875e43600dfe042452e0b868f7a48b817b9640b (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli c875e43600dfe042452e0b868f7a48b817b9640b (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core c875e43600dfe042452e0b868f7a48b817b9640b (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary           : prisma-fmt c875e43600dfe042452e0b868f7a48b817b9640b (at node_modules/@prisma/engines/prisma-fmt-darwin)
Format Wasm             : @prisma/prisma-fmt-wasm 4.3.0-32.c875e43600dfe042452e0b868f7a48b817b9640b
Default Engines Hash    : c875e43600dfe042452e0b868f7a48b817b9640b
Studio                  : 0.473.0
```

aqrln:
Fixes https://github.com/prisma/prisma/issues/15325

Doesn't transform `config.generator.output.value` because it's an absolute path and it's inherently non-portable regardless of whether the OS is the same.


packages/client/src/generation/generateClient.ts
```diff
@@ -272,7 +272,6 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
 
   // if users use a custom output dir
   if (copyRuntime || !path.resolve(outputDir).endsWith(`@prisma${path.sep}client`)) {
-    // TODO: Windows, / is not working here...
```
aqrln, line 275: Outdated, refers the previous line. Introduced in https://github.com/prisma/prisma/commit/871cc4d13856f2697d8b8772c569f57fb2d3d8ec, fixed in https://github.com/prisma/prisma/commit/e9647573c5e314fc6cdb3ebe1d839bfc5a3cfcf2.

packages/internals/src/utils/path.ts
```diff
@@ -0,0 +1,15 @@
...
+ * treated as a path specific to the current platform, so backslashes will only
+ * be replaced with forward slashes on Windows. On other platforms, where a
+ * backslash is a valid filename character, it will be treated as such and will
+ * not be replaced.
+ */
+export function pathToPosix(filePath: string): string {
+  if (path.sep === path.posix.sep) {
+    return filePath
+  }
+  return filePath.split(path.sep).join(path.posix.sep)
```
millsp, line 14: So Windows->Unix will now work, will Unix->Windows work?

aqrln: Windows supports both `/` and `\`, it just uses `\` by convention. Normalizing everything to `/` works on both Unix and Windows.

millsp (approved):

PR #18899, opened 2023-04-24
feat: stable types node and exit hooks

### LINKED ISSUE #17127: DeprecationWarning: Implicit coercion to integer for exit code is deprecated
### Bug description

i am running nodemon for an express server that uses prisma

every time nodemon restarts the server, i get this warning:

```console
(node:14591) [DEP0164] DeprecationWarning: Implicit coercion to integer for exit code is deprecated.
    at process.exit (node:internal/process/per_thread:200:7)
    at process.<anonymous> (/Users/dionsaur84/Sites/dio-framework/node_modules/@prisma/client/runtime/index.js:27577:17)
    at Object.onceWrapper (node:events:628:26)
    at process.emit (node:events:513:28)
```

this warning appears to be triggered by `@prisma/client/runtime/index.js:27577:17`:

```javascript
  installHook(event, shouldExit = false) {
    process.once(event, async (code) => {
      debug9(`exit event received: ${event}`);
      for (const listener of this.idToListenerMap.values()) {
        await listener();
      }
      this.idToListenerMap.clear();
      if (shouldExit && process.listenerCount(event) === 0) {
        process.exit(code); // <-- this line
      }
    });
  }
```

is there a way to surpress this? should i add an event listener for when my server closes to make sure prisma also disconects? how do i resolve this issue myself, or is it a prisma issue?

edit: i've added a `$disconnect` call to my server close listener and the warning still occurs:

```javascript
const server = app.listen(env.APP_PORT, () => {
    console.log(`app listening on http://localhost:${env.APP_PORT}`)
})

server.on('close', async () => {
    await prisma.$disconnect()
})
```

final edit: looks like this is only happening in node 19+ due to this change:

https://fossies.org/diffs/node/v19.2.0_vs_v19.3.0/lib/internal/process/per_thread.js-diff.html

jkomyno:
See [Slack thread](https://prisma-company.slack.com/archives/CUXLS0Z6K/p1681995929722919).
Fixes https://github.com/prisma/prisma/issues/17127.

This PR:
- aligns the devDependency on `@types/node` for the public Prisma packages to v14.18.42.
- fixes related type issues that emerged in ExitHooks.ts after stabilising the `@types/node` versions


packages/engine-core/src/library/ExitHooks.ts
```diff
@@ -48,8 +49,8 @@ export class ExitHooks {
     return this.idToListenerMap.get(id)
   }
 
-  private installHook(event: string, shouldExit = false) {
-    process.once(event, async (code) => {
+  private installHook(event: 'beforeExit' | 'exit' | NodeJS.Signals, shouldExit = false) {
```
SevInf, line 52: I think by that point, there is very little in common between the handler for signals and `exit`/`beforeExit` event and we would make this code much simpler by splitting `installHook` into two methods, one for exit events and one for signals. I needed to deal with `ExitHooks` problem in my other PR and took this approach: https://github.com/prisma/prisma/pull/18969/files#diff-24df258365237f8cecaaabdd47cbad94bf43846b3fbea549316e93f23d338505

jkomyno: Thanks for the feedback, I'll be splitting this `installHook` in two. At this point, I think we can also get rid of the `shouldExit` argument, and always apply the `shouldExit` callback on the hooks installed by `SIG*` signals. What do you think?

SevInf: Yep, `shouldExit` is always `true` for signals and always `false` for exit events, so totally makes sense 👍

packages/engine-core/src/library/ExitHooks.ts
```diff
@@ -48,20 +49,34 @@ export class ExitHooks {
...
-      if (shouldExit && process.listenerCount(event) === 0) {
-        process.exit(code)
+      // Only exit when there are no listeners left for this signal.
+      // If there is another listener, that other listener is responsible for exiting.
+      if (isSomeoneStillListening) {
+        return
       }
+
+      // the usual way to exit with a signal is to add 128 to the signal number
+      const exitCode = os.constants.signals[signal] + 128
```
Jolg42, line 69: People might rely on the old exitCode, not sure if we want to do this change in a standard release?

jkomyno:
Sigh, tests are failing due to unrelated snapshot changes

packages/engine-core/src/library/ExitHooks.ts
jkomyno: Considering this is a direct consequence of Node.js changing its behavior, that's not up to us.
Afaik, as long as we don't turn a `0` exit code into another number, everything's fine

SevInf: I am with Alberto on this, right now, you applications exit code changes to 0 if you set our `beforeExit` hook, this change would restore them to what they are by default.

SevInf (approved):

packages/engine-core/src/library/ExitHooks.ts
Jolg42: Not sure about this since it was always 0 until this PR.
Also, we would at least need a test for this? It's an important bit that is still untested based on the CI run here?

SevInf: It was not always 0. We set our signal handlers only when you add `beforeExit` event listener. If you don't do it, exit code will be default one, 128 + signal.

Jolg42: Which means it was always 0 for people using the `beforeExit` event listener 😄

PR #18900, opened 2023-04-24
perf(client): Replace `chalk` with `kleur/color`

SevInf:
Signgicantly reduces bundle size (kleur is <1KB in size, which reduces library runtime size from 280K to 255K) and
improves `require(runtime)` performance by 5-10ms:

image

Few implementation notes:
- `kleur` is limited to 16 colors, in particular, `*Bright` versions of the colors are not supported. I replaced them with non-bright versions, which still look ok in the several places I've checked, and IMHO, is a fair price to pay for size/startup performance improvements. If you disagree - let me know.
- `migrate` and `cli` technically did not need to migrate, but since we moving and sharing code between different packages from time to time, I've decided it would be better if we handle colors the same way everywhere and went for complete `chalk` removal.

Close prisma/client-planning#341


packages/client/src/__tests__/json.test.ts
```diff
@@ -1,13 +1,10 @@
 import { getDMMF } from '@prisma/internals'
-import chalk from 'chalk'
 
 import { DMMFClass, makeDocument, objectEnumValues, transformDocument } from '../runtime'
 import { MergedExtensionsList } from '../runtime/core/extensions/MergedExtensionsList'
 
 const describeIf = (condition: boolean) => (condition ? describe : describe.skip)
 
-chalk.level = 0
```
SevInf, line 9: Not necessary, because we strip ANSI escapes in this suite anyway.

packages/cli/src/bin.ts
```diff
@@ -88,11 +86,6 @@ const args = arg(
 // Redact the command options and make it a string
 const redactedCommandAsString = redactCommandArray([...commandArray]).join(' ')
 
-// because chalk ...
-if (process.env.NO_COLOR) {
```
SevInf, line 92: No need for it now, `kleur` reads `NO_COLOR` env var internally.

packages/cli/src/Generate.ts
```diff
@@ -214,9 +212,9 @@ Please run \`prisma generate\` manually.`
         const generator = prismaClientJSGenerator.options?.generator
         const isDeno = generator?.previewFeatures.includes('deno') && !!globalThis.Deno
         if (isDeno && !generator?.isCustomOutput) {
-          throw new Error(`Can't find output dir for generator ${chalk.bold(
-            generator?.name,
-          )} with provider ${chalk.bold(generator?.provider.value)}.
+          throw new Error(`Can't find output dir for generator ${bold(generator!.name)} with provider ${bold(
```
aqrln, line 215: note to other reviewers if anyone else stumbles here too: `isDeno` implies `generator` is not nullish, so `!` is safe here

jkomyno: Can't this be turned into a type guard that uses `isDeno` to avoid `!`?

packages/client/src/runtime/utils/createErrorMessageWithContext.ts
```diff
@@ -24,11 +24,11 @@ type Colors = {
...
-  red: (str) => chalk.red(str),
-  gray: (str) => chalk.gray(str),
-  dim: (str) => chalk.dim(str),
-  bold: (str) => chalk.bold(str),
-  underline: (str) => chalk.underline(str),
+  red: (str) => red(str),
+  gray: (str) => gray(str),
+  dim: (str) => dim(str),
+  bold: (str) => bold(str),
+  underline: (str) => underline(str),
```
aqrln, line 31:
```suggestion
  red,
  gray,
  dim,
  bold,
  underline,
```

packages/engine-core/src/binary/BinaryEngine.ts
```diff
@@ -188,7 +186,7 @@ export class BinaryEngine extends Engine<undefined> {
 
     if (removedFlagsUsed.length > 0 && !process.env.PRISMA_HIDE_PREVIEW_FLAG_WARNINGS) {
       console.log(
-        `${chalk.blueBright('info')} The preview flags \`${removedFlagsUsed.join(
+        `${blue(bold('info'))} The preview flags \`${removedFlagsUsed.join(
```
aqrln, line 189: these log level labels should probably be consistently bold or normal font, I like the idea of making them all bold

aqrln (commented):
Halfway so far, I'll continue reviewing further. Thanks for doing this!

packages/internals/src/get-generators/getGenerators.ts
```diff
@@ -189,8 +189,8 @@ export async function getGenerators(options: GetGeneratorOptions): Promise<Gener
         } else {
           if (!generatorInstance.manifest || !generatorInstance.manifest.defaultOutput) {
             throw new Error(
-              `Can't resolve output dir for generator ${chalk.bold(generator.name)} with provider ${chalk.bold(
-                generator.provider,
+              `Can't resolve output dir for generator ${bold(generator.name)} with provider ${bold(
+                generator.provider.value!,
```
Jolg42, line 193: You found a bug? 😄

packages/migrate/package.json
```diff
@@ -43,15 +43,16 @@
...
     "execa": "5.1.1",
     "get-stdin": "8.0.0",
     "has-yarn": "2.1.0",
     "indent-string": "4.0.0",
+    "kleur": "4.1.5",
     "log-update": "4.0.0",
     "mariadb": "3.1.1",
     "mongoose": "6.7.0",
     "mssql": "9.1.1",
+    "ora": "5.4.1",
```
Jolg42, line 55: Since "spinner" moved to migrate this makes sense ✅

Jolg42 (approved):
I can't say that I read every line, but at least most 😅 
Looks great 🙌🏼

packages/migrate/src/commands/DbCommand.ts
```diff
@@ -1,43 +1,43 @@
...
+import { bold, dim, red } from 'kleur/colors'
 
 export class DbCommand implements Command {
   public static new(cmds: Commands): DbCommand {
     return new DbCommand(cmds)
   }
 
   private static help = format(`
-${process.platform === 'win32' ? '' : chalk.bold('🏋️  ')}Manage your database schema and lifecycle during development.
+${process.platform === 'win32' ? '' : bold('🏋️  ')}Manage your database schema and lifecycle during development.
```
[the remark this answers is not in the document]
aqrln: although not sure, maybe it might depending on the font

packages/get-platform/src/link.ts
```diff
@@ -1,8 +1,8 @@
-import chalk from 'chalk'
+import { underline } from 'kleur/colors'
 import terminalLink from 'terminal-link'
 
 export function link(url): string {
   return terminalLink(url, url, {
-    fallback: (url) => chalk.underline(url),
+    fallback: (url) => underline(url),
```
aqrln, line 6: nit: 
```suggestion
    fallback: underline,
```

(feel free to ignore though)

aqrln (approved):

packages/migrate/src/commands/DbCommand.ts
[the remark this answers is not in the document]
aqrln: there's quite a lot of these bold emojis as far as I see, maybe there was some intention behind it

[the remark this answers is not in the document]
SevInf: my guess would be, they used to be some plain unicode symbol for which bold made sense and got changed to emoji later

PR #18904, opened 2023-04-24
chore: same version for @types/node

Jolg42:
@microsoft/api-extractor patch version was bumped to avoid: ✕ unmet peer @types/node@^14.18.36: found 18.16.0

Triggered by 
```
// @ts-ignore: `rm` exists since Node.js 14.14.0, but doesn't seem to appear in all @types/node packages used by Prisma
```
From https://github.com/prisma/prisma/pull/18848#pullrequestreview-1394261574

We can use the latest version everywhere and clean up, it's not perfect but better than now

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,13 +28,13 @@
     "prepare": "is-ci || husky install"
   },
   "devDependencies": {
-    "@microsoft/api-extractor": "7.34.2",
+    "@microsoft/api-extractor": "7.34.4",
     "@sindresorhus/slugify": "1.1.2",
     "@slack/webhook": "6.1.0",
     "@types/benchmark": "2.1.2",
     "@types/fs-extra": "9.0.13",
     "@types/graphviz": "0.0.35",
-    "@types/node": "14.18.42",
+    "@types/node": "18.16.0",
     "@types/node-fetch": "2.6.3",
     "@types/redis": "2.8.32",
     "@types/resolve": "1.20.2",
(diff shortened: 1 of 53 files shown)


jkomyno:
Hi, let's not upgrade `@types/node` to `18.*`, which is very misleading, as our minimum supported Node.js version is `14.17.*`. This implies that we cannot trust the types exported by such a recent `@types/node` version.

In https://github.com/prisma/prisma/pull/18899, I had already stabilised the `@types/node` versions across the public `@prisma/*` packages to `14.18.42`, i.e., the version already in use that is the closest to `14.17.*`.

With the context provided in this comment, I think this PR can be closed.

jkomyno (changes requested):
Hi, let's not upgrade `@types/node` to `18.*`, which is very misleading, as our minimum supported Node.js version is `14.17.*`. This implies that we cannot trust the types exported by such a recent `@types/node` version.

In https://github.com/prisma/prisma/pull/18899, I had already stabilised the `@types/node` versions across the public `@prisma/*` packages to `14.18.42`, i.e., the version already in use that is the closest to `14.17.*`.

With the context provided in this comment, I think this PR can be closed.

Jolg42:
@jkomyno I don't think that using v14 types everywhere is a great solution either, there is no perfect solution here unfortunately 😢

jkomyno:
Note: https://github.com/prisma/prisma/pull/18899 was approved and merged, and likely conflicts with this PR.

PR #18913, opened 2023-04-25
chore: remove deprecated set output

Jolg42:
Related: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


Jolg42:
It looked like `core.setOutput` was deprecated (https://github.com/actions/toolkit/issues/1218) but it looks more complicated than that, it has a compatibility mode.
So I'll close it as it works perfectly for us here and I definitely prefer one line

PR #18922, opened 2023-04-25
perf: improve mapObjectValues by ~50%

Weakky:
## Overview

- Improves performance of mapObjectValues by ~50% (no proper benchmark was done). This function is used when processing engine's result. For a 1000 records each having 50 related records, we go from ~32-34ms to ~11-14ms (locally). Using `Object.fromEntries` + `Object.entries` iterated the dataset twice, which can get costly on large datasets.


aqrln (approved):

Weakky:
I don't have the rights to merge folks, so feel free to click the holy button!

PR #18947, opened 2023-04-26
ci(buildkite): try log managing sections

Jolg42:
Currently, logs look like this 🙈 
See https://buildkite.com/prisma/test-prisma-typescript/builds/21081#0187bd6c-2c45-4f99-b842-975bf97bf5c2
Screenshot 2023-04-26 at 15 43 58

Trying out:
https://buildkite.com/docs/pipelines/managing-log-output

It looks like this after for the testing pipeline:
Screenshot 2023-04-26 at 16 46 18

We can see the time each step takes too, which is much better than trying to guesstimate with the timestamps 🙌🏼


Jolg42:
I'll merge, so we can try it out and not have the "1MB truncation warning" forcing us to download or look at the raw version

PR #18986, opened 2023-04-27
This is a security test as part of a bug bouty, plese ignore.  pwtaxi

bazookaz1:
This is a security test as part of a bug bouty, plese ignore.  pwtaxi


janpio:
Closing so no one accidentally clicks "Approve and run".
Checking if the jobs that did not need approval could be a problem here - let us know via security@prisma.io if you learn  anything interesting please.

PR #18990, opened 2023-04-28
refactor: remove some promisify to fs.promises

Jolg42:
Available since Node 10~ 👴🏼 

https://nodejs.org/api/fs.html#promises-api
Version | Changes
-- | --
v14.0.0 | Exposed as require('fs/promises').
v11.14.0, v10.17.0 | This API is no longer experimental.
v10.1.0 | The API is accessible via require('fs').promises only.
v10.0.0 | Added in: v10.0.0

### DIFF
--- a/packages/cli/helpers/build.ts
+++ b/packages/cli/helpers/build.ts
@@ -3,13 +3,10 @@ import fs from 'fs'
 import { copy } from 'fs-extra'
 import lineReplace from 'line-replace'
 import path from 'path'
-import { promisify } from 'util'
 
 import type { BuildOptions } from '../../../helpers/compile/build'
 import { build } from '../../../helpers/compile/build'
 import { run } from '../../../helpers/compile/run'
-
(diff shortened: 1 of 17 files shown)


packages/cli/helpers/build.ts
```diff
@@ -33,20 +30,20 @@ const cliLifecyclePlugin: esbuild.Plugin = {
       })
 
       // we copy the contents from checkpoint-client to build
-      await copyFile(
+      await fs.promises.copyFile(
```
aqrln, line 33: We could import `fs` from `fs/promises` instead to avoid writing `fs.promises` everywhere, but either is fine with me.

aqrln (approved):

packages/cli/helpers/build.ts
Jolg42: Yes! this PR prepares the move if we want to, I need to work on other things now.

PR #18992, opened 2023-04-28
refactor(cli): better type and error for tryToReadDataFromSchema

Jolg42:
See [thread](https://prisma-company.slack.com/archives/C03BP8ED5MF/p1677270819410049)

### DIFF
--- a/packages/cli/src/utils/checkpoint.ts
+++ b/packages/cli/src/utils/checkpoint.ts
@@ -21,11 +21,11 @@ export async function runCheckpointClientCheck({
   command,
   telemetryInformation,
 }: {
-  schemaPath: string
   isPrismaInstalledGlobally: 'npm' | 'yarn' | false
   version: string
   command: string
   telemetryInformation: string
+  schemaPath?: string
 }): Promise<Check.Result | 0> {
   try {
     // SHA256 identifier for the project based on the Prisma schema path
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/checkpoint.ts
```diff
@@ -78,39 +78,48 @@ export async function runCheckpointClientCheck({
...
+    } catch (e) {
+      debug(
+        'Error from tryToReadDataFromSchema() while processing the schema. This is not a fatal error. It will continue without the processed data.',
+      )
+      debug(e)
     }
-  } catch (e) {
-    debug('Error from tryToReadDataFromSchema()')
-    debug(e)
   }
```
jkomyno, line 123: I'd simplify this part with:

```ts
  try {
    const schema = await getSchema(schemaPath)

    try {
      const config = await getConfig({
        datamodel: schema,
        ignoreEnvVarErrors: true,
      })

      if (config.datasources.length > 0) {
      }

      // Example 'prisma-client-js'
      schemaGeneratorsProviders = config.generators
        // Check that value is defined
        .filter((generator) => generator && generator.provider)
        .map((generator) => parseEnvValue(generator.provider))

      // restrict the search to previewFeatures of `provider = 'prisma-client-js'`
      // (this was not scoped to `prisma-client-js` before Prisma 3.0)
      const prismaClientJSGenerator = config.generators.find(
        (generator) => parseEnvValue(generator.provider) === 'prisma-client-js',
      )
      if (prismaClientJSGenerator && prismaClientJSGenerator.previewFeatures.length > 0) {
        schemaPreviewFeatures = prismaClientJSGenerator.previewFeatures
      }
    } catch (e) {
      debug(
        'Error from tryToReadDataFromSchema() while processing the schema. This is not a fatal error. It will continue without the processed data.',
      )
      debug(e)
    }
  } catch (e) {
    // fail silently if schema can't be read
  }
```

Having a nested try/catch is already hard to visually parse, let's avoid the noise of `let schema: string | undefined` and `if (schema) {}`, since we can.

Jolg42: Sounds better to me too

jkomyno (approved):

PR #19001, opened 2023-04-28
chore(ci): Reduce the number of jobs, move some of them to once a day

SevInf:
Following builds will be done once a day after this is merged:
- all binary engine tests
- json protocol with non-minimal supported version

If PR contains `ci test all` comment, all of those things will be
checked. If not, the job that will test them will run at 4am UTC from Monday to Friday.

Since we now need two jobs on different schedules, we need a way to
differentiate them: for that reason, both buildpulse and new nightly
jobs are implemented as separate workflows, that call original one
passing the reason (daily/buildpulse) as a parameter.

### DIFF
--- a/.github/workflows/daily-buildpulse.yml
+++ b/.github/workflows/daily-buildpulse.yml
@@ -1,0 +1,15 @@
+name: Daily BuildPulse run
+
+on:
+  workflow_dispatch:
+  schedule:
+    # https://crontab.guru - “At every 30th minute past every hour from 3 through 6.”
+    - cron: '*/30 3-6 * * *'
+
+jobs:
+  run_tests:
+    uses: ./.github/workflows/test.yml
+    with:
+      reason: buildpulse
+      jobTimeout: 70
+      ubuntuRunner: ubuntu-latest
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -11,6 +11,8 @@ jobs:
     uses: ./.github/workflows/test.yml
     with:
       reason: daily-test
+      jobTimeout: 70
+      ubuntuRunner: ubuntu-latest
   failure:
     needs:
       - run_tests
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -1,0 +1,30 @@
+name: Daily Test
+
+on:
+  workflow_dispatch:
+  schedule:
+    # 8am daily Mon-Fri
+    - cron: '0 8 * * 1-5'
+
+jobs:
+  run_tests:
+    uses: ./.github/workflows/test.yml
+    with:
+      reason: daily-build
+  failure:
+    needs:
+      - run_tests
+    if: ${{ failure() }}
+    name: Communicate failure
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set current job url in SLACK_FOOTER env var
+        run: echo "SLACK_FOOTER=<$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID|Click here to go to the job logs>" >> $GITHUB_ENV
+
+      - name: Slack Notification on Failure
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_TITLE: 'prisma/prisma daily build failed :x:'
+          SLACK_COLOR: '#FF0000'
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_CHANNEL: tmp-daily-build-notification-tests
--- a/.github/workflows/daily-test.yml
+++ b/.github/workflows/daily-test.yml
@@ -3,8 +3,8 @@ name: Daily Test
 on:
   workflow_dispatch:
   schedule:
-    # 8am daily Mon-Fri
-    - cron: '0 8 * * 1-5'
+    # 4am daily Mon-Fri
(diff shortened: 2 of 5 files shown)


SevInf:
Since new workflows won't be visible until we merge this to main, I'd suggest we do a review now and then test it later.

.github/workflows/test.yml
```diff
@@ -74,6 +86,15 @@ jobs:
         continue-on-error: true # This allow this job to fail (e.g. for scheduled runs), and then the `detect` step below will default to `all`
       - id: detect
         run: ./.github/workflows/scripts/detect-jobs-to-run.js <<< '${{ steps.files.outputs.all }}'
+      - name: Find command comment
+        uses: peter-evans/find-comment@v2
+        id: fc
+        if: ${{ github.event_name == 'pull_request' }}
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body-includes: 'ci build all'
```
Jolg42, line 95: Technically, it's more "ci test all"?

.github/workflows/test.yml
```diff
@@ -115,19 +136,20 @@ jobs:
...
 
     strategy:
       fail-fast: false
       matrix:
-        queryEngine: ['library', 'binary']
+        queryEngine: ${{ fromJson(needs.detect_jobs_to_run.outputs.queryEngine) }}
         engineProtocol: ['graphql', 'json']
         os: [buildjet-4vcpu-ubuntu-2004]
         node: [16, 18]
+        exclude: ${{ fromJson(needs.detect_jobs_to_run.outputs.excludeClient) }}
```
Jolg42, line 152: Nice!

.github/workflows/daily-test.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Set current job url in SLACK_FOOTER env var
+        run: echo "SLACK_FOOTER=<$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID|Click here to go to the job logs>" >> $GITHUB_ENV
+
+      - name: Slack Notification on Failure
+        uses: rtCamp/action-slack-notify@v2.2.0
+        env:
+          SLACK_TITLE: 'prisma/prisma daily build failed :x:'
+          SLACK_COLOR: '#FF0000'
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          SLACK_CHANNEL: tmp-daily-build-notification-tests
```
Jolg42, line 30: Usually we name the channel like `feed-` so maybe `feed-prisma-daily-build-failures`

SevInf: Yea, I'd move it to a proper channel once I'll be sure it is working, just did not want to spam actual channels meanwhile

.github/workflows/test.yml
SevInf: Changed!

.github/workflows/daily-test.yml
```diff
@@ -0,0 +1,30 @@
...
+  workflow_dispatch:
+  schedule:
+    # 8am daily Mon-Fri
+    - cron: '0 8 * * 1-5'
+
+jobs:
+  run_tests:
+    uses: ./.github/workflows/test.yml
+    with:
+      reason: daily-build
```
Jolg42, line 13: Nit: replace by `daily-test`

.github/workflows/daily-test.yml
```diff
@@ -0,0 +1,30 @@
+name: Daily Test
```
Jolg42, line 1: This one could also run on "ubuntuRunner" with a longer timeout actually?

.github/workflows/daily-test.yml
janpio: I think you can change this already now to avoid the later PR - the channel will be a new one anyway.

.github/workflows/daily-buildpulse.yml
```diff
@@ -0,0 +1,15 @@
...
+
+on:
+  workflow_dispatch:
+  schedule:
+    # https://crontab.guru - “At every 30th minute past every hour from 3 through 6.”
+    - cron: '*/30 3-6 * * *'
+
+jobs:
+  run_tests:
+    uses: ./.github/workflows/test.yml
```
janpio, line 11: Cool, didn't know this was a thing!

.github/workflows/test.yml
```diff
@@ -74,6 +91,15 @@ jobs:
         continue-on-error: true # This allow this job to fail (e.g. for scheduled runs), and then the `detect` step below will default to `all`
       - id: detect
         run: ./.github/workflows/scripts/detect-jobs-to-run.js <<< '${{ steps.files.outputs.all }}'
+      - name: Find command comment
+        uses: peter-evans/find-comment@v2
+        id: fc
```
janpio, line 96: more descriptive name maybe?

.github/workflows/test.yml
```diff
@@ -74,6 +91,15 @@ jobs:
...
+        if: ${{ github.event_name == 'pull_request' }}
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body-includes: 'ci test all'
+      - id: matrix
```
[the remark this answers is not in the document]
SevInf: Renamed to `determineBuildMatrixParams`

.github/workflows/test.yml
SevInf: Renamed to `findTestAllComment`

Jolg42 (approved):
Thanks a lot ✨

PR #19006, opened 2023-04-28
ci: prevent concurrent tests

millsp:
Turns out this was already enabled! That said, it was not enabled for benchmarks, so I activated it there. We also had some old cancellation logic that I ported to the newer native GH way of doing it.


.github/workflows/benchmark.yml
```diff
@@ -37,6 +37,10 @@ on:
 env:
   PRISMA_TELEMETRY_INFORMATION: 'prisma benchmark.yml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
millsp, line 42: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

.github/workflows/test.yml
```diff
@@ -53,6 +53,10 @@ env:
   # To hide "Update available 0.0.0 -> x.x.x"
   PRISMA_HIDE_UPDATE_MESSAGE: true
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'schedule' }}
```
millsp, line 58: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

millsp: `${{ github.event_name != 'schedule' }}` so that we don't cancel on scheduled runs (for BuildPulse)

aqrln (approved):

.github/workflows/test.yml
Jolg42, line 58:
```suggestion
  // Don't cancel if it's on schedule (used for BuildPulse for flaky tests detection)
  cancel-in-progress: ${{ github.event_name != 'schedule' }}
```

Jolg42 (approved):

jkomyno (approved):

PR #19008, opened 2023-04-28
ci: restore buildpulse nightly runs on free runners

millsp:
This PR re-enables BuildPulse in a cost effective way. We stopped running BuildPulse because of incurred costs on paid runners. This gives us much less confidence wrt to flaky tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -124,7 +124,7 @@ jobs:
   #
   client:
     timeout-minutes: ${{ github.event_name == 'schedule' && 100 || 35 }}
-    runs-on: ${{ github.event_name == 'schedule' && contains(matrix.os, 'buildjet') && 'ubuntu-latest' || matrix.os }}
+    runs-on: ${{ github.event_name != 'schedule' && contains(matrix.os, 'buildjet') && 'ubuntu-latest' || matrix.os }}
 
     needs: detect_jobs_to_run
     if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
@@ -238,7 +238,7 @@ jobs:
   #
   client-dataproxy:
     timeout-minutes: ${{ github.event_name == 'schedule' && 100 || 35 }}
-    runs-on: ${{ github.event_name == 'schedule' && contains(matrix.os, 'buildjet') && 'ubuntu-latest' || matrix.os }}
+    runs-on: ${{ github.event_name != 'schedule' && contains(matrix.os, 'buildjet') && 'ubuntu-latest' || matrix.os }}
 
     needs: detect_jobs_to_run
     if: ${{ contains(needs.detect_jobs_to_run.outputs.jobs, '-all-') || contains(needs.detect_jobs_to_run.outputs.jobs, '-client-') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -123,8 +123,8 @@ jobs:
   # CLIENT (without types test)
   #
   client:
-    timeout-minutes: 35
-    runs-on: ${{ matrix.os }}
+    timeout-minutes: ${{ github.event_name == 'schedule' && 100 || 35 }}
+    runs-on: ${{ github.event_name == 'schedule' && contains(matrix.os, 'buildjet') && 'ubuntu-latest' || matrix.os }}
```
millsp, line 127: One downside with this approach is that the job names will still contain `buildjet-4vcpu-ubuntu-2004` as it came from the matrix, although on scheduled runs it will be run will the free `ubuntu-latest`. I could also replace these in the names, but thought the complexity is enough. Agree?

.github/workflows/test.yml
```diff
@@ -342,8 +342,8 @@ jobs:
   # CLIENT (memory tests)
   #
   client-memory:
-    timeout-minutes: 15
-    runs-on: ${{ matrix.os }}
+    timeout-minutes: ${{ github.event_name == 'schedule' && 60 || 15 }}
```
millsp, line 345: Roughly increased the timeouts on free runners, we can re-adjust later.

.github/workflows/test.yml
millsp: You can see which runner is active at the beginning of the job https://github.com/prisma/prisma/actions/runs/4834439592/jobs/8615666902?pr=19008#step:1:10

millsp: Also, I doubled doubled the timeouts when running on schedule.

.github/workflows/test.yml
```diff
@@ -42,9 +42,9 @@ on:
       - '*.bench.ts'
       - 'scripts/ci/publish.ts'
       - 'graphs/**'
-  #   schedule:
-  #     # https://crontab.guru - “At every 15th minute past every hour from 3 through 5.”
-  #     - cron: '*/15 3-5 * * *'
+  schedule:
+    # https://crontab.guru - “At every 15th minute past every hour from 3 through 5.”
+    - cron: '*/15 3-5 * * *'
```
millsp, line 47:
```suggestion
  schedule:
    # https://crontab.guru - “At every 30th minute past every hour from 3 through 6.”
    - cron: '*/30 3-6 * * *'
```

.github/workflows/test.yml
```diff
@@ -42,9 +42,9 @@ on:
       - '*.bench.ts'
       - 'scripts/ci/publish.ts'
       - 'graphs/**'
-  #   schedule:
-  #     # https://crontab.guru - “At every 15th minute past every hour from 3 through 5.”
-  #     - cron: '*/15 3-5 * * *'
+  schedule:
+    # https://crontab.guru - “At every 30th minute past every hour from 3 through 6.”
+    - cron: '*/30 3-6 * * *'
```
millsp, line 47: Should make it run 6 times

Jolg42 (approved):
Nice!

.github/workflows/test.yml
```diff
@@ -123,8 +123,8 @@ jobs:
   # CLIENT (without types test)
   #
   client:
-    timeout-minutes: 35
-    runs-on: ${{ matrix.os }}
+    timeout-minutes: ${{ github.event_name == 'schedule' && 70 || 35 }}
+    runs-on: ${{ github.event_name == 'schedule' && contains(matrix.os, 'buildjet') && 'ubuntu-latest' || matrix.os }}
```
janpio, line 127: This could deserve a comment now to explain what happens here.
Def will need a few minutes to parse in a few months...

PR #19011, opened 2023-04-28
ci: enable jest sharding on free runners

millsp:
Completely gets rid of BuildJet by sharding the tests into multiple runners. This currently creates a lot of jobs, but by moving the Binary Engine tests to a nightly run we can remove half of the sharded tests, that would make it more viable.

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -116,7 +116,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shard: ['1/4', '2/4', '3/4', '4/4']
+        shard: ['1/4', '2/4', '3/4', '4/4', '5/5']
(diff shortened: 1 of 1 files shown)


Jolg42:
I think it would be best to merge https://github.com/prisma/prisma/pull/19001 before.
Then fix conflicts and merge this.

janpio (approved):
Nice one!

Side question: Do these shards really all need individual machines or could one maybe even run multiple shards in parallel on the same machine? That's how we used to do it in my Ruby days...

millsp:
> Nice one!
> 
> Side question: Do these shards really all need individual machines or could one maybe even run multiple shards in parallel on the same machine? That's how we used to do it in my Ruby days...

Jest will already utilizeq all the cores efficiently, so the machine is at full capacity for each shard. Sharding helps us to make use of more cores by splitting into multiple runners.

millsp:
Follow up work on this will be to reduce the setup time, so that tests can start faster.

PR #19013, opened 2023-04-29
test: remove ts-jest

millsp:
Remove ts-jest on windows to use esbuild make windows tests faster, shaving 6 minutes of execution time.


Jolg42 (approved):
Nice cleanup!

PR #19016, opened 2023-04-29
ci: simplify setup

millsp:
Can save a minute per job on the build (done) and avoids checking typescript types over and over (todo), saving another minute per job. Instead, a single job will check all types, and won't do it for every node version or binary type.

### DIFF
--- a/.github/workflows/scripts/setup.sh
+++ b/.github/workflows/scripts/setup.sh
@@ -1,9 +1,0 @@
-#!/bin/bash
-
-set -ex
-
-npm i --silent -g pnpm@7 --unsafe-perm
-
-pnpm i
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/setup.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
-
-set -ex
-
-npm i --silent -g pnpm@7 --unsafe-perm
```
janpio, line 5: Not necessary because of `pnpm/action-setup` action usage?

Jolg42:
Closing in favor of https://github.com/prisma/prisma/pull/20616

PR #19017, opened 2023-04-29
ci: replace jest --verbose logging by --silent

millsp:
Only display test errors to avoid excessive logging and run tests faster. Saves 3 minutes per job. If verbose logging is useful, it mostly is locally, and CI is really only used to look at failures (which will be properly logged).


Jolg42 (approved):
⏩ Nice!
> `--silent Prevent tests from printing messages through the console.`

We had some noisy logging in some places indeed.

PR #19021, opened 2023-04-30
fix: callback type for the `beforeExit` event

Gamote:
This PR fixes the wrong callback types for the `beforeExit` event type. The new event was added in the [9033ccd](https://github.com/prisma/prisma/commit/9033ccd5bc84249457deb97326740406d9e8e0c0) and [18d76e2](https://github.com/prisma/prisma/commit/18d76e2d2af96ebc6b1e8a469ab44a693ccc6b63) commits but the callback type was not added in the `getPrismaClient.ts`, and in `PrismaClient.ts` was added but it is not the right one.

## Issue
The following `async` function throws this error: `Promise returned in function argument where a void return was expected.(@typescript-eslint/no-misused-promises` because it expects `void` as return type instead of `Promise<void>:
```ts
//                    In here
//                       ↓
this.$on('beforeExit', async () => {
  await app.close();
});
```

If we check the `BinaryEngine` code we can see that if the event type is `beforeExit` the callback will be awaited:
```ts
  on(event: EngineEventType, listener: (args?: any) => any): void {
    if (event === 'beforeExit') {
      this.beforeExitListener = listener // Defined here
    } else {
      this.logEmitter.on(event, listener)
    }
  }

  async emitExit() {
    if (this.beforeExitListener) {
      try {
        await this.beforeExitListener() // Awaited here
      } catch (e) {
        console.error(e)
      }
    }
  }
```

## Solution

Update the `$on` method to accept a callback which returns `Promise<void>` and update the `PrsimaClient` type generator to reflect that.


Gamote:
Does anyone know how can I deal with the CLA prompt?
I have signed it and I made sure the email used in the commit it is present in my GitHub account.

janpio:
That should be all that is need to get the check green.

Gamote:
> That should be all that is need to get the check green.

I have tried several times to do it without success. I have also revoked my consent and did it again with no luck. 🤔

janpio:
Let's get the PR into a potentiall mergable state, and then worry about it (or not, we can always skip it for merge).

Current lint error:
```
/home/runner/work/prisma/prisma/packages/client/src/__tests__/integration/happy/exit-hook/test.ts
  11:37  error  Async arrow function has no 'await' expression  @typescript-eslint/require-await
```

Druue:
Hey @Gamote,
Can you please update this PR with the latest changes on main and update `beforeExit` to be non-breaking, i.e. not a _required_ async.
```ts
return Promise<void> | void
```

Thanks 🙏

Druue:
Hey @Gamote,

Thanks for submitting this PR! This is an important bug to fix, however, the current state of the PR would be a breaking change and, due to the lack of response, we will now be closing this. Please feel free to open up a new PR with the changes outlined [here](https://github.com/prisma/prisma/pull/19021#issuecomment-2009828524)

Again, thank you 🙏

PR #19039, opened 2023-05-02
perf(client): Stop bundling @opentelemetry packages

SevInf:
Removes all @opentelemetry packages from client bundle.
Client would now interact with OTEL throgh type-only `TracingHelper`
dependency. Actual implementation is provided by
`@prisma/instrumentation` package. Technically it works like this:

- `@prisma/internals` package contains type definition for
  `TracingHelper` interface. This interface describes all OTEL functions
  client might need. `runInChildSpan`, `getTraceParent` and `createSpan`
  functions moved there.
- `@prisma/client` contains no-op `TracingHelper` implementation. It
  will be used if tracing is not active (either through disabled preview
  feature or through absence of instrumentation package).
- `@prisma/instrumentation` contains active implementation of
  `TracingHelper`. It is set on a global `PRISMA_INSTRUMENTATION`
  instead of the config. When active, client will use that
  implementation instead of no-op one.
  
For library runtime, that reduces size from 252 to 224 KB.
Performance-wise, unfortunately, not so big impact but `require` still gets around 5ms faster:

image

Close prisma/client-planning#340


packages/client/src/runtime/core/tracing/TracingHelper.ts
```diff
@@ -0,0 +1,59 @@
+import type { Context } from '@opentelemetry/api'
+import { EngineSpanEvent, ExtendedSpanOptions, SpanCallback, TracingHelper } from '@prisma/internals'
+
+export const disabledTracingHelper: TracingHelper = {
+  isEnabled() {
+    return false
+  },
+  getTraceParent() {
+    return `00-10-10-00`
+  },
```
jkomyno, line 10: Where does this "magic string" come from?

SevInf: See comment [here](https://github.com/prisma/prisma/pull/19039/files#diff-bf97d215759ee6d6552a229d2be2ea08ea271db90803c961f29b57438c2ba7bfR20). I should probably duplicate it here as well. Other option would be moving it to `internals`, but that would require making `internals` a proper dependency of `instrumentation` package (rather than `devcDependency`) and that might be too much for one constant

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -0,0 +1,127 @@
...
+  SpanContext,
+  SpanKind,
+  trace,
+  TraceFlags,
+} from '@opentelemetry/api'
+import { Span as SpanConstructor, Tracer } from '@opentelemetry/sdk-trace-base'
+import { EngineSpanEvent, ExtendedSpanOptions, SpanCallback, TracingHelper } from '@prisma/internals'
+
+// If true, will publish internal spans as well
+const showAllTraces = process.env.PRISMA_SHOW_ALL_TRACES === 'true'
```
millsp, line 15: Does it make sense to get this to be dynamic (ie. not store it in a module constant?)

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  getActiveContext(): Context | undefined {
+    return _context.active()
+  }
+
+  runInChildSpan<R>(options: string | ExtendedSpanOptions, callback: SpanCallback<R>): R {
+    if (typeof options === 'string') {
+      options = { name: options }
+    }
+    if (options.internal && !showAllTraces) {
```
millsp, line 88:
```suggestion
    }

    if (options.internal && !showAllTraces) {
```

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    const tracer = trace.getTracer('prisma')
+    const context = options.context ?? this.getActiveContext()
+    const name = `prisma:client:${options.name}`
+    if (options.active === false) {
+      const span = tracer.startSpan(name, options, context)
+      return endSpan(span, callback(span, context))
+    }
+
+    return tracer.startActiveSpan(name, options, (span) => endSpan(span, callback(span, context)))
```
millsp, line 104: Could you please restore the comments from the previous file for this line and line 99?

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -0,0 +1,127 @@
...
+// If true, will publish internal spans as well
+const showAllTraces = process.env.PRISMA_SHOW_ALL_TRACES === 'true'
+
+// https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers
+// If traceparent ends with -00 this trace will not be sampled
+// the query engine needs the `10` for the span and trace id otherwise it does not parse this
+const nonSampledTraceParent = `00-10-10-00`
+
+export class ActiveTracingHelper implements TracingHelper {
+  constructor(private traceMiddleware: boolean) {}
```
millsp, line 23: Could we make this an "options" object? As we might want to add more along middleware in the future.

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  getTraceParent(context?: Context | undefined): string {
+    const span = trace.getSpanContext(context ?? _context.active())
+    if (span) {
+      return `00-${span.traceId}-${span.spanId}-0${span.traceFlags}`
+    }
+    return nonSampledTraceParent
+  }
+
+  async createSpan(engineSpanEvent: EngineSpanEvent): Promise<void> {
```
millsp, line 37:
```suggestion
  async createEngine(engineSpanEvent: EngineSpanEvent): Promise<void> {
```

packages/internals/src/tracing/types.ts
millsp, line 1: I'm curious to know if these types could actually live in the instrumentation package, and if anything prevents that?

SevInf: We'd have to depend on `instrumentation` package in client. Which we already do tbh, but for tests only. Once we start to have this dependency, we'd need to also ensure that the rest of `instrumentation` and its dependencies is tree-shaken away and it stay so when package evolves. `internals` seem like a safer option.

packages/instrumentation/src/ActiveTracingHelper.ts
SevInf: Not sure. If it's not set when the process is started but assigned to `prisma.env` later it will not get propogated to engine - so at best, inlining it would make it half-working: client internal spans will be exported, but engine ones will not.

packages/internals/src/tracing/types.ts
millsp: On the other hand, I think types are relatively safe and will be tree shaken. Or is there anything else imported? In the case where a runtime util needs to be imported, we do have tree-shaking enabled, so importing a function A depending on import A in Prisma Client should only bundle these (and not theoretical function B and import B which are not imported by the Prisma Client runtime, for example).

millsp (approved):

PR #19050, opened 2023-05-03
perf(client): Properly tree-shake `arg` and `lz-string`

SevInf:
`arg` is not used by any of the runtime dependencies but was
not tree-shaked because it is not marked as side effects free and
some `internals` helpers required it.

`lz-string` is not used by any runtime except edge/data-proxy but
was still bundled.

That saves around 8KB from a bundle size, but unfortinately has no
visible effect on performance.

Close prisma/client-planning#342

### DIFF
--- a/helpers/compile/plugins/noSideEffectsPlugin.ts
+++ b/helpers/compile/plugins/noSideEffectsPlugin.ts
@@ -1,0 +1,29 @@
+import type { Plugin } from 'esbuild'
+
+/**
+ * Plugin that forces "sideEffects": false on third party packages
+ * that do not have it specified.
+ * Needed for esbuild to be able to tree-shake those packages if
+ * they are not used.
+ *
+ * @param pattern plugin applies only to the modules matching the pattern
+ * @returns
+ */
+export function noSideEffectsPlugin(pattern: RegExp): Plugin {
+  return {
+    name: 'noSideEffectsPlugin',
+    setup(build) {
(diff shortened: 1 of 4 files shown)


helpers/compile/plugins/noSideEffectsPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+        args.pluginData = { resolved: true }
+        const { path, ...rest } = args
+
+        const result = await build.resolve(path, rest)
+        result.sideEffects = false
+        return result
+      })
+    },
+  }
+}
```
jkomyno, line 29: Isn't a "one-size-fit-all" solution like this one a bit dangerous? What if some third-party packages have side effects without specifying that? What if we used an allowlist of plugins to override, instead?

SevInf: Allowlist is exactly what this plugin is doing. Note the `pattern` parameter and the way it is used in `build.ts` - it will only apply to the modules that match pattern.

millsp (approved):

helpers/compile/plugins/noSideEffectsPlugin.ts
jkomyno: Thank you! I had looked at it too quickly and didn't realise it

jkomyno (approved):

PR #19057, opened 2023-05-03
fix(client): request a free port from the engine

aqrln:
Follow-up to https://github.com/prisma/prisma-engines/pull/3939

Pushing this functionality to the engine side prevents the possibility
of some other server grabbing the free port between the time a free
port was found on the client side and the server is bound on the
engine side. Asking the kernel for a free port and binding the server
to that port needs to happen as a single indivisible operation. See
the referenced PR description for additional details.

An example of a recent failure on CI today due to this issue:
https://github.com/prisma/prisma/actions/runs/4871460934/jobs/8688448247

### DIFF
--- a/packages/client/src/runtime/core/engines/binary/BinaryEngine.ts
+++ b/packages/client/src/runtime/core/engines/binary/BinaryEngine.ts
@@ -9,11 +9,9 @@ import { spawn } from 'child_process'
 import execa from 'execa'
 import fs from 'fs'
 import { blue, bold, dim, green, red, underline, yellow } from 'kleur/colors'
-import net from 'net'
 import pRetry from 'p-retry'
 import path from 'path'
 import type { Readable } from 'stream'
-import { URL } from 'url'
 import { promisify } from 'util'
 
 import { PrismaClientInitializationError } from '../../errors/PrismaClientInitializationError'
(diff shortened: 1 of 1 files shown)


packages/client/src/runtime/core/engines/binary/BinaryEngine.ts
```diff
@@ -194,11 +191,6 @@ export class BinaryEngine extends Engine<undefined> {
     this.previewFeatures = this.previewFeatures.filter((e) => !removedFlags.includes(e))
     this.engineEndpoint = engineEndpoint
 
-    if (engineEndpoint) {
```
SevInf, line 197: Isn't this an internal thing that allowed to connect client to a running engine? Or engine people do not need this anymore?

SevInf (commented):
LGTM with one question.

SevInf (approved):

packages/client/src/runtime/core/engines/binary/BinaryEngine.ts
aqrln: It's still there and should be working, but assigning `this.port` is not really necessary here because the code path where `this.port` was used is completely bypassed when `engineEndpoint` is specified, so this `if` statement essentially was just dead code.

millsp (approved):

PR #19095, opened 2023-05-05
perf(esbuild): don't keep names during minification

SevInf:
Saves 18KB. For error classes, names actually matter, so we do preserve
them manually. Has small impact on performance:

image


packages/client/src/runtime/core/errors/PrismaClientRustError.ts
```diff
@@ -18,10 +20,12 @@ export class PrismaClientRustError extends Error {
   }
 
   get [Symbol.toStringTag]() {
-    return 'PrismaClientRustPanicError'
+    return 'PrismaClientRustError'
```
Jolg42, line 23: Found a bug? 😄

packages/client/src/runtime/query.ts
```diff
@@ -3,6 +3,7 @@ import indent from 'indent-string'
 import { bold, dim, green, red, white } from 'kleur/colors'
 import stripAnsi from 'strip-ansi'
 
+import { setClassName } from '../../../internals/dist/utils/setClassName'
```
Jolg42, line 6:
```suggestion
import { setClassName } from '@prisma/internals/dist/utils/setClassName'
```
Like this does not work?

SevInf: Oops, VS Code auto-insertion

Jolg42: actually `import { setClassName } from '@prisma/internals'` would be the best?

packages/client/src/runtime/core/errors/PrismaClientRustError.ts
janpio: This might actually explain a few confusing issues we had in the past...

SevInf: Yep, copy-paste from a sibling file, most likely.

packages/client/src/runtime/query.ts
SevInf: fixed

Jolg42 (approved):

PR #19132, opened 2023-05-08
ci(GitHub): improve daily tests/CI workflows 

SevInf:
1. Splits most of the workflow in a separate reusable template
file. This allows us to safely use defaults of workflow inputs
regardless of the reasons workflow got triggered and don't think about
fallback every time we need to use the input.
2. "Detect jobs to run" step, as well as the attempt to find "ci test
all" comment now will be done only in CI workflow. Scheduled jobs
will automatically test everything.


SevInf:
ci test all

.github/workflows/test-template.yml
```diff
@@ -0,0 +1,1368 @@
+name: CI Template job
```
Jolg42, line 1: I find the rename confusing as this file is not just a template, it's a workflow file (which has "on" triggers) but is also used as a template in the daily workflow files.

Jolg42: Forget about this, I saw the test workflow file as deleted but it's actually still there :D

SevInf: The only `on` trigger it still has is `workflow_call`, which means it can only be called from other workflows. What would be a better name for it?

.github/workflows/test-template.yml
```diff
@@ -0,0 +1,1368 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, windows-latest]
+        node: [14]
+        queryEngine: ${{ fromJson(needs.determineBuildMatrixParams.outputs.queryEngine) }}
+        # Commented out as we still have limited capacity for macOS runners
+        # When we have more capacity we can shard again
+        # shard: ['1', '2']
```
Jolg42, line 1047: Maybe we can shard these now as we should have capacity?
(can be done separately)

aqrln (approved):

.github/workflows/test-template.yml
```diff
@@ -0,0 +1,1368 @@
...
+        description: Timeout for the jobs. Default is 35 minutes
+        type: number
+        default: 35
+      jobs:
+        description: Jobs to run. See .github/workflows/scripts/detect-jobs-to-run.js script
+        type: string
+        default: '-all-'
+      testAll:
+        description: Whether or not this workflow should test all possible engine/protocol combination
+        type: boolean
```
Jolg42, line 28: Maybe add `required: bool` to make it explicit which ones are optional/required

.github/workflows/test-template.yml
```diff
@@ -0,0 +1,1368 @@
...
+        default: CI
+      ubuntuRunner:
+        description: Override default runner used for linux tests. Can be used to opt-out of buildjet
+        type: string
+        default: buildjet-4vcpu-ubuntu-2004
+      jobTimeout:
+        description: Timeout for the jobs. Default is 35 minutes
+        type: number
+        default: 35
+      jobs:
```
Jolg42, line 22: Nit: potentially confusing for people later that this has the same name has the regular GitHub "jobs" definition.
Maybe renaming to `jobsToRun` helps.

.github/workflows/test-template.yml
Jolg42: (no need to do anything, sorry)

.github/workflows/test-template.yml
SevInf: Yes, i'd rather do this separately

Jolg42 (approved):

Jolg42:
Looks good so far! ✨  https://github.com/prisma/prisma/actions/runs/4916696460 
Screenshot 2023-05-08 at 17 26 13

PR #19139, opened 2023-05-08
feat(client): Adjust some of the tracing spans and add more internal spans

SevInf:
Adjust `connect` span to include the time spent waiting for the library
loading.
Adds following internal spans:
- `instantiateLibrary` - overarching span over resolving openssl
version, looking up and loading library and isntantiatiating engine
isntance.
- `getPlatform` - OS/openssl version detection
- `engineConstructor` - creation of engine instaance after library have
  been loaded.
- `dmmf` - overarching span for all dmmf loading routines
- `getDmmf` - `getDmmf` call, as measured from the client
- `parseDmmf` - parsing dmmf string
- `processDmmf` - constructing `DMMFHelper` class
- `processResults` - unpacking/deserializing engine response after we
got it from the engine.

### DIFF
--- a/packages/client/src/runtime/core/engines/library/LibraryEngine.ts
+++ b/packages/client/src/runtime/core/engines/library/LibraryEngine.ts
@@ -189,21 +189,25 @@ Please help us by answering a few questions: https://pris.ly/bundler-investigati
     return response as Tx.InteractiveTransactionInfo<undefined> | undefined
   }
 
-  private async instantiateLibrary(): Promise<void> {
-    debug('internalSetup')
-    if (this.libraryInstantiationPromise) {
-      return this.libraryInstantiationPromise
-    }
+  private instantiateLibrary(): Promise<void> {
+    return this.config.tracingHelper.runInChildSpan({ name: 'instantiateLibrary', internal: true }, async () => {
+      debug('internalSetup')
+      if (this.libraryInstantiationPromise) {
+        return this.libraryInstantiationPromise
+      }
 
-    assertNodeAPISupported()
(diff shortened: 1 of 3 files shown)


packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -355,50 +364,52 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
...
+        } finally {
+          this.libraryStartingPromise = undefined
         }
-      } finally {
-        this.libraryStartingPromise = undefined
       }
-    }
 
-    this.libraryStartingPromise = this.config.tracingHelper.runInChildSpan('connect', startFn)
+      this.libraryStartingPromise = startFn()
```
janpio, line 409: ? Was that wrong?

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -981,18 +984,21 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
+          )
 
-        return this._tracingHelper.runInChildSpan({ name: 'processDmmf', internal: true }, () => {
-          return new DMMFHelper(getPrismaClientDMMF(dmmf))
-        })
-      } catch (error) {
-        this._fetcher.handleAndLogRequestError({ ...params, args: {}, error })
-      }
+          return this._tracingHelper.runInChildSpan(
+            { name: 'processDmmf', internal: true },
```
janpio, line 995: No problem that the same span name has been used before in line 433? Or are these code paths mutually exclusive?

PR #19152, opened 2023-05-09
test(get-platform): add a benchmark for getPlatform

aqrln:


Jolg42:
Looks good, I guess the limitation here is that it will only be a benchmark running on ubuntu-linux?

Jolg42 (approved):

aqrln:
> I guess the limitation here is that it will only be a benchmark running on ubuntu-linux?

True, but I think that's fine. We only have non-trivial detection logic on Linux.

millsp (approved):

PR #19154, opened 2023-05-09
fix(client): Correctly handle invalid dates in both protocols

### LINKED ISSUE #18970: GraphQL protocol: Invalid `Date` values silently turn into `nulls`
### Bug description

JS `Date` object can be constructed in a way so it will contain invalid date without throwing any exceptions. For example `new Date('Not a date sting')`. When GraphQL protocol is used, such dates get silently converted into `null`. Depending on whether or not corresponding field is nullable, either `null` value would be written in DB (when users intentions most likely have been different) or validation error will be thrown (which is confusing, because as far as user concerned, they are supplying non-null date.)

We should not allow such input and should throw validation error, telling user that the date is invalid instead. 

JSON protocol rejects such dates, but error message could be improved there as well.

### How to reproduce

Run provided code snippet, ensuring that GraphQL protocol is on. `null` value would be written into the database.

### Expected behavior

Runtime error, telling that the provided date is not valid.

### Prisma information

```prisma
model User {
  id   String    @id
  date DateTime? // optional part is really important here
}
```

```ts
await prisma.user.create({
    data: {
      id: '1',
      date: new Date('A am not a date'),
    },
  })
```

### Environment & setup

- OS: any
- Database: any
- Node.js version: any

### Prisma Version

```
4.13.0
```

SevInf:
For GraphQL protocol: stop converting invalid dates to `null`, generate
proper error message.
For JSON: show proper validation message pointing where exactly we
encountered invalid date.

Fix #18970

### DIFF
--- a/packages/client/src/runtime/utils/date.ts
+++ b/packages/client/src/runtime/utils/date.ts
@@ -1,0 +1,7 @@
+export function isDate(value: unknown): value is Date {
+  return Object.prototype.toString.call(value) === '[object Date]'
+}
+
+export function isValidDate(date: Date) {
+  return date.toString() !== 'Invalid Date'
+}
--- a/packages/client/src/runtime/utils/date.ts
+++ b/packages/client/src/runtime/utils/date.ts
@@ -1,5 +1,9 @@
 export function isDate(value: unknown): value is Date {
-  return Object.prototype.toString.call(value) === '[object Date]'
+  return (
+    value instanceof Date ||
+    // date created in other JS context (for example, worker)
+    Object.prototype.toString.call(value) === '[object Date]'
+  )
 }
 
 export function isValidDate(date: Date) {
--- a/packages/client/tests/functional/issues/18970-invalid-date/tests.ts
+++ b/packages/client/tests/functional/issues/18970-invalid-date/tests.ts
@@ -26,7 +26,7 @@ testMatrix.setupTestSuite(
         → XX       prisma.user.findMany({
                      where: {
                        date: new Date('Invalid Date')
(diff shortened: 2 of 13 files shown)


packages/client/src/runtime/utils/date.ts
```diff
@@ -0,0 +1,7 @@
+export function isDate(value: unknown): value is Date {
+  return Object.prototype.toString.call(value) === '[object Date]'
```
aqrln, line 2: There's a few other places in the client that either have this check hardcoded or contain a similar `isDate` function, it would be nice to de-duplicate them and make them use this one.

aqrln: it's fine to do that in another PR though

packages/client/tests/functional/issues/18970-invalid-date/tests.ts
```diff
@@ -0,0 +1,71 @@
...
+        Invalid \`prisma.user.findMany()\` invocation in
+        /client/tests/functional/issues/18970-invalid-date/tests.ts:0:0
+
+          XX () => {
+          XX   testIf(getQueryEngineProtocol() === 'graphql')('throws on invalid date (graphql)', async () => {
+          XX     await expect(
+        → XX       prisma.user.findMany({
+                     where: {
+                       date: new Date('Invalid Date')
+                       ~~~~~~~~~~~~~~~~~~~~~~~~~~
```
aqrln, line 29: Is it expected that the squiggly line doesn't cover the whole line or is shifted from the value towards the key?

aqrln (approved):
💯

Issue #17891, opened 2023-02-11
`Oops. We could not send the error report. Error report submission failed due to:  FetchError: invalid json response body at https://error-reports.prisma.sh/ reason: Unexpected token H in JSON at position 0`


janpio:
Tried to reproduce https://github.com/prisma/prisma/issues/17889, but could not submit error report because I got the above.

Jolg42:
I found these issues mentioning this unexpected error:
https://github.com/prisma/prisma/issues/19130
https://github.com/prisma/prisma/issues/18762
https://github.com/prisma/prisma/issues/17891
https://github.com/prisma/prisma/issues/19141

We should have a look one day

FGoessler:
Hi,

We’ve noticed there hasn’t been activity on this issue for quite some time, and we suspect it might no longer be relevant. As part of our ongoing effort to streamline open issues in this repository, we’re closing this one to focus on what’s most important to our community.

If you believe this issue is still relevant and requires attention, feel free to reopen it. Reopening the issue is a valuable signal to us about what matters most to you.

You can learn more about our approach and motivations in the [Prisma ORM Manifesto](https://www.prisma.io/blog/prisma-orm-manifesto#1-defining-first-class-databases).

With love,
Your Prisma ORM Team ❤️

PR #19170, opened 2023-05-10
fix(internals): better error handling for sendPanic

Jolg42:
Related https://github.com/prisma/prisma/issues/17891

### DIFF
--- a/packages/internals/src/sendPanic.ts
+++ b/packages/internals/src/sendPanic.ts
@@ -1,4 +1,3 @@
-import Debug from '@prisma/debug'
 import { getPlatform } from '@prisma/get-platform'
 import archiver from 'archiver'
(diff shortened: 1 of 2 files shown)


aqrln (approved):

packages/internals/src/sendPanic.ts
```diff
@@ -33,64 +31,68 @@ export async function sendPanic({
...
     if (error.schemaPath) {
       const zip = await makeErrorZip(error)
       await uploadZip(zip, signedUrl)
     }
-
-    const id = await makeErrorReportCompleted(signedUrl)
-    return id
-  } catch (e) {
-    debug(e)
-    throw e
```
Jolg42, line 92: Removed try catch, because it was double logging (debug + throw) and throwing in the end so quite unnecessary 😅

PR #19171, opened 2023-05-10
chore(docker): change restart always to unless-stopped

Jolg42:
It can be very confusing with always to stop some images, close and reopen Docker Desktop and find them on again. Also when rebooting a computer after stopping all images they would start all again (eating all the RAM).

"unless-stopped" is better for me and us in general I think.

Note:
Similar to always, except that when the container is stopped (manually or otherwise), it is not restarted even after Docker daemon restarts.


aqrln (approved):
yes please 💯

PR #19184, opened 2023-05-11
feat(client): esm support draft

### LINKED ISSUE #5030: Support ES6 modules export

### LINKED ISSUE #14484: Finding schema.prisma without __dirname
### Bug description

From slack: https://prisma.slack.com/archives/CCWDULGUW/p1657602998648169

I have an issue with creating lambda function bundles with prisma. I'm using ESM and `__dirname` is not defined, but used in the prisma client to find the `schema.prisma` file as best I can tell.

I can hack around this by creating an ESM shim for dealing with legacy CJS code like the prisma client. This defines __dirname using `import.meta`. See https://github.com/evanw/esbuild/pull/2067#issuecomment-1073039746 for details.

However I can't tell prisma where my bundle path is since it assumes my schema.prisma is in __dirname. I can work around this by hardcoding `__dirname` in my shim to `/var/task` and just always copying the schema.prisma file in there. It doesn't seem like the best solution.

### How to reproduce

1. Create lambda function that runs as ESM (output a .mjs bundle or set type:module in package.json of function bundle)
2. Import @prisma/client

Should get an error 
```
  errorType: 'ReferenceError',
  errorMessage: '__dirname is not defined in ES module scope\n' +
    `This file is being treated as an ES module because it has a '.js' file extension and '/var/task/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.`,
  stack: [
    'ReferenceError: __dirname is not defined in ES module scope',
    `This file is being treated as an ES module because it has a '.js' file extension and '/var/task/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.`,
    '    at node_modules/.prisma/client/index.js (file:///var/task/packages/infra/lib/lambdas/cognito/preTokenGeneration.js:91:53)',
    '    at __require2 (file:///var/task/packages/infra/lib/lambdas/cognito/preTokenGeneration.js:20:50)',
    '    at node_modules/@prisma/client/index.js (file:///var/task/packages/infra/lib/lambdas/cognito/preTokenGeneration.js:1472:10)',
    '    at __require2 (file:///var/task/packages/infra/lib/lambdas/cognito/preTokenGeneration.js:20:50)',
    '    at file:///var/task/packages/infra/lib/lambdas/cognito/preTokenGeneration.js:75914:29',
    '    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)',
    '    at async Promise.all (index 0)',
    '    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)',
    '    at async _tryAwaitImport (file:///var/runtime/index.mjs:660:16)',
    '    at async _tryRequire (file:///var/runtime/index.mjs:709:37)'
```

### Expected behavior

What would be nice (from best to least best):
* Not to have to include the schema.prisma file in my deployment
* For prisma to generate runtime code that compatible with ESM (i.e. not use __dirname)
* To have an env var to tell prisma where to find the schema.prisma file

### Prisma information

4.0.0

### Environment & setup

- OS: 
- Database: 
- Node.js version: 
Node 16 lambda

### Prisma Version

```
prisma                  : 4.0.0
@prisma/client          : 4.0.0
Query Engine (Node-API) : libquery-engine da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/migration-engine-darwin)
Introspection Engine    : introspection-core da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/introspection-engine-darwin)
Format Binary           : prisma-fmt da41d2bb3406da22087b849f0e911199ba4fbf11 (at node_modules/@prisma/engines/prisma-fmt-darwin)
Default Engines Hash    : da41d2bb3406da22087b849f0e911199ba4fbf11
Studio                  : 0.465.0
```

### LINKED ISSUE #15614: `__dirname is not defined` when generating PrismaClient to custom location, w/ SvelteKit
### Bug description

Prisma seems to have issues running with SvelteKit when generating the client to some other location other than the default.

When running a production build for the app, the following error will show:
```
ReferenceError: __dirname is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/<sveltekit-project-dir>/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file:///<sveltekit-project-dir>/build/server/chunks/2-6f7b7120.js:19960:27
    at file:///<sveltekit-project-dir>/build/server/chunks/2-6f7b7120.js:31969:3
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async Promise.all (index 1)
    at async render_page (file:///<sveltekit-project-dir>/build/server/index.js:2382:19)
    at async resolve (file:///<sveltekit-project-dir>/build/server/index.js:2872:22)
    at async respond (file:///<sveltekit-project-dir>/build/server/index.js:2910:22)
    at async Array.ssr (file:///<sveltekit-project-dir>/build/handler.js:19060:3)
```

### Notes: 
* This issue does not happen when running `vite dev` or `vite preview` for some reason. Prisma works as expected under those conditions. The error only happens when running the build output directly with node.
* If you remove the `output` option from the generator, then import the prisma client from '@prisma/client', Prisma will run as expected - but unfortunately that doesn't meet my needs.
* **Reason for this use case**: I need to connect to multiple schemas, and I want the benefit of prisma's generated types for both. As far as I know, the only way to connect to multiple schemas is to generate multiple prisma clients - which in turn means I can't use '@prisma/client' because that package will only point to one client at a time. (When generating multiple clients to the default location, they will just overwrite each other)

### How to reproduce

See my test repo where I've reproduced the error: [https://github.com/McGaelen/prisma-sveltekit](https://github.com/McGaelen/prisma-sveltekit)

### Expected behavior

No error and Prisma runs as expected.

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
  output = "../node_modules/custom-client-location"
}
...(the rest is an exact copy of https://www.prisma.io/docs/getting-started/quickstart)
```

```ts
import {PrismaClient} from 'custom-client-location' // <- importing from custom location specified in prisma.schema
```

### Environment & setup

- OS: macOS, CentOS
- Database: PostgreSQL, SQLite, but probably all of them
- Node.js version: v16.15.1

### Prisma Version

```
prisma                  : 4.4.0
@prisma/client          : 4.4.0
Query Engine (Node-API) : libquery-engine f352a33b70356f46311da8b00d83386dd9f145d6 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli f352a33b70356f46311da8b00d83386dd9f145d6 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core f352a33b70356f46311da8b00d83386dd9f145d6 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt f352a33b70356f46311da8b00d83386dd9f145d6 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Format Wasm             : @prisma/prisma-fmt-wasm 4.4.0-66.f352a33b70356f46311da8b00d83386dd9f145d6
Default Engines Hash    : f352a33b70356f46311da8b00d83386dd9f145d6
Studio                  : 0.474.0

```

### LINKED ISSUE #46070: Next.js in a monorepo does not include file referenced in code
### Verify canary release

- [X] I verified that the issue exists in the latest Next.js canary release

### Provide environment information

Operating System:
      Platform: linux
      Arch: x64
      Version: #66~20.04.1-Ubuntu SMP Wed Jan 25 09:41:30 UTC 2023
    Binaries:
      Node: 19.0.1
      npm: 8.19.2
      Yarn: 1.22.17
      pnpm: 7.24.2
    Relevant packages:
      next: 13.1.7-canary.17
      eslint-config-next: N/A
      react: 18.2.0
      react-dom: 18.2.0

### Which area(s) of Next.js are affected? (leave empty if unsure)

_No response_

### Link to the code that reproduces this issue

https://github.com/millsp/nextjs-monorepo-file-not-found

### To Reproduce

- Clone the provided reproduction repository
- Follow steps in `README` of each branch
- There are two branches:
    1. [A JS-based code branch](https://github.com/millsp/nextjs-monorepo-file-not-found/tree/js)
    Which shows the described bug and how the workaround fixes it
    2. [A TS-based code branch](https://github.com/millsp/nextjs-monorepo-file-not-found/tree/ts)
    Which shows the described bug and how the workaround is broken

### Describe the Bug

An unexpected file not found error happens from an imported package that reads from the local file system (using Next.js in a monorepo).

In this monorepo, I have a package called `service` and another one called `file-reader`. Here’s the representation of the project:

```
.
├── packages
│   ├── file-reader
│   │   ├── index.js
│   │   ├── node_modules
│   │   ├── package.json
│   │   └── file.txt
│   └── service
│       ├── next.config.js
│       ├── next-env.d.ts
│       ├── node_modules
│       ├── package.json
│       └── pages
│           └── api
│               └── test.js
├── pnpm-workspace.yaml
└── README.md
```

`service` needs to use `file-reader`, and `file-reader` needs to read `file.txt`, but running the code with `next build` or `next dev`, and opening the API endpoint, we will then encounter an `ENOENT: no such file or directory` error:

```
Server Error
Error: ENOENT: no such file or directory, open '/nextjs-monorepo-notfound/packages/service/.next/server/pages/api/file.txt'
```

### Analysis why this is happening

Since `v13.1.7-canary.11` (which included https://github.com/vercel/next.js/pull/45864), the result of `pnpm exec next build` includes the `file.txt` in the correct location:

```
.next/standalone
├── node_modules
├── package.json
└── packages
    ├── file-reader
    │   ├── file.txt
    │   ├── index.js
    │   └── package.json
    └── service
        ├── .env
		├── .next
		│   └── server
		│       ├── pages
		│       │   ├── api 
		│       │   │   ├── test.js // this bundle includes `file-reader`
		│       │   │   └── test.js.nft.json
        ├── package.json
        └── server.js
```

(Before `v13.1.7-canary.11` that was not the case and you had to use `outputFileTracingRoot: "../../"` manually to make it work - thanks @ijjk for that timely fix!).

So why is the file still not found when executing the code, although it is included directly?

Our current understanding is that because there is a bundling step, `file-reader` will get bundled as the API endpoint (or as a `chunk`). However, the files this bundle is looking for (here `file.txt`) are not present right next to it. Remember, `file-reader` wants to open a file called `file.txt` and expects it to be near its `index.js`. 

So for this to work, we would need to have a tree like this instead:

```diff
.next/standalone
├── node_modules
├── package.json
└── packages
    ├── file-reader
    │   ├── file.txt
    │   ├── index.js
    │   └── package.json
    └── service
        ├── .env
		├── .next
		│   └── server
		│       ├── pages
		│       │   ├── api
+		│       │   │   ├── file.txt
		│       │   │   ├── test.js // this bundle includes `file-reader`
		│       │   │   └── test.js.nft.json
        ├── package.json
        └── server.js
```

### Our attempt at finding a workaround

We tried marking `file-reader` as external:

```js
webpack: (config, { buildId, dev, isServer, defaultLoaders, webpack }) => {
   config.externals = [...config.externals, 'file-reader']
   return config
},
```

This works perfectly when your codebase is JS-only, but breaks completely as soon as you use TypeScript. Because marking as external skips the Next.js transpilation step, then we end up with new errors instead:

```
Server Error
SyntaxError: Cannot use import statement outside a module

This error happened while generating the page. Any console logs will be displayed in the terminal window.

/nextjs-monorepo-notfound/packages/file-reader/index.ts:1
import path from 'path'
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at Object.compileFunction (node:vm:360:18)
```

(We have also tried `transpilePackages` in this context, but that didn’t have any effect)

This shows how bundling is causing these path issues.

### Expected Behavior

Next.js should include the files so they are found at runtime, and the user does not get an error. 

In other words, Next.js should be able to understand how to copy the correct files near the generated bundles and include these in their respective `nft.json` files for a valid `standalone` build.

### Which browser are you using? (if relevant)

_No response_

### How are you deploying your application? (if relevant)

_No response_

millsp:
closes https://github.com/prisma/prisma/issues/5030
closes https://github.com/vercel/next.js/issues/46070
closes https://github.com/prisma/prisma/issues/14484
closes https://github.com/prisma/prisma/issues/15614


packages/client/package.json
```diff
@@ -19,9 +19,41 @@
...
-  "main": "index.js",
-  "browser": "index-browser.js",
   "types": "index.d.ts",
+  "main": "index.js",
+  "module": "index.mjs",
+  "browser": "index-browser.mjs",
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./index.d.ts",
```
Andarist, line 29: It's not 100% correct to use `types` condition like this. And in this specific case, it's even more apparent. Notice how your `index.d.ts` **doesn't** match `index.mjs`:
```ts
// index.d.ts
export * from '.prisma/client'

// index.mjs
export * from '../../.prisma/client/index.mjs'
export * as default from '../../.prisma/client/index.mjs'
```
It's clearly visible that this declaration file doesn't describe closely this runtime file and yet you are saying (through the `types` condition applied like this) that it does.

millsp: Ok, then this is just half of the solution? How can we reconciliate both requirements: that an import from `@prisma/client` points to the right types (this is what this is doing, taken from ts [docs](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing)), but also allow types to be matched on fully specified imports. I think this needs to be handled in the `*` pattern below. Though I haven't tried yet.

Andarist: > How can we reconciliate both requirements: that an import from @prisma/client points to the right types (this is what this is doing, taken from ts [docs](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing)),

That is not exactly what this PR is doing - although at some point the TS docs were showing what you have in this PR, that was a mistake and the docs were changed. Notice that now the docs show an example with smth like this:
```json
{
    "exports": {
        ".": {
            // Entry-point for `import "my-package"` in ESM
            "import": {
                // Where TypeScript will look.
                "types": "./types/esm/index.d.ts",
                // Where Node.js will look.
                "default": "./esm/index.js"
            },
            // Entry-point for `require("my-package")` in CJS
            "require": {
                // Where TypeScript will look.
                "types": "./types/commonjs/index.d.cts",
                // Where Node.js will look.
                "default": "./commonjs/index.cjs"
            },
        }
    },
}
```

We can see here that the types are kinda duplicated between import and require conditions - and that matches exactly what you are doing at runtime. You are providing 2 totally different files for the runtime so it kinda only makes sense to do the same for types (I recognize that this is quite annoying but that's what is *correct*).

> but also allow types to be matched on fully specified imports (what you are showing). I think this needs to be handled in the * pattern below, and this is not so much of an issue actually? Though I haven't tried yet.

It would be easier for me to discuss this using some examples and stuff.

millsp: Do the files actually differ in content? These entrypoint files/types redirect to our actual generated types. We want to keep the experience the same between ESM/CJS, so we will have the same default exports, and named exports, APIs, etc...

medz:
@millsp Thank you for your great contribution to this. Esm is something that many NodeJS backend engineers are migrating. I found that more and more developers packaged as esm by default, and many third-party packages only provide esm. It's not a problem to allow cjs to be imported in tsconfig, but newer bundlers always don't support it. Like swc, I have to keep track of whether my imports are `import` or `import type` which is very annoying. Looking forward to this pr can enter the official version soon.

My blessings online for your hard work~

toteto:
@millsp is this PR and version from this going to be replaced by #21858? Can we expect new ESM release anytime soon?

janpio:
See https://github.com/prisma/prisma/pull/21858

PR #19193, opened 2023-05-11
chore(internals)!: remove legacy photonResolver and "provider=photonjs" handling

### LINKED ISSUE #19319: Prisma Client: remove legacy `photonResolver` and `provider=photonjs` handling
Prisma Client used to be called "Photon" and we want to remove the error associated with that rename since it has been a few years now.

PR: https://github.com/prisma/prisma/pull/19193

aqrln:
Closes: https://github.com/prisma/prisma/issues/19319


packages/internals/src/get-generators/generatorResolvers/generatorResolvers.ts
```diff
@@ -19,6 +18,5 @@ export type GeneratorResolvers = {
  * as written in the schema, as well as finding their output paths.
  */
 export const generatorResolvers: GeneratorResolvers = {
```
SevInf, line 20: I guess we could also remove this map completely now and just use `prismaClientResolver` directly?

aqrln: Yes, good point.

janpio:
ping @Jolg42 about the expanded scope of also removing the `photonjs` handling

Jolg42:
Thanks @janpio, no issue with that for me, only that we need to document it.

Jolg42 (approved):

aqrln:
There was an unresolved review comment about removing the map and using `prismaClientResolver` directly which I completely forgot about. Should I open a follow-up PR?

Jolg42:
Sorry about that, I don't think it's super important / has much value but of course can be done anytime now if you feel like it.

aqrln:
No worries at all, just asking to know if this is something we still want to do.

PR #19205, opened 2023-05-11
test(dbpull): split tests in multiple files

Jolg42:
Next PR will be about https://github.com/prisma/schema-team/issues/407

Note: things might get renamed & rearranged later, this is a clean split.

It should also limit the blast radius of https://github.com/prisma/prisma/issues/8031


packages/get-platform/src/test-utils/jestSnapshotSerializer.js
```diff
@@ -1,4 +1,5 @@
 const path = require('path')
+const dedent = require('strip-indent')
```
Jolg42, line 2:
```suggestion
```

packages/get-platform/src/test-utils/jestSnapshotSerializer.js
```diff
@@ -154,6 +155,7 @@ module.exports = {
       normalizeMigrateTimestamps,
       // artificial panic
       normalizeArtificialPanic,
+      dedent,
```
Jolg42, line 158:
```suggestion
```

packages/migrate/src/__tests__/DbPull/cockroachdb.test.ts
```diff
@@ -0,0 +1,263 @@
...
+
+// To avoid the loading spinner locally
+process.env.CI = 'true'
+
+const originalEnv = { ...process.env }
+
+describeIf(!process.env.TEST_SKIP_COCKROACHDB)('cockroachdb', () => {
+  if (!process.env.TEST_SKIP_COCKROACHDB && !process.env.TEST_COCKROACH_URI_MIGRATE) {
+    throw new Error('You must set a value for process.env.TEST_COCKROACH_URI_MIGRATE. See TESTING.md')
+  }
```
Druue, line 28: I think this was how it was before too but it still reads weirdly that we already discriminate on `!process.env.TEST_SKIP_COCKROACHDB` in the `describeIf()` but then do it again in the `if`

Jolg42: Oh yes, this is weird but nothing new here. That code does run somehow if the env vars are not set 🙈 (which in theory should not happen)

Jolg42:
I'll merge as this is a "clean" split passing CI

PR #19223, opened 2023-05-12
ci: shard no-docker-client-functional

Jolg42:
Let's see how much time we can save waiting 👀

Now: https://github.com/prisma/prisma/actions/runs/4961140552/usage
Screenshot 2023-05-12 at 19 40 40

Before: https://github.com/prisma/prisma/actions/runs/4961106118/usage
Screenshot 2023-05-12 at 19 41 34

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -1059,9 +1059,7 @@ jobs:
         os: [macos-latest, windows-latest]
         node: [14]
         queryEngine: ${{ fromJson(needs.determineBuildMatrixParams.outputs.queryEngine) }}
-        # Commented out as we still have limited capacity for macOS runners
-        # When we have more capacity we can shard again
-        # shard: ['1', '2']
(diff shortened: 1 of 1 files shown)


millsp:
It does not look like there is a runner limit specifically with Windows after looking at this page. Do you know if there is one?
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

millsp (approved):
Nice! Time is almost cut by half :tada:

Jolg42:
No idea about windows, maybe that's the perk about running on Azure 😄

Jolg42:
Now these and Buildkite are the slow ones
Screenshot 2023-05-12 at 19 42 45

janpio:
The limit on MacOS is because of Apple's weird licencing and the more expensive hardware required - that does not exist for Windows and Linux so I am not aware of specific limits.

PR #19225, opened 2023-05-12
fix(client): don't run engines before copying

### LINKED ISSUE #4308: prisma generate is blocked by query-engine-rhel-openssl-1.0.x opening in Notepad on Windows
## Bug description

When I run `prisma generate` two instances of Notepad are launched, which block `prisma generate` from completing until I close them

image

Until I close those two windows, it will just stand there waiting.

Just as a test I did this, and it did get rid of the problem, but also doesn't generate the binary targets I need...
image

## How to reproduce

This is the command I run

```terminal
$ node_modules/.bin/prisma generate --schema=api/db/schema.prisma
```

This is what my schema looks like

```prisma
datasource DS {
  // optionally set multiple providers
  // example: provider = ["sqlite", "postgresql"]
  provider = "sqlite"
}

generator client {
  binaryTargets = ["native", "rhel-openssl-1.0.x"]
}

model Post {
}
```

## Expected behavior

I expect `prisma generate` to run to completion without launching Notepad

## Prisma information

See above

## Environment & setup

- OS: Windows 10 10.0.19041
- Database: SQLite
- Node.js version: 14.4.0
- Prisma version: 2.9.0

```
$ node_modules/.bin/prisma -v
@prisma/cli          : 2.9.0
@prisma/client       : 2.9.0
Query Engine         : query-engine 369b3694b7edb869fad14827a33ad3f3f49bbc20 (at node_modules\@prisma\cli\query-engine-windows.exe)
Migration Engine     : migration-engine-cli 369b3694b7edb869fad14827a33ad3f3f49bbc20 (at node_modules\@prisma\cli\migration-engine-windows.exe)
Introspection Engine : introspection-core 369b3694b7edb869fad14827a33ad3f3f49bbc20 (at node_modules\@prisma\cli\introspection-engine-windows.exe)
Format Binary        : prisma-fmt 369b3694b7edb869fad14827a33ad3f3f49bbc20 (at node_modules\@prisma\cli\prisma-fmt-windows.exe)
Studio               : 0.296.0
```

aqrln:
Don't try to run non-native engines querying them for version when
determining whether they should be overwritten during generation.

~~This PR takes a rather conservative approach to fix the issue by
adding a check to compare the binary target of each engine with the
current platform. This only fixes the bug without changing the
performance characteristics of the code in any way (`getPlatform` is
memoized so calling it is essentially free at this point).~~

However, we should benchmark and re-evaluate if these checks provide
any benefit compared to just always copying and overwriting the
engines upon generation. When using the library engine, the version
check also leads to increased memory usage, as unlike with the binary
engine, all of the successfully loaded libraries (usually two when
regenerating existing client, but potentially even more on Linux since
there may be multiple binary targets compatible with the current
system) will stay loaded in memory until the generator process
exits. This also runs the napi-rs initialization code for each of
them, and we don't use Query Engine for anything else anymore in the
generator, so this is the only time QE is currently loaded during
generation.

**UPD:** all the checks were removed as per discussion, now we just copy the files unconditionally. The difference is in a ballpark of around 4 ms per engine on my machine.

Fixes: https://github.com/prisma/prisma/issues/4308

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -5,6 +5,7 @@ import {
   ClientEngineType,
   getClientEngineType,
   getEngineVersion,
+  getPlatform,
   Platform,
   setClassName,
 } from '@prisma/internals'
@@ -304,12 +305,22 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
       await ensureDir('/tmp/prisma-engines')
     }
 
+    const currentPlatform = await getPlatform()
+
     for (const [binaryTarget, filePath] of Object.entries(enginePath)) {
       const fileName = path.basename(filePath)
       const target =
         process.env.NETLIFY && binaryTarget !== 'rhel-openssl-1.0.x'
           ? path.join('/tmp/prisma-engines', fileName)
           : path.join(finalOutputDir, fileName)
+
+      // If the engine is not native, copy it unconditionally regardless of whether it already exists since we can't
+      // check its version.
+      if (binaryTarget !== currentPlatform) {
+        await overwriteFile(filePath, target)
+        continue
+      }
+
       const [sourceFileSize, targetFileSize] = await Promise.all([fileSize(filePath), fileSize(target)])
 
       // If the target doesn't exist yet, copy it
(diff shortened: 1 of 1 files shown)


millsp:
I'd vote for removing the version check always and just copy as you suggested, and that would also avoid this inconsistency. Do you know if the generation benchmarks that we have are good to check this (the stable ones).

millsp (approved):

PR #19259, opened 2023-05-15
fix: Fix error messages stating "Please install it with [object Promise]"

oohwooh:
`getPackageCmd` was not being `await`ed everywhere it needed to be. This results in messages displaying the following:
image


millsp:
Thanks for the PR, very much appreciated :rocket:. I think we should definitely add a test for this. It might be a bit tricky, happy to do it if you want.

oohwooh:
I would appreciate you implementing the test, if that's alright.

Thank you!

Jolg42 (approved):

aqrln (approved):

Jolg42:
I merged because it's a good PR to take in, also @aqrln found out the same issue in a separate PR.

Optimally, we should have tests, but it's not a blocker for this PR, so I created a tracking issue about that https://github.com/prisma/prisma/issues/19814

aqrln:
I was writing a comment about tests too (since they were requested in previous reviews) and was going to merge afterwards but @Jolg42 beat me to it 😄

So, there are a few reasons this PR would be very tricky to test. One problem is this logic requires an existing Prisma version to install: so either we'd need to implement falling back to last stable version when detecting that this is running in tests, or this would need to be tested in the ecosystem-tests and not here. In the former case, another problem would be that we lack an e2e test setup similar to https://github.com/prisma/prisma/tree/main/packages/client/tests/e2e but for Windows (for the first error message).

The second error message is even trickier to test because if it was triggered at all, then it's probably a bug by itself.

Templer44:
Hello Everyone, my team and I will work on creating a test for this!

PR #19270, opened 2023-05-16
temp: Only search for engine in `datamodelPath`

janpio:
Test what happens when we only search in `this.config.datamodelPath`

### DIFF
--- a/packages/client/src/runtime/core/engines/library/DefaultLibraryLoader.ts
+++ b/packages/client/src/runtime/core/engines/library/DefaultLibraryLoader.ts
@@ -189,18 +189,8 @@ Read more about deploying Prisma Client: https://pris.ly/d/client-generator`
 
     const dirname = eval('__dirname') as string
     const searchLocations: string[] = [
-      // TODO: why hardcoded path? why not look for .prisma/client upwards?
-      path.resolve(dirname, '../../../.prisma/client'), // Dot Prisma Path
-      this.config.generator?.output?.value ?? dirname, // Custom Generator Path
-      path.resolve(dirname, '..'), // parentDirName
-      path.dirname(this.config.datamodelPath), // Datamodel Dir
-      this.config.cwd, //cwdPath
-      '/tmp/prisma-engines',
+      path.dirname(this.config.datamodelPath),
     ]
-
-    if (this.config.dirname) {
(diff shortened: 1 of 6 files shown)


packages/client/src/runtime/core/engines/library/DefaultLibraryLoader.ts
```diff
@@ -187,21 +187,10 @@ Read more about deploying Prisma Client: https://pris.ly/d/client-generator`
...
-    const dirname = eval('__dirname') as string
     const searchLocations: string[] = [
-      // TODO: why hardcoded path? why not look for .prisma/client upwards?
-      path.resolve(dirname, '../../../.prisma/client'), // Dot Prisma Path
-      this.config.generator?.output?.value ?? dirname, // Custom Generator Path
-      path.resolve(dirname, '..'), // parentDirName
-      path.dirname(this.config.datamodelPath), // Datamodel Dir
-      this.config.cwd, //cwdPath
-      '/tmp/prisma-engines',
+      path.dirname(this.config.datamodelPath),
```
millsp, line 191:
```suggestion
      this.config.dirname,
```
Code says it can be undefined, but is always present.

millsp: This with what we got the data model path in the first place

janpio: That would be the (removed) lines 201-203...
Why could we currently be using both?

janpio:
[ci test all]

packages/client/src/runtime/core/engines/library/DefaultLibraryLoader.ts
millsp: Maybe the logic was different a long time ago, I cannot tell for sure.

millsp: Comes from here https://github.com/prisma/prisma/blob/ea22c620c075a244734ff38c8ad66cbbe8538800/packages/client/src/runtime/getPrismaClient.ts/#L439

Jolg42:
Closing to clean up the open PRs list.

PR #19383, opened 2023-05-22
fix: enum field ref should work

### LINKED ISSUE #19449: `fieldReference` is not working with enums
Hum looks like `fieldReference` is not working with enums

```ts
enum Locale {
  en
  ja
  es
  fr
 ...
}

  // ...

  localeDefault Locale
  locale        Locale
```

if I try to match those
```ts
.findMany({
    where: {
      locale: { equals: db.table.fields.localeDefault  },
    }
})

/*
Argument equals: Provided value prisma.table.fields.localeDefault of type LocaleFieldRefInput<Table> on prisma.findManyTable is not a Locale.
*/
```

Looks like a field reference is not a valid enum value, or it doesn't pass some form of validation.
If I do exactly the same query with 2 strings, it works fine (I created a view which cast my fields from enum to string)

_Originally posted by @kefniark in https://github.com/prisma/prisma/issues/15068#issuecomment-1488913661_

Weakky:
## Overview

fixes https://github.com/prisma/prisma/issues/19449


SevInf (approved):

Jolg42:
Note: I created an issue, so it will be listed in the release changelog

PR #19385, opened 2023-05-22
fix(generator-helper): continuously handle generator errors

### LINKED ISSUE #3294: Prisma generate - `Error: write EPIPE` on WSL <-> Windows
## Bug description

In the internal introspection CI: https://github.com/prisma/introspection-ci/issues/309#issuecomment-674075918

We ran into `Error: write EPIPE` with MacOS. Here is the internal commit: https://github.com/prisma/introspection-ci/pull/316/commits/392528c430679bfd42ac8d09e98e8d9403c1502c#diff-99e681890513b0f406b692a64ece0ddb

This has been seen before but was closed due to no reproduction: https://github.com/prisma/prisma/issues/1816

Environment: 

- Multiple `prisma generate` commands were running in parallel
- One of the commands has `Error: write EPIPE`
- Some generate commands failed with no output (may or may not be related)

## How to reproduce

- Probably try to run the same command on MacOS as the CI did. Introspection CI, branch `fix_309`

```
yarn -s generator-ci:prepare-generate
yarn -s generator-ci:run-generate
```

## Prisma information

```
@prisma/cli          : 2.5.0-dev.52
Current platform     : darwin
Query Engine         : query-engine 8f08c6fa89afd00710e41d99621786150b7eebbe (at node_modules/@prisma/cli/query-engine-darwin)
Migration Engine     : migration-engine-cli 8f08c6fa89afd00710e41d99621786150b7eebbe (at node_modules/@prisma/cli/migration-engine-darwin)
Introspection Engine : introspection-core 8f08c6fa89afd00710e41d99621786150b7eebbe (at node_modules/@prisma/cli/introspection-engine-darwin)
Format Binary        : prisma-fmt 8f08c6fa89afd00710e41d99621786150b7eebbe (at node_modules/@prisma/cli/prisma-fmt-darwin)
Studio               : 0.254.0
```

### LINKED ISSUE #8314: Crashing community generators make `generate` process hang
### Bug description

This `package.json`:

```
{
  "name": "repro8264",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "prisma": "2.27.0",
    "prisma-erd-generator": "0.3.2"
  },
  "dependencies": {
    "@prisma/client": "2.27.0"
  }
}
```

and this schema file:

```
generator client {
    previewFeatures = ["referentialActions"]
}

generator erd {
  provider = "prisma-erd-generator"
}

datasource db {
    provider = "postgresql"
    url      = "postgresql://brianne:Td...Lrx7@pg-db-provision.cm0mkpwj8arx.eu-central-1.rds.amazonaws.com:5432/ap...ot_dragon"
}

model Data {
    tenantId   String    @db.VarChar
    id         String    @db.VarChar
    locationId String?   @db.VarChar
    location   Location? @relation(fields: [tenantId, locationId], references: [tenantId, id], onDelete: SetNull)

    @@id([tenantId, id])
}

model Location {
    tenantId String                     @db.VarChar
    id       String                     @db.VarChar
    data     Data?
    tenant   Tenant                     @relation(fields: [tenantId], references: [id], onDelete: Cascade)

    @@id([tenantId, id])
}


model Tenant {
    id            String          @id @db.VarChar
    locations     Location[]
}
```

make the `generate` process hang because of an implementation error in the `prisma-erd-generator`.

### Expected behavior

Instead of hanging, we should show a helpful error.

### Environment & setup

- OS: 
- Database: 
- Node.js version: 

### Prisma Version

```
2.27.0
```

### LINKED ISSUE #10291: `npx prisma generate` hangs after upgrading `@prisma/client` from 3.0.2
### Bug description

Running `npx prisma generate` from `postinstall` hangs on 3.4.2. Used to work fine with 3.0.2:

```sh
> npx prisma generate

^C
``` 

I had to CTRL+C out of it.

I upgraded `@prisma/client` from 3.0.2 to latest (3.4.2) and after that upgrade, the generate command just hangs.

Using:

```json
{
  "dependencies": {
    "@prisma/client": "^3.4.2",
    "graphql": "^15.5.3",
    "nexus": "^1.1.0",
    "nexus-prisma": "^0.34.0"
  }
}
```
(stripped dependencies I didn't think were relevant)

### How to reproduce

* Upgrade prisma from 3.0.2 to 3.4.2
* Run `npx prisma generate`
* Observe: it just hangs

### Expected behavior

Expect it to complete successfully like it used to do before

In the changeset I made for my project are only 2 files:
* package.json
* package-lock.json

The diff for package.json shows only the @prisma/client version number as changed.

### Prisma information

generator client {
  provider = "prisma-client-js"
}

generator nexusPrisma {
  provider = "nexus-prisma"
}

datasource db {
  provider = "mysql"
}

### Environment & setup

- OS: MacOS
- Database: MySQL
- Node.js version: v16.13.0

### Prisma Version

```
prisma                  : 3.0.2
@prisma/client          : 3.4.2
Query Engine (Node-API) : libquery-engine 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt 2452cc6313d52b8b9a96999ac0e974d0aedf88db (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : 2452cc6313d52b8b9a96999ac0e974d0aedf88db
```

aqrln:
* Remove `initWaitTime`, handle errors properly all the time
* Don't hang when a generator fails
* Fix `write EPIPE` error when the generator stopped but the CLI still tries to send messages
* Improve type safety

Fixes https://github.com/prisma/prisma/issues/3294
Fixes https://github.com/prisma/prisma/issues/8314
Fixes https://github.com/prisma/prisma/issues/10291
Fixes https://github.com/prisma/prisma/issues/14002

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -54,9 +54,11 @@ process.removeAllListeners('warning')
 const debug = Debug('prisma:cli')
 process.on('uncaughtException', (e) => {
   debug(e)
+  throw e
 })
 process.on('unhandledRejection', (e) => {
   debug(e)
+  throw e
 })
 // Listen to Ctr + C and exit
 process.once('SIGINT', () => {
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -51,15 +51,6 @@ const commandArray = process.argv.slice(2)
 
 process.removeAllListeners('warning')
 
-const debug = Debug('prisma:cli')
-process.on('uncaughtException', (e) => {
-  debug(e)
-  throw e
-})
-process.on('unhandledRejection', (e) => {
-  debug(e)
-  throw e
-})
 // Listen to Ctr + C and exit
 process.once('SIGINT', () => {
   process.exit(130)
--- a/packages/generator-helper/src/GeneratorProcess.ts
+++ b/packages/generator-helper/src/GeneratorProcess.ts
@@ -37,6 +37,7 @@ export class GeneratorProcess {
   private initPromise?: Promise<void>
   private isNode: boolean
   private errorLogs = ''
+  private pendingError: Error | undefined
 
   constructor(private pathOrCommand: string, { isNode = false }: GeneratorProcessOptions = {}) {
     this.isNode = isNode
@@ -80,10 +81,12 @@ export class GeneratorProcess {
           for (const listener of Object.values(this.listeners)) {
             listener(null, error)
           }
+          this.pendingError = error
         }
       })
 
       this.child.on('error', (err) => {
+        this.pendingError = err
         if (err.message.includes('EACCES')) {
           debug(err)
           reject(
@@ -157,48 +160,37 @@ export class GeneratorProcess {
     }
   }
 
-  getManifest(config: GeneratorConfig): Promise<GeneratorManifest | null> {
-    return new Promise((resolve, reject) => {
-      const messageId = this.getMessageId()
-
-      this.registerListener(messageId, (result, error) => {
-        if (error) {
-          return reject(error)
-        }
-        if (result.manifest) {
-          resolve(result.manifest)
-        } else {
-          resolve(null)
-        }
-      })
-
-      this.sendMessage({
-        jsonrpc: '2.0',
-        method: 'getManifest',
-        params: config,
-        id: messageId,
-      })
-    })
-  }
-
-  generate(options: GeneratorOptions): Promise<any> {
-    return new Promise((resolve, reject) => {
-      const messageId = this.getMessageId()
-
-      this.registerListener(messageId, (result, error) => {
-        if (error) {
-          reject(error)
+  private rpcMethod<T, U>(method: string, mapResult: (x: unknown) => U = (x) => x as U): (arg: T) => Promise<U> {
+    return (params: T): Promise<U> =>
+      new Promise((resolve, reject) => {
+        if (this.pendingError) {
+          reject(this.pendingError)
           return
         }
-        resolve(result)
-      })
 
-      this.sendMessage({
-        jsonrpc: '2.0',
-        method: 'generate',
-        params: options,
-        id: messageId,
+        const messageId = this.getMessageId()
+
+        this.registerListener(messageId, (result, error) => {
+          if (error) {
+            reject(error)
+          } else {
+            resolve(mapResult(result))
+          }
+        })
+
+        this.sendMessage({
+          jsonrpc: '2.0',
+          method,
+          params,
+          id: messageId,
+        })
       })
-    })
   }
+
+  getManifest = this.rpcMethod<GeneratorConfig, GeneratorManifest | null>(
+    'getManifest',
+    (result) => (result as { manifest?: GeneratorManifest | null }).manifest ?? null,
+  )
+
+  generate = this.rpcMethod<GeneratorOptions, void>('generate')
 }
--- a/packages/generator-helper/src/GeneratorProcess.ts
+++ b/packages/generator-helper/src/GeneratorProcess.ts
@@ -22,6 +22,7 @@ type GeneratorProcessOptions = {
 export class GeneratorError extends Error {
   public code: number
   public data?: any
+
   constructor(message: string, code: number, data?: any) {
     super(message)
     this.code = code
@@ -43,16 +44,19 @@ export class GeneratorProcess {
     resolve: (result: any) => void
     reject: (error: Error) => void
   }
+
   constructor(private executablePath: string, { isNode = false, initWaitTime = 200 }: GeneratorProcessOptions = {}) {
     this.isNode = isNode
     this.initWaitTime = initWaitTime
   }
+
   async init(): Promise<void> {
     if (!this.initPromise) {
       this.initPromise = this.initSingleton()
     }
     return this.initPromise
   }
+
   initSingleton(): Promise<void> {
     return new Promise((resolve, reject) => {
       try {
@@ -131,6 +135,7 @@ export class GeneratorProcess {
       }
     })
   }
+
   private handleResponse(data: any): void {
     if (data.jsonrpc && data.id) {
       if (typeof data.id !== 'number') {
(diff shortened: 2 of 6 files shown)


packages/cli/src/bin.ts
```diff
@@ -54,9 +54,11 @@ process.removeAllListeners('warning')
 const debug = Debug('prisma:cli')
 process.on('uncaughtException', (e) => {
   debug(e)
+  throw e
 })
 process.on('unhandledRejection', (e) => {
   debug(e)
+  throw e
```
aqrln, line 61: This is unrelated to the referenced issues and can be extracted to a separate PR. These handlers could also lead to the CLI process hanging in some cases as they silently swallowed and ignored any unhandled exceptions (for me the problem occurred when I had a `ReferenceError` in an `exit` event handler in `GeneratorProcess`).

We can also just remove these event handlers if we don't need to log these errors to the debug log specifically.

`git blame` traces all the way back to https://github.com/prisma/prisma/commit/8700714177ada7a8046919a74daab1a9ba567e70, and this code doesn't seem to be present in Prisma 1, so hard to tell what it could have been added for.

packages/generator-helper/src/__tests__/generatorHandler.test.ts
```diff
@@ -61,7 +61,9 @@ function getExecutable(name: string): string {
 }
 
 describe('generatorHandler', () => {
-  // TODO: Windows: this test fails with timeout.
+  // TODO: Windows: this test fails with ENOENT on CI because it can't file the .cmd file:
+  //   spawn D:\\a\\prisma\\prisma\\packages\\generator-helper\\src\\__tests__\\exiting-executable.cmd ENOENT
+  // This does not happen on Windows locally.
```
aqrln, line 66: The timeout was fixed, now this test has the same problem as the next one.

On CI Windows seemingly can't find `exiting-executable.cmd` and `invalid-executable.cmd` in this folder, but these two tests pass for me locally on Windows if I un-skip them, so I couldn't investigate.

The `.cmd` files look identical to the ones in the tests that pass on CI as well, and the paths aren't too long (in fact, they are even longer in some tests that pass) so no idea what could go wrong.

packages/generator-helper/src/__tests__/generatorHandler.test.ts
```diff
@@ -73,12 +75,15 @@ describe('generatorHandler', () => {
     }
   })
 
-  // TODO: Windows: this test fails with ENOENT even though the .cmd file is there and can be run manually.
+  // TODO: Windows: this test fails with ENOENT on CI because it can't file the .cmd file:
+  //   spawn D:\\a\\prisma\\prisma\\packages\\generator-helper\\src\\__tests__\\invalid-executable.cmd ENOENT
+  // This does not happen on Windows locally.
```
aqrln, line 80: Nothing changed here, just updated the comment to be more detailed. The the previous review comment for more info.

packages/cli/src/bin.ts
```diff
@@ -54,9 +54,11 @@ process.removeAllListeners('warning')
 const debug = Debug('prisma:cli')
 process.on('uncaughtException', (e) => {
```
Jolg42, line 55: About this it means that error code will switch to 0, which is confusing

https://nodejs.org/api/process.html#event-uncaughtexception

> The 'uncaughtException' event is emitted when an uncaught JavaScript exception bubbles all the way back to the event loop.
> **By default**, Node.js handles such exceptions by printing the stack trace to stderr and exiting with code 1, overriding any previously set [process.exitCode](https://nodejs.org/api/process.html#processexitcode_1).
> **Adding a handler for the 'uncaughtException' event overrides this default behavior.**
> Alternatively, change the [process.exitCode](https://nodejs.org/api/process.html#processexitcode_1) in the 'uncaughtException' handler which will result in the process exiting with the provided exit code. Otherwise, **in the presence of such handler the process will exit with 0.**

packages/cli/src/bin.ts
```diff
@@ -54,9 +54,11 @@ process.removeAllListeners('warning')
...
   debug(e)
+  throw e
 })
 process.on('unhandledRejection', (e) => {
   debug(e)
```
Jolg42, line 60: If we throw, then we should remove debug to avoid double logging?
```suggestion
```

packages/cli/src/bin.ts
```diff
@@ -54,9 +54,11 @@ process.removeAllListeners('warning')
 const debug = Debug('prisma:cli')
 process.on('uncaughtException', (e) => {
   debug(e)
+  throw e
 })
 process.on('unhandledRejection', (e) => {
```
Jolg42, line 59: About this one, I'm not sure why we need it 🤔 

https://nodejs.org/api/process.html#event-unhandledrejection
> The 'unhandledRejection' event is emitted whenever a Promise is rejected and no error handler is attached to the promise within a turn of the event loop. When programming with Promises, exceptions are encapsulated as "rejected promises". Rejections can be caught and handled using [promise.catch()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) and are propagated through a Promise chain. **The 'unhandledRejection' event is useful for detecting and keeping track of promises that were rejected whose rejections have not yet been handled.**

packages/generator-helper/src/GeneratorProcess.ts
```diff
@@ -20,192 +21,215 @@ type GeneratorProcessOptions = {
...
-            stdio: ['pipe', 'inherit', 'pipe'],
-            env: {
-              ...process.env,
-              PRISMA_GENERATOR_INVOCATION: 'true',
-            },
-            shell: true,
-          })
+      if (this.isNode) {
+        this.child = fork(this.pathOrCommand, [], {
+          stdio: ['pipe', 'inherit', 'pipe', 'ipc'],
```
millsp, line 62: Why is `ipc` needed here?

aqrln: It was already here, I didn't change it. What it does is it creates an IPC channel for passing messages or transferring file descriptors between parent and child, i.e., enables `child.send()` in the parent and `process.on('message')` and `process.send()` in the child. I don't think we use it, so we could try removing this, but I'd probably do that in another PR since it's a bit out of scope.

millsp (approved):
The logic looks sound and better than it was, thank you!

PR #19407, opened 2023-05-23
fix: Update warning message on 10 Client instances

janpio:
The previous  wording was still talking about "actively running". This was true for the Query Engine binary that was starting a web server, opening a port and waiting for requests to that API - but is not any more for the Node-API Query Engine. Here there is nothing "actively running", but it is true that very often starting 10 Prisma Client instances is probably not intentional and should be avoided.


AndrewSouthpaw (commented):
Thanks so much for this!

janpio:
ci test all

Jolg42 (approved):

PR #19414, opened 2023-05-24
fix(client): Ensure empty array argument is not turned into object

### LINKED ISSUE #18846: GraphQL protocol encoder incorrectly turns empty array into empty object.
Schema:

```prisma
generator client {
  provider = "prisma-client-js"
  output   = "../node_modules/.prisma/client"
}

datasource db {
  provider = "postgresql"
  url      = env("TEST_POSTGRES_URI")
}

model User {
  id     String  @id @map("_id")
  active Boolean
}
```

TS:

```ts
  await prisma.user.findMany({
    where: {
      // @ts-ignore
      AND: [[], undefined],
    },
  })
```

This input is invalid: arrays are not valid values for individual elements of `AND` filter. It is correctly rejected by generated TS types. Yet, if we are using it with GraphQL protocol and `@ts-ignore` from a snippet above, the following GraphQL query is produced:

```graphql
query {
  findManyUser(where: {
    AND: [{}]
  }) {
    id
    active
  }
}
```

So, empty array element got turned into an empty object, making the query valid. As soon as we add at least one element to the array, this correctly throws validation error again. 

We believe this is not intentional and validation error should be thrown for empty array case as well. Error is present when JSON protocol is used.

SevInf:
Fix #18846

### DIFF
--- a/packages/client/src/runtime/query.ts
+++ b/packages/client/src/runtime/query.ts
@@ -691,33 +691,53 @@ export class Arg {
     this.hasError =
       Boolean(error) ||
       (value instanceof Args ? value.hasInvalidArg : false) ||
-      (Array.isArray(value) && value.some((v) => (v instanceof Args ? v.hasInvalidArg : false)))
+      (Array.isArray(value) &&
+        value.some((v) => {
+          if (v instanceof Args) {
+            return v.hasInvalidArg
+          }
+          if (v instanceof Arg) {
+            return v.hasError
+          }
+          return false
+        }))
   }
   get [Symbol.toStringTag]() {
     return 'Arg'
   }
   public _toString(value: ArgValue, key: string): string | undefined {
-    if (typeof value === 'undefined') {
+    const strValue = this.stringifyValue(value)
+    if (typeof strValue === 'undefined') {
       return undefined
     }
 
+    return `${key}: ${strValue}`
+  }
+
+  public stringifyValue(value: ArgValue) {
+    if (typeof value === 'undefined') {
+      return undefined
+    }
     if (value instanceof Args) {
-      return `${key}: {
+      return `{
 ${indent(value.toString(), 2)}
 }`
     }
 
     if (Array.isArray(value)) {
       if (this.inputType?.type === 'Json') {
-        return `${key}: ${stringify(value, this.inputType)}`
+        return stringify(value, this.inputType)
       }
 
       const isScalar = !(value as any[]).some((v) => typeof v === 'object')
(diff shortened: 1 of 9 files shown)


packages/client/src/runtime/query.ts
```diff
@@ -1507,12 +1533,12 @@ function tryInferArgs(
...
   return new Arg({
     key,
-    value: value.map((v) => {
+    value: value.map((v, i) => {
       if (inputType.isList && typeof v !== 'object') {
         return v
       }
-      if (typeof v !== 'object' || !value) {
-        return getInvalidTypeArg(key, v, arg, inputType)
+      if (typeof v !== 'object' || !value || Array.isArray(v)) {
```
SevInf, line 1540: This line is an actual fix, the rest is just ensuring that the error renders correctly.

Jolg42: Maybe add some comments? Because it's hard to understand what's happening in general in `query.ts` 😅

packages/client/tests/functional/issues/18846-empty-array/_matrix.ts
```diff
@@ -0,0 +1,24 @@
...
+      provider: 'mongodb',
+    },
+    {
+      provider: 'cockroachdb',
+    },
+    {
+      provider: 'sqlserver',
+    },
+  ],
+])
```
Jolg42, line 24: Technically, we could only test on 1 database? (Is there something interesting to test on different databases for this?)

SevInf: I think generally our approach for functional tests have always been "test on all providers unless you have really good reasons not too", because we had suprirsing failures on selected providers where we were absolutely sure provider makes no difference

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -53,7 +53,7 @@ const args = arg(
 )
 
 async function main(): Promise<number | void> {
-  let jestCli = new JestCli(['--silent', '--config', 'tests/functional/jest.config.js'])
+  let jestCli = new JestCli(['--config', 'tests/functional/jest.config.js'])
```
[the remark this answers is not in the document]
SevInf: Was not meant to be commited, reverted

packages/client/src/runtime/query.ts
SevInf: Added. Not sure how much they help in a grand scheme of things though.

Jolg42 (approved):

aqrln (approved):

PR #19437, opened 2023-05-25
fix(cli): Remove `prisma2` bin

### LINKED ISSUE #19438: Prisma CLI: remove `prisma2` "executable"
Prisma CLI still includes code that it is usable via `prisma2`. One would think that to only be a link or something that needs little to no space, but on MacOS this leads to a 4.5 MB file `node_modules/.bin/prisma2`. 

Usually that does not matter much, but sometimes this [can also be included in serverless functions when not configured properly](https://github.com/prisma/prisma/issues/19358) - making those 4.5MB bigger than they would need to be (and should be, as the CLI of course usually is not needed at all).

We have deprecated the `prisma2` command ages ago, more than 3 years actually with version `2.0.0-beta.3`:
PR: https://github.com/prisma/prisma/pull/2215
https://github.com/prisma/prisma/blob/bfc2193876265c9e36312c661396245dff8a20d0/packages/cli/src/bin.ts#L66-L73

janpio:
closes https://github.com/prisma/prisma/issues/19438

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -56,8 +56,7 @@
     ]
   },
   "bin": {
-    "prisma": "build/index.js",
-    "prisma2": "build/index.js"
+    "prisma": "build/index.js"
   },
   "devDependencies": {
     "@prisma/client": "workspace:*",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -85,7 +85,6 @@
     "fs-jetpack": "5.1.0",
     "get-port": "5.1.1",
     "global-dirs": "3.0.1",
-    "is-installed-globally": "0.4.0",
     "jest": "29.5.0",
     "jest-junit": "16.0.0",
     "kleur": "4.1.5",
--- a/packages/cli/scripts/preinstall.js
+++ b/packages/cli/scripts/preinstall.js
@@ -24,11 +24,11 @@ function isPackageInstalledGlobally(name) {
 }
 
 function prismaIsInstalledGlobally() {
-  const prisma2InstalledGlobally = isPackageInstalledGlobally('prisma2')
-  if (prisma2InstalledGlobally) {
+  const prismaInstalledGlobally = isPackageInstalledGlobally('prisma')
+  if (prismaInstalledGlobally) {
     return {
-      ...prisma2InstalledGlobally,
-      name: 'prisma2',
+      ...prismaInstalledGlobally,
+      name: 'prisma',
     }
   }
 
--- a/packages/cli/scripts/preinstall.js
+++ b/packages/cli/scripts/preinstall.js
@@ -1,42 +1,4 @@
-const path = require('path')
-const globalDirs = require('global-dirs')
 const { drawBox } = require('@prisma/internals/dist/utils/drawBox')
-const isInstalledGlobally = require('is-installed-globally')
-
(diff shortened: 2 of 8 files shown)


packages/cli/package.json
```diff
@@ -56,8 +56,7 @@
     ]
   },
   "bin": {
-    "prisma": "build/index.js",
-    "prisma2": "build/index.js"
+    "prisma": "build/index.js"
```
janpio, line 59: This is the main change.

packages/cli/src/bin.ts
```diff
@@ -63,15 +63,6 @@ process.once('SIGINT', () => {
...
 })
 
-if (process.argv.length > 1 && process.argv[1].endsWith('prisma2')) {
-  console.log(
-    yellow('deprecated') +
-      `  The ${red('prisma2')} command is deprecated and has been renamed to ${green('prisma')}.\nPlease execute ${bold(
-        green('prisma' + (commandArray.length ? ' ' + commandArray.join(' ') : '')),
-      )} instead.\n`,
-  )
-}
```
janpio, line 73: This removes the deprecation warning code.

packages/cli/scripts/preinstall.js
```diff
@@ -24,11 +24,11 @@ function isPackageInstalledGlobally(name) {
...
-  if (prisma2InstalledGlobally) {
+  const prismaInstalledGlobally = isPackageInstalledGlobally('prisma')
+  if (prismaInstalledGlobally) {
     return {
-      ...prisma2InstalledGlobally,
-      name: 'prisma2',
+      ...prismaInstalledGlobally,
+      name: 'prisma',
     }
   }
```
janpio, line 33: This logic seems to have relied on the deprecated bin/executable, and can not any more.

Jolg42: Looks to me that the whole `prismaIsInstalledGlobally()` function and usage could go away as later in this file it's only used to check if `prisma2` is installed globally and error with a message (see line 62+)

janpio: Oh you are right, my change breaks that functionality.

I had considered removing that bit as well, but it is not about the `prisma2` bin of the current Npm package, but about a `prisma2` npm package we used to have. With us not having the`prisma2` bin any more, we honestly don't care if someone still has that old Npm package installed I think.

janpio: This feels pretty drastic, but along those lines?
https://github.com/prisma/prisma/pull/19439

Jolg42: Yes that's it, I think this removal is due and the clean slate very nice

packages/cli/src/bin.ts
Jolg42: I removed the green and yellow imports, which I found in the linting errors as unused

packages/cli/scripts/preinstall.js
janpio, line 1: This change completely removes handling of a `prisma2` Npm package or bin being available on the system.
(Also allowed us to get rid of `is-installed-globally`)

packages/cli/scripts/preinstall.js
janpio: Merged and also removed dependency.

Jolg42 (approved):

PR #19441, opened 2023-05-25
feat(db seed): pass extra arguments to the seed command

### LINKED ISSUE #9403: Pass extra arguments to `prisma db seed` to the seed command
## Problem

I can't pass extra arguments to the command defined in `prisma.seed` in the package.json to the seed command.

## Suggested solution

`prisma db seed -- --extra-args=extra-values`

## Alternatives

Env vars work as well but are not as nice.

Jolg42:
Closes https://github.com/prisma/prisma/issues/9403

### DIFF
--- a/packages/migrate/src/__tests__/DbSeed.test.ts
+++ b/packages/migrate/src/__tests__/DbSeed.test.ts
@@ -21,10 +21,16 @@ describe('seed', () => {
   it('seed.js with -- extra args should succeed', async () => {
     ctx.fixture('seed-sqlite-js-extra-args')
 
-    const result = DbSeed.new().parse(['--', '--my-custom-arg-from-cli=my-value', '-z'])
+    const result = DbSeed.new().parse([
+      '--',
+      '--my-custom-arg-from-cli-1',
+      'my-value',
+      '--my-custom-arg-from-cli-2=my-value',
+      '-z',
+    ])
     await expect(result).resolves.toContain(`The seed command has been executed.`)
     expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(
-      `Running seed command \`node prisma/seed.js --my-custom-arg-from-config=my-value --my-custom-arg-from-cli=my-value -z\` ...`,
+      `Running seed command \`node prisma/seed.js --my-custom-arg-from-config-1 my-value --my-custom-arg-from-config-2=my-value -y --my-custom-arg-from-cli-1 my-value --my-custom-arg-from-cli-2=my-value -z\` ...`,
     )
     expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
--- a/packages/migrate/src/__tests__/DbSeed.test.ts
+++ b/packages/migrate/src/__tests__/DbSeed.test.ts
@@ -18,7 +18,7 @@ describe('seed', () => {
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
   })
 
-  it('seed.js with extra args', async () => {
+  it('seed.js with -- extra args should succeed', async () => {
     ctx.fixture('seed-sqlite-js-extra-args')
 
     const result = DbSeed.new().parse(['--', '--my-custom-arg-from-cli=my-value', '-z'])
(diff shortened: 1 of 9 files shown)


packages/migrate/src/__tests__/DbSeed.test.ts
```diff
@@ -18,6 +18,18 @@ describe('seed', () => {
...
+    ctx.fixture('seed-sqlite-js-extra-args')
+
+    const result = DbSeed.new().parse(['--', '--my-custom-arg-from-cli=my-value', '-z'])
+    await expect(result).resolves.toContain(`The seed command has been executed.`)
+    expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(
+      `Running seed command \`node prisma/seed.js --my-custom-arg-from-config=my-value --my-custom-arg-from-cli=my-value -z\` ...`,
+    )
+    expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+    expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
+  })
```
jkomyno, line 31: This is a test for the successful case, we also need a test for the error case (as per [Slack](https://prisma-company.slack.com/archives/C02FNFLDUS3/p1685006712466749?thread_ts=1685004806.833549&cid=C02FNFLDUS3))

jkomyno: What if args are passed as `--my-custom-arg-from-cli my-value -z`, i.e., without the `=` symbol?

packages/migrate/src/utils/seed.ts
```diff
@@ -150,7 +150,22 @@ export async function getSeedCommandFromPackageJson(cwd: string) {
...
+  extraArgs?: string
+}): Promise<boolean> {
+  let command = commandFromConfig
+
+  // Used in `DbSeed.ts` to pass the extra params after a -- separator
+  // Example: db seed -- --custom-param
+  // Then extraArgs will be '--custom-param'
+  if (extraArgs) {
+    command = `${commandFromConfig} ${extraArgs}`
+  }
```
jkomyno, line 167: Nit: this could be replaced by a `const` and a ternary:

```ts
const command = extraArgs ? `${commandFromConfig} ${extraArgs}` : commandFromConfig
```

packages/migrate/src/__tests__/fixtures/seed-sqlite-js-extra-args/prisma/seed.js
```diff
@@ -0,0 +1,27 @@
+async function main() {
+  await new Promise((resolve) => setTimeout(resolve, 0))
+  console.log('Hello from seed.js')
+
+  if (!process.argv.includes('--my-custom-arg-from-config=my-value')) {
```
jkomyno, line 5: Is the usage of `=` here tied to the way the seed script was called? (E.g., `--my-custom-arg-from-cli=my-value`). Similarly to the comment above, what happens here if the seed script is called with `--my-custom-arg-from-cli my-value` (with no `=`)?

Rationale for this question: consumers of Node.js CLIs are used to be able to define values by either separating the parameter name from his value via a space, or by using a `=` symbol. It seems here we only support `=`, which can be fine per se, but should maybe be documented somewhere

Jolg42: It works the same though better use an args parser, I didn't want to add any extra dep so `.find()` works well for our test here.
Note:
```
// Using parseArgs would be better, but it's only available in v18.3.0, v16.17.0 and up
// https://nodejs.org/api/util.html#utilparseargsconfig
// import { parseArgs } from 'node:util';
```

jkomyno (approved):

packages/migrate/src/commands/DbSeed.ts
```diff
@@ -39,6 +45,11 @@ ${bold('Options')}
     )
 
     if (isError(args)) {
+      if (args instanceof ArgError && args.code === 'ARG_UNKNOWN_OPTION') {
+        throw new Error(`${args.message}
+Did you mean to pass these as arguments to your seed script? If so, add a -- separator before them:
+${dim('$')} prisma db seed -- --arg1 value1 --arg2 value2`)
```
janpio, line 51: In theory this could even use the real arg that created the error :D

Jolg42: We could, but then we would also need to parse the key/values to read the value, which we would "leak" in this error message. It's easy enough for the person to add a `--` where needed here.

PR #19457, opened 2023-05-25
fix(vsce): send reload signal to vscode extension

millsp:
This PR works with https://github.com/prisma/language-tools/pull/1424, we now send a simpler signal for when we need to reload types via our extension, instead of watching many possible paths.

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -1,6 +1,7 @@
 import { overwriteFile } from '@prisma/fetch-engine'
 import type { BinaryPaths, DataSource, DMMF, GeneratorConfig } from '@prisma/generator-helper'
 import { assertNever, ClientEngineType, getClientEngineType, Platform, setClassName } from '@prisma/internals'
+import paths from 'env-paths'
 import fs from 'fs'
 import { ensureDir } from 'fs-extra'
 import { bold, dim, green, red } from 'kleur/colors'
@@ -325,6 +326,11 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   if (!fs.existsSync(proxyIndexBrowserJsPath)) {
     await fs.promises.copyFile(path.join(__dirname, '../../index-browser.js'), proxyIndexBrowserJsPath)
   }
+
+  // we tell our vscode extension to reload the types by modifying this file
+  const prismaCache = paths('checkpoint').cache
+  const signalsPath = path.join(prismaCache, 'last-generate')
+  await fs.promises.writeFile(signalsPath, Date.now().toString()).catch(() => {})
 }
 
 function validateDmmfAgainstDenylists(prismaClientDmmf: PrismaClientDMMF.Document): Error[] | null {
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -327,11 +327,13 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
     await fs.promises.copyFile(path.join(__dirname, '../../index-browser.js'), proxyIndexBrowserJsPath)
   }
 
-  // we tell our vscode extension to reload the types by modifying this file
-  const prismaCache = paths('prisma').cache
-  const signalsPath = path.join(prismaCache, 'last-generate')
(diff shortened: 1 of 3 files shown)


packages/client/src/generation/generateClient.ts
```diff
@@ -325,6 +326,11 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   if (!fs.existsSync(proxyIndexBrowserJsPath)) {
     await fs.promises.copyFile(path.join(__dirname, '../../index-browser.js'), proxyIndexBrowserJsPath)
   }
+
+  // we tell our vscode extension to reload the types by modifying this file
+  const prismaCache = paths('checkpoint').cache
```
janpio, line 331: `checkpoint` would be misleading here, let's use `prisma` or something similar

millsp: I went for that at first, but I feared it would behave differently on windows because we don't use `env-paths` for the engine download, we have something custom. I can of course change that, but not sure what impact that would have
https://github.com/prisma/prisma/blob/ec75a1abb6f61b20e0cb7a90b0500390be6bcbf8/packages/fetch-engine/src/utils.ts/#L12-L33

janpio: I meant to use the string `checkpoint` to get a cache path, vs. using `prisma` as the cache path.
I think you are suggesting to use a completely different way to get a folder?

(I know that the approach that checkpoint uses is good on all OS. I know that the one we use to cache the downloaded files is weird on some where it just puts it into a `node_modules/.cache` or something local to the project instead of global on the machine)

millsp: > I think you are suggesting to use a completely different way to get a folder?
Basically, I wanted to rely on the fact that `.prisma/cache` is already created and use that, but the discrepancy on different platforms led me to abandon the idea.

With your suggestion, `const prismaCache = paths('prisma').cache`, we will obtain a `/home/.cache/prisma-nodejs` path basically, and I just need to ensure that the path exists/is created.

janpio: Yes, I recall that creating the directory the code also needs to take care of - it just gives you the valid paths as string.

packages/client/src/generation/generateClient.ts
```diff
@@ -325,6 +326,12 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   if (!fs.existsSync(proxyIndexBrowserJsPath)) {
     await fs.promises.copyFile(path.join(__dirname, '../../index-browser.js'), proxyIndexBrowserJsPath)
   }
+
+  // we tell our vscode extension to reload the types by modifying this file
+  const prismaCache = paths('prisma').cache
+  const signalsPath = path.join(prismaCache, 'last-generate')
+  await fs.promises.mkdir(prismaCache, { recursive: true }).catch(() => {})
+  await fs.promises.writeFile(signalsPath, Date.now().toString()).catch(() => {})
```
Jolg42, line 334: How does this work if you have multiple (different) projects open in VS Code? It would do a reload all of them?

Jolg42, line 334: We need to handle any error and not send an exitCode 1

Jolg42: done in the `catch()`

Druue (approved):

packages/client/src/generation/generateClient.ts
aqrln: I think it might be not a big deal if it reloads all of them, but if we want to optimise that, we could write a path to the project instead of or together with the timestamp, and then let each language server instance decide if the notification is relevant for it.

Jolg42: We decided that it could be a future optimisation

Jolg42, line 334: nitq: try catch is not pretty but would avoid writing the file if mkdir fails

Jolg42 (approved):

packages/client/src/generation/generateClient.ts
millsp: Added it

Jolg42 (approved):

aqrln (approved):

Druue (approved):

PR #19474, opened 2023-05-26
feat(client): Remove array shortcuts

### LINKED ISSUE #19303: Prisma Client: remove list shorthands (for `jsonProtocol`)
PRs:
- https://github.com/prisma/prisma-engines/pull/3966
- https://github.com/prisma/prisma/pull/19474

If a particular field in a query schema is a list, GraphQL encoder will allow to pass scalars and will auto-wrap it to a list with a single element. Comment around it suggest it was done to work-around engine not having a shortcut in one very specific place, but it does creates way more shortcuts than intended, some of them pretty nonsensical.

Since we don’t have a schema for JSON protocol, we need to add such shortcuts on engine side.

In Prisma 5, we should consider removing this auto-wrapping behavior from GraphQL encoder and some of the engine-level changes we introduced when working on JSON protocol:

- in: https://github.com/prisma/prisma/issues/18549
- Distinct: https://github.com/prisma/prisma/issues/18906 (this one actually has some merit, so might be worth adding it on the schema level instead)
- AND/OR: https://github.com/prisma/prisma/issues/18885
- composite filters: https://prisma-company.slack.com/archives/C04KV4HMEE5/p1676022822842409

💡 

TODO
- [ ] We should also ensure that TS rejects scalars at all places we remove shortcuts

SevInf:
Stop expanding `T[]` to `T | T[]` during types generation unless both
scalars and lists are allowed on the engine side.
GraphQL encoder: Stop converting scalar values into single-element lists
automatically.

BREAKING CHANGE: automatic conversion of scalar values to lists is no
longer performed unless explicitly allowed by the schema. Cases where
this is no longer allowed will be caught by both compile- and runtime
validation.

Sibling to prisma/prisma-engines#3966

Fix #19303

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -153,7 +153,7 @@
     }
   },
   "dependencies": {
-    "@prisma/engines-version": "4.17.0-2.2b04ac1c2af3bc5a845ff7a83ab8be8e0230cb43"
+    "@prisma/engines-version": "4.17.0-4.f6bb63d094cca54895abb071b92b6fcfe2e1d96a"
   },
   "sideEffects": false
 }
--- a/packages/client/src/__tests__/dmmf.test.ts
+++ b/packages/client/src/__tests__/dmmf.test.ts
@@ -51,12 +51,6 @@ describe('dmmf', () => {
                 namespace: model,
                 type: PostKind,
               },
-              {
-                isList: false,
-                location: enumTypes,
-                namespace: model,
-                type: PostKind,
-              },
             ],
             isNullable: false,
             isRequired: false,
@@ -66,12 +60,6 @@ describe('dmmf', () => {
             inputTypes: [
               {
                 isList: true,
-                location: enumTypes,
-                namespace: model,
-                type: PostKind,
-              },
-              {
-                isList: false,
                 location: enumTypes,
                 namespace: model,
                 type: PostKind,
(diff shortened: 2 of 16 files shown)


packages/client/src/__tests__/__snapshots__/noArgs.test.ts.snap
```diff
@@ -70,17 +70,24 @@ Invalid \`prisma.user()\` invocation:
...
   data: {
+    id: 5,
+    name: 'Harshit',
+    email: 'a@a.de',
     likedArticles: {
-?     create?: ArticleCreateWithoutLikedByInput | ArticleCreateWithoutLikedByInput[] | ArticleUncheckedCreateWithoutLikedByInput | ArticleUncheckedCreateWithoutLikedByInput[],
-?     connectOrCreate?: ArticleCreateOrConnectWithoutLikedByInput | ArticleCreateOrConnectWithoutLikedByInput[],
-?     connect?: ArticleWhereUniqueInput | ArticleWhereUniqueInput[]
+      connect: null
+               ~~~~
```
Druue, line 78: Errant red squiggly? 😅

Druue: Ah, looking at other tests, is this part of the snapshotting?

SevInf: Yes, this is how we render errors. And github somehow has a highlighting for squigglies.

Jolg42 (approved):

PR #19517, opened 2023-05-30
fix(clientExtensions): Include composites into default selection

### LINKED ISSUE #17388: [ClientExtensions, TypeScript] Fields of custom type aren't available when using the ClientExtensions preview-feature, unless `select`'d explicitly.
### Bug description

Only when the `clientExtensions` preview-feature is enabled (just by enabling it, no need to define custom extensions), fields of a user-defined custom `type` aren't available after querying the model, unless `select`'d explicitly. 
Note: They are available at runtime, only TypeScript doesn't have them, I'm using the MongoDB as the db `provider`. Also, might be related to #17349, but not sure.

### How to reproduce

Simply create a Prisma `model` that has one or more fields of an user-defined `type`. That field won't be available (TypeScript's types only, it **is** available at runtime) unless it was explicitly `select`d. Full example shown below.

### Expected behavior

Every field (TypeScript) should be available after executing a query, including the ones of custom types. Just like it is when **not** using the `clientExtensions` preview-feature.

### Prisma information

# schema.prisma

```prisma
type CustomType {
  data         String
  otherThings  String
  // ...
}

model ExampleUser {
  id        String @id @default(auto()) @map("_id") @db.ObjectId
  username  String
  password  String
  // ...
  otherData CustomType?
}
```

# Using the prisma client
```typescript

const client = new PrismaClient()

async function main() {
  await client.$connect()
  
  const userNotWorking = await client.exampleUser.findFirst({ 
    where: {
      username: 'test-user'
    } 
  })

 // @ts-expect-error : Here the `otherData` field is not present inside the `userNotWorking` object.
 const dataNotWorking = userNotWorking.otherData

 // To make it work, the query needs `select` to be added with all the fields wanted.
  const user = await client.exampleUser.findFirst({ 
    where: {
      username: 'test-user'
    },
   select: {
      otherData: true
   }
  })


 // Here the `otherData` field is present inside the `user` object.
 const data = user.otherData
}

main()
  .catch(console.error)
  .finally(() => {
    client.$disconnect()
  })
```

### Environment & setup

- OS: Ubuntu
- Database: MongoDB
- Node.js version: Both 14.20.0 and 18.13.0 (LTS)
- Preview Features: clientExtensions
- Typescript: 4.9.4

### Prisma Version

```
Query Engine (Node-API) : libquery-engine ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at ../../node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Migration Engine        : migration-engine-cli ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at ../../node_modules/@prisma/engines/migration-engine-debian-openssl-3.0.x)
Introspection Engine    : introspection-core ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at ../../node_modules/@prisma/engines/introspection-engine-debian-openssl-3.0.x)
Format Binary           : prisma-fmt ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at ../../node_modules/@prisma/engines/prisma-fmt-debian-openssl-3.0.x)
Format Wasm             : @prisma/prisma-fmt-wasm 4.9.0-42.ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5
Default Engines Hash    : ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5
```

SevInf:
Only scalars were included into the default selection when extensions
are on.
Fixed by adding a separate section for composites in a payload type and
changing default selection to `scalars & composites`.
After fixing that, another problem with composites got uncovered: when
client extensions are enabled, delegate types for composites are broken.
Since fluent api does not work for composites (and was not inteded too),
generating delegate types was a mistake to begin with. Fixed by removing
delegate types and fluent getters for composites. I will not call it a
breaking change since again, this never worked and our types were just
incorrect.

Fix #17388

### DIFF
--- a/packages/client/src/generation/TSClient/Model.ts
+++ b/packages/client/src/generation/TSClient/Model.ts
@@ -29,7 +29,6 @@ import {
   getSumAggregateName,
   Projection,
 } from '../utils'
-import { buildComment } from '../utils/types/buildComment'
 import { InputField } from './../TSClient'
 import { ArgsType, MinimalArgsType } from './Args'
 import { TAB_SIZE } from './constants'
@@ -38,12 +37,17 @@ import { TS } from './Generatable'
 import { getArgFieldJSDoc, getArgs, getGenericMethod, getMethodJSDoc, wrapComment } from './helpers'
 import { InputType } from './Input'
 import { ModelFieldRefs } from './ModelFieldRefs'
-import { ModelOutputField, OutputType } from './Output'
+import { buildModelOutputProperty, OutputType } from './Output'
 import { PayloadType } from './Payload'
 import { SchemaOutputType } from './SchemaOutput'
 import { buildIncludeType, buildScalarSelectType, buildSelectType } from './SelectInclude'
 import { getModelActions } from './utils/getModelActions'
 import { ifExtensions } from './utils/ifExtensions'
+
+const extArgsParam = ts
+  .genericParameter('ExtArgs')
+  .extends(ts.namedType('runtime.Types.Extensions.Args'))
+  .default(ts.namedType('runtime.Types.Extensions.DefaultArgs'))
 
 export class Model implements Generatable {
   protected outputType: OutputType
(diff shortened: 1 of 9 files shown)


Jolg42 (commented):
It looks like we do not have tests for a mongodb schema with a composite type and clientExtensions preview feature?

SevInf:
>It looks like we do not have tests for a mongodb schema with a composite type and clientExtensions preview feature?

I've extended `default-selection` test to also run with `clientExtensions` enabled.

packages/client/src/generation/TSClient/Model.ts
```diff
@@ -364,7 +381,7 @@ ${ifExtensions(
   new PayloadType(this.outputType, this.dmmf).toTS(),
 )}
 
-${new ModelDelegate(this.outputType, this.dmmf, this.generator).toTS()}
+${isComposite ? '' : new ModelDelegate(this.outputType, this.dmmf, this.generator).toTS()}
```
Jolg42, line 384:
```suggestion
// fluent api does not work for composites (and was not inteded too)
// so we don't generate delegates for them
${isComposite ? '' : new ModelDelegate(this.outputType, this.dmmf, this.generator).toTS()}
```

Jolg42 (approved):
Looks good and that's a nice cleanup 😉

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -1,76 +1,76 @@
...
+        ? O extends Payload
+          ? GetFindResult<O, S[K]>[]
+          : never
+        : P extends { objects: { [k in K]: infer O | null } }
+        ? O extends Payload
+          ? GetFindResult<O, S[K]> | (P['objects'][K] & null)
+          : never
+        : K extends '_count'
+        ? Count<GetFindResult<P, S[K]>>
+        : never
```
millsp, line 65: I find this flattening really harder to read

packages/client/src/generation/TSClient/Model.ts
```diff
@@ -282,54 +286,67 @@ export type ${getAggregateGetName(model.name)}<T extends ${getAggregateArgsName(
...
+        const objects = ts.objectType()
+        const scalars = ts.objectType()
+        const composites = ts.objectType()
+
+        for (const field of model.fields) {
+          if (field.kind === 'object') {
+            if (this.dmmf.typeMap[field.type]) {
+              composites.add(buildModelOutputProperty(field, this.dmmf))
+            } else {
+              objects.add(buildModelOutputProperty(field, this.dmmf))
```
millsp, line 298: Can this happen?

millsp (approved):

packages/client/src/generation/TSClient/Model.ts
SevInf: Yes. `typeMap` is little bit of a misleading name, but it lists all comosites types. So, if type is listed in `typeMap` it is a composite, if it is not it's just a regular relation.

packages/client/src/generation/TSClient/Output.ts
```diff
@@ -3,45 +3,41 @@ import indent from 'indent-string'
...
 
-export class ModelOutputField implements Generatable {
-  constructor(
-    protected readonly dmmf: DMMFHelper,
-    protected readonly field: DMMF.Field,
-    protected readonly useNamespace = false,
-  ) {}
-  public toTS(): string {
-    const { field, useNamespace } = this
-    // ENUMTODO
```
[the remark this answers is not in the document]
SevInf: Added 3 years ago, by the person who is no longer here, with a commit message "cleanup", so who knows 🤷

PR #19538, opened 2023-05-31
fix(client): Fix batch order for middleware/query combo

### LINKED ISSUE #18276: Sequential transactions run out of order when using client extensions and middleware
### Bug description

When using a batch transaction inside a client query extension, if you are also using a middleware that uses `await` then the batch transaction will run out of order.

### How to reproduce

The bug can be easily demonstrated using the example given [in the docs](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/query#wrap-a-query-into-a-batch-transaction) for batch transactions in queries, including `SET ROLE` statements.

Schema:
```prisma
datasource db {
  provider = "postgresql"
}

generator client {
}

model User {
  email     String   @unique
  name      String?
  role      Role     @default(USER)
}

enum Role {
  USER
  ADMIN
}
```

Minimal reproduction:
```ts

	log: ["query"],
});

	await "test";

	return next(params);
});

const run = async () => {
	try {
		await prisma.$executeRawUnsafe("CREATE ROLE test_user");
	} catch (e) {}

		query: {
			user: {
				async findFirst({ args, query }) {
					const [_, result] = await prisma.$transaction([
						prisma.$queryRawUnsafe("SET ROLE test_user"),
						query(args),
						prisma.$queryRawUnsafe("SET ROLE none"),
					]);
					return result;
				},
			},
		},
	});
	const result = await xprisma.user.findFirst();

	console.log(result);
};

run();
```

The SQL is then executed out of order:
```
prisma:query CREATE ROLE test_user
prisma:query BEGIN
prisma:query SELECT "public"."User"."id", "public"."User"."createdAt", "public"."User"."email", "public"."User"."name", "public"."User"."role" FROM "public"."User" WHERE 1=1 LIMIT $1 OFFSET $2
prisma:query SET ROLE test_user
prisma:query SET ROLE none
prisma:query COMMIT
```
If you comment out the `await 'test'` line, then the transaction will run sequentially.

### Expected behavior

I expect batch transactions to run sequentially when combining query extension and middleware with `await` statement.

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL 11
- Node.js version: 16.8.0

### Prisma Version

```
Query Engine (Node-API) : libquery-engine ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Format Wasm             : @prisma/prisma-fmt-wasm 4.9.0-42.ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5
Default Engines Hash    : ceb5c99003b99c9ee2c1d2e618e359c14aef2ea5
```

SevInf:
Root cause of the problem: data loader batches requests in the order
they come in, which in case of `$transaction` call happens to be the
order in which they are specified in the array. However, depending on
middlewares/extensions (and possibly other factors), it is possible that
they will be delivered to the data loader in a different order.
Fortunately, we already have an index of each request in a batch as a
part of request, so we can just sort the requests before sending them to
the engine.

Fix #18276

### DIFF
--- a/packages/client/src/runtime/DataLoader.ts
+++ b/packages/client/src/runtime/DataLoader.ts
@@ -8,6 +8,9 @@ export type DataLoaderOptions<T> = {
   singleLoader: (request: T) => Promise<any>
   batchLoader: (request: T[]) => Promise<any[]>
   batchBy: (request: T) => string | undefined
+  // Specifies the order in which requests in a batch would
+  // be sorted. See Array.prototype.sort callback
+  batchOrder: (requestA: T, requestB: T) => number
 }
 
 export class DataLoader<T = unknown> {
@@ -65,6 +68,7 @@ export class DataLoader<T = unknown> {
             batch[0].reject(e)
           })
       } else {
+        batch.sort((a, b) => this.options.batchOrder(a.request, b.request))
         this.options
           .batchLoader(batch.map((j) => j.request))
           .then((results) => {
--- a/helpers/compile/plugins/fill-plugin/fillers/process.ts
+++ b/helpers/compile/plugins/fill-plugin/fillers/process.ts
@@ -2,8 +2,6 @@ function noop<T = undefined>(v: T) {
   return () => v as T
 }
 
-const tickPromise = Promise.resolve()
-
 function getProcess() {
   return process
 }
@@ -94,13 +92,9 @@ export const process: NodeJS.Process = {
     rss: 0,
   }),
   nextTick: (fn: Function, ...args: unknown[]) => {
-    tickPromise
-      .then(() => fn(...args))
-      .catch((e) => {
-        setTimeout(() => {
-          throw e
-        }, 0)
-      })
+    setTimeout(() => {
+      fn(...args)
+    }, 0)
   },
   off: noop(getProcess()),
   on: noop(getProcess()),
--- a/packages/client/src/runtime/RequestHandler.ts
+++ b/packages/client/src/runtime/RequestHandler.ts
@@ -119,6 +119,13 @@ export class RequestHandler {
 
         return request.protocolMessage.getBatchId()
       },
+
+      batchOrder(requestA, requestB) {
+        if (requestA.transaction?.kind === 'batch' && requestB.transaction?.kind === 'batch') {
+          return requestA.transaction.index - requestB.transaction.index
+        }
+        return 0
+      },
     })
   }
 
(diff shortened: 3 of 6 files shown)


packages/client/src/runtime/DataLoader.ts
```diff
@@ -65,6 +68,7 @@ export class DataLoader<T = unknown> {
             batch[0].reject(e)
           })
       } else {
+        batch.sort((a, b) => this.options.batchOrder(a.request, b.request))
```
millsp, line 71: Is this a symptom of something we need to fix that lives more deeply, or is it the only way to fix it? I am curious to know why/how we could not preserve the order.

SevInf: So, what happens is:
- we call `prisma.$queryRaw` and await it. This triggers `PrismaPromise` and it passes through all the middlewares and gets to a query extension.
- extension wraps it into `$transaction` and adds 2 extra queries, 1 before and 1 after the original. 
- `$transaction` triggers all 3 `PrismaPromises`. Middle promise (original one) is already in progress and does not need pass through middleware. 2 new requests that extension added still do. As a result, middle promise reaches data loader before either of the outer ones.
-  Due to `lock` promise we won't flush the batch until all Promises have reached data loader (this part works correctly), but `DataLoader` still sends them to the engine in the order they come in, which means, middle promise is sent first. 

I think sorting is simplest solution for this case - we guaranteed to have all necessary promises by the time we flush the batch, we know what order they should be. Even simpler solution would've been re-implementing batching to bypass `DataLoader`, but that would require much more time.

millsp: I think this is a more general problem and is not limited to extensions. This simple code

```ts
  await new Promise((r) => setTimeout(r, Math.random() * 1000))
  return next(params)
})

await prisma.$transaction([
  prisma.$queryRawUnsafe('SELECT 1'),
  prisma.$queryRawUnsafe('SELECT 2'),
  prisma.$queryRawUnsafe('SELECT 3'),
])
```

millsp: Another way to fix it would be to use the locks to unlock in the right order, but let's rather remove all of this asap.

millsp (approved):

aqrln (approved):

millsp (approved):

PR #19554, opened 2023-06-01
fix: allow batchId to be calculated for findUniqueOrThrow in json protocol

hayes:
Currently batchId is only calculated for findUnqiue and not findUniqueOrThrow when using the json protocol.  The graphql protocol doesn't make this distinction. 

There is another bug that prevents findUniqueOrThrow from being batched (see https://github.com/prisma/prisma-engines/pull/4014) but this should prevent a regression once that is resolved.


SevInf:
So, now, when engine PR is merged, I think we are ready to revisit this.
@hayes thanks a lot for both contributions! Could you allow edits from the maintainers on this PR? I'd like to add a couple of functional tests for this.
Alternatively, if you'd like to, you can add them yourself. I'd just copy [this test](https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/batching/tests.ts#L31-L40) and adjust it to use `OrThrow` version.  But leaving that to us is perfectly fine too - both of your patches are already great, leaving final details to us is perfectly fine.

hayes:
Edit's should be allowed.  I probably won't have time today to add tests, but could add them later this week, but also would love to see this merged, so definitely no need to wait on me

SevInf (approved):

SevInf:
Thanks again for your contribution!
It is now merged in dev and will be available in next prisma release

PR #19556, opened 2023-06-01
chore(docker): add name + remove azure-sql-edge

Jolg42:
Adding a name is nicer, all our images are grouped under that name instead of the directory name `docker` (which might be the same in other projects).

By switching from `docker-compose` to `docker compose` we now use Version 2 and remove the old v1 usage which is deprecated:
> Compose V1 no longer receives updates and will not be available in new releases of Docker Desktop after June 2023.
https://docs.docker.com/compose/

Removing azure-sql-edge is now possible since "a while" as the regular mssql image can run in emulation (https://devblogs.microsoft.com/azure-sql/development-with-sql-in-containers-on-macos/)
Note that azure-sql-edge is quite different from the regular mssql image so we better use the regular one.
(See https://learn.microsoft.com/en-us/azure/azure-sql-edge/features, for example "Full-text indexes and search, and related DDL commands and Transact-SQL functions, catalog views, and dynamic management views." or "Spatial data type, and related DDL commands and Transact-SQL functions, catalog views, and dynamic management views." are not supported.)

Logging in to Docker Hub is only needed for Buildjet runners, side-effect: forks will be able to run the workflow without defining the secrets.


.github/workflows/test-template.yml
```diff
@@ -35,7 +35,7 @@ on:
         required: false
       BUILDPULSE_SECRET_ACCESS_KEY:
         required: false
-      # Docker Hub credentials must be passed to avoid rate limiting
+      # Docker Hub credentials must be passed to avoid rate limiting when using Buildjet runners
```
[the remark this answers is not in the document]
millsp: I don't think we use buildjet at all anymore, do we want to keep this in case?

docker/docker-compose.arm64.yml
```diff
@@ -1,11 +0,0 @@
-version: '3.7'
-
-services:
-  mssql:
-    image: mcr.microsoft.com/azure-sql-edge:latest
```
aqrln, line 5: Is overriding the SQL Server image not necessary on ARM anymore?

Jolg42: It's been a while indeed, like said in description.

More info in https://devblogs.microsoft.com/azure-sql/development-with-sql-in-containers-on-macos/

aqrln (approved):

PR #19558, opened 2023-06-01
ci: Silence jest output on CI only

SevInf:
`--silent` flag also silences all logging/debug output, which makes
it pretty inconveneint when trying to debug the test locally. This PR
changes it, so `--silent` flag is used on CI only.


Jolg42 (approved):
Thank you!

PR #19565, opened 2023-06-01
fix(client): bind itx to correct client when extensions are used

SevInf:
Previously, when extension callback was called, `this` within was bound to original client, not `transactionProxy`. This is relatively easy to fix by preserving `this` context of the function in which it was called. 
Consequence of this change: methods, that are not available on ITX client are now optional within the client extensions. Type change was necessary due to runtime behavior change: since extension context can now be bound to either ITX or normal client, there is no guarantee that those methods would exist.

Related fix: on a type elvel `ITXClient` was not receiving `client` components of the extension correctly. Fixed that as well.


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -1053,14 +1053,14 @@ function transactionProxy<T>(thing: T, transaction: PrismaPromiseInteractiveTran
...
-        return (...args: unknown[]) => {
+        return function (this: T, ...args: unknown[]) {
           // we hijack promise calls to pass txId to prisma promises
           if (prop === 'then') return target[prop](args[0], args[1], transaction)
           if (prop === 'catch') return target[prop](args[0], transaction)
           if (prop === 'finally') return target[prop](args[0], transaction)
 
           // if it's not the end promise, result is also tx-proxied
-          return transactionProxy(target[prop](...args), transaction)
+          return transactionProxy(target[prop].apply(this, args), transaction)
```
millsp, line 1063: Is `this` equal to `target`?

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -312,22 +312,23 @@ function interactiveTransactionDefinition(this: PrismaClientClass) {
   }
 
   const returnType = ts.promise(ts.namedType('R'))
+
+  const thisType = ifExtensions<ts.TypeBuilder>(
```
SevInf, line 316: `this` refers to `PrismaClient` type without client component applied (normally, we add later via `GetClient` type), so methods added through that type of extensions were not available to itx client.

packages/client/tests/functional/extensions/enabled/itx.ts
```diff
@@ -209,4 +210,54 @@ testMatrix.setupTestSuite((_0, _1, clientMeta) => {
...
+
+  test('methods from itx client denylist are optional within client extensions', async () => {
+    expect.assertions(10)
+    const xprisma = prisma.$extends({
+      client: {
+        testContextMethods(isTransaction: boolean) {
+          const ctx = Prisma.getExtensionContext(this)
+
+          expectTypeOf(ctx.$connect).not.toMatchTypeOf<typeof prisma.$connect>()
+          expectTypeOf(ctx.$connect).toMatchTypeOf<typeof prisma.$connect | undefined>()
```
millsp, line 222: I think you can combine both with an `.toEqualTypeOf`

millsp (approved):

PR #19570, opened 2023-06-02
fix(client): extensions' incorrect generic return types

millsp:
Client Extensions expose generic types for advanced type safety. They enable the type-safe extensibility for extensions, exposing parts of our input/output type logic processing. In other words, you can use them to get the input and the output types of a given call, generically. This is used by Accelerate and Pulse, but is also available to anyone building extensions.

Some of the return types were incorrect, so this is what this PR fixes, while also adding some tests. The tests aren't exhaustive enough in my opinion, but that's a time sink. Ideally, we would use this central type logic for everything, not just extensions. Let me know if you think a test case is worth expanding.

closes https://github.com/prisma/team-orm/issues/120


millsp:
I had to remove a hack that prevents TypeScript to resolve/expand types into the compiled extension types (in a real-world project, with an extension being output as a package). The test timing out is related to that, but that should not prevent a review already.

SevInf (approved):

PR #19571, opened 2023-06-02
fix(client): raw payload in MW and PCE

### LINKED ISSUE #18092: `$runCommandRaw` is not passing the expected data to middleware or client extension
### Bug description

If you run the following query, the object that you will get in a middleware or a client extension should be the same as the input.

```ts
await xprisma.$runCommandRaw({ aggregate: 'User', pipeline: [], explain: false })
```

```ts
  query: {
    $runCommandRaw({ args, query, operation }) {
      // args should be `{ aggregate: 'User', pipeline: [], explain: false }`
      return query(args)
    },
  },
})
```

However, the args we get are the ones we modify before preparing the query for the query engine. In this case it will be of this form, `command` is completely unexpected and is a correctness issue:

```ts
{ command: { aggregate: 'User', pipeline: [], explain: false } }
```

Originally found while writing a test https://github.com/prisma/prisma/pull/18077/files#diff-000e94f0a63b7522cfe7e2bef894955073865319a763af66c7d8404df538e39cR990

### How to reproduce

Uncomment https://github.com/prisma/prisma/pull/18077/files#diff-000e94f0a63b7522cfe7e2bef894955073865319a763af66c7d8404df538e39cR990

### Expected behavior

`args` in a middleware or in a client extension should be the same as the user input.

### Prisma information

n/a

### Environment & setup

- OS: n/a
- Database: MongoDB
- Node.js version: n/a

### Prisma Version

```
4.11.0-dev
```

millsp:
closes https://github.com/prisma/prisma/issues/18092

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -45,6 +45,7 @@ import { dmmfToJSModelName } from './core/model/utils/dmmfToJSModelName'
 import { ProtocolEncoder } from './core/protocol/common'
 import { GraphQLProtocolEncoder } from './core/protocol/graphql'
 import { JsonProtocolEncoder } from './core/protocol/json'
+import { rawCommandArgsMapper } from './core/raw-query/rawCommandArgsMapper'
 import { RawQueryArgs } from './core/raw-query/RawQueryArgs'
 import { rawQueryArgsMapper } from './core/raw-query/rawQueryArgsMapper'
 import { createPrismaPromise } from './core/request/createPrismaPromise'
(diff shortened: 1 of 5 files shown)


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -664,10 +665,11 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
 
       return createPrismaPromise((transaction) => {
         return this._request({
-          args: { command: command },
+          args: command,
           clientMethod: '$runCommandRaw',
           dataPath: [],
           action: 'runCommandRaw',
+          argsMapper: rawCommandArgsMapper,
```
Jolg42, line 672: Random note: found out in the code that we have one for `rawQueryArgsMapper` but none for runCommandRaw

Nice!

Jolg42 (approved):

SevInf:
Should we also add a test for middleware?

millsp:
Ideally yes, but I think that we'd also need to test all other methods and also test middlewares on mongo. I trust this is a good proxy for now. But we should test middlewares thoroughly for each provider, I think there's quite a gap there atm. I created a ticket for this.

millsp:
Added a missing case for middlewares.

PR #19579, opened 2023-06-02
TMP: run all tests with extensions enabled

SevInf:
Not meant for merging, just for checking that our type tests pass when extensions go GA


millsp:
Related https://github.com/prisma/prisma/pull/19626

PR #19584, opened 2023-06-03
fix(client): generic extensions' type inference

millsp:
Prisma Client extension authors did not get proper type information, although looking at TS code the constraints are all well set. I am convinced this is a shortcoming of the TS compiler. See the fix.

closes https://github.com/prisma/team-orm/issues/110

### DIFF
--- a/packages/client/scripts/default-index.d.ts
+++ b/packages/client/scripts/default-index.d.ts
@@ -44,7 +44,11 @@ export declare type PrismaClientExtends<
   >(
     args:
       | ((client: PrismaClientExtends<ExtArgs>) => { $extends: { extArgs: Args } })
-      | { name?: string; result?: R; model?: M; query?: Q; client?: C },
+      | { name?: string }
+      | { result?: R & runtime.Types.Extensions.UserArgs['result'] }
+      | { model?: M & runtime.Types.Extensions.UserArgs['model'] }
+      | { query?: Q & runtime.Types.Extensions.UserArgs['query'] }
+      | { client?: C & runtime.Types.Extensions.UserArgs['client'] },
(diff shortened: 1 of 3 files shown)


packages/client/scripts/default-index.d.ts
```diff
@@ -44,7 +44,11 @@ export declare type PrismaClientExtends<
   >(
     args:
       | ((client: PrismaClientExtends<ExtArgs>) => { $extends: { extArgs: Args } })
-      | { name?: string; result?: R; model?: M; query?: Q; client?: C },
+      | { name?: string }
+      | { result?: R & runtime.Types.Extensions.UserArgs['result'] }
```
millsp, line 48: The split into a union is just for the beauty of the code. The actual fix is to intersect `R` with `runtime.Types.Extensions.UserArgs['result']`, although `R` already extends `runtime.Types.Extensions.UserArgs['result']`. This is why I think it is more something to do with TS than the actual code.

packages/client/src/runtime/core/extensions/$extends.ts
```diff
@@ -11,11 +11,11 @@ import { OptionalFlat } from '../types/Utils'
 export type Args = OptionalFlat<RequiredArgs>
 export type RequiredArgs = NameArgs & ResultArgs & ModelArgs & ClientArgs & QueryOptions
 
-type NameArgs = {
+export type NameArgs = {
```
SevInf, line 14: These exports don't seem to be used?

millsp: Yep, I'll restore their previous state.

SevInf (approved):

PR #19610, opened 2023-06-05
fix(client): Consistent naming for GroupBy args

SevInf:
Most of other args use non-capitalized name.
Follow up to #19606


Jolg42:
@SevInf Is this ready for a review? It looks like some tests are failing

Jolg42 (approved):

aqrln (approved):

PR #19616, opened 2023-06-05
feat(client): Add version info to default-client

### LINKED ISSUE #18829: Client Extension: Make `.prismaVersion.client` available on "@prisma/client/scripts/default-index"
If I `import { Prisma as PrismaForVersion } from "@prisma/client"` I have `Prisma.prismaVersion.client` available in the extension. If I use  `import { Prisma } from "@prisma/client/scripts/default-index"` then I don't. So currently I use two imports.

SevInf:
Adds client and engine version information to the default-index file.
Versions a picked from `package.json` and `@prisma/engines-version`.

Fix #18829

### DIFF
--- a/packages/client/tests/e2e/default-version/readme.md
+++ b/packages/client/tests/e2e/default-version/readme.md
@@ -1,0 +1,3 @@
+# Readme
+
+This is an example of a readme file. Put a description of your test here.
--- a/packages/client/tests/e2e/default-version/readme.md
+++ b/packages/client/tests/e2e/default-version/readme.md
@@ -1,3 +1,3 @@
 # Readme
 
-This is an example of a readme file. Put a description of your test here.
+Tests that both the client version an the engine can be accessed via the default client.
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -59,7 +59,8 @@
     "edge.d.ts",
     "index.js",
     "index.d.ts",
-    "index-browser.js"
+    "index-browser.js",
+    "version.js"
   ],
   "devDependencies": {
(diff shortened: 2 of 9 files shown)


packages/client/tests/e2e/_utils/run.ts
```diff
@@ -33,6 +33,7 @@ async function main() {
   args['--verbose'] = args['--verbose'] ?? false
   args['--skipBuild'] = args['--skipBuild'] ?? false
   args['--clean'] = args['--clean'] ?? false
+  const filters = args._.flatMap((filter) => [`../*${filter}*/_steps.ts`, `../*${filter}*/**/_steps.ts`])
```
SevInf, line 36: allows to run singular e2e test.

millsp: Is this equivalent to this line https://github.com/prisma/prisma/pull/19616/files#diff-9f3d1a92028dfbb4e992fc36c89be444605b3a820dc36ef26207e31f6febeab0R110?

SevInf: Oh, right, missed that part, will restore to original then

packages/client/tests/e2e/default-version/readme.md
```diff
@@ -0,0 +1,3 @@
+# Readme
+
+This is an example of a readme file. Put a description of your test here.
```
millsp, line 3:
```suggestion
Tests that both the client version an the engine can be accessed via the default client.
```

millsp (approved):

PR #19625, opened 2023-06-05
fix(client): invalid $allModel trigger for raw query

### LINKED ISSUE #19550: PCE: Query extensions of `$queryRawUnsafe` also triggers `$allModels.$allOperations`
> Don't know if it is wanted by design, but I was surprised by the following:
> 
> In that kind of setup:
> ```
> const xprisma = prisma.$extends({
>   query: {
>     $queryRawUnsafe({ args, query, operation }) {
>       // A
>       return query(args);
>     },
>     $allModels: {
>       $allOperations({ model, operation, args, query }){
>         // B
>         return query(args);
>       },
>     },
>   },
> })
> ```
> When `xprisma.$queryRawUnsafe` is used, both A and B are executed, `$allOperations` receiving `undefined` model and `$queryRawUnsafe` operation.
> 
> It isn't very logical IMO, when I add something into $allModels, I don't expect that it will also concern the top-level raw queries.

_Originally posted by @Maxime-J in https://github.com/prisma/prisma/issues/16500#issuecomment-1527803270_

millsp:
closes https://github.com/prisma/team-orm/issues/119
closes https://github.com/prisma/prisma/issues/19550

### DIFF
--- a/packages/client/src/runtime/core/extensions/MergedExtensionsList.ts
+++ b/packages/client/src/runtime/core/extensions/MergedExtensionsList.ts
@@ -47,7 +47,7 @@ class MergedExtensionsListNode {
     })
   }
 
-  getAllQueryCallbacks(jsModelName: string, operation: string) {
+  getAllQueryCallbacks(jsModelName: string | '$none', operation: string) {
     return this.queryCallbacksCache.getOrCreate(`${jsModelName}:${operation}`, () => {
       const prevCbs = this.previous?.getAllQueryCallbacks(jsModelName, operation) ?? []
       const newCbs: QueryOptionsCb[] = []
@@ -69,7 +69,8 @@ class MergedExtensionsListNode {
       }
 
       // when the extension isn't model-bound, apply it to all models
-      if (query['$allModels'] !== undefined) {
+      // '$none' is a special case for top-level operations without model
+      if (jsModelName !== '$none' && query['$allModels'] !== undefined) {
         if (query['$allModels'][operation] !== undefined) {
           newCbs.push(query['$allModels'][operation])
         }
--- a/packages/client/tests/functional/extensions/enabled/query.ts
+++ b/packages/client/tests/functional/extensions/enabled/query.ts
@@ -990,6 +990,94 @@ testMatrix.setupTestSuite(
         expect(fnUser).toHaveBeenNthCalledWith(1, { aggregate: 'User', pipeline: [], explain: false })
       }
     })
+
+    testIf(provider !== 'mongodb')('extending with $allModels.$allOperations and a top-level query', async () => {
+      const fnOperation = jest.fn()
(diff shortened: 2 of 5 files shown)


packages/client/src/runtime/core/extensions/MergedExtensionsList.ts
```diff
@@ -47,7 +47,7 @@ class MergedExtensionsListNode {
     })
   }
 
-  getAllQueryCallbacks(jsModelName: string, operation: string) {
+  getAllQueryCallbacks(jsModelName: string | '*', operation: string) {
```
SevInf, line 50: Can we use some other way to indicate top level operation? I'd read `getAllQueryCallbacks('*')` as "query callbacks for all models"

I think, making `jsModelName` nullable would work better in this case.

packages/client/tests/functional/extensions/enabled/query.ts
```diff
@@ -990,6 +990,92 @@ testMatrix.setupTestSuite(
...
+
+      const xprisma = prisma.$extends({
+        query: {
+          $queryRawUnsafe({ args, query, operation, model }: any) {
+            fnOperation({ args, operation, model })
+
+            return query(args)
+          },
+          $allModels: {
+            async findFirst({ args, query, operation, model }) {
```
SevInf, line 1008: Should this be `$allOperations`?

packages/client/src/runtime/core/extensions/MergedExtensionsList.ts
millsp: Yes, we can, I think this was done for the caching logic, so we had to encode `undefined` into something else since it expects a string and would probably be incorrect to stringify `undefined` as `"undefined"` as that could match a model name (unlikely but possible). So maybe `"<undefined>"` works best? Or something else?

SevInf: Could be `$topLevel` or `$none` then.

packages/client/tests/functional/extensions/enabled/query.ts
millsp: Good catch, thanks.

packages/client/src/runtime/core/extensions/MergedExtensionsList.ts
millsp: Changed to `$none`

SevInf (approved):

SevInf:
I think new tests need a skip for Mongo, but otherwise LGTM

millsp:
FYI: I changed this test because it was invalid with this change. There is an issue to implement a top-level `$allOperations`. https://github.com/prisma/prisma/pull/19625/files#diff-a22ee5e3a3ee441dedf6b4fea394ef2b3820c5a84a4df1cecf27bad0803c75ebL27``

PR #19626, opened 2023-06-05
chore(client): remove extensions preview

millsp:
closes https://github.com/prisma/team-orm/issues/176

Additionally, the PR brings a new `--generate-only` option for testing for when working on type-level/generation code. Removing the preview feature also uncovered a two type bugs. The first one is about `rejectOnNotFound` not being taken into account, and the second was about allowing extraneous arguments to be passed without raising an error. With this, I also took the advantage to rename the very hard to read types via new aliases, along with the removal of `ifExtensions`.


packages/client/src/runtime/core/types/Utils.ts
```diff
@@ -68,10 +68,14 @@ export type Cast<A, W> = A extends W ? A : W
...
-  {[K in keyof W]: K extends keyof A ? LegacyExact<A[K], W[K]> : W[K]}>
+  { [K in keyof A]: K extends keyof W ? LegacyExact<A[K], W[K]> : never },
+  { [K in keyof W]: K extends keyof A ? LegacyExact<A[K], W[K]> : W[K] }>
   : never;
 
 export type JsonObject = { [Key in string]?: JsonValue }
 export interface JsonArray extends Array<JsonValue> {}
 export type JsonValue = string | number | boolean | JsonObject | JsonArray | null
+
+export type Record<T extends string | number | symbol, U> = {
```
SevInf, line 79: Is it different from built-in `Record`?

SevInf (approved):

packages/client/src/runtime/core/types/Utils.ts
millsp: Same

PR #19657, opened 2023-06-07
test: negative zero warning in dependencies

Jolg42:
Same as https://github.com/prisma/prisma/pull/14384 but for the newer version

It removes this log
```
{
  column: 27,
  file: '/Users/j42/Dev/prisma/packages/engines/dist/index.js',
  length: 2,
  line: 16286,
  lineText: '          } else if (x === -0) {',
  namespace: '',
  suggestion: ''
}
Comparison with -0 using the "===" operator will also match 0
```

when for example running `pnpm run test:functional:code`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -141,7 +141,7 @@ importers:
       '@types/fs-extra': 9.0.13
       '@types/jest': 29.5.1
       '@types/rimraf': 3.0.2
-      checkpoint-client: 1.1.23
+      checkpoint-client: 1.1.24
       debug: 4.3.4
       dotenv: 16.0.3
       esbuild: 0.15.13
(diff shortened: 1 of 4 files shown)


Druue (approved):

pnpm-lock.yaml
```diff
@@ -11947,8 +11947,9 @@ packages:
         optional: true
     dependencies:
       whatwg-url: 5.0.0
+    patched: true
 
-  /node-fetch/2.6.7_3wcp6bao3dfksocwsfev5pt7km:
+  /node-fetch/2.6.7:
```
Jolg42, line 11952: Only used via `"vercel": "28.13.0"` in `prisma/packages/client/tests/e2e/` tests like `nextjs-schema-not-found/16_monorepo-serverComponents-customOutput-reExportIndirect/packages/service/package.json`

Jolg42: and actually `checkpoint-client/1.1.23` so I'll fix that

SevInf (approved):

Issue #4703, opened 2020-12-17
Run prisma migrate command programmatically


dested:
## Problem

It would be nice to have a way to run the migrate programmatically. I am running into an issue where my api is using lambdas (serverless) and my database is only accessible from within an aws vpc. Because of this I cannot "execute commands" on deploy in an easy way. My current solution is to ssh into an ec2 that is within the vpc, execute the prisma commands, and then spin the ec2 back down. While this is working, it is far from ideal. If I was instead able to have a separate lambda that had all the migration data bundled with it I could execute the migration from a lambda (that is within the vpc). 

While this may seem like a specialized use case, databases that are not publicly accessible are common place, and solutions for migration that don't include running npm commands from the cli I think become critical.

## Solution
```
import {migrate} from '@prisma/cli';
migrate('up');
```

## Alternatives
Stated above, no great alternatives for common usecases.

PhilippMolitor:
I had some thoughts on this feature and already shared them with @albertoperdomo from the Prisma team.
To me, a very essential feature that would be possible with this addition would be to run a custom seeder.
Imagine freshly deploying the app onto a server, and when it detects that there are 0 migrations in the database yet, it runs a custom function creating some essential entries.

Imagine something like this:

```typescript
const connection = new PrismaClient();

connection.$connect()
  .then(() => {
    const migrations = connection.$migrations();

    const isFreshInstall = migrations.databaseIsInitialized;

    if(migrations.pending > 0) {
      migrations.runPendingMigrations();
    } else if (migrations.schemaNotMatching) {
      throw new Error('server schema newer or not compatible with application schema');
    }

    if(isFreshInstall) {
      connections.mymodel.create({data: { ... }});
      console.log('database was seeded!');
    }
  });
```

yardenshoham:
I'm packaging my app with `pkg` and I can't expect the environment to have node/npm/npx/yarn. How can I go about creating the sqlite file and initializing the database programmatically (basically `npx prisma migrate reset --preview-feature`) so in a new environment, the app could be run without any shell interaction?

johnnyoshika:
Automatic migration in code would be really nice for me as well.

AhmetHuseyinDOK:
I also need this feature for my projects too, love prisma, hope this will feature will be implemented soon

thdxr:
[...]

PR #19660, opened 2023-06-07
fix(deps): install missing arg dependency to migrate package

boennemann:
"@prisma/migrate" is directly importing "arg", yet not installing it as a dependency.

This will work most of the time, because it also depends on "@prisma/internals" which *most of the time* will cause it to incidentally be available in the top-level node_modules folder.

However when installing other packages that depend on a different version of arg (e.g. ts-node) the migrate package explodes ("Can not import module arg"), because arg gets pushed into `/node_modules/@prisma/internals` only.

This PR adds an explicit dependency definition.

### DIFF
--- a/packages/migrate/package.json
+++ b/packages/migrate/package.json
@@ -43,6 +43,7 @@
     "@prisma/debug": "workspace:*",
     "@prisma/get-platform": "workspace:*",
     "@sindresorhus/slugify": "1.1.2",
+    "arg": "5.0.2",
     "execa": "5.1.1",
     "fp-ts": "2.16.0",
     "get-stdin": "8.0.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -784,6 +784,7 @@ importers:
       '@types/pg': 8.10.2
       '@types/prompts': 2.4.4
       '@types/sqlite3': 3.1.8
+      arg: 5.0.2
       esbuild: 0.15.13
       execa: 5.1.1
       fp-ts: 2.16.0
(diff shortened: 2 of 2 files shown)


boennemann:
Test failure seems unrelated to what I have changed.

Jolg42:
Note: It looks like the only place where `arg` is used directly in `@prisma/migrate` is in `db seed` command for `import { ArgError } from 'arg'`

Jolg42:
Thanks for your first PR here @boennemann 🙌🏼 
On a sidenote, I'm quite curious about why you are using the `@prisma/migrate` directly, if you want to let me know, I'm very curious 👀

Jolg42 (approved):

boennemann:
> Thanks for your first PR here @boennemann 🙌🏼
🙌🙌 

> On a sidenote, I'm quite curious about why you are using the `@prisma/migrate` directly, if you want to let me know, I'm very curious 👀

It's honestly something that I would not be mad about at all if we wouldn't have to do it anymore.
When we first setup prisma we wanted the migrations to be automatically executed as part of our CI and CD, but there was no way to programatically access migrate to configure it. We set up a script based on https://github.com/prisma/prisma/issues/4703#issuecomment-879857806 and didn't have any issues with it since.

Jolg42:
Totally understand, thanks for sharing 🙌🏼
I think this (official & documented programmatic access to Migrate commands) needs to happen one day.

janpio:
Just to close the loop here: https://github.com/prisma/prisma/issues/4703 (Hopefully soon... but happy to hear your current approach works fine - unless when we forget some dependencies :D )

PR #19685, opened 2023-06-09
feat(client): Add private extension API for hooking into batching logic

SevInf:
Adds private `$__internalBatch` key for query extensions. This callback
will be exectured before the batch query is executed, after inidivdual
callbacks are done. It will allow Accelerate extension to proberly
populate cache info for batches.

Eventually, we'd like to make this API public, but for the first
iteration we keep it private. Only functionality, necessary for
Accelarate is implemented at the moment. Important missing feature for
publicizing this API: ability to change arguments within
`$__internalBatch` callback, similar to normal query extensions.

Fix prisma/team-orm#182

### DIFF
--- a/packages/client/src/runtime/RequestHandler.ts
+++ b/packages/client/src/runtime/RequestHandler.ts
@@ -106,7 +106,7 @@ export class RequestHandler {
               }
 
               try {
-                return this.processResponse(requests[i], result)
+                return this.mapQueryEngineResult(requests[i], result)
               } catch (error) {
                 return error
               }
@@ -124,7 +124,7 @@ export class RequestHandler {
           isWrite: request.protocolMessage.isWrite(),
           customDataProxyFetch: request.customDataProxyFetch,
         })
-        return this.processResponse(request, response)
+        return this.mapQueryEngineResult(request, response)
       },
 
       batchBy: (request) => {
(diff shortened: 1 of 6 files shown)


packages/client/src/runtime/RequestHandler.ts
```diff
@@ -59,15 +62,6 @@ export type HandleErrorParams = {
   transaction?: PrismaPromiseTransaction
 }
 
-export type Request = {
```
SevInf, line 62: We had to add more and more things from `RequestParams` here, so it made less and less sense to keep this object to me. I've decided we could just pass `RequestParams` to dataloader directly instead.

millsp: I agree, better this way.

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -77,41 +71,62 @@ type ApplyExtensionsParams = {
...
-
-        const containsWrite = requests.some((r) => r.protocolMessage.isWrite())
-
-        return this.client._engine.requestBatch(queries, {
-          traceparent,
-          transaction: getTransactionOptions(transaction),
-          containsWrite,
-          customDataProxyFetch: requests[0].customDataProxyFetch,
-        })
+        return applyQueryBatchExtensions(
```
millsp, line 82: Could we make this function a partial application so we don't need to pass request again?

millsp (approved):

PR #19653, opened 2023-06-07
fix(client): Fix queryRaw methods within itx when extensions are used

SevInf:
Stop passing `MergedExtensionList` property through itx proxy.

Fix #19651

PR #19691, opened 2023-06-09
fix(client): broken tests with extensions

millsp:
closes https://github.com/prisma/team-orm/issues/204

### DIFF
--- a/packages/client/src/runtime/core/extensions/$extends.ts
+++ b/packages/client/src/runtime/core/extensions/$extends.ts
@@ -91,7 +91,6 @@ export function $extends(this: Client, extension: Args | ((client: Client) => Cl
   const newClient = Object.create(oldClient, {
     _extensions: {
       value: this._extensions.append(extension),
-      configurable: true,
     },
   }) as Client
 
--- a/packages/client/src/runtime/core/extensions/$extends.ts
+++ b/packages/client/src/runtime/core/extensions/$extends.ts
@@ -92,7 +92,6 @@ export function $extends(this: Client, extension: Args | ((client: Client) => Cl
     _extensions: {
       value: this._extensions.append(extension),
       configurable: true,
-      enumerable: true,
     },
   }) as Client
 
--- a/packages/client/src/runtime/core/extensions/$extends.ts
+++ b/packages/client/src/runtime/core/extensions/$extends.ts
@@ -91,6 +91,8 @@ export function $extends(this: Client, extension: Args | ((client: Client) => Cl
   const newClient = Object.create(oldClient, {
     _extensions: {
       value: this._extensions.append(extension),
+      configurable: true,
+      enumerable: true,
     },
   }) as Client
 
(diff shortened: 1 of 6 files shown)


millsp:
TODO: Once you approve this PR, I'll turn this testing into a flag `--extensions-smoke-test` or something like that, better ideas for a name?

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -97,7 +97,15 @@ function setupTestSuiteMatrix(
         globalThis['newPrismaClient'] = (...args) => {
           const client = new global['loaded']['PrismaClient'](...args)
           clients.push(client)
-          return client
+          return new Proxy(client.$extends({}), {
+            get(target, prop) {
+              if (prop === '$use' || prop === '$on') {
```
SevInf, line 102: That feels a little bit hacky to me. If we are going to keep this permanently, would it be better to skip the tests for `use` and `prop` permanently?

packages/client/src/runtime/core/extensions/$extends.ts
```diff
@@ -91,6 +91,8 @@ export function $extends(this: Client, extension: Args | ((client: Client) => Cl
   const newClient = Object.create(oldClient, {
     _extensions: {
       value: this._extensions.append(extension),
+      configurable: true,
+      enumerable: true,
```
SevInf, line 95: I don't think it needs to be `enumerable`

SevInf:
Actually, I am not 100% sure we need this permanently. The only thing this run found already was fixed in other PR, just not merged. If that's the case, I'd rather vote for simplicity and less configuration options for test runner.

millsp:
> The only thing this run found already was fixed in other PR, just not merged.

Not actually, `configurable: true` was required for many tests to pass so that uncovered at at least missing test case.

millsp:
> Are we going to run it on CI as well?

I don't think so, I would keep it for future extension development. Right now if we add the missing test case, there is no immediate reason to continuously test it (or maybe only via nightly?).

SevInf:
>Not actually, configurable: true was required for many tests to pass so that uncovered at at least missing test case.
But that's the same issue as #19653, that has a test.

millsp:
> But that's the same issue as https://github.com/prisma/prisma/pull/19653, that has a test.

I think it's a new case, look a the tests failing.

SevInf:
I think that's caused by extra `Proxy`  in the test setup itself, not something client- or extenions specific.

millsp:
Ok, let me give that a try and report back.

SevInf (approved):

millsp:
After your suggestion, turned out that we had two tests failing. One was because for extensions, it required some extra default inputs which were missing. For the other, it was a legitimate failure not being able to cancel a transaction via middleware and with the combination of an empty extension. Added a test and a fix for that, and removed the smoke test.

PR #19695, opened 2023-06-09
chore(client): deprecate middlewares

millsp:
Closes https://github.com/prisma/team-orm/issues/181
Works with https://github.com/prisma/pris.ly/pull/110

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -777,6 +777,8 @@ export class PrismaClient<
 
   /**
    * Add a middleware
+   * @deprecated since 4.16.0. Use query extensions instead.
+   * @see https://pris.ly/d/extensions
    */
   $use(cb: Prisma.Middleware): void
 
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -777,7 +777,7 @@ export class PrismaClient<
 
   /**
    * Add a middleware
-   * @deprecated since 4.16.0. Use query extensions instead.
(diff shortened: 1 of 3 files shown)


packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -777,6 +777,8 @@ export class PrismaClient<
 
   /**
    * Add a middleware
+   * @deprecated since 4.16.0. Use query extensions instead.
```
SevInf, line 780: Since we want it to be a softer deprection, maybe, softer warining would work better instead. "For new code, prefer client extensions instead" for example.

Jolg42: Adding a link to docs of query extenstions would make it easier for users I think

millsp:
```suggestion
   * @deprecated since 4.16.0. For new code, prefer client extensions instead.
```

packages/client/src/__tests__/integration/happy/exhaustive-schema/__snapshots__/test.ts.snap
```diff
@@ -739,6 +739,8 @@ export class PrismaClient<
 
   /**
    * Add a middleware
+   * @deprecated since 4.16.0. Use query extensions instead.
```
millsp, line 742:
```suggestion
   * @deprecated since 4.16.0. For new code, prefer client extensions instead.
```

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -508,6 +508,8 @@ export class PrismaClient<
 
   /**
    * Add a middleware
+   * @deprecated since 4.16.0. Use query extensions instead.
```
millsp, line 511:
```suggestion
   * @deprecated since 4.16.0. For new code, prefer client extensions instead.
```

SevInf (approved):

PR #19696, opened 2023-06-09
feat(client): new extensions endpoint

millsp:
closes https://github.com/prisma/team-orm/issues/185


packages/client/scripts/default-index.js
```diff
@@ -25,10 +25,12 @@ module.exports = {
...
     get prismaVersion() {
-      // default-index might gets copied to a .prisma/client directory and
-      // no longer be able to require @prisma/engines-version package. Delegating
-      // version information to the file at stable location allows us to overcome this.
-      return require('@prisma/client/version')
+      const { version, dependencies } = require('@prisma/client/package.json')
+
+      return {
+        client: version,
+        engine: dependencies['@prisma/engines-version'].split('.')[3],
```
SevInf, line 32: I don't think that matches generated version for integration branches. Version in the `package.json` contains branch name, while `prismaVersion.engine` is always just a commit hash.

millsp: I tried to move everything together because it makes things simpler for latter ESM support. Now applied a solution to your suggestion.

SevInf (approved):
Should we also update `defineExtension` functional test to use new import?

packages/client/scripts/default-index.d.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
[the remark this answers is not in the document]
millsp: Somehow eslint complains on the exports if I don't

millsp:
> Should we also update defineExtension functional test to use new import?

I will try to change it, iirc I think there was a limitation there, but will double check. That said, there is a test that packs an extension in real conditions, at least.

millsp:
Updated to `../../../extension`

PR #19698, opened 2023-06-10
feat(client): improved validator

### LINKED ISSUE #18943: Prisma Client Extension: support `Prisma.validator`
> I got asked to make a feature request by @millsp https://github.com/prisma/prisma/issues/15074#issuecomment-1384354061            

I always define a `select:` when doing queries (because otherwise we get the equivalent of `select *`), and want to do the equivalent to the below but with my default prisma client:

```ts
const postSelect = Prisma.validator<Prisma.PostSelect>()({
  // ..
})
```

_Originally posted by @KATT in https://github.com/prisma/prisma/issues/15074#issuecomment-1382089974_

millsp:
closes https://github.com/prisma/prisma/issues/18943

Previously, you could use the `validator` in this way, eg. to validate a selection for reuse. This is convenient but perhaps a bit counter-intuitive to reference internal types, from the user's perspective. This is how you validate a select on the type-level:
```ts
const data1 = Prisma.validator<PrismaNamespace.UserSelect>()({
  id: true,
})
```

This PR introduces a new way to do the same, but without the use of types, while also being compatible with extensions. The semantics change a bit, but also resemble more the actual `PrismaClient` API. The equivalent to the code above would be:
```ts
const data2 = Prisma.validator(prisma, 'user', 'findFirst', 'select')({
  id: true,
})
```

Note that the operation name `findFirst` is required, so the new variant needs to be more specific even for `select`. But of course, this is not limited to `select` and is also meant to work with other data inputs:
 
 ```ts
const data2 = Prisma.validator(prisma, 'user', 'create', 'data')({
  name: 'John Doe',
})
```
 
The new signature is as follows, in pseudo-code:

```ts
function validator(
	client: /* a PrismaClient or an extended PrismaClient */
	model: /* a model name, which will show up on auto-completion */
	operation: /* an operation name, which will show up on auto-completion */
	prop?: /* an optional property name, to validate a sub-part of the operation input */
) => (
	args: /* the operation input or sub-part to be validated on the type-level */
) => args /* the same args, validated and narrowed as passed in the input */
```

https://www.prisma.io/docs/concepts/components/prisma-client/advanced-type-safety/prisma-validator

### DIFF
--- a/packages/client/src/generation/TSClient/common.ts
+++ b/packages/client/src/generation/TSClient/common.ts
@@ -529,12 +529,6 @@ export function validator<V>():
 
 export function validator<
   C,
-  M extends Exclude<keyof C, \`$\${string}\`>
->(client: C, model: M):
-  <S>(select: $Utils.Exact<S, $Public.Args<C[M], 'findFirstOrThrow'>['select']>) => S;
-
-export function validator<
-  C,
   M extends Exclude<keyof C, \`$\${string}\`>,
   O extends keyof C[M] & $Public.Operation,
 >(client: C, model: M, operation: O):
--- a/packages/client/src/generation/TSClient/common.ts
+++ b/packages/client/src/generation/TSClient/common.ts
@@ -529,7 +529,14 @@ type Cast<A, B> = A extends B ? A : B;
 
 export const type: unique symbol;
 
-export function validator<V>(): <S>(select: $Utils.LegacyExact<S, V>) => S;
+export function validator<V>():
+<S>(select: $Utils.LegacyExact<S, V>) => S;
+export function validator<T>(client: T):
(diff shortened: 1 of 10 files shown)


packages/client/src/generation/TSClient/common.ts
```diff
@@ -524,7 +524,23 @@ type Cast<A, B> = A extends B ? A : B;
 
 export const type: unique symbol;
 
-export function validator<V>(): <S>(select: $Utils.LegacyExact<S, V>) => S;
+export function validator<V>():
```
SevInf, line 527: I think it should be fine to move it to runtime since it became way more complex than it used to be and also does not really depend on DMMF

millsp: Agree, let's move as many parts as we can

SevInf (approved):

PR #19711, opened 2023-06-12
fix(client): Improve union error ranking heuristics

### LINKED ISSUE #19707: jsonProtocol: Wrong error message shown for validation error for checked/unchecked types
Schema:

```prisma
generator client {
  previewFeatures = ["jsonProtocol"]
}

datasource db {
  provider = "postgresql"
  url      = env("TEST_POSTGRES_URI")
}

model User {
  id             Int          @id @default(autoincrement())
  tags           String[]
  organizationId Int          @unique
  organization   Organization @relation(fields: [organizationId], references: [id])
}

model Organization {
  id   Int   @id @default(autoincrement())
  user User?
}
```

Code:

```ts
await prisma.organization.create({
  data: { id: 123 },
})

await prisma.user.create({
  data: {
    tags: 'admin',
    organizationId: 123,
  },
})
```

The actual error here: `tags` field is supposed to be an array. The error shown instead:

```
PrismaClientValidationError:
Invalid `prisma.user.create()` invocation in
/Users/serhii/projects/prisma/reproductions/workbench/index.ts:13:21

  10   data: { id: 123 },
  11 })
  12
→ 13 await prisma.user.create({
       data: {
         tags: "admin",
         organizationId: 123,
     +   organization: {
     +     create: OrganizationCreateWithoutUserInput | OrganizationUncheckedCreateWithoutUserInput,
     +     connectOrCreate: OrganizationCreateOrConnectWithoutUserInput,
     +     connect: OrganizationWhereUniqueInput
     +   }
       }
     })

Argument `organization` is missing.

```

which is misleading and unhelpful. If we change the code to use checked input, we'd get the opposite, "organizationId" is missing from unchecked variant.

SevInf:
When argument value does not match any of the allowed input types for a
particular field, engine returns a union of individual errors for each
possible types. We then use heuristics to pick the best one to render on
the client side.
Previously, this heuristics was based only on the length of argumentPath
and selectionPath. Logic was, the type which made it deeper into the
user's input before failing is probably closer to users intentions. This
worked quite well for the most cases, however, if the path length of two
errors is the same it just picked the first one and it occasionally
produced confusing result.

Consider #19707 for example. There, we receive a union of two errors:
1. Invalid input for checked type: `data.organization` field is missing
2. Invalid input for unchecked type: `data.tags` must be a list

Obviously, the second error is what user wanted to be told. However,
both `data.organization` and `data.tags` have a length of 2 and so we
pick the first one to show instead, completely misrepresenting what the
problem is.

This PR introduces another step to the heuristics: if path of two errors
has the same length, we rank some of the errors higher than the rest:

1. `InvalidArgumentValue` is ranked highest, since when this error
   happens we can be sure that particular field actually exists on the
   schema and the input has correct type, just that particular value can
   not be accepted.
2. `InvalidArgumentType` is ranked below `InvalidArgumentValue` but
   above everything else, since in that case we are at least sure that
   the field user supplied exists on the encolsing type and it still has
   higher chance of being helpful to the user than any other error.
3. Rest errors are ranked below `InvalidArgumentType` without a
   particular ordering between them.

Fix #19707


packages/client/src/runtime/core/errorRendering/applyValidationError.test.ts
```diff
@@ -2329,10 +2329,71 @@ describe('Union', () => {
               }
             }
 
-            Argument \`<bold>gt</intensity>\`: Invalid value provided. Expected <green>String</color> or <green>StringFilter</color>, provided <red>Int</color>.
+            Argument \`<bold>email</intensity>\`: Invalid value provided. Expected <green>String</color> or <green>StringFilter</color>, provided <red>Int</color>.
 
         `)
   })
+
+  test('longest path: InvalidArgumentType vs RequiredArgumentMissing', () => {
```
Jolg42, line 2337: Here it's actually not about the longest path? But about `InvalidArgumentType` score?

Jolg42 (approved):
Looks good to me (though I don’t remember the last time I touched something around there)

Issue #12469, opened 2022-03-21
Certain `model` names cause clashes in generated types


austincrim:
### Bug description

**existing issues**
- #12332 
- #10490
- #9773
- #9669
- #9568, #7518

The following are all exported types from the `index.d.ts` and will cause clashes if they have corresponding `models` in the Prisma schema.

- PrismaPromise
- PrismaVersion
- JsonObject
- JsonValue
- InputJsonObject
- InputJsonValue
- PromiseType
- PromiseReturnType
- Enumerable
- RequiredKeys
- TruthyKeys
- TrueKeys
- Subset
- SelectSubset
- SubsetIntersection
- UnEnumerate
- Union 
- IntersectOf
- Overwrite
- At
- ComputeRaw
- OptionalFlat
- Strict
- Merge
- Boolean 
- True 
- False 
- Not
- Extends
- Has
- Or
- Keys
- GetScalarType
- ModelName 
- Datasources
- RejectOnNotFound 
- RejectPerModel
- RejectPerOperation 
- HasReject
- ErrorFormat 
- Hooks
- LogLevel 
- LogDefinition
- GetLogType
- GetEvents
- QueryEvent
- LogEvent
- PrismaAction 
- MiddlewareParams
- Middleware
- Datasource
- SortOrder 
- QueryMode 
- IntFilter
- StringFilter
- IntWithAggregatesFilter
- StringWithAggregatesFilter
- StringFieldUpdateOperationsInput
- IntFieldUpdateOperationsInput
- NestedIntFilter
- NestedStringFilter
- NestedIntWithAggregatesFilter
- NestedFloatFilter
- NestedStringWithAggregatesFilter
- BatchPayload

### How to reproduce

1. Generate a schema with one of the model names listed above
2. Write Prisma query
3. Observe return type is incorrect

### Prisma information

n/a

### Environment & setup

- OS: macOS
- Node.js version: 17.7.1

### Prisma Version

```
3.11
```

psikoi:
`Metric` is now also causing conflicts with model names.

priithaamer:
Just got hit with this by migrating prisma from `3.x` to `4.3` on a project that had a model named `Metric`.

Renaming the model worked for me but is there another workaround besides that?

psikoi:
> Just got hit with this by migrating prisma from `3.x` to `4.3` on a project that had a model named `Metric`.
> 
> Renaming the model worked for me but is there another workaround besides that?

[...]

PR #19732, opened 2023-06-13
fix(client): Allow to use `Metric` as a model name

SevInf:
Problem: even though built-in `Metric` type is placed in
a different namespace, type conflict still occurs: `Delegate` classes
are in the same namespace and can't refer to non-namespaced name.

Implements a tiny part of [namespace overhaul proposal](https://www.notion.so/prismaio/Fix-type-name-clashes-in-TS-Client-9e9bbdee13f94a818ece060e55465267): introduces `$Models` namespace.

Ref #12469


SevInf:
Yes, we can.
Closing in favor of #20150

PR #19750, opened 2023-06-14
fix(client): Simplify iTX proxy and apply it only to the client

SevInf:
In #19565 we had to change how iTx client calls it's methods.

Previously, it was calling them on default client and appending
transaction information afterwards. After that change, it started
calling them on iTxClient.

Unitended consequence of that: now every cleint property is proxied,
including private properties, private properties of nested object,
returned results, etc. It breaks in a situation like #19653 and #19685.

This PR attempts to rework how itxProxy works. The only thing we need
from itxProxy is ability to pass transaction paramters to the queries
within the callback. Here, it is achieved by introducing a factory for
PrismaPromises that is bound to a specific client instance. When used on
default transaction, factory creates normal promise. ITX client replaces
the default factory with the one that always produces promises, bound to
a specific transaction.

So, instead of deeply proxying everything on itxClient, it creation is
now this:
- replace promise factory with the one that is bound to a transaction
- remove denylist properties
- add TX_ID property

This is better and safer solution to #19565, as well as code
simplification.

Unblocks #19685

### DIFF
--- a/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
+++ b/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
@@ -6,7 +6,7 @@ export interface CompositeProxyLayer<KeyType extends string | symbol = string | 
   /**
    * Returns a list of keys, defined by a layer
    */
-  getKeys(): KeyType[]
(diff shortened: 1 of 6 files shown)


millsp (approved):
LGTM!

PR #19772, opened 2023-06-14
fix(client): don't create package.json automatically

### LINKED ISSUE #19306: Prisma Client generator: remove creation of a package.json
Code: https://github.com/prisma/prisma/blob/0fe45ca9e013e607fe5d3cf0852ea9aca341466c/packages/internals/src/get-generators/generatorResolvers/prisma-client-js/prismaClientResolver.ts#L39

I don't think it was documented anywhere.

Instead, we could output a warning message?

aqrln:
Don't try to create `package.json` before automatically installing missing Prisma packages. Instead, delegate this to the package manager (which will find the closest `package.json` file or create one automatically if none was found).

This fixes an issue where a stray `package.json` file may be incorrectly created in the wrong directory because we only check the current working directory and its parent, while the `package.json` may already exist more levels up.

However, since we're not removing automatic package installation yet, with these changes it's important to properly determine the directory relative to which it is safe to run `npm install`. Otherwise, when there was indeed no `package.json` file anywhere up the hierarchy, it (and the `node_modules` directory) would've been created near the schema file, which is not something that we usually want.

Before the changes it was often fine (but still not always correct) to run `npm install` in `baseDir`: it would work if we guessed with placing `package.json` in the current working directory, but could break and potentially lead to two `package.json` files being created in total when running the command from a nested directory (e.g. when schema is `project/prisma/schema.prisma` and cwd is `project/src` but `project/package.json` does not exist, which is quite an edge case).

Now we infer the potential project/package root as the longest common ancestor of both the schema path and the current working directory. That's where we run `npm install` from. If the actual root already exists above, it will be found by the package manager, otherwise a new `package.json` will be created in that directory. I also added a warning printed in that case, but I'm not sure if it's helpful or too prone to false positives, I can remove it if others think that's better.

If the current working directory and the schema are on different disks on Windows and we can't calculate the common path, we print another warning and fall back to `baseDir`.

Closes: https://github.com/prisma/prisma/issues/19306


Jolg42:
@aqrln let me know once it's ready for a review, I'm happy to give it a read

aqrln:
@Jolg42 I added a description, this should be ready for review.

packages/internals/src/get-generators/generatorResolvers/prisma-client-js/prismaClientResolver.ts
```diff
@@ -66,18 +68,18 @@ export async function prismaClientResolver(baseDir: string, version?: string) {
         `Could not resolve ${missingCli(`${bold('prisma')} and `)}${bold(
           '@prisma/client',
         )} in the current project. Please install ${hasCli('it')}${missingCli('them')} with ${missingCli(
-          `${bold(green(`${getPackageCmd(baseDir, 'add', 'prisma', '-D')}`))} and `,
-        )}${bold(green(`${getPackageCmd(baseDir, 'add', '@prisma/client')}`))}, and rerun ${bold(
+          `${bold(green(`${await getPackageCmd(baseDir, 'add', 'prisma', '-D')}`))} and `,
+        )}${bold(green(`${await getPackageCmd(baseDir, 'add', '@prisma/client')}`))}, and rerun ${bold(
```
aqrln, line 72: These two lines used to print `[object Promise]` due to missing `await`s. I can extracted this to a separate PR for Prisma 4.16.

janpio: I think we have that already from a user: https://github.com/prisma/prisma/pull/19259

aqrln: ah, nice, thanks!

aqrln: will do that tomorrow then

packages/internals/src/get-generators/generatorResolvers/prisma-client-js/prismaClientResolver.ts
```diff
@@ -30,27 +31,28 @@ export async function prismaClientResolver(baseDir: string, version?: string) {
...
+      )
+      projectRoot = baseDir
+    }
+
+    if (!fs.existsSync(path.join(projectRoot, 'package.json'))) {
+      console.warn(
+        yellow(
+          `${bold('Warning:')} could not find the ${bold(
+            'package.json',
+          )} file in the inferred project root, it will be created by your package manager on the appropriate level if necessary.`,
```
Jolg42, line 53:
```suggestion
          `${bold('Warning:')} Prisma could not find a ${bold(
            'package.json',
          )} file in the inferred project root. During the next step, when an auto-install of Prisma package(s) will be attempted, it will then be created by your package manager on the appropriate level if necessary.`,
```

packages/internals/src/get-generators/generatorResolvers/prisma-client-js/prismaClientResolver.ts
```diff
@@ -30,27 +31,28 @@ export async function prismaClientResolver(baseDir: string, version?: string) {
...
-      console.info(`✔ Created ${bold(green('./package.json'))}`)
+    let projectRoot = longestCommonPathPrefix(baseDir, process.cwd())
+    debug('projectRoot', projectRoot)
+
+    if (projectRoot === undefined) {
+      console.warn(
+        yellow(
+          `${bold('Warning:')} schema directory ${bold(baseDir)} and current working directory ${bold(
+            process.cwd(),
+          )} have no common ancestor, using the schema path as the project root.`,
```
Jolg42, line 42:
```suggestion
          `${bold('Warning:')} [Prisma auto-install on generate] The Prisma schema directory ${bold(baseDir)} and the current working directory ${bold(
            process.cwd(),
          )} have no common ancestor. The Prisma schema direcory will be used as the project root.`,
```

Jolg42 (commented):
@aqrln Did you try it manually?

aqrln:
> @aqrln Did you try it manually?

Yes

Jolg42 (approved):

packages/internals/src/get-generators/generatorResolvers/prisma-client-js/prismaClientResolver.ts
```diff
@@ -30,27 +31,30 @@ export async function prismaClientResolver(baseDir: string, version?: string) {
...
+    const warningTag = `${bold('Warning:')} ${dim('[Prisma auto-install on generate]')}`
+
+    if (projectRoot === undefined) {
+      console.warn(
+        yellow(
+          `${warningTag} The Prisma schema directory ${bold(baseDir)} and the current working directory ${bold(
+            process.cwd(),
+          )} have no common ancestor. The Prisma schema directory will be used as the project root.`,
+        ),
+      )
```
millsp, line 46: Should we perhaps show this only when there is a hard error?

millsp (approved):

PR #19787, opened 2023-06-15
feat(client):  Pass `Sql` instance to $queryRaw/$executeRaw extensions

### LINKED ISSUE #18125: PCE: Improve runtime types in raw query extensions

SevInf:
Query extensions for safe SQL raw methods will now always receive `Sql`
instaces.
`*Unsafe` variants are left unchanged and would still receive a tuple of
query + values.
Middlewares are unchanged and would always receive tuples.

To allow us support different arguments for middlewares and query
extensions `middlewareArgsMapper` is introduced as a part of internal
request parameters. If present, it's method would be exectued for
transforming args from internal representation (which query extensions
will be used) to middleware arguments and back.

Close #18125

### DIFF
--- a/packages/client/src/generation/TSClient/PrismaClient.ts
+++ b/packages/client/src/generation/TSClient/PrismaClient.ts
@@ -106,8 +106,8 @@ function clientTypeMapOthersDefinition(this: PrismaClientClass) {
   })
 
   const argsResultMap = {
-    $executeRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
-    $queryRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
+    $executeRaw: { args: 'Prisma.Sql', result: 'any' },
+    $queryRaw: { args: 'Prisma.Sql', result: 'any' },
     $executeRawUnsafe: { args: '[query: string, ...values: any[]]', result: 'any' },
     $queryRawUnsafe: { args: '[query: string, ...values: any[]]', result: 'any' },
     $runCommandRaw: { args: 'Prisma.InputJsonObject', result: 'Prisma.JsonObject' },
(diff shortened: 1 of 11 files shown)


packages/client/tests/functional/middleware-raw-args/tests.ts
```diff
@@ -0,0 +1,66 @@
+import { NewPrismaClient } from '../_utils/types'
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { Prisma as PrismaNamespace, PrismaClient } from './node_modules/@prisma/client'
+
+declare let newPrismaClient: NewPrismaClient<typeof PrismaClient>
+declare let Prisma: typeof PrismaNamespace
+
+testMatrix.setupTestSuite(
```
SevInf, line 9: Test was created on `main` to ensire we are not changing middlewares.

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -106,8 +106,8 @@ function clientTypeMapOthersDefinition(this: PrismaClientClass) {
   })
 
   const argsResultMap = {
-    $executeRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
-    $queryRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
```
Jolg42, line 110: So this is only a change impacting clientExtensions? The function name is a bit unclear

Jolg42 (approved):
See comment, looks good to me @millsp could take a quick look before merging, I think

packages/client/src/generation/TSClient/PrismaClient.ts
SevInf: Yes, that's for extensions-only

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -689,15 +717,17 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
       return this._request({
         action: 'queryRaw',
         args,
         transaction,
         clientMethod,
-        argsMapper: rawQueryArgsMapper(this, clientMethod),
+        argsMapper: rawQueryArgsMapper(this._activeProvider, clientMethod),
         callsite: getCallSite(this._errorFormat),
         dataPath: [],
+        middlewareArgsMapper,
```
millsp, line 730: Instead of having a new mapper, could the middleware logic re-wrap `argsMapper`?

millsp (approved):
Nice cleanup

PR #19812, opened 2023-06-16
fix(cli): don't crash if opening browser for "prisma studio" fails

### LINKED ISSUE #1128: `Error: spawn xdg-open ENOENT` on `prisma studio`
Some users are reporting that they get `Error: spawn xdg-open ENOENT` when running `prisma studio` (see parts of https://github.com/prisma/prisma/issues/19010).

A workaround is to tell Studio not to open a browser window:
```
  Start Studio without opening in a browser
    prisma studio --browser none
    BROWSER=none prisma studio
```

Background: Prisma Studio uses the npm package `open` to open a browser window with Studio when you run the command by default (https://github.com/prisma/prisma/blob/c7e97a5c39584aa24f2b15a5c4d4d6ecbab90a95/packages/cli/src/Studio.ts#L114-L117). It could be that a recent update in some operating system might have removed the default tool that is used for that.

What docker image or operating system are you getting this error?

aqrln:
Unless `wait: true` is passed to `open`, it does not install an `error`
event listener to the child process, so it's on us to handle it.

Before https://github.com/prisma/prisma/pull/19385 all unhandled errors that
happen in CLI were silently suppressed, so this issue was hidden. This PR
fixes it.

The added e2e test fails with `Error: spawn xdg-open ENOENT` without
the changes in `Studio.ts`, and passes with them.

Adding special debug logs that the test depends on was necessary since
the error happens asynchronously later, so we can't depend on the
"Prisma Studio is up on http://localhost:5555" message, and we need to
depend on some output that's printed after the point where the ENOENT
error may have happened to determine we can send SIGINT to the child
process so that the test doesn't hang infinitely.

An alternative would've been to rely on a timeout to terminate the
process, but that would be unreliable/flaky.

Fixes: https://github.com/prisma/studio/issues/1128

### DIFF
--- a/packages/cli/src/Studio.ts
+++ b/packages/cli/src/Studio.ts
@@ -1,3 +1,4 @@
+import Debug from '@prisma/debug'
 import { enginesVersion } from '@prisma/engines'
 import { arg, checkUnsupportedDataProxy, Command, format, HelpError, isError, loadEnvFile } from '@prisma/internals'
 import { getSchemaPathAndPrint } from '@prisma/migrate'
@@ -6,6 +7,8 @@ import getPort from 'get-port'
 import { bold, dim, red } from 'kleur/colors'
 import open from 'open'
 import path from 'path'
+
+const debug = Debug('prisma:studio')
 
 const packageJson = require('../package.json') // eslint-disable-line @typescript-eslint/no-var-requires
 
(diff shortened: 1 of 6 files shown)


Jolg42 (approved):
Nice test!

janpio:
Use either `prisma@dev` or wait until tomorrow, when this will be included in `prisma@4.16.0`.

FreshImmuc:
Thank you this fixed it!

PR #19815, opened 2023-06-16
feat(get-platform): add support for new linux-static binary targets

aqrln:
Add support for `linux-static-x64` and `linux-static-arm64` binary targets.

A few gotchas:
* You need `engineType = "binary"` or `PRISMA_CLIENT_ENGINE_TYPE=binary` to be able to generate client with these added to `binaryTargets`, otherwise you'll get a 404 trying to download a non-existing library.
* There's no way to make the client or CLI use them without manually setting the env vars, we don't currently have a way to override the detected platform.


packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -180,211 +203,227 @@ It took ${timeInMsToDownloadAll}ms to execute download() for all binaryTargets.`
...
+            "name": "query-engine-linux-static-arm64",
+            "size": 15984368,
+          },
+          {
+            "name": "query-engine-linux-static-x64",
+            "size": 19274912,
+          },
+          {
+            "name": "query-engine-rhel-openssl-1.0.x",
+            "size": 19287104,
```
janpio, line 401: Wow, that is quite an increase. What happened there?

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -180,211 +203,227 @@ It took ${timeInMsToDownloadAll}ms to execute download() for all binaryTargets.`
...
+        ]
+      `)
 
       expect(await getVersion(queryEnginePath, BinaryType.QueryEngineBinary)).toMatchInlineSnapshot(
-        `"query-engine eac182fd33c63959a61946df56831625a9a39627"`,
+        `"query-engine b20ead4d3ab9e78ac112966e242ded703f4a052c"`,
       )
       expect(await getVersion(migrationEnginePath, BinaryType.MigrationEngineBinary)).toMatchInlineSnapshot(
-        `"migration-engine-cli eac182fd33c63959a61946df56831625a9a39627"`,
+        `"schema-engine-cli b20ead4d3ab9e78ac112966e242ded703f4a052c"`,
```
janpio, line 426: Isn't that a P5 change?

packages/engines/src/index.ts
```diff
@@ -62,6 +62,8 @@ path.join(__dirname, '../query-engine-darwin-arm64')
 path.join(__dirname, '../query-engine-debian-openssl-1.0.x')
 path.join(__dirname, '../query-engine-debian-openssl-1.1.x')
 path.join(__dirname, '../query-engine-debian-openssl-3.0.x')
+path.join(__dirname, '../query-engine-linux-static-x64')
+path.join(__dirname, '../query-engine-linux-static-arm64')
 path.join(__dirname, '../query-engine-rhel-openssl-1.0.x')
 path.join(__dirname, '../query-engine-rhel-openssl-1.1.x')
 path.join(__dirname, '../query-engine-rhel-openssl-3.0.x')
```
janpio, line 69: Why is there no binary equivalent to lines 77-81 + 85?

```
path.join(__dirname, '../libquery_engine-linux-arm64-openssl-1.0.x.so.node')
path.join(__dirname, '../libquery_engine-linux-arm64-openssl-1.1.x.so.node')
path.join(__dirname, '../libquery_engine-linux-arm64-openssl-3.0.x.so.node')
path.join(__dirname, '../libquery_engine-linux-musl.so.node')
path.join(__dirname, '../libquery_engine-linux-musl-openssl-3.0.x.so.node')
...
path.join(__dirname, '../query_engine-windows.dll.node')

```

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -180,211 +203,227 @@ It took ${timeInMsToDownloadAll}ms to execute download() for all binaryTargets.`
...
+            "size": 15984368,
+          },
+          {
+            "name": "query-engine-linux-static-x64",
+            "size": 19274912,
+          },
+          {
+            "name": "query-engine-rhel-openssl-1.0.x",
+            "size": 19287104,
+          },
```
janpio, line 402: Wow, that is quite an increase. What happened there?

packages/engines/src/index.ts
aqrln: Yep, I noticed that too and was going to open a separate PR.

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -180,211 +203,227 @@ It took ${timeInMsToDownloadAll}ms to execute download() for all binaryTargets.`
...
+            "name": "query-engine-linux-musl-arm64-openssl-3.0.x",
+            "size": 19120392,
+          },
+          {
+            "name": "query-engine-linux-musl-openssl-3.0.x",
+            "size": 16775312,
+          },
+          {
+            "name": "query-engine-linux-static-arm64",
+            "size": 15984368,
```
janpio, line 393: Surprisingly small, isn't it?

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -180,211 +203,227 @@ It took ${timeInMsToDownloadAll}ms to execute download() for all binaryTargets.`
...
+            "name": "query-engine-darwin",
+            "size": 17840936,
+          },
+          {
+            "name": "query-engine-darwin-arm64",
+            "size": 16655984,
+          },
+          {
+            "name": "query-engine-debian-openssl-1.0.x",
+            "size": 19287104,
```
janpio, line 353: Another surprising increase
```diff
- "size": 17525064,
+ "size": 19287104,

packages/engines/src/index.ts
aqrln: #19816

packages/fetch-engine/src/__tests__/download.test.ts
aqrln: This specifically was already changed on the engine side in April (https://github.com/prisma/prisma-engines/pull/3844). We actually have a hack in `prisma version` command that replaces `schema-engine-cli` with `migration-engine-cli` in the output. But here it's the raw command output.

packages/fetch-engine/src/__tests__/download.test.ts
aqrln: Yes actually, I was wondering for a long time already why all our ARM binaries so small given that we always vendor OpenSSL there. Intuitively I assumed ARM64 machine code was less compact than x86-64, but seems like the opposite is the case.

packages/fetch-engine/src/__tests__/download.test.ts
aqrln: https://github.com/prisma/prisma-engines/pull/3932.

Basically we've been statically linking more modern OpenSSL in 1.0.x engine variants since https://github.com/prisma/prisma-engines/pull/1770, but at some point it broke for QE binary and it has been using system OpenSSL while migration engine and QE library have been using their own one. Then the non-functional cargo feature was fixed, and we implemented https://github.com/prisma/ecosystem-tests/pull/3500 to prevent things like this happening without us knowing.

We discussed this on a meeting with you, Tom and Pierre a few months ago.

Jolg42 (approved):

packages/fetch-engine/src/__tests__/download.test.ts
janpio: Ok, so we were testing an version older than the one from April until now? Sounds suboptimal. Can't we use the engine hash of the current version in this test?

packages/fetch-engine/src/__tests__/download.test.ts
janpio: Understood.

packages/fetch-engine/src/__tests__/download.test.ts
aqrln: Changed `FIXED_ENGINE_HASH` to `CURRENT_ENGINE_HASH` in all tests except those that snapshot engine size.

janpio: And now, 13 minutes later I understand that this will lead to small size changes each commit which will make the snapshots fail I think. Sorry @aqrln.

aqrln: No no, all good, there's no problem. The tests that snapshot engine sizes still use the fixed commit hash that needs to be updated manually. I only changed the rest of the tests (including this one).

aqrln: ah not, wait, I'm wrong, I thought it was a separate test but it's a part of the test that snapshots sizes :/

aqrln: somehow I also didn't commit all changes in the first place

aqrln: also the commit hash would've been changing here and failing snapshots, so it wouldn't work even if it was a separate test which I also failed to realise — probably time to wrap up for today 😄

PR #19816, opened 2023-06-16
fix(engines): add missing node-file-trace annotations

aqrln:
Add missing NFT annotations for engine files.


packages/engines/src/index.ts
```diff
@@ -62,11 +62,19 @@ path.join(__dirname, '../query-engine-darwin-arm64')
 path.join(__dirname, '../query-engine-debian-openssl-1.0.x')
 path.join(__dirname, '../query-engine-debian-openssl-1.1.x')
 path.join(__dirname, '../query-engine-debian-openssl-3.0.x')
+path.join(__dirname, '../query-engine-linux-arm64-openssl-1.0.x')
```
millsp, line 65: Do we also have these builds for the schema engine? If so, should we add them to the CLI annotations we have?

aqrln: I don't think annotations for schema engine need to be added here because it is not supposed to be deployed. But I'm also not familiar with the tooling that requires these annotations, perhaps @Jolg42 can answer better.

aqrln: Actually if my understanding is correct that only the files that need to be deployed on Vercel need to be annotated, then perhaps we need to, on the contrary, trim this list down instead of expanding it? (i.e. remove darwin and unnecessary linux builds)

jkomyno:
Is there anything else that needs to happen here? @aqrln

aqrln:
@jkomyno I think it's mergeable, it's only waiting for an approval

wmadden:
Closing this for now @aqrln, feel free to reopen

PR #19837, opened 2023-06-19
refactor(client): extension dynamic types

### LINKED ISSUE #17843: VScode behavior is very slow while using clientExtensions
### Bug description

_Today I updated prisma to 4.10.0 and immediately experienced very slow VScode behavior in project where prisma is installed._

### **EDIT:**

After few hours of playing with VScode extensions, code, schema, ... I ended up with new observings.

Seems like the version is **not** causing the issue. Let me explain:

Before updating to version 4.10.0 I prepared this code for extending the prisma client:
```
.$extends({
  result: {
    iOrder: {
      orderNumber: {
        needs: { id: true, randomNumber: true },
        compute(order) {
          return +`${order.id}${order.randomNumber}`;
        }
      }
    }
  },
  model: {
    iOrder: {
      byOrderNumber: (vs: number) => {
        const str = `${vs}`;
        const split = [parseInt(str.slice(0, str.length-2)), parseInt(str.slice(str.length-2))];
        return {
          id: split[0],
          randomNumber: split[1]
        };
      }
    } satisfies Record<string, (...args: any) => Prisma.IOrderWhereInput>,
  }
});
```

Right after updating the version I extended the client with this code and added `clientExtensions` to `previewFeatures` and ran `npx prisma generate`. After that the VScode window was **very** slow (intellisense, error higlighting, ...).
For example after typing `(new PrismaClient()).` the intellisense would show up in +- 13 seconds. Because of that I wrongly blamed the version - sorry about that.

Now what I discovered:

1. I removed the extending code (see above) and the intellisense show time dropped to +- 3 seconds.
2. I removed `clientExtensions` from `previewFeatures` and the show time was almost instant (as it was). 

So it seems like that the `clientExtensions` preview feature is causing this problem.

Sorry about not posting our schema as we find it sensitive. But there are 24 models.

Btw [this](https://pcsupport.lenovo.com/us/en/products/LAPTOPS-AND-NETBOOKS/THINKPAD-T-SERIES-LAPTOPS/THINKPAD-T14-TYPE-20UD-20UE/20UE/20UES2WA00) is my machine.

### **EDIT2:**
This slow behavior is not immediate. It shows up after few minutes of working. And it seems, like `prismaTransactionClient.` takes longer to show intellisense than `prismaClient.`

### How to reproduce

1. use `clientExtensions` preview feature
2. extend the client
3. experience the slowness of VScode - for example intellisense

### Expected behavior

VScode should be quick with and without using `clientExtensions`.

### Environment & setup

- OS: Windows 11
- Database: PostgreSQL
- Node.js version: 16.15.1

### Prisma Version

```
prisma                  : 4.10.0
@prisma/client          : 4.10.0
Query Engine (Node-API) : libquery-engine ca7fcef713137fa11029d519a9780db130cca91d (at node_modules\@prisma\engines\query_engine-windows.dll.node)
Migration Engine        : migration-engine-cli ca7fcef713137fa11029d519a9780db130cca91d (at node_modules\@prisma\engines\migration-engine-windows.exe)
Format Wasm             : @prisma/prisma-fmt-wasm 4.10.0-84.ca7fcef713137fa11029d519a9780db130cca91d
Default Engines Hash    : ca7fcef713137fa11029d519a9780db130cca91d
```

([slack thread](https://prisma-company.slack.com/archives/C02PJ7JGM6K/p1676750364663279?thread_ts=1676709798.085849&cid=C02PJ7JGM6K))

millsp:
closes https://github.com/prisma/prisma/issues/17843
closes https://github.com/prisma/team-orm/issues/193

This PR mainly refactors `clientExtensions` to discard type generation in favor of full dynamic types, and opportunistically implements top-level operations. It turns out that the reported performance drops severity was proportional to the amount of models (see linked issue). `result` and `query` extensions were the biggest culprits as they previously generated quite a lot of generated types - types which are then evaluated by TypeScript. Adding to this, the complexity of `query` and its explosive combinatory explosion of unions for `$allModels` and `$allOperations` caused severe slowdowns. So users with many models would be doomed not to benefit from extensions (for this refactor I tested with a real-world 120 models schema). To add to that, I found that monkey patching type "instances" of `PrismaClient` or models (any class really) puts unnecessary strain on the compiler's memory due to the way `this` is tracked by the compiler, degrading the auto-completion experience. Similarly this is also what caused `const xprisma: PrismaClient = new PrismaClient().$extends()` to cause excessive depth comparison errors instead of a simple assignability error.

Given this, I completely changed the approach. We are not generating types in the `index.d.ts` any longer but it is happening dynamically and on demand. In other words, types will be dynamically calculated as you type/autocomplete within the `$extends` configuration, nothing more nothing less. That also means inference is not a vector of slow down anymore, and in turn, this prevents previous unnecessary computations or evaluation for all the existing models. In order to achieve this, this PR uses a lot of advanced TS techniques as well as some novel approaches. The extended clients are calculated on the spot thanks to a `TypeMap` which represents all information a `PrismaClient` needs. As counter-intuitive as it may seem, monkey patching is far less efficient than constructing the client properties via mapped types. This has also drastically reduced memory consumption for types.

When I first discovered the problem, I was worried that we would need to change the API in some breaking way, but luckily the tests are left mostly the same and nothing has fundamentally changed on the surface.

PS: Type performance seems to be a thing, so we will want to test it in the future.


packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -106,8 +48,8 @@ function clientTypeMapOthersDefinition(this: PrismaClientClass) {
   })
 
   const argsResultMap = {
-    $executeRaw: { args: 'Prisma.Sql', result: 'any' },
-    $queryRaw: { args: 'Prisma.Sql', result: 'any' },
+    $executeRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
+    $queryRaw: { args: '[query: TemplateStringsArray | Prisma.Sql, ...values: any[]]', result: 'any' },
```
Jolg42, line 52: This looks like a revert of what was done in https://github.com/prisma/prisma/pull/19787
Is that expected?

packages/client/src/runtime/core/extensions/$extends.ts
```diff
@@ -61,8 +61,8 @@ type QueryOptionsCbArgs = {
 type ModelQueryOptionsCbArgs = {
   model: string
   operation: string
-  args: JsArgs
-  query: (args: JsArgs) => Promise<unknown>
+  args: unknown
```
SevInf, line 64: Why this change? `JsArgs` properly types all possible user input values.

packages/client/src/runtime/core/types/Extensions.ts
```diff
@@ -21,30 +25,178 @@ export type Args = InternalArgs
...
+
+type DynamicResultExtensionNeeds<TypeMap extends Record<string, any>, M extends keyof any, S> = {
+  [K in keyof S]: K extends keyof TypeMap['model'][M]['findFirstOrThrow']['payload']['scalars'] ? S[K] : never
+} & {
+  [N in keyof TypeMap['model'][M]['findFirstOrThrow']['payload']['scalars']]?: boolean
+}
 
-export type GetClient<Base extends Record<any, any>, C extends Args['client'], _C extends Args['model'][string] = Record<string, any> extends C ? {} : C> =
-  { [K in keyof _C | Exclude<keyof Base, '$use' | '$on'>]: K extends keyof _C ? ReturnType<_C[K]> : Base[K] }
+type DynamicResultExtensionData<TypeMap extends Record<string, any>, M extends keyof any, S> =
```
SevInf, line 99:
```suggestion
type DynamicResultExtensionData<TypeMap extends Record<string, any>, M extends PropertyKey, S> =
```

packages/client/src/runtime/core/types/Extensions.ts
```diff
@@ -21,30 +25,178 @@ export type Args = InternalArgs
...
   { [K in keyof _R | keyof Base]?: K extends keyof _R ? boolean : Base[K] }
 
-export type GetModel<Base extends Record<any, any>, M extends Args['model'][string], _M extends Args['model'][string] = Record<string, any> extends M ? {} : M> =
-  { [K in keyof _M | keyof Base]: K extends keyof _M ? ReturnType<_M[K]> : Base[K] }
+/** Query */
+
+export type DynamicQueryExtensionArgs<Q_, TypeMap extends Record<string, any>> = {
+  [K in keyof Q_]:
+    K extends '$allOperations'
+    ? (args: { model?: string, operation: string, args: any, query: (args: any) => PrismaPromise<any> }) => Promise<any>
```
SevInf, line 38: So, for top-level `$allOperations` you are not getting inference on model/args, am I getting this right? Should we?

packages/client/src/runtime/core/types/Extensions.ts
```diff
@@ -21,30 +25,178 @@ export type Args = InternalArgs
...
+  [P in keyof TypeMap['other'] | keyof TypeMap['model'] | keyof ExtArgs['client'] as Uncapitalize<P & string>]:
+    P extends Operation
+    ? <A extends TypeMap['other'][P]['args']>(...args: A extends any[] ? A : [A]) =>
+      PrismaPromise<GetOperationResult<TypeMap['other'][P]['payload'], A, P>>
+    : P extends keyof TypeMap['model']
+      ? DynamicModelExtensionThis<TypeMap, P, ExtArgs>
+      : Return<ExtArgs['client'][P]>
+} & {
+  $extends: ExtendsHook<'extends', TypeMapCb, ExtArgs>
+  $transaction<R>(fn: (client: Omit<DynamicClientExtensionThis<TypeMap, TypeMapCb, ExtArgs>, ITXClientDenyList>, options?: { maxWait?: number, timeout?: number, isolationLevel?: string }) => Promise<R>): Promise<R>
```
SevInf, line 146: This loses strong typing for `isolationLevel` on extended client and makes it available where it should not (mongo).
I guess we should put `isolationLevel` into the type map as well?

packages/client/tests/functional/extensions/itx.ts
```diff
@@ -282,7 +282,7 @@ testMatrix.setupTestSuite(({ provider }, _, clientMeta) => {
 
           expectTypeOf(ctx.$connect).toEqualTypeOf<typeof prisma.$connect | undefined>()
           expectTypeOf(ctx.$disconnect).toEqualTypeOf<typeof prisma.$disconnect | undefined>()
-          expectTypeOf(ctx.$transaction).toEqualTypeOf<typeof prisma.$transaction | undefined>()
+          expectTypeOf(ctx.$transaction).toMatchTypeOf<Function | undefined>()
```
SevInf, line 285: Ideally, we'd not lose typing here

SevInf:
I think this used to narrow down correctly at some point?

image

packages/client/src/generation/TSClient/PrismaClient.ts
millsp: Yes, I needed to restore it so I can type regular calls as well. The `query` extensions have a spacial case instead.

packages/client/src/runtime/core/types/Extensions.ts
millsp: I'll update it.

packages/client/src/runtime/core/types/Extensions.ts
millsp: Good catch, thanks.

packages/client/src/runtime/core/extensions/$extends.ts
millsp: Yes, there is no reason to change this, just me testing something.

millsp:
> I think this used to narrow down correctly at some point?

Yes, I think we're missing a test. I've distributed the union independently instead of all together. Will fix that.

packages/client/tests/functional/extensions/itx.ts
millsp: The transaction types will not match anymore although correct (and equivalent) since we are not monkey patching the PrismaClient anymore. So besides adding a full tests to ensure $transaction works correctly, I don't know how it can be tested.

SevInf (approved):
Approving so it will be mergeable after you done with fixes

packages/client/src/runtime/core/types/Extensions.ts
millsp: So in some cases the union can get pretty crazy and I am not sure if this would be useful in terms of performance. I am not sure why the compiler is able to handle `$.allModels.$allOperations` though. I think it has to do with the share of args being a record always and now we'd be mixing arrays and classes potentially, so the intersection of that on access of a property can be pretty wild - well, that's my guess. So I am leaving it as is for now.

PR #19840, opened 2023-06-19
chore: remove deprecated warning part mentioning `node:lts-slim` Docker image

jkomyno:
Note: the `node:lts-slim` Docker image no longer ships with OpenSSL as of the latest Debian version.
Thus, we have removed this part of a warning that occurs when `@prisma/get-platform` failed to resolve the Prisma compilation target to download on a system featuring a Linux distro of the Debian family:

```
If you're running Prisma on Docker, you may also try to replace your base image with `node:lts-slim`, which already ships with OpenSSL installed.
```

We originally believed we had mentions of popular Docker images in Prisma docs, but that's not the case.

### DIFF
--- a/packages/get-platform/src/__tests__/getPlatform.test.ts
+++ b/packages/get-platform/src/__tests__/getPlatform.test.ts
@@ -128,7 +128,7 @@ describe('getPlatformInternal', () => {
       expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
       expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`
         prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".
-        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again. If you're running Prisma on Docker, you may also try to replace your base image with \`node:lts-slim\`, which already ships with OpenSSL installed.
+        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again.
       `)
       expect(ctx.mocked['console.error'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
     })
--- a/packages/get-platform/src/__tests__/getPlatform.test.ts
+++ b/packages/get-platform/src/__tests__/getPlatform.test.ts
@@ -128,7 +128,7 @@ describe('getPlatformInternal', () => {
       expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
       expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`
         prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".
-        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again. If you're running Prisma on Docker, you could also switch to an image that has OpenSSL already installed, such as \`node:lts-bullseye-slim\`.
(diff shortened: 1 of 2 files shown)


packages/get-platform/src/__tests__/getPlatform.test.ts
```diff
@@ -128,7 +128,7 @@ describe('getPlatformInternal', () => {
       expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
       expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`
         prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".
-        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again. If you're running Prisma on Docker, you may also try to replace your base image with \`node:lts-slim\`, which already ships with OpenSSL installed.
+        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again.
```
janpio, line 131:
```suggestion
        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again. If you're running Prisma on Docker, switch to an image that has OpenSSL installed.
```
Maybe?

jkomyno: Alright, I could also say "... that has OpenSSL installed, like `node:lts-bullseye-slim`"

janpio: I fear that will lead people to try the `bookworm` one though (as they always want to be on the newest) and then complain again :D Let's see.

I generally still like the idea better to not have a concrete name in the error message, but instead link to a docs page that can more easily be updated and provide some explanation.

jkomyno: Ok, final proposal:

- Add `If you're running Prisma on Docker, add this command to your Dockerfile, or switch to an image that already has OpenSSL installed.` to the message
- Revisit that old idea we had of adding a small Docker + Prisma tutorial to Prisma docs, and once we have that, add a pris.ly link in this warning

packages/get-platform/src/__tests__/getPlatform.test.ts
```diff
@@ -128,7 +128,7 @@ describe('getPlatformInternal', () => {
       expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
       expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`
         prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".
-        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again. If you're running Prisma on Docker, you may also try to replace your base image with \`node:lts-slim\`, which already ships with OpenSSL installed.
+        Please manually install OpenSSL via \`apt-get update -y && apt-get install -y openssl\` and try installing Prisma again. If you're running Prisma on Docker, add this command to your Dockerfile, or switch to an image that already has OpenSSL installed.
```
Jolg42, line 131: >add this command to your Dockerfile,
Where is the command 🤔 ?

jkomyno: `apt-get update -y && apt-get install -y openssl` would be the command. It's shown in the immediately precedent paragraph

Jolg42: ... Not sure how I didn't see it 🙈

Jolg42 (approved):

packages/get-platform/src/__tests__/getPlatform.test.ts
janpio: (I also had to reread. Hope users are smarter than us)

aqrln (approved):

PR #19845, opened 2023-06-19
fix(client): Fix composites selection

SevInf:
With extensions GA we've fixed default composites selections, but
explict ones are still broken.
This PR adjusts `GetResult` to pick selection fields from both
`composites` and `objects` fields of the `Payload`, ensures that
`Payload` types for composites are used and correctly picked up with
both explicit and default selections.

### DIFF
--- a/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
@@ -341,17 +341,17 @@ import $Extensions = runtime.Types.Extensions
 export type PrismaPromise<T> = $Public.PrismaPromise<T>
 
 
-export type EmbedPayload<ExtArgs extends $Extensions.Args = $Extensions.DefaultArgs> = {
+export type EmbedPayload = {
   objects: {}
-  scalars: $Extensions.GetResult<{
+  scalars: {
     text: string
     boolean: boolean
     scalarList: number[]
-  }, ExtArgs["result"]["embed"]>
+  }
   composites: {
-    embedEmbedList: EmbedEmbed[]
-    requiredEmbedEmbed: EmbedEmbed
-    optionalEmbedEmbed: EmbedEmbed | null
+    embedEmbedList: EmbedEmbedPayload[]
+    requiredEmbedEmbed: EmbedEmbedPayload
+    optionalEmbedEmbed: EmbedEmbedPayload | null
   }
 }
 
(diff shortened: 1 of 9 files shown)


packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
```diff
@@ -341,39 +341,39 @@ import $Extensions = runtime.Types.Extensions
 export type PrismaPromise<T> = $Public.PrismaPromise<T>
 
 
-export type EmbedPayload<ExtArgs extends $Extensions.Args = $Extensions.DefaultArgs> = {
```
millsp, line 344: I think this was added so we can later support composite result extensions.

SevInf: We could always add them back once we get to it - it's a matter of removing 2 if statements in generator. I just felt that since we are not using them at the moment it does not make sense to generate `ExtArgs` for composites.

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -44,28 +45,49 @@ type GetFindResult<P extends Payload, A> =
...
               ? O
               : K extends '_count'
                 ? Count<P['objects']>
                 : never
-    } & (A extends { include: any } & Record<string, unknown> ? P['scalars'] & P['composites'] : unknown)
-  : P['scalars'] & P['composites']
+    } & (A extends { include: any } & Record<string, unknown> ? DefaultSelection<P> : unknown)
+  : DefaultSelection<P>
+
+type SelectablePayloadFields<K extends PropertyKey, O> = { objects: { [k in K]: O } } | { composites: { [k in K]: O } }
```
millsp, line 67:
```suggestion
type SelectablePayloadFields<K extends PropertyKey, O> =
  | { objects: { [k in K]: O } }
  | { composites: { [k in K]: O } }
```

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -44,28 +45,49 @@ type GetFindResult<P extends Payload, A> =
...
+
+// prettier-ignore
+type SelectField<P extends SelectablePayloadFields<any, any>, K extends PropertyKey> = 
+    P extends { objects: Record<K, any> } 
+    ? P['objects'][K]
+    : P extends { composites: Record<K, any> }
+      ? P['composites'][K]
+      : never
```
millsp, line 75:
```suggestion
  P extends { objects: Record<K, any> } 
  ? P['objects'][K]
  : P extends { composites: Record<K, any> }
    ? P['composites'][K]
    : never
```

millsp (approved):
Nice!

millsp:
:fishsticks:

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/test.ts.snap
millsp: Could you please create a placeholder or issue in our private board? (we've got quite a few of these things we want to add later)

PR #19874, opened 2023-06-21
chore(readme): Add links to Discord server

jharrell:
Added links to Discord alongside Slack


janpio (approved):

Jolg42:
@jharrell This looks good!

I think we could add the link in more places, see where Slack is also mentioned https://github.com/search?q=repo%3Aprisma%2Fprisma%20slack.prisma.io&type=code

And question, should we use a shortlink instead? Like `http://pris.ly/discord` or one like Slack but as a redirect (pris.ly would be easier I think).
Asking because this invite link might break one day?

jharrell:
@Jolg42 switching to use `pris.ly/discord` and also adding to more locations

Jolg42 (approved):

PR #19876, opened 2023-06-21
fix(client): extensions $transaction result inference

### LINKED ISSUE #19867: 4.16.0 Count query is not returning the right type when in a transaction
### Bug description

When i use a transaction with an array arg, it doesnt return the correct type. It seems to be mixing up the types from both items in the transaction

### How to reproduce

Use prisma 4.16.0. This doesn't occur on earlier versions:
image

### Expected behavior

Returned type for total is `number`

### Prisma information

See above

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: 18.12.0

### Prisma Version

```
4.16.0
```

millsp:
closes https://github.com/prisma/prisma/issues/19867

### DIFF
--- a/packages/client/tests/functional/extensions/tx.ts
+++ b/packages/client/tests/functional/extensions/tx.ts
@@ -191,7 +191,22 @@ testMatrix.setupTestSuite((_0, _1, clientMeta) => {
         isolationLevel: 'Serializable',
       })
 
-      expectTypeOf(data).toEqualTypeOf<({ id: string } | null)[]>()
+      expectTypeOf(data).toEqualTypeOf<[{ id: string } | null]>()
+    }
+  })
+
+  test('type inference allows for destructuring the array', () => {
+    ;async () => {
+      const xprisma = prisma.$extends({})
+
+      // @ts-test-if: provider !== 'mongodb'
+      const [data, count] = await xprisma.$transaction([
+        xprisma.user.findFirst({ select: { id: true } }),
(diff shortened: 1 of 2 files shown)


SevInf (approved):

packages/client/tests/functional/extensions/tx.ts
```diff
@@ -191,7 +191,21 @@ testMatrix.setupTestSuite((_0, _1, clientMeta) => {
         isolationLevel: 'Serializable',
       })
 
-      expectTypeOf(data).toEqualTypeOf<({ id: string } | null)[]>()
+      expectTypeOf(data).toEqualTypeOf<[{ id: string } | null]>()
+    }
+  })
+
+  test('type inference allows for destructuring the array', () => {
+    ;async () => {
```
Jolg42, line 199: nit: I don't understand why this test needs `;async () => {` like this. We usually do it like this
```
test('description', async () => {
```

Jolg42 (approved):

Jolg42:
Note: out in `4.16.1-dev.2` and up

packages/client/tests/functional/extensions/tx.ts
millsp: Yep, so this is because we don't really care to run it when it's type-level only

millsp: But it is still convenient to write it as a jest test even if it does not matter

PR #19899, opened 2023-06-22
test(client): extension with generic and non-generic model

### LINKED ISSUE #17458: Client extensions incorrect typings when defined both specific model and all models methods
### Bug description

Prisma client produces a compile error when using `$allModels` method if I defined a specific model method too. It seems that a model which has specific model methods cannot use `$allModels` methods.

### How to reproduce

https://github.com/tekihei2317/prisma-pagination-example/blob/main/scripts/query.ts#L12

Here’s repro repo.

I defined a `signUp` method to the user model, and a `paginate` method to all models. `xprisma.post.paginate` works well but `xprisma.user.paginate` produces a compile error that the `paginate` property does not exist.

### Expected behavior

Models have both specific model methods and `$allModels` methods. In the previous example, `xprisma.user` has both `signUp` and `paginate`.

### Prisma information

```prisma
model User {
  email String  @unique
  name  String?
  posts Post[]
}

model Post {
  title     String
  content   String?
  author    User     @relation(fields: [authorId], references: [id])
  authorId  Int
}
```

```ts
import { Prisma, PrismaClient } from "@prisma/client";

export const xprisma = prisma.$extends({
  model: {
    user: {
      async signUp(email: string) {
        await prisma.user.create({ data: { email } });
      },
    },
    $allModels: {
      async paginate<T, A>(
        this: T,
        args: Prisma.Exact<A, Prisma.Args<T, "findMany">> & {
          page: number;
          perPage: number;
        }
      ): Promise<Prisma.Result<T, A, "findMany">> {
        const { page, perPage } = args;

        return await (this as any).findMany({
          ...omit(args, "page", "perPage"),
          skip: perPage * (page - 1),
          take: perPage,
        });
      },
    },
  },
});

// ok
const posts = await xprisma.post.paginate({
  where: { authorId: 1 },
  orderBy: { createdAt: "desc" },
  page: 1,
  perPage: 5,
});

// compile error
// Property 'paginate' does not exist on type 'GetModel<UserDelegate<RejectOnNotFound | RejectPerOperation | undefined, { result: {} & Record<string, {}>; model: {} & Record<string, { paginate: () => <T, A>(this: T, args: Exact<...> & { ...; }) => Promise<...>; }> & { ...; }; client: {}; query: {}; }>, { ...; }>'.ts(2339)
const users = await xprisma.user.paginate({
  page: 1,
  perPage: 5,
});
```

### Environment & setup

- OS: macOS
- Database: SQLite
- Node.js version: 16.18.0

### Prisma Version

```
4.9.0
```

millsp:
closes https://github.com/prisma/prisma/issues/17458

### DIFF
--- a/packages/client/tests/functional/extensions/model.ts
+++ b/packages/client/tests/functional/extensions/model.ts
@@ -741,6 +741,45 @@ testMatrix.setupTestSuite(
       expectTypeOf(ctx).toHaveProperty('myCustomCallB').toEqualTypeOf<() => void>()
       expectTypeOf(ctx).toHaveProperty('update').toMatchTypeOf<Function>()
     })
+
+    test('one specific user extension along a generic $allModels model extension', () => {
+      const myCustomCallA = jest.fn()
+      const myCustomCallB = jest.fn()
+
+      const xprisma = prisma.$extends({
+        model: {
+          user: {
+            myCustomCallB(input: string) {
+              myCustomCallB(input)
+              return input
+            },
+          },
+          $allModels: {
+            myCustomCallA(input: number) {
+              myCustomCallA(input)
+              return input
+            },
+          },
+        },
+      })
+
+      const results = [
+        xprisma.user.myCustomCallA(42),
+        xprisma.user.myCustomCallA(42),
+        xprisma.user.myCustomCallB('Hello'),
+      ] as const
+
+      // @ts-expect-error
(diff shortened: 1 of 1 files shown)


SevInf (approved):

Jolg42 (approved):

Jolg42:
Note: this is a just a test.

PR #19907, opened 2023-06-22
fix(validatePrismaClientOptions): Add space to error message

janpio:


aqrln (approved):
LGTM, `getDidYouMean` also returns a string that starts with a space.

PR #19924, opened 2023-06-23
fix(client): On Windows, when using binary engine, do not run chmod

Jolg42:
Internal Slack
https://prisma-company.slack.com/archives/CRGDU19K6/p1687269464620359


janpio:
Can we do a test in this repo for this somehow?

Jolg42:
🤔 Didn't think about that yet
I think we better check inside the `plusX` function and exclude Windows there, so this doesn't happen by accident again

Jolg42:
Test is now passing: 
See prisma@4.17.0-integration-chmod-windows.2
https://github.com/prisma/ecosystem-tests/actions/runs/5359213266/jobs/9722561816

aqrln (approved):

jkomyno (approved):
Nit for the future: we could be more explicit about what `plusX` does by renaming it into something like `chmodPlusX`. This way, it'd be immediately apparent that it doesn't make much sense to call it on Windows.

Or, we could avoid the naming problem by adding a no-op within `plusX` whenever `process.platform === 'win32'`, to avoid trying to call `chmod` on platforms where it doesn't exist. I think the first solution would be cleaner, though.

PR #19942, opened 2023-06-24
fix(client): extensions with fluent api

### LINKED ISSUE #19921: Prisma Client fluent API does not work with extends anymore on 4.16.1
### Bug description

can't use [fluent api](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#fluent-api)

since this last 4.16.1 version

```
Property '[REDACTED]' does not exist on type 'PrismaPromise<GetResult<{ ....}
```

### How to reproduce

try to use fluent api and latest version of Prisma

### Expected behavior

_No response_

### Prisma information

```ts
prisma.MODEL.findUnique().FLUENT_RELATION()
```

### Environment & setup

on macOS and node 20 but not the problem here

### Prisma Version

4.16.1

### LINKED ISSUE #19958: Getting wrong types with prisma client extensions
### Bug description

The findFirst interface doesnt report incorrect types on `4.16.1`. On `4.15` I didn't have this issue.

### How to reproduce

```ts
const x = await prisma.user.findFirst({ stripeCustomerId: customerId });
```

This code is definitely incorrect, it needs a `where` clause. But it doesn't report it as an error.

If I get rid of the extension code, it correctly reports a type error.

### Expected behavior

Type error is shown

### Prisma information

```ts
const client = (url: string | undefined) =>
  new PrismaClient(url ? { datasources: { db: { url } } } : undefined).$extends({
    model: {
      $allModels: {
        async exists<T>(this: T, where: Prisma.Args<T, 'findFirst'>['where']): Promise<boolean> {
          const context = Prisma.getExtensionContext(this);
          const result = await (context as any).count({ where });
          return result > 0;
        }
      }
    }
  });
```

Getting rid of the extension works

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: 18.12.1

### Prisma Version

```
4.16.1
```

millsp:
closes https://github.com/prisma/prisma/issues/19921
closes https://github.com/prisma/prisma/issues/19958
closes https://github.com/prisma/team-orm/issues/118

Note that the tests were upgraded, and currently the version coming from `$extends` seems to be more type compliant (some tests wrt nullability are skipped with the regular client).

### DIFF
--- a/packages/client/src/runtime/core/types/Extensions.ts
+++ b/packages/client/src/runtime/core/types/Extensions.ts
@@ -2,10 +2,10 @@ import { Sql } from 'sql-template-tag'
 
 import { ITXClientDenyList } from '../../itxClientDenyList'
 import { RequiredArgs as UserArgs } from '../extensions/$extends'
-import { GetFindResult, GetResult as GetOperationResult, Operation } from './GetResult'
+import { FluentOperation, GetFindResult, GetResult as GetOperationResult, Operation } from './GetResult'
 import { Payload } from './Payload'
 import { PrismaPromise } from './Public'
-import { Call, ComputeDeep, Fn, Optional, Return, ToTuple, UnwrapTuple } from './Utils'
+import { Call, ComputeDeep, Exact, Fn, Optional, Path, Return, ToTuple, UnwrapTuple } from './Utils'
 
 /* eslint-disable prettier/prettier */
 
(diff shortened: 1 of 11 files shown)


packages/client/src/runtime/core/types/Public.ts
```diff
@@ -15,13 +15,14 @@ export type Args<T, F extends Operation> =
...
-  ? GetResult<T[symbol]['types']['operations'][F]['payload'], A, F>
+  T extends { [K: symbol]: { types: { payload: any } } }
+  ? GetResult<T[symbol]['types']['payload'], A, F>
   : never
 
-export type Payload<T, F extends Operation> =
-T extends { [K: symbol]: { types: { operations: { [K in F]: { payload: any } } } } }
-  ? T[symbol]['types']['operations'][F]['payload']
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export type Payload<T, F extends Operation = never> =
```
millsp, line 23: `Operation` is virtually useless now, keeping it not to break the API

packages/client/src/runtime/core/types/Extensions.ts
```diff
@@ -60,11 +60,11 @@ export type DynamicQueryExtensionArgs<Q_, TypeMap extends TypeMapDef> = {
         : never
 }
 
-export type DynamicQueryExtensionCb<TypeMap extends TypeMapDef, _0 extends PropertyKey, _1 extends PropertyKey, _2 extends PropertyKey> =
+type DynamicQueryExtensionCb<TypeMap extends TypeMapDef, _0 extends PropertyKey, _1 extends PropertyKey, _2 extends PropertyKey> =
```
Jolg42, line 63: @SevInf previously added these exports to fix an issue with the TypeScript compiler, see https://github.com/prisma/prisma/pull/19885

Are we sure exports are not needed anymore?

SevInf: There is a test that will fail if declaration for extended client is not created, so i think it's fine.

SevInf (approved):

PR #19985, opened 2023-06-27
feat(client): add missing groupBy type string on middleware params action type

iurylippo:
If you are trying to filter a param action by groupBy in a middleware, it generates  a ts error:

image

This feature add missing groupBy type string on middleware params action type.


aqrln (approved):
Thank you for the contribution, the change looks good to me!

I've updated the branch and the snapshots since they have changed on `main` due to being differentiated by engine/runtime type now.

PR #20045, opened 2023-07-03
chore: rename prisma-fmt-wasm to prisma-schema-wasm

Jolg42:
Part of https://github.com/prisma/prisma-engines/pull/4033

### DIFF
--- a/packages/internals/src/__tests__/engine-commands/formatSchema.test.ts
+++ b/packages/internals/src/__tests__/engine-commands/formatSchema.test.ts
@@ -12,7 +12,7 @@ if (process.env.CI) {
 
 const ctx = jestContext.new().add(jestConsoleContext()).assemble()
 
-describe('format wasm', () => {
+describe('schema wasm', () => {
   describe('diff', () => {
     test('2-spaces', async () => {
       const schema = await fs.promises.readFile(path.join(fixturesPath, 'format', 'schema-2-spaces.prisma'), {
--- a/packages/internals/src/panic.ts
+++ b/packages/internals/src/panic.ts
@@ -36,6 +36,7 @@ export enum ErrorArea {
   PHOTON_STUDIO = 'PHOTON_STUDIO',
   // Unused since 4.9.0 and now using `LIFT_CLI`
   INTROSPECTION_CLI = 'INTROSPECTION_CLI',
+  // Note: prisma-fmt-wasm was renamed to prisma-schema-wasm in 5.0.0
   FMT_CLI = 'FMT_CLI',
   QUERY_ENGINE_BINARY_CLI = 'QUERY_ENGINE_BINARY_CLI',
   QUERY_ENGINE_LIBRARY_CLI = 'QUERY_ENGINE_LIBRARY_CLI',
--- a/packages/internals/src/wasm.ts
+++ b/packages/internals/src/wasm.ts
@@ -1,4 +1,4 @@
-import prismaFmt from '@prisma/prisma-fmt-wasm'
+import prismaSchemaWasm from '@prisma/prisma-schema-wasm'
 
 import { WasmPanicRegistry } from './WasmPanicRegistry'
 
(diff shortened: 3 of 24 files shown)


Jolg42:
Note: ignore the "The system cannot find the path specified." for Windows run, it looks like it started with https://github.com/prisma/prisma/pull/19427 and I'm checking what is the problem there.

packages/internals/src/wasm.ts
```diff
@@ -1,14 +1,14 @@
...
+import prismaSchemaWasm from '@prisma/prisma-schema-wasm'
 
 import { WasmPanicRegistry } from './WasmPanicRegistry'
 
 // Note: using `import { dependencies } from '../package.json'` here would break esbuild with seemingly unrelated errors.
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { dependencies } = require('../package.json')
 
-export { prismaFmt }
+export { prismaSchemaWasm }
```
jkomyno, line 9: Nit: do we really need a named export in a `wasm.ts` file to have a `-Wasm` postfix?

jkomyno (approved):
It seems that both the CI and the codebase are handled! Nice, approving with a small nit that doesn't block merges.

packages/internals/src/wasm.ts
Jolg42: `prismaSchema` is usually used in the codebase to refer to the Prisma schema file, so I think we can be very explicit here

packages/internals/src/__tests__/engine-commands/formatSchema.test.ts
```diff
@@ -12,7 +12,7 @@ if (process.env.CI) {
 
 const ctx = jestContext.new().add(jestConsoleContext()).assemble()
 
-describe('format wasm', () => {
+describe('schema wasm', () => {
```
janpio, line 15: doesn't this refer to the action of formatting?

packages/internals/src/panic.ts
```diff
@@ -36,6 +36,7 @@ export enum ErrorArea {
   PHOTON_STUDIO = 'PHOTON_STUDIO',
   // Unused since 4.9.0 and now using `LIFT_CLI`
   INTROSPECTION_CLI = 'INTROSPECTION_CLI',
+  // Note: prisma-fmt-wasm was renamed to prisma-schema-wasm in 5.0.0
```
janpio, line 39: What does this refer to here?

packages/internals/src/wasm.ts
janpio: I agree, just `prismaSchema` would be dangerous.

packages/internals/src/wasm.ts
```diff
@@ -1,14 +1,14 @@
...
 // e.g. 4.3.0-18.a39215673171b87177b86233206a5d65f2558857
-export const prismaFmtVersion: string = dependencies['@prisma/prisma-fmt-wasm']
+export const prismaSchemaVersion: string = dependencies['@prisma/prisma-schema-wasm']
```
janpio, line 11: Shouldn't this be `prismaSchemaWasmVersion` then?

jkomyno: Right, https://github.com/prisma/prisma/pull/20106

packages/internals/src/panic.ts
jkomyno: The error area is something used in our internal error reporting dashboard to categorise Rust panics' "topic areas". It mirrors this `ErrorArea` enum [here](https://github.com/prisma/error-handling-backend/blob/482bf538ec055f4ef41926736d72c5f08e3b9a07/src/types/enums/index.ts#L11-L27).

packages/internals/src/__tests__/engine-commands/formatSchema.test.ts
Jolg42: Yes, here actually a better name would be "format" but actually these tests are now duplicates of the ones below, and we could clean up there, though not urgent at all.

PR #20071, opened 2023-07-05
feat(client)!: remove beforeExit from LibraryEngine and DataProxyEngine

### LINKED ISSUE #19312: Prisma Client: remove "beforeExit" hook from LibraryEngine/DataProxyEngine
```tsx
prisma.$on('beforeExit', () => {
   doSomething()
})
```

Notes:
Binary engine uses process exit hooks to shut down child process automatically and exposes `beforeExit` event so users still have an opportunity to handle process exit.
Library engine does not need to subscribe to signals/process exit events to function, but still does so it can expose this event. If this event is used, engine instance can not be GCed correctly. Since there is no need to listen to process exit for our own needs, we can remove this event from our API and ask users to setup process handlers themselves instead - data proxy engine already does that.

- [ ] prisma/docs#4998
- [ ] Remove code from LibraryEngine
- [ ] Also remove TODO & `NotImplementedYetError` from DataProxyEngine: see `./packages/engine-core/src/data-proxy/DataProxyEngine.ts`
- [ ] Replace by "only implemented in Binary engine"?
- [ ] Remove/update tests
- [ ] Release notes draft
- [ ] Anything else?

aqrln:
- Remove beforeExit from LibraryEngine and DataProxyEngine
- Run the exit hooks test only with binary engine
- Don't generate beforeExit type for the data proxy engine and the library engine

Closes https://github.com/prisma/prisma/issues/19312


jkomyno (approved):

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -348,7 +337,11 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
 
   on(event: EngineEventType, listener: (args?: any) => any): void {
     if (event === 'beforeExit') {
-      this.beforeExitListener = listener
+      console.warn(
+        `${yellow(
+          'prisma:warn',
+        )} "beforeExit" hook is not applicable to the library engine type, it is only relevant and implemented for the binary engine type. Please add your event listener to the \`process\` object directly instead.`,
```
Jolg42, line 343:
```suggestion
        )} "beforeExit" hook is not applicable to the library engine type since Prisma 5.0.0, it is only relevant and implemented for the binary engine type. Please add your event listener to the \`process\` object directly instead.`,
```

Jolg42 (approved):
Looks good to me :shipit:

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -307,6 +307,14 @@ function runCommandRawDefinition(this: PrismaClientClass) {
   return ts.stringify(method, { indentLevel: 1, newLine: 'leading' })
 }
 
+function eventRegistrationMethodDeclaration(dataProxy: boolean) {
+  if (dataProxy) {
```
SevInf, line 311: Should we also generate the same on `getClientEngineType() === ClientEngineType.Library` ?

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
SevInf: Should we have a hard error here? `beforeExit` event will never work, `console.warn` might be pretty easy to miss.

packages/client/src/generation/TSClient/PrismaClient.ts
aqrln: Actually yes, I forgot that the engine type can't be switched without regenerating the client anymore!

packages/client/src/utils/getTestClient.ts
```diff
@@ -65,10 +65,20 @@ export async function getTestClient(schemaDir?: string, printWarnings?: boolean)
   return getPrismaClient(options)
 }
```
Jolg42, line 66: Maybe could use some comments to document this for us internally

Jolg42 (approved):

SevInf (approved):

PR #20075, opened 2023-07-05
test(integration/happy/enum): dbgenerated() - update schema to current introspection state

janpio:
No need to wait to update the schema until we remove `dbgenerated()`, this already does not introspect like that any more.

(Created via downloading the database and running Introspection with 4.16.2)

Part of https://github.com/prisma/prisma/issues/19313

### DIFF
--- a/packages/client/src/__tests__/integration/happy/enums/schema.prisma
+++ b/packages/client/src/__tests__/integration/happy/enums/schema.prisma
@@ -3,30 +3,30 @@ datasource my_db {
   url      = "file:dev.db"
 }
 
+model Location {
+  city String @default("")
+  id   Int    @id @default(autoincrement())
+  User User[]
+}
+
+model Post {
+  createdAt DateTime @default(dbgenerated("'1970-01-01 00:00:00'"))
+  email     String   @unique(map: "Post.email") @default("")
+  id        String   @id
+  name      String   @default("")
+  updatedAt DateTime @default(dbgenerated("'1970-01-01 00:00:00'"))
+  user      String?
+  User      User?    @relation(fields: [user], references: [id], onUpdate: NoAction)
+}
+
 model User {
-  email        String    @default("") @unique
+  email        String    @unique(map: "User.email") @default("")
   favoriteTree String    @default("ARBORVITAE")
   id           String    @id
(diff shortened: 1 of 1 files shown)


packages/client/src/__tests__/integration/happy/enums/schema.prisma
```diff
@@ -3,30 +3,30 @@ datasource my_db {
...
+  id        String   @id
+  name      String   @default("")
+  updatedAt DateTime @default(dbgenerated("'1970-01-01 00:00:00'"))
+  user      String?
+  User      User?    @relation(fields: [user], references: [id], onUpdate: NoAction)
+}
+
 model User {
-  email        String    @default("") @unique
+  email        String    @unique(map: "User.email") @default("")
```
Jolg42, line 23: Is that "@unique(map: "User.email")" really expected? 🤔

packages/client/src/__tests__/integration/happy/enums/schema.prisma
```diff
@@ -3,30 +3,30 @@ datasource my_db {
...
 
+model Location {
+  city String @default("")
+  id   Int    @id @default(autoincrement())
+  User User[]
+}
+
+model Post {
+  createdAt DateTime @default(dbgenerated("'1970-01-01 00:00:00'"))
+  email     String   @unique(map: "Post.email") @default("")
```
Jolg42, line 14: Is that map for "@unique(map: "Post.email")" really expected? 🤔

Jolg42 (approved):

packages/client/src/__tests__/integration/happy/enums/schema.prisma
janpio: At least for this database, yes. 
Either the SQLite default naming scheme is different from what Prisma expects, or this database was created with these explicit names that do not match our naming scheme for unique indexes - so it renders them.

packages/client/src/__tests__/integration/happy/enums/schema.prisma
janpio: https://github.com/prisma/prisma/pull/20075#discussion_r1254275417

PR #20096, opened 2023-07-06
fix(client): statically inject prismaVersion to default-index.js

### LINKED ISSUE #19972: Prisma Client: updating to 4.16.0 or 4.16.1 breaks Cloudflare worker, it errors with `The package "path" wasn't found on the file system but is built into node`
### Bug description

When you update prisma to the latest version (4.16.1) it breaks `wrangler`.

The error is:
```bash
  node_modules/@prisma/client/scripts/default-index.js:1:21:
      1 │ const path = require('path')
  
  The package "path" wasn't found on the file system but is built into node
```

A workaround is to set the `node_compat = true` flag in the wrangler.toml file.
However this comes with some downfalls.

### How to reproduce

1. Install prima 4.16.1
2. Generate edge client with the --data-proxy flag
3. Setup prisma using the edge client

### Expected behavior

Edge client works in edge environments. 

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
  previewFeatures = ["jsonProtocol"]
}

datasource db {
  provider = "mysql"
  url = env("DATABASE_URL")
  relationMode = "prisma"
}
```

```ts
import { PrismaClient } from "@prisma/client/edge";

export default function prisma(url: string) {
    datasources: {
      db: {
        url,
      },
    },
  });

  return client;
}

export type ExtendedPrismaClient = ReturnType<typeof prisma>;

```

### Environment & setup

- OS: macOS
- Database: PlanetScale
- Node.js version: 18.16

### Prisma Version

```
4.16.1
```

aqrln:
Resolving the client and engines version dynamically by reading
`package.json` files in `default-index.js` is not be possible in edge
runtimes like Cloudflare Workers, or might even be potentially problematic
when using bundlers with Node.js. We can inject the version statically
at build time instead to avoid the need to figure it out at run time.

Fixes: https://github.com/prisma/prisma/issues/19972


packages/client/helpers/build.ts
```diff
@@ -100,6 +101,20 @@ function writeDtsRexport(fileName: string) {
   fs.writeFileSync(path.join(runtimeDir, fileName), 'export * from "./index"\n')
 }
 
+function writeDefaultIndexFile() {
```
millsp, line 104: wdyt about having a `script` folder in `src` with a ts file, and using the bundler to inject the right version? It might be less lines of code, but overall I am fine with this approach too.

aqrln: that sounds better, yep

aqrln: After trying this one thing I'm not super happy with is that the bundler can't tree-shake the json files, which leads the whole `package.json` contents of both `@prisma/client` and `@prisma/engines-version` to be included into the bundle. Together with a little bit of other generated boilerplate, this inflates the `default-index.js` to 9 Kb.

Do you think it's acceptable or should we rather not do it?

millsp: I thought it could tree-shake things via imports like `import { version } from "./package.json"`, if that is not the case then drop it.

aqrln: Indeed! The problem was I used `clientVersion` from `runtime/utils`, and it used `require` rather than `import` (and, likewise, so does `@prisma/engines-version`). I pushed a commit now.

I also tried moving the types to the `.ts` file and generating the `.d.ts` but couldn't get it to work properly (the generated `.d.ts` file looked like `export * from "/Users/aqrln/prisma/prisma/packages/client/src/scripts/default-index.ts"`).

Jolg42 (approved):

Issue #14640, opened 2022-08-03
Preview feature feedback: Prisma tracing (`tracing`)


keerlu:
Please share your feedback about the `tracing` functionality released in [v4.2.0](https://github.com/prisma/prisma/releases/tag/4.2.0) in this issue.

- If you encounter a bug, please [open a bug report](https://github.com/prisma/prisma/issues/new?assignees=&labels=&template=bug_report.md&title=) in this repo.
- If the feature is working well for you, please share this in a comment below or leave a 👍 on this issue.

If you have any questions, don't hesitate to ask them in the [`#prisma-client`](https://prisma.slack.com/archives/CCWDULGUW) channel in the [Prisma Slack](https://slack.prisma.io).

safareli:
With this, it's still not possible to "tag" prisma queries with the request id they are executed as part of, if i'm wrong please share how?


For context please take a look at https://github.com/prisma/prisma/issues/10425

safareli:
Iin slack it this was shared:

> Tracing allows you to track _requests_ as they flow through your application. This is especially useful for debugging distributed systems where each request can span multiple services.


it would be very useful if we are able to attach request id of an http request to all queries which are created as part of the http request. (for logging, and for rate limiting)

Akxe:
Please add this to the documentation (in some form):

> To run setup locally one can use local installation of Jeager with this command:
> `.\jaeger-all-in-one.exe --collector.zipkin.host-port=:9411 --collector.otlp.enabled`

I struggled for so long because I missed the `--collector.otlp.enabled` flag.

millsp:
It would be great to have a way to "label" prisma queries so that name can appear in the traces. In other words, if your label is create_a_user, instead of having `prisma:client:operation`, you'd have `prisma:client:operation:create_a_user`.

millsp:
I'd like to have another option in for the instrumentation package serialize: boolean, because it might not always be useful to pollute the spans with these prisma:*:serialize traces, I would like to disable it by default but give the option to enable it as necessary.

brennan-airtime:
It appears that prisma is sending db spans with an attribute of `db.type: postgres` (for my pg example) but by the semantic conventions in the spec the key should be `db.system` and the value `postgresql`. 

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md

[...]

Issue #17953, opened 2023-02-16
`prisma:engine` spans do not respect `suppressTracing()`


uhyo:
### Bug description

When using Prisma's `tracing` preview feature, `prisma:engine` spans do not respect parent context being [suppressTracing()](https://github.com/open-telemetry/opentelemetry-js/blob/279458e/packages/opentelemetry-core/src/trace/suppress-tracing.ts#L23)-ed.

### How to reproduce

See https://github.com/uhyo/prisma-trace-bug-repro for a complete reproduction.

#### Correctly Working Example
Suppose a Prisma query is run inside another span. Prisma-related spans become children of the span, which is expected behavior.

```ts
  await tracer.startActiveSpan("main-outer", async (span) => {
    try {
      await tracer.startActiveSpan("main-inner", async (span) => {
        try {
          const prisma = new PrismaClient();
          const count = await prisma.user.count();
          console.log({ count });
          await prisma.$disconnect();
        } finally {
          span.end();
        }
      });
    } finally {
      span.end();
    }
  });
```

Generated spans are visualized as follows, where we can see that Prisma-related spans are correctly placed below our `main-inner`span.

```
----- traceId = 2ad011bfec91f3dc46af7701e87c7484 -----
main-outer (dd2f5e0728196a28) sampled=true
  main-inner (c760fcd3b1fa373d) sampled=true
    prisma:client:operation (df101553ab91918a) sampled=true
      prisma:engine (9a1236432ce66c79) sampled=true
        prisma:engine:connection (bbfb416e35154346) sampled=true
        prisma:engine:db_query (85cae0bd1d1f746d) sampled=true
        prisma:engine:serialize (182fd59131b12f83) sampled=true
      prisma:client:connect (f3d1c10e9a91d985) sampled=true
      prisma:client:serialize (11d98ef1631251d0) sampled=true
    prisma:client:disconnect (2bfb1ab20b42b487) sampled=true
```

#### Buggy Example

Suppose we wrapped the Prisma operation with a context which is the result of `suppressTracing`. We expect any spans inside this context to be suppressed.

```ts
  await tracer.startActiveSpan("main-outer", async (span) => {
    try {
      await context.with(suppressTracing(context.active()), async () => {
        await tracer.startActiveSpan("main-inner", async (span) => {
          try {
            const prisma = new PrismaClient();
            const count = await prisma.user.count();
            console.log({ count });
            await prisma.$disconnect();
          } finally {
            span.end();
          }
        });
      });
[...]

PR #20113, opened 2023-07-07
feat(instrumentation): add ability to filter span by `PrismaLayerType`

overbit:
Add ability to filter out some of the instrumentation span generated by prisma

Related #14640

https://github.com/prisma/prisma/issues/14640#issuecomment-1264641220


janpio:
(The failing test is unrelated)

aqrln:
The code changes look good, thanks so much for the contribution! This definitely needs to be tested, would you be able to add some tests, or do you need help with this?

overbit:
> The code changes look good, thanks so much for the contribution! This definitely needs to be tested, would you be able to add some tests, or do you need help with this?

I could add some unit tests but I'm not sure what is the strategy here. The only other tests are pretty basic.

Any suggestions? Otherwise I'm happy if you guys want to add specific tests.

overbit:
@aqrln tests added. There are failing tests in the ci but they shouldn't be related to these changes

janpio:
(Retriggered a run for the failing tests)

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -45,6 +50,10 @@ export class ActiveTracingHelper implements TracingHelper {
     const tracer = trace.getTracer('prisma') as Tracer
 
     engineSpanEvent.spans.forEach((engineSpan) => {
+      if (this.ignoreLayersTypes.map((value) => value.toString()).includes(engineSpan.name)) {
```
aqrln, line 53: Should we maybe do the mapping in the constructor instead to avoid allocating a new array each time a new span arrives?

aqrln (commented):
This looks very solid, thank you for adding the tests! What I was having in mind was rather adding some integration tests in `packages/client/test/functional/tracing` (sorry for not making this clear, it's not easy to navigate our test structure) but these unit tests should also be very useful. We can add the integration tests ourselves because I don't want to burden you with more work and I'd like to merge this PR soon — but if you are interested and want to, feel free to look into adding them, just let me know!

packages/instrumentation/src/ActiveTracingHelper.ts
overbit: Done

overbit:
@janpio @aqrln any update on this? We really need it otherwise we will disable the prisma instrumentation all together due to the amount of segments created

overbit:
@aqrln any news? In [@amplication](https://github.com/amplication/amplication) we ended up disabling prisma tracing at the moment as without this fix, it was creating too many segments

overbit:
@aqrln @SevInf I've update the branch from main as this PR is still awaiting for review

overbit:
@janpio @aqrln any update?

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -21,12 +23,15 @@ const nonSampledTraceParent = `00-10-10-00`
 
 type Options = {
   traceMiddleware: boolean
+  ignoreLayersTypes: PrismaLayerType[]
```
SevInf, line 26: Could you clarify why you use `layer` name for this feature?
We are filtering out spans, OTEL docs also call them spans. What would you think about renaming this property to `ignoreSpanNames`?

packages/instrumentation/src/types.ts
```diff
@@ -0,0 +1,12 @@
+export enum PrismaLayerType {
```
SevInf, line 1: This list is not exhaustive and it has some typos in it.
Actually, I'd prefer we remove enum type completely and type new option as the `string`. We have no good way to keep this list up to date without quite a bunch of maintenance burden

packages/instrumentation/src/__tests__/ActiveTracingHelper.test.ts
```diff
@@ -0,0 +1,148 @@
+import { trace } from '@opentelemetry/api'
```
SevInf, line 1: We'd prefer to test this over as a functional test, rather than unit test with a lot of mocking, so we can be sure this works with the rest of the OTEL stack.

You could add a test to an existing `tracing` test [here](https://github.com/prisma/prisma/blob/b333dcd4979d65dc34036d07f998566fb6f520c9/packages/client/tests/functional/tracing/tests.ts#L87) and documentation on our functional tests [here](https://github.com/prisma/prisma/blob/main/TESTING.md#prisma-client-functional-tests)

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -100,6 +109,11 @@ export class ActiveTracingHelper implements TracingHelper {
     const tracer = trace.getTracer('prisma')
     const context = options.context ?? this.getActiveContext()
     const name = `prisma:client:${options.name}`
+
+    if (this.ignoreLayersTypes.includes(name)) {
```
SevInf, line 113: More of a conceptual question rather than request to change: should ignoring the span also ignore all of it's children? If so, i think we'll need to suppress tracing for the callback.
See, for example, [how http instrumentation does it](https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L494).

Problem here is that it won't work for engine spans (see #17953), but that's something for us to solve, not for this PR.

SevInf (dismissed):
Hi @overbit.

I am very sorry for the delay. I've left my review, there are couple of things that would need to change. Let me know if you'll have time to do it - totally ok if you don't, i'll take over the PR and merge it if that's the case.

Thank you for submitting it and once more, apologies for a big delay from our side.

packages/instrumentation/src/ActiveTracingHelper.ts
overbit: No issues with that. The `layer` name was meant to follow Prisma naming as it was the target. Happy to update it

packages/instrumentation/src/ActiveTracingHelper.ts
overbit: That is a great question. One scenario is related to the need to trace the query information without the data related to connection, serialisation and co. 
If we follow the same approach of http instrumentation, we would need to update the prisma layers to allow a more flexible configuration.

Personally, the most useful span related to an ORM is wrapping the full ORM logic and presenting query data (keeping PII obfuscation in mind) considering that tools like AWS X-Ray will only display traces with up to 200 segments

nomego:
Any progress?

packages/instrumentation/src/ActiveTracingHelper.ts
lucas-gregoire, line 26: Wouldn't it be more flexible to use hooks instead of a list of span names to filter? 
Beyond filtering spans, we could consider adding or transforming attributes (on engine or client spans). 
This is what is done in the HTTP instrumentation: https://www.npmjs.com/package/@opentelemetry/instrumentation-http (`applyCustomAttributesOnSpan`, `requestHook`, ...) and it's quite an interesting approach.

overbit:
@SevInf apologies but I haven't got any time to proceed and resolve your comments. Feel free to take over from here

lewinskimaciej:
For people who got here like me after a while, as it seems Prisma team won't implement any filtering on their own, you can achieve what you want by creating a custom SpanProcessor(/BatchSpanProcessor) like this:
```ts
import {
  ReadableSpan,
  SpanExporter,
  BatchSpanProcessor,
} from '@opentelemetry/sdk-trace-base';

export class FilteringSpanProcessor extends BatchSpanProcessor {
  private spansToFilterOut: RegExp;

  constructor(exporter: SpanExporter, spanPatternToFilterOut: string) {
    super(exporter);
    this.spansToFilterOut = new RegExp(spanPatternToFilterOut);
  }

  onEnd(span: ReadableSpan): void {
    // Filter out spans based on custom criteria
    if (this.shouldExport(span)) {
      super.onEnd(span);
    }
  }

  private shouldExport(span: ReadableSpan): boolean {
    return !this.spansToFilterOut.test(span.name);
  }
}

```

and registering it with NodeSdk:
```ts
new FilteringSpanProcessor(
        traceExporter,
        // this will filter out anything that is not prisma:client:operation or prisma:engine:query/db_query spans
        '^(prisma:(?!client:operation|engine:(query|db_query)).*)', 
      ),
```

FGoessler:
Thanks again @overbit for this nice PR! I took the liberty to apply the suggested changes, improve the test suite and slightly updated the functionality to now also support regex based filtering. ;) 

Filters can be supplied on the `PrismaInstrumentation` constructor. E.g.: 
```typescript
new PrismaInstrumentation({ 
  ignoreSpanTypes: [
    /prisma:client:ser.*/, 
    'prisma:client:connect', 
    'prisma:engine:query',
  ], 
})
```

Note that if you filter out a parent span all its child spans will also be filtered out. E.g. `prisma:engine:query` has various child spans types as well as `prisma:client:transaction` and `prisma:client:operation`.

Tracing docs can be found here: https://www.prisma.io/docs/orm/prisma-client/observability-and-logging/opentelemetry-tracing

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -85,6 +88,11 @@ export class ActiveTracingHelper implements TracingHelper {
     const tracer = this.tracerProvider.getTracer('prisma')
     const context = options.context ?? this.getActiveContext()
     const name = `prisma:client:${options.name}`
+
+    if (shouldIgnoreSpan(name, this.ignoreSpanTypes)) {
+      return callback()
+    }
```
aqrln, line 94: What happens if we are in a parent span created by user in the application code, and `prisma:client:operation` is skipped? Won't this lead to the child spans being created and linked to the application span anyway?

FGoessler: Based on the functional tests child spans are fully filtered out if their parent is filtered out. Why do you think this might not be the case? 🤔

FGoessler: Updated the test as we discussed

aqrln (approved):

PR #20118, opened 2023-07-07
test(client): add generator-config-types test

aqrln:
Follow-up to https://github.com/prisma/prisma/pull/20119

### DIFF
--- a/packages/client/tests/e2e/_utils/run.ts
+++ b/packages/client/tests/e2e/_utils/run.ts
@@ -51,34 +51,49 @@ async function main() {
   console.log('🎠 Preparing e2e tests')
   // we first get all the paths we are going to need to run e2e tests
   const tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), 'prisma-build'))
+
   const cliPkgPath = path.join(monorepoRoot, 'packages', 'cli')
   const wpPluginPkgPath = path.join(monorepoRoot, 'packages', 'nextjs-monorepo-workaround-plugin')
   const clientPkgPath = path.join(monorepoRoot, 'packages', 'client')
+  const enginesPkgPath = path.join(monorepoRoot, 'packages', 'engines')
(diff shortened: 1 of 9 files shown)


packages/client/tests/e2e/_utils/run.ts
```diff
@@ -51,45 +51,66 @@ async function main() {
...
   const restoreOriginal = async () => {
     await $`cd ${tmpDir} && cp cli.package.json ${cliPkgJsonPath}`
     await $`cd ${tmpDir} && cp client.package.json ${clientPkgJsonPath}`
+    await $`cd ${tmpDir} && cp generator-helper.package.json ${generatorHelperPkgJsonPath}`
   }
 
   // if process is killed by hand, ensure that package.json is restored
   process.on('SIGINT', () => restoreOriginal().then(() => process.exit(0)))
 
-  // use bundleDependencies to directly include deps like @prisma/engines
```
aqrln, line 75: `bundleDependencies` didn't work with transitive dependencies of `@prisma/debug`, and those were missing after package installation.

I haven't tried recursively reading `package.json` files and adding transitive dependencies to `bundleDependencies` as well — perhaps that could have worked.

Instead, I replaced references to `@prisma/engines` and `@prisma/debug` with packaged tarballs. One downside of this approach that I can see is that since I used absolute paths, now they are only usable inside the docker container.

millsp (approved):

packages/client/tests/e2e/_utils/run.ts
millsp: Could you please open a discussion with Zoltan in the pnpm repo? (Otherwise I can do it if you have other priorities)

aqrln: forgot to post the link here, https://github.com/orgs/pnpm/discussions/6861

millsp: I just made that everything is put in /tmp locally and in the docker container, so the difference is eliminated for macOS and Linux.

PR #20120, opened 2023-07-07
fix(client): don't attempt to copy default-index.js if it doesn't exist

aqrln:
The file may not exist yet when running `pnpm install` in the freshly
cloned repo before building the client package for the first time.
It should always exist in the published package.

### DIFF
--- a/packages/client/scripts/postinstall.js
+++ b/packages/client/scripts/postinstall.js
@@ -211,7 +211,9 @@ async function createDefaultGeneratedThrowFiles() {
     await makeDir(dotPrismaClientDir)
     await makeDir(defaultDenoClientDir)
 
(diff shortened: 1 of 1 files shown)


millsp:
Could we disable the `postinstall` in the monorepo somehow (eg: version 0.0.0) so that the client isn't generated at all?

Jolg42 (approved):

aqrln:
> Could we disable the postinstall in the monorepo somehow (eg: version 0.0.0) so that the client isn't generated at all?

Checking for version 0.0.0 would disable postinstall in repros and e2e tests I think (which might be a problem if we want to, e.g., explicitly test the postinstall script).

We could probably try to come up with some other heuristics but that would add some complexity (and I don't see a clean way to do it).

PR #20131, opened 2023-07-09
fix(client): initialize generic parameters of ExtendsHook 

### LINKED ISSUE #20128: Client Extensions destroys types using Prisma with javascript
### Bug description

When using Prisma in a .js file types are erased once the Client is extended. This is happening whether the extension is mounted directly or created with `defineExtension`. 

### How to reproduce

1. Go to https://github.com/prisma/prisma-client-extensions/tree/main/static-methods
2. Change script.ts to script.js (and remove TS syntax)
3. See types being erased from the `prisma` variable

### Expected behavior

Types to be inferred just as they would when Prisma is used in Typescript file (.ts)

### Prisma information

```prisma
generator client {
}

datasource db {
  provider = "sqlite"
  url      = "file:./dev.db"
}

model User {
  id       String    @id @default(cuid())
  email    String
  password Password?
}
```

```js
// index.js

  model: {
    user: {
      async findSomething(arg) {
        return null
      }
    }
  }
});

const x = await prisma.user.findSomething()
// type is `any` ^^^
```

### Environment & setup

- OS: MacOS
- Editor: VSCode Version: 1.80.0
- Database: SQLite
- Node.js version: v18

### Prisma Version

```
Query Engine (Node-API) : libquery-engine 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules/@prisma/engines/migration-engine-darwin)
```

adamjkb:
Fixes #20128

Explicitly initializes the default state of the ExtendsHook interface so in a JS environment they don't default to `any`and breaking the Client's type. 

_The `any` fallback is one of the difference how JS is different from  TS: https://github.com/microsoft/TypeScript/issues/30009#issuecomment-469385244_

The change in the PR is just defining the default explicitly but—as far as I can tell— Typescript already does this implicitly.


adamjkb:
This is in fact breaking types from external plugins. So it might not be as simple as this. _(Briefly tested it w/ `withAccelerate` and it resulted in the same issue as with the original issue.)_

millsp:
@adamjkb what would be the next steps?

PR #20138, opened 2023-07-09
feat: remove error messages for valid NixOS setups

Gerschtli:
This PR fixes the issues described in https://github.com/prisma/prisma/issues/17900. Please have a look at this ticket regarding the objective.

Basically with these changes, the error messages regarding not available precompiled engine files is now only shown if there is an error in the setup, if the environment variables are correctly used, no error is shown.

### DIFF
--- a/packages/fetch-engine/src/download.ts
+++ b/packages/fetch-engine/src/download.ts
@@ -13,7 +13,7 @@ import { BinaryType } from './BinaryType'
 import { chmodPlusX } from './chmodPlusX'
 import { cleanupCache } from './cleanupCache'
 import { downloadZip } from './downloadZip'
-import { getBinaryEnvVarPath } from './env'
+import { allEngineEnvVarsSet, getBinaryEnvVarPath } from './env'
 import { getHash } from './getHash'
 import { getBar } from './log'
 import { getCacheDir, getDownloadUrl, overwriteFile } from './utils'
@@ -70,8 +70,12 @@ export async function download(options: DownloadOptions): Promise<BinaryPaths> {
   const platform = await getPlatform()
   const os = await getos()
 
-  if (os.targetDistro && ['nixos'].includes(os.targetDistro)) {
-    console.error(`${yellow('Warning')} Precompiled engine files are not available for ${os.targetDistro}.`)
+  if (os.targetDistro && ['nixos'].includes(os.targetDistro) && !allEngineEnvVarsSet(Object.keys(options.binaries))) {
+    console.error(
+      `${yellow('Warning')} Precompiled engine files are not available for ${
+        os.targetDistro
+      }, please provide the paths via environment variables, see https://pris.ly/d/custom-engines`,
+    )
   } else if (['freebsd11', 'freebsd12', 'freebsd13', 'openbsd', 'netbsd'].includes(platform)) {
     console.error(
       `${yellow(
--- a/packages/fetch-engine/src/env.ts
+++ b/packages/fetch-engine/src/env.ts
@@ -70,3 +70,13 @@ function getEnvVarToUse(binaryType: BinaryType): string {
 
   return envVar
 }
+
+export function allEngineEnvVarsSet(binaries: string[]): boolean {
+  for (const binaryType of binaries) {
+    if (!getBinaryEnvVarPath(binaryType as BinaryType)) {
+      return false
+    }
+  }
+
+  return true
+}


Gerschtli:
If someone has some pointers for me, why some tests are failing, that would be highly appreciated :)

janpio:
CI is flaky recently, so I just retriggered them. Let's see.

Gerschtli:
Ah okay, thanks, now it is only the test that already fails in main.

janpio:
Yes, so your code seems to be good form that perspective. Sorry for the confusion, we'll hopefully get rid of that both soon.

aqrln:
Thank you for the PR @Gerschtli!

The core change looks good but I think requiring an additional environment variable for OpenSSL path is a bit superfluous. Instead, I'd prefer for OpenSSL search to be bypassed completely when the engines env vars are set (and therefore the warning too). That might require introducing a new platform type for Linux with unknown OpenSSL, so that's something to think about.

Now, talking about the suggested `OPENSSL_LIB_DIR` in general (e.g. for the use case it was suggested for in the other PR), the fundamental problem with it is that it can be set to whatever with no guarantees that this directory is known by the dynamic linker. While I can see how it would work on NixOS with full paths to libraries embedded into elf headers (disregarding for a moment that searching for OpenSSL to determine its version is pointless on NixOS with custom binaries in the first place), generally it would be too easy to misuse it and set it to a path from which OpenSSL couldn’t be loaded. It’s basically equivalent to allowing to forcefully override the detected OpenSSL version without checking anything, but with extra steps.

I can only see two use cases for this env var:
1. To work around bugs or omissions in our detection logic.
2. To override the path to OpenSSL when that path is non-standard an not in `ld.so.conf`, but instead configured via `LD_LIBRARY_PATH` env var.

As for (1), optimally we’d get bug reports and fix those issues.

As for (2), it’s better to just read `LD_LIBRARY_PATH` directly instead. In fact, this is an omission in our detection logic which we should add.

I think fixing (2) would even allow using our stock pre-built `debian-openssl-3.0.x` engines on NixOS with something like `LD_LIBRARY_PATH = with pkgs; lib.makeLibraryPath [ openssl ];` in `devShell`.

Gerschtli:
Thank you for the review! That makes sense, I will have a look at replacing `OPENSSL_LIB_DIR` with `LD_LIBRARY_PATH`.

Gerschtli:
I force-pushed a rebase onto latest main and only changed my first commit, PTAL :)

packages/get-platform/src/getPlatform.ts
```diff
@@ -288,6 +288,11 @@ type ComputeLibSSLSpecificPathsParams = {
 }
 
 export function computeLibSSLSpecificPaths(args: ComputeLibSSLSpecificPathsParams) {
+  if (process.env.LD_LIBRARY_PATH) {
+    debug(`Found 'LD_LIBRARY_PATH' specified as env-var, ignoring platform specific lib-dir paths...`)
+    return process.env.LD_LIBRARY_PATH.split(':')
```
aqrln, line 293: While `LD_LIBRARY_PATH` does take precedence, it does not completely supplant the paths from `ld.so.conf`, so we still have to check the standard paths.

Consider the case when `LD_LIBRARY_PATH` is required for native dependencies of some other Node.js addon that an application uses (a real example from my experience is a proprietary library for interfacing with certain devices) but OpenSSL is installed via the package manager.

Gerschtli:
I split the PR into https://github.com/prisma/prisma/pull/20381 :)

packages/get-platform/src/getPlatform.ts
Gerschtli: Should be fixed in #20381 :)

packages/fetch-engine/src/download.ts
```diff
@@ -70,8 +70,12 @@ export async function download(options: DownloadOptions): Promise<BinaryPaths> {
   const platform = await getPlatform()
   const os = await getos()
 
-  if (os.targetDistro && ['nixos'].includes(os.targetDistro)) {
-    console.error(`${yellow('Warning')} Precompiled engine files are not available for ${os.targetDistro}.`)
+  if (os.targetDistro && ['nixos'].includes(os.targetDistro) && !allEngineEnvVarsSet(Object.keys(options.binaries))) {
+    console.error(
+      `${yellow('Warning')} Precompiled engine files are not available for ${
+        os.targetDistro
+      }, please provide the paths via environment variables, see https://www.prisma.io/docs/concepts/components/prisma-engines#using-custom-engine-libraries-or-binaries.`,
```
aqrln, line 77: Providing the link to the docs here is a great idea, thank you for doing this! Let's add a level of indirection via a shortlink to ensure the URL always stays relevant, and to make the console output cleaner, I'll open a PR in https://github.com/prisma/pris.ly

aqrln:
```suggestion
      }, please provide the paths via environment variables, see https://pris.ly/d/custom-engines`,
```

janpio: pris.ly PR is merged, so that link now works.

aqrln (approved):
This looks good to me!

I can confirm that with these changes the only warning that is printed is `prisma:warn Prisma failed to detect the libssl/openssl version to use, and may not work as expected. Defaulting to "openssl-1.1.x".`, which is already a big improvement.

While https://github.com/prisma/prisma/pull/20381 provides *a* way to get rid of that warning, I still don't think it is *the* right way when custom binaries are provided. I understand though if it's acceptable for you and you are okay with using it for this purpose.

Both PRs are neat and self-contained (and #20381 implements functionality that is useful for everyone, not just NixOS users), but there are some parts missing from the big picture if you'd like to continue working on this (or for us to tackle in the future).

Basically there are two separate tracks stemming from each of the two PRs:

1. Based on this PR, we could skip OpenSSL resolution when all engine env vars are set, as it's just useless work and the result of that resolution is not used in any meaningful way when the env vars are provided (except for things like printing the current platform in `prisma version` output — we will have to handle that somehow). This will get rid of the OpenSSL warning on NixOS, and is a nice optimization for everyone using custom engines. Also potentially useful for shaving off some extra milliseconds (or even tens of milliseconds) from cold starts in serverless environments.
2. Based on #20381, we could check if all the required shlibs are present in `LD_LIBRARY_PATH` (I guess we might also need to check for libgcc_s?) and download the debian binaries rather than bailing out with an error on NixOS. I think it might be a little convoluted but maybe worth exploring, at least as an experiment.

aqrln:
And thank you so much for doing this!

PR #20150, opened 2023-07-10
test: Test naming conflicts for models

SevInf:
Conflicts themselves should've were fixed in 14.16.x, but we had no
tests for them.

Note: While it is possible to use those names for models and composites,
enums still conflict. Fix for enums will be done in a separate PR.


millsp (approved):

Jolg42 (approved):
Nice!

PR #20159, opened 2023-07-10
chore: update to pnpm 8

aqrln:


aqrln:
Tests are failing because `findPrismaClientDir` is apparently failing to find the right directory, which triggers auto-installation during generation, which causes `@prisma/client@4.16.2` to be installed as a dependency and used.

aqrln:
actually looking at the logs `prismaClientDir` looks legit 🤔 
```
prisma:generator  prismaClientDir /Users/aqrln/prisma/prisma/packages/client/src/__tests__/generation/node_modules/@prisma/client 
```

packages/client/tests/e2e/default-version/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
aqrln, line 1: this one was downgraded from 6.1 somehow? all others were on 5.4, and I ran `pnpm install` with latest 8.6.7

.github/workflows/test-template.yml
```diff
@@ -68,7 +68,7 @@ jobs:
 
       - uses: pnpm/action-setup@v2.2.4
         with:
-          version: 7
+          version: 8
```
SevInf, line 71: Does not to change, just a suggestion: maybe we could move this value to the `inputs` and update it in a single place next time?

SevInf (approved):

PR #20161, opened 2023-07-10
fix(client): types oom and inferred length errors

### LINKED ISSUE #16536: `$extends` TS error: "Inferred type of this node exceeds the maximum length the compiler will serialize" with `"declaration": true` in `tsconfig`
`The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.`

This error also showed up in another project.

Screenshot from 2022-11-30 12-46-59

[Source file](https://github.com/codeledge/entitree-monorepo/blob/main/apps/entitree-images/prisma/prismaClient.ts). To reproduce, check out main, and upgrade to Prisma 4.7

### LINKED ISSUE #19999: 4.16.1 breaks type check
### Bug description

This will probably get dismissed because I'm not currently able to strip down from our repository what exactly is causing this. But from 4.16.0 -> 4.16.1 I'm not able to use next build because something on prisma is causing a cyclic type loop that is exhausting v8 memory.

I also noticed that types in some occasions (only happened in one model) stopped working. I can't go to type definition nor autocomplete. Don't dismiss it yet, I'm trying to reproduce what's causing this. I created this bug report in case anyone has the same issue (couldn't find others) and was able to recreate it in minimal reproduction.

**Rolling back to 4.16.0 fixes this issue**

An example of the error thrown while type checking:
```
<--- Last few GCs --->

[1960:00000221C0ABAC30]    85531 ms: Mark-sweep 4048.5 (4136.9) -> 4033.6 (4138.1) MB, 1230.4 / 0.0 ms  (average mu = 0.094, current mu = 0.025) allocation failure scavenge might not succeed
[1960:00000221C0ABAC30]    86821 ms: Mark-sweep 4049.6 (4138.1) -> 4034.9 (4139.4) MB, 1236.3 / 0.0 ms  (average mu = 0.068, current mu = 0.042) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
- info Linting and checking validity of types . 1: 00007FF6F615B34F v8::internal::CodeObjectRegistry::~CodeObjectRegistry+123599
 2: 00007FF6F60E8CB6 v8::internal::MicrotaskQueue::GetMicrotasksScopeDepth+65206
 3: 00007FF6F60E9D8D node::OnFatalError+301
 4: 00007FF6F6A1C36E v8::Isolate::ReportExternalAllocationLimitReached+94
 5: 00007FF6F6A0694D v8::SharedArrayBuffer::Externalize+781
 6: 00007FF6F68A9F2C v8::internal::Heap::EphemeronKeyWriteBarrierFromCode+1468
 7: 00007FF6F68B6BD9 v8::internal::Heap::PublishPendingAllocations+1129
 8: 00007FF6F68B3BAA v8::internal::Heap::PageFlagsAreConsistent+2842
 9: 00007FF6F68A6809 v8::internal::Heap::CollectGarbage+2137
10: 00007FF6F68A49C0 v8::internal::Heap::AllocateExternalBackingStore+2000
11: 00007FF6F68C9546 v8::internal::Factory::NewFillerObject+214
12: 00007FF6F65FBCE5 v8::internal::DateCache::Weekday+1797
13: 00007FF6F6AAA061 v8::internal::SetupIsolateDelegate::SetupHeap+494417
14: 00000221C2DC6AD0
error Command failed with exit code 1.
```

### Expected behavior

_No response_

### Prisma information

```prisma

generator client {
    provider = "prisma-client-js"
}

datasource db {
    provider = "postgresql"
}
```

Somehow I believe the extends may have something to do with this.
```ts
import { PrismaClient } from "@prisma/client"

import { env } from "@/helpers/env.mjs"

import type { Prisma} from "@prisma/client"

type PaginationArgs<Input, Output extends Record<string, any> = {}> = {
	page?: number;
	limit?: number;
	sort?: "asc" | "desc";
	order?: string;

	map?: (data: Input) => Output | Promise<Output>;
}

function getPrisma () {
	return (new PrismaClient({
		log: env.NODE_ENV === "development" ? ["error", "warn"] : ["error"],
	})).$extends({
		name: "paginate",
		model: {
			$allModels: {
				async paginate<
					T,
					Args,
					Result extends Prisma.Result<T, Args, "findMany">,
					Options extends PaginationArgs<Result[number]>,
				>(this: T, args: Prisma.Exact<Args, Omit<Prisma.Args<T, "findMany">, "cursor" | "take" | "skip">> & Options) {
					const { page = 1, limit = 25, select, where, order, sort } = (args || {}) as any

					return Promise.all([
						(this as any).findMany({
							select,
							where,
							skip: (page - 1) * limit,
							take: limit,
							...(order ? { orderBy: { [order]: sort || "asc" } } : {}),
						}).then(function (result: any) {
							return args.map ? Promise.all(result.map(args.map)) : result
						}),
						(this as any).aggregate({
							_count: { id: true },
							take: limit,
							select: { _count: true },
							where,
							skip: (page - 1) * limit,
						}).then((response: any) => ({
							page,
							pages: Math.ceil(response._count.id / limit),
							total: response._count.id,
							limit,
						})),
					]) as Promise<[
						Options["map"] extends Function ? Awaited<ReturnType<Options["map"]>>[] : Result,
						{ page: number; pages: number; total: number; limit: number },
					]>
				},
			},
		},
	})
}

const globalForPrisma = globalThis as unknown as { prisma: ReturnType<typeof getPrisma> }

export const prisma =
	globalForPrisma.prisma || getPrisma()

if (env.NODE_ENV !== "production") globalForPrisma.prisma = prisma
```

### Environment & setup

- OS: Windows 10
- Database: PostgreSQL
- Node.js version: v16.17.1

### Prisma Version

```
prisma                  : 4.16.1
@prisma/client          : 4.16.1
Query Engine (Node-API) : libquery-engine b20ead4d3ab9e78ac112966e242ded703f4a052c (at node_modules\@prisma\engines\query_engine-windows.dll.node)    
Migration Engine        : migration-engine-cli b20ead4d3ab9e78ac112966e242ded703f4a052c (at node_modules\@prisma\engines\migration-engine-windows.exe)
Format Wasm             : @prisma/prisma-fmt-wasm 4.16.0-66.b20ead4d3ab9e78ac112966e242ded703f4a052c
Default Engines Hash    : b20ead4d3ab9e78ac112966e242ded703f4a052c
```

millsp:
closes https://github.com/prisma/prisma/issues/16536
closes https://github.com/prisma/prisma/issues/19999

(maybe related https://github.com/prisma/prisma/issues/20308)

[Internal Slack](https://prisma-company.slack.com/archives/C043X6JBND6/p1689020469486759)


packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
```diff
@@ -1000,25 +1000,25 @@ export namespace Prisma {
...
+  export import Extension = $Extensions.UserArgs
   export import getExtensionContext = runtime.Extensions.getExtensionContext
-  export type Args<T, F extends $Public.Operation> = $Public.Args<T, F>
-  export type Payload<T, F extends $Public.Operation> = $Public.Payload<T, F>
-  export type Result<T, A, F extends $Public.Operation> = $Public.Result<T, A, F>
-  export type Exact<T, W> = $Public.Exact<T, W>
+  export import Args = $Public.Args
+  export import Payload = $Public.Payload
+  export import Result = $Public.Result
+  export import Exact = $Public.Exact
```
millsp, line 1021: Important part of the fix. I prevents the expansion of our types into the user's projects - does not work for recursive types. With this, a compiled user project using an extension would output types like

```ts
    $allModels: {
        paginate<T, A>(this: T, args: Prisma.Exact<A, Prisma.Args<T, "findMany">>): Promise<Prisma.Result<T, A, "findMany">>;
    };
```

With the previous version, tsc will actually try to inline the types with they actual implementation and potentially resolve them in-place depending on their input. This would be great if it did not actually produce a cryptic error message and could inline properly. tl;dr; only re-export via `export import A = B`.

> The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -48,13 +48,12 @@ export type FluentOperation =
 type Count<O> = { [K in keyof O]: Count<number> } & {}
 
 // prettier-ignore
-export type GetFindResult<P extends Payload, A> =
-  {} extends A ? DefaultSelection<P> :
-  A extends 
+export type GetFindResult<P extends Payload, _A, A = { 0: _A, 1: { select: undefined } }[Equals<_A, any>]> =
+  A extends
```
millsp, line 52: The removed condition `{} extends A ? DefaultSelection<P> :` caused a very huge performance hit. Instead we now have an ugly comparison to take some steps when args `A` equals `any`, `{}` worked but was too wide-matching, I guess.

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -82,31 +81,25 @@ type SelectablePayloadFields<K extends PropertyKey, O> =
...
-      : P[K]
-} & unknown
+export type UnwrapPayload<P> = {} extends P ? unknown : {
+  [K in keyof P]: 
+    P[K] extends { scalars: infer S, composites: infer C }[]
+    ? Array<S & UnwrapPayload<C>>
+    : P[K] extends { scalars: infer S, composites: infer C } | null
+      ? S & UnwrapPayload<C> | Select<P[K], null>
+      : never
+};
```
millsp, line 102: This implementation should be more light weight as it does less comparisons similarly to `GetFindResult` does. It improved things significantly on the user-reproduction since immediate evaluation is guarded by `infer`.

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
millsp: Actually: Apparently this is not valid in TypeScript, although it produces the desired outcome once compiled.

millsp: So I've used a workaround that we used for this same issue in the past. Intersect the type with ` & unknown` prevents TS from performing type inlining in dependent projects. Ideally, we would want something like this https://github.com/microsoft/TypeScript/issues/4336, but the current "[solution](https://github.com/microsoft/TypeScript/issues/4336#issuecomment-781862959)" we also use does not work with types, which was an assumption in my first comment.

millsp: Related slack thread https://prisma-company.slack.com/archives/C043X6JBND6/p1689103849934459

millsp:
Will add tests once I got the greenlight/feedback from the users.

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
millsp: Decided to use the first approach, and suppress the error message (which only happen in watch mode locally). The error is never suppressed on CI.

millsp:
Needs https://github.com/prisma/prisma/pull/20285

helpers/compile/plugins/tscPlugin.ts
```diff
@@ -92,6 +92,17 @@ export const tscPlugin: (emitTypes?: boolean) => esbuild.Plugin = (emitTypes?: b
...
           // we get the types generated by tsc and bundle them near the output
           bundleTypeDefinitions(typeOutPath, bundlePath)
+
+          // when types are already exported by us, it is not always guaranteed
+          // that the type bundler will also export them. It won't be able to
+          // trace them correctly back to runtime module and fail with either
+          // "... is using the type X but can not be named" or "The inferred
+          // type of this node exceeds the maximum length" error, or even worse
+          // the compiler could inline the types in the users' project causing
+          // very bad performance issues. To fix this, we export all the types.
```
millsp, line 102: This is really really bad. This means that when it did not fail, TS was potentially inlining our non-exported types utils into user projects (given that they are using or re-exporting a dependent type). This, as I found out, and as was written in a previous comment could not only cause errors in extreme cases but also slow down compilation...

packages/client/src/runtime/core/types/Utils.ts
```diff
@@ -34,21 +34,12 @@ export type Narrow<A> = {
 } | (A extends Narrowable ? A : never)
 
 // prettier-ignore
-export type Exact<A, W> = (W extends A ? {
+export type Exact<A, W> = (A extends unknown ? (W extends A ? {
   [K in keyof W]: K extends keyof A ? Exact<A[K], W[K]> : never
-} : W) | (A extends Narrowable ? A : never)
+} : W) : never) | (A extends Narrowable ? A : never)
```
millsp, line 39: Fix from another PR needed here.

packages/client/src/runtime/index.ts
```diff
@@ -35,16 +33,6 @@ export { empty, join, raw, Sql, default as sqltag } from 'sql-template-tag'
...
 export { Extensions }
 export { Public }
-export { warnOnce } from '@prisma/internals'
-
-/**
- * Payload, PrismaPromise and Extensions types are already exported via Types but tsc
- * won't be able to trace them correctly back to runtime module and fail with either
- * "... is using the type X but can not be named" or "The inferred type of this node exceeds the maximum length" error.
- * The issue lies with the type bundler which does not add exports for dependent types
- * TODO: Maybe simply exporting all types in runtime will do the trick
```
millsp, line 45: It's what we are doing now, exporting all bundled types.

packages/client/tests/e2e/issues/19999-tsc-extensions-oom/src/index.ts
millsp, line 1: These are all the examples provided by the users.

packages/client/src/generation/TSClient/common.ts
```diff
@@ -183,12 +184,12 @@ export type MetricHistogramBucket = runtime.MetricHistogramBucket
...
+export import Extension = $Extensions.UserArgs
 export import getExtensionContext = runtime.Extensions.getExtensionContext
-export type Args<T, F extends $Public.Operation> = $Public.Args<T, F>
-export type Payload<T, F extends $Public.Operation> = $Public.Payload<T, F>
-export type Result<T, A, F extends $Public.Operation> = $Public.Result<T, A, F>
-export type Exact<T, W> = $Public.Exact<T, W>
+export import Args = $Public.Args
+export import Payload = $Public.Payload
+export import Result = $Public.Result
+export import Exact = $Public.Exact
```
jkomyno, line 192: Is this `export import T1 = T2` desired?

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
SevInf: Can it conflict with user defined models now? I.e. is it still possible to name your model `Result`? Could you add those names [here](https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/naming-conflict/_builtInNames.ts)  to verify?

packages/client/src/runtime/core/types/Utils.ts
```diff
@@ -118,6 +109,9 @@ export type PayloadToResult<P, O extends Record<any, any> = RenameAndNestPayload
 
 export type Select<T, U> = T extends U ? T : never
 
+// prettier-ignore
+export type Equals<A, B> = (<T>() => T extends A ? 1 : 2) extends (<T>() => T extends B ? 1 : 2) ? 1 : 0
```
SevInf, line 113: Could you explain what this type is doing and add this in a comment? I have really hard time grasping it.

packages/client/src/runtime/core/types/Result.ts
```diff
@@ -82,31 +81,25 @@ type SelectablePayloadFields<K extends PropertyKey, O> =
...
-        : P[K]
-      : P[K]
-} & unknown
+type UnwrapPayload<P> = {} extends P ? unknown : {
+  [K in keyof P]: 
+    P[K] extends { scalars: infer S, composites: infer C }[]
+    ? Array<S & UnwrapPayload<C>>
+    : P[K] extends { scalars: infer S, composites: infer C } | null
+      ? S & UnwrapPayload<C> | Select<P[K], null>
+      : never
```
SevInf, line 101: Is case for non nullable `P[K]` missing here?

millsp: I think it would be covered since unions are distributed, wdyt? The simplest example I have in mind something like:

```ts
type Test<O> = {
  [K in keyof O]: O[K] extends number | null ? true : false
}

type result = Test<{ a: null, b: number, c: number | null }>
```

packages/client/src/runtime/core/types/Utils.ts
millsp: It comes from another PR see discussion https://github.com/prisma/prisma/pull/20180#discussion_r1270060230

packages/client/src/runtime/core/types/Result.ts
millsp: Yes, better to test. I'll check and add as needed.

packages/client/src/generation/TSClient/common.ts
millsp: Yes, it's a syntax for re-exporting things in namespaces without having to re-define it like it was done previously.

packages/client/src/runtime/core/types/Utils.ts
millsp: It comes from a former Microsoft employee
https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
millsp: I guess before and after will both conflict, I'll write a test.

millsp: Turns out no there's no conflict. Did I understand correctly that there was no conflict because `Result` is in `Prisma` namespace and the model type definition was top-level?

millsp: Tests added.

packages/client/src/runtime/core/types/Utils.ts
millsp: It abuses some of the variance rules in covariant positions to assert that a type is strictly equal to another.

packages/client/src/runtime/core/types/Result.ts
millsp: We have a test for a nested required composite [here](https://github.com/prisma/prisma/blob/9bab0b986d7e021c86069a42c1a950788891bcf9/packages/client/tests/functional/composites/selection/tests.ts/#L101)

millsp: I forgot to mention in the PR, I wanted to point out that sometimes when recursive types shows up this way it can be a sign that the compiler is too eagerly evaluating and can (not always, but is a good proxy when developing) be related to other performance issues/responsiveness down the line. Also, when you hover over the type, that experience should be reflected there as well.

This is the previous implementation as you and I implemented it
image

This is how the refactored/proposed replacement is evaluated (well, not over-evaluated)
image

millsp: (I have started working on type performance tests btw, but work is not yet ready)

packages/client/src/__tests__/integration/happy/exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
SevInf: >Turns out no there's no conflict. Did I understand correctly that there was no conflict because Result is in Prisma namespace and the model type definition was top-level?

This, and also within `Prisma` namespace we never refer to model type by the name anymore.

SevInf (approved):

PR #20180, opened 2023-07-12
fix(client): types for mixed input type selections

### LINKED ISSUE #19962: Prisma Client: wrong type when using spread operator to set default values on query args
### Bug description

I'm using spread operator to set default values on query args of my findMany() function, but when I use it the generated type is incorrect.

### How to reproduce

Repository class:
```ts
class ComponentRepository {
  /* ... */

  findAll<T extends Prisma.ComponentFindManyArgs>(
    params?: Prisma.SelectSubset<T, Prisma.ComponentFindManyArgs>,
  ) {
    return this.prisma.findMany({
      // default `include`
      include: { componentProp: true, componentStyle: true },
      ...params,
    });
  }

  /* ... */
}
```

If I pass only `params` as args of `findMany` the type of `component.componentStyle` (see below) is correct.

```ts
class ComponentRepository {
  /* ... */

  findAll<T extends Prisma.ComponentFindManyArgs>(
    params?: Prisma.SelectSubset<T, Prisma.ComponentFindManyArgs>,
  ) {
    return this.prisma.findMany(params);
  }

  /* ... */
}
```

### Expected behavior

`componentStyle` need exists in `component.componentStyle`. 

### Prisma information

```ts
const componentWithImages = await this.componentRepository.findAll({
  where: {
    overlayId,
    id: {
      notIn: params.map(({ id }) => id),
    },
    componentStyle: {
      some: { key: 'backgroundImage' },
    },
  },
  // override include
  include: { componentStyle: true },
});

for (const component of componentWithImages) {
  /**
    * error TS2339: Property 'componentStyle' does not exist on type '({ componentProp: (GetResult<{ componentId: string; key: string; value: string; variableAssign: string | null; type: ComponentPropType; }, unknown> & {})[]; componentStyle: (GetResult<...> & {})[]; } | {} | { ...; } | undefined) & GetResult<...> & {}'.
    */
  const style = component.componentStyle.find(
    (s) => s.key === 'backgroundImage',
  );

  /* ... */
}
```

### Environment & setup

- OS: WSL 2 Ubuntu 22.04.1 LTS
- Database: PostgreSQL
- Node.js version: v18.14.0

### Prisma Version

```
prisma                  : 4.16.1
@prisma/client          : 4.16.1
Query Engine (Node-API) : libquery-engine b20ead4d3ab9e78ac112966e242ded703f4a052c (at ../../node_modules/.pnpm/@prisma+engines@4.16.1/node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Migration Engine        : migration-engine-cli b20ead4d3ab9e78ac112966e242ded703f4a052c (at ../../node_modules/.pnpm/@prisma+engines@4.16.1/node_modules/@prisma/engines/migration-engine-debian-openssl-3.0.x)
Format Wasm             : @prisma/prisma-fmt-wasm 4.16.0-66.b20ead4d3ab9e78ac112966e242ded703f4a052c
Default Engines Hash    : b20ead4d3ab9e78ac112966e242ded703f4a052c
```

### LINKED ISSUE #20093: Using `Prisma.XyzFindManyArgs` breaks `findMany` typing in v4.16.0+
### Bug description

We sometimes generate find options dynamically and use the `Prisma.UserFindManyArgs` type for this, e.g. in a case like this:

```ts
    const userFindMany: Prisma.UserFindManyArgs = {};

    if(true) { // Some condition is met
      userFindMany.where = { email: 'mail@example.com' };
      userFindMany.take = 1;
    }

    const users = await prismaClient.user.findMany({
      ...userFindMany,
      include: {
        userRoles: true,
      },
    });

    console.log(users[0].email); // This works
    console.log(users[0].userRoles); // But this doesn't
```

Starting with Prisma v4.16.0, the code above throws a typing error: `TS2339: Property 'userRoles' does not exist on type [...]`. I think this is probably caused by the new typing because of client extensions - but in my opinion, the code above should still work.

As a workaround, it's possible to write a custom type for `userFindManyArgs`, e.g. `const userFindMany: { where?: Prisma.UserWhereInput; take?: number } = {};`, but that's only a workaround for a breaking change.

### How to reproduce

Use the code above, check if `users[0].userRoles` throws a typescript error.

### Expected behavior

Same behavior as before - adding an include should always include the relations in the result typing.

### Prisma information

```prisma
// Add your schema.prisma
model User {
  email     String   @unique
  userRoles               UserRole[]
}

model UserRole {
}
```

```ts
const prismaClient = new PrismaClient();
```

### Environment & setup

- OS: Windows
- Database: PostgreSQL
- Node.js version: v16.17.1
- Typescript version: v5.1.6

### Prisma Version

```
Query Engine (Node-API) : libquery-engine 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules\@prisma\engines\query_engine-windows.dll.node)
Migration Engine        : migration-engine-cli 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules\@prisma\engines\migration-engine-windows.exe)
```

millsp:
closes https://github.com/prisma/prisma/issues/20093
closes https://github.com/prisma/prisma/issues/19962


packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -48,21 +48,19 @@ export type FluentOperation =
...
-export type GetFindResult<P extends Payload, A> =
-  {} extends A ? DefaultSelection<P> :
-  A extends 
-  | { select: infer S } & Record<string, unknown>
-  | { include: infer S } & Record<string, unknown>
-  ? S extends undefined ? DefaultSelection<P> :
-    {
-      [K in keyof S as S[K] extends false | undefined | null ? never : K]:
-        S[K] extends object
+export type GetFindResult<P extends Payload, _A, A = { 0: _A, 1: { select: undefined } }[Equals<_A, any>]> =
```
millsp, line 51: `{ 0: _A, 1: { select: undefined } }[Equals<_A, any>]` this means if `_A` equals `any` then return `{ select: undefined }` otherwise return `_A`.

packages/client/tests/functional/extensions/result.ts
```diff
@@ -380,4 +380,34 @@ testMatrix.setupTestSuite(() => {
       expectTypeOf(data).toEqualTypeOf<User>()
     }
   })
+
+  test('when args have both include and select and one of them is optional, result includes both', () => {
```
millsp, line 384: This is probably incorrect anyways if we would think on a strict type-level, but the expected current behavior that was broken.

packages/client/tests/e2e/ts-version/4.4/package.json
millsp, line 1: https://github.com/prisma/prisma/pull/20180/files#diff-1c07144a6146c085293750cf7f2b615ef7eb7722eeee82e6e228bfe521898781R53 is not supported by TS 4.4 to 4.6 included.

packages/client/src/runtime/core/types/Utils.ts
```diff
@@ -118,6 +118,19 @@ export type PayloadToResult<P, O extends Record<any, any> = RenameAndNestPayload
...
+export type Or<A extends 1 | 0, B extends 1 | 0> = {
+  0: {
+    0: 0
+    1: 1
+  }
+  1: {
+    0: 1
+    1: 1
+  }
+}[A][B]
```
jkomyno, line 132: [Link to Playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDyUA8BBOobAB2AJgM5wCMcAPnAAwA0cAQniASeVbXQHxwBeOAG8AUHHoAuEeIlT6siRWkVZAX1nKZc+ap1a9cDWoDa2ALonm50aKQo4AJWCkArgBt4QjJkYVeQA) for reviewers.

packages/client/src/runtime/core/types/Utils.ts
```diff
@@ -118,6 +118,19 @@ export type PayloadToResult<P, O extends Record<any, any> = RenameAndNestPayload
 
 export type Select<T, U> = T extends U ? T : never
 
+export type Equals<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? 1 : 0
```
jkomyno, line 121: Does it come from [here](https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650)?

packages/client/src/runtime/core/types/Utils.ts
millsp: It's not used atm but decided to leave it anyways, as it can be a useful type anyways.

packages/client/src/runtime/core/types/Utils.ts
millsp: I had it in ts-toobelt, but it really originally was created by a former Microsoft employee that worked on TS

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -48,21 +48,19 @@ export type FluentOperation =
...
+  A extends
+  | { select: infer S extends object } & Record<string, unknown>
+  | { include: infer I extends object } & Record<string, unknown>
```
jkomyno, line 54: Good that `infer` now has two separate type variables

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -48,21 +48,19 @@ export type FluentOperation =
...
+  ? {
+      [K in keyof S | keyof I as (S & I)[K] extends false | undefined | null ? never : K]:
```
jkomyno, line 56: Would this work / be less ambiguous?

```suggestion
      [K in (keyof S | keyof I) as (S & I)[K] extends false | undefined | null ? never : K]:
```

jkomyno (approved):
Approved with a couple of questions

packages/client/src/runtime/core/types/GetResult.ts
millsp: fyi, this is more a perf improvement, so I moved it out of the PR to https://github.com/prisma/prisma/pull/20161

millsp: (I left the two utilities though)

packages/client/src/runtime/core/types/GetResult.ts
```diff
@@ -50,19 +50,18 @@ type Count<O> = { [K in keyof O]: Count<number> } & {}
...
-        S[K] extends object
+  A extends
+  | { select: infer S extends object } & Record<string, unknown>
+  | { include: infer I extends object } & Record<string, unknown>
+  ? {
+      [K in keyof S | keyof I as (S & I)[K] extends false | undefined | null ? never : K]:
+        (S & I)[K] extends object
         ? P extends SelectablePayloadFields<K, (infer O)[]>
-          ? O extends Payload ? GetFindResult<O, S[K]>[] : never
+          ? O extends Payload ? GetFindResult<O, (S & I)[K]>[] : never
```
SevInf, line 60: I am curious if that will now produce unexpected result:

```ts
include: {
   field: true
}
select: {
   field: false
}
```

Since we can't intersect `true & false`?

packages/client/src/runtime/core/types/Utils.ts
```diff
@@ -118,6 +118,20 @@ export type PayloadToResult<P, O extends Record<any, any> = RenameAndNestPayload
 
 export type Select<T, U> = T extends U ? T : never
 
+// prettier-ignore
+export type Equals<A, B> = (<T>() => T extends A ? 1 : 2) extends (<T>() => T extends B ? 1 : 2) ? 1 : 0
```
SevInf, line 122: Does not seem to be used?

packages/client/src/runtime/core/types/GetResult.ts
millsp: This case is in theory protected against on user input when both are provided and not optional. However, that does not apply to utility types. In that case, as you pointed out, the intersection is non-valid and the whole mapped type has no keys to map on, and resolves to `{}`. Then because `include` is present, the default selection is added to it.

packages/client/src/runtime/core/types/Utils.ts
millsp: Yes, it was left over from related work that will be merged soon.

packages/client/src/runtime/core/types/GetResult.ts
millsp: Not sure if this is something to test or document. Perhaps we should test it, as it is currently a bit undefined in behavior. Or perhaps it does not matter because it is invalid? wdyt

SevInf: I think we sort of have a test for `select` + `include` being invalid [here](https://github.com/prisma/prisma/blob/19fd17d21c9e2e49e5b4640deb8b355dbbe1246f/packages/client/tests/functional/query-validation/tests.ts#L10) and `@ts-expect-error` ensures we have a compile-time error as well, so I think we'd be fine without `GetResult`-specific test.

SevInf (approved):

PR #20181, opened 2023-07-12
chore: bump `prettier` and `eslint-plugin-prettier`

mrazauskas:
Closes #20082
Closes #20179

`prettier` and `eslint-plugin-prettier` should be updated together. Since this is major version of `prettier`, there are few changes in formatting.


.prettierrc.yml
```diff
@@ -1,5 +1,4 @@
 tabWidth: 2
-trailingComma: all
```
mrazauskas, line 2: Shipped as new default: https://prettier.io/docs/en/options.html#trailing-commas

Jolg42 (approved):
Thanks @mrazauskas for the PR 💚

mrazauskas:
@Jolg42 Sorry, I forgot one detail. Current, if Prettier v3 is installed, Jest fails to update inline snapshots. There are two solutions:

- downgrade Prettier to v2
- add `prettierPath: null` to Jest config. This will make sure Jest does not run Prettier, but new inline snapshots should be formatted outside Jest.

What would you prefer?

As you see in CI, all works with existing snaps. The problem accrues if they have to be updated or created.

Jolg42:
Ouch 😢 

It sounds to me like reverting is easier/best until this is figured out then.

Jolg42:
PR created from "Revert" UI button https://github.com/prisma/prisma/pull/20182

mrazauskas:
> It sounds to me like reverting is easier/best until this is figured out then.

Agreed. I will come back after the issue is fixed on Jest side.

Jolg42:
Thank you!

Jolg42:
Now Jest will error in next version
https://github.com/jestjs/jest/pull/14367

Tracking issue to monitor
https://github.com/jestjs/jest/issues/14305

PR #20194, opened 2023-07-12
fix(client): Enum type clashes

### LINKED ISSUE #20031: Generated enum type is different when model name changes
### Bug description

I created an enum called `LogLevel` and used in my model. Model's create type is different than LogLevel enum type. So, this:

```ts
  await prismaClient.log.create({
    data: {
      level: LogLevel.INFO
    }
  });
```

fails typescript with `Type '"INFO"' is not assignable to type 'LogLevel'. Did you mean '"info"'?ts(2820)`

---

When I change enum name to `Level` types are correct.

### How to reproduce

1. Create minimal ts project with just index.ts
2. Install `prisma` and `@prisma/client` and init + create + migrate enum and model

```schema.prisma
enum LogLevel {
  INFO
  WARN
}

model Log {
  id    Int      @id @default(autoincrement())
  level LogLevel
}
```

4. Write code in index.ts

```ts
import { LogLevel, PrismaClient } from '@prisma/client'

(async () => {
  const prismaClient = new PrismaClient();

  await prismaClient.log.create({
    data: {
      level: LogLevel.INFO
    }
  });
})()
```

6. Try to build ts and see error `Type '"INFO"' is not assignable to type 'LogLevel'. Did you mean '"info"'?ts(2820)`
7. Optionally, remove migration, change enum's name to `Level` and migrate + generate again to see no error with `Level.INFO` now.

### Expected behavior

Generated types should be correct regardless `enum`'s name

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
}

enum Level {
  INFO
  WARN
}

model Log {
  id    Int      @id @default(autoincrement())
  level Level
}

```

```ts
import { Level, PrismaClient } from '@prisma/client'

(async () => {
  const prismaClient = new PrismaClient();

  await prismaClient.log.create({
    data: {
      level: Level.INFO
    }
  });
})()

```

### Environment & setup

- OS: Kubuntu 22.04, Kernel version: 5.15.0-76-generic (64-bit)
- Database: PostgreSQL
- Node.js version: 18.16.0

### Prisma Version

```
Query Engine (Node-API) : libquery-engine 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Migration Engine        : migration-engine-cli 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules/@prisma/engines/migration-engine-debian-openssl-3.0.x)
```

SevInf:
Follow up to #20150.
Changes we did to extensions have fixed name clashes for models, but not
for enums. Enums problem is more complicated: in case of name clash, it
is impossible for contents of `Prisma` namespace to reference the value
from outer scope.
Solution is: all top-level enums are put into $Enums namespace. Types
within `Prisma` namespace always refer to them using fully-qualifed
name. Top-level enums are re-exported from `$Enums` namespace for
backward compatiblity.

Also, does opportunistic clean-ups/refactors

Fix #20031


packages/client/src/generation/TSClient/Enum.ts
```diff
@@ -23,7 +23,10 @@ export class Enum implements Generatable {
 ${indent(type.values.map((v) => `${v}: ${this.getValueJS(v)}`).join(',\n'), TAB_SIZE)}
 }`
     const enumBody = this.isStrictEnum() ? `makeStrictEnum(${enumVariants})` : enumVariants
-    return `exports.${this.useNamespace ? 'Prisma.' : ''}${type.name} = ${enumBody};`
+
+    return this.useNamespace
+      ? `exports.Prisma.${type.name} = ${enumBody};`
+      : `exports.${type.name} = exports.$Enums.${type.name} = ${enumBody};`
```
millsp, line 29: This seems like it brings a new feature to the runtime, should we document it, or is it only internal?

millsp (approved):

packages/client/src/generation/TSClient/Enum.ts
SevInf: Not sure how useful this feature would be to the general public, but we can indeed document it.

PR #20213, opened 2023-07-13
fix(createPrismaPromise): Fix loss of AsyncLocalContext within query extensions under Cloudflare Workers

### LINKED ISSUE #20104: Accessing `AsyncLocalStorage` within a query client extension (`$allOperations`) breaks under Vercel Edge functions/Cloudflare Workers
### Bug description

TL;DR: 
1. calling `.getStore()` on an `AsyncLocalStorage` (from `node:async_hooks`)..
2. ..*from within a Prisma client `query:` extension*..
3. ..(but not elsewhere, including a custom `client:` method within the same Prisma client extension) 
4. ..is always returning `undefined` for me (instead of returning the contents of the async store)
5. ..*when deployed to Vercel Edge functions (which runs on Cloudflare Workers)*
6. ...(but works locally + under Node.JS)

~

I'm using `AsyncLocalStorage` to pass request context around (current userId, etc).
I've been using this successfully within a Prisma Client extension to implement Row-level security with Postgres.

Works just fine under Node.JS - but I hit a weird bug(?) trying to make the same code work under the Vercel Edge functions runtime (which runs on top of Cloudflare Workers).

Vercel Edge / Cloudflare workers does support AsyncLocalStorage - and I've confirmed the same async storage is working in other parts of my code. But fails to load the async context store from within a Prisma client query extension.
(See illustrative code below.)

Sorry this isn't a good reproduction (yet) - and a very specific set of circumstances that might be caused by an incompatibility within Prisma, but may potentialy be a bug within Cloudflare workers' implementation of `AsyncLocalStorage`.

It's been driving me crazy trying to diagnose and I want to get a record written up here in case anyone else runs into this and knows something!

### How to reproduce

..illustrative code snippet for now (sorry - just wanted to start getting this out there); I'll debug further and update when I can, ideally with a MRE.

```typescript
import { AsyncLocalStorage } from "node:async_hooks";

globalThis.storage = new AsyncLocalStorage<{whatever: string}>;

const exampleExtension = Prisma.defineExtension((prisma) => {
  return prisma.$extends({
    client: {
      clientMethodsWork() {
        // within client methods works - value of `whatever` is retrieved from async store as expected
        console.log(globalThis.storage.getStore()?.whatever);
      }
    },
    query: {
      $allOperations(params) {
        // within query extension doesn't work when deployed to Vercel Edge functions (on Cloudflare Workers)
        // (i.e. using Edge/dataproxy client) - getStore() returns `undefined`
        // - works locally (specifically: under Next.JS simulation of Vercel Edge runtime)
        // - works when deployed under Node.JS runtime
        console.log(globalThis.storage.getStore()?.whatever);
        // ...
        // return query(args); etc
      }
    }
  });
});

// ...

const prisma = new PrismaClient().$extends(exampleExtension);

globalThis.storage.run({whatever: "some state to pass async"}, () => {
  // any method..
  prisma.exampleModel.findUnique(...);
});
```

### Expected behavior

Reading from AsyncLocalStorage should work anywhere within the async call stack, including within custom query client extensions like `$allOperations`.

### Prisma information

(I don't believe this is related to the schema or specific query usage.)

### Environment & setup

- OS: Vercel Edge Functions / Cloudflare Workers
- Database: N/A
- Node.js version: N/A

### Prisma Version

```
Query Engine (Node-API) : libquery-engine 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 4bc8b6e1b66cb932731fb1bdbbc550d1e010de81 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
```

andyjy:
Confirmed this patch fixes #20104 🎉  

May also affect Deno (unconfirmed)

This demonstrates the fix but would require further testing and ecosystem tests(?) before ready to merge.

I'm unsure about the addition of the `node:async_hooks` dependency and how that may interop with the build of the Prisma runtime for different Javascript runtimes (Edge, Deno). Next.js bundles the import from `node:async_hooks` for the Vercel Edge runtime just fine, but I'm not sure whether introducing this import may complicate/break things for users using custom bundler setups.

I believe this change has no impact under Node.js, based on reasoning about what it does - and anecdotally it doesn't break my project. But it's also not required and could be implemented for the Prisma Edge runtime only.

This is only required as a workaround for "incomplete" implementation of AsyncLocalStorage under Cloudflare Workers not passing the context through thenables. (And I suspect potentially Deno also - not checked). It would no longer be required if/when this is supported by these runtimes.

Also aware that if I'm the only reporter of this issue, it may be this is something to keep in draft if/until other users run into problems trying to use AsyncLocalContext within Prisma Query Extensions under edge runtimes.

### DIFF
--- a/packages/client/src/runtime/core/request/createPrismaPromise.ts
+++ b/packages/client/src/runtime/core/request/createPrismaPromise.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from "node:async_hooks"
 import type { PrismaPromise, PrismaPromiseTransaction } from './PrismaPromise'
 
 export type PrismaPromiseCallback = (transaction?: PrismaPromiseTransaction) => PrismaPromise<unknown>
@@ -20,6 +21,12 @@ export type PrismaPromiseFactory = (callback: PrismaPromiseCallback) => PrismaPr
  */
 export function createPrismaPromiseFactory(transaction?: PrismaPromiseTransaction): PrismaPromiseFactory {
   return function createPrismaPromise(callback) {
+    // bind the current Async Context to the callback function
+    // to ensure any use of AsyncLocalStorage.getStore() within
+    // is not broken by the implementation of PrismaPromise as a 'thenable'
+    // => Cloudflare Workers's implementation of AsyncLocalStorage does not
+    // currently pass async context within thenables
+    callback = AsyncLocalStorage.bind(callback);
     let promise: PrismaPromise<unknown> | undefined
     const _callback = (callbackTransaction = transaction) => {
       try {
--- a/packages/client/src/runtime/core/request/createPrismaPromise.ts
+++ b/packages/client/src/runtime/core/request/createPrismaPromise.ts
@@ -24,9 +24,29 @@ export function createPrismaPromiseFactory(transaction?: PrismaPromiseTransactio
     // bind the current Async Context to the callback function
     // to ensure any use of AsyncLocalStorage.getStore() within
     // is not broken by the implementation of PrismaPromise as a 'thenable'
+    //
     // => Cloudflare Workers's implementation of AsyncLocalStorage does not
-    // currently pass async context within thenables
-    callback = AsyncLocalStorage.bind(callback);
+    // currently pass async context within thenables:
+    // https://github.com/cloudflare/workerd/issues/870
+    //
+    // we only do this when we detect Vercel Edge functions or Cloudflare Workers
+    // since it has the side-effect of "locking" the async context to that set at
+    // the time the PrismaPromise is created. This alters the behaviour to prevent
+    // using AsyncLocalContext.run() to pass a new async context from within a
+    // Prisma Query Extension itself. Likely rare - but example use-case here:
+    // https://github.com/andyjy/prisma-local-proxy/blob/35494f512e9968a9bc4747e2a1614e4f3d214d1d/client.ts#L86
+    if (
+      // detect Cloudflare Workers:
+      // https://developers.cloudflare.com/workers/platform/compatibility-dates/#global-navigator
+      (globalThis.navigator?.userAgent === "Cloudflare-Workers") ||
+      
+      // detect Edge runtime, deployed to Vercel (which runs on Cloudflare Workers, but doesn't
+      // appear to set the `global_navigator` flag required to pass the test above):
+      (typeof EdgeRuntime === "string" && !!process.env.VERCEL)
+    ) {
+      callback = AsyncLocalStorage.bind(callback);
+    }
+    
     let promise: PrismaPromise<unknown> | undefined
     const _callback = (callbackTransaction = transaction) => {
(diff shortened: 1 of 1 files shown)


andyjy:
Updated with:

1. link to new upstream issue created to track the underlying cause under Cloudflare workers - https://github.com/cloudflare/workerd/issues/870

2. test to only apply the workaround when we detect running under Cloudflare workers - so leave existing behaviour unaffected under the Node.js runtime. 

In particular re: (2) I discovered/realised the workaround does alter existing behaviour in that it "locks" the async context to that set at the time `createPrismaPromise()` is called, which shouldn't affect most situations but does prevent the ability to update/create a new async context within a Prisma Query Extension extended method that itself calls other query extension methods. 

This seems an extremely(!) rare use-case but an example is the override of `$transaction` within the Prisma local proxy I shared, here: https://github.com/andyjy/prisma-local-proxy/blob/35494f512e9968a9bc4747e2a1614e4f3d214d1d/client.ts#L86

For my own project I am unblocked without requiring this PR to be merged using `patch-package` to patch this change into the Prisma runtime. (It's not a pretty patch and will require regular manual recreation on future Prisma upgrades due to the minified distribution runtime file concatenating the source into extremely long lines, but it gets the job done.)

It may be the appropriate action for Prisma is simply to wait for as/when the upstream issue is implemented by CloudFlare/V8 - but I figured useful to update this PR to share/document my updated approach anyhow.

millsp:
Hey @andyjy thanks for your PR! We'll be taking a look at it soon to get it merged. I had a question about how would this work for people that don't have node compatibility enabled? Should we have some kind of conditional/wrapper that checks whether the module exists (I guess, we can see that when we run ecosystem tests)?

andyjy:
@millsp:

> how would this work for people that don't have node compatibility enabled? Should we have some kind of conditional/wrapper that checks whether the module exists

Yes agreed, good call. Albeit I haven't figured out the best way to implement the import from `node:async_hooks` in a way that works under all bundler setups / environments - including graceful failure when not present.

Separately, thanks for giving this PR some attention! 🙏  Worth calling out that unless this begins to affect many other Prisma users, the best action may be to wait it out until the underlying issue is fixed upstream in V8, as I just enquired about in https://github.com/cloudflare/workerd/issues/870#issuecomment-1648735616

andyjy:
Update from https://github.com/cloudflare/workerd/issues/870#issuecomment-1648981159:

> The change to fix that in V8 is here: https://chromium-review.googlesource.com/c/v8/v8/+/4674242
>
> That should hopefully land soon. Can't comment on when cloudflare would backport that to their runtime. It's in a working state already, just need V8 folks to approve it. They may want me to write some proper tests for it though. The existing ContinuationPreservedEmbedderData API apparently doesn't have any tests. 🤔

PR #20232, opened 2023-07-14
feat(migrate): add back `getDatabaseVersion`

jkomyno:
This PR closes https://github.com/prisma/prisma-private/issues/203.
Related Rust PR: https://github.com/prisma/prisma-engines/pull/4071.
Example fake error report with `getDatabaseVersion`: https://prisma-errors.netlify.app/report/14837.

Screenshot 2023-07-14 at 22 39 10

## Notes for the reviewers

The tests in this PR have been updated by locally tampering with the `schema-engine` binary, so that the release version of the binary built from https://github.com/prisma/prisma-engines/pull/4071 was used.

Hence, please:
- first, merge https://github.com/prisma/prisma-engines/pull/4071
- then, rebase this PR with `main` after the pipeline pulls in the Rust changes above
- then, wait for the TS CI to kick in and proceed with the review

This should be easier / cheaper than waiting for integration versions.


packages/internals/src/migrateTypes.ts
```diff
@@ -0,0 +1,24 @@
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace MigrateTypes {
```
Jolg42, line 2: Nit: technically, I think it's better to name like "SchemaEngineTypes", but not blocking this PR for sure

Jolg42 (approved):
Nice!

~~Maybe quickly check that the error reporting backend gets the database version like expected~~

packages/internals/src/migrateTypes.ts
jkomyno: I see where you're coming from, and in principle, I'd agree with you. However, considering we still have the `@prisma/migrate` package, wouldn't that be confusing to have conflicting naming conventions within the same codebase?

jkomyno:
> Nice!
> 
> Maybe quickly check that the error reporting backend gets the database version like expected

Tested already! That's why I've mentioned the example fake report https://prisma-errors.netlify.app/report/14837 in this PR's description :)

jkomyno:
> Nice!
> 
> Maybe quickly check that the error reporting backend gets the database version like expected

Tested already! That's why I've mentioned the example fake report https://prisma-errors.netlify.app/report/14837 in this PR's description :)

Issue #19611, opened 2023-06-05
Prisma Client: Explicit wrapper for JSON values


Jolg42:
## Problem

JSON column in Prisma Client is provided already parsed for developer convenience - that way, they don’t have to parse it to read from it or stringify it to pass as an input. But, since most of the JS values are also valid JSON, that poses a risk of collision in case particular field can accept multiple different values as a filter.

Couple of example from the past:

DbNull/JsonNull/AnyNull used to be strings in Prisma 3:
```ts
prisma.model.create({
   data: {
      jsonColumn: 'DbNull' 
      // do we want to to write string "DbNull" into the column or use DbNull enum value?
   }
})
```
To avoid this ambiguity, null constants have been changed to special objects in Prisma 4 (see [docs](https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-4#dbnull-jsonnull-and-anynull-are-now-objects)).

Field references were using simpler `{ _ref: "columnName" }` form in initial spec:
```ts
prisma.model.findMany({
   where: {
      jsonColumn: { equals: { _ref: "otherColumn" } }
      // do we want the values where jsonColumn === otherColumn or jsonColumn = "{ "_ref": "otherColumn"}"
   }
})
```
To avoid this collision, references were changed to special `prisma.model.$fields.field` objects (see [docs](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#compare-columns-in-the-same-table)).

Json list in combination with array shortcut:
```ts
prisma.model.create({
   data: {
      jsonList: ['a', 'b', 'c']
      // do we want 3 JSON values "a", "b", "c" within the list or single ["a", "b", "c"]?
   }
})
```

Originally, these conflicts were purely client problems: the engine returned and accepted JSON as unparsed strings so each of those situations could be unambiguously expressed one way or another. But since for JSON protocol, we don’t have query schema during encoding time and therefore can not stringify JSON values automatically, it brought those conflicts to the engine side as well. To resolve them, we had to introduce special values on the protocol level (semantically similar to *Null enums or $fields objects on the client.

Going forward, we will have to find a way to resolve such conflicts every time we introduce a union of JSON value and any other non-JSON values. We will be forced to introduce more wrappers for each of those alternative inputs, complicating our TS public API and JSON protocol.

## Suggested solution

Instead of introducing a wrapper around each entity that could conflict with JSON value, this proposal suggests introducing and requiring a wrapper around JSON values themselves. 

[...]

Issue #15397, opened 2022-09-19
Improve the UX of null values with JSON fields


aqrln:
Many users seem to find our API with `Prisma.DbNull`/`Prisma.JsonNull` confusing. The intended and expected behavior was reported as a bug in https://github.com/prisma/prisma/issues/9247.

One possible reason is that the documentation about it is hard to find (https://github.com/prisma/docs/issues/3800) — we will fix that in the docs, and we should also consider how we can make it more discoverable in editor tooltips and error messages.

Another possible reason is that the API might not be very intuitive, and we could find a way to match the users' expectations better.

One of the things that particularly confuse people, judging from the discussion, is that Prisma returns JavaScript `null`s for JSON fields in results (without distinguishing between two kinds of nulls), however doesn't accept them in the arguments of create, insert and upsert operations, because they are more strict. Returning `DbNull` and `JsonNull` from the results would solve this discrepancy, but it would be a big breaking change that would hinder interoperability with other code, make Prisma records less "pure objects" and make users unhappy about it.

janpio:
To add to the description of why this is problematic right now:

```
TSError: ⨯ Unable to compile TypeScript:
script.ts:15:40 - error TS2322: Type 'JsonValue' is not assignable to type 'JsonNull | InputJsonValue | undefined'.
  Type 'null' is not assignable to type 'JsonNull | InputJsonValue | undefined'.

15     await prisma.user.update({ data: { jsonField: rec.jsonField }, where: { id: 1 }});
```

This (or similar) is the error you get when trying to write back an object that you got via Prisma Client, with a `Json` field with `null` in the database. It does in no way indicate what the problem is and makes it super hard to figure out what is going on.

zayenz:
I recently stumbled on this paper-cut in the usage of Prisma, where I wanted to copy the Json-field from one row to a newly created row.

In my data-model, the Json-value is never null, it is always an object with key-value pairs. While full validation of check constraints (#1604 and #3388) or Json Schema checks (#16654) would be nice, a simple way to "just" say that a Json-field is never null would really be good for ergonomics, as it would enable straight-forward copying of Json values.

raymondKelly:
> To add to the description of why this is problematic right now:
> 
> ```
> TSError: ⨯ Unable to compile TypeScript:
> script.ts:15:40 - error TS2322: Type 'JsonValue' is not assignable to type 'JsonNull | InputJsonValue | undefined'.
[...]

PR #20245, opened 2023-07-16
Prisma Client : Create Explict wrapper.js

xonx4l:
An solution to -:https://github.com/prisma/prisma/issues/19611

### DIFF
--- a/packages/client/src/__tests__/generation/__fixture__/@prisma/client/Explict wrapper.js
+++ b/packages/client/src/__tests__/generation/__fixture__/@prisma/client/Explict wrapper.js
@@ -1,0 +1,45 @@
+ prisma.model.create({
+     data: {
+         // equivalent to DbNull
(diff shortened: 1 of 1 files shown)


aqrln:
Hi @xonx4l and thanks for your desire to contribute, we appreciate it! However, we have to close this PR since it doesn't actually solve or implement anything.

Do I understand it correctly that you just wanted to share your perspective on what kind of API you'd like to see? If that's the case, it's best to leave a comment with your proposal in the issue. Of course opening a PR right away is also fine — but then that PR needs to contain at least a proof of concept implementation.

Also, perhaps a better issue to leave the feedback in might be https://github.com/prisma/prisma/issues/15397, since the linked issue (https://github.com/prisma/prisma/issues/19611) is about a proposal for explicit wrapper for JSON values, which your examples don't seem to be about.

PR #20263, opened 2023-07-17
fix(cli): preinstall message when Node.js < 16.13

### LINKED ISSUE #20260: npm install of Prisma CLI fails on preinstall with no logs when Node.js version is lower than minimum
> I create the project with npx prisma init and uploaded to Render with different error, i link you the github repo to check it out (im using neon.tech as DB provider)
>
> https://github.com/Dominguezd01/PrismaTest
>
> I hope this helps, peace

_Originally posted by @Dominguezd01 in https://github.com/prisma/prisma/issues/20193#issuecomment-1637046919_

Jolg42:
Looks like this now
Screenshot 2023-07-17 at 16 33 32

Because `width`, which is a required parameter, was not defined it caused the empty box issue 
Screenshot 2023-07-17 at 16 42 08

See https://github.com/prisma/prisma/blob/main/packages/internals/src/utils/drawBox.ts
```
export type BoxOptions = {
  title?: string
  width: number
  height: number
  str: string
  horizontalPadding?: number
  verticalPadding?: number
}
```

It is the only place where `drawBox` was called without the required parameters, since width was not passed it resulted in a `NaN` in the code.

Closes https://github.com/prisma/prisma/issues/20260


packages/internals/src/utils/drawBox.ts
```diff
@@ -26,7 +26,13 @@ function maxLineLength(str: string): number {
 
 export function drawBox({ title, width, height, str, horizontalPadding }: BoxOptions): string {
   horizontalPadding = horizontalPadding || 0
+  // This avoids having NaN later
+  // if these required param were not passed (in JS, for example)
```
aqrln, line 30: Let's maybe move this comment above the previous line since it applies to it as well.

aqrln (approved):

janpio:
Stating the obvious: We should have tests for this to avoid this happening again in the future.

packages/cli/scripts/preinstall.js
```diff
@@ -1,17 +1,19 @@
...
+var { red, underline } = require('kleur/colors')
 
 export function main() {
   // process.version (e.g. `v16.0.0`)
   const nodeVersions = process.version.split('.')
   // `.slice(1)` removes `v` from `v16`
   const nodeMajorVersion = parseInt(nodeVersions[0].slice(1))
   const nodeMinorVersion = parseInt(nodeVersions[1])
-  if (nodeMajorVersion < 16 || nodeMajorVersion === 16 && nodeMinorVersion < 13) {
+  if (nodeMajorVersion < 16 || (nodeMajorVersion === 16 && nodeMinorVersion < 13)) {
```
millsp, line 10: Should we add a test to snapshot this message for the future?

millsp (approved):

packages/cli/scripts/preinstall.js
```diff
@@ -1,17 +1,26 @@
-const { drawBox } = require('@prisma/internals/dist/utils/drawBox')
+// @ts-check
+const { drawBox, kleurColors } = require('@prisma/internals')
```
Jolg42, line 2: Using a re-export because kleur is a dev dependency and we bundle deps

PR #20296, opened 2023-07-19
feat(client): Add shortcut for `by` argument

### LINKED ISSUE #20261: Types wrongly accept non-array parameter `by` in `groupBy` in 5.0.0
With the following schema:

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
  url      = "file:dev.db"
}

model A {
  id Int @id @default(autoincrement())
}
```

and query:

```ts
await prisma.a.groupBy({
  by: 'id'
})
```

In Prisma 4.16.2 this is a type error (as expected, it must be `by: ['id']`). In Prisma 5, however, this passes the type checking and fails at run time with the following validation error:
```
PrismaClientValidationError:
Invalid `prisma.a.groupBy()` invocation in
/Users/aqrln/prisma/transient-repros/groupby/index.ts:6:33

  3 async function main() {
  4   const prisma = new PrismaClient()
  5
→ 6   const result = await prisma.a.groupBy({
        select: {
      ?   id?: true,
      ?   _count?: true,
      ?   _avg?: true,
      ?   _sum?: true,
      ?   _min?: true,
      ?   _max?: true
        },
        by: "id"
      })

The `select` statement for type AGroupByOutputType must not be empty. Available options are listed in green.
    at Ur (/Users/aqrln/prisma/transient-repros/groupby/node_modules/@prisma/client/runtime/library.js:115:5803)
    at zr.handleRequestError (/Users/aqrln/prisma/transient-repros/groupby/node_modules/@prisma/client/runtime/library.js:122:8007)
    at zr.handleAndLogRequestError (/Users/aqrln/prisma/transient-repros/groupby/node_modules/@prisma/client/runtime/library.js:122:7697)
    at zr.request (/Users/aqrln/prisma/transient-repros/groupby/node_modules/@prisma/client/runtime/library.js:122:7307) {
  clientVersion: '5.0.0'
}
```
(and the validation error suggestion is not accepted by the TS types)

SevInf:
This shortcut exists and works on the engine level, but was not
implemented on the client. Since 5.0, validation and types allowed to
have a scalar in `by`, but such messages were not encoded properly.

Fix #20261


Jolg42 (approved):

packages/client/tests/functional/issues/20261-group-by-shortcut/tests.ts
```diff
@@ -0,0 +1,55 @@
...
+        teamName: 'Red',
+      },
+    ])
+  })
+
+  test('works with a scalar in "by" and no other selection', async () => {
+    const result = await prisma.round.groupBy({
+      by: 'teamName',
+      orderBy: { teamName: 'asc' },
+    })
```
millsp, line 44: Should we also add type-level tests? Not sure if we needed to update the underlying implementation?

millsp: For Client extensions, groupBy types are "disconnected" from the centralized types we introduced via extensions in `GetResult.ts`, I think. So we may need to add tests for client extensions too (see [code](https://github.com/prisma/prisma/blob/62e09f34b9b0215817d085907d349eeaa842cc35/packages/client/src/runtime/core/types/GetResult.ts/#L127-L129)).

SevInf: Fixed extensions as well.

millsp: Nice, thank you.

millsp (approved):

aqrln (approved):

PR #20299, opened 2023-07-19
fix(client): Properly handle namespaces in DMMFHelper

### LINKED ISSUE #5749: [v2.17.0] Missing client TS types Aggregate*Args
Since upgrading to v2.17.0 from v2.15.0 for both `@prisma/client` and `prisma` i get an TypeScript error stating that the `Aggregate*Args` types are missing:

VSCode screenshot - missing TS types

After some digging around in `index.d.ts` I've found the type `AggregateUser` which in my understanding should replace  `AggregateUserArgs` in  `aggregateUser<T extends AggregateUserArgs>`.  

Unfortunately downgrading to **v2.15.0** has a similar error.

I've used:
- `prisma migrate dev --preview-feature` (added Table `Query`) after removing the `prisma/migrations` folder (as suggested [here](https://github.com/prisma/prisma/issues/5367), although i did not switch DB providers)
- `prisma generate`
- `prisma db seed --preview-feature`

My **prisma.schema**:

```prisma
datasource db {
    provider = "mysql"
}

generator client {
    provider = "prisma-client-js"
}

model User {
    id        Int       @id @default(autoincrement())
    name      String?   @unique
    email     String    @unique
    password  String
    createdAt DateTime  @default(now())
    updatedAt DateTime? @updatedAt
}

enum MediaSource {
    LOCAL
    EXTERNAL
}

model Media {
    id           Int         @id @default(autoincrement())
    path         String
    source       MediaSource @default(LOCAL)
    mime         String
    previewImage String?     @db.Text
    createdAt    DateTime    @default(now())
    updatedAt    DateTime?   @updatedAt
    product      Product?
    market       Market?
}

model Product {
    id         Int        @id @default(autoincrement())
    name       String     @unique
    media      Media?     @relation(fields: [mediaId], references: [id])
    mediaId    Int?
    createdAt  DateTime   @default(now())
    updatedAt  DateTime?  @updatedAt
    expenses   Expense[]
    categories Category[]
}

model Market {
    id        Int       @id @default(autoincrement())
    name      String    @unique
    media     Media?    @relation(fields: [mediaId], references: [id])
    mediaId   Int?
    createdAt DateTime  @default(now())
    updatedAt DateTime? @updatedAt
    expenses  Expense[]
}

model Expense {
    product   Product  @relation(fields: [productId], references: [id])
    productId Int
    market    Market?  @relation(fields: [marketId], references: [id])
    marketId  Int?
    cost      Float
    amount    Int      @default(1)
    date      DateTime @db.Date
}

model Category {
    id        Int       @id @default(autoincrement())
    name      String    @unique
    color     String?
    createdAt DateTime  @default(now())
    updatedAt DateTime? @updatedAt
    products  Product[]
}

model Query {
    id         Int       @id @default(autoincrement())
    color      String?
    collection String
    query      String    @db.Text
    createdAt  DateTime  @default(now())
    updatedAt  DateTime? @updatedAt
}
```

### LINKED ISSUE #8153: Models named Query cause an internal GraphQL Parse Error
### Bug description

When naming models any case of `query`, no operations can be performed. Originally discovered on an actual project of mine, and discussed with @janpio. Initial suspicion was that this was due to the actual content of my query, but once I managed to troubleshoot the begun, the name of the variable (which fixed the problem) became my theory.

Further confirmed by an extremely simple model with just one field producing the exact same error.

### How to reproduce

Setup a basic prisma project using NPM. This means the standard:
```
npm init
npm i -D typescript ts-node @types/node prisma
npx prisma init
npx prisma migrate
```
Just use a simple sqlite3 database, it won't even query the DB because it'll crash first.

### Expected behavior

This should work, and not cause this query error, or there should at least be a more descriptive error message which is specific to this error and indicates the troubleshooting step. The standard error (below) is totally useless.
```
C:\Users\david\Documents\Dev\prisma-querybug\node_modules\@prisma\client\runtime\index.js:33900
          throw new import_engine_core.PrismaClientUnknownRequestError(message, this.prisma._clientVersion);
                ^
Error:
Invalid `prisma.query.create()` invocation:


  Error parsing GraphQL query: query parse error: Parse error at 6:3
Unexpected `}[Punctuator]`
Expected `Name`
```

If the decision is made not to correct this behaviour, this should be documented.

### Prisma information

Prisma Schema:
```
datasource db {
  provider = "sqlite"
  url      = "file:./test.db"
}

generator client {
  provider = "prisma-client-js"
}

model Query {
  id String @id @default(uuid())
  // literally not including anything else, its not because of the model contents...
}
```

Test Code:
```ts

async function main() {
    prisma.$connect();
    const qry = await prisma.query.create({
        data: {}
    });

    console.log(qry);
}

main();
```
with package.json:
```
  "devDependencies": {
    "@types/node": "^16.0.1",
    "prisma": "^2.26.0",
    "ts-node": "^10.0.0",
    "typescript": "^4.3.5"
  },
  "dependencies": {
    "@prisma/client": "^2.26.0"
  }
```

### Environment & setup

- OS:  Windows, and likely all.
- Database: PostgreSQL and SQLite verified, and likely all.
- Node.js version:  15.3.0

### Prisma Version

```
prisma               : 2.26.0
@prisma/client       : 2.26.0
Query Engine         : query-engine 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\query-engine-windows.exe)
Migration Engine     : migration-engine-cli 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\migration-engine-windows.exe)
Introspection Engine : introspection-core 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary        : prisma-fmt 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash : 9b816b3aa13cc270074f172f30d6eda8a8ce867d
Studio               : 0.408.0
```

### LINKED ISSUE #9307: Naming an entity "Query" leads to an error
### Bug description

Naming an entity "Query" in prisma.schema leads to an error when trying to start such a project. The console output will not contain clear information about what's wrong. The auto-generated file _\node_modules\.prisma\client\index.d.ts_ contains errors after naming an entity "Query".

### How to reproduce

1. Set up the prisma quickstart project [https://www.prisma.io/docs/getting-started/quickstart](https://www.prisma.io/docs/getting-started/quickstart) 
2. In schema.prisma, change the name of the entity "Post" to "Query" wherever it occurs
3. Run a migration (that regenerates the client as well)
4. Run "npm run dev" and see the console error output
5. open \node_modules\.prisma\client\index.d.ts and see the TypeScript errors

### Expected behavior

The app starts without error and outputs an empty array to the console since it's the prisma quickstart project with an empty database (after our migration).

### Prisma information

[schema.prisma.txt](https://github.com/prisma/prisma/files/7186264/schema.prisma.txt)

### Environment & setup

I have run into this error on Windows 10, Node 14.17.6 and Prisma 3.0.2

### Prisma Version

```
Prisma 3.0.2
```

SevInf:
`DMMFHelper` did not take namespaces into account and merged them all
into single map, based on the name. This is incorrect, since it is
possible to have different type with the same name in different
namespaces. In particular, this broke models named either `Query` or
`Mutation` since they match built in names from `prisma` namespace.

Now we have a separate map for each namespace.

Fix #5749
Fix #9307
Fix #8153

### DIFF
--- a/packages/client/src/generation/TSClient/Model.ts
+++ b/packages/client/src/generation/TSClient/Model.ts
@@ -48,7 +48,7 @@ export class Model implements Generatable {
   constructor(protected readonly model: DMMF.Model, protected readonly context: GenerateContext) {
     this.dmmf = context.dmmf
     this.genericsInfo = context.genericArgsInfo
-    this.type = this.context.dmmf.outputTypeMap[model.name]
+    this.type = this.context.dmmf.outputTypeMap.model[model.name]
     this.mapping = this.context.dmmf.mappings.modelOperations.find((m) => m.model === model.name)!
   }
   protected get argsTypes(): Generatable[] {
@@ -96,7 +96,7 @@ export class Model implements Generatable {
   private getGroupByTypes() {
     const { model, mapping } = this
 
-    const groupByType = this.dmmf.outputTypeMap[getGroupByName(model.name)]
+    const groupByType = this.dmmf.outputTypeMap.prisma[getGroupByName(model.name)]
     if (!groupByType) {
       throw new Error(`Could not get group by type for model ${model.name}`)
     }
@@ -156,7 +156,7 @@ type ${getGroupByPayloadName(model.name)}<T extends ${groupByArgsName}> = Prisma
   }
   private getAggregationTypes() {
     const { model, mapping } = this
-    let aggregateType = this.dmmf.outputTypeMap[getAggregateName(model.name)]
+    let aggregateType = this.dmmf.outputTypeMap.prisma[getAggregateName(model.name)]
     if (!aggregateType) {
       throw new Error(`Could not get aggregate type "${getAggregateName(model.name)}" for "${model.name}"`)
     }
@@ -169,11 +169,11 @@ type ${getGroupByPayloadName(model.name)}<T extends ${groupByArgsName}> = Prisma
 
     const aggregateTypes = [aggregateType]
 
-    const avgType = this.dmmf.outputTypeMap[getAvgAggregateName(model.name)]
-    const sumType = this.dmmf.outputTypeMap[getSumAggregateName(model.name)]
-    const minType = this.dmmf.outputTypeMap[getMinAggregateName(model.name)]
-    const maxType = this.dmmf.outputTypeMap[getMaxAggregateName(model.name)]
-    const countType = this.dmmf.outputTypeMap[getCountAggregateOutputName(model.name)]
+    const avgType = this.dmmf.outputTypeMap.prisma[getAvgAggregateName(model.name)]
+    const sumType = this.dmmf.outputTypeMap.prisma[getSumAggregateName(model.name)]
+    const minType = this.dmmf.outputTypeMap.prisma[getMinAggregateName(model.name)]
+    const maxType = this.dmmf.outputTypeMap.prisma[getMaxAggregateName(model.name)]
+    const countType = this.dmmf.outputTypeMap.prisma[getCountAggregateOutputName(model.name)]
 
     if (avgType) {
       aggregateTypes.push(avgType)
--- a/packages/client/src/generation/dmmf.ts
+++ b/packages/client/src/generation/dmmf.ts
@@ -2,7 +2,7 @@ import type { DMMF } from '@prisma/generator-helper'
 
 import { applyMixins } from '../runtime/utils/applyMixins'
 import type { Dictionary } from './utils/common'
-import { keyBy, ScalarTypeTable } from './utils/common'
+import { keyBy } from './utils/common'
 
 export class DMMFDatamodelHelper implements Pick<DMMF.Document, 'datamodel'> {
   datamodel: DMMF.Datamodel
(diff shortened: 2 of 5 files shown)


packages/client/src/generation/TSClient/Model.ts
```diff
@@ -48,7 +48,7 @@ export class Model implements Generatable {
   constructor(protected readonly model: DMMF.Model, protected readonly context: GenerateContext) {
     this.dmmf = context.dmmf
     this.genericsInfo = context.genericArgsInfo
-    this.type = this.context.dmmf.outputTypeMap[model.name]
+    this.type = this.context.dmmf.outputTypeMap.model[model.name]
```
millsp, line 51: Can be model or composite, right?

packages/client/src/generation/dmmf.ts
```diff
@@ -108,96 +109,104 @@ class DMMFSchemaHelper implements Pick<DMMF.Document, 'schema'> {
...
+          // should not happen
+          continue
         }
-      }
-      type.fieldMap = keyBy(type.fields, 'name')
-    }
 
-    for (const type of this.outputTypes.prisma) {
-      for (const field of type.fields) {
-        if (typeof field.outputType.type === 'string' && !ScalarTypeTable[field.outputType.type]) {
```
millsp, line 127: Was the scalar table not needed anymore because you changed the way the condition is checked?

packages/client/src/generation/dmmf.ts
```diff
@@ -108,96 +109,104 @@ class DMMFSchemaHelper implements Pick<DMMF.Document, 'schema'> {
...
       for (const field of type.fields) {
         for (const fieldInputType of field.inputTypes) {
-          const fieldType = fieldInputType.type
-          if (
-            typeof fieldType === 'string' &&
-            !ScalarTypeTable[fieldType] &&
-            (this.inputTypeMap[fieldType] || this.enumMap[fieldType])
-          ) {
-            fieldInputType.type = this.inputTypeMap[fieldType] || this.enumMap[fieldType] || fieldType
+          if (typeof fieldInputType.type !== 'string') {
```
millsp, line 149: Is this mostly for type narrowing?

millsp (approved):
After taking a bit of time to remember this part of the code, it looks good to me. Nice and clean!

packages/client/src/generation/dmmf.ts
SevInf: Yes, on engine side, `type` is always string but we define it is `string | SchemaEnum | SchemaObject` and override the value in one of the `resolveX` methods. Since we don't have separate types for input/output DMMF, TS types force us to handle non-string cases here as well.
Ideally, I'd just remove all `resolve` methods and leave original DMMF untouched (already created tech debt card for this). This was useful mostly for `query.ts`, for generator, "flat" types are more convenient for majority of cases.

packages/client/src/generation/TSClient/Model.ts
SevInf: Actually, on the `schema` level of DMMF we don't differentiate between models and composites - derived types for both land here.

PR #20319, opened 2023-07-20
fix(client): intellisense type shapes

### LINKED ISSUE #20320: Hovering over types (intellisense) shows confusing `GetResult`
### Bug description

When you hover over the result type of a result (for instance a `findMany`), when it is stored into a variable, the intellisense popup shows a confusing `GetResult` wrapping the result types.

### How to reproduce

1. Create a project and a schema
2. Generate the client
3. Create an index.ts
4. Write a `const data = await prisma.<model>.findMany` query 
5. Hover over `data`
6. See the strange `GetResult` included in the types

Screenshot from 2023-07-20 13-16-02

### Expected behavior

`GetResult` should not be present. Either the selected fields or the model type.

### Prisma information

n/a

### Environment & setup

n/a

### Prisma Version

```
5.0.0
```

millsp:
Gets rid of the `GetResult` by forcing the evaluation of the types on the results. This cannot be tested easily because the types are purely equivalent and it's only a visual change in intellisense, so I left a comment in the code.

Before:
Screenshot from 2023-07-20 13-16-02

After:
Screenshot from 2023-07-20 13-15-31

An unrelated [PR](https://github.com/prisma/prisma/pull/20161) about performance will get rid of the extra `& {}`, so it will be easier to read than just the current fix. Keeping the PRs separate on purpose, although there's some overlap.

closes https://github.com/prisma/prisma/issues/20320
closes https://github.com/prisma/team-orm/issues/228

### DIFF
--- a/packages/client/src/runtime/core/types/Extensions.ts
+++ b/packages/client/src/runtime/core/types/Extensions.ts
@@ -27,6 +27,7 @@ export type DefaultArgs = InternalArgs<{}, {}, {}, {}>
 
 export type GetResult<Base extends Record<any, any>, R extends Args['result'][string], KR extends keyof R = string extends keyof R ? never : keyof R> =
   { [K in KR | keyof Base]: K extends KR ? R[K] extends (() => { compute: (...args: any) => infer C }) ? C : never : Base[K] } & unknown
+  // ! the intersection with unknown is important, it prevent the result types from showing an ugly `GetResult<...> & {}` type on hover
 
 export type GetSelect<Base extends Record<any, any>, R extends Args['result'][string], KR extends keyof R = string extends keyof R ? never : keyof R> =
   { [K in KR | keyof Base]?: K extends KR ? boolean : Base[K] }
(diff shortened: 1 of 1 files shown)


packages/client/src/runtime/core/types/Extensions.ts
```diff
@@ -26,7 +26,7 @@ export type Args = InternalArgs
 export type DefaultArgs = InternalArgs<{}, {}, {}, {}>
 
 export type GetResult<Base extends Record<any, any>, R extends Args['result'][string], KR extends keyof R = string extends keyof R ? never : keyof R> =
-  { [K in KR | keyof Base]: K extends KR ? R[K] extends (() => { compute: (...args: any) => infer C }) ? C : never : Base[K] }
+  { [K in KR | keyof Base]: K extends KR ? R[K] extends (() => { compute: (...args: any) => infer C }) ? C : never : Base[K] } & unknown
```
Jolg42, line 29: >so I left a comment in the code.
Did you push the comment? 😄

millsp: Nope :homer-simpson-disappears-into-bush:

millsp: https://github.com/prisma/prisma/pull/20319/files#diff-17bc9f33a3484a25fedb3eaaeb439e7cd240ec835fc020953241d0c2b075c3e5R30

Jolg42 (approved):

Issue #20325, opened 2023-07-21
Prisma 5 drops `undefined` from Arrays when using Json fields with Postgres


cpojer:
### Bug description

Since Prisma 5, `undefined` values in an array are dropped when writing them to a Json field using Postgres. In Prisma 4, 
`undefined` values were turned into `null`, mirroring `JSON.stringify` behavior in JavaScript.

I expect undefined values not to get dropped from arrays and the encoding behavior to match `JSON.stringify`.

### How to reproduce

Schema:
```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

model Todo {
  id    Int     @default(autoincrement()) @id
  content String 
  checked  Boolean @default(false)
  test Json?
}
```

Create call:
```js
await prisma.todo.create({
    data: {
      content,
      checked: false,
      test: ["here", undefined, "is", null, "a", 0, "value"],
    },
  });
```

Bug:
```js
const todos = await prisma.todo.findMany();

console.log(todos[0].test) // ["here, "is", null, "a", 0, "value"]
```

Repros:
Prisma 4: https://codesandbox.io/p/sandbox/next-js-server-actions-prisma-postgres-demo-forked-2l7nzz?embed=1&file=%2Fapp%2Fprisma.ts%3A1%2C1
Same container but with Prisma 5: https://codesandbox.io/p/sandbox/next-js-server-actions-prisma-postgres-demo-forked-xf42jz?embed=1&file=%2Fpackage.json%3A28%2C16

You can run `node` in those containers and execute this:
```js
const PrismaClient = require('@prisma/client').PrismaClient;
const prisma = new PrismaClient();
const t = await prisma.todo.findMany();
console.log(t);
```

To inspect the data returned from Prisma.

### Expected behavior

JavaScript turns `JSON.stringify([undefined, null])` into `'[null, null]'`. I expect Prisma 5 to match this behavior.

Concretely, the above code should return `["here, null, "is", null, "a", 0, "value"]` like in Prisma 4 instead of `["here, "is", null, "a", 0, "value"]`. Notice that the second element (`null`) is missing in the Prisma 5 version.

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
// Add your code using Prisma Client
```

### Environment & setup

Latest of literally everything, macOS and Alpine Linux.

### Prisma Version

```
5.0.0
```

Jolg42:
[...]

PR #20330, opened 2023-07-21
fix(client): return null when undefined in Json arr

Jolg42:
Related to https://github.com/prisma/prisma/issues/20325

TODO: Is that what we really want to do here?

/integration


aqrln:
Should this be closed in favour of https://github.com/prisma/prisma/pull/20777?

Jolg42:
I think so

PR #20332, opened 2023-07-21
chore: add template for discussions

ludralph:
I created a `new.yml` file to add a template for GitHub Discussion based on this [GitHub Documentation](https://docs.github.com/en/discussions/managing-discussions-for-your-community/syntax-for-discussion-category-forms)


Jolg42:
I'm a bit confused that this looks like a copy and paste of the bug template: I'll ping you on Slack

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,69 @@
+name: New Discussion
+description: Please fill out the following information to create a new discussion.
+  Make sure to provide clear and concise details to help others understand and address your concerns effectively.
+
+body:
+  - type: textarea
+    id: description
+    attributes:
+      label: Problem description
+      description: A clear and concise description of what the problem is
```
Jolg42, line 10: Maybe the phrasing is too restrictive here? It could also be an idea, for example.
```suggestion
      label: Idea/Problem description
      description: A clear and concise description of what the idea/problem is
```

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,69 @@
...
+      description: Steps to reproduce the behavior
+      value: |
+        <!--
+        1. Go to '...'
+        2. Change '....'
+        3. Run '....'
+        4. See error
+        -->
+    validations:
+      required: true
```
Jolg42, line 26: Note about all of these:
In the bug report template, it makes sense to have them required, but here I think they should be optional given that the subject of a discussion could be "anything"?

janpio: I agree, this does not sound like the right template for a generic "New Discussion" form.

janpio (changes requested):
See comments

.github/DISCUSSION_TEMPLATE/new.yml
ludralph: I have updated this based on your suggestion

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,61 @@
+name: New Prisma Discussion
```
nikolasburk, line 1: why not use "Question" to be more specific and clear about what exactly this is? 


```suggestion
name: Question
```

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,61 @@
...
+description: Please fill out the following information to create a new discussion related to Prisma.
+  Make sure to provide clear and concise details to help others understand and address your concerns effectively.
```
nikolasburk, line 3:
```suggestion
description: Use the following template to ask your question about Prisma. Make sure to provide clear and concise details to help others understand and address your concerns effectively.
```

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,61 @@
...
+
+body:
```
nikolasburk, line 5: I kind of like this intro in the bug reports template, maybe we should have smth similar?

image


```suggestion
body:
  - type: markdown
    attributes:
      value: |
        Thanks for raising the questions you have about Prisma, we are happy to help! 🙏 Please follow the sections in the template and provide as much information as possible about your problem, e.g. by [setting the `DEBUG="*"` environment variable](https://www.prisma.io/docs/concepts/components/prisma-client/debugging) and enabling [additional logging output](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging) in Prisma Client.   
```

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,61 @@
...
+  - type: textarea
+    id: description
+    attributes:
+      label: Description of the Idea / Problem
```
nikolasburk, line 9: we are primarily framing GitHub Discussion as a place for "Questions", so I think we should use this as our primary wording:

```suggestion
      label: Question
```

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,61 @@
...
+      description: Provide a clear and concise description of the idea or problem you want to discuss.
```
nikolasburk, line 10:
```suggestion
      description: Provide a clear and concise description of the question you have or problem you want to discuss.
```

.github/DISCUSSION_TEMPLATE/new.yml
```diff
@@ -0,0 +1,61 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: steps
+    attributes:
+      label: How to reproduce
+      description: Steps to reproduce the behavior
```
nikolasburk, line 17:
```suggestion
      description: If your question is about specific behaviour of Prisma, tell us how to reproduce it.
```

nikolasburk (changes requested):
this looks great Raphael, I left a few comments for the first sections which I'd like you to incorporate but I think we can publish this soon 🚀 🚀 🚀

Jolg42 (changes requested):
Something went wrong during a merge?
The PR diff got a lot bigger suddenly

nikolasburk (commented):
the changes in `.github/DISCUSSION_TEMPLATE/new.yml`  look good to me; but why does this PR touch so many other files? 🤔

nikolasburk:
> Something went wrong during a merge? The PR diff got a lot bigger suddenly

Yeah, I noticed this as well!

@ludralph are you able to resolve this? You can just create a new PR where only the schema template is changed as the only file 😅

PR #20340, opened 2023-07-21
ci: disable "typescript compilation" benchmark under CodSpeed

aqrln:


millsp:
CodSpeed is now completely disabled, was that intentional?

aqrln:
@millsp it didn't run because of a typo that made the workflow file invalid, should be fixed now

millsp (approved):

millsp:
image :trollface:

aqrln:
@millsp I think it's red because of a removed benchmark, not because of the regression. Our current configured threshold should be 10% I believe.

millsp:
Nice, thanks for the context.

PR #20381, opened 2023-07-25
feat: Add support for LD_LIBRARY_PATH

Gerschtli:
This PR fixes the issues described in https://github.com/prisma/prisma/pull/18012. Please have a look at this tickets regarding the objective. It is split from https://github.com/prisma/prisma/pull/20138 as recommended by @aqrln.

Basically, these changes allow users with a valid NixOS setup to set the openssl library directory. This also opens the possibility for more portability for non standard setups.

### DIFF
--- a/packages/get-platform/src/__tests__/getSSLVersion.test.ts
+++ b/packages/get-platform/src/__tests__/getSSLVersion.test.ts
@@ -6,11 +6,34 @@ const describeIf = (condition: boolean) => (condition ? describe : describe.skip
 const ctx = jestContext.new().assemble()
 
 describeIf(process.platform === 'linux')('computeLibSSLSpecificPaths', () => {
-  // eslint-disable-next-line jest/no-identical-title
-  it('should not return an error', () => {
+  const originalEnv = process.env
+
+  afterEach(() => {
+    process.env = originalEnv // Restore old environment
+  })
+
+  it('should return system specific search paths when LD_LIBRARY_PATH is not set', () => {
+    delete process.env.LD_LIBRARY_PATH
     const arch = 'x64'
     const archFromUname = 'x86_64'
-    computeLibSSLSpecificPaths({ familyDistro: 'debian', arch, archFromUname })
+    const paths = computeLibSSLSpecificPaths({ familyDistro: 'debian', arch, archFromUname })
+    expect(paths).toEqual(['/usr/lib/x86_64-linux-gnu', '/lib/x86_64-linux-gnu'])
+  })
+
+  it('should return system specific search paths when LD_LIBRARY_PATH is an empty string', () => {
+    process.env.LD_LIBRARY_PATH = ''
+    const arch = 'x64'
+    const archFromUname = 'x86_64'
+    const paths = computeLibSSLSpecificPaths({ familyDistro: 'debian', arch, archFromUname })
+    expect(paths).toEqual(['/usr/lib/x86_64-linux-gnu', '/lib/x86_64-linux-gnu'])
+  })
+
+  it('should respect LD_LIBRARY_PATH and fall back to system specific search paths', () => {
+    process.env.LD_LIBRARY_PATH = '/path/to/libs:/other/path'
+    const arch = 'x64'
+    const archFromUname = 'x86_64'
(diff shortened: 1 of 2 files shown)


packages/get-platform/src/getPlatform.ts
```diff
@@ -309,6 +309,15 @@ export function computeLibSSLSpecificPaths(args: ComputeLibSSLSpecificPathsParam
       debug(`Don't know any platform-specific paths for "${familyDistro}" on ${arch} (${archFromUname})`)
       return []
     })
+
+  if (process.env.LD_LIBRARY_PATH) {
+    debug(
+      `Found 'LD_LIBRARY_PATH' specified as env-var, these paths will take presendence over platform specific lib-dir paths...`,
```
aqrln, line 315:
```suggestion
      `Found 'LD_LIBRARY_PATH' specified as env-var, these paths will take precedence over platform specific lib-dir paths`,
```

aqrln (approved):
looks good to me, thank you!

aqrln:
Ideally this needs to be tested, but we can add the tests later (feel free to open a follow-up PR if you'd like to contribute the tests).

Gerschtli:
I updated the tests for both my changes :)

packages/get-platform/src/__tests__/getSSLVersion.test.ts
```diff
@@ -6,11 +6,26 @@ const describeIf = (condition: boolean) => (condition ? describe : describe.skip
 const ctx = jestContext.new().assemble()
 
 describeIf(process.platform === 'linux')('computeLibSSLSpecificPaths', () => {
-  // eslint-disable-next-line jest/no-identical-title
-  it('should not return an error', () => {
+  const OLD_ENV = process.env
+
+  afterAll(() => {
+    process.env = OLD_ENV // Restore old environment
+  })
```
aqrln, line 13:
```suggestion
  const ORIGINAL_ENV = process.env

  beforeEach(() => {
    process.env = ORIGINAL_ENV // Restore old environment
  })
```

let's make sure each test starts with the fresh env and nothing can leak between them

packages/get-platform/src/__tests__/getSSLVersion.test.ts
```diff
@@ -6,11 +6,26 @@ const describeIf = (condition: boolean) => (condition ? describe : describe.skip
...
+
+  it('should return system specific search paths', () => {
+    process.env.LD_LIBRARY_PATH = ''
```
aqrln, line 16: maybe

```suggestion
    delete process.env.LD_LIBRARY_PATH
```

Or perhaps even two separate tests for missing and empty `LD_LIBRARY_PATH`?

aqrln (commented):
looks great, a couple of comments regarding the tests and let's merge afterwards

packages/get-platform/src/__tests__/getSSLVersion.test.ts
Gerschtli: oh thanks, good catch :)

aqrln (approved):
looks great, thanks!

janpio:
Thanks @Gerschtli!

aqrln:
Hey @Gerschtli, I temporarily reverted this before the release (https://github.com/prisma/prisma/pull/20457) so this won't make it into 5.1.0, but this is still something I'd like to re-apply and merge for 5.2.0. The reason is this turned out to be a breaking change (flagged by our ecosystem-tests) for users deploying on AWS Lambda and other serverless functions providers powered by it (e.g. Netlify and Vercel) as their `binaryTargets` property in the schema file would no longer be correct. There is nothing wrong with this change per se, but we need to adjust the platform detection logic to make this non-breaking (see details in the revert PR).

VanCoding:
@aqrln After reading the revert & a lot of comments to it, I still don't get what was the problem this this PR.

Did it make problems with Vercel? Is there a LD_LIBRARY_PATH set that points to different openssl files? If not, how can this PR have an effect on this?

I'm very annoyed by the warnings about openssl not being found on NixOS and really would like to be able to fix them...

PR #20417, opened 2023-07-27
feat: Make Prisma Studio work with `directUrl`

### LINKED ISSUE #17358: Use `directUrl` in `prisma studio`
4.9.0 added `directUrl` for Migration and Introspection commands. We should also use `directUrl` for local Studio.

janpio:
Closes https://github.com/prisma/prisma/issues/17358

Depends on https://github.com/prisma/studio-code/pull/483
Related docs change: https://github.com/prisma/docs/pull/4953

Internal discussion: [prisma-company.slack.com/archives/CC8F34ALF/p1685742636214419?thread_ts=1685736658.218249&cid=CC8F34ALF](https://prisma-company.slack.com/archives/CC8F34ALF/p1685742636214419?thread_ts=1685736658.218249&cid=CC8F34ALF)


packages/internals/src/cli/checkUnsupportedDataProxy.ts
```diff
@@ -62,7 +60,7 @@ async function checkUnsupportedDataProxyMessage(command: string, args: Args, imp
 
       const datamodel = await fs.promises.readFile(argValue, 'utf-8')
       const config = await getConfig({ datamodel, ignoreEnvVarErrors: true })
-      const url = command === 'studio' ? config.datasources[0]?.url : getEffectiveUrl(config.datasources[0])
+      const url = getEffectiveUrl(config.datasources[0])
```
janpio, line 63: Before this explicitly skipped `studio` for the usage of the `directUrl` (via `getEffectiveUrl`).
Now `studio` does not get special handling, which allows the command to execute with a `prisma://` `url` and a direct `directUrl` value.

packages/cli/src/Studio.ts
```diff
@@ -93,6 +94,11 @@ ${bold('Examples')}
 
     const staticAssetDir = path.resolve(__dirname, '../build/public')
 
+    const schema = await fs.promises.readFile(schemaPath, 'utf-8')
+    const config = await getConfig({ datamodel: schema, ignoreEnvVarErrors: false })
+    const _directUrl = getDirectUrl(config.datasources[0])
+    const directUrl = _directUrl.value
```
janpio, line 100: This retrieves the `directUrl` explicitly from the schema and puts it into `StudioServer` to be used.

(This is a bit dirty, as Studio gets the `url` value by itself, but this was the best and most straightforward way to make Studio actually handle `directUrl` without completely refactoring it and the way it uses Prisma Client.)

packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -24,49 +21,6 @@ function sendRequest(message: any): Promise<any> {
...
-    process.env.STUDIO_URL = 'prisma://aws-us-east-1.prisma-data.com/?api_key=MY_API_KEY'
-    process.env.STUDIO_DIRECT_URL = 'postgres://localhost:1234'
-  })
-
-  afterEach(() => {
-    // Back to original env vars
-    process.env = { ...originalEnv }
-  })
-
-  it('should fail if url is prisma://', async () => {
```
janpio, line 41: This should not fail any more.

packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -24,49 +21,6 @@ function sendRequest(message: any): Promise<any> {
...
-    await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
-
-      Using the Data Proxy (connection URL starting with protocol prisma://) is not supported for this CLI command prisma studio yet. 
-
-      More information about Data Proxy: https://pris.ly/d/data-proxy-cli
-
-    `)
-  })
-
-  it('should fail if url is prisma:// and directUrl is set', async () => {
```
janpio, line 55: This should also not fail any more.

packages/cli/src/Studio.ts
janpio, line 100:
```suggestion
    let directUrl = null
    if(_directUrl)
    	directUrl = _directUrl.value
```

janpio:
This can manually be tested via `prisma@5.1.0-integration-studio-direct-url.8`.

packages/cli/src/__tests__/commands/Studio.test.ts
millsp: This should still be the case, I think. If you don't provide a `directUrl` then it should fail.

packages/client/src/runtime/utils/validatePrismaClientOptions.ts
```diff
@@ -20,8 +20,7 @@ const validators = {
 
     for (const [key, value] of Object.entries(options)) {
       if (!datasourceNames.includes(key)) {
-        const didYouMean =
-          getDidYouMean(key, datasourceNames) || ` Available datasources: ${datasourceNames.join(', ')}`
+        const didYouMean = getDidYouMean(key, datasourceNames) || ` Available datasources: ${datasourceNames.join(', ')}`
```
millsp, line 23: Prettier makes your code pretty :troll:

packages/cli/src/Studio.ts
millsp: Simpler is better.

millsp (approved):

janpio (commented):
Nice changes to the tests, thank you!

Issue #20302, opened 2023-07-20
CLI errors with `TypeError [ERR_INVALID_URL]: Invalid URL` when `HTTP(S)_PROXY` en var has is set to a URL without a protocol


JMH0113:
### Bug description

npx prisma studio
cue me：
Downloading Prisma engines for Node-API for windows [                    ] 0%Error: Invalid URL

"prisma": "^5.0.0"

### How to reproduce

npx prisma xxx

### Prisma information

```prisma
datasource db {
  provider = "mysql"
  url      = env("DATABASE_URL")
}
```

### Environment & setup

- OS: windows
- Database: MySQL
- Node.js version: V16.18.0

### Prisma Version

```
5.0.0
```

Jolg42:
@JMH0113 This is unexpected to me. 🤔 

Could you try to run `DEBUG=* npx prisma -v` and share the output here?

JMH0113:
I do not know run DEBUG=* npx prisma -v,please tell me step by step.
I run DEBUG=* npx prisma -v by vscode terminal,has error.

Jolg42:
@JMH0113 Sorry, this should work better on Windows 
```
set DEBUG=* && npx prisma -v
```

JMH0113:
@Jolg42 Here's what it looks like when you run `set DEBUG=* && npx prisma -v`
prisma:engines  binaries to download libquery-engine, migration-engine +0ms
  prisma:download  file D:\pyq-api-node\node_modules\@prisma\engines\query_engine-windows.dll.node does not exist and must be downloaded +0ms
  prisma:download  file D:\pyq-api-node\node_modules\@prisma\engines\migration-engine-windows.exe does not exist and must be downloaded +1ms
  prisma:download  Downloading https://binaries.prisma.sh/all_commits/4bc8b6e1b66cb932731fb1bdbbc550d1e010de81/windows/query_engine.dll.node.gz to D:\pyq-api-node\node_modules\@prisma\engines\query_engine-windows.dll.node +2ms
> Downloading Prisma engines for Node-API for windows [                    ] 0%Error: TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:387:5)
    at URL.onParseError (node:internal/url:565:9)
    at new URL (node:internal/url:641:5)
    at new HttpsProxyAgent2 (D:\pyq-api-node\node_modules\@prisma\engines\dist\index.js:20628:50)
[...]

Issue #20193, opened 2023-07-12
Postinstall script fails in 5.0.0 with ECONNRESET


Soulusions:
### Bug description

I just upgraded a working project to prisma 5 and now simply trying to run `pnpm install` causes an error as follows:
```
node_modules/.pnpm/@prisma+engines@5.0.0/node_modules/@prisma/engines: Running postinstall script, failed in 50.7s
.../node_modules/@prisma/engines postinstall$ node scripts/postinstall.js
│ node:internal/process/promises:288
│             triggerUncaughtException(err, true /* fromPromise */);
│ Error: read ECONNRESET
│     at TLSWrap.onStreamRead (node:internal/stream_base_commons:217:20) {
│   errno: -4077,
│   code: 'ECONNRESET',
│   syscall: 'read',
│   attemptNumber: 1,
│   retriesLeft: 2
│ }
│ Node.js v18.16.0
```

### How to reproduce

1. Set prisma CLI version as `^5.0.0`
2. Run `pnpm i`

### Expected behavior

Postinstall script should succeed

### Prisma information

```prisma
model User {
  id    Int     @id @default(autoincrement())
  email String  @unique
  name  String?
  posts Post[]
}

model Post {
  id        Int     @id @default(autoincrement())
  title     String
  content   String?
  published Boolean @default(false)
  author    User    @relation(fields: [authorId], references: [id])
  authorId  Int
}
```

### Environment & setup

- OS: Windows 11
- Database: pgsql
- Node.js version: 18.16.0

### Prisma Version

```
5.0.0
```

Jolg42:
@Soulusions Could you try 
- Running `DEBUG=* pnpm install`?
- Running `DEBUG=* pnpm prisma -v`?
- Downloading manually https://binaries.prisma.sh/all_commits/6b0aef69b7cdfc787f822ecd7cdc76d5f1991584/windows/schema-engine.exe.gz to see if that works?

Jolg42:
Note: this looks like a temporary network issue to us, so retrying might just work.

Soulusions:
Right, after further investigation this seemed to only be a problem when running the install command in the vsc cli, from a regular old terminal everything went fine, which is quite odd but whatever, thanks for the pointer though!

luizfelipelaviola:
I got the same error using the terminal inside VSCode. Using the terminal out of VSCode works. 🤷🏼‍♂️

janpio:
Also on Windows as @Soulusions?

[...]

PR #20419, opened 2023-07-27
feat(fetch-engine): better debug/errors & retry checksum too

Jolg42:
/integration

Related
https://github.com/prisma/prisma/issues/20302
https://github.com/prisma/prisma/issues/20193

This improves our download
- I added a retry for the checkshum
	- and added tests for retries for checksum and engine file 
- Better error / debug for `HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`
	- and added tests 
	- show a better error message if the proxy URL is invalid (because parsing it failed)
	- ignores the error and tries downloading without the proxy

Note about these environment variables in general, I checked if there was any specification for them, I didn't find any.

It looks like a very popular convention that might have originated with curl, see
https://superuser.com/a/1166790
https://curl.se/docs/manpage.html

> The environment variables can be specified in lower case or upper case. The lower case version has precedence. http_proxy is an exception as it is only available in lower case.
> 
> Using an environment variable to set the proxy has the same effect as using the [-x, --proxy](https://curl.se/docs/manpage.html#-x) option.
> 
> http_proxy [protocol://]<host>[:port]
> 
> Sets the proxy server to use for HTTP.
> 
> HTTPS_PROXY [protocol://]<host>[:port]
> 
> Sets the proxy server to use for HTTPS.
> 
> [url-protocol]_PROXY [protocol://]<host>[:port]
> 
> Sets the proxy server to use for [url-protocol], where the protocol is a protocol that curl supports and as specified in a URL. FTP, FTPS, POP3, IMAP, SMTP, LDAP, etc.
> 
> ALL_PROXY [protocol://]<host>[:port]
> 
> Sets the proxy server to use if no protocol-specific proxy is set.
> 
> NO_PROXY <comma-separated list of hosts/domains>
> 
> list of host names that should not go through any proxy. If set to an asterisk '*' only, it matches all hosts. Each name in this list is matched as either a domain name which contains the hostname, or the hostname itself.
> 
> This environment variable disables use of the proxy even when specified with the [-x, --proxy](https://curl.se/docs/manpage.html#-x) option. That is NO_PROXY=direct.example.com curl -x http://proxy.example.com/ http://direct.example.com/ accesses the target URL directly, and NO_PROXY=direct.example.com curl -x http://proxy.example.com/ http://somewhere.example.com/ accesses the target URL through the proxy.
> 
> The list of host names can also be include numerical IP addresses, and IPv6 versions should then be given without enclosing brackets.
> 
> Since 7.86.0, IP addresses can be specified using CIDR notation: an appended slash and number specifies the number of "network bits" out of the address to use in the comparison. For example "192.168.0.0/16" would match all addresses starting with "192.168".

I also checked around and from what I see the protocol is required when using `HTTP_PROXY` or `HTTPS_PROXY` env vars.

### DIFF
--- a/packages/fetch-engine/src/__tests__/download.test.ts
+++ b/packages/fetch-engine/src/__tests__/download.test.ts
@@ -694,7 +694,7 @@ It took ${timeInMsToDownloadAllFromCache2}ms to execute download() for all binar
           version: CURRENT_ENGINES_HASH,
         }),
       ).rejects.toThrow(
-        `Failed to fetch the engine file at https://binaries.prisma.sh/all_commits/b182127de581d0bf6ec89152a6e2fb4652be0f0a/rhel-openssl-3.0.x/libquery_engine.so.node.gz. 500 KO`,
+        `Failed to fetch the engine file at https://binaries.prisma.sh/all_commits/${CURRENT_ENGINES_HASH}/rhel-openssl-3.0.x/libquery_engine.so.node.gz. 500 KO`,
       )
 
       // Because we try to fetch 2 different checksum files before we even start downloading the binaries
--- a/packages/fetch-engine/src/__tests__/download.test.ts
+++ b/packages/fetch-engine/src/__tests__/download.test.ts
@@ -664,7 +664,7 @@ It took ${timeInMsToDownloadAllFromCache2}ms to execute download() for all binar
           version: CURRENT_ENGINES_HASH,
         }),
       ).rejects.toThrow(
-        `Failed to fetch sha256 checksum at https://binaries.prisma.sh/all_commits/b182127de581d0bf6ec89152a6e2fb4652be0f0a/rhel-openssl-3.0.x/libquery_engine.so.node.gz.sha256. 500 KO`,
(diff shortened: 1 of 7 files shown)


packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -642,6 +642,118 @@ It took ${timeInMsToDownloadAllFromCache2}ms to execute download() for all binar
...
+      await expect(
+        download({
+          binaries: {
+            [BinaryType.QueryEngineLibrary]: baseDirChecksum,
+          },
+          binaryTargets: ['rhel-openssl-3.0.x'],
+          version: CURRENT_ENGINES_HASH,
+        }),
+      ).rejects.toThrow(
+        `Failed to fetch sha256 checksum at https://binaries.prisma.sh/all_commits/b182127de581d0bf6ec89152a6e2fb4652be0f0a/rhel-openssl-3.0.x/libquery_engine.so.node.gz.sha256. 500 KO`,
```
janpio, line 667:
```suggestion
        `Failed to fetch sha256 checksum at https://binaries.prisma.sh/all_commits/${CURRENT_ENGINES_HASH}/rhel-openssl-3.0.x/libquery_engine.so.node.gz.sha256. 500 KO`,
```

Jolg42: Yep because tests are failing

SevInf (approved):

PR #20425, opened 2023-07-28
feat(client): allow extension to reference their $parent

millsp:
Improvement for Accelerate: it allows an override to call its parent. Currently Accelerate needs this pattern, but without it the way it is achieved breaks interactive transactions. Also introduces the helper as a public API (see tests).

```ts
client.$extends({
  model: {
    $allModels: {
      findFirst(args: any) {
        const ctx = Prisma.getExtensionContext(this)

        return ctx.$parent![ctx.$name!].findFirst(args)
      }
    }
  }
})
```

### DIFF
--- a/packages/client/src/runtime/core/model/applyModel.ts
+++ b/packages/client/src/runtime/core/model/applyModel.ts
@@ -48,7 +48,7 @@ export function applyModel(client: Client, dmmfModelName: string) {
     fieldsPropertyLayer(client, dmmfModelName),
     addObjectProperties(modelExtensions),
     addProperty('name', () => dmmfModelName),
-    addProperty('parent', () => client.parent),
+    addProperty('$parent', () => client._appliedParent),
   ]
 
   return createCompositeProxy({}, layers)
--- a/packages/client/src/runtime/core/model/applyModel.ts
+++ b/packages/client/src/runtime/core/model/applyModel.ts
@@ -34,9 +34,6 @@ const fluentProps = [
 ] as const
 const aggregateProps = ['aggregate', 'count', 'groupBy'] as const
 
-// used to retrieve the parent client from a model
-const MODEL_PARENT_CLIENT = Symbol.for('prisma.client.model.parent')
-
 /**
  * Dynamically creates a model interface via a proxy.
  * @param client to trigger the request execution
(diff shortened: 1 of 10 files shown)


packages/client/src/runtime/core/model/applyModel.ts
```diff
@@ -34,29 +34,45 @@ const fluentProps = [
...
   ]
 
-  const modelExtensions = client._extensions.getAllModelExtensions(dmmfModelName)
-  if (modelExtensions) {
-    layers.push(addObjectProperties(modelExtensions))
+  const model = createCompositeProxy({}, getDefaultLayers())
+  const modelKeys = Object.keys(model)
+
+  // when a model extension override calls its own name, we refer to its parent
+  // eg. calling findUnique in accelerate's own findUnique calls client.findUnique
```
janpio, line 60: ´client` here is always the Client how it was before the Accelerate extension was applied, not the fully unmodified one, correct?

millsp: It's the client with all extensions minus any override to the builtin methods.

SevInf:
Hm, what if I want to patch the method and call patched method from a different one? 

```ts
client.$extends({
    model: {
      someModel: {
        findFirst() {
        }
      
        myHelper() {
           Prisma.getExtensionContext(this).findFirst(...) // I want my find first here
        }
      }
    }
})
```

I am sorry, but the way this PR works looks too magical and unintuitive for me: some things in a context refer to current extension and some refer to parent, but only for model extensions, and not, for example, for client. Additionally, it means that if we are introducing new method/field on models, it's a potentially breaking change for extensions.

I think I'd rather prefer an explicit API for accessing the parent so that Accelerate could use it. Something like:

```ts
Prisma.getExtensionParent(this).findFirst()
```

or 

```ts
Prisma.getExtensionContext(this).parent.findFirst()
```

SevInf:
On a further thought, i think, we might not even need extra API and can just make current thing work with transactions. We are [re-applying model](https://github.com/prisma/prisma/blob/5585688b54d50cb8bcd43f75756f4cae44555c4a/packages/client/src/runtime/getPrismaClient.ts#L781) and client extensions to ITX client. So, the only reason why this:

```ts
Prisma.defineExtension(client => client.$extends({
   $allModels: {
        findFirst() {
             return client.someModel.findFirst()
        }
   }
}))
```

breaks in transaction is because we resolve callback in `$extends` and that captures non-itx client. If instead we re-execute callback every time we re-apply the extension (with  appropriate client) that would work correctly.

millsp:
> If instead we re-execute callback every time we re-apply the extension (with appropriate client) that would work correctly

Would you consider that a breaking change?

> Hm, what if I want to patch the method and call patched method from a different one?

I think that should call the override properly, could use a test.

> Prisma.getExtensionContext(this).parent.findFirst()

Sounds good as well

SevInf:
>Would you consider that a breaking change?

Probably. In some cases we suggested to init the state of the extension there and this might break if we start calling it multiple times. Maybe, signature for future should be `() => (client) => extension`, so you could have a place to keep a state (first callback) but still have ability to re-apply the extension to ITXClient.

So, what do you think if for now we'd go with `.parent` API on context and consider re-application for the future?

millsp:
Yes, I am happy with the parent API too. I guess we'd return `rawClient` in this case, right?

SevInf:
`rawClient` does not have model proxies yet. I think simplest thing to do probably would be to keep parent around when we do `$extends` and forward that to the callbacks.

millsp:
I implemented `ctx.parent` and introduced a bunch of extra tests on top of the previous ones, as this is now a public API.

millsp:
I went ahead and deprecated `name` in favor of `$name`, we can keep it around until Prisma 6.

SevInf (approved):

PR #20438, opened 2023-07-28
fix(client): Ensure result extensions are applied after all query extensions

### LINKED ISSUE #20437: Client extensions: result extensions should be applied after query extensions
f you are short-curcuting query call, result extensions would also be skipped. For example:

```ts
prisma.$extends({
    query: {
        $allOperations({ query, args }) {
              if (someCondition) return someMockedResponse
              return query(args)
         }
    }
}).$extends({
   result: {
      someModel: {
         computedField: { ...}
      }
   }
})
```

Imagine they are two separate extensions, not aware of each other and someMockedResponse perfectly mimics DB response for particular query. From the standpoint of mocking extension, it does everything correctly. However, return value of query extension acts as final decider on what gets returned from the query. So, in case mocking condition holds true, someMockedResponse would act as one and final result of the query and result extension will get skipped. So, mocked response won’t contain computedField, unlike normal request.

Result extensions should apply only and always after all query extensions are done and should not be observable in any way from within query extension.

[Internal discussion](https://prisma-company.slack.com/archives/C043X6JBND6/p1690462237248899)

SevInf:
Fix #20437

/integration

### DIFF
--- a/packages/client/src/runtime/getPrismaClient.ts
+++ b/packages/client/src/runtime/getPrismaClient.ts
@@ -33,10 +33,9 @@ import {
 } from './core/engines'
 import { prettyPrintArguments } from './core/errorRendering/prettyPrintArguments'
 import { $extends } from './core/extensions/$extends'
+import { applyAllResultExtensions } from './core/extensions/applyAllResultExtensions'
 import { applyQueryExtensions } from './core/extensions/applyQueryExtensions'
-import { applyResultExtensions } from './core/extensions/applyResultExtensions'
 import { MergedExtensionsList } from './core/extensions/MergedExtensionsList'
-import { visitQueryResult } from './core/extensions/visitQueryResult'
 import { checkPlatformCaching } from './core/init/checkPlatformCaching'
 import { serializeJsonQuery } from './core/jsonProtocol/serializeJsonQuery'
 import { MetricsClient } from './core/metrics/MetricsClient'
@@ -44,7 +43,6 @@ import {
   applyModelsAndClientExtensions,
   unApplyModelsAndClientExtensions,
 } from './core/model/applyModelsAndClientExtensions'
-import { dmmfToJSModelName } from './core/model/utils/dmmfToJSModelName'
 import { rawCommandArgsMapper } from './core/raw-query/rawCommandArgsMapper'
 import { RawQueryArgs } from './core/raw-query/RawQueryArgs'
 import {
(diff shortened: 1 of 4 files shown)


packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -893,6 +900,29 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
       })
     }
 
+    _applyResultExtensions(result: object | null, modelName: string, args: JsArgs) {
```
millsp, line 903: +1 for moving it to another file near applyQueryExtensions if possible.

millsp (approved):

packages/client/src/runtime/getPrismaClient.ts
SevInf: Done.

greg-nagy:
@SevInf I am not sure about the original intention but I assume,  that the result extensions should also work for nested queries as well. At the moment if the model is loaded through a select, then the result extensions are not applied because the model is the "parent" model. 

See my comment for a bit more detail: https://github.com/prisma/docs/issues/5217#issuecomment-1696474897

What would be a good next step here?

SevInf:
@greg-nagy [Result extensions](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/result) are intended to work with nested read and writes and we have several tests for that inside of our repo. If that's not the case - please, open a separate issue and share all reproduction steps. 
This PR generally has nothing to do with this. You are incorrect about `applyAllResultExtensions` function being introduced here - it existed ever since result extensions were introduced, just in the difference place under different name.

greg-nagy:
thanks for the response! 

>  are intended to work with nested read and writes ... If that's not the case - please, open a separate issue and share all reproduction steps.

That doesn't seem to be the case — although I am 1 version behind atm. I will create reproduction steps and open an issue when I get there. 

Until then

> You are incorrect about applyAllResultExtensions function being introduced here - it existed ever since result extensions were introduced, just in the difference place under different name.

Might be the case easily. Thanks for the clarification. I just checked the last PR that changed the relevant parts.

greg-nagy:
@SevInf You have just updated your docs with this limitation https://github.com/prisma/docs/pull/5246/files

Does this mean that you are aware of the issue and no need to open a new issue from me?

SevInf:
@greg-nagy, no, I believe that doc change was incorrect, we are working on it.
I'd appreciate if we continued this conversation in a dedicated issue with a reproduction, rather than on unrelated closed PR.

Issue #19819, opened 2023-06-16
Upgrade from 4.12 to 4.15: Prisma can no longer find schema.prisma file when launching Electron packaged executable


dfoley-greenlight:
### Bug description

App written in Electron. Build steps to produce .exe all still work as expected. However, on launch, error is thrown saying Prisma cannot find schema.prisma. Only change is upgrade from 4.12 to 4.15. Bundler still places all the relevant files (query-engine & schema.prisma) in the correct locations. No problems when running in Dev.

### How to reproduce

-

### Prisma information

generator client {
  provider        = "prisma-client-js"
  output          = "../src/main/db/generated"
  previewFeatures = ["multiSchema"]
  binaryTargets   = ["native"]
}

### Environment & setup

Windows 10
SQL Server
Node: 18.12.1

### Prisma Version

Prisma Version 4.15

dfoley-greenlight:
![image](https://github.com/prisma/prisma/assets/128077018/05834e0d-27ea-422a-8903-862034e42520)

![image](https://github.com/prisma/prisma/assets/128077018/68c70988-7142-4ad1-9bcb-6335f704b55f)

Left = 4.15
Right = 4.12.

Seems to be that i bundle to "dest/main/db/generated" which matches second dirname pattern, but 4.15 hard-codes "src/main/db/generated" if __dirname DNE - i.e. if bundled.

janpio:
Can you please share the actual error message you are getting?

lemonpole:
I'm running into this issue as well after upgrading from `prisma@9.x` to `prisma@16.x`.

- Electron: `25.2.0`
- Prisma: `4.16.2`
- SQLite3: `5.1.6`

This is my `prisma/schema.prisma` file:
```
// This is your Prisma schema file,
// learn more about it in the docs: https://pris.ly/d/prisma-schema

generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
  url      = "file:save_0.db"
}
```

After creating the production bundle, my resources folder looks like this:

```
$ ls -la out/alpha/app-win32-x64/resources/
total 28
drwxr-xr-x 1 Foo Bar 197121     0 Sep 15 10:59 .
drwxr-xr-x 1 Foo Bar 197121     0 Sep 15 10:59 ..
drwxr-xr-x 1 Foo Bar 197121     0 Sep 15 10:59 .prisma
drwxr-xr-x 1 Foo Bar 197121     0 Sep 15 10:59 app
[...]

PR #20447, opened 2023-07-29
fix(client): add extra schema search location

millsp:
Fixes schema not found issues for SST and Electron.

Closes https://github.com/prisma/team-orm/issues/224

/integration


packages/client/src/generation/utils/buildDirname.ts
```diff
@@ -34,7 +34,16 @@ const fs = require('fs')
 
 config.dirname = __dirname
 if (!fs.existsSync(path.join(__dirname, 'schema.prisma'))) {
-  config.dirname = path.join(process.cwd(), ${JSON.stringify(pathToPosix(relativeOutdir))})
+  const alternativePaths = [
+    ${JSON.stringify(pathToPosix(relativeOutdir))},
+    ${JSON.stringify(pathToPosix(relativeOutdir).split('/').slice(1).join('/'))},
```
janpio, line 39: This is the new path?
What does that "translate" to? Maybe a comment to describe for all the alternative paths as an example?

jkomyno: Hey @millsp, wonder if you've seen this comment? (Tagging you in Slack as well)

millsp: It will be equivalent to what we had before (see last picture [here](https://github.com/prisma/prisma/issues/19819#issuecomment-1597138099)).

Jolg42: Maybe add a comment in the code?

packages/client/tests/e2e/schema-not-found-sst-electron/tests/main.ts
```diff
@@ -0,0 +1,22 @@
...
+const consoleMock = jest.spyOn(global.console, 'warn').mockImplementation()
+
+afterEach(() => {
+  consoleMock.mockClear()
+})
+
+test('bundled prisma client will re-use the schema.prisma via cwd', async () => {
+  process.chdir(path.join(__dirname, '..', 'dist'))
+
+  console.log('cd', process.cwd())
```
Jolg42, line 12:
```suggestion
  // console.log('cd', process.cwd())
```
Maybe?

Jolg42 (approved):

packages/client/src/generation/utils/buildDirname.ts
millsp: Done

janpio:
This should probably get a test in ecosystem-tests?

millsp:
Since it can be kept local, I opted for reproducing the behavior in [e2e](https://github.com/prisma/prisma/pull/20447/files#diff-0da33ecfbc5e8afcbbf4842b24a00690ac5799e07c6934301131a072f74241eaR12). Ideally yes, we should have tests for SST as well and Electron.

PR #20454, opened 2023-07-31
ci(daily buildpulse): un-schedule some runs

Jolg42:
We currently schedule this 8 times every day, we could have the same result with fewer runs.

[skip ci]

See https://github.com/prisma/prisma/actions/workflows/daily-buildpulse.yml
Screenshot 2023-07-31 at 13 43 47

### DIFF
--- a/.github/workflows/daily-buildpulse.yml
+++ b/.github/workflows/daily-buildpulse.yml
@@ -3,8 +3,8 @@ name: Daily BuildPulse run
 on:
   workflow_dispatch:
   schedule:
-    # https://crontab.guru - “At every 30th minute past every hour from 3 through 6.”
-    - cron: '*/30 3-6 * * *'
+    # https://crontab.guru - “At every 30th minute past every hour from 3 through 4.”
+    - cron: '*/30 3-4 * * *'
 
 jobs:
   run_tests:


.github/workflows/daily-buildpulse.yml
```diff
@@ -3,8 +3,8 @@ name: Daily BuildPulse run
 on:
   workflow_dispatch:
   schedule:
-    # https://crontab.guru - “At every 30th minute past every hour from 3 through 6.”
-    - cron: '*/30 3-6 * * *'
+    # https://crontab.guru - “At every 30th minute past every hour from 3 through 4.”
```
SevInf, line 6: That means just 2 builds? I am not quite sure that would be enough for flakiness detection.

Jolg42: We only need 2 runs on the same commit for Buildpulse technically, but this would get for 4 runs, actually.
The other problem is that we are not doing anything with the reports currently anyway?

Before = 8 runs (UTC time)
- 3:00
- 3:30
- 4:00
- 4:30
- 5:00
- 5:30
- 6:00
- 6:30

After = 4 runs
- 3:00
- 3:30
- 4:00
- 4:30

jkomyno (approved):

SevInf (approved):

PR #20457, opened 2023-07-31
chore(get-platform): revert "search for openssl in LD_LIBRARY_PATH if provided"

aqrln:
This reverts commit d339a812b0b22298af01c4054974ee32aea1b256 (partial revert of https://github.com/prisma/prisma/pull/20381).

As it currently stands, it's a breaking change for AWS Lambda (including Netlify and Vercel Serverless functions) users. We will make the workaround from https://github.com/prisma/prisma/pull/17241 work across directories to avoid the breaking change and re-apply this patch and land it for 5.2.0 after the 5.1.0 release.

Long term, for Prisma 6, it would be nice to use OpenSSL 3 that Lambda with Node.js 18 layer provides in `/var/lang/lib`. Alternatively, we could provide infer all supported binary targets for a given platform, and make it possible to use either `rhel-openssl-1.0.x` or `rhel-openssl-3.0.x`, whichever was generated, and only potentially change what the `native` binary target would be. This can be done in a minor release.

### DIFF
--- a/packages/get-platform/src/__tests__/getSSLVersion.test.ts
+++ b/packages/get-platform/src/__tests__/getSSLVersion.test.ts
@@ -6,34 +6,11 @@ const describeIf = (condition: boolean) => (condition ? describe : describe.skip
 const ctx = jestContext.new().assemble()
 
 describeIf(process.platform === 'linux')('computeLibSSLSpecificPaths', () => {
-  const originalEnv = process.env
-
-  afterEach(() => {
-    process.env = originalEnv // Restore old environment
-  })
-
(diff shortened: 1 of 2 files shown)


Jolg42 (approved):

Jolg42:
Note: ecosystem-tests in progress
https://github.com/prisma/ecosystem-tests/actions/runs/5715430924
optional https://github.com/prisma/ecosystem-tests/actions/runs/5715430919

aqrln:
vercel and netlify-cli now pass, netlify-github fails for unrelated reason (404 with an html page that says "site not found")

PR #20467, opened 2023-07-31
ci: move PDP send event to prisma/prisma

Jolg42:
Before we switched to GitHub Actions we were using: https://github.com/prisma/prisma-engines/blob/main/.github/workflows/send-tag-event.yml

Now we can move that here where the publishing script is defined in GH


Jolg42:
Note: it's something we can do AFTER the release.

PR #20475, opened 2023-08-01
chore(*): use fs-extra everywhere as default

Jolg42:
WIP


packages/migrate/src/commands/DbExecute.ts
```diff
@@ -158,6 +158,7 @@ See \`${green(getCommandWithExecutor('prisma db execute -h'))}\``,
       datasourceType = {
         tag: 'url',
         url: args['--url'],
+        f,
```
Jolg42, line 161:
```suggestion
```

PR #20504, opened 2023-08-02
fix(client): Make sure result extensions don't break `.count`

### LINKED ISSUE #20499: Prisma Client errors with "TypeError: Cannot create proxy with a non-object as target or handler" when using result client extension with no `needs` and `count` method
### Bug description

Just upgraded to 5.1.0, and after adjusting the Array shortcut breaking change, I have run the testing suite and this test fails, I'm not quite sure how to debug this.

Any help or question is appreciated and I would love to help.

```bash
 FAIL  src/controllers/v1/users.controller.test.ts > Users > Search > should search
TypeError: Cannot create proxy with a non-object as target or handler
 ❯ xe ../../packages/prisma-client/client/runtime/library.js:26:5353
 ❯ Ds ../../packages/prisma-client/client/runtime/library.js:112:2696
 ❯ visitor ../../packages/prisma-client/client/runtime/library.js:112:3823
 ❯ $r ../../packages/prisma-client/client/runtime/library.js:112:3062
 ❯ Is ../../packages/prisma-client/client/runtime/library.js:112:3751
 ❯ l ../../packages/prisma-client/client/runtime/library.js:126:10330
 ❯ search src/controllers/v1/users.controller.ts:149:17
    147|     : { name: Prisma.SortOrder.asc };
    148|
    149|   const count = await prisma.user.count();
```

Our User schema is quite big so it's complicated to share.

I'm using Postgres, Node 18.13.0, Typescript 5.0.4 and a PNPM monorepo if it helps. 

### How to reproduce

-

### Expected behavior

_No response_

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
```

### Environment & setup

- OS: Macosx
- Database: Postgresql
- Node.js version: 18.13.0

### Prisma Version

```
5.1.0
```

SevInf:
Problem: after #20438, we were trying to apply result extensions to
"sugared" result of `.count` method - plain number. Theoretically,
extensions should not apply there at all, but field with no `needs`
applies to every result of a corresponding model. And since result
extensions use proxies and proxy can not be created over non-objects,
the code thrown very undescriptive error instead.

Fix #20499

### DIFF
--- a/packages/client/src/runtime/core/extensions/applyAllResultExtensions.ts
+++ b/packages/client/src/runtime/core/extensions/applyAllResultExtensions.ts
@@ -24,6 +24,10 @@ export function applyAllResultExtensions({
   extensions,
   runtimeDataModel,
 }: ApplyAllResultExtensionsParams) {
+  // We return the result directly (not applying result extensions) if
+  // - there is no extension to apply
+  // - result is `null`
+  // - result is not an object (e.g. `.count()`)
   if (extensions.isEmpty() || result == null || typeof result !== 'object') {
     return result
   }
--- a/packages/client/src/runtime/core/extensions/applyAllResultExtensions.ts
+++ b/packages/client/src/runtime/core/extensions/applyAllResultExtensions.ts
@@ -24,7 +24,7 @@ export function applyAllResultExtensions({
   extensions,
   runtimeDataModel,
 }: ApplyAllResultExtensionsParams) {
-  if (extensions.isEmpty() || result == null) {
+  if (extensions.isEmpty() || result == null || typeof result !== 'object') {
     return result
   }
   const model = runtimeDataModel.models[modelName]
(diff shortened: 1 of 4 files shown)


packages/client/src/runtime/core/extensions/applyAllResultExtensions.ts
```diff
@@ -24,7 +24,7 @@ export function applyAllResultExtensions({
   extensions,
   runtimeDataModel,
 }: ApplyAllResultExtensionsParams) {
-  if (extensions.isEmpty() || result == null) {
+  if (extensions.isEmpty() || result == null || typeof result !== 'object') {
```
Jolg42, line 27:
```suggestion
  // We return the result directly (not applying result extensions) if
  // - there is no extension to apply
  // - result is `null`
  // - result is not an object (e.g. `.count()`)
  if (extensions.isEmpty() || result == null || typeof result !== 'object') {
```
maybe?

Jolg42 (approved):

Issue #16418, opened 2022-11-23
Transpile of PrismaClient constructor taking a long time


kennytraction:
### Bug description

TypeScript is taking a long time (>3 minutes) to transpile code that includes creating a new instance of `PrismaClient`. It seems much of that time is spent type checking the `PrismaClient` constructor.

### How to reproduce

Use a larger schema with the following few lines of code. Run `tsc` on the project containing this code.

```
import { PrismaClient } from '@prisma/client'
export class MyPrisma {
  prisma: PrismaClient
  constructor() {
    this.prisma = new PrismaClient()  // This lines takes ~3 minutes to transpile
  }
}
```

Removing the line that creates an instance of `PrismaClient` results in the code transpiling very quickly.

### Expected behavior

I expect that transpiling code that includes using the `PrismaClient` constructor to not take minutes.

### Prisma information

Our schema currently has ~247 models. We may be able to share it privately.

### Environment & setup

- OS: Windows, MacOS, Docker
- Database: PostgreSQL
- Node.js version: 14.17.0

### Prisma Version

```
prisma                  : 4.2.1
@prisma/client          : 4.2.1
Current platform        : windows
Query Engine (Node-API) : libquery-engine 2920a97877e12e055c1333079b8d19cee7f33826 (at ..\node_modules\@prisma\engines\query_engine-windows.dll.node)
Migration Engine        : migration-engine-cli 2920a97877e12e055c1333079b8d19cee7f33826 (at ..\node_modules\@prisma\engines\migration-engine-windows.exe)
Introspection Engine    : introspection-core 2920a97877e12e055c1333079b8d19cee7f33826 (at ..\node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary           : prisma-fmt 2920a97877e12e055c1333079b8d19cee7f33826 (at ..\node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash    : 2920a97877e12e055c1333079b8d19cee7f33826
Studio                  : 0.469.0
```

millsp:
[...]

PR #20536, opened 2023-08-03
test(client): ensure that odoo schema can compile under a minute #16418

millsp:
#16418

### DIFF
--- a/packages/client/tests/e2e/16418-slow-compilation-big-schema/_steps.ts
+++ b/packages/client/tests/e2e/16418-slow-compilation-big-schema/_steps.ts
@@ -1,0 +1,27 @@
+import { $ } from 'zx'
+
+import { executeSteps } from '../_utils/executeSteps'
+
+void executeSteps({
+  setup: async () => {
+    process.env.PRISMA_SKIP_POSTINSTALL_GENERATE = 'true'
+    await $`pnpm install`
+    await $`pnpm prisma generate`
+  },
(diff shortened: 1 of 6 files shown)


Jolg42:
What do we want to do with this PR @millsp ?

millsp:
I think we can review, merge, and close the original issue?

packages/client/tests/e2e/16418-slow-compilation-big-schema/_steps.ts
```diff
@@ -0,0 +1,27 @@
...
+  test: async () => {
+    const timeBefore = Date.now()
+    await $`pnpm exec tsc`
+    const timeAfter = Date.now()
+
+    const timeDiff = timeAfter - timeBefore
+    console.log(`timeDiff: ${timeDiff}`)
+
+    if (timeDiff > 1000 * 60) {
+      throw new Error('Compilation took more than 1 minute')
```
Jolg42, line 20: How much time does it take at the moment?

Jolg42 (approved):

packages/client/tests/e2e/16418-slow-compilation-big-schema/_steps.ts
millsp: I think it's about 40s on CI. I shared that in the original issue we should probably measure instantiations, and memory instead of time. I hope to get to this some day :smile:

PR #20590, opened 2023-08-07
refactor(cli): remove data proxy preview feature checks in generate

millsp:
closes https://github.com/prisma/team-orm/issues/264


packages/internals/src/get-generators/utils/check-feature-flags/checkFeatureFlags.ts
```diff
@@ -1,26 +1,12 @@
...
-  options.dataProxy === true &&
-    config.generators.some((generatorConfig) => {
-      return generatorConfig.previewFeatures.some((feature) => {
-        if (feature.toLocaleLowerCase() === 'metrics'.toLocaleLowerCase()) {
-          throw new Error(forbiddenPreviewFeatureWithDataProxyFlagMessage('metrics'))
-        }
-      })
-    })
-}
+export function checkFeatureFlags(_config: ConfigMetaFormat, _options: GetGeneratorOptions) {}
```
Jolg42, line 9: Why do we need to keep `checkFeatureFlags`?

millsp: No really but thought we could leave it there for future. I can remove if you want.

Jolg42: It's ok for me to keep like this 👍🏼

Jolg42 (approved):

PR #20591, opened 2023-08-07
refactor(cli): remove dataproxy generate env var

millsp:
Will make `ecosystem-tests/vercel-cli-serverless-functions` fail terporarily, until dynamic switch via URL is implemented.

closes https://github.com/prisma/team-orm/issues/265

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -292,11 +292,7 @@ Please run \`${getCommandWithExecutor('prisma generate')}\` to see the errors.`)
               printDownloadProgress: !watchMode,
               version: enginesVersion,
               cliVersion: pkg.version,
-              dataProxy:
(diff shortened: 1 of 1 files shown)


Jolg42 (approved):
Do we plan to add a warning maybe when using the env vars saying that they can be removed?

millsp:
I asked Jan about his opinion on this. I am of the opinion that we say nothing, so we don't add extra work on users. Also, where we are headed to, `--data-proxy` and `--accelerate` will still exist but do nothing, so they seem to work as usual. We will probably remove these in Prisma 6. So the behavior is "kinda" equivalent between env vars and `--data-proxy`, they just do nothing.

janpio:
(We also have _very_ few users of these env vars, so the additional effort will not affect a lot of people - that was my killer argument :D )

PR #20594, opened 2023-08-08
feat(client, cli): dynamic prisma:// urls engine loading

### LINKED ISSUE #12734: Local Prisma Studio does not work with Data Proxy
### Bug description

Using the local `prisma studio` with mongodb via PDP fails.

### How to reproduce

1. Create a new prisma project locally using MongoDB (I used MongoDB Atlas)
2. Create a Prisma Data Platform project and configure it to use the mongodb
2. Make sure you enabled the `dataProxy` preview feature and generate the client locally (Instructions at https://www.prisma.io/docs/concepts/data-platform/data-proxy)
3. Open prisma studio (`npx prisma studio`) and try to see the first records
5. I got the following error 
```
Message: Error in Prisma Client request: 

Could not parse URL of the datasource
  
Query:
{
  "modelName": "User",
  "operation": "findMany",
  "args": {
    "take": 100,
    "skip": 0,
    "select": {
      "id": true,
      "name": true
    }
  }
}
```

<details>
  <summary>Image of the error</summary>
error

</details>

### Expected behavior

I was expecting to use the studio as I normally do.

##### Using MongoDB and `prisma studio` directly

I swapped the `prisma://...` url with the `mongodb+srv://...` and regenerated the client and the it worked.

##### Using MongoDB and PDP as an ORM

I followed the same steps to reproduce the bug but I used a simple script to interact with the DB and it worker.

<details>
  <summary>index.ts</summary>
  
```typescript
import { v4 } from `uuid`


async function main(){
        const name = v4();
        const user = await prisma.user.create({
                data: {
                        name: name
                }
        });
        console.log(user);

        const user_get = await prisma.user.findMany({
            "take": 100,
            "skip": 0,
            "select": {
              "id": true,
              "name": true
            }

        });
        console.log(user_get);
}

main()
```
</details>

### Prisma information
<details>
  <summary> The `schema.prisma`</summary>
  
```
generator client {
  provider = "prisma-client-js"
  previewFeatures = ["dataProxy"]

}

datasource db {
  provider = "mongodb"
  url      = "prisma://aws-us-east-1.prisma-data.com/?api_key=******"
}

model User {
  id String @id @default(auto()) @map("_id") @db.ObjectId
  name String
}
```
</details>

<details>
  <summary>And here is the full error logs (`DEBUG='*'`)</summary>
  
```
  express:router dispatching POST /api +8s
  express:router query  : /api +0ms
  express:router expressInit  : /api +0ms
  express:router textParser  : /api +0ms
  body-parser:text content-type "text/plain" +0ms
  body-parser:text content-encoding "identity" +0ms
  body-parser:text read body +1ms
  body-parser:text parse body +0ms
  express:router <anonymous>  : /api +0ms
  express:router serveStatic  : /api +0ms
  prisma:studio-server Incoming request:  {"requestId":6,"channel":"prisma","action":"clientRequest","payload":{"data":{"schemaHash":"a872c6d0ea25ff99e01fab453268d071","modelName":"User","operation":"count"}}} +8s
  prisma:studio-pcw [getPrismaClient] {
  prismaClientPath: '/home/mmicu/work/git/prisma-mongo/node_modules/prisma/prisma-client',
  queryEngine: {
    type: 'binary',
    path: '/home/mmicu/work/git/prisma-mongo/node_modules/prisma/engines/22b822189f46ef0dc5c5b503368d1bee01213980/query-engine-debian-openssl-1.1.x'
  },
  previewFeatures: []
} +8s
  prisma:tryLoadEnv Environment variables not found at  +8s
  prisma:tryLoadEnv Environment variables not found at  +0ms
  prisma:tryLoadEnv No Environment variables loaded +1ms
  prisma:client dirname /home/mmicu/work/git/prisma-mongo/prisma +8s
  prisma:client relativePath  +0ms
  prisma:client cwd /home/mmicu/work/git/prisma-mongo/prisma +1ms
  prisma:client clientVersion: in-memory +1ms
  prisma:client clientEngineType: dataproxy +1ms
  prisma:studio-server Outgoing response:  {
  requestId: 6,
  channel: '-prisma',
  action: 'clientRequest',
  payload: {
    error: {
      type: undefined,
      code: 'P5002',
      message: 'Could not parse URL of the datasource',
      stack: 'Could not parse URL of the datasource'
    },
    data: null
  }
} +16ms
  express:router dispatching POST /api +29ms
  express:router query  : /api +0ms
  express:router expressInit  : /api +1ms
  express:router textParser  : /api +0ms
  body-parser:text content-type "text/plain" +0ms
  body-parser:text content-encoding "identity" +1ms
  body-parser:text read body +0ms
  body-parser:text parse body +2ms
  express:router <anonymous>  : /api +0ms
  express:router serveStatic  : /api +0ms
  prisma:studio-server Incoming request:  {"requestId":7,"channel":"prisma","action":"clientRequest","payload":{"data":{"schemaHash":"a872c6d0ea25ff99e01fab453268d071","modelName":"User","operation":"findMany","args":{"take":100,"skip":0,"select":{"id":true,"name":true}}}}} +17ms
  prisma:studio-pcw [getPrismaClient] {
  prismaClientPath: '/home/mmicu/work/git/prisma-mongo/node_modules/prisma/prisma-client',
  queryEngine: {
    type: 'binary',
    path: '/home/mmicu/work/git/prisma-mongo/node_modules/prisma/engines/22b822189f46ef0dc5c5b503368d1bee01213980/query-engine-debian-openssl-1.1.x'
  },
  previewFeatures: []
} +40ms
  prisma:tryLoadEnv Environment variables not found at  +39ms
  prisma:tryLoadEnv Environment variables not found at  +0ms
  prisma:tryLoadEnv No Environment variables loaded +1ms
  prisma:client dirname /home/mmicu/work/git/prisma-mongo/prisma +37ms
  prisma:client relativePath  +0ms
  prisma:client cwd /home/mmicu/work/git/prisma-mongo/prisma +0ms
  prisma:client clientVersion: in-memory +1ms
  prisma:client clientEngineType: dataproxy +0ms
  prisma:studio-server Outgoing response:  {
  requestId: 7,
  channel: '-prisma',
  action: 'clientRequest',
  payload: {
    error: {
      type: undefined,
      code: 'P5002',
      message: 'Could not parse URL of the datasource',
      stack: 'Could not parse URL of the datasource'
    },
    data: null
  }
} +16ms
  express:router dispatching POST /api +35ms
  express:router query  : /api +0ms
  express:router expressInit  : /api +0ms
  express:router textParser  : /api +0ms
  body-parser:text content-type "text/plain" +0ms
  body-parser:text content-encoding "identity" +1ms
  body-parser:text read body +0ms
  body-parser:text parse body +1ms
  express:router <anonymous>  : /api +1ms
  express:router serveStatic  : /api +0ms
  prisma:studio-server Incoming request:  {"requestId":8,"channel":"telemetry","action":"send","payload":{"data":{"command":"error_throw","commandDetails":"{\"type\":\"client\",\"description\":\"Unable to run script\"}","commandContext":"{\"model_count\":1}"}}} +23ms
  prisma:studio-server Outgoing response:  {
  requestId: 8,
  channel: '-telemetry',
  action: 'send',
  payload: { error: null, data: null }
} +1ms
```
</details>

### Environment & setup

- OS: Linux Mint 19 Tara
- Database: MongoDb (using MongoDB Atlas)
- Node.js:`v16.14.2`

### Prisma Version

```
prisma                  : 3.12.0
@prisma/client          : 3.12.0
Query Engine (Node-API) : libquery-engine 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt 22b822189f46ef0dc5c5b503368d1bee01213980 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : 22b822189f46ef0dc5c5b503368d1bee01213980
Studio                  : 0.459.0
Preview Features        : dataProxy

```

### LINKED ISSUE #13608: Postinstall hook always generates non Data Proxy Prisma Client
When you install a new version of `@prisma/client` or run `npm install` on a project that already has Prisma Client in dependencies, a postinstall hook runs `prisma generate --postinstall` under the hood to create a Prisma Client so the project can instantly be used with Prisma Client without having to manually run `prisma generate`.

As the Data Proxy is currently "enabled" by setting an environment variable when running `prisma generate` (and via running `prisma generate --data-proxy` with no additional env var in the near future), that means that by default the generated Client can never be a Data Proxy Client. The worst situation for this could be when you have a project with a Data Proxy Client, update Prisma Client, and suddenly do _not_ have a Data Proxy Client any more.

The same is also true for any CLI commands that generate a new Client under the hood - they always just run `prisma generate`.

Solutions could be to configure the Data Proxy permanently (via a property in the schema file), reading the previous Client, reacting to the database connection string value - or generally getting rid of the difference between a Data Proxy and non Data Proxy Client.

Workaround: Just run `prisma generate` with the data proxy env var (or `prisma generate --data-proxy` soon) to explicitly create a Prisma Client for Data Proxy explicitly.

### LINKED ISSUE #17796: Change how Data Proxy Client deals with unsupported preview features
Internal Discussion: https://prisma-company.slack.com/archives/CEYCG2MCN/p1675804667291339

millsp:
This PR brings the removal `--data-proxy` and enables the handling of `prisma://` URLs to happen dynamically instead.

closes https://github.com/prisma/team-orm/issues/284
closes https://github.com/prisma/prisma/issues/12734
closes https://github.com/prisma/prisma/pull/18128
closes https://github.com/prisma/prisma/issues/13608
closes https://github.com/prisma/team-orm/issues/285
closes https://github.com/prisma/team-orm/issues/292
closes https://github.com/prisma/team-orm/issues/286
closes https://github.com/prisma/team-orm/issues/291
closes https://github.com/prisma/team-orm/issues/290
closes https://github.com/prisma/team-orm/issues/289
closes https://github.com/prisma/prisma/issues/17796
closes https://github.com/prisma/team-orm/issues/266
closes https://github.com/prisma/team-orm/issues/288
closes https://github.com/prisma/team-orm/issues/287

/integration

The ecosystem-tests run can be seen [here](https://github.com/prisma/ecosystem-tests/actions/runs/5862093491). As you can see, we have some tests failing which are related to:
1. File changes: We are creating more files by default
1. Engine not being detected due to the removal of `_clientEngineType`

So nothing is fundamentally broken, just some assertions need to be updated. I will open a PR tomorrow.


packages/cli/src/Generate.ts
```diff
@@ -245,26 +238,10 @@ ${dim('```')}
...
-            ? `
-
-${
-  isDeno
-    ? 'To use Prisma Client with Deno and the Data Proxy, import it like this:'
-    : 'To use Prisma Client in edge runtimes like Cloudflare Workers or Vercel Edge Functions, import it like this:'
-}
-${dim('```')} 
-${highlightTS(`\
-import { PrismaClient } from '${importPath}/${isDeno ? 'deno/' : ''}edge${isDeno ? '.ts' : ''}'`)}
```
janpio, line 259: This probably needs to stick around for Deno.

packages/cli/src/Generate.ts
```diff
@@ -245,26 +238,10 @@ ${dim('```')}
...
 ${dim('```')}
 
-You will need an Accelerate or a Prisma Data Proxy connection string. See documentation: ${link(
-                'https://pris.ly/d/data-proxy',
-              )}
-`
-            : ''
-        }${breakingChangesStr}${versionsWarning}`
+See other options for edge runtimes and more: ${link('http://pris.ly/d/importing-client') /** TODO create link */}
```
millsp, line 243: Add edge case for Deno

janpio, line 243: Mayyyybe we still want to output the edge import?

packages/cli/src/Studio.ts
```diff
@@ -106,7 +94,7 @@ ${bold('Examples')}
     const staticAssetDir = path.resolve(__dirname, '../build/public')
 
     const schema = await fs.promises.readFile(schemaPath, 'utf-8')
-    const config = await getConfig({ datamodel: schema, ignoreEnvVarErrors: false })
+    const config = await getConfig({ datamodel: schema, ignoreEnvVarErrors: true })
```
millsp, line 97: Not required to throw an error here for env vars, Studio will display one anyways.

packages/cli/src/Studio.ts
```diff
@@ -89,8 +79,6 @@ ${bold('Examples')}
       return this.help(args.message)
     }
 
-    await checkUnsupportedDataProxy('studio', args, true)
```
millsp, line 92: Studio now works with Accelerate URLs. There will be a test, but not via this PR as there is enough content to go through. Consider this an opportunistic removal.

packages/cli/src/Generate.ts
```diff
@@ -98,8 +96,8 @@ ${bold('Examples')}
       '-h': '--help',
       '--watch': Boolean,
       '--schema': String,
-      '--data-proxy': Boolean,
-      '--accelerate': Boolean,
+      '--data-proxy': Boolean, // remove in Prisma 6
+      '--accelerate': Boolean, // remove in Prisma 6
```
millsp, line 100: Still exists for backwards compatibility but does nothing. The switch is now dynamic and Data Proxy is loaded based on the URL.

packages/cli/src/Generate.ts
```diff
@@ -136,11 +134,6 @@ ${bold('Examples')}
         printDownloadProgress: !watchMode,
         version: enginesVersion,
         cliVersion: pkg.version,
-        dataProxy:
```
millsp, line 139: This property was the representation of `--data-proxy`, now is also going away.

packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -39,20 +39,6 @@ describe('studio with alternative urls and prisma://', () => {
...
-
-    await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
-
-      Using the Data Proxy (connection URL starting with protocol prisma://) is not supported for this CLI command prisma studio yet. Please use a direct connection to your database via the datasource 'directUrl' setting.
-
-      More information about Data Proxy: https://pris.ly/d/data-proxy-cli
-
-    `)
-  })
-
```
millsp, line 55: This test now fails because the Data Proxy actually works with Accelerate URLs. I will add tests in a follow-up PR.

packages/client/helpers/build.ts
```diff
@@ -55,7 +55,8 @@ const edgeRuntimeBuildConfig: BuildOptions = {
   define: {
     // that helps us to tree-shake unused things out
     NODE_CLIENT: 'false',
-    TARGET_ENGINE_TYPE: '"data-proxy"',
+    // tree shake the Library and Binary engines out
+    TARGET_ENGINE_TYPE: '"edge"',
```
millsp, line 59: Just a change of name because there is no runtime called `data-proxy.js` anymore.

packages/client/helpers/build.ts
```diff
@@ -112,12 +113,10 @@ void build([
   generatorBuildConfig,
   nodeRuntimeBuildConfig('binary'),
   nodeRuntimeBuildConfig('library'),
-  nodeRuntimeBuildConfig('data-proxy'),
```
millsp, line 115: We don't have a special Data Proxy runtime anymore, instead Data Proxy is always included in either Library or Binary. That enables a seamless switch between engines just based on the URL contents.

packages/client/src/__tests__/integration/errors/client-version-error/test.ts
```diff
@@ -6,7 +6,7 @@ test('client-version-error', async () => {
   try {
     await prisma.user.findMany({ invalidArg: true })
   } catch (e) {
-    expect(e.clientVersion).toMatchInlineSnapshot(`client-test-version`)
+    expect(e.clientVersion).toMatchInlineSnapshot(`0.0.0`)
```
millsp, line 9: Opportunistic cleanup in the `getTestClient` required to change the snapshot.

packages/client/src/__tests__/integration/happy/client-engine/test.ts
```diff
@@ -92,7 +92,7 @@ function buildTests() {
             ]
             `)
 
-          expect(prisma._clientEngineType).toMatch(expectedClientEngine)
```
millsp, line 95: This property was removed, as it was really only used for this test. Now using an equivalent based on the generator config.

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/common.ts
```diff
@@ -9,15 +9,14 @@ if (isMacOrWindowsCI) {
   jest.setTimeout(80_000)
 }
 
-export const testGeneratedClient = (runtimeName: 'binary' | 'library' | 'dataProxy') => async () => {
+export const testGeneratedClient = (runtimeName: 'binary' | 'library') => async () => {
```
millsp, line 12: As said above, there's only two runtimes now. In a follow up PR, we will likely remove displaying them completely.

packages/client/src/generation/TSClient/Datasources.ts
```diff
@@ -1,10 +1,10 @@
+import { DataSource } from '@prisma/generator-helper'
 import indent from 'indent-string'
 
-import type { InternalDatasource } from '../../runtime/utils/printDatasources'
 import type { Generatable } from './Generatable'
 
 export class Datasources implements Generatable {
-  constructor(protected readonly internalDatasources: InternalDatasource[]) {}
+  constructor(protected readonly internalDatasources: DataSource[]) {}
```
millsp, line 7: Types were duplicated in the code, now we refer to a single source of truth. Cleanup was necessary.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -66,18 +62,7 @@ export class TSClient implements Generatable {
   }
 
   public async toJS(edge = false): Promise<string> {
-    const {
-      platforms,
-      generator,
-      sqliteDatasourceOverrides,
```
millsp, line 72: Not used.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -91,18 +76,23 @@ export class TSClient implements Generatable {
       generator.config.engineType = engineType
     }
 
+    const inlineSchema = (await readFile(schemaPath)).toString('base64')
+    const inlineSchemaHash = crypto.createHash('sha256').update(inlineSchema).digest('hex')
     const config: Omit<GetPrismaClientConfig, 'runtimeDataModel' | 'dirname'> = {
       generator,
       relativeEnvPaths,
-      sqliteDatasourceOverrides,
```
millsp, line 97: Not used.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -91,18 +76,23 @@ export class TSClient implements Generatable {
       generator.config.engineType = engineType
     }
 
+    const inlineSchema = (await readFile(schemaPath)).toString('base64')
+    const inlineSchemaHash = crypto.createHash('sha256').update(inlineSchema).digest('hex')
```
millsp, line 80: Inline schema and hash needed for the data proxy to work correctly.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -91,18 +76,23 @@ export class TSClient implements Generatable {
...
       datasourceNames: datasources.map((d) => d.name),
       activeProvider: this.options.activeProvider,
-      dataProxy: this.options.dataProxy,
       postinstall: this.options.postinstall,
       ciName: ciInfo.name ?? undefined,
+      inlineDatasources: datasources.reduce((acc, ds) => {
+        return (acc[ds.name] = { url: ds.url }), acc
+      }, {} as GetPrismaClientConfig['inlineDatasources']),
+      inlineSchema,
+      inlineSchemaHash,
```
millsp, line 95: Simplified datasource information needed by the dataproxy. In a follow up PR we will warn users if they have a hardcoded URL in their schema, as it would appear here otherwise.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -132,15 +122,18 @@ const config = ${JSON.stringify(config, null, 2)}
 ${buildDirname(edge, relativeOutdir)}
 ${buildRuntimeDataModel(this.dmmf.datamodel)}
 
-${await buildInlineSchema(dataProxy, schemaPath)}
```
millsp, line 135: Done directly in the configuration above.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -132,15 +122,18 @@ const config = ${JSON.stringify(config, null, 2)}
...
-${buildInlineDatasource(dataProxy, datasources)}
```
millsp, line 136: Same.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -132,15 +122,18 @@ const config = ${JSON.stringify(config, null, 2)}
...
 const PrismaClient = getPrismaClient(config)
 exports.PrismaClient = PrismaClient
 Object.assign(exports, Prisma)${deno ? '\nexport { exports as default, Prisma, PrismaClient }' : ''}
-${buildNFTAnnotations(dataProxy, engineType, platforms, relativeOutdir)}
+${buildNFTAnnotations(
+  edge || false /** TODO after removal of dataProxy, do if --no-engine or edge */,
+  engineType,
+  platforms,
+  relativeOutdir,
+)}
```
millsp, line 136: We don't build annotations for edge runtime builds, and in a follow up PR we will also introduce `--no-engine` which will lock a user to use Data Proxy.

packages/client/src/generation/utils/buildInlineDatasources.ts
millsp, line 1: Now done directly via the generated client config.

packages/client/src/generation/utils/buildInlineSchema.ts
millsp, line 1: Now done directly via the generated client config.

packages/client/src/generation/utils/buildNFTAnnotations.ts
```diff
@@ -16,13 +16,13 @@ import { map } from '../../../../../helpers/blaze/map'
...
-  dataProxy: boolean,
+  noEngine: boolean,
   engineType: ClientEngineType,
   platforms: Platform[] | undefined,
   relativeOutdir: string,
 ) {
-  // We don't want to bundle engines when `--data-proxy` is enabled
-  if (dataProxy === true) return ''
+  // We don't want to bundle engines when `--no-engine is enabled
+  if (noEngine === true) return ''
```
millsp, line 25: Not yet implemented on the CLI so this value is currently always false.

packages/client/src/runtime/core/engines/binary/BinaryEngine.ts
```diff
@@ -124,6 +122,13 @@ export class BinaryEngine extends Engine<undefined> {
     this.activeProvider = config.activeProvider
     this.connection = new Connection()
 
+    // compute the datasource override for binary engine
+    const dsOverrideName = Object.keys(config.overrideDatasources)[0]
+    const dsOverrideUrl = config.overrideDatasources[dsOverrideName]?.url
+    if (dsOverrideName !== undefined && dsOverrideUrl !== undefined) {
+      this.datasourceOverrides = [{ name: dsOverrideName, url: dsOverrideUrl }]
+    }
```
millsp, line 130: As mentioned before, types were improved to simplify the amount of definitions we maintain for the datasources. Initially the duplicate was more oriented to work directly work with `BinaryEngine`, but today we have three engine implementations, so it made sense to normalize the types. As a side-effect, `BinaryEngine` needs to do a bit of extra work to keep its previous data structures.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -32,9 +33,6 @@ import { Fetch, request } from './utils/request'
 
 const MAX_RETRIES = 3
 
-// to defer the execution of promises in the constructor
-const P = Promise.resolve()
```
millsp, line 36: Not needed anymore, we now defer some initialization to happen when we need to get the URL.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -162,32 +161,39 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
+    return 'unknown'
+  }
 
+  /**
+   * This is not a real "start" method, but rather a deferred initialization. We
+   * will only parse the URL on the first request to match the behavior of other
+   * engines, they will only throw errors on their very first request. This is
+   * needed in case the URL is misconfigured.
+   */
+  async start() {
```
millsp, line 181: The URL validation and extraction was moved into `start`, then `url` needs this to be able to calculate the Accelerate URL.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -239,7 +245,14 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
   }
 
   private async url(s: string) {
-    return `https://${this.host}/${await this.remoteClientVersion}/${this.inlineSchemaHash}/${s}`
+    if (this.startPromise !== undefined) {
+      await this.startPromise
+    } else {
+      this.startPromise = this.start()
+      await this.startPromise
+    }
```
millsp, line 253: Perhaps I can move this into `start`, looks out of place here.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -447,30 +460,23 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
-    const datasources = this.mergeOverriddenDatasources()
-    const mainDatasourceName = Object.keys(datasources)[0]
-    const mainDatasource = datasources[mainDatasourceName]
-    const dataProxyURL = this.resolveDatasourceURL(mainDatasourceName, mainDatasource)
+    const serviceURL = resolveDatasourceUrl({
+      inlineDatasources: this.inlineDatasources,
+      overrideDatasources: this.config.overrideDatasources,
+      clientVersion: this.clientVersion,
+      env: this.env,
+    })
```
millsp, line 468: Most of the previous logic is now in a re-usable helper `resolveDatasourceUrl`.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -447,30 +460,23 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
-    const { protocol, host, searchParams } = url
-
-    if (protocol !== 'prisma:') {
-      throw new InvalidDatasourceError(
-        'Datasource URL must use prisma:// protocol when --accelerate or --data-proxy are used',
-        {
-          clientVersion: this.clientVersion,
-        },
-      )
-    }
```
millsp, line 473: This is not needed anymore because the decision to load the Data Proxy engine depends on the contents of the URL, so we are sure at this stage it will at least contain `prisma://...`.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -482,62 +488,10 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
-      return loadedEnvURL
-    }
-
-    throw new InvalidDatasourceError(
-      `Datasource "${name}" specification is invalid: both value and fromEnvVar are null`,
-      {
-        clientVersion: this.clientVersion,
-      },
-    )
-  }
```
millsp, line 535: Now fully handled by `resolveDatasourceUrl`.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -482,62 +488,10 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
-
   metrics(options: MetricsOptionsJson): Promise<Metrics>
   metrics(options: MetricsOptionsPrometheus): Promise<string>
   metrics(): Promise<Metrics> | Promise<string> {
-    throw new NotImplementedYetError('Metric are not yet supported for Data Proxy', {
+    throw new NotImplementedYetError('Metrics are not yet supported for Data Proxy', {
```
millsp, line 494: Fix typo.

packages/client/src/runtime/core/engines/data-proxy/utils/getClientVersion.ts
```diff
@@ -31,9 +31,11 @@ async function _getClientVersion(host: string, config: EngineConfig) {
...
   // if it is an integration or dev version, we resolve its dataproxy
   // for this we infer the data proxy version from the engine version
-  if (suffix !== undefined || clientVersion === '0.0.0') {
-    // when we are running in tests, then we are using the mini proxy
-    if (process.env.TEST_DATA_PROXY !== undefined) return '0.0.0'
+  if (suffix !== undefined || clientVersion === '0.0.0' || clientVersion === 'in-memory') {
+    // we use the generated engine version to infer if we're in a test
+    if (config.engineVersion === '0000000000000000000000000000000000000000') {
+      return '0.0.0' // when we are running in tests, we use mini proxy
+    }
```
millsp, line 38: Env vars are not available for the edge client tests, which means it actually tried to resolve the real data proxy url, wasting time on CI (considerably). Now instead we just check if the client engine version is the one from the test setup.

packages/client/src/runtime/core/engines/data-proxy/utils/getClientVersion.ts
```diff
@@ -18,7 +18,7 @@ async function _getClientVersion(host: string, config: EngineConfig) {
   }
 
   // for data proxy v2, or accelerate, resolution isn't needed
-  if (host.includes('accelerate') && clientVersion !== '0.0.0') {
+  if (host.includes('accelerate') && clientVersion !== '0.0.0' && clientVersion !== 'in-memory') {
```
millsp, line 21: We also handle `in-memory` now so you can actually test studio locally. I think this should not be needed anymore once I merge https://github.com/prisma/studio-code/pull/514

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -110,10 +109,17 @@ Please help us by answering a few questions: https://pris.ly/bundler-investigati
...
     if (config.enableDebugLogs) {
       this.logLevel = 'debug'
     }
+
+    // compute the datasource override for library engine
+    const dsOverrideName = Object.keys(config.overrideDatasources)[0]
+    const dsOverrideUrl = config.overrideDatasources[dsOverrideName]?.url
+    if (dsOverrideName !== undefined && dsOverrideUrl !== undefined) {
+      this.datasourceOverrides = { [dsOverrideName]: dsOverrideUrl }
+    }
```
millsp, line 121: Refactored logic from below.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -236,17 +234,17 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
...
+        // being GCed. Using weak ref helps to avoid this cycle
         const weakThis = new WeakRef(this)
         this.engine = new this.QueryEngineConstructor(
           {
             datamodel: this.datamodel,
             env: process.env,
             logQueries: this.config.logQueries ?? false,
             ignoreEnvVarErrors: true,
-            datasourceOverrides: this.datasourceOverrides,
+            datasourceOverrides: this.datasourceOverrides ?? {},
```
millsp, line 247: The type was optional, but turns out the engine always needs an object here.

packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -0,0 +1,35 @@
+import { ClientEngineType, getClientEngineType } from '@prisma/internals'
+
+import { BinaryEngine, DataProxyEngine, EngineConfig, LibraryEngine } from '../engines'
+import { PrismaClientValidationError } from '../errors/PrismaClientValidationError'
+import { resolveDatasourceUrl } from './resolveDatasourceUrl'
+
+export function getEngineInstance(_engineConfig: EngineConfig) {
```
millsp, line 7: Perhaps this could have a doc comment.

packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -0,0 +1,35 @@
...
+      clientVersion: _engineConfig.clientVersion,
+    })
+  } catch {
+    // the error does not matter, but that means we don't have a valid url which
+    // means we can't use the DataProxyEngine and will default to LibraryEngine
+  }
+
+  const engineType = getClientEngineType(_engineConfig.generator!)
+
+  if (url?.startsWith('prisma://') || TARGET_ENGINE_TYPE === 'edge') {
```
millsp, line 24: Notice the main change from this PR here: The switch to the Data Proxy engine is either dynamic and can happen with either Library or Binary engine builds; or the Data Proxy engine is completely locked when targeting edge runtimes.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -377,20 +346,6 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
-            return source && source.url
-          })
-          .map(([name, { url }]: any) => ({
-            name,
-            url,
-          }))
-
-        // TODO: isn't it equivalent to just `inputDatasources` if the first argument is `[]`?
-        const datasources = mergeBy([], inputDatasources, (source: any) => source.name)
-
```
millsp, line 393: Removed in favor of directly sending the datasource overrides to the engines.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -428,27 +382,22 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
+          engineVersion: config.engineVersion,
           previewFeatures: this._previewFeatures,
           activeProvider: config.activeProvider,
-          inlineSchema: config.inlineSchema,
-          inlineDatasources: config.inlineDatasources,
-          inlineSchemaHash: config.inlineSchemaHash,
+          inlineSchema: config.inlineSchema ?? '',
+          overrideDatasources: options.datasources ?? {},
+          inlineDatasources: config.inlineDatasources ?? {},
+          inlineSchemaHash: config.inlineSchemaHash ?? '',
```
millsp, line 391: Node: These should not require any defaults, they are required and not nullable.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -428,27 +382,22 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
 
-        if (this._dataProxy) {
-          const runtime = NODE_CLIENT ? 'Node.js' : 'edge'
-          debug(`using Data Proxy with ${runtime} runtime`)
-        }
-
-        this._engine = this.getEngine()
-
-        this._fetcher = new RequestHandler(this, logEmitter) as any
+        this._engine = getEngineInstance(this._engineConfig)
```
millsp, line 399: Simplified engine creation.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -476,20 +425,6 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
-        return new LibraryEngine(this._engineConfig)
-      } else if (this._clientEngineType === ClientEngineType.Binary && TARGET_ENGINE_TYPE === 'binary') {
-        return new BinaryEngine(this._engineConfig)
-      }
-
-      throw new PrismaClientValidationError('Invalid client engine type, please use `library` or `binary`', {
-        clientVersion: this._clientVersion,
-      })
-    }
-
```
millsp, line 492: Simplified engine creation.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -524,23 +459,14 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
     }
-    /**
-     * @private
-     */
-    async _runDisconnect() {
-      await this._engine.stop()
-      delete this._connectionPromise
-      this._engine = this.getEngine()
-      delete this._disconnectionPromise
-    }
```
millsp, line 543: Unsused.

packages/client/src/runtime/utils/printDatasources.ts
millsp, line 1: These are the types which were duplicated just for the datasources I mentioned earlier, removing it as it only causes confusion.

packages/client/src/utils/getTestClient.ts
```diff
@@ -54,12 +54,14 @@ export async function getTestClient(schemaDir?: string, printWarnings?: boolean)
     generator,
     dirname: absSchemaDir,
     relativePath: path.relative(outputDir, absSchemaDir),
-    clientVersion: 'client-test-version',
-    engineVersion: 'engine-test-version',
+    clientVersion: '0.0.0',
+    engineVersion: '0000000000000000000000000000000000000000',
```
millsp, line 58: Opportunistic cleanup.

packages/client/src/utils/getTestClient.ts
```diff
@@ -54,12 +54,14 @@ export async function getTestClient(schemaDir?: string, printWarnings?: boolean)
...
     relativeEnvPaths,
     datasourceNames: config.datasources.map((d) => d.name),
     activeProvider,
-    dataProxy: Boolean(process.env.TEST_DATA_PROXY),
+    inlineDatasources: { db: { url: config.datasources[0].url } },
+    inlineSchema: '',
+    inlineSchemaHash: '',
```
millsp, line 64: New metadata needed to instantiate a client.

packages/client/tests/functional/17797-no-env-error-init/tests.ts
```diff
@@ -14,12 +14,5 @@ testMatrix.setupTestSuite(
   {
     skipDefaultClientInstance: true,
     skipDb: true,
-    skipDataProxy: {
-      runtimes: ['node', 'edge'],
-      reason: `
-        Fails with Data Proxy: error is an instance of InvalidDatasourceError
-        Datasource "db" references an environment variable "INVALID_DATABASE_URI" that is not set.
-      `,
-    },
```
millsp, line 23: This is not an issue anymore with the refactor of `start` on `DataProxyEngine`

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -129,7 +129,7 @@ function setupTestSuiteMatrix(
             // disconnect might also fail, so ignoring the error here
           })
           if (clientMeta.dataProxy) {
-            await stopMiniProxyQueryEngine(client)
+            await stopMiniProxyQueryEngine(client, globalThis['datasourceInfo'])
```
millsp, line 132: I did some changes to improve miniproxy startup times, detected because I had some failing tests.

packages/client/tests/functional/_utils/stopMiniProxyQueryEngine.ts
```diff
@@ -1,23 +1,21 @@
...
 
-  const host = engine.host
-  const clientVersion = await engine.remoteClientVersion
-  const schemaHash = engine.inlineSchemaHash
+  debug('stopping mini-proxy query engine at', url.host)
 
-  debug('stopping mini-proxy query engine at', host)
-
-  const response = await nodeFetch(`https://${host}/_mini-proxy/${clientVersion}/${schemaHash}/stop-engine`, {
+  const response = await nodeFetch(`https://${url.host}/_mini-proxy/0.0.0/${schemaHash}/stop-engine`, {
```
millsp, line 15: When I replaced the version with `0.0.0` I realized that some tests were actually using real versions. That shows our tests were potentially slow for the edge client tests because we would resolve the actual URLs.

In that investigation, I just re-wrote the logic in here a bit, but it is completely equivalent, and we would be able to detect of our tests are ever trying to use a resolve url (they would fail).

packages/client/tests/functional/runtime-import/tests.ts
```diff
@@ -5,7 +5,6 @@ import testMatrix from './_matrix'
 
 const libraryRuntime = 'runtime/library'
 const binaryRuntime = 'runtime/binary'
-const dataProxyRuntime = 'runtime/data-proxy'
```
millsp, line 8: This string cannot be found in the generated client anymore, so we don't search for it anymore.

packages/client/tests/functional/invalid-env/tests.ts
```diff
@@ -29,10 +29,5 @@ testMatrix.setupTestSuite(
   {
     skipDb: true,
     skipDefaultClientInstance: true, // So we can manually call connect for this test
-    skipDataProxy: {
-      runtimes: ['node', 'edge'],
-      reason:
-        'InvalidDatasourceError: Datasource URL must use prisma:// protocol when --accelerate or --data-proxy are used',
-    },
```
millsp, line 36: Fixed.

packages/client/tests/functional/issues/10229/tests.ts
```diff
@@ -33,9 +33,8 @@ testMatrix.setupTestSuite(
       `,
     },
     skipDataProxy: {
-      runtimes: ['node', 'edge'],
-      reason:
-        'InvalidDatasourceError: Datasource URL must use prisma:// protocol when --accelerate or --data-proxy are used',
+      runtimes: ['edge'],
+      reason: 'InvalidDatasourceError will be shown on edge runtime, we should fix this in Prisma 6',
```
millsp, line 37: Changing the error type would be considered a breaking change imo.

packages/client/tests/functional/missing-env/tests.ts
```diff
@@ -10,18 +12,34 @@ testMatrix.setupTestSuite(
...
+        },
+      })
+
+      try {
+        await prisma.$connect()
+      } catch (e) {
+        const message = stripAnsi(e.message as string)
+        expect(e).toBeInstanceOf(Prisma.PrismaClientInitializationError)
+        expect(message).toContain('error: Environment variable not found: DATABASE_URI.')
+      }
```
millsp, line 38: Just making sure the logic still works with empty overrides.

packages/client/tests/functional/missing-env/tests.ts
```diff
@@ -10,18 +12,34 @@ testMatrix.setupTestSuite(
...
   {
     skipDb: true,
-    skipDefaultClientInstance: true, // So we can manually call connect for this test
-    skipDataProxy: {
-      runtimes: ['node', 'edge'],
-      reason: `
-        Fails with Data Proxy: error is an instance of InvalidDatasourceError
-        instead of Prisma.PrismaClientInitializationError.
-      `,
-    },
```
millsp, line 25: Fixed.

packages/internals/src/cli/getGeneratorSuccessMessage.ts
```diff
@@ -23,8 +23,7 @@ function formatVersion(generator: Generator): string | undefined {
 
   if (generator.getProvider() === 'prisma-client-js') {
     const engineType = getClientEngineType(generator.config)
-    const engineMode = generator.options?.dataProxy ? 'dataproxy' : engineType
-    return version ? `${version} | ${engineMode}` : engineMode
+    return version ? `${version} | ${engineType}` : engineType
```
millsp, line 26: This will likely be changed/removed completely in a future PR.

packages/internals/src/get-generators/utils/check-feature-flags/checkFeatureFlags.ts
```diff
@@ -1,26 +1,12 @@
...
-function checkForbiddenFeaturesWithDataProxyFlag(config: ConfigMetaFormat, options: GetGeneratorOptions) {
-  options.dataProxy === true &&
-    config.generators.some((generatorConfig) => {
-      return generatorConfig.previewFeatures.some((feature) => {
-        if (feature.toLocaleLowerCase() === 'metrics'.toLocaleLowerCase()) {
-          throw new Error(forbiddenPreviewFeatureWithDataProxyFlagMessage('metrics'))
-        }
-      })
-    })
-}
```
millsp, line 23: This will happen at instantiation time, it will be introduced in a future PR.

millsp:
The ecosystem-tests run can be seen [here](https://github.com/prisma/ecosystem-tests/actions/runs/5862093491). As you can see, we have some tests failing which are related to:
1. File changes: We are creating more files by default
1. Engine not being detected due to the removal of `_clientEngineType`

So nothing is fundamentally broken, just some assertions need to be updated.

packages/client/src/generation/generateClient.ts
```diff
@@ -296,7 +284,8 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
     )
   }
 
-  if (transpile === true && dataProxy !== true) {
+  if (transpile === true) {
+    // TODO after removal of dataProxy, do if --no-engine
```
millsp, line 288: I am already preparing the PR for this.

Jolg42:
ci test all

Jolg42:
After `ci test all` It looks like it fails on a few `binary` runs 🤔

packages/client/tests/memory/_utils/generateMemoryTestClient.ts
```diff
@@ -35,6 +35,5 @@ export async function generateMemoryTestClient(testDir: MemoryTestDir) {
       edge: [__dirname.replace(/\\/g, '/'), '..', '..', '..', 'runtime', 'edge'].join('/'),
     },
     projectRoot: testDir.basePath,
-    dataProxy: !!process.env.TEST_DATA_PROXY,
```
Jolg42, line 38: It looks like we are only testing the default (library) generated client in memory tests at the moment anyway.
It could be interesting to test more I think?

Jolg42:
Change in public package, removed `data-proxy.js` & `data-proxy.d.ts` see

Before
Screenshot 2023-08-15 at 11 37 17

After
Screenshot 2023-08-15 at 11 39 34

packages/cli/src/Generate.ts
Jolg42: Removing it later without a warning before does not sound "great" 🤔
I can see us potentially keeping this in Prisma 6 to avoid a "useless" breaking change.

packages/cli/src/Generate.ts
```diff
@@ -136,11 +134,6 @@ ${bold('Examples')}
...
-          !!args['--data-proxy'] ||
-          !!args['--accelerate'] ||
-          !!process.env.PRISMA_GENERATE_DATAPROXY ||
-          !!process.env.PRISMA_GENERATE_ACCELERATE,
```
Jolg42, line 143: I think we could update the docs at https://www.prisma.io/docs/reference/api-reference/environment-variables-reference#prisma_generate_dataproxy

To say that this env var was removed in `5.2.0`

packages/cli/src/Studio.ts
Jolg42: Why changing it if "it's the same"?

packages/client/src/generation/TSClient/Datasources.ts
Jolg42: I remember them being a bit different 🤔

packages/client/src/generation/TSClient/TSClient.ts
Jolg42: What's the impact of this on `prisma generate` for big schemas? since we would do this all the time even when not actually needed? (user using Node.js only, without accelerate)

I'm thinking that big schemas that were slow to generate would get worse? 🤔

packages/client/src/generation/TSClient/TSClient.ts
Jolg42: Maybe we could also have this warning surfaced in the Prisma VS Code extension / `prisma validate` 🤔

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
```diff
@@ -1,4 +1,4 @@
-import type { InternalDatasource } from '../../runtime/utils/printDatasources'
```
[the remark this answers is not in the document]
Jolg42: Did you check that the types were different?
See 
```
export interface DataSource {
  name: string
  provider: ConnectorType
  activeProvider: ConnectorType
  url: EnvValue
  // below is different from InternalDatasource
  directUrl?: EnvValue
  schemas: string[] | []
}

export interface InternalDatasource {
  name: string
  provider: ConnectorType
  activeProvider: ConnectorType
  url: EnvValue
  // below is different from Datasource
  config: any
}
```

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -140,16 +138,17 @@ class DataProxyHeaderBuilder {
...
   private logEmitter: EventEmitter
   private env: { [k in string]?: string }
 
   private clientVersion: string
   private tracingHelper: TracingHelper
-  readonly remoteClientVersion: Promise<string>
-  readonly host: string
-  readonly headerBuilder: DataProxyHeaderBuilder
+  private remoteClientVersion!: string
+  private host!: string
```
Jolg42, line 149: Is the `!` needed here 🤔

packages/client/src/runtime/core/engines/library/LibraryEngine.test.ts
```diff
@@ -110,6 +116,12 @@ test('responds to initialization error with PrismaClientInitializationError', as
       tracingHelper: disabledTracingHelper,
       env: {},
       cwd: process.cwd(),
+      inlineSchema: '',
+      inlineSchemaHash: '',
```
Jolg42, line 120: Technically, given how it's implemented, at runtime `inlineSchema` and `inlineSchemaHash` would always be provided but not used?

packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    url = resolveDatasourceUrl({
+      inlineDatasources: _engineConfig.inlineDatasources,
+      overrideDatasources: _engineConfig.overrideDatasources,
+      env: { ..._engineConfig.env, ...process.env },
+      clientVersion: _engineConfig.clientVersion,
+    })
+  } catch {
+    // the error does not matter, but that means we don't have a valid url which
+    // means we can't use the DataProxyEngine and will default to LibraryEngine
```
Jolg42, line 19: I'm not sure I get it 🤔 
Why not error then?

packages/client/tests/functional/invalid-env/tests.ts
```diff
@@ -30,9 +30,8 @@ testMatrix.setupTestSuite(
     skipDb: true,
     skipDefaultClientInstance: true, // So we can manually call connect for this test
     skipDataProxy: {
-      runtimes: ['node', 'edge'],
-      reason:
-        'InvalidDatasourceError: Datasource URL must use prisma:// protocol when --accelerate or --data-proxy are used',
+      runtimes: ['edge'],
+      reason: 'InvalidDatasourceError will be shown on edge runtime, we should fix this in Prisma 6',
```
Jolg42, line 34: Create an issue so we don't forget?

packages/client/tests/functional/runtime-import/tests.ts
```diff
@@ -16,24 +15,24 @@ testMatrix.setupTestSuite((suiteConfig, suiteMeta, clientMeta) => {
...
       expect(generatedClientContents).not.toContain(edgeRuntime)
+      expect(generatedClientContents).not.toContain(libraryRuntime)
     } else if (getClientEngineType() === ClientEngineType.Library) {
       expect(generatedClientContents).toContain(libraryRuntime)
-      expect(generatedClientContents).not.toContain(binaryRuntime)
-      expect(generatedClientContents).not.toContain(dataProxyRuntime)
       expect(generatedClientContents).not.toContain(edgeRuntime)
+      expect(generatedClientContents).not.toContain(binaryRuntime)
     } else if (getClientEngineType() === ClientEngineType.Binary) {
+      expect(generatedClientContents).toContain(edgeRuntime)
```
Jolg42, line 33:
```suggestion
      expect(generatedClientContents).not.toContain(edgeRuntime)
```
Would make more sense here, like it's done with Library?

packages/client/src/runtime/core/init/getEngineInstance.ts
millsp: It it errors, it means we have no URL and we default to `LibraryEngine`, to use the Data Proxy you need to provide the URL.

millsp: Also a bit more context on this, we have tests that assert that errors are never thrown in the constructor but only can happen on the first query, so it would be incorrect to error on init. Maybe that answers the question. If this is ever changed, the test will let us know.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -524,23 +459,14 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
       }
     }
 
-    $connect() {
+    async $connect() {
```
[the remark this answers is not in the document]
millsp: Left over from testing something, will remove.

packages/cli/src/Generate.ts
millsp: Actually, I will restore it via `--no-engine`. I think it is fair to assume that if you have such an env var, you'd also want the engine not to be included otherwise it would be a breaking change.

packages/client/src/runtime/core/engines/library/LibraryEngine.test.ts
millsp: Yes, they are required because you can switch to the dataproxy at any time, but is only used by the `DataProxyEngine`.

packages/client/scripts/default-deno-edge.ts
```diff
@@ -1,8 +1,8 @@
 class PrismaClient {
   constructor() {
     throw new Error(
-      `@prisma/client/deno/edge did not initialize yet. Please run "prisma generate --data-proxy" and try to import it again.
-In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues`,
+      `@prisma/client/deno/edge did not initialize yet. Please run "prisma generate" and try to import it again.
+      In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues`,
```
aqrln, line 5: Indentation is not part of the string on the previous line

packages/client/tests/functional/runtime-import/tests.ts
millsp: Yes, I agree. It was alwso caught by the tests.

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/common.ts
```diff
@@ -9,15 +9,14 @@ if (isMacOrWindowsCI) {
...
   const clientDir = path.join(__dirname, 'test-clients', runtimeName)
   await fs.promises.mkdir(clientDir, { recursive: true })
   await fs.promises.copyFile(path.join(__dirname, 'schema.prisma'), path.join(clientDir, 'schema.prisma'))
 
   const generatorOpts = {
```
aqrln, line 17: I think we can now pass `engineType: ClientEngineType` as the argument of this function instead of `runtimeName` and get rid of this block

packages/cli/src/Studio.ts
millsp: This was changed recently to make `directUrl` work but is not actually necessary because Studio will present an error on its own and in the UI, this happens via the CLI. Additionally, the reason why I needed to change this is because the engine also validates the Data Proxy URLs and errors.

packages/client/src/generation/TSClient/TSClient.ts
millsp: Turns out there is no impact.

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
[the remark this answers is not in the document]
millsp: Yes I know they were different and caused conflicts in type-checking in the code base. The original type that we get is always of type `DataSource` but somehow along the way it was replaced with `InternalDatasource` without adding or using these extra properties, essentially casting `DataSource` to `InternalDatasource`. I trust `DataSource` is the correct one because that is the type given by the generator:
image

packages/client/src/generation/utils/buildInlineDatasources.ts
millsp: I have restored this.

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/common.ts
millsp: Done.

packages/client/scripts/default-deno-edge.ts
millsp: Nice catch, thanks

millsp: Nice catch, thanks.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -91,18 +76,23 @@ export class TSClient implements Generatable {
...
       relativePath: pathToPosix(path.relative(outputDir, path.dirname(schemaPath))),
       clientVersion: this.options.clientVersion,
       engineVersion: this.options.engineVersion,
       datasourceNames: datasources.map((d) => d.name),
       activeProvider: this.options.activeProvider,
-      dataProxy: this.options.dataProxy,
       postinstall: this.options.postinstall,
       ciName: ciInfo.name ?? undefined,
+      inlineDatasources: datasources.reduce((acc, ds) => {
+        return (acc[ds.name] = { url: ds.url }), acc
```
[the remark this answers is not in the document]
millsp: How would you re-write this ideally?

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/common.ts
millsp: Also applied everywhere else where I could.

Jolg42 (approved):
:shipit:

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/common.ts
aqrln: Then we can now remove `generatorOpts` and replace `...generatorOpts` below with `{ engineType: runtimeName }`.

aqrln: That was my thinking initially but I didn't word the comment very clearly, and github doesn't allow adding a code suggestion here 😄

millsp: Oh right, thanks for pointing this out. Now fixed!

aqrln (approved):

packages/client/src/generation/TSClient/TSClient.ts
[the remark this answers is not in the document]
aqrln: I don't think this matters much.
I'd personally write
```ts
datasources.reduce((acc, ds) => {
  acc[ds.name] = { url: ds.url }
  return acc
})
```
but
```ts
datasources.reduce((acc, ds) => (acc[ds.name] = { url: ds.url }), acc)
```
or the way it currently is in the PR is totally fine with me too.

PR #20650, opened 2023-08-11
test(client): Rewrite `tracing` test to not use snapshots

SevInf:
Since traces are delivered asynchronously, we can't reliably snapshot them - relevant traces might've not been delivered yet and snapshot might be flaky. To attempt to alleviate that we had 500ms delay in each tracing test which unnecessary increased test runtime in some cases (70 sequential tests, 0.5s each = 35 extra seconds) and still was not enough in others. Moving them to `toEqual` instead of snapshots allowed us to wrap assertions into `waitFor` and make them more stable, while also reducing the runtime.

Unfortinately, that did not fix main source of flakiness - itx test still sometimes misses tracing spans when run on data proxy. Debugging shows that the relevant spans are never returned from the engine side in this case - so it's a real bug, not a test problem.


Jolg42 (commented):
I think you could rename this PR to either 
`test(client):`
or 
`fix(client)`

Feat would be for something new I think

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -850,7 +850,7 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
           attributes: {
             method: params.action,
             model: params.model,
-            name: `${params.model}.${params.action}`,
```
Jolg42, line 853: So for example, `$queryRaw`?

packages/client/tests/functional/tracing/tests.ts
```diff
@@ -510,72 +334,46 @@ testMatrix.setupTestSuite(({ provider }, suiteMeta, clientMeta) => {
...
+              { name: 'prisma:engine:itx_query_builder', children: [findManyDbQuery(), engineSerialize()] },
+            ],
+          },
+        ],
+      })
     })
   })
 
-  describe('tracing on $raw methods', () => {
+  describeIf(provider !== 'mongodb')('tracing on $raw methods', () => {
```
Jolg42, line 367: It looks like we're missing a test for https://www.prisma.io/docs/concepts/components/prisma-client/raw-database-access#runcommandraw

Jolg42: Can probably be done separately

Jolg42 (approved):
So much better than before 😍 

Thank you 🙌🏼

packages/client/src/runtime/getPrismaClient.ts
SevInf: Oh,  right, completely forgot that I did this part. 
Yes, this is a fix for raw methods.

PR #20654, opened 2023-08-11
fix: eval removed in Migrate.ts

karanssj4:


janpio:
Why this change?

karanssj4:
@janpio `eval('require('../package.json')')` can be rewritten directly as a `require('../package.json')`

I'm not sure if having it like that serves any purpose

since you're using TS, we can use import instead to please linter and have consistency.

One advantage though is this change made it possible for me to use a bundler (esbuild) on my project, because the package.json file wasnt imported. This cuts back on file system dependencies while bundling

karanssj4:
@aqrln @Druue @janpio is this getting merged?

karanssj4:
> @aqrln @Druue @janpio is this getting merged?

any update on this? We're maintaining a pnpm patch to make this bundleable in absense of this

karanssj4:
any update?

millsp:
Hey @karanssj4, thanks for your contribution. My colleague @SevInf is working on a refactor https://github.com/prisma/prisma/pull/22009/files#diff-16b0c068bdcb1485f34de451427766f4f5a1e0ae41091d3c1b3b60af50b67805R64 of this part of the code. The eval will be removed to give place to a proper `require`, in turn that will not evade any bundler's static analysis. With regards to your PR, your current change would cause us to bundle local versions that equal `0.0.0`, which is why we were hiding things behind `eval`. Does this make sense to you, and would you be ok closing your PR in favor of ours?

PR #20656, opened 2023-08-11
feat(client): Enable cleartext plugin for mySQL docker images

### LINKED ISSUE #10304: Error querying the database: Unknown authentication plugin `mysql_clear_password'.
## Problem

Getting "Error querying the database: Unknown authentication plugin `mysql_clear_password'." when trying to connect to an AWS RDS proxy (I can connect fine in other clients with proxy and I can connect fine with prisma directly to the database).

## Suggested solution

I couldn't find where we make mysql connections in the project (didn't spend a lot of time looking), but we need add a way to switch auth when making a connection.

mysql2 has the ability to do this with the authPlugins or on the connection config itself as authSwitchHandler. See below which I got from here https://gist.github.com/btbytes/cf1a1e4a90582924833f4d0a15acc1e9

```
connectionConfig.authSwitchHandler = (data, cb) => {
  if (data.pluginName === 'mysql_clear_password') {
	// See https://dev.mysql.com/doc/internals/en/clear-text-authentication.html
	console.log("pluginName: "+data.pluginName);
	let password = token + '\0';
	let buffer = Buffer.from(password);
	cb(null, password);
  }
};
```

I don't think there is alternative solution using prisma

MohamedAkl1:
Closes #10304

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -65,18 +65,20 @@ services:
       - MYSQL_ROOT_PASSWORD=root
       - MYSQL_DATABASE=tests
       - MYSQL_USER=prisma
+      - LIBMYSQL_ENABLE_CLEARTEXT_PLUGIN=1
     ports:
       - '3306:3306'
 
   mysql_isolated:
     image: mysql:8.0
-    command: --default-authentication-plugin=mysql_native_password --lower_case_table_names=1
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
https://github.com/prisma/prisma/issues/10304#issuecomment-1675512552

jagadeesh70 (approved):

Jolg42:
Closing as it's not implementing the functionality

PR #20680, opened 2023-08-14
fix: switch Date.now() to Math.round(performance.now())

Jolg42:
I didn't change
packages/client/src/runtime/core/engines/data-proxy/utils/time.ts

>Date.now() may have been impacted by system and user clock adjustments, clock skew, etc. as it is relative to the Unix epoch (1970-01-01T00:00:00Z) and dependent on the system clock. The performance.now() method on the other hand is relative to the timeOrigin property which is a [monotonic clock](https://w3c.github.io/hr-time/#dfn-monotonic-clock): its current time never decreases and isn't subject to adjustments.

See docs: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
https://www.w3.org/TR/hr-time-2/

### DIFF
--- a/helpers/compile/build.ts
+++ b/helpers/compile/build.ts
@@ -173,7 +173,7 @@ const watch =
 
     // triggers quick rebuild on file change
     const fastRebuild = debounce(async () => {
-      const timeBefore = Date.now()
+      const timeBefore = Math.round(performance.now())
 
       // we handle possible rebuild exceptions
       const rebuildResult = await handle.async(() => {
@@ -184,7 +184,8 @@ const watch =
         console.error(rebuildResult.message)
       }
 
-      console.log(`${Date.now() - timeBefore}ms [${options.name ?? ''}]`)
+      const timeAfter = Math.round(performance.now())
+      console.log(`${timeAfter - timeBefore}ms [${options.name ?? ''}]`)
     }, 10)
 
     // triggers a full rebuild on added file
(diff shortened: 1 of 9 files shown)


jkomyno:
What would be the impact of this PR in non-Node.js runtimes, like Deno? Afaik, `performance.now()` is Node.js-only.

Jolg42:
`performance.now()` is not specific to Node.js, it's implemented in all browsers now too (see links to docs I added in the description)

Concerning Deno it looks like it's supported?
https://deno.land/api@v1.36.1?s=Performance

millsp:
Just checked and it looks like we polyfill it with `Date.now`. So this looks good to me, just give a glance to ecosystem tests just in case.

millsp (approved):

helpers/compile/build.ts
```diff
@@ -184,7 +184,8 @@ const watch =
         console.error(rebuildResult.message)
       }
 
-      console.log(`${Date.now() - timeBefore}ms [${options.name ?? ''}]`)
+      const timeAfter = Math.round(performance.now())
+      console.log(`${timeAfter - timeBefore}ms [${options.name ?? ''}]`)
```
aqrln, line 188: A bit late here, but I think rounding should happen on the final expression (i.e. `Math.round(timeAfter - timeBefore)`) and not on each term so that we don't accumulate rounding errors.

It's nothing critical but would be nice to do, I can open a PR.

Jolg42: Feel free to open a PR. My current thinking is that we don't need to be precise in all the places I changed.

PR #20681, opened 2023-08-14
test(db execute): increase timeout in CI for 1 test

Jolg42:
Example found by @millsp 
https://github.com/prisma/prisma/actions/runs/5711429656/job/15473061854#step:8:744

### DIFF
--- a/packages/migrate/src/__tests__/DbExecute.test.ts
+++ b/packages/migrate/src/__tests__/DbExecute.test.ts
@@ -400,6 +400,11 @@ COMMIT;`,
       ctx.fixture('schema-only-postgresql')
       expect.assertions(2)
 
+      // In CI, sometimes 5s is not enough
+      if (process.env.CI) {
+        jest.setTimeout(10_000)
+      }
+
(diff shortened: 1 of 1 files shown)


packages/migrate/src/__tests__/DbExecute.test.ts
```diff
@@ -400,6 +400,11 @@ COMMIT;`,
       ctx.fixture('schema-only-postgresql')
       expect.assertions(2)
 
+      // In CI, sometimes 5s is not enough
+      if (process.env.CI) {
+        jest.setTimeout(10_000)
```
aqrln, line 405: Is this test really more heavy than the other ones in this file or is it just the only one that timed out so far? Would changing the timeout for the whole test suite be more appropriate maybe?

aqrln: Ah, I see, maybe DNS resolution takes too long before it gives up or something like that. Sounds good then.

aqrln (approved):

Issue #19694, opened 2023-06-09
RustPanic on `prisma generate` when Unsupported field defined in a Composite type


conneryn:
Hi Prisma Team! The following command just crashed.

## Command

`generate`

## Versions
      
| Name        | Version            |
|-------------|--------------------|
| Platform    | debian-openssl-1.1.x| 
| Node        | v16.14.0           | 
| Prisma CLI  | 4.15.0             | 
| Engine      | 8fbc245156db7124f997f4cecdd8d1219e360944| 

## Error
```
RustPanic: RuntimeError: panicked at 'internal error: entered unreachable code: No unsupported field should reach that path', query-engine/schema/src/build/output_types/field.rs:41:40
```

## Other details

The schema is generated without issues on version 4.8.1 (I have not tried any other version in between yet).  

It appears to be related to including an "Unsupported" field in a composite type.  A minimal schema that reproduces the issue for me is:
```
datasource db {
  provider = "mongodb"
  url      = env("DATABASE_URL")
}

type NestedType {
  this_causes_error Unsupported("RegularExpression")
}

model sample_model {
  id              String          @id @default(auto()) @map("_id") @db.ObjectId
  some_field NestedType
}
```

This may be related to: https://github.com/prisma/prisma/issues/19646, however, the panic occurs in a different place, and I specifically have issues when the Unsupported is in a composite type.

Let me know if/how I can provide more information.

PR #20684, opened 2023-08-14
test(cli): Unsupported types break generate when nested in composite types

Druue:
tests #19694


packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -47,13 +57,13 @@ describe('--schema from project directory', () => {
...
-      const prisma = new PrismaClient()
-      \`\`\`
-    `)
+                                          ✔ Generated Prisma Client (0.0.0 | TEST_ENGINE_TYPE) to ./@prisma/client in XXXms
+                                          You can now start using Prisma Client in your code. Reference: https://pris.ly/d/client
+                                          \`\`\`
+                                          import { PrismaClient } from './@prisma/client'
+                                          const prisma = new PrismaClient()
+                                          \`\`\`
+                            `)
```
Jolg42, line 66: Can you undo these changes? Looks like it's still a problem with inline snapshots 😢

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -37,6 +37,16 @@ describe('using cli', () => {
 
     expect(data.stdout).toContain(`I am a minimal generator`)
   }, 75_000) // timeout
+
+  it('should work with nested ', async () => {
```
Jolg42, line 41: Another way to do it would be to make the test assume that it should work, and make it `it.fail` so we can just remove `fail` when it's fixed
```suggestion
  it.fail('should work with nested ', async () => {
```

Jolg42: See https://jestjs.io/docs/api#testfailingname-fn-timeout

Jolg42 (dismissed):

Jolg42:
The tests show `Received promise resolved instead of rejected`

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -603,6 +603,19 @@ describe('using cli', () => {
...
+    ctx.fixture('composite-panic')
+    const result = ctx.cli('generate')
+    // This points to: fn map_scalar_input_type()
+    await expect(result).rejects.toThrow(
+      expect.objectContaining({
+        message: expect.stringContaining(
+          `panicked at 'internal error: entered unreachable code: No unsupported field should reach that path'`,
+        ),
+      }),
+    )
```
Jolg42, line 617:
```suggestion
    await expect(result).resolves
```
or something like that

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -603,6 +603,19 @@ describe('using cli', () => {
 
     expect(data.stdout).toContain(`I am a minimal generator`)
   }, 75_000) // timeout
+
+  it("doesn't work with nested unsupported in composite type", async () => {
```
Jolg42, line 607:
```suggestion
  it("should work with nested unsupported in composite type", async () => {
```

PR #20692, opened 2023-08-15
feat(client, cli): `--no-engine`

### LINKED ISSUE #9597: CFW: Avoid including Query Engine when data proxy is enabled

millsp:
This PR builds on https://github.com/prisma/prisma/pull/20594. It introduces a new `prisma generate --no-engine` flag which prevents engines to be uselessly copied if they are not used. `--accelerate`, `--data-proxy` are preserved for backwards compatibility as well as their env vars. Before this PR, the engine copy was already skipped but were never removed from the generated folder if they existed already, this is now fixed.

As a side-product, the PR also improves our `e2e` tests so that they can run faster on CI and enabling us to create lockfiles for the individual tests to make them faster. I have not created all the missing lockfiles, but I had to change the paths in all the projects which already helped cutting the total run time by 30%. Because of this change, I suggest that you suppress the noise with this file [filter](https://github.com/prisma/prisma/pull/20692/files?file-filters%5B%5D=.md&file-filters%5B%5D=.prisma&file-filters%5B%5D=.sh&file-filters%5B%5D=.ts&show-deleted-files=true&show-viewed-files=true).

Take a look at the issue list below to get an overview of the changes.

closes https://github.com/prisma/prisma/issues/9597
closes https://github.com/prisma/team-orm/issues/293
closes https://github.com/prisma/team-orm/issues/295
closes https://github.com/prisma/team-orm/issues/294
closes https://github.com/prisma/team-orm/issues/296
closes https://github.com/prisma/team-orm/issues/297
closes https://github.com/prisma/team-orm/issues/306
closes https://github.com/prisma/team-orm/issues/307

/integration


millsp:
TODO also add a test to ensure that we can boot in dataproxy mode when it is an accelerate url and `--noEngines` is not passed.

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -474,18 +475,18 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
-      })
+      throw new InvalidDatasourceError('Could not parse URL of the datasource', errorInfo)
     }
 
-    const { host, searchParams } = url
+    const { protocol, host, searchParams } = url
+
+    if (protocol !== 'prisma:') {
+      throw new InvalidDatasourceError('Datasource URL must use prisma:// protocol', errorInfo)
+    }
```
millsp, line 485: I restored this as it is actually useful in some test cases I added.

packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -8,35 +9,36 @@ import { resolveDatasourceUrl } from './resolveDatasourceUrl'
...
     // means we can't use the DataProxyEngine and will default to LibraryEngine
   }
 
-  const engineType = getClientEngineType(_engineConfig.generator!)
+  const engineType = getClientEngineType(engineConfig.generator!)
 
-  if (url?.startsWith('prisma://') || TARGET_ENGINE_TYPE === 'edge') {
-    return new DataProxyEngine(_engineConfig)
+  if (url?.startsWith('prisma://') || Boolean(clientConfig.noEngine) || TARGET_ENGINE_TYPE === 'edge') {
+    return new DataProxyEngine(engineConfig)
```
millsp, line 34: If it is an Accelerate url, or `--no-engine`, or was built for `edge`.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -324,7 +330,8 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
           config.relativeEnvPaths.schemaEnvPath && path.resolve(config.dirname, config.relativeEnvPaths.schemaEnvPath),
       }
 
-      const loadedEnv = NODE_CLIENT && tryLoadEnvs(envPaths, { conflictCheck: 'none' })
+      const loadedEnv = // for node we load the env from files, for edge only via env injections
+        (NODE_CLIENT && tryLoadEnvs(envPaths, { conflictCheck: 'none' })) || config.injectableEdgeEnv?.()
```
millsp, line 334: A test case uncovered that env vars for edge client happened on `require`, which is very different from the other builds. So now `config.injectableEdgeEnv` became a function.

packages/client/tests/functional/engine-selection-logic/tests.ts
```diff
@@ -0,0 +1,45 @@
...
+    })
+
+    testIf(clientMeta.dataProxy /** = --no-engine */)('--no-engine prevents from using the other engines', async () => {
+      process.env[`DATABASE_URI_${suiteConfig.provider}`] = 'postgresql://postgres:password@localhost:5432/db'
+
+      const prisma = newPrismaClient()
+      const promise = prisma.$connect()
+
+      // proof that the correct engine is used
+      await expect(promise).rejects.toThrowErrorMatchingInlineSnapshot(`Datasource URL must use prisma:// protocol`)
```
millsp, line 27: The error shows that when using `--no-engine` (formerly, `--data-proxy`), you can only use the Data proxy.

packages/client/tests/functional/engine-selection-logic/tests.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    // test that we can pass a prisma:// url when the tests is not run as a dataproxy
+    testIf(!clientMeta.dataProxy)('prisma:// url works as expected even when --no-engine is not used', async () => {
+      process.env[`DATABASE_URI_${suiteConfig.provider}`] = 'prisma://localhost:5432/db'
+
+      const prisma = newPrismaClient()
+      const promise = prisma.$connect()
+
+      // proof that the correct engine is used
+      await expect(promise).rejects.toThrowErrorMatchingInlineSnapshot(`No valid API key found in the datasource URL`)
```
millsp, line 38: This shows that when `--no-engine` is not used, it is still possible to use `prisma://` urls and uses the Data Proxy engine.

packages/client/tests/functional/invalid-env-file/tests.ts
millsp, line 1: This is a previous test that I expanded. We only had `invalid-env` which I renamed to `invalid-env-file` as it tests an invalid env file actually. Then I added the same test but for an actual invalid env value, below in `invalid-env-value`.

This captures what currently happens when an invalid env is given to the PrismaClient with different engines and builds.

packages/client/tests/functional/invalid-env-file/tests.ts
```diff
@@ -0,0 +1,58 @@
...
+
+      if (clientMeta.dataProxy && clientMeta.runtime === 'edge') {
+        // this error is expected because the edge client is not able to read files
+        await expect(promise).rejects.toBeInstanceOf(Prisma.PrismaClientInitializationError)
+        await expect(promise).rejects.toThrowErrorMatchingInlineSnapshot(
+          `error: Environment variable not found: DATABASE_URI.`,
+        )
+      } else if (clientMeta.dataProxy && clientMeta.runtime === 'node') {
+        // TODO Prisma 6: should be a PrismaClientInitializationError, but the message is correct
+        // await expect(promise).rejects.toBeInstanceOf(Prisma.InvalidDatasourceError)
```
millsp, line 35: This is not exposed, so the error instance cannot be matched.

packages/client/tests/functional/invalid-env-value/tests.ts
millsp, line 1: Same test as `invalid-env-file`, but with the env value being set into the environment.

packages/client/tests/functional/runtime-import/tests.ts
```diff
@@ -33,6 +34,28 @@ testMatrix.setupTestSuite((suiteConfig, suiteMeta, clientMeta) => {
       expect(generatedClientContents).toContain(binaryRuntime)
       expect(generatedClientContents).not.toContain(edgeRuntime)
       expect(generatedClientContents).not.toContain(libraryRuntime)
+    } else {
+      throw new Error('Unhandled case')
```
millsp, line 38: For future-proofing.

packages/client/tests/functional/runtime-import/tests.ts
```diff
@@ -33,6 +34,28 @@ testMatrix.setupTestSuite((suiteConfig, suiteMeta, clientMeta) => {
...
+    }
+  })
+
+  test('imported files have the expected annotations', async () => {
```
millsp, line 42: We also want to make sure we don't include unnecessary annotations.

packages/client/tests/functional/invalid-env-value/tests.ts
millsp, line 1: Same test as `invalid-env-file`, but with the env value being set into the environment.

packages/cli/src/Generate.ts
```diff
@@ -50,6 +50,7 @@ ${bold('Options')}
       --schema   Custom path to your Prisma schema
        --watch   Watch the Prisma schema and rerun after a change
    --generator   Generator to use (may be provided multiple times)
+   --no-engine   Generate a client for use with Accelerate only
```
millsp, line 53: For now this is its only purpose, but we might use it for other purposes in the future.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -128,12 +131,7 @@ ${buildDebugInitialization(edge)}
 const PrismaClient = getPrismaClient(config)
 exports.PrismaClient = PrismaClient
 Object.assign(exports, Prisma)${deno ? '\nexport { exports as default, Prisma, PrismaClient }' : ''}
-${buildNFTAnnotations(
-  edge || false /** TODO after removal of dataProxy, do if --no-engine or edge */,
-  engineType,
-  platforms,
-  relativeOutdir,
-)}
+${buildNFTAnnotations(Boolean(edge || noEngine), engineType, platforms, relativeOutdir)}
```
millsp, line 134: Edge runtime and `--no-engine` don't produce file annotations.

packages/client/src/generation/generateClient.ts
```diff
@@ -284,8 +291,7 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
     )
   }
 
-  if (transpile === true) {
-    // TODO after removal of dataProxy, do if --no-engine
+  if (transpile === true && noEngine !== true) {
```
millsp, line 294: Allow the engine copy if `--no-engine` was not passed.

packages/client/src/generation/generateClient.ts
```diff
@@ -539,5 +545,18 @@ async function copyRuntimeFiles({ from, to, runtimeName, sourceMaps }: CopyRunti
...
+/**
+ * Attempts to delete the output directory.
+ * @param finalOutputDir
+ */
+async function deleteOutputDir(finalOutputDir: string) {
+  try {
+    // we want to make sure that if we delete, we delete the right directory
+    if (require(`${finalOutputDir}/package.json`).name === GENERATED_PACKAGE_NAME) {
+      await fs.rmdir(finalOutputDir, { recursive: true })
+    }
```
millsp, line 560: `--no-engine` will delete the output directory of the client. Do you want me to perhaps delete each individual file instead?

packages/client/src/generation/utils/buildInjectableEdgeEnv.ts
```diff
@@ -46,7 +46,7 @@ function declareInjectableEdgeEnv(datasources: DataSource[]) {
   const injectableEdgeEnvCode = injectableEdgeEnvJson.replace(/"/g, '')
 
   return `
-config.injectableEdgeEnv = ${injectableEdgeEnvCode}`
+config.injectableEdgeEnv = () => (${injectableEdgeEnvCode})`
```
millsp, line 49: Explained later on.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -378,8 +385,7 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
                 ? options.log === 'query'
                 : options.log.find((o) => (typeof o === 'string' ? o === 'query' : o.level === 'query')),
             ),
-          // we attempt to load env with fs -> attempt edge env -> default
-          env: loadedEnv?.parsed ?? config.injectableEdgeEnv?.parsed ?? {},
+          env: loadedEnv?.parsed ?? {},
```
millsp, line 388: Combined above.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -478,10 +484,10 @@ export function getPrismaClient(config: GetPrismaClientConfig) {
...
         throw e
       } finally {
-        // Debug module keeps a list of last 100 logs regardless of environment variables.
-        // This can cause a memory leak. It's especially bad in jest environment where keeping an
-        // error in this list will prevent jest sandbox from being GCed. Clearing logs on disconnect
-        // helps to avoid that
+        // Debug module keeps a list of last 100 logs regardless of environment
+        // variables. This can cause a memory leak. It's especially bad in jest
+        // environment where keeping an error in this list prevents jest sandbox
+        // from being GCed. Clearing logs on disconnect helps to avoid that
```
millsp, line 490: Stylistic change.

packages/client/tests/e2e/_utils/run.ts
```diff
@@ -109,12 +109,12 @@ async function main() {
...
-      await $`cd ${enginesPkgPath} && pnpm pack --pack-destination ${__dirname}/../`
-      await $`cd ${debugPkgPath} && pnpm pack --pack-destination ${__dirname}/../`
-      await $`cd ${generatorHelperPkgPath} && pnpm pack --pack-destination ${__dirname}/../`
-      await $`cd ${wpPluginPkgPath} && pnpm pack --pack-destination ${__dirname}/../`
+      await $`cd ${clientPkgPath} && pnpm pack --pack-destination /tmp/`
+      await $`cd ${cliPkgPath} && pnpm pack --pack-destination /tmp/`
+      await $`cd ${enginesPkgPath} && pnpm pack --pack-destination /tmp/`
+      await $`cd ${debugPkgPath} && pnpm pack --pack-destination /tmp/`
+      await $`cd ${generatorHelperPkgPath} && pnpm pack --pack-destination /tmp/`
+      await $`cd ${wpPluginPkgPath} && pnpm pack --pack-destination /tmp/`
```
millsp, line 117: As mentioned in the title of the PR, this simplifies and makes running e2e tests much faster because we don't spend time copying the packaged client/cli/debug/generator to each test anymore. It also enables to be able to run tests locally again, and speed things up even more by creating lock files (in a future PR).

packages/client/tests/e2e/noengine-file-deletion/_steps.ts
```diff
@@ -0,0 +1,49 @@
...
+import { $ } from 'zx'
+
+import { executeSteps } from '../_utils/executeSteps'
+
+void executeSteps({
+  setup: async () => {
+    await $`pnpm install`
+    await $`pnpm prisma db push`
+  },
+  test: async () => {
```
millsp, line 11: Here we ensure that the engine is properly deleted between runs.

packages/client/tests/functional/accelerate-without-no-engine-hint/tests.ts
```diff
@@ -0,0 +1,39 @@
...
+    testIf(clientMeta.dataProxy === false)('using accelerate without --no-engine produces a warning at runtime', () => {
+      process.env[`DATABASE_URI_${suiteConfig.provider}`] = 'prisma://accelerate.net/?api_key=doesNotMatter'
+      const consoleWarnMock = jest.spyOn(console, 'warn').mockImplementation()
+
+      newPrismaClient()
+
+      expect(consoleWarnMock).toHaveBeenCalledTimes(1)
+      expect(consoleWarnMock.mock.calls[0]).toMatchInlineSnapshot(`
+        [
+          In production, we recommend using \`prisma generate --no-engine\` (See: \`prisma generate --help\`),
```
millsp, line 19: Users will get this suggestion as a warning if they are not using `--no-engine`.

packages/client/tests/functional/invalid-env-value/tests.ts
```diff
@@ -0,0 +1,58 @@
+import stripAnsi from 'strip-ansi'
```
millsp, line 1: This is actually a rename of the test below with some added test cases for each type of engine/runtime.

packages/client/tests/functional/issues/10229/tests.ts
```diff
@@ -33,8 +33,8 @@ testMatrix.setupTestSuite(
       `,
     },
     skipDataProxy: {
-      runtimes: ['edge'],
-      reason: 'InvalidDatasourceError will be shown on edge runtime, we should fix this in Prisma 6',
+      runtimes: ['node', 'edge'],
+      reason: 'InvalidDatasourceError is not compatible with asserted error // Change in Prisma 6',
```
millsp, line 37: Previously removed, but now restored as the functional test suite is now testing with `--no-engine`. This kinda shows the bad discrepancy we have, but we can only get it fixed in Prisma 6 probably.

packages/client/src/generation/generateClient.ts
```diff
@@ -539,5 +545,18 @@ async function copyRuntimeFiles({ from, to, runtimeName, sourceMaps }: CopyRunti
...
+}
+
+/**
+ * Attempts to delete the output directory.
+ * @param finalOutputDir
+ */
+async function deleteOutputDir(finalOutputDir: string) {
+  try {
+    // we want to make sure that if we delete, we delete the right directory
+    if (require(`${finalOutputDir}/package.json`).name === GENERATED_PACKAGE_NAME) {
```
Jolg42, line 558: We only have a name in the package.json since [`4.2.0`](https://github.com/prisma/prisma/releases/tag/4.2.0) released 1 year ago
https://github.com/prisma/prisma/pull/14361

So what happens when someone upgrades from an older version?

packages/client/src/generation/generateClient.ts
```diff
@@ -539,5 +545,18 @@ async function copyRuntimeFiles({ from, to, runtimeName, sourceMaps }: CopyRunti
...
+      await fs.rmdir(finalOutputDir, { recursive: true })
+    }
+  } catch {}
```
Jolg42, line 561: If we silence this and that somehow the deletion doesn't succeed, it would look like `--no-engine` did nothing? 
It would be hard to debug a case like this?

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -474,18 +475,18 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
     try {
       url = new URL(serviceURL)
     } catch {
-      throw new InvalidDatasourceError('Could not parse URL of the datasource', {
-        clientVersion: this.clientVersion,
-      })
+      throw new InvalidDatasourceError('Could not parse URL of the datasource', errorInfo)
```
Jolg42, line 478: Maybe a good opportunity to improve this a bit? This would give a hint that the expected url should start with prisma protocol
```suggestion
      throw new InvalidDatasourceError('Could not parse URL of the datasource. It must be a valid URL that starts with the "prisma://" protocol', errorInfo)
```

packages/client/tests/e2e/_utils/run.ts
```diff
@@ -88,8 +88,8 @@ async function main() {
     const dependencies = pkgJsonWithRuntimeDeps.dependencies as Record<string, string>
 
     // replace references to unbundled local packages with built and packaged tarballs
-    if (dependencies['@prisma/engines']) dependencies['@prisma/engines'] = '/test/prisma-engines-0.0.0.tgz'
-    if (dependencies['@prisma/debug']) dependencies['@prisma/debug'] = '/test/prisma-debug-0.0.0.tgz'
+    if (dependencies['@prisma/engines']) dependencies['@prisma/engines'] = '/tmp/prisma-engines-0.0.0.tgz'
+    if (dependencies['@prisma/debug']) dependencies['@prisma/debug'] = '/tmp/prisma-debug-0.0.0.tgz'
```
Jolg42, line 92: It looks like the tests are passing because e2e tests are not tested on Windows.
How interesting it is to run these tests on Windows? I guess we could open a tech-debt issue to investigate.

packages/cli/src/Generate.ts
```diff
@@ -136,6 +138,12 @@ ${bold('Examples')}
         cliVersion: pkg.version,
         generatorNames: args['--generator'],
         postinstall: Boolean(args['--postinstall']),
+        noEngine:
+          Boolean(args['--no-engine']) ||
+          Boolean(args['--data-proxy']) || // legacy
+          Boolean(args['--accelerate']) || // legacy
+          Boolean(process.env.PRISMA_GENERATE_DATAPROXY) || // legacy
+          Boolean(process.env.PRISMA_GENERATE_ACCELERATE), // legacy
```
Jolg42, line 146:
```suggestion
          Boolean(args['--data-proxy']) || // legacy, we need to keep it to avoid an uneeded breaking change
          Boolean(args['--accelerate']) || // legacy, we need to keep it to avoid an uneeded breaking change
          Boolean(process.env.PRISMA_GENERATE_DATAPROXY) || // legacy, we need to keep it to avoid an uneeded breaking change
          Boolean(process.env.PRISMA_GENERATE_ACCELERATE), // legacy, we need to keep it to avoid an uneeded breaking change
```

packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -8,35 +9,40 @@ import { resolveDatasourceUrl } from './resolveDatasourceUrl'
...
+      clientVersion: engineConfig.clientVersion,
     })
   } catch {
     // the error does not matter, but that means we don't have a valid url which
     // means we can't use the DataProxyEngine and will default to LibraryEngine
   }
 
-  const engineType = getClientEngineType(_engineConfig.generator!)
+  if (clientConfig.noEngine !== true && url?.startsWith('prisma://')) {
+    console.warn('In production, we recommend using `prisma generate --no-engine` (See: `prisma generate --help`)')
```
Jolg42, line 32: We end up in a similar case than `--data-proxy` here? We would need an env var for some cases for this to work?

Jolg42 (commented):
Note: I wanted to try this out, so I triggered an integration release, I'll try to check it later, though I'm not sure i'll have time before the release.

Now out at `5.2.0-integration-feat-no-engine.1`

packages/client/src/generation/generateClient.ts
aqrln: I was also wondering about this, as it makes the behaviour a bit different, but I'm not sure about whether there are any real implications.

One thing that comes to my mind, can it cause any issues for `prisma generate --watch`?

packages/client/src/generation/generateClient.ts
```diff
@@ -250,10 +257,10 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
       const filePath = path.join(finalOutputDir, fileName)
       // The deletion of the file is necessary, so VSCode
       // picks up the changes.
-      if (await exists(filePath)) {
-        await fs.promises.unlink(filePath)
+      if (existsSync(filePath)) {
+        await fs.unlink(filePath)
```
aqrln, line 261: We could unlink the file without checking and ignore the error if it doesn't exist. This is technically more correct in general case because doing this in two calls introduces a race condition as another process can remove the file in between, but I'm not sure if this argument stands here given that there are probably other things that could fail if another generator were to modify the directory concurrently. So feel free to ignore this if you think it doesn't apply here.

On the other hand, it could also be a tiny bit more efficient if the file is more likely to exist than not to.

packages/client/tests/e2e/noengine-file-deletion/_steps.ts
```diff
@@ -0,0 +1,49 @@
...
+  const prismaPath = path.dirname(
+    require.resolve('.prisma/client', {
+      paths: [path.dirname(require.resolve('@prisma/client'))],
+    }),
+  )
+
+  let hasEngineFile = false
+  for await (const line of $`ls -l ${prismaPath}`.stdout) {
+    if (line.includes('libquery_engine-debian')) {
+      hasEngineFile = true
```
aqrln, line 44:
```suggestion
      hasEngineFile = true
      break
```

aqrln: also, would
```ts
const hasEngineFile = (await fs.promises.readdir(prismaPath))
  .includes('libquery_engine-debian.so.node')
```
perhaps be simpler?

millsp: Probably :D

packages/client/src/runtime/core/init/getEngineInstance.ts
millsp: Good point, I'll add it.

packages/client/src/generation/generateClient.ts
millsp: Then nothing happens, a bit of an edge case but also its better safety.

packages/client/src/generation/generateClient.ts
millsp: Are we testing watch? If not I can test it by hand, report, and open an issue to get it tested.

packages/client/tests/e2e/_utils/run.ts
aqrln: It doesn't unfortunately.

FWIW, it wouldn't be very useful to run these tests in Docker with the same base image, as that would be just Linux in a VM on Windows and not actually Windows. Windows containers do exist though (https://hub.docker.com/_/microsoft-windows-base-os-images), and if I understand correctly, don't require spawning a VM as long as the version of the container OS matches the host OS version. Maybe there's a way to use them without Docker (which will refuse to install without Hyper-V support)?

aqrln: We can make this run this without containers too, of course, especially if we want to test this across operating systems. Right now these tests assume Linux (even Debian-based Linux specifically), and Docker allows to run them on macOS or Windows or RHEL or Alpine but not actually test native support for those platforms.

aqrln: Containers are still useful for isolation, or if we want to be able to run tests for different targets with different base images on the same system, but not fundamentally required.

Jolg42:
Note: ecosystem-test run is here https://github.com/prisma/ecosystem-tests/actions/runs/5889637734 the errors are see should be unrelated 
@millsp posted about it in [Slack](https://prisma-company.slack.com/archives/CRGDU19K6/p1692233763884269)

Jolg42 (approved):
I didn't have time to try the [integration release](https://github.com/prisma/prisma/pull/20692#pullrequestreview-1582287857) yet, I suggest that someone does some QA before the release to be safe.

packages/client/src/generation/generateClient.ts
janpio: At least some debug log that surfaces his would be good.

packages/client/src/generation/generateClient.ts
millsp: I checked and currently `--no-engine` is not implemented for `--watch`.

packages/client/src/generation/generateClient.ts
millsp: Added internal debug logs

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
millsp: Made these on par with the regular engine errors.

packages/client/src/generation/generateClient.ts
millsp: It's a valid point, but since this PR does not attempt to touch this, I suggest we improve it via another PR.

aqrln (approved):

PR #20715, opened 2023-08-17
fix(client): missing require for data proxy client and custom output

### LINKED ISSUE #15395: `Cannot fetch data from service: include is not a function` Error while using Next.js with Data Proxy
#### As reported by the end user:

> We are using Prisma 4.2.1 and our database is publicly accessible. Prisma itself was working perfectly fine with the direct database url. It only started failing recently in production when trying to integrate Data Proxy. I have attached screenshots of our configuration. We are using **Nextjs with Prisma** and the attached files show a simple backend request fetching data using Data Proxy, which should be hitting our database. We created a simple lib file to have a single instance of Prisma that we can import into our backend calls. When running the app in a dev environment and running the build files locally, there are no issues with connecting to Data Proxy and fetching and inserting data into our database. Debugging on our end, we have identified that the error is P5010, but are having trouble finding info on how to resolve this issue.
>
> Since we are using Serverless js, we have tried using the prisma/client/edge, as opposed to prisma/client because our application gets deployed to **Cloudfront with Lambda@Edge**, but we have noticed no difference. We have a .env file that imports the variables and have confirmed that the backend call is using the correct Data Proxy url.
> 
> CloudwatchLogs
> 

Message Link for Internal Slack: https://prisma-company.slack.com/archives/C02GF9J74SH/p1654854367720369 

### How to reproduce

> #### Reproduction code:
>
> [Sample-master.zip](https://github.com/prisma/prisma/files/9599800/Sample-master.zip)
> 
> Instructions to run the code are in the `README.md` file in zip folder.
>
> #### AWS Configuration Images:
>
> 1. database-config-sample
> 2. database-config-sample-2
> 3. database-config-sample-3
> 4. database-config-sample-4
> 5. database-config-sample-5
>
> #### Error Logs Images:
>
> 1. error-logs
> 2. error-logs-2

### Expected behavior

Should not get `include is not a function` error.

### Prisma information

The Schema file is in the attached zip folder.

### Environment & setup

- Next.js with AWS Lambda

### Prisma Version

Prisma Version `4.2.1`

### LINKED ISSUE #18227: Custom Prisma Client output location breaks Prisma Data Proxy in NextJS
### Bug description

When using a custom Prisma Client output location with Prisma Data proxy, the below error is received.
See https://github.com/borisno2/prisma-with-data-proxy-custom example of code not working when importing prisma client from custom location (this example works when using a direct DB connection)
While this example - importing from `@prisma/client` works - https://github.com/borisno2/prisma-with-data-proxy-standard

### How to reproduce

1. Create Next App with app directory install prisma
2. Change output location of Prisma Client
3. Import Prisma Client from custom location
4. Run `prisma generate --data-proxy`
5. Run `next dev` using a prisma data proxy url (`prisma://...` as DATABASE_URL )
6. See error
```
error - Error: 
Invalid `prisma.thing.findMany()` invocation:


Cannot fetch data from service:
Nf is not a function
    at Xr.handleRequestError (./prisma/__generated__/prisma-client/runtime/data-proxy.js:10175:52)
    at Xr.handleAndLogRequestError (./prisma/__generated__/prisma-client/runtime/data-proxy.js:10152:18)
    at Xr.request (./prisma/__generated__/prisma-client/runtime/data-proxy.js:10142:18)
    at async t._request (./prisma/__generated__/prisma-client/runtime/data-proxy.js:10701:20)
    at async Home (./app/page.tsx:28:20)
digest: "98910894"
```

### Expected behavior

Custom output location should work the same as importing from `@prisma/client` and no error should be received when using Prisma Data Proxy

### Prisma information

```prisma
datasource postgresql {
    url               = env("DATABASE_URL")
    shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
}

generator client {
    provider = "prisma-client-js"
    output   = "__generated__/prisma-client"
}

model Thing {
    name String @default("")
}

```

```ts
import { PrismaClient } from '../prisma/__generated__/prisma-client'

const inter = Inter({ subsets: ['latin'] })

export default async function Home() {
  const prisma = new PrismaClient({ log: ['query', 'info', 'warn', 'error'] })
  const things = await prisma.thing.findMany()
...
```

### Environment & setup

- OS: MacOS
- Database: Postgres and Prisma Data Proxy
- Node.js version:  18.13.0

### Prisma Version

```
prisma                  : 4.11.0
@prisma/client          : 4.11.0
Query Engine (Node-API) : libquery-engine 8fde8fef4033376662cad983758335009d522acb (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Migration Engine        : migration-engine-cli 8fde8fef4033376662cad983758335009d522acb (at node_modules/@prisma/engines/migration-engine-darwin)
Format Wasm             : @prisma/prisma-fmt-wasm 4.11.0-57.8fde8fef4033376662cad983758335009d522acb
Default Engines Hash    : 8fde8fef4033376662cad983758335009d522acb
Studio                  : 0.483.0
```

millsp:
Next.js always marks `@prisma/client` as external so it is not actually bundled. However, when you output in a custom location, it is bundled just as regular code. This bundling causes Next.js to make different assumptions and erases possible calls to `require`. In the problematic code, we obfuscate `require` so it resolves to `undefined` at runtime. While we can't remove the problematic code until we drop support for Node 16, we can improve it by not loading the problematic code when we can.

Fortunately, we can skip the problematic code path completely whenever `fetch` is globally available, in the user's case that would be when they run on Node 18 and above. So that is the fix, we try to avoid using our obfuscated code as much as possible by using the real `fetch` when it's available. Then when we drop Node 16, we can get rid of this completely.

closes https://github.com/prisma/prisma/issues/18227
closes https://github.com/prisma/prisma/issues/15395


packages/client/src/runtime/core/engines/data-proxy/utils/request.ts
```diff
@@ -142,3 +138,45 @@ async function nodeFetch(url: string, options: RequestOptions = {}): Promise<Req
 
 // trick to obfuscate require from bundlers, useful for Vercel Edge
 const include = typeof require !== 'undefined' ? require : () => {}
+
+class NodeHeaders {
```
millsp, line 142: Polyfill written by the features team to make our node-fetch actually compliant with the real fetch when it comes to header management.

packages/client/src/runtime/core/engines/data-proxy/utils/request.ts
```diff
@@ -40,11 +38,16 @@ export async function request(
...
 
   try {
-    if (jsRuntimeName === 'browser-like') {
-      return await customFetch(fetch)(url, options)
+    if (typeof fetch === 'function') {
+      const response = await customFetch(fetch)(url, options)
+
+      // needed to be able to call json() multiple times
+      response.json = () => response.clone().json()
+      response.text = () => response.clone().text()
```
millsp, line 48: A "fetch" weirdness https://developer.mozilla.org/en-US/docs/Web/API/Response/clone

millsp: iiuc, you cannot call `text` then `json`, vice-versa, or twice or more in a row hence the need for clone. The performance impact isn't clear. Perhaps I can try to find a better alternative that does not involve cloning, I will look into it again tomorrow.  Ideas welcome :)

Jolg42: Why do we need to call `.json()` multiple times? (from the comment)

SevInf: +1 to the question. I'd rather fix us calling `json`/`text` multiple then introduce non-standard behaviour even to standard fetch.

packages/client/src/runtime/core/engines/data-proxy/utils/request.ts
```diff
@@ -142,3 +146,45 @@ async function nodeFetch(url: string, options: RequestOptions = {}): Promise<Req
 
 // trick to obfuscate require from bundlers, useful for Vercel Edge
 const include = typeof require !== 'undefined' ? require : () => {}
+
+export class NodeHeaders {
+  readonly #headers = new Map<string, string>()
```
SevInf, line 151: Back in the day there were problems with private fields: https://prisma-company.slack.com/archives/CUXLS0Z6K/p1675164034311419.
We have no explicit decision on whether or not it's ok to break `"target": "ES3"` people, but maybe we should now.

Jolg42: It looks like the minimum would be `ES2022` for this.
I think to be safe, we would need to
- Check the impact of this
- Check it fails and document what to do?
- See what's the default in the important cases, for example is it es2021 by default for [Cloudflare Workers](https://github.com/cloudflare/workers-sdk/blob/624846d574746d9eca575ef0d24f32e20605980c/packages/wrangler/templates/tsconfig.init.json#L14)?
- Mention this in the docs
- Check ecosystem-tests
- Check our examples
- Anything else?

packages/client/src/runtime/core/engines/data-proxy/utils/request.ts
rtbenfield: Calling `clone` can buffer the response into memory if the two streams are not read concurrently and at similar speeds. When one of them is not read, which would always be the case here for the original response, that one will buffer entirely into memory and (I think) stay until the GC can clean up the response object. I would also +1 locating the extra `json`/`text` calls and resolving it there if possible.

packages/client/src/runtime/core/engines/data-proxy/utils/request.ts
millsp: Let me remove the private property for the time being, this will go away soon anyways.

packages/client/src/runtime/core/engines/data-proxy/utils/request.ts
millsp: Done

Jolg42 (approved):

PR #20734, opened 2023-08-18
fix(client): improve edge env var errors

### LINKED ISSUE #20589: Better error message if `@prisma/client/edge` can not find environment variable
`@prisma/client/edge` is designed for edge environments where no files can be accessed or read, so it does not try to interact with the `.env` file at all. This can lead to unexpected situations where a project worked fine locally with `@prisma/client`, but then after switching to `@prisma/client/edge` and/or deploying, suddenly you get an error message like:

```
Internal server error: Datasource "db" references an environment variable "DATABASE_URL" that is not set
```

An example for such confusion can be https://github.com/prisma/prisma/issues/18542

We have good solutions and reasons for this, but just the default error message is not super helpful in understanding and fixing this.

Instead we could output a better error and error message that explains the situation a bit and offers solutions (either directly or via link to our docs).

millsp:
closes https://github.com/prisma/team-orm/issues/144
closes https://github.com/prisma/prisma/issues/20589

### DIFF
--- a/packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
+++ b/packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
@@ -29,6 +29,16 @@ export function resolveDatasourceUrl({
 
   // env var is set for use but url is undefined
   if (datasourceUrl?.fromEnvVar !== undefined && resolvedUrl === undefined) {
+    if (TARGET_ENGINE_TYPE === 'edge') {
+      throw new PrismaClientInitializationError(
+        `error: Environment variable not found: ${datasourceUrl.fromEnvVar}.
+Only \`process.env\` and \`globalThis\` can be read, but not \`.env\`.
+
+To solve this, provide it directly: https://pris.ly/d/datasource-url`,
+        clientVersion,
+      )
+    }
+
     // error matches as much as possible the usual engine error
     throw new PrismaClientInitializationError(
       `error: Environment variable not found: ${datasourceUrl.fromEnvVar}.`,
--- a/packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
+++ b/packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
@@ -32,9 +32,9 @@ export function resolveDatasourceUrl({
     if (TARGET_ENGINE_TYPE === 'edge') {
       throw new PrismaClientInitializationError(
         `error: Environment variable not found: ${datasourceUrl.fromEnvVar}.
-Only \`process.env\` and \`globalThis\` can be read, but not \`.env\`.
+In edge runtimes, only \`process.env\` & \`globalThis\` are read, not \`.env\`.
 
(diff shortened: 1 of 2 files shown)


packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
```diff
@@ -29,6 +29,16 @@ export function resolveDatasourceUrl({
 
   // env var is set for use but url is undefined
   if (datasourceUrl?.fromEnvVar !== undefined && resolvedUrl === undefined) {
+    if (TARGET_ENGINE_TYPE === 'edge') {
+      throw new PrismaClientInitializationError(
+        `error: Environment variable not found: ${datasourceUrl.fromEnvVar}.
+Only \`process.env\` and \`globalThis\` can be read, but not \`.env\`.
+
+To solve this, provide it directly: https://pris.ly/d/datasource-url`,
```
SevInf, line 37:
```suggestion
      throw new PrismaClientInitializationError(
        `error: Environment variable not found: ${datasourceUrl.fromEnvVar}.
When using Prisma edge client, environment variables are read only from \`process.env\` or \`globalThis\` globals, \`.env\` file is not accessible.

To solve this, provide it directly: https://pris.ly/d/datasource-url`,
```

Just a suggestion, feel free to ignore.

SevInf (approved):
LGTM, with one suggestion regarding the message

aqrln (approved):

packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
```diff
@@ -29,6 +29,16 @@ export function resolveDatasourceUrl({
 
   // env var is set for use but url is undefined
   if (datasourceUrl?.fromEnvVar !== undefined && resolvedUrl === undefined) {
+    if (TARGET_ENGINE_TYPE === 'edge') {
+      throw new PrismaClientInitializationError(
+        `error: Environment variable not found: ${datasourceUrl.fromEnvVar}.
+In edge runtimes, only \`process.env\` & \`globalThis\` are read, not \`.env\`.
```
janpio, line 35: Wouldn't it be more correct to say that Prisma Client in edge runtimes does not read and load the `.env` file?
So only values that are already in `process.env` and/or `globalThis` via other sources can be accessed?

millsp: How would you change this sentence?

packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
```diff
@@ -29,6 +29,16 @@ export function resolveDatasourceUrl({
...
+
+To solve this, provide the URL directly: https://pris.ly/d/datasource-url`,
```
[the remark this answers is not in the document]
millsp: It means nothing was found via `process.env` or `globalThis`, and we couldn't read `.env`, so you need to pass the URL directly.

PR #20753, opened 2023-08-18
fix(cli): incorrect url hardcoding warning

millsp:
After doing some QA in a real project, I realized the new url hardcoding warnings were incorrectly firing. I mistakenly believed that we already had tests with `env("...")`. Now added.

### DIFF
--- a/packages/cli/src/generate/getHardcodedUrlWarning.ts
+++ b/packages/cli/src/generate/getHardcodedUrlWarning.ts
@@ -1,9 +1,12 @@
 import { ConfigMetaFormat, link } from '@prisma/internals'
 
 export function getHardcodedUrlWarning(config: ConfigMetaFormat) {
+  const datasource = config.datasources?.[0]
+
   if (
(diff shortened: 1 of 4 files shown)


packages/cli/src/generate/getHardcodedUrlWarning.ts
```diff
@@ -1,9 +1,12 @@
...
 
 export function getHardcodedUrlWarning(config: ConfigMetaFormat) {
+  const datasource = config.datasources?.[0]
+
   if (
-    config.datasources?.[0].provider !== 'sqlite' &&
-    (config.datasources?.[0].url.fromEnvVar == null || config.datasources?.[0].directUrl?.fromEnvVar == null)
+    datasource !== undefined &&
+    datasource.provider !== 'sqlite' &&
+    (datasource.url.fromEnvVar === null || datasource.directUrl?.fromEnvVar === null)
```
millsp, line 9: The previous check was not a strict null check, that was the issue.

PR #20781, opened 2023-08-21
test(cli): studio via mini-proxy

millsp:


packages/client/src/runtime/core/engines/data-proxy/utils/getClientVersion.ts
```diff
@@ -32,23 +32,14 @@ async function _getClientVersion(host: string, config: EngineConfig) {
   // if it is an integration or dev version, we resolve its dataproxy
   // for this we infer the data proxy version from the engine version
   if (suffix !== undefined || clientVersion === '0.0.0' || clientVersion === 'in-memory') {
-    // we use the generated engine version to infer if we're in a test
-    if (config.engineVersion === '0000000000000000000000000000000000000000') {
+    // if the host is local, then it means we are using the mini-proxy
```
aqrln, line 35: should we also check for `[::1]` here?

packages/client/src/runtime/core/engines/data-proxy/utils/getClientVersion.ts
```diff
@@ -32,23 +32,14 @@ async function _getClientVersion(host: string, config: EngineConfig) {
...
+    if (host.startsWith('localhost') || host.startsWith('127.0.0.1')) {
```
aqrln, line 36: should we also check for `[::1]` here?

packages/client/src/runtime/core/engines/data-proxy/utils/request.ts
```diff
@@ -45,6 +45,7 @@ export async function request(
       return await customFetch(nodeFetch)(url, options)
     }
   } catch (e) {
+    console.log(url)
```
aqrln, line 48: stray `console.log`

janpio: Uh, that is a bad one @millsp

PR #20782, opened 2023-08-22
fix(client): error for missing env on cloudflare

millsp:


packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
```diff
@@ -29,12 +31,12 @@ export function resolveDatasourceUrl({
...
   if (datasourceUrl?.fromEnvVar !== undefined && resolvedUrl === undefined) {
-    if (TARGET_ENGINE_TYPE === 'edge') {
+    if (TARGET_ENGINE_TYPE === 'edge' && detectRuntime() === 'workerd') {
       throw new PrismaClientInitializationError(
         `error: Environment variable not found: ${datasourceUrl.fromEnvVar}.
-In edge runtimes, only \`process.env\` & \`globalThis\` are read, not \`.env\`.
 
-To solve this, provide the URL directly: https://pris.ly/d/datasource-url`,
+In Cloudflare Module Workers, the URL is in the Worker's \`env\` parameter of \`fetch\`.
+To solve this, provide the URL directly: https://pris.ly/d/cloudflare-datasource-url`,
```
janpio, line 39:
```suggestion
In Cloudflare module Workers, environment variables are available only in the Worker's \`env\` parameter of \`fetch\`.
To solve this, provide the connection string directly: https://pris.ly/d/cloudflare-datasource-url`,
```

packages/client/tests/functional/missing-env/tests.ts
```diff
@@ -43,18 +43,37 @@ testMatrix.setupTestSuite(
...
         } catch (e) {
           const message = stripAnsi(e.message as string)
           expect(e).toBeInstanceOf(Prisma.PrismaClientInitializationError)
           expect(message).toMatchInlineSnapshot(`
             error: Environment variable not found: DATABASE_URI.
-            In edge runtimes, only \`process.env\` & \`globalThis\` are read, not \`.env\`.
 
-            To solve this, provide the URL directly: https://pris.ly/d/datasource-url
+            In Cloudflare Module Workers, the URL is in the Worker's \`env\` parameter of \`fetch\`.
+            To solve this, provide the URL directly: https://pris.ly/d/cloudflare-datasource-url
```
janpio, line 72:
```suggestion
            In Cloudflare module Workers, environment variables are available only in the Worker's \`env\` parameter of \`fetch\`.
            To solve this, provide the connection string directly: https://pris.ly/d/cloudflare-datasource-url`,
```

janpio (approved):
Small nitpicks on msg.

packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
```diff
@@ -29,12 +31,12 @@ export function resolveDatasourceUrl({
 
   // env var is set for use but url is undefined
   if (datasourceUrl?.fromEnvVar !== undefined && resolvedUrl === undefined) {
-    if (TARGET_ENGINE_TYPE === 'edge') {
+    if (TARGET_ENGINE_TYPE === 'edge' && detectRuntime() === 'workerd') {
```
jkomyno, line 34: Minor nit, non-blocking: let's define a `const jsRuntime = detectRuntime()` block above, and use that here in the conditional.
Realistically, we're going to have custom logic in here for other runtimes as well, and this would reduce the future `git diff`s.

packages/client/src/runtime/core/init/resolveDatasourceUrl.ts
jkomyno: I'm a bit puzzled by the terminology "Cloudflare module Workers" here.
"Cloudflare Workers (ES Format)" sounds a bit more like it: https://developers.cloudflare.com/workers/learning/migrate-to-module-workers/.

packages/client/tests/functional/missing-env/tests.ts
jkomyno: See https://github.com/prisma/prisma/pull/20782/files#r1301269378.

PR #20847, opened 2023-08-25
chore: use dotenv-cli for db urls

millsp:
This PR will make it easier to contribute as `direnv` is not strictly require anymore and was replaced with `dotenv-cli`, although it can still be used. `direnv` was especially annoying for windows users, but also yet another hurdle to get started. We also now include a new `.db.env` along with the old `.envrc` directly into the repo (because why not?), and also took advantage to cleanup passing the env vars on CI which was repetitive and not centralized.


Jolg42 (approved):
👍🏼 I guess it should be fast enough 🤔 (the previous explicit approach and using `direnv` didn't have an impact at runtime)

PR #20848, opened 2023-08-25
refactor: Simplify DMMFHelper, stop (most) DMMF patching

SevInf:
This merges 3 DMMFHelper back into one - since P5 they are used only
within generator and we no longer need to split into base/full dmmf for
runtime.
Also, removes a couple of functions/properties that are not used by the
generator and were added for runtime GraphQL validation.
Most notably, removes `resolve*` methods and leaves original DMMF mostly
untouched. Which means that `field.type.type` property will always be a
string now and we can remove a bunch of condintion for handing resolved
values all over the genrator. For 2 places where we actually need
resolved types, two helper methods for on-demand non-destructive
resolution are added.


millsp:
(This [test](https://github.com/prisma/prisma/actions/runs/5977949572/job/16219062493?pr=20848#step:5:219) is flaky and does not complete as it should, please ignore).

Jolg42:
@SevInf Did you try to reproduce the CodSpeed report locally?

SevInf:
@Jolg42 yes, that's a real regression, I roughly know where to look at (that specific benchmark we added just for this case), will handle it within a week

SevInf:
Note: performance regression is now fixed, codspeed refuses to update the comment. If you check it's dashboard instead than for the latest commit it's tiny bit faster than `main`.

Jolg42 (approved):
Nice cleanup! 💚

PR #20855, opened 2023-08-25
ci: type check codebase

millsp:


tsconfig.json
```diff
@@ -1,5 +1,18 @@
...
+    "**/fixtures",
+    "**/node_modules",
+    "**/.prisma/client",
+    "./reproductions",
+    "./packages/cli/preinstall",
+    "./packages/cli/prisma-client",
+    "./packages/client/runtime",
+    "./packages/client/declaration",
+    "./packages/client/generator-build"
+  ]
```
millsp, line 17: Unrelated, this is in a bid to make eslint faster in vscode.

tsconfig.utils.typecheck.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.build.bundle.json",
+  "compilerOptions": {
+    "emitDeclarationOnly": false,
+    "noEmit": true
+  }
+}
```
millsp, line 7: This is used to type check the whole codebase at once.

packages/migrate/tsconfig.json
```diff
@@ -1,3 +1,4 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
+  "include": ["./**/*"]
```
millsp, line 3: This ensures that only the files at the root and the project are included, making it clear that we don't need to look below. I think this could also help with some eslint slowdowns but not 100% sure. This should be mostly harmless anyways.

packages/cli/src/Version.ts
```diff
@@ -70,17 +70,20 @@ export class Version implements Command {
...
-        .exhaustive()
+      return (
+        match(engineMetaInfo)
+          .with({ 'query-engine': P.select() }, (currEngineInfo) => {
+            return [
+              `Query Engine${cliQueryEngineBinaryType === BinaryType.QueryEngineLibrary ? ' (Node-API)' : ' (Binary)'}`,
+              currEngineInfo,
+            ]
+          })
+          // @ts-ignore TODO @jkomyno, as affects the type of rows
```
Jolg42, line 81: Did you want to ping Alberto here?

Jolg42 (approved):
Thank you 💚

PR #20875, opened 2023-08-28
fix(client): assignability issues in model extensions leading stack depth errors

millsp:
This PR prevents excessive depth comparison when types cannot be resolved. Because we were allowing `GetSelect` to work directly with potentially `unknown` extensions (= extension not passed), in turn this allowed -rather not prevented- the compiler from deeply evaluating possible extensions. While on its own this is totally fine and efficient, if you ask the compiler to compare two recursive (or deep) types against each other then it won't know where to stop because `unknown` is an infinitely deep value, depending on how you treat it. In our case, we didn't treat it at all, so the compiler could compare without limits, until it errors. Now instead, we are telling the compiler to treat `unknown` as a special case and default to the base types when it is the case, essentially preventing it from going infinitely deeper.

/integration

### DIFF
--- a/packages/client/tests/functional/extensions/query.ts
+++ b/packages/client/tests/functional/extensions/query.ts
@@ -1199,7 +1199,7 @@ testMatrix.setupTestSuite(
                 const data = await query(args)
 
                 expectTypeOf(operation).toEqualTypeOf<'aggregate'>()
-                expectTypeOf(args).toEqualTypeOf<PrismaNamespace.UserAggregateArgs>()
+                expectTypeOf(args).toMatchTypeOf<PrismaNamespace.UserAggregateArgs>()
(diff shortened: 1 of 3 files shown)


packages/client/tests/functional/extensions/query.ts
```diff
@@ -1199,7 +1199,7 @@ testMatrix.setupTestSuite(
                 const data = await query(args)
 
                 expectTypeOf(operation).toEqualTypeOf<'aggregate'>()
-                expectTypeOf(args).toEqualTypeOf<PrismaNamespace.UserAggregateArgs>()
+                expectTypeOf(args).toMatchTypeOf<PrismaNamespace.UserAggregateArgs>()
```
millsp, line 1202: Assignability change as mentioned in the PR.

millsp:
Conflict of interests wrt to types, abandoning PR

PR #20886, opened 2023-08-29
test(e2e): remove wrangler in unsupported-edge-error 

millsp:
This PR fixes recent timeouts/flakiness on e2e tests. I noticed the blocking test was always the same and pin pointed to the newly added wrangler test. I believe this is the fix, I'll run it multiple times to confirm.

### DIFF
--- a/packages/client/tests/e2e/unsupported-edge-error/_steps.ts
+++ b/packages/client/tests/e2e/unsupported-edge-error/_steps.ts
@@ -8,19 +8,20 @@ void executeSteps({
     await $`pnpm prisma generate`
   },
   test: async () => {
-    const wrangler = $`pnpm wrangler dev`.nothrow()
+    // TODO: re-enable wrangler when https://github.com/cloudflare/workers-sdk/issues/3631 is fixed
+    // const wrangler = $`pnpm wrangler dev`.nothrow()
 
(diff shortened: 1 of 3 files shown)


Jolg42 (approved):

SevInf:
Do we have any other way of checking if the deployment is done, other than waiting? I think all unconditional `sleep`'s within a test eventually became a problem sooner or later.

millsp:
PR is still in draft, sorry for the ping :) We previously check it this way

```ts
for await (const line of wrangler.stdout) {
  if (line.includes('http://127.0.0.1:8787')) {
    break
  }
}
```

That seemed like a rather good way, so I am still at the debugging stage. I just know this is the unhealthy test so far.

Will keep you updated.

millsp:
I think it's something on CF's end. I've spotted this SEGFAULT in our [logs](https://github.com/prisma/prisma/actions/runs/6014728770/job/16315110628?pr=20886#step:5:347), so I think that could be it. I will retry again once the issue has been resolved. https://github.com/cloudflare/workers-sdk/issues/3631. For now, I did a "best effort" re-write of the test, but it is not run via `wrangler` anymore.

SevInf (approved):

PR #20900, opened 2023-08-30
feat(cli): add version, and help, to commands and flags output

Druue:


Druue:
I didn't see a test that actually checks the output of this message, is that something that we want?

janpio (approved):

Jolg42 (approved):

Jolg42:
I think it's good to merge. Thanks!

PR #20927, opened 2023-09-01
chore: support local envrc

aqrln:


SevInf (approved):
Nice, I wished we had that on multiple occasions

PR #20942, opened 2023-09-04
feat: emit warning on pasting private db url directly.

### LINKED ISSUE #17640: Emit a warning when database URL is pasted directly into the schema
## Problem

It should generally be considered a bad practice to paste a private database URL into the schema directly. If the user commits that schema by accident somewhere, or deploys it somewhere, they will be exposed to a security risk.

## Suggested solution

I suggest that we emit a warning on `prisma generate` (maybe others too?) when we detect any URL that isn't pointing to `localhost` or `127.0.0.1` is pasted in the schema. This way, we can raise awareness around security and suggest users to either use `.env` or environment variables.

kunal00000:
### Expected behavior
- If database url is pasted directly in schema file and doesn't contain `localhost` or `127.0.0.1` then a warning will be showed and confirmation will be taken before next steps.
Screenshot 2023-09-04 at 4 57 54 PM

### Commands 
- migrate dev

### Tested using
- postgres database url

resolves #17640 
/claim #17640

### DIFF
--- a/packages/migrate/src/commands/MigrateDev.ts
+++ b/packages/migrate/src/commands/MigrateDev.ts
@@ -14,7 +14,7 @@ import {
   validate,
 } from '@prisma/internals'
 import fs from 'fs'
-import { bold, dim, green, red } from 'kleur/colors'
+import { bold, dim, green, red, yellow } from 'kleur/colors'
 import prompt from 'prompts'
 
 import { Migrate } from '../Migrate'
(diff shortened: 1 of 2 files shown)


kunal00000:
Hello @millsp, could you please confirm if my understanding of the problem is correct and if my proposed solution is on the right track? If so, are there any additional commands or steps we should consider for this?

petradonka:
Thanks @kunal00000 🙌 

@millsp, would you take a look at this PR and see how it works together with the similar checks we've recently also added? Thank you!

petradonka:
Thanks for the PR! Closing as per https://github.com/prisma/prisma/issues/17640#issuecomment-1706115682

Issue #7771, opened 2021-06-19
Output path of current Prisma in `-v` 


janpio:
## Problem

This is a situation that can happen:
```
--- yarn prisma -v ---
warning package.json: No license field
prisma               : 2.26.0-dev.7
@prisma/client       : 2.26.0-dev.7
Current platform     : debian-openssl-1.1.x
Query Engine         : query-engine c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules/@prisma/engines/query-engine-debian-openssl-1.1.x)
Migration Engine     : migration-engine-cli c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine : introspection-core c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary        : prisma-fmt c838e79f39885bc8e1611849b1eb28b5bb5bc922 (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash : c838e79f39885bc8e1611849b1eb28b5bb5bc922
Studio               : 0.402.0
--- ls node_modules/@prisma/engines/ ---
ls: cannot access 'node_modules/@prisma/engines/': No such file or directory
```

As you can see `yarn prisma -v` tells us the engines comes from `node_modules/@prisma/engines`, but when you try to list that directory, it is actually not present.

What is happening here is that `yarn prisma` is actually from a few folders _above_ our project. This is kinda visible when you look to not run `yarn -s` but just `yarn`:
```
codespace ➜ /workspaces/e2e-tests/platforms-serverless/firebase-functions (janpio-patch-5 ✗) $ yarn prisma -v
yarn run v1.22.10
warning package.json: No license field
$ /workspaces/e2e-tests/node_modules/.bin/prisma -v
prisma               : 2.25.0-dev.36
@prisma/client       : 2.25.0-dev.36
...
```

## Suggested solution

Output the directory of where the executed `prisma` is from, either similar to Yarn the absolute path to the binary in `node_modules`, or the folder where the `package.json` that is responsible for the installation of Prisma comes from.

## Alternatives

Rely on `yarn` itself to give that information. Will not work with `npx` unfortunately for example.

vishalbalani:
Is this issue still open? /attempt #7771

[...]

PR #20950, opened 2023-09-04
Resolved Output path of current Prisma

MAYANK-TRIPATH:
/claim #7771
/fixes #7771


reproductions/basic-sqlite/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Prisma development playground",
   "main": "index.ts",
   "scripts": {
+    "prisma": "node -e 'console.log(require.resolve(\"@prisma/cli\"))'",
```
rishi-raj-jain, line 7: This seems like a hacky fix, why change the import path?

reproductions/basic-sqlite/package.json
```diff
@@ -17,7 +18,7 @@
   },
   "devDependencies": {
     "@types/node": "17.0.45",
-    "prisma": "../../packages/cli",
+    "prisma": "^5.2.0",
```
aqrln, line 21: Please don't change this, it must use the local package.

reproductions/basic-sqlite/package.json
aqrln, line 7: This package is not part of Prisma itself, it's merely a template for a local playground used for development. Changing anything here has no effect for users.

reproductions/basic-sqlite/package.json
jharrell, line 21: We would like to keep the relative import as that tests against the most up to date code.

reproductions/basic-sqlite/package.json
jharrell: I agree.

jharrell (changes requested):

reproductions/basic-sqlite/yarn.lock
```diff
@@ -0,0 +1,139 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
aqrln, line 1: Stray `yarn.lock` file in a pnpm workspace. If you need to add a test that specifically uses yarn for any reason, you could add support for yarn packages in `packages/client/tests/e2e` and create a test there, but I don't believe that should be necessary. Reading the issue, a regular test in `packages/cli/src/__tests__/commands/Version.test.ts` should be sufficient.

aqrln (changes requested):
Thank you for opening the PR and trying to tackle the issue, however, as it currently stands, it doesn't actually change anything in Prisma and only touches unrelated files. Please let us know if anything needs further clarifications in the issue so we can help you succeed 🚀

MAYANK-TRIPATH:
@aqrln @jharrell @rishi-raj-jain Thank you so much you reviewing this PR and guiding me.
Sorry for misunderstanding i guess it was just that we are encountering issues with inconsistent Prisma versions when running `yarn prisma -v` and need to ensure consistent Prisma versions across different environments. 
So i though its just some scripts and dependency faults.
Closing this PR for now and will raise new one with right solution :)
Thank you.

PR #20952, opened 2023-09-04
ci: add bundle size workflow

millsp:
Track file size variations on PRs. 

⚠️ The failure is normal, as it tries to checkout main to compare the sizes, but `size-limit` is not installed in the monorepo yet, so we need to merge it first.

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -1,0 +1,27 @@
+name: Bundle size
+
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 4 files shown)


janpio (approved):

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          node-version: 18
+          pnpm-version: 8
+          skip-tsc: true
+          skip-build: true
+
+      - uses: andresz1/size-limit-action@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          skip_step: install
```
Jolg42, line 28: Maybe we can put that in our current linting job to combine

PR #20953, opened 2023-09-04
ci: do not report bundle size by gzip

millsp:
Instead, the reported bundle size should now be in KB.


janpio:
That made things a lot bigger 😆

PR #20954, opened 2023-09-04
feat(cli): show absolute path for output path of current Prisma

mnmt7:
closes #7771 
/claim #7771


packages/internals/src/engine-commands/getEnginesMetaInfo.ts
```diff
@@ -145,7 +145,16 @@ export async function resolveEngine(binaryName: BinaryType): Promise<EngineInfo>
   )()
 
   const engineInfo: EngineInfo = {
-    path: enginePathEither,
+    path: pipe(
+      enginePathEither,
+      E.fold(
+        (error) => E.left(error),
+        (path) => {
+          const absolutePath = `${process.cwd()}/${path}`
```
rishi-raj-jain, line 153: you might want to do join here instead of forwardSlash as it'd break on Windows.

mnmt7:
@rishi-raj-jain Thanks for pointing that out. I have now changed the code to use join instead of concatenating.

jharrell:
@mnmt7 please fill out the CLA so that I can pass this along to engineering.

see: https://github.com/prisma/prisma/pull/20954#issuecomment-1705744203

mnmt7:
> @mnmt7 please fill out the CLA so that I can pass this along to engineering.
> 
> see: [#20954 (comment)](https://github.com/prisma/prisma/pull/20954#issuecomment-1705744203)

@jharrell Done

jharrell:
@mnmt7 I believe this approach is on the right path, but there are no tests confirming or denying that. Please write a test that fails on main and succeeds with your implementation. 

Thanks!

jharrell (changes requested):
@mnmt7 I believe this approach is on the right path, but there are no tests confirming or denying that. Please write a test that fails on main and succeeds with your implementation. 

Thanks!

mnmt7:
@jharrell While creating the test, I realized that I made the wrong change. The `resolveEngine()` function gives the absolute path of the engine and it was the `getEnginesInfo()` function that converted it into the relative path. Closing this in favor of #21115. Here I have added the test as well.

PR #20969, opened 2023-09-05
chore(client): Simplify driver adapter types

SevInf:
Reuses adapter types for `@jkomyno/driver-adapter-utils` instead of duplicating them in the generator so that they'll always stay in sync.

Ensures `adapter` property is not added to the option unless preview featire is on.


jkomyno (approved):
Thanks!

PR #21076, opened 2023-09-13
fix(telemetry): fix Span instanciation

Riron:
This PR aims at fixing a compatibility problem with dd-trace (and possibly other opentelemetry providers), the library provided by Datadog to instrument nodejs apps.

dd-trace is now compatible with opentelemetry by default and makes collecting opentelemetry data in Datadog a breeze. However, it uses it's own Provider / Span... implementation which doesn't work with the current `prisma:engine:*` Span implementation.
As stated in this issue, [engines spans don't show up in the instrumentation](https://github.com/DataDog/dd-trace-js/issues/1244#issuecomment-1652793489). `prisma:client:*` spans do appear though.

It seems to be linked to the fact that the current implementation instantiate an `@opentelemetry/sdk-trace-base` Span, which is not picked up by Datadog.
Instead, I suggest to let the tracer create the Span as it's the only one that knows which type of Span it expects.

With this change, `prisma:egine:*` spans do pop in Datadog.

Sidenotes:
- I have absolutely 0 experience with opentelemetry. Make sure to review thoroughly, even if it's a small change :)
- with versions >5.0 of @prisma/instrumentation, the code completely breaks the dd-trace integration, as instanciating an opentelemetry sdk-trace-base Span calls code that doesn't work with a dd-trace tracer. Dd-trace tracer doesn't implement the `getSpanLimits ` method as they don't use it. So their `Span` constructor doesn't have this `parentTracer.getSpanLimits()` call. See the error below

```
TypeError: parentTracer.getSpanLimits is not a function
at new Span (/app/node_modules/@opentelemetry/sdk-trace-base/build/src/Span.js:59:41)
at Ki.createEngineSpan (/app/node_modules/@prisma/client/runtime/library.js:123:1600)
at pr.logger (/app/node_modules/@prisma/client/runtime/library.js:114:1093)
at /app/node_modules/@prisma/instrumentation/dist/ActiveTracingHelper.js:59:20
at ActiveTracingHelper.createEngineSpan (/app/node_modules/@prisma/instrumentation/dist/ActiveTracingHelper.js:44:27)
at Array.forEach (<anonymous>)
at /app/node_modules/@prisma/client/runtime/library.js:114:850
```

### DIFF
--- a/packages/instrumentation/src/ActiveTracingHelper.ts
+++ b/packages/instrumentation/src/ActiveTracingHelper.ts
@@ -8,7 +8,7 @@ import {
   trace,
   TraceFlags,
 } from '@opentelemetry/api'
-import { Span as SpanConstructor, Tracer } from '@opentelemetry/sdk-trace-base'
+import { Tracer } from '@opentelemetry/sdk-trace-base'
 import { EngineSpanEvent, ExtendedSpanOptions, SpanCallback, TracingHelper } from '@prisma/internals'
 
 // If true, will publish internal spans as well
@@ -45,12 +45,6 @@ export class ActiveTracingHelper implements TracingHelper {
     const tracer = trace.getTracer('prisma') as Tracer
 
     engineSpanEvent.spans.forEach((engineSpan) => {
-      const spanContext: SpanContext = {
-        traceId: engineSpan.trace_id,
-        spanId: engineSpan.span_id,
-        traceFlags: TraceFlags.SAMPLED,
-      }
-
       const links = engineSpan.links?.map((link) => {
         return {
           context: {
@@ -61,20 +55,22 @@ export class ActiveTracingHelper implements TracingHelper {
         }
       })
 
-      const span = new SpanConstructor(
-        tracer,
-        ROOT_CONTEXT,
-        engineSpan.name,
-        spanContext,
-        SpanKind.INTERNAL,
-        engineSpan.parent_span_id,
-        links,
-        engineSpan.start_time,
-      )
-
-      if (engineSpan.attributes) {
-        span.setAttributes(engineSpan.attributes)
+      const parentSpanContext: SpanContext = {
+        traceId: engineSpan.trace_id,
+        spanId: engineSpan.parent_span_id,
+        traceFlags: TraceFlags.SAMPLED,
       }
+      const parentContext = trace.setSpanContext(ROOT_CONTEXT, parentSpanContext)
+      const span = tracer.startSpan(
+        engineSpan.name,
+        {
+          links,
+          kind: SpanKind.INTERNAL,
+          startTime: engineSpan.start_time,
+          attributes: engineSpan.attributes,
+        },
+        parentContext,
+      )
 
       span.end(engineSpan.end_time)
     })
--- a/packages/instrumentation/src/ActiveTracingHelper.ts
+++ b/packages/instrumentation/src/ActiveTracingHelper.ts
@@ -9,7 +9,7 @@ import {
   TraceFlags,
 } from '@opentelemetry/api'
 import { Tracer } from '@opentelemetry/sdk-trace-base'
-import { EngineSpanEvent, ExtendedSpanOptions, SpanCallback, TracingHelper } from '@prisma/internals'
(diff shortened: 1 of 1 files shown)


packages/instrumentation/src/ActiveTracingHelper.ts
janpio, line 1: Lines 4 and 6, import of `ROOT_CONTEXT` and `SpanContext` can be removed now.

Riron:
I think the current implementation in this PR doesn't propagate correctly the `engineSpan.parent_span_id` as it takes the `parentSpanId` from the active context in `tracer.startSpan()`.
I will look into that.

Riron:
So the `parent_span_id` is fixed. But there is one last (big) difference with the previous code: previously the spanId was forced with `spanId = engineSpan.spanId`. Which is not doable with `startSpan()`.
It's probably the reason the tests fail => The child - parent relationship must be broken at some point

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -61,21 +55,23 @@ export class ActiveTracingHelper implements TracingHelper {
...
-      const span = new SpanConstructor(
-        tracer,
-        ROOT_CONTEXT,
+      const parentSpanContext: SpanContext = {
+        traceId: engineSpan.trace_id,
+        spanId: engineSpan.parent_span_id,
+        traceFlags: TraceFlags.SAMPLED,
+      }
+      const parentContext = trace.setSpanContext(ROOT_CONTEXT, parentSpanContext)
+      const span = tracer.startSpan(
```
Riron, line 64: There is still one major problem with this code: the `engineSpan.spanId` is lost as we don't control the `spanId` when instanciating spans through the tracer. This is a problem as because of this the relationship between engine spans is lost.: we do keep the `parent_span_id` propagation, but this propagation is useless if the `spanId` themselves are not propagated.

I don't see an easy fix for that.
We could accumulate the engine spans, and only create them once we have a full tree of spans. That would enable us to create them from the root and remap the `parent_span_id` to the newly created ids. But I'm not sure if that would be an acceptable solution.

SevInf: That problem is, unfortunately, the main reason why we use `Span` constructor rather than `tracer`. 
The only way I could see we can solve it right now that will work with all kind of engines: ask `opentelemetry-js` people to allow to set explicit `spanId` via `tracer` or send them a pull request that would implement that.

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -42,42 +44,59 @@ export class ActiveTracingHelper implements TracingHelper {
...
   createEngineSpan(engineSpanEvent: EngineSpanEvent) {
+    const rootEngineSpans = ['prisma:engine', 'prisma:engine:itx_runner']
+
+    for (const span of engineSpanEvent.spans) {
+      // We receive engine spans as soon as they end. So the parent nodes are received after the children.
+      // Because of that we cannot process the spans immediately, as creating spans affects them a random SpanId and that
+      // breaks the `parent_span_id` relationship.
+      // To circumvent this issue we accumulate engine spans and only process them once a full span tree is acquired.
+      const currentValue = this.engineSpansToProcess.get(span.parent_span_id) ?? []
+      this.engineSpansToProcess.set(span.parent_span_id, [...currentValue, span])
```
Riron, line 55: To circumvent the SpanId problem, one way would be to accumulate traces and process them once we know we have a full "tree". 
Some notes:
- are spans always nested in either prisma:engine or prisma:engine:itx_runner ?
- spans should arrive in order, but who knows
- if a root span never arrives for some reason it could mean a small memory leak. A regular cleanup of old & unused spans might be needed

Anyway, I know it might not be an acceptable solution for you guys, but I just wanted to propose an alternative that could work on paper - or at least pass the tests :)

Riron: Any thoughts on whether or not this would be an acceptable solution @SevInf ? If it's not I will close the PR.

Riron:
Closing as this PR is getting old and the proposed solution is not optimal.
I reckon a better way to solve this problem would be to send Rust spans not only when they end, but when they start as well. So that the spans could be properly created in TS land, keeping the parent / child relationship.

PR #21119, opened 2023-09-18
fix: Ensure only selected types are exported from runime module

SevInf:
In #20161, in order to solve type declaration/OOM/performance issues,
we added forced exports for every type within the runime.

However, our type bundler renames certain types if the naming conflicts.
For example, if there are 2 `Result` types, one of them will be renamed
to `Result_2`.

Making renamed type public gives `tsc` green light for dependenig on it
when creating declaration for third-party library. This is wrong,
because rename is not guaranteed to be stable. In fact, we have the
problem with `accelerate` extension on a different branch: latest
Accelerate depends on `Result_2`, but in a branch corresponding type got
renamed to `Result_3`.

Internal discussion: https://prisma-company.slack.com/archives/C04TW4A2H8C/p1694778678856899

In this PR we revert to manual re-exports and ensure every type from
`core/types` gets re-exported without rename. Tests, fixed in #20161 are still green,
but it also ensures that only types we want will leak to third party
project.

Couple of public types with conflicting names got renamed.

/integration


SevInf:
Note: accelerate extension would need to be rebuilt to work correctly. We could add `type Result_2 = Result`  alias to circumvent this

SevInf:
Superseded by #21179

PR #21121, opened 2023-09-18
fix(client): Don't mention colors in the error messages if they are not on

### LINKED ISSUE #17706: Query validation messages reference color output even if colors are not enabled
image

Error mentions that "available options are listed in green", even though default output is colorless. With `pretty` [errorFormat](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#errorformat) the message is displayed in color and makes a little bit more sense:

image

Our options are:
- change error message to "available options are marked with ?" or something similar: works regardless of error format, better for colorblind users.
- change error message depending on error format, talk about color only when it is actually present. 
- make `pretty` default format so message would at least make more sense to the most users. Thing to mind: It is an `error.message` property, not some explicit terminal output and ANSI escape sequences for colors will be visible if it is rendered in anything other than terminal. We need at least check most popular frameworks 503 development pages to see how the message looks there.

jiashengguo:
change the error message depending on the error format, and talk about color only when it is actually present.

EDITED:

Resolves #17706


jiashengguo:
@quest-bot loot #{17706}

jiashengguo:
@quest-bot loot #17706

packages/client/src/runtime/core/errorRendering/applyValidationError.ts
```diff
@@ -469,7 +469,7 @@ function splitPath(path: string[]): [parentPath: string[], fieldName: string] {
 }
 
 function availableOptionsMessage({ green }: Colors) {
-  return `Available options are listed in ${green('green')}.`
+  return `Available options are ` + (green === noop ? `marked with ?` : `listed in ${green('green')}`) + '.'
```
SevInf, line 472: Hi @jiashengguo.
Generally, PR looks good. There is one thing I'd like to change: could we adjust the way we check for presence of colors? Green color function being equal to `noop` if colors are disabled is implementation detail that I would like to keep encapsulated.

What I'd suggest instead: add `readonly enabled: boolean` property to `Colors` interface and set it to corresponding values for both `activeColors` and `inactiveColors` implementations. Inside of `availableOptionsMessage`, check `enabled` property instead.

Thank you!

SevInf (changes requested):

packages/client/src/runtime/core/errorRendering/applyValidationError.ts
jiashengguo: sounds good! changed accordingly.

SevInf (approved):
Thanks a lot, looks good now!

PR #21124, opened 2023-09-18
fix(internals): semverLt compares Numbers instead of Strings

### LINKED ISSUE #18751: Semver checks for yarn and typescript could potentially fail
### Bug description

As pointed out in this [review comment](https://github.com/prisma/prisma/pull/18646#discussion_r1166402270), we are comparing semver numbers from strings and not as numbers, which could cause unexpected results.

The goals is to make these numbers instead of string and add a unit test. Also for `yarn`, I propose to check if `engines` can be a reliable replacement in the `package.json` instead of a runtime check. Similarly, `typescript` could be an optional peer dependency instead.

### How to reproduce

Write a unit test with a semver string that will cause a wrong comparison.

### Expected behavior

Semver parts should be compared as numbers, not strings.

### Prisma information

n/a

### Environment & setup

- n/a

### Prisma Version

```
dev
```

icanipa:
fix(internals): semverLt compares Numbers instead of Strings

fix semverLt function to compare versions with numbers instead of strings.

before semverLt compare string versions that could fall into error because '5' > '10' is true in 'Strings' but not in Numbers

@quest-bot loot #18751 

Resolves #18751


packages/internals/src/__tests__/getGenerators/generatorResolvers/prisma-client-js.test.ts
millsp, line 1: Hey @icanipa, thank you for this pull request! I think you could move this file directly next to `get-generators/generatorResolvers/prisma-client-js/check-dependencies/semverLt.ts`, and call it `semverLt.test.ts`. With that, you can then also remove the `describes` with this dedicated unit test. What do you think?

icanipa: sure, I'll fix that

packages/internals/src/__tests__/getGenerators/generatorResolvers/prisma-client-js.test.ts
```diff
@@ -0,0 +1,12 @@
...
+
+describe('compare semverLt as numbers', () => {
+  test('compare semverLt version is working correct', () => {
+    // compare semverLt as Number not as Strings. '5' > '10' is true in Strings
+    const minVersion = '1.1.10'
+    const testVersion = '1.1.5'
+    // semverLt returns true if currentVersion is lower than the minVersion
+    const results = semverLt(testVersion, minVersion)
+    expect(results).toBe(true)
+  })
```
millsp, line 11: I think we could test a little more in general to check this function works as expected (as this is also its first test). Not a requirement though, I'll leave it up to you :)

packages/internals/src/get-generators/generatorResolvers/prisma-client-js/check-dependencies/semverLt.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    const testVersion = '1.1.5'
+    // semverLt returns true if currentVersion is lower than the minVersion
+    const results = semverLt(testVersion, minVersion)
+    expect(results).toBe(true)
+  })
+  test('compare semverLt with higer versions', () => {
+    const minVersion = '1.19.2'
+    const testVersion = '1.19.3'
+    const results = semverLt(testVersion, minVersion)
+    expect(results).toBe(false)
```
millsp, line 16:
```suggestion

  test('compare semverLt with higher versions', () => {
    const minVersion = '1.19.2'
    const testVersion = '1.19.3'
    const results = semverLt(testVersion, minVersion)
    expect(results).toBe(false)
```

millsp: Just corrected a typo and added a line break

millsp (approved):

millsp:
Thanks @icanipa for your pull request :tada:

PR #21138, opened 2023-09-19
fix(client): extension types with mocking library

### LINKED ISSUE #21136: Prisma Client extension with method override and `jest-mock-extended` or `vitest-mock-extended` expects `never` input
### Bug description

When using Prisma Client for instance with Accelerate or any other extension that overrides the original client methods, and mocking this extended client with either `jest-mock-extended` or `vitest-mock-extended` will in turn make that `mockResolvedValue` expects an input of type `never`.

[Slack thread](https://prisma-company.slack.com/archives/C05SCP1DG8Z/p1695071134806409?thread_ts=1694726613.081089&cid=C05SCP1DG8Z)

### How to reproduce

```ts
import { mockDeep } from "vitest-mock-extended";
import type { PrismaClient } from "@prisma/client";
import { withAccelerate } from "@prisma/extension-accelerate"

const prismaMock = mockDeep<PrismaClient>().$extends(withAccelerate());

prismaMock.booking.count.mockResolvedValue(0) // see red squiggly line under `0`
```

### Expected behavior

The expected input should not be `never`, it should be the type of the result (in this case, a `number` for `.count`).

However, due to limitations, we can not properly infer `this` in the extension ahead of time and within the mocking library, which is necessary for the mocking library to preserve type safety (see slack thread). Instead, when `this` cannot be inferred, we could default to `any` instead of never.

### Prisma information

n/a

### Environment & setup

n/a

### Prisma Version

```
5.3.1
```

### Explanation

When you look at this line: `prismaMock.booking.count.mockResolvedValue(0);`, the value 0 is underlined with a squiggly, and TypeScript complains that the value should not be a number, but never.

So first, let's unpack how this vistest mocking library works. First of all, Calcom enables the mocking. What this enables, is that you can call method utilities for mocking on methods to be mocked. See count, originally it's a simple method, but after mocking it has methods available on the method itself.

To achieve this, the mocking library add the new properties at runtime, and on the other hand it also does so on the type level. It does so deeply and on any method sitting on the mocked object, in this case the prisma client. To achieve this on the type level, it uses recursive mapped types to make the added methods available on top of each prisma client operation. That's where things go wrong.

This library (and probably other mocking libraries) override, or re-create if you will, the mocked prisma client operation. In this process, it uses inference on the ...args and the return types of the operation. This is what causes the loss of types.

Going a bit deeper into why: Accelerate is a dynamic/generic extension, and as you may know, this dynamicity is made possible by Prisma.Args and Prisma.Result which respectively get the arguments and result types available on any client, dynamically via the underlying this (in this example, this would be the booking model). When mockResolvedValue is called, its internal infer forces the resolution of arg and result types while the underlying model is not yet known.

In theory I believe that this could have been known by the compiler, actually. However, this would make no sense for generic paramaters targeting this to behave differently than other generic parameters. So it is consistent, and this must be a reasonable design limitation in the compiler, totally understandable. So the conclusion is that in turn this forces Prisma.Args and Prisma.Result to return never. This only affects methods that are overrides or make use of generic parameters bound to this, Accelerate does both.

millsp:
closes https://github.com/prisma/prisma/issues/21136

/integration


packages/client/src/runtime/core/types/Utils.ts
```diff
@@ -36,7 +36,7 @@ export type Narrow<A> = {
 // prettier-ignore
 export type Exact<A, W> = (A extends unknown ? (W extends A ? {
   [K in keyof W]: K extends keyof A ? Exact<A[K], W[K]> : never
-} : W) : never) | (A extends Narrowable ? A : never)
+} : W) : never) | (A extends Narrowable ? A : never) | ([W] extends [never] ? any : never)
```
millsp, line 39: It's not really `Exact`'s role to do this, in fact I initially hoped that I could that `Prisma.Args` could simply return `any` if it reveived a bad input. However, because of the sensitivity of `Prisma.Args` and its interaction with `Prisma.Exact` as well as input inference, it's not possible to do so

millsp: We can also see this differently: It does not make sense to pass `never` as the shape to be matched by `Exact`, so it can be seen as some kind of special edge case that would then allow `any` value to be matched.

millsp: An alternative to this would be to let Accelerate handle the edge case (please read the slack conversation or explanation  in the issue), yet that also does not feel right if we can potentially make it slightly better. Arguably, Accelerate is also just using our API, so our role is to make it work as good as we possibly can with the described limitation.

packages/client/tests/functional/extensions/defineExtension.ts
```diff
@@ -619,14 +619,12 @@ testMatrix.setupTestSuite(() => {
...
 
-      // @ts-test-if: provider === 'mongodb'
       const _findRaw = xprisma.user._findRaw({})
       // @ts-test-if: provider === 'mongodb'
       const findRaw = await prisma.user.findRaw({})
       // @ts-test-if: provider === 'mongodb'
       expectTypeOf<typeof _findRaw>().toEqualTypeOf<typeof findRaw>()
 
-      // @ts-test-if: provider === 'mongodb'
       const _aggregateRaw = xprisma.user._aggregateRaw({})
```
millsp, line 628: Side-effect of the change, the input turned from `never` to `any` so skipping this type check does not make sense anymore.

packages/client/src/runtime/core/types/Utils.ts
millsp: Let me know if you think there is a better way

daniel-smyth:
Keen for this!

Jolg42:
About the ecosystem-tests:
https://github.com/prisma/ecosystem-tests/actions/runs/6344577190/job/17254939099
Only the driver adapters jobs failed but I guess that’s expected for now though we need to do something about that soon

Jolg42 (approved):

aqrln (approved):

PR #21140, opened 2023-09-19
fix(cli): context-aware import path snippet for `generate` command

### LINKED ISSUE #6409: `generate` shows me TS snippet although clearly in JS project
I get this output on `npx prisma generate`:

````

✔ Generated Prisma Client (2.20.1) to .\node_modules\@prisma\client in 115ms
```
```
````

Although I am clearly in a JS project:
- No tsconfig
- Only source code file is an index.js

It would be nice if Prisma Client could do some minimal detection of the project type and output the appropriate code snippet.

### PR DESCRIPTION
**Pull request by fiznool**
As discussed in #6409, this PR implements the original proposal of switching the snippet if the project appears to be a regular JavaScript one. A similar check already exists for `isDeno` and so it seemed appropriate to do the same thing for TypeScript on Node.js, too.

Fixes #6409

### CODE REVIEW

**Comment by fiznool** [comment]
@quest-bot loot #6409

**Comment by fiznool** [comment]
Apologies for all the commits - I couldn't get the test harness running on my local machine, so I've been relying on CI to run the tests.

**Comment by janpio** [comment]
No worries, we use `Squash and merge` anyway 😆

File: packages/cli/src/Generate.ts [hunk 1]
```diff
@@ -248,17 +248,28 @@ This might lead to unexpected behavior.
 Please make sure they have the same version.`
             : ''
 
+        const isTS = fs.existsSync(path.resolve(cwd, 'tsconfig.json'))
```
millsp, line 251: Hey @fiznool, thanks a lot for your contribution! I think this can work, yet I also think there will be cases where it does not:

* I think that a TS project does not necessarily have a `tsconfig.json` at the root of the package (or the file could have a different name), for instance in some monorepos it can be at the root on the monorepo only. Would it work to look if `typescript` is installed for instance?

* The JS side now changed the snippet to `require`, in ESM projects for instance you could actually want the `import`, so how do we detect ESM projects?

fiznool: Thanks for your review. On the two points:

1. I'm not quite sure how you'd detect the presence of typescript in a non-hacky way. Do you have any thoughts on how to do this? I can think of two possibilities: using `require.resolve('typescript)` to try and load typescript, or to run the `npm ls typescript--depth=0` command. Neither feels like an elegant solution!
2. I did originally consider the ES Modules possibility, however on consultation with the [docs](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/generating-prisma-client), it seems that the code snippets for JS projects still use CommonJS. I therefore decided to maintain parity with the documentation. Does that make sense?

millsp: I think `require.resolve` is a good solution, albeit not pretty. It may be more eager than looking for `tsconfig` which in some cases could also technically be wrong (eg. JS package in monorepo that partially uses TS), in that case we would get the current behavior and maybe that's fine.

On your second point, yes that makes sense. And you reminded me that we actually had some reports that ESM imports need to work with a workaround at the moment, so I think this is fine. Also, ESM detection should be straightforward.

fiznool: Just pushed a change to use `require.resolve` for detecting if TypeScript is installed, as discussed.

fiznool: The problem we have now is that we can’t really add a test for this, since in the test harness, typescript will always be resolved. I guess we’ll have to accept that this can’t be tested? In the same way, for example, that the `isDeno` branches also cannot be tested here.

millsp: In the CLI tests (`prisma/packages/prisma/src/__tests__`) we have fixtures that could help you simulate a project with and without TypeScript. That said, I don't think we install dependencies in there, so that may be a problem. Another way would be to create a fake TypeScript module in the node_modules of the one fixture and not in the other.

Alternatively, you can add a test in `prisma/packages/client/tests/e2e`. We also have a CLI for running these, via `pnpm run test:e2e example --verbose`, so you could duplicate the example folder in there. These are more costly to write and to run.

fiznool: I think the main problem here is that `require.resolve` is going to recursively walk the tree until it finds TypeScript, which will be installed at the root of the tree when the tests are run. So I don't see an easy way to mock this, unless I mock the `canResolveTypeScript` function? Alternatively, I'm not sure if this is really worth adding test coverage for.

millsp: `test:e2e` should be run in a completely isolated test in a docker container, while the fixtures would be run in a folder in `/tmp`

fiznool:
Apologies but I can't continue with this. I can't get my head around the test harness, and why this is failing on CI (it's working on my machine). Hopefully someone else can pick this up and finish it off.

PR #21141, opened 2023-09-20
chore(client-tests): fix data proxy fetch flakiness

millsp:


janpio (approved):
👍 And ugh, at node-fetch being the source for flakyness :/

PR #21163, opened 2023-09-20
feat(client): enforce `driverAdapters` preview feature

aqrln:
* Enforce `driverAdapters` preview feature to use driver adapters at run time and not just on type level.
* Minor refactoring along the way: make `getPrismaClient` options validators type-safe.
* Add functional tests helpers for testing infrastructure around driver adapters.


packages/client/tests/functional/_utils/types.ts
```diff
@@ -5,6 +5,7 @@ export type MatrixOptions = {
     from: `${Providers}`[]
     reason: string
   }
+  skipBinary?: boolean
```
SevInf, line 8: Should we require a  reason here as well?

packages/client/src/runtime/utils/validatePrismaClientOptions.ts
```diff
@@ -50,18 +60,26 @@ It should have this form: { url: "CONNECTION_STRING" }`,
       }
     }
   },
-  adapter: (_options: any) => {
-    return
+  adapter: (adapter, config) => {
+    if (adapter === undefined) {
```
SevInf, line 64: At some point me and Joël discussed if stricter validation for `adapter` makes things safer:

1. Always throw on explicit `undefined` value, but not on missing property.
2. Allow `null` for explicitly skipping driver adapter.

Rationale for that was that if `adapter` is `undefined` by accident, it many cases it is not really obvious that adapter is not used. We ourselves have been burned by that couple of times. Doing validation this way prevents this mistake, and those who want to use adapter conditionally can still pass `null`.

What would you think about that?

SevInf (approved):
I left a couple of suggestions. I don't think any of them are blocking and strictly required for this to get merged, feel free to ignore or do them in follow up PRs.

Druue (approved):

packages/client/src/runtime/utils/getPreviewFeatures.ts
millsp, line 1: Could we move this file to `runtime/core/init`?

millsp (approved):
Neat!

millsp:
The failing tests are expected as we are automatically adding the preview feature flags, that will mess with your test. You could use `skipTestIf(driverAdapter)(...`, this variable is available in `clientMeta`.

packages/client/src/runtime/utils/validatePrismaClientOptions.ts
aqrln: Absolutely, will do this in a follow-up PR
https://github.com/prisma/team-orm/issues/388

PR #20991, opened 2023-09-06
test(client): test driver adapters

Jolg42:
WIP

Based on https://github.com/prisma/prisma/pull/20709

Related
https://github.com/prisma/team-orm/issues/275
https://github.com/prisma/team-orm/issues/258

Continued at https://github.com/prisma/prisma/pull/21168

PR #21168, opened 2023-09-20
test(client): driver adapters

millsp:
needs https://github.com/prisma/prisma/pull/21200


packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -71,6 +79,18 @@ async function main(): Promise<number | void> {
     jestCli = jestCli.withEnv({ ONLY_TEST_PROVIDERS: providers.join(',') })
   }
 
+  if (args['--driver-adapter']) {
+    if (args['--data-proxy']) {
+      throw new Error('--driver-adapter is only available when --data-proxy is used')
```
janpio, line 84:
```suggestion
      throw new Error('--driver-adapter is not available when --data-proxy is used')
```

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -64,21 +105,45 @@ function shouldSkipProvider(
...
-  return excludedProviders.includes(provider)
+  // if there is a flavor to run and it's not included, skip
+  if (flavor && !includedProviderFlavors?.includes(flavor)) {
+    return true
+  }
+
+  // if there is a flavor to run and it's excluded, skip
+  if (flavor && excludedProviderFlavors.includes(flavor)) {
+    return true
+  }
```
miguelff, line 131: We have two mechanism to skip tests, lack of inclusion and explicit exclusion, Is it a correct assumption that by default, all flavors are included, and there's no need to specify all in the list, per specific suite?

If so, wouldn't it make sense to only have a list of exclusions?

miguelff (approved):
No objection on the changes.

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
```diff
@@ -130,7 +135,7 @@ function getTestSuiteParametersString(configs: Record<string, string>[]) {
         const providerFlavorStr = config.providerFlavor === undefined ? '' : `providerFlavor=${config.providerFlavor},`
         return `relationMode=${config.relationMode},provider=${config.provider},${providerFlavorStr}onUpdate=${config.onUpdate},onDelete=${config.onDelete},id=${config.id}`
       } else {
-        const firstKey = Object.keys(config)[0]
+        const firstKey = Object.keys(config)[0] // ! TODO this can actually produce incorrect tests and break type checks ! \\ Replace with hash
```
Jolg42, line 138: Todo for tech debt?

packages/client/tests/functional/_utils/setupTestSuiteClient.ts
```diff
@@ -1,5 +1,12 @@
+import { neonConfig, Pool as neonPool } from '@neondatabase/serverless'
+import { connect } from '@planetscale/database'
+import { PrismaNeon } from '@prisma/adapter-neon' // TODO rename to `@prisma/adapter-neon`
+import { PrismaPg } from '@prisma/adapter-pg' // rename to `@prisma/adapter-pg`
```
Jolg42, line 4:
```suggestion
import { PrismaNeon } from '@prisma/adapter-neon'
import { PrismaPg } from '@prisma/adapter-pg'
```

Jolg42 (approved):
Nice! I see some code from https://github.com/prisma/prisma/pull/20991 I'll close that one then.

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
jkomyno: Does the comment says it risks producing wrong results because of Object.keys relying on an unstable order for its keys? And, does it break type checks because `Object.keys(config)` could be empty, and thus its 0th entry would be undefined?

As of `ES2020` (supported in Node.js 14.x and onwards), `Object.keys` actually follows a consistent property order. Moreover, `Reflect.ownKeys` enforces this same consistent order even on `ES2019`.

packages/client/tests/functional/_utils/getTestSuitePlan.ts
jkomyno: I think it would make sense to use exclusion lists only, as Miguel says.

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -51,6 +54,8 @@ const args = arg(
     '--shard': String,
     // Passes the same flag to Jest to silence the output
     '--silent': Boolean,
+    // Passes the same flag to Jest to silence the output
+    '--runInBand': Boolean,
```
jkomyno, line 58:
```suggestion
    // Tell Jest to run tests sequentially
    '--runInBand': Boolean,
```

packages/client/tests/functional/_example/tests.ts
```diff
@@ -47,6 +47,7 @@ testMatrix.setupTestSuite(
     })
 
     test('getTestSuiteSchema', () => {
+      // @ts-expect-error
```
jkomyno, line 50: Is this due to that `Object.keys(...)[0]` referenced elsewhere? What's the actual type error?

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
```diff
@@ -35,11 +39,10 @@ export function getTestSuiteFullName(suiteMeta: TestSuiteMeta, suiteConfig: Name
...
  */
-export function getTestSuitePreviewFeatures(matrixOptions: Record<string, string>) {
-  return [
-    ...(matrixOptions['providerFeatures']?.split(', ') ?? []),
-    ...(matrixOptions['previewFeatures']?.split(', ') ?? []),
-  ]
+export function getTestSuitePreviewFeatures(schema: string): string[] {
+  const match = schema.match(schemaPreviewFeaturesRegex)
+
+  return match === null ? [] : JSON.parse(match[1])
```
jkomyno, line 45: Is it possible that `match[1]` is undefined?

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -116,18 +115,19 @@ export async function setupTestSuiteDatabase(
...
     const dbPushParams = ['--schema', schemaPath, '--skip-generate']
-    const providerFlavor = suiteConfig.matrixOptions['providerFlavor'] as ProviderFlavor | undefined
-    // `--force-reset` is great but only using it where it's necessary makes the
-    // tests faster Since we have full isolation of tests / database, we do not
-    // need to force reset but we currently break isolation for Vitess (for
-    // faster tests), so it's good to force reset in this case
-    if (providerFlavor === ProviderFlavors.VITESS_8) {
+
+    // we reuse and clean the db when running in single-threaded mode
+    if (process.env.JEST_MAX_WORKERS === '1') {
```
jkomyno, line 120: Do `process.env.JEST_MAX_WORKERS` and `--runInBand` achieve the same goals? If so, could we just use one of those to avoid redundancy?

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -116,18 +115,19 @@ export async function setupTestSuiteDatabase(
...
     }
     await DbPush.new().parse(dbPushParams)
 
+    if (suiteConfig.matrixOptions.providerFlavor === ProviderFlavors.VITESS_8) {
+      await new Promise((r) => setTimeout(r, 300)) // wait for vitess to catch up
+    }
+
     if (alterStatementCallback) {
-      const provider = suiteConfig.matrixOptions['provider'] as Providers
+      const provider = suiteConfig.matrixOptions.provider
```
jkomyno, line 130: Nit

```suggestion
      const { provider } = suiteConfig.matrixOptions
```

Jolg42:
Note: compared to https://github.com/prisma/prisma/pull/20991 this PR implements testing only for Linux, the driver adapter tests are not running on Windows or macOS.

packages/client/tests/functional/_utils/getTestSuitePlan.ts
millsp: It works the other way around. By default, no flavors are included. As such, the exclusion could be removed, and I mostly kept it to keep the code isomorphic.

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
millsp: This is because of the matrix keys actually, this assumes that every first key in the matrix is a different provider, and that is not always the case.

packages/client/tests/functional/_example/tests.ts
millsp: This is a small thing actually, don't worry about it.

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
millsp: On a type-level in theory yes, and in fact not because the regex has a group.

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
millsp: JEST_MAX_WORKERS provides the information for us to know if it's run in band.

.github/workflows/test-template.yml
```diff
@@ -238,6 +238,71 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    if: ${{ contains(inputs.jobsToRun, '-all-') || contains(inputs.jobsToRun, '-client-') }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        flavor: ['js_pg', 'js_neon']
+        shard: ['1/6', '2/6', '3/6', '4/6', '5/6', '6/6']
+        node: [20] #[16, 18, 20]
```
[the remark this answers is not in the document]
millsp: I agree, and yet doing so might congest the amount of available jobs (500) available at a time.

PR #21189, opened 2023-09-22
ci(driver-adapters): Run new driver adapters test suite

janpio:


.github/workflows/test-template.yml
```diff
@@ -238,6 +238,70 @@ jobs:
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - run: docker compose -f docker/docker-compose.yml up --detach postgres neon_wsproxy
+
+      - name: Install & build
+        uses: ./.github/actions/setup
+        with:
+          node-version: ${{ matrix.node }}
+          pnpm-version: ${{ inputs.pnpmVersion }}
+          # Fails if set to true
```
millsp, line 277: Why?

janpio: Copy-Paste from above :shrug:

millsp: I tend to think it should be skipped, as the other jobs will type-check already

PR #21195, opened 2023-09-22
feat(cli): Add a --with-model arg to generate an example schema

### LINKED ISSUE #20915: `prisma init --with-model`
When testing things, it is often needed that after `prisma init` you add a trivial model to your `schema.prisma` file. It would be amazing if there was a simple CLI argument that adds a model to the generated file by default, making that step obsolete.

Model can be as simple as our suggestion on `prisma generate` without a model:
```
model User {
  email String  @unique
  name  String?
}
```

Kayoshi-dev:
Hello! This is my first time contributing to an open source project that big!

This add a new --with-model params you can give to the CLI to generate a new project with a predefined schema containing a User model.

It should closes #20915.

Tried my best, feel free to tell me if something seems wrong!


packages/cli/src/Init.ts
```diff
@@ -236,7 +261,11 @@ export class Init implements Command {
       fs.mkdirSync(prismaFolder)
     }
 
-    fs.writeFileSync(path.join(prismaFolder, 'schema.prisma'), defaultSchema(provider))
+    if (args['--with-model']) {
+      fs.writeFileSync(path.join(prismaFolder, 'schema.prisma'), withModelSchema(provider))
```
janpio, line 265: To avoid duplication of the `datasource` and `generator` blocks, could we re-use `defaultSchema` here and just add the model at the bottom?

packages/cli/src/Init.ts
```diff
@@ -111,6 +131,7 @@ export class Init implements Command {
              -h, --help   Display this help message
   --datasource-provider   Define the datasource provider to use: postgresql, mysql, sqlite, sqlserver, mongodb or cockroachdb
                   --url   Define a custom datasource url
+           --with-model   Set an example schema.prisma
```
janpio, line 134:
```suggestion
           --with-model   Add example model to created schema file
```

packages/cli/src/Init.ts
```diff
@@ -111,6 +121,7 @@ export class Init implements Command {
              -h, --help   Display this help message
   --datasource-provider   Define the datasource provider to use: postgresql, mysql, sqlite, sqlserver, mongodb or cockroachdb
                   --url   Define a custom datasource url
+           --with-model   Add example model to created schema file
```
janpio, line 124: Turns out, this is of course a flag - and not an option:
```suggestion

    ${bold('Flags')}

           --with-model   Add example model to created schema file
```

(Also, you could add an empty line between line 118 and 119 - that seems to be missing for a long time!)

janpio (approved):
Looks good to me!

packages/cli/src/Init.ts
Kayoshi-dev: Done! Thanks a lot for your help!

janpio (approved):
Perfect, thank you so much.

The only remaining tiny thing you _could_ do is add a snapshot for the `prisma init --help` output , as you changed that but the tests did not have to change to reflect that - so we are missing that one. But I think we could also **merge without it**, as that is not your fault.

packages/cli/src/Init.ts
```diff
@@ -35,6 +35,16 @@ datasource db {
 `
 }
 
+export const withModelSchema = (provider: ConnectorType = 'postgresql') => {
+  return defaultSchema(provider).concat(`
+model User {
+  id    Int     @id @default(autoincrement())
+  email String  @unique
+  name  String?
+}
```
millsp, line 44: Hey thanks for your contribution! I think this will produce an invalid schema for `mongodb` and `cockroachdb` providers, so we would need to add that and test that so we don't accidentally suggest an invalid schema now and in the future if we add a new provider.

millsp: That would only matter if the provider is specified as well, otherwise it is `postgresql` by default and it's fine then.

Kayoshi-dev: Hello ! You're right, I just checked online and on your documentation and it will produce an incorrect schema for both the provider you mentioned.

I'm not sure about how I should do that, do you have any idea on how I should tackle this problem ? 

Maybe I could create a new file in the "utils" folder dedicated to the creation of schemas ? 

Thanks for your feedback 😊

janpio: I think you can use the `provider` parameter that is provided to `withModelSchema` already to include a _different_ model for MongoDB and CockroachDB that is valid for these.

packages/cli/src/Init.ts
```diff
@@ -160,6 +205,9 @@ export class Init implements Command {
   
   Set up a new Prisma project and specify the url that will be used
     ${dim('$')} prisma init --url mysql://user:password@localhost:3306/mydb
+
+  Set up a new Prisma projects with a model example
```
Jolg42, line 209:
```suggestion
  Set up a new Prisma project with an example model
```

Jolg42 (approved):

Jolg42:
This is now available in our dev version `5.14.0-dev.21` and will be part of the next official version `5.14.0` scheduled for May, 14th.

PR #21271, opened 2023-09-28
feat(cli): support for custom seeder file in prisma reset

### LINKED ISSUE #21184: Provide argument `migrate reset` so that it gets used by `prisma db seed`
### Discussed in https://github.com/prisma/prisma/discussions/21173

<div type='discussions-op-text'>

<sup>Originally posted by **LamprosPitsillos** September 21, 2023</sup>
### Question

https://www.prisma.io/docs/guides/migrate/seed-database#user-defined-arguments

I use as input to my seed a json, that is passed in as an argument to the ts file.
How do i pass this to `prisma migrate reset ` so that it gets used by seed ?

```
❯ npx prisma db seed ./STjsons/test_objects_used_by_methods.json
❯ npx prisma migrate reset ./STjsons/test_objects_used_by_methods.json # doesnt work 
```
</div>

# Feature Request :
Have `migrate reset` propagate arguments to underlying commands such as `prisma db seed`

dhayaleswaran:
This PR enables `--seed-file` argument in `prisma migrate reset` to support custom seed files which are not in prisma.seed.

Closes #21184


janpio (changes requested):
Hey, I don't think the issue was asking for a custom seed file. 
The issue was about being able to supply additional parameters to `migrate reset`, which would then be collected and supplied to the internal call of `db seed` that happens at the end of `migrate reset`.

dhayaleswaran:
> Hey, I don't think the issue was asking for a custom seed file. The issue was about being able to supply additional parameters to `migrate reset`, which would then be collected and supplied to the internal call of `db seed` that happens at the end of `migrate reset`.

Can you please provide an example ?

janpio:
It's in https://github.com/prisma/prisma/issues/21184 and https://github.com/prisma/prisma/discussions/21173. The user wants to be able to do `npx prisma migrate reset ./STjsons/test_objects_used_by_methods.json` - which he then should be able to read in  the `db seed` execution that is done under the hood of `migrate reset`.

dhayaleswaran:
> It's in #21184 and #21173. The user wants to be able to do `npx prisma migrate reset ./STjsons/test_objects_used_by_methods.json` - which he then should be able to read in the `db seed` execution that is done under the hood of `migrate reset`.

Hope I did the correct changes this time

janpio:
Yeah, I think that does now more what the user asked for.
(Seems there is a test regression though=

What motivated you to add `--allow-seed-args` instead of just forwarding all of them as we do when you use `db seed` directly?

A test would probably be good to show how the params end up being available in the seed script.

dhayaleswaran:
> What motivated you to add `--allow-seed-args` instead of just forwarding all of them as we do when you use `db seed` directly?

 I just want the user and the system to know `There are some custom seed args available.` This may help in the future for further processing if exists.

> A test would probably be good to show how the params end up being available in the seed script.

I just wanted to ensure that the expectations were met before adding tests.

dhayaleswaran:
@janpio , I added tests for the changed files can you have a look on it ?

janpio:
Generally looks good.

We'll have to decide with my colleagues if we need the additional param, or if we can just forward them all the times  as the user originally suggested.

You have a few test regressions that need to be cleaned up.

dhayaleswaran:
> Generally looks good.
> 
> We'll have to decide with my colleagues if we need the additional param, or if we can just forward them all the times as the user originally suggested.
> 
> You have a few test regressions that need to be cleaned up.

Can you point what are them needs to be cleaned up ?

janpio:
image
None of these tests should be failing.

dhayaleswaran:
@janpio None of the tests are failing now. Can you please check it ?

janpio:
Thank you, then it is our turn with:

> We'll have to decide with my colleagues if we need the additional param, or if we can just forward them all the times as the user originally suggested.

I think we should remove the additional param and make this happen all the time with no need for additional configuration.

PR #21272, opened 2023-09-28
ci: retry docker startup failures

millsp:
This PR aims to improve docker download failures by retrying up to 3 times. I've been seeing these errors for a while, let's suppress the noise a bit.

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -135,7 +135,8 @@ jobs:
         uses: nick-fields/retry@v2
         with:
           timeout_minutes: 10
-          max_attempts: 3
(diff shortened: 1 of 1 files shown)


millsp:
The first run failed, with three retries so I added a delay between the retries and increased the amount of retries
https://github.com/prisma/prisma/actions/runs/6343776103/job/17232435133?pr=21272

millsp:
The login is now also failing, that's less usual https://github.com/prisma/prisma/actions/runs/6343839728/job/17232625954?pr=21272#step:3:18, I don't think I want to also retry that one, agree?

Jolg42 (approved):

PR #21294, opened 2023-09-29
test(client): planetscale driver adapter

millsp:


docker/docker-compose.yml
```diff
@@ -43,11 +43,10 @@ services:
     ports:
       - 33807:33807
     healthcheck:
-      test: ['CMD', 'mysqladmin', 'ping', '-h127.0.0.1', '-P33577']
+      test: ['CMD', 'mysqladmin', 'ping', '-h127.0.0.1', '-P33807']
```
millsp, line 46: This was incorrect making that the container is always marked unhealthy. Health state is needed for `planetscale_proxy`.

docker/docker-compose.yml
```diff
@@ -137,3 +137,20 @@ services:
       - '5488:80'
     depends_on:
       - postgres
+    restart: always # restart between reboots
+
+  planetscale_proxy:
+    build: ./planetscale_proxy
```
millsp, line 143: Proxy utility to make planetscale driver adapter communicate with vitess.

docker/planetscale_proxy/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1
+
+RUN apt update && apt install netcat-openbsd -y
```
millsp, line 3: `netcat` is needed for the healthcheck.

.github/workflows/test-template.yml
```diff
@@ -762,7 +762,7 @@ jobs:
           timeout_minutes: 10
           max_attempts: 5
           retry_wait_seconds: 120
-          command: docker compose -f docker/docker-compose.yml up --detach postgres mysql mssql mongodb_migrate mongodb_migrate_seed cockroachdb
+          command: docker compose -f docker/docker-compose.yml up --wait --detach postgres mysql mssql mongodb_migrate mongodb_migrate_seed cockroachdb
```
millsp, line 765: `--wait` takes advantage of healthchecks, so with this and the other changes we are sure for instance that `vitess` and `planetscale` are started and fully ready by the time the tests start.

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
```diff
@@ -3,20 +3,27 @@ import path from 'path'
...
 import { ClientMeta, TestCliMeta } from './types'
 
 export type TestSuiteMatrix = { [K in string]: any }[][]
 export type NamedTestSuiteConfig = {
   parametersString: string
-  matrixOptions: Record<string, string> & { provider: Providers; providerFlavor?: ProviderFlavors }
+  matrixOptions: Record<string, string> & {
+    provider: Providers
+    providerFlavor?: ProviderFlavors
+    relationMode?: RelationModes
```
millsp, line 16: I formalized what relation mode is withing the test setup itself, so now an understood concept.

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -65,6 +65,11 @@ function getExpandedTestSuitePlanWithProviderFlavors(suiteConfig: NamedTestSuite
     newSuiteConfig.parametersString += `, ${flavor}`
     // ^^^ temporary until I get to the TODO in getTestSuiteParametersString
 
+    // if the test is not doing stuff with relation mode already, we set one
+    if (newSuiteConfig.matrixOptions.relationMode === undefined) {
+      newSuiteConfig.matrixOptions.relationMode = relationModesForFlavor[flavor]
+    }
+
```
millsp, line 72: This ensures that a given flavor will have what it needs to work, or be skipped when needed.

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -135,13 +141,18 @@ function shouldSkipSuiteConfig(
     return true
   }
 
-  // if flavors are enabled and test has no flavor, skip
-  if (includedProviderFlavors && flavor === undefined) {
+  // if the flavor is explicitly skipped in the matrix options, skip
+  if (flavor !== undefined && options?.skipProviderFlavor?.from.includes(flavor)) {
     return true
   }
```
millsp, line 147: This was just moved at the bottom

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -135,13 +141,18 @@ function shouldSkipSuiteConfig(
...
 
-  // if the flavor is explicitly skipped in the matrix options, skip
-  if (flavor && options?.skipProviderFlavor?.from.includes(flavor)) {
+  // if there is a relation mode set and the flavor doesn't support it, skip
+  if (flavor !== undefined && relationMode !== undefined && relationMode !== relationModesForFlavor[flavor]) {
+    return true
+  }
```
millsp, line 152: And this is new

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -122,8 +122,11 @@ export async function setupTestSuiteDatabase(
     }
     await DbPush.new().parse(dbPushParams)
 
-    if (suiteConfig.matrixOptions.providerFlavor === ProviderFlavors.VITESS_8) {
-      await new Promise((r) => setTimeout(r, 300)) // wait for vitess to catch up
+    if (
+      suiteConfig.matrixOptions.providerFlavor === ProviderFlavors.VITESS_8 ||
+      suiteConfig.matrixOptions.providerFlavor === ProviderFlavors.JS_PLANETSCALE
+    ) {
+      await new Promise((r) => setTimeout(r, 1000)) // wait for vitess to catch up
```
millsp, line 129: That's the tablet sync time set in the docker compose, basically.

packages/client/tests/functional/extensions/itx.ts
```diff
@@ -72,6 +73,7 @@ testMatrix.setupTestSuite(({ provider, providerFlavor }, _, clientMeta) => {
       const result = xprisma.$transaction(async (tx) => {
         const userA = await tx.user.create({
           data: {
+            id: copycat.uuid(0).replaceAll('-', '').slice(-24),
```
millsp, line 76: Without this, the id is random and cannot be snapshotted.

docker/docker-compose.yml
```diff
@@ -42,12 +57,6 @@ services:
     restart: unless-stopped
     ports:
       - 33807:33807
-    healthcheck:
-      test: ['CMD', 'mysqladmin', 'ping', '-h127.0.0.1', '-P33577']
```
millsp, line 46: This healthcheck was not good and always failed, see fixed version below.

janpio: Using the correct port helps :hidethepain:

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
```diff
@@ -176,6 +185,15 @@ export function getTestSuiteSchema(suiteMeta: TestSuiteMeta, matrixOptions: Name
     schema = schema.replace(defaultGeneratorMatch[0], `${defaultGeneratorMatch[0]}\n${previewFeaturesStr}`)
   }
 
+  // for PlanetScale and Vitess, we need to add `relationMode = "prisma"` to the schema
+  if (matrixOptions.relationMode && providerMatch !== null) {
```
janpio, line 189: Aside: There is a future where relationMode is optional for both.

packages/client/tests/functional/raw-queries/send-type-hints/tests.ts
```diff
@@ -6,9 +7,10 @@ declare let prisma: PrismaClient
 declare let Prisma: typeof PrismaNamespace
 
 testMatrix.setupTestSuite(
-  (suiteConfig) => {
-    test('Buffer ($queryRaw)', async () => {
-      if (suiteConfig.provider === 'mysql') {
+  ({ provider, providerFlavor }) => {
+    // TODO the buffer returned by the driver adapter seems to be an ArrayBufferView, not a Node.js Buffer
```
aqrln, line 11: Now this is weird as it has to go all the way through our code and the buffers are theoretically always constructed by us. I would expect this test to fail for other reasons (either some conversion error or wrong data like byte values of characters of `{}` string in the response) but not like this.

packages/client/tests/functional/metrics/enabled/tests.ts
```diff
@@ -218,7 +218,8 @@ testMatrix.setupTestSuite(
       skipTestIf(
         providerFlavor === ProviderFlavors.JS_NEON ||
           providerFlavor === ProviderFlavors.JS_PG ||
-          providerFlavor === ProviderFlavors.JS_LIBSQL,
+          providerFlavor === ProviderFlavors.JS_LIBSQL ||
```
aqrln, line 221: Should the whole test suite be skipped perhaps?

packages/client/tests/functional/tracing/tests.ts
```diff
@@ -554,9 +554,9 @@ testMatrix.setupTestSuite(
   },
   {
     skipProviderFlavor: {
-      from: ['js_libsql'],
+      from: ['js_libsql', 'js_planetscale'],
       reason:
-        'The spans are not consistent, or not in a way that makes sense compared to the order driver adapters. Needs investigation.',
+        'The spans are not consistent, or not in a way that makes sense compared to the other driver adapters. Needs investigation.',
```
aqrln, line 559: This is what I said before: libSQL and PlanetScale adapters use high-level APIs for transactions provided by the underlying drivers and we don't run `BEGIN` and `COMMIT`/`ROLLBACK` statements ourselves, hence no spans for them.

aqrln: If there are any other issues, those would need investigation, but if it's just missing spans for `BEGIN`/`COMMIT`, then it's well understood and expected (what we want to or even could do about it though is another question).

aqrln (commented):
left a few comments but overall looks very good!

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
millsp: Yes. I was hesitant to implement `relationModesForFlavor` as an object of lists instead of a single `relationMode` per provider, perhaps I should have done that. This can easily be integrated in the future.

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -135,13 +141,18 @@ function shouldSkipSuiteConfig(
...
-  if (includedProviderFlavors && flavor === undefined) {
+  // if the flavor is explicitly skipped in the matrix options, skip
+  if (flavor !== undefined && options?.skipProviderFlavor?.from.includes(flavor)) {
     return true
   }
 
-  // if the flavor is explicitly skipped in the matrix options, skip
-  if (flavor && options?.skipProviderFlavor?.from.includes(flavor)) {
+  // if there is a relation mode set and the flavor doesn't support it, skip
+  if (flavor !== undefined && relationMode !== undefined && relationMode !== relationModesForFlavor[flavor]) {
```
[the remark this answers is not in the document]
millsp: I think the condition is safe enough in this case with `relationMode !== undefined` and `relationMode !== relationModesForFlavor[flavor]` means they should not have an unexpected overlap. Did you have an example in mind that I could have missed?

packages/client/tests/functional/optimistic-concurrency-control/tests.ts
```diff
@@ -17,7 +18,8 @@ testMatrix.setupTestSuite(({ provider }) => {
     await prisma.resource.deleteMany()
   })
 
-  test('updateMany', async () => {
+  // TODO optimistic concurrency control is not working for JS_PLANETSCALE
```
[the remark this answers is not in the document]
millsp: Yes, I will open the issues today

packages/client/tests/functional/metrics/enabled/tests.ts
millsp: Yes, I think we could at this stage, or try to make things more granular?

packages/client/tests/functional/_utils/getTestSuitePlan.ts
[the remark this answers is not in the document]
aqrln: No, you are totally right, it does imply `relationModesForFlavor[flavor] !== undefined`. Sorry!

packages/client/tests/functional/metrics/enabled/tests.ts
aqrln: Leaving it granular might make things easier in the future so that we could un-skip them one by one, so no strong preference either way.

aqrln (approved):

PR #21312, opened 2023-10-02
chore: add `.husky` to `.gitignore`

aqrln:
Husky overrides `core.hooksPath` to `.husky` instead of using default
`.git/hooks` but only makes Git ignore the specific scripts that come
from husky itself but not from other tooling (which ends up in `.husky`
too because of the configuration override). This commit adds the whole
directory to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+.husky/
 
 yarn-error.log
 dist


Jolg42:
It doesn't look like we should do this to me 🤔 

Random find: https://stackoverflow.com/a/66218554/1345244

Jolg42:
Note: this was necessary because @aqrln is trying out `git-branchless` and it installs hooks to the repo to detect any changes to the commit graph, but husky takes over git configuration of the repo to make all git hooks (whether installed by husky or not) live in the .husky directory

Jolg42 (approved):

PR #21314, opened 2023-10-02
feat(cli): introduce multiple flags for init

### LINKED ISSUE #17737: Prisma init: allow setting generator provider value

steebchen:
Add new flags for `prisma init` to allow more detailed configuration,
specifically `--generator-provider`, `--preview-features` and `--output`.

Supersedes #17875

Fixes #17737

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -131,9 +131,9 @@ export class Init implements Command {
     
              -h, --help   Display this help message
   --datasource-provider   Define the datasource provider to use: postgresql, mysql, sqlite, sqlserver, mongodb or cockroachdb
-   --generator-provider   Define the generator provider to use (by default, it will be prisma-client-js)
-      --preview-feature   Define a preview feature to use (by default, there will be none)
-               --output   Define the output path to use (by default, it will not be defined)
+   --generator-provider   Define the generator provider to use. Default: \`prisma-client-js\`
+      --preview-feature   Define a preview feature to use.
+               --output   Define Prisma Client generator output path to use.
                   --url   Define a custom datasource url
 
   ${bold('Examples')}
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -20,16 +20,31 @@ import { isError } from 'util'
 
 import { printError } from './utils/prompt/utils/print'
 
-export const defaultSchema = (provider: ConnectorType = 'postgresql') => {
+export const defaultSchema = (props?: {
+  datasourceProvider?: ConnectorType
+  generatorProvider?: string
+  previewFeatures?: string[]
+  output?: string
+}) => {
+  const {
+    datasourceProvider = 'postgresql',
+    generatorProvider = defaultGeneratorProvider,
+    previewFeatures = defaultPreviewFeatures,
+    output = defaultOutput,
+  } = props || {}
   return `// This is your Prisma schema file,
 // learn more about it in the docs: https://pris.ly/d/prisma-schema
 
 generator client {
-  provider = "prisma-client-js"
-}
+  provider = "${generatorProvider}"
+${
+  previewFeatures.length > 0
+    ? `  previewFeatures = [${previewFeatures.map((feature) => `"${feature}"`).join(', ')}]\n`
+    : ''
+}${output != defaultOutput ? `  output = "${output}"\n` : ''}}
 
 datasource db {
-  provider = "${provider}"
+  provider = "${datasourceProvider}"
   url      = env("DATABASE_URL")
 }
 `
@@ -95,6 +110,12 @@ export const defaultGitIgnore = () => {
 `
 }
 
+export const defaultGeneratorProvider = 'prisma-client-js'
+
+export const defaultPreviewFeatures = []
+
+export const defaultOutput = 'node_modules/.prisma/client'
+
 export class Init implements Command {
   static new(): Init {
     return new Init()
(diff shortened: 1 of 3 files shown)


packages/cli/src/Init.ts
```diff
@@ -110,6 +131,9 @@ export class Init implements Command {
     
              -h, --help   Display this help message
   --datasource-provider   Define the datasource provider to use: postgresql, mysql, sqlite, sqlserver, mongodb or cockroachdb
+   --generator-provider   Define the generator provider to use: by default, it will be prisma-client-js
+      --preview-feature   Define the preview features to use: by default, there will be none
```
janpio, line 135: Adapting wording a bit more to match the fact this can be used multiple times:
```suggestion
      --preview-feature   Define a preview feature to use: by default, there will be none
```

packages/cli/src/Init.ts
```diff
@@ -110,6 +131,9 @@ export class Init implements Command {
     
              -h, --help   Display this help message
   --datasource-provider   Define the datasource provider to use: postgresql, mysql, sqlite, sqlserver, mongodb or cockroachdb
+   --generator-provider   Define the generator provider to use: by default, it will be prisma-client-js
+      --preview-feature   Define a preview feature to use: by default, there will be none
+               --output   Define the output path to use: by default, it will be node_modules/.prisma/client
```
janpio, line 136: Now actually looking at it in context, the `:` is to be used to list the options.
The default output is also implicit, so I suggest to change that wording as well:

```suggestion
   --generator-provider   Define the generator provider to use (by default, it will be prisma-client-js)
      --preview-feature   Define a preview feature to use (by default, there will be none)
               --output   Define the output path to use (by default, it will not be defined)
```

packages/cli/src/Init.ts
```diff
@@ -119,6 +143,15 @@ export class Init implements Command {
 
   Set up a new Prisma project and specify MySQL as the datasource provider to use
     ${dim('$')} prisma init --datasource-provider mysql
+
+  Set up a new Prisma project and specify prisma-client-go as the generator provider to use
+    ${dim('$')} prisma init --generator-provider prisma-client-go
```
steebchen, line 148:
```suggestion
  Set up a new Prisma project and specify 'go run github.com/steebchen/prisma-client-go' as the generator provider to use
    ${dim('$')} prisma init --generator-provider "go run github.com/steebchen/prisma-client-go"
```

as prisma-client-go would probably work but it's not recommended to do

cc @janpio

packages/cli/src/Init.ts
```diff
@@ -110,6 +131,9 @@ export class Init implements Command {
     
              -h, --help   Display this help message
   --datasource-provider   Define the datasource provider to use: postgresql, mysql, sqlite, sqlserver, mongodb or cockroachdb
+   --generator-provider   Define the generator provider to use (by default, it will be prisma-client-js)
+      --preview-feature   Define a preview feature to use (by default, there will be none)
+               --output   Define the output path to use (by default, it will not be defined)
```
Jolg42, line 136:
```suggestion
   --generator-provider   Define the generator provider to use. Default: `prisma-client-js`
      --preview-feature   Define a preview feature to use.
               --output   Define Prisma Client generator output path to use.
```

steebchen: Leaving this up to @janpio 😅

steebchen: @Jolg42 did you see Jan’s previous suggestions?

packages/cli/src/Init.ts
janpio: 🤷

packages/cli/src/Init.ts
janpio: Looks good to me, I just iterated on what was there before.
This does look better.

packages/cli/src/Init.ts
```diff
@@ -119,6 +143,15 @@ export class Init implements Command {
...
   Set up a new Prisma project and specify MySQL as the datasource provider to use
     ${dim('$')} prisma init --datasource-provider mysql
+
+  Set up a new Prisma project and specify prisma-client-go as the generator provider to use
+    ${dim('$')} prisma init --generator-provider prisma-client-go
+
+  Set up a new Prisma project and specify x and y as the preview features to use
+    ${dim('$')} prisma init --preview-feature x --preview-feature y
+
+  Set up a new Prisma project and specify ./generated-client as the output path to use
```
Jolg42, line 153:
```suggestion
  Set up a new Prisma project and specify `./generated-client` as the output path to use
```

packages/cli/src/Init.ts
```diff
@@ -119,6 +143,15 @@ export class Init implements Command {
 
   Set up a new Prisma project and specify MySQL as the datasource provider to use
     ${dim('$')} prisma init --datasource-provider mysql
+
+  Set up a new Prisma project and specify prisma-client-go as the generator provider to use
+    ${dim('$')} prisma init --generator-provider prisma-client-go
+
+  Set up a new Prisma project and specify x and y as the preview features to use
```
Jolg42, line 150:
```suggestion
  Set up a new Prisma project and specify `x` and `y` as the preview features to use
```

packages/cli/src/Init.ts
```diff
@@ -119,6 +143,15 @@ export class Init implements Command {
 
   Set up a new Prisma project and specify MySQL as the datasource provider to use
     ${dim('$')} prisma init --datasource-provider mysql
+
+  Set up a new Prisma project and specify prisma-client-go as the generator provider to use
```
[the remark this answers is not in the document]
Jolg42: Ok, to be done later because I can't commit that apparently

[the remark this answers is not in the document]
steebchen: you'd have to escape it anyway :D

steebchen:
Thanks for the reviews and merge!

PR #21315, opened 2023-10-02
feat(client): disallow explicit `undefined` value for `adapter`

aqrln:
Throw on explicit `undefined` value of `adapter` (but not on missing
property) and allow `null` to conditionally or explicitly disable driver
adapters, as suggested in <https://github.com/prisma/prisma/pull/21163#discussion_r1332724218>.

Closes https://github.com/prisma/team-orm/issues/388

### DIFF
--- a/packages/client/src/runtime/utils/validatePrismaClientOptions.ts
+++ b/packages/client/src/runtime/utils/validatePrismaClientOptions.ts
@@ -66,7 +66,7 @@ It should have this form: { url: "CONNECTION_STRING" }`,
     }
     if (adapter === undefined) {
       throw new PrismaClientConstructorValidationError(
-        `"adapter" property must not be undefined, use null to conditionally disable driver adapters`,
+        `"adapter" property must not be undefined, use null to conditionally disable driver adapters.`,
       )
     }
     const previewFeatures = getPreviewFeatures(config)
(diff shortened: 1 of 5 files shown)


millsp (approved):

packages/client/src/runtime/utils/validatePrismaClientOptions.ts
```diff
@@ -61,9 +61,14 @@ It should have this form: { url: "CONNECTION_STRING" }`,
     }
   },
   adapter: (adapter, config) => {
-    if (adapter === undefined) {
+    if (adapter === null) {
       return
     }
+    if (adapter === undefined) {
+      throw new PrismaClientConstructorValidationError(
+        `"adapter" property must not be undefined, use null to conditionally disable driver adapters.`,
```
Jolg42, line 69: Additionally, we could also mention: Or remove the `driverAdapters` preview feature flag?

Jolg42: (though that would not be an option once it's GA)

aqrln: Without the feature flag though it would be an unknown option.

aqrln: at least on the type level

PR #21333, opened 2023-10-04
test(client): attempt to fix some driver adapter flakiness

millsp:


packages/client/tests/functional/0-legacy-ports/batch-find-unique/tests.ts
```diff
@@ -70,6 +72,10 @@ testMatrix.setupTestSuite(({ providerFlavor }) => {
       prisma.user.findUnique({ where: { email: copycat.email(4) } }),
     ])
 
+    await waitFor(() => {
+      expect(executedBatchQuery).toMatchSnapshot()
+    })
```
millsp, line 77: This should be more resilient now, as we had failure with missing an assertion and needing to update the snapshot.

packages/client/tests/functional/_utils/setupTestSuiteClient.ts
```diff
@@ -1,13 +1,5 @@
-import { createClient as createLibSqlClient } from '@libsql/client'
-import { neonConfig, Pool as neonPool } from '@neondatabase/serverless'
-import { connect } from '@planetscale/database'
-import { PrismaLibSQL } from '@prisma/adapter-libsql'
-import { PrismaNeon } from '@prisma/adapter-neon'
-import { PrismaPg } from '@prisma/adapter-pg'
-import { PrismaPlanetScale } from '@prisma/adapter-planetscale'
```
millsp, line 7: I think that isolating their loading is better as they could affect the regular tests that don't use them. After adding them here, I had seen an error in jest that pointed out that something tried to do an import after the tests were finished, so I have a hunch that something like this may be happening in the drivers. Also, lately I am seeing more frequent timeouts on unrelated tests, so I am trying to isolate the problem. All in all, I think better isolation is good.

millsp: It looks like it fixed the timeouts after a few runs.

docker/planetscale_proxy/Dockerfile
```diff
@@ -9,7 +9,7 @@ ENTRYPOINT /go/bin/ps-http-sim \
   -http-port=8085 \
   -mysql-addr=$MYSQL_HOST \
   -mysql-port=$MYSQL_PORT \
-  -mysql-idle-timeout=5s \
+  -mysql-idle-timeout=1200s \
```
millsp, line 12: Aims to fix the following error, keep an eye out if you see this again and please let me know if you do. The source code of this planetscale proxy says that the timeout code has race conditions, so I aimed at a very high timeout to see if we can fix this.

```
 DatabaseError: Code: INTERNAL
    invalid sequence, expected X got Y
     (errno 2013) (sqlstate HY000) during query: SELECT `test-0000-00000000`.`User`.`id` FROM `test-0000-00000000`.`User` WHERE (`test-0000-00000000`.`User`.`id` = 'a1e8d6df4d7adb11db8b0ac9' AND 1=1) LIMIT 1 OFFSET 0

      at Connection.execute (../../../../node_modules/.pnpm/@planetscale+database@1.11.0/node_modules/@planetscale/database/dist/cjs/index.js:99:19)
      at PrismaPlanetScale.performIO (../../../../node_modules/.pnpm/@prisma+adapter-planetscale@0.4.0_@planetscale+database@1.11.0/node_modules/@prisma/adapter-planetscale/dist/index.js:143:22)
      at PrismaPlanetScale.queryRaw (../../../../node_modules/.pnpm/@prisma+adapter-planetscale@0.4.0_@planetscale+database@1.11.0/node_modules/@prisma/adapter-planetscale/dist/index.js:114:54)
      at ../../../../node_modules/.pnpm/@prisma+driver-adapter-utils@0.8.0/node_modules/@prisma/driver-adapter-utils/dist/index.mjs:71:14
```

packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
```diff
@@ -308,7 +308,13 @@ Invalid \`prisma.user.create()\` invocation in
...
 `;
 
-exports[`interactive-transactions (provider=sqlite, js_libsql) batching raw rollback 1`] = `: UNIQUE constraint failed: User.email`;
+exports[`interactive-transactions (provider=sqlite, js_libsql) batching raw rollback 1`] = `
+
+Invalid \`prisma.$executeRaw()\` invocation:
+
+
+SQLITE_CONSTRAINT_UNIQUE: UNIQUE constraint failed: User.email
+`;
```
millsp, line 317: This now seems to be a constant failure, and I can also reproduce on my laptop.

packages/client/tests/functional/_utils/setupTestSuiteClient.ts
millsp: Seems like we still have some of these new timeouts https://github.com/prisma/prisma/actions/runs/6400641262/job/17374539235?pr=21333#step:8:245

millsp: I have another idea for investigating the timeouts, will open a separate PR later.

jkomyno:
I see that "CI / Tests / Client (without types test) (3/5 ...)" now succeeds 🚀
But it looks that "CI / Tests / Client (without types test) (2/5, library, 20)" is flakier now?

packages/client/tests/functional/_utils/setupTestSuiteClient.ts
```diff
@@ -119,27 +111,36 @@ export function setupTestSuiteClientDriverAdapter({
   }
 
   if (providerFlavor === ProviderFlavors.JS_PG) {
-    const pool = new pgPool({
+    const { Pool } = require('pg') as typeof import('pg')
```
jkomyno, line 114: It's this `as typeof import` actually required? (no pun intended)

jkomyno (approved):

packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
Jolg42: Do we need to create an issue for tracking this?

packages/client/tests/functional/_utils/setupTestSuiteClient.ts
millsp: I will try to see if I can put it as a type import instead

millsp: `require` is not typed for packages, I was surprised

packages/client/tests/functional/interactive-transactions/__snapshots__/tests.ts.snap
millsp: We have an issue for improving errors with `libsql` I believe, I'll double check

PR #21363, opened 2023-10-05
test(client): fix timeout flakiness

millsp:
I noticed the timeouts we are encountering are exclusive to `mssql` and I think it is related to a recent change which optimizes for the reuse of the db when we are running in band. I think the timeout happens during the reset of the databasen I am thus changing the logic a bit so that it behaves more like before. The real fix of course would be to understand why `mssql` times out when we reset the db, of course. Not today. This PR also makes running the tests slightly faster.

### DIFF
--- a/packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
+++ b/packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
@@ -111,7 +111,7 @@ function setupTestSuiteMatrix(
 
         globalThis['db'] = {
           setupDb: () => setupTestSuiteDatabase(suiteMeta, suiteConfig, [], options?.alterStatementCallback),
-          dropDb: () => dropTestSuiteDatabase(suiteMeta, suiteConfig),
+          dropDb: () => dropTestSuiteDatabase(suiteMeta, suiteConfig).catch(() => {}),
         }
       })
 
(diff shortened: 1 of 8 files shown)


packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -143,7 +143,7 @@ function setupTestSuiteMatrix(
           }
         }
         clients.length = 0
-        if (options?.skipDb !== true || (options?.skipDb !== true && process.env.JEST_MAX_WORKERS !== '1')) {
+        if (options?.skipDb !== true && process.env.TEST_REUSE_DATABASE !== 'true' && process.env.CI !== 'true') {
```
millsp, line 146: Only drops the db if no skipped, and if the db does not need to be reused. And also, if we are on CI why care to waste time dropping the db, also avoid dropping it.

packages/client/tests/functional/_utils/types.ts
```diff
@@ -26,6 +26,11 @@ export type NewPrismaClient<T extends new (...args: any) => any> = (
   ...args: ConstructorParameters<T>
 ) => InstanceType<T>
 
+export type Db = {
+  setupDb: () => Promise<void>
+  dropDb: () => Promise<void>
+}
```
millsp, line 32: This is now available in the tests

Jolg42 (approved):

PR #21370, opened 2023-10-05
fix: @prisma/nextjs-monorepo-workaround-plugin creates "schema.prisma1 could not be found" bug

### LINKED ISSUE #19784: NextJS monoerpo "schema.prisma1 could not be found"
### Bug description

I'm having the issue with prisma files not found I cannot provide a reproduction right now. 

I'm using Next 13 with the app dir inside a monorepo and having `db` package.
I'm using the monorepo workaround which works almost everytime but it seems it hits some race conditions at times.

When going to a route the "compiling" messages shows and as soon as it's ready the page loads and throws the error. I can see that during compilation the .next/server/app/<example> get's created and before the prisma files get copied the app tries to load and fails. 

Most times a refresh solves the issue and I can see the prisma files being copied into the newly created .next folder, where some other times it wont.

This happens a lot more consistently when doing client side navigation to a route or requesting an API which is not compiled yet, and needs to be compiled on the fly.

The compilation most times shows a success message after copying the files. I've checed the Prisma plugin code and can verify that when the error happens it doesn't get to the copy file code.

This usually happens a lot more on API routes when requested. i.e  The next-auth route which get's requested from the client and/or the server.

### Expected behavior

I expect the compilation to complete correctly 100% of the time after the Prisma plugin copies the files into the correct `.next/<path_to_route` folder

### Prisma information

```prisma
generator client {
  provider             = "prisma-client-js"
  output               = "../generated/prisma-client"
}

```

```ts
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: v18.13.0

### Prisma Version

```
prisma                  : 4.14.0
@prisma/client          : 4.13.0
Query Engine (Node-API) : libquery-engine d9a4c5988f480fa576d43970d5a23641aa77bc9c (at ../../node_modules/.pnpm/@prisma+engines@4.14.0/node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Migration Engine        : migration-engine-cli d9a4c5988f480fa576d43970d5a23641aa77bc9c (at ../../node_modules/.pnpm/@prisma+engines@4.14.0/node_modules/@prisma/engines/migration-engine-debian-openssl-3.0.x)
Format Wasm             : @prisma/prisma-fmt-wasm 4.14.0-67.d9a4c5988f480fa576d43970d5a23641aa77bc9c
Default Engines Hash    : d9a4c5988f480fa576d43970d5a23641aa77bc9c
```

toniopelo:
fixes #19784;

The plugin overrides its own copy destination when several assets are output in the same compilation run.
It causes the "schema.prisma1 could not be found" bug on a "random" basis.

The problem essentially relies on the fact that for every compilation run we assume that one prisma file is copied only once to a single destination, so we keep a 1:1 mapping between original file and copy destination.
We build this mapping and at the end we perform the copying, but when a compilation run needs to output several source files that all imports the prisma files, our 1:1 mapping gets overridden and the last source file wins.
As a result, the prisma files will be copied only once instead of several times.
This bug is randomly happening because sometimes the source files that doesn't get their prisma files copied already had them because of a previous compilation run.

This PR should fix this.

There might be room for improvement here, as we copy the same files to several places.
I guess we could copy them only once and change the import path of the generated assets instead. In development, we might not care, but we might care for the production build ? I'm not sure how this impact production build and `nft.json` files, so I just kept the current implementation regarding this to avoid breaking things.

### DIFF
--- a/packages/nextjs-monorepo-workaround-plugin/index.js
+++ b/packages/nextjs-monorepo-workaround-plugin/index.js
@@ -149,7 +149,6 @@ class PrismaPlugin {
 
     // copy prisma files to output as the final step (for all users)
     compiler.hooks.done.tapPromise('PrismaPlugin', async () => {
-      console.log(origins, originAssetsToCopies)
       const asyncActions = Object.entries(originAssetsToCopies).map(async ([from, copies]) => {
         await Promise.all(
           copies.map(async (copy) => {
--- a/packages/nextjs-monorepo-workaround-plugin/index.js
+++ b/packages/nextjs-monorepo-workaround-plugin/index.js
@@ -117,9 +117,9 @@ class PrismaPlugin {
           name: 'PrismaPlugin',
           stage: Compilation.PROCESS_ASSETS_STAGE_ANALYSE,
         },
-        async (assets) => {
+        (assets) => {
           const nftAssetNames = Object.keys(assets).filter((k) => k.endsWith('.nft.json'))
-          const nftAsyncActions = nftAssetNames.map((assetName) => {
+          nftAssetNames.forEach((assetName) => {
             // prepare paths
             const outputDir = compiler.outputPath
             const assetPath = path.resolve(outputDir, assetName)
@@ -142,7 +142,7 @@ class PrismaPlugin {
             compilation.updateAsset(assetName, newRawSource)
           })
 
-          await Promise.all(nftAsyncActions)
+          return Promise.resolve()
         },
       )
     })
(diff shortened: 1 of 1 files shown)


stuible:
I would love to see this change merged.  I can confirm it fixes my issues with Prisma and Next.js 14 in a Turbo Repo monorepo.

5starkarma:
Any updates here?

stuible:
Our team is still waiting for this fix to be released.  Are there any blocking changes preventing this pull request from being merged?

FGoessler (approved):
Am merging this although for current Prisma versions the schema file seems to not be strictly required anymore. Users still having this issue with older version but cannot upgrade Prisma yet, can use the updated plugin version.

Issue #21369, opened 2023-10-05
Planetscale driver: unsupported column type NULL


JClackett:
Getting this error when running a raw query

```
Error: Unsupported column type: NULL
    at fieldToColumnType (/var/task/node_modules/@prisma/adapter-planetscale/dist/index.js:79:13)
    at /var/task/node_modules/@prisma/adapter-planetscale/dist/index.js:118:42
    at Array.map (<anonymous>)
    at PrismaPlanetScale.queryRaw (/var/task/node_modules/@prisma/adapter-planetscale/dist/index.js:118:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /var/task/node_modules/@prisma/client/runtime/library.js:26:4983 {
  clientVersion: '5.4.1'
```

query is:

```mysql
      SELECT
        Spot.id, Spot.name, Spot.type, Spot.address, null as image, null as blurHash,
        (SELECT AVG(rating) FROM Review WHERE Review.spotId = Spot.id) AS rating,
        (CAST(COUNT(ListSpot.spotId) as CHAR(32))) AS savedCount
      FROM
        Spot
      LEFT JOIN
        ListSpot ON Spot.id = ListSpot.spotId
      WHERE
        Spot.creatorId = ? AND Spot.deletedAt IS NULL AND (Spot.creatorId = ? OR (Spot.publishedAt IS NULL OR Spot.publishedAt < NOW()))
      GROUP BY
        Spot.id
      ORDER BY
        Spot.createdAt DESC, Spot.id
      LIMIT ?
      OFFSET ?;
```

aqrln:
Thanks for the report @JClackett! We need the schema to be able to reproduce this, could you please post it here as well?

aqrln:
@JClackett btw, are you using a `$queryRaw`, or did you post an SQL statement generated by some Prisma query? I don't think it should matter either way for the reproduction since the error happens when decoding the response in the driver adapter, but asking just in case.

JClackett:
Yeah using queryRaw!

JClackett:
just thinking about it, could it because im creating "null" columns manually in the query?

```sql
Spot.id, Spot.name, Spot.type, Spot.address, null as image, null as blurHash,
```

Im doing this because I then populate the result with each spots latest image, and wanted to make sure that the resulting query had the keys available (they are just unknown at that point)

aqrln:
I reproduced the issue, should be an easy fix.

[...]

PR #21377, opened 2023-10-05
test(client): add regression test for gh-21369

aqrln:
Add regression test for <https://github.com/prisma/prisma/issues/21369>.

Refs: https://github.com/prisma/prisma-engines/pull/4320

### DIFF
--- a/packages/client/tests/functional/issues/21369-select-null/_matrix.ts
+++ b/packages/client/tests/functional/issues/21369-select-null/_matrix.ts
@@ -1,0 +1,4 @@
+import { defineMatrix } from '../../_utils/defineMatrix'
+import { sqlProviders } from '../../_utils/providers'
+
+export default defineMatrix(() => [sqlProviders])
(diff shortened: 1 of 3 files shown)


packages/client/package.json
```diff
@@ -82,7 +82,7 @@
     "@prisma/adapter-libsql": "5.4.0",
     "@prisma/adapter-neon": "5.4.0",
     "@prisma/adapter-pg": "5.4.0",
-    "@prisma/adapter-planetscale": "5.4.0",
+    "@prisma/adapter-planetscale": "5.4.1-pre",
```
millsp, line 85: Does "pre" stand for pre-release version?

millsp (approved):

packages/client/package.json
Jolg42, line 85: We should switch to a dev version later (once automated publishing is done)

Jolg42 (approved):

packages/client/package.json
aqrln: Yes

PR #21378, opened 2023-10-05
test(client): introduce sqlProviders to use in test matrix

aqrln:
Introduce `sqlProviders` helper similar to `allProviders` but without MongoDB.

### DIFF
--- a/packages/client/tests/functional/_utils/providers.ts
+++ b/packages/client/tests/functional/_utils/providers.ts
@@ -37,9 +37,9 @@ export const relationModesForFlavor = {
   [ProviderFlavors.VITESS_8]: RelationModes.PRISMA,
 } as Record<ProviderFlavors, RelationModes | undefined>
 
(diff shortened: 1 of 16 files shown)


packages/client/tests/functional/_utils/providers.ts
```diff
@@ -37,9 +37,9 @@ export const relationModesForFlavor = {
   [ProviderFlavors.VITESS_8]: RelationModes.PRISMA,
 } as Record<ProviderFlavors, RelationModes | undefined>
 
-export type AllProviders = { provider: Providers }[]
```
aqrln, line 40: This type wasn't imported anywhere and wasn't strictly necessary for the annotation below as the same type was inferred.

millsp (approved):

Jolg42 (approved):
Thanks for bringing this idea back from my PR that wasn't merged 💚

PR #21382, opened 2023-10-06
feat(client): wasm engine

millsp:
/integration

This PR accommodates the upcoming wasm engine. It currently has a dummy query engine for testing that it has the necessary things in place to be able to run the wasm engine. You can run it yourself with the following instructions:

1. Install the latest integration version `5.5.0-integration-feat-client-wasm-engine.5`
1. In your schema, add `driverAdapters` to the preview features
1. In your schema, add `engineType = "wasm"`
1. Generate the client again
1. You can now use in in Node via `@prisma/client` or
1. You can use it in a Cloudflare Worker via `@prisma/client/edge`
1. Setup the driver adapter


packages/client/package.json
```diff
@@ -93,6 +93,7 @@
     "@prisma/internals": "workspace:*",
     "@prisma/migrate": "workspace:*",
     "@prisma/mini-proxy": "0.9.4",
+    "@prisma/query-engine-wasm": "5.5.0-3.6e26301fe272ba4ba0598fe43eb5d8df030be4db",
```
millsp, line 96: Fictitious package for now.

packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -37,14 +38,17 @@ export function getEngineInstance(clientConfig: GetPrismaClientConfig, engineCon
 
   const engineType = getClientEngineType(engineConfig.generator!)
 
-  if (url?.startsWith('prisma://') || clientConfig.noEngine || TARGET_ENGINE_TYPE === 'edge') {
+  if (engineType === ClientEngineType.Wasm && engineConfig.adapter !== undefined) {
+    return new WasmEngine(engineConfig)
```
millsp, line 42: The engine is currently always available on all the builds.

packages/client/src/runtime/core/engines/wasm/WasmEngine.ts
```diff
@@ -0,0 +1,42 @@
...
+  constructor(config: EngineConfig) {
+    super(config, {
+      async loadLibrary() {
+        const wasmMod = await config.getQueryEngineWasmModule?.()
+        // from https://developers.cloudflare.com/workers/runtime-apis/webassembly/rust/#javascript-plumbing-wasm-bindgen
+        // ./mywasmlib_bg.js is generated by wasm-bindgen and written _inside_ the wasm module itself so wasm can also use it
+        // on cloudflare workers and browsers, files cannot be read, so we need to pass the module directly for that filename
+        // TODO ./mywasmlib_bg.js needs to be replaced with the actual name from the wasm-bindgen build
+        const instance = new WebAssembly.Instance(wasmMod, { './mywasmlib_bg.js': wasmBindgenRuntime })
+        wasmBindgenRuntime.__wbg_set_wasm(instance.exports)
```
millsp, line 24: This works both for CF and Node.

packages/client/package.json
janpio: It actually exists now 😆 
https://prisma-company.slack.com/archives/C05Q1T8RJ8K/p1696608960033699?thread_ts=1696568642.779789&cid=C05Q1T8RJ8K

packages/client/src/generation/generateClient.ts
```diff
@@ -313,6 +313,14 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
     await fs.copyFile(schemaPath, schemaTargetPath)
   }
 
+  // copy the necessary engine files needed for the wasm/driver-adapter engine
+  if (getClientEngineType(generator) === ClientEngineType.Wasm) {
+    const queryEngineWasmModulePath = path.dirname(require.resolve('@prisma/query-engine-wasm'))
+    const queryEngineWasmFilePath = path.join(queryEngineWasmModulePath, 'query_engine.wasm')
```
millsp, line 319:
```suggestion
    const queryEngineWasmFilePath = path.join(queryEngineWasmModulePath, 'query-engine.wasm')
```

packages/client/src/runtime/core/engines/common/Engine.ts
```diff
@@ -147,6 +147,13 @@ export interface EngineConfig {
    * in the current working directory. This usually means it has been bundled.
    */
   isBundled?: boolean
+
+  /**
+   * Gets the raw wasm module for the query engine. This configurations is
+   * generated specifically for each type of client, eg. Node.js client and Edge
+   * clients will have different implementations.
```
millsp, line 154: Maybe also worth mentioning that is is a callback on *purpose*. We only load the wasm if we need it.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -80,36 +80,24 @@ export class LibraryEngine extends Engine<undefined> {
...
-      } else if (config.isBundled === true) {
-        throw new PrismaClientInitializationError(
-          `Prisma Client could not find its \`schema.prisma\`. This is likely caused by a bundling step, which leads to \`schema.prisma\` not being copied near the resulting bundle. We would appreciate if you could take the time to share some information with us.
-Please help us by answering a few questions: https://pris.ly/bundler-investigation-error`,
-          config.clientVersion!,
-        )
-      }
+    config.inlineSchema
 
-      throw e
```
millsp, line 104: The schema is inlined nowadays, we don't need to read it from disk anymore. At the same time, it is needed for the qe to work in edge environments.

packages/client/src/generation/utils/buildGetQueryEngineWasmModule.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (engineType !== ClientEngineType.Wasm) {
+    return `config.getQueryEngineWasmModule = undefined`
+  }
+
+  // this could work on cf workers without esm support because dynamic imports
+  // work both in CJS and ESM we need to try this out as soon as possible
+  if (edge === true) {
+    return `config.getQueryEngineWasmModule = async () => {
+  return (await import('./query-engine.wasm')).default
+}`
```
millsp, line 17: It works!

packages/client/src/runtime/core/engines/library/DefaultLibraryLoader.ts
```diff
@@ -60,23 +60,17 @@ export function load(libraryPath: string): Library {
...
+      return config.tracingHelper.runInChildSpan({ name: 'loadLibrary', internal: true }, () => load(enginePath))
     } catch (e) {
       const errorMessage = handleLibraryLoadingErrors({ e: e as Error, platformInfo, id: enginePath })
 
-      throw new PrismaClientInitializationError(errorMessage, this.config.clientVersion!)
+      throw new PrismaClientInitializationError(errorMessage, config.clientVersion!)
     }
-  }
+  },
 }
```
millsp, line 76: This does not need to be a class

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -80,36 +80,28 @@ export class LibraryEngine extends Engine<undefined> {
...
+    if (TARGET_BUILD_TYPE === 'library') {
+      if (engineType === ClientEngineType.Wasm && config.adapter !== undefined) {
+        this.libraryLoader = libraryLoader ?? wasmLibraryLoader
+      } else {
+        this.libraryLoader = libraryLoader ?? defaultLibraryLoader
+      }
+    } else {
+      this.libraryLoader = libraryLoader ?? wasmLibraryLoader
```
millsp, line 95: (I am aware that we are conflating multiple concepts into one and that this is not the prettiest. On the one hand we have the `TARGET_BUILD_TYPE` which is injected by `esbuild` for tree-shaking. And on the other hand, we have the engine type, which is what was put into the schema. These two used to be parallel concepts, and over time we moved `DataProxyEngine` into `library`, and now we new condition for `engineType = wasm`).

millsp: This enables two things: It tree-shakes the default library loader out for edge builds, meaning you can only use wasm at the edge. Then the opposite, for library builds you can either use the default one or the wasm one as well.

packages/client/src/runtime/core/engines/common/Engine.ts
Jolg42:
```suggestion
   * Gets the raw wasm module for the query engine. This configurations is
   * generated specifically for each type of client, eg. Node.js client and Edge
   * clients will have different implementations.
   * Note: that is is a callback on purpose. We only load the wasm if we need it.
```

packages/client/src/runtime/core/engines/library/WasmLibraryLoader.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable @typescript-eslint/require-await */
+import '@prisma/driver-adapter-utils'
+
+// this import points directly to ./mywasmlib_bg.js it is generated with >>>
```
Jolg42, line 4:
```suggestion
// this import points directly to ./query_engine_bg.js it is generated with >>>
```

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -278,6 +278,13 @@ export type GetPrismaClientConfig = {
    * runtime, this means the client will be bound to be using the Data Proxy.
    */
   noEngine?: boolean
+
+  /**
+   * Gets the raw wasm module for the query engine. This configurations is
+   * generated specifically for each type of client, eg. Node.js client and Edge
+   * clients will have different implementations.
+   */
```
Jolg42, line 286:
```suggestion
  /**
   * Gets the raw wasm module for the query engine. This configurations is
   * generated specifically for each type of client, eg. Node.js client and Edge
   * clients will have different implementations.
   * Note: that is is a callback on purpose. We only load the wasm if we need it.
   */
```

helpers/compile/plugins/fill-plugin/fillers/weakref.ts
```diff
@@ -0,0 +1,11 @@
...
+  value: T
+
+  constructor(value: T) {
+    this.value = value
+  }
+
+  deref() {
+    return this.value
+  }
+}
```
millsp, line 11: This does not exist on cloudflare workers, it cannot be polyfilled very well, we just make it do nothing.

packages/client/helpers/build.ts
```diff
@@ -69,13 +72,19 @@ const edgeRuntimeBuildConfig: BuildOptions = {
         define: 'fn',
         inject: functionPolyfillPath,
       },
-
-      // TODO no tree shaking on wrapper pkgs
-      '@prisma/get-platform': { contents: '' },
```
millsp, line 74: Outdated

packages/client/helpers/build.ts
```diff
@@ -69,13 +72,19 @@ const edgeRuntimeBuildConfig: BuildOptions = {
...
+      // we shim WeakRef, it does not exist on CF
+      WeakRef: {
+        inject: weakrefPolyfillPath,
+      },
       // these can not be exported anymore
       './warnEnvConflicts': { contents: '' },
-      './utils/find': { contents: '' },
```
millsp, line 77: Oudated

packages/client/tests/e2e/bundler-detection-error/tests/main.ts
```diff
@@ -22,8 +22,24 @@ test('bundled prisma client will fail if generated client is gone', async () =>
   const { somePrismaCall } = require('../dist/index.js')
 
   await expect(somePrismaCall()).rejects.toThrowErrorMatchingInlineSnapshot(`
-"Prisma Client could not find its \`schema.prisma\`. This is likely caused by a bundling step, which leads to \`schema.prisma\` not being copied near the resulting bundle. We would appreciate if you could take the time to share some information with us.
-Please help us by answering a few questions: https://pris.ly/bundler-investigation-error"
```
millsp, line 26: Some context https://prisma-company.slack.com/archives/C04TW4A2H8C/p1698789183882339?thread_ts=1698787796.270229&cid=C04TW4A2H8C

helpers/compile/plugins/fill-plugin/fillers/weakref.ts
SevInf: We might want to introduce explicit `.destroy` method to the engine later then, if we can not use WeakRefs universally.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -81,36 +81,30 @@ export class LibraryEngine extends Engine<undefined> {
...
-      }
+    const engineType = getClientEngineType(config.generator!)
 
-      throw e
+    if (TARGET_BUILD_TYPE === 'library') {
+      // for "library" builds, we can use both the wasm and native engines
+      if (engineType === ClientEngineType.Wasm && config.adapter !== undefined) {
+        this.libraryLoader = libraryLoader ?? wasmLibraryLoader
+      } else {
+        this.libraryLoader = libraryLoader ?? defaultLibraryLoader
```
SevInf, line 94: Should we throw if `engineType` is WASM, but adapter is not set?

packages/internals/src/client/getClientEngineType.ts
```diff
@@ -3,28 +3,36 @@ import type { GeneratorConfig } from '@prisma/generator-helper'
...
 }
 
-function getEngineTypeFromEnvVar() {
+function getEngineTypeFromEnvVar(generatorConfig?: GeneratorConfig) {
   const engineType = process.env.PRISMA_CLIENT_ENGINE_TYPE
   if (engineType === ClientEngineType.Library) {
     return ClientEngineType.Library
   } else if (engineType === ClientEngineType.Binary) {
     return ClientEngineType.Binary
+  } else if (engineType === ClientEngineType.Wasm && generatorConfig?.previewFeatures.includes('driverAdapters')) {
```
SevInf, line 34: Similar to the previuous comment, I feel like engineType=wasm witout driver adapters enabled should cause explicit validation error, rather than ignoring env var.

packages/client/src/generation/utils/buildGetQueryEngineWasmModule.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  // by default just make it work on node.js by turning the file into a module
+  return `config.getQueryEngineWasmModule = () => {
+  const queryEngineWasmFilePath = require('path').join(config.dirname, 'query-engine.wasm')
+  const queryEngineWasmFileBytes = require('fs').readFileSync(queryEngineWasmFilePath)
+
+  return new WebAssembly.Module(queryEngineWasmFileBytes)
```
SevInf, line 25:
```suggestion
  return Promise.resolve().then(() => {
    const queryEngineWasmFilePath = require('path').join(config.dirname, 'query-engine.wasm')
    const queryEngineWasmFileBytes = require('fs').readFileSync(queryEngineWasmFilePath)

    return new WebAssembly.Module(queryEngineWasmFileBytes)
 })   
```

So that both variants return promise.

packages/client/src/runtime/core/engines/common/Engine.ts
```diff
@@ -147,6 +147,14 @@ export interface EngineConfig {
...
    */
   isBundled?: boolean
+
+  /**
+   * Loads the raw wasm module for the wasm query engine. This configuration is
+   * generated specifically for each type of client, eg. Node.js client and Edge
+   * clients will have different implementations.
+   * @remarks this is a callback on purpose, we only load the wasm if needed.
+   */
+  getQueryEngineWasmModule?: () => Promise<any>
```
SevInf, line 157:
```suggestion
  getQueryEngineWasmModule?: () => Promise<Library>
```

From `packages/client/src/runtime/core/engines/library/types/Library.ts`

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -81,36 +81,30 @@ export class LibraryEngine extends Engine<undefined> {
...
+      this.libraryLoader = libraryLoader ?? wasmLibraryLoader
     }
 
     this.config = config
     this.libraryStarted = false
     this.logQueries = config.logQueries ?? false
     this.logLevel = config.logLevel ?? 'error'
-    this.libraryLoader = loader
     this.logEmitter = config.logEmitter
+    this.datamodel = atob(config.inlineSchema)
```
SevInf, line 106: Not necessary something for now, but we should consider removing base64-encoding for inline schema here - where for DataProxy it could be argued that we are saving some time on encoding it (although, I'd say DP itself has no reason to require schema to be base64-encoded), in case of NodeAPI/WASM it does nothing but increase generated bundle size by 33% and require us to spend some time on decoding during startup.

SevInf (approved):
Added a couple of comments. Approving to not block it further, but I'd really like if added validation errors for invalid combinations of `engineType`, `adapter` option and preview features before merging, while this code is fresh. 

Rest can wait, I think.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
millsp: I agree. We went base64 because the binary engine gets started with the schema that way. I think we could introduce a new cli option to start it with the schema as is. Do you know if there was any reason for the base64?

janpio: Possibly relevant: How do we upload the schema to DataProxy/Accelerate?

millsp: https://prisma-company.slack.com/archives/C04TW4A2H8C/p1698863742436789?thread_ts=1698787796.270229&cid=C04TW4A2H8C

millsp: This is also now tracked in an issue

packages/client/src/runtime/core/engines/common/Engine.ts
millsp: It's not exactly working this way, at this stage it is just a raw wasm module, the instance is later created here https://github.com/prisma/prisma/blob/cfd3296895a41a091c46db08ac501be44cdec74a/packages/client/src/runtime/core/engines/library/WasmLibraryLoader.ts/#L14-L16

packages/client/src/generation/utils/buildGetQueryEngineWasmModule.ts
millsp: I opted to make the function async

PR #21385, opened 2023-10-06
test(client): metrics, avoid flakiness due to histograms

Jolg42:
Example https://github.com/prisma/prisma/actions/runs/6423277178/job/17441479169?pr=21363#step:7:477
```
 Summary of all failing tests
FAIL metrics/enabled/tests.ts (7.195 s)
  ● metrics.enabled (provider=sqlite) › before a query › SQL Providers: should have the same keys, before and after a query

    expect(received).toMatchObject(expected)

    - Expected  - 1
    + Received  + 1

    @@ -25,11 +25,11 @@
                50,
                Any<Number>,
              ],
              Array [
                100,
    -           0,
    +           1,
              ],
              Array [
                500,
                0,
              ],

      375 |           },
      376 |         ])
    > 377 |         expect(histogramsAfter.sort((a, b) => a.key.localeCompare(b.key))).toMatchObject([
          |                                                                            ^
      378 |           {
      379 |             key: 'prisma_client_queries_duration_histogram_ms',
      380 |             labels: {},

      at Object.<anonymous> (metrics/enabled/tests.ts:377:76)

```


janpio:
Doesn't this point to a specific request not being counted at all though?
(Otherwise it would be 1 less somewhere else, it it was just a different bucket)

(Of course not need to temporarily remove the flakyness, but we should have a bug issue then as well)

Jolg42:
@janpio Here I think it was counted under a faster bucket and there was already an expect any number for that one.
So technically we could improve the test and expect 1 request in one bucket (not matter which one)

Jolg42:
See https://github.com/prisma/team-orm/issues/432

PR #21387, opened 2023-10-06
chore: un-ignore `.husky/pre-commit`

aqrln:
Reverts https://github.com/prisma/prisma/pull/21312 and only ignores
non-husky hooks instead.

See https://github.com/prisma/prisma/pull/21312#issuecomment-1750233232.

### DIFF
--- a/.husky/.gitignore
+++ b/.husky/.gitignore
@@ -1,0 +1,3 @@
+*
+!.gitignore
+!pre-commit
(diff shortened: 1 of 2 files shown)


.husky/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+!.gitignore
```
Jolg42, line 2: This should not be needed 🤔 

It's weird 😄

Jolg42: But a side effect of the `*`?

Jolg42 (approved):

PR #21393, opened 2023-10-06
chore: use underscore notation for >= 1000

Jolg42:
Easier to read and won't have any other impact than that.


aqrln (approved):
Do we want to also enable some eslint rule to enforce this?

Jolg42:
If you can find it 😄 I don't think it's worth creating one.

PR #21437, opened 2023-10-10
chore: giving up on patching node-fetch

Jolg42:
The patch was only useful to avoid some logs locally in our dev env.

We can remove node-fetch completely once we remove Node 16


Jolg42:
That should unblock https://github.com/prisma/prisma/actions/runs/6468260487/job/17563361433#step:7:1

Issue #19124, opened 2023-05-07
Prisma generate randomly fails on Ubuntu due to missing internal .so `libquery_engine-debian-openssl-1.1.x.so.node`


jakeleventhal:
### Bug description

Here is my package.json in the root of my repo:
```json
{
  "devDependencies": {
    "@prisma/client": "4.13.0",
    "pnpm": "8.4.0",
    "prettier": "2.8.8",
    "prisma": "4.13.0",
    "prisma-json-types-generator": "2.3.1",
    "ts-node": "10.9.1",
    "typescript": "5.0.4",
    "turbo": "1.9.3",
    "vercel": "29.0.0"
  },
  "scripts": {
    "postinstall": "pnpm run -r --parallel prisma-generate"
  }
}
```
I have two database packages with `prisma-generate` scripts that look like so:
```json
"prisma-generate": "prisma generate --schema=./prisma/schema.prisma"
```

In my GitHub actions, when I run `pnpm install`, I randomly (and only sometimes) get the following error:
```
. postinstall$ pnpm run -r --parallel prisma-generate
. postinstall: Scope: 26 of 27 workspace projects
. postinstall: packages/ec-database prisma-generate$ prisma generate --schema=./prisma/schema.prisma
. postinstall: packages/pl-database prisma-generate$ prisma generate --schema=./prisma/schema.prisma
. postinstall: packages/ec-database prisma-generate: Prisma schema loaded from prisma/schema.prisma
. postinstall: packages/pl-database prisma-generate: Prisma schema loaded from prisma/schema.prisma
. postinstall: packages/pl-database prisma-generate: Error: ENOENT: no such file or directory, stat '/home/runner/work/Ecominate/Ecominate/node_modules/.pnpm/prisma@4.13.0/node_modules/prisma/libquery_engine-debian-openssl-1.1.x.so.node'
. postinstall: packages/pl-database prisma-generate: Failed
. postinstall: undefined
. postinstall: /home/runner/work/Ecominate/Ecominate/packages/pl-database:
. postinstall:  ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @ecominate/pl-database@1.0.0 prisma-generate: `prisma generate --schema=./prisma/schema.prisma`
. postinstall: Exit status 1
. postinstall: Failed
 ELIFECYCLE  Command failed with exit code 1.
```

### How to reproduce

See above. Can't consistently reproduce

### Expected behavior

No errors

### Prisma information

See above

### Environment & setup

[...]

PR #21439, opened 2023-10-10
fix(fetch-engine): exclude non-mac platforms from removing files

lukahartwig:
Related #19124

Running multiple instances of Prisma CLI in parallel can cause race conditions among the fs calls.

In this PR I added a platform check for an edge case on Mac that deletes files before overwriting them. Deleting files can cause concurrent Prisma CLI instances to fail randomly.

The platform check only fixes the issue for non-mac platforms.


Jolg42 (approved):
Thanks @lukahartwig for your first PR to Prisma 💚

janpio:
Note: https://github.com/prisma/prisma/issues/19124#issuecomment-1758682017

PR #21446, opened 2023-10-11
test: use a fork of ps-http-sim for security & stability

Jolg42:
Before, we were doing a git clone on a repo that we don't own and its content could change at anytime.

### DIFF
--- a/docker/planetscale_proxy/Dockerfile
+++ b/docker/planetscale_proxy/Dockerfile
@@ -4,7 +4,7 @@ RUN apt update && apt install netcat-openbsd -y
 RUN cd /go/src && git clone https://github.com/prisma/planetscale-proxy.git
 RUN cd /go/src/planetscale-proxy && go install .
 
-ENTRYPOINT /go/bin/planetscale-proxy \
+ENTRYPOINT /go/bin/ps-http-sim \
   -http-addr=0.0.0.0 \
   -http-port=8085 \
   -mysql-addr=$MYSQL_HOST \
--- a/docker/planetscale_proxy/Dockerfile
+++ b/docker/planetscale_proxy/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang:1
 
 RUN apt update && apt install netcat-openbsd -y
-RUN cd /go/src && git clone https://github.com/mattrobenolt/ps-http-sim.git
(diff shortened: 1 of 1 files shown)


docker/planetscale_proxy/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM golang:1
 
 RUN apt update && apt install netcat-openbsd -y
-RUN cd /go/src && git clone https://github.com/mattrobenolt/ps-http-sim.git
+RUN cd /go/src && git clone https://github.com/prisma/ps-http-sim.git
 RUN cd /go/src/ps-http-sim && go install .
 
 ENTRYPOINT /go/bin/ps-http-sim \
```
miguelff, line 7:
```suggestion
RUN cd /go/src && git clone https://github.com/prisma/planetscale-proxy.git
RUN cd /go/src/planetscale-proxy && go install .

ENTRYPOINT /go/bin/planetscale-proxy \
```

miguelff (approved):
Can you please 🙏🏾 use https://github.com/prisma/planetscale-proxy fork instead? I think the name is more descriptive. I provided a patch in the review.

Jolg42:
Somehow it fails with the clone @miguelff, any idea? Do you see the same on your side?
```
container prisma-prisma-planetscale_proxy-1 is unhealthy
```
https://github.com/prisma/prisma/actions/runs/6481411961/job/17598870099?pr=21446#step:4:687

janpio (approved):

docker/planetscale_proxy/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM golang:1
 
 RUN apt update && apt install netcat-openbsd -y
-RUN cd /go/src && git clone https://github.com/mattrobenolt/ps-http-sim.git
-RUN cd /go/src/ps-http-sim && go install .
+RUN cd /go/src && git clone https://github.com/prisma/planetscale-proxy.git
+RUN cd /go/src/planetscale-proxy && go install .
 
 ENTRYPOINT /go/bin/ps-http-sim \
```
janpio, line 7: might be worth a comment that the name comes from the codebase above (which now has no more connection in the name itself)

PR #21450, opened 2023-10-11
feat(DataProxyEngine): Send Engine Hash as Header for requests

do4gr:
[Internal Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1697040018592629)

This adds the engine hash as a header to all requests to the QE when using the `DataProxyEngine`

Closes https://linear.app/prisma-company/issue/FT-1757/have-the-client-send-the-engine-hash-during-schema-upload

### DIFF
--- a/packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
+++ b/packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
@@ -79,14 +79,14 @@ class DataProxyHeaderBuilder {
   readonly tracingHelper: TracingHelper
   readonly logLevel: EngineConfig['logLevel']
   readonly logQueries: boolean | undefined
-  readonly engineHash : string
+  readonly engineHash: string
 
   constructor({
     apiKey,
     tracingHelper,
     logLevel,
     logQueries,
-    engineHash
+    engineHash,
   }: {
     apiKey: string
     tracingHelper: TracingHelper
@@ -165,11 +165,11 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
 
     this.config = config
     this.env = { ...this.config.env, ...process.env }
-    this.inlineSchema = config.inlineSchema ?? ''
-    this.inlineDatasources = config.inlineDatasources ?? {}
-    this.inlineSchemaHash = config.inlineSchemaHash ?? ''
-    this.clientVersion = config.clientVersion ?? 'unknown'
-    this.engineHash = config.engineVersion ?? 'unknown'
+    this.inlineSchema = config.inlineSchema
+    this.inlineDatasources = config.inlineDatasources
+    this.inlineSchemaHash = config.inlineSchemaHash
+    this.clientVersion = config.clientVersion
+    this.engineHash = config.engineVersion
     this.logEmitter = config.logEmitter
     this.tracingHelper = this.config.tracingHelper
   }
@@ -178,8 +178,9 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
     return this.headerBuilder.apiKey
   }
 
+  // The version is the engine hash
+  // that we expect to have on the remote QE
   version() {
-    // QE is remote, we don't need to know the exact commit SHA
     return this.engineHash
   }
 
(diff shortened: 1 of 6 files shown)


packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -66,6 +66,7 @@ type DataProxyHeaders = {
   Authorization: string
   'X-capture-telemetry'?: string
   traceparent?: string
+  'Prisma-Engine-Hash': string
```
janpio, line 69:
```suggestion
  'X-Prisma-Engine-Hash': string
```

Because:

> Some web applications use custom headers to add comments or annotations to an HTTP message. The convention is to prefix the header name with X- to indicate that it is non-standard

http://www.httpwatch.com/HTTPgallery/headers/#:~:text=Some%20web%20applications%20use%20custom,that%20it%20is%20non%2Dstandard.

Jolg42: >On June 2012, the deprecation of recommendation to use the "X-" prefix has become official as [RFC 6648](https://www.rfc-editor.org/rfc/rfc6648).
https://www.rfc-editor.org/rfc/rfc6648

`X-` is to be avoided since 2012, it was a recommendation before.

Jolg42: But we could keep it for consistency if we have others `X-` headers I guess...

packages/client/src/runtime/core/engines/data-proxy/DataProxyEngine.ts
```diff
@@ -158,10 +165,11 @@ export class DataProxyEngine extends Engine<DataProxyTxInfoPayload> {
...
     this.config = config
     this.env = { ...this.config.env, ...process.env }
-    this.inlineSchema = config.inlineSchema ?? ''
-    this.inlineDatasources = config.inlineDatasources ?? {}
-    this.inlineSchemaHash = config.inlineSchemaHash ?? ''
-    this.clientVersion = config.clientVersion ?? 'unknown'
+    this.inlineSchema = config.inlineSchema
+    this.inlineDatasources = config.inlineDatasources
+    this.inlineSchemaHash = config.inlineSchemaHash
+    this.clientVersion = config.clientVersion
```
Jolg42, line 171: They used to be optional before, but not anymore (example: https://github.com/prisma/prisma/commit/49a66fa70742f70c9fed376473082485ad02d497#diff-84f89062eba89dc0d9ef5a8d43ee3ae512a780a4da288e57b6b937763748e2a9R66-R71)

packages/client/tests/functional/dataproxy-engine/version/tests.ts
```diff
@@ -0,0 +1,21 @@
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(() => {
```
SevInf, line 7: I am not 100% sure we should assert on private properties here. Test below (the one that uses the extension) already verifies that correct header is sent through more public means. Do you think we still need this one?

packages/client/tests/functional/dotenv-loading/_matrix.ts
```diff
@@ -0,0 +1,4 @@
+import { defineMatrix } from '../_utils/defineMatrix'
```
SevInf, line 1: Does not seem to be related to the rest of this PR?

SevInf (approved):
After addressing the comments here - LGTM. Could also be done separately if we rather merge this as quickly as possible.

packages/client/tests/functional/dotenv-loading/_matrix.ts
Jolg42: Oopsie I need to remove that

packages/client/tests/functional/dataproxy-engine/version/tests.ts
Jolg42: It's also testing `version()` which is not covered by the other test. though `version()` is actually only used here.

So I think it's somewhat interesting to have this, though I agree testing private properties is not great it's maybe better than nothing.
Here it also shows how `remoteClientVersion` is undefined before calling `$connect()`.

Jolg42 (approved):

Jolg42:
It's released in `5.5.0-dev.54` and up.

PR #21577, opened 2023-10-21
fix(client): normalize outputDir

### LINKED ISSUE #21576: Prisma generation bug with pnpm on Windows if it's on postinstall.
### Bug description

Prisma generate is generating client to the wrong location if using pnpm and `prisma generate` is executed from postinstall on Windows.

This bug is caused by the getDefaultOutDir: https://github.com/prisma/prisma/blob/23d5ef0672372035a84552b6b457197ca19f486d/packages/client/src/generation/generateClient.ts#L160-L183

There are 2 if statements. I am not sure why does postinstall was treaten differently. Even on Linux, if `prisma generate` is executed from postinstall script, it would return from the first if statement.

Here's where the problem lies: The first if statement did not normalize the path. Windows uses `\`, not `/`. Therefore the result was, `prisma generate` executed by postinstall script on Windows would return on the second if statement, thus making the output `node_modules/.prisma`

Solution: use path.normalize like in #21577 

### How to reproduce

1. Use pnpm on Windows
2. Init Prisma and add some model
3. Try run `$ pnpm exec prisma generate`.
4. Prisma generated client should be generated at `node_modules/.pnpm/@prisma+client@xxx/node_modules/.prisma`
5. But now try adding `"postinstall": "prisma generate"` to package.json
6. Delete node_modules folder then reinstall using `pnpm install`
7. Wait for postinstall to be finished
8. Prisma client is generated to `node_modules/.prisma` instead of `node_modules/.pnpm/@prisma+client@xxx/node_modules/.prisma`
9. Broken Prisma client on pnpm by postinstall

### Expected behavior

When using pnpm, it should output to `node_modules/.pnpm/@prisma+client@xxx/node_modules/.prisma`, even on postinstall

### Prisma information

```prisma
// Example schema

generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
}

model User {
  email String  @unique
  name  String?
}
```

### Environment & setup

- OS: Windows 11 22621.2428
- Database: N/A (This issue only affects prisma client generation, not prisma itself)
- Node.js version: v18.18.2

### Prisma Version

```
prisma                  : 5.4.2
@prisma/client          : 5.4.2
Query Engine (Node-API) : libquery-engine ac9d7041ed77bcc8a8dbd2ab6616b39013829574 (at node_modules\.pnpm\@prisma+engines@5.4.2\node_modules\@prisma\engines\query_engine-windows.dll.node)
Schema Engine           : schema-engine-cli ac9d7041ed77bcc8a8dbd2ab6616b39013829574 (at node_modules\.pnpm\@prisma+engines@5.4.2\node_modules\@prisma\engines\schema-engine-windows.exe)
Schema Wasm             : @prisma/prisma-schema-wasm 5.4.1-2.ac9d7041ed77bcc8a8dbd2ab6616b39013829574
Default Engines Hash    : ac9d7041ed77bcc8a8dbd2ab6616b39013829574
```

Hazmi35:
Fixes https://github.com/prisma/prisma/issues/21576

There could be other code around comparing paths that needs to be normalized too, but I'm not sure if there are any. Haven't searched for it

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -648,8 +648,9 @@ async function getGenerationDirs({
   testMode,
 }: GenerateClientOptions) {
   const isCustomOutput = generator.isCustomOutput === true
+  const normalizedOutputDir = path.normalize(outputDir)
   let userRuntimeImport = isCustomOutput ? './runtime' : '@prisma/client/runtime'
-  let userOutputDir = isCustomOutput ? outputDir : await getDefaultOutdir(path.normalize(outputDir))
+  let userOutputDir = isCustomOutput ? normalizedOutputDir : await getDefaultOutdir(normalizedOutputDir)
 
   if (testMode && runtimeBase) {
     userOutputDir = outputDir
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -361,7 +361,7 @@ function getTypedSqlRuntimeBase(runtimeBase: string) {
 
 // TODO: explore why we have a special case for excluding pnpm
 async function getDefaultOutdir(outputDir: string): Promise<string> {
-  if (outputDir.endsWith('node_modules/@prisma/client')) {
+  if (outputDir.endsWith(path.normalize('node_modules/@prisma/client'))) {
     return path.join(outputDir, '../../.prisma/client')
   }
   if (
@@ -649,7 +649,7 @@ async function getGenerationDirs({
 }: GenerateClientOptions) {
   const isCustomOutput = generator.isCustomOutput === true
   let userRuntimeImport = isCustomOutput ? './runtime' : '@prisma/client/runtime'
-  let userOutputDir = isCustomOutput ? outputDir : await getDefaultOutdir(outputDir)
+  let userOutputDir = isCustomOutput ? outputDir : await getDefaultOutdir(path.normalize(outputDir))
 
   if (testMode && runtimeBase) {
     userOutputDir = outputDir


Hazmi35:
What's blocking this PR? Tests are passing, the CI / Tests / All pkgs (win+mac) (windows-latest) failed because of timeout, can someone rerun the workflow?

Hazmi35:
> The PR did not recieve a review yet from us. The issue it is fixing was not reproduced by anyone of the team, so we don't even have an understanding if something is really broken.
> 
> Additionally, as the bug existed and no tests were failing, it is possible that this change actually introduces more problems that would also not be covered by tests - so of course we prefer to be careful.
> 
> Do you have an idea how we can get more confidence in the change?
> 
> (I reran the failing test)

Fair, but the problem is this is a simple PR, which only patch the `prisma generate` command.

Please see the this image:
image

As you can see the outputDir2 on Windows uses back slash `\`. But it is being compared with forward slash `/`.
`F:\test\node_modules\.pnpm\@prisma+client@5.5.2_prisma@5.5.2\node_modules\@prisma\client` is obviously not ends with `node_modules/@prisma/client`.

Now what happened if it's not true? Then it will go to this if statement:
https://github.com/prisma/prisma/blob/23d5ef0672372035a84552b6b457197ca19f486d/packages/client/src/generation/generateClient.ts#L165-L180

As you can see it will forcefully generated `.prisma/client` in `node_modules`, which if you're using pnpm, it won't work.

The solution is, always normalize path using path.normalize before comparing them.

janpio:
Thanks for the additional information, that will be helpful when an Engineer reviews the PR.

Hazmi35:
What's really blocking this PR to be reviewed?

Hazmi35:
`prisma generate` invoked via postinstall is still broken on Windows, rebased.

aqrln (commented):
Hey @Hazmi35, sorry for such a long delay.

Even without looking at the linked issue and regardless of whether it was reproducible, this change actually makes total sense to me by itself: `outputDir.endsWith('node_modules/@prisma/client')` is not going to be true on Windows if `outputDir` contains native backslashes as path separators.

It looks like I can't update the branch (most likely because you opened a PR from the `main` branch of your fork) — would you mind rebasing on the latest `main`? The changes look good to me if the CI is green after the rebase.

Hazmi35:
> Hey @Hazmi35, sorry for such a long delay.
> 
> Even without looking at the linked issue and regardless of whether it was reproducible, this change actually makes total sense to me by itself: `outputDir.endsWith('node_modules/@prisma/client')` is not going to be true on Windows if `outputDir` contains native backslashes as path separators.
> 
> It looks like I can't update the branch (most likely because you opened a PR from the `main` branch of your fork) — would you mind rebasing on the latest `main`? The changes look good to me if the CI is green after the rebase.

Hello.

I've rebased my branch from the latest main branch.

aqrln (approved):
LGTM, thank you!

Issue #21565, opened 2023-10-19
Prisma installation failure


sharonpatta:
### Bug description

I have node version 18.18.2. However, when I try to run `npm install`, it fails saying that I need node version >= 16.13. I have attached the log.
[2023-10-19T21_10_10_343Z-debug-0.log](https://github.com/prisma/prisma/files/13047824/2023-10-19T21_10_10_343Z-debug-0.log)

### How to reproduce
1. Check node version using `node -v` and verify that it is >= 16.13
2. Run `npm i` or `npm install prisma@5.4.2`

### Environment & setup

- OS: MacOS
- Database: PostgreSQL
- Node.js version: 18.18.2

chrisvire:
@sharonpatta is a student working on a [senior capstone project](https://github.com/sillsdev/appbuilder-usercontent-management).  I started working with them yesterday getting the project setup with SvelteKit and Prisma.

This is only happening on her computer, but it is very odd that the log clearly shows that version of node as 18.18.2:

```
404 verbose stack Error: command failed
404 verbose stack     at ChildProcess.<anonymous> (/Users/sharonpatta1/.volta/tools/image/node/18.18.2/lib/node_modules/npm/node_modules/@npmcli/promise-spawn/lib/index.js:53:27)
404 verbose stack     at ChildProcess.emit (node:events:517:28)
404 verbose stack     at maybeClose (node:internal/child_process:1098:16)
404 verbose stack     at Socket.<anonymous> (node:internal/child_process:450:11)
404 verbose stack     at Socket.emit (node:events:517:28)
404 verbose stack     at Pipe.<anonymous> (node:net:350:12)
405 verbose pkgid prisma@5.4.2
406 verbose cwd /Users/sharonpatta1/cps491/appbuilder-usercontent-management
407 verbose Darwin 21.6.0
408 verbose node v18.18.2
409 verbose npm  v9.8.1
410 error code 1
411 error path /Users/sharonpatta1/cps491/appbuilder-usercontent-management/node_modules/prisma
412 error command failed
413 error command sh -c node scripts/preinstall-entry.js
414 error ┌──────────────────────────────────────────────┐
414 error │    Prisma only supports Node.js >= 16.13.    │
414 error │    Please upgrade your Node.js version.      │
414 error └──────────────────────────────────────────────┘
415 verbose exit 1
```

chrisvire:
FYI, we tried installing node via nvm and https://volta.sh and both had the same failure.

[...]

PR #21587, opened 2023-10-23
chore(tech-debt): specify minimum support Node.js version only once in preinstall.js

jkomyno:
Small tech-debt change that came up while replying to https://github.com/prisma/prisma/issues/21565.
Basically, you can now:
- avoid removing the `v` version prefix, by replacing `process.version` with `process.versions.node`
- avoid specifying the minimum support Node.js major and minor versions in multiple places

Note: there's no urgency in reviewing this PR. I'm just posting it here so that I don't forget it in my git stash.


packages/cli/scripts/preinstall.js
```diff
@@ -1,21 +1,33 @@
...
+    .slice(0, 2) // only major and minor version
+    .map((v) => parseInt(v, 10))
 }
 
 export function printMessageAndExitIfUnsupportedNodeVersion(nodeVersion) {
-  const nodeVersionAsParts = nodeVersion.split('.')
-  // `.slice(1)` removes `v` from `v16`
-  const nodeMajorVersion = parseInt(nodeVersionAsParts[0].slice(1))
-  const nodeMinorVersion = parseInt(nodeVersionAsParts[1])
+  const MIN_NODE_VERSION = '16.13'
```
jkomyno, line 20: Optionally:
```suggestion
  const MIN_NODE_VERSION = '16.13.0'
```

or 
```suggestion
  const MIN_NODE_VERSION = '16.13.x'
```

but this would require changing the test snapshots as well.

packages/cli/scripts/preinstall.js
```diff
@@ -1,21 +1,33 @@
+const process = require('node:process')
 // @ts-check
 const { drawBox } = require('@prisma/internals/dist/utils/drawBox')
 
 export function main() {
-  // process.version (e.g. `v16.0.0`)
-  printMessageAndExitIfUnsupportedNodeVersion(process.version)
+  // Node.js version, without the `v` prefix (e.g. `16.13.0`)
+  const nodeVersion = process.versions.node
```
Jolg42, line 7: We can keep using `process.version`, it's going to be the safest one to use and it doesn't bring value of changing this.
```suggestion
  const nodeVersion = process.versions
```

Jolg42:
@jkomyno Is this ready for a review?

packages/cli/scripts/preinstall.js
jkomyno: `process.versions` exists since Node.js 9.0.0, and arguably `process.versions.node` is more ergonomic and explicit than `process.version.slice(1)`, but discussing about this nit is of little value imho, so I don't have a strong opinion.

jkomyno: I've removed `process.versions.node` in favor of `process.version.slice(1)`.

packages/cli/scripts/preinstall.js
```diff
@@ -1,21 +1,40 @@
+const process = require('node:process')
 // @ts-check
 const { drawBox } = require('@prisma/internals/dist/utils/drawBox')
 
 export function main() {
-  // process.version (e.g. `v16.0.0`)
-  printMessageAndExitIfUnsupportedNodeVersion(process.version)
+  const nodeVersion = process.version.slice(1)
```
Jolg42, line 6:
```suggestion
  // process.version (e.g. `v20.0.0`) starts with `v`, we remove it.
  const nodeVersion = process.version.slice(1)
```
Keep the example it's nice to have it here

packages/cli/scripts/preinstall.js
```diff
@@ -1,21 +1,40 @@
...
+
+  printMessageAndExitIfUnsupportedNodeVersion(nodeVersion)
```
Jolg42, line 8:
```suggestion
  printMessageAndExitIfUnsupportedNodeVersion(process.version)
```
Pass the version directly and avoid slicing the first character twice

packages/cli/scripts/preinstall.js
Jolg42: We document `16.13` in
https://www.prisma.io/docs/reference/system-requirements

It's ok without the patch version I think.

packages/cli/scripts/preinstall.js
jkomyno: Thanks, this was a leftover

Jolg42 (approved):

Jolg42:
It looks like there is a merge conflict since a file was renamed recently

Issue #1961, opened 2020-03-27
Improve version output


schickling:
## Problem

Currently the version CLI output (when running `prisma2 -v`) is quite "overwhelming".

![image](https://user-images.githubusercontent.com/1567498/77776900-01042e80-704f-11ea-964b-78ac917c3b9f.png)

## Solution

1. I suggest to reformat the output by "segmenting" the output into two blocks. e.g.

```
@prisma/cli: 2.0.0-preview025

Current platform:     darwin
Query Engine:         prisma 8814060fa684793b73d07dbfccd4b7777b3361ae (at /usr/local/lib/node_modules/@prisma/cli/query-engine-darwin)
Migration Engine:     migration-engine-cli 8814060fa684793b73d07dbfccd4b7777b3361ae (at /usr/local/lib/node_modules/@prisma/cli/migration-engine-darwin)
Introspection Engine: introspection-core 8814060fa684793b73d07dbfccd4b7777b3361ae (at /usr/local/lib/node_modules/@prisma/cli/introspection-engine-darwin)
```

2. Additionally I'd print the "key" of the key/value-pairs (e.g. `@prisma/cli`) in **bold**.

## Additional context

- Related issue: https://github.com/prisma/prisma2/issues/981

ChristophDietrich:
What about to do it like node:
`node -v`

Output:
`v12.16.1`

----
Prisma2 (short version)
`prisma2 -v`

Output:
`@prisma/cli: 2.0.0-preview025`

----
Prisma2 (long version)
`prisma2 -v -full`

Output:
```
@prisma/cli: 2.0.0-preview025

Current platform:     darwin
Query Engine:         prisma 8814060fa684793b73d07dbfccd4b7777b3361ae (at /usr/local/lib/node_modules/@prisma/cli/query-engine-darwin)
Migration Engine:     migration-engine-cli 8814060fa684793b73d07dbfccd4b7777b3361ae (at /usr/local/lib/node_modules/@prisma/cli/migration-engine-darwin)
Introspection Engine: introspection-core 8814060fa684793b73d07dbfccd4b7777b3361ae (at /usr/local/lib/node_modules/@prisma/cli/introspection-engine-darwin)
```

janpio:
👍 on splitting the output.

> 1. Additionally I'd print the "key" of the key/value-pairs (e.g. `@prisma/cli`) in **bold**.

[...]

PR #21625, opened 2023-10-26
feat(version): add more information in version command output

Jolg42:
Related to https://github.com/prisma/prisma/issues/1961

Example output on my local computer
```
prisma                  : 0.0.0
@prisma/client          : 0.0.0
Node.js                 : v18.16.0
Query Engine (Node-API) : libquery-engine f365956fa36e50f1c89d8ffe3997d512ab2d6fec (at ../engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli f365956fa36e50f1c89d8ffe3997d512ab2d6fec (at ../engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.6.0-2.f365956fa36e50f1c89d8ffe3997d512ab2d6fec
Default Engines Hash    : f365956fa36e50f1c89d8ffe3997d512ab2d6fec
```

### DIFF
--- a/packages/cli/src/Version.ts
+++ b/packages/cli/src/Version.ts
@@ -17,6 +17,7 @@ import {
 } from '@prisma/internals'
 import { bold, dim, red } from 'kleur/colors'
 import { match, P } from 'ts-pattern'
+import os from 'os'
 
 import { getInstalledPrismaClientVersion } from './utils/getClientVersion'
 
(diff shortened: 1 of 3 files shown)


Jolg42:
Not ready to be merged, but review/feedback is welcome.

packages/cli/src/Version.ts
```diff
@@ -91,7 +92,10 @@ export class Version implements Command {
     const rows = [
       [packageJson.name, packageJson.version],
       ['@prisma/client', prismaClientVersion ?? 'Not found'],
-      ['Current platform', platform],
+      ['Current platform', os.platform()],
```
janpio, line 95: Changing the meaning of this is a bit "meh", but I think I am ok with it.

Maybe "Current OS" instead to avoid the repurposing?

(And then maybe start with "Computed binaryTarget", then OS and architecture)

Jolg42: Sounds good.

I changed the wording to 
```
      ['Computed binaryTarget', platform],
      ['Operating System', os.platform()],
      ['Architecture', os.arch()],
```

Jolg42: `Current` was not needed

janpio (approved):
Nice.

Optimally only to be merged with some tests that run on different platforms and create snapshots for these so we know all this works as expected.

millsp (approved):
Nice! Do you think it would be useful to also display the provider?

PR #21632, opened 2023-10-26
chore(cli): update message should be a warning not an error

gilles-yvetot:
Most of the observability tools will trigger errors in production when an upgrade is available. It should be considered as a warning and not an error

### DIFF
--- a/packages/cli/src/utils/printUpdateMessage.ts
+++ b/packages/cli/src/utils/printUpdateMessage.ts
@@ -40,7 +40,7 @@ export function printUpdateMessage(checkResult: { status: 'ok'; data: Check.Resp
     horizontalPadding: 2,
   })
 
-  console.error(boxedMessage)
+  console.warn(boxedMessage)
 }
 
 function makeInstallCommand(
--- a/packages/cli/src/__tests__/printUpdateMessage.test.ts
+++ b/packages/cli/src/__tests__/printUpdateMessage.test.ts
@@ -24,15 +24,15 @@ function printUpdateMessageFromTo(from: string, to: string): void {
   })
 }
 
-const consoleErrorMock = jest.spyOn(console, 'error').mockImplementation()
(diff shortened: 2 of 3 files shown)


packages/cli/src/utils/printUpdateMessage.ts
```diff
@@ -40,7 +40,7 @@ export function printUpdateMessage(checkResult: { status: 'ok'; data: Check.Resp
     horizontalPadding: 2,
   })
 
-  console.error(boxedMessage)
```
millsp, line 43: Could you add a test for this?

gilles-yvetot:
Hey @millsp, sorry I didn't expect that to be tested to be honest haha. Kudos to you! I have amended my previous commit to reflect that

Jolg42:
@gilles-yvetot good point! Though I don't see the difference, between `console.error` and `console.warn`, both will be directed to `stderr`. 
I think that to avoid problems with observability tools looking at logs, it needs to be sent to `stdout`, right?

Jolg42:
See https://nodejs.org/api/console.html#consolewarndata-args
>The console.warn() function is an alias for [console.error()](https://nodejs.org/api/console.html#consoleerrordata-args).

gilles-yvetot:
@Jolg42 oh! very interesting. Well then I can switch to a `console.log` or `console.info` or `console.debug`? Will you be okay with that?

gilles-yvetot:
No response

aqrln:
On which level does the observability tool you're facing this issue with work and how it detects the error messages? I'd assume that it can't be that any logs sent to stderr are a problem because that's pretty much where such messages are supposed to be sent (otherwise it would be problematic to use Prisma, e.g., in a CLI script that prints consumable data to stdout). If the tool specifically detects the calls to `console.error` on JS side by injecting code there, then I'd say these changes are good as they are and there's no need to change `console.warn` to anything else.

PR #21642, opened 2023-10-27
chore: adapter porting nits

millsp:
* Fixes integration tests
* Removes `tsup` in favor of consistency (we can have a discussion about `tsup`)
* Removes unnecessary `.gitignore`
* Removes extra test in `adapter-libsql`
* ~~Changes `debug` for `@prisma/debug` in adapters~~

/integration


Jolg42 (approved):

packages/adapter-planetscale/src/planetscale.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/require-await */
 import type planetScale from '@planetscale/database'
+import { Debug } from '@prisma/debug'
```
janpio, line 3: I see you decided to import `Debug` directly from `@prisma/debug`.
A benefit of the previous solution via `@prisma/driver-adapter-utils` was that `Debug` would also be available for any third party that develops a driver adapter and only has the utils as a dependency.

Worth importing via utils?

janpio:
I see you decided to import Debug directly from @prisma/debug.
A benefit of the previous solution via @prisma/driver-adapter-utils was that Debug would also be available for any third party that develops a driver adapter and only has the utils as a dependency.

Worth importing via utils?

millsp:
When I did this, I was thinking that it would be nice to use our own `debug` package for the consolidated crash reports with the last logs. Now thinking in retrospective, I am not even sure these would be triggered via a wasm engine. I'll restore it as it was.

jkomyno (approved):
Nice! Please make sure this PR doesn't possibly interfere with the fixes in https://github.com/prisma/prisma-engines/pull/4398 before merging. I can also do that if needed.

jkomyno:
In order to fix releases as well, we probably need to add `repository.url` fields to the newly added `*adapter*` packages (see [CI/CD failure](https://github.com/prisma/prisma/actions/runs/6691046387/job/18177517840#step:5:472)).

janpio:
@Jolg42 might know more about the provenance stuff.

Jolg42:
Right, that should be easy to fix.

Jolg42:
Here it is: https://github.com/prisma/prisma/pull/21683

janpio (changes requested):
Decision and potential change on this left:
> When I did this, I was thinking that it would be nice to use our own debug package for the consolidated crash reports with the last logs. Now thinking in retrospective, I am not even sure these would be triggered via a wasm engine. I'll restore it as it was.

PR #21653, opened 2023-10-27
ci(publish): fix getEnginesCommitHash

Jolg42:
Thanks @janpio for noticing that in https://github.com/prisma/prisma/actions/runs/6668971879/job/18125730860

### DIFF
--- a/scripts/ci/publish.ts
+++ b/scripts/ci/publish.ts
@@ -1,3 +1,4 @@
+import { devDependencies as devDependenciesPrismaEnginesPkg } from '@prisma/engines/package.json'
 import slugify from '@sindresorhus/slugify'
 import { IncomingWebhook } from '@slack/webhook'
 import arg from 'arg'
@@ -612,7 +613,7 @@ Check them out at https://github.com/prisma/ecosystem-tests/actions?query=workfl
 
       await publishPackages(packages, publishOrder, dryRun, prismaVersion, tag, args['--release'])
 
-      const enginesCommitHash = await getEnginesCommitHash()
+      const enginesCommitHash = getEnginesCommitHash()
       const enginesCommitInfo = await getCommitInfo('prisma-engines', enginesCommitHash)
       const prismaCommitHash = await getLatestCommitHash('.')
       const prismaCommitInfo = await getCommitInfo('prisma', prismaCommitHash)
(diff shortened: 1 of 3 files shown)


scripts/ci/publish.ts
```diff
@@ -657,9 +657,7 @@ Check them out at https://github.com/prisma/ecosystem-tests/actions?query=workfl
 async function getEnginesCommitHash(): Promise<string> {
   const prismaPath = path.resolve(process.cwd(), './packages/engines/package.json')
   const pkg = JSON.parse(await fs.promises.readFile(prismaPath, 'utf-8'))
-  // const engineVersion = pkg.prisma.version
```
janpio, line 660: Any idea why this was commented out and not used in the first place?

janpio: Ah, your suggestion actually goes into the `@prisma/engines-version` package.json and then gets `prisma.enginesVersion` there?

scripts/ci/publish.ts
```diff
@@ -657,9 +657,7 @@ Check them out at https://github.com/prisma/ecosystem-tests/actions?query=workfl
 async function getEnginesCommitHash(): Promise<string> {
   const prismaPath = path.resolve(process.cwd(), './packages/engines/package.json')
   const pkg = JSON.parse(await fs.promises.readFile(prismaPath, 'utf-8'))
```
millsp, line 659: If we use `import`, we can benefit from type-safety on the json properties

scripts/ci/publish.ts
```diff
@@ -657,9 +657,7 @@ Check them out at https://github.com/prisma/ecosystem-tests/actions?query=workfl
...
-  // const engineVersion = pkg.prisma.version
-  const engineVersion = pkg.devDependencies['@prisma/engines-version']?.split('.').slice(-1)[0]
-
+  const engineVersion = pkg.devDependencies['@prisma/engines-version'].prisma.enginesVersion
```
millsp, line 660: I don't think this property exists in `./packages/engines/package.json`. I'd suggest using a direct import from `import { prisma } from "@prisma/engines-version"` as suggested [here](https://github.com/prisma/prisma/pull/21653/files#r1374771640) or some similar approach.

scripts/ci/publish.ts
Jolg42: Yes, that's the best way to do this. We put the hash in the package.json

Jolg42: The comment is about a different hash, not the one we are interested here

millsp (approved):

PR #21657, opened 2023-10-27
refactor(ArgumentsRenderingTree): split buildInputValue function into smaller functions for better maintainability

AikoRamalho:
The first thing that stood out to me was that the code was doing a lot of type checking and then creating different types of objects based on those checks. This is a classic use case for the Factory pattern. The Factory pattern is a creational design pattern that provides an interface for creating objects in a superclass, but allows subclasses to alter the type of objects that will be created.

Another principle I kept in mind while refactoring was the Single Responsibility Principle. This principle states that a class should have only one reason to change. It should do one thing and do it well. So, I created three separate factory classes - ValueFactory, ArrayValueFactory, and ObjectValueFactory. Each of these factories are responsible for creating a specific type of object. The ValueFactory creates a value of the appropriate type based on the input, ArrayValueFactory creates an ArrayValue object, and ObjectValueFactory creates an ObjectValue object.

Refactoring the code in this way makes it more readable and maintainable. It's easier to understand what each part of the code is doing because each class has a clear, single responsibility. If you need to change how a certain type of object is created, you know exactly where to go to make that change.

### DIFF
--- a/packages/client/src/runtime/core/errorRendering/ArgumentsRenderingTree.ts
+++ b/packages/client/src/runtime/core/errorRendering/ArgumentsRenderingTree.ts
@@ -41,104 +41,85 @@ export class ArgumentsRenderingTree implements ErrorBasicBuilder {
  * @param args
  * @returns
  */
+class ValueFactory {
+  public static createValue(value: unknown) {
+    if (typeof value === 'string') {
+      return new ScalarValue(JSON.stringify(value))
+    }
+
+    if (typeof value === 'number' || typeof value === 'boolean') {
+      return new ScalarValue(String(value))
+    }
+
+    if (typeof value === 'bigint') {
+      return new ScalarValue(`${value}n`)
+    }
+
+    if (value === null) {
+      return new ScalarValue('null')
+    }
+
+    if (value === undefined) {
+      return new ScalarValue('undefined')
+    }
+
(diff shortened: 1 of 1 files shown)


SevInf:
Hi @AikoRamalho.
Thank you for submitting your contribution! 
I am sorry, but we've decided we will not merge it. Whether or not functional style (3 pure functions) this code is originally written in is inferior to OOP-style (3 factory classes) this PR introduces is up to debate and big part of it comes down to a personal preference. A lot of classic GoF design patterns were designed for working around Smalltalk and pre-ISO C++ weakness and might not be that relevant for different paradigms. If you are interested in this discussion, [this Stackoverflow thread](https://stackoverflow.com/questions/327955/does-functional-programming-replace-gof-design-patterns) will provide a nice starting point.

Lastly, if you are interested in contributing to Prisma, I'd suggest to pick up some small improvement or a bug fix, rather than refactoring - it is much less likely that those kinds of tasks will get merged afterward. Since you are already looking into error rendering code, [one of the existing issues on the topic](https://github.com/prisma/prisma/issues?q=is%3Aopen+is%3Aissue+label%3A%22topic%3A+error+handling%22+label%3Akind%2Fimprovement) might of interest. Let me know if you need any guidance.

PR #21662, opened 2023-10-28
tmp: Fix qe tests

miguelff:


janpio:
Pointing PR to https://github.com/prisma/prisma/pull/21642 instead as that already fixes some test failures of driver-adapters in this repo.

miguelff:
This is not needed anymore. engines can rely on @millsp changes in https://github.com/prisma/prisma/pull/21642

PR #21677, opened 2023-10-30
chore: remove node-fetch from publish and cleanup

Jolg42:
/integration

Note: Separately from this PR I switched the release action from v16 to v18 in https://github.com/prisma/prisma/commit/fdbedf325f614868aedf164f79c2ca78b89db09e
and later to v20 because of fetch types.


helpers/compile/plugins/fill-plugin/fillers/process.ts
```diff
@@ -126,6 +126,9 @@ export const process: NodeJS.Process = {
   setgid: noop(undefined),
   setgroups: noop(undefined),
   setuid: noop(undefined),
+  // See https://nodejs.org/api/process.html#processsourcemapsenabled
+  // Added in: v20.7.0
+  sourceMapsEnabled: false,
```
Jolg42, line 131: What do you think about this @millsp, is that ok?

package.json
```diff
@@ -40,8 +40,7 @@
     "@types/fs-extra": "9.0.13",
     "@types/graphviz": "0.0.37",
     "@types/jest": "29.5.6",
-    "@types/node": "18.18.6",
-    "@types/node-fetch": "2.6.7",
+    "@types/node": "20.8.9",
```
Jolg42, line 43: We run Node.js v20 for the publish script, for example.

helpers/compile/plugins/fill-plugin/fillers/process.ts
millsp: Sounds good!

millsp (approved):

PR #21678, opened 2023-10-30
feat: add support for nested transaction rollbacks via savepoints in sql

LucianBuzzo:
~~This adds a high level test that should pass if/when https://github.com/prisma/prisma-engines/pull/4375 is merged.~~
Transactions are now handled inside this repository, so the update to prisma-engines is no longer required.

This change adds support for handling rollbacks in nested transactions
in SQL databases. Specifically, the inner transaction should be rolled
back if the outer transaction fails.

To do this we keep track of the transaction ID and transaction depth so we can
re-use an existing open transaction in the underlying engine. This change also
allows the use of the `$transaction` method on an interactive transaction client.

## Summary by CodeRabbit

* **New Features**
  * Nested interactive transactions for SQL databases (Postgres, MySQL, SQLite, SQL Server) via savepoints; MongoDB remains without nested-transaction support.

* **Tests**
  * Added extensive nested-transaction tests covering multi-level nesting, savepoint behavior, rollback/commit propagation, concurrency and provider-specific syntax.

* **Documentation**
  * Updated runtime documentation to describe nested transaction behavior and provider-specific savepoint semantics.


packages/client/tests/functional/interactive-transactions/tests.ts
```diff
@@ -166,6 +166,46 @@ testMatrix.setupTestSuite(({ provider, providerFlavor }, _suiteMeta, clientMeta)
     await expect(result).resolves.toHaveLength(2)
   })
 
+  /**
+   * If a parent transaction is rolled back, the child transaction should also rollback
+   * - This is only supported in SQL derived servers
+   */
+  testIf(provider.toLowerCase().includes('sql'))('sql: nested rollback', async () => {
```
Jolg42, line 173:
```suggestion
  testIf(provider !== Providers.MONGODB)('sql: nested rollback', async () => {
```

Jolg42: Please add this import along this change
```
import { Providers } from '../_utils/providers'
```

LucianBuzzo:
@Jolg42 Is it possible to verify that https://github.com/prisma/prisma-engines/pull/4375 fixes the issue and gets the test to pass?

janpio:
`5.6.0-12.integration-sql-nested-transactions-a8af640da343c10e0f7421264c2838b302d3ce6d` via https://github.com/prisma/prisma/pull/21735/files should work for that.

janpio:
Ummm, seems it is not working @LucianBuzzo. Or I messed up royally. Asking some colleagues for advice.

LucianBuzzo:
@janpio I'm replicating the issue on my end - I think I might need to update the client code to maintain the same transaction ID in nested transactions 🤔 . I'm going to dig through and see what I can come up wiht.

docker/docker-compose.yml
```diff
@@ -60,17 +60,17 @@ services:
       timeout: 2s
       retries: 20
 
-  cockroachdb:
```
Jolg42, line 63: You'll need to undo the changes done here below to make tests pass in GitHub Actions.

janpio:
Replaced by https://github.com/prisma/prisma/pull/22161

packages/engines/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@prisma/debug": "workspace:*",
-    "@prisma/engines-version": "5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a",
+    "@prisma/engines-version": "5.7.0-22.integration-sql-nested-transactions2-f486ca20b160ea1fd92f8f351f6d05eff8b478b3",
```
janpio, line 25:
```suggestion
    "@prisma/engines-version": "5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8",
```

packages/fetch-engine/package.json
```diff
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "@prisma/debug": "workspace:*",
-    "@prisma/engines-version": "5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a",
+    "@prisma/engines-version": "5.7.0-22.integration-sql-nested-transactions2-f486ca20b160ea1fd92f8f351f6d05eff8b478b3",
```
janpio, line 46:
```suggestion
    "@prisma/engines-version": "5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8",
```

packages/internals/package.json
```diff
@@ -81,7 +81,7 @@
     "@prisma/fetch-engine": "workspace:*",
     "@prisma/generator-helper": "workspace:*",
     "@prisma/get-platform": "workspace:*",
-    "@prisma/prisma-schema-wasm": "5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a",
+    "@prisma/prisma-schema-wasm": "5.7.0-22.integration-sql-nested-transactions2-f486ca20b160ea1fd92f8f351f6d05eff8b478b3",
```
janpio, line 84:
```suggestion
    "@prisma/prisma-schema-wasm": "5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8",
```

packages/migrate/package.json
```diff
@@ -56,7 +56,7 @@
   },
   "dependencies": {
     "@prisma/debug": "workspace:*",
-    "@prisma/engines-version": "5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a",
+    "@prisma/engines-version": "5.7.0-22.integration-sql-nested-transactions2-f486ca20b160ea1fd92f8f351f6d05eff8b478b3",
```
janpio, line 59:
```suggestion
    "@prisma/engines-version": "5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8",
```

pnpm-lock.yaml
```diff
@@ -467,8 +467,8 @@ importers:
         specifier: workspace:*
         version: link:../engines
       '@prisma/engines-version':
-        specifier: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
-        version: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
+        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
+        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```
janpio, line 471:
```suggestion
        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```

pnpm-lock.yaml
```diff
@@ -763,8 +763,8 @@ importers:
         specifier: workspace:*
         version: link:../debug
       '@prisma/engines-version':
-        specifier: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
-        version: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
+        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
+        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```
janpio, line 767:
```suggestion
        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```

pnpm-lock.yaml
```diff
@@ -1088,8 +1088,8 @@ importers:
         specifier: workspace:*
         version: link:../get-platform
       '@prisma/prisma-schema-wasm':
-        specifier: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
-        version: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
+        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
+        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```
janpio, line 1092:
```suggestion
        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```

pnpm-lock.yaml
```diff
@@ -1245,8 +1245,8 @@ importers:
         specifier: workspace:*
         version: link:../debug
       '@prisma/engines-version':
-        specifier: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
-        version: 5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a
+        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
+        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```
janpio, line 1249:
```suggestion
        specifier: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
        version: 5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8
```

pnpm-lock.yaml
```diff
@@ -3166,17 +3166,17 @@ packages:
...
   /@prisma/mini-proxy@0.9.4:
     resolution: {integrity: sha512-QydFgafroCKNaLJ/79Zr9auEb2/87+v8gI8s6RdHyLkBL/iSRtv9btPgCvcpcm9IhN3uYHt6hloX/W16FdcJag==}
     engines: {node: '>=16'}
     hasBin: true
     dev: true
 
-  /@prisma/prisma-schema-wasm@5.7.0-21.f2e5886144b9dcd9774a6799317e94d76904001a:
-    resolution: {integrity: sha512-cWUGImR3ztSJpb2z+Vka/IS6rFEAvQ6AHlcH92btu+iAjcPpc46YlI23iujvfBpZySJALJmxdji+CNVG2ybWMg==}
+  /@prisma/prisma-schema-wasm@5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8:
+    resolution: {integrity: sha512-/bUANcYz7Zfkks62e/qWZscn369bqfPWmx86GkQ8AJ7HuJmds8lkOOOQ8vr7zPGvbHfn00SrVXtabLmQdfk0fQ==}
```
janpio, line 3179:
```suggestion
  /@prisma/prisma-schema-wasm@5.7.0-29.integration-sql-nested-transactions2-e943433a36b16524e1d88f0f0b93c3d8d4ddd4c8:
    resolution: {integrity: sha512-/bUANcYz7Zfkks62e/qWZscn369bqfPWmx86GkQ8AJ7HuJmds8lkOOOQ8vr7zPGvbHfn00SrVXtabLmQdfk0fQ==}

```

LucianBuzzo:
@janpio The tests are passing locally. I'm now working on an edge case where you use a batch transaction inside a client extensions 😬

Jolg42:
@LucianBuzzo to get the CI running, you would need to solve the git conflict. GitHub won't run any Action if there is a conflict with the base branch unfortunately 😢

LucianBuzzo:
@Jolg42 Cool, I'll get that sorted 👍

LucianBuzzo:
@Jolg42 I'm getting a lot of tests failing with an error about `relationJoins`, e.g.
```
issues.16535-select-enum (provider=mysql) › allows to select enum field

    GetDmmfError: Prisma schema validation - (get-dmmf wasm)
    Error code: P1012
    error: The preview feature "relationJoins" is not known. Expected one of: deno, driverAdapters, fullTextIndex, fullTextSearch, metrics, multiSchema, postgresqlExtensions, tracing, views
      -->  schema.prisma:6
       | 
     5 | engineType = "library"
     6 | previewFeatures = ["relationJoins"]
       | 
       
       ```
       
       Any idea why this is happening?

**Comment by Jolg42** [comment]
@LucianBuzzo 
>EDIT: I'm going to guess that the engine version I'm using doesn't have this feature - I'll work on getting the engine PR updated

Exactly, the `main` evolves rapidly, so getting your fork in sync would work I think.

**Comment by LucianBuzzo** [comment]
The new interactive rollback tests are now passing and behaving as expected, however there are now some failing tests elsewhere that I'll need to address.

**Comment by LucianBuzzo** [comment]
From what I can tell the transaction depth isn't getting decremented when using the `js_pg` adapter.
I've added a test to sanity check this:
```
 pnpm run test:functional --adapter js_pg --shard 6/6 --client-runtime wasm --preview-features driverAdapters --runInBand -t 'multiple interactive transactions'
 ```

LucianBuzzo:
@SevInf It looks like everything is now passing, except d1 and mini-proxy, which looks to be unreleated.

SevInf:
`mini-proxy` might actually be related: did we change binary-engine public interface in the engine PR?
Both `mini-proxy` and real Accelerate use it underneath, so we might take a closer look to ensure they won't break once it released. Of course, both of those projects are in the private repos, so there is nothing really actionable for you right now. I'll try to carve out some time to take a look myself, but realistically, that probably happen after 5.12.0 release next week.

LucianBuzzo:
@SevInf I've added a `begin()` method in the engine that is used to open the next transaction level on a transaction that has already been started. That might be the culprit. If there's nothing I can do to help, I'll sit tight until after the v5.12.0 release 👍

LucianBuzzo:
@SevInf Now that v5.12.0 is released, are you able to assist in the mini-proxy and d1 test failures? TIA 🙏

LucianBuzzo:
@SevInf @janpio Any updates on how to move this PR forwards?

janpio:
@LucianBuzzo I contacted you via LinkedIn (only method I could find) to talk about how getting a better way for us to talk to each other (Either invite you to our Slack, or via Discord?)

I merged, which of course lead to a few more tests failing - [`417db07` (#21678)](https://github.com/prisma/prisma/pull/21678/commits/417db0742e0f5b3fa0fa55dff3cdb3344adcd365) is the last commit before I did that to get to the test failures that probably exist and are real, and not just caused by new tests that the engine version here does not support yet.

LucianBuzzo:
@janpio Thanks, I'll check out the test failures - I've responded to you message on LinkedIn, slack is probably the best bet 👍

LucianBuzzo:
@SevInf Looks like the mini-proxy tests were failing because I was accidentally overwriting the transaction payload when making a request to the engine. This is fixed and everything appears to be working correctly. 
I think the `Client func&legacy-notypes` might be flakey - can you rerun the failed test?

packages/adapter-neon/src/neon.ts
```diff
@@ -154,6 +154,13 @@ class NeonTransaction extends NeonWsQueryable<neon.PoolClient> implements Transa
     super(client)
   }
 
+  async begin(): Promise<Result<void>> {
+    debug(`[js::begin]`)
+
+    this.client.release()
+    return Promise.resolve(ok(undefined))
+  }
```
aqrln, line 162: I'm not sure I understand this. Why does `begin` close the transaction?

packages/client/src/runtime/core/types/exported/itxClientDenyList.ts
```diff
@@ -1,4 +1,4 @@
-const denylist = ['$connect', '$disconnect', '$on', '$transaction', '$use', '$extends'] as const
+const denylist = ['$connect', '$disconnect', '$on', '$use', '$extends'] as const
```
aqrln, line 1: Shouldn't this depend on the provider now?

packages/adapter-neon/src/neon.ts
LucianBuzzo: This is a mistake, I think maybe a copypasta from the `commit` implementation 🤔 
I'll remove the release line.

packages/client/src/runtime/core/types/exported/itxClientDenyList.ts
LucianBuzzo: Yes correct, this is really only available for SQL servers. Do you know if there is a utility for determining if the provider is an SQL server?

packages/client/tests/functional/extensions/itx.ts
```diff
@@ -315,7 +315,7 @@ testMatrix.setupTestSuite(
             if (isTransaction) {
               expect(ctx.$connect).toBeUndefined()
               expect(ctx.$disconnect).toBeUndefined()
-              expect(ctx.$transaction).toBeUndefined()
+              expect(ctx.$transaction).toBeDefined()
```
aqrln, line 318: This should be undefined on MongoDB

LucianBuzzo: @aqrln My preference would be to leave it defined so there is a consistent interface, and throw a readable error, like what is done for the planetscale adapter.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -672,46 +678,110 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
-      options,
+      options = {},
     }: {
       callback: (client: Client) => Promise<unknown>
       options?: Options
     }) {
+      const isNested = Boolean(this[TX_ID])
+      const callerScopeId = isNested ? (this as any)[TX_SCOPE_ID] : undefined
+      const scopeState: ItxScopeState = isNested ? (this as any)[TX_SCOPE_STATE] : { stack: [] }
+      const scopeStack = scopeState?.stack
```
jacek-prisma: I think it'd be worth introducing a type for these 3 symbols to avoid repetetive and error prone casting to any

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -672,46 +678,110 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
+
+      if (isNested && !scopeStack) {
+        throw new Error('Internal error: missing transaction scope state for nested transaction.')
+      }
+
+      const scopeId =
+        typeof globalThis.crypto?.randomUUID === 'function'
+          ? globalThis.crypto.randomUUID()
+          : `${Date.now()}-${Math.random().toString(16).slice(2)}`
```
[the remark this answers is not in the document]
jacek-prisma: agreed, should be extracted from here

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -672,46 +678,110 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
+
       const headers = { traceparent: this._tracingHelper.getTraceParent() }
 
       const optionsWithDefaults: Options = {
         maxWait: options?.maxWait ?? this._engineConfig.transactionOptions.maxWait,
         timeout: options?.timeout ?? this._engineConfig.transactionOptions.timeout,
         isolationLevel: options?.isolationLevel ?? this._engineConfig.transactionOptions.isolationLevel,
+        newTxId: options.newTxId,
+      }
+      let info: any
```
jacek-prisma: Why not use the actual type? `InteractiveTransactionInfo`

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -672,46 +678,110 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
+          await this._engine.transaction('rollback', headers, info).catch(() => {})
+        }
 
         throw e // silent rollback, throw original error
+      } finally {
+        if (scopeStack.at(-1) === scopeId) {
+          scopeStack.pop()
+        } else {
+          // Reset the scope stack to avoid poisoning this transaction context after an ordering violation.
+          scopeStack.splice(0, scopeStack.length)
```
jacek-prisma:
```suggestion
          scopeStack.length = 0
```

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -109,6 +109,19 @@ export class RequestHandler {
...
+          const existingPayload =
+            typeof interactiveTransaction.payload === 'object' && interactiveTransaction.payload !== null
+              ? interactiveTransaction.payload
+              : {}
+
+          interactiveTransaction.payload = {
+            // If the interactive transaction has a payload, we need to merge it with the new_tx_id
+            ...(existingPayload as any),
+            new_tx_id: interactiveTransaction.id,
+          }
```
jacek-prisma: is this `new_tx_id` used anywhere?

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -39,6 +42,48 @@ const debug = Debug('prisma:client:transactionManager')
 const COMMIT_QUERY = (): SqlQuery => ({ sql: 'COMMIT', args: [], argTypes: [] })
 const ROLLBACK_QUERY = (): SqlQuery => ({ sql: 'ROLLBACK', args: [], argTypes: [] })
 
+const SAVEPOINT_QUERY = (provider: Transaction['provider'], name: string): SqlQuery => {
```
jacek-prisma: these should maybe just return the SQL string, since `args` and `argTypes` will always be empty

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -39,6 +42,48 @@ const debug = Debug('prisma:client:transactionManager')
...
+  switch (provider) {
```
jacek-prisma: it doesn't feel right to this here, I don't mind extending the driver adapter interface with optional methods and moving it all there

jacek-prisma: the same applies to `RELEASE_SAVEPOINT_QUERY` and `ROLLBACK_TO_SAVEPOINT_QUERY`

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -102,13 +147,53 @@ export class TransactionManager {
...
+      const savepointName = `prisma_sp_${existing.savepointCounter++}`
+      existing.savepoints.push(savepointName)
+
+      const query = SAVEPOINT_QUERY(existing.transaction.provider, savepointName)
+      try {
+        await this.#withQuerySpanAndEvent(query, existing.transaction, () => existing.transaction!.executeRaw(query))
+      } catch (e) {
+        // Keep state consistent if creating the savepoint fails.
+        existing.depth -= 1
+        existing.savepoints.pop()
```
jacek-prisma: Due to the `await` and a lack of synchronization, I think it's possible to end up in a race where `existing.savepoints.pop()` pops a different transaction

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -672,46 +678,110 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
+        const activeScope = scopeStack.at(-1)
+        if (activeScope !== callerScopeId) {
+          throw new Error('Concurrent nested transactions are not supported')
+        }
+
+        // Re-use the underlying transaction in the engine by reusing the same transaction id.
+        options.newTxId = this[TX_ID]
+        scopeStack.push(scopeId)
+      } else {
+        scopeStack.push(scopeId)
```
jacek-prisma: `scopeStack.push(scopeId)` can be pulled out of the `if`-`else`

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -672,46 +678,110 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
         const transaction = { kind: 'itx', ...info } as const
 
-        result = await callback(this._createItxClient(transaction))
+        result = await callback(this._createItxClient(transaction, scopeId, scopeState))
+
+        if (!isNested && scopeStack.length !== 1) {
+          throw new Error('Cannot close transaction while a nested transaction is still active.')
+        }
+        // Don't allow closing a transaction if we are not the active scope.
+        if (scopeStack.at(-1) !== scopeId) {
```
jacek-prisma: Doesn't this check cover the check above 100% of the time? I mean `!isNested && scopeStack.length !== 1`

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -672,46 +678,110 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
         await this._engine.transaction('commit', headers, info)
       } catch (e: any) {
-        // it went bad, then we rollback the transaction
-        await this._engine.transaction('rollback', headers, info).catch(() => {})
+        // If we try to close out-of-order (e.g. un-awaited nested transaction),
+        // the TransactionManager depth will be > 1 and a single rollback would
+        // only roll back to the latest savepoint, leaving the top-level transaction
+        // open and leaking it. Force rollback to depth 0 in that case.
+        const isOrderViolation = scopeStack.at(-1) !== scopeId
+        const rollbackAttempts = isOrderViolation ? Math.max(1, scopeStack.length) : 1
```
jacek-prisma: I'd rename this to make the intent clearer
```suggestion
        const rollbackScopeCount = isOrderViolation ? Math.max(1, scopeStack.length) : 1
```

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
LucianBuzzo: Yeah I see what you mean, it makes sense to have the driver adapter take responsibility for the savepoint SQL, and the manager remains fairly agnostic to the implementation of the savepoint itself.

packages/client/src/runtime/RequestHandler.ts
LucianBuzzo: No, this is leftover from when the savepoint code was handled in prisma-engines, I'll remove it.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -216,6 +216,27 @@ type EventCallback<E extends ExtendedEventType> = [E] extends ['beforeExit']
...
+
+type ItxScopeState = {
+  stack: string[]
+}
+
+type ItxScopeContext = {
+  txId: string | undefined
+  scopeId: string | undefined
+  scopeState: ItxScopeState | undefined
+}
```
jacek-prisma: I think these should always either all be undefined or all defined, the type should model these 2 states instead of handling each field separately

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -160,6 +161,16 @@ class BetterSQLite3Transaction extends BetterSQLite3Queryable<StdClient> impleme
...
   }
+
+  savepoint(action: SavepointAction, name: string): SqlQuery {
+    if (action === 'create') {
+      return { sql: `SAVEPOINT ${name}`, args: [], argTypes: [] }
+    }
+    if (action === 'rollback') {
+      return { sql: `ROLLBACK TO ${name}`, args: [], argTypes: [] }
+    }
+    return { sql: `RELEASE SAVEPOINT ${name}`, args: [], argTypes: [] }
```
jacek-prisma: For consistency with other APIs I think this should actually perform the IO (execute the underlying statement in the database) rather than returning a query

jacek-prisma: I mean doing that in every adapter and modifying the signature to allow `async`

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -39,6 +43,52 @@ const debug = Debug('prisma:client:transactionManager')
 const COMMIT_QUERY = (): SqlQuery => ({ sql: 'COMMIT', args: [], argTypes: [] })
 const ROLLBACK_QUERY = (): SqlQuery => ({ sql: 'ROLLBACK', args: [], argTypes: [] })
 
+// TODO: Remove this provider-based fallback once all adapters implement `transaction.savepoint`.
```
jacek-prisma: Why do we need a fallback? I think the new API method should be optional:
```
savepoint?(...): Promise<void>`
```
And adapters that don't implement should raise an error

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -102,13 +152,49 @@ export class TransactionManager {
...
+        }
+        if (!existing.transaction) {
+          throw new TransactionInternalConsistencyError(
+            `Transaction missing underlying driver transaction when starting a nested transaction.`,
+          )
+        }
+
+        existing.depth += 1
+
+        const savepointName = `prisma_sp_${existing.savepointCounter++}`
```
jacek-prisma: I think we should extract the generated names to dedicated functions, maybe in their own little file, so that they're not scattered all over the codebase

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -160,6 +161,16 @@ class BetterSQLite3Transaction extends BetterSQLite3Queryable<StdClient> impleme
     this.#unlockParent()
     return Promise.resolve()
   }
+
+  savepoint(action: SavepointAction, name: string): SqlQuery {
```
jacek-prisma: I think this would be better modelled as 3 methods
`createSavepoint?(name: string): Promise<void>`
`rollbackToSavepoint?(name: string): Promise<void>`
`releaseSavepoint?(name: string): Promise<void>`

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
jacek-prisma: I think it might also be worth extracting the counter into a little abstraction of it's own, like a name generator object

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -216,6 +216,56 @@ type EventCallback<E extends ExtendedEventType> = [E] extends ['beforeExit']
...
+
+  throw new Error('Internal error: inconsistent transaction scope context.')
+}
+
+function isItxScopeState(value: unknown): value is ItxScopeState {
+  if (typeof value !== 'object' || value === null) {
+    return false
+  }
+
+  return Array.isArray((value as Record<string, unknown>).stack)
```
jacek-prisma:
```suggestion
  return Array.isArray(value['stack'])
```

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -216,6 +216,56 @@ type EventCallback<E extends ExtendedEventType> = [E] extends ['beforeExit']
...
+  scopeState: ItxScopeState
+}
+
+type ItxScopeContext = TopLevelItxScopeContext | NestedItxScopeContext
+
+function getItxScopeContext(client: object): ItxScopeContext {
+  const symbolStorage = client as Record<symbol, unknown>
+  const txId = symbolStorage[TX_ID]
+  const scopeId = symbolStorage[TX_SCOPE_ID]
+  const scopeState = symbolStorage[TX_SCOPE_STATE]
```
jacek-prisma: For the sake of simplicity, I'd probably store all of this in one object that contains all of this information and gets stored under one symbol

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -237,20 +399,63 @@ export class TransactionManager {
...
+
+      await this.#runSerialized(tx, async () => {
+        const current = this.transactions.get(transactionId)
+        if (current && ['running', 'waiting'].includes(current.status)) {
+          await this.#closeTransaction(current, 'timed_out')
+        } else {
+          // Transaction was already committed or rolled back when timeout happened.
+          // Should normally not happen as timeout is cancelled when transaction is committed or rolled back.
+          // No further action needed though.
+          debug('Transaction already committed or rolled back when timeout happened.', transactionId)
```
jacek-prisma: instead of duplicating the comment + `debug` line, it might be worth extracting it to a little internal function

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
LucianBuzzo: @jacek-prisma For some reason I though that you could use custom/community provided adapters, and I didn't necessarily want to break them.

jacek-prisma: There are community adapters, but I think it's okay if they raise an error for this

jacek-prisma: There's only one external adapter that has any users (not many) that I'm aware of: https://github.com/tidbcloud/prisma-adapter
We could also add the new API there ahead of the release

LucianBuzzo: Ok cool 👍

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
LucianBuzzo: Are you thinking of having something that increments the counter and outputs a savepoint name?

packages/adapter-better-sqlite3/src/better-sqlite3.ts
LucianBuzzo: Yeah, sure and then the execution is fully delegated, rather than just the SQL generation.

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
jacek-prisma: Yeah

packages/adapter-better-sqlite3/src/better-sqlite3.ts
LucianBuzzo: Resolved by 6f750bfc712d8fe947ae4052b6203d6ade963a1c

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
LucianBuzzo: Resolved by 6f750bfc712d8fe947ae4052b6203d6ade963a1c

packages/adapter-better-sqlite3/src/better-sqlite3.ts
LucianBuzzo: Resolved by 6f750bfc712d8fe947ae4052b6203d6ade963a1c

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -215,7 +215,67 @@ type EventCallback<E extends ExtendedEventType> = [E] extends ['beforeExit']
...
+    isItxScopeState(objectValue['scopeState'])
+  )
+}
+
+function isItxScopeState(value: unknown): value is ItxScopeState {
+  if (typeof value !== 'object' || value === null) {
+    return false
+  }
+
+  return Array.isArray((value as Record<string, unknown>)['stack'])
```
jacek-prisma: `value['stack']` should be fine here

LucianBuzzo:
@jacek-prisma let's go for the hatrick? 😅 🤞

jacek-prisma (approved):

janpio:
Congratulations @LucianBuzzo!

LucianBuzzo:
Haha thanks @janpio we got there in the end! 🎉

Issue #21648, opened 2023-10-27
Postgres: too many bind variables in prepared statement on findAll


joaomlneto:
### Bug description

`findAll()` fails with error:
```js
Assertion violation on the database: `too many bind variables in prepared statement, expected maximum of 32767, received 32769`
```

### How to reproduce

1. Create table `LoyaltyStoreOffer` with a composite ID.
2. Create table `LoyaltyStoreOfferRequiredItem` that has an ID foreign key referencing `LoyaltyStoreOffer`.
3. Populate first table with _at least_ `16384` items.
4. Attempt to execute `findAll()` on the first table, while also selecting related entries rom the second table.

```ts
const isInitialized = (await prisma.loyaltyStoreOffer.count()) > 0;

if (!isInitialized) {
    console.log('going to initialize')

    await prisma.loyaltyStoreOffer.createMany({
        data: [...Array(16384).keys()].map(i => ({
            offerId: i,
            corporationId: i,
        }))
    })


    await prisma.loyaltyStoreOfferRequiredItem.createMany({
        data: [...Array(1).keys()].map(i => ({
            offerId: Math.floor(Math.sqrt(i)),
            corporationId: Math.floor(Math.sqrt(i)),
            typeId: Math.floor(Math.sqrt(i) + i),
        }))
    })

    console.log('initialized successfully!')
}

console.log('going to run buggy query')

await prisma.loyaltyStoreOffer.findMany({
    select: {
        offerId: true,
        corporationId: true,
        requiredItems: {
            select: {
                typeId: true,
                corporationId: true,
            },
        },
    },
})
```

### Expected behavior

Query executes successfully.

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

model LoyaltyStoreOffer {
  offerId       Int
  corporationId Int
  requiredItems LoyaltyStoreOfferRequiredItem[] @relation("LoyaltyStoreOfferRequiredItems")

  @@id([offerId, corporationId])
}

model LoyaltyStoreOfferRequiredItem {
  offer         LoyaltyStoreOffer @relation("LoyaltyStoreOfferRequiredItems", fields: [offerId, corporationId], references: [offerId, corporationId])
  offerId       Int
  corporationId Int
  typeId        Int

  @@id([offerId, corporationId, typeId])
}

```

```ts
await prisma.loyaltyStoreOffer.findMany({
    select: {
[...]

PR #21679, opened 2023-10-30
test(client): add "chunking-query" test suite, differentiating Driver Adapters from plain Rust drivers

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/394 and https://github.com/prisma/team-orm/issues/519.

In particular, we added a new functional test suite `chunking-query` that subsumes the test suites for issues 8832 and 9326, and adds `cockroachdb` to the test matrix.

Moreover, we now explicitly test scenarios in which Rust drivers get a `too many bind variables in prepared statement` database assertion error, but Driver Adapters don't.
These scenarios are explained by the new issues https://github.com/prisma/prisma/issues/21802 and https://github.com/prisma/prisma/issues/21803.

---

Note: the newly added tests are not applicable to PlanetScale, as Vitess supports at most 10k rows returned in a single query, and the test tries to retrieve 65k+ rows.


packages/client/tests/functional/too-many-bind-variables/tests.ts
```diff
@@ -0,0 +1,93 @@
...
+        try {
+          await selectWithIds(ids)
+
+          // unreachable
+          expect(true).toBe(false)
+        } catch (error) {
+          const e = error as Error
+
+          expect(e.message).toContain('Assertion violation on the database')
+          expect(e.message).toContain('too many bind variables in prepared statement')
```
janpio, line 64: Shouldn't these work as the engine should take care of chunking to split the queries into multiple?

Jolg42: That was also my expectation, same for this user in https://github.com/prisma/prisma/issues/21648#issuecomment-1783838900
Because we fixed something that looks similar in https://github.com/prisma/prisma/issues/8832#issuecomment-1260533913

miguelff: This is indeed a bug, the Postgres extended protocol uses a short to identify the number of query parameters, but we know that at the engine level, because we know wether rust drivers will run. We could fix that in this PR, but I don't think it's in the scope of this issue, but a correctness one, as it's not a bug in driver adapters code.

miguelff: Actually the investigation subject of https://github.com/prisma/team-orm/issues/394 is done without the need to merge this PR and the outcome should be a correctness issue to fix the binding problem when using rust drivers.

jkomyno: We have had previous indications of Driver Adapters not using the query chunking logic defined in the Prisma engines.
Keeping [this](https://github.com/prisma/team-orm/issues/394#issuecomment-1787218882) in mind, I've answered in a top-level comment for future visibility. See: https://github.com/prisma/prisma/pull/21679#issuecomment-1787315586.

packages/client/tests/functional/too-many-bind-variables/tests.ts
```diff
@@ -0,0 +1,93 @@
...
+      try {
+        await selectWithIds(ids)
+      } catch (error) {
+        const e = error as Error
+
+        expect(e.message).toContain(
+          'code: "08P01", message: "bind message has 32767 parameter formats but 0 parameters"',
+        )
+      }
+    })
```
jkomyno, line 77: @miguelff, this shows Driver Adapters failing on queries with `65536` parameters

packages/client/tests/functional/too-many-bind-variables/tests.ts
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    testIf(providerFlavor !== undefined)(
+      'Using Driver Adapters with massive filter list does not cause errors',
+      async () => {
+        const n = N_LARGE
+        const ids = generatedIds(n)
+        await selectWithIds(ids)
+      },
+    )
```
jkomyno, line 60: @miguelff, this shows Driver Adapters succeeding on queries with `32768` parameters, despite Rust Drivers failing on the same number

jkomyno:
Note: currently investigating some divergence between local execution and CI runs.

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+      })
+      return tags
+    }
+
+    async function clean() {
+      const cleanPrismaPromises = [prisma.tagsOnPosts.deleteMany(), prisma.post.deleteMany(), prisma.tag.deleteMany()]
+      for (const promise of cleanPrismaPromises) {
+        await promise
+      }
+    }
```
janpio, line 72: I think these are all specific to the first `describe`, could we move them there?

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+          // unreachable
+          expect(true).toBe(false)
+        } catch (error) {
+          const e = error as Error
+
+          if (['postgresql', 'cockroachdb'].includes(provider)) {
+            expect(e.message).toContain('bind message has 32767 parameter formats but 0 parameters')
+          }
+        }
+      })
```
janpio, line 242: These are the failing cases, can we wrap them into their own `describe`?
(That would also allow us to move `selectWith2InFilters` into here as it is used in both `describeIf` that will then be children of it)

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+            {
+              id: { in: ids },
+            },
+            {
+              id: { in: ids },
+            },
+          ],
+        },
+      })
+    }
```
janpio, line 28: See https://github.com/prisma/prisma/pull/21679/files#r1383648125

packages/client/tests/functional/issues/8832/tests.ts
```diff
@@ -1,207 +0,0 @@
...
-              expect(error.meta).toMatchObject({
-                database_error:
-                  'too many bind variables in prepared statement, expected maximum of 32767, received 65533',
-              })
-            }
-          },
-        )
-
-        // TODO this test does not fail as expected
-        skipTestIf(providerFlavor === ProviderFlavors.JS_NEON || providerFlavor === ProviderFlavors.JS_PG)(
```
janpio, line 157: This is the important bit that is missing in the new test file, right?

packages/client/tests/functional/issues/9326/tests.ts
```diff
@@ -1,87 +0,0 @@
...
-      test('should succeed when the number of params is 32767 or less', async () => {
-        const n = 32767
-        const ids = await createTags(n)
-        const tags = await getTagsParams(ids)
-
-        expect(tags.length).toBe(n)
-      })
-
-      // TODO this test does not fail on adapters, and has zero assertions. Are we creating prepared statements?
-      skipTestIf(providerFlavor === ProviderFlavors.JS_NEON || providerFlavor === ProviderFlavors.JS_PG)(
```
janpio, line 57: This is the important bit that is missing in the new test file, right?

packages/client/tests/functional/issues/8832/tests.ts
jkomyno: This troublesome bit has been reworked so that tests are run on `JS_NEON` / `JS_PG` as well as Rust drivers in the new `chunking-queries` test suite.

packages/client/tests/functional/issues/9326/tests.ts
jkomyno: This troublesome bit has been reworked so that tests are run on `JS_NEON` / `JS_PG` as well as Rust drivers in the new `chunking-queries` test suite.

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: > These are the failing cases

Not all of them are failing: "With Driver Adapters only > Selecting 32768 ids at once in two inclusive disjunct filters works" is succeeding. Where should that be? How many `describe` blocks would you recommend having here?

janpio: Enough to structure the code in a way that one can understand what the "helper functions" are used for.

Generally these are the queries where chunking currently does not work though, and for driver adapters it only works as the driver can handle more than what our chunking would create (as our chunking is based on the tokio-postgres capabilities, which are artificially low). So a grouping describe "chunking logic does not trigger" would work for me here.

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -255,5 +255,9 @@ testMatrix.setupTestSuite(
       // query server parameter.
       reason: 'Vitess supports at most 10k rows returned in a single query, so this test is not applicable.',
     },
+    skipDataProxy: {
+      runtimes: ['node', 'edge'],
+      reason: 'Unreachable assertions in the data proxy',
```
janpio, line 260: What does "unreachable assertions" mean here?
What happens for Data Proxy / Mini Proxy?

jkomyno: As written elsewhere, the CI fails on *Proxy, as the test execution runs "unreachable assertions" (which appear in the code as `expect(true).toBe(false)`). The `skipDataProxy` settings avoids this problem.

I'm very confused about why such gimmick is needed in the first place, though.

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: (I have implemented this refactoring)

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: I've found the problem, and I'm fixing it.

jkomyno: It's probably fixed now. I've discovered that, if we run the MySQL-equivalent snippet of https://github.com/prisma/prisma/issues/21802 (`n = 32767`), we don't get an error at all (as opposed to Postgres/CockroachDB, where we get the error "too many bind variables in prepared statement"). However, if we increase `n` by 1 (`n = 32767`), we get the error "Prepared statement contains too many placeholders", which is analogous to Postgres' "too many bind variables in prepared statement"

jkomyno: Now I'm investigating a potential race condition: sometimes (with Neon), we get an error message containing "bind message has 5 parameter formats but 0 parameters", although the expectation is "bind message has 32767 parameter formats but 0 parameters".

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,265 @@
...
+  ({ provider, providerFlavor }) => {
+    function generatedIds(n: number) {
+      // ["1","2",...,"n"]
+      const ids = Array.from({ length: n }, (_, i) => i + 1)
+      return ids
+    }
+
+    // Note: we're excluding `sqlite` from the success cases until `createMany` is added
+    // to the SQLite driver. See https://github.com/prisma/prisma/issues/10710.
+    // We're also excluding `js_neon` the `createMany` would take ages otherwise (while `js_pg` is much).
```
janpio, line 17: Why does it takes ages with Neon?
What is the `(while `js_pg` is much)` supposed to mean? Don't understand.

janpio: Where are we actually exclusing `js_neon`? Just outdated comment?

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,265 @@
...
+testMatrix.setupTestSuite(
+  ({ provider, providerFlavor }) => {
+    function generatedIds(n: number) {
+      // ["1","2",...,"n"]
+      const ids = Array.from({ length: n }, (_, i) => i + 1)
+      return ids
+    }
+
+    // Note: we're excluding `sqlite` from the success cases until `createMany` is added
+    // to the SQLite driver. See https://github.com/prisma/prisma/issues/10710.
```
janpio, line 16: This should probably be the explanation in the `optOut` at the bottom instead instead of a comment here?

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,265 @@
...
+        expect(tags.length).toBe(n - 1)
+      })
+
+      test('should succeed when the explicit number of params is 32767 or less', async () => {
+        const n = 32767
+        const ids = await createTags(n)
+        const tags = await getTagsParams(ids)
+
+        expect(tags.length).toBe(n)
+      })
```
janpio, line 160: Unrelated to this PR:
1. All these tests would benefit from an explanation _why_ they succeed. Either because the number of params is fine for one query, or because we detect the too high number and split into multiple queries, right?
2. Additionally it would be really nice if we would really confirm the number of queries that are executed. That would also "document" what the chunking is doing.

packages/client/tests/functional/chunking-query/tests.ts
janpio: Including MysQL definitely makes this more messy, but good learning! Seems the missing chunking behaves different there than with PostgreSQL. Good call 👍

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: Yes, the placement of this comment has historical reasons that no longer apply. I'm moving it below.

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: The comment for `js_neon` no longer applies. I had added the comment in a previous iteration, where my environment loaded Node.js rather than Node.js 20. This resulted in WebSockets silently failing to connect to the database, resulting in a stalled CI (until the 180s time limit of our test suite is reached).
I'll remove the comment, thanks for spotting it out.

Note: I've double-checked that the CI for Driver Adapters uses Node.js 20 only.

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: https://github.com/prisma/prisma/issues/21827

janpio (approved):
Look good to me now. 
Wait for actual Eng review though before merge, please.

miguelff (approved):

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+          },
+        })
+        return tags
+      }
+
+      async function clean() {
+        const cleanPrismaPromises = [prisma.tagsOnPosts.deleteMany(), prisma.post.deleteMany(), prisma.tag.deleteMany()]
+        for (const promise of cleanPrismaPromises) {
+          await promise
+        }
```
Jolg42, line 66: Nit: I'm more used to write a transaction for this, because the order is guaranteed and it makes the code easier I think like
```suggestion
        await prisma.$transaction([
          prisma.tagsOnPosts.deleteMany(), 
          prisma.post.deleteMany(), 
          prisma.tag.deleteMany()
        ])
```

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+            }
+          }
+        })
+      })
+
+      describeIf(providerFlavor !== undefined)('With Driver Adapters only', () => {
+        test('Selecting 32768 ids at once in two inclusive disjunct filters works', async () => {
+          const ids = generatedIds(32768)
+          await selectWith2InFilters(ids)
+        })
```
Jolg42, line 227: It's missing an expect like 
```suggestion
        test('Selecting 32768 ids at once in two inclusive disjunct filters works', async () => {
          const ids = generatedIds(32768)
          await expect(selectWith2InFilters(ids)).resolves.toBeTruthy()
        })
```
`toBeTruthy()` is an example, maybe there is a better thing to expect here but I still need to read through things

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+            // unreachable
+            expect(true).toBe(false)
+          } catch (error) {
+            const e = error as Error
+
+            if (['postgresql', 'cockroachdb'].includes(provider)) {
+              expect(e.message).toContain('bind message has 32767 parameter formats but 0 parameters')
+            }
+          }
+        })
```
Jolg42, line 244:
```suggestion
        test('Selecting 65536 ids at once in two inclusive disjunct filters results in error', async () => {
          expect.assertions(1)
          const ids = generatedIds(65536)

          try {
            // this should throw 
            await selectWith2InFilters(ids)
          } catch (error) {
            const e = error as Error

            if (['postgresql', 'cockroachdb'].includes(provider)) {
              expect(e.message).toContain('bind message has 32767 parameter formats but 0 parameters')
            }
          }
        })
```
When adding expects inside a `try catch` we can add an expectation for the number of assertions `expect.assertions(1)`
Also, it's the pattern we follow in other tests so for clarity and consistency it's better I think.

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: If that rejected, an error would be thrown, and the test would fail (so we'd see that in the CI).
Note: this is a straight-up copy of previously existing tests.

packages/client/tests/functional/chunking-query/tests.ts
jkomyno: While `expect.assertions(_)` is used elsewhere for the reasons you suggest, in this case, it would add more noise than benefits. Why? Because the `expect()` in the `catch` branch is only called conditionally, so the number of expected assertions should also be computed conditionally. Thanks for the comment, though

packages/client/tests/functional/chunking-query/tests.ts
```diff
@@ -0,0 +1,263 @@
...
+      })
+    })
+  },
+  {
+    optOut: {
+      from: ['sqlserver', 'mongodb', 'sqlite'],
+      reason:
+        'not relevant for this test. Sqlite is excluded due to it lacking `createMany` (see: https://github.com/prisma/prisma/issues/10710).',
```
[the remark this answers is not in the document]
jkomyno: In the original tests, only `postgres` was tested. Now, we test `cockroachdb` (as it uses the same Rust driver) and `mysql` (added originally to test `planetscale`, which is however skipped atm).

There was no specific reason to very `sqlserver` as well, as that's out of scope for the issue this PR solves, but that could change in the future.

PR #21688, opened 2023-10-30
fix(adapter-libsql): support DateTime values saved as integers

### LINKED ISSUE #21552: Inconsistent datetime column data importing to Turso
### Bug description

Creating a db on TURSO from the local sqlite file built from Prisma results in the following error on columns using DateTime type in schema.prisma

`Conversion failed: expected a string in column createdAt, found 1697658213523.0`

### Local Dev

1. Setup Prisma for local `sqlite` development
2. Create a model with DateTimes
3. Seed entries into that model

### Turso Setup

1. Create a TURSO db using the local file
  a. `turso db create test-db --from-file prisma/data.db`
2. Connect to turso via [these official instructions](https://www.prisma.io/blog/prisma-turso-ea-support-rXGd_Tmy3UXX#connect-to-turso-using-prisma)
#21546 

### Query

1. Find a row from your model by id or whatever

### Expected behavior

Row gets returned without error

### Prisma information

```prisma
generator client {
}

datasource db {
  provider = "sqlite"
}

model Test {
  id             String               @id @unique @default(cuid())
  createdAt      DateTime             @default(now())
  updatedAt      DateTime             @default(now()) @updatedAt
}


```

```ts
  let test = await prisma.Test.findMany({
    where: {
      id: 'knownID',
    }
  })
```

### Environment & setup

- OS: macos
- Database: sqlite / turso
- Node.js version: v18.16.1

### Prisma Version

```
prisma                  : 5.4.2
@prisma/client          : 5.4.2
Query Engine (Node-API) : libquery-engine ac9d7041ed77bcc8a8dbd2ab6616b39013829574 (at ../../node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli ac9d7041ed77bcc8a8dbd2ab6616b39013829574 (at ../../node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.4.1-2.ac9d7041ed77bcc8a8dbd2ab6616b39013829574
Default Engines Hash    : ac9d7041ed77bcc8a8dbd2ab6616b39013829574
```

aqrln:
Support datetime values saved as integer timestamps, which is the format
historically used by the native quaint connector. The native connector
itself supports reading datetime values saved as ISO strings and some
other formats.

Fixes: https://github.com/prisma/prisma/issues/21552

### DIFF
--- a/packages/client/tests/functional/issues/21552-datetime-representation-compatibility/tests.ts
+++ b/packages/client/tests/functional/issues/21552-datetime-representation-compatibility/tests.ts
@@ -15,6 +15,8 @@ declare const datasourceInfo: DatasourceInfo
 testMatrix.setupTestSuite(
   () => {
     test('can read back DateTime created via native connector', async () => {
+      // Use a separate process for seeding to use native drivers. We can't turn
+      // the native quaint executor back on in the current process.
       const seedOutput = await execa.node(path.join(__dirname, '_seed.ts'), [], {
         nodeOptions: ['-r', 'esbuild-register'],
         env: {
--- a/packages/client/tests/functional/issues/21552-datetime-representation-compatibility/tests.ts
+++ b/packages/client/tests/functional/issues/21552-datetime-representation-compatibility/tests.ts
@@ -1,0 +1,46 @@
+import execa from 'execa'
+import path from 'path'
+
+import { DatasourceInfo } from '../../_utils/setupTestSuiteEnv'
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './node_modules/@prisma/client'
+
+declare const prisma: PrismaClient
+declare const datasourceInfo: DatasourceInfo
+
+/**
+ * Regression test for https://github.com/prisma/prisma/issues/21552
+ */
+testMatrix.setupTestSuite(
+  () => {
+    test('can read back DateTime created via native connector', async () => {
+      const seedOutput = await execa.node(path.join(__dirname, '_seed.ts'), [], {
+        nodeOptions: ['-r', 'esbuild-register'],
+        env: {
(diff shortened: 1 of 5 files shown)


packages/client/tests/functional/issues/21552-datetime-representation-compatibility/tests.ts
```diff
@@ -0,0 +1,54 @@
...
+testMatrix.setupTestSuite(
+  (_suiteConfig, suiteMeta) => {
+    test('can read back DateTime created via native connector', async () => {
+      // Copy the seed script to the generated directory so that it imports the
+      // correct client, and so that the schema-relative sqlite database path is
+      // resolved the same way as in the test.
+      await fs.cp(path.join(__dirname, '_seed.ts'), path.join(suiteMeta.generatedFolder, '_seed.ts'))
+
+      // Use a separate process for seeding to use native drivers. We can't turn
+      // the native quaint executor back on in the current process.
```
millsp, line 25: Do you mean that you need this because `PRISMA_DISABLE_QUAINT_EXECUTORS` may be disabled when running against driver adapters?

millsp (approved):

packages/client/tests/functional/issues/21552-datetime-representation-compatibility/tests.ts
aqrln: Yes, and in fact that's the main thing that we are testing, the test is not even really relevant without driver adapters.

aqrln: I expanded the comment a tiny bit so it's hopefully more clear without the context.

PR #21691, opened 2023-10-30
fix(client): Fix itx forwarding and add missing tests

SevInf:
Before, we postponed some of the tests for future: error forwarding for
adapter and validation error if adapter is used in binary engine. This
PR adds missing tests and fixes couple of things they uncovered:

- error forwarding was not working correctly for interactive
  transactions.

- validation error for adapter in binary engine happened a little bit
  too late and did not follow our usual conventions for constructor
  options validation.

Close prisma/team-orm#363


packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -565,3 +575,7 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
     return this.parseEngineResponse(responseString)
   }
 }
+
+function isUserFacingError(e: unknown): e is RequestError['user_facing_error'] {
+  return typeof e === 'object' && e !== null && e['error_code']
```
aqrln, line 580:
```suggestion
  return typeof e === 'object' && e !== null && e['error_code'] !== undefined
```

so that the return value of the whole function is `true` or `false` and not `undefined` or some truthy value

packages/client/tests/functional/adapter-errors/tests.ts
```diff
@@ -0,0 +1,87 @@
...
+          executeRaw() {
+            throw executeRawError
+          },
+          startTransaction() {
+            throw transactionError
+          },
+          close() {
+            // method is not used at the moment
+            return Promise.resolve(ok(undefined))
+          },
```
aqrln, line 33: There's a mock adapter in `tests/functional/driver-adapters/_utils/mock-adapter.ts` but it currently throws the same error in each method, I think it would be better to update it to throw different errors like you did here and reuse it.

packages/client/tests/functional/adapter-errors/tests.ts
```diff
@@ -0,0 +1,87 @@
...
+    },
+    skipDataProxy: {
+      runtimes: ['edge', 'node'],
+      reason: 'Test requires driver adapters that are not available on Data Proxy',
+    },
+    optOut: {
+      from: ['postgresql', 'mysql', 'mongodb', 'cockroachdb', 'sqlserver'],
+      reason: 'Test uses mock adapter and does not touch an actual database',
+    },
+  },
```
aqrln, line 86: `defaultTestSuiteOptions` from `tests/functional/driver-adapters/_utils/test-suite-options.ts` might be helpful here but the opt out logic there follows a bit different pattern and allows the test to run for each provider flavour, even though the test suites those options are used in also use mock adapter and do not touch an actual database. Do you think it would be better to reduce the number of test runs there as well?

packages/client/tests/functional/large-floats/tests.ts
```diff
@@ -32,9 +32,9 @@ testMatrix.setupTestSuite(
...
   },
   {
-    skipProviderFlavor: {
-      from: ['js_neon', 'js_pg'],
-      reason: 'Expected: 9007199254740991 Received: 9007199254740990',
-    },
+    // skipProviderFlavor: {
+    //   from: ['js_neon', 'js_pg'],
+    //   reason: 'Expected: 9007199254740991 Received: 9007199254740990',
+    // },
```
aqrln, line 38: Should this be removed now?

packages/client/tests/functional/adapter-errors/tests.ts
aqrln: also, would it make sense to move these test suites to `test/functional/driver-adapters`?

SevInf: I've added an export for all the errors mock adapter throws and re-used it here now.
Also, moved the tests.

packages/client/tests/functional/adapter-errors/tests.ts
SevInf: In the end I've decided it does not have a good reason for skipping other providers - test is fast, it's better not to rely on our assumption that nothing there would ever depenend on provider flavour.
So, i'm just reused default options here now.

packages/client/tests/functional/large-floats/tests.ts
SevInf: Accidental unrelated change, reverted.

aqrln (approved):
Looks good to me, thank you!

PR #21697, opened 2023-10-30
test(client): unskip upsert/native-atomic for libsql adapter

aqrln:
`upsert/native-atomic` test was dependent on the timing of the logs
arriving before the assertion is made. This might not necessarily
happen, and in practice turned out to be flaky on CI with libSQL
adapter. This PR makes the test wait for the logs using the
`waitFor` utility and un-skips the tests for libSQL adapter.

Closes: https://github.com/prisma/team-orm/issues/398

---

Hide whitespace changes for ease of review:
image

### DIFF
--- a/packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
+++ b/packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
@@ -183,7 +183,6 @@ testMatrix.setupTestSuite(
       await checker.expectUsedNativeUpsert(true)
     })
 
-    // TODO flaky on CI, never fails locally, reason unknown
     test('should only use ON CONFLICT when there is only 1 unique field in the where clause', async () => {
       const name = faker.person.firstName()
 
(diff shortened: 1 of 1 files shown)


packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
```diff
@@ -66,218 +61,209 @@ testMatrix.setupTestSuite(
...
+        },
+      })
 
-        expect(checker.usedNative()).toBeTruthy()
-      },
-    )
+      await checker.expectUsedNativeUpsert(true)
+    })
 
     // TODO flaky on CI, never fails locally, reason unknown
```
millsp, line 265:
```suggestion
```

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
```diff
@@ -66,218 +61,209 @@ testMatrix.setupTestSuite(
     })
 
     // TODO flaky on CI, never fails locally, reason unknown
```
millsp, line 63:
```suggestion
```

millsp (approved):
Thanks for making the test easier to read!

packages/client/tests/functional/methods/upsert/native-atomic/tests.ts
aqrln: Thanks for catching this!

PR #21705, opened 2023-10-31
test(client): un-skip and fix `issues/6578` for libSQL adapter

aqrln:
- Un-skip the `DateTime` and `UUID` test in
  `test/functional/issues/6578` for libSQL adapter.
- Fix reliance on the timing of an asynchronous operation that makes the
  test fail on CI but not locally.

Closes: https://github.com/prisma/team-orm/issues/405

### DIFF
--- a/packages/client/tests/functional/issues/6578/tests.ts
+++ b/packages/client/tests/functional/issues/6578/tests.ts
@@ -1,5 +1,6 @@
 import { QueryEvent } from '../../../../src/runtime/getPrismaClient'
 import { ProviderFlavors, Providers } from '../../_utils/providers'
+import { waitFor } from '../../_utils/tests/waitFor'
 import { NewPrismaClient } from '../../_utils/types'
 import testMatrix from './_matrix'
 // @ts-ignore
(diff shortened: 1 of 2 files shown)


packages/client/tests/functional/issues/6578/tests.ts
```diff
@@ -60,8 +60,10 @@ testMatrix.setupTestSuite(
           })
         }
 
-        // This test is asserting that JSON.parse does not throw because quotes are used
-        const params = JSON.parse(paramsString)
+        // This test is asserting that JSON.parse does not throw because quotes are used.
+        // However, we may need to retry if parsing fails because logs haven't arrived yet
+        // and `paramsString` is empty.
+        const params = await waitFor(() => JSON.parse(paramsString))
```
aqrln, line 66: On the second thought it might've been simpler and perhaps more explicit to just do

```ts
await waitFor(() => paramsString !== '')

const params = JSON.parse(paramsString)
```

Happy to change it if that's what reviewers would prefer, I don't mind either way.

Jolg42: I like this version you suggest because it's more explicit and if something fails it would fail on a specific line (and not on a does multiple things line)

Jolg42 (approved):

millsp (approved):

packages/client/tests/functional/issues/6578/tests.ts
aqrln: Actually it won't work like this, it will have to be something like this instead:

```ts
await waitFor(() => {
  if (paramsString === '') {
    throw new Error('params not received')
  }
})
```

So more explicit but not simpler. Would you still prefer it?

jkomyno: I like the explicit approach.

jkomyno (approved):

PR #21713, opened 2023-10-31
test: Change how connection strings are created for and provided to integration tests

janpio:
As preparation for being able to run PgBouncer tests, I needed to clean up the way our Postgres tests work with the connection string. This lead to this cleanup:

- Don't modify and append to `process.env.TEST_POSTGRES_URI` or `TEST_MYSQL_URI`, replace database name instead and then store in other env var instead - so it will be easier to change the original values (for example to add a parameter)
- Remove `TEST_POSTGRES_BASE_URI`, `TEST_MYSQL_BASE_URI` and `TEST_MARIADB_BASE_URI`, as they were not really necessary and just made things more complicated


.db.env
```diff
@@ -1,28 +1,25 @@
 # PostgreSQL
-TEST_POSTGRES_BASE_URI="postgres://prisma:prisma@localhost:5432"
```
janpio, line 2: removed

docker/docker-compose.yml
```diff
@@ -80,7 +80,8 @@ services:
     environment:
       - MYSQL_ROOT_PASSWORD=root
       - MYSQL_DATABASE=tests
-      - MYSQL_USER=prisma
+      - MYSQL_USER=prisma # not used
```
millsp, line 83: Should we keep this if we always use the root user?

packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -69,7 +69,7 @@ describe('studio with alternative urls and prisma://', () => {
   })
 
   testIf(process.platform !== 'win32')('queries work if url is prisma:// via the mini-proxy', async () => {
-    process.env.DATABASE_URL = `${process.env.TEST_POSTGRES_URI}-${Date.now()}-studio`
+    process.env.DATABASE_URL = process.env.TEST_POSTGRES_URI!.replace('tests', `tests-${Date.now()}-studio`)
```
millsp, line 72: Could we use `PRISMA_DB_NAME` as a keyword instead?

millsp (approved):

docker/docker-compose.yml
janpio: I don't know, honestly. 
This not being used might actually be a smell we want to fix (in the future)?

packages/cli/src/__tests__/commands/Studio.test.ts
Jolg42: Very true

docker/docker-compose.yml
Jolg42: It's fine to not do anything about this

.db.env
```diff
@@ -1,28 +1,25 @@
...
 TEST_POSTGRES_URI="postgres://prisma:prisma@localhost:5432/tests"
```
Jolg42, line 2:
```suggestion
# Note that the database name must be `tests` (in our test setup we search and replace that value later)
TEST_POSTGRES_URI="postgres://prisma:prisma@localhost:5432/tests"
```

.db.env
```diff
@@ -1,28 +1,25 @@
...
-TEST_MYSQL_ISOLATED_URI="mysql://root:root@localhost:3307/tests"
 TEST_MYSQL_URI_MIGRATE="mysql://root:root@localhost:3306/tests-migrate"
 TEST_MYSQL_SHADOWDB_URI_MIGRATE="mysql://root:root@localhost:3306/tests-migrate-shadowdb"
+# Note: the isolated instance is only needed for one test (client/src/__tests__/integration/errors/connection-limit-mysql/test.ts)
+TEST_MYSQL_ISOLATED_URI="mysql://root:root@localhost:3307/tests"
 
 # MariaDB
-TEST_MARIADB_BASE_URI="mysql://root:root@localhost:4306"
-TEST_MARIADB_URI="mysql://prisma:prisma@localhost:4306/tests"
+TEST_MARIADB_URI="mysql://root:root@localhost:4306/tests"
```
Jolg42, line 16:
```suggestion
# Note that the database name must be `tests` (in our test setup we search and replace that value later)
TEST_MARIADB_URI="mysql://root:root@localhost:4306/tests"
```

.envrc
```diff
@@ -1,28 +1,25 @@
 # PostgreSQL
-export TEST_POSTGRES_BASE_URI="postgres://prisma:prisma@localhost:5432"
 export TEST_POSTGRES_URI="postgres://prisma:prisma@localhost:5432/tests"
-# Note: the isolated instance is only needed for one test (client/src/__tests__/integration/errors/connection-limit-postgres/test.ts)
-export TEST_POSTGRES_ISOLATED_URI="postgres://prisma:prisma@localhost:5435/tests"
 export TEST_POSTGRES_URI_MIGRATE="postgres://prisma:prisma@localhost:5432/tests-migrate"
```
Jolg42, line 3:
```suggestion
# Note that the database name must be `tests` (in our test setup we search and replace that value later)
export TEST_POSTGRES_URI_MIGRATE="postgres://prisma:prisma@localhost:5432/tests-migrate"
```

.envrc
```diff
@@ -1,28 +1,25 @@
...
-export TEST_MYSQL_ISOLATED_URI="mysql://root:root@localhost:3307/tests"
 export TEST_MYSQL_URI_MIGRATE="mysql://root:root@localhost:3306/tests-migrate"
 export TEST_MYSQL_SHADOWDB_URI_MIGRATE="mysql://root:root@localhost:3306/tests-migrate-shadowdb"
+# Note: the isolated instance is only needed for one test (client/src/__tests__/integration/errors/connection-limit-mysql/test.ts)
+export TEST_MYSQL_ISOLATED_URI="mysql://root:root@localhost:3307/tests"
 
 # MariaDB
-export TEST_MARIADB_BASE_URI="mysql://root:root@localhost:4306"
-export TEST_MARIADB_URI="mysql://prisma:prisma@localhost:4306/tests"
+export TEST_MARIADB_URI="mysql://root:root@localhost:4306/tests"
```
Jolg42, line 16:
```suggestion
# Note that the database name must be `tests` (in our test setup we search and replace that value later)
export TEST_MARIADB_URI="mysql://root:root@localhost:4306/tests"
```

.envrc
```diff
@@ -1,28 +1,25 @@
...
 
 # SQL Server
-export TEST_MSSQL_URI="mssql://SA:Pr1sm4_Pr1sm4@localhost:1433/master" # for `mssql` lib used in some tests
 export TEST_MSSQL_JDBC_URI="sqlserver://localhost:1433;database=master;user=SA;password=Pr1sm4_Pr1sm4;trustServerCertificate=true;"
```
Jolg42, line 19:
```suggestion
# Note that the database name must be `master` (in our test setup we search and replace that value later)
export TEST_MSSQL_JDBC_URI="sqlserver://localhost:1433;database=master;user=SA;password=Pr1sm4_Pr1sm4;trustServerCertificate=true;"
```

Jolg42 (approved):
Looks great! It's not fun to touch these details everywhere, and it looks better now.
See suggestions.

.db.env
```diff
@@ -1,28 +1,25 @@
 # PostgreSQL
```
janpio, line 1:
```suggestion
# Note: The default database name for the main connection string of each database has to be `tests` (or `master` for SQL Server (as we we search and replace that value later in our test setup)

# PostgreSQL
```

.envrc
```diff
@@ -1,28 +1,25 @@
 # PostgreSQL
```
janpio, line 1:
```suggestion
# Note: The default database name for the main connection string of each database has to be `tests` (or `master` for SQL Server (as we we search and replace that value later in our test setup)

# PostgreSQL
```

.db.env
janpio: Decided to add a note at top of file instead as otherwise we need to repeat this one too many times all over the files

.db.env
janpio: Decided to add a note at top of file instead as otherwise we need to repeat this one too many times all over the files

.envrc
janpio: Decided to add a note at top of file instead as otherwise we need to repeat this one too many times all over the files

.envrc
janpio: Decided to add a note at top of file instead as otherwise we need to repeat this one too many times all over the files

.envrc
janpio: Decided to add a note at top of file instead as otherwise we need to repeat this one too many times all over the files

PR #21745, opened 2023-11-02
test(driver-adapters): close connection on Driver Adapters test

jkomyno:
Closes https://github.com/prisma/team-orm/issues/415. Please read the comments therein.

See [comment](https://github.com/prisma/team-orm/issues/415#issuecomment-1790495743).


packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -134,10 +134,21 @@ function setupTestSuiteMatrix(
 
       afterAll(async () => {
```
jkomyno, line 135: Note: this `afterAll` hook here runs for every functional Client test, and actually disables opportunities for other, most specific `afterAll` hooks, as this is the first one that runs, and causes the global `prisma` variable to turn to `undefined` before the next `afterAll` hooks are hit.

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -134,10 +134,21 @@ function setupTestSuiteMatrix(
...
         for (const client of clients) {
+          // @ts-ignore
+          console.log('client', client)
+
+          // @ts-ignore
+          console.log('client._adapter', client._adapter)
```
jkomyno, line 141: TODO: remove this after seeing the logs in CI.

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -138,6 +140,13 @@ function setupTestSuiteMatrix(
             // sometimes we test connection errors. In that case,
             // disconnect might also fail, so ignoring the error here
           })
+
+          if (client._adapter) {
+            await client._adapter.close()
+            console.log('Waiting for Driver Adapters to close the connection...')
+            await setTimeout(10_000)
```
jkomyno, line 147:
```suggestion
            await setTimeout(20_000)
```

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -163,7 +168,7 @@ function setupTestSuiteMatrix(
         delete globalThis['prisma']
         delete globalThis['Prisma']
         delete globalThis['newPrismaClient']
-      }, 180_000)
+      }, 220_000)
```
jkomyno, line 171: This change may no longer be necessary

jkomyno, line 171:
```suggestion
      }, 180_000)
```

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -144,6 +144,11 @@ function setupTestSuiteMatrix(
             // sometimes we test connection errors. In that case,
             // disconnect might also fail, so ignoring the error here
           })
+
+          if (client._adapter) {
```
SevInf, line 148: Can we reuse the link we can from `newDriverAdapter` here, rather than add a property to a client that would be used only in test?

SevInf:
More global question: rather than doing something specific for the test suite, should `client.$disconnect()` call `adapter.close()`? My expectation would be indeed that `$disconnect` will close all connections to the DB.

jkomyno:
> More global question: rather than doing something specific for the test suite, should `client.$disconnect()` call `adapter.close()`? My expectation would be indeed that `$disconnect` will close all connections to the DB.

Btw, now I’ve remembered why I didn’t call .close() directly within $disconnect(): that would technically be a breaking change.
Basically Prisma would mutate some object that it is given.

E.g., imagine a scenario where users create a pg Pool via

```typescript
import { Pool } from 'pg'

const pool = new Pool({ connectionString })
```

then share it with a Prisma Adapter, use Prisma, and detach Prisma via $disconnect:

```typescript
import { PrismaClient } from '.prisma/client'

const adapter = new PrismaPg(pool)

// use prisma...

await prisma.$disconnect()
```

Currently, they can safely keep using the pool connection they had initialised even after Prisma disconnects or goes out of scope: 

```
await prisma.$disconnect()

await pool.query('SELECT 1')
```

If we called `this._adapter.close()` within the Prisma Client definition of `$disconnect()`, this would no longer use the same connection as before.

---

We could however argue that, if `prisma.$disconnect()` is called by the user, we should indeed close the currently open connection even on Driver Adapters.

janpio:
I suggest to split this question out of this PR and handle it in a new issue so this can be merged for now. @SevInf @jkomyno

packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
jkomyno: Actually, I will try removing the `client._adapter.close()` altogether, in the spirit of [this comment of yours](https://prisma-company.slack.com/archives/C04TW4A2H8C/p1699354430085669?thread_ts=1699349717.575299&cid=C04TW4A2H8C), and see whether all latent connections are actually released implicitly.

janpio (approved):
Both test failures are flaky and unrelated

PR #21747, opened 2023-11-02
test(driver-adapters): unskip PlanetScale/LibSQL tests

jkomyno:
This PR:
- closes https://github.com/prisma/team-orm/issues/512 (unskipped with no test changes)
- closes https://github.com/prisma/team-orm/issues/506 (refactored and unskipped)
- closes https://github.com/prisma/team-orm/issues/507 (unskipped with no test changes)

See related [Notion doc](https://www.notion.so/prismaio/Currently-skipped-tests-for-driver-adapters-in-prisma-prisma-b1abfd87118d4ac883a72a93525f5ed7).

### DIFF
--- a/packages/client/tests/functional/0-legacy-ports/query-raw/tests.ts
+++ b/packages/client/tests/functional/0-legacy-ports/query-raw/tests.ts
@@ -1,6 +1,6 @@
 import { copycat } from '@snaplet/copycat'
 
-import { ProviderFlavors, Providers } from '../../_utils/providers'
+import { Providers } from '../../_utils/providers'
 import testMatrix from './_matrix'
 // @ts-ignore
 import type $ from './node_modules/@prisma/client'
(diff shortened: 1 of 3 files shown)


packages/client/tests/functional/0-legacy-ports/query-raw/tests.ts
```diff
@@ -55,7 +55,15 @@ testMatrix.setupTestSuite(
         sqlserver: [{ '': 1 }],
       }
 
-      expect(result).toStrictEqual(results[provider])
+      const resultsByDriverAdapter = {
+        js_planetscale: [{ ':vtg1 /* INT64 */': BigInt('1') }],
+      }
```
janpio, line 60: Why do we expect this weird string as the returned key? Is this a normal thing for the serverless driver? Does this only affect queries like `SELECT 1` or also other, normal queries that would return some data from columns?

aqrln: I assume it's a Vitess thing, and we would have to do the same if we ran tests with Vitess, but let's wait for @jkomyno to confirm.

janpio: Good point that we are not running Vitess tests, so we have no real comparison.

aqrln: Either way, this is not something we control, that's just how it is regardless of whether this comes from Vitess or from the serverless driver. This only affects the raw queries that return values without column names, otherwise normal Prisma queries would not be able to work.

There's also no universally expected behaviour in this case, it is different for each native connector too (Postgres returns `{ '?column?': 1 }` for this query, MySQL without Vitess returns `{ '1': 1 }`, SQL Server returns `{ '': 1 }`).

miguelff (approved):

PR #21626, opened 2023-10-26
feat(CLI): add debug command

Jolg42:
Closes https://github.com/prisma/prisma/issues/21619

Example view from my terminal (slightly outdated already)
Screenshot 2023-10-26 at 12 21 08

PR #21760, opened 2023-11-03
chore(debug): change display & add 1 test

Jolg42:
Follows https://github.com/prisma/prisma/pull/21626

From feedback in [Internal Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1698772227588129)

### DIFF
--- a/packages/cli/src/__tests__/commands/DebugInfo.test.ts
+++ b/packages/cli/src/__tests__/commands/DebugInfo.test.ts
@@ -9,7 +9,7 @@ const ctx = jestContext.new().add(jestConsoleContext()).assemble()
 const originalEnv = { ...process.env }
 
 function cleanSnapshot(str: string): string {
-  str = str.replace(new RegExp('(Path: ).*', 'g'), '$1 REDACTED_PATH')
+  str = str.replace(new RegExp('(Path: ).*', 'g'), '$1REDACTED_PATH')
   return str
 }
 
@@ -76,10 +76,10 @@ describe('debug', () => {
 
     expect(cleanSnapshot(result as string)).toMatchInlineSnapshot(`
       -- Prisma schema --
-      Path:  REDACTED_PATH
+      Path: REDACTED_PATH
 
       -- Local cache directory for engines files --
-      Path:  REDACTED_PATH
+      Path: REDACTED_PATH
 
       -- Environment variables --
       When not set, the line is dimmed and no value is displayed.
@@ -168,10 +168,10 @@ describe('debug', () => {
 
     expect(cleanSnapshot(result as string)).toMatchInlineSnapshot(`
       -- Prisma schema --
-      Path:  REDACTED_PATH
+      Path: REDACTED_PATH
 
       -- Local cache directory for engines files --
-      Path:  REDACTED_PATH
+      Path: REDACTED_PATH
 
       -- Environment variables --
       When not set, the line is dimmed and no value is displayed.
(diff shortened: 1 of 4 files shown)


packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -149,14 +149,106 @@ describe('debug', () => {
...
+    Object.assign(process.env, envVarsSetToEmptyString)
+    // To make sure the terminal is always detected
+    // as non interactive, locally and in CI
+    process.env.TERM = 'dumb'
+
+    const result = await DebugInfo.new().parse([])
+
+    expect(cleanSnapshot(result as string)).toMatchInlineSnapshot(`
+      -- Prisma schema --
+      Path:  REDACTED_PATH
```
janpio, line 171: 2 spaces intentional?

packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -149,14 +149,106 @@ describe('debug', () => {
...
+
+      -- Local cache directory for engines files --
+      Path:  REDACTED_PATH
+
+      -- Environment variables --
+      When not set, the line is dimmed and no value is displayed.
```
janpio, line 177: Does this look intuitive with the backticks when seen in an actual terminal?

packages/cli/src/__tests__/commands/DebugInfo.test.ts
Jolg42: It's fixed, it was just the test snapshots anyway

packages/cli/src/__tests__/commands/DebugInfo.test.ts
Jolg42: Well, we need to print a few different things, keep in mind that env vars are always strings.
- `undefined` value, instead of printing undefined we leave this empty, because technically you could also pass the string `undefined` as a value.
- empty string ``, the env var is set to an empty string value
- some string `1`, the env var is set to a string value
    - some string with only blank spaces `  `   
	- some string `DEBUG="*" command` (random example showing that quotes are theoretically possible)  

So adding backticks helps here to distinguish what's actually the value.

packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -62,24 +62,24 @@ describe('debug', () => {
...
-  it('should succeed when env vars are NOT set', async () => {
+  it('should succeed when env vars are NOT set (undefined)', async () => {
     ctx.fixture('example-project/prisma')
 
     // Make sure all env vars are set to undefined
     const envVarsSetToUndefined = Object.fromEntries(Object.keys(envVars).map((key) => [key, undefined]))
     Object.assign(process.env, envVarsSetToUndefined)
     // To make sure the terminal is always detected
-    // as non interactive, localy and in CI
+    // as non interactive, llocaly and in CI
```
Jolg42, line 72:
```suggestion
    // as non interactive, locally and in CI
```

packages/cli/src/__tests__/commands/DebugInfo.test.ts
janpio: So in the test all the env vars are set to empty string?

Jolg42: I added tests for the different possibilities here, not only empty.

jkomyno (approved):
Nice work!

PR #21792, opened 2023-11-06
ci(renovate): Enable config migration

janpio:
With this configuration Renovate should open PRs to improve the configuration itself, so that we never use outdated keys or similar (as we do right now with `packageNames`, which was migrated to `matchPackageNames`)

see https://docs.renovatebot.com/configuration-options/#configmigration


janpio:
Replaced by https://github.com/prisma/prisma/pull/21793

PR #21793, opened 2023-11-06
ci(renovate): Enable config migration

janpio:
With this configuration Renovate should open PRs to improve the configuration itself, so that we never use outdated keys or similar (as we do right now with packageNames, which was migrated to matchPackageNames)

see https://docs.renovatebot.com/configuration-options/#configmigration

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -10,6 +10,7 @@
   "constraints": {
     "pnpm": "8"
   },
+  "configMigration": true,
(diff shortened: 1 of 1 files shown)


janpio:
At the same time, this might test this renovate feature: https://github.com/renovatebot/renovate/discussions/25604#discussioncomment-7486250

Jolg42 (approved):

PR #21794, opened 2023-11-06
test(client/functional): improve test CLI

Jolg42:
It's a long time painful point of this test CLI.

This adds more parameters and cleans it up a bit.

It also now ignores the order of the search string.
Before, this would not work 
```
pnpm run test:functional:code namOfFile -t "nameOfTestCase" --verbose 
```
Only this was valid, with `namOfFile` in last position
```
pnpm run test:functional:code -t "nameOfTestCase" --verbose namOfFile
```


SevInf (approved):

millsp (approved):
Nice!

PR #21811, opened 2023-11-06
test(client): enable tracing tests with all driver adapters

aqrln:
* Un-skip tracing tests for libSQL/Turso and PlanetScale driver
  adapters.

* Adapt the test expectations to support synthetic BEGIN and COMMIT
  spans produced for driver adapters that manage transactions using
  high-level APIs provided by third-party JavaScript drivers and don't
  emit normal query spans for BEGIN and COMMIT because those statements
  are not executed by Prisma itself.

* Update the `deleteMany` test to support `relationMode = "prisma"` as
  it is used with PlanetScale.

Closes: https://github.com/prisma/team-orm/issues/505

---

Hide whitespace changes for ease of review:
image

### DIFF
--- a/packages/client/tests/functional/tracing/tests.ts
+++ b/packages/client/tests/functional/tracing/tests.ts
@@ -14,7 +14,7 @@ import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
 import { PrismaInstrumentation } from '@prisma/instrumentation'
 import { ClientEngineType, getClientEngineType } from '@prisma/internals'
 
-import { Providers } from '../_utils/providers'
+import { Providers, RelationModes } from '../_utils/providers'
 import { waitFor } from '../_utils/tests/waitFor'
 import { NewPrismaClient } from '../_utils/types'
 import testMatrix from './_matrix'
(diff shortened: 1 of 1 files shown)


aqrln:
Should this PR also close https://github.com/prisma/team-orm/issues/404 or should we also add the `js:query:args` and `js:query:sql` child spans to the synthetic BEGIN/COMMIT spans to close that issue?

cc @millsp @miguelff

packages/client/tests/functional/tracing/tests.ts
```diff
@@ -84,308 +84,378 @@ afterAll(() => {
...
 
-        span['children'] = children
-      }
+    return span
+  }
 
-      return span
+  function txBegin() {
+    if (usesSyntheticTxQueries) {
+      return dbQuery('-- Implicit "BEGIN" query via underlying driver', QueryChildSpans.None)
```
aqrln, line 153: `expect.stringContaining('BEGIN')` would've covered both so only the second argument technically needs to be different but I decided to be explicit so that we can see the tests failing if the behaviour changes.

packages/client/tests/functional/tracing/tests.ts
```diff
@@ -84,308 +84,378 @@ afterAll(() => {
...
-        attributes.model = model
-      }
-      return {
-        name: 'prisma:client:operation',
-        attributes,
-        children,
-      }
+  function txCommit() {
+    if (usesSyntheticTxQueries) {
+      return dbQuery('-- Implicit "COMMIT" query via underlying driver', QueryChildSpans.None)
```
aqrln, line 161: ditto

janpio (commented):
Looks good to me, but I'll let actual engineer approve.

packages/client/tests/functional/tracing/tests.ts
```diff
@@ -84,308 +84,378 @@ afterAll(() => {
...
     })
+  }
 
-    async function waitForSpanTree(expectedTree: Tree): Promise<void> {
-      await waitFor(async () => {
-        await processor.forceFlush()
-        const spans = inMemorySpanExporter.getFinishedSpans()
-        const rootSpan = spans.find((span) => !span.parentSpanId) as ReadableSpan
-        const tree = buildTree(rootSpan, spans)
+  enum QueryChildSpans {
```
millsp, line 109: Maybe this suggestion could make it more explicit?
```suggestion
  enum AdapterQueryChildSpans {
```

millsp (approved):

PR #21883, opened 2023-11-10
fix(adapter-planetscale): Initialize adapter with a `Client`, rather than `Connection`

SevInf:
`Connection` is an abstraction over single DB connection and can not be
used in parallel. `Client` is an abstraction over remote connection
pool.
This change prevents parallel request done on a single connection and
is a proper fix for `invalid sequence` errors.

This is a breaking change for planetscale adapter, so it will need a mention in release notes.

Fix prisma/team-orm#540
Fix prisma/team-orm#494

/integration

### DIFF
--- a/packages/adapter-planetscale/src/planetscale.ts
+++ b/packages/adapter-planetscale/src/planetscale.ts
@@ -147,7 +147,6 @@ class PlanetScaleTransaction extends PlanetScaleQueryable<planetScale.Transactio
 export class PrismaPlanetScale extends PlanetScaleQueryable<planetScale.Client> implements DriverAdapter {
   constructor(client: planetScale.Client) {
     if (client['constructor']?.['name'] !== 'Client') {
-      console.log(Object.prototype.toString.call(client))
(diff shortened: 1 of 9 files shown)


Jolg42 (approved):

packages/adapter-planetscale/src/planetscale.ts
```diff
@@ -146,6 +146,14 @@ class PlanetScaleTransaction extends PlanetScaleQueryable<planetScale.Transactio
 
 export class PrismaPlanetScale extends PlanetScaleQueryable<planetScale.Client> implements DriverAdapter {
   constructor(client: planetScale.Client) {
+    if (client['constructor']?.['name'] !== 'Client') {
+      console.log(Object.prototype.toString.call(client))
+      throw new TypeError(`PrismaPlanetScale must be initialized with and instance of Client:
```
Jolg42, line 151:
```suggestion
      throw new TypeError(`PrismaPlanetScale must be initialized with an instance of Client:
```

Jolg42: Maybe add a test?

PR #21891, opened 2023-11-10
ci: Allow to use custom engine branch in client tests

SevInf:
Allows to use engine from a branch in engines repo without.

How does it work: if body of the PR contains `[slash]engine-branch [branch]` command, before running TS tests we'll pull in specified branch, build the engine for linux, mac and windows and set up environment variables to point to those versions instead of the ones installed from `npm`.

Build is cached, so unless engine branch receives new commits, engines won't be rebuilt on subsequent runs. Cargo dependencies are cached separately, so even if there is new commit, but `Cargo.lock` did not change, future build will be faster.  In worst possible case, engine build takes around 20min, but because of caches it is usually faster.

Number of tests are skipped if custom engine is used: for the most part, they expect to find engine at a certain path or expect certain version output. However, there is at least one actual bug (#21926).

If custom engine is used, separate failed status check and a comment on PR with reminder to remove the command before merging will be added. For people with admin permissions: it might be a good idea to add this check to required checcks in branch protection rules.

Some caveats: custom branch does not work for benchmark workflow at the moment. For it to work, we'll need to merge it with `test` or `test-template` workflows and I am not sure we want to do this now.

Close prisma/team-orm#538
/engine-branch test-client-ci


.github/workflows/test.yml
```diff
@@ -101,11 +101,64 @@ jobs:
...
+            <!-- custom-engine -->
+            🟢 All good, this PR is no longer uses custom engine.
+          edit-mode: replace
+      - name: Block merge
+        if: ${{ needs.build_custom_engine.outputs.engineHash != '' }}
+        run: |
+          echo "Custom engine is used, failing the job"
+          echo "Don't forget to remove /engine-branch command from PR body before merge"
+          exit 1
+
```
miguelff, line 153: This is neat, I might backport this to engines to prevent merges.

packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -27,89 +28,93 @@ function sendRequest(message: any): Promise<any> {
 }
 
 let studio: Studio
-describe('studio with alternative urls and prisma://', () => {
-  afterEach(() => {
-    // Back to original env vars
-    process.env = { ...originalEnv }
-  })
+// Prisma Studio ignores env vars for overriding engine paths, skipping test for now
+describeIf(!process.env.PRISMA_QUERY_ENGINE_LIBRARY && !process.env.PRISMA_QUERY_ENGINE_BINARY)(
```
miguelff, line 32: Should this condition be a helper that denotes the intent?

miguelff (approved):

packages/cli/src/__tests__/commands/Studio.test.ts
SevInf: This condition is slightly different in every test, not sure single helper would cut it.

.github/workflows/build-engine-branch.yml
```diff
@@ -0,0 +1,98 @@
...
+        with:
+          path: |
+            target/release/schema-engine
+            target/release/schema-engine.exe
+            target/release/query-engine
+            target/release/query-engine.exe
+            target/release/libquery_engine.so
+            target/release/libquery_engine.dylib
+            target/release/query_engine.dll
+          key: ${{ runner.os }}-engine-${{ steps.engine-hash.outputs.engineHash }}-${{ github.event.number }}
```
miguelff, line 74: If two different PRs need the same version of the engine and this was built already, what's the reason to not reuse it? In other words, why adding `github.event.number`  to the cache key?

SevInf: My only thinking for that case was that in the case they are isolated by the PR, it makes it easier to find where and how the engine was actually built - it is always one of the commits on the same pull request. Where in other case, it might be actually impossible to figure out where we are getting the cache from without additional work. Reusing the cache will save us some time though, so in the end I don't have strong opinion. If you prefer to reuse the cache instead I can change it.

miguelff: No, your reasoning is perfectly valid, I think it's fine as it is, and then we can reevaluate whether we build the same engine from different PRs often. (I don't think is as likely to happen)

Thanks for explaining.

packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -27,89 +28,93 @@ function sendRequest(message: any): Promise<any> {
 }
 
 let studio: Studio
-describe('studio with alternative urls and prisma://', () => {
-  afterEach(() => {
-    // Back to original env vars
-    process.env = { ...originalEnv }
-  })
+// Prisma Studio ignores env vars for overriding engine paths, skipping test for now
```
millsp, line 31: Would it help to store the engine path statically at build time (eg. inject the env var path contents at build time)?

millsp: Additionally, making it static may allow not have have to touch any test in general, that would be similar to the approach I took here https://github.com/prisma/prisma/blob/551986bc5c4900eaf3d71f2725b327ccef7154e1/packages/fetch-engine/src/download.ts/#L62-L67, basically is `enginesOverride` exists, we will copy the engine into the `.cache/prisma/_local_` folder, making it transparent to any test that uses it.

SevInf: For quite a bunch of skipped tests that will indeed help to unskip them. There are few of the tests which test that reported engine version is the same as specified in `package.json` and those would have to have stay skipped.

millsp (approved):

.github/workflows/build-engine-branch.yml
```diff
@@ -0,0 +1,98 @@
...
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Build custom engines with cargo
+        if: steps.artifacts-cache.outputs.cache-hit != 'true'
+        shell: bash
+        run: |
+          cargo build --release \
+            -p query-engine \
+            -p query-engine-node-api \
+            -p schema-engine-cli
```
jkomyno, line 98: Let's remember to build `query-engine-wasm` as well :)

jkomyno (approved):

.github/workflows/build-engine-branch.yml
jkomyno: Let's take that to a separate future PR, and use a custom Wasm for testing for the time being.

PR #21914, opened 2023-11-13
chore(reproductions): rename dir to sandbox & ignore for CI / Renovate

Jolg42:
Rename directory to "sandbox" as to not confuse it with bug reproductions

+ avoids Renovate PRs like https://github.com/prisma/prisma/pull/21899

This directory is only meant for local development and not tested.


sandbox/README.md
```diff
@@ -0,0 +1,5 @@
+## Reproductions
```
millsp, line 1:
```suggestion
## Sandbox
```

PR #21918, opened 2023-11-13
fix(adapter-neon, adapter-pg): support citext type

### LINKED ISSUE #21807: Using `citext` fields with neon database driver causes conversion error
### Bug description

Using the postgres extension `Citext` on a field causes the error:

```
Error converting field "slug" of expected non-nullable type "String", found incompatible value of "someslug".
```

### How to reproduce

1. Use citext postgres extension on any field in a model
2. Fetch data from that model
3. See error:

```
Error converting field "slug" of expected non-nullable type "String", found incompatible value of "someslug".
```

Some debug logs:
```
  prisma:client  Prisma Client call: +670ms
  prisma:client  prisma.someModel.findMany({
  where: {
    slug: {
      equals: "someslug"
    }
  },
  select: {
    id: true,
    slug: true
  }
}) +0ms
  prisma:client  Generated request: +0ms
  prisma:client  {
  "modelName": "SomeModel",
  "action": "findMany",
  "query": {
    "arguments": {
      "where": {
        "slug": {
          "equals": "someslug"
        }
      }
    },
    "selection": {
      "id": true,
      "slug": true
    }
  }
}
 +0ms
  prisma:client:libraryEngine  sending request, this.libraryStarted: false +669ms
  prisma:client:engines:resolveEnginePath  enginePath /Users/markus/prisma/node_modules/.pnpm/@prisma+client@5.5.2_prisma@5.5.2/node_modules/.prisma/client/libquery_engine-darwin-arm64.dylib.node +0ms
  prisma:client:libraryEngine  Using driver adapter: %O {
  errorRegistry: oc { registeredErrors: [] },
  queryRaw: [AsyncFunction (anonymous)],
  executeRaw: [AsyncFunction (anonymous)],
  flavour: 'postgres',
  startTransaction: [AsyncFunction: startTransaction],
  close: [AsyncFunction (anonymous)]
} +8ms
  prisma:client:libraryEngine  library starting +17ms
  prisma:client:libraryEngine  library started +2ms
  prisma:driver-adapter:neon [js::query_raw] {
  prisma:driver-adapter:neon   sql: 'SELECT "public"."SomeModel"."id", "public"."SomeModel"."slug" FROM "public"."SomeModel" WHERE "public"."SomeModel"."slug" = $1 OFFSET $2',
  prisma:driver-adapter:neon   args: [ 'someslug', 0 ]
  prisma:driver-adapter:neon } +0ms
  prisma:client:request_handler  PrismaClientKnownRequestError: Error converting field "slug" of expected non-nullable type "String", found incompatible value of "someslug".
    at Zt (/Users/markus/prisma/node_modules/.pnpm/@prisma+client@5.5.2_prisma@5.5.2/node_modules/@prisma/client/runtime/library.js:36:253)
    at mr.buildQueryError (/Users/markus/prisma/node_modules/.pnpm/@prisma+client@5.5.2_prisma@5.5.2/node_modules/@prisma/client/runtime/library.js:115:1216)
    at mr.request (/Users/markus/prisma/node_modules/.pnpm/@prisma+client@5.5.2_prisma@5.5.2/node_modules/@prisma/client/runtime/library.js:114:4091)
    at async Object.singleLoader (/Users/markus/prisma/node_modules/.pnpm/@prisma+client@5.5.2_prisma@5.5.2/node_modules/@prisma/client/runtime/library.js:123:5333)
    at async Cn.request (/Users/markus/prisma/node_modules/.pnpm/@prisma+client@5.5.2_prisma@5.5.2/node_modules/@prisma/client/runtime/library.js:123:5825)
    at async l (/Users/markus/prisma/node_modules/.pnpm/@prisma+client@5.5.2_prisma@5.5.2/node_modules/@prisma/client/runtime/library.js:128:9968)
    at async /Users/markus/prisma/script.ts:13:5 {
  code: 'P2032',
  clientVersion: '5.5.2',
  meta: { field: 'slug', expected_type: 'String', found: 'someslug' }
} +0ms
```

### Expected behavior

Prisma should correctly convert Citext fields to strings

### Prisma information

```prisma
datasource postgresql {
  url               = env("DATABASE_URL")
  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
  directUrl         = env("DIRECT_URL")
}

generator client {
}

model SomeModel {
  id                                     String                 @id @default(cuid())
  slug                                   String                 @unique @default("") @postgresql.Citext
}
```

```ts
await prisma.someModel.findMany({
      where: {
        slug: {
          equals: 'someslug',
        },
      },
      select: {
        id: true,
        slug: true,
      },
    })
  );
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL v14
- Node.js version: v18

### Prisma Version

```
prisma                  : 5.5.2
@prisma/client          : 5.5.2
Query Engine (Node-API) : libquery-engine aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a (at ../../node_modules/.pnpm/@prisma+engines@5.5.2/node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a (at ../../node_modules/.pnpm/@prisma+engines@5.5.2/node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.5.1-1.aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a
Default Engines Hash    : aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a
```

aqrln:
Fixes: https://github.com/prisma/prisma/issues/21807

### DIFF
--- a/packages/adapter-neon/src/conversion.ts
+++ b/packages/adapter-neon/src/conversion.ts
@@ -2,41 +2,9 @@ import { types } from '@neondatabase/serverless'
 import { type ColumnType, ColumnTypeEnum, JsonNullMarker } from '@prisma/driver-adapter-utils'
 import { parse as parseArray } from 'postgres-array'
 
-const ScalarColumnType = types.builtins
+import { ArrayColumnType, maxSystemCatalogOID, PgType } from './pg-types'
 
-/**
- * PostgreSQL array column types (not defined in ScalarColumnType).
- *
- * See the semantics of each of this code in:
- *   https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_type.dat
- */
-const ArrayColumnType = {
-  BIT_ARRAY: 1561,
-  BOOL_ARRAY: 1000,
-  BYTEA_ARRAY: 1001,
-  BPCHAR_ARRAY: 1014,
-  CHAR_ARRAY: 1002,
-  CIDR_ARRAY: 651,
-  DATE_ARRAY: 1182,
-  FLOAT4_ARRAY: 1021,
-  FLOAT8_ARRAY: 1022,
-  INET_ARRAY: 1041,
-  INT2_ARRAY: 1005,
-  INT4_ARRAY: 1007,
-  INT8_ARRAY: 1016,
-  JSONB_ARRAY: 3807,
-  JSON_ARRAY: 199,
-  MONEY_ARRAY: 791,
-  NUMERIC_ARRAY: 1231,
-  OID_ARRAY: 1028,
-  TEXT_ARRAY: 1009,
-  TIMESTAMP_ARRAY: 1115,
-  TIME_ARRAY: 1183,
-  UUID_ARRAY: 2951,
-  VARBIT_ARRAY: 1563,
-  VARCHAR_ARRAY: 1015,
-  XML_ARRAY: 143,
-}
+const ScalarColumnType = types.builtins
 
 export class UnsupportedNativeDataType extends Error {
   // map of type codes to type names
@@ -157,10 +125,10 @@ export class UnsupportedNativeDataType extends Error {
 
   type: string
 
-  constructor(code: number) {
+  constructor(type: PgType) {
     super()
-    this.type = UnsupportedNativeDataType.typeNames[code] || 'Unknown'
-    this.message = `Unsupported column type ${this.type}`
+    this.type = type.name ?? UnsupportedNativeDataType.typeNames[type.id] ?? 'Unknown'
+    this.message = `Unsupported column type ${this.type} (OID=${type.id})`
   }
 }
 
@@ -169,8 +137,8 @@ export class UnsupportedNativeDataType extends Error {
  * module to see how other attributes of the field packet such as the field length are used to infer
  * the correct quaint::Value variant.
  */
-export function fieldToColumnType(fieldTypeId: number): ColumnType {
-  switch (fieldTypeId) {
+export function fieldToColumnType(fieldType: PgType): ColumnType {
+  switch (fieldType.id) {
     case ScalarColumnType['INT2']:
     case ScalarColumnType['INT4']:
       return ColumnTypeEnum.Int32
@@ -251,11 +219,14 @@ export function fieldToColumnType(fieldTypeId: number): ColumnType {
     case ArrayColumnType.OID_ARRAY:
       return ColumnTypeEnum.Int64Array
     default:
-      if (fieldTypeId >= 10000) {
-        // Postgres Custom Types
+      // Postgres Custom Types
+      if (fieldType.id > maxSystemCatalogOID) {
+        if (fieldType.name !== undefined && ['citext', 'ltree', 'lquery', 'ltxtquery'].includes(fieldType.name)) {
+          return ColumnTypeEnum.Text
+        }
         return ColumnTypeEnum.Enum
       }
-      throw new UnsupportedNativeDataType(fieldTypeId)
+      throw new UnsupportedNativeDataType(fieldType)
   }
 }
 
--- a/packages/adapter-neon/src/conversion.ts
+++ b/packages/adapter-neon/src/conversion.ts
@@ -1,173 +1,47 @@
 import { types } from '@neondatabase/serverless'
-import { type ColumnType, ColumnTypeEnum, JsonNullMarker } from '@prisma/driver-adapter-utils'
+import { type ColumnType, ColumnTypeEnum, err, JsonNullMarker, ok, type Result } from '@prisma/driver-adapter-utils'
 import { parse as parseArray } from 'postgres-array'
 
 import { ArrayColumnType, maxSystemCatalogOID, PgType } from './pg-types'
 
 const ScalarColumnType = types.builtins
 
-export class UnsupportedNativeDataType extends Error {
-  // map of type codes to type names
-  static typeNames: { [key: number]: string } = {
-    16: 'bool',
-    17: 'bytea',
-    18: 'char',
-    19: 'name',
-    20: 'int8',
-    21: 'int2',
-    22: 'int2vector',
-    23: 'int4',
-    24: 'regproc',
-    25: 'text',
-    26: 'oid',
-    27: 'tid',
-    28: 'xid',
-    29: 'cid',
-    30: 'oidvector',
-    32: 'pg_ddl_command',
-    71: 'pg_type',
-    75: 'pg_attribute',
-    81: 'pg_proc',
-    83: 'pg_class',
-    114: 'json',
-    142: 'xml',
-    194: 'pg_node_tree',
-    269: 'table_am_handler',
-    325: 'index_am_handler',
-    600: 'point',
-    601: 'lseg',
-    602: 'path',
-    603: 'box',
-    604: 'polygon',
-    628: 'line',
-    650: 'cidr',
-    700: 'float4',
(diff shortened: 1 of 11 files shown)


packages/adapter-neon/src/conversion.ts
```diff
@@ -233,29 +75,35 @@ export function fieldToColumnType(fieldTypeId: number): ColumnType {
...
     case ArrayColumnType.OID_ARRAY:
-      return ColumnTypeEnum.Int64Array
+      return ok(ColumnTypeEnum.Int64Array)
     default:
-      if (fieldTypeId >= 10000) {
-        // Postgres Custom Types
-        return ColumnTypeEnum.Enum
+      // Postgres Custom Types
+      if (fieldType.id > maxSystemCatalogOID) {
+        if (fieldType.name !== undefined && ['citext', 'ltree', 'lquery', 'ltxtquery'].includes(fieldType.name)) {
```
aqrln, line 98: The list simply mirrors https://github.com/sfackler/rust-postgres/blob/10edbcb46c44933417e8d2e7a1c1d63c4119beb3/postgres-types/src/lib.rs#L707-L710, not actually tested for any types except `citext`.

packages/driver-adapter-utils/src/result.ts
```diff
@@ -40,3 +40,19 @@ export function err<T>(error: Error): Result<T> {
     },
   }
 }
+
+/**
+ * Collect an array of results into an array of values wrapped in a single result.
+ * Equivalent to `results.reduce((a, b) => a.flatMap((a) => b.map((b) => a.concat(b))), ok([] as T[]))`.
+ */
+export function sequence<T>(results: Result<T>[]): Result<T[]> {
```
aqrln, line 48: what name would y'all prefer?

packages/adapter-pg/src/pg-types.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (id <= maxSystemCatalogOID && knownSystemTypeNames[id]) {
+      return ok({
+        id,
+        name: knownSystemTypeNames[id],
+      })
+    }
+
+    const result = (
+      await this.client.performIO({
+        sql: 'SELECT typname FROM pg_catalog.pg_type WHERE oid = $1',
```
miguelff, line 86: Hum, this impacts coldstart performance. In order to support citext (or any other OID added by an extension) we do need to query the catalog indeed, however, I think we can do less queries provided that each user query selects various columns, and we have information about the oids for such columns as the user query is returned.  The algorithm will follow:

* user query returns oid set `O = { o1, o2, .., on. }`
* we look in the cache for individual oid type names, `oi | i belong {1..n}` was already in the cache from a previous query. 
* we query O - oi, and cache the results. individually. (Not their promises, tho)

What do you think? Will that improve cold start?

packages/adapter-neon/src/pg-types.ts
```diff
@@ -0,0 +1,226 @@
+import type neon from '@neondatabase/serverless'
```
miguelff, line 1: You already extracted this into its own file. Would it make sense to extract it into tis own file shared by both neon and pg?

Is it possible to do so via npm / pnpm at build time so both resulting packages contain a copy of the file, without the need to ship an extra package?

packages/driver-adapter-utils/src/result.ts
miguelff: `flatten`

packages/adapter-pg/src/pg-types.ts
miguelff: If the schema contains 4 tables each with 5 columns. And user queries are "SELECT * FROM...", instead of 20 roundtrips for the enums this would lead to 4 roundtrips.

packages/adapter-pg/src/conversion.ts
```diff
@@ -234,29 +76,35 @@ export function fieldToColumnType(fieldTypeId: number): ColumnType {
...
-      return ColumnTypeEnum.Int64Array
+      return ok(ColumnTypeEnum.Int64Array)
     default:
-      if (fieldTypeId >= 10000) {
-        // Postgres Custom Types
-        return ColumnTypeEnum.Enum
+      // Postgres Custom Types
+      if (fieldType.id > maxSystemCatalogOID) {
+        if (fieldType.name !== undefined && ['citext', 'ltree', 'lquery', 'ltxtquery'].includes(fieldType.name)) {
+          return ok(ColumnTypeEnum.Text)
```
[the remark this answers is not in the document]
aqrln: I just copied this list [from tokio-postgres](https://github.com/sfackler/rust-postgres/blob/10edbcb46c44933417e8d2e7a1c1d63c4119beb3/postgres-types/src/lib.rs#L707-L710) without thinking much about it but yeah, it's better to leave just `citext` here and then add support for other types separately with tests. I left a comment about it above as well: https://github.com/prisma/prisma/pull/21918#discussion_r1395672590

packages/adapter-pg/src/conversion.ts
```diff
@@ -1,206 +1,48 @@
...
-}
-
 /**
  * This is a simplification of quaint's value inference logic. Take a look at quaint's conversion.rs
  * module to see how other attributes of the field packet such as the field length are used to infer
  * the correct quaint::Value variant.
  */
-export function fieldToColumnType(fieldTypeId: number): ColumnType {
-  switch (fieldTypeId) {
+export function fieldToColumnType(fieldType: PgType): Result<ColumnType> {
```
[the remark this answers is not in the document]
aqrln: no strong preference either way, I can change it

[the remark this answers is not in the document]
miguelff: Forget this one, I don't like the breks either.

packages/adapter-pg/src/pg-types.ts
aqrln: That makes sense, we could indeed make less queries by looking up multiple OIDs at once.

packages/driver-adapter-utils/src/result.ts
aqrln: to me `flatten` sounds more like turning multiple nested results into a single one, rather than transposing an array of results into a result of array

aqrln: I don't know any standard name for this kind of transformation, [hoogle](https://hoogle.haskell.org) told me it's called `sequence` in haskell when I typed the signature in, so I used that, but I'm not a fan of the name

aqrln: in rust that would technically be `collect` but it's also very different of how generic it is, but maybe we could call it `collectResults` or `collectIntoResult` here?

miguelff: I know it's technically not exactly flatten but it's a simple good enough approximation to have a quick mental model, considering Result a higher-order wrapping of the value returned. Having said that, I don't have strong opinions. Go with what makes more sense to you.

janpio (changes requested):
Will write a message in #driver-adapters which we should clarify first before proceeding here.

Update: https://prisma-company.slack.com/archives/C04TW4A2H8C/p1700503909449549

aqrln:
Alternative approach: https://github.com/prisma/prisma/pull/22239

Issue #21221, opened 2023-09-25
`prisma_pool_connections_idle` reports misleading number


janpio:
https://github.com/importcjj/mobc/issues/82

janpio:
https://github.com/importcjj/mobc/pull/86 happened 👀

silentroach:
after upgrade to 5.6.0 our idle counter looks like this (with connection_limit = 10):

```
prisma_pool_connections_busy{app="xxx"} 0
prisma_pool_connections_idle{app="xxx"} 201
prisma_pool_connections_open{app="xxx"} 10
```

janpio:
Hmm, are you seeing a regression or continued bug here @silentroach? If so, could you open a _new_ issue so we can potentially reproduce this and figure out what is going on? (WIth a new issue, others are also more probably to chime in.) Thanks!

PR #21937, opened 2023-11-14
test(metrics): expect 0 or 1 for `prisma_pool_connections_idle`

Jolg42:
Related
https://github.com/prisma/prisma-engines/pull/4379 https://github.com/prisma/prisma/issues/21221

### DIFF
--- a/packages/client/tests/functional/metrics/enabled/tests.ts
+++ b/packages/client/tests/functional/metrics/enabled/tests.ts
@@ -271,7 +271,7 @@ testMatrix.setupTestSuite(
               labels: {},
               // This can sometimes be reported as 1 or 0,
               // as metrics are reported asynchronously.
-              value: 0 || 1,
+              value: expect.toBeOneOf([0, 1]),
               description: 'The number of pool connections that are not busy running a query',
             },
             {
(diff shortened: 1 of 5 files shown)


miguelff (approved):

janpio (approved):

packages/client/tests/functional/metrics/enabled/tests.ts
```diff
@@ -269,7 +269,9 @@ testMatrix.setupTestSuite(
             {
               key: 'prisma_pool_connections_idle',
               labels: {},
-              value: expect.any(Number),
+              // This can sometimes be reported as 1 or 0,
+              // as metrics are reported asynchronously.
+              value: 0 || 1,
```
SevInf, line 274: That will always evaluate to 1, so if the actual value would be 0 test won't pass.
It does not seem like there is built-in asymetric matcher for `either`, so it looks like we need to write our own, unfortunately.

SevInf (changes requested):
I think there is a bug in the code, it will need fixing before merging.

packages/client/tests/functional/metrics/enabled/tests.ts
Jolg42: True 😅 I also didn't see a built in but I'm also not sure is it will actually be flaky, it was reliably `1` when running locally

Jolg42: I pushed a fix

SevInf (approved):

PR #21939, opened 2023-11-14
fix(adapter-planetscale): Use instanceof for checking for `Client` instance

SevInf:
Prevoulsly, we tried to check for constructor name, which theoretically
should've been more reliable and work with several copies of
`@planetscale/database`. However, name gets mangled during minification
and check fails. Since minification is pretty common usecase (unlike
having multiple copies of the `@planetscale/database`), reverting back
to `instanceof` check.

### DIFF
--- a/packages/adapter-planetscale/src/planetscale.ts
+++ b/packages/adapter-planetscale/src/planetscale.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/require-await */
-import planetScale from '@planetscale/database'
+import * as planetScale from '@planetscale/database'
 import type {
   DriverAdapter,
   Query,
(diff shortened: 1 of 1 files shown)


Jolg42:
Note: found via https://github.com/prisma/ecosystem-tests/pull/4226

packages/adapter-planetscale/src/planetscale.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/require-await */
-import planetScale from '@planetscale/database'
+import * as planetScale from '@planetscale/database'
```
Jolg42, line 2: add a comment?

packages/adapter-planetscale/src/planetscale.ts
```diff
@@ -146,7 +146,7 @@ class PlanetScaleTransaction extends PlanetScaleQueryable<planetScale.Transactio
 
 export class PrismaPlanetScale extends PlanetScaleQueryable<planetScale.Client> implements DriverAdapter {
   constructor(client: planetScale.Client) {
-    if (client['constructor']?.['name'] !== 'Client') {
+    if (!(client instanceof planetScale.Client)) {
```
Jolg42, line 149: add a comment?

Since we don't have a test for the minification in this repo, it would help I think

Jolg42 (approved):

cstrnt:
Heya, this has caused some issues (#21956) for me :/

Jolg42:
Thanks! Let's continue the discussion in https://github.com/prisma/prisma/issues/21956
We definitely want to fix this.

PR #21962, opened 2023-11-15
fix: Add proper export maps for ESM packages

### LINKED ISSUE #21956: Planetscale Driver doesn't work anymore after updating to 5.6.0
### Bug description

I've updated all my dependencies (`@prisma/adapter-planetscale`, `@prisma/client` and `prisma`) to v5.6.0.

I'm now getting an error that the given PlanetScale client isn't an instance of of the class `Client` from `@planetscale/database`

The error is thrown here https://github.com/prisma/prisma/commit/29bd6252d2692c300fa2d81024837c5eff2b3f2f because the `instanceof` check fails. I've checked that I have the same version of `@planetscale/database` locally installed

### How to reproduce

1. Update given packages to 5.6.0
2. Install all packages with pnpm v8.6.9 (should potentially break with other versions as well)
3. Update the code according to the migration guide (https://github.com/prisma/prisma/releases/tag/5.6.0)
4. Breaks

### Expected behavior

Prisma should work. The error is invalid since the `Client` instance I'm passing to `PrismaPlanetScale` is in-fact an instance of the Client class.

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
```

### Environment & setup

- OS: MacOS
- Database: MySql / Vitesse
- Node.js version: v18.17.0
- Package Manager: pnpm v8.6.9

### Prisma Version

```
5.6.0
```

SevInf:
Adapter packages included `module` entires for bundlers, but not
`exports` for unbundled node.js code. Therefore, in that context
adapters were always treated as common.js, which means that they might
not always be compatible with ESM drivers.

Build process for the adapter packages was adjusted to emit separate
declaration file for ESM files. This is a TS 4.7+ requirement.

Fix #21956
/integration


Jolg42 (approved):

helpers/compile/configs.ts
```diff
@@ -18,7 +18,7 @@ export const adapterConfig: BuildOptions[] = [
     entryPoints: ['src/index.ts'],
     outfile: 'dist/index',
     outExtension: { '.js': '.mjs' },
-    emitTypes: false,
```
millsp, line 21: Are the types different or is it just about the extension of the file (in which case they could be duplicated instead of re-bundled)?

millsp (approved):

helpers/compile/configs.ts
SevInf: I think right now they are identical. They have to be separate files though, pointing both of them to the same `.d.ts` file will not work according to TS docs.

SevInf: Sure, created a follow up issue https://github.com/prisma/team-orm/issues/589

PR #21975, opened 2023-11-15
chore: Remove `--default-authentication-plugin=mysql_native_password` from MySQL docker

janpio:
We were using `--default-authentication-plugin=mysql_native_password` in the Docker MySQL config, which was not needed for Prisma itself but the `mariadb` Node driver we use to sometimes talk to MySQL and MariaDB.

I modified the `mariadb.createConnection` to be able to work with the new default auth plugin of MySQL.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -95,7 +95,7 @@ services:
 
   mysql:
     image: mysql:8.0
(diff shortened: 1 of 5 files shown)


Jolg42 (approved):
Nice! ✨

PR #21977, opened 2023-11-15
test(client): add --engine-type flag in test cli

millsp:
closes https://github.com/prisma/team-orm/issues/550


packages/client/tests/functional/_utils/getTestSuiteInfo.ts
```diff
@@ -182,15 +185,22 @@ export function getTestSuiteSchema(suiteMeta: TestSuiteMeta, matrixOptions: Name
...
   if (previewFeatureMatch === null && defaultGeneratorMatch !== null) {
-    schema = schema.replace(defaultGeneratorMatch[0], `${defaultGeneratorMatch[0]}\n${previewFeaturesStr}`)
+    const replacement = `${defaultGeneratorMatch[0]}\n${previewFeaturesStr}`
+    schema = schema.replace(defaultGeneratorMatch[0], replacement)
+  }
+
+  // if an engine type is specified, append it to the default generator block
+  if (matrixOptions.engineType !== undefined && defaultGeneratorMatch !== null) {
+    const replacement = `${defaultGeneratorMatch[0]}\nengineType = "${engineType}"`
+    schema = schema.replace(defaultGeneratorMatch[0], replacement)
```
millsp, line 195: This part is optional, I decided to properly write `engineType` into the schema. Optional in the sense that I could have used the env var instead. I prefer if it is as close as possible to reality.

packages/client/src/generation/generateClient.ts
```diff
@@ -317,7 +317,9 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   if (getClientEngineType(generator) === ClientEngineType.Wasm) {
     const queryEngineWasmFilePath = path.join(runtimeSourceDir, 'query-engine.wasm')
     const queryEngineWasmTargetPath = path.join(finalOutputDir, 'query-engine.wasm')
-    await fs.copyFile(queryEngineWasmFilePath, queryEngineWasmTargetPath)
+
+    const copyOrSymlink = testMode ? fs.symlink : fs.copyFile
+    await copyOrSymlink(queryEngineWasmFilePath, queryEngineWasmTargetPath)
```
millsp, line 322: Avoids spending time on making copies.

packages/client/tests/functional/binary-engine/reuse-binary-engine/prisma/_schema.ts
```diff
@@ -4,7 +4,6 @@ import testMatrix from '../_matrix'
 export default testMatrix.setupSchema(({ provider }) => {
   return /* Prisma */ `
     generator client {
-      engineType = "binary"
```
millsp, line 7: Small cleanup, this should never be hardcoded (and was ignored because of our reliance on the engine being selected via the env var).

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -40,6 +40,10 @@ export function getTestSuitePlan(
     return getExpandedTestSuitePlanWithProviderFlavors(config)
   })
 
+  expandedSuiteConfigs.forEach((config) => {
+    config.matrixOptions.engineType ??= testCliMeta.engineType
```
millsp, line 44: Makes this information available in the test suite itself (in case we need to skip tests for any reason)

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -139,24 +141,29 @@ async function main(): Promise<number | void> {
...
+      throw new Error(`Unknown flavors: ${unknownFlavors.join(', ')}`)
     }
 
     if (providerFlavors.some(isDriverAdapterProviderFlavor)) {
       jestCli = jestCli.withArgs(['--runInBand'])
       jestCli = jestCli.withEnv({ PRISMA_DISABLE_QUAINT_EXECUTORS: 'true' })
       jestCli = jestCli.withEnv({ TEST_REUSE_DATABASE: 'true' })
 
-      if (args['--data-proxy'] || process.env.PRISMA_CLIENT_ENGINE_TYPE === 'binary') {
+      if (args['--data-proxy'] || (args['--engine-type'] || process.env.PRISMA_CLIENT_ENGINE_TYPE) === 'binary') {
```
millsp, line 155:
```suggestion
      if (args['--data-proxy'] || args['--engine-type'] === 'binary' || process.env.PRISMA_CLIENT_ENGINE_TYPE === 'binary') {
```

packages/client/tests/functional/_utils/getTestSuiteInfo.ts
SevInf: I am fine with either, but can we do it in the same way for binary and wasm engines too? Right now binary uses env var.

millsp: Definitely. Part of my wider goal doing this also was that generally I consider the `engineType` to be one dimension of the  testing matrix. For a long time it has been an env var managed by the ci, that feels weird as other matrix elements are managed by the setup. I have put more structure in this matrix, making it clear and accessible on the type level and at runtime in tests, and making `getTestSuitePlan` the source of truth.

In other words, with the related changes of the PR, instead of doing
```ts
if (getClientEngineType() !== ClientEngineType.Binary) {
```
then you could have
```ts
if (suiteConfig.engineType === ClientEngineType.Library) {
```
(`suiteConfig` now has `provider`, `providerFlavor`, `relationMode`, and `engineType`)

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -202,8 +210,12 @@ async function main(): Promise<number | void> {
     jestCli = jestCli.withArgs(['--testPathIgnorePatterns', 'relationMode-in-separate-gh-action'])
   }
 
+  if (process.env.PRISMA_CLIENT_ENGINE_TYPE && !args['--engine-type']) {
+    throw new Error('Functional tests expect --engine-type to be explicitly set, not via env var')
+  }
+
   try {
-    if (args['-u']) {
+    if (args['--updateSnapshot']) {
```
millsp, line 218: `-u` did not work anymore, maybe because of a prev. PR. Fixed it.

SevInf (approved):

packages/client/helpers/functional-test/run-tests.ts
Jolg42: Most likely my PR https://github.com/prisma/prisma/pull/21794, thanks!

packages/client/src/generation/generateClient.ts
Jolg42: I'm not sure if that works well on Windows 🤔

Jolg42 (approved):

packages/client/src/generation/generateClient.ts
millsp: I think you might be right. I recall issues with rights to create symlinks on Windows. Once we hook it on CI if there are any issues, I'll enable `copyFile` for windows instead.

PR #21983, opened 2023-11-16
fix(adapter-pg): Fix ESM bundle

SevInf:
Fix prisma/team-orm#592
/integration


miguelff (approved):
Unblocking merge for when (if) ecosystem tests pass.

PR #21989, opened 2023-11-16
test(metrics): avoid flaky test on `prisma_pool_connections_idle`

Jolg42:
There is an issue to investigate how this can happen: https://github.com/prisma/team-orm/issues/587

See example
https://github.com/prisma/prisma/actions/runs/6892261253/job/18749160153?pr=21987#step:8:292
Tests / Client func/legacy-notypes (4/5, library, 18)
```
     -     "value": toBeOneOf<0,1>,
    +     "value": 2,
```

### DIFF
--- a/packages/client/tests/functional/metrics/enabled/tests.ts
+++ b/packages/client/tests/functional/metrics/enabled/tests.ts
@@ -269,9 +269,11 @@ testMatrix.setupTestSuite(
             {
               key: 'prisma_pool_connections_idle',
               labels: {},
-              // This can sometimes be reported as 1 or 0,
+              // This can sometimes be reported as 0, 1 or 2,
               // as metrics are reported asynchronously.
-              value: expect.toBeOneOf([0, 1]),
+              // Note: We want to investigate why these different values are reported in our test setup
+              // https://github.com/prisma/team-orm/issues/587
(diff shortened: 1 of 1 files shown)


packages/client/tests/functional/metrics/enabled/tests.ts
```diff
@@ -269,9 +269,11 @@ testMatrix.setupTestSuite(
             {
               key: 'prisma_pool_connections_idle',
               labels: {},
-              // This can sometimes be reported as 1 or 0,
+              // This can sometimes be reported as 0, 1 or 2,
               // as metrics are reported asynchronously.
-              value: expect.toBeOneOf([0, 1]),
+              // Note: We want to investigate why these different values are reported in our test setup
+              // https://github.com/prisma/team-orm/issues/587
+              value: expect.toBeOneOf([0, 1, 2]),
```
millsp, line 276: Do you think we could expect any number here, or is it fully limited to this set?

Jolg42: That's something to investigate, my guess is that this should cover most cases, because that would be weird if not but possible.

millsp (approved):

PR #22005, opened 2023-11-17
fix(tests): disconnect in seed script in issues/21552 regression test

aqrln:
Fixes the test failing with timeout on nightly CI with binary engine.

Ref: https://github.com/prisma/team-orm/issues/496


Jolg42 (approved):
Nice and easy 🙌🏼

PR #22040, opened 2023-11-20
chore: use packageManager field

jasonkuhrt:
This PR makes use of corepack by specifying the package manager to use in the `packageManager` field.

On trunk this is what happens when attempting to use corepack:

image

Once this change is made, using `pnpm` within he project should automatically use the correct version of pnpm, always.

The current approach on trunk leads to an experience like this:

CleanShot 2023-11-20 at 12 24 36@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,9 @@
   "name": "dev-env",
   "triggerEmptyDevReleaseByIncrementingThisNumber": 0,
   "license": "Apache-2.0",
+  "packageManager": "pnpm@8.6.7",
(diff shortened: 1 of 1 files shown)


jasonkuhrt:
This isn't a trivial fix, maybe someone else familiar with the project can figure it out.

Jolg42:
I think this confuses the `resolveYarnSchema()` function, like I said in Slack. And since we don't really need this change, so we can drop it until someone feels this really needs to be done.

PR #22017, opened 2023-11-17
chore: bundle vendors in internal packages

millsp:
Replaces https://github.com/prisma/prisma/pull/20857

Part of https://github.com/prisma/team-orm/issues/106

This PR aims to solve the fact that we are not able to update to newer ESM dependencies. To achieve this, we are switching to a bundling-only strategy (chunked). By bundling both CJS and ESM dependencies together with our published code, we can eliminate the semantic incompatibilities between both and produce a bundle that is CJS only (and possibly ESM bundles in the future, too).

In contrast with the previous PR, this one does not attempt to bundle into a single file. Instead it uses what is called bundle chunking, called `splitting` in `esbuild`. This basically allows a set of file to have their dependencies bundled (what we want to solve the ESM issue), and the bundle is efficiently shared between all existing files (chunked), and all files are efficiently chunked themselves. This fully preserves the multi-file structure that we have today, so we won't impact dependents with this change.

This PR is substantially simpler than the original one. 

/integration

PR #19447, opened 2023-05-25
fix(cli): remove deprecated flags/args & old migrate detection

Jolg42:
See `throwUpgradeErrorIfOldMigrate` for old migrate detection.

For db execute / seed / diff
        '--preview-feature': Boolean

For migrate deploy / dev / reset / resolve / status
        '--experimental': Boolean,
        '--early-access-feature': Boolean,

For db push
        // Deprecated
        // --force renamed to --accept-data-loss in 2.17.0
        '--force': Boolean,
        '-f': '--force',

For db pull
      '--experimental-reintrospection': Boolean,
      '--clean': Boolean,

PR #22041, opened 2023-11-20
feat(cli): introduce platform namespace

jasonkuhrt:
This PR introduces a no-op platform namespace into the CLI.


packages/cli/src/platform/namespace.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Command } from '@prisma/internals'
+
+export class Platform implements Command {
+  public static new(): Platform {
+    return new Platform()
+  }
+
+  public async parse(argv: string[]) {
+    // todo Molt Command does not support CJS
+    // const args = MoltCommand.create().parse({ line: argv })
```
jasonkuhrt, line 11: I would like to use `@molt/command` for parameter definition, validation, and help rendering.

It currently relies on an ESM only package though, so I need to go "fix" that, because IIUC @Jolg42 we're not going to be using ESM for the foreseeable future still for the CLI right?

jasonkuhrt (commented):
IIUC this PR still needs to add usage of feature flags.

packages/cli/package.json
```diff
@@ -114,6 +114,7 @@
     "zx": "7.2.3"
   },
   "scripts": {
+    "prisma": "ts-node src/bin.ts",
```
jasonkuhrt, line 117: This makes manual testing locally easy.

packages/cli/src/platform/namespace.ts
Jolg42: @millsp is actually working on something about that right now, watch out for https://github.com/prisma/prisma/pull/22017 I hope we can merge this soon.

Jolg42: From a step back, as a general advice, I think it's important, from a user point of view, that the CLI doesn't completely look and feel different for platform commands.

packages/cli/src/platform/namespace.ts
```diff
@@ -0,0 +1,19 @@
+// import { Command as MoltCommand } from '@molt/command'
```
Jolg42, line 1: Please rename the file to "PlatformCommand" to follow the existing name scheme, it can be changed, but consistency is also important to consider.

packages/cli/src/platform/namespace.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Command } from '@prisma/internals'
+
+export class Platform implements Command {
+  public static new(): Platform {
+    return new Platform()
+  }
```
Jolg42, line 7:
```suggestion
export class PlatformCommand implements Command {
  public static new(): PlatformCommand {
    return new PlatformCommand()
  }
```
Same as file name

packages/cli/src/CLI.ts
```diff
@@ -59,7 +59,8 @@ export class CLI implements Command {
     }
 
     const cmd = this.cmds[cmdName]
-    if (cmd) {
+    const isVisible = cmdName !== 'platform' || Boolean(args['--early-access-feature'])
+    if (cmd && isVisible) {
```
Jolg42, line 63: You can remove this here and handle it in the platform command file, it would be better I think

Jolg42: Check https://github.com/prisma/prisma/pull/19447 for how to do it, see “packages/migrate/src/commands/MigrateCommand.ts” file

jasonkuhrt: I made a change before seeing your second comment. I think the current approach does the job though well enough?

jasonkuhrt:
@Jolg42 I will try adding a test that shows the feature flag is working

packages/cli/src/platform/PlatformCommand.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Command, unknownCommand } from '@prisma/internals'
+
+export class PlatformCommand implements Command {
+  public static new(): PlatformCommand {
+    return new PlatformCommand()
+  }
+
+  public async parse(argv: string[]) {
+    const isHasEarlyAccessFeatureFlag = Boolean(argv.find((_) => _.match(/early-access-feature/)))
+    if (!isHasEarlyAccessFeatureFlag) return unknownCommand('You must pass --early-access-feature to use platform commands.', 'platform') // prettier-ignore
```
Jolg42, line 11: This is not the intended usage for this, we have an error for this case already here "EarlyAccessFlagError", it can be modified if needed.

Jolg42 (approved):
See comment
I'll approve this so you can merge and iterate later

packages/cli/src/platform/PlatformCommand.ts
jasonkuhrt: Thanks @Jolg42 found thanks to auto import!

packages/cli/src/platform/namespace.ts
jasonkuhrt: I need to get this comment resolved to merge. I think this thread is informational now and nothing actionable. Will resolve!

packages/cli/src/__tests__/commands/platform/Platform.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe('using cli', () => {
+  it('should not work without the --early-access-feature flag', async () => {
+    console.log(1)
+    const result = await PlatformCommand.new().parse([])
+    const resultIsError = isError(result)
+    expect(resultIsError).toBeTruthy()
+    if (resultIsError) {
+      expect(result.message).toMatch('This feature is currently in Early Access.')
+    }
```
Jolg42:
```suggestion
  await expect(
    PlatformCommand.new().parse([]),
  ).rejects.toMatchInlineSnapshot(``)
```

Can be simplified, it's best to not use a conditional
Alternatively, it can be a "toThrow()" with a matching message

jasonkuhrt: The conditional is to satisfy TS.

jasonkuhrt: Your suggestion assumes an error is thrown, but it is actually returned. Are you suggesting to throw an error instead of returning it?

jasonkuhrt: @Jolg42 I'll merge and we can of course continue changing things (transition to a throw and update this test, etc.)

Jolg42: We should have some consistency in the CLI, we use throw indeed.

packages/cli/src/platform/PlatformCommand.ts
```diff
@@ -0,0 +1,24 @@
...
+import { EarlyAccessFlagError } from '../../../migrate/src/utils/flagErrors'
+
+export class PlatformCommand implements Command {
+  public static new(): PlatformCommand {
+    return new PlatformCommand()
+  }
+
+  public async parse(argv: string[]) {
+    const isHasEarlyAccessFeatureFlag = Boolean(argv.find((_) => _.match(/early-access-feature/)))
+    if (!isHasEarlyAccessFeatureFlag) return new EarlyAccessFlagError() // prettier-ignore
```
Jolg42, line 12:
```suggestion
    // TODO: throw instead and adapt tests
    if (!isHasEarlyAccessFeatureFlag) return new EarlyAccessFlagError() // prettier-ignore
```

PR #22046, opened 2023-11-21
feat(client): Generating Prisma client without any model in its schema

### LINKED ISSUE #11582: Generating Prisma client without any model in its schema
## Problem

We have a server template including Prisma, which works fine so far. However, we now have another server extending this template without the need of models - just raw queries. While generating the Prisma client, via `npx prisma generate`, it is complaining about not having at least one model.

```
$ npx prisma generate
Error: 
You don't have any models defined in your schema.prisma, so nothing will be generated.
```

Obviously correct - but in this case not wanted. So, we would like to use Prisma with basically `$queryRaw` and `$queryRawUnsafe` only.

## Suggested solution

It should be allowed to generate the Prisma Client without existing models in its schema.

## Alternatives

First guess would be to add a dummy model.

kivanval:
The purpose of this PR is to allow you to create a project with Prisma without defining models.

The main difficulty was that the returned DMMF with empty models returns unexpected values, as one kind of solution we can intercept this on the client side of the module.

Also the documentation has changed a bit as it was customized for the fact that there is always one model.

I saw in comments under issue that it is possible to make more flag, it is not difficult to add, but we can discuss its necessity :)

closes #11582


Kivanval:
closes #11582

packages/internals/src/__tests__/getGenerators/getGenerators.test.ts
```diff
@@ -725,70 +725,20 @@ describe('getGenerators', () => {
...
-      `)
-    }
+    expect(generators.length).toEqual(1)
 
     expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(`""`)
     expect(ctx.mocked['console.info'].mock.calls.join('\n')).toMatchInlineSnapshot(`""`)
-    expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`""`)
+    expect(stripAnsi(ctx.mocked['console.warn'].mock.calls.join('\n'))).toMatchInlineSnapshot(`
+    "
+    warn You don't have any models defined in your schema.prisma, so nothing will be generated.
```
janpio, line 739: Isn't this what you wanted to change?

To me this message means "generate does not work".
But I think now generate does work, but it will just be an empty Client?

How about keeping the old message about a simple copy-pasteable model name, and instead describe clearer what is happening now?

Kivanval: > But I think now generate does work, but it will just be an empty Client?
Yeah, you're right.



> How about keeping the old message about a simple copy-pasteable model name, and instead describe clearer what is happening now?

I didn't really understand what you were trying to suggest here, except to make it clearer that everything works.
So, you're suggesting to leave the last message, where it was stated that you can specify the model and also say that there are no models?

```
You don't have any models defined in your 'schema.prisma', so PrismaClient will be empty.
You can define a model like this:

model User {
  id    String     @id @default(auto()) @map("_id") @db.ObjectId
  email String  @unique
  name  String?
},

More information in our documentation:
${link('https://pris.ly/d/prisma-schema')}
```

Something like this?

janpio: Yes, exactly. Otherwise that helpful output goes away.

Kivanval: I'm pushing change, you'll see when you have time. 🙃

Kivanval:
@janpio

PR #22048, opened 2023-11-21
fix(cli): throw early access error

jasonkuhrt:
This PR follows up on the feedback here https://github.com/prisma/prisma/pull/22041#discussion_r1399640784.

### DIFF
--- a/packages/cli/src/platform/PlatformCommand.ts
+++ b/packages/cli/src/platform/PlatformCommand.ts
@@ -9,14 +9,13 @@ export class PlatformCommand implements Command {
 
   public async parse(argv: string[]) {
     const isHasEarlyAccessFeatureFlag = Boolean(argv.find((_) => _.match(/early-access-feature/)))
-    // TODO: throw instead and adapt tests
-    if (!isHasEarlyAccessFeatureFlag) return new EarlyAccessFlagError() // prettier-ignore
+    if (!isHasEarlyAccessFeatureFlag) throw new EarlyAccessFlagError() // prettier-ignore
 
     // todo Molt Command does not support CJS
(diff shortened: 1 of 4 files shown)


packages/cli/src/platform/PlatformCommand.ts
jasonkuhrt, line 1: Renamed this module to be consistent with the module under test.

packages/cli/src/platform/PlatformCommand.ts
```diff
@@ -9,14 +9,13 @@ export class PlatformCommand implements Command {
 
   public async parse(argv: string[]) {
     const isHasEarlyAccessFeatureFlag = Boolean(argv.find((_) => _.match(/early-access-feature/)))
-    // TODO: throw instead and adapt tests
-    if (!isHasEarlyAccessFeatureFlag) return new EarlyAccessFlagError() // prettier-ignore
+    if (!isHasEarlyAccessFeatureFlag) throw new EarlyAccessFlagError() // prettier-ignore
```
Jolg42, line 12: Note it would be good to avoid disabling prettier completely like here for this line with "// prettier-ignore"

Jolg42 (approved):

packages/cli/src/platform/PlatformCommand.ts
jasonkuhrt: I prefer to keep guards on a single line. Do you want to avoid that style?

Jolg42: Using `// prettier-ignore` completely disables prettier for this line, so it's better to avoid when I think the value is low, like here.
The less `// prettier-ignore` in the codebase, the better.
There are some cases where it's actually required, these are rare.

PR #22053, opened 2023-11-21
feat(prisma): introduce initial skeleton commands

jasonkuhrt:
This PR introduces skeletons of the various initial commands for the Prisma CLI.


packages/cli/src/bin.ts
```diff
@@ -83,7 +83,25 @@ async function main(): Promise<number> {
   const cli = CLI.new(
     {
       init: Init.new(),
-      platform: PlatformCommand.new(),
+      platform: Platform.$.new({
+        workspace: Platform.Workspace.$.new({
```
janpio, line 87:
```suggestion
        platform: Platform.$.new({
          workspace: Platform.Workspace.$.new({
```

packages/cli/src/platform/$.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class $ implements Command {
+  public static new(commands: Commands): $ {
+    return new $(commands)
+  }
+
+  private constructor(private readonly commands: Commands) {}
+
+  public async parse(argv: string[]) {
+    const isHasEarlyAccessFeatureFlag = Boolean(argv.find((_) => _.match(/early-access-feature/)))
```
janpio, line 13: Not really new, but maybe we want to make this shorter as just `--early-access` (so without the `-feature` suffix)?

jasonkuhrt: That would be nice :)

janpio: Let's ask @Jolg42, but as we are not using that at all (and haven't for years) I think that should be fine.

packages/cli/src/__tests__/commands/platform/PlatformCommand.test.ts
```diff
@@ -1,17 +1,17 @@
 import { isError } from '@prisma/internals'
 
-import { PlatformCommand } from '../../../platform/PlatformCommand'
+import { $ } from '../../../platform/$'
 
 describe('--early-access-feature flag', () => {
   it('should not work without it', () => {
     // eslint-disable-next-line
```
Jolg42, line 7: Why is `// eslint-disable-next-line` needed?

packages/cli/src/platform/$.ts
Jolg42: Feel free to change it, shorter is better I think too, make sure to search everywhere for it in this repository 👍🏼

Jolg42 (approved):

packages/cli/src/bin.ts
```diff
@@ -27,7 +27,7 @@ import { DebugInfo } from './DebugInfo'
 import { Format } from './Format'
 import { Generate } from './Generate'
 import { Init } from './Init'
-import { PlatformCommand } from './platform/PlatformCommand'
+import { Platform } from './platform'
```
sampolahtinen, line 30: I like the idea of namespacing all the commands instead of following the pattern `<NAMESPACE><COMMAND>`, for example `MigrateDev`. I would like to hear @Jolg42 opinion on this one as well.

Would `PlatformCommand` be a better name? That reads quite nicely and communicates the intention of wrapping all commnds for the platform. Examples:
 - PlatformCommand.Auth.Login
 - PlatformCommand.Project.Create

sampolahtinen (approved):

packages/cli/src/platform/$.ts
```diff
@@ -0,0 +1,20 @@
+import { Command, Commands } from '@prisma/internals'
+
+import { EarlyAccessFlagError } from '../../../migrate/src/utils/flagErrors'
+
+export class $ implements Command {
```
sampolahtinen, line 5: In general would it make sense to rename `$`  to be `root` or `RootCommand` wdyt?

jasonkuhrt: Sure could be. The symbol denotes that it's a namespace which is different than a command. Technically could be named anything including "root" so $ tries to make that separation between namespace and command more explicit.

But maybe $ is just counter productive too!

packages/cli/src/platform/$.ts
jasonkuhrt: @Jolg42 @janpio I'll make the change but in a different PR if that's ok.

packages/cli/src/__tests__/commands/platform/PlatformCommand.test.ts
jasonkuhrt: CleanShot 2023-11-21 at 13 41 58@2x

packages/cli/src/bin.ts
jasonkuhrt: I'm not sure I follow the pattern. `PlatformCommand` but not `AuthPlatform` etc. What makes those namespaces different?

packages/cli/src/bin.ts
```diff
@@ -83,7 +83,30 @@ async function main(): Promise<number> {
   const cli = CLI.new(
     {
       init: Init.new(),
-      platform: PlatformCommand.new(),
+      platform: Platform.$.new({
+        login: Platform.Login.new(),
+        logout: Platform.Logout.new(),
```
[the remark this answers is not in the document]
jasonkuhrt: @luanvdw ?

PR #22056, opened 2023-11-21
feat(prisma): platform integration for project & accelerate cmds

jasonkuhrt:
This PR integrates the prototype PDP API into the Prisma CLI for a select few initial commands:

- workspace show
- accelerate enable
- accelerate disable
- project create
- project delete
- project show


Jolg42:
Note that tests are failing and it will need to be fixed before merging.
```
FAIL src/__tests__/commands/platform/PlatformCommand.test.ts
  ● --early-access-feature flag › should work with it

    ArgError: unknown or unexpected option: --early-access-feature

      68 | function arg(argv, spec, stopAtPositional = true, permissive = false) {
      69 |   try {
    > 70 |     return (0, import_arg.default)(spec, { argv, stopAtPositional, permissive });
         |                                 ^
      71 |   } catch (e) {
      72 |     return e;
      73 |   }

      at arg (../../node_modules/.pnpm/arg@5.0.2/node_modules/arg/index.js:132:13)
      at arg (../internals/dist/chunk-2ZMLAV6A.js:70:33)
      at $.parse (src/platform/$.ts:16:21)
      at Object.parse (src/__tests__/commands/platform/PlatformCommand.test.ts:14:34)
```
https://github.com/prisma/prisma/actions/runs/6949362943/job/18933113690?pr=22056#step:8:104

packages/cli/package.json
```diff
@@ -123,7 +123,8 @@
     "prepublishOnly": "pnpm run build"
   },
   "dependencies": {
-    "@prisma/engines": "workspace:*"
+    "@prisma/engines": "workspace:*",
+    "arg": "5.0.2"
```
Jolg42, line 127: This needs to be moved under `devDependencies`, it will be bundled then

packages/cli/src/platform/project/create.ts
```diff
@@ -1,12 +1,35 @@
...
+    const args = arg(argv, {
+      ...platformParameters.workspace,
+      '--display-name': String,
+      '-d': '--display-name',
+    })
+    if (isError(args)) throw args
+    const token = getRequiredParameterOrThrow(args, ['--token', '-t'], 'PRISMA_TOKEN')
+    const workspace = getRequiredParameterOrThrow(args, ['--workspace', '-w'])
+    const displayName = getOptionalParameter(args, ['--display-name', '-d'])
+    if (isError(args)) throw args
```
Jolg42, line 25: I don't understand why `if (isError(args)) throw args` is repeated 🤔

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const platformParameters = {
+  global: {
+    '--early-access-feature': Boolean,
+    '--token': String,
+    '-t': '--token',
+  },
+  workspace: {
+    '--early-access-feature': Boolean,
+    '--token': String,
```
sampolahtinen, line 13: Do we need to repeat global parameters for each command?

packages/cli/src/platform/project/$.ts
```diff
@@ -8,8 +10,8 @@ export class $ implements Command {
   private constructor(private readonly commands: Commands) {}
 
   public async parse(argv: string[]) {
-    await Promise.resolve('todo')
-
-    return JSON.stringify(argv)
+    const args = arg(argv, {})
+    if (isError(args)) throw args
```
Jolg42, line 14: What we normally do in the CLI is to return the help message along (you can check for the `public help` function definition of any command)
```suggestion
    if (isError(args)) return this.help(args.message)
```

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const platformRequestOrThrow = async (params: {
+  route: string
+  token: string
+  path: string
+  payload?: object
+}) => {
+  const { path, payload, token, route } = params
+  const url = new URL(`https://console.prisma.io/${path.replace(/^\//, '')}?_data=routes/${route}`)
```
Jolg42, line 65: `https://console.prisma.io` could probably be a const of its own?

packages/cli/src/helpers.ts
jasonkuhrt: Given how `arg` works, yes. Otherwise it complains about unknown parameters.

The other option would be to have an ancestor strip it, which brings its own complexity.

packages/cli/src/platform/project/$.ts
jasonkuhrt: Can you confirm this is correct:

1. Throw feature flag error
2. Return all other errors

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  environmentVariable?: string,
+): Exclude<$Args[$Names[number]], undefined> => {
+  const value = getOptionalParameter(args, names, environmentVariable)
+  if (value === undefined) {
+    throw new Error(`Missing ${names.join(' or ')}`)
+  }
+  return value
+}
+
+export const platformRequestOrThrow = async (params: {
```
sampolahtinen, line 58: Should we move Platform specific helpers to `./utils/platform`?

Auth utils could be placed in this same locations.

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  const url = new URL(`https://console.prisma.io/${path.replace(/^\//, '')}?_data=routes/${route}`)
+  const headers = new Headers({
+    'Content-Type': 'application/json',
+    Authorization: `Bearer ${token}`,
+  })
+  const response = await fetch(url, {
+    method: payload ? 'POST' : 'GET',
+    headers,
+    body: payload ? JSON.stringify(payload) : undefined,
+  })
```
Jolg42, line 74: You probably want to use a version for future-proofing?
For ideas https://restfulapi.net/versioning/

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  token: string
+  path: string
+  payload?: object
+}) => {
+  const { path, payload, token, route } = params
+  const url = new URL(`https://console.prisma.io/${path.replace(/^\//, '')}?_data=routes/${route}`)
+  const headers = new Headers({
+    'Content-Type': 'application/json',
+    Authorization: `Bearer ${token}`,
+  })
```
Jolg42, line 69: It could be interesting to pass 
- the Prisma CLI version
- potentially, later some data we have in `runCheckpointClientCheck()` like `project_hash`

packages/cli/src/platform/accelerate/disable.ts
```diff
@@ -1,12 +1,29 @@
...
   }
 
   public async parse(argv: string[]) {
-    await Promise.resolve('todo')
-    return JSON.stringify(argv)
+    const args = arg(argv, {
+      ...platformParameters.project,
+    })
+    if (isError(args)) throw args
+    const token = getRequiredParameterOrThrow(args, ['--token', '-t'], 'PRISMA_TOKEN')
```
sampolahtinen, line 15: `token` should be optional parameter right?

How I reason about the authentication check priorities is:
1. If `token` argument is present, read the value and verify the token
2. If `token` is not present, attempt to read credentials from disk
3. If none of the above, throw `Unauthorized`

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  const response = await fetch(url, {
```
Jolg42, line 70: It could be interesting to set a default timeout because it's not part of `fetch` spec, see:
https://www.npmjs.com/package/node-fetch#request-cancellation-with-abortsignal

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  })
+  const text = await response.text()
+  const json = JSON.parse(text)
+  return json
+}
+
+export const dispatchToSubCommand = async (commands: Commands, args: Arg.Result<{}>) => {
+  const commandName = args._[0]
+  const command = commands[commandName]
+  if (!command) throw new Error(`Command ${commandName} not found`)
```
sampolahtinen, line 83: Could we utilize existing `unknownCommand` @prisma/internals utility?

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const getRequiredParameterOrThrow = <$Args extends Record<string, unknown>, $Names extends (keyof $Args)[]>(
+  args: $Args,
+  names: $Names,
+  environmentVariable?: string,
+): Exclude<$Args[$Names[number]], undefined> => {
+  const value = getOptionalParameter(args, names, environmentVariable)
+  if (value === undefined) {
+    throw new Error(`Missing ${names.join(' or ')}`)
```
Jolg42, line 53:
```suggestion
    throw new Error(`Missing value for ${names.join(' or ')} parameter`)
```
Probably better like this?

packages/cli/package.json
Jolg42: But actually you should probably use 
```
import { arg } from '@prisma/internals'
```
instead

jasonkuhrt: I was, it was installed for the types, but I've since refactored anyways and relying directly on `argv` in the dispatch function now.

packages/cli/src/helpers.ts
jasonkuhrt: Let's put a TODO for now.

packages/cli/src/platform/project/$.ts
Jolg42: all Errors are thrown, here you can check for the public help function definition of any command like I said earlier

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (!entry) {
+    if (environmentVariable) {
+      const value = process.env[environmentVariable]
+      if (value) {
+        return value as any
+      }
+    }
+  }
+  return (entry?.[1] ?? undefined) as any
+}
```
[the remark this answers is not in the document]
jasonkuhrt: I'll put a TODO about that.

packages/cli/src/platform/accelerate/disable.ts
Jolg42: Indeed, if we have `login`, then the state will be persisted somewhere and token is not required

packages/cli/src/helpers.ts
jasonkuhrt: Too early, tbd

packages/cli/src/helpers.ts
jasonkuhrt: Yeah let's populate the `user-agent`. Is there an easy way to get at the current version?

packages/cli/src/helpers.ts
Jolg42: it doesn't mean you have to use it right away on the backend, but when you'll need it, it would be there

packages/cli/src/helpers.ts
jasonkuhrt: Found it, commit pushed!

packages/cli/src/platform/accelerate/disable.ts
jasonkuhrt: Its required for now, when we add login, yep, it should become optional as you described.

jasonkuhrt: I'll add TODO.

jasonkuhrt: Actually there are four callsites, no central place, we don't need to be reminded to do this anyways.

packages/cli/src/helpers.ts
jasonkuhrt: I'll put a TODO for now.

packages/cli/src/platform/$.ts
```diff
@@ -12,9 +13,7 @@ export class $ implements Command {
   public async parse(argv: string[]) {
     const isHasEarlyAccessFeatureFlag = Boolean(argv.find((_) => _.match(/early-access-feature/)))
     if (!isHasEarlyAccessFeatureFlag) throw new EarlyAccessFlagError()
-
-    await Promise.resolve('todo')
-
-    return ''
+    const result = await dispatchToSubCommand(this.commands, argv)
+    return JSON.stringify(result)
```
sampolahtinen, line 17: A nitty one:

```suggestion
return result
```,

or could be:
```suggestion
return await dispatchToSubCommand(this.commands, argv)
```,

**Review by sampolahtinen — APPROVED** [review]
Thanks for adding couple of todos! LGTM

File: packages/cli/src/platform/$.ts (hunk 14 above)
Comment (jasonkuhrt) [reply packages/cli/src/platform/$.ts:17]: Not sure I understand, removing the JSON stringification breaks the CLI output. This result goes direct to `console.log` so it cannot be the object since under the hood that runs `util.format` etc. which is not built for JSON.parse consumption that we need on the output (e.g. for `| jq`).

File: packages/cli/src/platform/project/$.ts (hunk 4 above)
Comment (Jolg42) [reply packages/cli/src/platform/project/$.ts]: Maybe I took a shortcut / misunderstood something by the way. Note that this is probably interesting for you 
See https://github.com/prisma/prisma/blob/ce9bb031140b994e04931d80847b6f04e745e860/packages/cli/src/bin.ts#L115-L127
If a command returns a `HelpError` then we print the message.
If it's another error, it prints the full error, which might not be what you want (you probably only want to print the error message as well?)

File: packages/cli/src/platform/$.ts (hunk 14 above)
Comment (sampolahtinen) [reply packages/cli/src/platform/$.ts:17]: The proposal was inspired from command parsing done by other commands. For example `MigrateCommand`

In this example, I like that the `unknownCommand` utility is used together with the `help` property of the class

https://github.com/prisma/prisma/blob/91cb420813b559bcf941b535f1cc606a23c0838c/packages/migrate/src/commands/MigrateCommand.ts#L85-L99

Comment (sampolahtinen) [reply packages/cli/src/platform/$.ts:17]: > We'd not doing help right now though.

Even if not, for now we could have an empty help block which we could then plug in to already existing pattern. Later on, we can add the help descriptions.

File: packages/cli/src/platform/project/$.ts (hunk 4 above)
Comment (jasonkuhrt) [reply packages/cli/src/platform/project/$.ts]: > all Errors are thrown, here you can check for the public help function definition of any command like I said earlier

@Jolg42 I'm confused

https://github.com/prisma/prisma/pull/22041#discussion_r1400238956

File: packages/cli/src/platform/$.ts (hunk 14 above)
Comment (jasonkuhrt) [reply packages/cli/src/platform/$.ts:17]: But we don't know if we're going to be using that pattern yet.

File: packages/cli/src/platform/workspace/$.ts [hunk 15]
```diff
@@ -8,8 +10,6 @@ export class $ implements Command {
   private constructor(private readonly commands: Commands) {}
 
   public async parse(argv: string[]) {
-    await Promise.resolve('todo')
-
-    return JSON.stringify(argv)
+    return dispatchToSubCommand(this.commands, argv)
```
sampolahtinen, line 13: This is exactly what I was hunting for in the other comment 🎉 Very nice!

packages/cli/src/platform/project/$.ts
jasonkuhrt: Will resolve for now, let's keep discussing though.

janpio (approved):

PR #22060, opened 2023-11-22
fix(client): create dummy wasm-bg file

millsp:
This PR fixes the following issue when trying to deploy with `engineType=wasm` as an Edge Middleware via Vercel and Next.js. 
>  ⨯ ./node_modules/.pnpm/@prisma+client@5.7.0-dev.37_prisma@5.7.0-dev.37/node_modules/.prisma/client/query-engine.wasm
> Module not found: Can't resolve './query_engine_bg.js'

Some bundlers (eg. webpack) need this file to exist, even if it's empty. This is because they analyze `query-engine.wasm` for references to other files (yes, `query-engine.wasm` has references to `query_engine_bg.js`). It does not matter for us, because we bundle `query_engine_bg.js`, so that bundler is worrying about something that is not an issue, actually. The actual test backing this is work in progress (via ecosystem-tests).

Perhaps a future solution that is cleaner would be to emit a file called `__query_engine_bg__`, that may be enough not to trigger the bundler.


Jolg42 (approved):

Jolg42:
It would be interesting to have an ecosystem-test that monitors this case specifically.

millsp:
Yes, I am busy writing the Next.js tests which this is for.

PR #22068, opened 2023-11-22
feat(prisma): add platform api calls for apikey cmds

jasonkuhrt:
This PR brings Platform API calls from `apikey` Prisma CLI commands.


packages/cli/src/platform/apikey/create.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public async parse(argv: string[]) {
+    const args = arg(argv, {
+      ...platformParameters.project,
+      '--display-name': String,
+      '-d': '--display-name',
+    })
+    if (isError(args)) return args
+    const token = getRequiredParameter(args, ['--token', '-t'], 'PRISMA_TOKEN')
```
janpio, line 22: Aside: This env var name should probably be more PDP specific.

jasonkuhrt: I also think its suffix should match the flag name, whatever it changes to.

@luanvdw this is the token I was mentioning before.

Based on some other ongoing discussions IIUC the most accurate term here would be `PRISMA_PLATFORM_MANAGEMENT_TOKEN` (but this is not settled yet).

packages/cli/src/platform/accelerate/disable.ts
```diff
@@ -26,6 +26,6 @@ export class Disable implements Command {
         intent: 'disable',
         projectId: project,
       },
-    })
+    }) as Promise<any>
```
jasonkuhrt, line 29: These have been added to make what was previously implicitly this type explicit.

It needs to be fixed later.

Right now the local CLI framework here only allows strings to be returned.

Changing this requires increasing the scope of this PR and I think we can defer that.

The casts here are not changing behaviour, so, its still progress, by making them explicit, in this PR.

sampolahtinen: Let's add some of your remarks as TODO comments?

jasonkuhrt: There's a bunch of callsites though and in this case the `as` casting serves as kind of todo anyways. But feel free to push a commit if that helps!

Jolg42 (approved):
Looks ok, though without tests 😉

packages/cli/src/utils/platform.ts
```diff
@@ -71,7 +71,7 @@ export const platformRequestOrThrow = async (params: {
   token: string
   path: string
   payload?: object
-}) => {
+}): Promise<object> => {
```
sampolahtinen, line 74: @jasonkuhrt lets re-iterate on this one. I think we could just remove explicit return type and avoid casting `as Promise<any>`

I will approve and merge this PR as it is.

sampolahtinen (approved):

PR #22082, opened 2023-11-23
chore: use dprint

tristan-f-r:
runs much, _much_ faster than prettier

(see [Microsoft/TypeScript's discussion on using dprint](https://github.com/microsoft/TypeScript/pull/54820))


aqrln:
Hey @LeoDog896, thanks for opening the PR and trying to improve our DX! While dprint is indeed faster and more flexible, we don't currently have a pressing need to migrate: Prettier is fast enough for us right now and doesn't really cause any problems. I would personally consider dprint or deno-fmt for new projects, but for the existing ones that already use Prettier the benefits must outweight the cost of migration, which may be substantial, as the code churn will cause merge conflicts in all in-flight work (and we will need to edit every commit in every relevant open PR or local branch to reformat it), make using `git blame` less straightforward (as we will need to `--ignore-rev` this commit which is not possible in GitHub UI, for example), and so on.

We appreciate the work you've done here but unfortunately I have to close this PR because we cannot merge it at this time, and if we want to proceed with migration to another formatter in the future, it will have to be redone as this kind of changes to the code base is quite time sensitive.

Going forward, before working on something that affects the whole codebase and hasn't been discussed, I'd suggest opening an issue or discussion to collect feedback first, or reaching out to us in the community Slack or Discord. We'd be happy to chat and provide guidance! When switching tools for performance reasons, it would also help to provide concrete numbers that demonstrate the value of switching, so that it can be assessed against the potential cost.

PR #22093, opened 2023-11-23
feat(prisma): platform cli terminal auth

sampolahtinen:
This PR introduces web based authentication for the Platform CLI.

NOTE: This feature will not work before needed changes have been applied on the Console's side: https://github.com/prisma/pdp-control-plane/pull/723.


packages/cli/package.json
```diff
@@ -97,6 +98,7 @@
     "fs-jetpack": "5.1.0",
     "get-port": "5.1.1",
     "global-dirs": "3.0.1",
+    "http": "0.0.1-security",
```
Jolg42, line 101:
```suggestion
```
`http` comes from Node.js, it’s not a package.

packages/cli/src/utils/platform.ts
```diff
@@ -101,3 +109,41 @@ export const dispatchToSubCommand = async (commands: Commands, argv: string[]) =
...
+/**
+ *
+ * Authentication related utils
+ *
+ */
+export interface AuthConfig {
+  token?: string | null
+}
+
+export const configDirectoryPath = new XdgAppPaths('prisma-platform-cli').config()
```
Jolg42, line 122: We’ll need to document the default paths used by the CLI somewhere.

packages/cli/package.json
sampolahtinen: Totally my bad. Thanks for spotting!

packages/cli/src/platform/login.ts
```diff
@@ -1,12 +1,111 @@
...
+      return ''
+    } catch (error) {
+      throw new Error(`Authentication failed: ${isError(error) ? error.message : ''}`)
+    }
+  }
+}
+
+const generateAuthSigninUrl = async (params: { connection: string; redirectTo: string }) => {
+  const prismaClientVersion = await getInstalledPrismaClientVersion().catch(() => null)
+  const state = { client: `prisma@${prismaClientVersion}`, ...params }
```
sampolahtinen, line 87: Added `client` property to the redirect state that includes Prisma client version. This can be useful for future implementations and backwards compatibility.

packages/cli/src/platform/$.ts
```diff
@@ -14,6 +14,7 @@ export class $ implements Command {
     const isHasEarlyAccessFeatureFlag = Boolean(argv.find((_) => _.match(/early-access-feature/)))
     if (!isHasEarlyAccessFeatureFlag) throw new EarlyAccessFlagError()
     const result = await dispatchToSubCommand(this.commands, argv)
+    // TODO: Consider removing JSON.stringify as it breaks if sub-command parse returns JSON.stringify
     return JSON.stringify(result)
```
tatianainama, line 18: we could check if the type of result is `string` and otherwise stringify?

packages/cli/src/platform/login.ts
```diff
@@ -1,12 +1,111 @@
...
   }
 
-  public async parse(argv: string[]) {
-    await Promise.resolve('todo')
-    return JSON.stringify(argv)
+  public async parse() {
+    console.log('Authenticating to Prisma Platform CLI via browser')
+
+    const server = http.createServer()
+    const { port } = await listen(server, 0, '127.0.0.1')
```
tatianainama, line 29: I'm a bit confused by the parameters here for `listen`, the documentation of the library is quite vague, and this options are not included there, would you mind adding some docs explaining what the `0` means for example?

On a side note, would it be better to just use the result instead for the full url?
I'm getting this snippet from the [library docs](https://www.npmjs.com/package/async-listen):

```ts
const address = await listen(server, port);
console.log(address); // => URL('http://127.0.0.1:3000')

// use address instead of concatenate the port later 
```

packages/cli/src/platform/login.ts
tatianainama, line 87: do we need the `prisma@` ? would the consumer of this need to remove it later to get the version?

tatianainama (approved):
I dropped a couple of nits, but lgtm!

packages/cli/src/platform/$.ts
sampolahtinen: That is a good solution!

packages/cli/src/platform/login.ts
sampolahtinen: I was bouncing between two options:
1.  
```ts
const state = { client: `prisma@<version number>`}
```
2.
```ts
const state = { client: `prisma`, version: <version number>}
```
@jasonkuhrt already established a pattern here and I thought to follow the same:
https://github.com/prisma/prisma/blob/edf305f55792a099bba340acffce37740cd030b3/packages/cli/src/utils/platform.ts#L91

However, you are right that consumer would need to remove the version number and/or add some extra string manipulation. I personally like the (1) more

packages/cli/src/platform/login.ts
sampolahtinen: Will update!

packages/cli/src/platform/$.ts
sampolahtinen: I will fix this in another PR as it touches files not relevant to this PR

packages/cli/src/platform/login.ts
tatianainama: > @jasonkuhrt already established a pattern here and I thought to follow the same pattern

Then it makes sense to keep the same pattern. Based on that, should we then also declare `state` the same way?
```ts
const state = { userAgent: `prisma@<version>` }
```

I guess this is because we are limited to send this data only through params, instead of setting headers as you showed in the snippet right?

sampolahtinen: > I guess this is because we are limited to send this data only through params, instead of setting headers as you showed in the snippet right?

Correct. We can't set the headers since we simply just open the address in the browser.

Reason why I was thinking to use name `client` is that it could become a pattern:
1. If console receives an external authentication request, the state is expected to include at least `client` and `redirectTo` properties

PR #22102, opened 2023-11-23
fix(client): wasm engine loading for edge-light

millsp:
`edge-light` needs to load `wasm` files in a different way that is not standard, with an additional `?module` on the path. This causes issues with the `workerd` runtime, as `wrangler` legitimately does not understand `?module` during its static analysis of the code and fails. This PR attempts to make it work for both runtimes by:
1. Conditionally load based on runtime detection
2. Hiding `?module` in a template string to avoid issues with `wrangler`

/integration

### DIFF
--- a/packages/client/src/runtime/index.ts
+++ b/packages/client/src/runtime/index.ts
@@ -2,6 +2,10 @@ import * as Extensions from './core/extensions'
 import * as Public from './core/public'
 import * as Types from './core/types'
 
+export { Types }
+export { Extensions }
+export { Public }
+
 export { DMMFHelper as DMMFClass } from '../generation/dmmf'
 export { type BaseDMMF, DMMF } from '../generation/dmmf-types'
 export { NotFoundError } from './core/errors/NotFoundError'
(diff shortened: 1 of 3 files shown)


packages/client/src/runtime/index.ts
```diff
@@ -18,21 +22,17 @@ export {
...
 export { objectEnumValues } from './core/types/exported/ObjectEnums'
 export type { PrismaClientOptions } from './getPrismaClient'
 export { getPrismaClient } from './getPrismaClient'
 export { makeStrictEnum } from './strictEnum'
 export { warnEnvConflicts } from './warnEnvConflicts'
 export { Debug } from '@prisma/debug'
 export type { DriverAdapter } from '@prisma/driver-adapter-utils'
+export { warnOnce } from '@prisma/internals'
 export { default as Decimal } from 'decimal.js'
+export { detectRuntime } from 'detect-runtime'
```
janpio, line 36: We should sooner or later upgrade here:
> Warning detect-runtime has been deprecated in favor of [std-env](https://github.com/unjs/std-env) (since version 3.4.0), which includes the same functionalities and much more.
via https://github.com/QuiiBz/detect-runtime

millsp: https://github.com/prisma/team-orm/issues/626

millsp: Thanks for the reminder

SevInf (approved):

jkomyno (approved):

PR #22126, opened 2023-11-27
feat(prisma): platform cli read tokens

sampolahtinen:
This PR includes a proposal for reading platform authentication token.


packages/cli/src/bin.ts
```diff
@@ -72,6 +73,7 @@ const redactedCommandAsString = redactCommandArray([...commandArray]).join(' ')
 
 const isPrismaInstalledGlobally = isCurrentBinInstalledGlobally()
 
+export let platformToken: string | undefined
```
sampolahtinen, line 76: This and LOC 85 - 93 is the main feature proposal.

PR #22128, opened 2023-11-27
test(client): skip "hijacking a batch transaction" tests with planetscale adapter

aqrln:
These tests are very flaky with PlanetScale adapter and it's disrupting.
This commit skips them with PlanetScale until we fix the underlying
issue. Although the functionality actually works as expected, the logs
don't always arrive in the order these tests expects, so they may fail.
The problem seems more general than PlanetScale as it was also reported
to happen with Neon and even native MySQL connector, but that's rare so
I only skipped PlanetScale here (which fails every other time on CI and
consistently locally).

Ref: https://github.com/prisma/team-orm/issues/495

### DIFF
--- a/packages/client/tests/functional/extensions/query.ts
+++ b/packages/client/tests/functional/extensions/query.ts
@@ -3,7 +3,7 @@ import { assertNever } from '@prisma/internals'
 import { randomBytes } from 'crypto'
 import { expectTypeOf } from 'expect-type'
 
-import { Providers } from '../_utils/providers'
+import { ProviderFlavors, Providers } from '../_utils/providers'
 import { wait } from '../_utils/tests/wait'
 import { waitFor } from '../_utils/tests/waitFor'
 import { NewPrismaClient } from '../_utils/types'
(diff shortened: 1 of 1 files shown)


packages/client/tests/functional/extensions/query.ts
```diff
@@ -529,121 +529,124 @@ testMatrix.setupTestSuite(
...
     )
 
-    testIf(provider !== Providers.MONGODB && process.platform !== 'win32')(
-      'hijacking a batch transaction into another one with a simple call',
-      async () => {
-        const fnEmitter = jest.fn()
+    testIf(
+      provider !== Providers.MONGODB &&
+        process.platform !== 'win32' &&
+        providerFlavor !== ProviderFlavors.JS_PLANETSCALE,
```
Jolg42, line 535: Add a comment here for why the test is skipped with `providerFlavor !== ProviderFlavors.JS_PLANETSCALE`?

Jolg42 (approved):

PR #22144, opened 2023-11-28
feat(cli): read platform token

sampolahtinen:
This PR introduces reading of authentication tokens from the disk and using it as an additional way for authenticating requests. 

The priority order of different authentication methods is as follows:
1. Attempt to read value from arguments `--token` or `-t`
2. Attempt to read value from environment variable `PRISMA_TOKEN`
3. Attempt to read token stored in `auth.json` on the disk
4. If none of the above is present, throw `ErrorPlatformUnauthorized` 

NOTE:
This feature requires https://github.com/prisma/pdp-control-plane/pull/730


packages/cli/src/utils/platform.ts
```diff
@@ -59,6 +59,21 @@ export const getRequiredParameter = <$Args extends Record<string, unknown>, $Nam
...
+  try {
+    const token =
+      getOptionalParameter(args, ['--token', '-t'], 'PRISMA_TOKEN') ||
+      (await readAuthConfig().catch(() => ({ token: '' }))).token
+    return token
+  } catch (error) {
+    return ``
+  }
+}
+
```
tatianainama, line 76: what do you think on making this one `getPlatformTokenOrThrow` ? seems like consumers are also throwing this error 
I think this can remove some code repetition across the different commands

packages/cli/src/utils/platform.ts
```diff
@@ -59,6 +59,21 @@ export const getRequiredParameter = <$Args extends Record<string, unknown>, $Nam
   return value
 }
 
+export const ErrorPlatformUnauthorized = new Error(
+  'No platform credentials found. Please provide a token via --token or -t, or add PRISMA_TOKEN environment variable or run `prisma platform login`.',
```
Jolg42, line 63:
```suggestion
  'No platform credentials found. Run `prisma platform login` first. Alternatively you can provide a token via the `--token` or `-t` parameters, or set the `PRISMA_TOKEN` environment variable with the token.',
```

I would first suggest login.

Jolg42: You could make it look like other instructions in other parts in the CLI
Like 

```
`Run ${green(getCommandWithExecutor('prisma platform login'))}`
```

`getCommandWithExecutor` is from `@prisma/internals`

packages/cli/src/utils/platform.ts
```diff
@@ -59,6 +59,21 @@ export const getRequiredParameter = <$Args extends Record<string, unknown>, $Nam
...
+)
+
+export const getPlatformToken = async <$Args extends Record<string, unknown>>(args: $Args) => {
+  try {
+    const token =
+      getOptionalParameter(args, ['--token', '-t'], 'PRISMA_TOKEN') ||
+      (await readAuthConfig().catch(() => ({ token: '' }))).token
+    return token
+  } catch (error) {
+    return ``
```
Jolg42, line 73: Maybe add some logging using `debug('Error from getPlatformToken()', error)`

packages/cli/src/utils/platform.ts
sampolahtinen: I agree with you. Initially I actually had this function throwing.

@Jolg42 are there any code guidelines/paradigms about returning an error vs throwing an error?

In this specific context, it would make sense to throw, avoiding code repetition as @tatianainama pointed out.

packages/cli/src/utils/platform.ts
sampolahtinen: Amazing, thanks @Jolg42! I will update the copy and highlight the command

packages/cli/src/utils/platform.ts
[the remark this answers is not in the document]
sampolahtinen: Valid point! This slipped past me. I will make it return `null`

packages/cli/src/utils/platform.ts
sampolahtinen: @Jolg42 did you intentionally use back ticks  around parameters or are single quotes ok? 

Example:  `'--token'`

Jolg42: In general, we use backticks for code quotes, so for consistency

packages/cli/src/utils/platform.ts
Jolg42: I think you'll be ok with throwing here since you know where this is used exactly. (at the beginning of a command).

packages/cli/src/utils/platform.ts
```diff
@@ -1,11 +1,15 @@
...
+import { Commands, getCommandWithExecutor, unknownCommand } from '@prisma/internals'
 import fs from 'fs-extra'
+import { green } from 'kleur/colors'
 import fetch, { Headers } from 'node-fetch'
 import path from 'path'
 import XdgAppPaths from 'xdg-app-paths'
 
 import { getInstalledPrismaClientVersion } from './getClientVersion'
 
+const debug = Debug('prisma:platform')
```
Jolg42, line 11:
```suggestion
const debug = Debug('prisma:cli:platform')
```
Would be better I think, we already have one like this. (and feel free to fix it for Studio later if you want, currently it's `prisma:studio` and would be better as `prisma:cli:studio`)

Jolg42 (approved):

packages/cli/src/utils/platform.ts
sampolahtinen: @tatianainama I converted this utility to throw! Anything else to change or good to go?

tatianainama (approved):

PR #22160, opened 2023-11-29
feat(CLI): improve platform auth, file i/o

sampolahtinen:
This PR introduces a new utility for reading JSON files that strips BOM. In addition, file system methods have been wrapped in try/catch block.

### DIFF
--- a/packages/cli/src/platform/logout.ts
+++ b/packages/cli/src/platform/logout.ts
@@ -1,4 +1,4 @@
-import { Command } from '@prisma/internals'
+import { Command, isError } from '@prisma/internals'
 
 import { deleteAuthConfig, readAuthConfig } from '../utils/platform'
 
@@ -8,8 +8,9 @@ export class Logout implements Command {
   }
 
   public async parse() {
-    const { token } = await readAuthConfig()
-    if (!token) {
+    const authJson = await readAuthConfig()
+    if (isError(authJson)) throw authJson
+    if (!authJson.token) {
       return 'You are not logged in.'
     }
     await deleteAuthConfig()
--- a/packages/cli/src/utils/platform.ts
+++ b/packages/cli/src/utils/platform.ts
@@ -71,9 +71,12 @@ export const ErrorPlatformUnauthorized = new Error(
 
 export const getPlatformTokenOrThrow = async <$Args extends Record<string, unknown>>(args: $Args) => {
   try {
-    const authJson = await readAuthConfig()
-    if (isError(authJson)) throw authJson
-    const token = getOptionalParameter(args, ['--token', '-t'], 'PRISMA_TOKEN') || authJson.token
+    let token = getOptionalParameter(args, ['--token', '-t'], 'PRISMA_TOKEN') as string
+    if (!token) {
+      const authJson = await readAuthConfig()
+      if (isError(authJson)) throw authJson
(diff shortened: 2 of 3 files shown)


packages/cli/src/platform/logout.ts
```diff
@@ -8,8 +8,9 @@ export class Logout implements Command {
   }
 
   public async parse() {
-    const { token } = await readAuthConfig()
-    if (!token) {
+    const authJson = await readAuthConfig()
+    if (isError(authJson)) throw authJson
```
Jolg42, line 12: I don't understand why this should throw here in logout 🤔 
It would better to delete the config instead I think

packages/cli/src/utils/platform.ts
```diff
@@ -71,9 +71,9 @@ export const ErrorPlatformUnauthorized = new Error(
 
 export const getPlatformTokenOrThrow = async <$Args extends Record<string, unknown>>(args: $Args) => {
   try {
-    const token =
-      getOptionalParameter(args, ['--token', '-t'], 'PRISMA_TOKEN') ||
-      (await readAuthConfig().catch(() => ({ token: '' }))).token
+    const authJson = await readAuthConfig()
+    if (isError(authJson)) throw authJson
+    const token = getOptionalParameter(args, ['--token', '-t'], 'PRISMA_TOKEN') || authJson.token
```
Jolg42, line 76: The order gets weird here I think, let me explain:

The order of preference for the token is mentioned in https://github.com/prisma/prisma/pull/22144
1. Attempt to read value from arguments `--token` or `-t`
2. Attempt to read value from environment variable `PRISMA_TOKEN`
3. Attempt to read token stored in `auth.json` on the disk
4. If none of the above is present, throw `ErrorPlatformUnauthorized` 

But here, if `--token` is passed, then the auth config will first be read and could throw.

packages/cli/src/utils/platform.ts
```diff
@@ -146,27 +146,77 @@ export const configDirectoryPath = new XdgAppPaths('prisma-platform-cli').config
...
 export async function writeAuthConfig(data: AuthConfig) {
-  await fs.mkdirp(configDirectoryPath)
-
-  return await fs.writeJSON(authConfigPath, data)
+  try {
+    await fs.mkdirp(configDirectoryPath)
+    return await fs.writeJSON(authConfigPath, data)
+  } catch (error) {
+    debug('Error from writeAuthConfig()', error)
+    return error as Error
```
Jolg42, line 154: nit: This could be done differently
```suggestion
  } catch (error: Error) {
    debug('Error from writeAuthConfig()', error)
    return error
```

sampolahtinen: True! Tho both ways are a bit bad 🙈 Better way would be to deal with unknown type and if it does not extend error, then convert it to an error

packages/cli/src/utils/platform.ts
sampolahtinen: Let me improve this a bit!

packages/cli/src/utils/platform.ts
Jolg42: A source of problems could be that if we have mixed behaviors throw vs return an Error we can easily forget to handle them?

Jolg42 (approved):

PR #22167, opened 2023-11-30
chore(client/data-proxy): update InvalidDatasourceError code

marcoieni:
Update this error code to show the Accelerate [error](https://www.prisma.io/docs/reference/api-reference/error-reference#p6001-invaliddatasource) instead of the Data Proxy one.

This will also affect Data Proxy users. The only inconvenience will be that they see this error code instead of the old one when they have an invalid URL. I don't think it's too annoying. 👍


Jolg42 (approved):
Shortly discussed in [Internal Slack](https://prisma-company.slack.com/archives/C03R3PJN004/p1701362818022559)

PR #22189, opened 2023-12-01
ci: Change func/legacy job name to func & legacy

SevInf:
Slash confuses me every time, my brain just reads it as "legacy
functional tests" and then I spend next couple of minute trying to find
modern ones.

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -107,7 +107,7 @@ jobs:
   # CLIENT (functional and legacy tests without types test)
   #
   client:
-    name: Client func & legacy-notypes
+    name: Client func&legacy-notypes
     timeout-minutes: ${{ inputs.jobTimeout }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-template.yml
```diff
@@ -107,7 +107,7 @@ jobs:
   # CLIENT (functional and legacy tests without types test)
   #
   client:
-    name: Client func/legacy-notypes
+    name: Client func & legacy-notypes
```
Jolg42, line 110:
```suggestion
    name: Client func&legacy-notypes
```

Shorter would be better because it cuts before we can see if it's "l..." for "library" here for example
https://github.com/prisma/prisma/actions/runs/7059347363/job/19216734376?pr=22189
Screenshot 2023-12-01 at 12 30 50

Jolg42 (approved):

PR #22190, opened 2023-12-01
feat(CLI): redact token argument in platform commands

sampolahtinen:
This PR introduces redaction of `--token` argument used in platform commands.

In addition, the alias `-t` has been removed for the simplicity.

### DIFF
--- a/packages/cli/src/utils/platform.ts
+++ b/packages/cli/src/utils/platform.ts
@@ -15,19 +15,16 @@ export const platformParameters = {
     // TODO Remove this from global once we have a way for parents to strip out flags upon parsing.
     '--early-access-feature': Boolean,
     '--token': String,
-    '-t': '--token',
   },
   workspace: {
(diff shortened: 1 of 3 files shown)


Jolg42 (approved):
✨

packages/cli/src/utils/platform.ts
```diff
@@ -15,19 +15,16 @@ export const platformParameters = {
     // TODO Remove this from global once we have a way for parents to strip out flags upon parsing.
     '--early-access-feature': Boolean,
     '--token': String,
-    '-t': '--token',
```
janpio, line 18: Huh, this is unexpected. Why?

janpio: Ah, so Prisma CLI or Checkpoint does not support short options yet?

If so, we should create an issue because that is quite a DX thing to have to drop. Maybe @sampolahtinen can do a PR on checkpoint as well.

PR #22191, opened 2023-12-01
chore(CLI): shorten early access flag

sampolahtinen:
This PR renames the `--early-access-feature` to `--early-access`


packages/migrate/src/utils/flagErrors.ts
```diff
@@ -23,7 +23,7 @@ export class EarlyAccessFlagError extends Error {
   constructor() {
```
sampolahtinen, line 23: Let's just keep this error here for the simplicity.

Jolg42 (approved):

PR #22192, opened 2023-12-01
feat(prisma): platform signup attribution

sampolahtinen:
This PR introduces  reading the signature and passing it to the console.


packages/cli/src/platform/login.ts
```diff
@@ -85,10 +92,15 @@ export class Login implements Command {
 
 const generateAuthSigninUrl = async (params: { connection: string; redirectTo: string }) => {
   const prismaClientVersion = await getInstalledPrismaClientVersion().catch(() => null)
-  const state = { client: `prisma@${prismaClientVersion}`, ...params }
+  const state = {
+    client: `prisma@${prismaClientVersion}`,
+    signature: await checkpoint.getSignature(),
```
sampolahtinen, line 97: @Jolg42  what happens if this fails?

Jolg42: I added a catch because in theory it can fail, it will be rare

Jolg42 (approved):

PR #22193, opened 2023-12-01
fix(CLI): platform --early-access should work in any position

Jolg42:
Before this change, the flag needed to be provided at the end.

`prisma platform --early-access login` was not working as expected before this change

Also instead of help output, the output was just `""`, this adds a placeholder.


packages/cli/src/platform/$.ts
```diff
@@ -13,7 +13,19 @@ export class $ implements Command {
...
+    // Since `dispatchToSubCommand`
+    // assumes that the first element of the array to be the command
+    // we must remove the flag before
+    //
+    // It makes it possible to run, for example:
+    // prisma platform --early-access login
+    // prisma platform login --early-access
+    const argvWithoutEarlyAccess = (argv = argv.filter(function (it) {
+      return it !== '--early-access'
+    }))
```
sampolahtinen, line 26: Nit: could we use arrow function here?

sampolahtinen (approved):

PR #22194, opened 2023-12-01
fix(cli:debug): load .env file if it exists

### LINKED ISSUE #21968: prisma debug command does not show env variables declared in `.env` file
### Bug description

I added a `DEBUG="*"` env variable defined in my `.env` - when I invoke the `npx prisma debug` command, the output doesn't show me that I have DEBUG env variable set (Attached Screenshot)
But, if I explicitly set `export DEBUG=*` in my command line, the command correctly shows that I have DEBUG env variable set.

Screenshot 2023-11-15 at 4 48 05 PM
Screenshot 2023-11-15 at 4 44 48 PM

### How to reproduce

1. Have an env variable called `DEBUG="*"` in the `.env` file.
2. Execute `npx prisma debug` command.

### Expected behavior

The value of `DEBUG` variable is correctly shown in output.

### Prisma information

N/A

### Environment & setup

In every environment.

### Prisma Version

```
prisma                  : 5.6.0
@prisma/client          : 5.6.0
Node.js                 : v18.17.1
Query Engine (Node-API) : libquery-engine e95e739751f42d8ca026f6b910f5a2dc5adeaeee (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli e95e739751f42d8ca026f6b910f5a2dc5adeaeee (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.6.0-32.e95e739751f42d8ca026f6b910f5a2dc5adeaeee
Default Engines Hash    : e95e739751f42d8ca026f6b910f5a2dc5adeaeee
Studio                  : 0.495.0
```

Jolg42:
It was not as easy as it looks, there were some rabbit holes on the way :smile:

Closes https://github.com/prisma/prisma/issues/21968


packages/cli/src/DebugInfo.ts
```diff
@@ -1,5 +1,15 @@
...
-import { arg, format, getSchemaPath, HelpError, isCi, isError, isInteractive, link } from '@prisma/internals'
+import {
+  arg,
+  format,
+  getSchemaPath,
+  HelpError,
+  isCi,
+  isError,
+  isInteractive,
+  link,
```
janpio, line 10: Are these necessary?

packages/cli/src/DebugInfo.ts
```diff
@@ -41,6 +51,10 @@ export class DebugInfo implements Command {
       return this.help()
     }
 
+    // Loads the `.env` file if it exists
+    // and prints a line to stdout to inform the users
+    loadEnvFile(undefined, true)
```
janpio, line 56: I see `undefined` seems to be normal here and taken from `packages/cli/src/Version.ts`.

packages/cli/src/DebugInfo.ts
Jolg42: What do you mean?

Jolg42: Yes, all of them are used here

packages/cli/src/DebugInfo.ts
Jolg42: Switching to `loadEnvFile(args['--schema'], true)`

janpio: Should we rename that function maybe to transport all that is in that comment?
`loadEnvFileIfExistsAndPrintToStdOut()`

janpio (commented):
👍

packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -330,6 +330,98 @@ describe('debug', () => {
...
+  it('should read the .env file if it exists', async () => {
+    ctx.fixture('dotenv-debug-cmd')
+
+    // To make sure the terminal is always detected
+    // as non interactive, locally and in CI
+    process.env.TERM = 'dumb'
+
+    const result = await DebugInfo.new().parse([])
+
+    expect(result).not.toContain('this_is_private')
```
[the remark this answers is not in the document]
Jolg42: Because "why not". It's a safety check, there is no reason for it based on the implementation.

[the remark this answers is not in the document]
janpio: But why _could_ there be `this_is_private` in the result? Why not `foo`?

packages/cli/src/DebugInfo.ts
Jolg42: I destructured the params instead

packages/cli/src/__tests__/commands/DebugInfo.test.ts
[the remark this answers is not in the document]
Jolg42: Because I set `this_is_private` in the `.env` file as value.

packages/internals/src/utils/tryLoadEnvs.ts
```diff
@@ -120,20 +120,22 @@ export function loadEnv(envPath: string | null | undefined): DotenvLoadEnvResult
...
+      // Set to any value to enable
+      // Example for empty .env file
+      // [dotenv][DEBUG] did not match key and value when parsing line 1:
+      //
+      // Value needs to be null or undefined, false is truthy
+      // https://github.com/motdotla/dotenv/blob/7301ac9be0b2c766f865bbe24280bf82586d25aa/lib/main.js#L89-L91
+      debug: process.env.DOTENV_CONFIG_DEBUG ? true : undefined,
+    })
+
+    const dotenvExpandOutput = dotenvExpand(dotenvOutput)
```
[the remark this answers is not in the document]
Jolg42: Totally

packages/cli/src/__tests__/commands/DebugInfo.test.ts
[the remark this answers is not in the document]
janpio: Ah, so checking if it loads unrelated env vars 👍

packages/internals/src/utils/loadEnvFile.ts
```diff
@@ -5,11 +5,14 @@ import { tryLoadEnvs } from './tryLoadEnvs'
  * Read .env file only if next to schema.prisma
  * .env found: print to console it's relative path
  */
-export function loadEnvFile(schemaPath?: string, print = false) {
+export function loadEnvFile({
+  schemaPath,
+  printMessage = false,
+}: { schemaPath?: string; printMessage?: boolean } = {}) {
```
janpio, line 11: So much nicer and clearer now when looking at usage!

packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -330,6 +331,98 @@ describe('debug', () => {
...
+
+      For downloading engines
+      - PRISMA_ENGINES_MIRROR: \`from_env_file\`
+      - PRISMA_BINARIES_MIRROR (deprecated): \`from_env_file\`
+      - PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING: \`from_env_file\`
+      - BINARY_DOWNLOAD_VERSION: \`from_env_file\`
+
+      For configuring the Query Engine Type
+      - PRISMA_CLI_QUERY_ENGINE_TYPE:
+      - PRISMA_CLIENT_ENGINE_TYPE:
```
janpio, line 388: Seems this is overwritten by the test setup, so can have different values.

packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -330,6 +331,98 @@ describe('debug', () => {
...
+      -- Local cache directory for engines files --
+      Path: REDACTED_PATH
+
+      -- Environment variables --
+      When not set, the line is dimmed and no value is displayed.
+      When set, the line is bold and the value is inside the \`\` backticks.
+
+      For general debugging
+      - CI: \`true\`
+      - DEBUG: \`prisma*\`
```
janpio, line 359: Seems this is already set when the .env file is loaded, hence has a different value than set in there.

packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -55,19 +55,17 @@ const envVars = {
 }
 
 describe('debug', () => {
-  beforeEach(() => {
-    process.env = { ...originalEnv, ...envVars }
-  })
-  afterAll(() => {
+  // clean up env vars after each individual test
```
millsp, line 58: This may be useful to ensure no other env vars leak
```suggestion
  // make sure the env is empty before each test
  beforeEach(() => {
    process.env = {}
  })
  // clean up env vars after each individual test
```

millsp (approved):
Looks good! (See suggestion to clear the issues about env pollution within tests)

PR #22236, opened 2023-12-04
feat(cli): platform root help

sampolahtinen:
This PR adds:
- Root help for the platform namespace
- A new utility that generates help output


packages/cli/src/utils/platform.ts
```diff
@@ -119,19 +119,71 @@ export const platformRequestOrThrow = async (params: {
   return json
 }
 
-export const dispatchToSubCommand = async (commands: Commands, argv: string[]) => {
+export const dispatchToSubCommand = async (commands: Commands, argv: string[], help?: (error?: string) => string) => {
   const commandName = argv[0]
-  // Placeholder for now
-  const helpPlaceholder = 'Coming soon: help output for this command.'
-  if (!commandName) return helpPlaceholder
+  if (['-h', '--help'].includes(commandName)) return help?.() ?? ''
```
sampolahtinen, line 124: I am still bouncing back and forth whether `help` method should be passed as an argument to `dispatchToSubCommand`

Benefits:
- Wraps the logic for all platform commands avoiding repeating logic and error checking on the command level

Con:
- This pattern does not work for commands without sub commands. These commands would anyways need to add logic to their command definition.  As a result we would have two different ways of handling help calls

jasonkuhrt (approved):
LGTM!

Jolg42 (approved):

packages/cli/src/__tests__/commands/platform/Platform.test.ts
```diff
@@ -19,4 +20,47 @@ describe('--early-access flag', () => {
...
+    const resultIsError = isError(result)
+    expect(resultIsError).toBeFalsy()
+    expect(spy).toHaveBeenCalledTimes(1)
+    spy.mockRestore()
+  })
+
+  it('should output the help', async () => {
+    const commandInstance = $.new({})
+    const result = await commandInstance.parse(['--early-access'])
+    expect(result).toMatchInlineSnapshot(`
```
sampolahtinen, line 37: Ill improve this a bit. It would be better to call the `createHelp` utility

PR #22242, opened 2023-12-04
feat(cli): platform command success outputs

sampolahtinen:
This PR introduces prettier outputs for platform commands

### DIFF
--- a/packages/cli/src/platform/accelerate/disable.ts
+++ b/packages/cli/src/platform/accelerate/disable.ts
@@ -6,6 +6,7 @@ import {
   getRequiredParameter,
   platformParameters,
   platformRequestOrThrow,
+  successMessage,
 } from '../../utils/platform'
 
 export class Disable implements Command {
@@ -36,9 +37,10 @@ export class Disable implements Command {
     if (payload.error?.message) {
       throw new Error(payload.error.message)
     }
-    // green "success" text
     log(
-      `Success! Accelerate disabled. Prisma clients connected to ${args['--project']} will not be able to send queries through Accelerate.`,
+      successMessage(
+        `Accelerate disabled. Prisma clients connected to ${args['--project']} will not be able to send queries through Accelerate.`,
+      ),
     )
     return ''
   }
--- a/packages/cli/src/platform/accelerate/disable.ts
+++ b/packages/cli/src/platform/accelerate/disable.ts
@@ -1,4 +1,5 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { log } from 'console'
 
 import {
   getPlatformTokenOrThrow,
@@ -23,7 +24,7 @@ export class Disable implements Command {
     if (isError(workspace)) return workspace
     const project = getRequiredParameter(args, ['--project', '-p'])
     if (isError(project)) return project
-    return platformRequestOrThrow({
+    await platformRequestOrThrow<{ data: {}; error: null }>({
       token,
       path: `/${workspace}/${project}/accelerate/settings`,
       route: '_app.$organizationId_.$projectId.accelerate.settings',
@@ -31,6 +32,11 @@ export class Disable implements Command {
         intent: 'disable',
         projectId: project,
       },
-    }) as Promise<any>
+    })
+    // green "success" text
+    log(
+      `Success! Accelerate disabled. Prisma clients connected to ${args['--project']} will not be able to send queries through Accelerate.`,
+    )
+    return ''
   }
 }
--- a/packages/cli/src/platform/accelerate/disable.ts
+++ b/packages/cli/src/platform/accelerate/disable.ts
@@ -1,5 +1,4 @@
 import { arg, Command, isError } from '@prisma/internals'
-import { log } from 'console'
 
 import {
   getPlatformTokenOrThrow,
@@ -37,7 +36,7 @@ export class Disable implements Command {
     if (payload.error?.message) {
       throw new Error(payload.error.message)
     }
-    log(
+    console.info(
       successMessage(
         `Accelerate disabled. Prisma clients connected to ${args['--project']} will not be able to send queries through Accelerate.`,
       ),
--- a/packages/cli/src/platform/accelerate/disable.ts
+++ b/packages/cli/src/platform/accelerate/disable.ts
@@ -24,7 +24,7 @@ export class Disable implements Command {
     if (isError(workspace)) return workspace
     const project = getRequiredParameter(args, ['--project', '-p'])
     if (isError(project)) return project
-    await platformRequestOrThrow<{ data: {}; error: null }>({
+    const payload = await platformRequestOrThrow<{ data: {}; error: null | { message: string } }>({
       token,
       path: `/${workspace}/${project}/accelerate/settings`,
       route: '_app.$organizationId_.$projectId.accelerate.settings',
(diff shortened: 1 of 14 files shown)


packages/cli/src/platform/workspace/show.ts
```diff
@@ -1,4 +1,5 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { table } from 'console'
```
Jolg42, line 2: If you could avoid this and use `console.table`, we don't use this at all in the codebase and it's better to be consistent.

packages/cli/src/platform/accelerate/enable.ts
```diff
@@ -1,10 +1,12 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { log } from 'console'
```
Jolg42, line 2: If you could avoid this and use `console.info`, we don't use this at all in the codebase and it's better to be consistent.

packages/cli/src/platform/apikey/create.ts
```diff
@@ -1,11 +1,13 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { log } from 'console'
```
Jolg42, line 2: If you could avoid this and use `console.info`, we don't use this at all in the codebase and it's better to be consistent.

packages/cli/src/platform/apikey/delete.ts
```diff
@@ -1,10 +1,12 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { log } from 'console'
```
Jolg42, line 2: If you could avoid this and use `console.info`, we don't use this at all in the codebase and it's better to be consistent.

packages/cli/src/platform/apikey/show.ts
```diff
@@ -1,4 +1,5 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { table } from 'console'
```
Jolg42, line 2: If you could avoid this and use `console.table`, we don't use this at all in the codebase and it's better to be consistent.

packages/cli/src/platform/project/create.ts
```diff
@@ -1,13 +1,25 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { log } from 'console'
```
Jolg42, line 2: If you could avoid this and use `console.info`, we don't use this at all in the codebase and it's better to be consistent.

packages/cli/src/platform/project/show.ts
```diff
@@ -1,4 +1,5 @@
 import { arg, Command, isError } from '@prisma/internals'
+import { table } from 'console'
```
Jolg42, line 2: If you could avoid this and use `console.table`, we don't use this at all in the codebase and it's better to be consistent.

packages/cli/src/platform/project/delete.ts
```diff
@@ -21,15 +38,21 @@ export class Delete implements Command {
...
       route: '_app.$organizationId_.$projectId.settings.general',
       payload: {
         intent: 'delete',
       },
-    }) as Promise<any>
+    })
+    if (payload.error) throw new Error(`${payload.error.name}: ${payload.error.message}`)
+
+    console.info(successMessage(`Project ${payload.data.displayName} - ${payload.data.id} deleted.`))
+    return ``
```
tatianainama, line 56:
```suggestion
    return ''
```

packages/cli/src/platform/accelerate/disable.ts
```diff
@@ -23,14 +24,23 @@ export class Disable implements Command {
...
+    })
+    if (payload.error?.message) {
+      throw new Error(payload.error.message)
+    }
+    console.info(
+      successMessage(
+        `Accelerate disabled. Prisma clients connected to ${args['--project']} will not be able to send queries through Accelerate.`,
+      ),
+    )
+    return ''
```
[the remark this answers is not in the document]
tatianainama: true 😄 I'll update my commands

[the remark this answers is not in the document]
sampolahtinen: Ill update all the rest!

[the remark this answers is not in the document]
sampolahtinen: For the commands printing tables, I just did a simple:
```ts
console.table(...data)
return ''
```

packages/cli/src/utils/platform.ts
```diff
@@ -117,6 +117,11 @@ export const platformRequestOrThrow = async (params: {
     body: payload ? JSON.stringify(payload) : undefined,
   })
   const text = await response.text()
+
+  if (response.status > 400) {
```
jasonkuhrt, line 121:
```suggestion
  if (response.status >= 400) {
```

?

sampolahtinen: Ill commit this!

packages/cli/src/platform/auth/show.ts
```diff
@@ -13,10 +14,20 @@ export class Show implements Command {
...
+    const payload = await platformRequestOrThrow<{
+      user: { id: string; handle: string; email: string; displayName: string }
+    }>({
       token,
       path: `/settings/account`,
       route: '_app._user.settings.account',
-    }) as Promise<any>
+    })
+    console.log(`Currently authenticated as ${green(payload.user.email)}`)
+    console.log('')
```
Jolg42, line 25:
```suggestion
    console.info(`Currently authenticated as ${green(payload.user.email)}\n`)
```
Use `console.info` when you want to log something to the user. It makes it easier to spot random `console.log` from debugging for example.

packages/cli/src/platform/login.ts
```diff
@@ -27,6 +27,13 @@ export class Login implements Command {
   }
 
   public async parse() {
+    const authConfig = await readAuthConfig()
+    if (isError(authConfig)) throw authConfig
+    if (authConfig.token)
+      return `Already authenticated. Run ${green(
+        '`prisma platform auth show --early-access`',
+      )} to see the current user.`
```
Jolg42, line 35:
```suggestion
      return `Already authenticated. Run ${green(
        getCommandWithExecutor('prisma platform auth show --early-access'),
      )} to see the current user.`
```
Use `getCommandWithExecutor()` for displaying commands

packages/cli/src/platform/login.ts
```diff
@@ -84,9 +91,7 @@ export class Login implements Command {
       const result = await writeAuthConfig({ token: authResult.token })
       if (isError(result)) throw result
 
-      console.info('Authenticated successfully as:')
-      console.info(JSON.stringify(authResult.user, null, 4))
-      return ''
+      return successMessage(`Authentication complete for ${authResult.user.email}`)
```
Jolg42, line 94:
```suggestion
      return successMessage(`Authentication successful for ${authResult.user.email}`)
```

packages/cli/src/platform/login.ts
Jolg42: and add the import

packages/cli/src/utils/platform.ts
```diff
@@ -117,6 +117,11 @@ export const platformRequestOrThrow = async (params: {
     body: payload ? JSON.stringify(payload) : undefined,
   })
   const text = await response.text()
+
+  if (response.status >= 400) {
+    throw new Error(text)
+  }
```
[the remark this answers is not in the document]
Jolg42: 🤔 I still find this weird, because it's definitely API related and a 3xx would break the CLI with no error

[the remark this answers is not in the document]
Jolg42: Trying to think about this, maybe it's a bit too late for me 😄 
In the end, I think 300 errors are actually ok they will not break anything (unless it redirects to something broken)

Jolg42 (approved):

PR #22246, opened 2023-12-04
test(client): add --preview-features to test setup CLI

millsp:
Makes it possible to automatically inject any test schema with a specific list of preview features.

For example, you could run it like this:

```pnpm run test:functional --provider postgresql --preview-features distinctOn```

The preview features are passed in one go and are comma separated:

```pnpm run test:functional --preview-features superFeature,amazingFeature,coolFeature```

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -316,7 +316,12 @@ jobs:
           engine-hash: ${{ inputs.engineHash }}
           skip-tsc: false
 
-      - run: pnpm run test:functional --flavor ${{ matrix.flavor }} --shard ${{ matrix.shard }} --engine-type ${{ matrix.engine-type }} --runInBand
+      - run: pnpm run test:functional \
+          --flavor ${{ matrix.flavor }} \
+          --shard ${{ matrix.shard }} \
+          --engine-type ${{ matrix.engine-type }} \
+          --preview-features driverAdapters \
+          --runInBand
         working-directory: packages/client
         env:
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: .github/workflows/test-template.yml [hunk 1]
```diff
@@ -316,7 +316,7 @@ jobs:
           engine-hash: ${{ inputs.engineHash }}
           skip-tsc: false
 
-      - run: pnpm run test:functional --flavor ${{ matrix.flavor }} --shard ${{ matrix.shard }} --engine-type ${{ matrix.engine-type }} --runInBand
+      - run: pnpm run test:functional --flavor ${{ matrix.flavor }} --shard ${{ matrix.shard }} --engine-type ${{ matrix.engine-type }} --preview-features driverAdapters --runInBand
```
millsp, line 319: Already used by `driverAdapters` now

Jolg42 (approved):
Nice one ✨

.github/workflows/test-template.yml
janpio, line 319: Does this _add_ another run with preview feature `driverAdapters` or does this add the preview flag to this run and change what is actually tests?

millsp: It is not a separate run or test suite name, and it is completely unknown to the tests. This is the same run as before and we now inject the preview feature via the CLI direct instead of via previously [hardcoded conditions](https://github.com/prisma/prisma/pull/22246/files#diff-1d8cafa403bb80bfd8c3562210e02f7ffd423c74e5a9064984a9b0e79e48ce06L178).

PR #22255, opened 2023-12-05
fix(cli): Platform commands fixes

tatianainama:
This PR fixes some responses from the cli platform commands:
- [x] `accelerate enable` now fails if project/workspace is invalid
- [x] `accelerate enable --apikey` flag now autogenerates a name for key generation
- [x] `apikey create` now returns the apikey correctly
- [x] make all `console.table` display the same headers and fix column order


sampolahtinen (approved):
LGTM! Thank you for these improvements! 🎉

Jolg42 (approved):
Nice ✨

PR #22256, opened 2023-12-05
feat(cli): Display full connection string for accelerate enable command

tatianainama:
This PR improves the response of `accelerate enable` when the param `--apikey` is used.
New output looks like the following: 

```
✦ prisma platform accelerate enable [ ... ] --apikey

Success! Accelerate enabled. Use this generated API key in your Accelerate connection string to authenticate requests: 
prisma://accelerate.prisma-data.net?api_key=eyJhbGciOi[redacted]
```

### DIFF
--- a/packages/cli/src/platform/accelerate/enable.ts
+++ b/packages/cli/src/platform/accelerate/enable.ts
@@ -1,6 +1,7 @@
 import { arg, Command, isError } from '@prisma/internals'
 
 import {
+  generateConnectionString,
   getOptionalParameter,
   getPlatformTokenOrThrow,
   getRequiredParameter,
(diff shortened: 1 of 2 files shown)


Jolg42 (approved):

packages/cli/src/platform/accelerate/enable.ts
```diff
@@ -61,7 +62,9 @@ export class Enable implements Command {
         throw new Error(payload.error.message)
       }
       return successMessage(
-        `Accelerate enabled. Use this generated API key in your Accelerate connection string to authenticate requests: ${payload.data.tenantAPIKey}`,
+        `Accelerate enabled. Use this generated API key in your Accelerate connection string to authenticate requests: \n${generateConnectionString(
+          payload.data.tenantAPIKey,
+        )}`,
```
Jolg42, line 67: We could also add a link (we would need a shortlink) to https://www.prisma.io/docs/data-platform/accelerate/getting-started ?

tatianainama: oh, to display users what are the next steps? that looks good ✨
I'll note this for the next iteration since we might want some product input here, what do you think?

Jolg42: Sounds good!

PR #22257, opened 2023-12-05
fix(cli): rename platform command option --display-name to --name

sampolahtinen:
This PR simply renames --display-name to be --name in the platform cli commands

### DIFF
--- a/packages/cli/src/platform/apikey/show.ts
+++ b/packages/cli/src/platform/apikey/show.ts
@@ -15,8 +15,6 @@ export class Show implements Command {
   public async parse(argv: string[]) {
     const args = arg(argv, {
       ...platformParameters.project,
-      '--display-name': String,
-      '-d': '--display-name',
     })
     if (isError(args)) return args
     const token = await getPlatformTokenOrThrow(args)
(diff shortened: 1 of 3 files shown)


packages/cli/src/platform/apikey/show.ts
```diff
@@ -15,8 +15,6 @@ export class Show implements Command {
   public async parse(argv: string[]) {
     const args = arg(argv, {
       ...platformParameters.project,
-      '--display-name': String,
-      '-d': '--display-name',
```
sampolahtinen, line 19: This was not used.

Jolg42 (approved):

tatianainama (approved):

tatianainama:
@sampolahtinen I just noticed the command `auth show` displays the following:

```
id                    :  <id>
handle           : <handle>
email              : <email>
displayName : <displayName>
```

Maybe not in this PR? but I think we should also change that `displayName` to be `name`

jasonkuhrt:
@luanvdw I don't think using `--name` is a good idea, as it becomes harder to understand if this is a `handle` or not. Many other interfaces out there treat `name` with the semantics of `handle` and that creates expectations like "I can find by this name" and "Choosing the right name matters and should not change often" etc.

PR #22258, opened 2023-12-05
feat(cli): move platform login and logout under auth command

sampolahtinen:
This PR moves platform login and logout under the `auth` command


packages/cli/src/bin.ts
```diff
@@ -84,12 +84,12 @@ async function main(): Promise<number> {
...
       init: Init.new(),
       platform: Platform.$.new({
-        login: Platform.Login.new(),
-        logout: Platform.Logout.new(),
         workspace: Platform.Workspace.$.new({
           show: Platform.Workspace.Show.new(),
         }),
         auth: Platform.Auth.$.new({
+          login: Platform.Login.new(),
+          logout: Platform.Logout.new(),
```
tatianainama, line 92: should we also change the methods to be:
```suggestion
          login: Platform.Auth.Login.new(),
          logout: Platform.Auth.Logout.new(),
```
so it can be consistent with the other commands and hierarchies

sampolahtinen: Yes absolutely! I moved the files and also updated some old copy references

Jolg42 (approved):
Looks good, make sure you update the docs at https://pris.ly/platform-cli-docs

tatianainama:
> Looks good, make sure you update the docs at https://pris.ly/platform-cli-docs

I'll take care of this 👍

tatianainama (approved):

PR #22267, opened 2023-12-05
feat: show auth commands as part of root help

sampolahtinen:
This PR proposes to include auth commands as part of the root help.


packages/cli/src/__tests__/commands/platform/Platform.test.ts
```diff
@@ -42,8 +42,9 @@ describe('--early-access flag', () => {
 
       Commands
 
-                login   Logs into your Prisma Data Platform account or creates a new one with GitHub authentication
-               logout   Logs out of your Prisma Data Platform account
+           auth login   Logs into your Prisma Data Platform account or creates a new one with GitHub authentication
+          auth logout   Logs out of your Prisma Data Platform account
+            auth show   Displays information about the currently authenticated user
```
Jolg42, line 47: This is weird to list the subcommands like this only for auth

janpio (approved):

Jolg42 (approved):

PR #22272, opened 2023-12-05
fix(e2e): fix failing studio-1128-spawn-enoent test

aqrln:
This test was broken and hanging infinitely after
https://github.com/prisma/prisma/pull/22192 because the namespace of
debug logs for the `prisma studio` command was changed from
`prisma:studio` to `prisma:cli:studio` and this test relied on the old
one.

It did not fail in the PR itself because the changes were localized
to Prisma CLI but the test itself is in the client package, so it was
skipped. We should probably move the test to the CLI package if it's
feasible.

I also added a timeout to fail the test after 5 minutes with an error
instead of hanging infinitely.

### DIFF
--- a/packages/client/tests/e2e/issues/studio-1128-spawn-enoent/_steps.ts
+++ b/packages/client/tests/e2e/issues/studio-1128-spawn-enoent/_steps.ts
@@ -9,8 +9,12 @@ void executeSteps({
   },
 
   test: async () => {
-    process.env.DEBUG = 'prisma:studio'
+    process.env.DEBUG = 'prisma:cli:studio'
     const studio = $`pnpm exec prisma studio`
+
+    const timer = setTimeout(() => {
+      throw new Error('"prisma studio" didn\'t request to open the browser')
+    }, 300_000)
 
     for await (const output of studio.stderr) {
       // Exit as soon as xdg-open subprocess in studio either spawns or reports failure to spawn
@@ -18,6 +22,7 @@ void executeSteps({
         output.includes('requested to open the url http://localhost:5555') ||
         output.includes('failed to open the url http://localhost:5555 in browser')
       ) {
+        clearTimeout(timer)
         await studio.nothrow().kill()
         return
       }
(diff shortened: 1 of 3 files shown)


millsp (approved):

packages/client/tests/e2e/issues/studio-1128-spawn-enoent/_steps.ts
```diff
@@ -9,15 +9,20 @@ void executeSteps({
   },
 
   test: async () => {
-    process.env.DEBUG = 'prisma:studio'
+    process.env.DEBUG = 'prisma:cli:studio'
```
Jolg42, line 12: I did that change 😅 

Thanks for the investigation & cleanup!

packages/client/tests/e2e/issues/studio-1128-spawn-enoent/_steps.ts
```diff
@@ -9,15 +9,20 @@ void executeSteps({
...
     const studio = $`pnpm exec prisma studio`
 
+    const timer = setTimeout(() => {
+      throw new Error('"prisma studio" didn\'t request to open the browser')
+    }, 300_000)
```
Jolg42, line 17:
```suggestion
    }, 30_000)
```

We could make it 30s, that should be reasonable?

Jolg42 (approved):

millsp:
Another test is failing, it may need an investigation, though all ecosystem-tests are passing so it it may not be critical.

packages/client/tests/e2e/issues/studio-1128-spawn-enoent/_steps.ts
aqrln: I think 30 seconds is a little bit too little. I could imagine studio maybe sometimes taking longer than that to start on slow CI under load — hopefully not but still not too far fetched.

This timeout does not influence the run time of the test in success case, only in failure case, so there's no reason to not make it generous enough to not be flaky.

The idea is to have proper error message instead of a mysterious timeout after half an hour, the test taking 5 minutes only if it’s broken (and being instantaneous otherwise) is fine IMO.

Issue #18292, opened 2023-03-12
Node hangs when `console.log(new PrismaClient())`


sc0ch:
### Bug description

Linux - node process hangs (stuck on prisma:client:libraryEngine  internalSetup +0ms)
Windows - javascript out of memory error.
Any schema. 
Affected versions of @prisma/client > 4.7.0

### How to reproduce

1. npx prisma init
2. npx prisma generate (npx prisma db push)
3. Create index.mjs with:
	```
	import { PrismaClient } from '@prisma/client'
	console.log(new PrismaClient())
	```
4. `export DEBUG="*" && node index.mjs` -> See error

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mysql"
  url      = env("DATABASE_URL")
}

model Session {
  id    Int    @id @default(autoincrement())
  key   String @unique
  value String

  @@map("sessions")
}
```

### Environment & setup

- OS: Windows 11, Ubuntu 22.04.2
- Database: MySQL
- Node.js version: 18.5.0, 19.7.0

### Prisma Version

```
prisma                  : 4.11.0
@prisma/client          : 4.11.0
Current platform        : debian-openssl-3.0.x
Query Engine (Node-API) : libquery-engine 8fde8fef4033376662cad983758335009d522acb (at node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Migration Engine        : migration-engine-cli 8fde8fef4033376662cad983758335009d522acb (at node_modules/@prisma/engines/migration-engine-debian-openssl-3.0.x)
Format Wasm             : @prisma/prisma-fmt-wasm 4.11.0-57.8fde8fef4033376662cad983758335009d522acb
Default Engines Hash    : 8fde8fef4033376662cad983758335009d522acb
Studio                  : 0.483.0
```

janpio:
Why do you want to `console.log(new PrismaClient())`?

AvIsBeastMC:
> Why do you want to `console.log(new PrismaClient())`?
fr

entrptaher:
I did this and then whenever I wanted to console.log(prisma) anywhere for any reason, it hung the whole process.

```
import { PrismaClient } from "@prisma/client";
let prisma = new PrismaClient();
console.log(prisma)
```

professor-l:
[...]

PR #22286, opened 2023-12-06
fix(client): set `enumerable: false` for `$parent`

Jolg42:
Related to https://github.com/prisma/prisma/issues/18292

This was started in my stash from a recent train ride :train:

When trying it in the `./sandbox` directory, `console.log(prisma)` works.

Please review, but do not merge, there is no test for this yet here.

### DIFF
--- a/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
+++ b/packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
@@ -41,11 +41,7 @@ const customInspect = Symbol.for('nodejs.util.inspect.custom')
  * @param layers
  * @returns
  */
-export function createCompositeProxy<T extends object>(
-  target: T,
-  layers: CompositeProxyLayer[],
-  disableInspect?: boolean,
-): T {
+export function createCompositeProxy<T extends object>(target: T, layers: CompositeProxyLayer[]): T {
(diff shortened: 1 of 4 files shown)


SevInf:
Could we check if just making it non-enumerable fixes the problem as well? Then we can fix it without extra settings

packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
```diff
@@ -104,6 +104,14 @@ export function createCompositeProxy<T extends object>(target: T, layers: Compos
             ...layer?.getPropertyDescriptor(prop),
           }
         }
+
+        if (prop === '$parent') {
```
SevInf, line 108: I think more appropriate way to do it would be to add optional way to define property descriptor for `addProperty` and set `enumerable: false` there. `createCompositeProxy` is meant to be very generic and know as little as possible about the contents of the objects it decorates.

Jolg42: Definitely! this was just me trying things out and isolate what works. Thanks for the suggestion

Jolg42:
Closing this, as it's not ready, also the latest state with `$parent` doesn't actually work I think, and I won't have time to do anything about it.

PR #22289, opened 2023-12-06
ci: always run e2e tests

millsp:
As @aqrln mentioned, pointed out to me in a DM, `client-e2e` depend on all the workspace packages, and we had some false positives because of our test skipping logic on CI. This then enables `e2e-tests` to always be run.

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -462,13 +462,13 @@ jobs:
     timeout-minutes: 35
     runs-on: ${{ matrix.os }}
 
-    if: ${{ contains(inputs.jobsToRun, '-all-') || contains(inputs.jobsToRun, '-client-') }}
+    # we always run this job, as it depends on all packages
 
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        node: [16]
+        node: [20]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -468,6 +468,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
+        # The Node.js version here is only used for the GitHub Actions job
+        # The version used by the tests is actually defined in
+        # ./packages/client/tests/e2e/_utils/standard.dockerfile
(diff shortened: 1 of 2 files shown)


.github/workflows/test-template.yml
```diff
@@ -462,13 +462,13 @@ jobs:
     timeout-minutes: 35
     runs-on: ${{ matrix.os }}
 
-    if: ${{ contains(inputs.jobsToRun, '-all-') || contains(inputs.jobsToRun, '-client-') }}
```
aqrln, line 465: should this be something like

```yaml
    if: |
      ${{
        contains(inputs.jobsToRun, '-all-') ||
        contains(inputs.jobsToRun, '-internals-') ||
        contains(inputs.jobsToRun, '-migrate-') ||
        contains(inputs.jobsToRun, '-cli-') ||
        contains(inputs.jobsToRun, '-client-')
      }}
```

?

millsp: Thanks for the proposal, I encoded your idea in a different way

.github/workflows/test-template.yml
```diff
@@ -462,13 +462,13 @@ jobs:
...
+    if: ${{ contains(inputs.jobsToRun, '-all-') || contains(inputs.jobsToRun, '-client-e2e-') }}
 
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        node: [16]
+        node: [20]
```
Jolg42, line 471: Why the switch now? (we still support Node.js 16)

I also want to switch, though for most cases I think it would be better to actually switch to 18 instead 🤔

aqrln (approved):

.github/workflows/test-template.yml
Jolg42: Right, then 20 is better with a comment

Jolg42, line 471:
```suggestion
	# The Node.js version here is only used for the GitHub Actions job
	# The version used by the tests is actually defined in
	# packages/client/tests/e2e/_utils/standard.dockerfile.
        node: [20]
```

Jolg42 (approved):
Thanks, please add the suggested comment before merging

PR #22338, opened 2023-12-11
feat(client): Implement `getConnectionInfo` method for driver adapters

SevInf:
Sibling to prisma/prisma-engines#4548.

Implements `getConnectionInfo` for PlanetScale, Neon and Pg adapters.
It is neccassary for getting engine test suite to work without URL in
schema file.

Implementation is the following:
- for postgres-based adapters, schema name can be specified as an
  optional parameter in adapter's constructor.
- for PlanetScale, DB name is parsed from connection url. HTTP client
  does not have a public method for getting the connection at the
  moment, so we are using the hack for now. I've requested a public
  method for getting it from PlanetScale people, hope we get it sorted
  out before release.

Contributes to  prisma/team-orm#662

### DIFF
--- a/packages/adapter-neon/src/neon.ts
+++ b/packages/adapter-neon/src/neon.ts
@@ -2,6 +2,7 @@
 import type neon from '@neondatabase/serverless'
 import type {
   ColumnType,
+  ConnectionInfo,
   DriverAdapter,
   Query,
   Queryable,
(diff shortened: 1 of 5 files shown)


janpio (commented):
See related discussion in https://prisma-company.slack.com/archives/C04TW4A2H8C/p1702373605197379

packages/adapter-neon/src/neon.ts
```diff
@@ -121,13 +122,23 @@ class NeonTransaction extends NeonWsQueryable<neon.PoolClient> implements Transa
   }
 }
 
+export type PrismaNeonOptions = {
+  schemaName?: string
```
janpio, line 126: I might suggest that we fully rename this from `schemaName` to `schema` in the public API to be consistent with the `schema` connection string parameter that we have in our connection strings.

No necessary to adapt yet, ad a higher level discussion is going on and I think this would be a simple search/replace.

Jolg42 (approved):
Looks good to me

packages/adapter-neon/src/neon.ts
SevInf: It was a quick change, done.

PR #22369, opened 2023-12-13
ci(macos): overwrite conflicting files rather than ignore errors

aqrln:
Overwrite the existing files to fix the Python issue instead of ignoring
all `brew install` errors.

Ref: https://github.com/prisma/prisma/pull/22339

### DIFF
--- a/.github/workflows/scripts/setup-mysql.sh
+++ b/.github/workflows/scripts/setup-mysql.sh
@@ -7,7 +7,7 @@ if [ "$RUNNER_OS" = "Windows" ]; then
 fi
 
 if [ "$RUNNER_OS" = "macOS" ]; then
-    brew install mysql || true
+    brew install --overwrite mysql
     brew services run mysql
 fi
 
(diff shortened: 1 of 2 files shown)


Jolg42:
It looks like tit's still failing with the original error 
```
==> Pouring python@3.12--3.12.1.ventura.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.12
Target /usr/local/bin/2to3-3.12
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.12'
```

Maybe try what was mentioned in https://prisma-company.slack.com/archives/C058VM009HT/p1702288460104659?thread_ts=1702287792.342269&cid=C058VM009HT
(https://github.com/actions/runner-images/issues/2322#issuecomment-749211076)

aqrln:
Ah well okay then, I thought it was gonna work but if it doesn't, I don't have capacity to look into it right now. Let's stay with Pierre's hotfix for now then.

Jolg42:
Same here, thanks for trying

PR #22426, opened 2023-12-15
chore(tech-debt): use "provider" and "adapter" instead of "flavour"/"flavor"

jkomyno:
This PR contributes to https://github.com/prisma/team-orm/issues/560.

closes: https://github.com/prisma/team-orm/issues/560.


packages/client/src/runtime/core/raw-query/rawQueryArgsMapper.ts
```diff
@@ -32,19 +31,11 @@ More Information: https://pris.ly/d/execute-raw
...
-  ({ clientMethod, activeProvider, driverAdapterProvider }: RawQueryArgsMapperInput) =>
+  ({ clientMethod, activeProvider }: RawQueryArgsMapperInput) =>
   (args: RawQueryArgs) => {
-    // TODO: remove this block, probably not needed anymore
-    if (driverAdapterProvider !== undefined) {
-      // When using the driver adapter, we need to use the flavour of the adapter,
-      // in order to avoid enumerating all the possible `@prisma/*` provider names.
-      activeProvider = driverAdapterProvider
-    }
-
```
jkomyno, line 47: Removed this while at it. This block was needed e.g. back when `@prisma/planetscale` was a thing.

jkomyno:
Follow up: I should update the code and see whether some more renames are necessary with the incoming code.

jkomyno:
I confirm this PR is rfr @Jolg42

packages/client/tests/functional/_utils/providers.ts
```diff
@@ -7,11 +7,13 @@ export enum Providers {
...
 
-export enum ProviderFlavors {
+export enum AdapterProviders {
   JS_PG = 'js_pg',
   JS_PLANETSCALE = 'js_planetscale',
   JS_NEON = 'js_neon',
   JS_LIBSQL = 'js_libsql',
+
+  // TODO: what to do with Vitess? It's not a driver adapter, but it's a flavor of MySQL.
   VITESS_8 = 'vitess_8',
```
Jolg42, line 17: Good question, for now it's an exception and that's ok I guess

Jolg42: That's why it was called "flavor" at first I think

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -277,21 +277,21 @@ function getDbUrl(provider: Providers): string {
...
- * Returns configured database URL for specified provider flavor, falling back to
- * `getDbUrl(provider)` if no flavor-specific URL is configured.
- * @param providerFlavor provider variant, e.g. `vitess` for `mysql`
+ * Returns configured database URL for specified provider, Driver Adapter, or provider variant (e.g., Vitess 8 is a known variant of the "mysql" provider),
+ * falling back to `getDbUrl(provider)` if no specific URL is configured.
+ * @param driverAdapter provider variant, e.g. `vitess` for `mysql`
  * @param provider provider supported by Prisma, e.g. `mysql`
  */
-function getDbUrlFromFlavor(providerFlavor: `${ProviderFlavors}` | undefined, provider: Providers): string {
+function getDbUrlFromFlavor(driverAdapterOrFlavour: `${AdapterProviders}` | undefined, provider: Providers): string {
```
Jolg42, line 285:
```suggestion
function getDbUrlFromFlavor(driverAdapterOrFlavor: `${AdapterProviders}` | undefined, provider: Providers): string {
```

Flavor or Flavour? both are used here

packages/client/tests/functional/driver-adapters/_utils/test-suite-options.ts
```diff
@@ -4,9 +4,9 @@ import { MatrixOptions } from '../../_utils/types'
  * Opt out from providers not specified in `driverAdaptersTestProviders` in `setupTestSuite`
  * (to avoid the error about missing providers in the test suite matrix).
  */
-export const optOutFromProvidersWithNoMatchingFlavours: MatrixOptions['optOut'] = {
+export const optOutFromProvidersWithNoMatchingDriverAdapters: MatrixOptions['optOut'] = {
   from: ['cockroachdb', 'sqlserver', 'mongodb'],
-  reason: 'no corresponding flavour of driver adapters yet',
+  reason: 'no availability of a Driver Adapter with these providers yet',
```
Jolg42, line 9: Off topic for this PR, but technically we could run cockroachdb through pg or neon adapter I think

Jolg42 (approved):
Nice cleanup!

For Vitess it's a bit weird now as it's also a database we want to run without a driver adapter.
But since it's improving the global state, we can merge this anyway and think about how we want to improve things later.

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
jkomyno: Good catch. Let's stick with the standard US-EN "flavor"

packages/client/tests/functional/driver-adapters/_utils/test-suite-options.ts
jkomyno: Possibly, but only if the PSL `provider` is `postgres`, not `cockroachdb`.

millsp:
I agree with Joël, this was really meant as a flavor, which in this context is a mechanism that provides a way to opt out of the conventional path (like running vitess). I don't disagree with the change, though leaving vitess as-is introduces inconsistency in the concepts.

PR #22430, opened 2023-12-15
ci(client): test with and without all preview features enabled

### LINKED ISSUE #22187: Allow to run whole client test suite with force-enabled preview features
At the moment for enabling a preview feature in a test suite we have to manually add it to each schema file. In the past, we have big overarching preview features where it would be very useful to be able to easily run the whole suite with them enabled. 
Additionally. we had conflicting preview features that worked when tested individually, but broke when used together.

We should implement the mechanism that would allow us to:

- Run test suite or particular test with the preview feature enabled, even if it is not explicitly specified in the tests schema
- Run whole test suite with all available preview features enabled
- Add CI job, running client suite with all preview features enabled, either as regular job or as nightly job that could also be triggered on PR on demand.

Druue:
Closes #22187


.github/workflows/test.yml
```diff
@@ -154,6 +154,28 @@ jobs:
...
+        with:
+          node-version: 20
+          pnpm-version: ${{ inputs.pnpmVersion }}
+          engine-hash: ${{ inputs.engineHash }}
+          skip-tsc: false
+
+      - name: Get Preview Features
+        id: get
+        run: ./.github/workflows/scripts/get-preview-features.js
+
```
millsp, line 178: I think we need to pass the preview features below

.github/workflows/test-template.yml
```diff
@@ -108,7 +108,8 @@ jobs:
         shard: ['1/5', '2/5', '3/5', '4/5', '5/5']
         queryEngine: ${{ fromJson(inputs.queryEngine) }}
         node: [16, 18, 20]
-        previewFeatures: ['', 'relationJoins']
+        previewFeatureSet: ['all', 'none']
```
millsp, line 111: We either test against all preview features or none.

.github/workflows/scripts/get-preview-features.js
```diff
@@ -0,0 +1,27 @@
...
+  previewFeatures = featureSet === 'all' ? getPreviewFeatures() : included
+  previewFeatures = previewFeatures.filter((pf) => !excluded.includes(pf))
+
+  const commaSeparatedPreviewFeaturesString = previewFeatures.join(',')
+
+  if (typeof process.env.GITHUB_OUTPUT == 'string' && process.env.GITHUB_OUTPUT.length > 0) {
+    fs.appendFileSync(process.env.GITHUB_OUTPUT, `previewFeatures=${commaSeparatedPreviewFeaturesString}\n`)
+    console.debug(`${commaSeparatedPreviewFeaturesString} added to GITHUB_OUTPUT`)
+  }
+}
```
millsp, line 23: Uses `@prisma/internals` to get the current engine's full list of preview features.

If the feature set is set to `full`, you get the full list of features minus the exclusions.
Additionally, it's possible to force the inclusion of a preview feature so it's always active regardless of the feature set.

.github/workflows/test-template.yml
```diff
@@ -141,8 +142,15 @@ jobs:
           # Fails if set to true
           skip-tsc: false
 
+      - name: Get Preview Features
+        id: gpf
+        run: ./.github/workflows/scripts/get-preview-features.js
+        env:
+          PREVIEW_FEATURE_SET: ${{ matrix.previewFeatureSet }}
+          EXCLUDED_PREVIEW_FEATURES: 'driverAdapters'
```
millsp, line 150: Driver adapters are a special feature, so in our "normal" tests, we don't enable it yet to keep things separated, also because they are tested in a separate job.

.github/workflows/test-template.yml
```diff
@@ -207,14 +217,21 @@ jobs:
           # Fails if set to true
           skip-tsc: false
 
-      - run: pnpm run test:functional --data-proxy --shard ${{ matrix.shard }}
+      - name: Get Preview Features
+        id: gpf
+        run: ./.github/workflows/scripts/get-preview-features.js
+        env:
+          PREVIEW_FEATURE_SET: ${{ matrix.previewFeatureSet }}
+          EXCLUDED_PREVIEW_FEATURES: driverAdapters,metrics
```
millsp, line 225: We don't enable these for the Data Proxy tests. Maybe driver adapters could make sense though?

.github/workflows/test-template.yml
```diff
@@ -280,7 +294,15 @@ jobs:
           engine-hash: ${{ inputs.engineHash }}
           skip-tsc: false
 
-      - run: pnpm run test:functional --adapter ${{ matrix.adapter }} --shard ${{ matrix.shard }} --engine-type ${{ matrix.engine-type }} --preview-features ${{ matrix.previewFeatures }} --runInBand
+      - name: Get Preview Features
+        id: gpf
+        run: ./.github/workflows/scripts/get-preview-features.js
+        env:
+          PREVIEW_FEATURE_SET: ${{ matrix.previewFeatureSet }}
+          INCLUDED_PREVIEW_FEATURES: driverAdapters
```
millsp, line 302: In the driver adapters job, their preview feature is always enabled.

.github/workflows/test-template.yml
```diff
@@ -280,7 +294,15 @@ jobs:
...
+          EXCLUDED_PREVIEW_FEATURES: ${{ matrix.engine-type == 'wasm' && 'relationJoins' || '' }}
```
millsp, line 303: Relation joins are excluded as before, so this is just an adaption of what we had.

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -120,6 +120,8 @@ const args = arg(
 async function main(): Promise<number | void> {
   let miniProxyProcess: ExecaChildProcess | undefined
 
+  if (args instanceof Error) throw args
```
millsp, line 123: Fixes type issues.

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -225,13 +227,18 @@ async function main(): Promise<number | void> {
 
   try {
     if (args['--updateSnapshot']) {
+      const providers = args['--provider'] ?? [...allProviders]
       const snapshotUpdate = jestCli.withArgs(['-u']).withArgs(args['_'])
-      snapshotUpdate.withEnv({ UPDATE_SNAPSHOTS: 'inline' }).run()
+
+      for (const provider of providers) {
+        snapshotUpdate.withEnv({ UPDATE_SNAPSHOTS: 'inline', ONLY_TEST_PROVIDERS: provider }).run()
+      }
```
millsp, line 235: I could not update inline snapshots for all providers at once, this fixes it (jest does not allow updating many inline snapshots in the same test in one go).

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -225,13 +227,18 @@ async function main(): Promise<number | void> {
...
+
       snapshotUpdate.withEnv({ UPDATE_SNAPSHOTS: 'external' }).run()
     } else {
       if (!args['--types-only']) {
         jestCli
-          .withArgs(['--testPathIgnorePatterns', 'typescript', '--', args['_']])
+          .withArgs(['--testPathIgnorePatterns', 'typescript', '--', ...args['_']])
```
millsp, line 241: Better type safety caught this

packages/client/tests/functional/_utils/getTestSuitePlan.ts
```diff
@@ -41,11 +41,28 @@ export function getTestSuitePlan(
...
+      if (skip === false && testWillPerformSnapshot === false) {
+        testWillPerformSnapshot = true
+        return [...acc, { name, skip, suiteConfig }]
+      } else {
+        return [...acc, { name, skip: true, suiteConfig }]
+      }
+    }, [] as TestPlanEntry[])
+  }
+
+  return testPlanEntries
```
millsp, line 65: Related to the code deleted below. I found another bug that prevented me from updating many inline snapshots at once (for a single provider this time). The previous code had the assumption that the first test suite to pass would update the inline snapshots. In practice it is possible that the first test suite could also be skipped, and no snapshot would be updated. The previous code did not account for that.

The new code makes sure that the first non-skipped test suite found will be the one performing the inline snapshot update, and the rest will be skipped (as mentioned before: jest cannot allow parallel test suites to update a common inline snapshot).

packages/client/tests/functional/metrics/disabled/tests.ts
```diff
@@ -6,14 +6,14 @@ import type { PrismaClient } from './node_modules/@prisma/client'
 
 declare let prisma: PrismaClient
 
-testMatrix.setupTestSuite(() => {
-  test('attempt to use $metrics a compile-time error', () => {
+testMatrix.setupTestSuite((_0, _1, _2, cliMeta) => {
+  skipTestIf(cliMeta.previewFeatures.includes('metrics'))('attempt to use $metrics a compile-time error', () => {
```
millsp, line 10: Because now we force the metrics flag, this test needs to handle it. It tests what should happen when metrics preview feature is not enabled.

packages/client/tests/functional/query-validation/tests.ts
```diff
@@ -158,51 +164,96 @@ testMatrix.setupTestSuite(
         notAnArgument: 123,
       })
 
-      if (cliMeta.previewFeatures.includes('relationJoins')) {
+      if (relationJoinsEnabled && fullTextSearchEnabled) {
```
millsp, line 167: `fullTextSearch` introduces a bit of a combinatorial explosion, and because we used inline snapshots now we need to handle the different cases of combinations. `fullTextSearch` introduces `UserOrderByWithRelationAndSearchRelevanceInput` vs `UserOrderByWithRelationInput`.

packages/client/tests/functional/relation-load-strategy/preview-feature-disabled.ts
```diff
@@ -8,418 +8,717 @@ declare let prisma: PrismaClient
 testMatrix.setupTestSuite(
   ({ provider }, _suiteMeta, _clientMeta, cliMeta) => {
     const relationJoinsDisabled = !cliMeta.previewFeatures.includes('relationJoins')
+    const fullTextSearchEnabled =
```
millsp, line 11: Same here with `fullTextSearch`.

packages/client/tests/functional/relation-load-strategy/unsupported-queries.ts
```diff
@@ -8,318 +8,536 @@ declare let prisma: PrismaClient
 testMatrix.setupTestSuite(
   ({ provider }, _suiteMeta, _clientMeta, cliMeta) => {
     const relationJoinsEnabled = cliMeta.previewFeatures.includes('relationJoins')
+    const fullTextSearchEnabled =
```
millsp, line 11: Same here with `fullTextSearch`.

.github/workflows/scripts/get-preview-features.js
Druue: Ah nice, this looks a lot cleaner than what we started with 😁

.github/workflows/test-template.yml
Jolg42, line 150:
```suggestion
          # Driver adapters are a special feature, 
          # so in our "normal" tests, we don't enable it yet to keep things separated, 
          # also because they are tested in a separate job.
          EXCLUDED_PREVIEW_FEATURES: 'driverAdapters'
```

.github/workflows/test-template.yml
Jolg42, line 303:
```suggestion
         # Exclude relationJoins tests with WASM because the WASM engine does not include the necessary changes yet.
          EXCLUDED_PREVIEW_FEATURES: ${{ matrix.engine-type == 'wasm' && 'relationJoins' || '' }}
```
this was the comment that existed before

.github/workflows/scripts/get-preview-features.js
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env node
+// @ts-check
```
SevInf, line 2: nice, we should do it more often in pure-js context

packages/client/tests/functional/_utils/getTestSuitePlan.ts
SevInf: Nice, stumbled upon this couple of times but never took time for investigating.

packages/client/tests/functional/metrics/disabled/tests.ts
SevInf: Do you think it would be common enough usecase? If it is, would be better to add it as the rest of `optOut` options with a forced `reason`?

.github/workflows/test-template.yml
Jolg42:
```suggestion
          # We skip `metrics` for Data Proxy tests
          EXCLUDED_PREVIEW_FEATURES: driverAdapters,metrics
```

Jolg42: I'm not sure for `driverAdapters` here if that actually changes something? 🤔

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -225,13 +227,18 @@ async function main(): Promise<number | void> {
 
   try {
     if (args['--updateSnapshot']) {
+      const providers = args['--provider'] ?? [...allProviders]
       const snapshotUpdate = jestCli.withArgs(['-u']).withArgs(args['_'])
-      snapshotUpdate.withEnv({ UPDATE_SNAPSHOTS: 'inline' }).run()
+
+      for (const provider of providers) {
+        snapshotUpdate.withEnv({ UPDATE_SNAPSHOTS: 'inline', ONLY_TEST_PROVIDERS: provider }).run()
```
Jolg42, line 234:
```suggestion
        # runs the snapshot update for each provider separately
        # because jest does not allow updating many inline snapshots in the same test in one go
        snapshotUpdate.withEnv({ UPDATE_SNAPSHOTS: 'inline', ONLY_TEST_PROVIDERS: provider }).run()
```

packages/client/tests/functional/engine-selection-logic/tests.ts
```diff
@@ -18,7 +18,7 @@ testMatrix.setupTestSuite(
     })
 
     describe('via env var', () => {
-      testIf(clientMeta.dataProxy /** = --no-engine */)(
+      testIf(clientMeta.dataProxy /** = --no-engine */ && !cliMeta.previewFeatures.includes('metrics'))(
```
Jolg42, line 21: I don't understand why we need to check that `metrics` is not enabled here?

millsp: It's not clear if you lack context indeed, related: https://github.com/prisma/prisma/pull/22430#discussion_r1452528050

packages/client/tests/functional/query-validation/tests.ts
```diff
@@ -1,11 +1,17 @@
...
 
 declare let prisma: PrismaClient
 
 testMatrix.setupTestSuite(
-  (_suiteConfig, _suiteMeta, _clientMeta, cliMeta) => {
+  ({ provider }, _suiteMeta, _clientMeta, cliMeta) => {
+    const relationJoinsEnabled = cliMeta.previewFeatures.includes('relationJoins')
+    const fullTextSearchEnabled =
+      cliMeta.previewFeatures.includes('fullTextSearch') &&
+      (provider === Providers.POSTGRESQL || provider === Providers.MYSQL)
```
Jolg42, line 13: Not sure if/how we should handle Vitess here?

Jolg42 (approved):
Looks good, (see comments before merging)

packages/client/tests/functional/engine-selection-logic/tests.ts
Jolg42:
```suggestion
      # Note that when using Accelerate, the client will fail to init if metrics is enabled.
      # so we only run this tests in this file when `metrics` is not enabled.
      testIf(clientMeta.dataProxy /** = --no-engine */ && !cliMeta.previewFeatures.includes('metrics'))(
```

packages/client/tests/functional/metrics/disabled/tests.ts
millsp: Yes, I think we should clean this up

.github/workflows/test-template.yml
millsp: Yes, that's a good question. I'll give it some thought

PR #22464, opened 2023-12-19
feat(client): pg-worker

millsp:
This PR brings in `@prisma/pg-worker` with is an optimized version of `pg` that weighs `47KB` less (gziped) compared to plain `pg`. In time, we might want to open a PR on the `pg` repository so we can get rid of this package.

https://github.com/prisma/ecosystem-tests/pull/4579

closes https://github.com/prisma/team-orm/issues/623

/integration


millsp:
Tests fail because I only built an ESM version atm

packages/adapter-pg/src/pg.ts
```diff
@@ -20,6 +20,7 @@ const debug = Debug('prisma:driver-adapter:pg')
 type StdClient = pg.Pool
 type TransactionClient = pg.PoolClient
 
+// eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
```
millsp, line 23:
```suggestion
```

jkomyno (approved):
Nice!

PR #22576, opened 2024-01-05
chore: cleanup actions following buildjet removal

Jolg42:
This is a cleanup of unused/unecessary things / leftovers.

Note that Docker login was only necessary when the runner was hosted by buildjet

### DIFF
--- a/.github/workflows/daily-buildpulse.yml
+++ b/.github/workflows/daily-buildpulse.yml
@@ -12,5 +12,4 @@ jobs:
     with:
       reason: buildpulse
       jobTimeout: 70
-      ubuntuRunner: ubuntu-latest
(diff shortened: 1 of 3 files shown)


millsp (approved):

janpio:
> Note that Docker login was only necessary when the runner was hosted by buildjet

Are you sure about that? I think we also had cases before where GH Actions runners we running into rate limits.

Jolg42:
Yes because I think Docker would be unusable on GitHub Actions without them doing a special agreement behind the scenes.
Docker login is only useful, on GitHub Actions when doing mutations or something private.

But also, we'll find out very quickly is that is not true. So far we didn't see anything about it.

PR #22624, opened 2024-01-11
chore(migrateDeploy): fix spelling mistake

victorgdb:
Fixing a typo when one single migration was applied the following text was displayed "The following migration have been applied".

This PR fixes it.

### DIFF
--- a/packages/migrate/src/__tests__/MigrateDeploy.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDeploy.test.ts
@@ -45,7 +45,7 @@ describe('sqlite', () => {
 
     const result = MigrateDeploy.new().parse([])
     await expect(result).resolves.toMatchInlineSnapshot(`
-            The following migration have been applied:
+            The following migration(s) have been applied:
 
             migrations/
               └─ 20201231000000_init/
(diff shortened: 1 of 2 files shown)


Jolg42:
It looks like it's currently inconsistent indeed.

Before we can merge, check the tests as they are failing, there is probably one or more snapshots to update.

I think this could be updated too:
- https://github.com/prisma/prisma/blob/1947b1aeabd5bfe20c9747803c6693075a03e99a/packages/migrate/src/commands/MigrateDeploy.ts#L113

victorgdb:
> **Milestone Check ** Pending — Please set the milestone!

Hey @Jolg42, thanks for your reply. Tests have been fixed, regarding the line you quoted, it seems coherent in English to me!

Jolg42:
Indeed! I meant to be consistent and use `migration(s)` there too, because technically it could be one or more.

victorgdb:
> Indeed! I meant to be consistent and use `migration(s)` there too, because technically it could be one or more.

Of course, but `All migrations have been applied` is valid both in terms of English and in terms of feature. Even if one migration is applied, all migrations have been applied 😇

Jolg42 (approved):
Thanks for the PR!

PR #22630, opened 2024-01-12
feat(client): exports map

millsp:
This PR enables the new DX that we want for our `driverAdapters` preview for edge functions. In order to learn more about this, please read this [document](https://www.notion.so/prismaio/Current-and-Future-Client-DA-DX-64b26c6b85c548288c29d4b0ef2b8ed4). Additionally, this [document](https://www.notion.so/prismaio/DX-rundown-Edge-Functions-support-d191797e03cd4b15b42fed07b1febebe) turns the previous document and other related research (successes and failures) into a summary and a plan, so you may be interested into reading it.

## Exports maps

In a nutshell, we want to get rid of `@prisma/client/edge` and `engineType=wasm`. In order to achieve this, we use an [exports map](https://nodejs.org/api/packages.html#conditional-exports). The exports map is responsible for loading the correct client for the correct runtime.

## Custom outputs

However, it is important to note that export maps don't work with custom outputs, because they only work from within `node_modules`. This is where things are more complicated: we want to make it work there as well so that our DX is equal in all directions.

## Warning channels

In order to make this work, a user using custom outputs will be prompted to add their local package folder as a dependency to their own `package.json`. We have 3 channels to let them know about that:
* `pnpm prisma generate` will emit the first piece of information with instructions and docs
* If they missed that, then they will get a deprecation warning on their imported client (via types)
* If they missed that, then they will get a warning with the same information as above (via runtime)

Once they have setup their local prisma client as a dependency, the exports map works and all the warnings disappear. A warning can happen if a user has either not imported their local client as a dependency, or they have misconfigured their `tsconfig.json`'s `moduleResolution`. In either of these cases the exports maps either don't work or don't work optimally and the warning will be displayed through one of the channels.

## Warning emission

In a few words, the mechanism that supports turning these warnings on and off is based on the exports map itself. First of all, such warnings can only appear on custom output clients, they should never appear otherwise.
When an exports map isn't working, and a user is importing their client like eg. `../prisma/client`, under the hood `../prisma/client/index` is actually loaded. We rely on this behavior to:
1. Include a runtime warning in `index`
2. Include a type-level warning in `index.d.ts`
3. (We always output instructions for custom outputs on generate)

## Warning suppression

So `index` becomes a "warning `index`" for custom outputs, to deliver further instructions. When instructions are applied, the exports map then works and now `../prisma/client` was replaced with a `db` import (since we ask users to make it a proper dependency now). So now when a user imports `db`, it is gated through the exports map and does not load `index` anymore, it loads `default` -- the warning-less counterpart of `index`.

Under the hood, the way this works is that `default` is the actual generated client and types, and `index` is just a thin wrapper on top of it to emit warnings as needed. For the types `index` overrides the `PrismaClient` type to mark it as deprecated and re-exports all the rest. For the runtime, `index` re-exports everything from `default` as well and has an additional `console.warn`.

## Better imports

This then supports our users so that they don't have to think about importing `@prisma/client` vs `@prisma/client/edge` vs `@prisma/client/wasm`. That also means that we could simplify our instructions on generate, so the part about `@prisma/client/edge` is gone there too (the Accelerate hint remains though).

## Exports mappings

Another thing that the exports map enables for us is the ability to remove `engineType=wasm`. This is no longer a requirement because on runtimes like `workerd`, the exports map will point to `@prisma/client/wasm` under the hood, which is a client capable of loading the wasm engine in edge runtimes. Let's recap our runtimes and entrypoints:
* `@prisma/client/default`: Our Node.js client that works with `engineType`
* `@prisma/client/edge`: Now a client that is solely meant to run with Accelerate
* `@prisma/client/wasm`: Similar to `edge`, but can run both Accelerate and Wasm

Of course, all these imports happen under the hood and are managed by the exports map. In a nutshell, these are the mappings that the exports map helps us achieve can be summarized like the following:
* `nodejs` -> `@prisma/client/default` (Accelerate, Library, Binary)
* `worker` -> `@prisma/client/wasm` (Accelerate, Wasm)

`@prisma/client/edge` could thus in theory show a deprecation warning too in the future, and we have a task for it for GA.

## Removal of `engineType=wasm`

Because the exports map is able to do the right decisions for us, it means we no longer need `engineType=wasm`. We will now always have a `wasm` client generated, and a small tradeoff of that means that the `query-engine.wasm` will always be included into the generated folder, unless `prisma generate --no-engine` is being used, which will restrict to only being able to use Accelerate in edge runtimes.

Another consequence of this is that Node.js will always use the `library` (or `binary`) engine, and after this PR will no longer be able to load Wasm files. This is a conscious choice, as we've measured that `driverAdapters` are slightly more efficient via N-API, we will only load the Wasm engine if we need to (ie. on edge runtimes). This is not set in stone and we can change it in the future, of course.

## New client runtime

As you might have guessed, as there is a new `@prisma/client/wasm` endpoint there is also a new client runtime: `client/runtime/wasm`. The `wasm` client runtime shares much of its configuration with the `edge` one and are nearly similar in the way they are built.

The key differences between `edge` and `wasm` after the changes from this PR is that `edge` is now exclusively used for Accelerate and thus only contains the `DataProxyEngine`. On the other hand `wasm` includes both the `LibraryEngine` and `DataProxyEngine` (that follows the usual pattern that we always include Accelerate in any of our runtime builds).

Additionally, as mentioned earlier, `library` will not include any Wasm loading abilities anymore.

## Changes to our test setup

Because we made that the Wasm engine cannot be loaded in Node.js anymore and because `engineType=wasm` went way, it required us to adapt our tests and functional tests to accommodate these changes. The main change in our test setup is the introduction of `--client-runtime` which under the hood will force loading one of the compiled client runtimes (found in `packages/client/runtime`). They can be `edge`, `wasm`, or `node`, and of course these namings are meant as a parallel to our exports map.

Being able to run `--client-runtime wasm` has enabled us to test our `wasm` client more thoroughly as it is the same runtime file that runs on Cloudflare Workers. Previously, we ran our tests against the `node` runtime even when testing Wasm via `driverAdapters` (ie. the un-polyfilled version of the runtime). Now only the Wasm loading remains Node.js based, as we run our tests in `jest`, not `workerd`.

### Generation cleanup

In order to implement this, I had to touch many places in our generation code. I felt like it was becoming critical to cleanup in order to be able to maneuver in this space now and in the future, as the code is getting more confusing and less structured over time and as we add new things on top of legacy. I took the opportunity to deep dive and improve things along the way.

### Generation performance

As I added more things, it degraded the performance by 40% (according to CodSpeed). I noticed that we were cloning the dmmf with `klona` multiple times in order to generate the clients. So I turned our `DMMF` into a `Readonly` type and got rid of `klona`. Additionally, I made the properties of `DMMFHelper` lazy, so they only compute when we need them. This resulted in a 30% generation speed increase (according to CodSpeed).

### Fix of a bug in `WasmLoader`

I found that there was a race condition in our Wasm loader, which could probably happen in production. I found this because it was exacerbated somehow by the changes in our functional test setup.

### Cleanup of "dummy file" logic

We have some code that creates dummy files before a client is generated. I needed to update this area and noticed quite some duplication there, which could make changes prone to error. So a small bonus refactor along the way.

### Better snapshots for edge/wasm clients

We've had a long standing issue where we had to skip some tests and snapshots because edge/wasm clients cannot read from disk, however in order to show pretty errors with line numbers we also read the source files from the path retrieved by the stacktrace. Now a `$fs` global utility is injected by jest to allow edge/wasm clients to do just that, so we don't need to skip tests.

/integration


package.json
```diff
@@ -131,6 +131,10 @@
       "path": "packages/client/runtime/edge-esm.js",
       "gzip": false
     },
+    {
+      "path": "packages/client/runtime/wasm.js",
+      "gzip": false
+    },
```
Jolg42, line 137: No need of a `wasm-esm.js`?

millsp: Not yet, one day when we get back to esm

packages/client/helpers/build.ts
```diff
@@ -64,6 +64,44 @@ const edgeRuntimeBuildConfig: BuildOptions = {
...
+      },
+      // we shim WeakRef, it does not exist on CF
+      WeakRef: {
+        globals: weakrefPolyfillPath,
+      },
+      // these can not be exported anymore
+      './warnEnvConflicts': { contents: '' },
+    }),
+  ],
+  logLevel: 'error',
```
millsp, line 83: Should share common config between `esm` and `wasm` runtilmes.

packages/client/src/generation/ts-builders/AnyDeclarationBuilder.ts
```diff
@@ -1,6 +1,7 @@
+import { ClassDeclaration } from './Class'
 import { ConstDeclaration } from './ConstDeclaration'
 import { InterfaceDeclaration } from './Interface'
 import { TypeDeclaration } from './TypeDeclaration'
 
-// TODO: enum, class
-export type AnyDeclarationBuilder = TypeDeclaration | ConstDeclaration | InterfaceDeclaration
+// TODO: enum
+export type AnyDeclarationBuilder = TypeDeclaration | ConstDeclaration | InterfaceDeclaration | ClassDeclaration
```
SevInf, line 7: I think `Export` and `Import` would also belong here.

packages/client/tests/functional/0-legacy-ports/malformed-id/tests.ts
```diff
@@ -21,23 +21,27 @@ testMatrix.setupTestSuite(
       if (clientMeta.runtime === 'edge') {
```
SevInf, line 21: I think this `if` does not make sense anymore, snapshot is identical in all branches.

packages/client/tests/functional/_example/tests.ts
```diff
@@ -78,9 +78,8 @@ testMatrix.setupTestSuite(
   //   },
   // },
   {
-    skipEngine: {
-      from: ['wasm'],
-      reason: 'Tracing preview feature creates a panic in the wasm engine',
+    skip(when, { clientRuntime }) {
```
SevInf, line 81: Nice!

packages/client/wasm.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ...require('.prisma/client/wasm'),
+}
```
SevInf, line 3: Nit, but is there a reasone why we don't do:
```suggestion
module.exports = require('.prisma/client/wasm');
```

just to save on a spread?

packages/generator-helper/src/dmmf.ts
```diff
@@ -1,56 +1,60 @@
+export type ReadonlyDeep<O> = {
+  +readonly [K in keyof O]: ReadonlyDeep<O[K]>
+}
+
 // eslint-disable-next-line @typescript-eslint/no-namespace
 export namespace DMMF {
-  export interface Document {
+  export type Document = ReadonlyDeep<{
```
SevInf, line 7: Did we got rid of last parts where we modify DMMF?

packages/client/scripts/default-wasm.d.ts
janpio, line 1: When exactly will these types be used/relevant?

packages/client/scripts/default-wasm.js
```diff
@@ -1,7 +1,7 @@
 class PrismaClient {
   constructor() {
     throw new Error(
-      `@prisma/client did not initialize yet. Please run "prisma generate" and try to import it again.
+      `To use @prisma/client in an edge runtime, enable \`driverAdapters\` in your Prisma schema and run \`prisma generate\`.
```
janpio, line 4: What will be the long term message here after the preview feature is removed?

packages/client/src/generation/TSClient/Generatable.ts
```diff
@@ -1,18 +1,19 @@
+// TODO rename this to Generable (correct english term)
```
janpio, line 1: https://en.wiktionary.org/wiki/generatable 😆

packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
```diff
@@ -0,0 +1,5 @@
+export const invalidImportWarning = [
+  'Add your client as a `db` dependency to `package.json` and import it as a dependency instead',
```
janpio, line 2: Is "`db` dependency" universally true?

janpio, line 2: > add ... as a ... dependency and import as a dependency instead
Is this correct?

packages/client/wasm.js
millsp: I think this was required by some framework for some reason

millsp: Found it: https://github.com/prisma/prisma/pull/12907

packages/client/scripts/default-wasm.d.ts
millsp: This is when your client did not run `prisma generate` yet

packages/client/scripts/default-wasm.js
millsp: It will the same as usual I guess, "Prisma client did not generate, please make sure to run prisma generate"

packages/client/src/generation/TSClient/Generatable.ts
millsp: Somehow it vexes me when I read it, I'm the able kind :laughing: 

https://en.wiktionary.org/wiki/generable

millsp: But you have a point, this awful word is real

packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
millsp: You could technically call it what you want, I am trying to set some convention that is closest to what I have seen so far from our users.

millsp: What seems incorrect to you?

packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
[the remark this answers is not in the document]
millsp: Both at runtime and on type-level when the exports map is not well configured.

packages/generator-helper/src/dmmf.ts
millsp: I got rid of the ones I found, there were very few.

packages/client/src/runtime/core/engines/library/WasmLibraryLoader.ts
```diff
@@ -31,20 +25,25 @@ export const wasmLibraryLoader: LibraryLoader = {
...
-
-      // from https://developers.cloudflare.com/workers/runtime-apis/webassembly/rust/#javascript-plumbing-wasm-bindgen
-      loadedWasmInstance = new WebAssembly.Instance(wasmMod, { './query_engine_bg.js': wasmBindgenRuntime }).exports
-      wasmBindgenRuntime.__wbg_set_wasm(loadedWasmInstance)
+      loadedWasmInstance = (async () => {
+        const wasmModule = await config.getQueryEngineWasmModule?.()
+
+        if (wasmModule === undefined || wasmModule === null) {
+          throw new PrismaClientInitializationError(
+            'The loaded wasm module was unexpectedly `undefined` or `null` once loaded',
```
janpio, line 33: When could this case happen?

packages/client/src/utils/generateInFolder.ts
```diff
@@ -24,19 +24,13 @@ const del = promisify(rimraf)
 
 export interface GenerateInFolderOptions {
   projectDir: string
-  useLocalRuntime?: boolean
-  transpile?: boolean
```
janpio, line 28: What was this used for?

packages/client/src/runtime/core/engines/library/WasmLibraryLoader.ts
millsp: Practically, I don't think it could ever happen unless a bundler did something awful to the bundle, and even then it's likely to be a different error, but here because it's technically possible according to the type system.

packages/client/src/utils/generateInFolder.ts
millsp: `useLocalRuntime` is dead code (always set to false) and was removed https://github.com/prisma/prisma/pull/22630/files#diff-48c06b9c0b442da6779812b9b0acdeaff470e209e1fd3ff31ba4037061c4cb8dL85

`transpile` is always set to true, which also generates the client as it does everywhere else, so it is not actually a useful setting nor needed. Setting `transpile` to `false` prevents the engine copy within the tests, and we never set it to false, also because we don't need this https://github.com/prisma/prisma/pull/22630/files#diff-3c7fec7e990cab78b875f4fcce9979640a5fe3004db9dd082578dac14874d5deL296

Both of these settings are testing-related only.

packages/client/wasm.js
janpio: We should add a comment for the future here then.

packages/client/src/generation/TSClient/Generatable.ts
janpio: "generable" is awful, it is missing the "te" part. "gener" is not even a word :p

packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
janpio: using "dependency" twice, once for the fact that it is a `dependency` in the `pkg.sjon`, and then how it is imported - also referred to as "dependency".

millsp: Could you help me reformulate/refine this?

millsp: Maybe an example package json would help to illustrate?

packages/generator-helper/src/dmmf.ts
SevInf: Great! Should `ReadonlyDeep` also turn arrays in `ReadonlyArray`?

SevInf:
I think we should also adjust a generated name in `.prisma/client/package.json`. Currently it is invalid and some package managers complain about it when you try to add custom path as a dependency
Also worth double-checking: in case of multiple generated clients, how would package managers react to the packages with identical names?

millsp:
Yes, I agree. We discussed that [here](https://prisma-company.slack.com/archives/C0632RWQYMA/p1704751683202459). After looking into it, it seems the package needs to be both private and have a better name. From my first exploration, it seems the name does not actually matter for package managers, though I think we should make it unique for other tooling.

SevInf:
Just so it won't get lost. `pnpm` actually complains about invalid name of generated package if it is added as dependency.
To reproduce:
- generate client at custom path
- add it as dependency as our new warning suggests
- remove generated client and re-run `pnpm i`
- generate client at same location and run `pnpm i` again

packages/client/src/generation/ts-builders/AnyDeclarationBuilder.ts
millsp: If we allow `Export` in the list, then it means one could end up with `export export...`?

packages/generator-helper/src/dmmf.ts
millsp: I think it does, see eg.
image

packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
millsp: I updated the message so it is more packed with all necessary information and steps.

millsp: I updated the message so it is more packed with all necessary information and steps.

jsumners-nr:
For what it's worth, this PR is not a minor compatible change as demonstrated in https://github.com/newrelic/node-newrelic/actions/runs/7712498218/job/21020203083?pr=1983#step:6:16711

janpio:
Thanks for the comment @jsumners-nr, we're looking into it.
We have a regression reported here, don't know yet if this is related: https://github.com/prisma/prisma/issues/22870

andyjy:
sooo much simpler now 🙏 🎉 
(Next.js, using both Edge early access + normal Node engine, locally + deployed to Vercel)

Screenshot 2024-01-30 at 22 47 38

PR #22636, opened 2024-01-12
chore: Ensure BinaryEngine correctly handles top level request errors

SevInf:
`request` and `requestBatch` methods handled only GraphQL-style errors
(`errors` array in JSON response). This did not work correctly in case
of top level error (for example, incoming message not being parsed
correctly).

Contributes to prisma/team-orm#710


SevInf:
ci test all

Jolg42 (approved):

PR #22637, opened 2024-01-12
feat(tech-debt): [get-platform] remove outdated distro resolution warning

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/814.

It removes the warning:

> Prisma doesn't know which engines to download for the Linux distro "...". Falling back to Prisma engines built "debian".

which is currently displayed whenever the `ID_LIKE` field of `/etc/os-release` can't be read or does not match a certain set of predetermined strings.

This PR should allow us to reduce noise in the public Github issues.

### DIFF
--- a/packages/get-platform/src/__tests__/getPlatform.test.ts
+++ b/packages/get-platform/src/__tests__/getPlatform.test.ts
@@ -166,10 +166,6 @@ describe('getBinaryTargetForCurrentPlatformInternal', () => {
         }),
       ).toBe('debian-openssl-3.0.x')
       expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
-      expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`
-        prisma:warn Prisma doesn't know which engines to download for the Linux distro "unknown". Falling back to Prisma engines built "debian".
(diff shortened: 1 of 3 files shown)


packages/get-platform/src/__tests__/getPlatform.test.ts
```diff
@@ -166,10 +166,6 @@ describe('getBinaryTargetForCurrentPlatformInternal', () => {
         }),
       ).toBe('debian-openssl-3.0.x')
       expect(ctx.mocked['console.log'].mock.calls.join('\n')).toMatchInlineSnapshot(``)
-      expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toMatchInlineSnapshot(`
-        prisma:warn Prisma doesn't know which engines to download for the Linux distro "unknown". Falling back to Prisma engines built "debian".
-        Please report your experience by creating an issue at https://github.com/prisma/prisma/issues so we can add your distro to the list of known supported distros.
-      `)
```
millsp, line 172: I am ok removing this if in case of failure (engine loading) we do display the necessary information for the user to understand we defaulted to debian because xyz

jkomyno: So, should this be moved to the logic that tests for panics?

millsp: Not sure if it is a panic, but I think we have some code to try/catch `dlopen` maybe it should happen there?

PR #22687, opened 2024-01-17
feat(adapter-pg,adapter-neon): Enforce usage of connection pool

SevInf:
Adapters don't work with non-pooled `Client` instances. For `pg` and
`neon` we enforce it only on type level, so check won't work for JS
projects. In this PR, we are checking the passed instance and ensuring
it is a `Pool`, similarly to what we already do for `planetscale`.

Fix prisma/team-orm#616

### DIFF
--- a/packages/adapter-pg/jest.config.js
+++ b/packages/adapter-pg/jest.config.js
@@ -1,0 +1,25 @@
+module.exports = {
+  transform: {
+    '^.+\\.(m?j|t)s$': '@swc/jest',
+  },
+  transformIgnorePatterns: [],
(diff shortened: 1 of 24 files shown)


Jolg42 (approved):
Looks good to me

packages/adapter-pg/jest.config.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
```
millsp, line 1: Could we share one jest file across?

millsp (approved):

PR #22699, opened 2024-01-18
revert(client): Revert "feat(adapter-pg,adapter-neon): Enforce usage of connection pool (#22687)"

SevInf:
The change broke engines, we'll fix it and reintroduce PR later.

This reverts commit e58b7506656272fa2c21f3682ac16197cc2046a2.


miguelff (approved):

Jolg42 (approved):

jkomyno (approved):
Let's merge this to unblock the pipeline.

In a follow-up PR, we should likely try substituting

```typescript
import * as pg from 'pg'
```

with

```typescript
import pg from 'pg'
```

as that's been tested to work locally when joint with `prisma-engines` tests.

PR #22700, opened 2024-01-18
feat(adapter-pg,adapter-neon): Enforce usage of connection pool (#22687)

SevInf:
PR was already reviewed in #22687, this iteration just
fixes PG import without breaking engine executor.


SevInf:
confirmed that engines are fine here https://github.com/prisma/prisma-engines/pull/4655

helpers/test/presets/default.js
millsp, line 1: Not sure why this is showing up again, wasn't this added already?

SevInf: We reverted the PR it was added in

Jolg42 (approved):

packages/adapter-neon/src/neon.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/require-await */
-import type neon from '@neondatabase/serverless'
+import * as neon from '@neondatabase/serverless'
```
Jolg42, line 2: The 1 line change?

millsp (approved):

packages/adapter-pg/src/pg.ts
```diff
@@ -11,7 +11,7 @@ import type {
   TransactionOptions,
 } from '@prisma/driver-adapter-utils'
 import { Debug, err, ok } from '@prisma/driver-adapter-utils'
-import type pg from 'pg'
+import pg from 'pg'
```
Jolg42, line 14: I think that was the change actually

packages/adapter-neon/src/neon.ts
Jolg42: (Nope)

packages/adapter-pg/src/pg.ts
Jolg42: Previously `import * as pg from 'pg'`

packages/adapter-neon/src/neon.ts
jkomyno: Given we moved from `import * as pg from 'pg'` to `import pg from 'pg'`, shouldn't we do the same with `neon` as well?

Issue #21454, opened 2023-10-11
Nested json is transformed to string when used “$type” key in object


tomasz-py:
### Bug description

Json value which has array of objects where one of keys is "$type" is transformed to string

`
  const data = await prisma.test.create({
    data: {
       value: {
         options: [{ $type: "z" }, { $type: 1 }],
        },
    },
  });
`

data.value will looks like: 
`{ options: [ '{"$type":"z"}', '{"$type":1}' ] }`

instead of 
`{ options: [ {'$type': 'z'}, {'$type': 1 } ] }`

the same situation is for object in object:
`  const data = await prisma.test.create({
    data: {
         value: {
          options: {
            $type: "test",
          },
        },
    },
  });`

data.value will contain options as string instead of json
`{ options: '{"$type":"test"}' }`

Additional info:
its stored properly if object key is $test, so its only breaks when key is "$type"

### How to reproduce

1. Use the following schema
2. Generate the client and run the following script
3. Check value in database

### Expected behavior

Prisma should store array of jsons as jsons instead of transforming them into string

### Prisma information

```prisma
model Test {
  id  Int  @id @default(autoincrement())
  value  Json?  @db.JsonB
}
```

```ts
import { PrismaClient } from "@prisma/client";

async function main() {
  const prisma = new PrismaClient();

  const data = await prisma.test.create({
    data: {
       value: {
         options: [{ $type: "z" }, { $type: 1 }],
        },
    },
  });

 const otherData =  await prisma.test.create({
    data: {
        value: {
          options: {
            $type: "test",
          },
        },
    },
  });

  console.log(data.value); // { options: [ '{"$type":"z"}', '{"$type":1}' ] }
 
  console.log(otherData.value) // { options: '{"$type":"test"}' }

  prisma.$disconnect();
}

main();
```

### Environment & setup

- OS: Windows 11
- Database: PostgreSQL
- Node.js version: v20.5.1

### Prisma Version

```
prisma                  : 5.1.1
@prisma/client          : 5.1.1
Current platform        : windows
Query Engine (Node-API) : libquery-engine 6a3747c37ff169c90047725a05a6ef02e32ac97e (at node_modules\@prisma\engines\query_engine-windows.dll.node)
Schema Engine           : schema-engine-cli 6a3747c37ff169c90047725a05a6ef02e32ac97e (at node_modules\@prisma\engines\schema-engine-windows.exe)
[...]

PR #22714, opened 2024-01-19
test: Reproduce #21454

SevInf:
Reproduction for $type key in JSON issue. It works correctly if $type is
a top level key in JSON field, but fails if it is nested.


SevInf:
Close in favor of #22850

PR #22716, opened 2024-01-19
chore: Fix new-test script

SevInf:
Removes yeoman dependency and fixes the script.
Also adds a couple of nice additions on top of what old script did, like
correctly resolving import paths for nested dirs and running prettier
after creation.

Fix prisma/team-orm#828


Jolg42 (approved):
Nice!

PR #22770, opened 2024-01-23
feat(adapters): cloudflare-d1 - transactions

Druue:
closes https://github.com/prisma/team-orm/issues/849


Jolg42:
Note: we found that this doesn't work and errors with `not authorized`

It matches what I found in workerd https://github.com/cloudflare/workerd/blob/main/src/workerd/api/sql-test.js

PR #22780, opened 2024-01-24
test(client): add test for accelerate and driver adapter errors

millsp:
closes https://github.com/prisma/team-orm/issues/863

### DIFF
--- a/packages/client/tests/e2e/driver-adapters-accelerate/prisma/schema.prisma
+++ b/packages/client/tests/e2e/driver-adapters-accelerate/prisma/schema.prisma
@@ -1,0 +1,36 @@
+// This is your Prisma schema file,
+// learn more about it in the docs: https://pris.ly/d/prisma-schema
+
+generator client {
+  provider = "prisma-client-js"
+  previewFeatures = ["driverAdapters"]
+}
+
+datasource db {
+  provider = "sqlite"
+  url      = "file:./db"
+}
+
+model User {
+  id        Int      @id @default(autoincrement())
+  email     String   @unique
+  name      String?
+  posts     Post[]
+  profile   Profile?
+}
+
+model Post {
+  id        Int      @id @default(autoincrement())
+  title     String
+  content   String?
+  published Boolean  @default(false)
+  author    User     @relation(fields: [authorId], references: [id])
+  authorId  Int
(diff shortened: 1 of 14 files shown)


packages/client/tests/e2e/driver-adapters-accelerate/prisma/schema.prisma
```diff
@@ -0,0 +1,36 @@
...
+
+generator client {
+  provider = "prisma-client-js"
+  previewFeatures = ["driverAdapters"]
+}
+
+datasource db {
+  provider = "sqlite"
+  url      = "file:./db"
+}
```
Jolg42, line 12: Technically, Accelerate does not support the `sqlite` provider.
So it would be best if we had a special error for that, which I guess we don't have at the moment?

packages/client/tests/e2e/driver-adapters-accelerate/tests/url.ts
```diff
@@ -0,0 +1,45 @@
...
+    expect(newClient).toThrowErrorMatchingInlineSnapshot(`
+"Prisma Client was configured to use the \`adapter\` option but the URL was a \`prisma://\` URL.
+Please either use the \`prisma://\` URL or remove the \`adapter\` from the Prisma Client constructor."
+`)
+  })
+})
+
+test('driver adapters cannot be used with accelerate via @prisma/client/default', () => {
+  jest.isolateModules(() => {
+    expect.assertions(1)
```
Jolg42, line 25: I'm not sure why `expect.assertions(1)` is actually needed here?

Jolg42 (approved):

packages/client/tests/e2e/driver-adapters-accelerate/prisma/schema.prisma
millsp: Yeah, I am not sure what would happen in this case. I'll change it to `postgresql` so it is cleaner at least.

packages/client/tests/e2e/driver-adapters-accelerate/tests/url.ts
millsp: Left over from another test construct, thanks for catching

PR #22795, opened 2024-01-24
test(client): Test diffierent moduleResolution options in TS

SevInf:
For new driver DX, if your TS is misconfigured, client class will be
marked as deprecated. This PR adds tests for different combinations of
`module`/`moduleResolution` options and ensures that client is marked as
deprecated when we want it and is not when we don't.

Fix prisma/team-orm#854


packages/client/tests/e2e/custom-output-ts/tests/main.test.ts
```diff
@@ -0,0 +1,96 @@
...
+const prismaClientName = ts.escapeLeadingUnderscores('PrismaClient')
+
+const deprecatedOptions = [
+  { module: 'commonjs', moduleResolution: 'Node' },
+  { module: 'ES2015', moduleResolution: 'Node' },
+  { module: 'ES2020', moduleResolution: 'Node' },
+  { module: 'ES2022', moduleResolution: 'Node' },
+  { module: 'ESNext', moduleResolution: 'Node' },
+  { module: 'Node16', moduleResolution: 'Node' },
+  { module: 'NodeNext', moduleResolution: 'Node' },
```
SevInf, line 16: These two point out that we maybe need to adjust our warning a little bit: changing just `module` without `moduleResultion` does not actually fixes the problem.

SevInf: Actually, TS API seems to be more forgiving than `tsc`, these options are invalid when checked through `tsc. Should we remove them?

packages/client/tests/e2e/custom-output-ts/tests/main.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  { module: 'NodeNext', moduleResolution: 'NodeNext' },
+
+  { module: 'ES2015', moduleResolution: 'Bundler' },
+  { module: 'ES2020', moduleResolution: 'Bundler' },
+  { module: 'ES2022', moduleResolution: 'Bundler' },
+  { module: 'ESNext', moduleResolution: 'Bundler' },
+  { module: 'Node16', moduleResolution: 'Bundler' },
+  { module: 'NodeNext', moduleResolution: 'Bundler' },
+]
+
```
Jolg42, line 42: Maybe add `ES2023`?
I think it's the same as `ESNext` for the moment, but it will change later

millsp: Somewhat related: Bundler seems to break our types (found in the QA session)

packages/client/tests/e2e/custom-output-ts/tests/main.test.ts
```diff
@@ -0,0 +1,108 @@
...
+})
+
+describe('direct import', () => {
+  // for direct import, client should always be deprecated
+  for (const options of deprecatedOptions.concat(nonDeprecatedOptions)) {
+    test(`${JSON.stringify(options)}: PrismaClient is deprecated`, () => {
+      expect(isPrismaClientDeprecated(noDepTs, options)).toBe(true)
+    })
+  }
+})
```
millsp, line 64: Could we add another case for the default import? Then I think with that the PR is complete.

packages/client/tests/e2e/custom-output-ts/tests/main.test.ts
millsp: Yes, if we can improve our error message I am all for it, how would you change it?

millsp: And yes, if the option is invalid then I guess we should remove it.

packages/client/tests/e2e/custom-output-ts/tests/main.test.ts
SevInf: `moduleResolution` and `moduleResolution` is different from `target` and goes only up to `ES2022`.

packages/client/tests/e2e/custom-output-ts/tests/main.test.ts
SevInf: done

packages/client/tests/e2e/custom-output-ts/tests/main.test.ts
SevInf: Warning suggestion was coming from me not getting validation error from invalid combinations and thinking they are valid. Since then I figured out how to make TS API throw errors on those, our present warning is correct.

millsp (approved):

PR #22810, opened 2024-01-25
test(driver-adapters): add test for `Bytes`

jkomyno:
This PR contributes to https://github.com/prisma/team-orm/issues/687.

Before this PR, we were testing the following `Bytes` pipeline in Rust only:

1. Create a record with a `Bytes` field in the `schema.prisma`
2. Retrieve such `Bytes` record

The problem is that we were modeling these tests in GraphQL, which - afaik - only supports submitting `Bytes` fields as strings, which are later `base64` encoded and decoded.

In PlanetScale (both Napi.rs and Wasm), the reproduction test of https://github.com/prisma/team-orm/issues/687#issue-2024070923 leads to test failures like:

```diff
- {"data":{"findManyChild":[{"id":1,"bytes":"AQID"},{"id":2,"bytes":"FSDF"},{"id":3,"bytes":"AA"}]}}
+ {"data":{"findManyChild":[{"id":1,"bytes":"AQID"},{"id":2,"bytes":"FSDDr8K/wr0="},{"id":3,"bytes":"AA=="}]}}
```

Let's consider the string `"AQID"` for the `Bytes` field of a new record. Call it `inputStr`.

The Rust Query Engine does pretty much the following (expressed by means of JS functions for convenience):
- [INSERT]
  - It assumes `inputStr` is already base64-encoded
  - It attempts decoding the text via e.g. `atob(inputStr)`, which for `"AQID"` is `"\x01\x02\x03"`
  - It encodes it in bytes via e.g. `Buffer.from(atob(inputStr))`, which for `"AQID"` is `<Buffer 01 02 03>`
  - It inserts the encoded string into a new record in the database
- [SELECT]
  - It identifies `Bytes` as the column type
  - When using `PlanetScale`, [this line](https://github.com/prisma/prisma-engines/blob/c2e5bf22be25fc20437ad83be3d21897645f9651/query-engine/driver-adapters/src/conversion/js_to_quaint.rs#L237) will be hit, as the retrieved value is JSON-encoded as `String("\u{1}\u{2}\u{3}")` [Rust] / `"\x01\x02\x03"` [JS]. 
  - When using `pg`, [this line](https://github.com/prisma/prisma-engines/blob/main/query-engine/driver-adapters/src/conversion/js_to_quaint.rs#L238-L245) will be hit, as the retrieved value is JSON-encoded as `[Number(1), Number(2), Number(3)]`
  
When the retrieved value is JSON-encoded as a string (like in PlanetScale's case), further string processing happens, which sometimes corrupts the string. I haven't pinpointed exactly how that happens, but `base64` encoding definitely plays a part.
In fact, creating a new record with "`FSDF`" for the `Bytes` fields results in `"FSDDr8K/wr0="` being retrieved by Prisma.

This divergence in Driver Adapters behavior completely goes away if we use `@prisma/client` directly.

---

When using the TS `@prisma/client` with JSON protocol (the only one supported by `driver-adapters`), however, `Bytes` fields are required to be instances of `Buffer` (which, btw, is Node.js-specific, and does not exist in EcmaScript).
As a consequence, `Bytes` fields submitted and retrieved via `@prisma/client` never undergo the `base64` encoding/decoding steps.

This means that the following holds true for every Driver Adapter (Napi.rs + Wasm) and native Driver:

```typescript
await prisma.binary.create({
  data: {
    id: '1',
    bytes: Buffer.from('Some string to encode in binary')
  }
})

const record = await prisma.binary.findFirst({ where: { id: '1' } })
// ==>
// {
//   id: '1',
//   bytes: Buffer.from('Some string to encode in binary'), // (*)
// }
```

(*) Technically, the received value isn't a `Buffer`, but it can be converted into one such object.


jkomyno:
Here's a debug-intensive log of the `planetscale-js` run on the Rust tests associated to https://github.com/prisma/team-orm/issues/687.

[out.planetscale.txt](https://github.com/prisma/prisma/files/14064256/out.planetscale.txt)

---

Here's a debug-intensive log of the `pg-js` run on the Rust tests associated to https://github.com/prisma/team-orm/issues/687.

[out.pg.txt](https://github.com/prisma/prisma/files/14064260/out.pg.txt)

jkomyno:
Note: this is blocked until https://github.com/planetscale/database-js/issues/161 is resolved.

Jolg42 (approved):
@jkomyno I'm interested in this test for testing D1.

Do you think it would be ok to tweak this to expect it to fail for planetscale's adapter? And change the expectation once fixed?

jkomyno:
> @jkomyno I'm interested in this test for testing D1.
> 
> Do you think it would be ok to tweak this to expect it to fail for planetscale's adapter? And change the expectation once fixed?

Hi @Jolg42, it's totally ok to tweak this if it's convenient for you to test D1.

More likely than not, this PR's title will be renamed in "Test `bytes` on TypeScript as well as on Rust PR", so D1 is welcome here.

Jolg42 (approved):
Note: we discussed that we could merge this now.

About tests: "TypeScript is passing and Rust failing for PlanetScale"

After that, I'm curious to see how it goes for D1 🚀

PR #22818, opened 2024-01-25
fix(cli): exclude Client wasm files from CLI

millsp:
closes https://github.com/prisma/team-orm/issues/799

It's complemented by https://github.com/prisma/prisma/pull/22819 which we would merge before hand so we can inspect the results.


millsp:
(failing test is unrelated)

SevInf (approved):

PR #22832, opened 2024-01-26
fix(cli, client): deletion of engines when `--no-engine` is passed

### LINKED ISSUE #17685: .prisma/client: Name can't start with a dot
### Bug description

When prima client is generated with custom output path, it throws error `.prisma/client: Name can't start with a dot`
Screenshot 2023-02-02 at 12 29 24 AM

### Expected behavior

I want to run yarn normally after manually generating the client.

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
  output   = "../src/generated/client"
}

datasource db {
  provider = "postgresql"
}

```

```ts
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: v16.13.2

### Prisma Version

```
```

millsp:
We knew we introduced a small regression when we fixed https://github.com/prisma/prisma/issues/17685, this fixes it.

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -19,7 +19,6 @@ import { BrowserJS, JS, TS, TSClient } from './TSClient'
 import { TSClientOptions } from './TSClient/TSClient'
 import type { Dictionary } from './utils/common'
 
-const GENERATED_PACKAGE_NAME = 'prisma-client'
 const debug = Debug('prisma:client:generateClient')
 
 type OutputDeclaration = {
@@ -632,5 +631,5 @@ async function deleteOutputDir(finalOutputDir: string, datamodel: string) {
 function getUniquePackageName(datamodel: string) {
   const hash = createHash('sha256')
   hash.write(datamodel)
-  return `${GENERATED_PACKAGE_NAME}-${hash.digest().toString('hex')}`
+  return `prisma-client-${hash.digest().toString('hex')}`
 }
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -289,7 +289,7 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   }
 
   if (noEngine === true) {
-    await deleteOutputDir(outputDir)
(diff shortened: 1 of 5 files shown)


packages/client/tests/e2e/_utils/run.ts
```diff
@@ -14,9 +14,6 @@ const args = arg(
     // like run jest in band, useful for debugging and CI
     '--runInBand': Boolean,
     '--run-in-band': '--runInBand',
-    // do not fully build cli and client packages before packing
-    '--skipBuild': Boolean,
-    '--skip-build': '--skipBuild',
```
millsp, line 19: Not needed. Everyone builds their repo before running tests.

packages/client/src/generation/generateClient.ts
```diff
@@ -604,11 +604,11 @@ async function copyRuntimeFiles({ from, to, runtimeName, sourceMaps }: CopyRunti
  * Attempts to delete the output directory.
  * @param finalOutputDir
  */
-async function deleteOutputDir(finalOutputDir: string) {
+async function deleteOutputDir(finalOutputDir: string, datamodel: string) {
   try {
     debug(`attempting to delete ${finalOutputDir} recursively`)
     // we want to make sure that if we delete, we delete the right directory
-    if (require(`${finalOutputDir}/package.json`).name === GENERATED_PACKAGE_NAME) {
+    if (require(`${finalOutputDir}/package.json`).name === getUniquePackageName(datamodel)) {
```
millsp, line 611: The fix.

Jolg42 (approved):

packages/client/src/generation/generateClient.ts
SevInf: Does not necessary needs changing right now, just food for thought. Would it make sense to change it to something like this?

```suggestion
    if (require(`${finalOutputDir}/package.json`).startsWith(GENERATED_PACKAGE_NAME)) {
```

Otherwise, if schema changed at the same time `--no-engine` flag was added to invocation, engine won't be deleted.

millsp: Good catch, I'll add a test for that

PR #22850, opened 2024-01-29
fix(client): Fix $type key nested in JSON column

### LINKED ISSUE #21454: Nested json is transformed to string when used “$type” key in object

SevInf:
Works on top of prisma/prisma-engine#4670.
Adds `$type: Raw` special case to JSON protocol and replaces `$type:
Json` encoding with it.

See engine PR for detailed description of the problem.

Fix #21454

### DIFF
--- a/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.test.ts
+++ b/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.test.ts
@@ -577,24 +577,29 @@ test('args - object with $type field', () => {
       args: { where: { jsonColumn: { $type: 'Decimal', value: '123' } } },
     }),
     // not using inline snapshot here because our default serializer mangles backslashes on windows
-  ).toBe(`{
-  "modelName": "User",
-  "action": "findMany",
-  "query": {
-    "arguments": {
-      "where": {
-        "jsonColumn": {
-          "$type": "Json",
-          "value": "{\\"$type\\":\\"Decimal\\",\\"value\\":\\"123\\"}"
+  ).toMatchInlineSnapshot(`
+    {
+      "modelName": "User",
+      "action": "findMany",
+      "query": {
+        "arguments": {
+          "where": {
+            "jsonColumn": {
+              "$type": "Raw",
+              "value": {
+                "$type": "Decimal",
+                "value": "123"
+              }
+            }
+          }
+        },
+        "selection": {
+          "$composites": true,
+          "$scalars": true
         }
       }
-    },
-    "selection": {
-      "$composites": true,
-      "$scalars": true
     }
-  }
-}`)
+  `)
 })
 
 test('args - JsonNull field', () => {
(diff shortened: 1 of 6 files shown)


SevInf:
Note: wasm tests are failing because `/engine-branch` does not build custom wasm engine at the moment.

packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.test.ts
```diff
@@ -577,24 +577,29 @@ test('args - object with $type field', () => {
...
         }
       }
-    },
-    "selection": {
-      "$composites": true,
-      "$scalars": true
     }
-  }
-}`)
+  `)
```
Jolg42, line 602: Suggestion: keep the old test too so making it
- `args - object with $type field (Json)`
- `args - object with $type field (Raw)`

Jolg42 (approved):
Nice one!

PR #22859, opened 2024-01-29
fix(client): clarify import warnings' instructions

millsp:
This PR clarifies and simplifies instructions on some of the DX changes.

(This does not need to make it into the release)


packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
```diff
@@ -9,10 +9,11 @@ export const invalidImportWarningTs = [
   ' - For pnpm, run `pnpm add db@link:./path/to/your/client`',
   ' - For yarn, run `yarn add db@link:./path/to/your/client`',
   '-----',
-  '3. If you have a `tsconfig.json`, set `moduleResolution` to `node16`, `nodenext`, or `bundler`.',
-  'Alternatively, if you prefer, setting `module` to `node16` or `nodenext` equally works.',
+  '3. If you have a `tsconfig.json`, you will need to set `module` and `moduleResolution`.',
+  'Set `moduleResolution` to `nodenext`, or `bundler` and set `module` to `nodenext`.',
```
millsp, line 13:
```suggestion
  'Set `moduleResolution` to `nodenext` or `bundler`, and set `module` to `nodenext`.',
```

packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
```diff
@@ -9,10 +9,11 @@ export const invalidImportWarningTs = [
   ' - For pnpm, run `pnpm add db@link:./path/to/your/client`',
   ' - For yarn, run `yarn add db@link:./path/to/your/client`',
   '-----',
-  '3. If you have a `tsconfig.json`, set `moduleResolution` to `node16`, `nodenext`, or `bundler`.',
-  'Alternatively, if you prefer, setting `module` to `node16` or `nodenext` equally works.',
```
millsp, line 13: We don't need to mention `Node16`, it's deprecated and `NodeNext` is equivalent to it (as per TS [docs](https://www.typescriptlang.org/tsconfig#node16nodenext)), less is better here.

millsp: Also, now we are pointing users to set both rather than explaining all possible and invalid combinations, and when  and when it is not ok to have just the one or the other.

Druue (approved):
✨

Jolg42 (approved):

sandbox/d1/package-lock.json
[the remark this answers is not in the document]
millsp: It's a small off-topic cleanup indeed, we don't use npm so we shouldn't have it there.

packages/client/src/generation/TSClient/utils/invalidImportWarning.ts
Silvenga: @millsp my impression was that `Node16` was a "breaking change" flag meaning "Node 16+", where `NodeNext` would always hug bleeding edge (the ts docs just claim both are the same). It might make sense to add the old language because most people likely should be on `Node16` (supported by the [example tsconfig](https://www.npmjs.com/package/@tsconfig/node22)).

PR #22864, opened 2024-01-30
fix(client): smaller `util` polyfill

millsp:
This PR ditches the `util` polyfill in favor of a hand written approach (we only implement/polyfill what we need). This improves gzip compression.

### DIFF
--- a/helpers/compile/plugins/fill-plugin/fillers/util.ts
+++ b/helpers/compile/plugins/fill-plugin/fillers/util.ts
@@ -1,0 +1,52 @@
+export function promisify(fn: (...args: any[]) => any) {
+  return (...args: any[]) => {
+    return new Promise((resolve, reject) => {
+      fn(...args, (err: any, result: any) => {
+        if (err) {
+          reject(err)
+        } else {
+          resolve(result)
+        }
+      })
+    })
+  }
+}
+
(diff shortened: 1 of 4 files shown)


helpers/compile/plugins/fill-plugin/fillers/util.ts
millsp, line 1: Should we add unit tests for these, or are functional tests enough in your opinion?

helpers/compile/plugins/fill-plugin/fillers/util.ts
```diff
@@ -0,0 +1,52 @@
...
+      return {
+        ...value,
+        message: value.message,
+        stack: value.stack,
+      }
+    }
+
+    return value
+  })
+}
```
jkomyno, line 41: Is `inspect` only used for DataProxy? If so, is there a way of excluding it from the bundle directly when targeting `driverAdapters` with Wasm? Does the current implementation achieve so?

jkomyno (approved):

helpers/compile/plugins/fill-plugin/fillers/util.ts
millsp: It is also used by dependencies.

PR #22931, opened 2024-02-02
feat(get-platform): add FreeBSD 15 as a new platform

obiwac:


packages/get-platform/src/binaryTargets.ts
```diff
@@ -26,6 +26,7 @@ export type BinaryTarget =
   | 'freebsd12'
   | 'freebsd13'
   | 'freebsd14'
+  | 'freebsd15'
```
aqrln, line 29: You are also missing it in the `binaryTargets` array below

aqrln (commented):
Thank you! Looks good except one comment

packages/get-platform/src/binaryTargets.ts
obiwac: yes, my bad, it was unstaged ;)

aqrln (approved):
LGTM, thank you!

PR #22953, opened 2024-02-05
feat: bump @planetscale/database to 1.15.0

### LINKED ISSUE #161: bug: `BLOB` / `Buffer` input parameters result in corrupted data
Hey again 👋🏻, Alberto from @prisma here.

While working on our new [Driver Adapters](https://planetscale.com/docs/tutorials/planetscale-serverless-driver-prisma-example) feature, we noticed that `@planetscale/database` doesn't handle `Bytes` fields correctly, when compared to other JavaScript drivers. In particular, the binary values we insert do not always match the values retrieved by `@planetscale/database` (see https://github.com/prisma/team-orm/issues/687). Here is a TypeScript reproduction.

## Scaffolding

Consider a PlanetScale database with the following table,

```sql
CREATE TABLE `binary_test` (
  `id` varchar(191) NOT NULL,
  `bytes` longblob NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE InnoDB,
  CHARSET utf8mb4,
  COLLATE utf8mb4_unicode_ci;
```

generated from this Prisma model:

```prisma
model binary_test {
  id String @id
  bytes Bytes
}
```

Let us also assume that the connection string to such PlanetScale database is available in the `DATABASE_URL` environment variable.

## The problem

Let's say we want to store an input string in `binary_test`.`bytes`, a `BLOB` field.
To do so, we:

1. Wrap the string in a Node.js `Buffer`, and apply `base64` encoding.
2. Pass the buffer as a query parameter to `@planetscale/database`.
3. Query the record we've just created.
4. Read the raw `bytes` column value
5. Place this value into a `Buffer`
6. Apply the `base64` decoding to the `Buffer`  

Here's an example with the input string `FSDF`

```typescript
import { Client } from '@planetscale/database'

const connectionString = `${process.env.DATABASE_URL}`

async function mainPlanetscale() {
  const input = 'FSDF'
  console.log('input', input)

  const inputAsBuffer = Buffer.from(input, 'base64')
  console.log('inputAsBuffer', inputAsBuffer)

  await client.execute('DELETE FROM `binary_test`')
  await client.execute('INSERT INTO `binary_test` (`id`, `bytes`) VALUES (1, ?)', [inputAsBuffer])
  const result = await client.execute('SELECT `id`, `bytes` FROM `binary_test`')

  const outputRaw = result.rows[0]['bytes']
  console.log('outputRaw', outputRaw)
  console.log('typeof outputRaw', typeof outputRaw)

  const outputAsBuffer = Buffer.from(outputRaw)
  console.log('outputAsBuffer', outputAsBuffer)

  const output = outputAsBuffer.toString('base64')
  console.log('output', output)
  console.log('`input === output`', input === output)
}

mainPlanetscale().catch((e) => {
  console.error(e)
  process.exit(1)
})
```

which yields the following to stdout:

```
input FSDF
inputAsBuffer <Buffer 15 20 c5>
outputRaw  ï¿½
typeof outputRaw string
outputAsBuffer <Buffer 15 20 c3 af c2 bf c2 bd>
output FSDDr8K/wr0=
`input === output` false
```

Notice the discrepancy between the first and the second-last line, which doesn't match our expectations:

```diff
- FSDF
+ FSDDr8K/wr0=
```

It is interesting to observe that only the first 2 bytes of the input are preserved in the output.

---

Let's compare this to the `mysql2` driver, run against the same PlanetScale instance:

```typescript
// install via `npm i -S mysql2`
import mysql from 'mysql2/promise'

const connectionString = `${process.env.DATABASE_URL}`

async function mainMysql2() {
  const client = await mysql.createConnection({
    uri: connectionString,
    ssl: {
      verifyIdentity: false,
      rejectUnauthorized: false,
    }
  })

  const input = 'FSDF'
  console.log('input', input)

  const inputAsBuffer = Buffer.from(input, 'base64')
  console.log('inputAsBuffer', inputAsBuffer)

  await client.execute('DELETE FROM `binary_test`')
  await client.execute('INSERT INTO `binary_test` (`id`, `bytes`) VALUES (1, ?)', [inputAsBuffer])
  const result = await client.query('SELECT `id`, `bytes` FROM `binary_test`')

  const outputRaw = result[0][0]['bytes']
  console.log('outputRaw', outputRaw)
  console.log('typeof outputRaw', typeof outputRaw)

  const outputAsBuffer = Buffer.from(outputRaw)
  console.log('outputAsBuffer', outputAsBuffer)

  const output = outputAsBuffer.toString('base64')
  console.log('output', output)
  console.log('`input === output`', input === output)

  await client.end()
}

mainMysql2().catch((e) => {
  console.error(e)
  process.exit(1)
})
```

which yields the following to stdout:

```
input FSDF
inputAsBuffer <Buffer 15 20 c5>
outputRaw <Buffer 15 20 c5>
typeof outputRaw object
outputAsBuffer <Buffer 15 20 c5>
output FSDF
`input === output` true
```

In particular, we see that `input` (first line) matches `output` (second-last line) when using `mysql2`,.

---

Does this `@planetscale/database-js` byte discrepancy problem occur for every byte input? Not really. For instance, let's run `mainPlanetScale` with the input string `AQID`:

```
input AQID
inputAsBuffer <Buffer 01 02 03>
outputAsString 
typeof outputAsString string
outputAsBuffer <Buffer 01 02 03>
output AQID
`input === output` true
```

jkomyno:
This PR contributes to https://github.com/prisma/team-orm/issues/687, as `@planetscale/database@1.15.0` fixes https://github.com/planetscale/database-js/issues/161.

### DIFF
--- a/packages/adapter-planetscale/src/conversion.ts
+++ b/packages/adapter-planetscale/src/conversion.ts
@@ -101,5 +101,12 @@ export const cast: typeof defaultCast = (field, value) => {
   if (field.type === 'JSON' && value === 'null') {
     return JsonNullMarker
   }
-  return defaultCast(field, value)
+
+  const defaultValue = defaultCast(field, value)
+
+  if (defaultValue instanceof Uint8Array) {
+    return Array.from(defaultValue)
+  }
+
+  return defaultValue
 }
--- a/packages/adapter-planetscale/src/planetscale.test.ts
+++ b/packages/adapter-planetscale/src/planetscale.test.ts
@@ -1,10 +1,11 @@
 import { Client, connect } from '@planetscale/database'
+import { fetch as undiciFetch } from 'undici'
 
 import { PrismaPlanetScale } from './planetscale'
 
 describe('validation', () => {
   test('throws if passed Connection instance', () => {
-    const connection = connect({ url: 'http://example.com' })
+    const connection = connect({ url: 'http://example.com', fetch: undiciFetch })
 
     expect(() => {
       // @ts-ignore
--- a/packages/adapter-planetscale/src/planetscale.test.ts
+++ b/packages/adapter-planetscale/src/planetscale.test.ts
@@ -1,11 +1,10 @@
 import { Client, connect } from '@planetscale/database'
-import { fetch as undiciFetch } from 'undici'
 
 import { PrismaPlanetScale } from './planetscale'
 
 describe('validation', () => {
   test('throws if passed Connection instance', () => {
-    const connection = connect({ url: 'http://example.com', fetch: undiciFetch })
+    const connection = connect({ url: 'http://example.com' })
 
     expect(() => {
       // @ts-ignore
--- a/packages/adapter-planetscale/package.json
+++ b/packages/adapter-planetscale/package.json
@@ -43,8 +43,7 @@
     "@swc/core": "1.3.96",
     "@swc/jest": "0.2.31",
     "jest": "29.7.0",
-    "jest-junit": "16.0.0",
-    "undici": "5.28.1"
+    "jest-junit": "16.0.0"
   },
   "peerDependencies": {
     "@planetscale/database": "^1.15.0"
--- a/packages/adapter-planetscale/package.json
+++ b/packages/adapter-planetscale/package.json
@@ -39,11 +39,12 @@
     "@prisma/driver-adapter-utils": "workspace:*"
   },
   "devDependencies": {
+    "@planetscale/database": "1.15.0",
     "@swc/core": "1.3.96",
     "@swc/jest": "0.2.31",
-    "@planetscale/database": "1.15.0",
     "jest": "29.7.0",
-    "jest-junit": "16.0.0"
+    "jest-junit": "16.0.0",
+    "undici": "5.28.1"
   },
   "peerDependencies": {
     "@planetscale/database": "^1.15.0"
--- a/packages/adapter-planetscale/package.json
+++ b/packages/adapter-planetscale/package.json
@@ -41,11 +41,11 @@
   "devDependencies": {
     "@swc/core": "1.3.96",
     "@swc/jest": "0.2.31",
-    "@planetscale/database": "1.14.0",
+    "@planetscale/database": "1.15.0",
     "jest": "29.7.0",
     "jest-junit": "16.0.0"
   },
   "peerDependencies": {
-    "@planetscale/database": "^1.14.0"
+    "@planetscale/database": "^1.15.0"
   }
 }
--- a/packages/bundle-size/package.json
+++ b/packages/bundle-size/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "@libsql/client": "0.4.2",
     "@neondatabase/serverless": "0.7.2",
-    "@planetscale/database": "1.14.0",
+    "@planetscale/database": "1.15.0",
     "@prisma/adapter-neon": "workspace:*",
     "@prisma/adapter-pg": "workspace:*",
     "@prisma/adapter-planetscale": "workspace:*",
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -158,7 +158,7 @@
     "@opentelemetry/resources": "1.21.0",
     "@opentelemetry/sdk-trace-base": "1.21.0",
     "@opentelemetry/semantic-conventions": "1.21.0",
-    "@planetscale/database": "1.14.0",
+    "@planetscale/database": "1.15.0",
     "@prisma/adapter-d1": "workspace:*",
     "@prisma/adapter-libsql": "workspace:*",
     "@prisma/adapter-neon": "workspace:*",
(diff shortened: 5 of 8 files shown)


packages/adapter-planetscale/src/planetscale.test.ts
```diff
@@ -1,10 +1,11 @@
 import { Client, connect } from '@planetscale/database'
+import { fetch as undiciFetch } from 'undici'
 
 import { PrismaPlanetScale } from './planetscale'
 
 describe('validation', () => {
   test('throws if passed Connection instance', () => {
-    const connection = connect({ url: 'http://example.com' })
+    const connection = connect({ url: 'http://example.com', fetch: undiciFetch })
```
jkomyno, line 8: This change was somehow required for me to build & test locally.
It doesn't seem to hurt, though, so I'd keep it.

packages/adapter-planetscale/src/conversion.ts
```diff
@@ -101,5 +101,12 @@ export const cast: typeof defaultCast = (field, value) => {
...
   }
-  return defaultCast(field, value)
+
+  const defaultValue = defaultCast(field, value)
+
+  if (defaultValue instanceof Uint8Array) {
+    return Array.from(defaultValue)
+  }
+
+  return defaultValue
```
jkomyno, line 111: This will probably end up in PlanetScale's upstream, but for the time being it lets the TypeScript functional Client test `driver-adapters/team-orm-687-bytes` pass.

packages/adapter-planetscale/src/conversion.ts
```diff
@@ -101,5 +101,12 @@ export const cast: typeof defaultCast = (field, value) => {
   if (field.type === 'JSON' && value === 'null') {
     return JsonNullMarker
   }
-  return defaultCast(field, value)
+
+  const defaultValue = defaultCast(field, value)
+
+  if (defaultValue instanceof Uint8Array) {
+    return Array.from(defaultValue)
+  }
```
millsp, line 109: Why does it need to be turned into an array?

jkomyno: Otherwise tests fail. A `Uint8Array` passed to Napi / Wasm would look like an `serde_json::Value::Object` shaped as:

```json
{
  "0": 21,
  "1": 25,
  "2": ...
}
```

jkomyno: Applying the `if` above makes `PlanetScale` behave exactly like `Pg`.

SevInf: Hm... Can we somehow force napi to accept to `Uint8Array` in there and adjust neon as well? Transfering array buffer between JS and napi is really cheap (it basically amounts to transfering a pointer) where array requires a copy.

packages/adapter-planetscale/src/planetscale.test.ts
SevInf: Weird, this test should not get anywhere where `fetch` would've been required. But maybe planetscale changed something.

packages/adapter-planetscale/src/conversion.ts
jkomyno: Improving performance would indeed be better.
Would you be open to:
- merge this PR and its companion https://github.com/prisma/prisma-engines/pull/4695 now
- open an tech-debt issue to share `Uint8Array`s directly between `@prisma/client` and engines, which we can work on this Friday?

(I'm proposing this because [this issue](https://github.com/prisma/team-orm/issues/687) has been open for a while, and with this PR at least PlanetScale works with `Bytes`. Temporary closing this "byte correctness context" would also allow me to better focus on https://github.com/prisma/team-orm/issues/892).

packages/adapter-planetscale/src/planetscale.test.ts
jkomyno: I will try removing this and try again on the CI.

jkomyno: https://github.com/prisma/prisma/pull/22953/commits/1a41661b0da4a6b8d0a10742af87c6b3dc91cfa4

jkomyno: Looks like this was just a local problem of mine. I'm closing this conversation as the change has been reverted, and the CI is still green.

packages/adapter-planetscale/src/conversion.ts
SevInf: Yes, the plan is fine with me.

SevInf (approved):
Approved, provided that we'll do a [follow-up](https://github.com/prisma/prisma/pull/22953#discussion_r1479869536) rather soon.

packages/adapter-planetscale/src/conversion.ts
jkomyno: Thanks. I've created https://github.com/prisma/team-orm/issues/926.

PR #22957, opened 2024-02-05
feat(client): Use split WASM engine

SevInf:
Implements client part of WASM engine splitting.
All 3 engines are bundled together during build of `@prisma/client`. During generation, we pick appropriate one depending on datasource provider, specified in schema.

Since each engine now it's own `wasm_bg.js` file and it's generally not possible to reuse it, we are no longer bundling it. Instead, we are copying one of the 3 possible files into generated client and loading it inside of generated js.

Close prisma/team-orm#891

/integration


packages/client/helpers/build.ts
```diff
@@ -34,6 +38,16 @@ function nodeRuntimeBuildConfig(targetBuildType: typeof TARGET_BUILD_TYPE): Buil
   }
 }
 
+function wasmBindgenRuntimeConfig(provider: DriverAdapterSupportedProvider): BuildOptions {
```
SevInf, line 41: Doing this build for 2 reasons:
1. convert ESM to CJS
2. Minify the file.

packages/client/src/generation/generateClient.ts
```diff
@@ -360,18 +362,16 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   }
 
   // copy the necessary engine files needed for the wasm/driver-adapter engine
-  if (generator.previewFeatures.includes('driverAdapters') && noEngine !== true) {
-    const queryEngineWasmFilePath = path.join(runtimeDir, 'query-engine.wasm')
-    const queryEngineWasmTargetPath = path.join(outputDir, 'query-engine.wasm')
-    // some bundlers (eg. webpack) need this file to exist, even if it's empty
-    // this is because they analyze `query-engine.wasm` for references to other
-    // files. It does not matter for us, because we bundle query_engine_bg.js.
-    const dummyQueryEngineBgTargetPath = path.join(outputDir, 'query_engine_bg.js')
```
SevInf, line 369: We no longer need a dummy one, since we are not bundling `_bg.js` file anymore and real file will be present at bundling time.

SevInf:
Ecosystem tests need adjustments, but this PR is ready for review.

packages/client/src/generation/generateClient.ts
```diff
@@ -360,18 +362,16 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
...
-    const dummyQueryEngineBgContents = '/** Dummy file needed by some bundlers when using `query-engine.wasm` */'
-
-    const copyOrNoop = testMode ? function noop() {} : fs.copyFile
-    await copyOrNoop(queryEngineWasmFilePath, queryEngineWasmTargetPath)
-    await fs.writeFile(dummyQueryEngineBgTargetPath, dummyQueryEngineBgContents)
+  if (generator.previewFeatures.includes('driverAdapters') && isWasmEngineSupported(provider)) {
+    const suffix = provider === 'postgres' ? 'postgresql' : provider
```
millsp, line 366: Could we fix this upstream?

packages/client/src/generation/generateClient.ts
```diff
@@ -360,18 +362,16 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
...
+    if (noEngine !== true && !testMode) {
+      await fs.copyFile(
+        path.join(runtimeDir, `query_engine_bg.${suffix}.wasm`),
+        path.join(outputDir, `query_engine_bg.wasm`),
+      )
+    }
+
+    await fs.copyFile(path.join(runtimeDir, `query_engine_bg.${suffix}.js`), path.join(outputDir, `query_engine_bg.js`))
```
millsp, line 374: Actually, I think this line could go into the condition above, correct?

packages/client/helpers/build.ts
```diff
@@ -34,6 +38,16 @@ function nodeRuntimeBuildConfig(targetBuildType: typeof TARGET_BUILD_TYPE): Buil
...
+  return {
+    name: `query_engine_bg.${provider}`,
+    entryPoints: [`@prisma/query-engine-wasm/${provider}/query_engine_bg.js`],
+    outfile: `runtime/query_engine_bg.${provider}`,
+    minify: true,
+    sourcemap: 'linked',
```
millsp, line 47: Do we need a sourcemap?

packages/client/src/runtime/core/engines/common/Engine.ts
```diff
@@ -147,14 +148,31 @@ export interface EngineConfig {
...
+   */
+  engineWasm?: WasmLoadingConfig
+}
+
+export type WasmLoadingConfig = {
+  /**
+   * WASM-bindgen runtime for corresponding module
+   */
+
+  runtime: WasmRuntime
```
millsp, line 162:
```suggestion
  /**
   * WASM-bindgen runtime for corresponding module
   */
  runtime: WasmRuntime
```

packages/client/src/runtime/core/engines/common/Engine.ts
```diff
@@ -147,14 +148,31 @@ export interface EngineConfig {
...
    * @remarks this is a callback on purpose, we only load the wasm if needed.
    * @remarks only used by LibraryEngine.ts
    */
-  getQueryEngineWasmModule?: () => Promise<unknown>
+  getQueryEngineWasmModule: () => Promise<unknown>
+}
+
+type WasmRuntime = {
+  __wbg_set_wasm(exports: unknown)
+  QueryEngine: QueryEngineConstructor
```
millsp, line 175: Is it useful on its own of could we inline it above?

packages/client/src/generation/generateClient.ts
SevInf: I think you are right, we'll just need to adjust generation to skip `require` for this with `--no-engine` as well.

packages/client/helpers/build.ts
SevInf: Not really, I'll remove it

packages/client/src/runtime/core/engines/library/WasmLibraryLoader.ts
```diff
@@ -21,12 +21,19 @@ export const wasmLibraryLoader: LibraryLoader = {
       )
     }
 
+    if (engineWasm === undefined) {
+      throw new PrismaClientInitializationError(
+        `WASM engine loading was not configured correctly. Please, re-run \`prisma generate\``,
```
millsp, line 26: How could `prisma generate` fix the issue?

packages/client/tests/functional/_utils/setupTestSuiteClient.ts
```diff
@@ -113,20 +114,26 @@ export function setupTestSuiteClientDriverAdapter({
...
+
+              // TODO: cleanup and remove ts-expect-error
+              // Pierre thinks that this is an upstream problem with the type definitions
+              // (provided by TS or @types/node) and will get fixed at some point.
+              // @ts-expect-error (2511) - Cannot create an instance of an abstract class.
+              return new globalThis.WebAssembly.Module(queryEngineWasmFileBytes)
+            },
+          }
+        : undefined,
+    })
```
millsp, line 136: nit: A mutation would be easier to read, is there a preference needed in this case?

millsp (approved):
🚀

millsp:
There is an error at runtime in ecosystem [tests](https://github.com/prisma/ecosystem-tests/actions/runs/7804095155/job/21285496804?pr=4519#step:8:181), you might need to apply our filler plugin to the wasm-bg file.

packages/client/src/runtime/core/engines/library/WasmLibraryLoader.ts
SevInf: Leftover from an earlier version of the code where `runtime` was not nested under wasm and it was possible to use older client with newer runtime accidentally. Now you are right, the message is misleading, I'll change it.

SevInf:
@millsp FYI: at some point I got lost while reading `noEngine` double negatives and replaced it with `copyEngine` property that is `true` by default and set to `false` when `--no-engine` flag is passed. Change is purely internal, CLI flag and parameter passed to generator remained unchanged.

millsp (approved):

PR #22958, opened 2024-02-05
fix(client): engine loading on vercel & improved custom output dx

millsp:
This PR enables us to remove the complex wasm configuration required to deploy `driverAdapters` with Next.js. Additionally, it also removed the need to configure anything when generating in a custom output location. On this last point, it means even paths like `import { PrismaClient } from "../my/custom/output"` will work with no extra configuration in workers. For context: when using custom output paths, one cannot directly import from the path, but must install the custom output as a local package of the project. Thanks to the `imports` map, we can avoid this friction.

closes https://github.com/prisma/team-orm/issues/920
closes https://github.com/prisma/team-orm/issues/982

ecosystem-tests https://github.com/prisma/ecosystem-tests/pull/4613

/integration


packages/client/package.json
```diff
@@ -30,25 +30,25 @@
...
         "node": "./default.js",
-        "edge-light": "./wasm.js",
-        "workerd": "./wasm.js",
-        "worker": "./wasm.js",
+        "edge-light": "./default.js",
+        "workerd": "./default.js",
+        "worker": "./default.js",
         "browser": "./index-browser.js"
       },
       "default": "./default.js"
```
millsp, line 54: We point everything to `./default.js` which is a trampoline.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -77,19 +74,16 @@ export class TSClient implements Generatable {
...
-      importWarning,
       reusedJs,
     } = this.options
 
-    if (reusedJs && importWarning) {
-      const topExports = `module.exports = { ...require('./${reusedJs}.js') }`
-      const warning = `console.warn('${invalidImportWarningJs.join('\\n')}')`
-
-      return `${[topExports, warning].join('\n\n')}`
-    }
```
millsp, line 89: Import warnings go away, both JS and TS.

packages/client/src/generation/generateClient.ts
```diff
@@ -112,7 +111,7 @@ export async function buildClient({
   const defaultClient = new TSClient({
     ...nodeClientOptions,
     reusedTs: 'index',
-    reusedJs: 'index',
+    reusedJs: '.',
```
millsp, line 114: Forces it through the exports map, this is temporary. When we GA, this `default` client will be deleted/replaced.

packages/client/src/generation/generateClient.ts
```diff
@@ -124,12 +123,38 @@ export async function buildClient({
     edge: true,
   })
 
+  const trampolineTsClient = new TSClient({
+    ...nodeClientOptions,
+    reusedTs: 'index',
+    reusedJs: '#main-entry-point',
+  })
```
millsp, line 130: This is the core surface of the change. When `require('#main-entry-point')` is executed, it goes through an `imports` map.

packages/client/src/generation/generateClient.ts
```diff
@@ -124,12 +123,36 @@ export async function buildClient({
...
+
   const pkgJson = {
     name: getUniquePackageName(datamodel),
     main: 'index.js',
     types: 'index.d.ts',
     browser: 'index-browser.js',
-    exports: clientPackageExports,
+    exports: {
+      ...clientPkg.exports,
+      ...{ '.': exportsMapDefault },
```
millsp, line 154: This is just for backwards compatibility and must be removed before `driverAdapters` GA.

packages/client/src/generation/generateClient.ts
```diff
@@ -141,26 +164,38 @@ export async function buildClient({
   fileMap['default.js'] = await JS(defaultClient)
   fileMap['default.d.ts'] = await TS(defaultClient)
   fileMap['index-browser.js'] = await BrowserJS(nodeClient)
-  fileMap['package.json'] = JSON.stringify(pkgJson, null, 2)
```
millsp, line 144: Moved later below.

packages/client/src/generation/generateClient.ts
```diff
@@ -141,26 +164,38 @@ export async function buildClient({
...
+    // - defaultClient can be deleted since trampolineTsClient will replace it.
+    //   - Special handling of . paths in TSClient.ts can also be removed.
+    // - The main @prisma/client exports map can be simplified:
+    //   - Everything can point to `default.js`, including browser fields.
+    //   - Exports map's `.` entry can be made like the others (e.g. `./edge`).
+    // - exportsMapDefault can be deleted as it's only needed for defaultClient:
+    //   - #main-entry-point can handle all the heavy lifting on its own.
+    //   - Always using #main-entry-point is kept for GA (small breaking change).
+    //   - exportsMapDefault can be inlined down below and removed everywhere else.
+    // In short: A lot can be simplified, but can only happen in GA & P6.
```
millsp, line 183: Summary of what can be simplified/done before GA.

packages/client/src/generation/generateClient.ts
```diff
@@ -141,26 +164,38 @@ export async function buildClient({
...
+    fileMap['wasm-edge-light-loader.js'] = `export default (await import('./query_engine_bg.wasm?module')).default`
+    pkgJson['browser'] = 'default.js' // also point to the trampoline client otherwise it is picked up by cfw
+    pkgJson['imports'] = {
+      // when `import('#wasm-engine-loader')` is called, it will be resolved to the correct file
+      '#wasm-engine-loader': {
+        'edge-light': './wasm-edge-light-loader.js',
+        workerd: './wasm-worker-loader.js',
+        worker: './wasm-worker-loader.js',
+        default: './wasm-worker-loader.js',
+      },
```
millsp, line 196: This fixes the DX for the engine loading on Vercel. By having a custom engine laoder for Vercel and Cloudflare we can remove the extra configuration needed in `next.config.js` to be able to load the wasm. Now it will load out of the box.

packages/client/src/generation/generateClient.ts
```diff
@@ -141,26 +164,38 @@ export async function buildClient({
...
+      // when `require('#main-entry-point')` is called, it will be resolved to the correct file
+      '#main-entry-point': exportsMapDefault,
```
millsp, line 198: This will basically act as an exports map, except it is done on internal `import`s, while the `exports` map acted on the `import` of this library (so: internal imports vs user imports). This will improve the DX around custom output paths because `imports` map are able to work outside `node_modules`, while `exports` maps can't. In other words, by switching to `imports`, users will still be able to `import { PrismaClient } from "../my/custom/output"` and it will be able to dynamically load the `wasm` client if on a eg. worker. Hence, we can also delete all the warnings that we added (both in TS and JS!).

packages/client/package.json
```diff
@@ -30,25 +30,25 @@
...
     ".": {
       "require": {
         "types": "./default.d.ts",
         "node": "./default.js",
-        "edge-light": "./wasm.js",
-        "workerd": "./wasm.js",
-        "worker": "./wasm.js",
+        "edge-light": "./default.js",
+        "workerd": "./default.js",
+        "worker": "./default.js",
```
millsp, line 43: While these are redundant, they are actually needed. If `wrangler` does not see its key, it will default to `browser` as a consequence. We will be able to simplify this before GA though, comments below.

packages/cli/src/Generate.ts
```diff
@@ -290,33 +288,13 @@ import { PrismaClient } from '${importPath}/${isDeno ? 'deno/' : ''}edge${isDeno
...
-${dim('```')}
-
 More information: https://pris.ly/d/client`
           } else {
             hint = `
 ${bold('Start using Prisma Client')}
 ${dim('```')}
 ${highlightTS(`\
-import { PrismaClient } from ${importPath}
+import { PrismaClient } from '${importPath}'
```
janpio, line 297: Was this a bug before?

jkomyno: It seems it was a typo in our stdout message, yes

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -169,42 +163,7 @@ ${buildNFTAnnotations(edge || !copyEngine, clientEngineType, binaryTargets, rela
...
-      prismaClientClassConstructor.addParameter(prismaClientClassConstructorArg.optional())
-
-      prismaClientClass.add(prismaClientClassConstructor)
-
-      const prismaClientClassExport = ts.moduleExport(prismaClientClass)
-      prismaClientClassExport.setDocComment(deprecationComment)
-
-      return [topExports, topImports, prismaClientClassExport].map((v) => ts.stringify(v)).join('\n\n')
-    }
+    const { reusedTs } = this.options
```
jkomyno, line 166: Nice cleanup

jkomyno (approved):
Reviewed the code, and tested it locally on our sandbox 🚀

packages/client/src/generation/generateClient.ts
```diff
@@ -124,12 +123,36 @@ export async function buildClient({
...
+    reusedJs: '#main-entry-point',
+  })
+
+  const exportsMapBase = {
+    node: './index.js',
+    'edge-light': './wasm.js',
+    workerd: './wasm.js',
+    worker: './wasm.js',
+    browser: './index-browser.js',
+    default: './index.js',
```
millsp, line 138: Order is important! Add a comment

packages/client/src/generation/generateClient.ts
```diff
@@ -141,26 +164,38 @@ export async function buildClient({
...
+    // - defaultClient can be deleted since trampolineTsClient will replace it.
+    //   - Special handling of . paths in TSClient.ts can also be removed.
+    // - The main @prisma/client exports map can be simplified:
+    //   - Everything can point to `default.js`, including browser fields.
+    //   - Exports map's `.` entry can be made like the others (e.g. `./edge`).
+    // - exportsMapDefault can be deleted as it's only needed for defaultClient:
+    //   - #main-entry-point can handle all the heavy lifting on its own.
+    //   - Always using #main-entry-point is kept for GA (small breaking change).
+    //   - exportsMapDefault can be inlined down below and MUST be removed elsewhere.
+    // In short: A lot can be simplified, but can only happen in GA & P6.
```
millsp, line 183: leave a note: `jest` 28 will break with `#main-entry-point` so we might want to change this name to match a file eg. `./main-entry-point`. @SevInf

SevInf: Entires in import map have to start with `#`, according to [Node docs](https://nodejs.org/api/packages.html#subpath-imports:~:text=Entries%20in%20the%20%22imports%22%20field%20must%20always%20start%20with%20%23%20to%20ensure%20they%20are%20disambiguated%20from%20external%20package%20specifiers)

PR #22960, opened 2024-02-05
ci(bundle-size): properly diff sizes on PRs

millsp:
I have this PR, which reduces the [size](https://github.com/prisma/prisma/pull/22864), but as you can see, the report says it does not on the gzip size while it does on the bundle. These numbers are incorrect obviously, and it's because I think I am not letting the github action do its thing properly. This PR attempts to fix that.


jkomyno (commented):
✅

PR #22961, opened 2024-02-05
fix(client): schema compression

millsp:
This PR moves away from base64 inline schema string encoding in favor of storing it as a plain string. In turns, this results in better gzip compression overall.

Needs https://github.com/prisma/studio-code/pull/538 for CLI tests to pass


packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -108,8 +108,8 @@ export class TSClient implements Generatable {
         ? (Object.keys(binaryPaths.libqueryEngine ?? {}) as BinaryTarget[])
         : (Object.keys(binaryPaths.queryEngine ?? {}) as BinaryTarget[])
 
-    const inlineSchema = (await readFile(schemaPath)).toString('base64')
-    const inlineSchemaHash = crypto.createHash('sha256').update(inlineSchema).digest('hex')
+    const inlineSchema = await readFile(schemaPath, 'utf8')
+    const inlineSchemaHash = crypto.createHash('sha256').update(btoa(inlineSchema)).digest('hex')
```
SevInf, line 112: Do we actually need to `btoa` schema here? IIRC, accelerate treats hash as opaque, so sending hash of a textual schema might work as well.

SevInf (approved):

jkomyno (approved):

packages/client/src/generation/TSClient/TSClient.ts
millsp: That was not the case at least with miniproxy, so I went with this.

PR #22962, opened 2024-02-05
fix(client): prune generated dmmf

millsp:
This PR prunes the generated dmmf in the client to its bare minimal version, and only does it when we are generating the `wasm` client. This improves gzip compression.


SevInf:
Do we know if that will affect any of the third party projects? `runtimeDataModel` itself is private, but is also used to re-create `Prisma.dmmf`.

millsp:
This change only targets `wasm` build. So to your point, I think we should at least throw an error for `Prisma.dmmf` there, as it may only be partially true otherwise?

SevInf:
Yes, i think error for `Prisma.dmmf` would be an ok way to go, if it is only about wasm build

SevInf (approved):

alexdoan102:
@millsp  @SevInf  @hayes  since prisma now support cloudflare 
   ==>  https://www.prisma.io/docs/orm/prisma-client/deployment/edge/deploy-to-cloudflare
   
   we need the Prisma.dmmf to use with 3rd party [@pothos/plugin-prisma] plugin.
    ==> this is the issue: https://github.com/hayes/pothos/issues/1190 
    
    can you  please put this property back, so that the graphql plugins will work for edge./cloudflare?

Issue #20021, opened 2023-06-30
Engines file missing when ignoring missing checksum


spudly:
### Bug description

I recently added the PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING environment variable because we use Artifactory, which has its own way of storing checksum files that doesn't match what the Prisma binary downloader expects. After doing so, I had difficulty getting the binaries downloaded and into the right location.

## First Issue:

Installing works as expected the first time, but the next time I need to reinstall (or install prisma into a different repo), it sees the binary in `~/.cache/prisma`, but never copies the files from cache into node_modules. The only way I've found to get the binaries installed is to run `rm -rf ~/.cache/prisma` before running `npm install`.

The problem is here: https://github.com/prisma/prisma/blob/main/packages/fetch-engine/src/download.ts#L278-L291

In that code, it ignores the missing checksum and allows the script to continue, but it never copies the binary file from cache. It just logs the checksum failure and returns.

## Second issue

When running `prism a generate`, I get this error:

```
ENOENT: no such file or directory, copyfile '<my-package>/node_modules/prisma/libquery_engine-darwin.dylib.node' -> '<my-package>/node_modules/.prisma/client/libquery_engine-darwin.dylib.node'
```

The file exists, but it's not in the `node_modules/prisma` folder, where it's trying to copy from. It's in `node_modules/@prisma/engines/`. I'm not sure where the problem is for this one, or if this is related to the first issue or not.

To work around this, I'm running the following command after installing dependencies:
```
find node_modules -name '*.node' -exec cp -n "{}" node_modules/prisma/ \;
```

### How to reproduce

1.  `cd` into a project that uses prisma
2.  run `export PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING=true`
3.  run `rm -rf ~/.cache/prisma` (clear cache so existing cache doesn't interfere with these steps)
4. `rm -rf node_modules` (remove node modules so that it actually does a full reinstall)
5. run `npm install` (this is the first install, and should add prisma binaries to cache, but it may fail to get checksums for those binaries)
6. To simulate an environment (like mine) where the binaries downloaded but the checksum files didn't, you'll probably have to run something like this: `find ~/.cache/prisma -name '*.sha256' -delete`
7. `rm -rf node_modules`
9. `npm install` (this is the second install, which should use the cached binaries, but it doesn't copy the binaries from cache)

### Expected behavior

Binaries should be installed during postinstall regardless of whether the binaries are cached or not

[...]

PR #23001, opened 2024-02-07
fix(client): prisma cannot be used in corporate environments without Internet access

RaHehl:
this should fix the bug #20021 introduced with pr #17312


janpio:
Can you explain the reasoning behind this change a bit? Why is the error message not needed at all? Why do your think your logic is valid?

RaHehl:
> Can you explain the reasoning behind this change a bit? Why is the error message not needed at all? Why do your think your logic is valid?

@janpio 

Yes of course:

My initial situation is similar to that described in the bug:

In this specific case, I'm trying to build a custom Hoppscotch backend container, as bugs make it impossible to use the unmodified version.

However, this fails because Prisma has problems downloading the binaries, which is nothing unusual as in my case the build environment is behind a corporate firewall that does not allow direct internet access.

But it's not dramatic at first, so I just changed PRISMA_ENGINES_MIRROR to our Prisma JFrog Artifactory mirror url (It's mainly a pull through cache).

The next problem is that the sha256 files are not loaded, apparently a problem with Artifactory that cannot be avoided at first.

But there's PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING to obviously ignore missing checksums in this case the sha256 files, which is exactly what I'm missing.

However, this means that the binaries apparently are no longer stored correctly, exactly as described here https://github.com/prisma/prisma/issues/20021

So I just looked in the prisma code to see what PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING does.

Then I noticed “packages/fetch-engine/src/download.ts”

Download definitely sounds right, from the area where the bug should be.

PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING appears exactly 1 time here, in a method called binaryNeedsToBeDownloaded which returns a boolean.

So from the name and the returns there is obviously a method that is supposed to determine whether a download of the binaries is necessary

So it looks like a hot lead for the problem.

And specifically it says (currently in line 308) if PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING is set, just don't do anything, give back, nothing has to be downloaded.

So again exactly what is described in the bug.

So first of all, this looks like the right place.

The debug message can be left in, I don't want to contradict you now.

But let's first look at the control flow in the method:

The first thing that happens is "if (cachedFile) {" since I haven't seen anything that activates skipCacheIntegrityCheck, we can ignore that for now.

The next thing that happens is "if (await exists(sha256FilePath)) {" and we obviously don't have the file in my scenario for which PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING was set.

And then we are back at the assumed point where absolutely nothing happens except to say that nothing should be downloaded.

So what is missing is:

Check whether the file is already there
If not, check whether there is such a cachedFile (I don't know the background of the cachedFile, but it sounds logical at first) if the cachedFile exists, just use it, then nothing has to be downloaded, so return false.

But if there is nothing there, a download obviously has to take place, so return true.

But if the target file exists, it doesn't have to be downloaded, so continue to return false.

RaHehl:
@janpio Do you have any idea how things could continue here?

janpio:
No, this is for the engineering team to pick up. No ETA on that, sorry.

Emulator000:
This is still an issue and the fix looks fine to me, I'm wondering the reason behind the initial logic: why put this new environment variable `PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING` if then the binaries are not copied in the `prisma` folder?

I think that the name is misleading in that case and it had to be something like `PRISMA_ENGINES_IGNORE_CHECKSUM_AND_SKIP_MISSING`.

I think that it's pretty obvious that the initial logic is not right if we want to skip&force the download of these binaries under a corporate environment.

RaHehl:
@janpio Is there anything you can do to get engineering to start working?

janpio:
No, they are aware of this PR and will get to it when they have time for it.

packages/fetch-engine/src/download.ts
```diff
@@ -306,10 +306,14 @@ async function binaryNeedsToBeDownloaded(
...
-      )
+      if (!targetExists) {
+        if (cachedFile) {
+            debug(`copying ${cachedFile} to ${job.targetFilePath}`)
+            await overwriteFile(cachedFile, job.targetFilePath)
+            return false
+          }
+        return true
+      }
       return false
```
aqrln, line 317:
```suggestion
      if (targetExists) {
        return false
      }
      if (cachedFile) {
        debug(`copying ${cachedFile} to ${job.targetFilePath}`)
        await overwriteFile(cachedFile, job.targetFilePath)
        return false
      }
      return true
```

aqrln (approved):
Thanks for the PR! The change looks good, I only have a style suggestion to make it more readable.

PR #23003, opened 2024-02-07
chore: use tsx instead of esbuild-register

millsp:
This PR trades `esbuild-register` for `tsx` three main reasons:
* `tsx` is more performant than `esbuild-register`
* `tsx` does not display uncontrolled warnings
* `tsx` CLI is simpler to use


millsp:
(failing test is unrelated)

jkomyno (approved):
I love `tsx` so much it's unreal

PR #23008, opened 2024-02-07
fix(client): optimize events polyfill

millsp:
This PR ditches `eventemitter3` in favor of a hand written polyfill. Contributes to better gzip sizes.

### DIFF
--- a/helpers/compile/plugins/fill-plugin/fillers/events.ts
+++ b/helpers/compile/plugins/fill-plugin/fillers/events.ts
@@ -9,6 +9,8 @@ export class EventEmitter {
     }
 
     this.events[event].push(listener)
+
+    return this
   }
 
   emit(event: string, ...args: any[]) {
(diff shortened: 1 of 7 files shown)


helpers/compile/plugins/fill-plugin/fillers/events.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  on(event: string, listener: Listener) {
+    if (!this.events[event]) {
+      this.events[event] = []
+    }
+
+    this.events[event].push(listener)
+
+    return this
+  }
```
jkomyno, line 14: The current API only allows event listeners to be added.
Shouldn't event listeners also be removed somehow? Isn't there a risk of memory leaks?

jkomyno: Nevermind, I see we never call `.off()` anyway.

jkomyno (approved):

helpers/compile/plugins/fill-plugin/fillers/events.ts
millsp: Prisma Client does not allow to remove listeners, so I don't think we need it until it is the case.

PR #23019, opened 2024-02-08
fix(instrumentation): do not pin OTEL dependencies

### LINKED ISSUE #21473: Fixed version of `@opentelemetry/*` dependencies
### Bug description

Currently prisms locks versions of OTEL libraries:

```json
    "@opentelemetry/api": "1.6.0",
    "@opentelemetry/instrumentation": "0.43.0",
    "@opentelemetry/sdk-trace-base": "1.17.0",
```

This creates problem with hoisting: if there are other versions of OTEL libs required in the project package manager will end up installing two copies of "@opentelemetry/api" in order to satisfy constraints:
- one in root `node_modules`
- second one in `node_modules/<another_package>/node_modules`

This will make either external or internal consumers of prisma to receive no-op implementation of a tracer.

### Expected behavior

A package should keep versions of it's dependencies as open as possible https://github.com/prisma/prisma/blob/811babfebbd8c573109f0dc083634bc2e865a87e/packages/instrumentation/package.json

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
```

### Environment & setup

- OS: 
- Database: 
- Node.js version: 

### Prisma Version

```

```

maxkomarychev:
Fixes issues with telemetry https://github.com/prisma/prisma/issues/21473 by using non-pinned versions of OTEL libraries.

Fixes #21473

### DIFF
--- a/packages/instrumentation/package.json
+++ b/packages/instrumentation/package.json
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "@opentelemetry/api": "^1.8",
-    "@opentelemetry/instrumentation": "^0.49 || ^0.50",
+    "@opentelemetry/instrumentation": "^0.49 || ^0.50 || ^0.51",
     "@opentelemetry/sdk-trace-base": "^1.22"
   },
   "files": [
--- a/packages/instrumentation/package.json
+++ b/packages/instrumentation/package.json
@@ -22,9 +22,9 @@
     "typescript": "5.3.3"
   },
   "dependencies": {
-    "@opentelemetry/api": "1.8.0",
-    "@opentelemetry/instrumentation": "0.50.0",
-    "@opentelemetry/sdk-trace-base": "1.23.0"
+    "@opentelemetry/api": "^1.8",
+    "@opentelemetry/instrumentation": "^0.49 || ^0.50",
+    "@opentelemetry/sdk-trace-base": "^1.22"
   },
   "files": [
     "dist"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1087,11 +1087,11 @@ importers:
         specifier: 1.8.0
         version: 1.8.0
       '@opentelemetry/instrumentation':
-        specifier: 0.50.0
+        specifier: ^0.49 || ^0.50
         version: 0.50.0(@opentelemetry/api@1.8.0)
       '@opentelemetry/sdk-trace-base':
-        specifier: 1.23.0
-        version: 1.23.0(@opentelemetry/api@1.8.0)
+        specifier: ^1.22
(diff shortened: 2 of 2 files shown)


packages/instrumentation/package.json
```diff
@@ -22,9 +22,9 @@
     "typescript": "5.3.3"
   },
   "dependencies": {
-    "@opentelemetry/api": "1.8.0",
-    "@opentelemetry/instrumentation": "0.49.1",
-    "@opentelemetry/sdk-trace-base": "1.22.0"
+    "@opentelemetry/api": "^1.7",
+    "@opentelemetry/instrumentation": "0.48 || 0.49",
```
Jolg42, line 26: Note that v0 behaves differently than v1
See online example 
https://jubianchi.github.io/semver-check/#/^0.48/0.49.1
Screenshot 2024-03-06 at 18 09 46

https://jubianchi.github.io/semver-check/#/0.48%20||%200.49/0.49.1
Screenshot 2024-03-06 at 18 09 29

maxkomarychev: I would actually make it open too

image


what do you think?

maxkomarychev: I think 0.48 appeared here only because pr is very old

maxkomarychev: I've pushed an update to just make it `^0.49`

maxkomarychev:
Dear prisma maintainers, could you please provide specific feedback regarding prospects of this being merged? Are you willing to accept this pr, should I keep keeping it up to date etc?

packages/instrumentation/package.json
```diff
@@ -22,9 +22,9 @@
     "typescript": "5.3.3"
   },
   "dependencies": {
-    "@opentelemetry/api": "1.8.0",
-    "@opentelemetry/instrumentation": "0.49.1",
-    "@opentelemetry/sdk-trace-base": "1.22.0"
+    "@opentelemetry/api": "^1.8",
+    "@opentelemetry/instrumentation": "^0.49",
```
aqrln, line 26:
```suggestion
    "@opentelemetry/instrumentation": "^0.49 || ^0.50",
```

aqrln (approved):
Sorry for the delay here — let's finally ship this

aqrln:
I think this may also need some adjustments in the Renovate config to avoid the versions being automatically pinned.
@Jolg42 could you help with this?

packages/instrumentation/package.json
maxkomarychev: thanks!

I appreciate your feedback. I've updated version ranges to include current and previous versions:



image


Thanks!

maxkomarychev:
thanks!

PR #23021, opened 2024-02-08
ci: use CODECOV_TOKEN

Jolg42:
Needed since https://github.com/prisma/prisma/pull/22908

To avoid failures that are reported as warnings, for example
https://github.com/prisma/prisma/actions/runs/7830436419
Screenshot 2024-02-08 at 15 35 49

While looking at this I requested adding the GitHub app which added 2 “checks”, these will always pass as I configured them to be informational.

We can remove this later if that is not useful
Screenshot 2024-02-08 at 15 49 26

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,6 +1,16 @@
 # Enable/Disable PR comments
 comment: false
 
+# https://docs.codecov.com/docs/commit-status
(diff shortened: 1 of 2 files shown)


Jolg42:
I disabled both for now.

PR #23029, opened 2024-02-08
fix(client): Reuse buffer polyfill in wasm-bindgen runtime

SevInf:
During engine split, we unbundled query_engine_bg.js file. That means,
that polyfills were no longer reused between the two. `fillPlugin`
helped to insert the import for `buffer`, but it did not inline it.

If we are to just enable bundling on wasm-bindgen runtime files,
polyfill will be duplicated between the two. So, instead we are doing
following:
- building `buffer` polyfill as the separate bundle
- inserting that import as external to both edge runtime and
  wasm-bindgen runtime files.

This should keep a single copy of a polyfill, while also fixing
ecosystem-tests.

### DIFF
--- a/helpers/compile/plugins/fill-plugin/fillPlugin.ts
+++ b/helpers/compile/plugins/fill-plugin/fillPlugin.ts
@@ -1,48 +1,13 @@
-import crypto from 'crypto'
 import * as esbuild from 'esbuild'
-import os from 'os'
 import path from 'path'
-import resolve from 'resolve'
-
-type LoadCache = { [K in string]: string }
 
 type Fillers = {
   [k in string]: {
-    imports?: string
+    imports?: string | { path: string; external: boolean }
     globals?: string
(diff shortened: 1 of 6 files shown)


helpers/compile/plugins/fill-plugin/fillPlugin.ts
```diff
@@ -1,50 +1,15 @@
...
 }
 
-/**
- * Bundle a polyfill with all its dependencies. We use paths to files in /tmp
- * instead of direct contents so that esbuild can include things once only.
- * @param cache to serve from
- * @param module to be compiled
- * @returns the path to the bundle
- */
-const loader = (cache: LoadCache) => (module: string) => {
```
SevInf, line 25: No longer used

SevInf:
Not needed anymore, `Buffer` is no longer referenced in wasm-bindgen runtime.

PR #23032, opened 2024-02-08
fix(client): optimize engine panic handling for wasm

millsp:
This PR prunes the `LibraryEngine`'s `RustPanic` related codepaths which cannot happen in our `wasm` runtime. This contributes to better gzip sizes.


Jolg42 (approved):

jkomyno:
> This PR prunes the `LibraryEngine`'s `RustPanic` related codepaths which cannot happen in our `wasm` runtime. This contributes to better gzip sizes.

Could you please remind us why "RustPanic related codepaths which cannot happen"?
The Query Engine can still crash unexpectedly at times, can it not?

millsp:
I don't have full knowledge there, this is after chatting with @SevInf, he will have the context.

SevInf:
>The Query Engine can still crash unexpectedly at times, can it not?

It can, but it can't report panic back to the client in the same way napi does.

PR #23087, opened 2024-02-12
chore: Add bundled-js-drivers package

SevInf:
Background: when we update driver version in prisma/prisma, more often
than not this breaks engine-side test executor until we update driver
there as well. This PR attempts to mitigate the problem in the following
way:
- adds internal meta-package with all of the drivers re-exported
- engine's will then remove direct depenency on drivers and pick them up
  from this package instead.
  
See https://github.com/prisma/prisma-engines/pull/4721 for usage.

Contribute to prisma/team-orm#940


Jolg42 (approved):
Nice one!

PR #23090, opened 2024-02-12
feat(cli): next version

jasonkuhrt:
This PR introduces the next version of the CLI. This PR includes BREAKING CHANGES of the current CLI interface.

3. [x] Rename APIKey to Service Token
1. [x] Use the PDP API
2. [x] Introduce environments
4. [x] Send client particulars over `user-agent` header


Jolg42:
@jasonkuhrt Note that this was merged today https://github.com/prisma/prisma/pull/23056

It should make it easier for you to run tests. And you have to put tests inside the platform directory for that (if you crate test files).

Let me know if you have any question

jasonkuhrt:
@Jolg42 Thanks will take a look! We're still at a stage of shipping a prototype so not sure we're going to have tests like last time. Appreciate you keeping us in the loop on developments there! 🙏

jasonkuhrt:
It looks like there are conflicts, I didn't expect that. Will investigate.

jasonkuhrt:
The conflict was trivially resolved by Git, it looked more intimidating than it was :D

packages/cli/src/platform/auth/login.ts
```diff
@@ -104,6 +104,7 @@ const generateAuthSigninUrl = async (params: { connection: string; redirectTo: s
     return null
   })
 
+  // todo change how we pass signature. Put into header user-agent, format idea: <sig>.<ver>.client.prisma
```
jasonkuhrt, line 107: todo

packages/cli/src/platform/auth/login.ts
```diff
@@ -27,111 +28,107 @@ export class Login implements Command {
...
+    client: userAgent,
     ...params,
   }
-  const stateEncoded = Buffer.from(JSON.stringify(state), `utf-8`).toString(`base64`)
-  const url = new URL('/auth/cli', platformConsoleUrl)
+  const stateEncoded = encodeState(state)
+  const url = new URL('/auth/cli', consoleUrl)
   url.searchParams.set('state', stateEncoded)
   return url
 }
```
jasonkuhrt, line 114: @prisma/platform This will require some changes in PDP to correctly read the signature. Though, actually, I am not sure we use the signature at all on login. @sampolahtinen ?

Jolg42:
@jasonkuhrt For info, all tests are failing which is a big blocker for this PR, we can't merge without fixing this.
See for example https://github.com/prisma/prisma/actions/runs/7922500498/job/21630227904?pr=23090#step:4:285
```
Error: s/cli build: ✘ [ERROR] Could not resolve "./platform/*Platform"
packages/cli build:     src/bin.ts:30:25:
packages/cli build:       30 │ import { Platform } from './platform/*Platform'
packages/cli build:          ╵                          ~~~~~~~~~~~~~~~~~~~~~~
packages/cli build: Failed
```

I'll review the code now

packages/cli/src/platform/auth/login.ts
sampolahtinen: > not sure we use the signature at all on login. @sampolahtinen ?

Previously, we used signature only on signups.

packages/cli/package.json
```diff
@@ -116,7 +116,7 @@
     "zx": "7.2.3"
   },
   "scripts": {
-    "prisma": "ts-node src/bin.ts",
+    "prisma": "tsx src/bin.ts",
```
sampolahtinen, line 119: @Jolg42 is this change ok? 

Without this change, one enters ESM rabbit hole.

packages/cli/src/platform/_lib/userAgent.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { version as PRISMA_CLI_VERSION } from '../../../package.json'
+import { unknownToError } from './prelude'
+
+const debug = Debug('prisma:cli:platform:_lib:userAgent')
+
+export const getUserAgent = async () => {
+  const signature = await Checkpoint.getSignature().catch(unknownToError)
+  if (isError(signature)) debug(`await checkpoint.getSignature() failed silently with ${signature.message}`) // prettier-ignore
+  return `${isError(signature) ? 'error' : signature}.${PRISMA_CLI_VERSION}.cli.prisma`
```
sampolahtinen, line 13: I think in case of signature error we should default the user-agent to something else than `error`?

packages/cli/package.json
Jolg42: We already use tsx already, we're very happy with it!

Jolg42: Anyway, this is only an internal change for a convenience script

packages/cli/src/bin.ts
```diff
@@ -27,7 +27,7 @@ import { DebugInfo } from './DebugInfo'
 import { Format } from './Format'
 import { Generate } from './Generate'
 import { Init } from './Init'
-import { Platform } from './platform'
+import { Platform } from './platform/*Platform'
```
Jolg42, line 30: This is what is breaking all the tests I guess

Jolg42 (commented):
Note
- make sure to update the test in https://github.com/prisma/prisma/blob/feat/cli/next-version/packages/cli/src/platform/%24.test.ts because I don't think it's passing at the moment

packages/cli/src/platform/accelerate/enable.ts
```diff
@@ -17,67 +12,112 @@ export class Enable implements Command {
...
+          serviceTokenCreateInput: { environmentId },
+        },
+      },
+    })
+
+    const gettingStartedUrl = link('https://pris.ly/d/accelerate-getting-started')
+
+    if (serviceTokenCreate) {
+      return messages.success(
+        `Accelerate enabled. Use this generated API key in your Accelerate connection string to authenticate requests:\n` +
```
Jolg42, line 110:
```suggestion
        `Accelerate enabled. Use this Accelerate connection string to authenticate requests:\n` +
```

Jolg42: From what I see `generateConnectionString` actually returns the connection string and not just the API key, so I think it makes sense to update the message

packages/cli/src/platform/_lib/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  },
+  serviceToken: {
+    '--token': String,
+    '--serviceToken': String,
+    '-s': '--serviceToken',
+  },
+} as const
+
+export const ErrorPlatformUnauthorized = new Error(
+  `No platform credentials found. Run ${green(getCommandWithExecutor('prisma platform login'))} first. Alternatively you can provide a token via the \`--token\` or \`-t\` parameters, or set the 'PRISMA_TOKEN' environment variable with a token.`, // prettier-ignore
```
Jolg42, line 35:
```suggestion
  `No platform credentials found. Run ${green(getCommandWithExecutor('prisma platform auth login --early-access'))} first. Alternatively you can provide a token via the \`--token\` or \`-t\` parameters, or set the 'PRISMA_TOKEN' environment variable with a token.`, // prettier-ignore
```
Is it actually `auth login`?
I think it needs the `--early-access` flag too

packages/cli/src/platform/_lib/userAgent.ts
jasonkuhrt: Well it's error for the signature, not the whole UA. Any other proposal? I think error for the signature calculation is a reasonable fallback. It could also be `unknown` but then we're just bike shedding, not really changing the overall approach. WDYT?

Jolg42:
@jasonkuhrt it looks like something went wrong with your git changes.
Screenshot 2024-02-16 at 15 34 51
And GitHub Actions are not running as a result, which is expected when there is a conflict.

I can recommend using the simple "update branch" button for syncing `main`:
Screenshot 2024-02-16 at 15 35 14

packages/cli/src/platform/_lib/userAgent.ts
sampolahtinen: Oh! I actually misread the code.

I think `unknown` makes sense, so in the case of a signature error the user-agent would be `unknown@1.2.3`

jasonkuhrt: @sampolahtinen Ok agreed regarding `unknown` instead of `error`. I don't think `<a>@<b>` captures enough information for us. I would prefer to use a domain name syntax? Any reason not to? My thinking is order of specificity: `<sig>.<ver>.cli.prisma`. Given `sig` is a different category of identifier, perhaps it makes sense to use `<ver>.cli.prisma@<sig>` where from a machine (the sig) different things could be executed from it.

jasonkuhrt:
@Jolg42 meetings done and I am now focused on this PR!

packages/cli/src/bin.ts
jasonkuhrt: Surprised. I am not aware of file name rules. Trying different things now. Whatever tools are being broken by this are the tools' fault AFAICT. Anyways we don't need to enter this rabbit hole. If `*` is the issue it shall be removed at present.

Jolg42: Not the best time to fix the tooling indeed

jasonkuhrt: Working with `_` :)

jasonkuhrt:
Investigating `CI / Tests / CLI commands (library, ubuntu-latest, 20) (pull_request) Failing after 1m` now.

packages/cli/src/platform/_lib/userAgent.ts
jasonkuhrt: Agreed on a format in Slack. Realized in [`e97b0ab` (#23090)](https://github.com/prisma/prisma/pull/23090/commits/e97b0abf47e8b2e2c953de9d2a807a761f1a49bf).

packages/cli/src/platform/auth/login.ts
jasonkuhrt: We agreed in standup today to use the UA going forward.

packages/cli/src/platform/_lib/utils.ts
```diff
@@ -29,6 +29,10 @@ export const platformParameters = {
     '--serviceToken': String,
     '-s': '--serviceToken',
   },
+  apikey: {
+    '--token': String,
+    '--apikey': String,
```
Jolg42, line 34: I thought `--apikey` was a boolean?
Because if it's a value (and I guess a secret) then it must be added in
- https://github.com/prisma/prisma/blob/main/packages/cli/src/utils/checkpoint.ts#L155
- with an addition to the test here https://github.com/prisma/prisma/blob/main/packages/cli/src/__tests__/checkpoint.test.ts#L119

Jolg42: Same for `--serviceToken`

jasonkuhrt: It is a resource ID, so not a secret.

Jolg42 (approved):
Nice one!

jasonkuhrt:
Thanks @prisma/orm! Will merge as soon as CI is green 🚀 .

sampolahtinen (approved):

packages/cli/src/platform/$.test.ts
```diff
@@ -45,8 +45,10 @@ describe('--early-access flag', () => {
                  auth   Manage authentication with your Prisma Data Platform account
             workspace   Manage workspaces
               project   Manage projects
+          environment   Manage environments
                apikey   Manage API keys
```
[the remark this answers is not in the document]
jasonkuhrt: @janpio correct! @luanvdw can share more but basically dev connections will be ready for that change next release.

PR #23100, opened 2024-02-13
refactor(client): optimized `buffer` polyfill

millsp:
Implements a `Buffer` polyfill with no dependencies, built on top of `Uint8Array`, and ditches the `buffer` polyfill for `/wasm` runtime builds. The implemented polyfill is more modern and weighs `2.36KB` gziped, while `buffer` weighed `8.76KB` gziped.

To ensure that the polyfill works, I copied the relevant tests from Node.js [repository](https://github.com/nodejs/node/tree/main/test) on top of the ones I had written by hand. For full transparency, I have left ❌ marks everywhere where the tests don't pass. 95% of the skipped tests are about errors not happening as usual, as this implementation is more lenient than `Buffer`. The remaining 5% are about old deprecated features, stricter errors in few cases, and a handful of quirks for the `utf16le` encoding. I don't expect you to review the whole test file (6K lines), but rather look at the 45 test skips.

Having such a complete polyfill will prevent any necessary breaking changes in the future (ie. vs plain `Uint8Array`), while still achieving a huge runtime size gain. This PR results in a `7KB` size reduction (gziped).


helpers/compile/plugins/fill-plugin/fillPlugin.ts
```diff
@@ -123,6 +123,12 @@ function onResolve(fillers: Fillers, args: esbuild.OnResolveArgs): esbuild.OnRes
  * @param args from esbuild
  */
 function onLoad(fillers: Fillers, args: esbuild.OnLoadArgs): esbuild.OnLoadResult {
+  if (fillers[args.path] === undefined) {
+    console.log('fillers', fillers[args.path])
+    console.log('args', args)
+    process.exit(1)
+  }
+
```
millsp, line 131:
```suggestion
```

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
```diff
@@ -0,0 +1,504 @@
...
+    for (let i = 0; i < byteLength; i++) {
+      val = val * 0x100 + view.getUint8(i)
+    }
+
+    if (view.getUint8(0) & 0x80) {
+      val -= Math.pow(0x100, byteLength)
+    }
+
+    return val
+  }
```
millsp, line 165: Not written by me, just taken from the Node.js 0.x repository and upgraded with a `DataView`.

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
```diff
@@ -0,0 +1,504 @@
+/* eslint-disable @typescript-eslint/no-unsafe-declaration-merging */
+export interface Buffer {
```
millsp, line 2: The types are manually set here because they are added dynamically and not hand written.

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
```diff
@@ -0,0 +1,504 @@
...
+  | 'base64'
+  | 'ascii'
+  | 'binary'
+  | 'base64url'
+  | 'ucs2'
+  | 'ucs-2'
+  | 'utf16le'
+  | 'utf-16le'
+  | 'latin1'
+  | 'latin-1'
```
millsp, line 455: All of them are supported!

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
```diff
@@ -0,0 +1,504 @@
...
+
+function initReadMethods(prototype: Buffer) {
```
millsp, line 457: This should be called `createDataViewMethods`. It dynamically creates all the methods for the types that you saw on line 2.

package.json
```diff
@@ -47,7 +47,7 @@
     "@typescript-eslint/utils": "7.0.1",
     "arg": "5.0.2",
     "batching-toposort": "1.2.0",
-    "buffer": "6.0.3",
+    "buffer": "^6.0.3",
```
Jolg42, line 50:
```suggestion
    "buffer": "6.0.3",
```
Keep the pin, we can let renovate handle the updates

Jolg42:
>This PR results in a 7KB size reduction.
It looks like it's actually more about 21 but you mean 7KB compressed I guess?

millsp:
> It looks like it's actually more but you mean 7KB compressed I guess?

Yes, uncompressed-minified is `6.8KB` for the new one and `28.4KB` for the old one.

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
```diff
@@ -0,0 +1,504 @@
...
+      for (let i = 0; i < view.byteLength; i += 2) {
+        string += c2s(view.getUint16(i, true))
+      }
+      return string
+    }
+    if (encoding === 'hex') {
+      return this.slice(start, end).reduce((s, v) => s + v.toString(16).padStart(2, '0'), '')
+    }
+
+    bufferPolyfillDoesNotImplement(`encoding "${encoding}"`)
```
SevInf, line 399: Would using `TextEncoder` here make sense? Or is it something `TextEncoder` does not support either?

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
```diff
@@ -0,0 +1,504 @@
...
+  }
+
+  toLocaleString() {
+    return this.toString()
+  }
+}
+
+function stringToBuffer(value: string, encoding: string) {
```
SevInf, line 407: Same question here, would `TextEncoder`/`TextDecoder` would be of any use here.

packages/client/src/__tests__/buffer-small.test.ts
```diff
@@ -0,0 +1,6197 @@
...
+  })
+
+  test('should create a new Buffer from an ArrayBuffer', () => {
+    const arr = new Uint16Array([0x48, 0x65, 0x6c, 0x6c, 0x6f])
+    const buf = Buffer.from(arr.buffer)
+    expect(buf.toString()).toEqual('H\u0000e\u0000l\u0000l\u0000o\u0000')
+  })
+
+  test('should throw an error if input is invalid', () => {
+    // expect(() => Buffer.from(null)).toThrow() ❌
```
SevInf, line 51: Does not seem like adding validation here would have that big of an impact on the size.

packages/client/src/__tests__/buffer-small.test.ts
```diff
@@ -0,0 +1,6197 @@
...
+
+  test('should handle empty buffers', () => {
+    const buf1 = Buffer.from('')
+    const buf2 = Buffer.from('')
+    expect(Buffer.compare(buf1, buf2)).toEqual(0)
+  })
+
+  test('should throw an error if either argument is not a Buffer', () => {
+    const buf = Buffer.from('Hello, World!')
+    // expect(() => Buffer.compare(buf, 'not a buffer' as any)).toThrow() ❌
```
SevInf, line 82: What is the result in the this case?
And pretty much same comment as before, I don't thing adding "is a string, ArrayBuffer view or Buffer" validation would have a big impact on the size, but it would bring polyfill closer to the actual thing

packages/client/src/__tests__/buffer-small.test.ts
```diff
@@ -0,0 +1,6197 @@
...
+  test('should compare specified portions of the buffers', () => {
+    const buf1 = Buffer.from('Hello, World!')
+    const buf2 = Buffer.from('Hello, Universe!')
+    expect(buf1.compare(buf2, 0, 5, 0, 5)).toEqual(0)
+  })
+
+  test('should handle negative and out of bounds values', () => {
+    const buf1 = Buffer.from('Hello, World!')
+    const buf2 = Buffer.from('Hello, Universe!')
+    // expect(() => buf1.compare(buf2, -1)).toThrow() ❌
```
SevInf, line 823: What would happen instead?

packages/client/src/__tests__/buffer-small.test.ts
```diff
@@ -0,0 +1,6197 @@
...
+  test('should write a string to the buffer with specified encoding', () => {
+    const buf = Buffer.allocUnsafe(20)
+    const bytesWritten = buf.write('48656c6c6f2c20576f726c6421', 0, 'hex')
+    expect(buf.toString('utf8', 0, bytesWritten)).toEqual('Hello, World!')
+    expect(bytesWritten).toEqual(13)
+  })
+
+  test('should throw an error if offset is out of bounds', () => {
+    const buf = Buffer.allocUnsafe(20)
+    // expect(() => buf.write('Hello, World!', 21)).toThrow() ❌
```
SevInf, line 930: Same question here, what would happen instead of an error?

packages/client/src/__tests__/buffer-small.test.ts
```diff
@@ -0,0 +1,6197 @@
...
+      assert.strictEqual(c[i], b[i])
+    }
+  }
+
+  {
+    // Floats will be converted to integers via `Math.floor`
+    b.fill(++cntr)
+    c.fill(++cntr)
+    const copied = b.copy(c, 0, 0, 512.5)
+    // assert.strictEqual(copied, 512) ❌
```
SevInf, line 2006: What happens instead?

SevInf (changes requested):
I feel like leaving out bounds validations here would be wrong here — we are leaving our user potentially silently swallowed  errors in this case, unlike real buffer or old polyfill. 

I also think it would be worth adding a test for interactions between built-in buffer and polyfilled one - for the cases WASM engine is used in node or `Buffer` from `node_compat` is used alongside our polyfill. For example, `Buffer.from(polyfilledBuffer)` should work and so on.

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
```diff
@@ -0,0 +1,704 @@
...
+  }
+
+  reverse() {
+    super.reverse()
+
+    return this
+  }
+
+  readIntBE(offset: number, byteLength: number) {
+    assertNumber(offset, 'offset')
```
SevInf, line 182: Does not need to change, but do you thing we can get a little bit of size win if we pull all those strings into constants? Var name can be than minified.

SevInf (approved):
Looks good to me now, I am ok with remaining failing tests. Only one small optional comment that would not block the merge.

helpers/compile/plugins/fill-plugin/fillers/buffer-small.ts
millsp: I checked and turns out not so much: Since the gzip size is what matters and these are put into a dictionary, putting them into a variable produced minimal gains unfortunately.

PR #23114, opened 2024-02-13
fix(adapter-*): use Uint8Array

millsp:


SevInf (changes requested):
Unfortinately, this code was added recently by me in #23047 to work around pg bug (https://github.com/brianc/node-postgres/pull/2930). It got fixed, but not released yet. Without that coercion, lists of bytes will not be encoded correctly when passed to the DB (we probably need test for that on the client side).

When the fix is available in `pg`, we won't need `fixArrayBufferValues` function at all.

Note: `pg` will convert all typed arrays to Buffers anyway, so for them we might actually suggest to go `node_compat` mode even after the fix is released.

Issue #23153, opened 2024-02-15
`runtimeDescription` is not defined error after upgrading to 5.9.1 (from 5.8.1)


david0673:
### Bug description

I get it whenever I try to access `Prisma.sql` which makes sense given that after running `prisma generate` it is generated as:
```typescript
Prisma.sql = () => {
  throw new Error(`sqltag is unable to be run ${runtimeDescription}.
In case this error is unexpected for you, please report it in https://github.com/prisma/prisma/issues`,
)}
```

not to mention that `runtimeDescription` is never being defined to begin with.

### How to reproduce

Afaik:
1. upgrade from 5.8.1 to 5.9.1
2. run `prisma generate`
3. look at `node_modules/.prisma/client/index-browser.js`

### Expected behavior

I'd expect the prisma client to generate normally

### Prisma information

I don't feel comfortably sharing my schema

### Environment & setup

- OS: macOS & ubuntu
- Database: PostgreSQL
- Node.js version: v20

### Prisma Version

v5.9.1

david0673:
https://discord.com/channels/937751382725886062/1206636663284178955

Jolg42:
@david0673 Hi! Thanks for the bug report, I'll fix the error message, but that won't help in your case because it would still error.
Could you help us to figure out how you ended up importing these from index-browser.js?
The best would be to provide a minimal reproduction.
I don't think we would need your schema, but we would need to have your setup and how this is imported.

david0673:
I'm not sure what context you'd like to know about my setup, but here is how I imported Prisma.sql:
```typescript
import { Prisma } from '@prisma/client';
...
  const rows = items.map((item) => {
    const values = order.map((key) => {
      const value = item[key];
      const cast = casts[key];
      if (cast) return cast(value) || Prisma.sql`NULL`;
      return Prisma.sql`${value}`;
    });
    return Prisma.join(values, ', ', '(', ')');
  });

  return Prisma.join(rows, ', ');
...
```

& here is my tsconfig.json
```json
{
  "compilerOptions": {
    "allowJs": true,
    "baseUrl": "src",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "incremental": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "lib": ["dom", "dom.iterable", "esnext"],
    "module": "esnext",
    "moduleResolution": "node",
    "noEmit": true,
    "noErrorTruncation": true,
[...]

PR #20683, opened 2023-08-14
fix(client): use Proxy for "browser" import

Jolg42:
[Internal Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1691661424805119)

closes https://github.com/prisma/prisma/issues/16153

/integration

PR #23164, opened 2024-02-16
fix(client): `runtimeDescription` is not defined #23153

### LINKED ISSUE #23430: `runtimeDescription` is not defined error
### Bug description

Found in https://github.com/prisma/prisma/issues/23153

`runtimeDescription` is not defined error after upgrading to 5.9.1 (from 5.8.1)

Jolg42:
This fixes the state introduced by myself in 5.2.0 in
https://github.com/prisma/prisma/pull/20683

I took this opportunity to add tests for `Prisma.sql` & co.

Contributes to #23153


Jolg42:
Note: I'll update the snapshots after we reviewed the text

packages/client/src/generation/TSClient/common.ts
```diff
@@ -126,7 +126,9 @@ Prisma.NullTypes = {
 export const notSupportOnBrowser = (fnc: string, browser?: boolean) => {
   if (browser)
     return `() => {
-  throw new Error(\`${fnc} is unable to be run \${runtimeDescription}.
+  throw new Error(\`${fnc} was imported from @prisma/client/index-browser.
```
SevInf, line 129: I am a little bit confused about the new message.

Bundlers wold route `@prisma/client` imports to `@prisma/client/index-browser` automatically if they are building frontend build. So, from users standpoint, they already do what the error message suggesting and would probably be confused even more.

SevInf (approved):

PR #23191, opened 2024-02-19
feat(cli): delete management token on logout

jasonkuhrt:
This PR makes the CLI invalidate a management token when the user logs out.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -117,6 +117,8 @@
   },
   "scripts": {
     "prisma": "tsx src/bin.ts",
+    "platform": "tsx src/bin.ts platform --early-access",
+    "pm": "tsx src/bin.ts platform --early-access",
     "dev": "DEV=true tsx helpers/build.ts",
     "build": "tsx helpers/build.ts",
     "test": "dotenv -e ../../.db.env -- tsx helpers/run-tests.ts",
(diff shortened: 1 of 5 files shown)


packages/cli/package.json
```diff
@@ -126,7 +128,8 @@
     "prepublishOnly": "pnpm run build"
   },
   "dependencies": {
-    "@prisma/engines": "workspace:*"
+    "@prisma/engines": "workspace:*",
+    "jose": "^5.2.2"
```
janpio, line 132: Is this in the correct place? All the other things seem to be in `devDep`.

jasonkuhrt: Its not that complex, let's inline it.

https://github.com/panva/jose/blob/05c4351be3a356da2a0e882fbbd8006f2725ec7b/src/util/decode_jwt.ts#L22

tatianainama (approved):

jasonkuhrt:
Refactored, looking good:

CleanShot 2024-02-19 at 12 35 37@2x

packages/cli/package.json
millsp: If you move it to `devDepenbdencies` we will bundle it automatically. And if it has an ESM export then we would also tree-shake the unused imports. If it's small enough you don't need to worry about inlining.

millsp (approved):

PR #23243, opened 2024-02-22
fix(getPlatform): Stop asking people to open issues for Linux distros

SevInf:
`debian` fallback works in all of those cases.


miguelff (approved):
God bless you Serhii 🙌🏽😅

Jolg42:
So many issues we can close now 🎊

PR #23251, opened 2024-02-23
Replace Neon Serverless Driver with @vercel/postgres

janpio:


packages/bundle-size/da-workers-neon/index.js
```diff
@@ -1,12 +1,12 @@
-import { Pool } from '@neondatabase/serverless'
 import { PrismaNeon } from '@prisma/adapter-neon'
+import { createPool } from '@vercel/postgres'
```
millsp, line 2:
```suggestion
import { createPool } from '@vercel/postgres/index.cjs'
```

PR #23260, opened 2024-02-23
test(client): Add e2e test for "Too many connections" error (MySQL)

SevInf:
Core feature that is added here: e2e tests now can run custom services
alongisde main test container. For example, databases.
This is done by placing `docker-compose.yaml` file into the test
directory and specifying link and dependency between `test-e2e` service
and any custom service.
In this PR, this functionality is used to add a test for "Too many
connections" that uses custom-configured MySQL instance with low max
connections number.

Close prisma/team-orm#955


packages/client/tests/e2e/_utils/run.ts
```diff
@@ -97,13 +98,28 @@ async function main() {
...
+    )
+    if (existsSync(localComposePath)) {
+      composeFileArgs.push('-f', localComposePath)
+    }
 
-  const dockerJobs = e2eTestNames.map((path) => {
+    const projectName = testPath.replaceAll('/', '-')
     return async () =>
-      await $`docker run --rm ${dockerVolumeArgs.split(' ')} -e "NAME=${path}" prisma-e2e-test-runner`.nothrow()
+      await $`docker compose ${composeFileArgs} -p ${projectName} run --rm ${dockerVolumeArgs} -e "NAME=${testPath}" test-e2e`.nothrow()
```
millsp, line 122: Clever!

millsp (approved):
Nice! (btw: I feel that someday we should document this custom test runner :smile:)

PR #23366, opened 2024-03-05
fix(adapter-pg*): fix import & test pipeline 

jkomyno:
This PR:
- reverts https://github.com/prisma/prisma/pull/23311
- introduces https://github.com/prisma/prisma/pull/23352 in its place (which is [already known to work on ecosystem-tests](https://github.com/prisma/prisma/pull/23352#issuecomment-1978554531))

PR #23399, opened 2024-03-07
fix(adapter-pg): Continue to try to fix imports

SevInf:
Follow up to #23366 

/integration


Jolg42 (approved):

Jolg42:
Let's see if all tests pass everywhere before merging 😄

Jolg42:
Looks like it's on its way to `5.11.0-integration-fix-adapter-pg-again.3`

PR #23411, opened 2024-03-08
feat(adapters): cloudflare-d1 - use `.raw({ columnNames: true})`

Druue:
Now using new `.raw({ columnNames: true })`. This means we lose the metrics we were retrieving from using `.all()`, however, we currently only use those for `executeRaw` so I added a hatch when running from execute raw to use `.run()` instead which only gives metrics

closes https://github.com/prisma/team-orm/issues/991

### DIFF
--- a/packages/adapter-d1/src/d1.ts
+++ b/packages/adapter-d1/src/d1.ts
@@ -58,6 +58,10 @@ class D1Queryable<ClientT extends StdClient> implements Queryable {
 
     return {
       columnNames,
+      // * Note: without Object.values the array looks like
+      // * columnTypes: [ id: 128 ],
+      // * and errors with:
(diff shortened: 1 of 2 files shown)


packages/adapter-d1/src/d1.ts
```diff
@@ -35,24 +36,25 @@ class D1Queryable<ClientT extends StdClient> implements Queryable {
...
         columnNames: [],
         columnTypes: [],
         rows: [],
       }
     }
 
-    const results = ioResult.results as Object[]
-    const columnNames = Object.keys(results[0])
+    const results = ioResult[1]
+    debug(results)
```
Jolg42, line 54: TODO: remove debug

Jolg42 (approved):

PR #23487, opened 2024-03-14
feat(client): hard error if iTx is used with adapter-d1

Jolg42:
This adds `adapterName` and the corresponding value from the `package.json` (the package name).

I thought this would be more useful in the current state, as this could be used for debugging (and more? ... share your other ideas)
We discussed previously with Alberto about a similar thing, a "tag", while it could be useful too, we don't have the need for this yet, only to get this error out.
So I opted for the most useful option, let's see how things go in the future.

Closes https://github.com/prisma/team-orm/issues/1010


jkomyno (approved):

jkomyno:
Nice cleanup!

PR #23488, opened 2024-03-14
fix(adapter-pg, adapter-neon): Correctly handle non-postgres errors with `code`

### LINKED ISSUE #23390: `node-postgres` (pg) errors with misleading `P2010 PrismaClientKnownRequestError` when using `@prisma/adapter-pg` with SSL (`?sslmode=require`)
### Bug description

I'm using:

* Supabase with Enforced SSL
* `pg` as the driver adapter
* `?sslmode=require` in the DB connection URL

And seeing this error:

```
Invalid `prisma.todo.findMany()` invocation:

Raw query failed. Code: `InvalidArg`. Message: `invalid type: unit value, expected a string`
 code: "P2010"
```

### How to reproduce

Follow these instructions https://www.prisma.io/docs/orm/overview/databases/postgresql#using-the-node-postgres-driver and use the `sslmode=require` connection argument

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
  url      =  env("DATABASE_URL")
}

model Todo {
  @@map("todos")

  task      String   @db.VarChar(255)
  isCompleted Boolean  @default(false)
}
```

```ts
import { Pool } from 'pg'

const connectionString = `${process.env.DATABASE_URL}`

const pool = new Pool({ connectionString })
const adapter = new PrismaPg(pool)
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL 15.1
- Node.js version: v20.8.0

### Prisma Version

```
5.9.1
```

SevInf:
In our postgres driver error handling code, we assumed that if exception
has a `code` property, it is one of the standard postgres codes and
handled it accordingly. However, `error.code` is widely used for other
even outside of PG errors and we handled this case incorrectly.
Particulary, our rust handling code expects `code`, `severity` and
`message` to always be set.
This PR fixes that by checking all of 3 properties. If either of them is
missing or invalid we keep original JS error intact.

Close #23390

### DIFF
--- a/packages/adapter-neon/src/neon.ts
+++ b/packages/adapter-neon/src/neon.ts
@@ -86,7 +86,7 @@ class NeonWsQueryable<ClientT extends neon.Pool | neon.PoolClient> extends NeonQ
       return ok(await this.client.query({ text: sql, values: fixArrayBufferValues(values), rowMode: 'array' }))
     } catch (e) {
       debug('Error in performIO: %O', e)
-      if (e && e.code) {
+      if (e && typeof e.code === 'string' && typeof e.severity === 'string' && typeof e.message === 'string') {
         return err({
           kind: 'Postgres',
           code: e.code,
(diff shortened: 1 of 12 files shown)


Jolg42 (approved):
Nice! Great that we have a test for this now, well done!

PR #23491, opened 2024-03-14
chore(d1): update README for Preview

Jolg42:
Closes https://github.com/prisma/team-orm/issues/1011


Jolg42:
We can probably update this after https://github.com/prisma/prisma/pull/23662

PR #23504, opened 2024-03-15
fix(adapter-d1): map more SQLite errors

Jolg42:
Found while pairing @jkomyno 

This adds the mapping for 
```
(tested in prisma-engines only)
D1_ERROR: NOT NULL constraint failed
-> extendedCode = 1299

(probably not tested)
D1_ERROR: CHECK constraint failed
-> extendedCode = 1811

(probably tested in engines?)
D1_ERROR: PRIMARY KEY constraint failed
-> extendedCode = 1555
```

### DIFF
--- a/packages/adapter-d1/src/d1.ts
+++ b/packages/adapter-d1/src/d1.ts
@@ -93,8 +93,8 @@ class D1Queryable<ClientT extends StdClient> implements Queryable {
         return ok([columnNames, rows])
       }
     } catch (e) {
-      console.error('Error in performIO: %O', error)
-      const { message } = error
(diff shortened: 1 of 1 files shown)


Jolg42:
For context: note how the error is a string in
https://github.com/cloudflare/miniflare/blob/c2ed3afdc1fed9f78d5cb6c50edc793a9a43a850/packages/d1/src/d1js.ts#L22-L24
https://github.com/cloudflare/workerd/blob/b165f80ac38d441ef9e357057853a4a1fcfa113e/src/cloudflare/internal/d1-api.ts#L140-L142

jkomyno (approved):

PR #23530, opened 2024-03-18
test(client): use `Providers` instead of strings

Jolg42:
It's not an exhaustive cleanup, but a good start.

### DIFF
--- a/packages/client/tests/functional/0-legacy-ports/malformed-id/tests.ts
+++ b/packages/client/tests/functional/0-legacy-ports/malformed-id/tests.ts
@@ -25,7 +25,7 @@ testMatrix.setupTestSuite(
         /client/tests/functional/0-legacy-ports/malformed-id/tests.ts:0:0
 
            XX testMatrix.setupTestSuite(
-           XX   () => {
+          XX   () => {
           XX     test('should throw Malformed ObjectID error: in 2 different fields', async () => {
         → XX       const user = prisma.user.create(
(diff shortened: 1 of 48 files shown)


packages/client/tests/functional/0-legacy-ports/malformed-id/tests.ts
```diff
@@ -24,7 +25,7 @@ testMatrix.setupTestSuite(
         /client/tests/functional/0-legacy-ports/malformed-id/tests.ts:0:0
 
            XX testMatrix.setupTestSuite(
-           XX   () => {
+          XX   () => {
```
Druue, line 28: random removed white-space 🤔 probably not related to the PR itself?

Druue (approved):

packages/client/tests/functional/0-legacy-ports/malformed-id/tests.ts
Jolg42: It is related and needed here actually.

Jolg42: But it's really annoying ...

SevInf:
Could you also adjust [template](https://github.com/prisma/prisma/blob/b333dcd4979d65dc34036d07f998566fb6f520c9/packages/client/helpers/new-test/new-test.ts#L62) for new tests?

Jolg42:
Good idea @SevInf - I think you are the only one using it? 😄 😉 

I think we could merge this without this change though.

PR #23544, opened 2024-03-19
ci(adapter-*): run relationMode tests in CI

Jolg42:
Related to https://github.com/prisma/team-orm/issues/1050#issuecomment-2007780481

Will not get merged soon, we just want to take a look at the results for now.


janpio:
Why was this marked as `Preview` in the board @Jolg42?

Jolg42:
We discussed on Slack that having a rough idea of what's happening here would be good to know.

But this can be put to later, that's ok.

PR #23637, opened 2024-03-27
feat: React Native support

ospfranco:


packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
```diff
@@ -559,6 +559,11 @@ export class PrismaClient<
    */
   $runCommandRaw(command: Prisma.InputJsonObject): Prisma.PrismaPromise<Prisma.JsonObject>
 
+
+  /**
+   * Tries to apply pending migrations to the database.
+   */
+  $applyPendingMigrations(): Prisma.PrismaPromise<void>
```
SevInf, line 566: Rather than a new method in a public API, could client just automatically do it on startup if it is built for React Native?

packages/client/package.json
```diff
@@ -141,6 +141,8 @@
     "wasm.d.ts",
     "index.js",
     "index.d.ts",
+    "rn.js",
```
SevInf, line 144: Lets rename all instances of `rn` to `react-native`, here and everywhere.

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -307,6 +307,22 @@ function runCommandRawDefinition(this: PrismaClientClass) {
   return ts.stringify(method, { indentLevel: 1, newLine: 'leading' })
 }
 
+function applyPendingMigrationsDefinition(this: PrismaClientClass) {
```
SevInf, line 310: See suggestion above about migrations being automatically applied on startup rather than requiring explicit `applyPendingMigrations` call from the user. Rest of the comments on this function apply only if we can't/do not want to do this.

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -307,6 +307,22 @@ function runCommandRawDefinition(this: PrismaClientClass) {
...
+  const method = ts.method('$applyPendingMigrations').setReturnType(ts.prismaPromise(ts.voidType))
```
SevInf, line 311: In case we want to keep explicit method:

```suggestion
  const method = ts.method('$applyPendingMigrations').setReturnType(ts.promise(ts.voidType))
```

`PrismaPromise` is a special thing that among other things means it can be passed to `$transaction`, which is not the case for `applyPendingMigrations`

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -307,6 +307,22 @@ function runCommandRawDefinition(this: PrismaClientClass) {
...
+
+  if (this.runtimeNameTs !== 'rn') {
+    method.setDocComment(ts.docComment`
+  Tries to apply pending migrations to the database.
+  `)
+  } else {
```
SevInf, line 317: Rather than telling people not to use and hoping they'll read a doc let's not generate the types for this method at all if it is not react-native build.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -35,9 +35,9 @@ import { PrismaClientClass } from './PrismaClient'
 
 export type TSClientOptions = O.Required<GenerateClientOptions, 'runtimeBase'> & {
   /** More granular way to define JS runtime name */
-  runtimeNameJs: 'binary' | 'library' | 'wasm' | 'edge' | 'edge-esm' | 'index-browser' | String
```
SevInf, line 38: Same comment as above, let's rename `rn` to `react-native`

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -123,6 +132,26 @@ export class LibraryEngine implements Engine<undefined> {
     }
   }
 
+  async applyPendingMigrations(): Promise<void> {
+    if (TARGET_BUILD_TYPE === 'rn') {
+      await this.start()
+      // @ts-ignore
```
SevInf, line 138: Please, create global `__PrismaProxy` definition rather than `@ts-ignoring` every usage.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -151,11 +180,26 @@ export class LibraryEngine implements Engine<undefined> {
         isolation_level: arg.isolationLevel,
       })
 
-      result = await this.engine?.startTransaction(jsonOptions, headerStr)
+      if (TARGET_BUILD_TYPE === 'rn') {
```
SevInf, line 183: This conditionals scattered all over the class look brittle and hard to maintain, let's abstract them out.

We already have an abstraction for using different kind of libraries within `LibraryEngine` and this is `libraryLoader`/ `Library`. What I would suggest:

- We'll write an adapter that would wrap `__PrismaProxy` and make it conform to `Library` interface
- We implement `libraryLoader` for react native. 

If we do that, we can revert most of the `LibraryEngine` to it's previous state and concentrate all the difference in different `Library` implementations. This is how NAPI/WASM library abstraction is implemented and I'd prefer React Native to work the same way too.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -123,6 +132,26 @@ export class LibraryEngine implements Engine<undefined> {
...
+    } else {
+      throw new Error('Cannot call this method from this type of engine instance')
+    }
+  }
+
+  async pushSchema(): Promise<void> {
+    if (TARGET_BUILD_TYPE === 'rn') {
+      await this.start()
+      // @ts-ignore
+      __PrismaProxy.pushSchema(this.engine, this.datamodel)
```
SevInf, line 149: Why do we need separate `pushSchema` call? We already passing the `datamodel` on to `__PrismaProxy.create`

SevInf (changes requested):
Nice job overall, besides the tests it will need a couple of changes before we can merge it. In order of importance:

- Using existing abstractions for React Native library
- Can `applyPendingMigrations` be done implicitly, rather than being public API? If not, couple of type generation things need to change.
- `react-native` name, rather than `rn`

Details are in specific comments. We also might need a product decision on whether or not React Native should be hidden under preview feature, I'll talk to product people.

packages/client/helpers/build.ts
```diff
@@ -143,6 +143,30 @@ const wasmRuntimeBuildConfig: BuildOptions = {
...
+  target: 'ES2022',
+  entryPoints: ['src/runtime/index.ts'],
+  outfile: 'runtime/rn',
+  bundle: true,
+  minify: false,
+  sourcemap: 'linked',
+  legalComments: 'none',
+  emitTypes: false,
+  define: {
+    ...commonEdgeWasmRuntimeBuildConfig.define,
```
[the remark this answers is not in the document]
ospfranco: I think we tried that with Pierre, I will give it a try

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
ospfranco: We talked about this with Soren, and unlike the server version, this process needs to be throwabl, catchable and automatable in case a migration fails since you cannot revert or manually change the databases on the user phones once something has been deployed.

By having it separated it makes this process more explicit and less prone to errors.

packages/client/src/generation/TSClient/PrismaClient.ts
ospfranco: I tried that, but there was an error and Pierre said you guys needed to merge some changes first. I will give it another go. Is there some sort of target I could access to avoid generating the types?

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
SevInf: Ok, then let's make sure those types won't be generated for non-RN clients

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
ospfranco: The __PrismaProxy is defined in `react-native-prisma`. I can create a type, but that would mean they both need to be kept in Sync whenever the C/Rist bindings change. I can do it, just want to confirm you want to take this route.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
ospfranco: This is outdated, I removed the push schema call when I removed the dependency on schema-core.

Just as documentation, this was posible because schema-core was included and that used to contain the push function, which would be used for initializing the db. Now the applyMigrations step is always required. For now there are no plans to bring back pushSchema.

I will take care of this.

packages/client/helpers/build.ts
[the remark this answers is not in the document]
ospfranco: works

packages/client/helpers/build.ts
```diff
@@ -143,6 +143,30 @@ const wasmRuntimeBuildConfig: BuildOptions = {
...
+    TARGET_BUILD_TYPE: '"rn"',
+  },
+  plugins: [
+    fillPlugin({
+      fillerOverrides: { ...commonEdgeWasmFillerOverrides },
```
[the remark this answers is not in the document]
ospfranco: renamed it to commonRuntimesOverrides

packages/client/src/generation/TSClient/PrismaClient.ts
ospfranco: Got it to work I believe, please review. Returning null and later filtering the list of methods to generate

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
ospfranco: I've implemented a `reactNativeLibraryLoader`, put the class that wraps the bindings inside, since I didn't know where else to put this. Let me know if this is correct.

ospfranco:
The failing tests seem to be d1, which seem to be heisentests

packages/client/src/runtime/core/engines/library/ReactNativeLibraryLoader.ts
```diff
@@ -0,0 +1,112 @@
...
+
+// unlike other implementations, on react-native the library needs to be loaded
+// before the engine can be created, so this loader just checks the bindings are there
+// and returns a dummy constructor that just wraps the methods so that the libraryEngine remains agnosti
+export const reactNativeLibraryLoader: LibraryLoader = {
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async loadLibrary(config) {
+    if (!__PrismaProxy) {
+      throw new PrismaClientInitializationError(
+        '__PrismaProxy not detected make sure rn bindings are installed',
```
SevInf, line 94:
```suggestion
        '__PrismaProxy not detected make sure React Native bindings are installed',
```

SevInf (approved):
Approved with one small error message correction, great job!

packages/client/package.json
```diff
@@ -174,7 +183,7 @@
     "@prisma/debug": "workspace:*",
     "@prisma/driver-adapter-utils": "workspace:*",
     "@prisma/engines": "workspace:*",
-    "@prisma/engines-version": "5.13.0-2.fc73af6cd8932d6a618a6c74cdaa015542d7adae",
```
SevInf, line 177: TODO: Don't forget to revert this change before merge.

Nik-Novak:
How would I integrate this into my RN app?

vearvip:
How should I use it in React Native

vladimiremi:
How should I use it in React Native?

sorenbs:
Hey there @Nik-Novak, @vearvip and @vladimiremi 👋 

To learn more about Prisma for Expo and React Native, join us at [App.js](https://appjs.co/) in Krakow, or keep an eye on the Prisma Blog end of May.

octalpixel:
> Hey there @Nik-Novak, @vearvip and @vladimiremi 👋
> 
> To learn more about Prisma for Expo and React Native, join us at [App.js](https://appjs.co/) in Krakow, or keep an eye on the Prisma Blog end of May.

Is the Prisma direction heading towards realtime sync as well ? Something to do with Pulse ?

PR #23649, opened 2024-03-28
fix(adapter-d1): Correctly treat CURRENT_TIMESTAMP result as date

SevInf:
Value inserted by `DEFAULT CURRENT_TIMESTAMP` on sqlite is not an iso
string, but rather 'YY-MM-DD HH:MM:SS' string.

Fix prisma/team-orm#1057

### DIFF
--- a/packages/client/helpers/functional-test/run-tests.ts
+++ b/packages/client/helpers/functional-test/run-tests.ts
@@ -154,7 +154,12 @@ async function main(): Promise<number | void> {
     )
 
     if (unknownAdapterProviders.length > 0) {
-      throw new Error(`Unknown adapter providers: ${unknownAdapterProviders.join(', ')}`)
+      const allAdaptersStr = Array.from(allAdapterProviders)
+        .map((provider) => `  - ${provider}`)
+        .join('\n')
+      throw new Error(
+        `Unknown adapter providers: ${unknownAdapterProviders.join(', ')}. Available options:\n${allAdaptersStr}\n\n`,
+      )
     }
 
     if (adapterProviders.some(isDriverAdapterProviderLabel)) {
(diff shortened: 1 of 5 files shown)


packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -154,7 +154,12 @@ async function main(): Promise<number | void> {
     )
 
     if (unknownAdapterProviders.length > 0) {
-      throw new Error(`Unknown adapter providers: ${unknownAdapterProviders.join(', ')}`)
+      const allAdaptersStr = Array.from(allAdapterProviders)
```
SevInf, line 157: Not strictly related to this PR, it just took me a while to figure out that proper `--adapter` argument is `js_d1` so I think it would be quicker if we list all supported options when we see unknown value.

jkomyno (approved):

jkomyno:
> ## [CodSpeed Performance Report](https://codspeed.io/prisma/prisma/branches/fix/d1-default-date)
> ### Merging #23649 will **degrade performances by 21.93%**
> Comparing `fix/d1-default-date` ([499b7e5](https://github.com/prisma/prisma/commit/499b7e55d45bfb872e61691ef020ce9c07ac15e1)) with `main` ([dfac408](https://github.com/prisma/prisma/commit/dfac40844e06dad97c79130ec712fbca41d27764))
> 
> ### Summary
> `❌ 1` regressions `✅ 2` untouched benchmarks
> 
> > ⚠️ _Please fix the performance issues or [acknowledge them on CodSpeed](https://codspeed.io/prisma/prisma/branches/fix/d1-default-date)._
> 
> ### Benchmarks breakdown
> Benchmark	`main`	`fix/d1-default-date`	Change
> ❌	`client generation 100 models with relations`	30.7 s	39.3 s	-21.93%

This doesn't seem accurate at all.

PR #23690, opened 2024-04-02
ci: run "build" before running CLI tests

Jolg42:
fixes
```
 FAIL src/__tests__/commands/Studio.test.ts (12.12 s)
  ● studio with alternative urls and prisma:// › queries work if url is prisma:// and directUrl is set

    expect(received).toMatchSnapshot()

    Snapshot name: `studio with alternative urls and prisma:// queries work if url is prisma:// and directUrl is set 1`

    - Snapshot  -  2
    + Received  + 35

      {
        action: clientRequest,
        channel: -prisma,
        payload: {
    -     data: [],
    -     error: null,
    +     data: null,
    +     error: {
    +       message:
    +     GeneratorError: Generator "/cli/prisma-client/generator-build/index.js" failed:
    +     node:internal/modules/cjs/loader:1031
    +       throw err;
```

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -55,7 +55,7 @@ runs:
           target/release/libquery_engine.so
           target/release/libquery_engine.dylib
           target/release/query_engine.dll
-    - name: Set environment variables for custom engine
+    - name: Set environment variables for custom engines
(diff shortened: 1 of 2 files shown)


Jolg42:
Replaced by https://github.com/prisma/prisma/pull/23691 (waiting for tests to pass to confirm there)

PR #23691, opened 2024-04-02
ci: pin pnpm to 8.15.5

Jolg42:
https://github.com/pnpm/pnpm/releases/tag/v8.15.6 was released 14h ago and tests are failing since.

Re-running old runs results in the same error = CLI Studio tests were passing with 8.15.5 and below.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -30,7 +30,7 @@ runs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ inputs.node-version }}
-        # cache: 'pnpm'
+        cache: 'pnpm'
         # This makes it possible to publish to npm later
         # The NPM_TOKEN secret needs to be defined in the job
         registry-url: 'https://registry.npmjs.org'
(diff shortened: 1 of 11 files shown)


.github/actions/setup/action.yml
```diff
@@ -30,7 +30,7 @@ runs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ inputs.node-version }}
-        cache: 'pnpm'
+        # cache: 'pnpm'
```
Jolg42, line 33:
```suggestion
        cache: 'pnpm'
```

.github/workflows/test-template.yml
```diff
@@ -783,7 +783,7 @@ jobs:
           node-version: ${{ matrix.node }}
           pnpm-version: ${{ inputs.pnpmVersion }}
           engine-hash: ${{ inputs.engineHash }}
-          skip-tsc: true
+          skip-tsc: false
```
Jolg42, line 786:
```suggestion
          skip-tsc: true
```

PR #23715, opened 2024-04-03
feat(migrate diff): add -o --output option

Jolg42:
Using `-o` or `--output` as they are well known options, see https://clig.dev/#arguments-and-flags

Since the Schema engine does not return the content of the diff, I had to intercept it.
To achieve that, it was best, and something we wanted to do a few years ago, to use `process.stdout`, and since many tests rely on the output the git diff is quite large.

Note: I did not want to use our Jest mock for process.stdout because it captures everything, which makes it very ugly.
Example: `DEBUG=prisma*` is captured, making it very hard to debug anything 🙈 

~~TODO: cherry pick this to main later~~

PR #23725, opened 2024-04-04
fix(migrate): fix --from/to-local-d1 & --local-d1 on Windows (#23700) #23725 

Jolg42:
See original PR against a patch branch: https://github.com/prisma/prisma/pull/23715
(cherry picked from commit 2c09e725df635a69c7065deeb0c40d6306f402a8)

### DIFF
--- a/packages/migrate/src/commands/MigrateDiff.ts
+++ b/packages/migrate/src/commands/MigrateDiff.ts
@@ -28,7 +28,7 @@ const helpOptions = format(
 ${bold('Options')}
 
   -h, --help               Display this help message
-  -o, --output             Path to the file where the diff output will be written to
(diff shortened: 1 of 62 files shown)


packages/migrate/src/commands/MigrateDiff.ts
```diff
@@ -26,6 +28,7 @@ const helpOptions = format(
 ${bold('Options')}
 
   -h, --help               Display this help message
+  -o, --output             Path to the file where the diff output will be written to
```
jkomyno, line 31: Small inspiration from `curl -o`'s help message.

```suggestion
  -o, --output             Writes to a file instead of stdout
```

PR #23735, opened 2024-04-04
test: Fix jest snapshot drifts

SevInf:
Our custom serializer returned string as-is. This confused the hell out
of jest's indentation detection algorithm and caused the drift of the
snapshots on every update we all know and love.
Default snapshot serializer wraps strings in quotes and I do the same
for this PR. This seems to fix the drift, for the price of having to
update every single inline snapshot we have.


packages/get-platform/src/test-utils/jestSnapshotSerializer.js
```diff
@@ -120,6 +121,11 @@ function prepareSchemaForSnapshot(str) {
     .join('\n')
 }
 
+// needed for jest to correctly handle indentation on multiline snapshot updates
+function wrapWithQuotes(str) {
```
SevInf, line 125: So here is an actual change, rest is just snapshot updates.

jkomyno:
Looks like a few more snapshot updates are necessary, due to quote symbols in snapshots

Jolg42 (changes requested):
Blocking this for now, as I see it's reverting changes I did in https://github.com/prisma/prisma/pull/23715
Screenshot 2024-04-04 at 19 08 23

Sorry 😄 
Good luck 🤞🏼

SevInf:
@Jolg42 this should be fixed now

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -235,7 +235,9 @@ async function main(): Promise<number | void> {
 
   try {
     if (args['--updateSnapshot']) {
-      const snapshotUpdate = jestCli.withArgs(['-u']).withArgs(args['_'])
+      const snapshotUpdate = jestCli
```
SevInf, line 238: Second meaningful change in the PR: `-u` flag now correctly exludes type checks

SevInf:
ci test all

packages/cli/src/__tests__/incomplete-schemas.test.ts
```diff
@@ -30,71 +30,6 @@ DROP TABLE IF EXISTS 'test-dbexecute';
...
- Prisma CLI Version : 0.0.0
-   `
-
-const aDatasourceBlockIsMissingError = `A datasource block is missing in the Prisma schema file.`
-const thereIsNoDatasourceError = `
-There is no datasource in the schema.
-
-
-`
-
```
Jolg42, line 97: Why did you do these changes?

SevInf: Pulling inline snapshot into separate variable does not make sense - it can't be updated this way.

Jolg42 (approved):

packages/cli/src/__tests__/incomplete-schemas.test.ts
Jolg42: I think it was made like this for a reason though if I remember right, I think it was to make sure the snapshots are the same, this breaks this, though it's not a big deal I guess.

PR #23770, opened 2024-04-08
feat(client): Add `omit` option to the queries

### LINKED ISSUE #5042: Add way to exclude fields via query arguments or globally
## Problem

Sometimes it is helpful to just remove some fields from the selection set like `password` so that we accidentally don't expose them to the frontend.

## Proposals

Adding a `exclude` key in the arguments of the queries to easily exclude fields from the selectionSet just like `include` or `select`.

```js
prisma.user.findMany({
  exclude:{
    password: true
  }
})
```

Or

```js
await prisma.user.findMany({
  omit: {
    password: true,
  },
})
```

## Alternatives

1. Use query and list all the fields you want
1. Schema annotation
2. Manual omission via lodash or something !?

## Related
- https://github.com/prisma/prisma/issues/2170
- https://github.com/prisma/prisma/issues/3246

SevInf:
Counterpart to prisma/prisma-engines#4807

Can be used standalone or combined with `include`, allows to exclude
the fields that normally would be included by default. Available in all
methods that return actual database records.

It is not useable together with `select` and attempt to do so would
cause type check and validation error.

When using together with result extensions, excluded dependency of a
computed field will be queried from a DB, but will not be returned to
the end user, unless computed field is exlucded as well (see "omit
with extensions" tests in this PR). This behaviour is equivalent to what
we do if depenency of a computed field is not mentioned in explicit
`select`.

TODO:
- [x] preview feature
- [x] validation of non-existing fields in exclude

Close prisma/team-orm#1080
#5042

/integration


packages/client/package.json
```diff
@@ -174,7 +174,7 @@
     "@prisma/debug": "workspace:*",
     "@prisma/driver-adapter-utils": "workspace:*",
     "@prisma/engines": "workspace:*",
-    "@prisma/engines-version": "5.13.0-6.dcdb692a8946281c0c85b4a6f7081984eec92b92",
+    "@prisma/engines-version": "5.13.0-3.feat-exclude-selection-e76495825e7475c1f31801a6b5e233caa6edebb8",
```
SevInf, line 177: TODO: revert before merging

packages/client/src/generation/TSClient/Args.ts
```diff
@@ -1,144 +1,118 @@
...
+  getLegacyModelArgName,
+  getModelArgName,
+  getSelectName,
+} from '../utils'
 import { GenerateContext } from './GenerateContext'
 import { getArgFieldJSDoc } from './helpers'
-import { InputField } from './Input'
+import { buildInputField } from './Input'
+
+export class ArgsTypeBuilder {
```
SevInf, line 15: This replaces both `ArgsType` and `MinArgsType` type classes - we now can selectively decide how to build different args types and which fields to include.

packages/client/src/runtime/core/errorRendering/applyValidationError.test.ts
```diff
@@ -622,8 +776,6 @@ describe('UnknownSelectionField', () => {
         include: {
           notThere: true,
           ~~~~~~~~
-      ?   id?: true,
```
SevInf, line 625: Opprtunistic fix: we previously suggested to use scalar fields in `include`. Only relations are valid in include, so those suggestions were incorrect.

SevInf:
Note: tests are failing because https://github.com/prisma/prisma/pull/23776 needs integration first, it should not block the review though.

packages/client/src/runtime/core/types/ValidationError.ts
```diff
@@ -9,10 +9,12 @@ import { EngineValidationError, OutputTypeDescription } from '../engines'
...
+ * Pair of mutually exclusive fields are found on selection (for example select + include or select + exclude)
  */
-export type IncludeAndSelectError = {
-  kind: 'IncludeAndSelect'
+export type MutuallyExclusiveFieldsError = {
+  kind: 'MutuallyExclusiveFields'
+  firstField: string
+  secondField: string
   selectionPath: string[]
 }
```
Druue, line 19: ✨

packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts
```diff
@@ -90,29 +101,50 @@ function serializeFieldSelection(
   { select, include, ...args }: JsArgs = {},
   context: SerializeContext,
 ): JsonFieldSelection {
+  let exclude: Exclusion | undefined
+  if (context.isPreviewFeatureOn('excludeApi')) {
+    exclude = args.exclude
+    delete args.exclude
```
Druue, line 107: TIL about the `delete` keyword, pretty cool :o

packages/client/src/runtime/core/types/exported/Result.ts
```diff
@@ -48,7 +48,16 @@ export type FluentOperation =
...
+export type TrueKeys<T> = {
+  [K in keyof T]: T[K] extends true ? K : never
+}[keyof T]
+
+export type GetFindResult<P extends OperationPayload, A> = A extends { exclude: infer Exclusion }
+  ? Compute<Omit<GetSelectIncludeResult<P, A>, TrueKeys<Exclusion>>>
+  : GetSelectIncludeResult<P, A>
+
+// prettier-ignore
+export type GetSelectIncludeResult<P extends OperationPayload, A> =
```
Druue, line 60: 🤯

Reading through the [infer docs](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#inferring-within-conditional-types) but I really appreciate the explanation of:
```ts
type ElementType<A> = A extends Array<infer X>
    ? x
    : never
```
translating to 
```rs
match A {
    Array(x) => x,
    _ => never,
}
```

packages/client/src/runtime/core/errorRendering/applyValidationError.ts
```diff
@@ -439,6 +486,22 @@ function addSelectionSuggestions(selection: ObjectValue, outputType: OutputTypeD
...
+}
+
+function addExclusionSuggestions(selection: ObjectValue, outputType: OutputTypeDescription) {
+  for (const field of outputType.fields) {
+    if (!selection.hasField(field.name) && !field.isRelation) {
+      selection.addSuggestion(new ObjectFieldSuggestion(field.name, 'true'))
+    }
+  }
+}
+
```
Druue, line 504: Are these for the typescript language-server or am I misunderstanding what `suggestion` means

packages/client/src/generation/ts-builders/TypeBuilder.ts
```diff
@@ -3,7 +3,10 @@ import type { KeyType } from './KeyType'
 import { Writer } from './Writer'
 
 export abstract class TypeBuilder implements BasicBuilder {
+  // TODO: this should be replaced with precedence system that would
+  // automatically add parenthesis where they are needed
```
Druue, line 7:
```suggestion
  // TODO(@SevInf): this should be replaced with precedence system that would
  // automatically add parenthesis where they are needed
```

Druue (commented):
🗒️ to be continued... 👀

packages/client/src/runtime/core/extensions/resultUtils.ts
```diff
@@ -122,3 +122,24 @@ export function applyComputedFieldsToSelection(
   }
   return result
 }
+
+export function applyComputedFieldsToExclusion(
```
Druue, line 126: Add doc comment and explain what it does
MAybe rename to `getEngineExclusion`

packages/client/src/runtime/core/extensions/resultUtils.ts
```diff
@@ -122,3 +122,24 @@ export function applyComputedFieldsToSelection(
...
+  exclusion: Exclusion,
+  computedFields: ComputedFieldsMap | undefined,
+): Selection {
```
Druue, line 129:
```suggestion
): Exclusion {
```

Druue (approved):

packages/client/package.json
```diff
@@ -174,7 +174,7 @@
     "@prisma/debug": "workspace:*",
     "@prisma/driver-adapter-utils": "workspace:*",
     "@prisma/engines": "workspace:*",
-    "@prisma/engines-version": "5.13.0-16.6d35870e9578a3e7406168f61f89f84e78c1b38d",
+    "@prisma/engines-version": "5.13.0-20.feat-exclude-selection-8eed76cebeb3eb147fdfbf46e79211a543caa112",
```
SevInf, line 177: TODO: remvoe before merging

PR #23918, opened 2024-04-22
chore: new promo message in main help

janpio:


SevInf (approved):

Druue:
Cool! :) The indentation looks a little off however? feels like it should be inline with `Flags` and `Examples`

image

janpio:
Decided against that, as the way the help is currently created makes this really hard - so a tradeoff that I think is ok to take right now.

PR #23965, opened 2024-04-25
ci: improve GitHub action workflow

ceddy4395:
Hey! 🙂
I want to contribute the following changes to your workflow:

- Avoid uploading artifacts on forks
- Use 'if' for upload-artifact action
- Avoid uploading artifacts on forks
- Run tests on multiple OS's

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))


Jolg42:
Hi @ceddy4395 

Thanks for your PR 🙌🏼 

I have one question: I'm not sure I understand the reason behind "Run tests on multiple OS's", could you explain why running on `[ubuntu-latest, ubuntu-20.04]` instead of only `ubuntu-latest` would be better?

ceddy4395:
Hi @Jolg42 

The reasoning behind this smell is that during testing, differences in OS's in terms of default installed dependencies etc. should be taken into account. 

I'm interested to hear your thoughts!

Jolg42:
@ceddy4395 Got it.

Could you revert the `runs-on: ${{matrix.os}}` changes? 

Why?

On `main` we have ~ 206 jobs
https://github.com/prisma/prisma/actions/runs/8891585684/usage

With your changes I see ~380 jobs
https://github.com/prisma/prisma/actions/runs/8831591209/usage?pr=23965

So this is a massive increase in the number of jobs, and it looks like it did not catch any bug.
This means every new PR and merge to main will trigger all these additional jobs "for nothing", so I see this as a waste of time and resources. It also increases the number of things a developer needs to look after and probably more false positive as flaky tests might show up more often.
In the end, I don't see these as useful at the moment personally.

Note that we have some specific tests in https://github.com/prisma/ecosystem-tests/ for a lot of things.
Example: the docker tests on different OSes https://github.com/prisma/ecosystem-tests/tree/dev/docker

Jolg42 (approved):

PR #24005, opened 2024-04-29
fix: set specific d1 max bind values

Weakky:
## Overview

Pass max bind values thresold to the Query Engine


jkomyno:
Deprecated by: https://github.com/prisma/prisma/pull/24242

PR #24026, opened 2024-04-30
feat(get-platform): Output detected but unsupported arch in warning

janpio:
Small improvement to warning message to make it easier to know what exaclty went wrong (by including the actual architecture in the message).


Jolg42 (approved):

Jolg42:
Windows tests are now fixed on `main`, let's check that it's true before merging this

Issue #23600, opened 2024-03-23
Next.js app build fails when using Prisma with DB driver in Server Action


nzws:
### Bug description

When building a Next.js application using Prisma on the edge runtime with the Database driver, the build fails with the following error. This problem seems to occur when using Prisma client on the server action imported by the client component. Also, this problem did not occur with Prisma v5.10.2 (in Early Access), and occurred with v5.11.0.

```
Failed to compile.

app/page.js from Terser
  x await isn't allowed in non-async function
       ,-[26410:1]
 26410 | /* harmony export */ __webpack_require__.d(__webpack_exports__, {
 26411 | /* harmony export */   "default": () => (__WEBPACK_DEFAULT_EXPORT__)
 26412 | /* harmony export */ });
 26413 | /* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ((await Promise.resolve(/* import() */).then(__webpack_require__.t.bind(__webpack_require__, 9737, 23))).default);
       :                                                                         ^^^^^^^
 26414 | 
 26415 | /***/ }),
       `----

Caused by:
    0: failed to parse input file
    1: Syntax Error
Error: 
  x await isn't allowed in non-async function
       ,-[26410:1]
 26410 | /* harmony export */ __webpack_require__.d(__webpack_exports__, {
 26411 | /* harmony export */   "default": () => (__WEBPACK_DEFAULT_EXPORT__)
 26412 | /* harmony export */ });
 26413 | /* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ((await Promise.resolve(/* import() */).then(__webpack_require__.t.bind(__webpack_require__, 9737, 23))).default);
       :                                                                         ^^^^^^^
 26414 | 
 26415 | /***/ }),
       `----

Caused by:
    0: failed to parse input file
    1: Syntax Error


> Build failed because of webpack errors
```

### How to reproduce

1. `git clone https://github.com/nzws/prisma-with-nextjs-edge-import-problem-repro`
2. `cd prisma-with-nextjs-edge-import-problem-repro`
3. `yarn install`
4. `yarn build`
5. See error

### Expected behavior

The build passes normally and the application can run.

### Prisma information

I created a repository that can reproduce this issue: https://github.com/nzws/prisma-with-nextjs-edge-import-problem-repro (https://github.com/nzws/prisma-with-nextjs-edge-import-problem-repro/commit/cc513a74a2fc5235409db82e46c51ddf84be4b07)

[...]

PR #23754, opened 2024-04-07
fix(client): replace dynamic import for `wasm-worker-loader` with static export

smorimoto:
Fixes #23600

As I understand, since dynamic import is not necessary here, we can easily fix the linked issue.
Is there anything I am missing?
I'm actually using this patch in my runtime environment and I don't see any performance issues either.

Issue #23929, opened 2024-04-23
Incompatibility with NextJS app dir, CloudFlare Pages and D1


steebchen:
### Bug description

There are issues when deploying NextJS (app dir) with CloudFlare pages and D1.

- warnings pop up in react server components
- server actions do not work at all with this combination

### How to reproduce

Reproduction at https://github.com/steebchen/next-cf-prisma-d1

- fork https://github.com/steebchen/next-cf-prisma-d1 or `git clone git@github.com:steebchen/next-cf-prisma-d1.git`
- `pnpm install`
- adapt your own cloudflare details in wrangler.toml
- create a pages project from git if forked or deploy using CLI if cloned
- run `pnpm dev`
- visit localhost:3000 and create a simple poll
  - warnings will popup
  - when creating the poll using the server action, it breaks

When running `pnpm build`, it results in this:

```
Creating an optimized production build ...
Failed to compile.

edge-chunks/274.js from Terser
  x await isn't allowed in non-async function
     ,-[167:1]
 167 | /* harmony export */ __webpack_require__.d(__webpack_exports__, {
 168 | /* harmony export */   "default": () => (__WEBPACK_DEFAULT_EXPORT__)
 169 | /* harmony export */ });
 170 | /* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ((await Promise.resolve(/* import() */).then(__webpack_require__.t.bind(__webpack_require__, 92[39](https://github.com/steebchen/next-cf-prisma-d1/actions/runs/8805198567/job/24167263548#step:8:40), 23))).default);
     :                                                                         ^^^^^^^
 171 | 
 172 | /***/ }),
     `----

Caused by:
    0: failed to parse input file
    1: Syntax Error
Error: 
  x await isn't allowed in non-async function
     ,-[167:1]
 167 | /* harmony export */ __webpack_require__.d(__webpack_exports__, {
 168 | /* harmony export */   "default": () => (__WEBPACK_DEFAULT_EXPORT__)
 169 | /* harmony export */ });
 170 | /* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = ((await Promise.resolve(/* import() */).then(__webpack_require__.t.bind(__webpack_require__, 9239, 23))).default);
     :                                                                         ^^^^^^^
 171 | 
 172 | /***/ }),
     `----

Caused by:
    0: failed to parse input file
    1: Syntax Error
```

Executing a server action results in the following error:

```
[...]

PR #24029, opened 2024-04-30
feat(client): try CJS-friendly alternative to https://github.com/prisma/prisma/pull/23754

jkomyno:
This PR is an alternative version of https://github.com/prisma/prisma/pull/23754, specifically aimed to guarantee that Prisma Client keeps working when distributed as a CommonJS package.

This PR addresses [this failing e2e test](https://github.com/prisma/prisma/actions/runs/8586883085/job/23555310627?pr=23754), aiming to avoid the esbuild error:

```
[esbuild Error]: Top-level await is currently not supported with the "cjs" output format
    at /usr/local/lib/node_modules/zx/build/vendor.js:2:17
```


jkomyno:
This PR seems to fix https://github.com/prisma/prisma/issues/23600, but not https://github.com/prisma/prisma/issues/23929.

nzws:
I'm not sure if this is the cause of [the error](https://github.com/prisma/prisma/issues/23600#issuecomment-2120636463), but it probably requires a fix in the import destination.

In this file, I think it needs to dynamically import the loader first and then dynamically import the engine.

https://github.com/prisma/prisma/blob/02deb429265bab06b5e6207a26d259f15d0776da/packages/client/src/generation/utils/buildGetQueryEngineWasmModule.ts#L31-L33

```
  getQueryEngineWasmModule: async () => {
    const loader = (await import('#wasm-engine-loader')).default
    const engine = (await loader).default
    return engine
  }
```

jkomyno:
Deprecated by https://github.com/prisma/prisma/pull/24312.

PR #24057, opened 2024-05-01
feat(cli): ground work for prismaSchemaFolder, implementing `validate`, `format`, and adapting `migrate *` + `db *` cli commands 

jkomyno:
This PR contributes to:
- closes: https://github.com/prisma/team-orm/issues/1095 (missing: tests in follow-up PR)
- closes: https://github.com/prisma/team-orm/issues/1096 (missing: tests in follow-up PR)

Most notable changes:
- `getSchemaPath` now considers the relative `./prisma/schema/` path as well, if it contains a *.prisma file with `previewFeatures = [..., "prismaSchemaFolder", ...]`
- `prisma format` now supports multiple schema files
- `prisma validate` now supports multiple schema files
- `getConfig` now supports multiple schema files
- lots of changes in tests


packages/internals/src/cli/getSchema.ts
```diff
@@ -16,38 +29,84 @@ export async function getSchemaPath(
...
+    const schemaPath = await fn(sourcePath!)
+    debug(`${strategy} resolved to ${schemaPath}`)
+
+    if (!schemaPath) {
+      continue
     }
 
-    return customSchemaPath
-  }
+    const schemaPathResult = await Promise.any([
```
SevInf, line 99: Right now, this is undefined behavoir if both file and folder exist: whoever completes first wins and gets used.
Instead we need check if both exist and throw if that's the case.

packages/internals/src/cli/getSchema.ts
```diff
@@ -16,38 +29,84 @@ export async function getSchemaPath(
...
+      // 2. Try the package.json `prisma.schema` custom path
+      { strategy: 'package.json', fn: getSchemaPathFromPackageJson, sourcePath: opts.cwd },
+
+      // 3. Try the conventional ./schema.prisma or ./prisma/schema.prisma paths
+      { strategy: 'relative', fn: getRelativeSchemaPath, sourcePath: opts.cwd },
+
+      // 4. Try resolving yarn workspaces and looking for a schema.prisma file there
+      { strategy: 'yarn', fn: resolveYarnSchema, sourcePath: opts.cwd },
+    ] as const
+  ).filter(({ sourcePath }) => sourcePath !== null)
```
SevInf, line 87:
```suggestion
  ).filter(({ sourcePath }) => sourcePath != null)
```

I think that meant to check for `undefined` as well?

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -164,13 +164,14 @@ See \`${green(getCommandWithExecutor('prisma db execute -h'))}\``,
...
       // throws an error if it doesn't
-      const schemaPath = await getSchemaPath(args['--schema'])
+      const schemaPathResult = await getSchemaPath(args['--schema'])
 
       // Execute command(s) to url from schema
       datasourceType = {
         tag: 'schema',
-        // if schemaPath is undefined, getSchemaPath will error
-        schema: schemaPath!,
+        // TODO: `schema` should have type `Datamodel`, see:
```
SevInf, line 172: I think we wanted to stop using `Datamodel` in reference to prisma schema in a new code.

packages/internals/src/cli/getSchema.ts
jkomyno: Correct: this should run `readSchemaFromMultiFiles` and fall back to `readSchemaFromSingleFile`.

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -164,13 +164,14 @@ See \`${green(getCommandWithExecutor('prisma db execute -h'))}\``,
     else {
       // validate that schema file exists
       // throws an error if it doesn't
```
jkomyno, line 166: This comment is wrong.

packages/migrate/src/commands/DbExecute.ts
Jolg42: Oh yes, that would be nice to get rid of the old "datamodel" mentions

packages/cli/src/Studio.ts
```diff
@@ -97,20 +96,19 @@ ${bold('Examples')}
...
     const staticAssetDir = path.resolve(__dirname, '../build/public')
 
-    const schema = await fs.promises.readFile(schemaPath, 'utf-8')
-    const config = await getConfig({ datamodel: schema, ignoreEnvVarErrors: true })
+    const config = await getConfig({ datamodel: schemas, ignoreEnvVarErrors: true })
 
     process.env.PRISMA_DISABLE_WARNINGS = 'true' // disable client warnings
     const studio = new StudioServer({
-      schemaPath,
+      schemaPath: path.resolve(process.cwd(), schemaPath),
```
SevInf, line 111: I think we don't need this anymore since `schemaPath` is now always absolute.
```suggestion
      schemaPath
```

packages/cli/src/__tests__/artificial-panic.test.ts
```diff
@@ -83,9 +83,20 @@ describe('artificial-panic formatter', () => {
       expect(e.rustStack).toBeTruthy()
       expect(e.schemaPath.replace(/\\/g, '/')) // replace due to Windows CI
         .toContain('prisma/schema.prisma')
-      expect(e).toMatchObject({
-        schema: undefined,
-      })
+      expect(e.schema).toMatchInlineSnapshot(`
```
SevInf, line 86: I know we talked about it yesterday, but could you remind me why this changed?

packages/internals/src/__tests__/engine-commands/validate.test.ts
```diff
@@ -172,25 +176,21 @@ describe('validate', () => {
...
           // @ts-expect-error
-          validate({ datamodel: true })
+          validate({ schemas: true })
         } catch (e) {
-          expect(isRustPanic(e)).toBe(true)
-          expect(serialize(e.message)).toMatchInlineSnapshot(`
-            ""RuntimeError: panicked at prisma-fmt/src/validate.rs:0:0:
-            Failed to deserialize ValidateParams: data did not match any variant of untagged enum SchemaFileInput at line 1 column 20""
-          `)
-          expect(e.rustStack).toBeTruthy()
```
SevInf, line 187: Where did this go?

SevInf (changes requested):
That was quite an undertaking, great job!.

Most of the comments I left are small, however there is one thing: `validate` and format now can not handle invalid multi-file schema - it just falls back to "file not found" message:

image


I feel like we should fix this and display proper validation messages in this case before we can consider `validate` issue done. I am fine with doing it in follow-up PR, in that case just let me know and remove `closes` text from this PR.

packages/cli/src/__tests__/artificial-panic.test.ts
jkomyno: Previously, `formatSchema` had two mutually exclusive input arguments, which would also appear in the `RustPanic` object: `schema` and `schemaPath`. Now, we pass both as input arguments in the form of an array of tuples.

SevInf: Got it, thank you!

SevInf (approved):
Migrate snapshots need updating now, but other than that, I think we are good to go!

Issue #8131, opened 2021-07-07
`createMany()` should return the created records


arcticmatt:
Initially brought up here: https://github.com/prisma/prisma/issues/4998#issuecomment-787206842

## Problem

The fact that the `createMany` API does not return the created records (which should be possible for Postgres) makes it difficult/impossible to use in some circumstances, meaning we must fall back to raw SQL. Here's an example.

Let's say I have these two tables

```
CREATE TABLE IF NOT EXISTS post (
  "id" SERIAL PRIMARY KEY,
  ...
);

CREATE TABLE IF NOT EXISTS comment (
  "id" SERIAL PRIMARY KEY,
  "postId" INTEGER NOT NULL REFERENCES post ("id") ON DELETE CASCADE,
  ...
);
```

Let's say I want to create many posts and many comments. The data to be created could be represented like so: `Array<{post: PostData, comments: Array<CommentData>}>`. If `createMany` returned the created data, I could do something like this:

```
const inputs: Array<{post: PostData, comments: Array<CommentData>}> = [...];
const posts = prisma.post.createMany({data: [...]});
const commentCreateData = inputs.map((input, index) => input.comments.map((commentData) => ...))).flat();
prisma.comment.createMany({data: commentCreateData});
```

However, since `createMany` does not return the created data, it is difficult to create many posts and then create many comments linked to those posts in an efficient manner. The remaining options are to create records one-by-one or use raw SQL. Using a combination of `createMany` and `findMany` is possible if you manually specify IDs, but does not work if you rely on Postgres (or some other DB) to generate IDs by default. 

## Suggested solution

`createMany` should return an array of created records ([like how `create` returns the single created record](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#create)).

## Alternatives

* Use `create` (inefficient if creating many records)
* Use raw SQL (efficient but inconvenient)
* Use `createMany` + `findMany` (only works if you manually specify IDs, and also less efficient than if `createMany` returned created records)

## Additional context

Personally, I've used raw SQL as a workaround since it's the most efficient alternative.

timvandam:
This would be great to have.

From what I've gathered, Postgres, SQLite and MariaDB support `INSERT RETURNING` which could be used for this, but MySQL does not.

I suppose this would open the door to also supporting nested writes inside `createMany`

haisapan:
great if have this feature

joeyfigaro:
+1 would love support for this.

jemink:
[...]

Issue #5455, opened 2021-02-03
issues with createMany


yousufhasan:
As per my understanding  `await prisma.accoutns.createMany()` works only on a single entity and not on sub-entities. hence i tried to use create along with createMany as mentioned in this ticket.

https://github.com/prisma/prisma/issues/4998

 but currently it is not working at sub levels as well which is an issue for example.

![image](https://user-images.githubusercontent.com/13927457/106824448-f4248280-66d6-11eb-86a1-4fe9b1e53ade.png)

jalik:
@yousufhasan  As I understand the doc says when using `createMany`, you cannot nest relation queries, so given your example, even if you called `create` in the top parent query, nesting `lineItems` in `subAccounts` is not supported, once you use `createMany`, no other query can be nested into it.

yousufhasan:
@jalik Shouldn't it be supported? The problem we are facing currently is that `create` is highly inefficient when it comes to creating large nested objects. is there any other way to efficiently save nested objects like in other orm tools like TypeORM?

jalik:
@yousufhasan I am not from the Prisma team, but I guess since it's a preview feature it's still incomplete, but it may be supported later.

pantharshit00:
I am marking this as a new feature request. Currently nested createMany many is not implemented.

AHBruns:
Hey, I just wanted to add my support for this. It would make a major difference in my project.

renatoaraujoc:
This feature would be amazing!

yoni2356:
This would be ground-breaking for my use case, would save tons of redundant code.

ulo:
As nested createMany is not implemented, how should one efficiently load big amounts of relational data (e.g. Users & Posts)?
The logical approach would be to run a createMany on the Users first, and then a createMany on the Posts.
But in order to connect them properly, the first createMany would need to return the inserted Users (or at least the IDs), which it unfortunately does not...
What are your workarounds on this?

giacomocerquone:
> As nested createMany is not implemented, how should one efficiently load big amounts of relational data (e.g. Users & Posts)?
> The logical approach would be to run a createMany on the Users first, and then a createMany on the Posts.
> But in order to connect them properly, the first createMany would need to return the inserted Users (or at least the IDs), which it unfortunately does not...
> What are your workarounds on this?

Many "create" in a for of / foreach (depending on how you want to await the promises) loop for now

And as per @yoni2356 comment

[...]

PR #24064, opened 2024-05-02
feat(client): add `.createManyAndReturn()` top level method

Jolg42:
Based on https://github.com/prisma/prisma-engines/pull/4842

Contributes to https://github.com/prisma/team-orm/issues/1107

Related to https://github.com/prisma/prisma/issues/8131


packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema/__snapshots__/binary.test.ts.snap
```diff
@@ -13853,6 +14203,22 @@ export namespace Prisma {
       args?: SelectSubset<T, DCreateManyArgs<ExtArgs>>
     ): Prisma.PrismaPromise<BatchPayload>
 
+    /**
+     * Create many Ds and returns the data saved in the database.
+     *     @param {DCreateManyAndReturnArgs} args - Arguments to create many Ds.
```
SevInf, line 14208: Should it be aligned this way?

SevInf (changes requested):
Great job, only a few things need to be added/changed:

- Type tests for new code are now done in functional test setup rather than `tsd`
- I think we'll need a functional test that ensures `createManyAndReturn` is not present on unsupported providers, both on type and runtime levels.

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema/__snapshots__/binary.test.ts.snap
Jolg42: Fixed

packages/client/src/__tests__/types/createManyAndReturn/index.test-d.ts
```diff
@@ -0,0 +1,9 @@
+import { expectError } from 'tsd'
```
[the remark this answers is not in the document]
Jolg42: I just deleted this instead.
We already have a test in the functional test suite as `packages/client/tests/functional/methods/createManyAndReturn/_matrix.ts`

SevInf (approved):

Jolg42:
Note: This can be tried out in our dev version `5.14.0-dev.48` and up, it will be part of the official release on May 14th.

Jolg42:
Thanks @marcus13371337 for trying it out!
I think what you mention here is that you would like support for creating nested relations for `.createMany()` and `.createManyAndReturn()`?
There is a tracking issue about that if you are interested: https://github.com/prisma/prisma/issues/5455

marcus13371337:
@Jolg42 No not really, my point was that it would be nice if the return types (and some arguments) of `createManyAndReturn` and `findMany` gave you compatible results. I.e, `items` in this example would be of the same type!
```ts
const items = await prisma.myModel.createManyAndReturn({
  data: [...],
  select: { id: true, name: true }, 
  include: { children: true }
})

const items = await prisma.myModel.findMany({
  where: { ... },
  select: { id: true, name: true },
  include: { children: true }
})
```

Jolg42:
@marcus13371337 You can now try `5.14.0-dev.61` and up, I fixed the types, `select` and `include` (and `omit` is the preview feature flag is set) now work 🎉

Jolg42:
@marcus13371337 Thanks for this, we will check this, it might take a few days, there are some holidays planned in our team 📆 
Expect some news early next week.

Note: `include` `supplier` from the `voucher` side works.
```
      const vouchers = await prisma.voucher.createManyAndReturn({
        data: [
          {
            legalEntityId: legalEntity1.id,
            supplierId: suppliers[0].id,
          },
        ],
        include: {
          legalEntity: true,
          supplier: true,
        },
      })
```

PR #24071, opened 2024-05-03
feat: add a test with `notablescan` for MongoDB

laplab:
Closes https://github.com/prisma/team-orm/issues/1121


packages/client/tests/e2e/mongodb-notablescan/package-lock.json
```diff
@@ -0,0 +1,791 @@
+{
```
SevInf, line 1: This is the lockfile from `npm`, we should remove it.
I think you might've run `npm` instead of `pnpm` at some point.

packages/client/tests/e2e/mongodb-notablescan/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "private": true,
+  "version": "0.0.0",
+  "main": "index.js",
+  "scripts": {},
+  "dependencies": {
+    "@prisma/client": "/tmp/prisma-client-0.0.0.tgz",
+    "mongodb": "6.5"
```
SevInf, line 8: I don't think this package is used

SevInf (approved):
Cleanup is needed in few places, after that LGTM

Jolg42:
Note: to be merged after the release only (happening very soon)

PR #24084, opened 2024-05-06
add usage trend link to README

cwtuan:
I'm an active user of the Prisma package, and I've observed its remarkable growth and rising popularity in the developer community over the past few years. As someone deeply engaged with Prisma, I strongly believe that offering usage trends to newcomers is vital for guiding their decisions regarding ORM technologies.

I propose a minor addition to the Prisma README: including a link that focuses on Prisma's popularity over the last three years.

Thank you for taking the time to consider my suggestion.


janpio:
Thanks, but we are currently not interested in adding this to our README.

PR #24085, opened 2024-05-06
feat(schema-files-loader): Customize file loading logic

SevInf:
Allows to swap files resolution logic used by the package.
By default, package still uses real fs (see `realFsResolver`).
Additinal implementatins are:

- `InMemoryFilesResolver` as the name suggests, keeps virtual file tree
in memory.
- `CompsoiteFilesResolver` combines two resolvers together.

Both will be used for language-tools, where some of the source files
might diverge from their on-disk content while editing.

We had to make them both aware of potential case-sensitivity and adjust their logic
accordingly, because on case-insensitive file systems URIs we getting from VSCode
will not neccessary match casing of files on disk. Both new resolvers need to be aware of that
so they can de-duplicate in-memory and on-disk entries correctly on each system.

Contribute to prisma/team-orm#1042
For usage, see https://github.com/prisma/language-tools/pull/1718
/integration


packages/schema-files-loader/src/resolver/CompositeFilesResolver.ts
```diff
@@ -0,0 +1,40 @@
+import { createFileNameToKeyMapper, FileNameToKeyMapper } from './caseSensitivity'
+import { CaseSensitivityOptions, FilesResolver, FsEntryType } from './types'
+
+/**
+ * Files resolver that combines two other resolvers
+ * together. Files existing in either one of those will be
+ * reported. Content existing in
```
Druue, line 7: Content existing in...?

Druue (approved):
lgtm! :) ✨

Druue:
Seems unrelated, gonna re-run
image

Druue:
This seems like something broke here
https://github.com/prisma/prisma/actions/runs/9003113770/job/24738228716?pr=24085
image

packages/schema-files-loader/src/resolver/CompositeFilesResolver.ts
SevInf: Updated

PR #24099, opened 2024-05-07
fix(client/createManyAndReturn): fix TS input types

Jolg42:
Related to https://github.com/prisma/team-orm/issues/1107


packages/client/src/generation/TSClient/Model.ts
```diff
@@ -67,7 +67,6 @@ export class Model implements Generatable {
         action === 'updateMany' ||
         action === 'deleteMany' ||
         action === 'createMany' ||
-        action === 'createManyAndReturn' ||
```
Jolg42, line 70: This was the reason why the types did not allow `include` / `exclude` / `omit`

SevInf (approved):

PR #24127, opened 2024-05-08
feat(cli): add `prismaSchemaFolder` support to `prisma generate` via `mergeSchemas`

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/1043.
It likely works out of the box to Prisma Studio due to backward compatibility (to check later, once `migrate` is done), so this PR also contributes to https://github.com/prisma/team-orm/issues/1044.


packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -97,7 +98,6 @@ export class TSClient implements Generatable {
         ? (Object.keys(binaryPaths.libqueryEngine ?? {}) as BinaryTarget[])
         : (Object.keys(binaryPaths.queryEngine ?? {}) as BinaryTarget[])
 
-    const inlineSchema = await readFile(schemaPath, 'utf8')
```
jkomyno, line 100: I don't understand why this read from disk were here. We are already given a `this.options.datamodel`.

packages/client/src/generation/generateClient.ts
```diff
@@ -411,9 +411,7 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   }
 
   const schemaTargetPath = path.join(outputDir, 'schema.prisma')
-  if (schemaPath !== schemaTargetPath) {
-    await fs.copyFile(schemaPath, schemaTargetPath)
```
jkomyno, line 415: We can no longer use `fs.copyFile`: `schemaPath` may refer to a schema folder with many `*.prisma` files to treat as a whole Prisma schema.

packages/client/src/runtime/utils/SourceFileSlice.ts
```diff
@@ -19,6 +19,7 @@ export class SourceFileSlice {
   static read(filePath: string): SourceFileSlice | null {
     let content: string
     try {
+      // TODO: consider using `getSchemaPath` here?
       content = fs.readFileSync(filePath, 'utf-8')
```
jkomyno, line 23: Before I start digging, do we know what `SourceFileSlice` is actually used for? Does it read schemas or what?

SevInf: It does not read schema files, it reads JS sources.
Used when we create a fancy stacktraces for errors.

packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -61,6 +60,7 @@ export class TSClient implements Generatable {
     this.genericsInfo = new GenericArgsInfo(this.dmmf)
   }
 
+  // eslint-disable-next-line @typescript-eslint/require-await -- must be async to match previous interface
```
SevInf, line 63: It is not a public class (it is not even possible to load it through any of the private modules), we can change the interface.

packages/internals/src/cli/getSchema.ts
```diff
@@ -24,14 +24,18 @@ export type GetSchemaResult = {
  * Async
  */
 
-type ReturnWithNonNullArg<In, Out> = In extends undefined ? Out | null : Out
-
+export async function getSchemaPath(): Promise<GetSchemaResult | null>
```
SevInf, line 27: Just for my understanding: overaloads were added to exclude "empty schema path, non-empty options" situation, right?

packages/internals/src/get-generators/getGenerators.ts
```diff
@@ -84,6 +85,8 @@ export async function getGenerators(options: GetGeneratorOptions): Promise<Gener
   if (!fs.existsSync(schemaPath)) {
     throw new Error(`${schemaPath} does not exist`)
   }
+
+  const schemas = (await getSchemaPath(schemaPath))!.schemas
```
SevInf, line 89: I think rather than using `!` we can get rid of `existsSync` check above and throw `${schemaPath} does not exist` if it `getSchemaPath` returns `null`.

SevInf (approved):
Few nitpicks and questions, other than that - looks good. Once tests are green we can merge it, i think.

packages/internals/src/cli/getSchema.ts
jkomyno: They were added so that the `Promise` result is only nullable when no first argument (the optional schema path to look in) is provided. This makes object destructuring more convenient

Issue #11986, opened 2022-02-22
`$on` argument type error when extending prisma client


smitssjors:
### Bug description

When extending the `PrismaClient` class it is not possible to pass the `query` event type to `prisma.$on()` even though the correct options are passed to the `PrismaClient` in the `super()` call.

```ts
class Test extends PrismaClient {
  constructor() {
    super({
      log: [
        {
          emit: "event",
          level: "query",
        },
      ],
    });

    this.$on("query", (e) => {
      console.log(e);
    });
  }
}
```
The code above produces the following error:
```
TS2345: Argument of type '"query"' is not assignable to parameter of type '"beforeExit"'.
```

### How to reproduce

1. Download the quick-start project
2. Paste in the above code snippet
3. Run `yarn dev`

### Expected behavior

The `query` parameter is valid.

### Prisma information

Nothing relevant

### Environment & setup

- OS: Ubuntu (WSL on Windows 11)
- Database: SQLite
- Node.js version: 16.14.0

### Prisma Version

```
prisma                  : 3.7.0
@prisma/client          : 3.7.0
Current platform        : debian-openssl-1.1.x
Query Engine (Node-API) : libquery-engine 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
Migration Engine        : migration-engine-cli 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/migration-engine-debian-openssl-1.1.x)
Introspection Engine    : introspection-core 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/introspection-engine-debian-openssl-1.1.x)
Format Binary           : prisma-fmt 8746e055198f517658c08a0c426c7eec87f5a85f (at node_modules/@prisma/engines/prisma-fmt-debian-openssl-1.1.x)
Default Engines Hash    : 8746e055198f517658c08a0c426c7eec87f5a85f
Studio                  : 0.445.0
```
Also happens on
```
prisma                  : 3.9.2
@prisma/client          : 3.9.2
Current platform        : debian-openssl-1.1.x
Query Engine (Node-API) : libquery-engine bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at ../../node_modules/@prisma/engines/libquery_engine-debian-openssl-1.1.x.so.node)
[...]

Issue #7068, opened 2021-05-13
Prisma Client Type Issues with NestJS


Thorbenl:
### Bug description

I tried to enable logging within my NestJS project and Prisma to no avail
I am trying to use Prisma with the latest version of NestJS, but am getting errors on trying to implement logging. Prisma Client types don't seem to work correctly. This is just a suspicion of mine.

The error:
```
TSError: ⨯ Unable to compile TypeScript:
src/test.service.ts:9:31 - error TS2345: Argument of type '"query"' is not assignable to parameter of type '"beforeExit"'.

9        this.prismaService.$on('query', e => {
                                ~~~~~~~
src/test.service.ts:10:39 - error TS2339: Property 'query' does not exist on type '() => Promise<void>'.

10             console.log("Query: " + e.query)
                                         ~~~~~
src/test.service.ts:11:42 - error TS2339: Property 'duration' does not exist on type '() => Promise<void>'.

11             console.log("Duration: " + e.duration + "ms")
```

### How to reproduce

Project is a copy of https://github.com/fivethree-team/nestjs-prisma-starter

The `prisma.service.ts`:
```
import { Injectable, OnModuleInit, OnModuleDestroy } from '@nestjs/common'
import { PrismaClient } from '@prisma/client'

@Injectable()
export class PrismaService extends PrismaClient implements OnModuleInit, OnModuleDestroy {
    constructor() {
        super();
    }

    async onModuleInit() {
        await this.$connect()
    }

    async onModuleDestroy() {
        await this.$disconnect()
    }
}
```
Works fine, I can use it across the API and access the DB. However, according to the Prisma 2 Docs on [Logging](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging#event-based-logging), I need to pass

```
import { PrismaClient } from '@prisma/client'

const prisma = new PrismaClient({
  log: [
    { level: 'warn', emit: 'event' },
    { level: 'info', emit: 'event' },
    { level: 'error', emit: 'event' },
  ],
```

and then use it like this:
```
@Injectable()
export class TestService {
    constructor(private prismaService: PrismaService) {
        this.prismaService.$on('query', e => {
            console.log("Query: " + e.query)
            console.log("Duration: " + e.duration + "ms")
        })
    }
```

Sadly, when compiling, I get these errors:

```
TSError: ⨯ Unable to compile TypeScript:
[...]

Issue #6279, opened 2021-03-27
Wrong type for $on method in PrismaClient


igor-mardari-ft:
## Bug description

When using `prisma.$on('error'` I get error a compilation error 
![image](https://user-images.githubusercontent.com/68684530/112715112-3b650d80-8ee7-11eb-81db-280c940293b9.png)

## How to reproduce

1. Just do somewhere in the code `prisma.$on('error'`. Instead of `error` it could be any `LogLevel`.

## Expected behaviour

I shouldn't get a compilation error when using `'error'` because it's matching `LogLevel` type.

## Prisma information

I've investigated the issue a little bit and it comes from the `GetLogType` 
![image](https://user-images.githubusercontent.com/68684530/112715207-ce05ac80-8ee7-11eb-9690-e366953dc0d2.png)

If I change
```
T['emit'] extends string
```
instead of 
```
T['emit'] extends 'event'
```
then, the error is gone. 

No matter what configuration I set to `log` property. It could be `LogLevel` or `LogDefinition` or nothing. It behaves the same. In this case, it doesn't make any sense to have such a complex type if it doesn't work properly.

At a simple level, this line,:
```
$on<V extends (U | 'beforeExit')>
```
could be rewritten to 
```
$on<V extends (LogLevel | 'beforeExit')>
```

## Environment & setup

- OS: MacOs
- Database: PostgreSQL
- Node.js version: v12.18.3
- Prisma version: 2.19.0

GarryOne:
EDIT: I found this similar issue: https://github.com/prisma/prisma/issues/4746 now, if I add 
```<Prisma.PrismaClientOptions, 'error'>``` it works, but this info is nowhere in the documentation. Would be helpful to be included asa example because it's not an obvious thing.

Thanks.

pabacham:
The same issue happens with such construction in Nest.js
import { Injectable, OnModuleInit, OnModuleDestroy } from '@nestjs/common';
import { PrismaClient } from '@prisma/client';
import { logQueryDuration } from './middlewares/logQueryDuration';

@Injectable()
export class PrismaService
  extends PrismaClient
  implements OnModuleInit, OnModuleDestroy {
  constructor() {
    super({
      log: [
        {
          emit: 'event',
          level: 'query',
        },
        {
          emit: 'stdout',
          level: 'error',
        },
        {
          emit: 'stdout',
          level: 'info',
        },
        {
          emit: 'stdout',
          level: 'warn',
        },
      ],
    });

[...]

Issue #4746, opened 2020-12-21
Invalid argument type with this.$on(...) in an extended class of PrismaClient


devalnor:
## Bug description

Typescript return `Argument of type 'string' is not assignable to parameter of type 'never'` when we want to use `this.$on()`in a extend class of PrismaClient

## How to reproduce

Bug can be reproduce based  on official starter 

`curl -L https://pris.ly/quickstart | tar -xz --strip=2 quickstart-master/typescript/starter`

```ts 
// index.ts
class Test extends PrismaClient {
  constructor() {
    super({
      log: [
        {
          emit: 'event',
          level: 'info',
        },
      ],
    });
    // Return Argument of type 'string' is not assignable to parameter of type 'never'.
    this.$on('info', (e) => {
      console.log(e);
    });
    this.post.findMany()
  }
  
  test() {
    // Return Argument of type 'string' is not assignable to parameter of type 'never'.
    this.$on('info', (e) => {
      console.log(e);
    });
  }
} 
```

## Expected behavior

 `this.$on()` may accept  strings `info`,`query` or `warn` as log level like it does with `prisma.$on()`

More info: https://www.prisma.io/docs/concepts/components/prisma-client/logging#the-log-option

## Environment & setup

- OS: Mac OS 10.14.6
- Database:SQLite
- Node.js version: v14.15.1
- Prisma version: 2.13.1

```

```

pantharshit00:
@devalnor 

Do this:
```ts
import { PrismaClient, Prisma } from "@prisma/client";

class Test extends PrismaClient<Prisma.PrismaClientOptions, "info"> {
  constructor() {
    super({
      log: [
        {
          emit: "event",
          level: "info",
        },
      ],
    });
    // Return Argument of type 'string' is not assignable to parameter of type 'never'.
    this.$on("info", (e) => {
      console.log(e);
    });
    this.post.findMany();
  }

  test() {
    // Return Argument of type 'string' is not assignable to parameter of type 'never'.
    this.$on("info", (e) => {
      console.log(e);
    });
  }
}
```


You need to pass generics yourself as our types can't infer that from usage here. We have these generics in place as you limit type of first argument of `$on` which is only enabled when user passes some option to the client.

devalnor:
It work!
Thank you @pantharshit00

nikelborm:
May be related https://github.com/prisma/prisma/pull/24133

PR #24133, opened 2024-05-08
fix(client): type inference for `prismaClient.$on(...)`

nikelborm:
# Problem

1. incorrect inference in generic `GetLogType` returned `never` instead of `T` on line 6. ([original piece of code here](https://github.com/prisma/prisma/blob/main/packages/client/src/generation/TSClient/PrismaClient.ts#L555))
```typescript
/*1.*/export type GetLogType<T extends LogLevel | LogDefinition> =
/*2.*/  T extends LogDefinition
/*3.*/    ? T['emit'] extends 'event'
/*4.*/      ? T['level']
/*5.*/      : never
/*6.*/  : never
```
This code led to error when some of the elements in `T` argument of `GetEvents` which are being later passed to `GetLogType` were `LogLevel` string literals instead of `LogDefinition` objects.

2. Case when developer forgot to add `as const` (and there's no suggestion in documentation to add it) to the argument of `PrismaClient` constructor.

[Example in documentation:](https://github.com/prisma/docs/blob/main/content/200-orm/200-prisma-client/600-observability-and-logging/130-logging.mdx?plain=1#L34-L37)
```typescript
  log: ['query', 'info', 'warn', 'error'],
})
```

Since neither developer, nor `PrismaClient` generic itself (`const` before `U` is absent [here](https://github.com/prisma/prisma/blob/main/packages/client/src/generation/TSClient/PrismaClient.ts#L467)) didn't tell typescript to treat argument of `new PrismaClient(...)` as constant, the type information of string literals will be erased and typescript will handle it just as `{ log: string[] }`.

3. `GetEvents` was hardcoded to expect arrays with length at max 4-elements [here](https://github.com/prisma/prisma/blob/main/packages/client/src/generation/TSClient/PrismaClient.ts#L557)

# Solution

1. `GetLogType` now returns those literals as is when they don't extend `LogDefinition` (object), but also with additional check which then forces `T` of `GetLogType<T>` to extend `LogLevel`
2. Added useful `const` modifier to the second `U` argument of `PrismaClient` generic which preserves string literal type info taken from the argument of `PrismaClient` constructor
3. `GetEvents` generic is now a ~~recursive~~ **UPD distributed** type and doesn't depend on length of `PrismaClientOptions.log` array

Might fix some of the following issues:
1. https://github.com/prisma/prisma/issues/19463
2. https://github.com/prisma/docs/issues/5021
3. https://github.com/prisma/docs/issues/1564
4. https://github.com/prisma/prisma/issues/4746
5. https://github.com/prisma/prisma/issues/7068
6. https://github.com/notiz-dev/nestjs-prisma/issues/56
7. https://github.com/prisma/prisma/issues/11986
8. https://github.com/prisma/prisma/issues/6279
9. https://github.com/prisma/docs/issues/3424

## Before
`U` argument of `PrismaClient` generic is `never` mainly because of the 1st problem mentioned above and we can't call `prismaClient.$on(...)` passing any string literal to it:
Screenshot from 2024-03-05 14-23-28

## After
The 1st screenshot shows that the second inferred argument `U` of generic `PrismaClient` stored in `DB` constant is now `"info" | "query" | "error"` according to IDE's tooltip above the line `DB.$on('', () => {})` and typescript also tells to replace first argument `''` to any of `U` union members: 
Screenshot from 2024-03-05 14-37-44

2nd screenshot shows that when we pass correct 1st argument to `DB.$on` it correctly infers type of the 2nd function argument:
Screenshot from 2024-03-05 14-41-01


kduprey:
Any updates or progress on this? Are we able to get this merged? 
Running into this issue when attempting to customise the `prismaClient` event logging.

packages/client/src/generation/TSClient/PrismaClient.ts
```diff
@@ -464,7 +464,7 @@ export class PrismaClientClass implements Generable {
     return `${this.jsDoc}
 export class PrismaClient<
   ClientOptions extends Prisma.PrismaClientOptions = Prisma.PrismaClientOptions,
-  U = 'log' extends keyof ClientOptions ? ClientOptions['log'] extends Array<Prisma.LogLevel | Prisma.LogDefinition> ? Prisma.GetEvents<ClientOptions['log']> : never : never,
+  const U = 'log' extends keyof ClientOptions ? ClientOptions['log'] extends Array<Prisma.LogLevel | Prisma.LogDefinition> ? Prisma.GetEvents<ClientOptions['log']> : never : never,
```
aqrln, line 467: This is a breaking change. Const generics require TypeScript 5.0 or later but Prisma supports 4.7+ right now. We're raising the minimal required version to 5.1 soon in Prisma 6 though.

aqrln (commented):
Thanks for the PR and sorry it was overlooked! While this can't land in Prisma 5.x (unless the solution can be changed to not rely on TypeScript 5 features), this should definitely be merged for upcoming Prisma 6.

(cc @jkomyno)

packages/client/src/generation/TSClient/PrismaClient.ts
nikelborm: Yeah... Unfortunately it (`Const generics require TypeScript 5.0 or later`) does...
But you can cherry pick one commit and not touch the other one.
Or I can create 2 split PRs. One of them for the closest version and the second one is for 6.0
What do you think?

aqrln: Prisma 6 is around the corner so let's just merge the whole PR.

aqrln:
@nikelborm could you please rebase your PR onto latest `main` so you don't have TS 4.7 failures (we already bumped the minimal required TypeScript version on main)? There are also other things failing in https://github.com/prisma/prisma/actions/runs/11617514866/job/32352930786?pr=24133 that need to be looked into:

```
  ● valid types › /home/runner/work/prisma/prisma/packages/client/src/__tests__/types/deprecation

    expect(received).resolves.not.toThrow()

    Received promise rejected instead of resolved
    Rejected to value: [Error: Command failed with exit code 1: /opt/hostedtoolcache/node/16.20.2/x64/bin/node /home/runner/work/prisma/prisma/packages/client/src/utils/compilerWorker.js /home/runner/work/prisma/prisma/packages/client/src/__tests__/types/deprecation/test.ts
    /home/runner/work/prisma/prisma/packages/client/src/utils/compilerWorker.js:20
        throw new Error('Compilation Error\n' + formatted)
        ^·
    Error: Compilation Error
    src/__tests__/types/deprecation/test.ts(20,14): error TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.·
```

nikelborm:
> @nikelborm could you please rebase your PR onto latest `main` so you don't have TS 4.7 failures (we already bumped the minimal required TypeScript version on main)? There are also other things failing in [prisma/prisma/actions/runs/11617514866/job/32352930786?pr=24133](https://github.com/prisma/prisma/actions/runs/11617514866/job/32352930786?pr=24133) that need to be looked into:
> 
> ```
>   ● valid types › /home/runner/work/prisma/prisma/packages/client/src/__tests__/types/deprecation
> 
>     expect(received).resolves.not.toThrow()
> 
>     Received promise rejected instead of resolved
>     Rejected to value: [Error: Command failed with exit code 1: /opt/hostedtoolcache/node/16.20.2/x64/bin/node /home/runner/work/prisma/prisma/packages/client/src/utils/compilerWorker.js /home/runner/work/prisma/prisma/packages/client/src/__tests__/types/deprecation/test.ts
>     /home/runner/work/prisma/prisma/packages/client/src/utils/compilerWorker.js:20
>         throw new Error('Compilation Error\n' + formatted)
>         ^·
>     Error: Compilation Error
>     src/__tests__/types/deprecation/test.ts(20,14): error TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.·
> ```

Rebased and fixed the issue. Not sure about those 2 failing CI's left. They seem to be failing not because of my changes

packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
```diff
@@ -2277,10 +2277,15 @@ export namespace Prisma {
...
 
-  export type GetLogType<T extends LogLevel | LogDefinition> = T extends LogDefinition ? T['emit'] extends 'event' ? T['level'] : never : never
-  export type GetEvents<T extends any> = T extends Array<LogLevel | LogDefinition> ?
-    GetLogType<T[0]> | GetLogType<T[1]> | GetLogType<T[2]> | GetLogType<T[3]>
-    : never
+  export type CheckIsLogLevel<T> = T extends LogLevel ? T : never;
+
+  export type GetLogType<T> = T extends LogDefinition
+    ? (T['emit'] extends 'event' ? CheckIsLogLevel<T['level']> : never)
+    : CheckIsLogLevel<T>;
```
aqrln, line 2284: So, about this. String literals are a shorthand for `{ level: "...", emit: "stdout" }`, so it was intentional that `never` was returned in this case, since [according to the docs](https://www.prisma.io/docs/orm/prisma-client/observability-and-logging/logging) it's either events or logs in stdout but not both.

However, this is not the actual behavior. You will indeed receive both events and console logs when using either the shorthand or the full `emit: "stdout"` form.

Technically this is a bug, as it contradicts both the types and the documentation. However, since we're doing a major release anyway, I'm fine with changing the semantics the way you propose and updating the types and the docs to match the actual behavior because it makes more sense to me too. In that case, though, we shouldn't restrict the `LogDefinition` case to `T['emit'] extends 'event'` either.

nikelborm: Do you mean changing it from `T['emit'] extends 'event' ? CheckIsLogLevel<T['level']> : never` to just `CheckIsLogLevel<T['level']>`, right?

Also did I understand correctly that you want me to change the docs too?

aqrln: > Do you mean changing it from T['emit'] extends 'event' ? CheckIsLogLevel<T['level']> : never to just CheckIsLogLevel<T['level']>, right?

Yes

> Also did I understand correctly that you want me to change the docs too? I'm happy to do it

That would be amazing, but up to you, we can also do this. Probably a small note that tells that `emit: 'stdout'` logs are also emitted as events is sufficient.

aqrln (approved):
LGTM, sorry for this falling through the cracks — don't hesitate to ping the reviewers in the future. I've updated the PR to port the changes to the new generator and updated the documentation comments.

Issue #24208, opened 2024-05-15
Bug: Migrations not compatible with D1


hrueger:
### Bug description

Generated migrations for D1 (following [this guide](https://www.prisma.io/docs/orm/overview/databases/cloudflare-d1#migration-workflows)) are not compatible with D1. When changing a table with relations, foreign keys are disabled and re-enabled with `PRAGMA foreign_keys=OFF;` and `PRAGMA foreign_keys=ON;`.
However, [D1 seems to need `PRAMGA defer_foreign_keys=ON;` and `PRAGMA defer_foreign_keys=OFF;`](https://developers.cloudflare.com/d1/reference/database-commands/#pragma-defer_foreign_keys--onoff) (also notice the swapped `ON` and `OFF`).

When applied, the migration fails with `✘ [ERROR] FOREIGN KEY constraint failed`.

### How to reproduce

Minimal reproduction: https://github.com/hrueger/prisma-24208

### Expected behavior

Migrations are generated with `PRAGMA defer_foreign_keys` instead so that the migration can be applied.

### Prisma information

```prisma

generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["driverAdapters"]
}

datasource db {
  provider = "sqlite"
  url      = env("DATABASE_URL")
}

model User {
  id        Int      @id @default(autoincrement())
  createdAt DateTime @default(now())
  email     String   @unique
  name      String?
  posts     Post[]
  // rename this field, let's say to `newField`
  oldField  String?
}

model Post {
  id        Int      @id @default(autoincrement())
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  title     String
  author    User    @relation(fields: [authorId], references: [id])
  authorId  Int
}
```

### Environment & setup

- OS: macOS
- Database: Cloudflare D1 / SQLite
- Node.js version: 21.6.1

### Prisma Version

```
prisma                  : 5.14.0
@prisma/client          : 5.14.0
Computed binaryTarget   : darwin-arm64
Operating System        : darwin
Architecture            : arm64
Node.js                 : v21.6.1
Query Engine (Node-API) : libquery-engine e9771e62de70f79a5e1c604a2d7c8e2a0a874b48 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
[...]

PR #24214, opened 2024-05-15
test(cli): e2e migrations for D1

jkomyno:
This PR adds automated tests for Cloudflare D1 migrations with Prisma https://www.prisma.io/docs/orm/overview/databases/cloudflare-d1#migration-workflows.
It also showcases that https://github.com/prisma/prisma/issues/24208 is now fixed via https://github.com/prisma/prisma-engines/pull/4871.

This PR relies on an integration release of https://github.com/prisma/prisma-engines/pull/4871 (currently being built).

**Note**: It's currently added to the `cli` package because of a previous attempt at using `prisma generate` within the test. Still, since at some point we will add `prisma db seed` support for D1, I think we can keep the CLI e2e tests there.


Jolg42:
Note: the base branch needs to be changed later

jkomyno:
> Note: the base branch needs to be changed later

Yep, I was waiting for the engines to compile, so you could see the list of green marks on the CI :)

packages/cli/src/__tests__/commands/CLI.test.ts
```diff
@@ -80,3 +83,705 @@ it('introspect should include deprecation warning', async () => {
   `)
   expect(ctx.mocked['console.error'].mock.calls).toHaveLength(0)
 })
+
+describe('e2e sqlite', () => {
```
Jolg42, line 87:
```suggestion
describe('local d1 db', () => {
```

Jolg42: It's only about D1 after all right?

packages/cli/src/__tests__/commands/CLI.test.ts
```diff
@@ -1,5 +1,7 @@
-import { jestConsoleContext, jestContext } from '@prisma/get-platform'
-import { DbPull } from '@prisma/migrate'
+import { type BaseContext, jestConsoleContext, jestContext } from '@prisma/get-platform'
+import { DbPull, MigrateCommand, MigrateDiff } from '@prisma/migrate'
+import execa from 'execa'
+import path from 'path'
 
 import { CLI } from '../../CLI'
```
Jolg42, line 6: In general: it looks like this file is the wrong place for these tests.

It would be better to create a new file in the `migrate` package here, for example: 
```
prisma/packages/migrate/src/__tests__/local-clouflare-d1-db-e2e.test.ts
```

packages/cli/src/__tests__/commands/CLI.test.ts
```diff
@@ -80,3 +83,705 @@ it('introspect should include deprecation warning', async () => {
...
+        "
+      `)
+
+      // Apply `init` migration to D1
+      const wranglerExecuteInitMigration = await applyWranglerD1Migration(ctx)
+      expect(wranglerExecuteInitMigration.exitCode).toBe(0)
+
+      // Seed some data into D1.
+      // Once we move the `PRIMARY KEY` from `User.id` to `User.count`, we would like to observe the `FOREIGN KEY` constraint failure,
+      // demonstrating that Prisma migrations check the integrity of the data, and rollback the migration if the data is not consistent.
```
Jolg42, line 501: This is untested, right?

Jolg42 (approved):
Looks good!

The tests in general are really long because they intend to test end to end. 

For some cases it would be nice, especially if we want to add more tests later to make them shorter by manually creating an init migration which contains all what is needed.

Example: (empty) -> apply init migration which also contains some data -> try to apply the next migration that should fail

Because it's ok to have a couple of long tests for e2e purposes, but it's not easy to maintain if we copy and paste them many times.

PR #24221, opened 2024-05-16
chore(renovate): Ensure JS drivers updates happen synchronously

SevInf:
At the moment, dev and production dependencies are split into different
groups. That means that JS driver in adapter packages (where it is a dev
dependcy) might update at a different time than production dependency in
`bundled-js-drivers`. The whole point of `bundled-js-drivers` packages
was to avoid version conflicts when adapter are used for engine tests
and that breaks it.


jkomyno (approved):

Jolg42 (approved):

Jolg42:
Maybe add `@types/pg`?

PR #24222, opened 2024-05-16
feat: display the socket path during migration

brian-dlee:
In this update, I've modified the migrate tool to display the socket path if it configured in the database connection. The problem that this solves is, for organizations like mine that use Google's Cloud SQL, our connection strings always include `localhost:3306` and that actually does not indicate the true target database. Using Google `cloud-sql-proxy`, a unix socket is created to facilitate encrypted traffic to the database with Google IAM integration. This socket is named after the database you are connected to (e.g. `orgname:us-west2:db-instance-name`). In our adoption of Prisma, we've actually had accidents occur where someone runs a migration against a cloud environment thinking they are doing it locally since the socket name is not advertised.

An assumption I made in this PR is that any time a socket is supplied in the URL, it is the key indicator of the target database and completely takes over for `host:port` as the displayed identifier.

I would have liked to update the underlying prisma client connection messaging as well because database errors uninformatively indicate a connection failure to `localhost:3306` and omit the socket name which is more important when debugging connectivity issues. I wasn't able to easily track down where this code lives and I was afraid of the widespread use of this. Truthfully, I'm way more worried about incorrectly applying migrations or performing database resets then debugging the Prisma client at runtime.

### DIFF
--- a/packages/migrate/src/utils/ensureDatabaseExists.ts
+++ b/packages/migrate/src/utils/ensureDatabaseExists.ts
@@ -11,6 +11,7 @@ import {
 import { bold } from 'kleur/colors'
 
 import { ConnectorType } from './printDatasources'
+import { getSocketFromDatabaseCredentials } from './unixSocket'
 
 export type MigrateAction = 'create' | 'apply' | 'unapply' | 'dev' | 'push'
 export type PrettyProvider = 'MySQL' | 'PostgreSQL' | 'SQLite' | 'SQL Server' | 'CockroachDB' | 'MongoDB'
@@ -210,8 +211,10 @@ export function getDbLocation(credentials: DatabaseCredentials): string | undefi
     return credentials.uri!
   }
 
-  if (credentials.socket) {
-    return `unix:${credentials.socket}`
+  const socket = getSocketFromDatabaseCredentials(credentials)
+
+  if (socket) {
+    return `unix:${socket}`
   } else if (credentials.host && credentials.port) {
     return `${credentials.host}:${credentials.port}`
   } else if (credentials.host) {
--- a/packages/migrate/src/utils/ensureDatabaseExists.ts
+++ b/packages/migrate/src/utils/ensureDatabaseExists.ts
@@ -210,7 +210,9 @@ export function getDbLocation(credentials: DatabaseCredentials): string | undefi
     return credentials.uri!
   }
 
-  if (credentials.host && credentials.port) {
+  if (credentials.socket) {
+    return `unix:${credentials.socket}`
+  } else if (credentials.host && credentials.port) {
     return `${credentials.host}:${credentials.port}`
   } else if (credentials.host) {
     return `${credentials.host}`
--- a/packages/migrate/src/utils/unixSocket.ts
+++ b/packages/migrate/src/utils/unixSocket.ts
@@ -1,0 +1,17 @@
+import { DatabaseCredentials } from '@prisma/internals'
+
+const simpleUnixPathPattern = /^\.{0,2}\//
+
+export function getSocketFromDatabaseCredentials(credentials: DatabaseCredentials): string | null {
+  if (credentials.type === 'postgresql') {
+    const host = credentials.host
+
+    if (typeof host === 'string' && simpleUnixPathPattern.test(host)) {
+      return host
+    }
+
+    return null
+  }
+
+  return credentials.socket ?? null
+}
--- a/packages/migrate/src/utils/unixSocket.ts
+++ b/packages/migrate/src/utils/unixSocket.ts
@@ -1,5 +1,21 @@
 import { DatabaseCredentials } from '@prisma/internals'
 
+/**
+ * match a string that looks like a unix file path
+ * implicit relative paths are ignored due to their ambiguity with hostnames
+ * (.e.g my-database-socket can both be a hostname and a file path missing the ./ prefix)
+ *
+ * example 1: an explicit relative path in the same directory
+ * ./sockets/my-database-socket
+ *
+ * example 2: an explicit relative path in a parent directory
+ * ../sockets/my-database-socket
+ *
+ * example 3: an absolute path
+ * /sockets/my-database-socket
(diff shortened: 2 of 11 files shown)


janpio:
That sounds pretty useful. Can you share an (of course anonymized) example of what such a connection string effectively looks like?

What "underlying prisma client connection messaging" do you mean? Most probably it comes from the Prisma engines (query and schema engine) with their codebases at https://github.com/prisma/prisma-engines.

brian-dlee:
@janpio Here's an anonymized output from me running the CLI after my changes are applied.

```
node ./packages/cli/build/index.js migrate dev
Datasource "db": MySQL database "my-db-name" at "unix:/Users/my-username/.cloudsql/my-project-name:us-west1:my-instance-name"

Error: P1017: Server has closed the connection.
```

The "underlying prisma client connection messaging" I meant is the error messaging that occurs if a database connection cannot be established. If we ever have problems with Prisma connecting to the database from Google Cloud Run we see error messages like "unable to connect to database at `localhost:3306`. If these error messages included the socket name (when appropriate) it could mean easier debugging for us. Maybe, for instance, we configured the incorrect cloud sql mount. It makes sense that I couldn't find it given that it's in the other repository 😅 .

packages/migrate/src/utils/ensureDatabaseExists.ts
```diff
@@ -210,7 +210,9 @@ export function getDbLocation(credentials: DatabaseCredentials): string | undefi
     return credentials.uri!
   }
 
-  if (credentials.host && credentials.port) {
+  if (credentials.socket) {
```
janpio, line 213: This should also include the `host` query parameter for PostgreSQL: https://www.prisma.io/docs/orm/overview/databases/postgresql#connecting-via-sockets Not sure how/if this is accessible here as `credentials.host` seems to be the parsed hostname though.

janpio:
So `localhost:3306` becomes `unix:/Users/my-username/.cloudsql/my-project-name:us-west1:my-instance-name`, where `unix:` is hardcoded and `/Users/my-username/.cloudsql/my-project-name:us-west1:my-instance-name` is what was provided via the query string paramter `socket`. That should be the necessary information to transfer the same over to prisma-engines.

brian-dlee:
Yep, that's right. 

I'm happy to make the same code changes to `prisma-engines` if you think that's necessary here. I'm not exactly an expert in Rust. My skill level is moderate, though. I've written a handful of small apps.

packages/migrate/src/utils/ensureDatabaseExists.ts
brian-dlee: I'll handle the socket path supplied as host for PostgreSQL here. I think given that it's a unix socket I'll look for a host value that appears to be a unix-style path pattern.

janpio:
Not necessary, but the right thing for consistency - as you raised yourself about the error message,, that would otherwise be misleading.

packages/migrate/src/utils/ensureDatabaseExists.ts
brian-dlee: All set

packages/migrate/src/utils/unixSocket.ts
```diff
@@ -0,0 +1,17 @@
+import { DatabaseCredentials } from '@prisma/internals'
+
+const simpleUnixPathPattern = /^\.{0,2}\//
+
+export function getSocketFromDatabaseCredentials(credentials: DatabaseCredentials): string | null {
+  if (credentials.type === 'postgresql') {
```
Druue, line 6: This also affects `cockroachdb`

> ```
> postgres://root@?host=/path/to/client&port=26257
> ```
> [source](https://www.cockroachlabs.com/docs/stable/connection-parameters#example-uri-for-a-unix-domain-socket)

Druue:
Also, no changes should be needed in engines :)
We already use the socket path for errors from there, e.g.

```
❯ npx prisma migrate dev
Datasource "db": CockroachDB database "crdb", schema "public" at "localhost"

Error: P1001: Can't reach database server at `/tmp:5432`

Please make sure your database server is running at `/tmp:5432`.
```

Druue:
I find it curious that you mention running into:
"""
we see error messages like "unable to connect to database at `localhost:3306`"
"""

Do you have a specific PXYZ error code for this?

Druue:
Ah, it seems we don't detect this properly for mysql in engines
DATABASE_URL="mysql://root:prisma@localhost:3309/database?socket=/path/to/sock.sock"
```
❯ npx prisma migrate dev
Datasource "db": MySQL database "database" at "localhost:3309"

Error: P1001: Can't reach database server at `localhost:3309`

Please make sure your database server is running at `localhost:3309`
```

packages/migrate/src/utils/unixSocket.ts
brian-dlee: Okay, I've made that update. I also noticed there is an alternate value for PostgreSQL in the list of connection types, `postgres`, and I handled that one as well.

packages/migrate/src/utils/unixSocket.ts
```diff
@@ -0,0 +1,17 @@
+import { DatabaseCredentials } from '@prisma/internals'
+
+const simpleUnixPathPattern = /^\.{0,2}\//
```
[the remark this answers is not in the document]
brian-dlee: A comment has been added to this regex. 

Note: I'm explicitly ignoring relative paths that do not begin with `./`. This may be problematic for some. I'm not sure. In theory, I could look for anything that includes a `/` and interpret it as a unix path which would allow someone to omit a `./` prefix when supplying a relative path.

brian-dlee:
@Druue thanks for the update in the prisma engines for MySQL! I'm indeed using MySQL (migrating to cockroach actually).

All your comments have been addressed. I appreciate your attention to the PR!

Druue (approved):
Hey @brian-dlee!
Thanks so much for the changes, this looks good to me now :)

PR #24232, opened 2024-05-17
feat(cli): Add silent / quiet global CLI flag

### LINKED ISSUE #23192: CLI: `--silent` to silence any output
## Problem

Prisma CLI outputs things. Often that output is not needed in CI and other automated scenarios, and usage can rely on the exit codes of the CLI instead.

## Suggested solution

Add global option `--silent` option to CLI that hides all output.

## Additional context

Related issue:
- https://github.com/prisma/prisma/issues/3390

paulomarg:
Closes #23192 

As mentioned in the issue, there are some scenarios where we want to omit the regular message returned from a command, where the output might make it harder to spot legitimate issues, or bloats the output of some scripts.

To achieve that, this PR introduces a new global `--quiet` flag (aliased to `--silent` if I understood the setup correctly 😄) that simply ignores the message returned from a command, unless that is an `Error`.

With this, users can toggle the regular output off but still catch any errors thrown by commands so we don't accidentally hide problems. Messages logged directly by a command (assuming no normal flow messages are logged that way) are still output normally.

### DIFF
--- a/packages/cli/src/CLI.ts
+++ b/packages/cli/src/CLI.ts
@@ -26,7 +26,6 @@ export class CLI implements Command {
       '--early-access': Boolean,
       '--telemetry-information': String,
       '--quiet': Boolean,
-      '--silent': '--quiet',
     })
 
     if (isError(args)) {
@@ -43,8 +42,12 @@ export class CLI implements Command {
       return this.help()
     }
 
+    const nonQuietCommands = ['init', 'studio', 'validate', 'version', 'debug']
+    const isQuietMode = args['--quiet']
+
     // check if we have that subcommand
     const cmdName = args._[0]
+
     // Throw if "lift"
     if (cmdName === 'lift') {
       throw new Error(`${red('prisma lift')} has been renamed to ${green('prisma migrate')}`)
@@ -58,9 +61,9 @@ export class CLI implements Command {
         )}`,
       )
       logger.warn('')
+    } else if (isQuietMode && nonQuietCommands.includes(cmdName)) {
+      throw new Error(`The ${cmdName} command does not support --quiet`)
     }
-
-    const isQuietMode = args['--quiet']
 
     const cmd = this.cmds[cmdName]
     if (cmd) {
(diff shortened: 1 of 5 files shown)


SevInf (changes requested):
Hi @paulomarg 
Thank you for submitting the PR. `--quiet` flag makes sense, however, I am pretty sure this PR won't quite achieve the functionality you want - unfortunately, we are pretty bad at centralizing our logging and there are bunch of other places within the CLI where rogue `console.log` might slip in. 
If you want to do an attempt at centralized logging, we'll take it. However, that might be a lot of work and I think there is way for us to proceed with this PR - just that before we can merge it, we should at least be sure that we won't make this logging situation worse in the future. And for that we'd need to snapshot output from real commands, not the fake one. You can find the examples of how to do it in `cli/src/__tests__` and `migrate/src/__tests__`.

Additionally, small nitpick - I don't think there is a good reason to have both `--silent` and `--quiet` flag, I'd prefer we'd just keep `--quiet` for now.

Thank you again for this PR and let me know if you need any help.

paulomarg:
Hey, thanks for getting back to me! That was a good callout, I added some test that would check the actual output and it does seem like a lot of it isn't coming from the command return value (see the new snapshots), so my proposal was too shallow.

I agree that the right solution would have to be centralizing all of the logging and using the flag to turn _that_ off instead, but unfortunately I won't have the time to do that :(

Thanks for taking the time to review, and I hope this PR helps whatever solution you end up implementing! I'm going to close it since it doesn't feel like it'll solve the problem.

SevInf:
No worries, @paulomarg, thank you for submitting the PR anyway. It won't be able to be merged is more of our failure than yours, to be honest, so, my apologies about that.

davecarlson:
@SevInf , surely "some" fuctionality is better than a perfect solution for now. It can always be adjusted and improved in the future. 

as per https://github.com/prisma/prisma/issues/23192#issuecomment-2097636494

SevInf:
@davecarlson I was not the one who closed the PR, but if you look at updated snapshots here, it is pretty far even from a baseline — `generate` message you are mentioning in the other issue is not affected at all, as an example, and it is still there.  For that specific case though, 5.16.0 will have `--no-hints` flag.

davecarlson:
sweet, thank you . that hopefully will do :)

paulomarg:
I agree with @SevInf, unfortunately this PR just didn't do what it intended, since we still had a lot of output going out even with the flag. I don't think it should be merged in the state it was in as having the flag omit only _some_ messages would probably make it worse.

PR #24269, opened 2024-05-23
fix(pg / neon/ PS): Do not parse JSON - send it directly to engines

### LINKED ISSUE #23926: `Inconsistent column data: Unexpected conversion failure from Number to BigInt` error when using `@prisma/adapter-pg`
Hi all, I've hit an issue when trying out the `@prisma/adapter-pg` preview feature. 

I'm using the latest `v5.13.0` and I can resolve the issue by removing the `"relationJoins"` from the preview features

Here's the error I'm seeing in the console:
```
  20     include: { user: true },
  21 };
  22 
→ 23 const result = await this.#prisma.baseUser.findUnique(
Inconsistent column data: Unexpected conversion failure for field User.intId from Number(1374511782084.0) to BigInt.
    at In.handleRequestError (/**/node_modules/@prisma/client/runtime/library.js:122:6854)
    at In.handleAndLogRequestError (/**/node_modules/@prisma/client/runtime/library.js:122:6188)
    at In.request (/**/node_modules/@prisma/client/runtime/library.js:122:5896)
    at l (/**/node_modules/@prisma/client/runtime/library.js:127:11167)
    at PrismaUserRepository.getUnsafe (/**/src/infrastructure/repository/v2/prisma/prisma-user.repository.ts:23:24)
```

Here's roughly how I'm calling Prisma:
```ts
async function getUnsafe(id: string | bigint): Promise<User | null> {
    const query = {
        where: typeof id === 'bigint' ? { intId: id } : { id },
        include: { user: true },
    };

    const result = await prisma.baseUser.findUnique(query);
    if (!result) {
        return null;
    }

    return toUser(result);
}

type PrismaBaseUser = Prisma.BaseUserGetPayload<{ include: { user: true } }>;

function toUser(user: PrismaBaseUser): User {
    return {
        id: user.id,
        intId: user.intId,
        name: user.user.name,
        email: user.user.email,
        avatarUrl: user.user.avatar,
    };
}
```

Here's a subset of my schema:
```prisma
generator client {
  previewFeatures = ["relationJoins"]
}

datasource db {
  provider = "postgresql"
}

model BaseUser {
  id                String       @id @db.Uuid
  intId             BigInt       @unique
  createdAt         DateTime     @default(now())
  updatedAt         DateTime     @updatedAt
  favorites         Repository[] @relation("UserFavorites")
  ownedRepositories Repository[]
  user              User         @relation("userById", fields: [id], references: [id])
  userByIntId       User         @relation("userByIntId", fields: [intId], references: [intId])

  // Indexing intId for faster lookups
  @@index([intId(ops: Int8BloomOps)], type: Brin)
}

model User {
  id              String    @id @db.Uuid
  correlationId   String    @unique @db.Uuid
  intId           BigInt    @unique
  name            String
  email           String?
  locale          String?
  avatar          String?
  createdAt       DateTime  @default(now())
  updatedAt       DateTime
  deletedAt       DateTime?
  baseUser        BaseUser? @relation("userById")
  baseUserByIntId BaseUser? @relation("userByIntId")

  // Indexing intId for faster lookups
  @@index([intId(ops: Int8BloomOps)], type: Brin)
  // Indexing email for faster lookups
  @@index([email], type: Hash)
}
```

Druue:
We ran into an issue where the handling for JSON parsing was inconsistent for bigints nested inside of JSON relations:

This is the result of a native PG query. Note how the outer value is a string as expected but the inner is a number which would yield conversion errors
image

As we now send handle JSON parsing in engines, rust is fully capable of handling `i64` values and can therefore parse it correctly leading to correct bigint values

```js
{
  id: 'a0eeea10-7911-4af2-0000-01400b5718d2',
  intId: 1374511782084n,
  user: {
    id: 'a0eeea10-7911-4af2-0000-01400b5718d2',
    intId: 1374511782084n,
    name: 'Christopher Addison'
  }
}
```

fix https://github.com/prisma/prisma/issues/23926
related https://github.com/prisma/prisma-engines/pull/4883

Note to be aware of:
- If we end up supporting further driver adapters in the future that support `JSON` but do not allow for custom type parsing as PG, Neon, and PS do, we may run into issues


packages/driver-adapter-utils/src/index.ts
```diff
@@ -1,5 +1,5 @@
 export { bindAdapter } from './binder'
-export { ColumnTypeEnum, JsonNullMarker } from './const'
```
SevInf, line 2: I think `JsonNullMarker` can be removed completely

Druue: huh, I thought I removed this

SevInf (approved):
LGTM, when engine update makes it into the repo and we see green tests

SevInf:
Should be good to merge once green

PR #24273, opened 2024-05-23
chore: Upgrade ESLint to v9

nzakas:
This is an attempt to upgrade the repo to ESLint v9.

I've verified that the new config file works and ESLint runs. However, there are some outstanding ESLint errors that need review from someone familiar with the codebase. I believe these are valid, but they didn't show up before the migration, so it's possible you may want to ignore them.

```text
C:\Users\nzaka\projects\third-party\prisma\helpers\blaze\omit.ts      
  11:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression
  17:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression

C:\Users\nzaka\projects\third-party\prisma\helpers\blaze\pick.ts      
  11:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression
  17:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression

C:\Users\nzaka\projects\third-party\prisma\packages\nextjs-monorepo-workaround-plugin\index.js
  121:57  error  'from' is defined 
but never used. Allowed unused args must match /^_/u  @typescript-eslint/no-unused-vars
```


eslint.config.cjs
```diff
@@ -0,0 +1,166 @@
...
+    'plugin:@typescript-eslint/recommended-requiring-type-checking',
+    'plugin:prettier/recommended',
+    'plugin:jest/recommended',
+  ),
+  {
+    plugins: {
+      '@typescript-eslint': typescriptEslint,
+      jest,
+      'simple-import-sort': simpleImportSort,
+      import: fixupPluginRules(_import),
```
Jolg42, line 67: Oh! That's interesting, because I identified this as a blocker recently and this unblocks it.

(In https://github.com/prisma/prisma/pull/24148#pullrequestreview-2052063099)

Jolg42:
Thanks a lot @nzakas! 🙌🏼 

I think we can ignore these errors for the `blaze` utils and rename `from` to `_from` for `nextjs-monorepo-workaround-plugin`.

Could you do that? I could not push a commit because of a permission error.

Also, we would need to have the CLA signed to be able to merge the PR later. Could you check this [in the message above](https://github.com/prisma/prisma/pull/24273#issuecomment-2127796996)?

Ping @millsp for info and for the `blaze` utils where we need to ignore the eslint errors.

millsp:
Thanks @nzakas for taking the time, this is very much appreciated.

@Jolg42 @nzakas  these eslint errors should be ignored as it is intentional in the implementation.
```
C:\Users\nzaka\projects\third-party\prisma\helpers\blaze\omit.ts      
  11:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression
  17:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression

C:\Users\nzaka\projects\third-party\prisma\helpers\blaze\pick.ts      
  11:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression
  17:45  error  Unexpected constant truthiness on the left-hand side of a `&&` expression  no-constant-binary-expression
```
and the variable producing an error here can be prefixed with an underscore to suppress the error.
```
C:\Users\nzaka\projects\third-party\prisma\packages\nextjs-monorepo-workaround-plugin\index.js
  121:57  error  'from' is defined 
but never used. Allowed unused args must match /^_/u  @typescript-eslint/no-unused-vars
```

nzakas:
CLA signed.

Updated the config file to ignore the `blaze` directory and renamed `from` to `_from`.

eslint.config.cjs
```diff
@@ -0,0 +1,167 @@
...
+    ],
+  },
+  ...compat.extends(
+    'eslint:recommended',
+    'plugin:@typescript-eslint/eslint-recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:@typescript-eslint/recommended-requiring-type-checking',
+    'plugin:prettier/recommended',
+    'plugin:jest/recommended',
+  ),
```
gutyerrez, line 62: I see only one problem in this part. eslintrc will be long not supported and not maintened again. The compat "transforms" the configuration to support eslintrc old configuration. In this case i really want the best thing is to adapt fully to new eslint format.

eslint.config.cjs
```diff
@@ -0,0 +1,167 @@
...
+  recommendedConfig: js.configs.recommended,
+  allConfig: js.configs.all,
+})
+
+const project = path.resolve(__dirname, 'tsconfig.json')
+
+module.exports = [
+  {
+    files: ['**/*.ts'],
+  },
```
gutyerrez, line 27: With the v9 release eslint give's a way to define rules for specific file extensions. in this block you have only a file extensions without inserting rules and others things. Are that really working good?

eslint.config.cjs
nzakas: That's a good approach in the long run. For right now, there are still too many plugins that don't officially support the new config format, so we need this. You can always revisit later as plugins are updated.

eslint.config.cjs
nzakas: Yes. This is telling ESLint to search for `**/*.ts` files by default. This replaces the `--ext .ts` command line flag. You can optionally include additional configuration if you want, but the primary focus of this entry is to ensure that TypeScript files are search for by default.

eslint.config.cjs
gutyerrez: TypeScript eslint now support Eslint v9. Here's a explain https://github.com/typescript-eslint/typescript-eslint/issues/8211#issuecomment-2106427365

Prettier too have a Eslint v9 support. Here's a merged PR https://github.com/prettier/eslint-plugin-prettier/pull/616

And you can see all supported parsers/plugins in the issue [18391](https://github.com/eslint/eslint/issues/18391)

gutyerrez (approved):

Jolg42 (approved):
Thanks a lot 🙌🏼

PR #24274, opened 2024-05-23
feat(cli): `generate`, add `--no-hint` argument 

### LINKED ISSUE #22513: Disable "Start using Prisma Client" hint logs on `prisma generate`
## Problem

This log is appended to the log of every `prisma generate` call:

```
Start using Prisma Client in Node.js (See: https://pris.ly/d/client)
\`\`\`
\`\`\`
or start using Prisma Client at the edge (See: https://pris.ly/d/accelerate)
\`\`\`
import { PrismaClient } from '@prisma/client/edge'
\`\`\`
```

...and there is no way to turn it off, even in config. In mature repositories I have alreadyb configured Prisma Client and am confident in how to use it, and this just spams the logs.

## Suggested solution / Alternatives

A configuration option somewhere that disables this. Either:

- setting an environment variable e.g. `PRISMA_DISABLE_HINT`
- an option in `package.json` under the `"prisma": { "hint": false }` key
- a flag to `prisma generate` e.g. `prisma generate --no-hint`
- some configuration option in `schema.prisma`

...or something similar. As far as I can tell, Prisma has no standard config file (e.g. `prisma.config.json` or `.prismarc`, which would otherwise be the optimal place to put something like this.

## Additional context

The code that logs the text question is located in `packages/cli/src/Generate.ts`:
https://github.com/prisma/prisma/blob/1eacb61c495b2e82ba8d8c77e2189c3df632a335/packages/cli/src/Generate.ts#L263-L283

nickcarnival:
This PR aims to remove some of the bloat that's output everytime you run `prisma generate` by adding a `--no-hint` flag to the `generate` command:

Closes #22513

```sh
prisma generate --no-hint
```

Screenshot 2024-05-23 at 12 50 34 PM

```sh
prisma generate
```

Screenshot 2024-05-23 at 12 50 04 PM

I've also created a sister PR in the `withfig/autocomplete` repo: https://github.com/withfig/autocomplete/pull/2345

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -56,7 +56,7 @@ ${bold('Options')}
              --watch   Watch the Prisma schema and rerun after a change
          --generator   Generator to use (may be provided multiple times)
          --no-engine   Generate a client for use with Accelerate only
-         --no-hints    Hides the hint messages but still outputs errors and warnings
+         --no-hints    Hides the hint messages but still outputs errors, warnings and update available messages
    --allow-no-models   Allow generating a client without models
 
 ${bold('Examples')}
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -249,7 +249,7 @@ When using Deno, you need to define \`output\` in the client generator section o
 ${breakingChangesMessage}`
           : ''
 
-        const hideHintMode = args['--no-hints'] || false
+        const hideHints = args['--no-hints'] ?? false
 
         const versionsOutOfSync = clientGeneratorVersion && pkg.version !== clientGeneratorVersion
         const versionsWarning =
@@ -261,8 +261,9 @@ This might lead to unexpected behavior.
 Please make sure they have the same version.`
             : ''
 
-        if (hideHintMode) hint = `${getHardcodedUrlWarning(config)}${breakingChangesStr}${versionsWarning}`
-        else {
+        if (hideHints) {
+          hint = `${getHardcodedUrlWarning(config)}${breakingChangesStr}${versionsWarning}`
+        } else {
           const tryAccelerateMessage = `Deploying your app to serverless or edge functions?
 Try Prisma Accelerate for connection pooling and caching.
 ${link('https://pris.ly/cli/--accelerate')}`
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -56,7 +56,7 @@ ${bold('Options')}
              --watch   Watch the Prisma schema and rerun after a change
          --generator   Generator to use (may be provided multiple times)
          --no-engine   Generate a client for use with Accelerate only
-        --no-hint     Hides the hint messages but still outputs errors and warnings
+         --no-hint     Hides the hint messages but still outputs errors and warnings
    --allow-no-models   Allow generating a client without models
 
 ${bold('Examples')}
@@ -107,7 +107,6 @@ ${bold('Examples')}
       '--data-proxy': Boolean,
       '--accelerate': Boolean,
       '--no-engine': Boolean,
-      // https://github.com/prisma/prisma/issues/22513
       '--no-hint': Boolean,
       '--generator': [String],
       // Only used for checkpoint information
@@ -161,6 +160,7 @@ ${bold('Examples')}
           Boolean(process.env.PRISMA_GENERATE_ACCELERATE) || // legacy, keep for backwards compatibility
           Boolean(process.env.PRISMA_GENERATE_NO_ENGINE),
         allowNoModels: Boolean(args['--allow-no-models']),
+        noHints: Boolean(args['--no-hints']),
       })
 
       if (!generators || generators.length === 0) {
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -56,7 +56,7 @@ ${bold('Options')}
              --watch   Watch the Prisma schema and rerun after a change
          --generator   Generator to use (may be provided multiple times)
          --no-engine   Generate a client for use with Accelerate only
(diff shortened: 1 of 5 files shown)


packages/cli/src/Generate.ts
```diff
@@ -48,12 +48,12 @@ ${bold('Usage')}
   ${dim('$')} prisma generate [options]
 
 ${bold('Options')}
-
           -h, --help   Display this help message
             --schema   Custom path to your Prisma schema
              --watch   Watch the Prisma schema and rerun after a change
          --generator   Generator to use (may be provided multiple times)
          --no-engine   Generate a client for use with Accelerate only
+        --no-hint     Hides the hint messages but still outputs errors and warnings
```
gutyerrez, line 56:
```suggestion
         --no-hint     Hides the hint messages but still outputs errors and warnings
```

packages/cli/src/Generate.ts
```diff
@@ -104,6 +104,8 @@ ${bold('Examples')}
       '--data-proxy': Boolean,
       '--accelerate': Boolean,
       '--no-engine': Boolean,
+      // https://github.com/prisma/prisma/issues/22513
```
gutyerrez, line 107: This is really relevant?

gutyerrez (approved):

packages/cli/src/Generate.ts
nickcarnival: 56bde3bd4ca1f716cf6c337f73c7ce5ec39577b7

packages/cli/src/Generate.ts
nickcarnival: 56bde3bd4ca1f716cf6c337f73c7ce5ec39577b7

nickcarnival:
@gutyerrez I've addressed your comments and the CI tests are now passing. It seems like someone has to add a milestone before the last check will pass but other than that it should be good to go :)

nickcarnival:
@Jolg42 I see that you removed the milestone. Is there something I need to do to get this added back?
It seems like that's the only check left for this PR.

packages/cli/src/platform/auth/login.ts
nickcarnival, line 1: This change was made via `pnpm run lint`

packages/client/tests/e2e/prisma-client-imports/src/default.ts
nickcarnival, line 1: This change was made via `pnpm run lint`

packages/client/tests/functional/omit/test.ts
nickcarnival, line 1: This change was made via `pnpm run lint`

packages/get-platform/src/test-utils/index.ts
nickcarnival, line 1: This change was made via `pnpm run lint`

nickcarnival:
@SevInf is there anything more I need to do to get this merged?

packages/cli/src/Generate.ts
```diff
@@ -154,6 +155,7 @@ ${bold('Examples')}
           Boolean(process.env.PRISMA_GENERATE_DATAPROXY) || // legacy, keep for backwards compatibility
           Boolean(process.env.PRISMA_GENERATE_ACCELERATE) || // legacy, keep for backwards compatibility
           Boolean(process.env.PRISMA_GENERATE_NO_ENGINE),
+        noHints: Boolean(args['--no-hints']),
```
SevInf, line 158: Why forward this to generators? At no point this value actually get used right now.

packages/cli/src/Generate.ts
```diff
@@ -309,6 +315,8 @@ More information: https://pris.ly/d/client`
 ${boxedTryAccelerateMessage}
 ${getHardcodedUrlWarning(config)}${breakingChangesStr}${versionsWarning}`
         }
+        const hideHintMode = args['--no-hints'] || false
```
SevInf, line 318: Right now, the code still goes through all of the trouble of creating hint message just to throw it away at the last moment if `--no-hint` is set. Can we instead refactor the code so that those code paths would be completely skipped in `--no-hints` case?

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -365,6 +365,60 @@ describe('using cli', () => {
     }
   })
 
+  it('should hide hints with --no-hints', async () => {
```
SevInf, line 368: Thanks for adding the test!

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -365,6 +365,60 @@ describe('using cli', () => {
...
+    if (engineType === ClientEngineType.Library) {
+      expect(data.stdout).toMatchInlineSnapshot(`
+      "Prisma schema loaded from prisma/schema.prisma
+
+      ✔ Generated Prisma Client (v0.0.0) to ./generated/client in XXXms
+      "
+    `)
+    }
+
+    if (engineType !== ClientEngineType.Library && engineType !== ClientEngineType.Binary) {
```
SevInf, line 396: This can never happen, engine type is always either binary or library. Where does this snapshot come from?

SevInf (changes requested):
Hi @nickcarnival!
Thank you for submitting the PR. I've left a few comments, after my feedback is addressed this should be good to merge.

packages/cli/src/Generate.ts
nickcarnival: I added another level of indentations and a large else block that accomplishes this 😆

packages/cli/src/Generate.ts
nickcarnival: It's gone now, thanks for catching this.

nickcarnival:
@SevInf thanks so much for the review, I believe I've addressed all of your comments.

SevInf (approved):

SevInf:
Thank you, waiting for final CI run and merging

PR #24297, opened 2024-05-27
feat(migrate): Support split schema files in migrate

SevInf:
Companion to https://github.com/prisma/prisma-engines/pull/4874/

Close prisma/team-orm#1040


packages/internals/src/cli/getSchema.ts
```diff
@@ -24,19 +24,19 @@ export type GetSchemaResult = {
  * Async
  */
 
-export async function getSchemaPath(): Promise<GetSchemaResult | null>
```
SevInf, line 27: Better name for what function returns now, imho.

packages/internals/src/cli/getSchema.ts
```diff
@@ -355,46 +309,6 @@ export async function getSchema(schemaPathFromArgs?: string): Promise<MultipleSc
   return schemaPathResult.schemas
 }
 
-/**
- * Sync.
- * TODO: this is only used by `Migrate` to synchronously read the schema path from its (sync) constructor.
- */
-
-export function getSchemaPathSync(schemaPathFromArgs?: string): string | null {
```
SevInf, line 363: No longer the case and we don't need it all `Sync` methods anymore.

packages/migrate/src/Migrate.ts
```diff
@@ -18,9 +22,6 @@ export class Migrate {
   private schemaPath?: string
   public migrationsDirectoryPath?: string
 
-  // TODO: make the constructor private, and only allow instantiation via an async static method.
```
SevInf, line 21: I don't think it is neccessary anymore.

packages/migrate/src/__tests__/DbPull/postgresql-multischema.test.ts
```diff
@@ -88,7 +88,7 @@ describe('postgresql-multischema', () => {
       "Prisma schema validation - (get-config wasm)
       Error code: P1012
       error: If provided, the schemas array can not be empty.
-        -->  schema.prisma:4
+        -->  with-schemas-in-datasource-0-value.prisma:4
```
SevInf, line 91: Similar to Alberto's PR for CLI before - this is actually a fix, `schema.prisma` was a hardcoded name before, now we use actual file names.

packages/migrate/src/types.ts
```diff
@@ -186,27 +176,18 @@ export namespace EngineArgs {
...
-    tag: 'schemaDatasource'
-    schema: string
-  }
+  type MigrateDiffTargetSchemaDatamodel = MigrateTypes.Tagged<'schemaDatamodel', MigrateTypes.SchemasContainer>
+  type MigrateDiffTargetSchemaDatasource = MigrateTypes.Tagged<'schemaDatasource', MigrateTypes.SchemasWithConfigDir>
+  //   // The path to a Prisma schema.
+  //   // The datasource url will be considered, and the live database it points to introspected for its schema.
+  //   tag: 'schemaDatasource'
+  //   schema: MigrateTypes.SchemasWithConfigDir
+  // }
```
SevInf, line 190:
```suggestion
```
dead code

SevInf:
Note: e2e test fails because it does not pick up correct engine `/engine-branch` command

packages/cli/src/__tests__/artificial-panic.test.ts
```diff
@@ -1,13 +1,17 @@
...
 import { DbPull } from '@prisma/migrate'
 
 import { Format } from '../Format'
 import { Validate } from '../Validate'
 
 const ctx = jestContext.new().assemble()
 
+function sanitizeSchemas(schemas: MultipleSchemas | undefined): MultipleSchemas | undefined {
+  return schemas?.map(([path, content]) => [relativizePathInPSLError(path), content])
+}
```
jkomyno, line 13: Maybe add a comment about why we need to relative paths in PSL?
Iirc, it's to ensure the paths displayed in Prisma snapshot tests remain in "relative" form, because `prisma-engines` would treat them as absolute paths instead.

packages/migrate/src/utils/getDatabaseVersionSafe.ts
```diff
@@ -1,15 +1,17 @@
...
 
 const debug = Debug('prisma:cli')
 
 /**
  * Retrieve the database version from the given schema or url.
  * This function never throws, and was introduced to prevent circular dependencies in `@prisma/internals`.
  */
-export async function getDatabaseVersionSafe(args: EngineArgs.GetDatabaseVersionParams): Promise<string | undefined> {
+export async function getDatabaseVersionSafe(
+  args: MigrateTypes.GetDatabaseVersionParams | undefined,
```
jkomyno, line 13: Why the `undefined` here?

SevInf: `| undefined` was a part of `EngineArgs.GetDatabaseVersionParams` definition before already.
Since that does not quite agree with what `schema-engine` expects, i've changed the definition and so we had to add `| undefined` to a couple of signatures where we actually pass undefined values every once in a while

jkomyno (approved):

packages/migrate/src/utils/replaceDatasource.ts
```diff
@@ -0,0 +1,17 @@
...
+
+function replaceDatasourceSingle(newDatsource: string, content: string) {
+  const noDatasource = removeDatasource(content)
+  if (content === noDatasource) {
+    // no datasource in this file
+    return content
+  }
+
+  return `${newDatsource}\n${noDatasource}`
+}
```
Jolg42, line 17:
```suggestion
function replaceDatasourceSingle(newDatasource: string, content: string) {
  const noDatasource = removeDatasource(content)
  if (content === noDatasource) {
    // no datasource in this file
    return content
  }

  return `${newDatasource}\n${noDatasource}`
}
```

PR #24318, opened 2024-05-28
fix(migrate): Introspection of schemas with no datasource

SevInf:
It was broken since #24297. Reason for that: for introsepction from url,
we remove existing datasource from the schema and add a new, fake one.
While working on multifile schemas support, i've introduce
`replaceDatasource` function for that. It did pretty much what it said:
found the file which had a datasource defined and replaced that with a
new one. However, if it did not find one, it would not add a new one -
so if schema had no datasource at all, it will stay invalid and schema
engine will boot us off.
Some people might've noticed that it worked correctly when there were
empty line breaks at the end of the schema file. That's a case of two
bugs leading to somewhat correct result: `replaceDatasource` function
relied on `content !== removeDatasource(content)` to determine if the
datasource was indeed removed from a file. And besides remiving the
datasource, it turns out, `removeDatasource` also trimmed newlines.


Jolg42:
I think the tests are failing now after the removal of `trim()` @SevInf

SevInf:
@Jolg42 I realized `removeOrAddDatasource` is still a bit flimsy and error-prone and reworked it, completely removing `removeDatasource` and `trimBlocksFromSchema`

packages/migrate/src/utils/replaceOrAddDatasource.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  let endLine = -1
+  for (let index = startLine; index < lines.length; index++) {
+    const lineTrimmed = lines[index].trim()
+    if (lineTrimmed.endsWith('}') && !lineTrimmed.startsWith('//')) {
+      endLine = index
+      break
+    }
+  }
```
Jolg42, line 66: I find this logic brittle, but since it was already like this before, it's ok for now.
It's assuming the file is formatted in a certain way, if there is one more spaces before `//` or after `{`/`}` it would break the assumption.

Jolg42 (approved):
Thanks for the fix and the cleanup!

packages/migrate/src/utils/replaceOrAddDatasource.ts
SevInf: We trim the line, so spaces are no problem here. But I agree, ideally we'd use `schema-wasm` and real parser for all kinds of such manipulations.

Jolg42: Trimming does help a lot for sure.
There are some cases where this could break still but a lot less likely (like a comment on a line `{ // something`)

PR #24378, opened 2024-05-31
feat(schema-files-loader): Support subfolders

SevInf:
Logic is the following:
- When asked to load files from folder, it is simple - we just
  recursively walk the tree and load all `.prisma` files.
- When asked to find all files belonging to the same schema, we walk up
  the directory tree and stop and the last directory that had any
  `.prisma` files - this directory will be the root of the schema.

So:

```
first
├── a.prisma
└── b.prisma
second
├── a.prisma
└── b.prisma
```

`first` and `second` describe different schema.

```
root.prisma
first
├── a.prisma
└── b.prisma
second
├── a.prisma
└── b.prisma
```

all files belong to the same schema.

### DIFF
--- a/packages/schema-files-loader/src/__fixtures__/subfolder/a.prisma
+++ b/packages/schema-files-loader/src/__fixtures__/subfolder/a.prisma
@@ -1,1 +1,5 @@
 // this is a
+generator client {
+    provider = "prisma-client-js"
+    previewFeatures = ["prismaSchemaFolder"]
+}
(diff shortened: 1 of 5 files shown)


Jolg42:
Note: Linting error can be ignore, I already fixed it on `main`

Jolg42 (approved):

SevInf:
Ok, I am going to merge this now, but if we find anything wrong with this feature before release we should revert it rather than rush a fix - it's nothing that can't be postponed until next release.

PR #24380, opened 2024-05-31
temp: Playing around to understand QE lifecycle

janpio:
- Do not call Node-API library for specific methods, run some TS code instead
	- Make `client` in driver adapters accessible
	- Store `adapter` in `LibraryEngine` (if present)
	- New `sandbox/node-engine` that is configured the right way to test this
	- New `tests/function/node-engine` that tests all the individual functionality that was already implemented (equivalent to sandbox, but automated to catch regressions)

Implemented functionality:
- `findMany()` with no further arguments
- `findMany()` with `_count` selection (1-n and m-n, one or multiple)


packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -455,16 +457,38 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
...
+          try {
+            const result = await this.adapter.client.execute(query)
+            return JSON.stringify(result.rows)
+          } catch (error) {
+            throw new Error(error)
+          }
+        })()
+      } else {
+        this.executingQueryPromise = this.engine?.query(queryStr, headerStr, interactiveTransaction?.id)
+      }
```
janpio, line 485: Minimal NodeEngine 🧌

Issue #23505, opened 2024-03-15
`@prisma/adapter-pg` modifies node-postgres global type parsers


matteosb:
### Bug description

Using the Prisma `node-postgres` adapter breaks queries run directly through `node-postgres` `Pool` and `Client` instances, even when the instance of `Pool` is not shared with the Prisma adapter. This is a problem for us because we'd like to use Prisma with the adapter, but we use `node-postgres` extensively, including for use-cases that Prisma doesn't support (streaming in particular).

Fortunately I think this should be very easy to fix: @prisma/adapter-pg registers its type parsers globally [here](https://github.com/prisma/prisma/blob/main/packages/adapter-pg/src/conversion.ts). It could instead pass the type parsers to its queries explictly as described [here](https://node-postgres.com/features/queries#types).

### How to reproduce

1. import `@prisma/adapter-pg`
2. Run a query via node-postgres 

This script illustrates the problem with timestamp parsing and doesn't require a Prisma schema:

```ts
import { Pool } from "pg";

const pgPool = new Pool({ connectionString: process.env.DATABASE_URL });

async function main() {
  const before = await pgPool.query(`SELECT NOW() as ts`);
  console.log(
    "NOW() before @prisma/adapter-pg - value:",
    before.rows[0].ts,
    "type:",
    typeof before.rows[0].ts
  );

  require("@prisma/adapter-pg");

  const after = await pgPool.query(`SELECT NOW() as ts`);
  console.log(
    "NOW() after @prisma/adapter-pg - value:",
    after.rows[0].ts,
    "type:",
    typeof after.rows[0].ts
  );
}

main().finally(() => pgPool.end());
```

output: 
```
NOW() before @prisma/adapter-pg - value: 2024-03-15T15:48:47.231Z type: object
NOW() after @prisma/adapter-pg - value: 2024-03-15 15:48:47.237668 type: string
```

### Expected behavior

`@prisma/adapter-pg` shouldn't modify `node-postgres`'s global type parsers. Instead, [performIO](https://github.com/prisma/prisma/blob/main/packages/adapter-pg/src/pg.ts#L88) could pass Prisma's custom type parsers as described in the [node-postgres documentation](https://node-postgres.com/features/queries#types).

### Prisma information

This can be reproduced without any Prisma schema

### Environment & setup

- OS: tested on macOS and Debian
- Database: PostgreSQL
- Node.js version: 20.11.0

### Prisma Version

```
prisma                  : 5.10.2
[...]

PR #24434, opened 2024-06-05
fix(adapter-pg/pg-worker/neon): pass customParsers to the query

Jolg42:
Related to https://github.com/prisma/prisma/issues/23505

### DIFF
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -18,6 +18,7 @@ const types = pg.types
 
 import { name as packageName } from '../package.json'
 import { fieldToColumnType, fixArrayBufferValues, UnsupportedNativeDataType, customParsers } from './conversion'
+import { Type } from 'ts-toolbelt/out/Any/Type'
 
 const debug = Debug('prisma:driver-adapter:pg')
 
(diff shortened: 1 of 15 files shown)


Jolg42:
TODO: check adapter-neon (& others)

packages/adapter-pg/src/pg.ts
```diff
@@ -87,7 +89,31 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements Quer
...
 
     try {
-      const result = await this.client.query({ text: sql, values: fixArrayBufferValues(values), rowMode: 'array' })
+      const result = await this.client.query(
+        {
+          text: sql,
+          values: fixArrayBufferValues(values),
+          rowMode: 'array',
+          types: {
+            // Error: packages/adapter-pg build: src/pg.ts(98,13): error TS2769: No overload matches this call.
```
SevInf, line 98: Would adding explicit type annotation to `customParsers` fix it?
`Record<number, types.TypeParser<string, unknown>>` should be the correct type.
Alternatively, maybe `as const` after `customParsers` declaration?

SevInf: Ah, I see the issue. Apparantly, `pg-types` types expect you to handle both binary and text protocol where versions, where as far as i can see, `pg` will ever pass only text version. We can put that into the comment instead, and then add assertion that `format === 'text'` before reaching out for our parsers.

Jolg42: I did the update

SevInf (approved):

PR #24437, opened 2024-06-05
refactor(client): Move non-aggregate methods generation to ts-builder

SevInf:
Will make future `globalOmit` changes easier.


packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
```diff
@@ -2918,10 +2916,8 @@ export namespace Prisma {
      *     // ... provide filter here
      *   }
      * })
-    **/
-    findUniqueOrThrow<T extends PostFindUniqueOrThrowArgs<ExtArgs>>(
-      args?: SelectSubset<T, PostFindUniqueOrThrowArgs<ExtArgs>>
```
SevInf, line 2923: `args` being optional for `findUniqueOrThrow` was actually a bug before

Jolg42 (approved):
Nice that the type files lost some empty lines!

PR #24448, opened 2024-06-06
feat(cli): add Driver Adapters and underlying drivers to `prisma -v`

jkomyno:
This PR closes https://github.com/prisma/team-orm/issues/379.

It modifies `prisma -v` so that:
- (Well-known, `@prisma/adapter-*`) Driver Adapters' versions are read and displayed
- For each Driver Adapter, the underlying driver and its version is also displayed

Driver Adapters are not shown when using the binary Query Engine.

---

Consider the following `package.json`:

```json
{
  "dependencies": {
    "@prisma/adapter-d1": "file:../adapter-d1",
    "@prisma/adapter-neon": "file:../adapter-neon",
    "@prisma/adapter-pg": "file:../adapter-pg",
    "@prisma/adapter-planetscale": "file:../adapter-planetscale",
    "@prisma/adapter-not-recognized": "file:../adapter-not-recognized",
    "@prisma/non-adapter": "file:../non-adapter",
    "@prisma/client": "file:../client",
    "@neondatabase/serverless": "<NEON_VERSION>"
  },
  "devDependencies": {
    "wrangler": "<WRANGLER_DEV_VERSION>",
    "pg": "<PG_DEV_VERSION>"
  }
}
```

Then, `prisma -v` would now display:

```diff
prisma                      : 0.0.0
@prisma/client              : 0.0.0
+ @prisma/adapter-d1          : file:../adapter-d1
+  ↳ wrangler                 : <WRANGLER_DEV_VERSION>
+ @prisma/adapter-neon        : file:../adapter-neon
+  ↳ @neondatabase/serverless : <NEON_VERSION>
+ @prisma/adapter-pg          : file:../adapter-pg
+  ↳ pg                       : Not found
+ @prisma/adapter-planetscale : file:../adapter-planetscale
+  ↳ @planetscale/database    : Not found
Computed binaryTarget       : TEST_PLATFORM
Operating System            : OS
Architecture                : ARCHITECTURE
Node.js                     : NODEJS_VERSION
Query Engine (Node-API)     : libquery-engine ENGINE_VERSION (at sanitized_path/libquery_engine-TEST_PLATFORM.LIBRARY_TYPE.node, resolved by PRISMA_QUERY_ENGINE_LIBRARY)
Schema Engine               : schema-engine-cli ENGINE_VERSION (at sanitized_path/schema-engine-TEST_PLATFORM, resolved by PRISMA_SCHEMA_ENGINE_BINARY)
Schema Wasm                 : @prisma/prisma-schema-wasm CLI_VERSION.ENGINE_VERSION
Default Engines Hash        : ENGINE_VERSION
Studio                      : STUDIO_VERSION"
....
```

Notes:
- `@prisma/adapter-not-recognized` is not a valid / well-known Driver Adapter, so it's not included in the output.
- `@prisma/adapter-planetscale` is found, but its underlying driver, `@planetscale/database`, isn't. That's because it's not installed in `package.json`.
- `@prisma/adapter-pg` is found, but its underlying driver, `pg`, isn't. That's because it doesn't appear in `dependencies`.
- `@prisma/adapter-d1` is found, and its underlying driver, `wrangler`, is also found. This is the only case in which the underlying driver can be retrieved from `devDependencies` (`@prisma/client` needs these drivers at runtime).

### DIFF
--- a/packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
+++ b/packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
@@ -74,9 +74,9 @@ exports[`version version with driver adapters, it only shows official adapters a
 "prisma                      : 0.0.0
 @prisma/client              : 0.0.0
 @prisma/adapter-d1          : file:../adapter-d1
- ↳ wrangler                 : 3.5.9
+ ↳ wrangler                 : <WRANGLER_DEV_VERSION>
 @prisma/adapter-neon        : file:../adapter-neon
- ↳ @neondatabase/serverless : 0.9.3
+ ↳ @neondatabase/serverless : <NEON_VERSION>
 @prisma/adapter-pg          : file:../adapter-pg
  ↳ pg                       : Not found
 @prisma/adapter-planetscale : file:../adapter-planetscale
--- a/packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
+++ b/packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
@@ -55,3 +55,37 @@ Schema Wasm           : @prisma/prisma-schema-wasm CLI_VERSION.ENGINE_VERSION
 Default Engines Hash  : ENGINE_VERSION
 Studio                : STUDIO_VERSION"
 `;
+
+exports[`version version with driver adapters, it does not show any adapter 1`] = `
+"prisma                : 0.0.0
+@prisma/client        : 0.0.0
+Computed binaryTarget : TEST_PLATFORM
+Operating System      : OS
+Architecture          : ARCHITECTURE
+Node.js               : NODEJS_VERSION
+Query Engine (Binary) : query-engine ENGINE_VERSION (at sanitized_path/query-engine-TEST_PLATFORM, resolved by PRISMA_QUERY_ENGINE_BINARY)
+Schema Engine         : schema-engine-cli ENGINE_VERSION (at sanitized_path/schema-engine-TEST_PLATFORM, resolved by PRISMA_SCHEMA_ENGINE_BINARY)
+Schema Wasm           : @prisma/prisma-schema-wasm CLI_VERSION.ENGINE_VERSION
+Default Engines Hash  : ENGINE_VERSION
+Studio                : STUDIO_VERSION"
+`;
+
+exports[`version version with driver adapters, it only shows official adapters and their underlying drivers 1`] = `
+"prisma                      : 0.0.0
+@prisma/client              : 0.0.0
+@prisma/adapter-d1          : file:../adapter-d1
+ ↳ wrangler                 : 3.5.9
+@prisma/adapter-neon        : file:../adapter-neon
+ ↳ @neondatabase/serverless : 0.9.3
+@prisma/adapter-planetscale : file:../adapter-planetscale
+ ↳ @planetscale/database    : Not found
+Computed binaryTarget       : TEST_PLATFORM
+Operating System            : OS
+Architecture                : ARCHITECTURE
+Node.js                     : NODEJS_VERSION
+Query Engine (Node-API)     : libquery-engine ENGINE_VERSION (at sanitized_path/libquery_engine-TEST_PLATFORM.LIBRARY_TYPE.node, resolved by PRISMA_QUERY_ENGINE_LIBRARY)
+Schema Engine               : schema-engine-cli ENGINE_VERSION (at sanitized_path/schema-engine-TEST_PLATFORM, resolved by PRISMA_SCHEMA_ENGINE_BINARY)
+Schema Wasm                 : @prisma/prisma-schema-wasm CLI_VERSION.ENGINE_VERSION
+Default Engines Hash        : ENGINE_VERSION
+Studio                      : STUDIO_VERSION"
+`;
--- a/packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
+++ b/packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
@@ -77,6 +77,8 @@ exports[`version version with driver adapters, it only shows official adapters a
  ↳ wrangler                 : 3.5.9
 @prisma/adapter-neon        : file:../adapter-neon
  ↳ @neondatabase/serverless : 0.9.3
+@prisma/adapter-pg          : file:../adapter-pg
+ ↳ pg                       : Not found
 @prisma/adapter-planetscale : file:../adapter-planetscale
  ↳ @planetscale/database    : Not found
 Computed binaryTarget       : TEST_PLATFORM
--- a/packages/cli/src/utils/getClientVersion.ts
+++ b/packages/cli/src/utils/getClientVersion.ts
@@ -1,3 +1,9 @@
+import {
+  isOfficialDriverAdapter,
+  type OfficialDriverAdapters,
+  type OfficialUnderlyingDrivers,
+  underlyingDriverAdaptersMap,
+} from '@prisma/driver-adapter-utils'
 import fs from 'fs'
 import Module from 'module'
 import pkgUp from 'pkg-up'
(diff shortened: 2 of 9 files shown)


packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
```diff
@@ -55,3 +55,39 @@ Schema Wasm           : @prisma/prisma-schema-wasm CLI_VERSION.ENGINE_VERSION
...
+Studio                : STUDIO_VERSION"
+`;
+
+exports[`version version with driver adapters, it only shows official adapters and their underlying drivers 1`] = `
+"prisma                      : 0.0.0
+@prisma/client              : 0.0.0
+@prisma/adapter-d1          : file:../adapter-d1
+ ↳ wrangler                 : 3.5.9
+@prisma/adapter-neon        : file:../adapter-neon
+ ↳ @neondatabase/serverless : 0.9.3
```
janpio, line 79: These versions will not be stable?

jkomyno: Those are just example versions from the `package.json` file in the test fixture added by this PR. I can rename them to `<WRANGLER_VERSION>` and `<NEON_VERSION>`, for added clarity.

packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
```diff
@@ -55,3 +55,39 @@ Schema Wasm           : @prisma/prisma-schema-wasm CLI_VERSION.ENGINE_VERSION
...
+Schema Engine         : schema-engine-cli ENGINE_VERSION (at sanitized_path/schema-engine-TEST_PLATFORM, resolved by PRISMA_SCHEMA_ENGINE_BINARY)
+Schema Wasm           : @prisma/prisma-schema-wasm CLI_VERSION.ENGINE_VERSION
+Default Engines Hash  : ENGINE_VERSION
+Studio                : STUDIO_VERSION"
+`;
+
+exports[`version version with driver adapters, it only shows official adapters and their underlying drivers 1`] = `
+"prisma                      : 0.0.0
+@prisma/client              : 0.0.0
+@prisma/adapter-d1          : file:../adapter-d1
```
janpio, line 76: For users this will show the actual version of the adapter?

packages/driver-adapter-utils/src/underlying-drivers.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const underlyingDriverAdaptersMap = {
+  '@prisma/adapter-d1': 'wrangler',
+  '@prisma/adapter-libsql': '@libsql/client',
+  '@prisma/adapter-neon': '@neondatabase/serverless',
+  '@prisma/adapter-planetscale': '@planetscale/database',
+  '@prisma/adapter-pg': 'pg',
+  '@prisma/adapter-pg-worker': '@prisma/pg-worker',
+} as const satisfies Record<OfficialDriverAdapters, string>
```
janpio, line 26: Nit: Maybe combine these two lists to avoid the duplication?

janpio (dismissed):

packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
jkomyno: I've updated the tests and description.

packages/cli/src/__tests__/commands/__snapshots__/Version.test.ts.snap
jkomyno: Yes

packages/driver-adapter-utils/src/underlying-drivers.ts
jkomyno: Right, done

packages/cli/src/utils/getClientVersion.ts
```diff
@@ -33,6 +39,68 @@ async function getPrismaClientVersionFromNodeModules(cwd: string = process.cwd()
...
+ * Try reading the Prisma Driver Adapter versions from the local package.json.
+ * This also includes the underlying driver versions, if found. They are looked up
+ * among the `dependencies` only, with the exception of `wrangler` for `@prisma/adapter-d1`,
+ * which can also be a `devDependency`.
+ * Only official, well-known Driver Adapters are considered.
+ */
+export async function getPrismaDriverAdapterVersionsFromLocalPackageJson(
+  cwd: string = process.cwd(),
+): Promise<Array<DriverAdapterVersion> | null> {
+  cwd = cwd ?? process.cwd()
```
SevInf, line 61: `cwd` already has a default in the signature

packages/cli/src/utils/getClientVersion.ts
```diff
@@ -33,6 +39,68 @@ async function getPrismaClientVersionFromNodeModules(cwd: string = process.cwd()
...
+      devDependencies?: Record<string, string>
+    }
+
+    const adapterVersions = Object.entries((pkgJson.dependencies ?? {}) as Record<string, string>)
+      // Find all known Driver Adapter dependencies, usually `@prisma/adapter-*`:
+      .filter(([key]) => isOfficialDriverAdapter(key)) as Array<[OfficialDriverAdapters, string]>
+
+    const adapterVersionsWithUnderlyingDrivers = adapterVersions.map(([adapterName, adapterVersion]) => {
+      const underlyingDriver = underlyingDriverAdaptersMap[adapterName]
+      let underlyingDriverVersion = pkgJson.dependencies?.[underlyingDriver] ?? null
```
SevInf, line 80: So, that potentially will give us a version range, not the actual version: (`^5.0.0` can really be any version of Prisma 5). Can we try to require (`const require = createRequire(pkgJsonPath); require(adapterName).version` the package and fallback onto the range only if that fails?

packages/cli/src/utils/getClientVersion.ts
```diff
@@ -33,6 +39,67 @@ async function getPrismaClientVersionFromNodeModules(cwd: string = process.cwd()
...
+
+      // `wrangler` is the only underlying driver that can be just a `devDependency`.
+      if (underlyingDriverVersion === null && underlyingDriver === 'wrangler') {
+        underlyingDriverVersion = pkgJson.devDependencies?.[underlyingDriver] ?? null
+      }
```
Jolg42, line 85: We should look in `dependencies` too because people will sometimes put it there as well

packages/driver-adapter-utils/src/underlying-drivers.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * This module provides a map from the official Prisma Driver Adapter names to their underlying driver names.
+ * E.g., `@prisma/adapter-planetscale` -> `@planetscale/database`
+ */
+
+export const underlyingDriverAdaptersMap = {
```
SevInf, line 6: I think this module would make more sense as a part of `cli`, not `driver-adapter-utils`:
- `cli` is the only one using it and I don't see that changing anytime soon
- it's weird that generic library with building blocks and abastractions for specific drivers now has a list of all specific adapters hardcoded.

packages/cli/src/utils/getClientVersion.ts
jkomyno: Wouldn't the behavior you're suggesting be different than what `prisma version` currently does?
I agree that having the precise version would be better, but if this is a new kind of logic, perhaps that should be applied in a different PR (?)

SevInf: Is it different? Engines and prisma cli version are known to cli for sure and for `@prisma/client` version we actually try to require it in a similar way I am suggestion

packages/cli/src/utils/getClientVersion.ts
Jolg42: Right, then I would say the opposite 😄 

In general this makes sense.

But it's possible that people misplace the adapter packages and place them in `devDependencies`, and we should look in both places without assuming they would only be for sure in "devDependencies" or "dependencies".

Jolg42: TLDR: we better not have any assumption on where the package is placed

PR #24469, opened 2024-06-07
test(client): fix PlanetScale snapshot tests

jkomyno:
This PR fixes https://github.com/prisma/team-orm/issues/510.


janpio:
Do I understand correctly that the ID is now not part of the result any more, so the problem went away and we can snapshot as any other one?

Jolg42 (approved):

jkomyno:
> Do I understand correctly that the ID is now not part of the result any more, so the problem went away and we can snapshot as any other one?

Exactly.

PR #24479, opened 2024-06-10
fix(cli): `Generate` should exit with 0 without a schema

SevInf:
This should've applied to `postintstall` only. However, 5.15.0 and below
(last 3 years, according to git history) had no hard error in all
`Generate` cases. That changed with `5.16.0-dev.5`. Since that would be
a breaking change, restoring past behaviour and fixing ecosystem-tests
in the process.

Close prisma/team-orm#1183
/integration

### DIFF
--- a/packages/cli/src/__tests__/commands/Generate.test.ts
+++ b/packages/cli/src/__tests__/commands/Generate.test.ts
@@ -775,6 +775,12 @@ describe('--schema from project directory', () => {
       `"Could not load \`--schema\` from provided path \`doesnotexists.prisma\`: file or directory not found"`,
     )
   })
+
+  it('should not throw errors if schema does not exist at default path', async () => {
+    ctx.fixture('empty')
+    const output = await Generate.new().parse([])
+    expect(output).toMatchInlineSnapshot(`""`)
+  })
 })
 
 describe('--schema from parent directory', () => {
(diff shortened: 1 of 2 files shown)


Jolg42:
Looks like a bunch of tests in `migrate` are failing @SevInf 

Example
https://github.com/prisma/prisma/actions/runs/9445855570/job/26014381660?pr=24479#step:7:862
```
FAIL src/__tests__/DbPush.test.ts (7.175 s)
  ● push › should fail if no schema file

    expect(received).rejects.toThrowErrorMatchingInlineSnapshot(snapshot)

    Snapshot name: `push should fail if no schema file 1`

    - Snapshot  - 9
    + Received  + 2

    - Could not find Prisma Schema that is required for this command.
    + Could not find a schema.prisma file that is required for this command.
    - You can either provide it with `--schema` argument, set it as `prisma.schema` in your package.json or put it into the default location.
    - Checked following paths:
    -
    - schema.prisma: file not found
    - prisma/schema.prisma: file not found
    - prisma/schema: directory not found
    -
    - See also https://pris.ly/d/prisma-schema-location
    + You can either provide it with --schema, set it as `prisma.schema` in your package.json or put it into the default location ./prisma/schema.prisma https://pris.ly/d/prisma-schema-location

      41 |
      42 |     const result = DbPush.new().parse([])
    > 43 |     await expect(result).rejects.toThrowErrorMatchingInlineSnapshot(`
         |                                  ^
      44 |       "Could not find Prisma Schema that is required for this command.
      45 |       You can either provide it with \`--schema\` argument, set it as \`prisma.schema\` in your package.json or put it into the default location.
      46 |       Checked following paths:

      at Object.toThrowErrorMatchingInlineSnapshot (../../node_modules/.pnpm/expect@29.7.0/node_modules/expect/build/index.js:218:22)
      at Object.toThrowErrorMatchingInlineSnapshot (src/__tests__/DbPush.test.ts:43:34)
```

SevInf:
Closing for a different appraoch

PR #24481, opened 2024-06-10
feat(client): added provider compatibility check for Driver Adapters

jkomyno:
This PR closes https://github.com/prisma/prisma/issues/21864.

When the Driver Adapter passed to the Prisma Client has an underlying `provider` incompatible with the provider declared in `schema.prisma`, we now throw a runtime validation error. Here's an example when passing `@prisma/adapter-libsql` to a Prisma Client generated from a schema with `provider = "postgres"` / `provider = "postgresql"`:

> "The Driver Adapter \`@prisma/adapter-libsql\`, based on \`sqlite\`, is not compatible with the provider \`postgres\` specified in the Prisma schema."

---

Note: In Prisma 6, we should probably standardise the `provider` values in the `Queryable` interface for Driver Adapters, especially to avoid having two strings for Postgres databases (`postgres`, `postgresql`).
- `getConfig(..).datasources[0].provider` can be `postgresql`, `postgres`, `mysql`, or other known providers
- `getConfig(..).datasources[0].activeProvider`, stored in `config.activeProvider`, can be `postgresql`, `mysql`, or other known providers
- `adapter.provider` can be `postgres`, `mysql`, or `sqlite`, and changing this requires changes to Rust as well,
    see https://github.com/prisma/prisma-engines/blob/d116c37d7d27aee74fdd840fc85ab2b45407e5ce/query-engine/driver-adapters/src/types.rs#L22-L23.

PR #24513, opened 2024-06-12
feat(client): Implement global `omit` configuration

SevInf:
Allows to set `omit` option in `PrismaClient` constructor:

```ts
  omit: {
    user: {
      password: true
    }
  }
})
```

If developer does so, specified field will be omitted from all model
results by default as if local `omit` option was specified for every
query. It is still possible to include field explicitly either by using
`omit: { field: false }` or `select: { field: true }` on a query.

Also, just like in case of local omit, if result extension needs omitted
field, we will still query it from DB, but drop before returning the
final result.

Feature implemented under the same `omitApi` preview flag as local `omit`.

Implementation-wise, done by forwarding `ClientOptions` generic all the
way from `PrismaClient` to individual query results. Both local and
global `omit` are now applied at the moment of `DefaultSelection`
creation. Previously, local omit was applied after `select` and `include`,
but that does not work if we want to be able to undo global omit in
individual queries.

Close prisma/team-orm#1109
Ref #5042

### DIFF
--- a/packages/client/src/generation/ts-builders/DocComment.ts
+++ b/packages/client/src/generation/ts-builders/DocComment.ts
@@ -25,18 +25,26 @@ export class DocComment implements BasicBuilder {
   }
 }
 
-function docComment(strings: TemplateStringsArray): DocComment
+function docComment(strings: TemplateStringsArray, ...args: string[]): DocComment
 function docComment(startingText?: string): DocComment
-function docComment(firstParameter: string | TemplateStringsArray | undefined): DocComment {
+function docComment(firstParameter: string | TemplateStringsArray | undefined, ...args: string[]): DocComment {
   if (typeof firstParameter === 'string' || typeof firstParameter === 'undefined') {
     return new DocComment(firstParameter)
   }
-  return docCommentTag(firstParameter)
+  return docCommentTag(firstParameter, args)
 }
 
-function docCommentTag(strings: TemplateStringsArray) {
+function docCommentTag(strings: TemplateStringsArray, args: string[]) {
   const docComment = new DocComment()
-  const lines = trimEmptyLines(strings.join('').split('\n'))
+  const fullText = strings
+    .flatMap((str, i) => {
+      if (i < args.length) {
+        return [str, args[i]]
+      }
+      return [str]
+    })
+    .join('')
+  const lines = trimEmptyLines(fullText.split('\n'))
   if (lines.length === 0) {
     return docComment
   }
--- a/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts
+++ b/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts
@@ -1,3 +1,4 @@
+import { lowerCase } from '../../../utils/lowerCase'
 import { ErrorFormat } from '../../getPrismaClient'
 import { CallSite } from '../../utils/CallSite'
 import { isDate, isValidDate } from '../../utils/date'
@@ -52,6 +53,12 @@ const jsActionToProtocolAction: Record<Action, JsonQueryAction> = {
   aggregateRaw: 'aggregateRaw',
 }
 
+export type GlobalOmitOptions = {
+  [modelName: string]: {
+    [fieldName: string]: boolean
+  }
+}
+
 export type SerializeParams = {
   runtimeDataModel: RuntimeDataModel
   modelName?: string
@@ -63,6 +70,7 @@ export type SerializeParams = {
   clientVersion: string
   errorFormat: ErrorFormat
   previewFeatures: string[]
+  globalOmit?: GlobalOmitOptions
 }
 
 export function serializeJsonQuery({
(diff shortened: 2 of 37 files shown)


SevInf:
Note: failing `createManyAndReturn` test will need an engine-side fix, TS code is not expected to change for this, so marking this as RFR

Jolg42:
I did a re-run because the e2e test was canceled

jkomyno:
> I did a re-run because the e2e test was canceled

Then this feels very much related to https://github.com/prisma/prisma/pull/24481!

packages/client/tests/functional/globalOmit/test.ts
```diff
@@ -0,0 +1,546 @@
...
+    const client = clientWithOmit({
+      omit: {
+        user: {
+          password: true,
+        },
+      },
+    })
+
+    client.$extends
+
```
SevInf, line 442:
```suggestion
```

packages/client/tests/functional/globalOmit/test.ts
```diff
@@ -0,0 +1,546 @@
...
+declare const prisma: PrismaClient
+
+// wrapper around newPrismaClient to correctly infer generic arguments.
+// `newPrismaClient` by itself is not smart enough for that and I don't think
+// we can make it smarter in a generic way, without having `PrismaClient` on hands.
+function clientWithOmit<O extends Prisma.PrismaClientOptions>(options: O): PrismaClient<O> {
+  return newPrismaClient(options) as unknown as PrismaClient<O>
+}
+
+testMatrix.setupTestSuite(({ provider }) => {
```
SevInf, line 19: TODO:
- test excluding more than one field (different types)
- test excluding every field.

packages/client/src/runtime/core/types/exported/Result.ts
```diff
@@ -111,6 +105,12 @@ export type UnwrapPayload<P> = {} extends P ? unknown : {
       : never
 };
 
+export type ApplyOmit<T, Omit> = Compute<{
```
SevInf, line 108: TODO: rename `OmitConfig`

packages/client/src/runtime/core/types/exported/Result.ts
```diff
@@ -99,7 +88,12 @@ export type SelectField<P extends SelectablePayloadFields<any, any>, K extends P
     : never
 
 // prettier-ignore
-export type DefaultSelection<P> = UnwrapPayload<{ default: P }>['default']
+export type DefaultSelection<P extends OperationPayload, Args = {}, ClientOptions = {}> =
```
SevInf, line 91: Rename to `Payload`

packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts
```diff
@@ -166,27 +176,37 @@ function createImplicitSelection(
...
 function addIncludedRelations(selectionSet: JsonSelectionSet, include: Selection, context: SerializeContext) {
   for (const [key, value] of Object.entries(include)) {
-    const field = context.findField(key)
+    if (value === false) {
+      selectionSet[key] = false
+      continue
+    }
 
-    if (field && field?.kind !== 'object') {
+    const field = context.findField(key)
```
SevInf, line 184: See if we can flatten it more

packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts
```diff
@@ -357,14 +385,17 @@ type ContextParams = {
...
+  globalOmit?: GlobalOmitOptions
 }
 
 class SerializeContext {
   public readonly model: RuntimeModel | undefined
   constructor(private params: ContextParams) {
     if (this.params.modelName) {
       // TODO: throw if not found
-      this.model = this.params.runtimeDataModel.models[this.params.modelName]
+      this.model =
```
SevInf, line 396:
```suggestion
      this.modelOrType =
```

SevInf:
@jkomyno @Druue your feedback is addressed

packages/client/src/generation/ts-builders/DocComment.ts
```diff
@@ -25,18 +25,26 @@ export class DocComment implements BasicBuilder {
...
-  const lines = trimEmptyLines(strings.join('').split('\n'))
+  const fullText = strings
+    .flatMap((str, i) => {
+      if (i < args.length) {
+        return [str, args[i]]
+      }
+      return [str]
+    })
+    .join('')
+  const lines = trimEmptyLines(fullText.split('\n'))
```
jkomyno, line 47: Is there any particular reason why `docCommentTag` was changed here? I notice the additional `docComment` test, do we need an additional `docCommentTag` test as well?

jkomyno (approved):
Large diff, hopefully we didn't miss anything during reviews (both in pair and individually).

Great work @SevInf!

packages/client/src/generation/ts-builders/DocComment.ts
SevInf: Yes. before it did not allow to use substitutions inside of the tag. Now it does and there is a test for it [here](https://github.com/prisma/prisma/pull/24513/files#diff-a8d3b3d1f355c90776f15ada80a71c10328d676529bfbe927e0fb86298b3631cR76)

PR #24570, opened 2024-06-18
feat(cli): use new login endpoint for optimize

aqrln:
Optimize now has its own CLI authentication endpoint that makes use of the new third-party client authentication API in Console, its own login flow pages and support for signing in with both GitHub and Google.


millsp (approved):

millsp:
Reminder to update the `optimize` tag.

PR #24609, opened 2024-06-21
feat(prismaSchemaFolder): Resolve paths in datasource blocks relative to the file DS is defined in

SevInf:
Similar to change to the [one we did for generators](https://github.com/prisma/prisma/pull/24598), this time for
datasource. Before the relative paths were resolved to the root folder, now it will be
relative to a file datasource is defined in. Will affect following paths:

- relative sqlite db path
- ssl certificates paths

Will not affect non-split schema files.

Close prisma/team-orm#1210


packages/cli/src/__tests__/incomplete-schemas.test.ts
```diff
@@ -92,16 +90,19 @@ describe('[wasm] incomplete-schemas', () => {
         await DbExecute.new().parse(['--file=./script.sql'])
       } catch (e) {
         expect(stripAnsi(e.message)).toMatchInlineSnapshot(`
-          "P1012
-
+          "Prisma schema validation - (get-config wasm)
```
SevInf, line 93: Output here changed a bit - we now fail earlier. But the information we output here is the same.

Weakky (approved):

jkomyno (approved):

Issue #24611, opened 2024-06-22
ReadMe link for the current Roadmap is invalid


raymondKelly:
### Bug description

When I go to the ReadMe at the top level of this project and I click the link for roadmap under the [submit-a-feature-request ](https://github.com/prisma/prisma/blob/main/README.md#submit-a-feature-request) section I am taken to a not found 404 page 
<img width="1146" alt="Screenshot 2024-06-21 at 6 13 41 PM" src="https://github.com/prisma/prisma/assets/44653931/e9e434fa-00dd-467e-b122-e3ecac385372">

I presume this is an old link since rework of the website. I think the new URL that is best to use is probably https://www.prisma.io/docs/orm/more/releases#roadmap as that is static and wont change for a while.

### How to reproduce

Go to https://github.com/prisma/prisma/blob/main/README.md#submit-a-feature-request and click the readme link

### Prisma information

N/A

### Environment & setup

N/A

### Prisma Version

N/A

laplab:
Thank you for pointing that out, @raymondKelly! This has now been fixed.

PR #24612, opened 2024-06-22
docs(prisma): incorrect link to the roadmap

raymondKelly:
Resolves Issue https://github.com/prisma/prisma/issues/24611 . Perhaps there is a roadmap page that encompasses more than just the ORM but if there is I could not find it. I'm happy to update this PR if there is.


raymondKelly:
Resolved by Prisma team

PR #24613, opened 2024-06-22
feat(cli): Add format --check

### LINKED ISSUE #4267: Add `check` Option to `format` CLI Command
## Problem
Currently, one can only format a schema using `prisma format` but can't validate the schema is formatted, in CI for example.

## Suggested solution
Add a `--check` option to the CLI for validating the schema is properly formatted.

## Alternatives
Add integration with Prettier which already has this functionality as described here: https://github.com/prisma/prisma/issues/1761

## Additional context
In https://github.com/amplication/amplication we are currently struggling with keeping consistent formatting of our Prisma Schema file. I can write a script to validate it in our CI but I think this functionality is relevant for more users and would be much simpler to implement if it was part of the CLI.

### LINKED ISSUE #12412: `prisma format` for CI environments
## Problem

Want to add Prisma format check to CI.

Only way to run `prisma format` is such that it changes files on disk and always exits 0.

That does not suffice in CI.

## Suggested solution

New flag or subcommand to check-only.

If check discovers any formatting violations then exit non-zero.

## Alternatives

None considered.

baileywickham:
Adds a --check option to format which will error if there are files which do not match prisma format. Useful in CI and pre commit hooks. Closes #4267

### DIFF
--- a/packages/cli/src/__tests__/commands/Format.test.ts
+++ b/packages/cli/src/__tests__/commands/Format.test.ts
@@ -41,6 +41,10 @@ describe('format', () => {
         // implicit: single schema file (`schema.prisma`)
         await expect(Format.new().parse([])).resolves.toMatchInlineSnapshot(`"Formatted schema.prisma in XXXms 🚀"`)
 
+        await expect(Format.new().parse(['--check'])).resolves.toMatchInlineSnapshot(
+          `"All files are formatted correctly!"`,
+        )
+
         // explicit: single schema file (`schema.prisma`)
         await expect(Format.new().parse(['--schema=schema.prisma'])).resolves.toMatchInlineSnapshot(
           `"Formatted schema.prisma in XXXms 🚀"`,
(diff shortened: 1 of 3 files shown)


baileywickham:
Let me know if there is anything I need to change/update to get this in. Happy to help here.

packages/cli/src/__tests__/commands/Format.test.ts
```diff
@@ -352,4 +357,11 @@ describe('format', () => {
     expect(ctx.mocked['console.warn'].mock.calls.join('\n')).toEqual('')
     expect(ctx.mocked['console.error'].mock.calls.join('\n')).toEqual('')
   })
+
+  it('check should fail on unformatted code', async () => {
+    ctx.fixture('example-project/prisma-unformatted')
+    await expect(Format.new().parse(['--schema=unformatted.prisma', '--check'])).resolves.toThrow(
+      `There are unformatted files. Run ${underline('prisma format')} to format them.`,
+    )
```
Druue, line 365:
```suggestion
    await expect(Format.new().parse(['--schema=unformatted.prisma', '--check'])).resolves.toMatchInlineSnapshot(
      `"! There are unformatted files. Run prisma format to format them."`,
    )
```

Druue (commented):
Hey @baileywickham, this looks good but returning an error directly ends up including the stacktrace which isn't needed here. We tend to opt for using an internal error display method that hides this :)

baileywickham:
@Druue Thanks for the tips, I implemented your suggestions and rebased. Let me know if there is anything else I can help with!

Druue (approved):

Druue:
Awesome stuff, thank you so much!

PR #24738, opened 2024-07-09
ci(release): Remove useless engine clone

SevInf:
Despite what function name and logs suggested, it was not actually doing
any tagging. It tried to produce the changelog, which then was never
written anywhere. Cloning itself took quite a bunch of time on the
release, so this should speed up the releases.

/integration


SevInf:
Release is roughly 2.5 min faster

Druue (approved):

PR #24739, opened 2024-07-09
chore(cli): cleanup import message when running generate

Druue:
consolidate importing prisma to one single line that sends users to `pris.ly/d/importing-client`

Removed specificity of `in Node.js` as we support other runtimes

closes https://github.com/prisma/team-orm/issues/1228

### DIFF
--- a/packages/cli/src/__tests__/commands/Generate.test.ts
+++ b/packages/cli/src/__tests__/commands/Generate.test.ts
@@ -668,8 +668,7 @@ describe('--schema from project directory', () => {
         "
         ✔ Generated Prisma Client (v0.0.0) to ./@prisma/client in XXXms
 
-        Start using Prisma Client (See: https://pris.ly/d/client)
-        Importing Prisma Client: http://pris.ly/d/importing-client
+        Start using Prisma Client (See: http://pris.ly/d/importing-client)
 
         ┌─────────────────────────────────────────────────────────────┐
         │  Deploying your app to serverless or edge functions?        │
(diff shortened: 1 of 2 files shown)


Druue:
Hm, I’m not sure how to approach testing this, it doesn’t seem possible in our sandbox set-up but manually checking this, `--silent` seems to be what we want

cc: @SevInf 

https://github.com/prisma/prisma/assets/29753584/d93e1f63-a188-42af-8d67-aa5ec67683fc

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -49,18 +49,7 @@ describe('using cli', () => {
...
-        const prisma = new PrismaClient()
-        \`\`\`
-        or start using Prisma Client at the edge (See: https://pris.ly/d/accelerate)
-        \`\`\`
-        import { PrismaClient } from './generated/client/edge'
-        const prisma = new PrismaClient()
-        \`\`\`
-
-        See other ways of importing Prisma Client: http://pris.ly/d/importing-client
+        Start using Prisma Client (See: https://pris.ly/d/client)
```
SevInf, line 52: Does the link lead to a correct place now? Would it make sense for it to point to https://www.prisma.io/docs/orm/prisma-client/setup-and-configuration/introduction instead?

Druue: So we have the following link `http://pris.ly/d/importing-client` for that. As far as I understood from @apolanc, we want to link to `https://pris.ly/d/client` and we can look at streamlining these pages a little bit so it's easier to navigate to importing new clients.

apolanc: I thought we were pointing to http://pris.ly/d/importing-client as a way to get rid of /client and /client/edge scenarios. What do you @SevInf and @Druue think of rephrasing this to:

`Start by importing your prisma client (See:  http://pris.ly/d/importing-client)`

Which is the next action if you run prisma migrate for the first time.

Druue: Sounds good to me

jkomyno:
Local output, for reference:

Screenshot 2024-07-12 at 11 49 02

jkomyno (approved):

Issue #24737, opened 2024-07-09
Prisma generate option --allow-no-models doesn't work.


MansurAliKoroglu:
### Bug description

`npx prisma generate --allow-no-models` doesn't work in `v16`. It works in `v15`

## Repository

https://github.com/MansurAliKoroglu/prisma-bug

## Steps

- Run `npm i` to install packages.
- Run `npx prisma generate --allow-no-models` and see it doesn't work
- Edit `package.json` file and change prisma packages to version `5.15.0` 
- Run `npm i` to install packages.
- Run `npx prisma generate --allow-no-models` and see it works.

### Expected behavior

It should work in `v16` too.

### Prisma information

```prisma.schema
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
  url      = env("DATABASE_URL")
}
```

No code needed.

### Environment & setup

OS: Kubuntu 22.04
Node: 20.15.0

### Prisma Version

16

jkomyno:
Hi @MansurAliKoroglu, thanks for informing us about this issue.
I've fixed this regression in https://github.com/prisma/prisma/pull/24741, and added more tests to avoid similar issues in the future.
This fix will be part of next week's stable release, Prisma 5.17.0.

Cheers 👋🏻

PR #24741, opened 2024-07-09
fix(cli): restored `prisma generate --allow-no-models` after regression

jkomyno:
This PR fixes the regression in https://github.com/prisma/prisma/issues/24737, caused by https://github.com/prisma/prisma/pull/24274.
It also complements the older PR https://github.com/prisma/prisma/issues/24160, by adding missing `cli` integration tests.


SevInf (approved):
Nice, thanks for adding all of the tests

PR #24745, opened 2024-07-09
feat(cli): Support `prisma platform auth show --sensitive`

justinvdm:
## Changes
When the user adds a `--sensitive` option when running `prisma platform auth token show`, we include the current auth token in the output

```sh
npx prisma platform auth show --sensitive

# Output

Currently authenticated as foo@bar.org

displayName : Foo Bar
id          : 123
email       : foo@bar.org
token       : abc123
```

 ## Context
This would allows users to provide the token explicitly for us to authenticate for Optimize in cases like docker or codespaces, where we aren't able to hit the temporary local server the cli creates for auth purposes.

For example:

```ts
new PrismaClient().$extends(withOptimize({
  // in this case the user stored the token the cli logged in (1), then provided it as an env var
  token: process.env.PLATFORM_TOKEN
}))
```

 ## Links
* Discussion: https://prisma-company.slack.com/archives/C04FDERNP32/p1720522326387089
* Ticket: https://linear.app/prisma-company/issue/EXP-135/prisma-optimize-login-in-docker
* Related issue: https://github.com/prisma/optimize-feedback/issues/3


packages/cli/src/platform/auth/__tests__/show.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    jest.spyOn(credentialsFile, 'load').mockImplementation(
+      (): ReturnType<typeof credentialsFile.load> =>
+        Promise.resolve({
+          token: 'testToken',
+        }),
+    )
+
+    await expect(Show.new().parse([])).resolves.not.toContain('testToken')
+  })
+})
```
justinvdm, line 75: Not sure if this test is in the right place or following the right conventions for how we usually test cli commands. Feedback welcome!

millsp, line 75:
```suggestion
import { credentialsFile } from '../../_lib/credentials'
import { requestOrThrow } from '../../_lib/pdp'
import { Show } from '../show'

jest.mock('../../_lib/pdp')
jest.mock('../../_lib/credentials')

describe('platform auth show', () => {
  test('should output token with --sensitive', async () => {
    jest.mocked(requestOrThrow).mockResolvedValue(
      Promise.resolve({
        me: {
          user: {
            __typename: 'User',
            id: '23',
            email: 'foo@bar.org',
            displayName: 'Foo Bar',
          },
        },
      }),
    )

    jest.mocked(credentialsFile).load.mockReturnValue(
      Promise.resolve({
        token: 'testToken',
      }),
    )

    await expect(Show.new().parse(['--sensitive'])).resolves.toContain('testToken')
  })

  test('should not output token if --sensitive not given', async () => {
    jest.mocked(requestOrThrow).mockResolvedValue(
      Promise.resolve({
        me: {
          user: {
            __typename: 'User',
            id: '23',
            email: 'foo@bar.org',
            displayName: 'Foo Bar',
          },
        },
      }),
    )

    jest.mocked(credentialsFile).load.mockReturnValue(
      Promise.resolve({
        token: 'testToken',
      }),
    )

    await expect(Show.new().parse([])).resolves.not.toContain('testToken')
  })
})
```

packages/cli/src/platform/auth/show.ts
```diff
@@ -41,9 +45,36 @@ export class Show implements Command {
         `,
       },
     })
+
+    let sensitiveData: { token: string | null } = {
+      token: null,
+    }
```
millsp, line 51:
```suggestion
    let sensitiveData: { token: string | undefined } = {
      token: undefined,
    }
```

justinvdm: Coincidentally I had this initially, but then `messages.resouce()` ends up logging `undefined` to the console. In contrast, it filters out nulls, thus the `null`: https://github.com/prisma/prisma/blob/0e92f9d40caf8888ca8bb04fbe893b48689168b9/packages/cli/src/platform/_lib/messages.ts#L35

I'd instead simply just add the property if the condition is met to avoid it needing to filter out nulls in the first place, but then I'd also need to add the relevant logic for the second param of `message.resource()` (e.g. `shouldShowSensitive ? { token } : {}`).

Overall just seemed simpler to go for null.

aqrln: You can also expand that condition to filter out both `null` and `undefined`. I personally don't have a strong preference and the code is fine with me as it currently is too.

jasonkuhrt (approved):
Generally LGTM, we can bike shed on code style but the gist is here 🎉

packages/cli/src/platform/auth/show.ts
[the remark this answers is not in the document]
justinvdm: That sounds good! Will do, thanks @jasonkuhrt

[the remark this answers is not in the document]
justinvdm: @jasonkuhrt I actually just realised now that we already get the token earlier on in this function, just a few lines up. Sorry for missing that 🤦:
https://github.com/prisma/prisma/blob/247b4b9d24fc900e503af34cbc9f76b061d216f7/packages/cli/src/platform/auth/show.ts#L19

 So in the end I didn't end up adding `loadOrThrow()`, since I didn't need it anymore. The changes are just as simple though.

aqrln (approved):

Issue #21631, opened 2023-10-26
$transaction weird behaviour


jwld:
### Bug description

```
const [emailConflict, phoneConflict] = await prisma.$transaction([
  prisma.worker.findUnique({ where: { email: data.email } }),
  prisma.worker.findUnique({ where: { phone: data.phone } })
])
```

Both are `null`, even when there's an email conflict. If I take the second query out:

```
const [emailConflict] = await prisma.$transaction([
  prisma.worker.findUnique({ where: { email: data.email } })
])
```

then `emailConflict` is not null and returns the conflicting worker as expected.

### How to reproduce

As above

### Prisma information

```
model Worker {
  email     String  @unique
  phone     String  @unique
  id        String   @id @default(cuid())
}
```

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: 18.9.0

### Prisma Version

```
5.5.2
```

Murtatrxx:
Do you use driver adapters, because they also cause weird problems in my case (#21592)

janpio:
What do you mean by "even when there's an email conflict" @jwld?

jwld:
@Murtatrxx nope, nothing like that

@janpio Sorry, I mean if there's a worker with that email in the DB already. For context we run this to check for potential conflicts before creating a new worker, rather than letting prisma throw a unique conflict error.

jkomyno:
Hi @jwld, I confirm that this issue was happening in Prisma version you reported, 5.5.2.

We actually fixed this problem in [Prisma 5.12.0](https://github.com/prisma/prisma/releases/tag/5.12.0), albeit not explicitly (it was likely a by-product of the fix for https://github.com/prisma/prisma/issues/23343).
I hence added explicit tests for the scenario you highlighted in [here](https://github.com/prisma/prisma/pull/24764), to ensure no future user will stumble upon the same problem.

Thanks!

jwld:
Thanks for the update, good to hear!

PR #24764, opened 2024-07-11
test(client): add functional tests for #21631

jkomyno:
#21631 got fixed in Prisma 5.12.0, but we were missing explicit tests, as explained in [here](https://github.com/prisma/prisma/issues/21631#issuecomment-2222957013).


Weakky (approved):
Wasn't this rather fixed in 5.17 from this PR ? https://github.com/prisma/prisma-engines/pull/4911 - Regardless, LGTM

PR #24792, opened 2024-07-14
feat(cli, generate command): Added promo array to rotate and make cli messaging more engaging

apolanc:
- Included array of promotions
- function to handle promos and show a random small and discrete text on each prisma generate run
- modifications to prisma init to remove the pdp promotion and add a 5th step instead

closes: https://github.com/prisma/team-orm/issues/1243


Druue (commented):
Hm, something doesn't seem to be rendering quite right @apolanc
image

I ran generate a few times and managed to get this result too which shouldn't be possible from the looks of the code
image

apolanc:
> Hm, something doesn't seem to be rendering quite right @apolanc image
> 
> I ran generate a few times and managed to get this result too which shouldn't be possible from the looks of the code image

Thanks @Druue I just pushed a change that fixes this behaviour. I was just calling the random function twice random().text random().link [facepalm]. All should be good now

apolanc:
image

Druue:
Is the gap at the bottom intended? It looks a little off @apolanc

apolanc:
> Is the gap at the bottom intended? It looks a little off @apolanc

No. I'll push a change for it now.

Druue (approved):

packages/cli/src/Generate.ts
```diff
@@ -255,21 +256,10 @@ Please make sure they have the same version.`
...
-            horizontalPadding: 2,
-          })
-
           hint = `
-Start by importing your prisma client (See:  http://pris.ly/d/importing-client)
+Start by importing your Prisma Client (See: http://pris.ly/d/importing-client)
 
-${boxedTryAccelerateMessage}
+${promotion.text} ${promotion.link}
 ${getHardcodedUrlWarning(config)}${breakingChangesStr}${versionsWarning}`
```
Druue, line 263:
```suggestion
${promotion.text} ${promotion.link}${getHardcodedUrlWarning(config)}${breakingChangesStr}${versionsWarning}`
```

Druue: and then update `getHardcodedUrlWarning` like so:
```diff
export function getHardcodedUrlWarning(config: ConfigMetaFormat) {
  const datasource = config.datasources?.[0]

  if (
    datasource !== undefined &&
    datasource.provider !== 'sqlite' &&
    (datasource.url.fromEnvVar === null || datasource.directUrl?.fromEnvVar === null)
  ) {
-    return `\n🛑 Hardcoding URLs in your schema poses a security risk: ${link('https://pris.ly/d/datasource-env')}\n`
+    return `\n\n🛑 Hardcoding URLs in your schema poses a security risk: ${link('https://pris.ly/d/datasource-env')}\n`
  }

  return ''
}
```

Druue: This removes an extra newline when this hardcoded url doesn't exist

apolanc:
```
${promotion.text} 
${promotion.link}
```

This would put the link bellow the line. Why do you suggest this change? 

I haven't replicated this scenario to see how it looks in the cli

Druue: Ah, no, it doesn't they're on the same line, that's just how it got rendered on github

Druue: image

Goop-arsh (commented):
R

Goop-arsh (commented):
T

PR #24818, opened 2024-07-17
feat: Added a tip to remove tips

gniting:
- Added additional tip to give users the option to remove/hide tips
- Change the prisma init pris.ly link to point to the right one


packages/cli/src/utils/handlePromotions.ts
```diff
@@ -23,6 +23,10 @@ const promotions = [
     text: `Tip: Curious about the SQL queries Prisma ORM generates? Optimize helps you enhance your visibility:`,
     link: 'https://pris.ly/tip-2-optimize',
   },
+  {
+    text: `Tip: Want to turn off tips and other hints?`,
```
gniting, line 27: Pls change/update the text if what I have suggested is not ideal.

apolanc (approved):

PR #24878, opened 2024-07-24
feat(driver-adapters): add TransactionContext

### LINKED ISSUE #4469: [DA] Planetscale engine tests: interactive_tx
This is a subclaster of failures after having enabled planetscale tests in https://github.com/prisma/prisma-engines/pull/4423
 
 
Failing tests in this cluster:
 

Follow the process in [This notion scratchpad page](https://www.notion.so/prismaio/Process-for-tackling-Planetscale-failing-tests-75e280e923e64e2aadf9e367a42aef8b?pvs=4)

jkomyno:
This PR:
- closes https://github.com/prisma/prisma-engines/issues/4469
- closes https://github.com/prisma/team-orm/issues/953

**Warning**: this changes the public `DriverAdapter` interface. It's technically not a breaking change (since `driverAdapters` is a preview feature), but we should notify the third-parties who started building their own Prisma Driver Adapters.

---

Before this PR, the workflow for starting a transaction was:

```mermaid
sequenceDiagram
    autonumber
    participant Client
    participant DriverAdapter
    participant Transaction
    
    Client->>DriverAdapter: startTransaction()
    DriverAdapter-->>Transaction: Transaction interface
    Client->>Transaction: executeRaw()
```

In other words, we'd instantiate a new transaction directly from the `DriverAdapters` interface.
But due to a MySQL limitation, this prevented us from properly setting the isolation level before starting the transaction itself (see https://github.com/prisma/prisma-engines/pull/4966).

So, this PR introduces the `TransactionContext` interface, which offers that guarantee that every query executed in it is preserved in the context of the transaction spawned immediately after. This directly addresses this comment from Serhii:

> `client.execute` is not guaranteed to be executed on the same connection transaction will later start and it only works it tests because there are no concurrent requests.

Here's how the workflow looks like now:

```mermaid
sequenceDiagram
    autonumber
    participant Client
    participant DriverAdapter
    participant TransactionContext
    participant Transaction
    
    Client->>DriverAdapter: transactionContext()
    DriverAdapter-->>TransactionContext: TransactionContext interface
    Client->>TransactionContext: executeRaw('SET TRANSACTION ISOLATION LEVEL ...')
    Client->>TransactionContext: startTransaction()
    TransactionContext-->>Transaction: Transaction interface
    Client->>Transaction: executeRaw()
```


packages/adapter-planetscale/src/planetscale.ts
```diff
@@ -193,3 +178,32 @@ const adapter = new PrismaPlanetScale(client)
...
+    const dbName = new URL(url).pathname.slice(1) /* slice out forward slash */
+    return ok({
+      schemaName: dbName,
+    })
+  }
+
+  async transactionContext(): Promise<Result<TransactionContext>> {
+    const conn = this.client.connection()
+    const ctx = new PlanetScaleTransactionContext(conn)
+    return Promise.resolve(ok(ctx))
```
aqrln, line 207:
```suggestion
    return ok(ctx)
```
it's already an `async` function

jkomyno: Not quite:

Screenshot 2024-09-03 at 12 20 00

aqrln: Then it's a problem with the eslint rule being wrong and not taking the implemented interfaces into account 😆
It's fine though, let's leave `Promise.resolve` to shut it up then.

aqrln: You did `  // eslint-disable-next-line @typescript-eslint/require-await` in other places in this PR to solve the same problem though.

packages/adapter-d1/src/d1.ts
```diff
@@ -129,6 +130,25 @@ class D1Transaction extends D1Queryable<StdClient> implements Transaction {
...
+
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async startTransaction(): Promise<Result<Transaction>> {
+    const options: TransactionOptions = {
+      // TODO: D1 does not have a Transaction API.
+      usePhantomQuery: true,
+    }
+
+    const tag = '[js::startTransaction]'
+    debug(`${tag} options: %O`, options)
```
aqrln, line 146:
```suggestion
    const tag = '[js::startTransaction]'
    debug(`%s options: %O`, tag, options)
```

packages/adapter-planetscale/src/planetscale.ts
jkomyno: I've looked more into it. I wasn't the original person who used `// eslint-disable-next-line @typescript-eslint/require-await` (Joel did). Afaik, this rule can be safely disabled as we're not using generator functions (`yield` behaves differently depending on whether a Promise is explicitly returned afaik, that's why this rule exists in the first place).

I will disable it at the top of the files, rather than once for each of these methods.

aqrln: Sounds good!

aqrln (approved):
Nice work!

PR #24916, opened 2024-07-31
fix(migrate): correct types import path in @prisma/migrate

boredland:
The actual package only contains the types in `dist/migrate/src`, but the package.json `types` pointed to a file that doesn't exist.

image


boredland:
> Curiosity: how did you stumble on this issue? 

Hi! We're using it to programmatically migrate. The biggest reason currently is, that we have a certain logic which database connection string to use (for vercel preview deployments) and populating this to an env-variable would be much more effort.

jkomyno (approved):

PR #24934, opened 2024-08-02
fix: Update libsql/client version

haaawk:
New version contains bugfixes for some serious performance problems


jkomyno:
Hi @haaawk, thanks for this PR. It looks like this update caused most tests to fail, though

haaawk:
> Hi @haaawk, thanks for this PR. It looks like this update caused most tests to fail, though

I'm not sure how the upgrade of libsql client could have caused failures like this @jkomyno 
image

SevInf (approved):

PR #24935, opened 2024-08-02
Add cross-env to support windows in basic-sqlite template

RobertSandiford:
The basic-sqlite template in sandbox was not windows compatible - the env var set in the generate command. Added cross-env dependency to resolve.


RobertSandiford:
modifying commit message

PR #24936, opened 2024-08-02
chore: Add cross-env to support windows in basic-sqlite template

RobertSandiford:
The basic-sqlite template in sandbox was not windows compatible - the env var set in the generate command. Added cross-env dependency to resolve.


wmadden (approved):
Hi @RobertSandiford,

Thanks for taking the time to open a pull request. This one seems a little underdeveloped since I'm sure there are _many_ other points of incompatibility with windows for developers in the Prisma codebase, but I suppose it doesn't hurt.

RobertSandiford:
Yep, some of the templates had cross-env, some didn't.

I just submitted the changes that I needed to make for my own needs, which was this basic template, probably the most important one.

wmadden:
@RobertSandiford in hindsight a better solution to your problem would be to use `pnpm config set script-shell` to set your shell to bash to solve this problem. We'll probably correct this at some point, but for now this is fine

PR #24941, opened 2024-08-04
fix: add actual current directory to search path of engine

rubenfiszel:
The error message say to have the engine next to the bundle but actually it only search one level up.

In our case, that's a big issue because we create ephemeral folder per executions and untar the bundle in that single folder. We cannot put anything in the folder above.

Looking at the code and the docs make me believe there was a typo and instead of using dirname, it uses config.dirname. 

To not break existing code, config.dirname was preserved


JasaDesain-chemproject:
The error message indicates that the engine should be placed next to the bundle, but it actually only searches one level up. In our case, this poses a significant problem because we generate ephemeral folders for each execution and extract the bundle within that specific folder. We are unable to place anything in the folder above.

After reviewing the code and the documentation, it appears that there might have been a typo; instead of using dirname, it uses config.dirname. To avoid breaking existing code, the use of config.dirname has been preserved.

<a href="https://bit.ly/3Aa5e7f">Jasa Desain Grafis</a>

cookrn:
This is related to https://github.com/prisma/prisma/discussions/19499 cc @millsp

tyler-dot-earth:
@aqrln i see this was tagged for `v6.14` in July and `v6.15` in August - did this fix slip through the cracks?

aqrln:
Closing since this is no longer relevant in Prisma 7

aqrln:
Sorry that this PR slipped through the cracks and thanks for the contribution! Feel free to re-open targeting 6.19.x instead of main if by any slim chance this is still relevant for you.

Issue #24780, opened 2024-07-12
Selection with `_count: undefined` not working properly


edobrb:
### Bug description

Settings `_count: undefined` on selection throws an error: `Unknown field at "wallet._count selection"`

result1 and result2 works properly, result3 no

### How to reproduce

```
const prisma = new PrismaClient()
const result1 = await prisma.authenticator.findMany({ select: { wallet: { select: { id: true, _count: { select: { sent: true } } } } } })
const result2 = await prisma.authenticator.findMany({ select: { wallet: { select: { id: true } } } })
const result3 = await prisma.authenticator.findMany({ select: { wallet: { select: { id: true, _count: undefined } } } })
await prisma.$disconnect()
  ```

### Expected behavior

result3 should not throw and behave like result2

### Prisma information

```prisma
model Wallet {
  id          String  @id @default(auto()) @map("_id") @db.ObjectId
}

model Authenticator {
  id         String  @id @default(auto()) @map("_id") @db.ObjectId
  
  walletId       String? @db.ObjectId
  wallet         Wallet? @relation(fields: [walletId], references: [id])
}
```

### Environment & setup

- OS: macOS
- Database: MongoDB
- Node.js version: 20.10

### Prisma Version

```
prisma                  : 5.16.2
@prisma/client          : 5.16.2
Computed binaryTarget   : darwin-arm64
Operating System        : darwin
Architecture            : arm64
Node.js                 : v20.10.0
Query Engine (Node-API) : libquery-engine 34ace0eb2704183d2c05b60b52fba5c43c13f303 (at ../../node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli 34ace0eb2704183d2c05b60b52fba5c43c13f303 (at ../../node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.16.0-24.34ace0eb2704183d2c05b60b52fba5c43c13f303
Default Engines Hash    : 34ace0eb2704183d2c05b60b52fba5c43c13f303
Studio                  : 0.502.0
```

Pressidentt:
I tried to recreate your problem. I tried running **findMany** request with **__count_**: undefined both on postrgres and mongodb. It works fine for both. It actually generates same SQLs for second and third request. Would you provide more information? Try adding logging and share your results.

 ` log: ['query', 'info', 'warn', 'error'] `
 
[...]

PR #25062, opened 2024-08-22
fix(client): Passing undefined _count in select

igort99:
## Description

As _count arg in select args is conditional it can be passed as undefined and in that case query fails. The expected behavior is that the query is executed as _count is not passed. This fix aims to accomplish that.

## Related Issue

- Issue: [#24780](https://github.com/prisma/prisma/issues/24780)

## Changes Made

-  Added functions for desugaring UserArgs in case of select args
-  Make sure if _count: undefined is passed it is cleared from args
-  Added test that covers this use case

## Screenshots (if applicable)

The error log looks like this.

```typescript
const links = await prisma.link.findMany({
  select: {
    user: {
      select: {
        id: true,
        _count: undefined
      }
    }
  }
})
Unknown field at "user._count selection"
```
Engines that is being thrown.

```json
[
  {
    "error": "Error in query graph construction: QueryParserError(ValidationError { kind: EmptySelection, message: \"Expected a minimum of 1 field to be present, got 0\", meta: Some(Object {\"outputType\": Object {\"name\": String(\"UserCountOutputType\"), \"fields\": Array [Object {\"name\": String(\"links\"), \"typeName\": String(\"Int\"), \"isRelation\": Bool(false)}]}, \"selectionPath\": Array [String(\"findManyLink\"), String(\"user\"), String(\"_count\")]}) })",
    "user_facing_error": {
      "is_panic": false,
      "message": "Expected a minimum of 1 field to be present, got 0",
      "meta": {
        "outputType": {
          "name": "UserCountOutputType",
          "fields": [
            {
              "name": "links",
              "typeName": "Int",
              "isRelation": false
            }
          ]
        },
        "selectionPath": [
          "findManyLink",
          "user",
          "_count"
        ]
      },
      "error_code": "P2009"
    }
  }
]
```

## Checklist

- [X] I have run the code and it works as expected.
- [X] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
+++ b/packages/client/src/__tests__/integration/happy/not-so-exhaustive-schema-mongo/__snapshots__/binary.test.ts.snap
@@ -557,9 +557,9 @@ export class PrismaClient<
    */
   $runCommandRaw(command: Prisma.InputJsonObject): Prisma.PrismaPromise<Prisma.JsonObject>
 
-  $extends: $Extensions.ExtendsHook<"extends", Prisma.TypeMapCb, ExtArgs, $Utils.Call<Prisma.TypeMapCb, {
+  $extends: $Extensions.ExtendsHook<"extends", Prisma.TypeMapCb<ClientOptions>, ExtArgs, $Utils.Call<Prisma.TypeMapCb<ClientOptions>, {
     extArgs: ExtArgs
-  }>, ClientOptions>
+  }>>
 
       /**
    * \`prisma.post\`: Exposes CRUD operations for the **Post** model.
@@ -1163,11 +1163,14 @@ export namespace Prisma {
     db?: Datasource
   }
 
-  interface TypeMapCb extends $Utils.Fn<{extArgs: $Extensions.InternalArgs, clientOptions: PrismaClientOptions }, $Utils.Record<string, any>> {
-    returns: Prisma.TypeMap<this['params']['extArgs'], this['params']['clientOptions']>
+  interface TypeMapCb<ClientOptions = {}> extends $Utils.Fn<{extArgs: $Extensions.InternalArgs }, $Utils.Record<string, any>> {
+    returns: Prisma.TypeMap<this['params']['extArgs'], ClientOptions extends { omit: infer OmitOptions } ? OmitOptions : {}>
   }
 
-  export type TypeMap<ExtArgs extends $Extensions.InternalArgs = $Extensions.DefaultArgs, ClientOptions = {}> = {
+  export type TypeMap<ExtArgs extends $Extensions.InternalArgs = $Extensions.DefaultArgs, GlobalOmitOptions = {}> = {
+    globalOmitOptions: {
+      omit: GlobalOmitOptions
+    }
     meta: {
       modelProps: "post" | "user" | "embedHolder" | "m" | "n" | "oneOptional" | "manyRequired" | "optionalSide1" | "optionalSide2" | "a" | "b" | "c" | "d" | "e"
       txIsolationLevel: never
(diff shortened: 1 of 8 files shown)


SevInf (changes requested):
Hi @igort99 


we really appreciate your contributions, but `argsMapper` looks like a wrong place to do it for me. 

JSON protocol serializer looks like a better and simpler place to do it. In particular [this](https://github.com/prisma/prisma/blob/aee77e9344c1c0cfcd71cd5086d4fc6c11a06b61/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts#L181) and [this](https://github.com/prisma/prisma/blob/aee77e9344c1c0cfcd71cd5086d4fc6c11a06b61/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts#L235) conditions needs to be adjusted to include explicitly `undefined` fields too.

Would you be interested in adjusting your PR this way? Thank you!

packages/client/tests/functional/issues/24780-desugar-select-aggregate-args/tests.ts
```diff
@@ -0,0 +1,31 @@
+// @ts-ignore
+import type { PrismaClient } from '@prisma/client'
+
+import testMatrix from './_matrix'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(
+  (_suiteMeta, _clientMeta) => {
+    test('select _count that is undefined', async () => {
```
SevInf, line 10: This would also need additional test for `include: { _count: undefined } `

packages/client/tests/functional/issues/24780-desugar-select-aggregate-args/tests.ts
```diff
@@ -0,0 +1,31 @@
+// @ts-ignore
+import type { PrismaClient } from '@prisma/client'
+
+import testMatrix from './_matrix'
+
+declare let prisma: PrismaClient
+
+testMatrix.setupTestSuite(
+  (_suiteMeta, _clientMeta) => {
```
SevInf, line 9:
```suggestion
  () => {
```

packages/client/tests/functional/issues/24780-desugar-select-aggregate-args/tests.ts
```diff
@@ -0,0 +1,31 @@
...
+            },
+          },
+        },
+      })
+
+      expect(result).toEqual([])
+    })
+  },
+  {
+    skipDataProxy: {
```
SevInf, line 26: There are no errors thrown in this test, so no need to skip data proxy.

packages/client/tests/functional/issues/24780-desugar-select-aggregate-args/_matrix.ts
```diff
@@ -0,0 +1,11 @@
+import { defineMatrix } from '../../_utils/defineMatrix'
+import { allProviders } from '../../_utils/providers'
+
+export default defineMatrix(() => [
+  allProviders,
+  [
+    {
+      previewFeatures: '',
+    },
+  ],
```
SevInf, line 10:
```suggestion
  allProviders,
```

igort99:
> Hi @igort99
> 
> we really appreciate your contributions, but `argsMapper` looks like a wrong place to do it for me.
> 
> JSON protocol serializer looks like a better and simpler place to do it. In particular [this](https://github.com/prisma/prisma/blob/aee77e9344c1c0cfcd71cd5086d4fc6c11a06b61/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts#L181) and [this](https://github.com/prisma/prisma/blob/aee77e9344c1c0cfcd71cd5086d4fc6c11a06b61/packages/client/src/runtime/core/jsonProtocol/serializeJsonQuery.ts#L235) conditions needs to be adjusted to include explicitly `undefined` fields too.
> 
> Would you be interested in adjusting your PR this way? Thank you!

Thank you for your feedback and review. As this is my first contribution to Prisma, I was uncertain about the appropriate location for the sanitization. I will update the pull request to align with your suggestions and work on enhancing the tests as well.

PR #25070, opened 2024-08-23
fix(client): .$extends prevents typescript documentation

### LINKED ISSUE #24648: `.$extends` prevents typescript documentation
### Bug description

Adding doc comments on models/fields with `///` show in the intellisense, **until** the client is extended:

<details>
  <summary>Without Extension</summary>

  
image

</details>

<details>
  <summary>With Extension</summary>

image

</details>

### How to reproduce

1. Create a `schema.prisma` with a `model` with a documented field
2. Instantiate a new client
3. Use `$extends` to extend client
4. Hover over documented field, and see no documentation in intellisense

### Expected behavior

Extending the client should not break the type docs.

### Prisma information

```prisma
model user {
  /// `lastUpdated` is used to track the last time the user's profile was updated.
  ///
  /// This has the main benefit of allowing concurrent ops to check for race conditions.
  lastUpdated                  Int @default(0)
}
```

```ts
import { PrismaClient } from '@prisma/client';

function getExtendedClient() {
    datasources: {
      db: {
        url: MONGOHQ_URL
      }
    }
  }).$extends({
    query: {
      user: {
        async update({ args, query }) {
          args.data.lastUpdated = Date.now();
          return query(args);
        },
        async updateMany({ args, query }) {
          args.data.lastUpdated = Date.now();
          return query(args);
        },
        async upsert({ args, query }) {
          const lastUpdated = Date.now();
          args.update.lastUpdated = lastUpdated;
          args.create.lastUpdated = lastUpdated;
          return query(args);
        },
        async create({ args, query }) {
          args.data.lastUpdated = Date.now();
          return query(args);
        },
        async createMany({ args, query }) {
          const data = args.data;
          const lastUpdated = Date.now();
          if (Array.isArray(data)) {
            data.forEach(data => {
              data.lastUpdated = lastUpdated;
            });
          } else {
            data.lastUpdated = lastUpdated;
          }

          return query(args);
        }
        // NOTE: raw ops are untouched, as it is meant to be a direct passthrough to mongodb
        // async findRaw({ model, operation, args, query }) {}
        // async aggregateRaw({ model, operation, args, query }) {}
      }
    }
  });

  const user = (await prisma.user.findMany())[0];

  user?.lastUpdated; // Hovering over should show intellisense

  return prisma;
}
```

### Environment & setup

- OS: Windows 11, WSL 2, Ubuntu 22.04
- Database: MongoDB
- Node.js version: 22.2.0

### Prisma Version

```
prisma                  : 5.5.2
@prisma/client          : 5.5.2
Query Engine (Node-API) : libquery-engine aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a (at ../node_modules/.pnpm/@prisma+engines@5.5.2/node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Schema Engine           : schema-engine-cli aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a (at ../node_modules/.pnpm/@prisma+engines@5.5.2/node_modules/@prisma/engines/schema-engine-debian-openssl-3.0.x)
Schema Wasm             : @prisma/prisma-schema-wasm 5.5.1-1.aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a
Default Engines Hash    : aebc046ce8b88ebbcb45efe31cbe7d06fd6abc0a
```

Probably should go with: https://github.com/prisma/prisma/issues/18561

HRM:
Closes #24648

The fix changes GetPayloadResult utility type from a mapped type to a combination of an intersection and a mapped type. This way documentation doesn't get erased in the presence of an extension.

### Tested cases:

1. no extension
2. extension without result extension
3. extension with result extension
4. result extension overwriting a field on base type
5. result extension with the omit feature


jkomyno:
Hi @HRM, thank you for this PR!

I've broken down each type from before an after, to properly understand what's going on in these types.

## Shared Definitions

```typescript
export type InternalArgs<
  R = { [K in string]: { [K in string]: unknown } },
  M = { [K in string]: { [K in string]: unknown } },
  Q = { [K in string]: { [K in string]: unknown } },
  C = { [K in string]: unknown },
> = {
  result: { [K in keyof R]: { [P in keyof R[K]]: () => R[K][P] } }
  model: { [K in keyof M]: { [P in keyof M[K]]: () => M[K][P] } }
  query: { [K in keyof Q]: { [P in keyof Q[K]]: () => Q[K][P] } }
  client: { [K in keyof C]: () => C[K] }
}
```

`InternalArgs` defines structures for different types of Prisma client extensions (`result`, `model`, `query`, `client`).

For `client`, it's an object whose:
- keys `K` are a subset of the keys in `C = { [K in string]: unknown }`, which implies `K extends string`, `C[K] extends unknown`
- values have type `() => C[K]`, so effectively `() => unknown`

For `result`, it's an object whose:
- keys `K` are a subset of the keys in `R = { [K in string]: { [P in string]: unknown } }`, which implies `K extends string`, `R[K] extends { [P in string]: unknown }`, and `R[K][P] extends unkown`
- values have type `{ [P in keyof R[K]]: () => R[K][P] }`, so an object whose keys `P` are a subset of `R[K]` aka `{ [P in string]: unknown }`, and whose inner values are `R[K][P] extends unkown`

`C` describes the extensions to the Prisma Client itself.
`R` describes the additional properties to the Prisma query result due to a Prisma Client extension.

## Before

```typescript
export type GetPayloadResult<
  Base extends Record<any, any>,
  R extends InternalArgs['result'][string],
  KR extends keyof R = string extends keyof R ? never : keyof R,
> = unknown extends R
  ? Base
  : {
      [K in KR | keyof Base]: K extends KR
        ? R[K] extends () => {
          compute: (...args: any) => infer C
        }
          ? C
          : never
        : Base[K]
    }
```

The `GetPayloadResult` type transforms the result's payload on extensions. It takes 3 parameters:
1. `Base`: The base type representing the original data structure (e.g., the model's shape).
2. `R`: The result extension type (from `InternalArgs`). It's type has the following shape: `{ [P in keyof R[K]]: () => R[K][P] }`
3. `KR`: A derived type that represents the keys of `R`, defaulting to `never` if `string extends keyof R`. It's effectively a type safety check.

There two top-level cases:
- `unknown extends R ? Base : ...` If `R` is unknown, it simply returns `Base`. This is a fallback for cases where no extension is provided.
- If `R` is not `unknown`, the result is a new object type, whose
  - keys `(K in KR | keyof Base)` includes may be from both the original `Base` type and `KR` (keys from the result extensions)
  - values depend. If the key `K` is part of the extension keys `KR`, we check whether `R[K]` is a function returning an object with a compute method. If `compute` exists, it infers the type of `C` from `compute`'s return type. Otherwise, it returns never (indicating an invalid extension). If the key is not in `KR`, it simply takes the value from `Base[K]`.

**Example**: [TypeScript Playground](https://www.typescriptlang.org/play/?ssl=51&ssc=2&pln=32&pc=1#code/PQKgBAQgogYg8gJSmAFgFzQBwM4C5jADmAlmigK4BGAdAMYD2AtsJgE7HaMCGL7nPmcgBshwAEwBWAAwB2KWBDAAUEoCmAD0z1WaMGgCemVWACSAOzSrWZrkICCrQtgA8SsGARgAvGADeYAG0AaTBiMzBsNHYzQgBdXD9AkLCIqLC4hPIzAGszegB3cIBfMCKAGjcwAFlvRODQ8MjojLrkxrSY+LAs3ILi0or3AEVa-3qUpvSusbbU5q6evMLSgcqAYVGkhrmpzJyl4oqAPlHK1lVsYTQEme3s1X16ADMPacCABTuH549g2K6ABQASm8JwQfwC71iKyKlUY9AAJqohDctil7o8XlU3pCvpjqn9ASCvCcqhCoTDKgBHchWfSo8bhDE-IY4z7o74vIaEhLA0FgblBWKQ6ElWHuWhCYiqCwM2bMl5rIn8tZ-UpKWFqTTaXQGIxgADiqjQ7y4+iE9C4CIQFyurncEC42GMGksZgR2A8qgYrARzi4Zn0ZTAAf0R0GHjArplHtMFisNnsjmwAQA5OdLkI0KnhZNOhGgp5o+7PQrIz484Qo+o3bGy54APxgMyqABuVjACXrx1qiz61drnoQlSbjudlRulXc7kZYELYAAPmAy2PVF0QsXY4Wp9P3E3wUKBzHPXySX4d7uJUxBJZedR71xkwlQ8STmEnh21hfp+LL7um1+f67gkLbtqw37uAkq5-DumpKL4O4MGYkRgJQTrGD4CF-qAIAQeAaxMIwMq6PQ4RoPQmBgD8AAGxAItReHKH+dEJAALGIEbuJqu5IShrCnH+NhEbyr6JAwjA3qoIn8qmaxwAAcgAygAKnY8nKam6q7tx05CMaYCYGaFpWjama6Jh373tQaHOpx05CVJYCsGmDk5sCdDXuQljAnZJROoaxqmualrWraWbOHqqg-DZqjBpFPysEcKi7oZwUmWFaDUHRO4EKg9BgagVjGAAesVGpKEAA).

## After

```typescript
export declare type GetPayloadResultExtensionKeys<
  R extends InternalArgs['result'][string],
  KR extends keyof R = string extends keyof R ?
    never : 
    keyof R,
> = KR

export declare type GetPayloadResultExtensionObject<R extends InternalArgs['result'][string]> = {
  [K in GetPayloadResultExtensionKeys<R>]: R[K] extends () => {
    compute: (...args: any) => infer C
  }
    ? C
    : never
}

export declare type GetPayloadResult<
  Base extends Record<any, any>,
  R extends InternalArgs['result'][string],
> = Omit<
  Base,
  GetPayloadResultExtensionKeys<R>
> &
  GetPayloadResultExtensionObject<R>
```

The `GetPayloadResultExtensionKeys` type extracts the keys (`KR`) from the result extension `R`. It ensures that only valid keys (those present in `R`) are considered. If `R` doesn’t have specific keys or if string is a valid key of `R`, `KR` defaults to `never`, i.e., it shows no keys can be extracted.

The `GetPayloadResultExtensionObject` type constructs a new object type based on the keys extracted by `GetPayloadResultExtensionKeys`. For each key `K` in `KR`:
- It checks if `R[K]` is a function that returns an object containing a `compute` method. If so, it infers the return type `C` of `compute`.
- Otherwise, if `R[K]` doesn’t match this pattern, the value is set to `never`, meaning it’s an invalid extension.

The `GetPayloadResult` type combines the base model (`Base`) with the extensions defined in `R`:
- `Omit<Base, GetPayloadResultExtensionKeys<R>>`: First, it removes any keys from `Base` that are also defined in `R` (result extensions).
- `GetPayloadResultExtensionObject<R>`: Then, it adds the computed types for these keys from the extension (`R`).

Before, `GetPayloadResult` handled these cases in a single opaque step (`K in KR | keyof Base`).

**Example**: [TypeScript Playground](https://www.typescriptlang.org/play/?#code/PQKgBAggYgKgogJTACwC6oA4GcBcxgDmAlqsgK4BGAdAMYD2AtsBgE5FYMCGzbH3GZADaDgAJgCsABgDsksCGAAoRQFMAHhjotUYVAE8MKsAEkAdqhUtTnQRBYEsAHkVgwSALxgA3mADaAaTAiUzAsVDZTAgBdHG8-QODQ8ODo2LJTAGtTOgB3EIBfMHyAGhcwAFkwTx8AoJCwiNS42sSGlJiwdKzcgqLS1wBFKuaE+uTIjprRpMaOruy8or6ygGFhqbqZ9rTMhYLSgD51spYVLCFUWI3EjJU9OgAzN0m-AAVN2-unhACojoAKACUVSOP38UV8ryiS3yZQYdAAJipBFd4h87o8Ki9IeivhVfgDge4juVfpDoYVYa4AI5kSx6VEtEKfTEDbHvG4Yp4DAmxIEgsA88HkmFlGiCIgqcyM6Ysp4rQkCla-IqKWGqDRaHRI8WcU66AxGADiKlQr04ekEdE4CIQZwucDUFlMWCIdFM-juTjKSHUzoRWBM5ks1ls9iwvgA5KdzoJUJGIW0Jv0wP5fU6pQGwHK3MMkwQwH7M4Gc0gAPxlVymFQAN0sYFilezXLch2GaeU6k02jAOsEeqM+kMYBNZotVptdtjqEdztd7oA8hQAFYqGioRzp-2BswWKw2OwOKMxi4J3z5qJHaplJkj03my3W232uOzqXzj1ezcHDpg6FF0ws35YlvCbegGAECw+SoGC9QcWJOFMPQiSOYIHnrFYyipVwwDLMBMJwhswGrOsWDVTtNR7PsBwNYdRwfCdn2nZxXAAIU4LAjAArM7XoFgEUcRC9GKMAhIOFMt2LIM91DQ8I2jF940TcZojbTwFwYEgWLAdjOJTejxyfKcHQzF03U-PQnAQA5FCOAAyMoDMfSdFLfMzFxXNcN2s5RFC8MD3TCMAKA4oxr0I1xQBAJscPAFZGAYKUdHdXQ6AwMBMQAAyIBFMpi1wFHynLYgAFlEFNXHVQj6BdHQWGOCLiM4RK+RQuJwMglRWoFSMVgXAA5ABlGAIH6mBI1VQiqpwwRTTADBDJc6cGoimCqBCvT8usFqwBYKNtpUBMgVoRhOqBCqlg4u8x2cpiLkcIcVExDaVBEx7MRYGymwW27jLjKgcqbfAUDoUiUEsIwAD1IfIoA).

jkomyno (approved):

HRM:
Hello thanks for quality description. Yes everything you described is correct, next time I will provide a description in the same fashion. Sorry for the slow reaction time.

jkomyno:
> Hello thanks for quality description. Yes everything you described is correct, next time I will provide a description in the same fashion. Sorry for the slow reaction time.

No worries! The only concern we currently have is whether the new type definition will be resolved fast enough by the TypeScript compiler, especially when large Prisma schemas (like [odoo.prisma](https://github.com/prisma/prisma/blob/4904c9cbf837003f616ea76b7bb55346d567d3b5/packages/internals/src/__tests__/__fixtures__/odoo.prisma)) are involved.

Do you have experience writing TypeScript performance tests, using tools specified in https://github.com/microsoft/TypeScript/wiki/Performance#performance-tracing?

HRM:
I didn't do performance tracing so far, but I can do a few measurments with the prisma file you provided. I will add the results here.

jkomyno:
> I didn't do performance tracing so far, but I can do a few measurments with the prisma file you provided. I will add the results here.

Thank you! We're eventually looking to add this kind of test to this repository, for tracking changes to `@prisma/client` in future PRs. We encourage you to try performance tracing and report back with something we can reproduce step-by-step :)

Thanks so much for your contribution!

HRM:
# Benchmark

## How to reproduce

For testing purposes, I used the following simple cases.

With extensions:

```ts
import { PrismaClient } from ".prisma/client";

async function main() {
    result: {
      res_users: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
      base_import_import: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
      email_template_preview: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
      ir_mail_server: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
    },
  });

  let user = (await prisma.res_users.findFirst({
    include: {
      base_import_import_base_import_import_create_uidTores_users: true,
      email_template_preview_email_template_preview_create_uidTores_users: {
        include: { ir_mail_server: true },
      },
    },
  }))!;
  let a = user.test_field;
  let b = user.base_import_import_base_import_import_create_uidTores_users[0]
    .test_field;
  let c = user
    .email_template_preview_email_template_preview_create_uidTores_users[0]
    .test_field;
  let d = user
    .email_template_preview_email_template_preview_create_uidTores_users[0]
    .ir_mail_server?.test_field;
  let e = user.login;
}

void main().catch((e) => {
  console.log(e.message);
  process.exit(1);
});

```

Without extensions:

```ts
import { PrismaClient } from ".prisma/client";

async function main() {

    let user = (await prisma.res_users.findFirst({
        include: {
            base_import_import_base_import_import_create_uidTores_users: true,
            email_template_preview_email_template_preview_create_uidTores_users:
                {
                    include: { ir_mail_server: true },
                },
        },
    }))!;
    let a = user.id;
    let b = user.base_import_import_base_import_import_create_uidTores_users[0]
        .file_name;
    let c = user
        .email_template_preview_email_template_preview_create_uidTores_users[0]
        .email_from;
    let d = user
        .email_template_preview_email_template_preview_create_uidTores_users[0]
        .ir_mail_server?.active;
    let e = user.login;
}

void main().catch((e) => {
    console.log(e.message);
    process.exit(1);
});
```

I did the following:

1.  Create a copy of the `basic-sqlite` folder in the `sandbox` directory.
2. Replace the contents of the `index.ts` file with one of the codes provided above.
3. From the `tracing` folder, in the same `sandbox` directory, copy the `tsconfig.json` file into our newly created directory.
4. Copy the contents of [odoo.prisma](https://github.com/prisma/prisma/blob/4904c9cbf837003f616ea76b7bb55346d567d3b5/packages/internals/src/__tests__/__fixtures__/odoo.prisma) into `prisma\schema.prisma`.
5. Run `pnpm i` and `pnpm run generate`.
6. Run the following code `tsc -p tsconfig.json --generateTrace tracing_output_old` (the name of the output directory can be anything of course).
7. Open `edge://tracing/` or `chrome://tracing/` in edge or chrome depending on your preferred browser. On the top left side of the page there is a `Load` button. Press it then select the created `trace.json` file from your tracing output folder.
8. Run the [analyze-trace](https://github.com/microsoft/typescript-analyze-trace) trace command on the trace output folder.
9. Try out both the old and the new implementations, with and without extensions. (It should be enough to change the implementation inside `node_modules\.prisma\client\runtime\library.d.ts`)

## Results

all

left side: old implementation, right side: new implementation, bottom: without extension, top: with extension

The trace analyzer gave `No hot spots found` result for all the traces.

[Here](https://github.com/user-attachments/files/16773557/client-test.zip) you can download the results themselves.

## Interpretation of the results:

As visible in the image above, the difference between the old and the new implementation is within the margin of error. The `checkSourceFile` step (which is also selected on the picture) should be the most affected from the changes. The presence of extensions causes a much larger difference. In the variants with the extension, the aforementioned step takes more than twice as long to finish. I'm not sure how representative the results are, but the difference between the variants with and without extensions implies that if there were some significant scaling issues with the new implementation, the traces should show it.

HRM:
# Benchmark

## How to reproduce

For testing purposes, I used the following simple cases.

With extensions:

```ts
import { PrismaClient } from ".prisma/client";

async function main() {
    result: {
      res_users: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
      base_import_import: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
      email_template_preview: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
      ir_mail_server: {
        test_field: {
          compute(data) {
            return "this is a test";
          },
        },
      },
    },
  });

  let user = (await prisma.res_users.findFirst({
    include: {
      base_import_import_base_import_import_create_uidTores_users: true,
      email_template_preview_email_template_preview_create_uidTores_users: {
        include: { ir_mail_server: true },
      },
    },
  }))!;
  let a = user.test_field;
  let b = user.base_import_import_base_import_import_create_uidTores_users[0]
    .test_field;
  let c = user
    .email_template_preview_email_template_preview_create_uidTores_users[0]
    .test_field;
  let d = user
    .email_template_preview_email_template_preview_create_uidTores_users[0]
    .ir_mail_server?.test_field;
  let e = user.login;
}

void main().catch((e) => {
  console.log(e.message);
  process.exit(1);
});

```

Without extensions:

```ts
import { PrismaClient } from ".prisma/client";

async function main() {

    let user = (await prisma.res_users.findFirst({
        include: {
            base_import_import_base_import_import_create_uidTores_users: true,
            email_template_preview_email_template_preview_create_uidTores_users:
                {
                    include: { ir_mail_server: true },
                },
        },
    }))!;
    let a = user.id;
    let b = user.base_import_import_base_import_import_create_uidTores_users[0]
        .file_name;
    let c = user
        .email_template_preview_email_template_preview_create_uidTores_users[0]
        .email_from;
    let d = user
        .email_template_preview_email_template_preview_create_uidTores_users[0]
        .ir_mail_server?.active;
    let e = user.login;
}

void main().catch((e) => {
    console.log(e.message);
    process.exit(1);
});
```

I did the following:

1.  Create a copy of the `basic-sqlite` folder in the `sandbox` directory.
2. Replace the contents of the `index.ts` file with one of the codes provided above.
3. From the `tracing` folder, in the same `sandbox` directory, copy the `tsconfig.json` file into our newly created directory.
4. Copy the contents of [odoo.prisma](https://github.com/prisma/prisma/blob/4904c9cbf837003f616ea76b7bb55346d567d3b5/packages/internals/src/__tests__/__fixtures__/odoo.prisma) into `prisma\schema.prisma`.
5. Run `pnpm i` and `pnpm run generate`.
6. Run the following code `tsc -p tsconfig.json --generateTrace tracing_output_old` (the name of the output directory can be anything of course).
7. Open `edge://tracing/` or `chrome://tracing/` in edge or chrome depending on your preferred browser. On the top left side of the page there is a `Load` button. Press it then select the created `trace.json` file from your tracing output folder.
8. Run the [analyze-trace](https://github.com/microsoft/typescript-analyze-trace) trace command on the trace output folder.
9. Try out both the old and the new implementations, with and without extensions. (It should be enough to change the implementation inside `node_modules\.prisma\client\runtime\library.d.ts`)

## Results

all

left side: old implementation, right side: new implementation, bottom: without extension, top: with extension

The trace analyzer gave `No hot spots found` result for all the traces.

[Here](https://github.com/user-attachments/files/16773557/client-test.zip) you can download the results themselves.

## Interpretation of the results:

As visible in the image above, the difference between the old and the new implementation is within the margin of error. The `checkSourceFile` step (which is also selected on the picture) should be the most affected from the changes. The presence of extensions causes a much larger difference. In the variants with the extension, the aforementioned step takes more than twice as long to finish. I'm not sure how representative the results are, but the difference between the variants with and without extensions implies that if there were some significant scaling issues with the new implementation, the traces should show it.

---

Maybe a larger test case on the ts side would produce

HRM:
It came to my mind that there is a chance generic types are not checked until they are concretized, so I created a new test, this time it includes two queryes for every model, one without related models and one with all the related models (where it is aplicable). Of course, writing this by hand in the case of a large schema is inconvenient, so I wrote a generator that generates two files based on the scheme `test.ts` and `test-extended.ts`. In the extended version every model's query result is extended with a test field.


---
## Setup instructions

1. Download this file: [client-ts-test.zip](https://github.com/user-attachments/files/16790428/client-ts-test.zip)
2. Extract it into the `sandbox folder`
3. Run `pnpm i` and `pnpm run generate`, the later command will create the above mentioned two test files.
6. Run the following two lines `tsc -p tsconfig.test.json --generateTrace tracing_output`, `tsc -p tsconfig.test-extended.json --generateTrace tracing_output_extended`
7. Open `edge://tracing/` or `chrome://tracing/` in edge or chrome depending on your preferred browser. On the top left side of the page there is a `Load` button. Press it then select the created `trace.json` file from your tracing output folder.
8. Run the [analyze-trace](https://github.com/microsoft/typescript-analyze-trace) trace command on the trace output folder.
9. Try out both the old and the new implementations. (It should be enough to change the implementation inside `node_modules\.prisma\client\runtime\library.d.ts`)

## Results

results_with_include

left side: new implementation, right side: old implementation, bottom: with extension, top: without extension

The trace analyzer marked all the test files as hot spots (both: new, old, extended, not extended), but nothing else.

I would say the difference is still negligible, for the whole compilation, and for source file checking. I did a few runs, a few milliseconds here and there, but all the results hover around the one shown above.

jkomyno:
Hi @HRM, thanks for taking the time for this trace benchmark! I agree that the difference in compilation time is negligible for this fix. @millsp, what do you think, can we merge this one?

Issue #19367, opened 2023-05-21
Mongo replica set dockerfile giving error on mongo 6.0


trolliama:
### Bug description

I am using the mongodb version 6 and because of this I've made some modification to the dockerfile. 
With those modifications the container is throwing this error:

`MongoServerError: already initialized`

### Expected behavior

Change the current repository dockerfile to this one:

```:dockerfile
FROM mongo:6

# we take over the default & start mongo in replica set mode in a background task
ENTRYPOINT mongod --port $MONGO_REPLICA_PORT --replSet rs0 --bind_ip 0.0.0.0 & MONGOD_PID=$!; \
# we prepare the replica set with a single node and prepare the root user config
INIT_REPL_CMD="rs.initiate({ _id: 'rs0', members: [{ _id: 0, host: '$MONGO_REPLICA_HOST:$MONGO_REPLICA_PORT' }] })"; \
INIT_USER_CMD="db.getUser('$MONGO_INITDB_ROOT_USERNAME') || db.createUser({ user: '$MONGO_INITDB_ROOT_USERNAME', pwd: '$MONGO_INITDB_ROOT_PASSWORD', roles: [ 'root' ] })"; \
# we wait for the replica set to be ready and then submit the commands just above
until (mongosh admin --port $MONGO_REPLICA_PORT --eval "$INIT_REPL_CMD && $INIT_USER_CMD"); do sleep 1; done; \
# we are done but we keep the container by waiting on signals from the mongo task
echo "REPLICA SET ONLINE"; wait $MONGOD_PID;
```
> I have chanaged the version at the top to the sixth version and changed the line 9 to use `mongosh`instead of `mongo` as it is required by the version 6

After that build it and run it

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
// Add your code using Prisma Client
```

### Environment & setup

- OS: Ubuntu
- Database: MongoDB 6.0
- Node.js version: 18

### Prisma Version

```
4.14.1
```

janpio:
Do you refer to https://github.com/prisma/prisma/blob/main/docker/mongodb_replica/Dockerfile?

millsp:
Thanks for the report @trolliama, here's the updated Dockerfile

```sh
FROM mongo:6

# we take over the default & start mongo in replica set mode in a background task
ENTRYPOINT mongod --port $MONGO_REPLICA_PORT --replSet rs0 --bind_ip 0.0.0.0 & MONGOD_PID=$!; \
[...]

PR #25120, opened 2024-08-29
fix(client): mongodb_replica dockerfile with replica status check to …

rabinthapa01:
### Description

The Docker build works as expected on the initial run. However, subsequent runs encounter a `MongoServerError: already initialized` error.

### Related Issue

This addresses the issue documented here: [Prisma Issue #19367](https://github.com/prisma/prisma/issues/19367).

### Changes Made

- Added a check for the replica set status to conditionally initialize the replica set, preventing the `MongoServerError` on subsequent runs.


aqrln:
All tests are failing with timeout with this change.

@RabinThapa1998 I'm assuming you're not interested in working on this anymore given that the PR has been in a broken state since August last year and still isn't ready for review. If that's not the case and you'd like to give it another try, feel free to reopen.

PR #25147, opened 2024-09-02
fix(client): render not is required by optional property

### LINKED ISSUE #24527: Not possible to omit optional fields in Mongo
### Bug description

When defining some nested property as optional, TypeScript actually makes you pass `null` to it, and doesn't allow you to omit it.

### How to reproduce

For this prisma.schema:

```
type Nested {
  something   String?
}

type SomeType {
  nested Nested?
}
```

it generates TS types that look like

```
interface Nested {
  something: null | string
}
```

### Expected behavior

I'd expect these types to be generated:

```
interface Nested {
  something?: null | string
}
```

Explicitly passing `null` for optional fields spams db with these nulls, makes the api cumbersome to work with etc.

### Prisma information

see above

### Environment & setup

Macos, node 20

### Prisma Version

5.14.0

zzau13:
Fixes #24527


zzau13:
What is the criterion for using undefined and null in generated models?

wmadden:
Hi @zzau13, 

Since this PR has been open for a few months without changes and if I understand your question correctly, you're questioning the purpose of the fix, I'm going to close this PR.

If you fix the tests or you feel this is a mistake, please don't hesitate to reopen this.

PR #25180, opened 2024-09-05
feat(migrate): add `--skip-empty` flag to skip empty migrations when using `--create-only`

sidharthv96:
Adds a `--skip-empty` flag to `prisma migrate dev`, which can be used when we don't want `--create-only` to create empty migration files. 

Fixes: https://github.com/prisma/prisma/discussions/9691#discussioncomment-10294922

## Use case

We are using supabase, and have a script which automatically enables RLS on all newly created tables, by appending SQL to the migration created by prisma. 
In the current situation, if we migrate, run rls, and then do a migration later, all the data needs to be reset, as the applied migration was modified by the rls script. 
By having a `--skip-empty` flag, we can first create the migration (if necessary), append RLS SQL, and then run `migrate dev` to actually do the migration. 

## Alternatives considered

Add RLS to a new migration file. But this would add numerous unnecessary migrations, as we would have to create a new migration file for every table we create.

I will add tests and documentation, if the proposed approach is acceptable.

### DIFF
--- a/packages/migrate/src/commands/MigrateDev.ts
+++ b/packages/migrate/src/commands/MigrateDev.ts
@@ -53,8 +53,10 @@ ${bold('Options')}
        -n, --name   Name the migration
     --create-only   Create a new migration but do not apply it
                     The migration will be empty if there are no changes in Prisma schema
+                    Use --skip-empty to skip creating empty migration
   --skip-generate   Skip triggering generators (e.g. Prisma Client)
       --skip-seed   Skip triggering seed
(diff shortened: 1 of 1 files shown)


wmadden:
Hi @sidharthv96,

Thank you for your contribution and for proposing the --skip-empty flag to the prisma migrate command. After careful consideration, we’ve decided not to implement this change as proposed. We plan to decompose the migrate dev command into smaller operations, allowing users to achieve the same result without adding another flag. Therefore, we’re closing this PR. We appreciate your understanding and look forward to your future contributions.

PR #25181, opened 2024-09-05
[WIP]: Ensure client can correctly find query engine file after bundling

### LINKED ISSUE #18069: `@prisma/webpack-plugin` is not able to handle multiple clients
### Bug description

`@prisma/webpack-plugin` is not able to handle multiple generated clients in a monorepo setup, which are then most commonly bundled together via a Next.js app. The current logic poorly attempts to manage multiple schemas, and thus ends up updating the paths to only one schema across all the bundled clients. This obviously does not work and leads to various errors depending on how much the two schemas differ.

https://github.com/prisma/prisma/issues/12853#issuecomment-1416383546
https://github.com/prisma/prisma/issues/12921#issuecomment-1437345309
https://github.com/prisma/prisma/issues/17687#issuecomment-1438256384
https://github.com/prisma/prisma/issues/17687#issuecomment-1440826128

### How to reproduce

https://github.com/trentonsnyder/nextjs-prisma-repro (thanks @trentonsnyder)

### Expected behavior

Multiple schemas can be handled without any issue when multiple clients are bundled.

### Prisma information

n/a

### Environment & setup

- OS: n/a
- Database: n/a
- Node.js version: n/a

### Prisma Version

```
4.11.0-dev
```

SevInf:
This extends engine lookup logic via some of the bundler-specific constructs. In particular, webpack, parcel and vite recognize `new URL('./relative-path', import.meta.url)` call. If encountered, bundlers will:

- copy referenced file to the correct location.
- renames according to the rules.
- replaces `new URL(..)` with a correct path pointing to a copied file.

Without the bundling, this construct will just create a file url, relative to the current module and will work as expected.

This PR adjust generation to include the mapping between binary target name and and an absolute path, created via `new URL(..., import.meta.url)` pattern. This map will look similar to this:

```js
export default {
  ["darwin-arm64"]: () => new URL("./libquery_engine-darwin-arm64.dylib.node", import.meta.url).pathname,
}
```

with one entry per configured binary target (or default one if none are configured explicitly).

Engine resolution logic is now adjusted to look at the path as the first option and fall back to a previous logic only if the engine does not exist there.

This allows us to:
- work in next-js monorepos without the need for `@prisma/nextjs-monorepo-workaround-plugin`
- works with non-next.js webpack projects without `CopyWebpackPlugin`
- work with vite or parcel without additional instructions about copying the engine.
- have multiple different engine versions bundled correctly without the conflicts.

There were a few hurdles to implementation though:
- `import.meta.url` expression is syntactically valid only in ESM context. As of now, generated client is a CJS module and if use it there directly, node will throw syntax error upon parsing the file and refuse to execute it. To work around that, we put binary targets map into separate `.mjs` file and load it dynamically via `import(...)`. We won't need to do this when we have proper ESM support.

- dynamic `import()` does not work in jest without `--experimental-vm-modules` flag. To continue work with Jest out of the box we ignore new code path if we detect that we are running from jest and fall back to the old logic.
- `esbuild` when outputting a CJS module will replace `import.meta` with an empty object and so, our `new URL()` construct will throw an error and won't work. To avoid this, we check if `import.meta.url` is defined and if it is not, return and empty `{}` instead of a map.

Close prisma/team-orm#1272
Fix #18069

### DIFF
--- a/packages/client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/library.test.ts
+++ b/packages/client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/library.test.ts
@@ -52,10 +52,10 @@ testIf(!process.env.PRISMA_QUERY_ENGINE_LIBRARY)('missing-engine-native-binaryTa
     Please help us by answering a few questions: https://pris.ly/engine-not-found-tooling-investigation
 
     The following locations have been searched:
-      /client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/node_modules/.prisma/client
-      /client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/node_modules/@prisma/client
-      /client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/node_modules/@prisma/client/runtime
-      /tmp/prisma-engines
-      /client/src/__tests__/integration/errors/missing-engine-native-binaryTarget"
+      /client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/node_modules/.prisma/client/libquery_engine-TEST_PLATFORM.LIBRARY_TYPE.node
+      /client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/node_modules/@prisma/client/libquery_engine-TEST_PLATFORM.LIBRARY_TYPE.node
+      /client/src/__tests__/integration/errors/missing-engine-native-binaryTarget/node_modules/@prisma/client/runtime/libquery_engine-TEST_PLATFORM.LIBRARY_TYPE.node
+      /tmp/prisma-engines/libquery_engine-TEST_PLATFORM.LIBRARY_TYPE.node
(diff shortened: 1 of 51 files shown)


davecarlson:
I am so happy you are working on this. THANK YOU.

davecarlson:
@jkomyno Is this no longer being worked on ?

We still have a requirement to pull in the workaround plugin for our apps atm ( 3 next 15 apps in a monorepo, without a shared package/database.

would love to be able to remove the plugin

jkomyno:
Hi @davecarlson, OP of this PR no longer works at Prisma, at as this PR was highly experimental, we're investigating an alternative strategy to improve the developer experience on Next.js monorepos. Thanks for your interest!

davecarlson:
> ting an alternative strategy to improve the developer experience on Next.js monorepos. Thanks for your interest!

Perfect, thnak you. So long as there's still an end game to remove this plugin, it's literally the only webpack plug in our whole stack, and everything else runs on turbobuild (these apps have to eject to the webpack config)

Issue #16309, opened 2022-11-16
Prisma is using the internal `Span` constructor


aqrln:
In `createSpan.ts`: https://github.com/prisma/prisma/blob/011e81d5d94fab836a0a7421a2c628937114f1f2/packages/engine-core/src/tracing/createSpan.ts#L30

However:

```ts
    /**
     * Constructs a new Span instance.
     *
     * @deprecated calling Span constructor directly is not supported. Please use tracer.startSpan.
     * */
    constructor(parentTracer: Tracer, context: Context, spanName: string, spanContext: api.SpanContext, kind: api.SpanKind, parentSpanId?: string, links?: api.Link[], startTime?: api.TimeInput, clock?: Clock);
```

Is this intentional (e.g., the other API can't do what we need), or is it just something that needs to be updated?

PR #25182, opened 2024-09-05
feat(instrumentation)!: make otel deps peer dependencies

### LINKED ISSUE #24373: Make `@prisma/instrumentation` dependencies peer dependencies
https://github.com/prisma/prisma/pull/23019 + https://github.com/prisma/prisma/pull/23971 switched `@prisma/instrumentation` dependencies to use open ranges instead of exact versions to fix https://github.com/prisma/prisma/issues/21473 and similar issues. This alleviates the issue but it's not a full solution. As per our conversation with @millsp, `@opentelementry/sdk-trace-base` and `@opentelementry/api` must be peer dependencies and not direct dependencies.

aqrln:
Make `@opentelemetry/api` a peer dependency and keep `@opentelemetry/instrumentation` a regular dependency in `@prisma/instrumentation`, as per the [Instrumentations Implementation Guide](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b09cb40242f09c0a52bacd15a7d8c84e954d57e0/GUIDELINES.md?plain=1#L49-L59).

Also make `@opentelemetry/sdk-trace-base` a peer dependency. Instrumentation must not depend on it at all, and won't after we solve https://github.com/prisma/prisma/issues/16309, but for now keeping it as a peer dependency makes it slightly less bad and will somewhat alleviate the problems that arise because of it.

Closes: https://github.com/prisma/prisma/issues/24373

### DIFF
--- a/packages/instrumentation/package.json
+++ b/packages/instrumentation/package.json
@@ -13,6 +13,9 @@
   },
   "bugs": "https://github.com/prisma/prisma/issues",
   "devDependencies": {
+    "@opentelemetry/api": "1.9.0",
+    "@opentelemetry/instrumentation": "0.53.0",
+    "@opentelemetry/sdk-trace-base": "1.26.0",
     "@prisma/internals": "workspace:*",
     "@swc/core": "1.6.13",
     "@types/jest": "29.5.12",
@@ -21,7 +24,7 @@
     "jest-junit": "16.0.0",
     "typescript": "5.4.5"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@opentelemetry/api": "^1.8",
     "@opentelemetry/instrumentation": "^0.49 || ^0.50 || ^0.51 || ^0.52.0 || ^0.53.0",
     "@opentelemetry/sdk-trace-base": "^1.22"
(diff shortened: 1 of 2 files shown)


packages/instrumentation/package.json
```diff
@@ -21,7 +24,7 @@
     "jest-junit": "16.0.0",
     "typescript": "5.4.5"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@opentelemetry/api": "^1.8",
     "@opentelemetry/instrumentation": "^0.49 || ^0.50 || ^0.51 || ^0.52.0 || ^0.53.0",
     "@opentelemetry/sdk-trace-base": "^1.22"
```
aqrln, line 30: This one should just be gone, an instrumentation package must not depend on the SDK. The users likely won't even use `sdk-trace-base` directly but `sdk-trace-node` or even a third-party SDK not based on `sdk-trace-base`. In the latter case we may break completely because we'll be using a completely wrong `Span` class. Removing this depends on https://github.com/prisma/prisma/pull/25164.

SevInf (approved):

PR #25197, opened 2024-09-09
chore(e2e): Update next.js to 14 in "schema-not-found" suite

SevInf:
Updates next.js and adjusts few things for tests to work afterwards:

- `app` directory is no longer experimental and does not need config
  option.
- `standalone` server location can now be different depending on
  repository structure.
- code inside of server components was not idempotent. Calling `create` from
  a server component have always been wrong, but so far we've been lucky
  and that caused no issues. That changed after update. Replaced `delete`/`create` calls
  with `SELECT 1` query instead to avoid consistency issues in case
  component rendering gets aborted at restarted.

Additionally, this adjusts e2e setup to be able to talk to the databases
running on a host machine on macOS as well - previously, this worked
only on Linux.

Close prisma/team-orm#1280


packages/client/tests/e2e/_utils/standard.cmd.sh
```diff
@@ -3,16 +3,16 @@
 set -euo pipefail
 
 # Get access to localhost
-export DOCKER_BRIDGE_IP=$(ip route | grep default | awk '{print $3}') # 172.17.0.1
+export DOCKER_BRIDGE_HOST="host.docker.internal"
```
SevInf, line 6: This change makes host access work correctly on both Mac and Linux. Probably Windows too, but I ahve not tested that

packages/client/tests/e2e/nextjs-schema-not-found/_shared/test.ts
```diff
@@ -42,26 +47,43 @@ async function test(endpoint: string) {
...
-export async function testServerComponents() {
-  process.env.WORKAROUND = 'true'
-  await test('test/42')
-  process.env.WORKAROUND = 'false'
-  await test('test/42')
+export async function getHttpCode(endpoint: string): Promise<number | undefined> {
+  return new Promise((resolve, reject) => {
+    http
+      .get(new URL(endpoint, 'http://localhost:3000'), (res) => {
+        res.resume() // consume body
```
jkomyno, line 64: Nice touch!

jkomyno (approved):
Very nice! 🚀

PR #25224, opened 2024-09-13
feat(client): implement `strictUndefinedChecks`

### LINKED ISSUE #20169: `where: { shop: undefined }`: The entire table data was accidentally deleted
### Bug description

```nodejs
let shop = undefined; // shop is variable, and in some cases, it may be undefined

await prisma.theme.deleteMany({
  where: {
    shop,
  }
})
```
I left out the shop parameter check, causing the shop to be undefined in some cases, triggering the deletion of the entire table. This operation of Prisma is very risky, I hope to improve it.

### LINKED ISSUE #20268: Select filter broke and fetched all data from DB
### Bug description

While the user was trying to fetch data that they created using https://tier-vercel-openai.vercel.app/ ([Repo](https://github.com/tierrun/tier-vercel-openai)), they were able to see data from all users in the DB from the table requested.

[Bug reported by the user](https://github.com/tierrun/tier-vercel-openai/issues/3), shows a video where all data from the table is fetched.

Code used to fetch

```ts
const res = await db.content.findMany({
    where: {
      userId: user?.id,
    },
    orderBy: {
      generatedAt: "desc",
    },
  });
```
You can see the snippet [here](https://github.com/tierrun/tier-vercel-openai/blob/def03fa0bf4131b2d3ac66869c9a738b78d5508a/src/app/(app)/history/page.tsx#L19C3-L26C6)

### How to reproduce

We were not able to reproduce the error at all, this was a one time thing which was recorded by our user as shown above.

Repro steps provided by customer.

1. Go to https://tier-vercel-openai.vercel.app/
2. Generate a copy content
3. Go to history
4. Be idle and let your computer go to sleep
5. When awake it should fetch all data from DB

### Expected behavior

It should only fetch data which satisfies the filter, or this will lead to massive data leak.

### Prisma information

```prisma

generator client {
  provider = "prisma-client-js"
}

datasource db {
  url               = env("POSTGRES_PRISMA_URL") // uses connection pooling
  directUrl         = env("POSTGRES_URL_NON_POOLING") // uses a direct connection
}

model Account {
  id                String   @id @default(cuid())
  userId            String
  type              String
  provider          String
  providerAccountId String
  refresh_token     String?  @db.Text
  access_token      String?  @db.Text
  expires_at        Int?
  token_type        String?
  scope             String?
  id_token          String?  @db.Text
  session_state     String?
  createdAt         DateTime @default(now()) @map(name: "created_at")
  updatedAt         DateTime @default(now()) @map(name: "updated_at")

  user User @relation(fields: [userId], references: [id], onDelete: Cascade)

  @@unique([provider, providerAccountId])
  @@index([userId])
  @@map(name: "accounts")
}

model Session {
  id           String   @id @default(cuid())
  sessionToken String   @unique
  userId       String
  expires      DateTime
  user         User     @relation(fields: [userId], references: [id], onDelete: Cascade)

  @@index([userId])
  @@map(name: "sessions")
}

model User {
  id            String    @id @default(cuid())
  name          String?
  email         String?   @unique
  emailVerified DateTime?
  image         String?
  createdAt     DateTime  @default(now()) @map(name: "created_at")
  updatedAt     DateTime  @default(now()) @map(name: "updated_at")

  accounts Account[]
  sessions Session[]
  contents Content[]

  @@map(name: "users")
}

model VerificationToken {
  identifier String
  token      String   @unique
  expires    DateTime

  @@unique([identifier, token])
  @@map(name: "verification_tokens")
}

model Content {
  id               String   @id @default(cuid())
  prompt           String
  generatedContent String?  @db.Text
  generatedAt      DateTime @default(now()) @map(name: "generated_at")
  userId           String
  user             User     @relation(fields: [userId], references: [id], onDelete: Cascade)

  @@index([userId])
  @@map(name: "posts")
}
```

```ts
const res = await db.content.findMany({
    where: {
      userId: user?.id,
    },
    orderBy: {
      generatedAt: "desc",
    },
  });
```

### Environment & setup

- OS: Mac OS 10.15.7
- Browser: Firefox

We do not have other details from our customer.

### Prisma Version

```
4.16.1
```

SevInf:
If preview feature is enabled, client-side validator will disallow
explicitly setting fields to `undefined`. Newly introduced `Prisma.skip`
could be used for skipping fields explicitly.

This should protect against accidental empty filters on destructive
operations, for example:

```ts
prisma.table.deleteMany({
  where: {
    // If `nullableThing` is `null` or `undefined`, this query will remove
    // all data from the table
    property: nullableThing?.property,
  }
})
```

Intention is to make this behaviour default in Prisma 6.

[Full spec](https://github.com/prisma/prisma/issues/20169#issuecomment-2338360300)

Close prisma/team-orm#1281
Close #20169
Close #20268

/integration


aqrln (approved):
Looks great to me 💯

comments are non-blocking

jkomyno (approved):

packages/client/src/runtime/core/types/exported/Skip.ts
```diff
@@ -0,0 +1,23 @@
+// private symbol that prevents direct construction of a class
+const secret = Symbol()
+
+export class Skip {
+  constructor(param?: symbol) {
+    if (param !== secret) {
+      throw new Error(`"Skip instance can not be constructed directly`)
```
SevInf, line 7:
```suggestion
      throw new Error(`Skip instance can not be constructed directly`)
```

Nope

PR #25263, opened 2024-09-23
test(e2e): Update types in e2e tests

SevInf:
Currently fails becuase the most recent typescript version fails on node
16 types.

----

Context: [Slack](https://prisma-company.slack.com/archives/C03R3PJN004/p1727099097079679).

For posterity, here's a short conversation we had on this topic.

@SevInf wrote:
> Ok, we are in pickle here. I think we have to either update node types to 18 in all e2e tests or downgrade tsc. Both are not ideal, but i think testing up-to-date typescript version is more important than types for node version that is EOL for year. WDYT?

@jkomyno wrote:
> I agree with your reasoning. Let's follow the bleeding edge version for tsc, and the lts version for Node.js


jkomyno:
It is still failing, e.g.: https://github.com/prisma/prisma/actions/runs/10995615963/job/30526875002?pr=25263#step:5:5200

jkomyno (approved):

PR #25270, opened 2024-09-24
feat(driver-adapters): add fallback to avoid breaking changes of `driverAdapters` preview feature in 5.20.0

jkomyno:
This PR avoids a breaking change to `driverAdapters` caused by https://github.com/prisma/prisma/pull/24878 and surfaced in [Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1727163471327249) / ecosystem-tests.

/integration

PR #25328, opened 2024-09-30
chore: revert fallback to `driverAdapters` breaking change, added via  #25270 in 5.20.0

jkomyno:
Reverts prisma/prisma#25270, as third-party `driverAdapters` contributors have been alerted of the API changes.

Context: [Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1727363441231399).


aqrln (approved):
We also need to update the adapters themselves in ecosystem-tests so that they aren't failing

PR #25483, opened 2024-10-22
fix(adapter-d1): throws Error when string column contains any ISO date

### LINKED ISSUE #25404: SQL Injection bug - D1 adaptor throws "Conversion failed: expected a datetime string in column" when string column contains any ISO date
### Bug description

The D1 adaptor attempts to find a valid ISO date in any 'string' db column (see https://github.com/prisma/prisma/blob/7aae7ec2146f03c58a639f69dfe48f093e1e8250/packages/adapter-d1/src/conversion.ts#L56) . If it doest, it attempts to convert it to a date format. The regex does not check to see if the ISO date is the only data in the column.

If a string column contains any ISO date along with other string data, any create or find will throw the following error:

Invalid `prisma.chat.create()` invocation: Inconsistent column data: Conversion failed: expected a datetime string in column

Note that this is potentially a serious issue as ISO dates are not sanitized by SQL libraries. If any user generated string contains a valid ISO date, e.g. 2024-10-09T16:05:08.547Z anywhere, it will cause this error to occur. 

### How to reproduce

```import { PrismaClient } from '@prisma/client'
import { PrismaD1 } from '@prisma/adapter-d1'

export default {
  async fetch(request, env, ctx) {
    // Setup Prisma Client with the adapter
    const adapter = new PrismaD1(env.MY_DB)

    const messages = "This is user input, 2024-10-09T16:05:08.547Z"

    // Execute a Prisma Client query
    const chat = await prisma.chat.create({
      data: {
       messages : JSON.stringify(messages),
      },
    })

    // Return result
    return new Response(usersCount)
  },
}
```

### Expected behavior

Data is written to database

But instead, fatal error is thrown - Invalid `prisma.chat.create()` invocation: Inconsistent column data: Conversion failed: expected a datetime string in column

### Prisma information

```// This is your Prisma schema file,

generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "sqlite"
}

model Chat {
  id          String    @id @default(cuid())
  createdAt   DateTime? @default(now())
  updatedAt   DateTime? @updatedAt
  messages    String?
}
```

### Environment & setup

- OS: macOS, cloudflare 
- Database: Cloudflare D1 
- Node.js version: v22.9.0

### Prisma Version

```
prisma                  : 5.20.0
@prisma/client          : 5.20.0
Node.js                 : v22.9.0
Query Engine (Node-API) : libquery-engine 06fc58a368dc7be9fbbbe894adf8d445d208c284 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli 06fc58a368dc7be9fbbbe894adf8d445d208c284 (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.20.0-12.06fc58a368dc7be9fbbbe894adf8d445d208c284
Default Engines Hash    : 06fc58a368dc7be9fbbbe894adf8d445d208c284
Studio                  : 0.502.0
```

Takur0:
This PR closes #25404


xPapla (approved):

aqrln:
Hey @Takur0, thanks so much for the PR, the change looks good to me. Would you mind adding a test as well?

Takur0:
Hi @aqrln, thank you for reviewing my PR! I'm happy to add a test for this change. I'll update the PR soon.

Takur0:
Hi @aqrln, I've added the test as requested. Please let me know if there's anything else you'd like me to adjust.

aqrln:
Thanks @Takur0, this looks great. My only comment is that although it's a regression test for a specific issue that happened with D1 adapter, I think there's still value in running it with all providers and adapters. Maybe some hypothetical future adapter could introduce a similar issue, and we'll be able to catch it early with the test already in place.

Takur0:
Thank you for your detailed comment, @aqrln ! I've updated the test code according to your suggestions. Please take a look!

packages/client/tests/functional/issues/25404/test.ts
```diff
@@ -0,0 +1,19 @@
+// @ts-ignore
+import type { PrismaClient } from '@prisma/client'
+
+// @ts-ignore
```
aqrln, line 4:
```suggestion
```

aqrln (approved):
Looks good to me, thanks!

acidjazz:
is this in a version I can use it? I'm not sure how to work around this

jkomyno:
> is this in a version I can use it? I'm not sure how to work around this

Hi, this issue was fixed in Prisma 5.22.0.

PR #25571, opened 2024-11-04
fix: allow batching of more than one query inside interactive transactions

### LINKED ISSUE #25570: Automatic batching doesn't work inside interactive transaction
### Bug description

I've encountered an issue where automatic batching (as described [here](https://www.prisma.io/docs/orm/prisma-client/queries/query-optimization-performance) ) in interactive transactions doesn't work as expected. This is a major issue for me as I'm running a graphQL server and processing each request inside a transaction so I can apply PostgreSQL RLS policies.
If I use the fluent API to query a single relationship, the batching happens as expected, but if I try to query more than one relationship, the batching doesn't happen.

 

### How to reproduce

I'm using a simple schema where there is a `User` model with two relationships `Post` and `Hat`:

```
datasource db {
  provider = "postgresql"
}

generator client {
  provider = "prisma-client-js"
}

model User {
  id             String           @id
  createdAt      DateTime         @default(now())
  email          String           @unique
  name           String?
  posts          Post[]
  hat            Hat?
}

model Post {
  updatedAt DateTime @updatedAt
  title     String   @db.VarChar(255)
  author    User?    @relation(fields: [authorId], references: [id])
  authorId  String?
}

model Hat {
  id     Int     @id @default(autoincrement())
  style  String?
  user   User?   @relation(fields: [userId], references: [id])
  userId String? @unique
}
```

If I run the following script, the batching doesn't occur:
```
const { PrismaClient } = require("@prisma/client");
const client = new PrismaClient();

const run = async () => {
    const ids = ['1', '2', '3', '4', '5'];
    const results = await client.$transaction(async (tx) => {
        return Promise.all([
            ...ids.map((id) => tx.user.findUniqueOrThrow({ where: { id } }).posts()),
            ...ids.map((id) => tx.user.findUniqueOrThrow({ where: { id } }).hat()),
        ]);
    })

    console.log(results);
};

run();
```

Note that if I remove the query for `hat` the batching works as expected.

### Expected behavior

I'd expect the batching to work regardless of how many relationships are qeuried.

### Prisma information

See reproduction above.

### Environment & setup

- OS: macOS Sonoma 14.6.1
- Database: PostgreSQL 11/14
- Node.js version: v20.11.0

### Prisma Version

```
prisma                  : 5.1.0
@prisma/client          : 5.11.0
Query Engine (Node-API) : libquery-engine a9b7003df90aa623086e4d6f4e43c72468e6339b (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli a9b7003df90aa623086e4d6f4e43c72468e6339b (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.1.0-28.a9b7003df90aa623086e4d6f4e43c72468e6339b
Default Engines Hash    : a9b7003df90aa623086e4d6f4e43c72468e6339b
Studio                  : 0.492.0
```

LucianBuzzo:
Fixes: #25570

This change fixes an issue where running different queries inside an interactive transaction in the same tick would not be batched together automatically by the engine.
By accounting for the protocolQuery when batching interactive transactions, the engine can optimize the queries correctly, as the queries will be batched by both the transaction ID and the query.

## Summary by CodeRabbit

* **Improvements**
  * Batching for interactive transactions now groups requests by transaction and protocol batch identifier when available for more precise execution.

* **Database**
  * Added a Comment model and a comments relation on User for test scenarios.

* **Tests**
  * Expanded interactive-transaction batching tests with detailed engine/batch metrics and PostgreSQL-only scenarios.


jkomyno:
> Fixes: #25570
> 
> This change fixes an issue where running different queries inside an interactive transaction in the same tick would not be batched together automatically by the engine. By accounting for the protocolQuery when batching interactive transactions, the engine can optimize the queries correctly, as the queries will be batched by both the transaction ID and the query.

Hi @LucianBuzzo, thanks for this PR! Do you think you can maybe add a few tests for this improvement in https://github.com/prisma/prisma/tree/main/packages/client/tests/functional/batching? You can run tests locally by cd'ing to `./packages/client` and running:

```
pnpm run test:functional --provider=${DB_PROVIDER} --client-runtime node --engine-type library --runInBand extensions
```

You can set `DB_PROVIDER` to e.g. `sqlite`, to make things easier.

LucianBuzzo:
@jkomyno No problem, I've added tests to check the base case of batching, and batching with multiple models queried.

packages/client/tests/functional/batching/tests.ts
```diff
@@ -79,6 +80,68 @@ testMatrix.setupTestSuite(
 
       await waitFor(() => expect(queriesExecuted).toBeGreaterThan(1))
     })
+
+    testIf(provider === Providers.POSTGRESQL)(
```
aqrln, line 84: Why only PostgreSQL?

aqrln: ah we're counting the number of queries, makes sense

SAGV:
@LucianBuzzo could you suggest the easiest way to monkeypatch prisma while this fix is on the way? We're facing a very similar issue with our own graphql+rls server unfortunately…

LucianBuzzo:
@SAGV You can see the patching operation here https://github.com/cerebruminc/yates/blob/v3.6.3/src/index.ts#L245
Getting this to function "natively" also required copying in some utility functions from the prisma client source: `getBatchId()` and `buildKeysString()`.

SAGV:
@LucianBuzzo wow, huge thanks! 👍🏻 👍🏻 👍🏻 

I've tried so many things yesterday trying to figure out why the batching is broken inside transactions… The patch has just made our graphql requests 10x faster!

CleanShot 2025-06-17 at 17 15 38@2x 

@aqrln even if there are 2 queries, it's still insanely faster…

packages/client/tests/functional/batching/tests.ts
```diff
@@ -18,30 +19,98 @@ const user2 = {
...
 
     beforeAll(async () => {
       prisma = newPrismaClient({
         log: [{ emit: 'event', level: 'query' }],
       })
 
+      // Count batching at the transport layer: this is what actually verifies that Prisma Client
+      // grouped multiple operations into a single engine roundtrip.
+
+      const engine = (prisma as any)._engine as any
```
jacek-prisma: To make it less brittle and more type safe, we could try to type it against `getPrismaClient`:

```suggestion
      const engine = (client as InstanceType<ReturnType<typeof getPrismaClient>>)._engine
```

packages/client/tests/functional/batching/tests.ts
```diff
@@ -56,7 +56,7 @@ testMatrix.setupTestSuite(
       // Count batching at the transport layer: this is what actually verifies that Prisma Client
       // grouped multiple operations into a single engine roundtrip.
 
-      const engine = (prisma as any)._engine as any
+      const engine = (client as InstanceType<ReturnType<typeof getPrismaClient>>)._engine
       if (!engine || typeof engine.request !== 'function' || typeof engine.requestBatch !== 'function') {
```
jacek-prisma: the `typeof` checks here are no longer needed
also some of the other checks and casting in this file should be unnecessary now, because engine is now precisely typed

jacek-prisma (approved):

PR #25590, opened 2024-11-05
feat(cli): add new promo survey for Prisma 5.22

jkomyno:
Context: [Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1730806441613279?thread_ts=1730792769.605449&cid=C058VM009HT).

### DIFF
--- a/packages/cli/src/__tests__/commands/Generate.test.ts
+++ b/packages/cli/src/__tests__/commands/Generate.test.ts
@@ -4,21 +4,26 @@ import { BaseContext, jestConsoleContext, jestContext } from '@prisma/get-platfo
 import { ClientEngineType, getClientEngineType } from '@prisma/internals'
 
 import { Generate } from '../../Generate'
-import { promotions } from '../../utils/handlePromotions'
+import { promotions, renderPromotion } from '../../utils/handlePromotions'
 
 const ctx = jestContext.new().add(jestConsoleContext()).assemble()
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -1,16 +1,24 @@
...
   it('should work with a custom output dir', async () => {
     ctx.fixture('example-project')
     const data = await ctx.cli('generate')
-    const stdout = data.stdout.replace(/Tip:.*/s, 'Tip: MOCKED RANDOM TIP')
+
+    // Replace any possible entry in `promotions`'s texts with a fixed string to make the snapshot stable
+    const stdout = Object.values(promotions)
+      .map(({ text }) => text)
+      .reduce((acc, curr) => {
+        return acc.replace(new RegExp(`${curr}.*`, 's'), 'Tip: MOCKED RANDOM TIP')
```
aqrln, line 20: This may explode with special characters in the tip that are valid regex syntax.

I think this should be enough:

```suggestion
        return acc.replace(curr, 'Tip: MOCKED RANDOM TIP')
```

jkomyno: That won't work, as links are appended to it. I can however simplify this by removing the regex altogether. Hold up

aqrln (approved):

PR #25591, opened 2024-11-05
fix: integrate backwards compatibility change for Optimize

aqrln:
Will be reverted after the release.


jkomyno (approved):
Please add a Linear task so we don't forget about this, thanks!

PR #25609, opened 2024-11-08
chore!: drop Node.js 16.x support and remove obsolete code

aqrln:
Prisma 6 will support Node.js 18.18 and later.

Since `fetch` is natively supported in these Node.js versions, the fallback implementation would never be used, and therefore is removed, together with all auxiliary types.

Using the native `Response` instead of our own changed the type of `Response#json()` to `Promise<unknown>` from `Promise<any>`. This uncovered that result handling in `DataProxyEngine` was almost completely untyped, which i fixed, together with some opportunistic refactoring along the way:
- Fixed inaccurate types of the query engine response.
- Moved the types for QE result extensions from `DataProxyEngine` to shared types, this will come in handy for Tracing GA, as we'll enable traces in response for other engine types too.
- Removed the dead code related to `elapsed` which has never even been a real feature, and is just some cruft left over from early Prisma 2.0 prototypes.

Closes: https://github.com/prisma/team-orm/issues/1340

### DIFF
--- a/.github/workflows/scripts/setup-mysql.sh
+++ b/.github/workflows/scripts/setup-mysql.sh
@@ -1,4 +1,4 @@
-#/bin/bash
+#!/usr/bin/env bash
 
 set -ex
 
--- a/.github/workflows/scripts/setup-mysql.sh
+++ b/.github/workflows/scripts/setup-mysql.sh
@@ -9,7 +9,7 @@ fi
 if [ "$RUNNER_OS" = "macOS" ]; then
     brew install mysql
 
-    cat <<EOF > /usr/local/etc/my.cnf
+    cat <<EOF > /etc/my.cnf
 [mysqld]
 bind-address = ::1
 mysqlx-bind-address = ::1
--- a/.github/workflows/scripts/setup-mysql.sh
+++ b/.github/workflows/scripts/setup-mysql.sh
@@ -11,7 +11,7 @@ if [ "$RUNNER_OS" = "macOS" ]; then
     brew services run mysql
 fi
 
-echo 'TEST_MYSQL_URI=mysql://root@localhost:3306/tests' >> "$GITHUB_ENV"
-echo 'TEST_FUNCTIONAL_MYSQL_URI=mysql://root@localhost:3306/PRISMA_DB_NAME' >> "$GITHUB_ENV"
-echo 'TEST_MYSQL_URI_MIGRATE=mysql://root@localhost:3306/tests-migrate' >> "$GITHUB_ENV"
-echo 'TEST_MYSQL_SHADOWDB_URI_MIGRATE=mysql://root@localhost:3306/tests-migrate-shadowdb' >> "$GITHUB_ENV"
+echo 'TEST_MYSQL_URI=mysql://root@127.0.0.1:3306/tests' >> "$GITHUB_ENV"
+echo 'TEST_FUNCTIONAL_MYSQL_URI=mysql://root@127.0.0.1:3306/PRISMA_DB_NAME' >> "$GITHUB_ENV"
+echo 'TEST_MYSQL_URI_MIGRATE=mysql://root@127.0.0.1:3306/tests-migrate' >> "$GITHUB_ENV"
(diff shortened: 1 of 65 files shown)


.github/workflows/scripts/setup-mysql.sh
```diff
@@ -7,7 +7,14 @@ if [ "$RUNNER_OS" = "Windows" ]; then
...
 
 if [ "$RUNNER_OS" = "macOS" ]; then
-    brew install mysql || true
+    brew install mysql
+
+    cat <<EOF > /opt/homebrew/etc/my.cnf
+[mysqld]
+bind-address = ::1
+mysqlx-bind-address = ::1
+EOF
```
aqrln, line 16: DNS resolution changed between Node.js 16 and 18, `localhost` now resolves to `::1` and not `127.0.0.1`, while the Homebrew-installed MySQL listens on IPv4 only by default.

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -134,6 +136,10 @@ async function main(): Promise<number | void> {
     }
   }
 
+  if (args['--inspect-brk']) {
+    jestCli = jestCli.withDebugger()
+  }
```
aqrln, line 141: Unrelated changes, I can split them off to a separate PR.

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -153,16 +153,15 @@ export class RequestHandler {
...
     const data = response?.data
-    const elapsed = response?.elapsed
 
     /**
      * Unpack
      */
     const result = this.unpack(data, dataPath, unpacker)
     if (process.env.PRISMA_CLIENT_GET_TIME) {
-      return { data: result, elapsed }
+      return { data: result }
```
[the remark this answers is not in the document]
jkomyno: Should this line change cause any updates in our docs?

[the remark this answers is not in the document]
aqrln: No, this is not a real feature that ever existed, just some leftover code from early Prisma 2 prototypes. `RequestHandler` is the last place where this value is accessible, it never reaches the `PrismaClient`.

jkomyno (approved):

packages/client/helpers/functional-test/run-tests.ts
jkomyno: I've figured they were unrelated, but it's fine to keep them here.

packages/client/src/runtime/utils/serializeRawParameters.ts
```diff
@@ -51,8 +51,7 @@ function encodeParameter(parameter: any, objectSerialization: 'fast' | 'slow'):
   if (isArrayBufferLike(parameter) || ArrayBuffer.isView(parameter)) {
     return {
       prisma__type: 'bytes',
-      // TODO Can we change this now in v16?
-      // TODO: node typings do not include ArrayBufferView as of 14.x
+      // TODO: node typings do not include ArrayBufferView as of 20.x
       prisma__value: Buffer.from(parameter as ArrayBuffer).toString('base64'),
```
jkomyno, line 55: Should we use `Uint8Array` to avoid explicit usage of `Buffer` here?

Something like:

```ts
function arrayBufferToBase64(arrayBuffer: ArrayBuffer): string {
  const uint8Array = new Uint8Array(arrayBuffer)
  const binaryString = uint8Array.reduce((acc, byte) => acc + String.fromCharCode(byte), '')
  return btoa(binaryString)
}

// Usage
{
  prisma__value: arrayBufferToBase64(parameter as ArrayBuffer)
}
```

aqrln: We have many other places where we use `Buffer` anyway, and the suggested `arrayBufferToBase64` function is too inefficient to be a replacement, although writing a function that directly transforms binary data to base64 is not a problem if we wanted to get rid of the `Buffer`s at some point. I wouldn't change it here in this PR though.

PR #25610, opened 2024-11-08
chore: switch to pnpm 9

aqrln:
With Node.js 16 being dropped, we can now upgrade pnpm.

Bundle size CI job is expected to fail in this PR, this is because it tries to use pnpm 9 with the code in `main`. It should work fine after this is merged.


aqrln:
I still want to do this, please don't close

jkomyno (approved):

PR #25734, opened 2024-11-27
fix: Specify 'type': 'commonjs' in generated client package.json

bartlomieju:
This commit changes the generated client's "package.json"
file to include `"type": "commonjs"`.

While in Node.js lack of `"type"` field implies CommonJS,
in Deno it defaults to ESM. This change will help fix
https://github.com/denoland/deno/issues/27079 and is completely
backwards compatible.

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -181,6 +181,7 @@ export async function buildClient({
 
   const pkgJson = {
     name: getUniquePackageName(datamodel),
+    type: "commonjs",
     main: 'index.js',
     types: 'index.d.ts',
     browser: 'index-browser.js',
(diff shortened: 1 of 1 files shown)


aqrln (approved):

packages/client/src/generation/generateClient.ts
```diff
@@ -181,6 +181,7 @@ export async function buildClient({
 
   const pkgJson = {
     name: getUniquePackageName(datamodel),
+    type: "commonjs",
```
aqrln, line 184:
```suggestion
    type: 'commonjs',
```

weird that the lint job on ci doesn't complain 🤔

bartlomieju:
Thanks for a quick merge!

PR #25756, opened 2024-11-29
feat(init): setup example .env files if they exist

### LINKED ISSUE #25655: feature request: prisma init: optionally setup the .env.example file, if it exists
## Problem

Currently when running `prisma init`, it appends the existing .env file if it exists or creates one. However, some repos have `.env.example` file, which is usually copied to create the .env file

## Suggested solution

Add the 'This was inserted by `prisma init`...' stuff to `.env.example` file, if it exists

sharunkumar:
Fixes #25655


aqrln:
Could you please add a test in [e2e tests](https://github.com/prisma/prisma/tree/main/packages/client/tests/e2e)?

sharunkumar:
> Could you please add a test in [e2e tests](https://github.com/prisma/prisma/tree/main/packages/client/tests/e2e?rgh-link-date=2024-11-30T12%3A01%3A58Z)?

let me know if the currently added test is sufficient

packages/client/tests/e2e/init-example-env/src/index.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const exampleEnvFiles = ['.env.example', '.example.env']
+
+exampleEnvFiles.forEach((file) => {
+  const fileContents = readFileSync(file, 'utf8')
+
+  if (!fileContents.includes('DATABASE_URL="file:./dev.db"')) {
+    console.error(`${file} does not contain env var from init command"`)
+    process.exit(1)
+  }
```
wmadden, line 11: This doesn't actually test that the example file was used. You probably need to assert that `FOO=BAR` is present.

wmadden (changes requested):

wmadden:
Hi @sharunkumar,

Thank you for your contribution and for proposing enhancements to the .env file setup during project initialization. After careful consideration, we’ve decided not to extend Prisma’s behavior around .env file management. Our future direction is to allow users to set up their own environment configuration loading behavior, without Prisma defaulting to loading it for them. Therefore, we’re closing this PR. We appreciate your understanding and look forward to your future contributions.

PR #25767, opened 2024-12-02
fix(client): revert `"type": "commonjs"` addition in generated `package.json`

### LINKED ISSUE #25761: Prisma 6.0.0 client with commonjs crashes on Next.js
### Bug description

After upgrading to 6.0.0, our Next.js with Prisma generated into a custom output folder crashes on dev mode.

The issue is caused by this pull-request : #25734 , Next.js talks about this error here: https://github.com/vercel/next.js/issues/35110#issuecomment-2016567633

Screenshot 2024-12-01 at 16 36 09

The error is caused when we're using `transpilePackages` in `next.config.js` looks like it's also transpiling the prisma client to esm or something like that.

I've checked and removing `type: commonjs` works fine, so not sure how to fix it. 

### Expected behavior

To not crash. 

### Prisma information

Prisma version: 6.0.0, with any prisma schema.

### Environment & setup

- OS:  Mac
- Database:  Any
- Node.js version:  22.11.0

### Prisma Version

```
6.0.0
```

aqrln:
Reverts prisma/prisma#25734 as discussed to fix https://github.com/prisma/prisma/issues/25761.

### DIFF
--- a/packages/client/src/generation/generateClient.ts
+++ b/packages/client/src/generation/generateClient.ts
@@ -181,7 +181,6 @@ export async function buildClient({
 
   const pkgJson = {
     name: getUniquePackageName(datamodel),
-    type: 'commonjs',
     main: 'index.js',
     types: 'index.d.ts',
     browser: 'index-browser.js',


wmadden (approved):

jkomyno (approved):

P4sca1:
I think specifying type: commonjs is totally valid and helps in environment like deno.
To me it seems like it is an issue with Next.js if packages that correctly specify type: commonjs (prisma is a CJS package) no longer work.

dodas:
This should have been fixed in Next.js rather than reverting a valid change in prisma.
Reverting this change just supports Next.js and their invalid handling of different package formats.

aqrln:
You both are right and I share the sentiment, but our goal is to not break the majority of our users, and realistically we have significantly more Next.js users. This is absolutely a bug that needs to be fixed upstream and it's unfortunate that we had to revert this to work around it (thus inconveniencing Deno users again), but we have to be pragmatic. If the breakage was introduced by a sudden change in Next.js without any changes on our side, then the considerations could have been different. Sorry for this, I understand that it is frustrating to you. We'll find a way forward.

semoal:
@dodas @P4sca1 you're absolutely right, I'm pushing the issue forward on with the Next.js team, once it's resolved upstream will ping the Prisma team, even I'll introduce the cjs change again to get this resolved asap!

mediaformat:
@semoal is there a tracking ticket or PR for this change?

PR #25775, opened 2024-12-02
fix(client): execute single-statement batch transactions correctly, allowing `SET ISOLATION LEVEL`

### LINKED ISSUE #17555: Batch `$transaction` with a single request is not executed within the transaction
```ts
prisma.$transaction([prisma.user.create()], { isolationLevel: Prisma.TransactionIsolationLevel.Serializable })
```

If there is only one request in a batch, it will never be executed within the transaction. Isolation level will not be set. This might be necessary, for example, for setting isolation level for raw queries or more complex prisma operations that result in more that DB query.

jkomyno:
This PR:
- fixes https://github.com/prisma/prisma/issues/17555
- deprecates https://github.com/prisma/prisma/pull/17556


jkomyno:
This PR turned out to be the wrong direction for solving the problem.

Issue #24715, opened 2024-07-05
Instrumentation: Problems in working with Elastic APM


samirotiv:
### Bug description

I am instrumenting our NestJS application using the Elastic APM OpenTelemetry Bridge. It seems to work fine with all other instrumentation libraries, such as Redis and NestJS, but is not working with Elastic.

Perhaps related to #14887 

When I enable Elastic APM OpenTelemetry Bridge and I enable Prisma Instrumentation, I get an error:

```
TypeError: parentTracer.getSpanLimits is not a function
    at new Span (/home/samir/git/dir/server/node_modules/@opentelemetry/sdk-trace-base/src/Span.ts:126:37)
    at /home/samir/git/dir/server/node_modules/@prisma/instrumentation/dist/chunk-VVAFFO6L.js:59:20
    at Array.forEach (<anonymous>)
    at ActiveTracingHelper.createEngineSpan (/home/samir/git/dir/server/node_modules/@prisma/instrumentation/dist/chunk-VVAFFO6L.js:44:27)
    at ro.createEngineSpan (/home/samir/git/dir/server/node_modules/@prisma/client/runtime/library.js:122:1645)
    at vr.logger (/home/samir/git/dir/server/node_modules/@prisma/client/runtime/library.js:112:1167)
    at /home/samir/git/dir/server/node_modules/@prisma/client/runtime/library.js:112:922
```

### How to reproduce

1. Set up Elastic APM either using a cloud trial or using Elastdocker
2. Instrument the code as shown below
3. Observe the logs

otel.ts:
```
import { NodeSDK } from '@opentelemetry/sdk-node';
import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-node';
import { getNodeAutoInstrumentations } from '@opentelemetry/auto-instrumentations-node';
import { registerInstrumentations } from '@opentelemetry/instrumentation';
import { PrismaInstrumentation } from '@prisma/instrumentation';

export function registerOtelInstrumentations() {
    registerInstrumentations({
        instrumentations: [
            getNodeAutoInstrumentations(),
            new PrismaInstrumentation(),
        ],
    });
    console.log('Registered instrumentation');
}
```

initelastic.ts:
```
import * as apm from 'elastic-apm-node';

import {
    APM_API_KEY,
    APM_SECRET_TOKEN,
    APM_SERVER_URL,
    APM_SERVICE_NAME,
} from './env';
import { registerOtelInstrumentations } from './otel';

[...]

PR #25828, opened 2024-12-10
fix: integrate new tracing model

### LINKED ISSUE #14887: Instrumentation error in production: `TypeError: parentTracer.getSpanLimits is not a function`
### Bug description

I am attempting to connect to `honeycomb.io` with the new Prisma Telemetry options. In a dev environment this works just fine. But in production, I receive the following error:

```
TypeError: parentTracer.getSpanLimits is not a function
  at new Span (/build/node_modules/@prisma/client/runtime/index.js:21694:37)
  at /build/node_modules/@prisma/client/runtime/index.js:23527:18
  at Array.forEach (<anonymous>)
  at createSpan (/build/node_modules/@prisma/client/runtime/index.js:23511:25)
  at runNextTicks (node:internal/process/task_queues:61:5)
  at listOnTimeout (node:internal/timers:528:9)
  at processTimers (node:internal/timers:502:7)
```

### How to reproduce

My code (after following these instructions: https://www.prisma.io/docs/concepts/components/prisma-client/opentelemetry-tracing#register-tracing-in-your-application)

```js
import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-proto";
import { PrismaInstrumentation } from "@prisma/instrumentation";
import { registerInstrumentations } from "@opentelemetry/instrumentation";
import { Resource } from "@opentelemetry/resources";
import { SemanticResourceAttributes } from "@opentelemetry/semantic-conventions";
import { SimpleSpanProcessor } from "@opentelemetry/sdk-trace-base";

function setupTelemetry() {
  if (process.env.NODE_ENV !== "production") {
    return;
  }
  const provider = new NodeTracerProvider({
    resource: new Resource({
      [SemanticResourceAttributes.SERVICE_NAME]: "SERVICE_NAME"
    })
  });

  // Setup how spans are processed and exported. In this case we're sending spans
  // as we receive them to an OTLP-compatible collector (e.g. Jaeger).
  provider.addSpanProcessor(new SimpleSpanProcessor(new OTLPTraceExporter()));

  // Register your auto-instrumentors
  registerInstrumentations({
    instrumentations: [new PrismaInstrumentation()],
    tracerProvider: provider
  });

  provider.register();
  // eslint-disable-next-line no-console
  console.log("📈 Telemetry set up");
}

export default setupTelemetry;
```

I then import `setupTelemetry` at the entrypoint for my application and execute it:

```js
import setupTelemetry from './server/telemetry.ts';
// ....
setupTelemetry();
```

### Expected behavior

In a production environment, I should not receive an error AND I should receive telemetry data to Honeycomb. Right now, I get the aforementioned error and receive no telemetry data.

In a development (local) environment, everything behaves as expected.

### Prisma information

Packages:
```
    "@prisma/client": "4.2.1",
    "@prisma/instrumentation": "4.2.1",
    "prisma": "4.2.1",
```

### Environment & setup

- OS: macOS (local) and Linux (Azure App Services)
- Database: Azure PostgreSQL DB
- Node.js version: v16.14.2 (locally and in Azure)

### Prisma Version

```
prisma                  : 4.2.1
@prisma/client          : 4.2.1
Query Engine (Node-API) : libquery-engine 2920a97877e12e055c1333079b8d19cee7f33826 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 2920a97877e12e055c1333079b8d19cee7f33826 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core 2920a97877e12e055c1333079b8d19cee7f33826 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt 2920a97877e12e055c1333079b8d19cee7f33826 (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Default Engines Hash    : 2920a97877e12e055c1333079b8d19cee7f33826
Studio                  : 0.469.0
```

### LINKED ISSUE #16309: Prisma is using the internal `Span` constructor

### LINKED ISSUE #17953: `prisma:engine` spans do not respect `suppressTracing()`

### LINKED ISSUE #19088: `tracing: engine` spans don't pass through `Sampler`
See [`tracing-no-sampling` test](https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/tracing-no-sampling/tests.ts).

It uses `new TraceIdRatioBasedSampler(0)` with 0% rate so it is expected that no spans will actually be sampled. But since we create engine spans via `Span` constructor, rather than `tracer.startSpan` method they never get passed through sampler and are always logged.

We can't use `startSpan` because it does not allow us to set `spanId`, which we need for forwarding engine spans to the client.

### LINKED ISSUE #19224: Add tests for tracing and different ways to open the first connection (including internal spans)
Currently there is unclarity how binary and library compare on tracing of new connections, either with or without an explicit `$connect()`.
We should add tests to https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/tracing/tests.ts that ensure via snapshots that we understand and monitor what these traces look like. This should include the internal spans that communicate if a database connection was created, vs. if it was just gotten from the existing pool.

- library
	- implicit connect
	- explicit `$connect()`
- binary
	- implicit connect
	- explicit `$connect()`

Optimally the snapshots between both are identical.
Might be possible though that the binary start takes care of the first connection, which might not be surfaced the same way in tracing as in the library.

### LINKED ISSUE #21397: `parentTracer.getSpanLimits is not a function`
### Bug description

Upgrading to Prisma v5.4.1 caused the following error to appear

```
TypeError: parentTracer.getSpanLimits is not a function
    at new Span (/var/task/node_modules/@prisma/instrumentation/node_modules/@opentelemetry/sdk-trace-base/build/src/Span.js:59:41)
    at /var/task/node_modules/@prisma/instrumentation/dist/ActiveTracingHelper.js:59:20
    at Array.forEach (<anonymous>)
    at ActiveTracingHelper.createEngineSpan (/var/task/node_modules/@prisma/instrumentation/dist/ActiveTracingHelper.js:44:27)
    at Wi.createEngineSpan (/var/task/node_modules/@prisma/client/runtime/library.js:123:1687)
    at dr.propagateResponseExtensions (/var/task/node_modules/@prisma/client/runtime/library.js:112:4691)
    at Object.callback (/var/task/node_modules/@prisma/client/runtime/library.js:112:6604)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at dr.withRetry (/var/task/node_modules/@prisma/client/runtime/library.js:112:8713)
    at Object.singleLoader (/var/task/node_modules/@prisma/client/runtime/library.js:123:5333) {
  clientVersion: '5.4.1'
}
```

Before the upgrade, everything was running smoothly and the only change was upgrading Prisma

### How to reproduce

- Upgrade to Prisma v5.4.1 (you must have open telemetry configured too

### Expected behavior

_No response_

### Prisma information

```
"@prisma/client": "^5.4.1",
"@prisma/instrumentation": "^5.4.1",
"prisma": "^5.4.1"
```

### Environment & setup

- OS: macOs
- Database: PostgreSQL
- Node.js version: 18

### Prisma Version

```
8.11.0
```

### LINKED ISSUE #22450: Support tracers not derived from `opentelemetry-sdk-trace-base` (e.g. Datadog tracer)
### Bug description

When trying to use Prisma instrumentation with Datadog, an exception is thrown:

```
Fatal: Uncaught exception
    TypeError: parentTracer.getSpanLimits is not a function
```

This is due to `ActiveTracingHelper` assuming that the tracer is an instance of [sdk-trace-base Tracer](https://github.com/open-telemetry/opentelemetry-js/blob/3e5929132129ed6022adbd05d085b998cb03e3d5/packages/opentelemetry-sdk-trace-base/src/Tracer.ts) instead of a [generic Tracer interface from opentracing api](https://github.com/open-telemetry/opentelemetry-js/blob/3e5929132129ed6022adbd05d085b998cb03e3d5/api/src/trace/tracer.ts). The tracer is [being cast to the sdk-trace-base Tracer](https://github.com/prisma/prisma/blob/075d31287c90b757fd9bd8d9b36032e6349fa671/packages/instrumentation/src/ActiveTracingHelper.ts#L45) in `ActiveTracingHelper.ts`, which is a wrong assumption in case of Datadog, since they implement their tracer from scratch, not using the sdk-trace-base implementation as a base. So the end result is that `PrismaInstrumentation` works with `NodeTracerProvider` but not with `dd-trace`.

A proposed fix for this would be to stop casting Tracer to a specific implementation base from `sdk-trace-base` and use [Tracer.startSpan](https://github.com/open-telemetry/opentelemetry-js/blob/3e5929132129ed6022adbd05d085b998cb03e3d5/api/src/trace/tracer.ts#L39) instead of the [deprecated new Span() constructor](https://github.com/open-telemetry/opentelemetry-js/blob/3e5929132129ed6022adbd05d085b998cb03e3d5/packages/opentelemetry-sdk-trace-base/src/Span.ts#L89-L94). This should resolve the issue for Datadog tracer.

### How to reproduce

1. Set up Datadog tracing like this:

```
import { registerInstrumentations } from "@opentelemetry/instrumentation";
import { PrismaInstrumentation } from "@prisma/instrumentation";
import tracer from "dd-trace";

export const ddTracer = tracer.init();
const provider = new ddTracer.TracerProvider();

registerInstrumentations({
  tracerProvider: provider,
  instrumentations: [new PrismaInstrumentation()],
});

provider.register();
```

2. Start the app and perform an operation with PrismaClient
3. See error
```
Fatal: Uncaught exception
    TypeError: parentTracer.getSpanLimits is not a function
```

### Expected behavior

Using Datadog tracer shouldn't result in an exception.

### Prisma information

The exception starts to manifest itself first in `5.3.0-integration-engines-5-3-0-7-qe-remove-unsafe-log-callback-1eb2a814ca332858858b479ede3d45f80c68d293.1` version of prisma:

`5.3.0` -crashes
`5.3.0-dev.44` - crashes
`5.3.0-dev.22` - crashes
`5.3.0-dev.11` - crashes
`5.3.0-dev.9` - crashes
`5.3.0-dev.8` - crashes
`5.3.0-integration-engines-5-3-0-7-qe-remove-unsafe-log-callback-1eb2a814ca332858858b479ede3d45f80c68d293.1` - crashes
`5.3.0-dev.7` - works
`5.3.0-dev.6` - works
`5.2.0` - works

However the exception from `ActiveTracingHelper` is being thrown even when it works, just in earlier versions it seems that the exception is swallowed, and doesn't bubble up crashing the app.

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: v18.17.1

### Prisma Version

Crashes in all versions since `5.3.0-dev.8`, current version used is:
```
5.7.0
```

### LINKED ISSUE #24285: `@prisma/instrumentation` Type Error: `'InstrumentionNodeModuleDefintion' is not generic`
### Bug description

Installing latest Prisma cause TS type error: `'InstrumentionNodeModuleDefintion' is not generic`

Here's the definition inside the installed package (`..../PrismaInstrumentation.d.ts`):
```ts
declare class PrismaInstrumentation extends InstrumentationBase {
    constructor(config?: Config);
    init(): InstrumentationNodeModuleDefinition<unknown>[];
    enable(): void;
    disable(): void;
    isEnabled(): boolean;
}
```

### How to reproduce

Installation step from https://www.prisma.io/docs/orm/prisma-client/observability-and-logging/opentelemetry-tracing should be sufficient:

```bash
npm install prisma@latest --save-dev
npm install @prisma/client@latest --save
npm install @prisma/instrumentation@latest --save
```
and ensure `skipLibCheck` inside tsconfig are set to false.
### Expected behavior

either `InstrumentationNodeModuleDefinition` class implements a generic type or the init() type should not have `unknown` as its type declaration

### Prisma information

N/A

### Environment & setup

OS: Windows 11 + Ubuntu 20.04

### Prisma Version

```
prisma                  : 5.14.0
@prisma/client          : 5.14.0
Architecture            : x64
Node.js                 : v20.12.2
Query Engine (Node-API) : libquery-engine e9771e62de70f79a5e1c604a2d7c8e2a0a874b48 (at node_modules\@prisma\engines\query_engine-windows.dll.node)
Schema Engine           : schema-engine-cli e9771e62de70f79a5e1c604a2d7c8e2a0a874b48 (at node_modules\@prisma\engines\schema-engine-windows.exe)
Schema Wasm             : @prisma/prisma-schema-wasm 5.14.0-25.e9771e62de70f79a5e1c604a2d7c8e2a0a874b48
Default Engines Hash    : e9771e62de70f79a5e1c604a2d7c8e2a0a874b48
Studio                  : 0.500.0
Preview Features        : tracing
```

### LINKED ISSUE #24373: Make `@prisma/instrumentation` dependencies peer dependencies

### LINKED ISSUE #24488: [Instrumentation] `registerInstrumentations` uses the global provider instead of the one passed in
### Bug description

I want to sample the traces coming from from Prisma, but I do not want to sample the traces that we manually instrument via `wrapWithSpan`. When I call `registerInstrumentations` and pass a `tracerProvider` it is not used and Prisma instead uses the global provider.

I've also opened this as [an issue](https://github.com/open-telemetry/opentelemetry-js/issues/4783) on the opentelemetry package as I don't know where the fault lies

### How to reproduce

Create an open telemetry setup with 2 providers. Register one globally and pass the other as tracerProvider to registerInstrumentations

Then run a process which results in the instrumentation generating traces.

Our setup code here:
```ts
import * as api from "@opentelemetry/api";
import { AsyncHooksContextManager } from "@opentelemetry/context-async-hooks";
import { OTLPTraceExporter } from "@opentelemetry/exporter-trace-otlp-http";
import { registerInstrumentations } from "@opentelemetry/instrumentation";
import { Resource } from "@opentelemetry/resources";
import {
  BasicTracerProvider,
  BatchSpanProcessor,
  ConsoleSpanExporter,
  SimpleSpanProcessor,
  TraceIdRatioBasedSampler,
} from "@opentelemetry/sdk-trace-base";
import {
  SEMRESATTRS_DEPLOYMENT_ENVIRONMENT,
  SEMRESATTRS_SERVICE_NAME,
  SEMRESATTRS_SERVICE_VERSION,
} from "@opentelemetry/semantic-conventions";
import { PrismaInstrumentation } from "@prisma/instrumentation";
import { ReploError } from "schemas/errors";

import { gitSha, reploEnv } from "../env";

export function openTelemetrySetup() {
  // NOTE (Gabe 2024-05-07): a context manager is required to propagate the
  // context for span nesting
  const contextManager = new AsyncHooksContextManager().enable();
  api.context.setGlobalContextManager(contextManager);

  // NOTE (Gabe 2024-05-07): This uses OTEL_EXPORTER_OTLP_XXXX env vars for
  // configuration.
  const otlpTraceExporter = new OTLPTraceExporter();

  const defaultProvider = new BasicTracerProvider({
    resource: new Resource({
      [SEMRESATTRS_SERVICE_NAME]: "replo-publisher",
      [SEMRESATTRS_DEPLOYMENT_ENVIRONMENT]: reploEnv,
      [SEMRESATTRS_SERVICE_VERSION]: gitSha,
    }),
  });
  const sampledProvider = new BasicTracerProvider({
    // NOTE (Gabe 2024-05-07): Sampling is enabled here to limit the performance
    // impact.
    sampler: new TraceIdRatioBasedSampler(0.1),
    resource: new Resource({
      [SEMRESATTRS_SERVICE_NAME]: "replo-publisher-prisma",
      [SEMRESATTRS_DEPLOYMENT_ENVIRONMENT]: reploEnv,
      [SEMRESATTRS_SERVICE_VERSION]: gitSha,
    }),
  });

  if (reploEnv === "production") {
    sampledProvider.addSpanProcessor(new BatchSpanProcessor(otlpTraceExporter));
    defaultProvider.addSpanProcessor(new BatchSpanProcessor(otlpTraceExporter));
  } else {
    // NOTE (Gabe 2024-05-07): this can be uncommented to write traces to the
    // console during development.
    // const consoleExporter = new ConsoleSpanExporter();
    sampledProvider.addSpanProcessor(
      new SimpleSpanProcessor(new ConsoleSpanExporter()),
    );
    defaultProvider.addSpanProcessor(
      new SimpleSpanProcessor(new ConsoleSpanExporter()),
    );
  }

  defaultProvider.register();
  api.trace.setGlobalTracerProvider(defaultProvider);

  registerInstrumentations({
    instrumentations: [new PrismaInstrumentation()],
    // This sampledProvider is not used, instead the defaultProvider is
    tracerProvider: sampledProvider,
  });
}
const tracer = api.trace.getTracer("replo-publisher");

export async function wrapWithSpan<R>(
  fn: () => Promise<R>,
  name: string,
  parentContext: api.Context,
): Promise<R> {
  const span = tracer.startSpan(name, {}, parentContext);
  try {
    const result = await fn();
    span.setStatus({ code: api.SpanStatusCode.OK });
    return result;
  } catch (error: unknown) {
    let message: string | undefined;
    if (error instanceof ReploError) {
      message = error.message;
    }
    span.setStatus({
      code: api.SpanStatusCode.ERROR,
      message,
    });
    throw error;
  } finally {
    span.end();
  }
}

```

All of the prisma spans end up being written to the `defaultProvider` and looks like this:
```
{
  "resource": {
    "attributes": {
      "service.name": "replo-publisher",
      "telemetry.sdk.language": "nodejs",
      "telemetry.sdk.name": "opentelemetry",
      "telemetry.sdk.version": "1.24.1",
      "deployment.environment": "development",
      "service.version": undefined
    }
  },
  "traceId": "c599bf0c727b33e92c689937a80fb7ec",
  "parentId": "c12bcce8d51f28c7",
  "traceState": undefined,
  "name": "prisma:engine:response_json_serialization",
  "id": "be557a7fb7e2ccb4",
  "kind": 0,
  "timestamp": 1718071948974729,
  "duration": 17,
  "attributes": {},
  "status": { "code": 0 },
  "events": [],
  "links": []
}

```

### Expected behavior

The  Prisma instrumentation will only write to the provider passed as `tracerProvider`

### Prisma information

I don't believe this is relevant, but if desired I can share.

### Environment & setup

- OS: macOS
- Database: Postgres
- Node.js version: v20.12.1Ï

### Prisma Version

```
prisma                  : 5.13.0
@prisma/client          : 5.13.0
Node.js                 : v20.12.1
Query Engine (Node-API) : libquery-engine b9a39a7ee606c28e3455d0fd60e78c3ba82b1a2b (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli b9a39a7ee606c28e3455d0fd60e78c3ba82b1a2b (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 5.13.0-23.b9a39a7ee606c28e3455d0fd60e78c3ba82b1a2b
Default Engines Hash    : b9a39a7ee606c28e3455d0fd60e78c3ba82b1a2b
Studio                  : 0.500.0
Preview Features        : metrics, tracing
```

aqrln:
Integration PR for https://github.com/prisma/prisma-engines/pull/5049.

Fixes the following client side issues, in addition to the ones closed by the engines PR:

Fixes: https://github.com/prisma/prisma/issues/14887
Fixes: https://github.com/prisma/prisma/issues/16309
Fixes: https://github.com/prisma/prisma/issues/17953
Fixes: https://github.com/prisma/prisma/issues/19088
Fixes: https://github.com/prisma/prisma/issues/21397
Fixes: https://github.com/prisma/prisma/issues/22450
Fixes: https://github.com/prisma/prisma/issues/24373
Closes: https://github.com/prisma/prisma/pull/25182
Closes: https://github.com/prisma/prisma/issues/24373
Closes: https://github.com/prisma/prisma/issues/24285
Fixes: https://github.com/prisma/prisma/issues/24715
Closes: https://github.com/prisma/prisma/issues/19224
Fixes: https://github.com/prisma/prisma/issues/24488
Closes: https://github.com/prisma/team-orm/issues/1367


packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -113,6 +124,41 @@ export class LibraryEngine implements Engine<undefined> {
...
+      connect: this.withRequestId(engine.connect?.bind(engine)),
+      disconnect: this.withRequestId(engine.disconnect?.bind(engine)),
+      dmmf: engine.dmmf?.bind(engine),
+      metrics: engine.metrics?.bind(engine),
+      query: this.withRequestId(engine.query?.bind(engine)),
+      rollbackTransaction: this.withRequestId(engine.rollbackTransaction?.bind(engine)),
+      sdlSchema: engine.sdlSchema?.bind(engine),
+      startTransaction: this.withRequestId(engine.startTransaction?.bind(engine)),
+      trace: engine.trace?.bind(engine),
+    }
```
jkomyno, line 140: Why does this use `?` operators (e.g., in `engine.connect?.bind(engine)`)?

`QueryEngineInstance` has type

```typescript
export type QueryEngineInstance = {
  connect(headers: string, requestId: string): Promise<void>
  disconnect(headers: string, requestId: string): Promise<void>
 // ...
}
```

packages/instrumentation/src/ActiveTracingHelper.ts
```diff
@@ -51,45 +55,14 @@ export class ActiveTracingHelper implements TracingHelper {
...
-      const span = new SpanConstructor(
-        tracer,
-        ROOT_CONTEXT,
-        engineSpan.name,
-        spanContext,
-        spanKind,
-        engineSpan.parent_span_id,
-        links,
-        engineSpan.start_time,
-      )
```
jkomyno, line 85: So long, `SpanConstructor`! 🚀

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
aqrln: The type is not exactly correct, not all of them are actually present in all engines :(
For example `applyPendingMigrations` is only present in c-abi engine. `dmmf` and `sdlSchema` are not everywhere IIRC. I just put `?` everywhere instead of figuring it out properly and fixing the type, thanks for calling this out, it's just lazy coding. I'll fix it.

aqrln: 9debd4a54dab59b97bb148e34d5ce74d49bb8e91

jkomyno (approved):

PR #25901, opened 2024-12-20
feat: introduce prisma postgres datasource to init cli command

sampolahtinen:
This PR introduces the basic flow for provisioning a PPG instance via the CLI.

Next  up:
- Introduce interactive prompts
- Write created keys to `.env.prisma`
- Scaffold a schema.prisma

### How to test:
1. Follow the prisma [CONTRIBUTING.md](https://github.com/prisma/prisma/blob/main/CONTRIBUTING.md) for setting up the project.
2. You can execute the CLI command with:
```
pnpm --filter prisma prisma init --datasource-provider=prisma+postgres
```
3. If you want to run the command against locally running console, manually update the following URLs:
https://github.com/prisma/prisma/blob/ff5acca38bd59e5d90472f236d91d7188e955ae0/packages/cli/src/platform/_lib/pdp.ts#L5-L6

Closes PLT-5373
Closes PLT-5370


packages/cli/src/Init.ts
```diff
@@ -18,10 +18,15 @@ import { bold, dim, green, red, yellow } from 'kleur/colors'
 import path from 'path'
 import { match, P } from 'ts-pattern'
 
+import { poll } from './platform/_'
+import { credentialsFile } from './platform/_lib/credentials'
+import { successMessage } from './platform/_lib/messages'
 import { printError } from './utils/prompt/utils/print'
 
+type ConnectorTypeOrPrismaPostgres = ConnectorType | 'prismapostgres'
```
sampolahtinen, line 26: I tried adding the `prismapostgres` to ConnectorType definition [here](https://github.com/prisma/prisma/blob/a97413b1e14c24190cf85d2fd3fe2b3953ebc3de/packages/generator-helper/src/types.ts#L66-L73).

This produced a bunch of other type errors. I was hesitant to add `prismapostgres` through out the code base.

Should we? Or is this temporary local type fine?

packages/cli/src/Init.ts
```diff
@@ -327,6 +334,81 @@ export class Init implements Command {
...
+      // TODO: Add interactive login/signup prompt
+      if (!credentials) {
+        const authenticationResult = await PlatformCommands.loginOrSignup()
+        console.log(authenticationResult.message)
+      }
+
+      const platformToken = await PlatformCommands.getTokenOrThrow(args)
+
+      const workspaces = await PlatformCommands.Workspace.getUserWorkspaces({ token: platformToken })
+      const defaultWorkspace = workspaces.find((_) => _.isDefault)
```
ellotheth, line 364: Aw man could this be `Workspace.getDefaultOrThrow()`?

packages/cli/src/Init.ts
```diff
@@ -327,6 +334,81 @@ export class Init implements Command {
...
+        'Checking the status of Prisma Postgres instance...',
+      )
+      console.log(successMessage('Prisma Postgres provisioning complete'))
+
+      console.log('Creating PPG API key...')
+      const serviceToken = await PlatformCommands.ServiceToken.create({
+        token: platformToken,
+        environmentId: project.defaultEnvironment.id,
+        displayName: `database-setup-prismaPostgres-api-key`,
+      })
```
ellotheth, line 399: Does this need error handling?

packages/cli/src/platform/_lib/utils.ts
```diff
@@ -63,3 +65,32 @@ export const generateConnectionString = (apiKey: string) => {
...
+  const waitTime = new Date().getMilliseconds() + waitMs
+
+  const wait = () =>
+    new Promise((resolve) => {
+      setTimeout(resolve, waitMs)
+    })
+
+  let result = await fn()
+  while (!until(result)) {
+    if (waitTime > endTime) {
```
ellotheth, line 86: `waitTime` needs to change in every loop to be "current timestamp + `waitMs`", otherwise it's always "poll start + `waitMs`"

packages/cli/src/Init.ts
sampolahtinen: Yes! Good catch. I will chat with Luan about what would be the ideal behavior in terms of the flow.

sampolahtinen: I will keep it as throwing errors and implement error branching flow in a separate PR if we choose to do so.

packages/cli/src/Init.ts
```diff
@@ -327,6 +339,77 @@ export class Init implements Command {
...
+      if (!credentials) {
+        const authenticationResult = await PlatformCommands.loginOrSignup()
+        console.log(authenticationResult.message)
+      }
+
+      const platformToken = await PlatformCommands.getTokenOrThrow(args)
+
+      const defaultWorkspace = await PlatformCommands.Workspace.getDefaultWorkspaceOrThrow({ token: platformToken })
+
+      console.log('Creating a project and provisioning a Prisma Postgres© instance...')
```
ellotheth, line 370: `®` or `™` if we're going to put a thing here, I don't think it's a copyright 😅

ellotheth (approved):
I get a lil `undefined` console output at the end of the process, otherwise looks great:

```
Checking the status of Prisma Postgres© instance...
Success! Prisma Postgres© provisioning complete
Creating PPG API key...
Success! Project has been successfully created!
Your database URL is:
prisma+postgres://accelerate.prisma-data.net/?api_key=herpderp
undefined
```

packages/cli/src/Init.ts
sampolahtinen: I noticed this during a huddle as well. Just updated.

sampolahtinen:
> I get a lil `undefined` console output at the end of the process, otherwise looks great:

Fixed this one as well.

packages/cli/src/Init.ts
```diff
@@ -327,6 +341,99 @@ export class Init implements Command {
...
+        console.log(authenticationResult.message)
+      }
+
+      const platformToken = await PlatformCommands.getTokenOrThrow(args)
+      const defaultWorkspace = await PlatformCommands.Workspace.getDefaultWorkspaceOrThrow({ token: platformToken })
+      const regions = await getPrismaPostgresRegionsOrThrow({ token: platformToken })
+
+      const ppgRegionSelection = await select({
+        message: 'Select your region:',
+        default: 'us-east-1',
```
ellotheth, line 380: (You might want to set this up with fallback for the next available region, in case it's unavailable.)

sampolahtinen: The select component is smart enough. If the default option satisfies the disabled condition, the default value will be changed to next in the list 👍

jkomyno (approved):

PR #25911, opened 2024-12-23
feat(fetch-engine): download cache directory customization support

smertnik3sh:
This feature allows to customize prisma download cache directory with `PRISMA_DOWNLOAD_CACHE_DIR` environment variable.

We use the prisma-cli 'generate' command in our build process (we use [bazel](https://bazel.build/)). To make our builds as "hermetic" as possible, we want the all cache to be controlled by bazel.


packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -862,4 +863,41 @@ It took ${timeInMsToDownloadAllFromCache2}ms to execute download() for all binar
       expect(await getVersion(queryEnginePath, BinaryType.QueryEngineLibrary)).toContain(CURRENT_ENGINES_HASH)
     })
   })
+
+  describe(`env.PRISMA_DOWNLOAD_CACHE_DIR=${baseCustomBinaryCache}`, () => {
```
aqrln, line 867: Let's keep the names of tests and `describe` blocks static. This will cause troubles if, e.g., some (non-inline) snapshot tests are added here in the future since the name will be different on every machine.

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -862,4 +863,41 @@ It took ${timeInMsToDownloadAllFromCache2}ms to execute download() for all binar
...
+
+    afterAll(() => {
+      delete process.env.PRISMA_DOWNLOAD_CACHE_DIR
+    })
+
+    beforeEach(async () => {
+      await del(path.posix.join(baseCustomBinaryCache, '*engine*'))
+      if (fs.existsSync(path.posix.join(baseCustomBinaryCache, 'master'))) {
+        fs.rmSync(path.posix.join(baseCustomBinaryCache, 'master'), { recursive: true })
+      }
```
aqrln, line 880:
```suggestion
      fs.rmSync(path.posix.join(baseCustomBinaryCache, 'master'), { recursive: true, force: true })
```

packages/fetch-engine/src/__tests__/download.test.ts
```diff
@@ -862,4 +863,41 @@ It took ${timeInMsToDownloadAllFromCache2}ms to execute download() for all binar
...
+    })
+
+    test('check schema-engine will be downloaded into user-specified directory', async () => {
+      // const files = getFiles(baseCustomBinaryCache).map((f) => f.name)
```
aqrln, line 884:
```suggestion
```

aqrln:
Closing as stale and unmergeable. Can't finish this myself and merge it because the CLA isn't signed.

PR #25914, opened 2024-12-23
fix(client): allow "passing" across rpc boundary in cloudflare workers

millsp:
This PR enables an internal use case, which is to pass instances of `PrismaClient` across RPC boundaries between workers. I initially though proxies were not serializable across Workers, but after a few experiments it turns out they are. Prisma Client's proxy was actually missing a base Object prototype to achieve that. I think this change is a safe one, but let me know what you think.


aqrln (approved):

aqrln:
Nice!

jkomyno (approved):

PR #25968, opened 2025-01-03
feat(cli): report typescript version via `prisma --version`

FGoessler:
This PR:
- adds the TypeScript version to the `prisma --version` output
- closes https://github.com/prisma/team-orm/issues/1371

It also handles corner cases for Bun and Deno, although I did not add explicit test cases for this. I am not aware of test infrastructure for Deno and Bun to cover this easily either.

### DIFF
--- a/packages/internals/src/cli/getTypescriptVersion.ts
+++ b/packages/internals/src/cli/getTypescriptVersion.ts
@@ -1,8 +1,6 @@
 async function getProcessObject() {
-  if (typeof process === 'object') return process
-
-  // On deno 1.X the process global is not directly available but has to be imported explicitly.
   try {
+    // On deno 2.X `process` is directly globally available but not in deno 1.X!
     return await import('node:process')
   } catch (_) {
     return null
--- a/packages/internals/src/cli/getTypescriptVersion.ts
+++ b/packages/internals/src/cli/getTypescriptVersion.ts
@@ -1,0 +1,22 @@
+async function getProcessObject() {
+  if (typeof process === 'object') return process
+
+  // On deno 1.X the process global is not directly available but has to be imported explicitly.
+  try {
+    return await import('node:process')
+  } catch (_) {
+    return null
+  }
+}
+
+export async function getTypescriptVersion(): Promise<string> {
+  try {
+    // On node.js this will import the project defined typescript package.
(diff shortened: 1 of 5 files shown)


packages/internals/src/cli/getTypescriptVersion.ts
```diff
@@ -0,0 +1,22 @@
...
+  try {
+    // On node.js this will import the project defined typescript package.
+    // On bun this will import the bun bundled typescript package.
+    return (await import('typescript')).default.version
+  } catch (_) {
+    // On deno the dynamic import of typescript will fail but typescript is natively available and its version is available in process.versions.
+    // If none of this worked typescript is likely not installed / available.
+    return (await getProcessObject())?.versions.typescript || 'unknown'
+  }
+}
```
jkomyno, line 22: Can we use [subpath imports](https://nodejs.org/api/packages.html#subpath-imports) instead, and have different `getTypeScriptVersion.[].ts` files for each major runtime that behaves differently? You can refer to [1](https://jkomyno-armada-js-2024.vercel.app/25?clicks=1), [2](https://jkomyno-armada-js-2024.vercel.app/26) for additional resources.

E.g., you may have a `getTypeScriptVersion.deno.ts` for Deno only, and a `getTypeScriptVersion.ts` for Node.js & Bun. Ideally, you should also import `node:process` in both files, regardless of whether `process` exist as a global variable.

FGoessler: Just tried this out but had no success in verifying that it actually works as our whole sandbox infrastructure is fitted for node.js and am not able right now to run `prisma --version` via deno based on my local build.

I also feel like using subpath imports here might be a bit of an overkill and actually spreads complexity around from the single `getTypeScriptVersion.ts` file into the `package.json` and maybe even further into the build and import system. But it's also the first time I am using subpath imports myself so I likely not have the full picture.

jkomyno: Alright, thanks for trying. Let's then avoid using subpath imports (we already use them in `@prisma/client`, but in fact, it'd be the first time we'd use them within `@prisma/internals`.

jkomyno (approved):

PR #25971, opened 2025-01-03
fix: node 23 support

### LINKED ISSUE #25463: prisma init error (node v 23)
### Bug description

$ npx prisma init
(node:10452) ExperimentalWarning: Support for loading ES Module in require() is an experimental feature and might change at any time
Error: (0 , DSe.isError) is not a function

but is working version 21 

### Expected behavior

_No response_

### Prisma information

```prisma
// Add your schema.prisma
```

```ts
```

### Environment & setup

- OS: 
- Database: 
- Node.js version: 

### Prisma Version

```
    "prisma": "^5.21.1",
```

FGoessler:
This PR resolves https://github.com/prisma/prisma/issues/25463 and thereby introduces support for node.js 23.

Reason for the failure was the use of a deprecated node.js api. See https://github.com/nodejs/node/pull/52744.
I assume this was also just an import error in the first place as we use an internal `isError` helper function everywhere else in the codebase.

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -107,7 +107,7 @@ jobs:
       matrix:
         shard: ['1/5', '2/5', '3/5', '4/5', '5/5']
         queryEngine: ${{ fromJson(inputs.queryEngine) }}
-        node: [18, 20, 22]
+        node: [18, 20, 22, 23]
         previewFeatures: ['', 'relationJoins']
     env:
       NODE_OPTIONS: '--max-old-space-size=8096'
@@ -180,7 +180,7 @@ jobs:
       fail-fast: false
       matrix:
         shard: ['1/6', '2/6', '3/6', '4/6', '5/6', '6/6']
-        node: [18, 20, 22]
+        node: [18, 20, 22, 23]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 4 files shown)


jkomyno (approved):
Good catch!

Let's fix the failing test https://github.com/prisma/prisma/actions/runs/12600363914/job/35119300223?pr=25971 before merging.

FGoessler:
I dug into these memory tests and my conclusion is that the updates in node.js v23 change the V8 JIT compiler and optimization behavior. It now either kicks in way later or performs additional optimizations only after a lot of executions which slightly increase the memory footprint due to the additional generated compiled code.

> [!NOTE]
> Usually we run 1k iterations per test plus 200 "warmup iterations".


### Why do I believe this is the case?

Running the memory profiler over a longer set of runs (~10k runs in the below) shows some spikes of additional and not freed memory during the first 500 to 1500 iterations. All in `compiled code` and `system` categories:

> [!NOTE]
> Bar heights indicate amount of allocated memory. The grey part was freed again. Blue parts means it is still active and not garbage collected at the the end of the recording.

Screenshot 2025-01-06 at 07 54 16
Screenshot 2025-01-06 at 07 54 41

Compare this to a a node 22 run of the same tests:

Screenshot 2025-01-06 at 08 00 34

Running the tests with the `--jitless` flag instantly stabilizes them. The memory profiler also shows this nicely:

Screenshot 2025-01-06 at 07 58 53


> Heap Timeline Files for the record:
> [heap-timelines.zip](https://github.com/user-attachments/files/18314922/heap-timelines.zip)

### How can we fix this for these tests?

- A) Increase the number of warmup iterations.
- B) Run the memory tests with the `--jitless` flag.

For now I choose Option A) as it more closely mimics production behavior, althought it slightly increases test runtimes.

PR #26092, opened 2025-01-14
feat(query compiler): create ClientEngine

FGoessler:
This PR creates a new `Engine` interface implementation called `ClientEngine` that leverages the Rust engine only to create a query plan but executes it via the typescript based QueryInterpreter. 
Mostly refactoring work to tie together some existing pieces from the hackaton into our expected architecture. Removes some of the hackaton workarounds to expose the functionality via "shortcuts".

Note that most of the code in `ClientEngine` is actually just copied from `LibraryEngine`. Esp. all the initialization code. We decided to go with this duplication for now as things might change down the line and we don't want to create a premature abstraction.

Leverage this by using:

```
generator client {
  // ...
  engineType      = "client"
}
```

See https://github.com/prisma/language-tools/pull/1812 for adding it to the language tools for autocomplete etc..


packages/client/src/generation/generateClient.ts
```diff
@@ -728,6 +728,10 @@ function getNodeRuntimeName(engineType: ClientEngineType) {
     return 'library'
   }
 
+  if (engineType === ClientEngineType.Client) {
+    return 'library' // TODO: should be library or introduce client here, too?
```
FGoessler, line 732: @aqrln ? 🤔

packages/client/src/runtime/utils/validatePrismaClientOptions.ts
```diff
@@ -100,7 +106,7 @@ It should have this form: { url: "CONNECTION_STRING" }`,
       )
     }
 
-    if (getClientEngineType() === ClientEngineType.Binary) {
+    if (getClientEngineType(config.generator) === ClientEngineType.Binary) {
```
FGoessler, line 109: ℹ️ Actually fixing a bug here I think: It always fell through to the env vars or default value and fully ignored the values in the generator block. Likely not a big issue though as `library` was the default anyways.

packages/client/src/generation/generateClient.ts
aqrln: We should change the build script to produce new, separate runtime bundle(s) for `ClientEngine`. However, doing that might not make sense until we actually separate a new WASM library. Let's keep this as is and tackle it in a follow up task.

aqrln (approved):
A few non-blocking comments; this is already enough for us to start iterating on it

PR #26109, opened 2025-01-17
feat(compiler): ORM-467 implement client side transaction manager

FGoessler:
This PR implements transaction handling for query execution on the client side.

### DIFF
--- a/packages/client/src/runtime/core/engines/client/ClientEngine.ts
+++ b/packages/client/src/runtime/core/engines/client/ClientEngine.ts
@@ -2,7 +2,7 @@ import Debug from '@prisma/debug'
 import { type ErrorCapturingDriverAdapter } from '@prisma/driver-adapter-utils'
 import type { BinaryTarget } from '@prisma/get-platform'
 import { assertNodeAPISupported, binaryTargets, getBinaryTargetForCurrentPlatform } from '@prisma/get-platform'
-import { TracingHelper } from '@prisma/internals'
+import { assertNever, TracingHelper } from '@prisma/internals'
 import { bold, green, red } from 'kleur/colors'
 
 import { PrismaClientInitializationError } from '../../errors/PrismaClientInitializationError'
@@ -34,6 +34,7 @@ import { defaultLibraryLoader } from '../library/DefaultLibraryLoader'
 import { reactNativeLibraryLoader } from '../library/ReactNativeLibraryLoader'
 import type { Library, LibraryLoader, QueryEngineConstructor, QueryEngineInstance } from '../library/types/Library'
 import { wasmLibraryLoader } from '../library/WasmLibraryLoader'
+import { TransactionManager } from './TransactionManager'
 
 const CLIENT_ENGINE_ERROR = 'P2038'
 
@@ -64,6 +65,7 @@ export class ClientEngine implements Engine<undefined> {
   libraryLoader: LibraryLoader
   library?: Library
   driverAdapter: ErrorCapturingDriverAdapter
+  transactionManager: TransactionManager
   logEmitter: LogEmitter
   libQueryEnginePath?: string
   binaryTarget?: BinaryTarget
@@ -135,6 +137,8 @@ export class ClientEngine implements Engine<undefined> {
     }
 
     this.libraryInstantiationPromise = this.instantiateLibrary()
+
+    this.transactionManager = new TransactionManager(this.driverAdapter)
   }
 
   applyPendingMigrations(): Promise<void> {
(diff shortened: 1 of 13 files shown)


packages/client/src/runtime/core/engines/client/TransactionManager.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+//  TODO: this test is failing and I have no idea why...
+test('transaction times out during starting', async () => {
+  const start = transactionManager.startTransaction({ timeout: 500, maxWait: START_TRANSACTION_TIME / 2 })
+
+  await jest.advanceTimersByTimeAsync(START_TRANSACTION_TIME + 100)
+
+  // TODO: This line asserts the error but somehow the error is still thrown beyond it and the test fails???
+  await expect(start).rejects.toBeInstanceOf(TransactionTimedOutError)
+})
```
FGoessler, line 89: ❓ I can't really figure out why this test keeps failing... maybe someone has a good idea?

packages/client/src/runtime/core/engines/client/TransactionManager.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  constructor(driverAdapter: DriverAdapter) {
+    this.driverAdapter = driverAdapter
+  }
+
+  async startTransaction(options: Tx.Options): Promise<Tx.InteractiveTransactionInfo<undefined>> {
+    // Supplying timeout default values is cared for upstream already.
+    if (!options.timeout) throw new TransactionManagerError('Timeout is required')
+    if (!options.maxWait) throw new TransactionManagerError('maxWait is required')
+    // TODO: care about transaction isolation levels?
```
FGoessler, line 45: ❓ I am not sure how we actually plan to handle transaction isolation levels with compiler approach. Somehow we'll have to pass this info to the query compiler? Or shall this transaction manager or the driver adapter inject the necessary SQL statements (and know about dialect/db differences?...)? 🤔

packages/client/src/runtime/core/engines/client/TransactionManager.ts
```diff
@@ -0,0 +1,169 @@
...
+      timeout: undefined,
+      transaction: undefined,
+    }
+    this.transactions.set(transaction.id, transaction)
+
+    // Start timeout to wait for transaction to be started.
+    transaction.timeout = this.startTransactionTimeout(transaction.id, options.maxWait)
+
+    const txContext = await this.driverAdapter.transactionContext()
+    if (!txContext.ok) throw new TransactionDriverAdapterError('Failed to start transaction.', txContext.error)
```
FGoessler, line 59: 🤔 I have to say that I am not sure what to think of that `Result` type that we use here in the DriverAdapter interfaces. Seems very Rust inspired and I get the idea but feels a bit weird when I use it in combination with promises. Also its error type is very restrictive so it cannot be really used as a generic Result type so far.

Am hence also wondering if that is really the best interface that we can expose to our community to implement their own driver adapters? 🤔

jacek-prisma: I agree, I think we should be trying to write idiomatic Typescript, and while there are some FP TS libraries with a `Result` type, they're not very commonly used

jacek-prisma: From my own bit of experience, I tried to write functional code in Typescript in the past, and it feels like trying to fit a square peg into a round hole, all standard library and utilities use mutable and non-persistent types, there are some alternatives but they don't get much traction from the community and they're not efficiently implemented in JS.

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -404,10 +390,47 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
...
+    } else {
+      assertNever(action, 'Invalid transaction action.')
+    }
+
+    return result
+  }
+
   async request<T>(
     query: JsonQuery,
     // TODO: support traceparent and interactiveTransaction!
```
jacek-prisma, line 432: seems we can remove the todo now?

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -446,12 +473,41 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
...
+        ? { ...this.config.transactionOptions, isolationLevel: transaction.options.isolationLevel }
+        : this.config.transactionOptions
+      txInfo = await this.transaction('start', {}, txOptions)
+    }
+
+    const results: BatchQueryEngineResult<T>[] = []
+    for (const { query, plan } of queriesWithPlans) {
+      const queryable = this.transactionManager.getTransaction(txInfo, query.action)
+      const interpreter = new QueryInterpreter(queryable, {})
+      results.push((await interpreter.run(plan)) as QueryEngineResultData<T>)
```
jacek-prisma, line 503: nitpick but we could maybe run these in parallel

jacek-prisma: or maybe some transaction logic depends on these being serial?

packages/client/src/runtime/core/engines/client/transactionManager/TransactionManager.test.ts
```diff
@@ -0,0 +1,240 @@
...
+test('for SQLite with unsupported isolation level', async () => {
+  const driverAdapter = new MockDriverAdapter({ provider: 'sqlite' })
+  const transactionManager = new TransactionManager({ driverAdapter, clientVersion: '1.0.0' })
+
+  await expect(
+    startTransaction(transactionManager, { isolationLevel: IsolationLevel.RepeatableRead }),
+  ).rejects.toBeInstanceOf(InvalidTransactionIsolationLevelError)
+})
+
+test('with the only by MS SQL Server supported isolation level "snapshot"', async () => {
```
jacek-prisma, line 190:
```suggestion
test('with the only isolation level supported in MS SQL Server, "snapshot"', async () => {
```
?

packages/client/src/runtime/core/engines/client/ClientEngine.ts
FGoessler: I would argue that at least when run as part of a transaction the ordering might be important for the end users so we should only run them serially. We could try to optimize for executing parallel queries when not in a transaction but am not sure how much of a benefit or breaking behavior this would be. 

=> I'll add a TODO: comment here for now to consider optimizing it and discuss this with @aqrln when he's back.

packages/client/src/runtime/core/engines/client/ClientEngine.ts
FGoessler: At least partially :)

jacek-prisma (approved):

PR #26115, opened 2025-01-17
chore(platform-cli): ppg copy changes

sampolahtinen:
Closes PLT-5559


packages/cli/src/platform/_lib/utils.ts
```diff
@@ -94,3 +94,43 @@ export const poll = async <F extends () => Promise<R>, R>(
...
+  workspaceId,
+  projectId,
+  environmentId,
+}: {
+  databaseUrl: string
+  workspaceId: string
+  projectId: string
+  environmentId: string
+}) => `
+Database URL:
```
sampolahtinen, line 109: Implementation for the PPG output.

packages/cli/src/Init.ts
```diff
@@ -362,10 +366,11 @@ export class Init implements Command {
         console.log(
           `Successfully authenticated as ${bold(
             authenticationResult.email,
-          )}. Let's set up your Prisma Postgres database!`,
+          )}. Let's set up your Prisma Postgres® database!`,
         )
       }
 
+      console.log("Let's set up your Prisma Postgres® database!")
```
ellotheth, line 373: Does this show the phrase twice if there were no saved credentials?

packages/cli/src/platform/_lib/utils.ts
```diff
@@ -94,3 +94,43 @@ export const poll = async <F extends () => Promise<R>, R>(
...
+${green(databaseUrl)}
+
+Project link:
+${link(`https://console.prisma.io/${workspaceId}/${projectId}/${environmentId}/dashboard`)}
+
+${bold('Next steps')}
+${bold('1. Define your database schema')}
+Open the ${green('schema.prisma')} file and define your first models. Go to ${link(
+  underline('pris.ly/ppg-quickstart'),
```
ellotheth, line 118: Do you want to set the protocol on this for consistency? (Or does `link()` do it automatically 😅)

packages/cli/src/Init.ts
sampolahtinen: Very good catch! It does. Will push a fix.

packages/cli/src/platform/_lib/utils.ts
sampolahtinen: Let's add it!

ellotheth (approved):

packages/cli/src/__tests__/commands/__snapshots__/Init.test.ts.snap
```diff
@@ -13,8 +13,7 @@ Next steps:
 5. Tip: Explore how you can extend the ORM with scalable connection pooling, global caching, and real-time database events. Read: https://pris.ly/cli/beyond-orm
 
 More information in our documentation:
-https://pris.ly/d/getting-started
-    "
```
sampolahtinen, line 17: I was battling against this change. In the end decided to update the snapshots. 

The default output looks good:

image

FGoessler (approved):

PR #26134, opened 2025-01-20
feat(cli): implement generic sub-commands (+ policy)

millsp:


packages/cli/src/__tests__/commands/SubCommand.test.ts
```diff
@@ -0,0 +1,107 @@
...
+        ],
+      ]
+    `)
+
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  expect(execa.command).toHaveBeenCalled()
+  expect(ni.getCommand).toHaveBeenCalled()
+
+  consoleLogSpy.mockRestore()
+})
```
millsp, line 107: Add a test to ensure cache evicts every 24h?

FGoessler: Am personally fine with either.

FGoessler (approved):
Nice to see this come into place that fast! 🎉

PR #26137, opened 2025-01-21
chore(compiler): ORM-517 integrate query compiler wasm into ClientEngine

FGoessler:
This is mostly a copy-and-adjust from the existing library-engine-wasm builds.

I am a bit concerned about the complexity of our build matrix and the potential configuration variants and how they are interact with each other. We e.g. have `TARGET_BUILD_TYPE`, `runtimeNameTs`, `runtimeNameJs`, `engineType`, and certain `wasm` flags that all need to be in correct combination and have implicit invariants.

### DIFF
--- a/packages/client/src/generation/TSClient/TSClient.ts
+++ b/packages/client/src/generation/TSClient/TSClient.ts
@@ -32,16 +32,7 @@ import { InputType } from './Input'
 import { Model } from './Model'
 import { PrismaClientClass } from './PrismaClient'
 
-type RuntimeName =
-  | 'binary'
-  | 'library'
-  | 'wasm'
-  | 'edge'
-  | 'edge-esm'
-  | 'index-browser'
-  | 'react-native'
-  | 'client'
-  | String
+type RuntimeName = 'binary' | 'library' | 'wasm' | 'edge' | 'edge-esm' | 'index-browser' | 'react-native' | 'client'
 
 export type TSClientOptions = O.Required<GenerateClientOptions, 'runtimeBase'> & {
   /** More granular way to define JS runtime name */
@@ -173,6 +164,7 @@ ${buildNFTAnnotations(edge || !copyEngine, clientEngineType, binaryTargets, rela
 `
     return code
   }
+
   public toTS(): string {
     const { reusedTs } = this.options
 
--- a/packages/client/src/generation/TSClient/TSClient.ts
+++ b/packages/client/src/generation/TSClient/TSClient.ts
@@ -32,13 +32,29 @@ import { InputType } from './Input'
 import { Model } from './Model'
 import { PrismaClientClass } from './PrismaClient'
 
-type RuntimeName = 'binary' | 'library' | 'wasm' | 'edge' | 'edge-esm' | 'index-browser' | 'react-native' | 'client'
-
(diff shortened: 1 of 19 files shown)


packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -33,9 +34,9 @@ import { PrismaClientClass } from './PrismaClient'
 
 export type TSClientOptions = O.Required<GenerateClientOptions, 'runtimeBase'> & {
   /** More granular way to define JS runtime name */
-  runtimeNameJs: 'binary' | 'library' | 'wasm' | 'edge' | 'edge-esm' | 'index-browser' | 'react-native' | String
+  runtimeNameJs: string // Usually one of: 'binary' | 'library' | 'wasm' | 'edge' | 'edge-esm' | 'index-browser' | 'react-native' | 'client'
   /** More granular way to define TS runtime name */
-  runtimeNameTs: 'binary' | 'library' | 'wasm' | 'edge' | 'edge-esm' | 'index-browser' | 'react-native' | String
+  runtimeNameTs: string
```
FGoessler, line 39: ℹ️ These types actually did not match what is used in code but leveraged the `String` overwrite to put all sorts of variants in for these config values. I think I do not really understand all edge cases and intentions behind this but having a strict union type does not work here.

packages/client/src/runtime/core/engines/client/ClientEngine.ts
FGoessler, line 1: ℹ️ Am dropping all the engine loading and starting logic here that was originally copied from the LibraryEngine. QueryCompiler is much easier to start up. ;)

packages/client/src/runtime/core/engines/client/WasmQueryCompilerLoader.ts
FGoessler, line 1: ℹ️ Modeled after the existing `WasmLibraryLoader.ts`.

packages/client/src/generation/generateClient.ts
```diff
@@ -735,7 +741,7 @@ function getNodeRuntimeName(engineType: ClientEngineType) {
       )
     }
 
-    return 'library' // TODO: eventually use dedicated client runtime bundle
+    return 'client' // TODO: eventually use dedicated client runtime bundle
```
jacek-prisma, line 744: can we remove the todo now?

FGoessler: Yes indeed - thought I dropped that one but seems to got lost in a branch switch...

packages/client/src/runtime/core/engines/client/types/QueryCompiler.ts
```diff
@@ -0,0 +1,22 @@
...
+  compile(request: string): Promise<string>
+}
+
+export type QueryCompilerOptions = {
+  datamodel: string
+  flavour: 'mysql' | 'postgres' | 'sqlite'
+  connectionInfo: {
+    schemaName?: string
+    maxBindValues?: number
+  }
```
jacek-prisma, line 13: would be nice to reuse an existing type, this is supposed to align with the driver adapter API

jacek-prisma: same for `flavour`, the strings are meant to match driver adapters

jacek-prisma (approved):

PR #26145, opened 2025-01-22
fix(build): make typescript optional peer dependency to not accidentally bundle it

FGoessler:
Issue with it was introduced by requiring it in https://github.com/prisma/prisma/pull/25968

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -271,7 +271,7 @@
   },
   "peerDependencies": {
     "prisma": "*",
-    "typescript": ">5.1.0"
+    "typescript": ">=5.1.0"
   },
   "peerDependenciesMeta": {
     "prisma": {
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -270,10 +270,14 @@
     "zx": "7.2.3"
   },
   "peerDependencies": {
-    "prisma": "*"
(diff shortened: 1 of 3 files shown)


packages/client/package.json
```diff
@@ -270,11 +270,15 @@
     "zx": "7.2.3"
   },
   "peerDependencies": {
-    "prisma": "*"
+    "prisma": "*",
+    "typescript": ">5.1.0"
```
FGoessler, line 274: Choose min version based on https://www.prisma.io/docs/orm/reference/system-requirements

millsp, line 274:
```suggestion
    "typescript": ">=5.1.0"
```

packages/internals/package.json
```diff
@@ -90,5 +90,13 @@
     "arg": "5.0.2",
     "prompts": "2.4.2"
   },
+  "peerDependencies": {
+    "typescript": ">5.1.0"
```
millsp, line 94: According to the docs

```suggestion
    "typescript": ">=5.1.0"
```

millsp (approved):
Thanks for the quick fix!

millsp:
Note: May be useful to manually test this in a real project to confirm the dep change didn't break anything.

PR #26159, opened 2025-01-23
refactor(client engine): client engine cleanups

FGoessler:
This PR performs some minor cleanups on our new `ClientEngine`.

- Remove unused stuff that was originally copied from the `LibraryEngine` but has no use-case in the `ClientEngine` anymore.
- Simplify and tidy up some error handling.
- Move types for the QueryGraph from the general purpose `Engine` interface file to the `ClientEngine`'s directory
- Move the `QueryInterpreter` to the `ClientEngine`'s directory


packages/client/src/generation/TSClient/TSClient.ts
```diff
@@ -32,11 +32,22 @@ import { InputType } from './Input'
...
+type RuntimeName =
+  | 'binary'
+  | 'library'
+  | 'wasm'
+  | 'edge'
+  | 'edge-esm'
+  | 'index-browser'
+  | 'react-native'
+  | 'client'
+  | (string & {}) // workaround to also allow other strings while keeping auto-complete intact
```
FGoessler, line 44: ℹ️  I dropped a similar workaround that was previously used in a PR yesterday as it was violating linters. I think this should work better now. 🤞

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -42,25 +37,18 @@ export class ClientEngine implements Engine<undefined> {
   QueryCompilerConstructor?: QueryCompilerConstructor
   queryCompilerLoader: QueryCompilerLoader
 
-  executingQueryPromise?: Promise<any>
```
FGoessler, line 45: ℹ️ In the `LibraryEngine` this was used to kinda await the last running query when stopping the engine. But as multiple queries can be in progress at a time I think this was never really accurate. As we no big stopping logic anymore it's imo not necessary to track our in progress queries like this.

If full query termination is really important to use we imo should rather look into using `AbortController` or so. 🤔

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -42,25 +37,18 @@ export class ClientEngine implements Engine<undefined> {
...
   driverAdapter: ErrorCapturingDriverAdapter
   transactionManager: TransactionManager
-  datamodel: string
 
   logEmitter: LogEmitter
   logQueries: boolean // TODO: actually implement
   logLevel: QueryEngineLogLevel
-  lastQuery?: string // TODO: actually implement
-  loggerRustPanic?: any // TODO: needed?
+  lastStartedQuery?: string
```
FGoessler, line 49: ℹ️ Imo a more correct term than `lastQuery`.

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -241,59 +226,54 @@ export class ClientEngine implements Engine<undefined> {
...
-    if (transaction?.kind === 'itx') {
-      // If we are already in an interactive transaction we do not nest transactions
-      txInfo = transaction.options
-    } else {
-      const txOptions = transaction?.options.isolationLevel
-        ? { ...this.config.transactionOptions, isolationLevel: transaction.options.isolationLevel }
-        : this.config.transactionOptions
-      txInfo = await this.transaction('start', {}, txOptions)
-    }
+    try {
```
FGoessler, line 239: ℹ️ Wrapping batch execution into a try-catch.

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -308,8 +288,8 @@ function getErrorMessageWithLink(engine: ClientEngine, title: string) {
     binaryTarget: undefined,
     title,
     version: engine.config.clientVersion!,
-    engineVersion: engine.versionInfo?.commit,
+    engineVersion: 'unknown', // WASM engines do not export their version info
```
FGoessler, line 291: ℹ️ Not sure if this something we want to eventually change but for WASM the underlying methods used by the `LibraryEngine` all just report `unknown`. 🤷

packages/client/src/runtime/core/engines/client/QueryPlan.ts
FGoessler, line 1: All copied 1:1 from the `Engine.ts` file.

jacek-prisma (approved):

PR #26161, opened 2025-01-23
fix(cli, sub-commands): better handling of esm packages

millsp:
/integration


packages/cli/src/SubCommand.ts
```diff
@@ -33,12 +33,13 @@ export class SubCommand implements Command {
 
     // if the package is not installed yet, we install it otherwise we skip
     if (existsSync(prefix) === false) {
-      const installCmd = getCommand('npm', 'install', [pkg, '--no-save', '--prefix', prefix])
+      const installCmd = getCommand('npm', 'install', [pkg, '--no-save', '--prefix', prefix, '--userconfig', prefix])
```
millsp, line 36: This prevents using the current projects .npmrc

packages/cli/src/SubCommand.ts
```diff
@@ -33,12 +33,13 @@ export class SubCommand implements Command {
...
       await command(installCmd, { stdout: 'ignore', stderr: 'inherit', env: process.env })
     }
 
     // load the module and run it via the Runnable interface
-    const module: Runnable = await import(`${prefix}/node_modules/${this.pkg}`)
+    const modulePath = [prefix, 'node_modules', this.pkg, 'dist', 'index.js']
+    const module: Runnable = await import(modulePath.join('/'))
```
millsp, line 42: ESM projects need a fully formed path

packages/cli/src/__tests__/fixtures/sub-command/node_modules/sub-command/package.json
```diff
@@ -1,5 +1,5 @@
 {
   "name": "sub-command",
   "version": "1.0.0",
-  "main": "index.mjs"
+  "main": "./dist/index.js"
```
millsp, line 4: This is a ne requirement for sub-clis, they must export an `index.js` in a `dist` folder. The `index.js` can be esm or cjs, that does not matter.

packages/cli/src/SubCommand.ts
millsp: Previously, I was using just the folder name for the import, but ESM packages require a fully formed path. To fix this, all imports will now point to `<installPath>/dist/index.js` and this is a requirement for any sub-cli, both in ESM or CJS.

FGoessler (approved):

PR #26167, opened 2025-01-24
feat(instrumentation): Use versioned global for tracing helper

lforst:
### Background and Problem

Hi :) We maintain the official Sentry SDK and would love to have first class support for Prisma instrumentation. We are currently depending on a specific v5 version of the `@prisma/instrumentation` package as a dependency of our Node.js SDK (`@sentry/node`).

With the release of Prisma v6 we examined applications crash that used Prisma v6 in combination with our SDK with our Prisma instrumentation enabled. The reason for that was that the `TracingHelper`'s interface was changed with Prisma v6, renaming a method (`createEngineSpan` -> `dispatchEngineSpans`) and the Sentry SDK registered the version 5 `TracingHelper`, which caused Prisma to call a method on it that doesn't exist.

We would love for our SDK to be compatible with multiple Prisma versions, and we assume other telemetry vendors would do so too! This PR doesn't necessarily fix the crashing, but it introduces a non-breaking mechanism to facilitate registering tracing helpers for specific versions of Prisma in a forwards compatible way.

### Changes

Currently, Prisma picks up whatever tracing helper is on `globalThis.PRISMA_INSTRUMENTATION` when an instrumentation library like `@prisma/instrumentation` put a helper there. There are no checks whether the helper is compatible with the current prisma version or not.

Because doing checks for compatibility is annoying, this PR updates Prisma to pick up the tracing helpers from a versioned field instead. The field is named `V${prismaMajorVersion}_PRISMA_INSTRUMENTATION`. This allows instrumentation libraries to register tracing helpers for specific versions of Prisma, without introducing any sort of incompatibilities.

To make this change actually non-breaking, we still need to read from `globalThis.PRISMA_INSTRUMENTATION`, for example when people would upgrade `@prisma/client` (where the reading happens) but not `@prisma/instrumentation` (where the writing happens). In the next major, I would recommend dropping the fallback to reduce the risk of Prisma crashing when libraries put incompatible helpers on the fallback global.


aqrln (approved):
This looks great to me, thank you!

aqrln:
Hey @lforst, do you want us to make a patch release with this change so you could make use of it right away?

PR #26181, opened 2025-01-27
feat: allow init --db if prisma folder exists

sampolahtinen:
Closes PLT-5645


FGoessler (approved):

packages/cli/src/Init.ts
```diff
@@ -428,6 +402,44 @@ export class Init implements Command {
       console.log(successMessage('Prisma Postgres® is ready to be used! 🚀'))
     }
 
+    if (
+      (isPpgCommand && (fs.existsSync(path.join(outputDir, 'schema.prisma')) || fs.existsSync(prismaFolder))) ||
+      fs.existsSync(path.join(prismaFolder, 'schema.prisma'))
```
ellotheth, line 407: Do we want to show the message if there's an empty `prisma` folder but no schema file?

sampolahtinen: We will not proceed with the schema.prisma or .env creation if:
- prisma folder exists
- prisma/schema.prisma exists
- ./schema.prisma exists

If any of these exists, we print success output when --db flag is present.

sampolahtinen: I have updated the copy to include mentioning of the folder.

packages/cli/src/Init.ts
```diff
@@ -398,7 +398,7 @@ export class Init implements Command {
         displayName: `database-setup-prismaPostgres-api-key`,
       })
 
-      prismaPostgresDatabaseUrl = `${PRISMA_POSTGRES_PROTOCOL}://accelerate.prisma-data.net/?api_key=${serviceToken.value}`
+      prismaPostgresDatabaseUrl = `${PRISMA_POSTGRES_PROTOCOL}//accelerate.prisma-data.net/?api_key=${serviceToken.value}`
```
sampolahtinen, line 401: Database url had double colons.

PR #26213, opened 2025-01-30
chore(query compiler): ORM-556 add CI step for query compiler tests

FGoessler:
This PR enables running our functional test against the new query compiler based client engine.

As tests are expected to fail for now, it applies a "known failures list" approach. The list can be found in `/packages/client/tests/functional/client-engine-known-failures-${suffix}.txt`. 
Where `suffix` can be any matrix based identifier to hold different failure lists for e.g. different driver adapters. 
For now the suite only runs against the `js_pg` driver adapter though.

You can update the list of known failures in two ways:
- A) Manually update the text file and remove the names of tests you fixed.
- B) Run all tests locally and generate an updated known failures file based on the jest output.

For B) you can use the additional package.json scripts:
- `pnpm test:functional:client` => Runs ALL tests in the same config as CI does.
- `pnpm test:knownQueryCompilerFailures:record js_pg` => Recreates the updated `client-engine-known-failures-js_pg.txt` file.

Use `pnpm test:knownQueryCompilerFailures:check js_pg` to check if your last local test run introduced any new test failures.


packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
```diff
@@ -78,6 +83,14 @@ export function renderQueryTemplate({
...
       templatePos++
     }
+    maxLoopCounter--
+
+    if (maxLoopCounter <= 0) {
+      debug('renderQueryTemplate ran into potential infinite loop during parameter substitution', query)
+      throw new Error(
+        'renderQueryTemplate ran into potential infinite loop during parameter substitution => execution aborted! Check query graph for correct `unique` statements.',
+      )
+    }
```
FGoessler, line 93: ℹ️ We have some tests that run into an endless loop during the replacement logic here. While a proper fix for that might require changes in multiple locations/different cases, with this change here we ensure our tests fail and not just hang indefinitely.

jacek-prisma: I think this might be adressed here: https://github.com/prisma/prisma/pull/26293

jacek-prisma (approved):

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
FGoessler: I hope so - am rebasing my changes on top of that now and let's see if the test hang or not ;)

PR #26224, opened 2025-01-31
feat: implement NPS survey

jacek-prisma:
Closes [ORM-534](https://linear.app/prisma-company/issue/ORM-534/implement-an-nps-rating-prompt-in-the-prisma-cli)


packages/cli/src/__tests__/nps.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    const status = jest.fn()
+    const readline = {
+      question: jest.fn(),
+      write: jest.fn(),
+    }
+    const capture = jest.fn()
+
+    process.env.CI = 'true'
+    await handleNpsSurveyImpl(currentDate, { status }, readline, { capture })
+    delete process.env.CI
```
aqrln, line 42: I don't think we necessarily want to delete it, it might have had been set before us assigning `true` when the tests themselves are running on CI.
```suggestion
    process.env.CI = originalEnv.CI
```

packages/cli/src/utils/nps/capture.ts
```diff
@@ -0,0 +1,43 @@
+import fetch from 'node-fetch'
+
+export interface EventCapture {
+  capture<Event>(id: string, name: string, payload: Event): Promise<void>
```
aqrln, line 4: In TypeScript, unlike other languages, generics are not so much an instrument of polymorphism but rather of propagating type constraints; polymorphism is primarily achieved through subtyping. A lone type variable that is not mentioned in the signature again is equivalent to just specifying its bound literally, or, in the case of an unconstrained type variable, the top type:
```suggestion
  capture(id: string, name: string, payload: unknown): Promise<void>
```

packages/cli/src/utils/nps/capture.ts
```diff
@@ -0,0 +1,43 @@
+import fetch from 'node-fetch'
```
aqrln, line 1: I don't think we need the `node-fetch` dependency anymore, all versions of Node.js we support in Prisma 6 ship `fetch` out of the box.

packages/cli/src/utils/nps/status.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const npsStatusUrl = new URL(
+  'https://raw.githubusercontent.com/jacek-prisma/nps-timeframe/refs/heads/main/timeframe.json',
+)
+
+export class ProdNpsStatusLookup implements NpsStatusLookup {
+  async status(): Promise<NpsStatus> {
+    const resp = await fetch(npsStatusUrl.href)
+    return await (resp.ok
+      ? ((await resp.json()) as NpsStatus)
```
aqrln, line 22: We should probably validate it using `zod` or `@effect/schema` or whatever and not just assert the type

packages/cli/src/utils/nps/survey.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    return promise.then((result) => {
+      clearTimeout(timeoutId)
+      resolve(result)
+    })
+  })
+}
+
+function isWithinTimeframe(date: Date, timeframe: Timeframe): boolean {
+  return new Date(timeframe.start) < date && new Date(timeframe.end) > date
```
aqrln, line 162: Should the range be inclusive?

packages/cli/src/utils/nps/status.ts
jacek-prisma: I don't think we use either right now, but since there's only 2 fields, maybe I can just check the schema manually. It might be worth doing the same for config

packages/cli/src/utils/nps/survey.ts
jacek-prisma: the date comparison compares up to milliseconds, so probably wont't matter, but I can make it inclusive

packages/cli/src/utils/nps/survey.ts
```diff
@@ -0,0 +1,163 @@
...
+  const question = rl.question(
+    'If you would like to rate how likely you are to recommend Prisma to a friend or colleague, ' +
+      'enter a number from 0 to 10, where 0 means "not likely at all" and 10 means ' +
+      '"extremely likely" and then press "Enter". This prompt will close automatically after ' +
+      `${promptTimeoutSecs} seconds.\n` +
+      'Rating: ',
+  )
+  const ratingAnswer = await timeout(question, promptTimeoutSecs * 1000)
+  if (ratingAnswer === undefined) {
+    rl.write(`No response received within ${promptTimeoutSecs} seconds. Exiting.\n`)
```
aqrln, line 99:
```suggestion
    rl.write(`No response received within ${promptTimeoutSecs} seconds. Exiting the survey.\n`)
```

packages/cli/src/utils/nps/status.ts
aqrln: > I don't think we use either right now, but since there's only 2 fields, maybe I can just check the schema manually

Since it's in the CLI and not the Client (where bundle size matters), I wouldn't worry too much about introducing new dependencies. Introducing it now lowers the barrier to using it where appropriate in the future, as well as finding and fixing other places where we might be missing validation.

> It might be worth doing the same for config

true

packages/cli/src/__tests__/nps.test.ts
jacek-prisma: done

packages/cli/src/utils/nps/capture.ts
jacek-prisma: done

packages/cli/src/utils/nps/capture.ts
jacek-prisma: done

packages/cli/src/utils/nps/status.ts
jacek-prisma: I added manual validation, if you wanna replace it with a library let me know which one, I've not used them extensively

packages/cli/src/utils/nps/survey.ts
jacek-prisma: done

packages/cli/src/__tests__/nps.test.ts
jacek-prisma: note that the originalEnv is captured from before the test suite runs, and these individial tests are expected to run with `process.env = {}`, which is set from `beforeAll`/`beforeEach`, so I don't think it matters what we do here

packages/cli/src/utils/nps/capture.ts
```diff
@@ -0,0 +1,41 @@
...
+
+type PosthogCapture<Props> = {
+  api_key: string
+  event: string
+  distinct_id: string
+  properties: Props
+}
+
+export class PosthogEventCapture implements EventCapture {
+  async capture(id: string, name: string, payload: Event) {
```
aqrln, line 22: Stray `Event` that probably refers to some global now? (though that shouldn't typecheck in a contravariant position either 🤔)

jacek-prisma: oops

packages/cli/src/utils/nps/status.ts
```diff
@@ -0,0 +1,42 @@
...
+    const obj = await resp.json()
+    if (!this.checkSchema(obj)) {
+      throw new Error('Invalid NPS status schema')
+    }
+    return obj
+  }
+
+  checkSchema(obj: any): obj is NpsStatus {
+    return (
+      obj.currentTimeframe === undefined ||
```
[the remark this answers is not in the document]
jacek-prisma: done

aqrln (approved):
Looks good to me (modulo the non-production link to JSON), nice work!

mmahalwy:
how was this okay to ship? it's literally blocking stuff on server deployments

mmahalwy:
@jacek-prisma @aqrln i've never seen a NPS survey on CLI... ever

aqrln:
@mmahalwy sorry it caused inconveniences for you. This shouldn't have been blocking anything in server deployments because it was supposed to be skipped in non-interactive environments. Sounds like our `process.stdin.isTTY` check might not have been enough. We'll investigate and ship a patch release quickly. Could you tell us more about the environment it causes issues for you? (e.g. is it a Docker build, a script running on CI, something else?)

PR #26239, opened 2025-02-03
fix(client): RPC stubs for top-level client methods

millsp:
Follow up on https://github.com/prisma/prisma/pull/25914. For full RPCs support, top-level methods need to be enumerable to work across RPCs.


packages/client/tests/functional/globals.d.ts
millsp, line 1: Unsolicited fix. Types were incorrect and out of date.

packages/client/src/runtime/core/metrics/MetricsClient.ts
```diff
@@ -22,7 +32,9 @@ export class MetricsClient {
    * @returns
    */
   prometheus(options?: MetricsOptions): Promise<string> {
-    return this._engine.metrics({ format: 'prometheus', ...options })
+    checkPreviewFeatureFlag(this._client)
```
millsp, line 35: Check now happens if you actually use it.

packages/client/src/runtime/core/model/applyModelsAndClientExtensions.ts
```diff
@@ -35,6 +36,21 @@ export function applyModelsAndClientExtensions(client: Client) {
...
+  return {
+    getKeys() {
+      return allKeys
+    },
+
+    getPropertyValue(prop) {
+      return client[prop]
+    },
+  }
+}
```
millsp, line 52: Makes base props and prototype props enumarable.

packages/client/src/runtime/getPrismaClient.ts
```diff
@@ -818,13 +815,15 @@ Or read our docs at https://www.prisma.io/docs/concepts/components/prisma-client
...
-        ]),
+      return createCompositeProxy(
+        applyModelsAndClientExtensions(
+          createCompositeProxy(unApplyModelsAndClientExtensions(this), [
+            addProperty('_appliedParent', () => this._appliedParent._createItxClient(transaction)),
+            addProperty('_createPrismaPromise', () => createPrismaPromiseFactory(transaction)),
+            addProperty(TX_ID, () => transaction.id),
+          ]),
+        ),
+        [removeProperties(itxClientDenyList)],
```
millsp, line 826: Same as before but we first build the tx client then remove disallowed methods.

FGoessler (approved):

PR #26247, opened 2025-02-05
chore(e2e tests): fix e2e tests with pnpm v10

FGoessler:
This PR fixes an issue with our E2E tests caused by `pnpm` v10 which requires to allowlist any post install scripts.

This PR also improves our test log output to avoid future confusions like the ones I had while investigating this issue. 

Also adding Typescript v5.7 explicit tests as v5.8. will soon be released and replace the `latest` tag.

### DIFF
--- a/packages/client/tests/e2e/_utils/run.ts
+++ b/packages/client/tests/e2e/_utils/run.ts
@@ -166,8 +166,10 @@ async function main() {
 
   if (args['--verbose'] === true) {
     for (const result of failedJobResults) {
-      console.log(`\n\n🛑🛑🛑 Test "${result.name}" failed with exit code ${result.exitCode} 🛑🛑🛑\n\n`)
+      console.log(`-----------------------------------------------------------------------`)
+      console.log(`🛑🛑🛑 Test "${result.name}" failed with exit code ${result.exitCode} 🛑🛑🛑`)
       console.log(`\t\t⬇️ Container Log output below ⬇️\n\n`)
+      console.log(`-----------------------------------------------------------------------`)
 
       const logsPath = path.resolve(__dirname, '..', result.name, 'LOGS.txt')
       const dockerLogsPath = path.resolve(__dirname, '..', result.name, 'LOGS.docker.txt')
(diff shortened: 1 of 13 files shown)


packages/client/tests/e2e/_utils/run.ts
```diff
@@ -198,7 +202,13 @@ async function restoreOriginalState() {
 }
 
 async function printFile(filePath: string) {
-  await pipeline(createReadStream(filePath), process.stdout)
+  try {
+    const fileStream = createReadStream(filePath)
+    fileStream.pipe(process.stdout, { end: false })
```
FGoessler, line 207: ℹ️ By default piping closes the target stream once its done. In this case closing stdout. Which lead to no further log output happening. That caused a very confusing picture in the CI logs that send me in the wrong direction for a while...

packages/client/tests/e2e/default-version/package.json
```diff
@@ -9,5 +9,10 @@
   "devDependencies": {
     "@types/jest": "29.5.12",
     "@types/node": "18.19.50"
+  },
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "@prisma/client"
+    ]
```
FGoessler, line 16: ℹ️ This is the actual fix for the failing test. `pnpm` v10 enforces that post install script need to be allowlisted. Our other tests passed because they don't rely on the script but perform a `prisma generate` explicitly. This test cannot do it as it doe snot install the prisma CLI itself. => We allowlist the script.

packages/client/tests/e2e/publish-extensions/simple-ext/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "tsc"
   },
   "peerDependencies": {
-    "@prisma/client": "/tmp/prisma-client-0.0.0.tgz"
+    "@prisma/client": "workspace:/tmp/prisma-client-0.0.0.tgz"
```
FGoessler, line 10: ℹ️ This is the other fix: `pnpm` v10 changed validation for peerDependencies and does not allow a plain path anymore.

packages/client/tests/e2e/_utils/run.ts
aqrln: oh wow

aqrln (approved):
Nice investigation and fix!

PR #26290, opened 2025-02-10
feat: serialize dot-separated column names into objects

jacek-prisma:


FGoessler (commented):
How is this / will this be tested? 🤔

jacek-prisma:
I can add a specific test case for this, but it'd be nice to introduce a more broad test suite for `QueryInterpreter`, I think it doesn't have any tests yet @FGoessler

jacek-prisma:
(this will be covered by the aggregate tests too)

aqrln:
Unit tests for the `serialize` function itself (at least this new functionality) would still be nice

jacek-prisma:
@aqrln done

aqrln (approved):
LGTM, thanks!

packages/client/src/runtime/core/engines/client/interpreter/serializer.ts
```diff
@@ -3,7 +3,20 @@ import { ResultSet } from '@prisma/driver-adapter-utils'
 export function serialize(resultSet: ResultSet): Record<string, unknown>[] {
   return resultSet.rows.map((row) =>
     row.reduce<Record<string, unknown>>((acc, value, index) => {
-      acc[resultSet.columnNames[index]] = value
+      const splitByDot = resultSet.columnNames[index].split('.')
+
+      let nested: any = acc
```
aqrln, line 8: fwiw, I try to avoid `any` like plague because it's more of an escape hatch to disable the type checker than a "real" type (it's a top and bottom type at the same time, which is a logical contradiction and source of unsoundness) but I couldn't come up with any well-behaved type that would work here so I guess it's fine. The function is small and easy to audit for correctness.

jacek-prisma: maybe `{}` type?

aqrln: actually yeah, that should work!

PR #26291, opened 2025-02-10
feat: ORM-607 disable automatic env loading during CLI cmds

FGoessler:
This PR disables the automatic env file loading during our CLI command execution if a `prisma.config.ts` file is detected.

It also refactors the config loading so it happens only at the top in the CLI. The parsed config then gets passed as argument to each subcommand. This follows our architecture strategy of inversion of control.


packages/config/src/loadConfigFromFile.ts
```diff
@@ -106,7 +106,10 @@ export async function loadConfigFromFile({
     }
 
     return {
-      config: required['default'] as PrismaConfig,
+      config: {
+        ...(required['default'] as PrismaConfig),
+        loadedFromFile: resolvedPath,
```
FGoessler, line 111: ℹ️ Based on this property we can detect whether a `prisma.config.ts` file was used/found.

packages/config/src/defaultConfig.ts
```diff
@@ -0,0 +1,14 @@
+import { PrismaConfig } from './PrismaConfig'
+
+/**
+ * All default values for the config shall be set here.
+ * Modules should not have to deal with missing config values and determining a default themselves as far as possible.
+ * => Consistent defaults and centralized top-level control of configuration via the CLI.
+ */
+export function defaultConfig(): PrismaConfig {
```
FGoessler, line 8: ℹ️ Important principle I'd love to apply. All agree?

packages/internals/src/utils/loadEnvFile.ts
```diff
@@ -1,14 +1,26 @@
...
-}: { schemaPath?: string; printMessage?: boolean } = {}) {
+}: {
+  schemaPath?: string
+  printMessage?: boolean
+  config: PrismaConfig
+}) {
+  if (config.loadedFromFile) {
+    console.debug(`Prisma config detected, skipping environment variable loading`)
+    return
+  }
```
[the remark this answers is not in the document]
jacek-prisma: It seems we normally don't use `console.debug` outside of tests/CI,
not sure if there's a reason for it but it's probably better to try to stay consistent throughout the codebase where we use `prisma/debug`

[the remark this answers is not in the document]
FGoessler: Good catch of a copy paste error - thanks!

jacek-prisma (approved):

packages/cli/src/__tests__/commands/SubCommand.test.ts
```diff
@@ -37,6 +38,10 @@ test('@<version>', async () => {
           [
             "--help",
           ],
+          {
+            "experimental": true,
+            "loadedFromFile": null,
```
jkomyno, line 43: `loadedFromFile` is no longer used.

packages/internals/src/utils/loadEnvFile.ts
```diff
@@ -1,14 +1,29 @@
...
+  config,
   printMessage = false,
-}: { schemaPath?: string; printMessage?: boolean } = {}) {
+}: {
+  schemaPath?: string
+  printMessage?: boolean
+  config: PrismaConfig
+}) {
+  if (config.loadedFromFile) {
+    debug(`Prisma config detected, skipping environment variable loading`)
```
jkomyno, line 23: We should probably write this to user in normal `stdout`, with the same format that we use to tell them that "Prisma has loaded the env vars for them" in the other conditional branch.

packages/cli/src/utils/loadConfig.ts
```diff
@@ -0,0 +1,26 @@
...
+      case 'ConfigFileParseError':
+        return new HelpError(`Failed to parse config file at "${resolvedPath}"`)
+      case 'TypeScriptImportFailed':
+        return new HelpError(`Failed to import config file as TypeScript from "${resolvedPath}". Error: ${error.error}`)
+      case 'UnknownError':
+        return new HelpError(`Unknown error during config file loading: ${error.error}`)
+      default:
+        assertNever(error, `Unhandled error '${JSON.stringify(error)}' in 'loadConfigFromFile'.`)
+    }
+  }
```
jkomyno, line 23: `resolvePath` may be `null` if no `prisma.config.ts` is given.
I'll need to adjust the type ´ConfigFromFile´ to avoid this ambiguity.

packages/cli/src/utils/loadConfig.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe('loadConfig', () => {
+  it('loads config from file', async () => {
+    ctx.fixture('prisma-config')
+
+    const config = await loadConfig('./prisma.config.ts')
+
+    expect(config).toMatchObject({
+      experimental: true,
+      loadedFromFile: path.join(ctx.fs.cwd(), './prisma.config.ts'),
```
jkomyno, line 18: When passing files to path where we specify a basedir explicitly, we should not use the `./` prefix in strings.

packages/config/src/defaultTestConfig.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * This default config can be used as basis for unit and integration tests.
+ */
+export function defaultTestConfig(): PrismaConfig {
+  return {
+    experimental: true,
+    loadedFromFile: null,
+  }
+}
```
jkomyno, line 11: What does this accomplish that `packages/config/src/defaultConfig.ts` doesn't already?

packages/config/src/defaultConfig.ts
```diff
@@ -0,0 +1,14 @@
+import { PrismaConfig } from './PrismaConfig'
```
jkomyno, line 1:
```suggestion
import type { PrismaConfig } from './PrismaConfig'
```

packages/config/src/defaultTestConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { PrismaConfig } from './PrismaConfig'
```
jkomyno, line 1:
```suggestion
import type { PrismaConfig } from './PrismaConfig'
```

packages/config/src/__tests__/loadConfigFromFile.test.ts
```diff
@@ -61,6 +61,7 @@ describe('loadConfigFromFile', () => {
       expect(resolvedPath).toMatch(path.join(ctx.fs.cwd(), 'prisma.config.ts'))
       expect(config).toMatchObject({
         experimental: true,
+        loadedFromFile: path.join(ctx.fs.cwd(), 'prisma.config.ts'),
```
jkomyno, line 64: Just use:

```suggestion
          loadedFromFile: resolvedPath,
```

packages/config/src/defaultTestConfig.ts
FGoessler: For the moment they are the same, indeed. I wanted to create a clear distinction from the start between a default config suitable for tests, and the default config for production use.
In my experience you'll likely have different config (e.g. paths, logging flags, ...) during tests eventually and then having to untangle which defaultConfig call is used in tests and which in actual code gets tedious.

packages/config/src/__tests__/loadConfigFromFile.test.ts
FGoessler: Addressed in https://github.com/prisma/prisma/pull/26315

packages/internals/src/utils/loadEnvFile.ts
FGoessler: Addressed in https://github.com/prisma/prisma/pull/26315

packages/config/src/defaultConfig.ts
FGoessler: Addressed in https://github.com/prisma/prisma/pull/26303/files already

packages/config/src/defaultTestConfig.ts
FGoessler: Addressed in https://github.com/prisma/prisma/pull/26303/files already

PR #26293, opened 2025-02-10
fix(client-engine): don't loop infinitely in query template

aqrln:
Fix the infinite loop with scalar lists, generally make the code more robust and add tests.

### DIFF
--- a/packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.test.ts
+++ b/packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.test.ts
@@ -1,0 +1,97 @@
+import { renderQueryTemplate } from './renderQueryTemplate'
+
+test('no template', () => {
+  expect(
+    renderQueryTemplate({
+      query: `SELECT * FROM users WHERE id = $1`,
+      params: [1],
+    }),
+  ).toEqual({
+    query: `SELECT * FROM users WHERE id = $1`,
+    params: [1],
+  })
+})
+
+test('no template and scalar list parameter', () => {
+  expect(
+    renderQueryTemplate({
+      query: `SELECT * FROM users WHERE id = $1 AND numbers = $2`,
+      params: [1, [1, 2, 3]],
+    }),
+  ).toEqual({
+    query: `SELECT * FROM users WHERE id = $1 AND numbers = $2`,
+    params: [1, [1, 2, 3]],
+  })
+})
+
+test('transforms IN template', () => {
+  expect(
+    renderQueryTemplate({
+      query: `SELECT * FROM users WHERE "userId" IN /* prisma-comma-repeatable-start */$1/* prisma-comma-repeatable-end */ OFFSET $2`,
+      params: [[1, 2, 3], 0],
+    }),
+  ).toEqual({
+    query: `SELECT * FROM users WHERE "userId" IN ($1, $2, $3) OFFSET $4`,
+    params: [1, 2, 3, 0],
+  })
+})
+
+test('transforms IN template with empty list', () => {
+  expect(
+    renderQueryTemplate({
+      query: `SELECT * FROM users WHERE "userId" IN /* prisma-comma-repeatable-start */$1/* prisma-comma-repeatable-end */ OFFSET $2`,
+      params: [[], 0],
+    }),
+  ).toEqual({
+    query: `SELECT * FROM users WHERE "userId" IN (NULL) OFFSET $1`,
+    params: [0],
+  })
+})
+
+test('handles singleton list in IN template', () => {
+  expect(
+    renderQueryTemplate({
+      query: `SELECT * FROM users WHERE "userId" IN /* prisma-comma-repeatable-start */$1/* prisma-comma-repeatable-end */ OFFSET $2`,
+      params: [[1], 0],
+    }),
+  ).toEqual({
+    query: `SELECT * FROM users WHERE "userId" IN ($1) OFFSET $2`,
+    params: [1, 0],
+  })
+})
+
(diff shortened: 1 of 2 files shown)


packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
```diff
@@ -38,50 +38,120 @@ export function renderQueryTemplate({
...
+  if (!query.includes(BEGIN_REPEAT)) {
+    return { query, params }
+  }
+
+  const flattenedParams: Value[] = []
   let lastParamId = 1
   let result = ''
   let templatePos = 0
 
+  type State = 'normal' | 'quoted' | 'repeating'
```
jacek-prisma, line 50: why not use an enum? I think by default they're number based so more optimal for comparisons and switch cases

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
```diff
@@ -38,50 +38,120 @@ export function renderQueryTemplate({
...
+          flattenedParams.push(paramValue)
+          result += `$${lastParamId++}`
+          break
+        }
+
+        case 'repeating': {
+          const paramArray = Array.isArray(paramValue) ? paramValue : [paramValue]
+
+          if (paramArray.length === 0) {
+            result += 'NULL'
```
jacek-prisma, line 125: wouldn't this 'accidentally' trigger for relational queries against an optional relation?

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
aqrln: I don't think it matters either way because strings are interned in JS so we should always be comparing integers anyhow (whether it's enum variants or pointers to `v8::String`) but I can change it to an enum if you'd prefer

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
aqrln: Very good question, I'm not sure actually. I'll check this.

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
jacek-prisma: ok, I guess it doesn't matter here then (I suspect enums are probably still more efficient if one is doing switches because their ordinals can be represented as a jump table)

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  ).toEqual({
+    query: `SELECT * FROM users WHERE "userId" IN ($1) OFFSET $2`,
+    params: [1, 0],
+  })
+})
+
+test("transforms IN template, doesn't touch scalar list", () => {
+  expect(
+    renderQueryTemplate({
+      query: `SELECT * FROM users WHERE "userId" IN /* prisma-comma-repeatable-start */$1/* prisma-comma-repeatable-end */ AND numbers = $2 OFFSET $3`,
```
jacek-prisma, line 78: probably not needed right now but just leaving it here: we'll eventually want to have the ability to take two sets of rows and do something like a product between them, this will happen when we have an m2m connect where we generate a multi value insert where each inserted value refers to columns read from two different queries

FGoessler (approved):

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
aqrln: `(NULL) IN (NULL)`, just like `NULL = NULL`, is false in SQL, only `NULL IS NULL` is true:

```
postgres@localhost:postgres> select * from tmp;
+--------+
| a      |
|--------|
| 1      |
| <null> |
+--------+
SELECT 2
Time: 0.009s
postgres@localhost:postgres> select * from tmp where a in (null);
+---+
| a |
|---|
+---+
SELECT 0
```

packages/client/src/runtime/core/engines/client/interpreter/renderQueryTemplate.ts
aqrln: Yeah, in this case V8 (Ignition) generates identical bytecode in both cases, the only difference being `LdaConstant` emitted in the case of strings or `LdaZero`/`LdaSmi` emitted for enums. The generated machine code for ARM64, however, ends up being 5 CPU instructions shorter with enums because small constants can be used as immediates:

```
0xffff90007424   3a4  7100067f       cmp w19, #0x1 (1)
```

but strings need to be loaded from the constant table into a register first:

```
0xffff8dfc746c   3ac  580080f6       ldr x22, pc+4124 (addr 0x0000ffff8dfc8488)    ;; object: 0x1fde6cc134f1 <String[6]: #quoted>
```

aqrln: I guess there technically is tiny benefit in using enums here then 😄

aqrln: Important note thought that this only definitely applies to const enums (as I just wrote 0, 1, 2 in a JS file I copied the code to literally), regular non-const TypeScript enums probably generate worse code than either of these due to an extra layer of indirection through the enum object.

jacek-prisma (approved):

PR #26303, opened 2025-02-11
feat(config): parse Prisma Config's structure for conformance

jkomyno:
This PR:
- Closes [ORM-618](https://linear.app/prisma-company/issue/ORM-618/ensure-that-prismaconfigts-conforms-to-the-expected-schema)
- Replaces `experimental: true` with `earlyAccess: true`, according to [this Slack thread](https://prisma-company.slack.com/archives/C058VM009HT/p1739288326017889) created after this PR was initially conceived.
- Yields a `ConfigFileParseError` error if the default export in `prisma.config.ts` doesn't conform to the expected schema
- Schema validation is performed via [Effect Schema](https://effect.website/docs/schema/introduction/)


packages/config/package.json
```diff
@@ -16,6 +16,7 @@
     "@prisma/get-platform": "workspace:*",
     "@swc/core": "1.10.11",
     "@swc/jest": "0.2.37",
+    "effect": "3.12.10",
```
FGoessler, line 19: This seems like a pretty big additional dependency with a lot of implications on how we write our TS code moving forward. Would love to have a more in depth discussion in the team before we introduce such a large additional dependency. Wdyt?

packages/config/src/loadConfigFromFile.ts
```diff
@@ -76,43 +85,49 @@ export async function loadConfigFromFile({
...
-  debug(`Config file loaded in %s`, getTime())
+    debug(`Config file loaded in %s`, getTime())
 
-  try {
-    // TODO: We should parse `configExport` to ensure it conforms to `PrismaConfig`'s shape at runtime
-    if (required['default'] === undefined) {
+    // Ensure the config file conforms to the expected PrismaConfig schema.
+    const parseResultEither = parsePrismaConfig(required['default'])
+
+    if (Either.isLeft(parseResultEither)) {
```
FGoessler, line 110: I think I don't get the concept of how effect treats left vs right here. 🤔 To me their definition is inverted and this should be a `Either.right(...)`. What am I missing?

FGoessler (approved):
As discussed with Alberto: We are OK with using `effect` here for the schema/shape definitions but do NOT want to leak its usages and patterns further out to the codebase.

Note: We already use the schema parts of `effect` in other parts of our codebases.

packages/config/package.json
jkomyno: (Solved in a call)

PR #26317, opened 2025-02-12
feat(config): locate prisma schemas via `prisma.config.ts`

jkomyno:
This PR:
- contributes to [ORM-622](https://linear.app/prisma-company/issue/ORM-622/locate-prisma-schemas-via-prismaconfigts-if-possible)
- defines a new top-level `schema` attribute in `prisma.config.ts`'s `defineConfig`'s input object, allowing to specify the location of a `single` or `multi` schema files
 - moves some `prismaSchemaFolder` functionality from `@prisma/internals` to `@prisma/schema-files-loader`
   - this was originally done because we initially loaded PSL files directly in `@prisma/config`, and later decided to deviate in agreement with @FGoessler


packages/internals/src/cli/getSchema.ts
```diff
@@ -256,11 +192,15 @@ async function getSchemaWithPathInternal(
     return customSchemaResult
   }
 
+  // 2. Try the `schema` from `PrismaConfig`
```
FGoessler, line 195: Guess that part is left for me to add? ;)

FGoessler (commented):
Not sure why this has some test failures but other than that looks good to me. :+1:

packages/internals/src/cli/getSchema.ts
jkomyno: Yess

FGoessler (approved):

PR #26321, opened 2025-02-12
feat: migration adapter/query adapter interfaces

jacek-prisma:
Closes [ORM-637](https://linear.app/prisma-company/issue/ORM-637/introduce-a-minimal-interface-for-migration-adapters).

Partially moves us to the new interfaces from the migration adapter proposal.
Also adds crude implementations of the new interface functions for D1, LibSQL and pg (we might need to iterate on them). It's been left unimplemented for others.


packages/adapter-d1/package.json
```diff
@@ -38,9 +38,8 @@
   "license": "Apache-2.0",
   "sideEffects": false,
   "dependencies": {
+    "@cloudflare/workers-types": "4.20240614.0",
     "@prisma/driver-adapter-utils": "workspace:*",
-    "@cloudflare/workers-types": "4.20240614.0"
-  },
-  "devDependencies": {},
-  "peerDependencies": {}
+    "miniflare": "3.20250204.0"
```
aqrln, line 43: Should it be a peer dependency instead? Or could we have a more open version range here?

jacek-prisma: I think it should probably be a dependency because it's a hard requirement for migrations, but I can check what version range we can support

jacek-prisma: specified the miniflare version with a caret, with the base version based on the worker types

packages/driver-adapter-utils/src/types.ts
```diff
@@ -129,70 +129,120 @@ const officialPrismaAdapters = [
...
+
+  /**
+   * Dispose of the connection and release any resources.
+   */
+  dispose(): Promise<void>
+}
+
+export interface TransactionContext extends AdapterInfo, SqlQueryable {
+  /**
+   * Starts new transaction.
```
jkomyno, line 193:
```suggestion
   * Start new transaction.
```

To make it consistent with other changes elsewhere.

packages/adapter-d1/src/d1.ts
```diff
@@ -79,54 +75,54 @@ class D1Queryable<ClientT extends StdClient> implements Queryable {
...
+  protected onError(error: any): never {
+    console.error('Error in performIO: %O', error)
+    const { message } = error
+
+    throw new PrismaAdapterError({
+      kind: 'sqlite',
+      extendedCode: matchSQLiteErrorCode(message),
+      message,
+    })
+  }
```
jkomyno, line 112: - `onError` is not using any field/method via `this.`, so it should be a free function rather than a method, no?
- destructuring fields from an `any` type is not type-safe. Errors caught by `catch(e) {}` are `e: unknown` and should be explicitly cast to type `Error`

packages/driver-adapter-utils/src/binder.ts
```diff
@@ -98,12 +109,16 @@ function wrapAsync<A extends unknown[], R>(
...
 ): (...args: A) => Result<R> {
   return (...args) => {
     try {
-      return fn(...args)
+      return ok(fn(...args))
     } catch (error) {
+      // unwrap the cause of exceptions thrown by driver adapters if there is one
+      if (error.cause && typeof error.cause === 'object') {
+        return err(error.cause)
+      }
```
jkomyno, line 121:
```suggestion
      if (error instanceof Error) {
        return err(error.cause)
      }
```

packages/driver-adapter-utils/src/error.ts
```diff
@@ -0,0 +1,11 @@
+import { Error as ErrorObject } from './types'
+
+export class PrismaAdapterError extends Error {
+  name = 'PrismaAdapterError'
```
jkomyno, line 4: They're called `Driver Adapters`, not `Prisma Adapters`. Let's use consistent naming conventions.

```suggestion
export class DriverAdapterError extends Error {
  name = 'DriverAdapterError'
```

packages/adapter-d1/src/d1.ts
jacek-prisma: > destructuring fields from an any type is not type-safe. Errors caught by catch(e) {} are e: unknown and should be explicitly cast to type Error
yeah I don't like it, but I just extracted the existing code here

packages/driver-adapter-utils/src/binder.ts
```diff
@@ -84,12 +91,16 @@ const bindTransaction = (errorRegistry: ErrorRegistryInternal, transaction: Tran
...
 ): (...args: A) => Promise<Result<R>> {
   return async (...args) => {
     try {
-      return await fn(...args)
+      return ok(await fn(...args))
     } catch (error) {
+      // unwrap the cause of exceptions thrown by driver adapters if there is one
+      if (error.cause && typeof error.cause === 'object') {
+        return err(error.cause)
+      }
```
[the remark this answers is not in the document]
jacek-prisma: do you mean `error.cause instanceof Error`? this is responsible for unwrapping the cause for the query engine to handle errors correctly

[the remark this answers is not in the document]
jacek-prisma: I think the causes aren't even necessarily instances of js Error, it's just the error payload we pass back to Rust

packages/adapter-d1/src/d1.ts
jacek-prisma: updated

packages/driver-adapter-utils/src/binder.ts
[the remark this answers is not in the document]
jacek-prisma: updated

packages/driver-adapter-utils/src/binder.ts
jacek-prisma: updated

packages/driver-adapter-utils/src/error.ts
jacek-prisma: done

jkomyno (approved):

PR #26329, opened 2025-02-13
feat(studio): ORM-578 enable use of driver adapters in prisma studio

FGoessler:
Enable `prisma studio` to work with driver adapters based on `prisma.config.ts`.

Relies on work in Prisma Studio itself to be merged first:
- https://github.com/prisma/studio-code/pull/596
- https://github.com/prisma/studio-code/pull/597

Example `prisma.config.ts` file:

```typescript
import path from 'node:path'
import { defineConfig } from '@prisma/config'
import { PrismaLibSQL } from '@prisma/adapter-libsql'
import { createClient } from '@libsql/client'

export default defineConfig({
  earlyAccess: true,
  schema: {
    kind: 'single',
    filePath: './prisma/schema.prisma',
  },
  studio: {
    adapter: async (env: unknown) => {
      const connectionString = `file:./dev.db'
      const libsql = createClient({
        url: connectionString,
      })
      return new PrismaLibSQL(libsql)
    },
  },
})
```


packages/cli/package.json
```diff
@@ -80,8 +81,8 @@
     "@prisma/internals": "workspace:*",
     "@prisma/migrate": "workspace:*",
     "@prisma/mini-proxy": "0.9.5",
-    "@prisma/studio": "0.509.0",
-    "@prisma/studio-server": "0.509.0",
+    "@prisma/studio": "0.509.1-integration-orm-578-driver-adapters.0",
+    "@prisma/studio-server": "0.509.1-integration-orm-578-driver-adapters.0",
```
FGoessler, line 85: 🚧 Update once new prisma studio version is released!

packages/cli/src/__tests__/commands/Studio.test.ts
```diff
@@ -676,3 +676,68 @@ describeIf(process.env.PRISMA_CLIENT_ENGINE_TYPE !== 'binary')('studio with sche
...
+    jest.setTimeout(20_000)
+
+    beforeAll(async () => {
+      // Before every test, we'd like to reset the DB.
+      // We do this by duplicating the original SQLite DB file, and using the duplicate as the datasource in our schema
+      rimraf.sync(path.join(__dirname, '../fixtures/studio-test-project-driver-adapter/dev_tmp.db'))
+      fs.copyFileSync(
+        path.join(__dirname, '../fixtures/studio-test-project-driver-adapter/dev.db'),
+        path.join(__dirname, '../fixtures/studio-test-project-driver-adapter/dev_tmp.db'),
+      )
```
jacek-prisma, line 692: I think we can just truncate the file, SQLite should be able to deal with an empty db file

packages/config/src/PrismaConfig.ts
```diff
@@ -67,7 +91,7 @@ const createPrismaConfigShape = () =>
  * The configuration for the Prisma Development Kit, before it is passed to the `defineConfig` function.
  * Thanks to the branding, this type is opaque and cannot be constructed directly.
  */
-export type PrismaConfig = {
+export type PrismaConfig<Env = any> = {
```
jacek-prisma, line 94: is there a good reasons for the `any` defaults? rather than something less of a potential footgun like `object` or `unknown`

packages/adapter-pg/src/pg.ts
```diff
@@ -184,7 +184,8 @@ export type PrismaPgOptions = {
 
 export class PrismaPg extends PgQueryable<StdClient> implements SqlConnection {
   constructor(client: StdClient, private options?: PrismaPgOptions, private readonly release?: () => Promise<void>) {
-    if (!(client instanceof pg.Pool)) {
+    // Note: Full `client instanceof pg.Pool` check would sometimes give false negatives depending on the package resolution.
+    if (!client) {
       throw new TypeError(`PrismaPg must be initialized with an instance of Pool:
 import { Pool } from 'pg'
 const pool = new Pool({ connectionString: url })
```
jkomyno, line 191: @aqrln I remember we had introduced these explicit errors for a reason. Is this reason still valid nowadays?

packages/cli/src/__tests__/fixtures/studio-test-project-driver-adapter/prisma.config.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createClient } from '@libsql/client'
+
+export default defineConfig({
+  earlyAccess: true,
+  schema: {
+    kind: 'single',
+    filePath: path.join(__dirname, 'schema-c.prisma'),
+  },
+  studio: {
+    adapter: async (env: unknown) => {
```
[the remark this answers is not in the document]
jkomyno: We could just have `adapter: async (env: Env)`, with

```typescript
import 'dotenv/config'

type Env = {
  LIBSQL_DATABASE_URL: 'string'
}

export default defineConfig<Env>({
  studio: {
    adapter: async (env) => {
      const { PrismaLibSQL } = await import('@prisma/adapter-libsql')
      const { createClient } = await import('@libsql/client')

      const libsql = createClient({
        url: env.LIBSQL_DATABASE_URL,
      })
      return new PrismaLibSQL(libsql)
  },
  // ...
})
```

and create a new `.env` file as a sibling of `prisma.config.ts`:

```
# .env
LIBSQL_DATABASE_URL="file:./dev_tmp.db"
```

This would then become a test representative of what the users would actually do in the wild.

packages/config/src/PrismaConfig.ts
```diff
@@ -104,10 +132,11 @@ const createPrismaConfigInternalShape = () =>
   Shape.Struct({
     earlyAccess: Shape.Literal(true),
     schema: Shape.optional(PrismaSchemaConfigShape),
+    studio: Shape.optional(PrismaStudioConfigShape),
     loadedFromFile: Shape.NullOr(Shape.String),
   })
 
-type _PrismaConfigInternal = {
+type _PrismaConfigInternal<Env = any> = {
```
jkomyno, line 139: Same consideration about `any` as above: https://github.com/prisma/prisma/pull/26329/files#r1973659418

packages/cli/src/__tests__/commands/Studio.test.ts
FGoessler: You mean instead of creating a fresh copy of the test db to delete the file or wipe it empty?

The point is that it's not am empty database here but contains preset test data that is queried in the test. This is the same test setup principle as used for the other studio tests here.

Or do you mean something different with "truncate" @jacek-prisma  ? 🤔

packages/adapter-pg/src/pg.ts
FGoessler: From what I saw in the tests it seems very easy to by accident pass a Client instead of Pool instance. Imo checking this via typechecks in the users code is sufficient though.

packages/cli/src/__tests__/fixtures/studio-test-project-driver-adapter/prisma.config.ts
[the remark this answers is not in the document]
FGoessler: Good point. Updated the test except for not using dotenv but setting `process.env` explicitly. `dotenv`'s logic for finding the right `.env` file felt too fiddly for me in this test setup. ;)

packages/cli/src/Studio.ts
```diff
@@ -77,9 +77,9 @@ ${bold('Examples')}
    * Parses arguments passed to this command, and starts Studio
    *
    * @param argv Array of all arguments
-   * @param _config The loaded Prisma config
+   * @param config The loaded Prisma config
    */
-  public async parse(argv: string[], config: PrismaConfigInternal): Promise<string | Error> {
+  public async parse(argv: string[], config: PrismaConfigInternal<any>): Promise<string | Error> {
```
FGoessler, line 82: ℹ️ This is now the only remaining use of `any` for the config stuff. I think we also cannot really avoid it here, or? 🤔

packages/cli/src/__tests__/commands/Studio.test.ts
FGoessler: I'll leave it like this - also because I'd need to change all the other tests that use the same pattern then, too. Not worth the effort imo.

jacek-prisma: okay

jkomyno (approved):

packages/cli/src/Studio.ts
FGoessler: Exactly. I think I also tried `unknown` here but it didn't let me :/

PR #26330, opened 2025-02-13
refactor: extract reusable parts of the client engine into a new package

aqrln:
`@prisma/client-engine-runtime` being built by `@prisma/client`'s build script meant that we still needed to include the client package and all of its transitive dependencies into the workspace in `libs/driver-adapters` in the `prisma-engines` repo, which was causing some issues.

This commit properly extracts the reusable parts of the client engine into a package for real. This is also important to make Accelerate work with the query compiler.

Additionally, `@prisma/debug` is changed to be ESM compatible. This is because `@prisma/client-engine-runtime` needs to be ESM compatible and depends on `@prisma/debug`.


packages/client/src/runtime/core/engines/common/types/Transaction.ts
```diff
@@ -1,16 +1,4 @@
...
-  Snapshot = 'Snapshot',
-  Serializable = 'Serializable',
-}
-
-export type Options = {
-  maxWait?: number
-  timeout?: number
-  isolationLevel?: IsolationLevel
-}
+export { IsolationLevel, type TransactionOptions as Options } from '@prisma/client-engine-runtime'
```
aqrln, line 1: I didn't want to extract a new package just for these, and I can't put them in `@prisma/internals` because I want to avoid using it in `@prisma/client-engine-runtime`.

packages/client-engine-runtime/src/transactionManager/TransactionManagerErrors.ts
```diff
@@ -0,0 +1,66 @@
+import { Error as DriverAdapterError } from '@prisma/driver-adapter-utils'
+
+export class TransactionManagerError extends Error {
+  code = 'P2028'
```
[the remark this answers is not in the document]
FGoessler: I think for now it's fine to keep it in here.

packages/client/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["./**/*"]
+  "include": ["./**/*", "../client-engine-runtime/src/interpreter/serialize.test.ts"]
```
FGoessler, line 3: Why is this needed? 🤔

FGoessler (approved):

packages/client/tsconfig.json
aqrln: uhh, looks like the language server did it when I moved the file, thanks for catching 🤦

PR #26333, opened 2025-02-14
feat(prisma-config): ORM-628 use schema path from config file

FGoessler:
This PR adjusts all our CLI commands to also respect the configured schema path location based on `prisma.config.ts`.

I intentionally left out larger refactorings like hoisting the actual schema loading further up. This also meant that for now a bunch of slightly inconsistent logic or implicit dependencies about the schema loading are still in the codebase. E.g. the "sneaky" implicit setting of the `--schema` cmd line arg in [checkUnsupportedDataProxyMessage](https://github.com/prisma/prisma/blob/23d8251da96e068ad45d09b2aef3b625f0fc4269/packages/internals/src/cli/checkUnsupportedDataProxy.ts#L54-L57).


packages/internals/src/cli/getSchema.ts
```diff
@@ -57,6 +57,8 @@ type PackageJsonLookupResult =
       }
     }
 
+export type SchemaPathFromConfig = { kind: 'single'; filenamePath: string } | { kind: 'multi'; folder: string }
```
jkomyno, line 60: Did you copy the type from `@prisma/config` because you didn't want to add a dependency to its package?

FGoessler: Not really the direct package dependency itself (internals already depends on config in the package.json), but to not let an "inner" module depend on a "sibling"/"higher" module moving forward. So the schema loading logic has its own interface that it exposes based on what it needs. Which "just happens" to align very well with the config format. ;)

packages/internals/src/cli/getSchema.ts
```diff
@@ -193,6 +199,10 @@ async function getSchemaWithPathInternal(
   }
 
   // 2. Try the `schema` from `PrismaConfig`
+  const prismaConfigResult = await readSchemaFromPrismaConfigBasedLocation(cwd, schemaPathFromConfig)
```
jkomyno, line 202: `cwd` shouldn't be taken into account. The path provided by `defineConfig` is absolute.

packages/config/src/__tests__/loadConfigFromFile.test.ts
```diff
@@ -42,6 +42,24 @@ describe('loadConfigFromFile', () => {
...
+        expect(error).toBeUndefined()
+        expect(config).toMatchObject({
+          earlyAccess: true,
+          loadedFromFile: resolvedPath,
+          schema: {
+            kind: 'single',
+            filenamePath: path.join(cwd, 'prisma', 'schema.prisma'),
+          },
+        })
+        expect(error).toBeUndefined()
```
jkomyno, line 60:
```suggestion
```

packages/config/src/__tests__/loadConfigFromFile.test.ts
```diff
@@ -80,6 +98,24 @@ describe('loadConfigFromFile', () => {
...
+        expect(config).toMatchObject({
+          earlyAccess: true,
+          loadedFromFile: resolvedPath,
+          schema: {
+            kind: 'multi',
+            folder: path.join(cwd, 'prisma', 'schema'),
+          },
+        })
+        expect(error).toBeUndefined()
```
jkomyno, line 116:
```suggestion
```

packages/internals/src/utils/getEnvPaths.ts
```diff
@@ -25,6 +25,8 @@ export async function getEnvPaths(
 ): Promise<EnvPaths> {
   const rootEnvPath = getProjectRootEnvPath({ cwd: opts.cwd }) ?? null
   const schemaEnvPathFromArgs = schemaPathToEnvPath(schemaPath)
+  // NOTE: We intentionally do NOT check based on the schema path from `prisma.config.ts` as having a
+  // `prisma.config.ts` file disables automatic env loading anyway.
```
jkomyno, line 29: Nice one!

jkomyno (approved):

PR #26336, opened 2025-02-14
fix(config): fix issue with --config cli arg

FGoessler:
The arg had to be "allowlisted" for each command individually.


jkomyno (changes requested):
We need tests using `--config` for each for each command.
A single fixture containing a valid `prisma.config.ts` and used by each command will do.

Actually, we need at least 2 fixtures: one in `cli`, one in `migrate`.

jkomyno (approved):

PR #26361, opened 2025-02-18
fix(cli): add esbuild and esbuild-register to runtime dependencies

aqrln:
/integration

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -133,7 +133,9 @@
     "prepublishOnly": "pnpm run build"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


jkomyno (approved):
Note for the future: we'll need to get rid of this duplication between `cli` and `@prisma/config`

aqrln:
> Note for the future: we'll need to get rid of this duplication between `cli` and `@prisma/config`

Yep. This is just to unblock the release. IMO we shouldn't be bundling `@prisma/config` in CLI and should depend on it as a normal dependency.

PR #26374, opened 2025-02-20
fix(config): e2e tests + unbundling

jkomyno:
This PR:
- closes [ORM-658](https://linear.app/prisma-company/issue/ORM-658/broken-e2e-test-on-main)
- introduces `sandbox/basic–sqlite-config`
- stops bundling `@prisma/config`
- only exposes the `Effect/Brand` type from the `effect` package

/integration

### DIFF
--- a/packages/cli/helpers/build.ts
+++ b/packages/cli/helpers/build.ts
@@ -60,7 +60,7 @@ const cliBuildConfig: BuildOptions = {
   outfile: 'build/index',
   plugins: [cliLifecyclePlugin],
   bundle: true,
-  external: ['fsevents', 'esbuild', 'esbuild-register'],
+  external: ['fsevents', '@prisma/config'],
   emitTypes: false,
   minify: true,
 }
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -72,7 +72,6 @@
     "@antfu/ni": "0.21.12",
     "@inquirer/prompts": "5.0.5",
     "@prisma/client": "workspace:*",
-    "@prisma/config": "workspace:*",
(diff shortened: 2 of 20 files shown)


aqrln (approved):

aqrln (changes requested):
Actually this will break the users since you're not bundling `@prisma/config` anymore and don't have `effect` in dependencies, so it will fail at run time trying to import `effect`.

aqrln:
Can we have an integration release once we sort out the npm stuff? I still don't see how this could work without `effect` in the dependencies of the CLI. It might also be a good opportunity to stop bundling `@prisma/config` in the CLI to get rid of the duplication.

jkomyno:
> Can we have an integration release once we sort out the npm stuff? I still don't see how this could work without `effect` in the dependencies of the CLI. It might also be a good opportunity to stop bundling `@prisma/config` in the CLI to get rid of the duplication.

I'm in a meeting rn, but here you go: https://github.com/prisma/prisma/actions/runs/13439159464

aqrln:
@jkomyno 

```
$ npx prisma validate
Failed to import config file as TypeScript from "/Users/aqrln/Downloads/config-tests/prisma.config.ts". Error: Error: Cannot find module '@prisma/driver-adapter-utils'
Require stack:
- /Users/aqrln/Downloads/config-tests/node_modules/@prisma/config/dist/defineConfig.js
- /Users/aqrln/Downloads/config-tests/node_modules/@prisma/config/dist/index.js
- /Users/aqrln/Downloads/config-tests/prisma.config.ts
- /Users/aqrln/Downloads/config-tests/node_modules/prisma/build/index.js
```

PR #26378, opened 2025-02-20
feat(migrate): Not offering a database reset

### LINKED ISSUE #21483: Configuration to disallow database reset
## Problem

I've started using prisma recently and it seems insanely scary that not only its super easy to reset the entire db but also the fact that doing a migration if it fails due to some sync issues its the default behaviour of prisma to suggest a full reset and data wipe?

## Suggested solution

Some configuration on the connection or something should exist to prevent this behaviour.
If in the datasource there's a preserveDB or preventDBReset or something like it, prisma should never suggest to reset db when running any command besides prisma reset db and even then it should ask more then once if we are sure of the concequences.

### LINKED ISSUE #26076: Schema Drift Migration errors produce a dangerous error message
### Bug description

I'm on the Supabase Team. When users edit their database through the Supabase Dashboard, they often encounter schema drift errors:

drift

Instead of only offering to reset the schema, could Prisma propose two solutions: baseline (non-destructive) or reset (destructive)?

Also, while the CLI tool does show which objects were changed, the text isn’t emphasized. Since this context is critical, it would be helpful if the output were highlighted:
drift 2

### How to reproduce

Run the below command after updating a schema outside of Prisma:
```sh
npx prisma migrate dev --name first_prisma_migration
```

### Expected behavior

Offer to baseline or reset instead of just reset

### Prisma information

I just follow the basic [Supabase Guide](https://supabase.com/docs/guides/database/prisma?queryGroups=initiate&initiate=bun_initiate&queryGroups=sync&sync=bun_sync&queryGroups=migrate&migrate=npm_migrate) for creating Prisma Projects.

### Environment & setup

Postgres

### Prisma Version

```
prisma                  : 6.2.1
@prisma/client          : Not found
Node.js                 : v22.0.0
TypeScript              : 5.4.5
```

viktor-ferenczi:
Modified CLI behavior for safer defaults regarding resetting (dropping) the whole database:
- Require `--allow-reset` for `migrate dev` to drop the database.
- Require `--force-reset` for `db push` to drop the database.

**Remarks**
- No extra confirmation prompt is presented if the above flags are used.
- No configuration is required to avoid resetting the database.

Fixes #26076
Fixes #21483

### DIFF
--- a/packages/internals/src/utils/canPrompt.ts
+++ b/packages/internals/src/utils/canPrompt.ts
@@ -5,9 +5,16 @@ import { isInteractive } from './isInteractive'
 
 // If not TTY or in CI we want to throw an error and not prompt.
 // Because:
-// Prompting when non interactive is not possible.
+// Prompting when non-interactive is not possible.
 // Prompting in CI would hang forever / until a timeout occurs.
+// We use prompts.inject() for testing prompts in our tests.
 export const canPrompt = (): boolean => {
-  // Note: We use prompts.inject() for testing prompts in our CI
-  return Boolean((prompt as any)._injected?.length) === true || (isInteractive() && !isCi())
+  const injectedCount = (prompt as any)._injected?.length
+  if (injectedCount) {
+    // This is debug logging you can enable, but it will cause the affected test cases to fail output validation.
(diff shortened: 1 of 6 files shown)


FGoessler (approved):
Nice first work! 🎉

Issue #25889, opened 2024-12-18
RFC: `prisma migrate commit`


aqrln:
# RFC: `prisma migrate commit`

## Problem statement

When working on a new feature in your application that involves changing the database schema you often need need multiple iterations until you get it right, especially when adding new models.

The case of initial prototyping before creation of the first migration is supported in Prisma well but iterating on new changes without thorough planning is not as ergonomic.

Applying each and every change with `migrate dev` will lead to long and messy migration history that needs to be squashed and rewritten before merging, especially if some of the steps may be undoing the others. Unnecessary intermediate migrations can lead to longer deployment and migration times, increased risk of errors and increased risk of degraded user experience due to locking if they touch existing tables. Squashing them is not hard but it’s a hassle.

Using `db push` during development, like during the initial prototyping, and then creating a migration using `migrate dev` before committing the code is more straightforward and leads to better results. The downside, however, is that creating that migration requires resetting the development database:

```
Drift detected: Your database schema is not in sync with your migration history.

The following is a summary of the differences between the expected database schema given your migrations files, and the actual schema of the database.

It should be understood as the set of changes to get from the expected schema to the actual schema.

[*] Changed the `Post` table
  [+] Added column `banana`

? We need to reset the PostgreSQL database "db" at "localhost:5432".
Do you want to continue? All data will be lost. > (y/N)
```

While it is certainly a good practice to have a seed script with good test data that can be applied any moment and treat the development database as disposable, it is not always the case, and sometimes it is desirable to keep the local data.

## Proposed solution

I propose a new command for creating migrations: `migrate commit`. Unlike `migrate dev`, it requires the database schema to be in sync with the Prisma schema but not with the latest migration.

This command writes the pending changes in the database schema into a new migration file, and marks this migration as applied. This allows to use `db push` to interactively iterate on the schema, and then commit the changes at once without losing any data.

Essentially this command *expects* the database schema to be different than the state recorded in the migration history, and diffs from the migrations directory to the database rather than to the schema file (hence the requirement that the database schema and the Prisma schema are in sync, which is not required or even expected by `migrate dev`).

![8ffb5b22-95ca-45c8-b442-6a2ac0d8db31](https://github.com/user-attachments/assets/8a71ae59-20f3-4157-8f32-ab07a4081b7f)

## Other use case: baselining

Mechanically this is very similar to baselining, the only difference being that for baselining we diff from the empty state and not from the existing migrations.

[...]

PR #26406, opened 2025-02-22
Schema Drift Baseline Option

Hrios05:
>Co-authored-by: Kaulan Serzhanuly <kaulan.2004@gmail.com>
>Co-authored-by: Amara Whitson <amarawhitson@gmail.com>

Issue https://github.com/prisma/prisma/issues/26076 was closed but we worked on this and hoping for feedback for coursework.


packages/migrate/src/commands/Untitled-1.jsonc
aqrln, line 1: What are these two `jsonc` files?

aqrln (commented):
Thank you for the PR! So far our thinking was to offer a separate command for this (see https://github.com/prisma/prisma/issues/25889) but having an interactive prompt to choose between three options makes total sense given that we had a yes/no prompt until just now. I don't know if we want to bring this prompt back in a safer way that you you suggest or have users pass the CLI flags explicitly — that's a UX question we need to think about.

That said, I think we should offer the standalone baseline/commit command regardless, so it should be implemented first, and then you'd just call it here rather than inline the implementation ad-hoc.

packages/migrate/src/commands/Untitled-1.jsonc
KaulanSerzhanuly: Hello 
Sorry i forgot to delete that file. Its for pretty and prettier extension.

aqrln:
Folks, I'm going to close this PR for now. Would you like to have a go at implementing https://github.com/prisma/prisma/issues/25889? We could then revisit this PR and consider adding this prompt once we have the actual primitive it will use implemented.

Regardless, thanks for your enthusiasm and the desire to contribute!

KaulanSerzhanuly:
Thank you! 
If we have time we will work on that issue!
Best Regards
Kaulan

PR #26420, opened 2025-02-24
feat(client): replace `detect-runtime` with `std-env`

jkomyno:
This PR:
- Closes [ORM-624](https://linear.app/prisma-company/issue/ORM-624/replace-deprecated-detect-runtime-dependency)
- Closes https://github.com/prisma/team-orm/issues/419

PR #9561, opened 2021-10-01
feat(client): data proxy

millsp:
This is still a work in progress (at prototype stage), as we don't have any kind of testing yet besides the manual ones we do.

PS: Save yourself the pain of going through the diff, use your favorite git diff tool and watch commit per commit for the changes.

From a higher view, this PR also brings in many necessary changes:
* A refactor of our whole compile pipeline, now with a centralized build system.
* Support for ESM modules (not enabled publicly), but picked up by the bundlers.
* A new runtime build in order to support CFW, with its all node polyfill injections.
* Tree-shaking of CommonJS modules via the compilation of tree-shaken ESM modules.
* Some cleanup here and there in order to make things work more easily.
* A new utility library that is aimed towards performance and type-safety.
* And a bit of documentation.

PR #26421, opened 2025-02-24
feat(client): replace "detect-runtime" with internal implementation

jkomyno:
Size-sensitive alternative to https://github.com/prisma/prisma/pull/26420.

This PR:
- Closes [ORM-624](https://linear.app/prisma-company/issue/ORM-624/replace-deprecated-detect-runtime-dependency)
- Closes https://github.com/prisma/team-orm/issues/419

### DIFF
--- a/packages/client/src/runtime/utils/getRuntime.ts
+++ b/packages/client/src/runtime/utils/getRuntime.ts
@@ -8,41 +8,40 @@ export type RuntimeName =
   | 'edge-light'
   | '' /* unknown */
 
+/**
+ * Indicates if running in Node.js or a Node.js compatible runtime.
+ *
+ * **Note:** When running code in Bun and Deno with Node.js compatibility mode, `isNode` flag will be also `true`, indicating running in a Node.js compatible runtime.
+ */
+const isNode = () => globalThis.process?.release?.name === 'node'
+
+/**
+ * Indicates if running in Bun runtime.
+ */
+const isBun = () => !!globalThis.Bun || !!globalThis.process?.versions?.bun
+
+/**
+ * Indicates if running in Deno runtime.
+ */
+const isDeno = () => !!globalThis.Deno
+
+/**
+ * Indicates if running in Netlify runtime.
+ */
+const isNetlify = () => typeof globalThis.Netlify === 'object'
+
+/**
+ * Indicates if running in EdgeLight (Vercel Edge) runtime.
+ */
+const isEdgeLight = () => typeof globalThis.EdgeRuntime === 'object'
+
+/**
+ * Indicates if running in Cloudflare Workers runtime.
+ * See: https://developers.cloudflare.com/workers/runtime-apis/web-standards/#navigatoruseragent
+ */
+const isWorkerd = () => globalThis.navigator?.userAgent === 'Cloudflare-Workers'
   
 function detectRuntime(): RuntimeName {
-  /**
-   * Indicates if running in Node.js or a Node.js compatible runtime.
-   *
-   * **Note:** When running code in Bun and Deno with Node.js compatibility mode, `isNode` flag will be also `true`, indicating running in a Node.js compatible runtime.
-   */
-  const isNode = globalThis.process?.release?.name === 'node'
-  
-  /**
-   * Indicates if running in Bun runtime.
-   */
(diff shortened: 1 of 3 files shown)


packages/client/src/runtime/utils/getRuntime.ts
```diff
@@ -1,4 +1,69 @@
...
+  
+  /**
+   * Indicates if running in Netlify runtime.
+   */
+  const isNetlify = typeof globalThis.Netlify === 'object'
+  
+  /**
+   * Indicates if running in EdgeLight (Vercel Edge) runtime.
+   */
+  const isEdgeLight = typeof globalThis.EdgeRuntime === 'object'
```
aqrln, line 38: Why different styles of checks (i.e. `!!globalThis.X` vs `typeof globalThis.X === 'object')?

packages/client/src/runtime/utils/getRuntime.ts
```diff
@@ -1,4 +1,69 @@
...
+    [isNode, 'node'],
+  ] as const
+
+  const detectedRuntime = runtimeChecks
+    // TODO: Transforming destructuring to the configured target environment ('chrome58', 'edge16', 'firefox57', 'safari11') is not supported yet,
+    // so we can't write the following code yet:
+    // ```
+    // .flatMap(([detected, runtime]) => detected ? [runtime] : [])
+    // ```
+    .flatMap(check => check[0] ? [check[1]] : [])
```
jkomyno, line 62: Side-note: the `'chrome58', 'firefox57', 'safari11', 'edge16'` targets for `browser` builds of Prisma have been set at least in October 2021, in https://github.com/prisma/prisma/pull/9561.

We should update them, but that would technically be a breaking change.

packages/client/src/runtime/utils/getRuntime.ts
```diff
@@ -1,4 +1,68 @@
...
+/**
+ * Indicates if running in EdgeLight (Vercel Edge) runtime.
+ */
+const isEdgeLight = () => typeof globalThis.EdgeRuntime === 'object'
+
+/**
+ * Indicates if running in Cloudflare Workers runtime.
+ * See: https://developers.cloudflare.com/workers/runtime-apis/web-standards/#navigatoruseragent
+ */
+const isWorkerd = () => globalThis.navigator?.userAgent === 'Cloudflare-Workers'
```
jkomyno, line 42: The values returned by these functions are accessed lazily by `detectRuntime()`. This is necessary for tests to properly set up a mocked `globalThis` object. Without this, we'd have failures in `data-proxy` functional tests.

packages/client/src/runtime/utils/getRuntime.ts
```diff
@@ -1,4 +1,68 @@
-import { detectRuntime, Runtime } from 'detect-runtime'
```
jkomyno, line 1: Had we replaced this with `std-env` altogether`, we'd have a 3KB larger `@prisma/client` and adapters bundle, see https://github.com/prisma/prisma/pull/26420.

packages/client/src/runtime/utils/getRuntime.ts
jkomyno: Because the second is more accurate, and was originally used in `detect-runtime`

packages/client/src/runtime/utils/getRuntime.ts
```diff
@@ -1,4 +1,68 @@
...
+    [isDeno, 'deno'],
+    [isBun, 'bun'],
+    [isNode, 'node'],
+  ] as const
+
+  const detectedRuntime = runtimeChecks
+    // TODO: Transforming destructuring to the configured target environment ('chrome58', 'edge16', 'firefox57', 'safari11') is not supported yet,
+    // so we can't write the following code yet:
+    // ```
+    // .flatMap(([isCurrentRuntime, runtime]) => isCurrentRuntime() ? [runtime] : [])
```
aqrln, line 59: What why? Shouldn't esbuild do exactly the transformation you wrote yourself?

And why do we even target such old environments when we require Node.js 18+ or an equivalent modern runtime?

jkomyno: I had a comment about this elsewhere: https://github.com/prisma/prisma/pull/26421#discussion_r1967539908

aqrln (approved):

PR #26429, opened 2025-02-25
fix(omit api): workaround to speed up tsc if global omit api is not used

FGoessler:
While investigating the source for slow typescript compilation and autocomplete when using client extension I noticed:

- The compiler gets slower when a options object is passed to the `PrismaClient` constructor.
- Using explicit `ReturnType<typeof getClient>` of an extended `PrismaClient` slows down the compiler.
- We pass down the `ClientOptions` object type from the constructor all the way down into various method calls to compute the type of the returned models based on global omit configurations.
- It does not matter which keys are set on the options object and it gets slow - e.g. `new PrismaClient({log: ["info"] })` is enough to make it slow.
- The problem only occurs in combination of client extensions + options object.
- Client extensions alone are OK. Options object / omit API alone is OK.

=> This PR attempts a workaround to narrow the usage of ClientOptions to only be passed down if a global omit configuration is actually present.
=> Only in the case of global omit config + client extensions the compiler gets slow.

This is still not ideal of course but should significantly lower the number of affected users until we have a proper solution in place.


FGoessler:
CI test failure should be fixed through https://github.com/prisma/prisma/pull/26437

jkomyno (approved):

FGoessler:
Closing in favor of: https://github.com/prisma/prisma/pull/26470

PR #26450, opened 2025-02-26
feat(client): fix read-replicas extension

jkomyno:
This PR should fix [ORM-662](https://linear.app/prisma-company/issue/ORM-662/fix-broken-read-replicas-extension).

Context:
- [`@prisma/extension-read-replicas`](https://github.com/prisma/extension-read-replicas) uses [`Object.getPrototypeOf(client).constructor`](https://github.com/prisma/extension-read-replicas/blob/0be44f860d0a8144afaf266f2aefc5b3a50514e9/src/extension.ts#L27) to create a new `PrismaClient` instance for each replica URL, in [`ReplicaManager`](https://github.com/prisma/extension-read-replicas/blob/141978067900c7fb9380a39a80eeea1239d0dcf7/src/ReplicaManager.ts#L33-L35).
- https://github.com/prisma/prisma/pull/25914, which was introduced to allow (de)serialising Prisma Clients across Cloudflare Workers RPC boundaries, added the following prototype getter function to `createCompositeProxy` and `defaultProxyHandlers`:

  ```typescript
  getPrototypeOf: () => Object.prototype
  ```

- This PR replaces its usage with

  ```typescript
  getPrototypeOf: () => Object.getPrototypeOf(client._originalClient)
  ```

  and adjusted both `@prisma/client` and tests to pass a `client: PrismaClient` along to `createCompositeProxy` and `defaultProxyHandlers`.

/integration


jkomyno:
The integration version is being released [here](https://github.com/prisma/prisma/actions/runs/13544987377/job/37854312489).

packages/client/src/runtime/core/compositeProxy/createCompositeProxy.ts
```diff
@@ -39,7 +40,7 @@ const customInspect = Symbol.for('nodejs.util.inspect.custom')
  * @param layers
  * @returns
  */
-export function createCompositeProxy<T extends object>(target: T, layers: CompositeProxyLayer[]): T {
+export function createCompositeProxy<T extends object>(client: Client, target: T, layers: CompositeProxyLayer[]): T {
```
aqrln, line 43: I don't think this utility should be aware of `Client` (and the need to unsoundly cast empty objects to `Client` in all tests is a good evidence of this), it's on a different level of abstraction.

Instead, it should take an abstract non-nullable value to get the prototype from:
```suggestion
export function createCompositeProxy<T extends object>(prototypeProvider: {}, target: T, layers: CompositeProxyLayer[]): T {
```

and then on line 177:

```diff
- getPrototypeOf: () => Object.getPrototypeOf(client._originalClient),
+ getPrototypeOf: () => Object.getPrototypeOf(prototypeProvider),
```

Then you can pass `client._originalClient` or `this._originalClient` when you actually do have a client, or an empty object without having to cast anything in the tests.

aqrln: alternatively, you can pass the prototype itself, that should also be fine

PR #26453, opened 2025-02-26
feat: allow chaining $on with $extends

jacek-prisma:
[ORM-614](https://linear.app/prisma-company/issue/ORM-614/method-dollaron-of-new-prismaclient-instance-returns-void)

### DIFF
--- a/packages/client/tests/functional/extensions/model.ts
+++ b/packages/client/tests/functional/extensions/model.ts
@@ -62,7 +62,6 @@ testMatrix.setupTestSuite(
       xprisma.user.extMethod()
 
       expect(extMethod).toHaveBeenCalledTimes(1)
-      expect((xprisma.post as any).extMethod).toBeUndefined()
     })
 
     test('extend all models', () => {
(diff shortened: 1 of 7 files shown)


packages/client/tests/functional/extensions/model.ts
```diff
@@ -43,6 +43,28 @@ testMatrix.setupTestSuite(
...
+            user: {
+              extMethod,
+            },
+          },
+        })
+
+      xprisma.user.extMethod()
+
+      expect(extMethod).toHaveBeenCalledTimes(1)
+      expect((xprisma.post as any).extMethod).toBeUndefined()
```
jkomyno, line 65: Not sure why this line is needed

jkomyno (approved):

packages/client/tests/functional/extensions/model.ts
jacek-prisma: not really needed, just carried over from another test, I'll remove it

PR #26455, opened 2025-02-27
feat(migrate): Removed --allow-reset

viktor-ferenczi:
This is a follow-up on #26378 

In order to simplify the CLI the `--allow-reset` command line option introduced to `migrate dev` in the above PR is removed. Instead the CLI suggests using `migrate reset` directly.

Changes:
- Removed `--allow-reset` from `migrate dev`
- Adjusted the affected test cases

### DIFF
--- a/packages/migrate/src/__tests__/MigrateDev.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDev.test.ts
@@ -9,6 +9,7 @@ import prompt from 'prompts'
 
 import { DbExecute } from '../commands/DbExecute'
 import { MigrateDev } from '../commands/MigrateDev'
+import { MigrateReset } from '../commands/MigrateReset'
 import { CaptureStdout } from '../utils/captureStdout'
 import { setupCockroach, tearDownCockroach } from '../utils/setupCockroach'
 import { setupMSSQL, tearDownMSSQL } from '../utils/setupMSSQL'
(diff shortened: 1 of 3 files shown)


packages/migrate/src/__tests__/MigrateDev.test.ts
```diff
@@ -502,7 +465,8 @@ describe('sqlite', () => {
 
       We need to reset the SQLite database "dev.db" at "file:dev.db"
       
-      You may use --allow-reset to drop the database. All data will be lost.
+      You may use prisma migrate reset to drop the development database.
```
jacek-prisma, line 468:
```suggestion
      You may use the `prisma migrate reset` command to drop the development database.
```
I'd wrap it in some quotes to clarify it's a command

jacek-prisma (approved):
left just 1 minor comment

packages/migrate/src/__tests__/MigrateDev.test.ts
viktor-ferenczi: The command has already been wrapped in red:
```ts
      process.stdout.write(
        '\n' +
          `You may use ${red('prisma migrate reset')} to drop the development database.\n` +
          `${bold(red('All data will be lost.'))}\n`,
      )
```

The stdout capture in the test is just clearing all the coloring, therefore it does not show up in the expected output.

PR #26483, opened 2025-03-03
feat(cli): define `prisma.config.ts` without requiring `@prisma/config`

jkomyno:
This PR:
- closes [ORM-648](https://linear.app/prisma-company/issue/ORM-648/prisma-config-export-type-prismaconfig-and-defineconfig-from-prisma)
- introduces types export from Prisma, specifically to expose `PrismaConfig`:
  ```typescript
  import path from 'node:path'
  import type { PrismaConfig } from 'prisma'
  
  export default {
    earlyAccess: true,
    schema: {
      kind: 'single',
      filePath: path.join('..', 'prisma', 'schema.prisma'),
    },
  } satisfies PrismaConfig
  ```
- introduces `prisma/config`, which exposes a subset of `@prisma/config` useful to Prisma users, without forcing them to manually download yet another Prisma package:
  ```typescript
  import path from 'node:path'
  
  export default defineConfig({
    earlyAccess: true,
    schema: {
      kind: 'single',
      filePath: path.join('..', 'prisma', 'schema.prisma'),
    },
  })

- updates the `basic-sqlite-config` sandbox to mirror the new recommended strategy for defining Prisma config files

/integration


packages/cli/helpers/build.ts
```diff
@@ -74,7 +109,12 @@ const preinstallBuildConfig: BuildOptions = {
   minify: true,
 }
 
-void build([cliBuildConfig, preinstallBuildConfig])
+const optionalPlugins = process.env.DEV === 'true' ? [] : [cliTypesBuildConfig, cliConfigBuildConfig]
```
jkomyno, line 112: Important! Without this, `pnpm run -r dev` will fail.

packages/cli/src/config.ts
```diff
@@ -0,0 +1 @@
+export { defineConfig, type PrismaConfig, type PrismaConfigInternal } from '@prisma/config'
```
jkomyno, line 1: `PrismaConfigInternal` must be exported, otherwise

```typescript
```

will result in a

```
The inferred type of `default` cannot be named without a reference to `...`. This is likely not portable. A type annotation is necessary.
```

error.

packages/cli/src/platform/auth/login.ts
```diff
@@ -106,6 +106,7 @@ export class Login implements Command {
 
     {
       const writeResult = await credentialsFile.save({ token: callbackResult.token })
+      // @ts-ignore: Expected 0-1 arguments, but got 2.
```
aqrln, line 109: Why is this necessary? Do we have outdated typings or target older environment than we actually require? `Error` does take 0-2 arguments in modern JavaScript.

jkomyno: Fixing this in a follow-up PR.

aqrln (approved):
Nice work!

packages/cli/src/platform/auth/login.ts
jkomyno, line 109:
```suggestion
```

packages/cli/src/platform/auth/login.ts
```diff
@@ -222,6 +222,7 @@ export const loginOrSignup = async () => {
 
   {
     const writeResult = await credentialsFile.save({ token: callbackResult.token })
+    // @ts-ignore: Expected 0-1 arguments, but got 2.
```
jkomyno, line 225:
```suggestion
```

PR #26485, opened 2025-03-03
fix: use webcrypto to fix WASM compat

jacek-prisma:
Attempts to fix ecosystem tests

/integration

### DIFF
--- a/packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
+++ b/packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
@@ -1,6 +1,5 @@
 import Debug from '@prisma/debug'
 import { ErrorCapturingSqlConnection, ErrorCapturingTransaction, SqlQuery } from '@prisma/driver-adapter-utils'
-import { webcrypto } from 'crypto'
 
 import { assertNever } from '../utils'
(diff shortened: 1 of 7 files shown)


packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
```diff
@@ -62,7 +61,7 @@ export class TransactionManager {
     const validatedOptions = this.validateOptions(options)
 
     const transaction: TransactionWrapper = {
-      id: crypto.randomUUID(),
+      id: webcrypto.randomUUID(),
```
jkomyno, line 64: The following is available in the minimum Node.js version we require, 18.18.x:
```suggestion
      id: globalThis.crypto.randomUUID(),
```

packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
```diff
@@ -1,7 +1,6 @@
-import crypto from 'node:crypto'
-
 import Debug from '@prisma/debug'
 import { ErrorCapturingSqlConnection, ErrorCapturingTransaction, SqlQuery } from '@prisma/driver-adapter-utils'
+import { webcrypto } from 'crypto'
```
jkomyno, line 3:
```suggestion
```

PR #26488, opened 2025-03-03
feat: allow readonly arrays as input on where and similar args

FGoessler:


FGoessler:
This was an experiment to see how easy it would be to address #5025 but we run into various type issues here in the test suites. => Not a quick fix and gonna shelf this for now.

PR #26497, opened 2025-03-04
fix: use web crypto api instead of node:crypto - better compatibility with CF workers

FGoessler:
Integration test run demonstrating the fixed behavior: https://github.com/prisma/ecosystem-tests/actions/runs/13658260462

/integration


helpers/compile/plugins/fill-plugin/fillers/decimal-small.ts
```diff
@@ -1,5 +1,3 @@
-import crypto from 'node:crypto'
```
FGoessler, line 1: ℹ️ That's the main culprit that broke things. Added it in https://github.com/prisma/prisma/pull/26437/files based on editor/lint suggestions...

jkomyno (approved):

PR #26536, opened 2025-03-06
fix: export index.js for compatibility

jacek-prisma:
Exports `build/index.js`, since it's used by at least one prisma integration that is currently broken: https://github.com/redwoodjs/redwood/blob/11c88f0f7bde98ba0c9d4726a548a5da500f6334/packages/cli/src/lib/generatePrismaClient.js#L25.

/integration

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -95,6 +95,13 @@
         "default": "./config.js"
       },
       "default": "./config.js"
+    },
+    "./build/index.js": {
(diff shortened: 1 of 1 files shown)


jacek-prisma:
The ecosystem passes on the integration branch: https://github.com/prisma/ecosystem-tests/actions/runs/13704867902/job/38327808804

FGoessler (approved):

PR #26544, opened 2025-03-07
feat(client): get rid of patched dependencies

jkomyno:
This PR:
- closes [ORM-797](https://linear.app/prisma-company/issue/ORM-797/tech-debt-get-rid-of-patches-in-prismaprisma)
- removes old patch of `@neonserverless/database`, initially introduced due to `globalThis.crypto` shims
- removes old patch of `node-fetch`, of a version not currently used
  - Q: Why do we still use `node-fetch` in the first place, now that `fetch` is available?
  - A: We still use it because `node-fetch` allows passing custom `Agent`s and `ProxyAgent`s to it, while std `fetch` does not. We can probably get rid of it once we get rid of `@prisma/fetch-engine`.


jacek-prisma:
might be worth running /integration to see if any ecosystem tests are affected

jkomyno:
Integration release workflow here: https://github.com/prisma/prisma/actions/runs/13760555973

jacek-prisma (approved):

jkomyno:
Note: let's not merge right now. An e2e test is failing consistently.

PR #26561, opened 2025-03-10
feat(neon): support neon serverless v1

FGoessler:
Adding support for upcoming neon v1 as per https://github.com/neondatabase/serverless/pull/145.

### DIFF
--- a/packages/adapter-neon/package.json
+++ b/packages/adapter-neon/package.json
@@ -47,6 +47,6 @@
     "jest-junit": "16.0.0"
   },
   "peerDependencies": {
-    "@neondatabase/serverless": "^0.6.0 || ^0.7.0 || ^0.8.0 || ^0.9.0 || ^0.10.0"
+    "@neondatabase/serverless": ">0.6.0 <2"
   }
 }
--- a/packages/adapter-neon/src/neon.test.ts
+++ b/packages/adapter-neon/src/neon.test.ts
@@ -1,6 +1,6 @@
 import { Client, Pool } from '@neondatabase/serverless'
 
-import { PrismaNeon } from './neon'
+import { PrismaNeon, PrismaNeonHTTP } from './neon'
 
 describe('validation', () => {
   test('throws if passed Client instance', () => {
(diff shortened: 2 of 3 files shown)


packages/adapter-neon/src/neon.test.ts
```diff
@@ -26,3 +26,44 @@ describe('validation', () => {
     }).not.toThrow()
   })
 })
+
+describe('neon version support', () => {
+  it('supports @neondatabase/serverless < 1.0.0', async () => {
+    // Before v0 client is directly callable to execute the SQL
+    const v0NeonClientMock = jest
+      .fn()
+      .mockResolvedValue({ fields: [], command: '', rowCount: 0, rows: [], rowAsArray: false })
```
jkomyno, line 35: Why would we test a mock, rather than testing how the actual `@neondatabase/serverless` client works?

We can store multiple versions of it in `package.json`'s `devDependencies`, as long as we mark them via aliases:

```json
{
  "devDependencies": {
    "serverless-v1": "npm:@neondatabase/serverless@^0.10.0",
    "serverless-v2": "npm:@neondatabase/serverless@^1.0.0"
  }
}
```

packages/adapter-neon/package.json
```diff
@@ -47,6 +47,6 @@
     "jest-junit": "16.0.0"
   },
   "peerDependencies": {
-    "@neondatabase/serverless": "^0.6.0 || ^0.7.0 || ^0.8.0 || ^0.9.0 || ^0.10.0"
+    "@neondatabase/serverless": ">0.6.0 <2"
```
jkomyno, line 50: This reminds me: we have no policy about what versions of third-party drivers we want to support.

packages/adapter-neon/src/neon.test.ts
FGoessler: Oh - I wasn't aware you could even do that. Will adjust it that way :+1:

FGoessler: I actually can't improve the test much in terms of not mocking things. If I want to fully avoid mocking we'd have to connect to real neon database here (or a containerized setup using the neon-proxy).

I think that's not worth the overhead and effort here. I rather instead update our neon ecosystem tests to use neon v1 once it is released.

jkomyno (approved):
Approved to unblock the release, but there's leftover work to do on the testing side

PR #26577, opened 2025-03-11
feat: change default output location in `prisma init`

aqrln:
Don't suggest generating in `node_modules` by default in `prisma init` and instead generate in the source tree.

The base directory for the location of the generated client is chosen based on the following rules:
- If there's a `tsconfig.json` and it explicitly defines the root directory, we use that
- If there's a directory called `src`, `lib` or `app`, we use that (in this order)
- Otherwise we use the current directory

In this directory, we create `generated/prisma`.

The directory is chosen in anticipation of getting rid of the `index.js` barrel file in the future ESM client and introducing the `client.ts` entrypoint, so that the full import path would be `./generated/prisma/client`. If we used `generated/prisma/client` as the output now, we'd have to change it again, otherwise we'd end up with `./generated/prisma/client/client`. To ease the transition, we'll add the alternative `client` entrypoint in the current client as well.

Closes: https://linear.app/prisma-company/issue/ORM-777/update-prisma-init-to-include-a-custom-outputpath-in-the-default

### DIFF
--- a/packages/cli/src/init/client-output-path.ts
+++ b/packages/cli/src/init/client-output-path.ts
@@ -1,0 +1,98 @@
+import fs from 'node:fs'
+import path from 'node:path'
+
+import { Schema as Shape } from 'effect'
+import { readConfigFile } from 'typescript'
+
+/**
+ * Determines the client output path relative to the schema directory.
+ *
+ * @param schemaDir absolute path to the schema directory
+ * @returns client output path relative to the schema directory
+ */
+export function determineClientOutputPath(schemaDir: string): string {
+  const sourceDir = getSourceDir()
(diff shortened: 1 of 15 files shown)


packages/cli/src/init/client-output-path.ts
```diff
@@ -0,0 +1,98 @@
+import fs from 'node:fs'
+import path from 'node:path'
+
+import { Schema } from 'effect'
```
FGoessler, line 4: 💡 I think we used to alias this import to `Shape` in the `prisma/config` stuff to avoid confusion with the overloaded term `Schema` in our domain. Maybe worthwhile here, too.

FGoessler (approved):
I like the logic - code wise also looks good. Just a minor nitpick idea. :)

PR #26601, opened 2025-03-12
feat(qc): Refactor dynamic placeholder approach

viktor-ferenczi:
- Rendering SQL queries from template fragments
- Removed `prisma-comma-repeatable` temporary hack

Implements [ORM-566](https://linear.app/prisma-company/issue/ORM-566/refactor-dynamic-placeholder-approach)

[Engine PR](https://github.com/prisma/prisma-engines/pull/5272)


packages/client-engine-runtime/jest.config.js
```diff
@@ -1,3 +1,5 @@
+require('dotenv').config({ path: '../../.db.env' })
```
aqrln, line 1: Why do we need the connection strings in these tests? I don't think we do any real IO here, there are only unit tests

packages/client-engine-runtime/src/interpreter/renderQuery.test.ts
```diff
@@ -0,0 +1,178 @@
...
+import { ScopeBindings } from './scope'
+
+test('no template', () => {
+  expect(
+    renderQuery(
+      {
+        type: 'rawSql',
+        sql: 'SELECT * FROM users WHERE id = $1',
+        params: [1],
+      } as QueryPlanDbQuery,
```
aqrln, line 12:
```suggestion
      } satisfies QueryPlanDbQuery,
```

aqrln: `as` asserts the type (even if it's not fully compatible) and can cause us to miss errors. `satisfies` actually checks it's assignable to the specified type.

packages/client-engine-runtime/src/interpreter/renderQuery.test.ts
```diff
@@ -0,0 +1,178 @@
...
+      {} as ScopeBindings,
```
aqrln, line 13:
```suggestion
      {} satisfies ScopeBindings,
```

aqrln: and everywhere else

packages/client-engine-runtime/jest.config.js
aqrln: That makes sense, e.g., for the functional tests in `packages/client`. What I don't understand is why this is necessary for the tests to work inside IDE in this package specifically.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -1,11 +1,23 @@
...
+export function renderQuery(dbQuery: QueryPlanDbQuery, scope: ScopeBindings): SqlQuery {
+  switch (dbQuery.type) {
+    case 'rawSql':
+      return renderRawSql(dbQuery.sql, substituteParams(dbQuery.params, scope))
+
+    case 'templateSql':
+      return renderTemplateSql(dbQuery.fragments, dbQuery.placeholder, substituteParams(dbQuery.params, scope))
+
+    default:
+      return dbQuery // never
```
aqrln, line 15:
```suggestion
      assertNever(dbQuery.type, 'Invalid query type')
```
so that we get a compile error if we don't handle any newly added case (and an additional failsafe at runtime)

aqrln: `asserNever` is defined in `utils.ts` in this package

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -15,16 +27,62 @@ export function renderQuery({ query, params }: QueryPlanDbQuery, scope: ScopeBin
...
+                  .map((value) => {
+                    flattenedParams.push(value)
+                    return formatPlaceholder(placeholderFormat, placeholderNumber++)
+                  })
+                  .join(',')
+          return `(${placeholders})`
+        }
+
+        default:
+          return fragment // never
```
aqrln, line 68:
```suggestion
          assertNever(fragment.type, 'Invalid fragment type')
```

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -15,16 +27,62 @@ export function renderQuery({ query, params }: QueryPlanDbQuery, scope: ScopeBin
...
+}
 
-  const { query: renderedQuery, params: expandedParams } = renderQueryTemplate({ query, params: substitutedParams })
+function formatPlaceholder(placeholderFormat: PlaceholderFormat, placeholderNumber: number): string {
+  return placeholderFormat.hasNumbering ? `${placeholderFormat.prefix}${placeholderNumber}` : placeholderFormat.prefix
+}
 
-  const argTypes = expandedParams.map((param) => toArgType(param as PrismaValue))
+function renderRawSql(sql: string, params: PrismaValue[]): SqlQuery {
+  const argTypes = params.map((param) => toArgType(param as PrismaValue))
```
aqrln, line 81: why is `as PrismaValue` necessary here? `params` is already `PrismaValue[]`

aqrln (commented):
The logic and tests look solid! I only have a bunch of cosmetic comments, other than that it looks great to me.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -15,16 +27,62 @@ export function renderQuery({ query, params }: QueryPlanDbQuery, scope: ScopeBin
...
+          const paramArray = Array.isArray(paramValue) ? paramValue : [paramValue]
+          const placeholders =
+            paramArray.length == 0
+              ? 'NULL'
+              : paramArray
+                  .map((value) => {
+                    flattenedParams.push(value)
+                    return formatPlaceholder(placeholderFormat, placeholderNumber++)
+                  })
+                  .join(',')
```
aqrln, line 63:
```suggestion
                  .join(', ')
```
otherwise some snapshot tests are probably gonna fail

packages/client-engine-runtime/jest.config.js
viktor-ferenczi: You're right, it is not needed here, since there is no real DB access. Removed.

packages/client-engine-runtime/src/interpreter/renderQuery.test.ts
viktor-ferenczi: Thanks, I wasn't aware of this potential issue. I fix this in all places.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
viktor-ferenczi: Eh, I had the runtime fail-safe there, then removed it as "unnecessary". Looks like it is actually good practice then.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
viktor-ferenczi: Okay, so we prefer readability of the generated SQL here to slightly less memory usage. Understood.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
aqrln: I'm mostly interested in compile time file-safe, e.g. I'd be okay with just
```ts
  default:
    dbQuery.type satisfies never
```

`assertNever` combines both this and a runtime exception (that should be unreachable at run time) — this is indeed redundant but doesn't hurt.

aqrln: I guess your variant also works because otherwise there would be a type error due to `dbQuery` not being assignable to the return type of the function, but it's a little bit more brittle than explicitly requiring that `dbQuery.type` is `never` in this branch (e.g. what if the types changed in the future and the type of `dbQuery` became a subtype of the return type?)

packages/client-engine-runtime/src/interpreter/renderQuery.ts
viktor-ferenczi: Nice catch. It is interesting that the unnecessary cast wasn't cough by the code inspection here.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -15,16 +27,62 @@ export function renderQuery({ query, params }: QueryPlanDbQuery, scope: ScopeBin
...
+          const paramArray = Array.isArray(paramValue) ? paramValue : [paramValue]
+          const placeholders =
+            paramArray.length == 0
+              ? 'NULL'
+              : paramArray
+                  .map((value) => {
+                    flattenedParams.push(value)
+                    return formatPlaceholder(placeholderFormat, placeholderNumber++)
+                  })
+                  .join(', ')
```
aqrln, line 63: Actually no, sorry, I was wrong. I thought we were formatting it like this but it's not the case, e.g.:
https://github.com/prisma/prisma/blob/0fa294aa0c67759375aeb92de78c6dc6b3135338/packages/client/tests/functional/0-legacy-ports/batch-find-unique/tests.ts#L87-L95

Let's revert this change:
```suggestion
                  .join(',')
```

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -15,16 +27,62 @@ export function renderQuery({ query, params }: QueryPlanDbQuery, scope: ScopeBin
...
+          return `(${placeholders})`
+        }
+
+        default:
+          assertNever(fragment.type, 'Invalid fragment type')
```
aqrln, line 68: the function needs to be imported from `../utils.ts`

packages/client-engine-runtime/src/interpreter/renderQuery.ts
viktor-ferenczi: No problem. Removed the space.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
viktor-ferenczi: Yeah, I figured it out. Tried to import it from another package, while we had a copy in `utils.ts`. Found it when I wanted to copy it exactly there :)

packages/client-engine-runtime/src/interpreter/renderQuery.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        placeholder: {
+          prefix: '$',
+          hasNumbering: true,
+        } as PlaceholderFormat,
+        params: [[], 0],
+      } as QueryPlanDbQuery,
+      {} as ScopeBindings,
+    ),
+  ).toEqual({
+    sql: 'SELECT * FROM users WHERE "userId" IN (NULL) OFFSET $1',
```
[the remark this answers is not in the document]
viktor-ferenczi: Added as [ORM-820](https://linear.app/prisma-company/issue/ORM-820/do-not-run-empty-queries)

jacek-prisma (approved):

packages/client-engine-runtime/package.json
```diff
@@ -31,7 +31,8 @@
     "@types/jest": "29.5.14",
     "@types/node": "18.19.76",
     "jest": "29.7.0",
-    "jest-junit": "16.0.0"
+    "jest-junit": "16.0.0",
+    "dotenv": "16.4.7"
```
jacek-prisma, line 35: why the `dotenv` addition?

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -15,16 +30,63 @@ export function renderQuery({ query, params }: QueryPlanDbQuery, scope: ScopeBin
...
+      switch (fragmentType) {
+        case 'parameter':
+          flattenedParams.push(params[paramIndex++])
+          return formatPlaceholder(placeholderFormat, placeholderNumber++)
+
+        case 'stringChunk':
+          return fragment.value
+
+        case 'parameterTuple': {
+          const paramValue = params[paramIndex++]
```
jacek-prisma, line 57: might be worth having a param length check here to ensure we're not looking past the array

packages/client-engine-runtime/src/interpreter/renderQuery.ts
```diff
@@ -15,16 +30,63 @@ export function renderQuery({ query, params }: QueryPlanDbQuery, scope: ScopeBin
...
+): SqlQuery {
+  let paramIndex = 0
+  let placeholderNumber = 1
+  const flattenedParams: PrismaValue[] = []
+  const sql = fragments
+    .map((fragment) => {
+      const fragmentType = fragment.type
+      switch (fragmentType) {
+        case 'parameter':
+          flattenedParams.push(params[paramIndex++])
```
jacek-prisma, line 50: similar as below:

might be worth having a param length check here to ensure we're not looking past the array

packages/client-engine-runtime/package.json
viktor-ferenczi: Removed.

packages/client-engine-runtime/src/interpreter/renderQuery.ts
viktor-ferenczi: Added explicit bounds checking with a useful error message.

jacek-prisma (approved):

PR #26612, opened 2025-03-13
feat(schema-engine): move file-disk access to TypeScript

jkomyno:
This is the TypeScript PR for https://github.com/prisma/prisma-engines/pull/5276.

Merge that before this one.
Warning: we will need to update the engines' version in this PR before merging it.


packages/internals/src/migrateTypes.ts
```diff
@@ -1,5 +1,71 @@
 // eslint-disable-next-line @typescript-eslint/no-namespace
 export namespace MigrateTypes {
+  export type MigrationDirectory = {
```
FGoessler, line 3: 💡 What about naming this simply `Migration`? 

That an individual Migration is a directory and not a single file is somewhat of a low level implementation detail. `MigrationDirectory` on the other hand is easily confused with the more common term `MigrationsDirectory` (that little `s` is easily overread 😬 ).

packages/migrate/src/Migrate.ts
```diff
@@ -55,44 +57,95 @@ export class Migrate {
...
 
-  public createMigration(params: EngineArgs.CreateMigrationInput): Promise<EngineResults.CreateMigrationOutput> {
-    return this.engine.createMigration(params)
+  public async createMigration(
+    params: Omit<EngineArgs.CreateMigrationInput, 'migrationsList'>,
+  ): Promise<{ generatedMigrationName: string | undefined }> {
+    if (!this.migrationsDirectoryPath) throw new Error('this.migrationsDirectoryPath is undefined')
+
+    const migrationsList = await listMigrations(this.migrationsDirectoryPath)
+    const { connectorType, generatedMigrationName, extension, migrationScript } = await this.engine.createMigration({
```
FGoessler, line 66: ❓ As `listMigrations` does not throw or abort on file read errors all the error handling for broken files etc is done on the Rust side of things, correct? Does this make our lifes significantly easier or actually harder as we have to serialize all the errors back and forth? 🤔

jkomyno: Life for us is unchanged. `listMigrations` treats loads most errors lazily, rather than eagerly, so that we don't need to change too many things at once. Our life is neither observably harder nor easier.

packages/internals/src/migrateTypes.ts
jkomyno: That would create more confusion without serving any obvious purpose: the fact that migrations are backed by disk files in a directory is not a low-level implementation detail at all, and e.g. the distinction between `MigrationDirectory` and `MigrationFile` is important to know what part of the migration workflow we actually want to adjust.

Besides, these names are merely copied from the Rust side, and will later be automatically generated by Rust.

PR #26624, opened 2025-03-14
refactor: drop legacy seed migration script warnings

FGoessler:
This PR drops a warning / helper message that we output to help users migrate from an old seed script setup to a new one. This was added over 3 years ago => I think by now all users who care about this have migrated and we can drop this logic. 

Done in scope of the `schemaPath` refactoring to get rid of the various schemaPath usages in this code. ;)


aqrln (approved):
I love seeing the code removed 🎉

PR #26657, opened 2025-03-17
chore(config): fix Windows tests

jkomyno:


packages/config/src/__tests__/loadConfigFromFile.test.ts
```diff
@@ -8,6 +9,14 @@ import { loadConfigFromFile, type LoadConfigFromFileError } from '../loadConfigF
 
 const ctx = jestContext.new().assemble()
 
+/**
+ * Normalize `filePath` to use forward slashes as a separator. `filePath` is
+ * treated as a path specific to the current platform, so backslashes will only
+ * be replaced with forward slashes on Windows. On other platforms, where a
+ * backslash is a valid filename character, it will be treated as such and will
+ * not be replaced.
+ */
```
FGoessler, line 18: 🧹 let's drop that comment again ;)

packages/config/src/__tests__/loadConfigFromFile.test.ts
```diff
@@ -157,9 +158,20 @@ describe('loadConfigFromFile', () => {
...
+        } else {
+          return { normalisedPath: resolvedPath, columnNumber: 3 }
+        }
+      })()
+
+      // eslint-disable-next-line jest/no-interpolation-in-snapshots
+      expect(error.error.message.replaceAll(normalisedPath, '<prisma-config>.ts')).toMatchInlineSnapshot(`
         "  [31mx[0m Unexpected eof
-           ,-[[36;1;4m<prisma-config>.ts[0m:5:3]
+           ,-[[36;1;4m<prisma-config>.ts[0m:5:${columnNumber}]
```
aqrln, line 174: If we're going to do this, then we should use `expect(...).toEqual(...)` and not a snapshot.

FGoessler (approved):

PR #26663, opened 2025-03-18
feat(schema folder): ORM 665 rework schema finding logic

FGoessler:
[ORM-665 Rework schema folder finding logic](https://linear.app/prisma-company/issue/ORM-665/rework-schema-folder-finding-logic)

## Summary

- `./prisma/schema` as a folder is not checked by default anymore.
- Migrations, views and typed sql folders are placed next to the schema file that contains the datasource block.
- We use a universal `schemaPath` property in `prisma.config.ts` to point to a folder or file.

## Remove default folder resolving

Instead of trying to find a Prisma Schema folder implicitly based on default locations we want to simplify the resolve logic to:

- `--schema` CLI option with either folder or filename
- `prisma.config.ts` - either folder or filename explicitly given
- package.json  -> `prisma.schema` - path to folder or filename given
- Default locations of `./prisma/schema.prisma` or `./schema.prisma` - no folder support 

(So far we would also check `./prisma/schema` if the preview feature is on.)

Why this change? => So it is very explicit for the user that they switched from a default location of a single file to a folder / custom based setup which will change where we look for things.

## Change default placement of migrations, views and typed sql folders

Also so far the schema folder has been the reference location to which we place the migrations, views and typed sql folders NEXT to it. Not into it. E.g.:

```
<<Current/old new state>>
/project-dir
|- /prisma
  |- /schema
    |- schema.prisma
  |- /migrations
  |- /views
  |- /sql
```

This was likely done this way to preserve the location of the migrations folder when switching from default single file (`./prisma/schema.prisma`) to default folder location (`./prisma/schema`).

Ideally we would also support glob patterns to find schema files across a larger variety of paths and enable more flexibility for the user.Even with the current refactoring, that is not easily possible, as we still have some legacy places that rely on the schema path structure (e.g. generators and env var loading).
To eventually support glob pattern based loading we cannot build our logic around a single file or directory path being passed in.

=> Determine the default location of those directories based on the location of the schema file that contains the datasource block. We already do this to find SQLite database files.

```
<<Expected new state  - given a schema path of e.g. `--schema=./prisma/schema`>>
/project-dir
|- /prisma
  |- /migrations
  |- /views
  |- /sql
  |- /more-schemas
    |- foo.schema
    |- bar.schema
  |- data-source-schema.prisma
```

> [!NOTE]
> Note: With [ORM-663](https://linear.app/prisma-company/issue/ORM-663/add-options-to-specify-migrations-folder-via-prismaconfigts) and [ORM-664](https://linear.app/prisma-company/issue/ORM-664/add-option-to-specify-views-folder-via-prismaconfigts) we will get additional customizable options.

Why this change? => Consistent behavior of how paths work in relation to schema files. Also most people should have their main schema.prisma file on the root of their specified schema path folder. 

## Unify config options

Most schema config options can take either a directory or a file. E.g. via the cli args or the `package.json` field. But for the `prisma.config.ts` we went with an explicit distinction. With the new code structure we don't really need this. We can easily detect whether a given path is a file or a folder and adjust our logic based on that.

Also moving forward we should treat everything as "multi schema file". Having a single schema file is just a special case of that whose details are only relevant during the initial file loading.

So prisma.config.ts becomes:
```
export default defineConfig({
  earlyAccess: true,
  schema: path.join(process.cwd(), 'prisma', 'schema.prisma'),
})
```

Why this change? => It's easier to configure, reduces config and code complexity on our end, is aligned with how most other tools accept paths and opens an easy way forward should we later on want to support glob patterns.


jkomyno (approved):

AmreeshTyagi:
Thanks for this improvement. I just observed one issue with different use case.

When I am using multiple databases without `prismaSchemaFolder` feature, migrations were generated in respected schema folders as I am putting different db schema in different folder in single file.

But when I enable prismaSchemaFolder feature so that I can create multiple schema for a single db, all migrations are generated in `prisma/migrations folder`, which is not correct. Ideally they should be created in respective schema folders.

Here:
`prisma/app_db/migrations` was created for `app_db/schema.prisma` with single schema (`prismaSchemaFolder` was disabled)
`prisma/migrations` was created for `master_db/schema.prisma` with multi schema enabled (`prismaSchemaFolder` was enabled)

image

FGoessler:
@AmreeshTyagi I assume you did your testing on Prisma version 6.5.0 or earlier, right?

And do I get your use-case correct when I say:
- You have multiple independent databases.
- For each database you have a dedicated folder (`/prisma/app_db` vs `/prisma/log_db`).
- In each folder you want to have: 1 or more `.prisma` files with the schema and the migrations folder for this database.

And the issue being that once you want to have more than 1 `.prisma` file per folder aka database you have to activate the `prismaSchemaFolder` preview feature and then the migrations folder gets placed one level higher in the `/prisma` directory for all databases. Correct?

If that's all true than the change from this PR should fix this for you with the upcoming 6.6.0 release as we now place the migrations folder next to the `.prisma` file with the datasource block and not next to the given schema path folder. 

(Feel free to try a dev build already - e .g [6.6.0-dev.95](https://www.npmjs.com/package/prisma/v/6.6.0-dev.95))

AmreeshTyagi:
> @AmreeshTyagi I assume you did your testing on Prisma version 6.5.0 or earlier, right?
> 
> And do I get your use-case correct when I say:
> 
> * You have multiple independent databases.
> * For each database you have a dedicated folder (`/prisma/app_db` vs `/prisma/log_db`).
> * In each folder you want to have: 1 or more `.prisma` files with the schema and the migrations folder for this database.
> 
> And the issue being that once you want to have more than 1 `.prisma` file per folder aka database you have to activate the `prismaSchemaFolder` preview feature and then the migrations folder gets placed one level higher in the `/prisma` directory for all databases. Correct?
> 
> If that's all true than the change from this PR should fix this for you with the upcoming 6.6.0 release as we now place the migrations folder next to the `.prisma` file with the datasource block and not next to the given schema path folder.
> 
> (Feel free to try a dev build already - e .g [6.6.0-dev.95](https://www.npmjs.com/package/prisma/v/6.6.0-dev.95))

Yes. I have the same use case and I am using v.6.5.0.  Good to know that it is fixed and planned in next release soon.
Thanks a lot. Will try dev build.

PR #26673, opened 2025-03-18
refactor: ORM-818 remove sending schema data in error reports

FGoessler:
Removing the logic that uploads (anonymized) schema files as part of rust error reports to our error backend. We did not leverage this data in recent times anyways. => Reduces a bunch of file handling complexity related to schema paths.


viktor-ferenczi (approved):
Excellent job!

The most productive days are when we remove more legacy code than we write.

PR #26678, opened 2025-03-18
refactor: move generator out of client

### LINKED ISSUE #25730: `bunx prisma generate` just hangs (bun)
### Bug description

Currently doing some dev work on my laptop running linux to make a quick dev server using prisma for database connections and my friend wanted to help me so i gave him the repo but for some reason `prisma generate` just hangs and when i try it also hangs (we both are using windows)

I also get this from using `DEBUG="*"`
```
prisma:engines binaries to download libquery-engine, schema-engine +900ms
prisma:loadEnv project root found at F:\stuff\pulse\packages\server\package.json +1305ms
prisma:getSchema prismaConfig {
  "packagePath": "F:\\stuff\\pulse\\packages\\server\\package.json"
} +3ms
prisma:tryLoadEnv Environment variables loaded from F:\stuff\pulse\packages\server\.env +1ms
prisma:getSchema prismaConfig {
  "packagePath": "F:\\stuff\\pulse\\packages\\server\\package.json"
} +4ms
prisma:getSchema Checking existence of F:\stuff\pulse\packages\server\schema.prisma +2ms
prisma:getSchema Reading schema from single file F:\stuff\pulse\packages\server\schema.prisma +0ms
prisma:getSchema Checking existence of F:\stuff\pulse\packages\server\prisma\schema.prisma +1ms
prisma:getSchema Checking existence of F:\stuff\pulse\packages\server\prisma\schema.prisma +1ms
prisma:getSchema Reading schema from single file F:\stuff\pulse\packages\server\prisma\schema.prisma +1ms
prisma:getSchema Reading schema from multiple files F:\stuff\pulse\packages\server\prisma\schema +1ms
prisma:getConfig Using getConfig Wasm +2ms
prisma:getConfig config data retrieved without errors in getConfig Wasm +11ms
prisma:getSchema Reading schema from single file F:\stuff\pulse\packages\server\prisma\schema.prisma +3ms
prisma:getConfig Using getConfig Wasm +214ms
prisma:getConfig config data retrieved without errors in getConfig Wasm +4ms
prisma:getDMMF Using getDmmf Wasm +1ms
prisma:getDMMF Using given datamodel +1ms
prisma:getDMMF dmmf data retrieved without errors in getDmmf Wasm +39ms
prisma:generator prismaCLIDir F:\stuff\pulse\node_modules\prisma +8ms
prisma:generator prismaClientDir F:\stuff\pulse\node_modules\@prisma\client +0ms
prisma:generator baseDir F:\stuff\pulse\packages\server\prisma +0ms
prisma:generator typescriptPath F:\stuff\pulse\node_modules\typescript +4ms
prisma:GeneratorProcess prisma:client:generator requiredEngine: libqueryEngine +53ms +144ms                                                                                                                                                                                
prisma:GeneratorProcess {"jsonrpc":"2.0","result":{"manifest":{"defaultOutput":".prisma/client","prettyName":"Prisma Client","requiresEngines":["libqueryEngine"],"version":"5.22.0","requiresEngineVersion":"605197351a3c8bdd595af2d2a9bc3025bca48ea2"}},"id":1} +1ms
```

### How to reproduce

- Run `bunx prisma generate`
- Observe: it just hangs

### Expected behavior

Expect it to complete

### Prisma information

```prisma

// Looking for ways to speed up your queries, or scale easily with your serverless or edge functions?
// Try Prisma Accelerate: https://pris.ly/cli/accelerate-init

generator client {
  previewFeatures = ["omitApi"]
}

//generator drizzle {
//  provider = "drizzle-prisma-generator"
//  output   = "./drizzle" // Where to put generated Drizle tables
//}

datasource db {
  provider = "postgresql"
}

enum StatType {
  View
}

model Stat {
  type    StatType
  special String
  data    Json?

  @@id([type, special])
}

model User {
  id            String    @id @unique
  flags         String[]
  access_token  String
  refresh_token String
  token         String
  public        Boolean   @default(true)
  finished      Boolean   @default(false)
  shadowed      Boolean   @default(false)
  profile       Profile[]
  extra         Json      @default("{}")
  createdAt     DateTime  @default(now())
  updatedAt     DateTime  @updatedAt
}

model Profile {
  id          String    @id @unique
  user        User      @relation(fields: [id], references: [id], onDelete: Cascade)
  pfp         String    @default("https://i.scdn.co/image/ab676161000051747baf6a3e4e70248079e48c5a")
  name        String    @default("dummy_text")
  username    String    @unique @default("dummy_username")
  verified    Boolean   @default(false)
  staff       Boolean   @default(false)
  artist      Boolean   @default(false)
  incomingReq Profile[] @relation("Req")
  outgoingReq Profile[] @relation("Req")
  friends     Profile[] @relation("FriendToFriend")
  friendsOf   Profile[] @relation("FriendToFriend")
  color       String    @default("#EB459E")
  bio         String    @default("im new to dis")
  state       Json?
  createdAt   DateTime  @default(now())
  updatedAt   DateTime  @updatedAt
}

```

```ts
```

### Environment & setup

- OS: Windows 10 Pro 22H2
- Database: PostgreSQL
- Bun version: 1.1.37

### Prisma Version

```
prisma                  : 5.22.0
@prisma/client          : 5.22.0
Architecture            : x64
Node.js                 : v22.6.0
Query Engine (Node-API) : libquery-engine 605197351a3c8bdd595af2d2a9bc3025bca48ea2 (at ..\..\node_modules\@prisma\engines\query_engine-windows.dll.node)
Schema Engine           : schema-engine-cli 605197351a3c8bdd595af2d2a9bc3025bca48ea2 (at ..\..\node_modules\@prisma\engines\schema-engine-windows.exe)  
Schema Wasm             : @prisma/prisma-schema-wasm 5.22.0-44.605197351a3c8bdd595af2d2a9bc3025bca48ea2
Default Engines Hash    : 605197351a3c8bdd595af2d2a9bc3025bca48ea2
Studio                  : 0.503.0
Preview Features        : omitApi
```

### LINKED ISSUE #26560: `prisma generate` hangs on Bun if Node.js is not installed
### Feature Summary

I've been using Prisma for a long time and understand that it relies on Node.js for its internal functionality. Recently, after setting up a fresh OS installation (Ubuntu), I started a new project using Bun as my primary runtime. When I ran bun prisma db push, the process froze at the generation phase.

Since I was already aware that Prisma requires Node.js, I eventually realized the issue, but it took some time to remember. I believe it would be helpful if Prisma provided a clear message indicating that Node.js is required, preventing confusion for users in similar situations.

### Use Cases  
- **Alternative Runtimes (Bun, Deno, etc.)** – Users running Prisma with non-Node.js runtimes may not realize it depends on Node.js.  
- **Fresh OS Installations** – Developers setting up a new system may forget to install Node.js, leading to silent failures.  
- **Faster Debugging** – A clear error message can help users quickly identify missing dependencies instead of troubleshooting a frozen process.  

### Problem Description  
Currently, running Prisma CLI commands (e.g., `prisma db push`) with alternative runtimes like Bun or Deno on a system without Node.js causes the process to freeze at the generation phase without a clear error message. Explicitly notifying users about the Node.js dependency would prevent confusion and improve the developer experience.  

### Improve Error Handling  
Detect when Prisma is executed in a non-Node.js runtime (e.g., Bun, Deno) and check if Node.js is available. If not, immediately display a clear error message instead of allowing the process to freeze.  

### Example Error Message  
If Node.js is missing, Prisma should fail early with a message like:  

### Alternative Solutions

-

### Potential Considerations

_No response_

### Prisma Version

6.3.1

Other

### Additional Context

_No response_

### Pre-Submission Checklist

aqrln:
## Motivation

This PR implements all of the infrastructure necessary for the new ESM-native cross-runtime TypeScript generator.

## Changes

* Implement the new generator contract based on JS objects and not external processes with JSON-RPC. The old style generators are still supported and tested.
* Move Prisma Client generator to a separate package bundled within the CLI. No need to search for `@prisma/client` on the disk to find the executable script anymore. We'll be able to get the full benefit of this in the new generator, as currently we still need to search for the runtime directory in the client package to copy the files over to the generated client, which we won't be doing in the modern generator.
* Remove the auto-installation logic for `@prisma/client`.
* Move the code shared between the client and the generator into `@prisma/client-common`.
* Implement the generators registry infrastructure for the future configuration option in `prisma.config.ts`. Right now it is only prepopulated with the built-in generators but the interface is geared towards supporting the config option.
* Migrate from Jest to Vitest in the files moved over to new packages, and backfill some missing tests.

## Notes

Studio passes a path to the client generator executable it resolves on its own to `getGenerators` in `@prisma/internals` via `providerAliases`. To keep backward compatibility with Studio (and anything else in PDP), we still build a JSON-RPC generator bundle (now implemented as a thin wrapper around `PrismaClientJsGenerator` from `@prisma/client-generator-js`), even though it is not used in any way by the ORM itself, and have `getGenerators` accept the deprecated `providerAliases` option which it automatically converts to a `registry`. All of this can be removed once Studio is updated to make use of the new APIs.

Closes: https://linear.app/prisma-company/issue/ORM-811/move-the-client-generator-out-of-the-client-package
Fixes: https://github.com/prisma/prisma/issues/26560
Fixes: https://github.com/prisma/prisma/issues/25730

/integration


packages/client-generator-js/src/utils/runtimeImport.ts
```diff
@@ -0,0 +1,21 @@
+import * as ts from '@prisma/ts-builders'
+
+// @ts-ignore This doesn't currently work at build time when we don't have the `paths` set in `tsconfig.json`.
+// Adding `@prisma/client` as a dev dependency leads to circular dependency issues.
+// TODO: figure out if we can reuse the runtime types without moving all of them to the common package.
+// eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+type RuntimeExport = keyof typeof import('@prisma/client/runtime/library') & string
```
FGoessler, line 7: ❓ I think I don't really understand what's going on here. Can we chat about it?

FGoessler (approved):
Pretty massive move but looks good overall 👍 🎉

packages/client-common/src/Cache.test.ts
```diff
@@ -1,3 +1,5 @@
+import { expect, test, vi } from 'vitest'
```
jkomyno, line 1: Welcome, vitest!

packages/client-generator-js/src/utils/runtimeImport.ts
aqrln: Sure!

The quick explanation is this type was used as the list of all valid exports of the runtime bundle, so that when we generate the corresponding _import_ in the generated code, the code in the generator that does this is statically type checked.

It was easy when both the generator and the runtime were in the same package, but now doing this would require depending on `@prisma/client` (at least as a dev dependency) in `@prisma/client-generator-js` which leads to a cyclic dependency and cryptic build errors. This could probably be solved by extracting the runtime into a yet another package, or perhaps making it part of `@prisma/client-common`, but it was a bit too much at this point.

In the editor, however, the `tsconfig.json` we use is not the same that we use for building the production packages. It has the `paths` compiler option defined, so that the imports from the other packages are mapped directly to their `src` directories and bypass the `node_modules`. This allows jumping to definitions and finding references across packages in the workspace. The side effect of it, however, is you can import anything from any package without depending on it (from TypeScript's point of view, it will not work at run time though).

I use this here to at least have autocompletion and type checking for the usages of `RuntimeExport` in the editor, even if we fall back to `type RuntimeExport = string` when we build the packages on CI. It's not great, but it's at least something.

Now, the combination of `@ts-ignore` and the intersection with `string` is exactly what implements this build-time fallback. The specific error being silenced with here is "module not found". The `@ts-ignore` makes tsc proceed regardless and infer the type of `import('...')` as `any` and `keyof typeof import('...')` (i.e. `keyof any`) as the widest possible type here, i.e., `PropertyKey`, which is defined as `string | number | symbol`. The intersection with a string is what narrows it down to just `string`.

Finally, the `eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents` directive is necessary because this intersection is redundant when we do have the correct, narrower, type of `keyof typeof import('@prisma/client/runtime/library')` in the editor.

I'll gladly change this if you have a better suggestion, I'm not super proud of this.

PR #26683, opened 2025-03-19
feat: split index.d.ts

FGoessler:


FGoessler:
Only doing this for ESM. See https://github.com/prisma/prisma/pull/26727

PR #26700, opened 2025-03-21
feat(cli): add MCP server

sorenbs:
/integration


packages/cli/src/Init.ts
```diff
@@ -344,26 +352,49 @@ export class Init implements Command {
         console.log(`Successfully authenticated as ${bold(authenticationResult.email)}.`)
       }
 
+      if (args['--generate']) {
```
sampolahtinen, line 355: What is the intented usage of `--generate` flag? I could not see it being used by the MCP command.

packages/cli/src/Init.ts
```diff
@@ -344,26 +352,49 @@ export class Init implements Command {
...
+        const spinner = ora(`Generating a Prisma Schema based on your description ${bold(args['--generate'])} ...`).start()
+
+        try {
+          ({ generatedSchema, generatedName } = await (await fetch(`https://prisma-generate-server.prisma.workers.dev/`, {
```
sampolahtinen, line 359: What does this worker do? Is it possible to peek into the underlying code?

packages/cli/src/Init.ts
```diff
@@ -232,6 +231,10 @@ export class Init implements Command {
       '--output': String,
       '--with-model': Boolean,
       '--db': Boolean,
+      '--region': String,
+      '--name': String,
+      '--non-interactive': Boolean,
+      '--generate': String,
```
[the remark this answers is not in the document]
sorenbs: I think they can just silently do nothing. They are only used by the MCP server, and I am not documenting them at this time. If we choose to document them, we should make sure they are ergonomic in all cases.

packages/cli/src/Init.ts
sorenbs: I am going to rename it to `--prompt` with an alias for `--vibe` as we discussed. This flag will use AI to generate a schema based on the provided prompt:

```
npx prisma init --prompt "A todo app specifically for tracking birds"
```

packages/cli/src/Init.ts
```diff
@@ -344,26 +356,51 @@ export class Init implements Command {
...
+
+        try {
+          ;({ generatedSchema, generatedName } = (await (
+            await fetch(`https://prisma-generate-server.prisma.workers.dev/`, {
+              method: 'POST',
+              body: JSON.stringify({
+                description: prompt,
+              }),
+            })
+          ).json()) as { generatedSchema: string; generatedName: string })
```
aqrln, line 371: Could we validate it using Effect Schema or zod instead of just asserting the type? We generally use Effect Schema in our code, so it's probably better to stick to using it here as well for consistency, and keep zod isolated to `@modelcontextprotocol/sdk` integration.

packages/cli/src/Init.ts
```diff
@@ -326,13 +332,19 @@ export class Init implements Command {
     const outputDir = process.cwd()
     const prismaFolder = path.join(outputDir, 'prisma')
 
+    let generatedSchema,
```
aqrln, line 335: I believe the inferred type of `generatedSchema` might be `any` here and it might need an explicit type annotation (`string | undefined`). We unfortunately currently have `noImplicitAny` disabled in tsconfig for some mysterious reason, which would have caught this — something for us to fix.

packages/cli/src/MCP.ts
```diff
@@ -0,0 +1,183 @@
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js'
+import type { PrismaConfigInternal } from '@prisma/config'
+import { Command, link } from '@prisma/internals'
+import { spawn } from 'child_process'
+import { z } from 'zod'
+
+import { createHelp } from './platform/_lib/help'
+
+function spawnAsPromise(
```
aqrln, line 10: You can use `execa` (we already depend on it, albeit an old version) so you don't have to re-implement this manually

packages/cli/src/MCP.ts
```diff
@@ -0,0 +1,183 @@
...
+  })
+
+  public async parse(_argv: string[], _config: PrismaConfigInternal): Promise<string | Error> {
+    const server = new McpServer({
+      name: 'Prisma',
+      version: '1.0.0',
+    })
+
+    server.tool(
+      'migrate-status',
```
aqrln, line 61: Should these be prefixed with `prisma-` or the tools from different servers are inherently namespaced already?

aqrln (approved):

jkomyno (approved):

PR #26709, opened 2025-03-21
feat(adapter-better-sqlite3): create `@prisma/adapter-better-sqlite3`

jkomyno:
This PR:
- Contributes to [ORM-731](https://linear.app/prisma-company/issue/ORM-731/new-driver-adapter-better-sqlite3-for-provider-=-sqlite)
- Creates a new `@prisma/adapter-better-sqlite3` Driver Adapter
- Adds `--adapter js_better_sqlite3` to the functional client test suite

Next up:
- use `@prisma/adapter-better-sqlite3` in `prisma-engines`


.github/workflows/scripts/detect-jobs-to-run.js
```diff
@@ -1,29 +1,11 @@
...
-  }
-
-  stdin.setEncoding('utf8')
-
-  for await (const chunk of stdin) {
-    result += chunk
-  }
-
-  return result
-}
```
FGoessler, line 22: ℹ️  We got some `detectOpenHandles` issues due to the get-stdin package and hence switched to the more modern node native approach. Also changing this code that was copied out of that lib for consistency.

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -0,0 +1,244 @@
...
+    const options: TransactionOptions = {
+      usePhantomQuery: false,
+    }
+
+    const tag = '[js::startTransaction]'
+    debug('%s options: %O', tag, options)
+
+    const release = await this[LOCK_TAG].acquire()
+
+    this.client.prepare('BEGIN').run()
```
jacek-prisma, line 207: I think we should be handling the potential exception here through `onError`

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -0,0 +1,244 @@
...
+    const errCodeNum = error.code && ERROR_CODE_STRING_TO_CODE_NUM[error.code]
+    if (errCodeNum) {
+      throw new DriverAdapterError({
+        kind: 'sqlite',
+        extendedCode: errCodeNum,
+        message: error.message,
+      })
+    }
+    throw error
+  }
```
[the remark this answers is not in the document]
FGoessler: Ah those are your recent additions for the error mapping? I'll have a look into it!

packages/adapter-better-sqlite3/src/better-sqlite3.ts
FGoessler: Added ✅

jacek-prisma (approved):

PR #26719, opened 2025-03-24
fix(ci): make `/engine-branch` command set up custom WASM modules

aqrln:
The `/engine-branch` command only built and set the environment variables for the native query engine and schema engine but ignored the WASM packages. Back when it was introduced, this limitation only affected prisma-schema-wasm and was only a problem for changes in DMMF, but as more of our Rust code is moved to WebAssembly, this became a lot bigger issue.

This PR makes this command also build and use prisma-schema-wasm, query-engine-wasm, schema-engine-wasm and query compiler from the specified engines branch.

Closes: https://linear.app/prisma-company/issue/ORM-826/fix-engine-branch-command-not-building-wasm-modules

/engine-branch main

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -49,6 +49,11 @@ runs:
           target/release/libquery_engine.so
           target/release/libquery_engine.dylib
           target/release/query_engine.dll
+          target/prisma-schema-wasm
+          query-engine/query-engine-wasm/pkg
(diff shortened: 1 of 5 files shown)


viktor-ferenczi (dismissed):

FGoessler (approved):
Approved but seems like we gotta exclude some build/target dirs from eslint rules 😬

viktor-ferenczi (approved):

PR #26720, opened 2025-03-24
feat(qc): Transactional graphs

viktor-ferenczi:
- Interpreting transactions in the query plan

Implements [ORM-742](https://linear.app/prisma-company/issue/ORM-742/run-transactional-graphs-in-transactions)

[Engine PR](https://github.com/prisma/prisma-engines/pull/5291)

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
+++ b/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
@@ -9,39 +9,39 @@ import { PrismaObject, ScopeBindings, Value } from './scope'
 import { serialize } from './serialize'
 
 export type QueryInterpreterOptions = {
-  queryable: SqlQueryable
   transactionManager: TransactionManager
+  allowTransaction: boolean
   placeholderValues: Record<string, unknown>
   onQuery?: (event: QueryEvent) => void
 }
 
 export class QueryInterpreter {
-  #queryable: SqlQueryable
   #transactionManager: TransactionManager
+  #allowTransaction: boolean
   #placeholderValues: Record<string, unknown>
   #onQuery?: (event: QueryEvent) => void
   readonly #generators: GeneratorRegistry = new GeneratorRegistry()
 
-  constructor({ queryable, transactionManager, placeholderValues, onQuery }: QueryInterpreterOptions) {
-    this.#queryable = queryable
+  constructor({ transactionManager, allowTransaction, placeholderValues, onQuery }: QueryInterpreterOptions) {
+    this.#allowTransaction = allowTransaction
     this.#transactionManager = transactionManager
     this.#placeholderValues = placeholderValues
     this.#onQuery = onQuery
   }
 
-  async run(queryPlan: QueryPlanNode): Promise<unknown> {
-    return this.interpretNode(queryPlan, this.#placeholderValues, this.#generators.snapshot(), this.#queryable)
+  async run(queryPlan: QueryPlanNode, queryable: SqlQueryable): Promise<unknown> {
+    return this.interpretNode(queryPlan, queryable, this.#placeholderValues, this.#generators.snapshot())
   }
 
   private async interpretNode(
     node: QueryPlanNode,
+    queryable: SqlQueryable,
     scope: ScopeBindings,
     generators: GeneratorRegistrySnapshot,
-    queryable: SqlQueryable,
   ): Promise<Value> {
     switch (node.type) {
       case 'seq': {
-        const results = await Promise.all(node.args.map((arg) => this.interpretNode(arg, scope, generators, queryable)))
+        const results = await Promise.all(node.args.map((arg) => this.interpretNode(arg, queryable, scope, generators)))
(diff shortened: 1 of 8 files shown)


packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
```diff
@@ -39,26 +39,35 @@ export class TransactionManager {
...
-  async startTransaction(options: Options): Promise<TransactionInfo> {
-    const validatedOptions = this.validateOptions(options)
+  async startTransaction(options?: Options): Promise<TransactionInfo> {
+    const validatedOptions = options !== undefined ? this.validateOptions(options) : this.transactionOptions
+    console.log(
+      `!!! TransactionManager.startTransaction: options = ${
+        options === undefined ? 'undefined' : JSON.stringify(options)
+      }`,
+    )
+    console.log(`!!! TransactionManager.startTransaction: validatedOptions = ${JSON.stringify(validatedOptions)}`)
```
jacek-prisma, line 57: I don't think we always want to log this, maybe a `Debug` log?

viktor-ferenczi: That's for temporary debugging of my PR, because transactions fail to start with a timeout. Want to know the options used there. It will not stay in the code, certainly.

packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
```diff
@@ -39,26 +39,35 @@ export class TransactionManager {
   // Used to provide better error messages than a generic "transaction not found".
   private closedTransactions: TransactionWrapper[] = []
   private readonly driverAdapter: SqlDriverAdapter
+  private readonly transactionOptions: Options
```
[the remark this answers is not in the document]
viktor-ferenczi: Default transaction options are received from `EngineConfig`, because they are not available from the transaction node in `QueryInterpreter`.

In order to support per-transaction control of these options we would need to pass these options all the way from the generated client. If we decide to do this, then I suggest making this a separate task to avoid scope creep here.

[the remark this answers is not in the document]
jacek-prisma: alright, I think that's okay, I didn't know we have `transactionOptions` in the config

packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
viktor-ferenczi: Removed the logging, since the source of the low timeout values was found in the `prisma-engines` repo by Alexey.

.github/workflows/test-template.yml
```diff
@@ -316,7 +316,7 @@ jobs:
           engine-hash: ${{ inputs.engineHash }}
           skip-tsc: false
 
-      - run: pnpm run test:functional --adapter ${{ matrix.flavor }} --shard ${{ matrix.shard }} --client-runtime client --preview-features ${{ matrix.previewFeatures }} --engine-type client --runInBand --json --outputFile ./tests/functional/results.json # TODO: `--json --outputFile` can be dropped once we remove the checkAgainstKnownFailures step below
+      - run: pnpm run test:functional --adapter ${{ matrix.flavor }} --shard ${{ matrix.shard }} --client-runtime client --preview-features ${{ matrix.previewFeatures }} --engine-type client --detectOpenHandles --runInBand --json --outputFile ./tests/functional/results.json # TODO: `--json --outputFile` can be dropped once we remove the checkAgainstKnownFailures step below
```
viktor-ferenczi, line 319: This change will be removed by rebasing, it has been separated to another PR.

packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
```diff
@@ -2,41 +2,46 @@ import { SqlQuery, SqlQueryable } from '@prisma/driver-adapter-utils'
...
+import { type TransactionManager } from '../transactionManager/TransactionManager'
 import { GeneratorRegistry, GeneratorRegistrySnapshot } from './generators'
 import { renderQuery } from './renderQuery'
 import { PrismaObject, ScopeBindings, Value } from './scope'
 import { serialize } from './serialize'
 
 export type QueryInterpreterOptions = {
-  queryable: SqlQueryable
+  transactionManager: TransactionManager
+  allowTransaction: boolean
```
jacek-prisma, line 13:
```suggestion
  transactionManager: TransactionManager
  allowTransaction: boolean
```
maybe it makes sense to not pass a transaction manager at all if we don't allow transactions, I don't like having a transaction manager here when it should not be used, so it could maybe be optional

it might even be worth making it explicit via a type 
```ts
type InterpreterTransactionManager
  = { enabled: true,  manager: TransactionManager }
  | { enabled: false }
```

viktor-ferenczi: Good point, let's make invalid state unrepresentable.

jacek-prisma (approved):

PR #26727, opened 2025-03-25
feat: ORM-596 file splitting in ESM generator

FGoessler:
Based on previous work https://github.com/prisma/prisma/pull/26683 this PR implements generated file splitting along model boundaries.

Example File Tree based off our sandbox:
Screenshot 2025-04-02 at 23 32 59


packages/client-generator-ts/src/TSClient/file-generators/CommonInputTypesFile.ts
```diff
@@ -0,0 +1,49 @@
...
+        const innerName = needsGeneric ? `${inputType.name}Base<$PrismaModel>` : `${inputType.name}Base`
+        const typeName = needsGeneric ? `${inputType.name}<$PrismaModel = never>` : inputType.name
+        // This generates types for JsonFilter to prevent the usage of 'path' without another parameter
+        const baseName = `Required<${innerName}>`
+        return `export type ${typeName} = 
+| Prisma.PatchUndefined<
+    Prisma.Either<${baseName}, Exclude<keyof ${baseName}, 'path'>>,
+    ${baseName}
+  >
+| Prisma.OptionalFlat<Omit<${baseName}, 'path'>>
```
aqrln, line 35: This seems to be the same as in `ModelFile.ts`, should we extract this to a function?

packages/client-generator-ts/src/TSClient/file-generators/EnumsFile.ts
```diff
@@ -0,0 +1,23 @@
...
+`
+
+export function createEnumsFile(context: GenerateContext): string {
+  const modelEnums: string[] = []
+  for (const datamodelEnum of context.dmmf.datamodel.enums) {
+    modelEnums.push(new Enum(datamodelEnumToSchemaEnum(datamodelEnum), false).toTS())
+  }
+
+  // Cannot have a fully empty file, so we export a constant to indicate that there are no enums
+  if (modelEnums.length === 0) return 'export const __NO_ENUMS__ = true'
```
aqrln, line 20: Do we actually use this constant anywhere? Should this be just `export {}` instead?

packages/client-generator-ts/src/TSClient/file-generators/PrismaNamespaceFile.ts
```diff
@@ -0,0 +1,266 @@
...
+import type { TSClientOptions } from '../TSClient'
+import { clientTypeMapDefinition } from '../TypeMap'
+
+const jsDocHeader = `/**
+ * WARNING: This is an internal file that is subject to change!
+ * 
+ * 🛑 Under no circumstances should you import this file directly! 🛑
+ * 
+ * All exports from this file are wrapped under a \`Prisma\` namespace object in the client.ts file.
+ * While this enables partial backward compatibility, it is not part of the stable public API.
```
aqrln, line 18: Many of the types here are definitely part of the public API. If we don't want users to use the `Prisma` namespace at all, we'll need to put them to separate files as well.

aqrln: (can happen in a follow up PR)

packages/client-generator-ts/src/generateClient.ts
```diff
@@ -132,10 +132,7 @@ export function buildClient({
 
   const client = new TSClient(clientOptions)
 
-  // we store the generated contents here
-  let fileMap: FileMap = {}
-  fileMap[outputName('client')] = client.toTS()
-  fileMap[outputName('index')] = `export * from '${importName('./client')}'`
+  let fileMap: FileMap = client.generateClientFiles()
```
aqrln, line 135:
```suggestion
  let fileMap = client.generateClientFiles()
```

aqrln (commented):
Amazing work 🎉

packages/client-generator-ts/src/TSClient/file-generators/PrismaNamespaceFile.ts
FGoessler: Yes we need further work if we want to double down on this. But at the moment not really a priority. Compile time speed, ESM compatibility and ease of adoption for our users are the prio at the moment.

packages/client-generator-ts/src/TSClient/file-generators/EnumsFile.ts
FGoessler: Updated it to use a comment and an `export {}`

packages/client-generator-ts/src/TSClient/file-generators/CommonInputTypesFile.ts
FGoessler: Correct. Originally thought those two might evolve differently over my refactoring. Not the case though. Moved this logic into the `InputType` class. I checked other usages of this class and seems to be safe to have that codepath in there always.

aqrln (approved):
Looks great

PR #26729, opened 2025-03-25
feat(generator-ts): require explicit output directory

aqrln:
Refs: https://linear.app/prisma-company/issue/ORM-689/implement-new-typescript-client-generator

### DIFF
--- a/packages/client-generator-ts/src/generator.ts
+++ b/packages/client-generator-ts/src/generator.ts
@@ -4,6 +4,7 @@ import Debug from '@prisma/debug'
 import { enginesVersion } from '@prisma/engines-version'
 import { EngineType, Generator, GeneratorConfig, GeneratorManifest, GeneratorOptions } from '@prisma/generator'
 import { ClientEngineType, getClientEngineType, parseEnvValue } from '@prisma/internals'
+import { bold, dim, green } from 'kleur/colors'
 import { match } from 'ts-pattern'
 
 import { version as clientVersion } from '../package.json'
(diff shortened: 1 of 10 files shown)


packages/client-generator-ts/src/generator.ts
```diff
@@ -13,23 +14,35 @@ import { resolvePrismaClient } from './resolvePrismaClient'
...
 type PrismaClientTsGeneratorOptions = {
-  shouldResolvePrismaClient?: boolean
   runtimePath?: string
 }
 
+const missingOutputErrorMessage = `An output path is required for the \`prisma-client-ts\` generator. Please provide an output path in your schema file:
+
+${dim(`generator client {
+  provider = "prisma-client-ts"`)}
+${green('  output   = "../src/generated"')}
```
viktor-ferenczi, line 24: It may be worth mentioning what directory the output path is relative to, since you used `..` in the example directory path. I guess this is relative to the schema file's directory.

packages/client-generator-ts/tests/generator.test.ts
```diff
@@ -230,27 +243,24 @@ describe('generator', () => {
...
-      8 output   = "./__fixture__/.prisma/client"
+      await generator.generate()
+    }).rejects.toThrowErrorMatchingInlineSnapshot(`
+      [Error: An output path is required for the \`prisma-client-ts\` generator. Please provide an output path in your schema file:
 
-      You won't need to change your imports.
-      Imports from \`@prisma/client\` will be automatically forwarded to \`.prisma/client\`]
+      generator client {
+        provider = "prisma-client-ts"
+        output   = "../src/generated"
```
viktor-ferenczi, line 261: Same as above.

viktor-ferenczi (approved):

packages/client-generator-ts/src/generator.ts
aqrln: great point

aqrln: done
image

PR #26732, opened 2025-03-25
feat(generator-ts): don't generate package.json

aqrln:
The client outputted by the new generator is a collection of source files and not a package in `node_modules`, and it doesn't need `package.json`. If users need a `package.json` (e.g. they want the generated client to be a package in its own right), they need to write that `package.json` themselves: we can't generate one because we don't know how they want to build the package and where the transpiled JavaScript sources or the bundle is in the package. We never delete any files in the output directory (other than the engine binaries when regenerating with the `--no-engine` CLI flag) and users are free to mix them with their own files as needed.

Refs: https://linear.app/prisma-company/issue/ORM-689/implement-new-typescript-client-generator

### DIFF
--- a/packages/client-generator-ts/src/generateClient.ts
+++ b/packages/client-generator-ts/src/generateClient.ts
@@ -1,4 +1,6 @@
-import Debug from '@prisma/debug'
+import fs from 'node:fs/promises'
+import path from 'node:path'
+
 import type * as DMMF from '@prisma/dmmf'
 import { overwriteFile } from '@prisma/fetch-engine'
 import type { ActiveConnectorType, BinaryPaths, DataSource, GeneratorConfig, SqlQueryOutput } from '@prisma/generator'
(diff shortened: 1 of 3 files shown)


packages/client-generator-ts/src/generateClient.ts
```diff
@@ -402,7 +289,7 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
   }
 
   if (!copyEngine) {
-    await deleteOutputDir(outputDir)
+    await deleteEnginesInOutputDir(outputDir)
```
FGoessler, line 292: 💭 If we really only delete the engines files but no other files that we generate we might imo run into issues. E.g. assume I delete a model, then I expect the model file in my generated output to be gone. With this change this won't be the case and as a user I might not even know that I need to delete it manually.

Maybe the model files are the only such case? Then we could just always wipe the models directory on generate but maybe there are other cases, too? 🤔

aqrln: Good point. Right now this would be a problem even without changing this function — this code path is only reached when passing the `--no-engine` CLI flag to `prisma generate`. There are two options I can see:
- Always delete the output directory: simpler but might be unexpected to users and lead to losing their files if they don't expect this behaviour
- Generate a file with the list of generated files which we can read on next generate, so we only remove those files

packages/client-generator-ts/src/generateClient.ts
```diff
@@ -598,39 +483,27 @@ function getNodeRuntimeName(engineType: ClientEngineType) {
...
-  const hash = createHash('sha256')
-  hash.write(datamodel)
-  return `${GENERATED_PACKAGE_NAME_PREFIX}${hash.digest().toString('hex')}`
-}
+    // TODO: replace with `client.ts`
+    if (!files.includes('client.d.ts')) {
+      // Make sure users don't accidentally wipe their source code or home directory.
+      throw new Error(
+        `${outputDir} exists and is not empty but doesn't look like a generated Prisma Client. ` +
+          'Please check your output path and remove the existing directory if you indeed want to generate the Prisma Client in that location.',
```
FGoessler, line 497: ❤️

FGoessler (approved):

viktor-ferenczi (changes requested):

packages/client-generator-ts/src/generateClient.ts
```diff
@@ -598,39 +481,27 @@ function getNodeRuntimeName(engineType: ClientEngineType) {
...
+      )
+    }
 
-const GENERATED_PACKAGE_NAME_PREFIX = 'prisma-client-'
+    await fs.rm(outputDir, { recursive: true, force: true })
```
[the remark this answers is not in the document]
aqrln: done

viktor-ferenczi (approved):

PR #26737, opened 2025-03-26
chore: ORM-828 ts generator test setup

FGoessler:
This PR makes our client functional test suite runable with the existing JS generator as well as the new ESM/TS generator.

- Update the tests to use a custom output path (ESM generator requires it)
- Add new matrix option 

> [!NOTE]
> Review note: Reviewing commit by commit might be useful here. The initial commit changes hundreds of files just to perform some path changes across all test files.

### DIFF
--- a/packages/client/tests/functional/_utils/providers.ts
+++ b/packages/client/tests/functional/_utils/providers.ts
@@ -23,6 +23,8 @@ export enum RelationModes {
   PRISMA = 'prisma',
 }
 
+export type GeneratorTypes = 'prisma-client-js' | 'prisma-client-ts'
+
 export const adaptersForProvider = {
   [Providers.POSTGRESQL]: [AdapterProviders.JS_PG, AdapterProviders.JS_NEON],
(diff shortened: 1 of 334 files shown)


packages/client/tests/functional/_utils/providers.ts
```diff
@@ -23,6 +23,8 @@ export enum RelationModes {
   PRISMA = 'prisma',
 }
 
+export type GeneratorTypes = 'prisma-client-js' | 'prisma-client-ts'
```
aqrln, line 26: Fine to keep it as is for consistency with other types in this file, but they all should be changed to singular

aqrln (approved):

packages/client/tests/functional/_utils/providers.ts
FGoessler: Yeah we could also all change them away from being ts enums I think? 🤔

FGoessler: I'll put it on my cleanup list for during the flight next week 😅

PR #26746, opened 2025-03-27
Fix: vite server crashing 

brrock:
Fixes
https://github.com/wakujs/waku/issues/1175
https://github.com/vitejs/vite/issues/19036

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -72,6 +72,12 @@
       "import": "./extension.js",
       "default": "./extension.js"
     },
+    "./browser": {
+      "types": "./index.d.ts",
+      "require": "./index-browser.js",
+      "import": "./index-browser.js",
+      "default": "./index-browser.js"
+    },
     "./index-browser": {
       "types": "./index.d.ts",
       "require": "./index-browser.js",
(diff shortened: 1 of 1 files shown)


packages/client/package.json
```diff
@@ -72,6 +72,12 @@
       "import": "./extension.js",
       "default": "./extension.js"
     },
+    "./browser": {
+      "types": "./index.d.ts",
+      "require": "./index-browser.js",
+      "import": "./index-browser.js",
+      "default": "./index-browser.js"
+    },
```
aqrln, line 80: Where do we reference this entrypoint?

packages/client/package.json
```diff
@@ -124,7 +130,7 @@
         "node": "./sql.mjs",
         "default": "./sql.mjs"
       },
-      "default": "./sql.js"
+      "default": "./default.js"
```
aqrln, line 133: This is incorrect, the `/sql` bundle is not a client

aqrln (changes requested):
Sorry for the delay! You PR description is rather minimal, could you please explain why these changes are necessary and what exactly they fix? Thank you.

PR #26747, opened 2025-03-27
test(async): Detecting stuck async code in functional tests

viktor-ferenczi:
- Added `--detectOpenHandles` test option to identify stuck async code

Functional test runs now provide useful information on any async code being stuck at the end of test suite execution. This also terminates the test right away instead of waiting until timeout.

For example:
```
Jest has detected the following 1 open handle potentially keeping Jest from exiting. 

      45 |   stdin.setEncoding("utf8");
      46 |   for await (const chunk of stdin) {
    > 47 |     result += chunk;
         |   ^
      48 |   }
      49 |   return result;
      50 | }
```

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -316,7 +316,7 @@ jobs:
           engine-hash: ${{ inputs.engineHash }}
           skip-tsc: false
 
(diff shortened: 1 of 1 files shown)


FGoessler (approved):
Are you intentionally only adding this to the query compiler tests? 🤔

viktor-ferenczi:
> Are you intentionally only adding this to the query compiler tests? 🤔

Yes, because I had such issues only with the QC tests so far.
Would you suggest adding it to all of our pnpm tests on CI?

PR #26761, opened 2025-03-27
feat: query compiler batching

jacek-prisma:
[ORM-793](https://linear.app/prisma-company/issue/ORM-794/native-batching-support-in-qc)

Depends on https://github.com/prisma/prisma-engines/pull/5300.

I've had to translate some Rust code into Typescript (from `query-engine/request-handlers/src/handler.rs`). 
This behavior is covered in existing tests that I've run manually to confirm it works like `batch-find-unique`.


packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -294,18 +300,19 @@ export class ClientEngine implements Engine<undefined> {
     queries: JsonQuery[],
     { transaction, traceparent: _traceparent }: RequestBatchOptions<undefined>,
   ): Promise<BatchQueryEngineResult<T>[]> {
-    this.lastStartedQuery = JSON.stringify(queries)
+    if (queries.length === 0) {
+      return []
+    }
+    const firstAction = queries[0].action
```
viktor-ferenczi, line 306: Is it guaranteed that each query has the same `action` in the batch?

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -318,23 +325,38 @@ export class ClientEngine implements Engine<undefined> {
...
+      })
+      const queryable = transactionManager.getTransaction(txInfo, firstAction)
+
+      let results: BatchQueryEngineResult<unknown>[] = []
+      switch (response.type) {
+        case 'multi': {
+          results = await Promise.all(
+            response.plans.map(async (plan) => {
+              const rows = await interpreter.run(plan as QueryPlanNode, queryable)
+              return { data: { [firstAction]: rows } }
```
viktor-ferenczi, line 343: Related to the question above.

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -345,6 +367,44 @@ export class ClientEngine implements Engine<undefined> {
...
+   * Converts the result of a compacted query back to result objects analogous to what queries
+   * would return when executed individually.
+   */
+  #convertCompactedRows(
+    rows: {}[],
+    response: CompactedBatchResponse,
+    action: string,
+  ): BatchQueryEngineResult<unknown>[] {
+    // a list of objects that contain the keys of every row
+    const keysPerRow = rows.map((item) => response.keys.reduce((acc, key) => ({ [key]: item[key], ...acc }), {}))
```
viktor-ferenczi, line 381: It may pose O(N^2) time complexity, where N is the number of rows. I suggest updating the same object instead.

```
const keysPerRow = rows.map((item) => {
  const obj: Record<string, any> = {};
  for (const key of response.keys) {
    obj[key] = item[key];
  }
  return obj;
});
```

viktor-ferenczi (changes requested):

packages/client/src/runtime/core/engines/client/ClientEngine.ts
jacek-prisma: this has the same time complexity as my code - O(M*N) (it just avoids re-creating objects)

packages/client/src/runtime/core/engines/client/ClientEngine.ts
jacek-prisma: added a check

packages/client/src/runtime/core/engines/client/ClientEngine.ts

viktor-ferenczi (approved):

PR #26767, opened 2025-03-27
feat(generator-ts): introduce runtime target switch

aqrln:
- Introduce `runtime` generator options in `prisma-client-ts` generator.

- Only generate client for a single target. Also make WASM imports on edge work without `package.json` and import subpaths, the correct platform-specific loading code is now injected directly into the generated client without trampoline files.


.github/workflows/test-template.yml
```diff
@@ -59,7 +59,7 @@ jobs:
   #
   lint:
     name: Lint
-    timeout-minutes: 7
```
FGoessler, line 62: Does our lint take that long? 🫣

packages/client-generator-ts/src/generateClient.ts
```diff
@@ -82,10 +80,14 @@ export function buildClient({
   copyEngine,
   envPaths,
   typedSql,
+  target = 'nodejs',
```
FGoessler, line 83: 💡 Would it be easily possible to have the default value only at the "entry level". Right now it seems like we have multiple functions with an optional `target` param that redefine the default. 🤔

packages/client-generator-ts/src/utils/buildGetWasmModule.ts
```diff
@@ -0,0 +1,68 @@
...
+  const wasmPathBase = `${runtimeBase}/query_compiler_bg.${activeProvider}`
+  const wasmBindingsPath = `${wasmPathBase}.mjs`
+  const wasmModulePath = `${wasmPathBase}.wasm`
+
+  const buildNodeLoader = match(runtimeName)
+    .with('library', () => component === 'engine' && !!process.env.PRISMA_CLIENT_FORCE_WASM)
+    .with('client', () => component === 'compiler')
+    .otherwise(() => false)
+
+  // We're missing an edge bundle for client engine right now
```
FGoessler, line 34:
```suggestion
  // TODO: We're missing an edge bundle for client engine right now
```
Will we need one though? Or can we "rest" on the fact that we have the query compiler + driver adapters soon enough and it's all compatible out of the box? 🤔

.github/workflows/test-template.yml
aqrln: Yeah, just the build itself now takes that long, it doesn't even get to the linting 😅

packages/client-generator-ts/src/utils/buildGetWasmModule.ts
aqrln: Yes, we still need one because we can't use the Node.js APIs and filesystem there

FGoessler (approved):

PR #26776, opened 2025-03-28
feat: complete libsql client adapter

jacek-prisma:
[ORM-694](https://linear.app/prisma-company/issue/ORM-694/libsqlclient-create-migration-adapter)

Primarily adding tests and exposing a separate web client

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -987,6 +987,33 @@ jobs:
         working-directory: packages/generator
 
   #
+  # Driver Adapter Unit Tests
+  #
+  driver-adapter-unit-tests:
+    name: Driver Adapter Unit Tests
+    timeout-minutes: 10
+    runs-on: ${{ matrix.os }}
+    if: contains(inputs.jobsToRun, '-all-')
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        node: [18, 20, 22, 23]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install & build
+        uses: ./.github/actions/setup
+        with:
+          node-version: ${{ matrix.node }}
+          engine-hash: ${{ inputs.engineHash }}
+          skip-tsc: true
+
+      - run: pnpm run test
+        name: 'adapter-libsql'
+        working-directory: packages/adapter-libsql
+
+  #
   # Run all tests on macOS and Windows.
   #
   # Unlike the other jobs, this job doesn't use Docker (and thus skips some
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -998,7 +998,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        node: [18, 20, 22, 23]
+        node: [18, 23]
     steps:
       - uses: actions/checkout@v4
 
--- a/packages/adapter-libsql/package.json
+++ b/packages/adapter-libsql/package.json
@@ -46,15 +46,12 @@
   "license": "Apache-2.0",
   "sideEffects": false,
   "dependencies": {
+    "@libsql/client": "^0.3.5 || ^0.4.0 || ^0.5.0 || ^0.6.0 || ^0.7.0 || ^0.8.0",
     "@prisma/driver-adapter-utils": "workspace:*",
     "async-mutex": "0.5.0"
   },
   "devDependencies": {
-    "@libsql/client": "0.8.0",
     "jest": "29.7.0",
     "jest-junit": "16.0.0"
-  },
-  "peerDependencies": {
-    "@libsql/client": "^0.3.5 || ^0.4.0 || ^0.5.0 || ^0.6.0 || ^0.7.0 || ^0.8.0"
   }
 }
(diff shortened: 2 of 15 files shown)


packages/adapter-libsql/package.json
```diff
@@ -39,7 +50,9 @@
     "async-mutex": "0.5.0"
   },
   "devDependencies": {
-    "@libsql/client": "0.8.0"
+    "@libsql/client": "0.8.0",
+    "jest": "29.7.0",
+    "jest-junit": "16.0.0"
   },
   "peerDependencies": {
     "@libsql/client": "^0.3.5 || ^0.4.0 || ^0.5.0 || ^0.6.0 || ^0.7.0 || ^0.8.0"
```
jkomyno, line 58: Do we even still need `peerDependencies`?

.github/workflows/test-template.yml
```diff
@@ -986,6 +986,33 @@ jobs:
...
+  driver-adapter-unit-tests:
+    name: Driver Adapter Unit Tests
+    timeout-minutes: 10
+    runs-on: ${{ matrix.os }}
+    if: contains(inputs.jobsToRun, '-all-')
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        node: [18, 20, 22, 23]
```
viktor-ferenczi, line 1001: Are we sure we really need to run this test on all those Ubuntu versions? 
Is this for consistency with the other adapters?

packages/adapter-libsql/package.json
viktor-ferenczi, line 1: Fill `"author"` with your name :)

packages/adapter-libsql/src/libsql.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    expect(tx.options.usePhantomQuery).toBe(true)
+
+    await expect(
+      tx.executeRaw({
+        sql: "INSERT INTO test (id, name) VALUES (20, 'Jane')",
+        args: [],
+        argTypes: [],
+      }),
+    ).resolves.toBe(1)
+    await expect(tx.rollback()).resolves.toBeUndefined()
```
viktor-ferenczi, line 155: The test case does not verify that no rows remain in the table, e.g. the rollback really happened.

viktor-ferenczi (changes requested):

packages/adapter-libsql/src/libsql.test.ts
jacek-prisma: I can't test that with an in-memory DB due to a bug in the library: https://github.com/tursodatabase/libsql-client-ts/issues/229

I covered that we do the rollback at the bottom of the file in `rollback rolls back the underlying transaction`

jacek-prisma: we could cover it thought if we bump the library because it was fixed in `0.9`

packages/adapter-libsql/package.json
jacek-prisma: the initial code wasn't written by me!

.github/workflows/test-template.yml
jacek-prisma: I think it makes sense to only run this on oldest and latest

packages/adapter-libsql/package.json
jacek-prisma: true, the driver should probably be converted to an actual dependency (in all the other adapters as well), I'll change it

packages/adapter-libsql/src/libsql.test.ts
jacek-prisma: I've just tried with 0.9 but the in-memory transaction behavior there is buggy as well, I think the test at the end of the file will be sufficient

packages/adapter-libsql/package.json
```diff
@@ -2,18 +2,28 @@
...
+    },
+    "./web": {
+      "require": {
+        "types": "./dist/index-web.d.ts",
+        "default": "./dist/index-web.js"
+      },
+      "import": {
+        "types": "./dist/index-web.d.mts",
+        "default": "./dist/index-web.mjs"
       }
```
jkomyno, line 27: Could we use explicit JS runtime keys like shown in [here](https://jkomyno-armada-js-2024.vercel.app/25), and pair this with an integration release + ecosystem test to see whether the correct import is selected in runtimes like Cloudflare Workers and Vercel Edge? We can have a manual test right now, and an ecosystem test next sprint, but we should check this before merging it.

jacek-prisma: Do you mean it should automatically pick the bundle based on runtime? I don't know if that's better because web and node implementations aren't interchangeable, so the `web` implementation is basically an opt-in.

jacek-prisma: As for a test, I agree it makes sense to have an ecosystem test (and we'll have to update the turso lambda test anyway to get it to pass), while not a proper test the bundle size script is run against the web adapter in this PR: https://github.com/prisma/prisma/pull/26776/files#diff-502f8d135ffba468d44ae1553239a364576436fda5ff4108067c6230d549fd04L5, which confirms that the export works

jkomyno: Note: my comment above is not blocking. As discussed irl, let's create a follow-up Linear ticket for creating ecosystem-tests to assert how these exports work in Cloudflare Workers and Vercel Edge. It's possible that the Node.js version actually works in these edge environments, as long as a `node_compat` attribute (or similar) is set.

jkomyno (approved):

viktor-ferenczi (approved):

PR #26777, opened 2025-03-28
feat(generator-js): warn if output path is not set

aqrln:
Closes: https://linear.app/prisma-company/issue/ORM-669/warn-if-output-path-is-not-set


packages/client-generator-js/src/generator.ts
```diff
@@ -17,6 +18,11 @@ type PrismaClientJsGeneratorOptions = {
   runtimePath?: string
 }
 
+const MISSING_CUSTOM_OUTPUT_PATH_WARNING = `\
+${yellow('Warning:')} You did not specify an output path for your \`generator\` in schema.prisma. \
+This behavior is deprecated and will no longer be supported in Prisma 7.0.0. To learn more \
+visit <URL>`
```
aqrln, line 24: needs the URL before merging

FGoessler (approved):

PR #26778, opened 2025-03-28
feat(config): read and bind adapters eagerly

jkomyno:
Currently, we require plain Driver Adapters from `prisma.config.ts`, and we forward them as they are to our `PrismaConfigInternal` object, that is then consumed by individual commands. Since e.g. `config.studio.adapter` and `config.migrate.adapter` are async factory functions, this implies that we only try calling these functions within the logic of existing commands (for instance, after validating command line arguments, etc).

However, we actually need to bind the this instances of these adapters in order for them to work properly with the Rust bindings we expose via Wasm. For studio, this is handled by bindAdapter within `@prisma/client` itself. For `prisma migrate` / `prisma db`, however, calling `bindSqlAdapterFactory` in the `Migrate` class can is discomforting, as:
there are cases where multiple instances of `Migrate` exist within the same command -> duplication of CPU work
we create one `Migrate` instance for each command, like 10 different times, with different arguments

Moreover, I'm not sold on the current semi-lazy nature of Prisma's config, where the values returned by the `adapter()` async functions aren't validated.

Here's a PR that doesn't change the public-facing API of `prisma.config.ts`, but it moves the invocations of config functions, such as `config.migrate.adapter()` up by one level, within `@prisma/config` itself, and changes PrismaConfigInternal so that we can have easily reference already instantiated and this-bound adapters via .e.g. config.migrate.adapter.
This makes the overall boundaries between `@prisma/config` and cli clearer, centralises potential error cases to the same code location (any async function invocation, like `config.migrate.adapter()`, may potentially fail; right now this invocation is scattered within `Db` and `Migrate` commands), and feels clearer to me too.


jkomyno:
Deprecated by https://github.com/prisma/prisma/pull/26781.

PR #26782, opened 2025-03-28
feat(migrate): introduce `@prisma/schema-engine-wasm` and `SchemaEngineWasm`

jkomyno:
This PR:
- closes [ORM-700](https://linear.app/prisma-company/issue/ORM-700/create-schemaenginewasm-class-that-replaces-the-schemaengine-json-rpc) and [ORM-699](https://linear.app/prisma-company/issue/ORM-699/opt-in-for-schema-engine-wasm-when-its-provided-in-prisma-config)
- installs `@prisma/schema-engine-wasm` into Prisma, sets it up in `packages/internals/src/wasm.ts`, and hooks it into a new `SchemaEngineWasm` class
- renames `SchemaEngine` into `SchemaEngineCLI`
- forces `SchemaEngineWasm` and `SchemaEngineCLI` to adhere to the same interface, `SchemaEngine`
- updates `Migrate` so that it conditionally creates a `SchemaEngineWasm` whenever an `adapter` is provided; this is done in an async `setup` function
- updates tests and commands that were using `new SchemaEngine` and `new Migrate`
- deprecates https://github.com/prisma/prisma/pull/26753 and https://github.com/prisma/prisma/pull/26760


jkomyno:
After this is merged, we should merge https://github.com/prisma/prisma/pull/26755 as well.

packages/migrate/src/SchemaEngineWasm.ts
```diff
@@ -0,0 +1,268 @@
...
+  debug?: boolean
+  engine: wasm.SchemaEngineWasm
+}
+
+/**
+ * Wrapper around `@prisma/schema-engine-wasm`.
+ */
+export class SchemaEngineWasm implements SchemaEngine {
+  private engine: wasm.SchemaEngineWasm
+  private schemaContext?: SchemaContext
```
FGoessler, line 37: 🧹Unused property? Same for `enabledPreviewFeatures` below. We only need them in the `setup` call but get them passed there explicitly.

packages/migrate/src/SchemaEngineWasm.ts
```diff
@@ -0,0 +1,268 @@
...
+   * Diffs have a direction. Which source is from and which is to matters.
+   * The resulting diff should be thought of as a migration from the schema in `args.from` to the schema in `args.to`.
+   * By default, we output a human-readable diff. If you want an executable script, pass the "script": true param.
+   */
+  public async migrateDiff(input: SchemaEngineInput<'diff'>) {
+    const { stdout, ...rest } = await this.runCommand('diff', input)
+
+    if (stdout) {
+      // Here we print the content from the Schema Engine to stdout directly
+      // (it is not returned to the caller)
```
FGoessler, line 204: ❓ According to the types there is never output returned to the called other than the exit-code. Or am I missing something? 🤔

FGoessler (changes requested):
Minor nitpicks and cleanup suggestions overall looks great! 👍

packages/migrate/src/SchemaEngineWasm.ts
jkomyno: They are useful, but left to a follow-up PR.
You can see the TODO list in the [linked Linear ticket](https://linear.app/prisma-company/issue/ORM-700/create-schemaenginewasm-class-that-replaces-the-schemaengine-json-rpc#comment-0d3a904f).

packages/migrate/src/SchemaEngineWasm.ts
```diff
@@ -0,0 +1,268 @@
...
+   * `compositeTypeDepth` is optional, and only required for MongoDB.
+   */
+  public async introspect({
+    schema,
+    force = false,
+    baseDirectoryPath,
+    viewsDirectoryPath,
+    compositeTypeDepth = -1, // cannot be undefined
+    namespaces,
+  }): SchemaEngineOutput<'introspect'> {
```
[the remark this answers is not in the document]
jkomyno: No, this change would result in a `Promise<Promise<...>>`.

> Also: Why don't we specify a type here for the args object? 🤔
Right, we could. I started this class by fully typing input args and and later implementing an interface. Now that we have a backing interface. I expect the interface to living only as long as `SchemaEngineCLI` lives, so I'll add the explicit input type here.

packages/migrate/src/SchemaEngineWasm.ts
jkomyno: > According to the types there is never output returned to the called other than the exit-code
Which is precisely what this implementation does :)
We print `stdout` to `process.stdout`, and return the rest (which is currently only `exitCode`) to the caller.
Compare it with https://github.com/prisma/prisma/blob/0d23f0a834e44045ea9ca07f2d22f82f90cb45e9/packages/migrate/src/SchemaEngine.ts#L200-L211 + https://github.com/prisma/prisma/blob/0d23f0a834e44045ea9ca07f2d22f82f90cb45e9/packages/migrate/src/SchemaEngine.ts#L298-L301 + https://github.com/prisma/prisma-engines/blob/fbda4d61f6cc9c7361b803e72f3a0ffeb87db447/schema-engine/connectors/schema-connector/src/connector_host.rs and its usages in https://github.com/prisma/prisma-engines/blob/fbda4d61f6cc9c7361b803e72f3a0ffeb87db447/schema-engine/core/src/commands/diff_cli.rs#L73.

packages/internals/src/WasmSchemaEngineLoader.ts
```diff
@@ -0,0 +1,48 @@
+// this import points directly to ./schema_engine_bg.js it is generated with >>>
```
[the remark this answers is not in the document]
jkomyno: Yeah, this was related to `import type { SchemaEngine } from '@prisma/schema-engine-wasm'`

packages/migrate/src/SchemaEngineWasm.ts
FGoessler: What confused me about the comment was the fact that it is placed inside the `if (stdout)` block. I was thinking at first: "Ah ok so if we have `stdout` we print it, but based on this comment it seems like there might be a code path where we do not print it or somehow return it. Let me have a closer look into this."

Fine for me to keep it like this now. 👍

jkomyno: I see. To be fair, the existing logic was very weird/cursed to begin with. I discovered how it worked in details last week only

packages/internals/src/WasmSchemaEngineLoader.ts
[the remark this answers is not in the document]
FGoessler: Let's move this comment then and you got my approval ;)

FGoessler (approved):

PR #26785, opened 2025-03-29
feat(client): always use inline schema

aqrln:
`BinaryEngine` was the only `Engine` implementation that still needed the schema to be present on disk. This changes it to use the inline schema, just like the other engines. This will allow not copying the schema or embedding its path in the new client.


aqrln:
https://github.com/prisma/prisma/pull/26783 was accidentally closed by GitHub automatically due to git branch shenanigans

jkomyno (approved):

PR #26788, opened 2025-03-30
ci: Fixes failing Auto Close GH Discussions Action

nurul3101:


nurul3101:
This fixes the failing GitHub Action: https://github.com/prisma/prisma/actions/runs/14150877067/job/39643862285

.github/workflows/auto-close-github-discussions.yml
```diff
@@ -17,6 +17,12 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Setup PNPM
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10
```
aqrln, line 23:
```suggestion
```
No need to specify the version here, there's a `packageManager` entry in `package.json` so it should download the right version, and, fwiw, we use pnpm 9 and not 10.

nurul3101: Thanks for letting me know. Removed the version.

aqrln (approved):

PR #26799, opened 2025-04-02
feat(migrate): conditionally drop support for CLI flags when using `@prisma/schema-engine-wasm`

jkomyno:
This PR:
- closes [ORM-701](https://linear.app/prisma-company/issue/ORM-701/prisma-config-conditionally-drop-support-for-flags-when-using-a)

When a Schema Engine Driver Adapter is given in prisma.config.ts, we should drop support for the following subsets of Prisma commands:'

- [x] prisma migrate diff: --from-url, --to-url, --from-schema-datasource, --to-schema-datasource, --shadow-database-url, --to-local-d1, --from-local-d1
- [x] prisma db execute: --url
- [x] prisma db pull: --url, --local-d1

This is needed to make the Wasm-ification feasible for schema-core, and to conform to the new D1 migration strategy.

### DIFF
--- a/packages/migrate/src/__tests__/MigrateDiff.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDiff.test.ts
@@ -4,7 +4,7 @@
 import os from 'node:os'
 import path from 'node:path'
 
-import { defaultTestConfig } from '@prisma/config'
+import { defaultTestConfig, loadConfigFromFile } from '@prisma/config'
 import { jestConsoleContext, jestContext } from '@prisma/get-platform'
 
 import { MigrateDiff } from '../commands/MigrateDiff'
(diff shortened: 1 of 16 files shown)


packages/migrate/src/__tests__/MigrateDiff.test.ts
```diff
@@ -32,6 +32,162 @@ describe('migrate diff', () => {
     captureStdout.stopCapture()
   })
 
+  describe('using Prisma Config', () => {
```
jacek-prisma, line 35: I think using a table-based test with
```ts
describe.each([
  ['--from-url', 'file:./dev.db'],
  ['--to-url', 'file:./dev.db']
...
])
```
could remove a lot of duplication here

jacek-prisma (approved):

PR #26802, opened 2025-04-02
fix(client): revert cwd change in `BinaryEngine`

aqrln:
Revert `cwd` related changes from <https://github.com/prisma/prisma/pull/26785>. Although we don't need the current working directory to be set to find the schema file, it is still necessary to find the client TLS certificates when those are used.

/integration


jkomyno (approved):

aqrln:
Failing before: https://github.com/prisma/ecosystem-tests/actions/runs/14217917245/job/39838921456
Fixed after: https://github.com/prisma/ecosystem-tests/actions/runs/14222461341/job/39853914999

PR #26833, opened 2025-04-08
feat: driver adapter error mappings

jacek-prisma:
[ORM-522](https://linear.app/prisma-company/issue/ORM-522/migrate-database-error-mapping-from-rust-to-typescript)


packages/adapter-d1/src/errors.ts
```diff
@@ -0,0 +1,66 @@
...
+
+// When we receive the result, we only get the error message, not the error code.
+// Example:
+// "name":"Error","message":"D1_ERROR: UNIQUE constraint failed: User.email"
+// So we try to match some errors and use the generic error code as a fallback.
+// https://www.sqlite.org/rescode.html
+export function convertDriverError(error: any): DriverAdapterErrorObject {
+  if (!isDbError(error)) {
+    throw error
+  }
```
jkomyno, line 11: We used to have `throw new DriverAdapterError(error.message)`, now we'd have `throw error` in this case.
Is this something we actually want?

jkomyno (approved):

packages/adapter-d1/src/errors.ts
jacek-prisma: `DriverAdapterError` was meant to wrap a valid database error, in this case we don't have a valid database error payload, so we throw the error we received instead,

This case is basically unreachable in this adapter, but by convention I made all the `convertDriverError` accept `any` error type and do a runtime type check (this could be changed to just accept the javascript Error type)

PR #26851, opened 2025-04-09
refactor(client-engine-runtime): use `assertNever` in `QueryInterpreter`

aqrln:
Use `assertNever` instead of inline implementation in `QueryInterpreter#interpretNode`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ The Prisma CLI includes a [Prisma MCP server](https://www.prisma.io/docs/postgre
 ```
 npx prisma mcp
 ```
- 
+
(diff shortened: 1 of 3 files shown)


jacek-prisma (approved):

jacek-prisma:
oh, there are inline implementations of it in `client-engine-runtime/src/UserFacingError.ts#L79` and `client-engine-runtime/src/UserFacingError.ts#L138` that I added, you can replace them too if you want

aqrln:
@jacek-prisma done

jacek-prisma (approved):

PR #26861, opened 2025-04-10
feat(client-engine-runtime): implement tracing support

aqrln:
Add tracing support to `@prisma/client-engine-runtime` and `ClientEngine`.

Closes: https://linear.app/prisma-company/issue/ORM-557/implement-tracing-support-in-the-clientengine


packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
```diff
@@ -39,13 +40,27 @@ export class TransactionManager {
   private closedTransactions: TransactionWrapper[] = []
   private readonly driverAdapter: SqlDriverAdapter
   private readonly transactionOptions: Options
-
-  constructor({ driverAdapter, transactionOptions }: { driverAdapter: SqlDriverAdapter; transactionOptions: Options }) {
+  private readonly tracingHelper: TracingHelper
+
+  constructor({
```
viktor-ferenczi, line 45: We also have `new TransactionManager` in `worker.ts` in the `prisma-engines` repo. We need a PR on the `prisma-engines` side, so the tests will pass there once this PR is on `main`.

viktor-ferenczi (changes requested):
Great job.

jacek-prisma (approved):

packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
aqrln: Good call out, done in https://github.com/prisma/prisma-engines/pull/5329

viktor-ferenczi (approved):

PR #26862, opened 2025-04-10
feat: ORM-668 GA prismaSchemaFolder

FGoessler:
This PR brings the `prismaSchemaFolder` feature to GA by removing the need to specify the `previewFeatures` flag in the schema file.

Also see https://github.com/prisma/prisma-engines/pull/5322


packages/internals/src/__tests__/getSchema.test.ts
```diff
@@ -190,43 +190,6 @@ it('finds the conventional prisma/schema path without configuration', async () =
   expect(res).toMatchInlineSnapshot(`"src/__tests__/__fixtures__/getSchema/conventional-path/prisma/schema.prisma"`)
 })
 
-it('throws error if both schema file and folder exist at default locations', async () => {
```
jacek-prisma, line 193: isn't this test case still valid?

FGoessler: Nope. There is no "default folder" location anymore. Only a default file location. If you wanna use a folder you gotta be explicit about it. This test should have been nuked earlier already though. 😅

jacek-prisma (approved):

PR #26871, opened 2025-04-11
chore: shorten CI names of test matrixes

FGoessler:


viktor-ferenczi (approved):

.github/workflows/test-template.yml
```diff
@@ -990,7 +990,7 @@ jobs:
   # Driver Adapter Unit Tests
   #
   driver-adapter-unit-tests:
-    name: Driver Adapter Unit Tests
+    name: DA Unit Tests [v${{ matrix.node }}]
```
jacek-prisma, line 993:
```suggestion
    name: Driver Adapter [v${{ matrix.node }}]
```
I think it's more useful to highlight that it's a Driver Adapter job than that its a unit test

.github/workflows/test.yml
```diff
@@ -157,7 +157,7 @@ jobs:
           exit 1
 
   run_tests:
-    name: Tests
+    name: 🧪
```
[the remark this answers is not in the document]
FGoessler: That level here imo gives no relevant info or context to the test name. It's just there because the higher level grouping requires it. Hence I wanted to get this as short as possible so it takes up as little space as possible. And I doubt you would ever search for this term. ;)

It's especially helpful in this view to give more space for the actual test name:
Screenshot 2025-04-14 at 14 36 36

.github/workflows/test-template.yml
jacek-prisma: @FGoessler I don't like the DA acronym, I don't think we use it anywhere so it's not obvious what it means here

FGoessler: Any other ideas how we can make the tests readable in those sidebars? 🤔 
If I just read "Tests / Driver Adapters..." repeated 20+ times I have no value of it and cannot navigate test failures efficiently :/

PR #26891, opened 2025-04-14
fix(da): Fixed better-sqlite3 database file path

viktor-ferenczi:
- Getting the better-sqlite3 database file path by removing the `file:` prefix from the `url`

In `prisma-engines` CI tests it receives URLs with the `file:` prefix, for example:
`file:/home/runner/work/prisma-engines/prisma-engines/db/order_by_aggr_one2m_count_asc_field_desc.db`

### DIFF
--- a/packages/adapter-better-sqlite3/src/better-sqlite3.ts
+++ b/packages/adapter-better-sqlite3/src/better-sqlite3.ts
@@ -222,7 +222,8 @@ export class PrismaBetterSQLite3AdapterFactory implements SqlMigrationAwareDrive
 
 function createBetterSQLite3Client(input: BetterSQLite3InputParams): StdClient {
   const { url, ...config } = input
-  const db = new Database(url, config)
+  const dbPath = url.replace(/^file:/, '')
+  const db = new Database(dbPath, config)
   db.defaultSafeIntegers(true)
   return db
 }


packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -200,8 +200,8 @@ export class PrismaBetterSQLite3Adapter extends BetterSQLite3Queryable<StdClient
 }
 
 type BetterSQLite3InputParams = BetterSQLite3Options & {
-  url: ':memory:' | (string & {})
-  shadowDatabaseURL?: ':memory:' | (string & {})
```
FGoessler, line 204: Let's leave that in to make it clear we support those special in memory database syntax. ;)

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -200,8 +200,8 @@ export class PrismaBetterSQLite3Adapter extends BetterSQLite3Queryable<StdClient
...
+  url: string & {}
+  shadowDatabaseURL?: string & {}
```
FGoessler, line 204:
```suggestion
  url: ':memory:' | (string & {})
  shadowDatabaseURL?: ':memory:' | (string & {})
```

FGoessler (approved):

packages/adapter-better-sqlite3/src/better-sqlite3.ts
viktor-ferenczi: Where does the requirement for this type come from?

It may well receive `file::memory:` in practice.

ineshbose:
Can we look into reverting this after https://github.com/WiseLibs/better-sqlite3/commit/674ce6be68a26742d9e24f8672da7888cea0aebb please?

There seem to be a conflicts here, and it breaks fileURLs having `file://` (with two slashes).

PR #26905, opened 2025-04-15
fix(client): propagate `dirname` from `client.ts` to `class.ts`

aqrln:
<https://github.com/prisma/prisma/pull/26892> added the missing `__dirname` ESM shim to `class.ts` but one more issue remained: moving the client configuration to `internals/class.ts` changed the value of `__dirname` to point at the `internals` directory and not at the client directory itself, causing issues like SQLite databases not being found in new e2e tests. At the same time, we cannot use
`path.dirname(__dirname)` because it will only work as long as the client is not bundled, and will lead to incorrect results if the client is bundled. Therefore it's actually not correct to use `__dirname` in `internals/*` at all, and we must propagate it from the `client.ts` file downstream to whatever other files need it.

This commit replaces the `PrismaClient` export of `internals/class.ts` with a function that takes `dirname` as an argument and returns the `PrismaClient` class. Calling this function is deferred to the `client.ts` file.


viktor-ferenczi (approved):
Reviewed it during a call.

PR #26914, opened 2025-04-15
feat: surface user facing compile errors

jacek-prisma:
surfaces errors implemented in https://github.com/prisma/prisma-engines/pull/5338 as `UserFacingError`


packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -421,6 +427,14 @@ export class ClientEngine implements Engine<undefined> {
   }
 }
 
+function rethrowCompileErrorAsUserFacing(error: any): never {
+  if (typeof error['message'] === 'string' && typeof error['code'] === 'string') {
+    throw new UserFacingError(error['message'], error['code'], error.meta)
```
aqrln, line 432: Shouldn't it instead be mapped to `PrismaClientKnownRequestError` here?

packages/client-common/src/QueryCompiler.ts
```diff
@@ -17,8 +17,8 @@ export type CompactedBatchResponse = {
 }
 
 export type QueryCompiler = {
-  compile(request: string): Promise<string>
-  compileBatch(batchRequest: string): Promise<BatchResponse>
```
jacek-prisma, line 21: the types here and invalid, the compiler doesn't return promises, our code somehow worked because we never used any promise methods on it!

aqrln (approved):

jacek-prisma:
will be merged in https://github.com/prisma/prisma/pull/26967

PR #26918, opened 2025-04-16
build(instrumentation): Emit ESM

### LINKED ISSUE #23410: Add esm support to `@prisma/instrumentation`
## Problem

`@prisma/instrumentation` is not ESM compatible.

https://github.com/prisma/prisma/blob/caeb9ba88cba8bafca8654e2c903160aa16a3fd5/packages/instrumentation/src/constants.ts#L1

The require call above is one part of it, but the other is that the package needs to be emitted with `type: module` or `*.mjs`.

## Suggested solution

Make the package ESM compatible.

lforst:
Emit the prisma instrumentation package as esm.

Fixes https://github.com/prisma/prisma/issues/23410


aqrln (approved):
LGTM, thanks!

PR #26929, opened 2025-04-17
feat(qc): Data mapper

viktor-ferenczi:
A tree structure of `ResultNode` instances represent the expected shape of the result data. It includes the output field names, database field name and the data types to cast to.

The root `ResultNode` is stored in a `DataMap` expression node, which wraps the entire query. This node drives the transformation of the query result to the expected shape and coercing to the expected data types as needed inside the Query Interpreter on the client side.

Implements [ORM-734](https://linear.app/prisma-company/issue/ORM-734/design-and-implement-data-mapper-nodes)

Depends on https://github.com/prisma/prisma-engines/pull/5343


packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -0,0 +1,166 @@
...
+  switch (structure.type) {
+    case 'object':
+      return mapArrayOrObject(data, structure.fields)
+
+    case 'value':
+      return mapValue(data, structure.resultType)
+
+    default:
+      // @ts-ignore
+      assertNever(structure, `Invalid data mapping type: '${structure.type}'`)
```
aqrln, line 15:
```suggestion
      assertNever(structure, `Invalid data mapping type: '${(structure as ResultNode).type}'`)
```

packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+          throw new Error(
+            `DataMapper[3]: Missing data field: '${name}'; ` +
+              `node: ${JSON.stringify(node)}; data: ${JSON.stringify(data)}`,
+          )
+        }
+        break
+
+      case 'value':
+        if (node.dbName in data) {
```
aqrln, line 56:
```suggestion
        if (Object.hasOwn(data, node.dbName)) {
```

packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -0,0 +1,166 @@
...
+        assertNever(node, `Invalid data mapping node type: '${node.type}'`)
+    }
+  }
+  return result
+}
+
+function mapValue(value: unknown, resultType: PrismaValueType): unknown {
+  if (value === null) return null
+
+  if (typeof resultType === 'string') {
```
aqrln, line 77: We may want to choose a different encoding (e.g. `#[serde(tag = "type")]`) to make it easier to work with this enum in TypeScript.

aqrln (approved):

packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -0,0 +1,111 @@
...
+    case 'BigInt':
+      return typeof value === 'bigint' ? value : BigInt(`${value}`)
+    case 'Float':
+      return typeof value === 'number' ? value : parseFloat(`${value}`)
+    case 'Boolean':
+      return typeof value === 'boolean' ? value : value !== '0'
+    case 'Decimal':
+      return typeof value === 'number' ? value : parseFloat(`${value}`)
+    case 'Date':
+      return value instanceof Date ? value : new Date(`${value}`)
```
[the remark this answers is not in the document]
viktor-ferenczi: Thanks, added a comment on that ticket.

PR #26947, opened 2025-04-22
feat(cli): make `--allow-no-models` the default behaviour

aqrln:
Make `--allow-no-models` the default behaviour for `generate` command, keep the old flag as no-op for backwards compatibility (so users' scripts that use it will not break) and add an opposite flag `--disallow-no-models` to restore the old behaviour.

/integration

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -64,12 +64,13 @@ ${bold('Options')}
           -h, --help   Display this help message
             --config   Custom path to your Prisma config file
             --schema   Custom path to your Prisma schema
+               --sql   Generate typed sql module
              --watch   Watch the Prisma schema and rerun after a change
          --generator   Generator to use (may be provided multiple times)
          --no-engine   Generate a client for use with Accelerate only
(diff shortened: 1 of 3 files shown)


packages/cli/src/Generate.ts
```diff
@@ -126,9 +127,19 @@ ${bold('Examples')}
       '--postinstall': String,
       '--telemetry-information': String,
       '--allow-no-models': Boolean,
+      '--disallow-no-models': Boolean,
```
viktor-ferenczi, line 130: Double negation. Maybe `--require-model` ?

viktor-ferenczi (approved):

packages/cli/src/Generate.ts
viktor-ferenczi: Or `--must-have-a-model` if "require" would be confusing.

aqrln: Made it `--require-models`, in plural. Singular makes sense logically but plural sounds a bit better to me, and is consistent with the opposite flag.

PR #26953, opened 2025-04-23
feat: put engineType="client" behind a preview feature and not env var

aqrln:
Closes: https://linear.app/prisma-company/issue/ORM-864/add-qc-preview-feature-flag


FGoessler (approved):

FGoessler:
We'll also need to add the previewFeature to the list in prisma-engines (so it works with the schema validator etc.) , right? 🤔

PR #26983, opened 2025-04-25
feat(cli): add error registry lookup for driver adapter errors during schema engine wasm execution

FGoessler:
So far during CLI time we would often just get a `Error: External error id#0` if something went wrong on the JS side of a driver adapter during CLI execution via the new Schema Engine WASM.

This is not only confusing for the user, but also hinders debugging and adjusting our CLI/Migrate test suite (see ORM-708).

Relates to ORM-708.

### DIFF
--- a/packages/driver-adapter-utils/src/binder.ts
+++ b/packages/driver-adapter-utils/src/binder.ts
@@ -16,11 +16,34 @@ import type {
 
 const debug = Debug('driver-adapter-utils')
 
+const DRIVER_ADAPTER_EXTERNAL_ERROR = 'P2036'
+
 class ErrorRegistryInternal implements ErrorRegistry {
   private registeredErrors: ErrorRecord[] = []
 
-  consumeError(id: number): ErrorRecord | undefined {
-    return this.registeredErrors[id]
+  lookupError(
+    error: number | { error_code: string; meta?: { id: number } } | { message: string },
+  ): ErrorRecord | undefined {
+    if (typeof error === 'number') {
+      const errorRecord = this.registeredErrors[error]
+      if (!errorRecord) throw `External error with reported id was not registered`
+      return errorRecord
+    } else if ('error_code' in error && error.error_code === DRIVER_ADAPTER_EXTERNAL_ERROR) {
+      // Query Engine Runtime Errors will look like this
+      const id = error.meta?.id
+      if (typeof id !== 'number') throw 'Malformed external JS error received from the engine'
+      const errorRecord = this.registeredErrors[id]
+      if (!errorRecord) throw `External error with reported id was not registered`
+      return errorRecord
+    } else if ('message' in error && error.message.startsWith('External error id#')) {
+      // Schema Engine CLI Runtime Errors will look like this
+      const id = parseInt(error.message.split('id#')[1])
+      const errorRecord = this.registeredErrors[id]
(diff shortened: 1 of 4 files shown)


packages/driver-adapter-utils/src/binder.ts
```diff
@@ -16,11 +16,34 @@ import type {
...
+      // Query Engine Runtime Errors will look like this
+      const id = error.meta?.id
+      if (typeof id !== 'number') throw 'Malformed external JS error received from the engine'
+      const errorRecord = this.registeredErrors[id]
+      if (!errorRecord) throw `External error with reported id was not registered`
+      return errorRecord
+    } else if ('message' in error && error.message.startsWith('External error id#')) {
+      // Schema Engine CLI Runtime Errors will look like this
+      const id = parseInt(error.message.split('id#')[1])
+      const errorRecord = this.registeredErrors[id]
```
jacek-prisma, line 41: we should probably add some mapping to return a properly structured error here, but we can handle it like this in the interim (might be worth a TODO)

jacek-prisma (approved):

packages/migrate/src/SchemaEngineWasm.ts
```diff
@@ -67,14 +67,18 @@ export class SchemaEngineWasm implements SchemaEngine {
...
+      const engine = await wasmSchemaEngineLoader.loadSchemaEngine(
+        {
+          datamodels,
+        },
+        debug,
+        adapter,
+      )
+      return new SchemaEngineWasm({ ...rest, engine, errorRegistry: adapter.errorRegistry })
+    } catch (e) {
+      throw adapter.errorRegistry.lookupError(e as Error) || e
```
aqrln, line 80: nit
```suggestion
      throw adapter.errorRegistry.lookupError(e as Error) ?? e
```

packages/driver-adapter-utils/src/binder.ts
```diff
@@ -16,11 +16,34 @@ import type {
...
   private registeredErrors: ErrorRecord[] = []
 
-  consumeError(id: number): ErrorRecord | undefined {
-    return this.registeredErrors[id]
+  lookupError(
+    error: number | { error_code: string; meta?: { id: number } } | { message: string },
+  ): ErrorRecord | undefined {
+    if (typeof error === 'number') {
+      const errorRecord = this.registeredErrors[error]
+      if (!errorRecord) throw `External error with reported id was not registered`
```
aqrln, line 29: I feel like the logic is backwards compared to the function signature. We throw if the specified error is not found (although the signature suggests `undefined` would be returned in this case) and only return `undefined` in the final `else` branch if the shape of the input was completely wrong and we couldn't interpret it (which it would be more appropriate to throw an error about).

packages/migrate/src/SchemaEngineWasm.ts
aqrln: Aso `throw adapter.errorRegistry.lookupError(e as Error)` is probably wrong, because `errorRegistry.lookupError` doesn't return an `Error`, it returns an `ErrorRecord`, which is defined as `{ error: unknown }`.

```suggestion
      throw adapter.errorRegistry.lookupError(e as Error)?.error ?? e
```

FGoessler:
Closing for now. Can revisit this when we continue on schema engien wasm.

PR #26995, opened 2025-04-28
fix: fix datasource.url warning

FGoessler:
The warning was printed incorrectly when using a driver adapter and not having a URL specified. The drawbacks of the workaround to not make it truly optional 😬

### DIFF
--- a/packages/migrate/src/utils/warnDatasourceDriverAdapter.ts
+++ b/packages/migrate/src/utils/warnDatasourceDriverAdapter.ts
@@ -2,6 +2,9 @@ import { ErrorCapturingSqlDriverAdapterFactory } from '@prisma/driver-adapter-ut
 import { SchemaContext } from '@prisma/internals'
 
 const DEPRECATED_PROPERTIES = ['url', 'directUrl', 'shadowDatabaseUrl']
+// Due to the complexity of making the datasource.url truly optional across the codebase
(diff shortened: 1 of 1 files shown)


packages/migrate/src/utils/warnDatasourceDriverAdapter.ts
```diff
@@ -2,6 +2,9 @@ import { ErrorCapturingSqlDriverAdapterFactory } from '@prisma/driver-adapter-ut
 import { SchemaContext } from '@prisma/internals'
 
 const DEPRECATED_PROPERTIES = ['url', 'directUrl', 'shadowDatabaseUrl']
+// Due to the complexity of making the datasource.url truly optional across the codebase
+// this placeholder value is used in the PSL parser if no URL property is given.
+const NO_URL_PLACEHOLDER = '<invalid>'
```
viktor-ferenczi, line 7: Shouldn't is be `'<N/A>'`? 

It is not "invalid", just not specified.

viktor-ferenczi (approved):

packages/migrate/src/utils/warnDatasourceDriverAdapter.ts
FGoessler: Fair point - I just went with what prisma-engines sets it to.I'll merge this in anyways for now and see down the line how to improve this also in line with our test setup.

PR #27011, opened 2025-04-29
chore: ORM-708 run migrate package tests with driver adapters

FGoessler:
This PR enables our `prisma/migrate` test suite to be run using driver adapters and our new WASM SchemaEngine. For this some significant refactoring and cleanup of the test suite had to be performed.

To run the test suite with driver adapters you can now do:
```sh
cd packages/migrate
PRISMA_MIGRATE_TEST_ADAPTER=libsql pnpm test
```

> [!NOTE]
> There are still 33 test failures left when running with the libsql driver adapater. Most of them relate to missing/broken `prisma migrate dev` and `prisma migrate reset` functionality. Fixing those is planned to be done via [ORM-901](https://linear.app/prisma-company/issue/ORM-901/fix-remaining-prismamigrate-pkg-test-failures) & [ORM-897](https://linear.app/prisma-company/issue/ORM-897/make-prisma-migrate-dev-work).

# Main problems that had to be solved:

- A) Injecting a driver adapter on demand based on the current test matrix configuration.
- B) Running certain tests only for driver adapters that support them. Incl. dealing with empty test suites.
- C) Normalization of captured stdout output as it differs between driver adapters/WASM engine and legacy engine.
- D) Standardizing loading sqlite databases as `datasource.url` in schema files is not considered by driver adapters.
- E) Tests used various different setups, mocking patterns etc.
- F) Jest will fail if it detects "obsolete" snapshots. Not running a test makes its snapshots that are recorded in a snapshot file marked as obsolete. => We cannot conditionally run tests with regular snapshots. Inline snapshots work though.

# Main conceptual changes:

- We are now consistently using the `jestContext` (`.../test-utils/jest.Context.ts` & `.../__tests__/__helpers__/context.ts`) approach to inject various behavior into the test context. In particular:
  - Capturing stdout/stderr/console.log - incl normalization.
  - Setting up test fixtures and tmp directories.
  - Creating a PrismaConfig object (`.../__tests__/__helpers__/prismaConfig.ts`) => `defaultTestConfig()` is replaced by `await ctx.config()`. This also injects the driver adapter based on the given `PRISMA_MIGRATE_TEST_ADAPTER` env var. Also merges the default config wit any config file present in the current test fixture.
  - Mocking and asserting `process.exit`.
- Created the `describeOnly` (`.../__tests__/__helpers__/conditionalTests.ts`) helper to mark specific tests as runnable only for a specific database provider or driver adapter.
- Certain log output is stripped. Some is normalized via placeholders. See `.../__tests__/__helpers__/stdoutNormalizationRules.ts`.
- All sqlite db files are now moved to the "root" of their fixture directory. This way the driver adapters can load those files in a uniform way. All schema files have been updated with relative paths. Mostly this meant moving the db file one folder level up outside of the `/prisma` directory.
- `WasmSchemaEngineLoader` had to be modified to allow multiple different instantiations of the WASM SchemaEngine with different arguments for the different test runs.
- Transformed all snapshot tests to inline snapshots tests.

# Minor side issues:

- `handlePanic.migrate.test.ts` for the interactive CLI case was actually a non-functional test as it was not actually testing an interactive CLI case. => dropped this test
- Various `try-catch` based tests were missing `expect.assertions(<n>)` statements.
- Various test cleanups were not performed consistently or at all.


packages/internals/helpers/build.ts
```diff
@@ -1,6 +1,30 @@
+import type * as esbuild from 'esbuild'
+import fs from 'fs'
+import path from 'path'
+
 import { build } from '../../../helpers/compile/build'
 import { esmSplitCodeToCjs } from '../../../helpers/compile/plugins/esmSplitCodeToCjs'
 
+const copySchemaEngineWasmPlugin: esbuild.Plugin = {
```
FGoessler, line 8: ❓ I am not sure if this should be done like this. But otherwise the tests had trouble to find the WASM files.

@aqrln can you have look into this approach briefly?

aqrln: Yeah this looks about right given that we bundle `packages/internals`. A better approach though would be to stop bundling `packages/internals` IMO, I don't see how it's useful or why we do it, while there clearly are downsides, like this one, or the fact it can lead to code bloat and duplication of shared dependencies when the bundle is being re-bundled again as part of the CLI or client. Only the top level user-facing packages should be bundled.

packages/internals/src/WasmSchemaEngineLoader.ts
aqrln, line 1: Do we install the panic handler anywhere?

packages/migrate/src/__tests__/__helpers__/conditionalTests.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Mostly used to run tests specific to a particular database dialect only with driver adapters who support it.
+ * The test suite without driver adapters executes test from all providers at once using the legacy schema engine.
+ * To skip tests for all driver adapters, use `driverAdapter: false`.
+ * To skip tests for the legacy schema engine, use `driverAdapter: true`.
+ *
+ * Examples:
+ * - describeOnly({ sqlite: true}, "test will run only with all SQLite driver adapters and the legacy schema engine", () => { ... })
+ * - describeOnly({ driverAdapter: true}, "test will only run with driver adapters", () => { ... })
+ * - describeOnly({ driverAdapter: false}, "test will only run on the legacy schema engine", () => { ... })
+ * - describeOnly({ libsql: true}, "test will only run with libsql driver adapter", () => { ... })
```
aqrln, line 21: style nit

```suggestion
 * - describeOnly({ sqlite: true }, "test will run only with all SQLite driver adapters and the legacy schema engine", () => { ... })
 * - describeOnly({ driverAdapter: true }, "test will only run with driver adapters", () => { ... })
 * - describeOnly({ driverAdapter: false }, "test will only run on the legacy schema engine", () => { ... })
 * - describeOnly({ libsql: true }, "test will only run with libsql driver adapter", () => { ... })
```

packages/migrate/src/__tests__/__helpers__/conditionalTests.ts
```diff
@@ -0,0 +1,57 @@
...
+  // Skip tests that shall not run for any driver adapter if a driver adapter is used
+  if (adapterName && matrix.driverAdapter === false) return skip(name)
+  // Skip tests that shall run only for driver adapters if none is used
+  if (adapterName === undefined && matrix.driverAdapter === true) return skip(name)
+
+  if (
+    adapterName === undefined ||
+    matrix[adapterName] ||
+    (provider && matrix[provider]) ||
+    (adapterName !== undefined && matrix.driverAdapter === true)
```
aqrln, line 40: Should this be `matrix.driverAdapter !== false` (i.e. `true` or `undefined`)? It also feels like it's already covered by the checks before.

packages/migrate/src/__tests__/__helpers__/conditionalTests.ts
```diff
@@ -0,0 +1,57 @@
...
+  ) {
+    // eslint-disable-next-line jest/valid-describe-callback
+    describe(name, fn)
+  } else {
+    skip(name)
```
aqrln, line 45: Why do we use this custom `skip` function instead of the builtin `describe.skip`? The latter probably also wouldn't have issues with the snapshots.

packages/migrate/src/__tests__/__helpers__/conditionalTests.ts
FGoessler: Yeah I guess you're right. Will adjust it. :+1:

packages/internals/src/WasmSchemaEngineLoader.ts
FGoessler: I'll check 👀

FGoessler: Doesn't seem like it. We should fix this once we get back to working on the SchemaEngineWASM stuff 😬

packages/internals/helpers/build.ts
FGoessler: Ok - then I'll leave this for now and we optimize later. :+1:

packages/migrate/src/__tests__/__helpers__/conditionalTests.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (matrix.mongodb && process.env.TEST_SKIP_MONGODB) return skip(name)
+
+  const adapterName = currentDriverAdapterName()
+  const provider = currentProvider()
+  // Skip tests that shall not run for any driver adapter if a driver adapter is used
+  if (adapterName && matrix.driverAdapter === false) return skip(name)
+  // Skip tests that shall run only for driver adapters if none is used
+  if (adapterName === undefined && matrix.driverAdapter === true) return skip(name)
+
+  if (adapterName === undefined || matrix[adapterName] || (provider && matrix[provider])) {
```
aqrln, line 36: This still doesn't look right to me. As per this condition,

1. If `adapterName === undefined`, the tests will always run regardless of the contents of `matrix`; i.e., `describeOnly({ cockroachdb: true })` will run for all providers and not just CockroachDB.

   Judging from the comment above:
   ```
    * - describeOnly({ sqlite: true }, "test will run only with all SQLite driver adapters and the legacy schema engine", () => { ... })
    ```
   it looks like it might be intentional but it's quite confusing if that's the case.

2. The `provider &&` part of the expression is always true and doesn't check anything, because `adapterName !== undefined` implies `provider !== undefined` (since `provider` here is a driver adapter provider specifically, and is undefined when not using driver adapters).

3. If `adapterName !== undefined`, either `matrix[adapterName]`or `matrix[provider]` is required. In other words, `describeOnly({ driverAdapter: true })` will never run and will always be skipped, which contradicts the doc comment above that says it's supposed to run with all driver adapters.

FGoessler: Very valid points. The matrix logic was indeed to convoluted and inverse. 🤔 

I now refactored this matrix logic to be more explicit. No you basically pass an object to the describe method specifying under which conditions you want it to run. (Instead of telling it when to skip)

@aqrln let me know what you think!

aqrln: looks great!

packages/migrate/src/__tests__/__helpers__/conditionalTests.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Ensure type wise that `allDriverAdapters` actually contains all supported driver adapters
+declare const __testAllDriverAdaptersA: keyof typeof allDriverAdapters
+declare const __testAllDriverAdaptersB: DriverAdapterName
+// eslint-disable-next-line no-constant-condition
+if (false) {
+  __testAllDriverAdaptersB satisfies keyof typeof allDriverAdapters
+  __testAllDriverAdaptersA satisfies DriverAdapterName
+}
```
aqrln, line 54: This pattern with type tests is an overkill here, you can achieve the same thing easier:
```suggestion
export const allProviders = {
  mysql: true,
  postgres: true,
  sqlite: true,
  mongodb: true,
  cockroachdb: true,
  sqlserver: true,
  d1: true,
} satisfies Record<SupportedProviders, true>

export const allDriverAdapters = {
  pg: true,
  'pg-worker': true,
  planetscale: true,
  neon: true,
  libsql: true,
  d1: true,
} satisfies Record<DriverAdapterName, true>
```

aqrln (approved):

PR #27045, opened 2025-04-30
test(client): add e2e test for native TS support in Node.js

aqrln:
Opening as draft so as not to lose it and finish it later once we are done with QE stuff.

The functionality itself works fine:

```
aqrln@patchouli ~/p/p/p/c/t/e/t/node-erasable-syntax-only ((f0eb8571))> node src/test.ts
(node:12423) ExperimentalWarning: Type Stripping is an experimental feature and might change at any time
{ id: 1, email: 'john@doe.io', name: null }
```

but the test is broken because custom Dockerfile support in our e2e tests doesn't seem to work and needs to be fixed. A custom Dockerfile is used because the default one uses Node.js 18, which doesn't support TypeScript yet.

Closes: https://linear.app/prisma-company/issue/ORM-744/ts-erasablesyntaxonly-support

### DIFF
--- a/packages/client/tests/e2e/ts-client/node-erasable-syntax-only/.gitignore
+++ b/packages/client/tests/e2e/ts-client/node-erasable-syntax-only/.gitignore
@@ -1,0 +1,2 @@
+/src/generated
+/prisma/db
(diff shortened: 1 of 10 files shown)


aqrln:
Closing for now, feel free to reopen whoever picks up https://linear.app/prisma-company/issue/ORM-744/e2e-test-ts-erasablesyntaxonly-support

PR #27063, opened 2025-05-02
fix(addPreamble): change ts-nocheck to act on whole file

AlvesJorge:
Using `/* @ts-nocheck */` only affects the following code block.

Using `// @ts-nocheck` affects the whole file, which is what the original implementor intended I believe @FGoessler


aqrln (approved):

aqrln:
Looks good to me, thanks! Could you please [sign the CLA](https://github.com/prisma/prisma/pull/27063#issuecomment-2846691306) so we can merge it?

PR #27075, opened 2025-05-03
fix(config): use jiti to load prisma.config.ts

### LINKED ISSUE #26952: Error in Prisma Language Server after adding prisma.config.ts
### Bug description

After updating to the latest Prisma version, I tried adding a prisma.config.ts file to my project. However, once the file was created, the Prisma Language Server started throwing a series of errors in the VSCode window:

Image

Image

### Severity

### Reproduction

1. Create a project:  
   ```shell  
   pnpm dlx nuxi init prisma-demo  
   ```  

2. Install dependencies:  
   ```shell  
   pnpm add -E prisma @prisma/client  
   ```  

3. Initialize Prisma:  
   ```shell  
   pnpm prisma init --datasource-provider sqlite --output ../generated/prisma  
   ```  

4. Add `prisma.config.ts`:  
   ```ts  
   import path from 'node:path';  

   export default defineConfig({  
     earlyAccess: true,  
     schema: path.join('prisma', 'schema.prisma'),  
   });  
   ```  

5. Restart VSCode, reopen the project, and click on `schema.prisma`.

**Expected behavior:** Everything works normally, with `@esbuild/win32-x64` in the extension folder.  

**Actual behavior:** After installing the extension in a Windows environment, it generates `@esbuild/linux-x64` instead of `@esbuild/win32-x64`:

Image

### Workaround

**Temporary Fix:**

By examining the error logs, I located the extension folder and found only `@esbuild/linux-x64` inside. Since I'm on Windows 11 24H2, I manually copied `@esbuild/win32-x64` from another source into `~\.vscode\extensions\prisma.prisma-6.6.0\node_modules\@esbuild`. After this, everything worked normally. 

Image 

> No matter how many times I uninstalled/reinstalled the Prisma extension or cleared its folder, only `@esbuild/linux-x64` would appear. Manually adding `@esbuild/win32-x64` was necessary.

### Frequency

Consistently reproducible

### Environment & Setup

- **OS:** Windows 11 24H2
- **Database:** SQLite
- **Node.js version:** 22.14.0
- **Prisma**: 6.6.0  
- **VSCode**: 1.99.3  

### Prisma Version

```
Loaded Prisma config from "E:\demo\prisma-test\prisma.config.ts".
Prisma config detected, skipping environment variable loading.
prisma                  : 6.6.0
@prisma/client          : 6.6.0
Architecture            : x64
Node.js                 : v22.14.0
TypeScript              : 5.8.3
Query Engine (Node-API) : libquery-engine f676762280b54cd07c770017ed3711ddde35f37a (at node_modules\.pnpm\@prisma+engines@6.6.0\node_modules\@prisma\engines\query_engine-windows.dll.node)
Schema Engine           : schema-engine-cli f676762280b54cd07c770017ed3711ddde35f37a (at node_modules\.pnpm\@prisma+engines@6.6.0\node_modules\@prisma\engines\schema-engine-windows.exe)
Schema Wasm             : @prisma/prisma-schema-wasm 6.6.0-53.f676762280b54cd07c770017ed3711ddde35f37a
Default Engines Hash    : f676762280b54cd07c770017ed3711ddde35f37a
```

nissy-dev:
Fix https://github.com/prisma/prisma/issues/26952
Fix https://github.com/prisma/language-tools/issues/1825

Prisma is currently using `esbuild` to load `prisma.config.ts`, but in that case, the VSCode extension must [distribute platform-specific binaries appropriately](https://github.com/evanw/esbuild/blob/5959289d90667c5a4026e6fb32cc58bbed9fc88a/npm/esbuild/package.json#L20-L46). At the moment, the binary is not being distributed correctly, so the VSCode extension does not work on platforms other than Linux when `prisma.config.ts` is used.

To resolve this issue, I changed the implementation to load `prisma.config.ts` using `jiti`. [ESLint uses a similar approach, whichI used as a reference for this implementation.](https://github.com/eslint/eslint/blob/0fa2b7a3666f1eedcc091446dc860037c9bafa5c/lib/config/config-loader.js#L141-L170)

### DIFF
--- a/packages/config/src/PrismaConfig.ts
+++ b/packages/config/src/PrismaConfig.ts
@@ -15,7 +15,7 @@ type EnvVars = Record<string, string | undefined>
 const sqlMigrationAwareDriverAdapterFactoryShape = <Env extends EnvVars = never>() =>
   Shape.declare(
     (input: any): input is (env: Env) => Promise<SqlMigrationAwareDriverAdapterFactory> => {
-      return input instanceof Function
+      return typeof input === 'function'
     },
     {
       identifier: 'SqlMigrationAwareDriverAdapterFactory<Env>',
@@ -27,7 +27,7 @@ const sqlMigrationAwareDriverAdapterFactoryShape = <Env extends EnvVars = never>
 const errorCapturingSqlMigrationAwareDriverAdapterFactoryShape = <Env extends EnvVars = never>() =>
   Shape.declare(
     (input: any): input is (env: Env) => Promise<ErrorCapturingSqlMigrationAwareDriverAdapterFactory> => {
-      return input instanceof Function
+      return typeof input === 'function'
     },
     {
       identifier: 'ErrorCapturingSqlMigrationAwareDriverAdapterFactory<Env>',
--- a/packages/migrate/src/__tests__/DbExecute.test.ts
+++ b/packages/migrate/src/__tests__/DbExecute.test.ts
@@ -1,7 +1,7 @@
 // describeIf is making eslint unhappy about the test names
 /* eslint-disable jest/no-identical-title */
 
-import { defaultTestConfig, loadConfigFromFile } from '@prisma/config'
+import { defaultTestConfig, PrismaConfigInternal } from '@prisma/config'
 import { jestConsoleContext, jestContext } from '@prisma/get-platform'
 import fs from 'fs'
 import path from 'path'
@@ -24,7 +24,7 @@ describe('db execute', () => {
   describe('using Prisma Config', () => {
     it('--url is not supported', async () => {
       ctx.fixture('prisma-config-validation/sqlite-d1')
-      const config = (await loadConfigFromFile({ configFile: 'prisma.config.ts', configRoot: ctx.fs.cwd() })).config!
+      const config = (await import(`${ctx.fs.cwd()}/prisma.config.ts`)).default as PrismaConfigInternal<any>
 
       try {
         await DbExecute.new().parse(['--url', 'file:./dev.db'], config)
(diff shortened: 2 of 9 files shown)


packages/config/src/PrismaConfig.ts
```diff
@@ -15,7 +15,7 @@ type EnvVars = Record<string, string | undefined>
 const sqlMigrationAwareDriverAdapterFactoryShape = <Env extends EnvVars = never>() =>
   Shape.declare(
     (input: any): input is (env: Env) => Promise<SqlMigrationAwareDriverAdapterFactory> => {
-      return input instanceof Function
+      return typeof input === 'function'
```
nissy-dev, line 18: jiti uses the vm module internally to load files. As a result, the execution context in which the parsing occurs differs from the one in which prisma.config.ts is loaded, causing instanceof Function to return false. To avoid this issue, we modified the logic to use typeof for the type check instead.

packages/migrate/src/__tests__/DbExecute.test.ts
```diff
@@ -24,7 +24,7 @@ describe('db execute', () => {
   describe('using Prisma Config', () => {
     it('--url is not supported', async () => {
       ctx.fixture('prisma-config-validation/sqlite-d1')
-      const config = (await loadConfigFromFile({ configFile: 'prisma.config.ts', configRoot: ctx.fs.cwd() })).config!
```
nissy-dev, line 27: In Jest, loadConfigFromFile cannot be used unless the experimental-vm-modules flag is enabled. However, enabling experimental-vm-modules requires the entire codebase to be compatible with ESM in order to run tests. Therefore, we handle this by loading the config directly instead.

ineshbose:
Looking at the old TODO comment within loadConfigFromFile - is there a reason using just jiti would be preferred over [c12](https://github.com/unjs/c12) (that uses jiti under the hood)?

nissy-dev:
We only need to read the TypeScript config, so jiti should be sufficient. Most of c12's features aren't needed and would just introduce unnecessary dependencies.

FGoessler:
Thanks @nissy-dev! We're aiming to include this in our 6.8.0 release. 🚀

FGoessler (approved):

PR #27078, opened 2025-05-03
feat(client-generator-ts): add getDmmf method to Prisma namespace

### LINKED ISSUE #27028: Get a list of model fields without `Prisma.dmmf` in Prisma 6.7.0
Hello, we'd been using `Prisma.dmmf` to get a list of model fields, like so:

```ts
const modelFields = Prisma.dmmf.datamodel.models.find(
  (m) => m.name === modelName,
)?.fields
```

However, I saw in [this comment](https://github.com/prisma/prisma/pull/26779#issue-2956326639) that `Prisma.dmmf` has been removed.

Is there another way to get a list of model fields without `Prisma.dmmf`? I looked into the Prisma source code, and played around with `getDMMF`, but wasn't able to get it working for my needs.

If not, it would be great to add a method for `Prisma.getDmmf()` as mentioned in the comment.

nissy-dev:
fix #27028


michaelhays:
FYI, it's a little inconsistent, but it looks like the capitalization `getDMMF` is more common than `getDmmf` in the codebase

rhyek:
Looking forward to this so I can finish my version update!

Rusty-Beard:
Is there any update on this?

allhailcheese:
until this is merged i cannot upgrade to prisma-client. any update?

nissy-dev:
@FGoessler @jkomyno @aqrln
It seems that many people would want this feature, and [the original implementation comment](https://github.com/prisma/prisma/pull/26779#issue-2956326639) even mentions that they would consider exporting it if there were enough requests. If you have a moment, I’d appreciate it if you could review it.

> Prisma.dmmf has been removed. Preserving the old behaviour was tricky in pure TS (previously we defined a getter on the the JS object the shape of which was described by the TS namespace declaration) and evaluating it eagerly wasn't nice because it would be wasting CPU cycles on something that most users won't need, so I dropped it for now. If necessary, it can be re-introduced as a plain function (e.g. Prisma.getDmmf()) later.

rhyek:
Of course...

RubenNL:
So, what's gonna happen? will there be another PR for this feature, or is this the end of DMMF?

michaelhays:
They're planning on addressing this with a different approach: https://github.com/prisma/prisma/issues/27028#issuecomment-3265334095

PR #27090, opened 2025-05-05
feat(CLI): add dev subcommand.

igalklebanov:
Hey :wave:

part of STR-5

Adding `prisma dev` to the CLI. Will be maintained in another package.

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -156,6 +156,8 @@ async function main(): Promise<number> {
       debug: DebugInfo.new(),
       // TODO: add rules subcommand to --help after EA
       rules: new SubCommand('@prisma/cli-security-rules'),
+      // TODO: add dev subcommand to --help after it works.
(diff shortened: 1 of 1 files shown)


millsp (approved):

packages/cli/src/bin.ts
```diff
@@ -156,6 +156,8 @@ async function main(): Promise<number> {
       debug: DebugInfo.new(),
       // TODO: add rules subcommand to --help after EA
       rules: new SubCommand('@prisma/cli-security-rules'),
+      // TODO: add dev subcommand to --help after it works.
+      dev: new SubCommand('@prisma/cli-dev'),
```
millsp, line 160: Should it be `@prisma/cli-ppg-dev`?

igalklebanov: @wmadden wdyt?

igalklebanov: decided internally.

FGoessler (approved):

PR #27106, opened 2025-05-07
fix: ORM-924 fix leaking env vars

### LINKED ISSUE #27074: security 6.7.0 prisma migrate update client files with db url value in clear
### Bug description

I just updated my project from prisma 6.6.0 to 6.7.0, my datasource is mysql. I added as suggested in the documentation 

```json
"prisma": {
    "schema": "./prisma"
  }
```

in my package.json

Now, "prisma generate" updates correctly the output folder of the prisma client generated files.
But when I execute "prisma migrate dev", 2 of the generated files are updated with the DB URL set in my .env file in clear:
`edge.js` and `index.js`

Image

I also tried with "prisma migrate dev --schema ./prisma", and the problem is still reproduced.

### Severity

### Reproduction

1. being in v6.6.0
2. add output in the schema.prisma for the generated client files
3. prisma generate
4. prisma migrate db
5. commit
6. update to v6.7.0
7. add prisma section in package.json
8. split schema.prisma into one main schema for the datasource and the models inside the models subfolder
9. prisma generate
10. commit
11. prisma migrate db
12. => check updated generated client files

Sensitive information such as the value of DATABASE_URL written in the .env file must not end up written in clear in the generated files for the prisma client.

### Frequency

Consistently reproducible

### Is this a regression?

Yes, as in 6.6.0 the values of the .env files where not written in the client folder

### Workaround

none found so far

### Prisma Schema & Queries

```prisma
generator client {
  provider = "prisma-client-js"
  output   = "../lib/generated/prisma/client"
}

datasource db {
  provider = "mysql"
}

```

```ts
```

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** macOS
- **Database:** MySQL
- **Node.js version:** 22.14.0

### Prisma Version

```
6.7.0
```

FGoessler:
Due to recent refactoring changes in our env var and schema loading we introduced a bug (see https://github.com/prisma/prisma/issues/27074) where during the code generation step at the end of `prisma migrate dev` values from `env(...)` statements in the schema file could end up in the generated code output.

Note that this was NOT the case during a regular `prisma generate`. => As users are NOT expected to run `prisma migrate dev` with production credentials and are not expected to commit the generated code this is unlikely to have leaked actual production credentials into generated code or source control systems.

This bug was likely introduced in version 6.6.0.

This PR includes an E2E tests that checks our generated output against any content from the provided env var.

Fixes #27074 
Fixes ORM-924

### DIFF
--- a/packages/client/tests/e2e/env-var-security/tests/index.ts
+++ b/packages/client/tests/e2e/env-var-security/tests/index.ts
@@ -1,0 +1,39 @@
+import * as fs from 'fs'
+import * as path from 'path'
+
+const DATABASE_URL = 'file:./secret-database-url-env-var-value.db'
+const FILE_EXTENSION_TO_CHECK = ['.js', '.mjs', '.cjs', '.ts', '.mts', '.cts', '.json', '.prisma', '.d.ts']
+
+function getFilesToCheckInDir(dir: string): string[] {
+  const files: string[] = []
+  const items = fs.readdirSync(dir)
+
+  for (const item of items) {
+    const fullPath = path.join(dir, item)
+    const stats = fs.statSync(fullPath)
+
+    if (stats.isDirectory()) {
+      files.push(...getFilesToCheckInDir(fullPath))
+    } else if (stats.isFile() && FILE_EXTENSION_TO_CHECK.includes(path.extname(fullPath))) {
+      files.push(fullPath)
+    }
+  }
+
+  return files
+}
+
+test('no generated files contain ENV var values', () => {
+  const generatedDir = path.join(process.cwd(), 'generated')
+  const filesToCheck = getFilesToCheckInDir(generatedDir)
+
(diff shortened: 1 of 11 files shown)


packages/client/tests/e2e/env-var-security/tests/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  expect(filesToCheck.length).toBeGreaterThan(0)
+
+  for (const file of filesToCheck) {
+    console.log(`checking file ${file}...: `)
+    const content = fs.readFileSync(file, 'utf8')
+    expect(content).not.toContain(DATABASE_URL)
+  }
+})
+
+export {}
```
aqrln, line 39: Shouldn't be necessary since we have `import` statements

aqrln (approved):

PR #27117, opened 2025-05-07
fix(platform-cli): ORM-925 complete logout also if token expired or invalidated

FGoessler:
Having an expired or invalidated token would give users an error for API interactions. But also won't allow to logout or freshly login. Now we ignore token expired/invalid errors during logout and finish the logout and wipe the local credential so users can login freshly as expected.

### DIFF
--- a/packages/cli/src/platform/auth/logout.ts
+++ b/packages/cli/src/platform/auth/logout.ts
@@ -19,35 +19,48 @@ export class Logout implements Command {
     if (credentials.token) {
       const jwt = decodeJwt(credentials.token)
       if (!isError(jwt) && jwt.jti) {
-        await requestOrThrow<
-          {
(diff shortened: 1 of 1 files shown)


packages/cli/src/platform/auth/logout.ts
```diff
@@ -19,35 +19,48 @@ export class Logout implements Command {
...
+                input: {
+                  id: jwt.jti,
+                },
               },
             },
-          },
-        })
+          })
+        } catch (e) {
+          console.log('FLOOO: >>>', e)
```
aqrln, line 53: stray console.log

FGoessler: ups - thanks!

aqrln (approved):

PR #27123, opened 2025-05-07
feat(cli): enhance SubCommand to accept options for binary target configuration

millsp:


aqrln (approved):

aqrln (commented):
Snapshot tests need to be updated. Maybe also worth adding a test that will have `options` passed.

PR #27127, opened 2025-05-08
fix: fix various invalid results

jacek-prisma:
[ORM-916](https://linear.app/prisma-company/issue/ORM-916/investigate-invalid-result-structure)


packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -89,22 +91,28 @@ function mapValue(value: unknown, resultType: PrismaValueType): unknown {
     case 'Boolean':
       return typeof value === 'boolean' ? value : value !== '0'
     case 'Decimal':
-      return typeof value === 'number' ? value : parseFloat(`${value}`)
+      return typeof value === 'number' ? new Decimal(value) : new Decimal(`${value}`)
```
semoal, line 94: probably big.js is a more lightweight alternative with the same feature, you want here that is precision.

decimal.js : 31.2kb
big.js: 6.8kb

jacek-prisma: the prisma client package already pulls in decimal.js as a dependency, so I think most users will end up with the dependency anyway
https://github.com/prisma/prisma/blob/12313b5a279daf8e0a454a7105b254c0cb670ad7/packages/client/package.json#L263

semoal: Ahh cool in that case, thought that was installed for the new qc 👌

aqrln (approved):

PR #27171, opened 2025-05-13
feat: ORM-941 create credentials store package

FGoessler:
Adds a dedicated package to handle credential storage for our CLI and VSCode extension.

As per the described in https://www.notion.so/prismaio/Technical-Design-10xPPg-VSCode-Extension-1ea9e8aecef780ee94c4e555c8bc568b


packages/credentials-store/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  async getCredentials(): Promise<Credentials[]> {
+    if (this.loadedCredentials === undefined) {
+      await this.reloadCredentialsFromDisk()
+    }
+    return this.loadedCredentials || []
+  }
+
+  async getCredentialsForWorkspace(workspaceId: string): Promise<Credentials | undefined> {
+    return (await this.getCredentials()).filter((c) => c.workspaceId === workspaceId)[0]
```
jacek-prisma, line 67: we're basically using `this.loadedCredentials` like a hash map, maybe we should switch to `Map`

jacek-prisma (approved):

packages/credentials-store/src/index.ts
FGoessler: The more common case (based on what I can tell from the vscode extension work) is that we'll need the list of all workspaces. Then the users chooses one and we'll likely take the existing object from there already rather than looking it up again.
Also the list of workspaces won't be long - not more than 1-5 items. So any datatype optimization we do here basically doesn't matter. Using a list/array though keeps the order "consistent".

jacek-prisma: fair enough

PR #27180, opened 2025-05-14
feat: keep track of the time of first Prisma command

jacek-prisma:
[ORM-975](https://linear.app/prisma-company/issue/ORM-975/nps-improvements)

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -40,6 +40,7 @@ import { Studio } from './Studio'
 import { SubCommand } from './SubCommand'
 import { Telemetry } from './Telemetry'
 import { redactCommandArray, runCheckpointClientCheck } from './utils/checkpoint'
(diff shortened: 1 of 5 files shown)


packages/cli/src/bin.ts
```diff
@@ -161,6 +162,10 @@ async function main(): Promise<number> {
     ['version', 'init', 'migrate', 'db', 'introspect', 'studio', 'generate', 'validate', 'format', 'telemetry'],
   )
 
+  await loadOrInitializeCommandState().catch((err) => {
+    debug(`Failed to initialize the command state: ${err}`)
+  })
```
FGoessler, line 167: Why do we need to add this here and expose this "command state" stuff beyond the NPS survey module? 🤔

jacek-prisma: because we want to keep track of the time any command is issued, to be able to tell when the user started using prisma

FGoessler (approved):

PR #27197, opened 2025-05-16
feat: ORM-934 add delete method to credentials store

FGoessler:
Forgot in my initial implementation that one might want to logout... xD

### DIFF
--- a/packages/credentials-store/src/index.ts
+++ b/packages/credentials-store/src/index.ts
@@ -56,6 +56,14 @@ export class CredentialsStore {
     await writeFile(this.authFilePath, JSON.stringify(data, null, 2))
   }
 
+  async deleteCredentials(workspaceId: string): Promise<void> {
+    await this.reloadCredentialsFromDisk()
+    this.loadedCredentials = this.loadedCredentials?.filter((c) => c.workspaceId !== workspaceId) || []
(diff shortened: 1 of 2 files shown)


packages/credentials-store/src/index.ts
```diff
@@ -56,6 +56,14 @@ export class CredentialsStore {
     await writeFile(this.authFilePath, JSON.stringify(data, null, 2))
   }
 
+  async deleteCredentials(workspaceId: string): Promise<void> {
+    await this.reloadCredentialsFromDisk()
+    this.loadedCredentials = this.loadedCredentials?.filter((c) => c.workspaceId !== workspaceId) || []
+    const data: AuthFile = { tokens: this.loadedCredentials }
+    await mkdir(path.dirname(this.authFilePath), { recursive: true })
+    await writeFile(this.authFilePath, JSON.stringify(data, null, 2))
```
jacek-prisma, line 64: this is duplicated twice, might be worth putting in a function of its own

jacek-prisma (approved):

packages/credentials-store/src/index.ts
FGoessler: Good call - done :+1:

PR #27199, opened 2025-05-16
fix: use absolute path file URL for subcommand loading

### LINKED ISSUE #27195: 6.8.1 hotfix init error still exists
### Bug description

As per #27192 with v6.8.0. I had the same problem and even with the v6.8.1 the issue still exists.

npx prisma init
Error: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'

Is this a nextjs issue? or something else?

node -v 22.15.1
npm -v 11.3.0

### Severity

### Reproduction

npx create-next-app@latest
npm -D i prisma 
npx prisma init

Unable to init. Even manually creating the folder and schema it still get's error when trying to migrate.

### Frequency

Consistently reproducible

### Is this a regression?

Just this one and 6.8.0

### Prisma Schema & Queries

```prisma
```
```ts
```

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** 
- **Database:** 
- **Node.js version:** 

### Prisma Version

```
// Prisma version output
```

aqrln:
Prior to 6.8.1, the subcommand loading mechanism was failing on Windows
because it was using an absolute path without the `file://` protocol,
making the disk name looks like a protocol and thus the whole path
becoming an invalid URL.

In 6.8.1, this was worked around by using a relative path, but this
again broke on Windows if the CLI installation path and the subcommand
temporary directory are on different drives.

The correct fix for the issue is to use a `file://` URL with the
absolute path to the subcommand. This was not done in 6.8.1 because it
broke the tests.

However, the reason that broke the tests was not in our code, but in
Jest itself, which has atrocious ESM support and doesn't know how to
handle the `file://` URLs.

This commit fixes the issue by using a `file://` URL for the absolute
path, and changes the `SubCommand.test.ts` test to run under Vitest
instead of Jest. Going forward, we can gradually migrate the rest of the
tests to Vitest.

Fixes: https://github.com/prisma/prisma/issues/27195
Closes: https://linear.app/prisma-company/issue/ORM-983/bug-subcommand-esm-module-resolution-fails-on-windows


packages/cli/helpers/run-tests.ts
```diff
@@ -23,7 +23,16 @@ export async function main() {
     process.env.NODE_EXTRA_CA_CERTS = miniProxy.defaultCertificatesConfig.caCert
   }
 
-  execa.sync('jest', ['--silent', ...process.argv.slice(2)], {
+  // We pass the --passWithNoTests flag to jest and vitest for them not to fail
+  // when the filter provided via CLI only matches tests in one of the two suites.
```
wmadden, line 27: It might be better to use a more exclusive file filter, e.g. `subcommand.vitest.ts`

wmadden (approved):
Changes look good 👍🏻

packages/cli/helpers/run-tests.ts
aqrln: We'd need this flag regardless, because it's exactly when only vitest or only jest matches, the other would fail by default.

But I like the idea of using `.vitest.ts` suffix for tests that should be running under Vitest instead of harcoding the specific file names in `jest.config.js` and `vitest.config.ts` 👍

FGoessler (approved):

PR #27211, opened 2025-05-19
fix: execute seq sequentially

jacek-prisma:
[ORM-948](https://linear.app/prisma-company/issue/ORM-948/fix-set-for-nested-records)

Required to fix a number of nested set tests in QE

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
+++ b/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
@@ -69,12 +69,11 @@ export class QueryInterpreter {
   ): Promise<Value> {
     switch (node.type) {
       case 'seq': {
-        const results: Value[] = []
+        let result: Value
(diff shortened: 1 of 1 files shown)


packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
```diff
@@ -69,7 +69,11 @@ export class QueryInterpreter {
   ): Promise<Value> {
     switch (node.type) {
       case 'seq': {
-        const results = await Promise.all(node.args.map((arg) => this.interpretNode(arg, queryable, scope, generators)))
+        const results: Value[] = []
```
aqrln, line 72: We don't actually need to store the array, we only need to return the last value.

jacek-prisma: true, updated the code

aqrln (approved):

PR #27229, opened 2025-05-21
fix: various batching fixes

jacek-prisma:
[ORM-918](https://linear.app/prisma-company/issue/ORM-918/transactional-batching)

Fixes all important batching tests.
Some notes:
- `batching raw rollback` works, but it's ignored because we generate a different (better) error message than the driver adapter and it uses a shared snapshot
- `invalid level generates run- and compile- time error` fails because of a problem related to isolation levels, not batching
- other batching tests that are failing are failing because they use raw queries


packages/client-engine-runtime/src/utils.ts
```diff
@@ -21,3 +24,30 @@ export function isDeepStrictEqual(a: unknown, b: unknown): boolean {
...
+  const lhsKeys = Object.keys(lhs)
+  const rhsKeys = Object.keys(rhs)
+  const smallerKeyList = lhsKeys.length < rhsKeys.length ? lhsKeys : rhsKeys
+
+  return smallerKeyList.every((key) => {
+    if (typeof lhs[key] !== typeof rhs[key]) {
+      if (typeof lhs[key] === 'number' || typeof rhs[key] === 'number') {
+        return `${lhs[key]}` === `${rhs[key]}`
+      } else if (lhs[key] instanceof Date || rhs[key] instanceof Date) {
+        return new Date(`${lhs[key]}`).getTime() === new Date(`${rhs[key]}`).getTime()
```
aqrln, line 43: Why re-convert them to dates if we already checked they are?

jacek-prisma: @aqrln only one of them is a Date

packages/client-engine-runtime/src/utils.ts
```diff
@@ -21,3 +24,30 @@ export function isDeepStrictEqual(a: unknown, b: unknown): boolean {
...
+      } else if (lhs[key] instanceof BigInt || rhs[key] instanceof BigInt) {
```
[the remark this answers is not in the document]
jacek-prisma: good catch, this one wasn't covered by a test so it wasn't caught, updated it

packages/client-engine-runtime/src/utils.ts
jacek-prisma: @aqrln we know only one of them is a Date, because of the previous `typeof lhs[key] !== typeof rhs[key]`

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -419,10 +419,12 @@ export class ClientEngine implements Engine<undefined> {
...
         case 'multi': {
           results = await Promise.all(
-            batchResponse.plans.map(async (plan, i) => {
-              const rows = await interpreter.run(plan as QueryPlanNode, queryable)
-              return { data: { [queries[i].action]: rows } }
-            }),
+            batchResponse.plans.map((plan, i) =>
+              interpreter.run(plan as QueryPlanNode, queryable).then(
+                (rows) => ({ data: { [queries[i].action]: rows } }),
+                (err) => err,
```
aqrln, line 425: Do we actually handle the errors returned this way somewhere? Should we maybe use `Promise.allSettled` instead of `Promise.all`?

jacek-prisma: these errors are handled in `DataLoader.ts`, this change fixes the `correctly reports location of a batch error` test

packages/client/tests/functional/extensions/model.ts
```diff
@@ -414,6 +413,10 @@ testMatrix.setupTestSuite(
           if (isSqlServer) {
             expectation.unshift([{ query: expect.stringContaining('SET TRANSACTION') }])
           }
+          if (cliMeta.engineType !== 'client') {
+            // Client engine issues COMMIT directly from the TransactionManager.
```
aqrln, line 417: It still must be logged though. `TransactionManager` should accept an `onQuery` callback and call it whenever it executes a query, just like `QueryInterpreter` does.

jacek-prisma: maybe, but I think that's out of the scope of the PR, I can make it a todo

jacek-prisma: I guess I could just leave the test failing 🤔 it's good to at least have confirmed that it works with this change

aqrln: leaving the test failing and fixing this in a separate PR sounds good to me

aqrln: actually it's fine, I've added a task so we won't forget either way, you can leave it as is

aqrln (approved):

PR #27238, opened 2025-05-22
fix: fix raw queries

jacek-prisma:
[ORM-1001](https://linear.app/prisma-company/issue/ORM-1001/correct-the-serializationdeserialization-of-raw-queries-so-that-its)

The mapping is the combining of 2 pieces of code in Rust:
https://github.com/prisma/prisma-engines/blob/c504ff10504e09dba57f0439835ce4b8efefbcdb/quaint/src/connector/column_type.rs#L58-L95
+
https://github.com/prisma/prisma-engines/blob/c504ff10504e09dba57f0439835ce4b8efefbcdb/libs/driver-adapters/src/types.rs#L114-L245

### DIFF
--- a/packages/client/tests/functional/extensions/query.ts
+++ b/packages/client/tests/functional/extensions/query.ts
@@ -532,7 +532,6 @@ testMatrix.setupTestSuite(
           const expectation = [
             [{ query: expect.stringContaining('SELECT') }],
             [{ query: expect.stringContaining('SELECT') }],
-            [{ query: expect.stringContaining('COMMIT') }],
           ]
           if (driverAdapter === undefined) {
             // Driver adapters do not issue BEGIN through the query engine.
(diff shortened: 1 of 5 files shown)


packages/client/tests/functional/extensions/query.ts
```diff
@@ -541,6 +540,10 @@ testMatrix.setupTestSuite(
           if (isSqlServer) {
             expectation.unshift([{ query: expect.stringContaining('SET TRANSACTION') }])
           }
+          if (cliMeta.engineType !== 'client') {
+            // Client engine issues COMMIT directly from the TransactionManager.
```
aqrln, line 544: fixed in https://github.com/prisma/prisma/pull/27235

aqrln (approved):

PR #27239, opened 2025-05-22
fix: make data mapper return tagged values in JSON protocol format

aqrln:
Port much of the [prisma-engines driver adapters data mapping logic](https://github.com/prisma/prisma-engines/blob/e17d3735a57aa4748b05b824e366c69cf2cc774f/libs/driver-adapters/src/conversion/js_to_quaint.rs#L68-L297) to `DataMapper` and return non-primitive values in the JSON protocol format, to be decoded later in `deserializeJsonResponse`.

Closes: https://linear.app/prisma-company/issue/ORM-1002/fix-serializationdeserialization-for-regular-queries-in-prismaprisma

/engine-branch push-xtstzzwznooo


packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -90,39 +90,130 @@ function mapObject(data: PrismaObject, fields: Record<string, ResultNode>): Pris
...
     case 'Bytes': {
-      if (!Array.isArray(value)) {
-        throw new DataMapperError(`Bytes data is invalid, got: ${typeof value}`)
+      if (typeof value === 'string') {
+        return { $type: 'Bytes', value }
       }
-      return new Uint8Array(value)
+      if (Array.isArray(value) || value instanceof Uint8Array) {
+        return { $type: 'Bytes', value: Buffer.from(value).toString('base64') }
+      }
```
jacek-prisma, line 200: do these cases ever happen? I think driver adapters always return bytes as an array of numbers

I'd remove the cases that are not reachable

packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -90,39 +90,130 @@ function mapObject(data: PrismaObject, fields: Record<string, ResultNode>): Pris
...
+          throw new DataMapperError(`Expected an integer in column '${columnName}', got string: ${value}`)
+        }
+
+        default:
+          throw new DataMapperError(`Expected an integer in column '${columnName}', got ${typeof value}: ${value}`)
+      }
+    }
+
+    case 'BigInt': {
+      if (typeof value !== 'bigint' && typeof value !== 'number' && typeof value !== 'string') {
```
jacek-prisma, line 140: I think driver adapters can't return an actual BigInt

jacek-prisma:
I see this includes a number of cases that might've been ported from engines, I think it might be worth trying to simplify and only handle types that driver adapters are allowed return, I'd try to remove some of the cases and see if the tests still pass

packages/client-engine-runtime/src/interpreter/DataMapper.ts
aqrln: Good question, I'm not sure.

packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -90,39 +90,135 @@ function mapObject(data: PrismaObject, fields: Record<string, ResultNode>): Pris
...
+      }
+      throw new DataMapperError(`Expected a byte array in column '${columnName}', got ${typeof value}: ${value}`)
     }
+
     default:
       assertNever(resultType, `DataMapper: Unknown result type: ${(resultType as PrismaValueType).type}`)
   }
 }
+
+const TIMEZONE_PATTERN = /Z$|[+-]\d{2}:?\d{2}$/
```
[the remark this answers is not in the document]
jacek-prisma: https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-OUTPUT

jacek-prisma (approved):

PR #27260, opened 2025-05-26
feat(CLI): allow downstream commands to handle shutdowns gracefully.

igalklebanov:
Hey :wave:

closes STR-18

This PR allows for downstream CLI commands to handle shutdowns gracefully, while still preserving the original decision to standardize on `130` when receiving `SIGINT`.

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -55,9 +55,14 @@ const commandArray = process.argv.slice(2)
 
 process.removeAllListeners('warning')
 
-// Listen to Ctr + C and exit
 process.once('SIGINT', () => {
   process.exitCode = 130
+
+  // no further downstream listeners for SIGINT, exit immediately with the code set above.
+  if (process.listenerCount('SIGINT') === 0) {
+    process.exit()
+  }
+  // otherwise, let the downstream listeners handle it.
 })
 
 // Parse CLI arguments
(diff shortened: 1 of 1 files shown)


packages/cli/src/bin.ts
```diff
@@ -57,7 +57,7 @@ process.removeAllListeners('warning')
 
 // Listen to Ctr + C and exit
```
aqrln, line 58: The comment is outdated now

packages/cli/src/bin.ts
```diff
@@ -57,7 +57,7 @@ process.removeAllListeners('warning')
...
 process.once('SIGINT', () => {
-  process.exit(130)
+  process.exitCode = 130
```
aqrln, line 60: This will prevent the CLI from exiting at all when _not_ using subcommands with their own sigint handlers, at least when running it directly and not in npm script. Should we maybe check if there are any other handlers, and exit if there are none?

```suggestion
  if (process.listenerCount('SIGINT') > 0) {
    process.exitCode = 130
  } else {
    process.exit(130)
  }
```

(comparing to 0 and not to 1 because this event handler is attached with `once` and not `on`, and the count will have already been decremented by the time the callback is invoked)

igalklebanov: Great idea to preserve original behavior for downstream that doesn't listen as well!

I have pushed a slightly different variation to what you suggested. 
`process.exit()` exits with `process.exitCode` if it was set before.

aqrln (approved):
LGTM!

PR #27311, opened 2025-06-02
feat(cli): skip downloading Query Engine when Query Compiler is turned on, and Schema Engine when config.migrate.adapter is set

jkomyno:
This PR:
- closes [ORM-1005](https://linear.app/prisma-company/issue/ORM-1005/make-sure-qe-binary-is-not-downloaded-when-using-qc)
- in `packages/engines`:
  - gets rid of the old `ensureBinariesExist`, replacing it with `ensureNeededBinariesExist`
  - add first test, using `vitest`.
- in `packages/cli`:
  - add new `getClientGeneratorInfo` utility, to safely extract info (preview features' list, engine type) from Prisma's official client generator(s) PSL block
  - combine `ensureBinariesExist` with `getClientGeneratorInfo` in the `CLI` barrel command file, ensuring no stdout regression

After this PR:
 - we no longer attempt downloading Query Engine binaries when the client generator has the `queryCompiler` feature (which implies `engineType` is inferred as `client`):
  ```prisma
  // ./prisma/schema.prisma

  generator client {
    output          = "../generated/client"
    previewFeatures = ["queryCompiler"]
  }
  ```
- we no longer attempt downloading Schema Engine binary when `prisma.config.ts` has a `migrate.adapter` set:
  ```typescript
  // ./prisma.config.ts
  
  import 'dotenv/config'
  
  export default defineConfig({
    earlyAccess: true,
    schema: './prisma/schema.prisma',
    migrate: {
      // @ts-ignore
      adapter: async () => {
        return Promise.resolve({ ... })
      },
    }
  })
  ```
  This was necessary, as during the manual QA, I realised that StackBlitz always fails with "socket hang up" errors when downloading Prisma binaries.

Note: I had previously attempted to re-use the `vitest.config.ts` file that was originally created in `packages/cli`, and install `vitest` only once, at the workspace level. I reverted that in https://github.com/prisma/prisma/pull/27311/commits/964aae7b37691a1863b579952236596abaed189f, as I realised that Vitest tests are sometimes stored in `*.vitest.ts`, and other times in `.test.ts`, depending on whether Jest is also run in the same package directory. Until we completely get rid of Jest, there is little we can do to share a single `vitest.config.ts` file.


packages/cli/src/CLI.ts
```diff
@@ -1,11 +1,16 @@
 import type { PrismaConfigInternal } from '@prisma/config'
-import { ensureBinariesExist } from '@prisma/engines'
+import { Debug } from '@prisma/debug'
+import { ensureNeededBinariesExist } from '@prisma/engines'
+import { download } from '@prisma/fetch-engine'
```
jkomyno: `download` is passed explicitly to `ensureNeededBinariesExist` to mock and perform assertions on it in `packages/engines/ensureNeededBinariesExist.test.ts`

packages/engines/src/__tests__/ensureNeededBinariesExist.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      await ensureNeededBinariesExist({
+        clientEngineType: 'client',
+        previewFeatures: [],
+        download,
+      })
+      expect(download).toHaveBeenCalledWith(
+        expect.objectContaining({
+          binaries: {
+            'schema-engine': expect.any(String),
+            'libquery-engine': expect.any(String),
```
jkomyno: We care about the keys (type of engine to download), not the values (i.e., the paths on disk).

packages/migrate/src/commands/DbPull.ts
```diff
@@ -154,6 +154,7 @@ Set composite types introspection depth to 2 levels
 
     // Print to console if --print is not passed to only have the schema in stdout
     if (schemaContext && !args['--print']) {
+      // TODO: the following line is likely redundant
```
jkomyno, line 157: Non-related thing spotted during PR.

packages/bundle-size/create-gzip-files.ts
```diff
@@ -34,7 +34,13 @@ void (async () => {
 
     // Install deps & copy schema & generate Prisma Client
     await $`cp ${getSchemaFile(project)} ${projectDir}/schema.prisma`
-    await $`pnpm prisma generate --schema=${projectDir}/schema.prisma`
+    await $`pnpm prisma generate --schema=${projectDir}/schema.prisma`.catch((error) => {
+      const e = error as Error
+      console.error(
+        `Failed to generate Prisma Client from ${getSchemaFile(project)} (copied to ${projectDir}/schema.prisma)`,
+      )
+      throw e
```
aqrln, line 42: Doesn't look like `e` is being used in any way in the error message

```suggestion
      console.error(
        `Failed to generate Prisma Client from ${getSchemaFile(project)} (copied to ${projectDir}/schema.prisma)`,
      )
      throw error
```

packages/cli/src/CLI.ts
```diff
@@ -34,16 +39,31 @@ export class CLI implements Command {
...
     }
 
+    // Extract client generator info once, use it for either `prisma --version`, or
+    // for any other supported command. If no client generator is successfully found,
+    // use sensible default values.
+    const { previewFeatures, engineType } = await getClientGeneratorInfo({
+      schemaPathFromConfig: config.schema,
+      schemaPathFromArg: args['--schema'],
+    }).catch((error) => {
+      const e = error as Error
```
aqrln, line 76: similarly, `e` is not used as `Error` (i.e. `message`/`stack` fields aren't accessed) so this type assertion and the extra variable serve no purpose

packages/engines/package.json
```diff
@@ -17,8 +17,8 @@
     "@types/jest": "29.5.14",
     "@types/node": "18.19.76",
     "execa": "5.1.1",
-    "jest": "29.7.0",
-    "typescript": "5.4.5"
+    "typescript": "5.4.5",
+    "vitest": "3.2.0"
```
aqrln, line 21: nice!

packages/engines/src/index.ts
```diff
@@ -13,38 +14,52 @@ export const DEFAULT_CLI_QUERY_ENGINE_BINARY_TYPE = BinaryType.QueryEngineLibrar
...
 
   const binaries = {
-    [cliQueryEngineBinaryType]: binaryDir,
+    // Schema engine is always needed, as `SchemaEngineWasm` is currently not gated behind a preview feature
     [BinaryType.SchemaEngineBinary]: binaryDir,
   }
+
+  // query engine should only be downloaded if the queryCompiler preview feature is not enabled, or if
+  // QE is enabled explicitly by specifying the engineType generator property.
+  const usesQueryCompiler = previewFeatures.includes('queryCompiler') && clientEngineType === 'client'
```
aqrln: `clientEngineType === 'client'` implies `previewFeatures.includes('queryCompiler')` so it should be enough to just check the `clientEngineType`

aqrln (approved):

packages/fetch-engine/src/download.ts
```diff
@@ -65,6 +65,11 @@ type BinaryDownloadJob = {
 }
 
 export async function download(options: DownloadOptions): Promise<BinaryPaths> {
+  // no need to do anything, if there are no binaries
+  if (!options.binaries || Object.values(options.binaries).length === 0) {
+    return {} // we don't download anything if nothing is provided
+  }
```
jkomyno, line 71: Moved this to the top, it doesn't make sense to run any logic before this check runs.

packages/cli/src/__tests__/commands/CLI.test.ts
```diff
@@ -1,84 +1,213 @@
...
-it('unknown command', async () => {
-  await expect(cliInstance.parse(['doesnotexist'], defaultTestConfig())).resolves.toThrow()
-})
+    beforeAll(() => {
+      process.env = {
+        ...originalEnv,
+        PRISMA_CLI_QUERY_ENGINE_TYPE: undefined,
+        PRISMA_CLIENT_ENGINE_TYPE: undefined,
+      }
+    })
```
jkomyno, line 67: I'm nuking variables set by the CI, as they override logic we need to check (which reads the client engine type).

PR #27318, opened 2025-06-03
fix(client): build edge runtime for query compiler

aqrln:
Closes: https://linear.app/prisma-company/issue/ORM-1016/client-runtime-is-not-built-for-qc-edge-combination

/integration

### DIFF
--- a/helpers/compile/plugins/fill-plugin/fillPlugin.ts
+++ b/helpers/compile/plugins/fill-plugin/fillPlugin.ts
@@ -139,10 +139,11 @@ const defaultFillersConfig: Fillers = {
   path: { imports: path.join(__dirname, 'fillers', 'path.ts') },
   tty: { imports: path.join(__dirname, 'fillers', 'tty.ts') },
   util: { imports: path.join(__dirname, 'fillers', 'util.ts') },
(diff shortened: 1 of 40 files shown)


jkomyno (approved):

helpers/compile/plugins/fill-plugin/fillPlugin.ts
```diff
@@ -139,10 +139,11 @@ const defaultFillersConfig: Fillers = {
   path: { imports: path.join(__dirname, 'fillers', 'path.ts') },
   tty: { imports: path.join(__dirname, 'fillers', 'tty.ts') },
   util: { imports: path.join(__dirname, 'fillers', 'util.ts') },
+  crypto: { imports: path.join(__dirname, 'fillers', 'crypto.ts') },
+  'node:crypto': { imports: path.join(__dirname, 'fillers', 'crypto.ts') },
```
jkomyno, line 143: I hope we can get rid of this in the future, and use `globalThis.crypto` only throughout our codebase

aqrln: I had to add this not because of our code but because of third-party dependencies (cuid and uuid generators)

PR #27328, opened 2025-06-04
feat: MySQL update support

jacek-prisma:
[ORM-604](https://linear.app/prisma-company/issue/ORM-604/support-mysql-updates)

/engine-branch feat/mysql-update-support

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
+++ b/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
@@ -1,7 +1,7 @@
 import { SqlQuery, SqlQueryable, SqlResultSet } from '@prisma/driver-adapter-utils'
 
 import { QueryEvent } from '../events'
-import { JoinExpression, Pagination, QueryPlanNode } from '../QueryPlan'
+import { FieldInitializer, FieldOperation, JoinExpression, Pagination, QueryPlanNode } from '../QueryPlan'
 import { type TracingHelper, withQuerySpanAndEvent } from '../tracing'
 import { type TransactionManager } from '../transactionManager/TransactionManager'
 import { rethrowAsUserFacing } from '../UserFacingError'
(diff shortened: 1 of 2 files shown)


packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
```diff
@@ -429,3 +438,41 @@ function paginate(list: {}[], { cursor, skip, take }: Pagination): {}[] {
...
+    case 'set':
+      return evaluateParam(op.value, scope, generators)
+    case 'add':
+      return asNumber(value) + asNumber(evaluateParam(op.value, scope, generators))
+    case 'subtract':
+      return asNumber(value) - asNumber(evaluateParam(op.value, scope, generators))
+    case 'multiply':
+      return asNumber(value) * asNumber(evaluateParam(op.value, scope, generators))
+    case 'divide':
+      return asNumber(value) / asNumber(evaluateParam(op.value, scope, generators))
```
aqrln: (not blocking)
Do we want to check if the second operand is zero and throw an error?

aqrln (approved):

packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
jacek-prisma: had a look at the Rust code and it'd panic there if the left operand is an int , but not if it's a float, not sure if it's a good reference to use 
also checked MySQL behavior and it seems to always return NULL for division by zero regardless of the data type - I think we should do the same since we're emulating a database update here

jacek-prisma: tried sqlite and postgres, the standard behavior seems to be to return null for division by zero for any data type with one exception - postgres raises an error when dividing an `INT` by zero

jacek-prisma: even if we do have this expression in postgres (and we currently won't), postgres raising an error would make this code unreachable either way!
so I think returning `null` for division by zero is the best

jacek-prisma: wdyt @aqrln ?

aqrln: makes sense, let's return null to be consistent with the database behaviour then

PR #27331, opened 2025-06-04
feat(client): skip download in `getGenerators` when `queryCompiler` is turned on

jkomyno:
This PR:
- closes [ORM-1028](https://linear.app/prisma-company/issue/ORM-1028/dont-download-binaries-in-prisma-generate-when-querycompiler-and)


packages/internals/src/get-generators/getGenerators.ts
```diff
@@ -313,10 +285,20 @@ generator gen {
         }
       }
     }
+
+    const queryEngineBinaryType = getCliQueryEngineBinaryType()
+    const queryEngineType = binaryTypeToEngineType(queryEngineBinaryType)
```
aqrln, line 290: I think these two lines and the `if` statement in which they are used can probably also be deleted (generating the DMMF has nothing to do with engines anymore so this is either dead code or it runs and does nothing useful because it produces the exact same DMMF). Fine to do this in a follow up PR if you'd prefer.

sandbox/basic-sqlite/.generated/prisma/client.ts
```diff
@@ -0,0 +1,45 @@
+
```
aqrln: Unintentionally committed

sandbox/basic-sqlite/prisma/schema.prisma
```diff
@@ -1,31 +1,14 @@
 generator client {
```
aqrln, line 1: These changes probably also weren't supposed to be committed

packages/internals/src/get-generators/getGenerators.ts
jkomyno: Let's do it in another PR, so that ecosystem-tests can run on prisma before and after this proposed change.

sandbox/basic-sqlite/.generated/prisma/client.ts
jkomyno: Intentionally committed 😛

It was useful to locally test the behavior of Prisma with `DEBUG="prisma:*"` turned on. If you prefer, I can move https://github.com/prisma/prisma/pull/27331/commits/173eb1d5dcf58d36765885616be415b338a3fe51 to another PR, but it would mean wasting time imho

sandbox/basic-sqlite/prisma/schema.prisma
jkomyno: See above

sandbox/basic-sqlite/.generated/prisma/client.ts
aqrln: why commit the generated client then?

jkomyno: I'll get rid of the generated client

jkomyno: There was a typo in the .gitignore I had edited. I literally missed a "." before `generated`. Fixed now.

aqrln (approved):

PR #27346, opened 2025-06-06
feat: relation join support flag

jacek-prisma:
[ORM-1038](https://linear.app/prisma-company/issue/ORM-1038/address-remaining-planetscale-test-failures)


packages/adapter-pg-worker/src/pg.ts
```diff
@@ -160,6 +160,7 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
   getConnectionInfo(): ConnectionInfo {
     return {
       schemaName: this.options?.schema,
+      supportsRelationJoins: true,
```
jkomyno, line 163: Please don't touch this package, it caused Git conflicts on my PR already

jkomyno (approved):

packages/adapter-pg-worker/src/pg.ts
jacek-prisma: without this change it no longer compiles!

PR #27358, opened 2025-06-07
feat(CLI): anonymously track all commands, regardless of success/failure.

igalklebanov:
Hey :wave:

closes EXP-1552

This PR moves the checkpoint client invocation to before CLI command invocation, in order to track command attempts, regardless of success/failure.

This is a decision that originates from the need to track `prisma dev` invocations, but the command being a long-running process that exits on user input is making the current tracking method not viable.

Other methods were discussed, including subcommands using a config-passed `onSuccess` callback to call when they're considered successful. This adds an opt-in mechanism which adds complexity and maintenance burden without much value.

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -182,6 +182,15 @@ async function main(): Promise<number> {
     return 1
   }
 
+  const checkResult = await runCheckpointClientCheck({
+    command: redactedCommandAsString,
+    isPrismaInstalledGlobally,
+    schemaPath: args['--schema'],
+    schemaPathFromConfig: config.schema,
+    telemetryInformation: args['--telemetry-information'],
+    version: packageJson.version,
+  })
+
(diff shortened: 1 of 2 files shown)


FGoessler (approved):

packages/cli/src/bin.ts
```diff
@@ -182,6 +182,15 @@ async function main(): Promise<number> {
     return 1
   }
 
+  const checkResult = await runCheckpointClientCheck({
```
FGoessler: Can we just kick this off here, but await only later? So we won't block user interaction for network requests or so?

igalklebanov: Good idea, check [`b8e5ff0` (#27358)](https://github.com/prisma/prisma/pull/27358/commits/b8e5ff06d490309c91928b058fd5c2e237dc28e2).

FGoessler (approved):

Issue #27163, opened 2025-05-13
Prisma 6.7.0 Schema  lookup change


jasonlewicki:
### Bug description

Around 6.6 the linter is starting to complain about schema files in the migrations folder that are being picked up by the linter. I'm using a data-migrations package https://github.com/Softjey/prisma-dm that packs up the schema and bundles it in the migration folder to use as migration specific prisma clients.

The errors are about duplicate generators, tables, and such.

i've tried specifying a schema in package.json (which I didn't have before and worked without)

I'm not sure how to get the functionality that I had before which was this:

```
/migrations
  /923989123_1
        /migration.sql
        /schema.prisma
/schema
  /schema.prisma
  /account.prisma
  /user.prisma
```
and not have the linter explode that schemas in the migrations folder are interfering and viceversa. I would have thought migrations would have been ignored.

### Severity

⚠️ Major: Breaks core functionality (e.g., migrations fail)

### Reproduction

having a schema file in the migrations folder causes this issue.

### Expected vs. Actual Behavior

linting should not be complaining.

### Frequency

Consistently reproducible

### Does this occur in development or production?

Both development and production

### Is this a regression?

this worked on 5.18.0 with `previewFeatures = ["prismaSchemaFolder"]` specified and now with it removed and on 6.7.0 does not work

### Workaround

downgrading to 5.18.0

### Prisma Schema & Queries

not needed

### Prisma Config

i don't have one.

### Logs & Debug Info

none

### Environment & Setup

doesn't matter

### Prisma Version

not applicable

jasonlewicki:
Agentic mode responded with this (I have not tested yet, will try later this week):

GitHub Copilot: This issue is caused by the Prisma linter picking up `schema.prisma` files inside your `migrations` folders, which leads to duplicate model/generator errors. This is a known pain point after the changes in Prisma 6.x, especially with the removal of the `prismaSchemaFolder` preview feature.

**Why this happens:**  
Prisma's linter and tooling now recursively look for `.prisma` files, including those in `migrations/**`, unless explicitly told otherwise. This causes conflicts when you have multiple schema files (like those bundled by `prisma-dm`) in your migration directories.

---

## Solutions

### 1. Exclude Migration Schemas from Linting

If you are using ESLint with a Prisma plugin, you can configure it to ignore the `migrations/**/schema.prisma` files.  
Add or update your `.eslintignore` file at the root of your project:

```
**/migrations/**/schema.prisma
```

[...]

PR #27424, opened 2025-06-16
feat: ORM-663 / ORM-664 add directory config options

FGoessler:
This PR adds config options via `prisma.config.ts` for:
- migrations directory
- views directory
- typed-sql directory

```ts
// prisma.config.ts
export default defineConfig({
  earlyAccess: true,
  viewsDirectory: './custom/views',
  typedSqlDirectory: './custom/sql',
  migrate: {
    migrationsDirectory: './custom/migrations',
  },
})
```

This gives more flexibility to our users to structure their project. 

It also provides a workaround for users where their migrations directory contains schema files but collides with the prisma-schema-folder setup that then accidentally picks up additional schema files from the migration directory. See https://github.com/prisma/prisma/issues/27163.

Closes ORM-663
Closes ORM-664

### DIFF
--- a/packages/config/src/PrismaConfig.ts
+++ b/packages/config/src/PrismaConfig.ts
@@ -116,6 +116,8 @@ const createPrismaConfigShape = <Env extends EnvVars = never>() =>
   Shape.Struct({
     earlyAccess: Shape.Literal(true),
     schema: Shape.optional(Shape.String),
+    viewsDirectory: Shape.optional(Shape.String),
+    typedSqlDirectory: Shape.optional(Shape.String),
     studio: Shape.optional(createPrismaStudioConfigShape<Env>()),
     migrate: Shape.optional(createPrismaMigrateConfigShape<Env>()),
   })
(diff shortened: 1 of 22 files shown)


packages/config/src/PrismaConfig.ts
```diff
@@ -127,6 +135,14 @@ export type PrismaConfig<Env extends EnvVars = never> = {
    * The path to the schema file or path to a folder that shall be recursively searched for .prisma files.
    */
   schema?: string
+  /**
+   * The path to the directory where view definition files are stored.
+   */
+  viewsDirectory?: string
```
aqrln, line 141: Why is this not part of the Migrate config?

FGoessler: Because creating views is part of the introspection logic of `db pull`. Not of the "migration" behavior.

FGoessler: Discussion on Slack: https://prisma-company.slack.com/archives/C058VM009HT/p1750754049304759

PR #27436, opened 2025-06-17
test(client): fix running e2e tests in parallel

aqrln:
When running tests locally, unless the `--maxWorkers` option was passed, all tests were started at the same time. This caused most of them to fail because Docker ran out of IP address ranges to assign to the networks created by Docker Compose.

This commit changes the default value of `--maxWorkers` from `Infinity` to `os.cpus().length`, and moves the hardcoded alternative default that was applied when `CI=true` environment variable was set to the CLI command used in the CI workflow as an explicit option.

Additionally, the pooling implementation was inefficient, it was unnecessarily looping every 100 ms to check if the next iteration can start, instead of starting as soon as it becomes possible and not doing anything before that. I refactored it to a proper semaphore.

### DIFF
--- a/packages/client/tests/e2e/_utils/run.ts
+++ b/packages/client/tests/e2e/_utils/run.ts
@@ -1,3 +1,4 @@
+import os from 'node:os'
 import { finished } from 'node:stream/promises'
 
 import { arg } from '@prisma/internals'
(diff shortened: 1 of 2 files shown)


packages/client/tests/e2e/_utils/run.ts
```diff
@@ -142,17 +143,15 @@ async function main() {
...
     for (const [i, job] of dockerJobs.entries()) {
-      while (availableWorkers === 0) {
-        await new Promise((resolve) => setTimeout(resolve, 100))
-      }
+      await semaphore.acquire()
 
-      --availableWorkers // borrow worker
       const pendingJob = (async () => {
         console.log(`💡 Running test ${i + 1}/${dockerJobs.length}`)
         jobResults.push(Object.assign(await job(), { name: e2eTestNames[i] }))
```
jacek-prisma, line 153: I guess this assumes that job never throws? i'd put `semaphore.release()` in a `finally` block just to make sure we never end up hanging forever waiting for a job that failed to release it

aqrln: @jacek-prisma done

jacek-prisma (approved):

PR #27439, opened 2025-06-17
feat(cli): TypeScript implementation for db pull

jharrell:
**NOTE: I heavily used Claude Code to write this code. Here be dragons.**

The purpose of this pull request is to be a proof of concept for a "native TS" implementation of various CLI commands with the end goal being commands that do not require calling out to prisma-engines and which are runnable programmatically.

This proof of concept starts with a TS implementation of `db pull` which is a relatively straightforward command. The goal is to:
- create a native TS implementation of `db pull`
- test the new implementation against the existing implementation
- create a TypeScript API for this command.

The new implementation and the existing one should have the exact same output, obviously. 

I am pushing this up as a draft PR because local testing is proving to be quite difficult. I'm hoping that I can use the tests here to help.


packages/migrate/src/commands/DbPullTypeScript.ts
```diff
@@ -0,0 +1,479 @@
...
+        )} for more info.`)
+      }
+    }
+
+    // KEY DIFFERENCE: Use SchemaEngineWasm directly instead of Migrate.setup()
+    debug('Setting up TypeScript-native schema engine with WASM')
+    const engine = await SchemaEngineWasm.setup({
+      adapter,
+      schemaContext: schemaContext ?? undefined,
+    })
```
aqrln, line 326: This still uses the schema engine though, there's no actual `db pull` implementation in TypeScript.

How is this different to the regular `db pull` with schema-engine-wasm?

jharrell: Yup, that’s true. IIUC, using the WASM engine would still be lighter weight and more portable than “true” Rust binaries, right? 

Would you look more favorably on this approach or a re-implementation of the underlying `engine.introspect` command?

aqrln: > Yup, that’s true. IIUC, using the WASM engine would still be lighter weight and more portable than “true” Rust binaries, right?

Yes, but that's exactly what we are already doing when you configure a driver adapter for Migrate, the normal `db pull` command uses it already

aqrln: > Would you look more favorably on this approach or a re-implementation of the underlying engine.introspect command?

Depends on what you want to achieve. This approach mostly requires work on the Rust side and doesn't require rewriting the commands like you did here. This is also something that we have already started working on, although sadly haven't had capacity to work on lately, so if you want to help out, that would definitely be appreciated.

A TypeScript implementation of `db pull`, however, could be an interesting experiment. A naive quick and dirty implementation that runs introspection queries and directly translates them to a Prisma schema might not be very hard, but it's going to quickly become a lot more complex if you want it to be something that could be extended to support more commands. That would need to start with a TypeScript definition of the abstract data model, implement schema parsing (since `db pull` doesn't just create a new schema, it needs the current content as well) and rendering, and only then you could implement the actual introspection. If you then implement datamodel diffing in addition to that, that would essentially be a prototype of a TypeScript-based schema engine and Prisma Migrate.

jharrell: > Depends on what you want to achieve. This approach mostly requires work on the Rust side and doesn't require rewriting the commands like you did here. This is also something that we have already started working on, although sadly haven't had capacity to work on lately, so if you want to help out, that would definitely be appreciated.

Happy to help! Do you have an existing branch or notes I can take a look at?

> A TypeScript implementation of db pull, however, could be an interesting experiment...

This was my original intention. I totally understand that a single command POC would quickly balloon once more commands are added, but I wanted to show how moving pieces of functionality could b done.

aqrln: > Do you have an existing branch or notes I can take a look at?

I wanted to send you a bunch of links to Notion and Linear and only then realized you can't see those anymore 😭

All the code is on `main` in both repos though.

> This was my original intention.

Feel free to experiment with that as well! I think at least designing a good TypeScript representation of the data model would already be immensely valuable (and if we could make it interoperable with schema-engine's abstractions, we could even port it piece by piece; e.g., you could have introspection in TypeScript and diffing in WASM).

jharrell: @aqrln shot you a message on Discord but just in case you're not active there:

I'm working through an implementation of the datamodel in TypeScript. No PR yet, but here's a super rough example of how it could be used to take data from postgres https://gist.github.com/jharrell/06450f499ef2611eb8e5971d5d67e107

Does that look right to you? Any more info you'd like to see? Let me know what you think!

PR #27440, opened 2025-06-18
feat: support multiple pools

nuintun:
The adapter-mssql currently does not support multiple pools, which causes projects connecting to multiple databases to fail to run properly. Now, support for multiple pools has been added.


jacek-prisma:
lgtm, thanks for the contribution

jacek-prisma (approved):

PR #27443, opened 2025-06-18
perf: in-memory join optimization

jacek-prisma:
[ORM-1037](https://linear.app/prisma-company/issue/ORM-1037/qc-performance-regression-with-in-memory-joins)

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
+++ b/packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
@@ -198,14 +198,7 @@ export class QueryInterpreter {
           })),
         )) satisfies JoinExpressionWithRecords[]
 
-        if (Array.isArray(parent)) {
-          for (const record of parent) {
-            attachChildrenToParent(asRecord(record), children)
-          }
-          return { value: parent, lastInsertId }
-        }
-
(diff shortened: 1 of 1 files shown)


packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
```diff
@@ -384,41 +377,44 @@ type JoinExpressionWithRecords = {
...
 
-function attachChildrenToParent(parentRecord: PrismaObject, children: JoinExpressionWithRecords[]) {
+function attachChildrenToParents(parentRecords: unknown, children: JoinExpressionWithRecords[]) {
   for (const { joinExpr, childRecords } of children) {
-    parentRecord[joinExpr.parentField] = filterChildRecords(childRecords, parentRecord, joinExpr)
-  }
-  return parentRecord
-}
+    const parentKeys = joinExpr.on.map(([k]) => k)
+    const childKeys = joinExpr.on.map(([, k]) => k)
```
aqrln, line 383: I guess we could change the representation in the query plan to make it more convenient. IIRC the pretty printer also ends up doing the same transformation.

aqrln (approved):

PR #27460, opened 2025-06-19
fix(qc): fix Bytes ID issue

### LINKED ISSUE #27455: [6.8.0~6.10.x queryCompilier] Nested select returns null
### Bug description

Image
Image

This code works fine with the Prisma client without the query compiler, but not when using the query compiler.
Whenever there is a nested `select`, it always returns `null`.

I've been waiting a long time for this problem to be fixed, but it still hasn't been resolved.

### Severity

### Reproduction

1. You should run `find`, `create`, `update`, or `delete` queries with nested select scopes.
2. If you use the `relationJoins` preview feature with `queryCompiler`, this issue occurs with all types of queries.
3. If you don’t use `relationJoins`, then deeply nested select queries may cause issues. (Most of these are find queries.)
4. I’ve encountered this issue across multiple environments, including MacOS, Windows, and Linux. and never solved. (i tested it with all versions of prisma client.)

Expected: All the queries **should** returns **exactly same with the types**! I don't care about what environment that we use. It must be ensured.
Actual: returns `null`.

`queryCompiler` is really most-in-case stable preview feature? i don't think so.

### Frequency

Consistently reproducible

### Is this a regression?

Never works

### Workaround

Remove `queryCompiler`

### Prisma Schema & Queries

```prisma
generator client {
  binaryTargets       = ["native", "linux-arm64-openssl-1.0.x"]
  moduleFormat        = "esm"
  output              = "../common/generations/prisma/"
  previewFeatures     = ["queryCompiler", "relationJoins", "nativeDistinct", "driverAdapters"]
  importFileExtension = "ts"
}

generator json {
  provider = "prisma-json-types-generator"
}

datasource database {
  provider = "postgresql"
}
```

```prisma
model Accommodation {
  id         Bytes              @id @database.ByteA
  name       String
  type       AccommodationTypes @default(MOTEL)
  /// [Address]
  address    Json
  isLaunched Boolean            @default(false)
  launchedAt DateTime           @default(now())

  updatedAt DateTime @updatedAt

  timeTables AccommodationTimeTable[]

  roles                 Role[]
  policies              Policy[]
  roomType              RoomType[]
  rooms                 Room[]
  channels              ChannelAccount[]
  properties            ChannelProperty[]
  packages              Package[]
  channelPackages       ChannelPackage[]
  bookings              Booking[]
  inventories           Inventory[]
  ratePlans             RatePlan[]
  periods               Period[]
  rates                 Rate[]
  Inclusions            Inclusion[]
  audits                Audit[]
  notifications         Notification[]
  notificationAlarms    NotificationAlarm[]
  integrations          IntegrationAdapter[]
  IntegrationEvent      IntegrationEvent[]
  billings              Billing[]
  billingPaymentMethods BillingPaymentMethod[]
  packageTimeTables     PackageTimeTable[]
  tags                  Tag[]
}

model AccommodationTimeTable {
  id Bytes @id @database.ByteA

  stayType StayType @default(NIGHTUSE)

  monCheckInTime  String
  monCheckOutTime String

  tueCheckInTime  String
  tueCheckOutTime String

  wedCheckInTime  String
  wedCheckOutTime String

  thuCheckInTime  String
  thuCheckOutTime String

  friCheckInTime  String
  friCheckOutTime String

  satCheckInTime  String
  satCheckOutTime String

  sunCheckInTime  String
  sunCheckOutTime String

  updatedAt DateTime @updatedAt

  accommodationId Bytes @database.ByteA

  accommodation Accommodation @relation(fields: [accommodationId], references: [id], onDelete: Cascade)

  @@unique([accommodationId, stayType])
}

```
```ts
```

### Prisma Config

```ts
import type { Prisma } from '@/clients';

export const AccommodationTimeTableScopes = {
  id: true,
  stayType: true,
  monCheckInTime: true,
  monCheckOutTime: true,
  tueCheckInTime: true,
  tueCheckOutTime: true,
  wedCheckInTime: true,
  wedCheckOutTime: true,
  thuCheckInTime: true,
  thuCheckOutTime: true,
  friCheckInTime: true,
  friCheckOutTime: true,
  satCheckInTime: true,
  satCheckOutTime: true,
  sunCheckInTime: true,
  sunCheckOutTime: true,
  createdAt: true,
  updatedAt: true,
} satisfies Prisma.AccommodationTimeTableSelect;

export const AccommodationScopes = {
  id: true,
  name: true,
  type: true,
  address: true,
  options: true,
  isLaunched: true,
  timeTables: {
    select: AccommodationTimeTableScopes,
  },
  launchedAt: true,
  createdAt: true,
  updatedAt: true,
} satisfies Prisma.AccommodationSelect;

async function createAccommodation(options: CreateAccommodationOptions) {
  return await prismaClient.accommodation.create({
    data: options,
    select: AccommodationScopes,
  });
}
```

### Logs & Debug Info

_No response_

### Environment & Setup

- **OS:** macOS, Windows, Debian
- **Database:** PostgreSQL
- **Node.js version:** 22 LTS

### Prisma Version

```
6.8.0~6.10.1
```

jacek-prisma:
The problem in this case is that the standard format for byte arrays in the driver adapters is `number[]`, so when we read an array of bytes out of one query and then forward it to another it gets passed as `number[]` arg, but that causes comparisons to fail, since the drivers do not map the arg into a byte array. This does not involve the DataMapper because this happens before the query plan completes.

The change modifies `serializeSql` to convert byte array columns into `Uint8Array`.

[ORM-1123](https://linear.app/prisma-company/issue/ORM-1123/fix-nested-select-returns-null)

Fixes https://github.com/prisma/prisma/issues/27455


packages/client-engine-runtime/src/interpreter/serializeSql.ts
```diff
@@ -3,25 +3,36 @@ import { type ColumnType, ColumnTypeEnum, type SqlResultSet } from '@prisma/driv
...
+        return (value: unknown) => (Array.isArray(value) ? new Uint8Array(value) : value)
+      default:
+        return (value: unknown) => value
+    }
+  })
+
+  return resultSet.rows
+    .map((row) => row.map((value, index) => mappers[index](value)))
+    .map((row) =>
+      row.reduce<Record<string, unknown>>((acc, value, index) => {
```
aqrln: Probably better to inline the first `map`:

```suggestion
    .map((row) =>
      row
        .map((value, index) => mappers[index](value)))
        .reduce<Record<string, unknown>>((acc, value, index) => {
```

The `rows` array could be large, and `map` is not lazy in JavaScript but will allocate a whole new array each time.

jacek-prisma: makes sense, updated

aqrln (approved):

PR #27393, opened 2025-06-11
feat(adapter-mysql2): add adapter-mysql2

jkomyno:
**DRAFT**

This PR:
- closes [ORM-761](https://linear.app/prisma-company/issue/ORM-761/new-driver-adapter-mysql2-for-provider-=-mysql)
- closes [ORM-1041](https://linear.app/prisma-company/issue/ORM-1041/include-mariadb-relation-join-check-in-getconnectioninfo-for-mysql2)
- adds support for `mysql2` for the `driverAdapters` preview feature
  - only sets `supportsRelationJoins: true` is the database version is not MariaDB AND it is > 8.0.13
- introduces `@prisma/adapter-mysql2`
- adds `js_mysql2` to the available drivers for the functional test suite

PR #27462, opened 2025-06-20
feat(adapter-mysql2): add adapter-mysql2 (2nd attempt)

jkomyno:
Second attempt at: https://github.com/prisma/prisma/pull/27393.

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -135,7 +135,7 @@ jobs:
           timeout_minutes: 10
           max_attempts: 5
           retry_wait_seconds: 120
-          command: docker compose -f docker/docker-compose.yml up --wait --detach postgres postgres_isolated mysql mysql_isolated mssql mongo cockroachdb vitess-8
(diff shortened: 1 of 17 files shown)


medz:
@jkomyno Any progress? We are migrating the old MySQL process to workers via hyperdrive. Through testing, mysql2 can connect normally, while Prisma MariaDB adapter always times out.

jkomyno:
@medz We may retry this later on. Right now, trying to add `adapter-mysql2` was a time-sink to due memory leaks and undocumented runtime behavior in mysql2. However, supporting MySQL on Cloudflare Workers is a significantly lower priority for us, compared to Prisma Postgres / Neon / PlanetScale / D1

medz:
@jkomyno thank you for your work, although it has not been officially merged. Although Mariadb can link to MySQL, it doesn't seem to be that good.

I know that Prisma Mariadb connection timeout is not the reason for Prisma, so I created https://github.com/mariadb-corporation/mariadb-connector-nodejs/issues/322 Perhaps the authorities will not pay attention.

But when to enable mysql2 adapter would be great for us, and choosing which database to use in the real world is not so idealistic. After most users join, the product has already taken shape and there is no choice at all.

PR #27464, opened 2025-06-20
feat(tracking): ORM-1125 only track known top level commands

FGoessler:
This PR moves the checkpoint logic - which is responsible for update notifcations and tracking - further down into the CLI execution so it only runs if at least the top level subcommand is known.

This shall remove unnecessary spam on our tracking backends.

It was considered to also avoid tracking unknown sub-subcommands (e.g. `prisma migrate foobar`). But that gets tricky due to the nested and async nature of our command execution system. Especially as we also want to track long running commands like `prisma dev` by firing the tracking call at the beginning of the execution and not just its end.

Expected outcome:
- Unknown subcommand: `prisma foobar` => does not track
- Known subcommand: `prisma version` => does track
- Known long running subcommand: `prisma dev` => does track
- Known sub-subcommand: `prisma migrate dev` => does track
- Unknown sub-subcommand: `prisma migrate foobar` => does track

### DIFF
--- a/packages/cli/src/__tests__/update-message.test.ts
+++ b/packages/cli/src/__tests__/update-message.test.ts
@@ -5,24 +5,143 @@ import { printUpdateMessage } from '../utils/printUpdateMessage'
 const ctx = jestContext.new().add(jestConsoleContext()).assemble()
 
 describe('update available message', () => {
-  it('dev tag - minor', () => {
+  let originalPrismaHideUpdateMessageEnv: string | undefined
+
+  beforeEach(() => {
+    originalPrismaHideUpdateMessageEnv = process.env.PRISMA_HIDE_UPDATE_MESSAGE
+    delete process.env.PRISMA_HIDE_UPDATE_MESSAGE
+  })
+
+  afterEach(() => {
+    process.env.PRISMA_HIDE_UPDATE_MESSAGE = originalPrismaHideUpdateMessageEnv
+  })
+
+  describe('update available', () => {
+    it('dev tag - minor', () => {
+      printUpdateMessage({
+        status: 'ok',
+        // @ts-ignore
+        data: {
+          previous_version: '2.6.1-dev.18',
+          current_version: '2.16.0-dev.8',
+          package: 'prisma',
+          release_tag: 'dev',
+          outdated: true,
+        },
+      })
+      const message = ctx.mocked['console.error'].mock.calls[0][0]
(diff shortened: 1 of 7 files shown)


packages/cli/src/utils/printUpdateMessage.ts
```diff
@@ -4,7 +4,12 @@ import { blue, bold } from 'kleur/colors'
 
 const isPrismaInstalledGlobally = isCurrentBinInstalledGlobally()
 
-export function printUpdateMessage(checkResult: { status: 'ok'; data: Check.Response }): void {
+export function printUpdateMessage(checkResult: Check.Result | 0 | void): void {
+  const shouldHide = process.env.PRISMA_HIDE_UPDATE_MESSAGE
+  if (!checkResult || checkResult.status !== 'ok' || shouldHide || !checkResult.data.outdated) {
+    return
+  }
```
FGoessler, line 11: ℹ️ Moved logic from the only call site into this function to better encapsulate it. Now this part is also unit tested.

packages/cli/src/utils/checkpoint.ts
```diff
@@ -1,18 +1,18 @@
...
+const packageJson = require('../../package.json')
 
-export interface CheckpointClientCheckOptions {
-  isPrismaInstalledGlobally: 'npm' | false
-  version: string
-  command: string
-  telemetryInformation: string
-  schemaPath?: string
-  schemaPathFromConfig?: string
-}
```
FGoessler, line 15: ℹ️ Moved the logic to derive this information from the call site into this function to better encapsulate it.

packages/cli/src/__tests__/update-message.test.ts
```diff
@@ -5,73 +5,157 @@ import { printUpdateMessage } from '../utils/printUpdateMessage'
...
+  beforeEach(() => {
+    originalPrismaHideUpdateMessageEnv = process.env.PRISMA_HIDE_UPDATE_MESSAGE
+    delete process.env.PRISMA_HIDE_UPDATE_MESSAGE
+  })
+
+  afterEach(() => {
+    process.env.PRISMA_HIDE_UPDATE_MESSAGE = originalPrismaHideUpdateMessageEnv
+  })
+
+  describe('update available', () => {
```
FGoessler, line 19: ℹ️ Github diff is a bit stupid here: These tests already existed, I just nested them into a describe block. The tests below are the actual new ones.

jacek-prisma (approved):

PR #27471, opened 2025-06-20
feat(client-engine-runtime): Support KSUID and Prefixed TypeID

### LINKED ISSUE #24992: Add support for typeid
## Problem

Implement support for TypeID:

https://github.com/jetify-com/typeid

## Suggested solution

Have `typeid()` as alternative to `uuid()` and `cuid()`.

## Alternatives

I tried to use the custom typeid_generate_text function defined in https://github.com/jetify-com/typeid-sql and use it with `dbgenerated` but I wasn't able to make it work.

## Additional context

This would allow for ids with prefixes (`@default(typeid('prefix'))`)

kayandra:
This PR introduces support for two new id generators
- adds [KSUID](https://github.com/novemberborn/ksuid) generator
- adds typeid generator with prefix and closes #24992.

Checklists
- [x] I've included backing tests for the generators. 
- [x] I've update the vscode extension and language server. see https://github.com/prisma/language-tools/pull/1886
- [ ] For typeid's it'd be better for the type returned to be of type `TypeID<'prefix'>` instead of string. I may need some guidance on this.
- [ ] Show suggestions for prefix based on the model name name in lowercase. i.e `model User {}` gets `typeid(<user>)` as suggestion. 

P.S if there's something important I have to do to get these merged please let me know.

### DIFF
--- a/packages/client-engine-runtime/package.json
+++ b/packages/client-engine-runtime/package.json
@@ -31,6 +31,8 @@
     "@prisma/driver-adapter-utils": "workspace:*",
     "decimal.js": "10.5.0",
     "nanoid": "5.1.5",
+    "ksuid": "3.0.0",
+    "typeid-js": "1.2.0",
     "ulid": "3.0.0",
     "uuid": "11.1.0"
   },
--- a/packages/client-engine-runtime/src/interpreter/generators.test.ts
+++ b/packages/client-engine-runtime/src/interpreter/generators.test.ts
@@ -103,6 +103,42 @@ test('should generate different and valid Nano IDs with custom length', () => {
   expect(nanoid1).toMatch(/^[A-Za-z0-9_-]{7}$/)
 })
 
+test('should generate different and valid KSUIDs', () => {
+  const registry = new GeneratorRegistry()
+  const snapshot = registry.snapshot()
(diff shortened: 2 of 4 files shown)


aqrln (changes requested):
Thanks a lot for the PR! This is a good start, however this also needs:

* Schema validation
* Proper integration tests in `packages/client/tests/functional`
* Support for it in https://github.com/prisma/prisma-engines/blob/c717abe28d5e8e01b58d43e0c0134630791630c4/query-engine/query-structure/src/field/scalar.rs#L239 so the query compiler knows they exist and can embed them to the generated query plans (which are interpreted by the code you modified here)
* The implementation in https://github.com/prisma/prisma-engines/blob/c717abe28d5e8e01b58d43e0c0134630791630c4/query-engine/query-structure/src/default_value.rs#L306 so that the query engine also knows what to do with them when the query compiler is not used.

kayandra:
Alright, I'll make the changes and push

paul-mais:
I'd like to help with this PR, if it's still active

aqrln:
Closing as stale. Please feel free to reopen and continue if you're still interested in implementing this.

PR #27478, opened 2025-06-20
fix(client-engine-runtime): support MySQL BIT type

aqrln:
`BIT(M)` columns in MySQL are returned by JS drivers as `Uint8Array`s of length `Math.ceil(M / 8)`. Although the query engine contains logic to decode these values as arbitrary integers, we actually only allow `Bytes` and `Boolean` fields to be annotated with this native type, so we only needed to add support for decoding `Uint8Array`s as booleans in the `DataMapper`.

Closes: https://linear.app/prisma-company/issue/ORM-900/support-mysql-bit-columns

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/DataMapper.ts
+++ b/packages/client-engine-runtime/src/interpreter/DataMapper.ts
@@ -185,6 +185,12 @@ function mapValue(
           throw new DataMapperError(`Expected a boolean in column '${columnName}', got ${typeof value}: ${value}`)
         }
       }
+      if (value instanceof Uint8Array) {
+        for (const byte of value) {
+          if (byte !== 0) return true
+        }
+        return false
+      }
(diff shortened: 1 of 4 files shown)


packages/client-engine-runtime/src/interpreter/DataMapper.ts
```diff
@@ -185,6 +185,12 @@ function mapValue(
           throw new DataMapperError(`Expected a boolean in column '${columnName}', got ${typeof value}: ${value}`)
         }
       }
+      if (value instanceof Uint8Array) {
+        for (const byte of value) {
+          if (byte !== 0) return false
```
jacek-prisma:
```suggestion
          if (byte === 0) return false
```
seems you meant

aqrln: almost 😅
fixed now, thanks for catching

FGoessler (approved):

PR #27489, opened 2025-06-23
feat: implement a MariaDB driver adapter

jacek-prisma:
[ORM-1129](https://linear.app/prisma-company/issue/ORM-1129/new-driver-adapter-mariadb-for-provider-=-mysql)


packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -451,6 +456,8 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
     }
 
     if (!this.libraryStarted) {
+      await (await this.adapterPromise)?.dispose()
```
jacek-prisma, line 459: adapters not disposed here were leaked, this would cause tests to hang forever, because mariadb pool keeps track of idle connections and prevents the process from exiting

packages/adapter-mariadb/jest.config.js
aqrln, line 1: I'd prefer to use vitest in all new packages we add

packages/adapter-mariadb/src/conversion.ts
```diff
@@ -0,0 +1,140 @@
...
+      // Fall back to Int32 for consistency with quaint.
+      return ColumnTypeEnum.Int32
+    default:
+      throw new Error(`Unsupported column type: ${field.type}`)
+  }
+}
+
+export function mapArg(arg: unknown): unknown {
+  if (arg instanceof Uint8Array) {
+    return Buffer.from(arg)
```
aqrln:
```suggestion
    return Buffer.from(arg.buffer, arg.byteOffset, arg.byteLength)
```
to share the underlying `ArrayBuffer` instead of cloning the data

packages/adapter-mariadb/jest.config.js
jacek-prisma: done

packages/adapter-mariadb/src/conversion.ts
jacek-prisma: done

packages/adapter-mariadb/src/mariadb.ts
```diff
@@ -0,0 +1,201 @@
...
+async function getCapabilities(pool: mariadb.Pool): Promise<{ supportsRelationJoins: boolean }> {
+  const tag = '[js::getCapabilities]'
+
+  try {
+    const [rows] = await pool.query({
+      sql: `SELECT VERSION()`,
+      rowsAsArray: true,
+    })
+
+    const version = rows[0][0] as `${number}.${number}.${number}${'-MariaDB' | ''}`
```
aqrln: the suffix can also be `-vitess` or `-Vitess` (at least; perhaps could be something else as well)

jacek-prisma: removed it and cleaned up the code

packages/client/tests/functional/_utils/setupTestSuiteClient.ts
```diff
@@ -317,5 +317,24 @@ export function setupTestSuiteClientDriverAdapter({
     }
   }
 
+  if (driverAdapter === 'js_mariadb') {
+    const { PrismaMariaDb } = require('@prisma/adapter-mariadb') as typeof import('@prisma/adapter-mariadb')
+
+    const [, user, password, host, port, database] =
+      datasourceInfo.databaseUrl.match(/^mysql:\/\/([^:]+):([^@]+)@([^:;]+):(\d+)\/(.+)$/) || []
```
aqrln: Doesn't matter much in tests since we control the connection string and know the regex will work for it, but might've been easier to use `URL` to parse it:

```suggestion
    const url = new URL(datasourceInfo.databaseUrl)
    const { username: user, password, hostname: host, port } = url
    const database = url.pathname && url.pathname.slice(1)
```

jacek-prisma: done

aqrln (approved):

Issue #27157, opened 2025-05-12
PrismaPG: prismaClient unusable after calling $disconnect since 6.6.0


mnahkies:
### Bug description

Since the new driver adapter interface from `6.6.0`, calling `$disconnect` calls `end()` on the underlying `pg.Pool`, and doesn't construct a fresh pool when new queries are issued.

This causes any subsequent queries to fail with an obtuse exception.

### Severity

🔹 Minor: Unexpected behavior, but does not block development

### Reproduction

```typescript
import {PrismaPg} from "@prisma/adapter-pg";
import {PrismaClient} from "@prisma/client";
import {Pool} from "pg";

async function main() {
  // const pool = new Pool({
  //   connectionString: process.env.CONNECTION_STRING,
  //   max: 10,
  //   allowExitOnIdle: true,
  // });

  const pgAdapter = new PrismaPg({
    connectionString: process.env.CONNECTION_STRING,
    max: 10,
    allowExitOnIdle: true,
  });
  const prismaClient = new PrismaClient({
    log: [
      {
        emit: 'event',
        level: 'query',
      },
    ],
    adapter: pgAdapter,
  });

  console.info(await prismaClient.$executeRaw`SELECT 1`)
  await prismaClient.$disconnect()
  // on 6.5.0 this works, 6.6.0 and up it doesn't.
  console.info(await prismaClient.$executeRaw`SELECT 1`)
}

main()

```

### Expected vs. Actual Behavior

**Expected**
The client should reconnect next time a query is issued.

**Actual**
An obtuse error is thrown, which if you add logging to https://github.com/prisma/prisma/blob/18e17cc0f6dfea6a4de44342f47f7a59ddd7a3ba/packages/adapter-pg/src/errors.ts#L6 you'll find it's a error from `pg` saying the `pool` has already been ended.

```
PrismaClientKnownRequestError: 
Invalid `prisma.$executeRaw()` invocation:


Error in external connector (id 0)
    at zn.handleRequestError (/Users/michaelnahkies/projects/node-libs/node_modules/@prisma/client/runtime/library.js:121:7459)
    at zn.handleAndLogRequestError (/Users/michaelnahkies/projects/node-libs/node_modules/@prisma/client/runtime/library.js:121:6784)
    at zn.request (/Users/michaelnahkies/projects/node-libs/node_modules/@prisma/client/runtime/library.js:121:6491)
[...]

Issue #25714, opened 2024-11-24
Possible memory leak using driver adapters with cloudflare D1


tellisnz:
### Bug description

Around the 16th november we started to get errors when making prisma calls in our production prisma/cloudflare workers/d1 setup.

<img width="1793" alt="image" src="https://github.com/user-attachments/assets/d71c7419-e8e3-4129-a062-c5a6197e2800">

The errors consisted of either:

"memory access out of bounds"
or
"Invalid typed array length: undefined"

We had made no recent deployments nor had our load profile changed, this was pretty out of the blue so might suggest there was a change made on the cloudflare side.

At the moment, our workaround when we start getting these errors is to redeploy the same code base, and this (usually) seems to trigger some sort of worker restart and clears the problem for a short period.

We have been going backwards and forwards with their support but it's fairly slow going, and their current response is:

>Our Workers team confirmed this is coming from the Prisma library you're using.
>They seem to be using WASM, and it's causing these memory access violations.
>You should be able to see this locally too with an up-to-date Wrangler.
>We'd recommend reaching out to the Prisma team and reporting the issue to them. It would be best to provide a reproduction case to them if you can as well.

I think we can reproduce some sort of memory leak locally, and a reproduction case is available [here](https://github.com/tellisnz/prisma-d1-example) - but it is fairly trivial and basically a reproduction of instructions from [cloudflare](https://developers.cloudflare.com/d1/tutorials/d1-and-prisma-orm/) / [prisma](https://www.prisma.io/docs/orm/overview/databases/cloudflare-d1).

If you run it, and start making requests against it, e.g. with

`while true; do curl http://localhost:8787 ; done`

and take memory snapshots every now and again, you see typed array usage increasing rapidly

<img width="779" alt="image" src="https://github.com/user-attachments/assets/2a5cd659-3e15-4988-8ef2-f4c0f79e0b81">

and can inspect:

<img width="2050" alt="image" src="https://github.com/user-attachments/assets/0cb2b449-f925-4645-bbf8-1e6129d6921b">

### How to reproduce

1. Clone https://github.com/tellisnz/prisma-d1-example and `npm i`
2. Create your own demo db i.e. `npx wrangler d1 create prisma-demo-db` (although not sure if completely necessary if just attempting to run locally) - but if using remote change database_id in wranger.toml 
4. Run `npx wrangler d1 migrations apply prisma-demo-db --local`
5. Run `npx prisma generate`
6. Run `npm run dev`
[...]

PR #27493, opened 2025-06-24
fix(mem): ORM-922 fix memory leak with driver adapters

FGoessler:
This PR fixes a memory leak as described in https://github.com/prisma/prisma/issues/25714.

Reason for the memory leak:
- When using the edge/workerd client variant with driver adapters we leverage a WASM query engine.
- This WASM query engines received a `DriverAdapter` instance and a log callback which holds a reference to its JS `LibraryEngine` wrapper.
- Due to this a reference cycle between JS and WASM exists that cannot be resolved by itself as it seems.
- => Every newly created `PrismaClient` causes more `DriverAdapter`, `LibraryEngine`, etc. instances to hang around and pile up memory.

Fix for this:
- We have to call `free` on the WASM query instance explicitly to free all its memory. This resolves the reference cycle and the `DriverAdapter` and `LibraryEngine` in turn also get freed.

Drawbacks / Constraints:
- We cannot call `free` automatically as we do not know when the `PrismaClient` is "done".
- I tried to experiment with [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) but it seems to be nor supported in Worker runtimes anyways.
- => We can only call `free` during an explicit `$disconnect` call.
- => ⚠️  Users have to call `client.$disconnect()` explicitly when using short lived `PrismaClient` instances.
- => As the WASM engine is now freed the `PrismaClient` could not be used again to `$connect`. I did some fixes and added test so this works.
- ⚠️  The query compiler does not yet support reconnecting though as it created the driver adapter in its constructor but disconnects upon `$disconnect`. => Additional work required to fix this.

> [!NOTE]
> These changes also fix the "reconnect" issue described in https://github.com/prisma/prisma/issues/27157 - but still require work to fix for the query compiler -> https://linear.app/prisma-company/issue/ORM-972/bug-regression-driver-adapter-disconnect-issues

Testing Notes:
- I tried to add a memory test for this, but failed in doing so.
- Our memory test suite setup does not play nicely with WASM modules yet.
- Am getting errors like:
  - `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".wasm"`
  - `PrismaClientInitializationError: The loaded wasm module was unexpectedly 'undefined' or 'null' once loaded`
- => Using a test with some mocks asserting free, dispose and disconnect are called

> [!NOTE]
> I tested the query compiler and it does not suffer from this memory leak issue. I assume because its WASM instance does not receive any long-lived JS callbacks or adapter objects. 🎉

### DIFF
--- a/packages/client-common/src/QueryEngine.ts
+++ b/packages/client-common/src/QueryEngine.ts
@@ -22,6 +22,11 @@ export type QueryEngineInstance = {
   connect(headers: string, requestId: string): Promise<void>
   disconnect(headers: string, requestId: string): Promise<void>
   /**
+   * Frees any resources allocated by the engine's WASM instance. This method is automatically created by WASM bindgen.
+   * Noop for other engines.
+   */
+  free?(): void
+  /**
    * @param requestStr JSON.stringified `QueryEngineRequest | QueryEngineBatchRequest`
    * @param headersStr JSON.stringified `QueryEngineRequestHeaders`
    */
--- a/packages/client/src/runtime/core/engines/library/LibraryEngine.test.ts
+++ b/packages/client/src/runtime/core/engines/library/LibraryEngine.test.ts
@@ -1,4 +1,5 @@
 import { QueryEngineInstance } from '@prisma/client-common'
+import { SqlDriverAdapter, SqlDriverAdapterFactory } from '@prisma/driver-adapter-utils'
 import { EventEmitter } from 'events'
 
 import { PrismaClientInitializationError } from '../../errors/PrismaClientInitializationError'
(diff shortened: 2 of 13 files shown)


aqrln:
> I tried to experiment with [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) but it seems to be nor supported in Worker runtimes anyways.

Also it is already used in JS code generated by `wasm-bindgen`, this is why you don't normally need to call `free` explicitly when running in Node.js

aqrln (approved):
LGTM, I'll fix reconnect in `ClientEngine` in a follow up PR. Thanks for adding a test!

PR #27502, opened 2025-06-24
feat(client-engine-runtime): support nested objects with mapped names

aqrln:
Ref: https://github.com/prisma/prisma-engines/pull/5495
Ref: https://linear.app/prisma-company/issue/ORM-966/mapped-db-level-field-name-conflicts-with-relation-field

/engine-branch push-ukmoovqomqvs

### DIFF
--- a/packages/client-engine-runtime/src/QueryPlan.ts
+++ b/packages/client-engine-runtime/src/QueryPlan.ts
@@ -65,7 +65,7 @@ export type ResultNode =
   | {
       type: 'Object'
       fields: Record<string, ResultNode>
-      flattened: boolean
+      serializedName: string | null
     }
   | {
       type: 'Value'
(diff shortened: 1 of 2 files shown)


packages/client-engine-runtime/src/QueryPlan.ts
```diff
@@ -65,7 +65,7 @@ export type ResultNode =
   | {
       type: 'Object'
       fields: Record<string, ResultNode>
-      flattened: boolean
+      from: string | null
```
jacek-prisma: I'd probably be more explicit and call it `dbName`

aqrln: renamed to `serializedName`

jacek-prisma (approved):

PR #27510, opened 2025-06-25
fix(adapter-pg): skip empty statements in `executeScript`

pclokcer:
…mpty ones


packages/adapter-pg/src/pg.ts
```diff
@@ -183,8 +183,12 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
   }
 
   async executeScript(script: string): Promise<void> {
-    // TODO: crude implementation for now, might need to refine it
-    for (const stmt of script.split(';')) {
+    const statements = script
+      .split(new RegExp(/;[\t\r\n ]*/g))
```
aqrln: `new RegExp` looks superfluous, you can just pass the regex literal directly.

A few questions about the regex itself:
1. Why is it important to skip the whitespace? If it is, why don't we skip the whitespace before the first statement?
2. Does the `g` flag really change anything here?

packages/adapter-pg/src/pg.ts
```diff
@@ -183,8 +183,12 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
   }
 
   async executeScript(script: string): Promise<void> {
-    // TODO: crude implementation for now, might need to refine it
-    for (const stmt of script.split(';')) {
+    const statements = script
+      .split(/;[\t\r\n ]*/)
```
aqrln: one more simplification:

```suggestion
      .split(/;\s*/)
```

(see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class_escape#s)

packages/adapter-pg/src/pg.ts
```diff
@@ -183,8 +183,12 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
...
+      .map(stmt => stmt.trim())
```
aqrln: Although, again, since we're trimming whitespace separately here, and the regex above doesn't handle whitespace before the first statement and before each semicolon anyway, I don't think there's much value in the regex and it could be replaced with just `split(';')`.

pclokcer: you rigth 👍

aqrln:
@pclokcer formatting check is failing on CI

pclokcer:
formatting failed resolved. it can be merge 👍

packages/adapter-pg/src/pg.ts
```diff
@@ -183,8 +183,8 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
   }
 
   async executeScript(script: string): Promise<void> {
-    // TODO: crude implementation for now, might need to refine it
-    for (const stmt of script.split(';')) {
+    const statements = script.split(/;\s*/).filter((stmt) => stmt.length > 0)
```
aqrln:
```suggestion
    const statements = script.split(';').map(stmt => stmt.trim()).filter((stmt) => stmt.length > 0)
```

aqrln (approved):

PR #27508, opened 2025-06-25
fix(client-generator-ts): fix dynamic import on Webpack 5 (default bundler on Next.js 15)

jkomyno:
This PR modifies the `prisma-client` generator by introducing

```typescript
const dynamicRequireFn = async <const T extends string>(name: T) =>
  typeof __non_webpack_require__ === 'function'
    ? Promise.resolve(__non_webpack_require__(name))
    : await import(/* webpackIgnore: true */ name)
```

This function evaluates to:
- `import(name)` for all bundler targets, except Webpack, but including Turbopack.
- `__non_webpack_require__(name)` for Webpack targets.

This is used to dynamically import a module at runtime, while also excluding it from Webpack's bundle.
 It allows to mitigate https://github.com/webpack/webpack/issues/19607 and
 - fix https://github.com/prisma/prisma/issues/27049
 - fix https://github.com/prisma/prisma/issues/27343

In practice, it's used to import `node:fs/promises` and `node:module`.

---

At the present state, this PR seems to do the trick with Next.js 15 / Webpack 5 without causing regressions in other frameworks or environments. E.g., since `__non_webpack_require__` is only defined on Webpack, it doesn't affect Turbopack, which will still use the native `await import`.

We can also decide to introduce a new top-level `runtime = "webpack"` configuration option for the `prisma-client` generator, and remove the build-time conditionality in `dynamicRequireFn`. However, that would introduce additional friction, and would need to be deprecated once Turbopack (Webpack's replacement) becomes generally available, so I'd advise against it.

PR #27516, opened 2025-06-25
fix(client-generator-ts): ensure _require is defined for CJS

JAD3N:
It seems during the changes made in https://github.com/prisma/prisma/pull/27508, `require` was renamed to `_require` but for CJS environments the definition wasn't also renamed. I've applied a simple change to ensure _require is defined.


aqrln (approved):
Thank you!

jkomyno:
Thank you @JAD3N! I discovered this PR thanks to a git conflict error in https://github.com/prisma/prisma/pull/27558, where  - among other things - I had discovered and fixed this problem 😄

PR #27528, opened 2025-06-26
fix(client): revamp ClientEngine initialization and shutdown

### LINKED ISSUE #27157: PrismaPG: prismaClient unusable after calling $disconnect since 6.6.0

aqrln:
Refactor the initialization and shutdown logic in the `ClientEngine`:

* Get rid of messy interdependent `Promise`s stored as class fields and introduce an explicit state machine with a discrete number of states and clearly defined transitions between them. All methods are safely callable concurrently and in any sequence because they know how to handle each of the four possible states. `ClientEngine` now starts disconnected and connects lazily on the first query, just like `LibraryEngine`.
	```mermaid
	stateDiagram-v2
	    [*] --> disconnected
	
	    disconnected --> connecting: #ensureStarted called
	    connecting --> connected: Connection promise resolved
	    connecting --> disconnected: Connection promise rejected
	    connected --> disconnecting: stop called
	    disconnecting --> disconnected: Disconnect promise settled
	```
* Make sure we discard the query compiler on panic and never use the same instance again to avoid memory leaks.
* Immediately drop the query compiler instance on WASM side on disconnect instead of relying on `FinalizationRegistry` to maybe free it some time later. Previously we would have a guaranteed memory leak on Cloudflare Workers here but they [shipped `FinalizationRegistry` support a couple of weeks ago](https://blog.cloudflare.com/we-shipped-finalizationregistry-in-workers-why-you-should-never-use-it/). However, just like the linked article says, the cleanup callbacks are not guaranteed to run, and we should call `free` directly if possible and only rely on GC hooks as a safety net.
* Fix `getErrorMessageWithLink` being passed a random query that happened to be started last instead of the query that actually failed, when multiple queries are in flight.

Refs: https://github.com/prisma/prisma/pull/27493
Fixes: https://github.com/prisma/prisma/issues/27157
Closes: https://linear.app/prisma-company/issue/ORM-972/bug-regression-driver-adapter-disconnect-issues


packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -138,46 +159,91 @@ export class ClientEngine implements Engine<undefined> {
...
+              driverAdapter,
+              transactionManager,
+              queryCompiler,
+            }
+
+            this.#state = { type: 'connected', engine }
+
+            return engine
+          } catch (error) {
+            this.#state = { type: 'disconnected' }
```
jacek-prisma, line 181: probably would not be a problem in practice, but if `await this.#instantiateQueryCompiler` fails, the `driverAdapter` gets leaked here

jacek-prisma (approved):
left 1 minor comment, but looks very good

packages/client/src/runtime/core/engines/client/ClientEngine.ts
aqrln: done in [`abbb2ea` (#27528)](https://github.com/prisma/prisma/pull/27528/commits/abbb2ea2e8d0c63265283cc16b099715e130b383)

jacek-prisma: 👍

jacek-prisma (approved):

PR #27532, opened 2025-06-26
fix: Remove Model suffix from ESM generator

### LINKED ISSUE #27531: ESM Model Name suffix
### Bug description

Hi there,

This PR changed the ESM model exports to have a `Model` suffix: https://github.com/prisma/prisma/pull/26727/files#diff-264ddc34afc7c904dfe0af24b2b265ed27299067e62c224a007c5541d2a38614R355

This means that I cannot import `User` and instead need to import `UserModel`.

The docs are also incorrect and suggest we should be able to import without the model suffix.

https://www.prisma.io/docs/orm/prisma-schema/overview/generators#importing-generated-model-types

### Severity

### Reproduction

1. Use new prisma-client generator
2. Observe generated code

Can I open a PR to remove this suffix? It doesn't seem necessary to me. For context, I want to use Enums and Models in a frontend app, and right now I am forced to import the `client.ts` file which has all the runtime stuff I don't want.

### Frequency

Consistently reproducible

### Is this a regression?

Worked in 6.6.0, broke in 6.7.0

### Workaround

No workaround

### Prisma Schema & Queries

```prisma
generator client {
  provider = "prisma-client"
  output   = "../generated"
}
```

### Prisma Config

_No response_

### Logs & Debug Info

_No response_

### Environment & Setup

- **OS:** macOS
- **Database:** Postgres
- **Node.js version:** v20.9.0

### Prisma Version

```
prisma                  : 6.10.1
@prisma/client          : 6.10.1
Node.js                 : v20.9.0
TypeScript              : 5.6.2
```

tonybruess:
Fixes #27531

### DIFF
--- a/packages/client-generator-ts/src/TSClient/Model.ts
+++ b/packages/client-generator-ts/src/TSClient/Model.ts
@@ -352,7 +352,7 @@ export type ${getAggregateGetName(model.name)}<T extends ${getAggregateArgsName(
     const modelTypeExport = ts
       .moduleExport(
         ts.typeDeclaration(
-          `${this.model.name}Model`,
+          this.model.name,
           ts
             .namedType(`runtime.Types.Result.DefaultSelection`)
             .addGenericArgument(ts.namedType(getPayloadName(this.model.name))),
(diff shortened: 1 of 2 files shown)


FGoessler:
Thanks for bringing this up but closing this as per https://github.com/prisma/prisma/issues/27531#issuecomment-3023459815

PR #27542, opened 2025-06-27
fix: add error mapping for value out of range errors

aqrln:
Closes: https://linear.app/prisma-company/issue/ORM-963/validation-error-for-out-of-range-integer

/engine-branch push-pvwlrwvnzzrn

### DIFF
--- a/packages/client-engine-runtime/src/UserFacingError.ts
+++ b/packages/client-engine-runtime/src/UserFacingError.ts
@@ -125,7 +125,7 @@ function renderErrorMessage(err: DriverAdapterError): string | undefined {
     case 'NullConstraintViolation':
       return `Null constraint violation on the ${renderConstraint(err.cause.constraint)}`
     case 'ValueOutOfRange':
-      return `Value out of range for the type. ${err.cause.cause}`
+      return `Value out of range for the type: ${err.cause.cause}`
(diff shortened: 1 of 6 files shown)


packages/client-engine-runtime/src/UserFacingError.ts
```diff
@@ -125,7 +125,7 @@ function renderErrorMessage(err: DriverAdapterError): string | undefined {
     case 'NullConstraintViolation':
       return `Null constraint violation on the ${renderConstraint(err.cause.constraint)}`
     case 'ValueOutOfRange':
-      return `Value out of range for the type. ${err.cause.cause}`
+      return `Value out of range for the type: ${err.cause.cause}`
```
aqrln, line 128: Changed this (and the corresponding QE error) to use a colon because an uppercase letter after a colon looks less ugly than a lowercase letter after a period.

Before:
* `Value out of range for the type. value "4294967296" is out of range for type oid` (Postgres)
* `Value out of range for the type. Out of range value for column 'tinyint'` (MySQL)

After:
* `Value out of range for the type: value "4294967296" is out of range for type oid` (Postgres)
* `Value out of range for the type: Out of range value for column 'tinyint'` (MySQL)

jacek-prisma (approved):

PR #27546, opened 2025-06-28
fix: enable running generators with bun

fehnomenal:
When running a generator with bun (like `provider = "bun src/kysely-mssql-generator.ts"`) the rpc is hanging forever.

Apparently bun outputs a red-colored string (with ANSI escape sequences) to `stderr` when calling `console.error`. This in turn is no valid JSON and parsing fails. This is circumvented by directly writing to `process.stderr`.

I found the problem by calling `NO_COLOR=1 bun prisma generate` which suppresses colored output from bun (and thus a plain json string on `stderr`).


aqrln (approved):
Thanks for figuring this out and fixing it!

PR #27558, opened 2025-06-30
test(client-generator-ts): test and snapshot the result of 80 possible combinations of `buildGetWasmModule`

jkomyno:
This PR:
- Closes [ORM-1143](https://linear.app/prisma-company/issue/ORM-1143/test-buildgetwasmmodule-in-prisma-client)
- Creates 80 new tests for `buildGetWasmModule` in `@prisma/client-generator-ts`
  - Each test performs assertions on a specific combination of input arguments
  - For each generated TypeScript definition, we run an in-memory instance of `typescript`'s compiler to verify we're emitting valid TypeScript
- Fixes reference errors in `buildGetWasmModule` in `@prisma/client-generator-ts` when `moduleFormat === 'cjs'`
  - Update: this was already fixed by https://github.com/prisma/prisma/pull/27516 
- Fixes types when users are not adopting Webpack and their `tsconfig.json` contains `{ "compilerOptions": { "skipLibCheck": false } }`


packages/client-generator-ts/tests/assert-typescript-is-valid.ts
```diff
@@ -0,0 +1,203 @@
...
+export function assertTypeScriptIsValid({ files, moduleFormat }: AssertTypeScriptIsValidInput) {
+  const configFromModuleFormat = match(moduleFormat)
+    .with('cjs', () => ({
+      module: ts.ModuleKind.Node16,
+      moduleResolution: ts.ModuleResolutionKind.Node16,
+      types: ['node'],
+    }))
+    .with('esm', () => ({
+      module: ts.ModuleKind.ES2022,
+      moduleResolution: ts.ModuleResolutionKind.Bundler,
```
aqrln, line 24: Should we also test Node.js + ESM? (with both `module` and `moduleResolution` set to either `Node16` or `NodeNext`)

FGoessler (approved):
Nice work with in memory ts checks! 🎉

packages/client-generator-ts/tests/assert-typescript-is-valid.ts
jkomyno: I'd say no, unless we have strong reasons to do so. At least, not in this PR

PR #27561, opened 2025-07-01
chore: deprecate PrismaD1HTTP

jacek-prisma:
Deprecating the class since we can't deprecate the re-export. In the next release we can remove the re-export and undeprecate the class.

### DIFF
--- a/packages/adapter-d1/src/d1-http.ts
+++ b/packages/adapter-d1/src/d1-http.ts
@@ -295,6 +295,7 @@ export class PrismaD1HTTPAdapter extends D1HTTPQueryable implements SqlDriverAda
   }
 }
 
+/** @deprecated Use PrismaD1 instead */
 export class PrismaD1HTTPAdapterFactory implements SqlMigrationAwareDriverAdapterFactory {
   readonly provider = 'sqlite'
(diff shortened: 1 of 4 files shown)


FGoessler (commented):
We are also missing a reexport of `PrismaD1Http` in `index.ts` so users can actually import this.

jacek-prisma:
@FGoessler it's already there:
https://github.com/prisma/prisma/blob/196c4d068333704508068213633243d9530eeb45/packages/adapter-d1/src/d1.ts#L9

FGoessler:
> @FGoessler it's already there:
> 
> https://github.com/prisma/prisma/blob/196c4d068333704508068213633243d9530eeb45/packages/adapter-d1/src/d1.ts#L9

Yes, but the exported file according to package.json is https://github.com/prisma/prisma/blob/06fe03625b1fcbcfb1ed2362e3f9d23f4fee257e/packages/adapter-d1/src/index.ts where it's only exporting `PrismaD1`.

jacek-prisma:
@FGoessler you're right, I didn't notice, re-exported it now

aqrln (approved):

PR #27566, opened 2025-07-01
fix(client-engine-runtime): use `db.system.name=cockroachdb` in tracing

aqrln:
`@prisma/client-engine-runtime` used `provider` as defined by the driver adapter and not the schema in its tracing implementation. This caused the `db.system.name` span attribute to be wrong for CockroachDB.

Now `QueryInterpreter` and `TransactionManager` accept a `provider` value defined in the schema. This property is optional and a fallback value is inferred from the driver adapter if not provided for compatibility with the QE test setup and with the query plan executor for Accelerate (the latter doesn't even have a schema).

QE with CockroachDB and `pg` driver adapter needs to be fixed separately on the engine side.

Closes: https://linear.app/prisma-company/issue/ORM-1144/fix-reported-dbsystem-for-pg-used-with-cockroachdb

### DIFF
--- a/packages/client-engine-runtime/package.json
+++ b/packages/client-engine-runtime/package.json
@@ -35,6 +35,7 @@
     "uuid": "11.1.0"
   },
   "devDependencies": {
+    "@prisma/generator": "workspace:*",
     "@types/jest": "29.5.14",
     "@types/node": "18.19.76",
     "jest": "29.7.0",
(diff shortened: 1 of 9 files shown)


jkomyno (approved):

packages/client-engine-runtime/package.json
```diff
@@ -35,6 +35,7 @@
     "uuid": "11.1.0"
   },
   "devDependencies": {
+    "@prisma/generator": "workspace:*",
```
aqrln, line 38: this may break the pnpm workspace in QE tests, need to check that

aqrln: https://github.com/prisma/prisma-engines/pull/5509 needs to be merged first

Issue #23495, opened 2024-03-14
Critical dependency Webpack warning and missing source map in `node_modules/@prisma/client/runtime/library.js`


ernestostifano:
### Bug description

When bundling my application with Webpack, I get the following 3 warnings:

```bash
WARNING in ../../node_modules/@prisma/client/runtime/library.js 12:37424-37431
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ../../node_modules/@prisma/client/runtime/library.js 12:37438-37445
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ../../node_modules/@prisma/client/runtime/library.js
Module Warning (from ../../node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from '/repositories/my-repo/node_modules/@prisma/client/runtime/library.js.map' file: Error: ENOENT: no such file or directory, open '/repositories/my-repo/node_modules/@prisma/client/runtime/library.js.map'
```

The third one, about the source map, is simple: the file declares a source map, but the source map is not included in the package:

<img width="1106" alt="Screenshot 2024-03-14 at 19 15 09" src="https://github.com/prisma/prisma/assets/34403992/8e537c9c-71ab-4350-ae07-455bcebbb655">

For the moment, I just suppressed that specific warning in that specific file using this Webpack config:

```js
{
        ignoreWarnings: [
            {
                module: /node_modules[\\/]@prisma[\\/]client[\\/]runtime[\\/]library\.js/,
                message: /Failed to parse source map/
            }
        ]
}
```

As for the other two warnings, unfortunately, since the code of that file is minified, I could not find exactly which `require` is causing the issue. The odd thing is that using a regexp I could not find non-static requires.

Obviously, I feel less confident to suppress these two warnings compared with the other one.

Any thoughts?

### How to reproduce

It is enough to import "@prisma/client" and try building with Webpack.

### Expected behavior

No warnings building with Webpack

### Prisma information

N/A

### Environment & setup

- OS: macOS 11.7.3
- Database: N/A
- Node.js version: v21.6.1

### Prisma Version

```
"@prisma/client": "5.11.0"
```

VittoriDavide:
having same issue, any updates @Weakky ?

rujorgensen:
I'm seeing this after switching from webpack to rspack in a NestJS project.

yorch:
[...]

PR #27567, opened 2025-07-01
chore(client): include runtime sourcemaps in the published package

rockingskier:
I've been suffering from https://github.com/prisma/prisma/issues/23495 and am searching for solutions

From what I can tell, the runtime directory is generated with [sourcemaps enabled](https://github.com/prisma/prisma/blob/main/packages/client/helpers/build.ts#L58), but the sourcemaps are then [not published](https://github.com/prisma/prisma/blob/main/packages/client/package.json#L188).  This leads to the source referencing the sourcemaps, but failing to find them.

Feels like there are two options:
a) publish the sourcemaps (this PR)
b) don't generate the sourcemaps - probably done [here](https://github.com/prisma/prisma/blob/main/packages/client/helpers/build.ts#L58) - I'm also open to this if it removes the warning

Consider this a conversation starter/[continuation](https://github.com/prisma/prisma/issues/23495) for the correct solution.

If the sourcemaps do get published, I _think_ developers will want/need to use the `PRISMA_COPY_RUNTIME_SOURCEMAPS` envvar to copy the sourcemaps to remove the warning.  Might even make sense making this the default

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -185,7 +185,6 @@
   "files": [
     "README.md",
     "runtime",
-    "!runtime/*.map",
     "scripts",
     "generator-build",
     "edge.js",


aqrln (approved):

aqrln:
Looks good to me 👍

PR #27576, opened 2025-07-02
fix(cli): always return valid JSON from MCP server

luanvdw:
/integration

### Problem
The MCP server was causing "Unexpected end of JSON input" errors in Claude Desktop due to stdout pollution. The MCP protocol requires stdout to contain only valid JSON-RPC messages, but console output from the CLI was corrupting the protocol stream.

### Solution
This PR implements a targeted console redirection that only applies when running in MCP mode (process.argv.includes('mcp')). All console methods (log, warn, info, debug) are redirected to stderr to preserve stdout for clean JSON-RPC communication.

### Key changes:
- Added conditional console redirection that only activates for MCP commands
- Preserves normal CLI console output for all other commands
- Maintains all existing MCP tool functionality

### Testing
- ✅ Confirmed working with Claude Desktop (no more JSON parsing errors)
- ✅ All MCP tools function correctly

### Technical Context

We explored several approaches including scoped console management within the MCP class, but discovered that stdout pollution occurs at module import time, before any class methods execute. The current solution applies console redirection early enough to catch all sources of output pollution while being targeted to MCP mode only.

### Future Work
We've documented a long-term solution internally and will address it in follow-up PRs to avoid scope creep while ensuring immediate compatibility with Claude Desktop.


packages/cli/src/MCP.ts
```diff
@@ -8,9 +8,29 @@ import { z } from 'zod'
 import { version } from '../package.json'
 import { createHelp } from './platform/_lib/help'
 
-async function runCommand({ args, cwd }: { args: string[]; cwd: string }) {
-  const result = await execa.node(process.argv[1], args, { cwd })
-  return `${result.stdout}\n${result.stderr}`
+// Only apply console redirection when running in MCP mode
+// This prevents stdout pollution that breaks MCP's JSON-RPC protocol
+if (process.argv.includes('mcp')) {
+  console.log = (...a: unknown[]) => process.stderr.write(a.map(String).join(' ') + '\n')
```
aqrln: This technically breaks much of the `console.log` functionality: this implementation will only work when printing primitive values and not using any format specifiers. Should it be just this instead?

```suggestion
  console.log = console.error.bind(console)
```

aqrln: Also, where do those logs come from? Can we just print them to stderr in the first place if they're in our code instead of monkey-patching `console`?

luanvdw: @aqrln you're right, thanks! I've modified the console redirection with your proposed changes 🙏

> Also, where do those logs come from? Can we just print them to stderr in the first place if they're in our code instead of monkey-patching console?
The real source of the stdout pollution is that when the MCP server is running it prints an empty string.

  1. MCP.parse() returns '' (empty string)
  2. bin.ts line 190 does console.log(result)
  3. console.log('') outputs an empty string to stdout
  4. This breaks the MCP JSON-RPC protocol

I agree with you, the real fix should be to fix this at the source instead of monkey-patching console.

aqrln (approved):

PR #27592, opened 2025-07-04
fix: fix prisma#27580

jacek-prisma:
[ORM-1184](https://linear.app/prisma-company/issue/ORM-1184/fix-bytes-batching-in-client-runtime)

Fixes https://github.com/prisma/prisma/issues/27580

Also refactored the code a bit to avoid stringifying everything and instead convert them to common types

### DIFF
--- a/packages/client-engine-runtime/src/utils.ts
+++ b/packages/client-engine-runtime/src/utils.ts
@@ -35,20 +35,81 @@ export function doKeysMatch(lhs: {}, rhs: {}): boolean {
   const smallerKeyList = lhsKeys.length < rhsKeys.length ? lhsKeys : rhsKeys
 
   return smallerKeyList.every((key) => {
-    if (typeof lhs[key] !== typeof rhs[key]) {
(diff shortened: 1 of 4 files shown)


packages/client-engine-runtime/src/utils.ts
```diff
@@ -35,22 +35,83 @@ export function doKeysMatch(lhs: {}, rhs: {}): boolean {
...
 }
 
+function asDecimal(value: unknown): Decimal | undefined {
+  if (Decimal.isDecimal(value)) {
+    return value
+  } else if (typeof value === 'number' || typeof value === 'string') {
+    return new Decimal(value)
+  } else {
+    return
+  }
```
jkomyno, line 70: Nit, this branch could be removed. This comment applies to other functions down here as well

jkomyno (approved):

packages/client-engine-runtime/src/utils.ts
jacek-prisma: typescript complains when there's no explicit `return`

PR #27594, opened 2025-07-04
feat(client): support Accelerate in ClientEngine

aqrln:
Closes: https://linear.app/prisma-company/issue/ORM-1138/allow-using-query-compiler-with-accelerate-in-prisma-client

### DIFF
--- a/packages/internals/src/utils/prismaPostgres.ts
+++ b/packages/internals/src/utils/prismaPostgres.ts
@@ -2,16 +2,31 @@ export const PRISMA_POSTGRES_PROVIDER = 'prisma+postgres'
 
 export const PRISMA_POSTGRES_PROTOCOL = `${PRISMA_POSTGRES_PROVIDER}:`
 
-export function isPrismaPostgres(connectionString?: string | URL): connectionString is string | URL {
+declare const prismaPostgresBrand: unique symbol
+declare const prismaPostgresDevBrand: unique symbol
+
+export type PrismaPostgresUrl<T extends string | URL> = T & {
+  readonly [prismaPostgresBrand]: true
+}
+
+export type PrismaPostgresDevUrl<T extends string | URL> = PrismaPostgresUrl<T> & {
+  readonly [prismaPostgresDevBrand]: true
+}
+
+export function isPrismaPostgres<T extends string | URL>(
+  connectionString?: T,
+): connectionString is PrismaPostgresUrl<T> {
   return connectionString?.toString().startsWith(`${PRISMA_POSTGRES_PROTOCOL}//`) ?? false
 }
 
-export function isPrismaPostgresDev(connectionString?: string | URL): connectionString is string | URL {
+export function isPrismaPostgresDev<T extends string | URL>(
+  connectionString?: T,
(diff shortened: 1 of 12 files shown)


packages/internals/src/utils/prismaPostgres.ts
```diff
@@ -2,16 +2,35 @@ export const PRISMA_POSTGRES_PROVIDER = 'prisma+postgres'
...
+  readonly [prismaPostgresBrand]: true
+}
+
+export type PrismaPostgresDevUrl<T extends string | URL> = PrismaPostgresUrl<T> & {
+  readonly [prismaPostgresDevBrand]: true
+}
+
+export function isPrismaPostgres<T extends string | URL>(
+  connectionString?: T,
+): connectionString is PrismaPostgresUrl<T> {
```
aqrln, line 18: The signatures of these functions were incorrect in a way that prevented them from being used with if-else statements, which I needed in `getUrlAndApiKey`.

The root cause was that the type assertion expressed a consequence implied by the condition that was checked but not the actual condition itself. If an optional connection string is a Prisma Postgres connection string, then it indeed implies it was defined. However, the reverse isn't true: if the connection string being tested is not a Prisma Postgres connection string, it does not necessarily mean that it was undefined! It could have just not been a Prisma Postgres URL but still a defined Accelerate URL.

This logic mistake broke the type of the connection string whenever `!isPrismaPostgres(url)` or `!isPrismaPostgresDev(url)` would be used. In my case, I passed a non-nullable `URL` value to `isPrismaPostgresDev` function, which caused TypeScript to consider the `else` branch unreachable and narrow the URL type to `never` (since it could not be `undefined` by its definition).

To fix this and make the signatures type safe and logically sound while preserving the `undefined`-removing semantics, the actual condition being checked had to be moved to the type level: these functions check that the argument is defined _and_ is a particular kind of a URL. Then the reverse condition becomes correct as well: if the function returns false, then the argument was either undefined, or a different kind of a URL. The actual semantics of these "kinds" and the mechanics of checking don't matter on the type level as long as the subtype relations are correct and express the underlying logical propositions (any PPg URL is a URL, and any local PPg URL is a PPg URL, but most URLs are generally neither PPg URLs nor local PPg URLs).

The easiest way to do this is to use branded types.

packages/internals/src/utils/prismaPostgres.ts
```diff
@@ -2,16 +2,35 @@ export const PRISMA_POSTGRES_PROVIDER = 'prisma+postgres'
...
+  connectionString?: T,
+): connectionString is PrismaPostgresDevUrl<T> {
   if (!isPrismaPostgres(connectionString)) {
     return false
   }
 
   const { host } = new URL(connectionString)
 
-  return host.includes('localhost') || host.includes('127.0.0.1')
+  return host.includes('localhost') || host.includes('127.0.0.1') || host.includes('[::1]')
```
aqrln, line 31: Unrelated fix along the way, can be extracted to a separate PR

jkomyno: Doesn't need as separate PR, dw, let's do it here.

jkomyno (approved):

Issue #27324, opened 2025-06-03
Cannot bundle the new `prisma-client` generator with esbuild


aniravi24:
### Bug description

When I use the new `prisma-client` generator with esbuild, I get an error like this the moment I try to run the bundled javascript.

```
node:internal/url:1507
    throw new ERR_INVALID_ARG_TYPE('path', ['string', 'URL'], path);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of URL. Received undefined
    at Object.fileURLToPath (node:internal/url:1507:11)
    at Object.<anonymous> (/home/user/projects/prisma/prisma-bug-template/postgis-express/build/index.js:23559:36)
    at Module._compile (node:internal/modules/cjs/loader:1730:14)
    at Object..js (node:internal/modules/cjs/loader:1895:10)
    at Module.load (node:internal/modules/cjs/loader:1465:32)
    at Function._load (node:internal/modules/cjs/loader:1282:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

### Severity

🚨 Critical: Data loss, app crash, security issue

### Reproduction

https://github.com/aniravi24/prisma-new-client-esbuild-repro

Clone this repo and follow the instructions. This is based on the Prisma [postgis-express example](https://github.com/prisma/prisma-examples/tree/latest/orm/postgis-express).

Here is the commit that shows exactly what I changed.

https://github.com/aniravi24/prisma-new-client-esbuild-repro/commit/5495c3a816697fa5fa2288eb45bb424d90a9c025

### Expected vs. Actual Behavior

Server crashes immediately.

I would expect the server to start up normally.

### Frequency

Consistently reproducible

### Does this occur in development or production?

Both development and production

### Is this a regression?

It's always been broken since the new client was announced.

### Workaround

I have to use the older `prisma-client-js` generator.

### Prisma Schema & Queries

```prisma
// Add relevant schema.prisma snippet
```
```ts
// Add relevant Prisma Client query
```

### Prisma Config

```ts
// Add your `prisma.config.ts`
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

[...]

PR #27610, opened 2025-07-09
fix(client): fixed queryCompiler on esbuild + cjs

jkomyno:
Consider the following `tsconfig.json` file:

```json
{
  "compilerOptions": {
    "strict": true,
    "module": "node18",
    "target": "es2022",
    "moduleResolution": "node16",
    "sourceMap": true,
    "outDir": "dist",
    "lib": [
      "ES2022",
      "DOM"
    ],
    "esModuleInterop": true,
    "isolatedModules": true,
    "resolveJsonModule": true
  }
}
```

Consider the following `esbuild` compilation file for Node.js + CJS:

```typescript
// esbuild.node.cjs.ts
import esbuild, { type BuildOptions } from 'esbuild'

const buildOpts = {
  bundle: true,
  sourcemap: true,
  entryPoints: ['./src/index.ts'],
  outdir: './dist',
  outExtension: {
    '.js': '.cjs',
  },
  format: 'cjs',
  platform: 'node',
  target: 'es2022',
} satisfies BuildOptions

async function build() {
  await esbuild.build(buildOpts);
}

build()
```

By default, after an `esbuild` pass, the bundled module looks at the following fields under `exports.<path>` in the transitive dependency's `package.json`:
- `node.import`
- `node.default`
- `import`
- `default`

Note how `node.require` and `require` are never reached.

> [!TIP]
> One could address this problem by adding `{ conditions: ['node', 'require'] }` to `buildOpts` in `esbuild.node.cjs.ts`.

Since we are mapping the `exports["./runtime/client"].import` field of `@prisma/client` to a generated file designed for ESM only, we were hitting the error shown in https://github.com/prisma/prisma/issues/27324 in CJS + `queryCompiler` context. Namely, we were mistakenly loading `NODE_ESM_BANNER` in a CJS project transpiled with `esbuild`.

Note about the risk of introducing a regression:
- `vite` never looks for the `node` key in exports map
- `webpack` prefers `node.require` to `require`, but it doesn't break when built locally or deployed ([proof](https://nextjs-starter-webpack.vercel.app/)) 
- `turbopack` prefers `node.import` to `import`, but it doesn't break when built locally or deployed ([proof](https://nextjs-starter-turbopack-three.vercel.app//))

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -92,6 +92,10 @@
     },
     "./runtime/client": {
       "types": "./runtime/client.d.ts",
+      "node": {
+        "require": "./runtime/client.js",
+        "default": "./runtime/client.js"
+      },
       "require": "./runtime/client.js",
       "import": "./runtime/client.mjs",
       "default": "./runtime/client.mjs"


jkomyno:
Integration release in progress: https://github.com/prisma/prisma/actions/runs/16160970278/job/45612488787 (6.12.0-dev.19).

FGoessler (approved):

PR #27630, opened 2025-07-11
feat(adapter): add PrismaMariaDb.fromUrl() for connection string support

ankur-arch:
### 📝 PR Description:

This PR introduces a static `fromUrl()` method on the `PrismaMariaDbAdapterFactory` class that allows users to instantiate the adapter factory directly from a database connection string (e.g., `DATABASE_URL` from environment variables).

#### 🔧 What’s Added:

* `fromUrl(connectionString, options?)` static method
* Support for parsing protocol, host, port, user, password, database, and query params

#### ✅ Benefits:

* Simplifies adapter setup using a single environment variable
* Aligns with how many users configure DB access in production

#### 📌 Example Usage:

```ts
import { PrismaMariaDb } from '@prisma/adapter-mariadb'
import { PrismaClient } from './generated/prisma'

const adapter = PrismaMariaDb.fromUrl(process.env.DATABASE_URL!)
```

### DIFF
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -15,6 +15,21 @@ import * as mariadb from 'mariadb'
 import { name as packageName } from '../package.json'
 import { mapArg, mapColumnType, mapRow, typeCast } from './conversion'
 import { convertDriverError } from './errors'
+import { URL } from 'url'
+
+function parseMySQLConnectionString(connStr: string) {
+  const url = new URL(connStr)
+  return {
+    protocol: url.protocol.replace(':', ''),
+    host: url.hostname,
+    port: url.port ? parseInt(url.port, 10) : undefined,
+    user: url.username,
+    password: url.password,
+    database: url.pathname.replace('/', ''),
+    // Optional: support search params like ?ssl=true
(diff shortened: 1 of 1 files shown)


jy-joung-bitor:
I didn't know `fromUrl` function, but you can do that like this:

```ts
const adapter = new PrismaMariaDb(env["DATABASE_URL"]!);
```

And it works.

tensordreams:
Thanks for the contribution, and apologies for the slow response here!

This is now superseded by #27970 ("feat(mysql driver adapter): ORM-1137 support mysql connection strings"), which landed connection-string support directly on the `PrismaMariaDbAdapterFactory` **constructor** rather than via a separate `fromUrl()` method:

```ts
import { PrismaMariaDb } from '@prisma/adapter-mariadb'
import { PrismaClient } from './generated/prisma'

const adapter = new PrismaMariaDb(process.env.DATABASE_URL!)
```

The merged implementation also handles a few extra cases on top of what this PR covered: defaulting `prepareCacheLength`, rewriting `mysql://` connection strings to `mariadb://`, and guarding against credential leakage in parse-error messages.

Since the use case is fully covered now, I'm closing this one. Thanks again for taking the time to open it!

PR #27631, opened 2025-07-11
chore(cli): Moved MCP.ts to mcp directory, added mcp/README.md, and r…

tylerhogarth:
- Moved MCP.ts into its own folder in the cli directory.
- Added a dedicated README for MCP that we can direct to in other repos
- Removed PDP MCP tools which will be managed exclusively by the remote MCP server


tylerhogarth:
FYI @luanvdw

packages/cli/src/mcp/README.md
```diff
@@ -0,0 +1,13 @@
+# Prisma MPC
```
aqrln:
```suggestion
# Prisma MCP
```

jkomyno (approved):

language-tools
aqrln, line 1: What is this added for?

PR #27639, opened 2025-07-13
fix(adapter-pg): get correct database name from error code 28000

barryhagan:
Error code 28000 is returned from a Postgres server when client authentication fails and turns into Prisma error P1010.  This is often the result of the server forcing SSL while the client cannot verify the server's cert and is not using `sslmode=no-verify`.

A Postgres server will return multiple error messages for code 28000 including:

- connection requires a valid client certificate
- pg_hba.conf rejects replication connection for host "%s", user "%s", %encryption_state
- pg_hba.conf rejects connection for host "%s", user "%s", database "%s", %encryption_state
- no pg_hba.conf entry for replication connection from host "%s", user "%s", %encryption_state
- no pg_hba.conf entry for host "%s", user "%s", database "%s", %encryption_state

The last message is used when SSL verification fails and the basis for the hostname parsing in the existing code, however, host here is the client IP resolved from the server not the database.

This PR updates the parsing to find the correct database name when it is available.

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -1055,6 +1055,9 @@ jobs:
         name: 'adapter-d1'
         working-directory: packages/adapter-d1
 
+      - run: pnpm run test
+        name: 'adapter-pg'
+        working-directory: packages/adapter-pg
   #
   # Run type performance benchmark tests
   #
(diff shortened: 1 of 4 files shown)


jacek-prisma:
Hi @barryhagan ,
Since this isn't covered by any existing tests, could you add an `errors.test.ts` with some basic unit tests that call `convertDriverError`?

Then add it to the CI:
```yml
      - run: pnpm run test
        name: 'adapter-pg'
        working-directory: packages/adapter-pg
```
in `test-template.yml > jobs > driver-adapter-unit-tests > steps`

barryhagan:
tests added and passing CI.

jacek-prisma (approved):
lgtm

PR #27640, opened 2025-07-13
fix: resolve wrong type inference for AND clauses with include

nehaaprasad:
Hey, I just made a Pull Request!

## What does this PR do?

Fixes #27637

Fixes a TypeScript type inference bug where using `AND` clauses in `where` conditions with `include` would incorrectly return the base model type instead of including the specified relations.

## How to test
```typescript
const assets = await prisma.asset.findMany({
  where: {
    AND: [
      { date: { not: null } },
      { date: { gte: "2024-01-01", lte: "2024-12-31" } }
    ]
  },
  include: { items: true }
})

// Before: assets[0].items would show TypeScript error
// After: assets[0].items is correctly typed

## Root Cause
The `GetFindResult` type was using `Equals<A, any> extends 1` which was too broad and caught cases where TypeScript inferred complex types as `any` due to complexity, causing fallback to `DefaultSelection` (base model without includes).

## Solution
Replaced the type check with a more precise one that only falls back to `DefaultSelection` when the type is explicitly `any`, not when it's complex:


jacek-prisma:
Hi @naaa760,
It looks like your change has caused a number of tests to fail, could you look into them?

aqrln:
ping @naaa760

nehaaprasad:
@jacek-prisma  and @aqrln 
I tried to resolve solve the test cases fail , and some fail due to dependency.

aqrln:
> and some fail due to dependency

What do you mean, which ones?

tensordreams:
Hi @nehaaprasad — thanks for this, and apologies that it sat for so long.

We are closing this one as inactive: the CLA has not been signed, and the 285 failing jobs raised in review were never explained or resolved.

PR #27643, opened 2025-07-14
fix(config): speed up tests on windows

jkomyno:
This PR tries to speed up `prisma.config.ts` tests on Windows by using `path.join`.
It attempts to prevent errors such as https://github.com/prisma/prisma/actions/runs/16222312665/job/45805707895?pr=27635.


jkomyno:
Apparently, we still need to bump the timeout

FGoessler (approved):

PR #27644, opened 2025-07-14
fix(client): never ignore the configured adapter in ClientEngine

aqrln:
Fixes: https://linear.app/prisma-company/issue/ORM-1210/fix-qcda-regression-caused-by-accelerate-support

/integration

### DIFF
--- a/packages/client/src/runtime/core/init/getEngineInstance.ts
+++ b/packages/client/src/runtime/core/init/getEngineInstance.ts
@@ -59,11 +59,17 @@ export function getEngineInstance({ copyEngine = true }: GetPrismaClientConfig, 
   // - Delete DataProxyEngine and all related files
   // - Update the DataProxy tests to use the /wasm endpoint, but keep ecosystem-tests as they are
 
+  // When a local driver adapter is configured, the URL from the datasource
+  // block in the Prisma schema is no longer relevant as driver adapters don't
+  // use it. Therefore, a configured driver adapter takes precedence over the
+  // Accelerate or PPg URL in the schema file.
+  const clientEngineUsesRemoteExecutor = (isUsing.accelerate || isUsing.ppg) && !isUsing.driverAdapters
+
   if (TARGET_BUILD_TYPE === 'react-native') return new LibraryEngine(engineConfig)
   else if (clientEngineConfigured && TARGET_BUILD_TYPE === 'client')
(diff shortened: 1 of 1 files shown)


packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -59,11 +59,17 @@ export function getEngineInstance({ copyEngine = true }: GetPrismaClientConfig,
   // - Delete DataProxyEngine and all related files
   // - Update the DataProxy tests to use the /wasm endpoint, but keep ecosystem-tests as they are
 
+  // When a local driver adapter is configured, the URL from the datasource
+  // block in the Prisma schema is no longer relevant as driver adapters don't
+  // use it. Therefore, a configured driver adapter takes precedence over the
+  // Accelerate or PPg URL in the schema file.
```
jacek-prisma, line 65: maybe it's worth warning the user if they provide both

jkomyno: > maybe it's worth warning the user if they provide both
This means that every user following our examples would have warnings. I don't like that tbh

For instance, in https://github.com/prisma/prisma-examples/blob/latest/generator-prisma-client/nextjs-starter-turbopack/prisma/schema.prisma#L5-L15:
- `DATABASE_URL` is only used by `prisma migrate`
- `DIRECT_URL` is used by Prisma Client at runtime and by `prisma db seed`

aqrln: @jkomyno

> DATABASE_URL is only used by prisma migrate

I don't think that's the case. `DIRECT_URL` [is set as the `directUrl` in the schema file](https://github.com/prisma/prisma-examples/blob/latest/generator-prisma-client/nextjs-starter-turbopack/prisma/schema.prisma#L14), so Migrate will ignore `url` and use `directUrl` instead. `DATABASE_URL` does not appear to be used in this example at all, it's only set as the `url` in the datasource block for seemingly no reason.

aqrln: I'm also not quite following why you would need to use both URLs even if this was true. You either use Accelerate with PPg or you don't, there's no point in mixing both ways.

jkomyno (approved):

PR #27652, opened 2025-07-15
feat(multi schema): ORM-1107 tables.external config

FGoessler:
This PR adds the `tables.external` config attribute that is used to configure which tables are externally managed and hence shall not be included during migration creation.

For the actual feature implementation in the engine see https://github.com/prisma/prisma-engines/pull/5529.


packages/migrate/src/types.ts
```diff
@@ -149,12 +150,14 @@ export namespace EngineArgs {
 
   export interface DevDiagnosticInput {
     migrationsList: MigrateTypes.MigrationList
+    schemaFilter: MigrateTypes.SchemaFilter | null
```
FGoessler: ℹ️ Yes there is a naming inconsistency here. Already fixing this in https://github.com/prisma/prisma-engines/pull/5533.

FGoessler:
Failing CI right now as this PR now also includes the interface changes for:
- https://github.com/prisma/prisma-engines/commit/ca9f504d0ba857487374e18da1a08d421d5d48dd
- https://github.com/prisma/prisma-engines/commit/1bb32ed4b585343f36604dde1aee6721602bc1ce
- https://github.com/prisma/prisma-engines/commit/e25e9a0d14a343c8d1495c0d2444a1bd271e8176

packages/migrate/src/Migrate.ts
```diff
@@ -26,26 +26,30 @@ interface MigrateSetupInput {
   migrationsDirPath?: string
   enabledPreviewFeatures?: string[]
   schemaContext?: SchemaContext
+  schemaFilter?: MigrateTypes.SchemaFilter
```
jkomyno, line 29: Does `schemaFilter` *really* need to be `nullable` or undefined? This will cause a chain of `schemaFilter?['prop'] property key access when trying to read such filter.

It seems more explicit to me to pass a default filter object, like the following:
```typescript
const defaultSchemaFilter = {
  externalTables: [],
  // in the future:
  exclude: [],
} satisfies MigrateTypes.SchemaFilter
```

packages/migrate/src/__tests__/fixtures/external-tables/prisma.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from '@prisma/config/src'
```
jkomyno:
```suggestion
import { defineConfig } from '@prisma/config'
```

packages/migrate/src/Migrate.ts
jkomyno: In fact, you used this in other parts of this PR:

```typescript
const schemaFilter: MigrateTypes.SchemaFilter = {
  externalTables: config.tables?.external ?? [],
}
```

jkomyno: In general, if we can avoid introducing `null`s, let's not.

FGoessler: It's something that is by far not used by all migrate related commands. In fact most `Migrate.setup` calls will not need this parameter. Hence I followed the existing pattern of having all attributes of this input object optional.

I could make leave it optional here on the input object but not make it optional on the `Migrate` class and create a default filter during the setup method?

Then all relevant read access "downstream" does not have to deal with the optionality (although that is also just in this class here).

FGoessler: I might change the interface between TS and Rust again in a dedicated PR to remove the optionality in the input types there. Was using optional there at first because it might ease the adoption (similar to how you would add a new field to a REST api).

FGoessler: See the changes I just pushed

jkomyno (approved):

PR #27657, opened 2025-07-15
feat(config): port test setup from Jest to Vitest

jkomyno:
This PR deprecates https://github.com/prisma/prisma/pull/27656.

This PR:
- closes [ORM-1220](https://linear.app/prisma-company/issue/ORM-1220/prismaconfigts-tech-debt-port-prismaconfig-tests-to-vitest)
- moves `@prisma/config` from Jest to Vitest, preserving Jest's unique `moduleMapper` emulation in `jiti`
- adds `vitestContext` to `@prisma/get-platform/src/test-utils/vitest`, for internal use only
  - Note: in Prisma 7, once we stop needing native platform resolution, we should rename `@prisma/get-platform` to something like `@prisma/test-utils`. This package already handles test things that are not pertaining to "platform" per se.

### DIFF
--- a/packages/config/vitest.config.ts
+++ b/packages/config/vitest.config.ts
@@ -1,8 +1,12 @@
-import tsconfigPaths from 'vite-tsconfig-paths'
 import { defineConfig } from 'vitest/config'
 
 export default defineConfig({
-  plugins: [tsconfigPaths({ ignoreConfigErrors: true })],
+  resolve: {
+    alias: {
+      'src/*': './src/*',
+      'test-utils/*': './src/__tests__/_utils/*',
+    },
+  },
   test: {
     include: ['**/*.test.ts'],
     unstubEnvs: true,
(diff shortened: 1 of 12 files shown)


packages/config/vitest.config.ts
```diff
@@ -0,0 +1,15 @@
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  resolve: {
+    alias: {
+      'src/*': './src/*',
+      'test-utils/*': './src/__tests__/_utils/*',
+    },
```
FGoessler, line 8: ❓ Do we really need these aliases? I see that we used them before via jest as well. However, I am usually not a big fan of resolve aliases as they in my experience more often than not cause issues down the line as they create a level of indirection that can cause confusion or at times lack proper tool support or trigger edge cases.

More a nitpick though. ;)

FGoessler (approved):

packages/config/vitest.config.ts
jkomyno: We do, we can't otherwise have module imports within our fixtures tests.

PR #27669, opened 2025-07-16
fix: fix transaction manager race condition

### LINKED ISSUE #27660: Transaction API Error: "Closed transaction found in active transactions map" with Query Compiler and Driver Adapter
### Bug description

When using the query compiler with PrismaPg driver adapter, I'm encountering a transaction error on a specific query in production only.

### Error
```
Transaction API error: Internal Consistency Error: Closed transaction found in active transactions map
```

### Failing Query
```
const piecesOfInformation = await coreBypassRlsPrisma.pieceOfInformation.findMany({
  where: {
    processingStatus: 'PENDING'
  }
});
```

Other queries on the same table and all other tables work fine with the same client.

### Severity

### Reproduction

The issue only happens in our production environment. I can't replicate this problem in my local dev and our test environment. Our test setup the same as production. The only difference is the data. 

The `findMany` should return 300 records. A SQL query on the table completes in less than 20ms, so the amount of data shouldn't be the issue either.

I also tried setting a timeout and wait time on the transaction, but the issue persists.

Expected: That prisma returns the rows
Actual: Prisma client throws transaction error

### Frequency

Consistently reproducible

### Is this a regression?

No, the same issue persists with 6.10 as well.

### Workaround

Switching back to the Rust-based client resolves the issue:
```
// This works without errors
return new PrismaClientBase({ datasourceUrl: connectionString }).$extends(
  useExtendedPrismaModel()
);

### Prisma Schema & Queries

### Prisma Schema

```prisma
generator client {
  provider               = "prisma-client"
  output                 = "./generated"
  previewFeatures        = ["postgresqlExtensions", "relationJoins", "queryCompiler"]
  runtime                = "nodejs"
  moduleFormat           = "esm"
  generatedFileExtension = "ts"
  importFileExtension    = "ts"
}

generator json {
  provider  = "prisma-json-types-generator"
  output    = "./generated"
  namespace = "PrismaJson"
}

datasource db {
  provider   = "postgresql"
  url        = env("CORE_DATABASE_URL")
  directUrl  = env("CORE_DATABASE_DIRECT_URL")
  extensions = [pgsodium(schema: "pgsodium"), supabase_vault, uuid_ossp(map: "uuid-ossp", schema: "extensions"), pgvector(map: "vector")]
}

model PieceOfInformation {
  id String @id @default(dbgenerated("typeid_generate_text('infopiece'::text)"))

  ownedByOrganization   Organization @relation(fields: [ownedByOrganizationId], references: [id])
  ownedByOrganizationId String       @default(dbgenerated("current_organization_id()"))

  processingStatus PieceOfInformationProcessingStatus @default(PENDING)
  status           LifecycleStatus                    @default(ACTIVE)
  name             String

  createdOn DateTime @default(now())
  updatedOn DateTime @updatedAt

  @@index([ownedByOrganizationId])
  @@index([ownedByOrganizationId, status])
  @@index([ownedByOrganizationId, itemId])
  @@index([processingStatus])
}
```

### Client Initialization

```ts
const extendedPrismaClient = (connectionString: string) => {
  const pgAdapter = new PrismaPg({ connectionString });
  return new PrismaClientBase({ adapter: pgAdapter }).$extends(useExtendedPrismaModel());
};

const baseCoreByPassRlsPrisma = 
  globalThis.baseCoreByPassRlsPrisma ?? extendedPrismaClient(process.env.CORE_DATABASE_URL!);

export const coreBypassRlsPrisma = baseCoreByPassRlsPrisma.$extends(
  byPassRls()
) as unknown as CorePrismaClient;
```

### RLS Middleware (wraps all operations in transactions)

```ts
const rlsExtension = (claimSql: Prisma.Sql, user: SimpleAuthUser) => {
  const rlsStore = new AsyncLocalStorage<RlsStore>();

  return Prisma.defineExtension(client => {
    return client.$extends({
      name: 'rlsExtension',
      query: {
        $allModels: {
          async $allOperations({ args, query }) {
            const alreadyInRls = rlsStore.getStore()?.alreadyInRls;
            if (alreadyInRls) return query(args);

            const [, result] = await client.$transaction([
              client.$executeRaw(claimSql),
              query(args)
            ]);

            return result;
          }
        }
     }
   }
}
```

### Environment & Setup

- **OS:** Debian
- **Database:** PostgreSQL
- **Node.js version:** 22

### Prisma Version

```
 6.11
```

jacek-prisma:
Should fix https://github.com/prisma/prisma/issues/27660

The changes here ensure that we never have a closed transaction inside of `this.transactions` in between any `await` points, so users don't end up seeing `TransactionInternalConsistencyError`s.
Also adds a `closing` state that ensures we do not attempt to rollback multiple times.

### DIFF
--- a/packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
+++ b/packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
@@ -95,7 +95,8 @@ export class TransactionManager {
     this.transactions.set(transaction.id, transaction)
 
     // Start timeout to wait for transaction to be started.
-    const startTimer = setTimeout(() => (transaction.status = 'timed_out'), validatedOptions.maxWait!)
+    let hasTimedOut = false
+    const startTimer = setTimeout(() => (hasTimedOut = true), validatedOptions.maxWait!)
 
     transaction.transaction = await this.driverAdapter.startTransaction(validatedOptions.isolationLevel)
 
@@ -104,13 +105,16 @@ export class TransactionManager {
     // Transaction status might have changed to timed_out while waiting for transaction to start. => Check for it!
     switch (transaction.status) {
       case 'waiting':
+        if (hasTimedOut) {
+          await this.closeTransaction(transaction, 'timed_out')
+          throw new TransactionStartTimeoutError()
+        }
+
         transaction.status = 'running'
         // Start timeout to wait for transaction to be finished.
         transaction.timer = this.startTransactionTimeout(transaction.id, validatedOptions.timeout!)
         return { id: transaction.id }
       case 'timed_out':
-        await this.closeTransaction(transaction, 'timed_out')
-        throw new TransactionStartTimeoutError()
       case 'running':
       case 'committed':
       case 'rolled_back':
@@ -202,8 +206,6 @@ export class TransactionManager {
   private async closeTransaction(tx: TransactionWrapper, status: 'committed' | 'rolled_back' | 'timed_out') {
     debug('Closing transaction.', { transactionId: tx.id, status })
 
-    tx.status = status
-
     try {
       if (tx.transaction && status === 'committed') {
         if (tx.transaction.options.usePhantomQuery) {
@@ -223,6 +225,7 @@ export class TransactionManager {
         }
       }
     } finally {
+      tx.status = status
       clearTimeout(tx.timer)
       tx.timer = undefined
 
--- a/packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
+++ b/packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
@@ -102,7 +102,7 @@ export class TransactionManager {
 
     clearTimeout(startTimer)
 
-    // Transaction status might have changed to timed_out while waiting for transaction to start. => Check for it!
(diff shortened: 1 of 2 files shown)


packages/client-engine-runtime/src/transactionManager/TransactionManager.ts
```diff
@@ -104,13 +105,17 @@ export class TransactionManager {
     // Transaction status might have changed to timed_out while waiting for transaction to start. => Check for it!
```
aqrln: nit:

```suggestion
    // Transaction status might have timed out while waiting for transaction to start. => Check for it!
```
since we're not actually doing that by checking for the state transition to `timed_out` in `case 'waiting'` anymore

aqrln (approved):
LGTM with a nit

AjaniBilby:
I don't see a git tag for 6.12.1, did this merge make it into that update?
Or is this fix not published to npm yet?

Because I'm having this issue in production - and I'm trying to work out if I just need to get this patch into my system, or if possibly it's a new issue I need to investigate.

AjaniBilby:
Sorry, never mind, there is no 6.12.1 - idk what I was thinking of.

jacek-prisma:
@AjaniBilby it's currently fixed on dev, a new stable version with the fix will release next week

PR #27673, opened 2025-07-17
fix(adapter-pg): subscribe to pool error events to prevent node crashing

### LINKED ISSUE #27641: adapter-pg will crash Node process if there is an open connection when the database shuts down
### Bug description

When using adapter-pg, the `PrismaPg` object internally creates a `pg.Pool`, but it never subscribes to the error event from this emitter.

Since this is not handled, any error emitted from the pool will crash the entire Node process unless you have also wired up a default global handler like this:

```ts
process.on('uncaughtException', (err) => {});
```

At a minimum this needs to do some kind of handling and logging in the adapter, but likely should allow the caller to provide a callback handler as well.

Right now without a global handler, pool errors can crash your app.

### Severity

### Reproduction

- Create a project using the `driverAdapters` and `queryCompiler` features with a postgres database.
- In the configuration of the PrismaPg object, set `idleTimeoutMillis` to 0 to keep connections alive
- Do not set a global handler for the process 'uncaughtException' event
- Start your application and make at least on database query
- Stop your postgres database
- Crash

Expected: pg.Pool error events should be handled so they don't cause an uncaught exception from an async callback.

Actual: my application container crashed.

### Frequency

Consistently reproducible

### Is this a regression?

No

### Workaround

Configure a global uncaughtException handler.
```ts
process.on('uncaughtException', (err) => {});
```

### Prisma Schema & Queries

```prisma
generator client {
  output          = "../src/prisma-generated"
  previewFeatures = ["driverAdapters", "fullTextSearchPostgres", "queryCompiler", "views"]
  moduleFormat    = "esm"
}
```

### Prisma Config

_No response_

### Logs & Debug Info

_No response_

### Environment & Setup

- **OS:** macOS, CentOS
- **Database:** PostgreSQL
- **Node.js version:** 20.19.2

### Prisma Version

6.11.1

barryhagan:
fixes #27641

### DIFF
--- a/packages/adapter-pg/src/__tests__/pg.test.ts
+++ b/packages/adapter-pg/src/__tests__/pg.test.ts
@@ -1,7 +1,27 @@
 import { getLogs } from '@prisma/debug'
+import EventEmitter from 'events'
 import pg, { DatabaseError } from 'pg'
 
 import { PrismaPgAdapterFactory } from '../pg'
+
+jest.mock('pg', () => {
+  class MockPool extends EventEmitter {
+    end = jest.fn(() => {})
+    listenerCount(event: string | symbol): number {
+      return super.listenerCount(event)
+    }
+    on(event: string | symbol, listener: (...args: any[]) => void): this {
+      return super.on(event, listener)
+    }
+    emit(event: string | symbol, ...args: any[]): boolean {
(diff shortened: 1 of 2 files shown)


aqrln (approved):

aqrln:
Thank you!

PR #27682, opened 2025-07-17
fix(adapter-pg): Add handling of socket and TLS level errors

### LINKED ISSUE #27626: adapter-pg doesn't properly handle DB connection errors
### Bug description

The adapter / Prisma client doesn't throw an error at `await prisma.$connect()` if it can't connect to the server. You only get an error when trying to run a query. You would get a `PrismaClientKnownRequestError` with code set to `ECONNREFUSED`/`ENOTFOUND` as opposed to getting known error codes such as `P1000`, `P1001`, etc.

It seems to create the pool instance, but don't call `.connect()` off of it, so it doesn't fail on incorrect configurations:
https://github.com/prisma/prisma/blob/c6936205a0cfb0d56887733f84e061197efba37c/packages/adapter-pg/src/pg.ts#L221-L223

This behavior is different compared to using it without driver adapters. It also differs from some other adapters such as MSSQL:

https://github.com/prisma/prisma/blob/c6936205a0cfb0d56887733f84e061197efba37c/packages/adapter-mssql/src/mssql.ts#L159-L163

### Severity

### Reproduction

- Create a new project with Prisma
- Enable `queryCompiler` and `driverAdapters` preview features
- Install `@prisma/adapter-pg` and `@types/pg`(_optional, for typechecking in adapter params_)

```ts
import { PrismaClient } from "src/@generated/prisma/client";

const adapter = new PrismaPg(
  {
    host: 'non-existing-host', // <-- not existing, so $connect() should logically fail
    port: 5432,
    user: 'postgres',
    password: 'pass',
    database: 'postgres',
  },
  {
    schema: 'public',
  },
);

  adapter,
});

console.log('Connecting to DB');
await prisma.$connect();
console.log('Successfully to DB'); // <-- You will see this being logged, whereas it should have thrown above
```

Expected: `await prisma.$connect()` to throw an error with P1000/P1001/etc. on DB connection errors.
Actual: `await prisma.$connect()` doesn't throw any errors on DB connection errors. If executing queries in that state, the queryies throw an error with code set to `ECONNREFUSED`/`ENOTFOUND` instead of P1000/P1001/etc.

### Frequency

Consistently reproducible

### Is this a regression?

Yes, relative to pre driver adapters behavior, if that counts.

### Workaround

Moving away from driver adapters.
Or messing with the internal stuff, like doing `await (await adapter.connect()).client.connect()` to trigger a connection.
Or creating a `pg` pool instance to manually check for connection.

### Prisma Schema & Queries

N/A

### Prisma Config

_No response_

### Logs & Debug Info

_No response_

### Environment & Setup

- **OS:** macOS
- **Database:** PostgreSQL
- **Node.js version:** v22.16.0

### Prisma Version

```
prisma                : 6.11.1
@prisma/client        : 6.11.1
Computed binaryTarget : darwin-arm64
Architecture          : arm64
Node.js               : v22.16.0
TypeScript            : 5.8.3
PSL                   : @prisma/prisma-schema-wasm 6.11.1-1.f40f79ec31188888a2e33acda0ecc8fd10a853a9
Schema Engine         : schema-engine-cli f40f79ec31188888a2e33acda0ecc8fd10a853a9 (at ../../node_modules/.pnpm/@prisma+engines@6.11.1/node_modules/@prisma/engines/schema-engine-darwin-arm64)
Default Engines Hash  : f40f79ec31188888a2e33acda0ecc8fd10a853a9
Preview Features      : driverAdapters, queryCompiler
```

barryhagan:
Add the missing mappings from adapter-pg to Prisma errors P1001, P1011, and P1017.

This doesn't completely resolve #27626, but it makes the error codes on queries correct for connection level issues.

### DIFF
--- a/packages/adapter-pg/src/__tests__/errors.test.ts
+++ b/packages/adapter-pg/src/__tests__/errors.test.ts
@@ -78,6 +78,33 @@ describe('convertDriverError', () => {
     expect(convertDriverError(error)).toEqual({ kind: 'TooManyConnections', cause: 'too many connections' })
   })
 
+  it.each([
+    ['UNABLE_TO_VERIFY_LEAF_SIGNATURE', 'unable to verify the first certificate'],
+    [
+      'ERR_TLS_CERT_ALTNAME_INVALID',
+      `Hostname/IP does not match certificate's altnames: Host: localhost. is not in the cert's altnames: DNS:*.localdev.com`,
+    ],
+    [undefined, 'The server does not support SSL connections'],
+  ])('should handle TLS error code %s', (code, message) => {
+    const error = { code, message }
+    expect(convertDriverError(error)).toEqual({
+      kind: 'TlsConnectionError',
+      reason: message,
+    })
+  })
+
+  it.each([
+    ['ENOTFOUND', 'DatabaseNotReachable', 'getaddrinfo ENOTFOUND locohost'],
+    ['ECONNREFUSED', 'DatabaseNotReachable', `connect ECONNREFUSED 127.0.0.1:6699`],
+    ['ECONNRESET', 'ConnectionClosed', 'read ECONNRESET'],
+    ['ETIMEDOUT', 'SocketTimeout', 'connect ETIMEDOUT 127.0.0.1:9100'],
+  ])('should handle socket error code %s', (code, kind, message) => {
+    const error = { code, message, syscall: 'syscallname', errno: -1 }
+    expect(convertDriverError(error)).toEqual({
+      kind,
+    })
+  })
+
   it('should handle default (unknown code)', () => {
     const error = {
       code: '99999',
@@ -99,6 +126,6 @@ describe('convertDriverError', () => {
   })
 
   it('should throw if not a db error', () => {
-    expect(() => convertDriverError({ message: 'The server does not support SSL connections' })).toThrow()
+    expect(() => convertDriverError({ message: 'Unknown driver message' })).toThrow()
   })
 })
--- a/packages/adapter-pg/src/errors.ts
+++ b/packages/adapter-pg/src/errors.ts
@@ -1,7 +1,67 @@
 import { Error as DriverAdapterErrorObject } from '@prisma/driver-adapter-utils'
 import type { DatabaseError } from 'pg'
 
+const TLS_ERRORS = new Set([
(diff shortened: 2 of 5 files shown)


barryhagan:
failing e2e is because this will need an update in prisma-engines to handle the DriverAdapterError mappings.   I guess `DatabaseNotReachable`, `TlsConnectionError`, and `ConnectionClosed` were never a thing for adapters even though they have Prisma error codes.

jacek-prisma:
Hi @barryhagan, this change requires a minor adjustment in prisma-engines, the new variants need to be added to this data type:
https://github.com/prisma/prisma-engines/blob/8ffdf9be03ef5d01959364990a78e7f816d6dc0b/libs/driver-adapters/src/error.rs#L55
Then, they need to be mapped here:
https://github.com/prisma/prisma-engines/blob/8ffdf9be03ef5d01959364990a78e7f816d6dc0b/libs/driver-adapters/src/lib.rs#L28

After the prisma-engines change is ready you can run the tests against the prisma-engines branch by adding `/engine-branch my-branch-name` to the description of this PR.

barryhagan:
You may have a bigger issue at hand with your CI job and e2e tests.  The use of a custom engine branch is not obeyed when running e2e tests in docker containers.  The PRISMA_QUERY_ENGINE_BINARY env var is not supplied and it doesn't look like you add a volume to mount`/target/release`.

So this e2e test will fail until the engine PR is merged. 

I added version output to the failing test to prove that even though the custom branch from my pull was built (bff345156758e55d5c51d6eaf300c3714808f13b), the e2e test is still using the default resolved engine (0b2e493791181e99e767ee7c566626d82da01481)

https://github.com/prisma/prisma/actions/runs/16390247346/job/46315265533#step:5:6296

aqrln:
> You may have a bigger issue at hand with your CI job and e2e tests. The use of a custom engine branch is not obeyed when running e2e tests in docker containers. The PRISMA_QUERY_ENGINE_BINARY env var is not supplied and it doesn't look like you add a volume to mount`/target/release`.

Ouch, yeah, you are right, very good catch. I think you could make this test pass in the PR by switching it to use `queryCompiler` or WASM query engine (assuming it won't cause any other problems in this test). But it's also fine to just merge the engines PR first.

jacek-prisma (approved):

PR #27689, opened 2025-07-18
feat: ORM-1112 `setupExternalTables` script

FGoessler:
Adds the `migrations.setupExternalTables` option to prisma.config.ts and passes it on to the engine.

/integration feat/orm-1112-init-sql-script

### DIFF
--- a/packages/config/src/PrismaConfig.ts
+++ b/packages/config/src/PrismaConfig.ts
@@ -37,15 +37,32 @@ const ErrorCapturingSqlMigrationAwareDriverAdapterFactoryShape = Shape.declare(
   },
 )
 
+const SetupExternalTablesShape = Shape.declare(
+  (input: any): input is () => Promise<string> => {
+    return typeof input === 'function'
+  },
+  {
+    identifier: 'SetupExternalTables',
+    encode: identity,
+    decode: identity,
+  },
+)
+
 export type MigrationsConfigShape = {
   /**
    * The path to the directory where Prisma should store migration files, and look for them.
    */
   path?: string
+  /**
+   * Provide a function to pass a SQL script that will be used to setup external tables during migration diffing.
+   * Also see `tables.external`.
+   */
+  setupExternalTables?: () => Promise<string>
 }
 
 const MigrationsConfigShape = Shape.Struct({
   path: Shape.optional(Shape.String),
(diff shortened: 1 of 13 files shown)


packages/config/src/PrismaConfig.ts
```diff
@@ -37,15 +37,32 @@ const ErrorCapturingSqlMigrationAwareDriverAdapterFactoryShape = Shape.declare(
...
 export type MigrationsConfigShape = {
   /**
    * The path to the directory where Prisma should store migration files, and look for them.
    */
   path?: string
+  /**
+   * Provide a function to pass a SQL script that will be used to setup external tables during migration diffing.
+   * Also see `tables.external`.
+   */
+  setupExternalTables?: () => Promise<string>
```
jacek-prisma: I'd expect this to just be a `string` 🤔 
I think if we want to allow users to to do async computations for config creation maybe we should just allow them to wrap entire config in a Promise rather than using it for individual fields

jkomyno: +1 on what Jacek said. See [plan](https://www.notion.so/prismaio/Planning-prisma-config-ts-GA-2249e8aecef780fbb6fdc14dbd49f7f2?source=copy_link#2249e8aecef78075b42de8f9d9cbd7d5).

```suggestion
  setupExternalTables?: () => string
```

FGoessler: Fair point :+1:

FGoessler: Updated ✅

jkomyno (approved):

packages/migrate/src/__tests__/fixtures/external-tables/prisma.config.ts
```diff
@@ -5,4 +5,14 @@ export default defineConfig({
...
   },
+  migrations: {
+    setupExternalTables: async () => {
+      return `
+        CREATE TABLE "User" (
+          "id" SERIAL PRIMARY KEY,
+          "name" TEXT NOT NULL
+        );
+      `
+    },
```
jacek-prisma: isn't this invalid now?

FGoessler: fixed

jacek-prisma (approved):

PR #27674, opened 2025-07-17
feat(config): add skeleton of "migrations.seed" implementation

jkomyno:
**DRAFT**

This PR:
- closes [ORM-1214](https://linear.app/prisma-company/issue/ORM-1214/prismaconfigts-implement-replacement-for-prismaseed-in-packagejson)
- Add `migrations.seed` optional string attribute in Prisma Config file (e.g., `prisma.config.ts`)
- Stop reading `package.json#prisma` config when a Prisma Config file is detected and loaded

Example:
- From `package.json` (old way)
  ```diff
    // package.json
    {
      "type": "module",
      "devDependencies": {
        "tsx": "latest"
      },
  -   "prisma": {
  -     "seed": "tsx ./prisma/seed.ts"
  -   }
    }
  ```
- To `prisma.config.ts` (new way)
  ```diff
    // prisma.config.ts
    import { defineConfig } from 'prisma/config'
  
    export default defineConfig({
      earlyAccess: true,
  +   migrations: {
  +     seed: 'tsx ./prisma/seed.ts'
  +   }
    })
  ```

PR #27709, opened 2025-07-21
feat: warn on prisma.schema in package.json

jacek-prisma:
[ORM-1217](https://linear.app/prisma-company/issue/ORM-1217/prismaconfigts-warn-when-prisma-key-is-found-in-packagejson)

### DIFF
--- a/packages/cli/src/__tests__/__snapshots__/package-json-schema.test.ts.snap
+++ b/packages/cli/src/__tests__/__snapshots__/package-json-schema.test.ts.snap
@@ -1,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`load schema from package.json and generate a deprecation warning 1`] = `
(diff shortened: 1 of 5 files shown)


jkomyno:
Right now, the deprecation message is only shown after successfully using `getSchemaFromPackageJson`.
I'd rather have the deprecation directly included in `getSchemaFromPackageJson`, like https://github.com/prisma/prisma/pull/27674/ already did

jacek-prisma:
@jkomyno you mean showing the deprecation even if the package.json file lookup fails? Or do you mean including it in the function in case it gets called from somewhere else? (right now I think we only need it in that single place)

jkomyno:
I think we can close this PR now. The only main different between this one and what's merged in main is the use of `warnOnce`. Arguably, we want to put pressure on the users to adopt the Prisma config file, so it's find to raise a warning until they do that. Wdyt?

PR #27724, opened 2025-07-22
feat(config): add migrations.seed support

jkomyno:
This PR:
- closes [ORM-1214](https://linear.app/prisma-company/issue/ORM-1214/prismaconfigts-implement-replacement-for-prismaseed-in-packagejson)
- Add `migrations.seed` optional string attribute in Prisma Config file (e.g., `prisma.config.ts`)
- Stop reading `package.json#prisma` config when a Prisma Config file is detected and loaded

Example:
- From `package.json` (old way)
  ```diff
    // package.json
    {
      "type": "module",
      "devDependencies": {
        "tsx": "latest"
      },
  -   "prisma": {
  -     "seed": "tsx ./prisma/seed.ts"
  -   }
    }
  ```
- To `prisma.config.ts` (new way)
  ```diff
    // prisma.config.ts
    import { defineConfig } from 'prisma/config'
  
    export default defineConfig({
      earlyAccess: true,
  +   migrations: {
  +     seed: 'tsx ./prisma/seed.ts'
  +   }
    })
  ```

---

Note: you may want to review https://github.com/prisma/prisma/pull/27724/commits/c33aee9e847b82b0d635a78f6874ebdbdcfc5640 separately.


helpers/compile/plugins/fill-plugin/fillers/path.ts
```diff
@@ -24,6 +24,24 @@ export function dirname(path: string) {
 }
 
 export const sep = '/'
+export const delimiter = ':'
+
+export function normalize(path: string) {
+  // Simple normalization - remove duplicate slashes and resolve . and ..
+  const parts = path.split('/').filter((part) => part !== '' && part !== '.')
```
aqrln, line 31: I'd do this in one pass instead of two by moving this filter into the loop below (i.e. `if (part !== '' && part !== '.') continue`) but it's fine with me either way, feel free to ignore this comment.

packages/config/src/loadConfigFromPackageJson.ts
```diff
@@ -47,14 +47,28 @@ export async function loadConfigFromPackageJson(cwd: string = process.cwd()): Pr
...
+  // We avoid accidentally emitting deprecation warnings when the `prisma` property in package.json is
+  // ```json
+  // {
+  //   "prisma": {
+  //     "prismaCommit": "placeholder-for-commit-hash-replaced-during-publishing-in-publish-ts"
+  //   }
+  // }
+  // ```
+  // which is the case in `packages/cli/package.json` only.
+  if (Object.keys(deprecatedConfig).length === 1 && deprecatedConfig['prismaCommit']) {
```
aqrln: This will fail if `deprecatedConfig.prismaCommit` is an empty string fwiw. We probably don't need to care about it, but comparing to undefined feels kinda better.

packages/migrate/src/__tests__/fixtures/seed-from-prisma-config/seed-sqlite-js-extra-args/prisma/seed.js
```diff
@@ -0,0 +1,37 @@
+// Using parseArgs would be better, but it's only available in v18.3.0, v16.17.0 and up
```
aqrln, line 1: Why is that a problem? This is in tests and not in user facing code.

packages/migrate/src/__tests__/fixtures/seed-from-prisma-config/seed-sqlite-js-extra-args/prisma/seed.js
```diff
@@ -0,0 +1,37 @@
...
+    throw new Error(`Missing custom arg -z (from CLI call)`)
+  }
+}
+
+main()
+  .then(() => console.log('Goodbye from seed.js'))
+  .catch((e) => {
+    console.error(e)
+    process.exit(1)
+  })
```
aqrln, line 37: Doesn't look like a useful `catch`, the way it's written it's no different than having no `catch` at all except it will also hide the stacktrace

packages/migrate/src/__tests__/fixtures/seed-from-prisma-config/seed-sqlite-js/prisma/seed.js
```diff
@@ -0,0 +1,11 @@
...
+  await new Promise((resolve) => setTimeout(resolve, 0))
+  console.log('Hello from seed.js')
+}
+
+main()
+  .then(() => console.log('Goodbye from seed.js'))
+  .catch((e) => {
+    console.error(e)
+    process.exit(1)
+  })
```
aqrln, line 11: likewise here and everywhere

packages/migrate/src/__tests__/fixtures/seed-from-prisma-config/seed-sqlite-js-extra-args/prisma/seed.js
aqrln: Also, why not make these scripts `.mjs` so you can use top-level await? This `async function main` pattern feels a bit archaic.

packages/config/src/loadConfigFromPackageJson.ts
jkomyno: > We probably don't need to care about it
Yeah, `prismaCommit` is only considered when a proper non-empty value is defined, which is the only actual instance in which you'll find it. I can adjust it, but it doesn't matter w.r.t. our codebase

packages/migrate/src/__tests__/fixtures/seed-from-prisma-config/seed-sqlite-js-extra-args/prisma/seed.js
jkomyno: This is copied verbatim from existing tests. The original comment is from Joel, from this PR from 2 years ago: https://github.com/prisma/prisma/pull/19441

packages/migrate/src/__tests__/fixtures/seed-from-prisma-config/seed-sqlite-js-extra-args/prisma/seed.js
jkomyno: This file will get deleted in Prisma 7. No need to waste time on it imho

packages/migrate/src/__tests__/fixtures/seed-from-prisma-config/seed-sqlite-js-extra-args/prisma/seed.js
jkomyno: Btw, this file will get deleted in Prisma 7. No need to waste time on it imho

aqrln (approved):

PR #27727, opened 2025-07-22
feat: expose underlying drivers where possible

jacek-prisma:
[ORM-1126](https://linear.app/prisma-company/issue/ORM-1126/allow-some-driver-adapters-to-accept-driver-instance-as-input-like-660)

Adds `underlyingDriver` on drivers that can be safely accessed from outside the adapter. Also updates `connect` signatures to return the concrete adapter type (to allow users to discover these methods).


packages/adapter-mariadb/src/mariadb.ts
```diff
@@ -149,6 +149,10 @@ export class PrismaMariaDbAdapter extends MariaDbQueryable<mariadb.Pool> impleme
   async dispose(): Promise<void> {
     await this.client.end()
   }
+
+  underlyingDriver(): mariadb.Pool {
```
aqrln, line 153: I'm wondering if we really want to use a generic name consistent between adapters or if it would actually be better to specialize them based on the type name that the method returns (i.e. `getPool`, `getClient` etc).

jacek-prisma: I think using a consistent name makes it more obvious and easier to discover for users, `getPool` could be interpreted to return some Prisma type, unless we make it super explicit i.e `getPgPool`, `getNeonClient` etc

aqrln (approved):

PR #27730, opened 2025-07-22
feat: share model delegate method generation between generators

aqrln:
Port the changes made in [#27714][] to `prisma-client-js` generator by moving the modified logic and everything it depends on to a shared package and reusing it in both generators.

[#27714]: https://github.com/prisma/prisma/pull/27714


packages/client-generator-common/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@prisma/client-generator-common",
+  "version": "0.0.0",
+  "description": "This package is intended for Prisma's internal use",
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts",
```
aqrln, line 7: remove these

PR #27747, opened 2025-07-25
fix: handle included enum arrays in data mapper

### LINKED ISSUE #27511: e.map is not a function error with prisma-client generator + queryCompiler
### Bug description

After upgrading from 5.22 to 6.10.1 with the new prisma-client generator, I'm getting "e.map is not a function" errors on some queries involving the existing many-to-many relations and the include property. P.S The expected migration for many-to-many relations (e.g drop index + create PK) has been applied.

### Severity

### Reproduction

Run the provided query on the provided schema.

Note that only when including the relation I get the error.

Expected no error.

Actual "e.map is not a function"

### Frequency

Consistently reproducible

### Is this a regression?

Yes, it used to worked in 5.22. Not working anymore in 6.10 with the new generator.

### Workaround

use with select instead of returning all relation data

```ts
    include: {
      users: {
        select: {
          id: true,
        },
      },
    },
```

### Prisma Schema & Queries

```prisma
model User {
  id                       String                    @id @unique @default(uuid())
  packages                 Package[]
}

model Package {
  id            String      @id @unique @default(uuid())
  type          PackageType
  numOfRequests Int
  users         User[]
}
```
```ts
prisma.package.findFirstOrThrow({
    where: {
      type: "API",
      users: {
        some: {
          id: userId,
        },
      },
    },
    include: {
      users: true,
    },
  });
```

### Prisma Config

```ts
generator client {
  previewFeatures     = ["queryCompiler", "driverAdapters"]
  output              = "../src/prisma"
  moduleFormat        = "cjs"
  importFileExtension = ""
}

datasource db {
  provider = "postgresql"
}


import { Prisma, PrismaClient } from "../prisma/client";

const adapter = new PrismaPg({ connectionString: process.env.DATABASE_URL });
const dbClient = new PrismaClient({
  adapter,
});
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** MacOs or Docker node:22-alpine
- **Database:**PostgreSQL
- **Node.js version:** 22

### Prisma Version

```
// Prisma version output
prisma                : 6.10.1
@prisma/client        : 6.10.1
Computed binaryTarget : darwin
Architecture          : x64
Node.js               : v22.16.0
TypeScript            : 5.8.3
PSL                   : @prisma/prisma-schema-wasm 6.10.1-1.9b628578b3b7cae625e8c927178f15a170e74a9c
Schema Engine         : schema-engine-cli 9b628578b3b7cae625e8c927178f15a170e74a9c (at node_modules/@prisma/engines/schema-engine-darwin)
Default Engines Hash  : 9b628578b3b7cae625e8c927178f15a170e74a9c
Preview Features      : driverAdapters, queryCompiler
```

baileywickham:
Possibly closes [this](https://github.com/prisma/prisma/issues/27511) issue. 

We noticed this issue with a query that included an array of enums in postgres. This query would fail and throw an exception, which has blocked us from adopting the query compiler. This PR fixes the issue in our project.

In the cases I tested for postgres, the format of the enum array was always a string that looked like `"{a,b,c}"`, I'm not sure if that is true for all providers, I'm also not sure if this should be handled at the driver/adapter level, I'd appreciate some advice here.

Happy to add tests as well, just wanted to run this PR by the team and make sure it looks correct before I go too far down that!

Thanks!

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/DataMapper.ts
+++ b/packages/client-engine-runtime/src/interpreter/DataMapper.ts
@@ -219,6 +219,15 @@ function mapValue(
     }
 
     case 'Array': {
+      // The pg adapter returns enum arrays as a string like "{a,b,c}" when the enum array is the result of a join.
+      // This is because pg can't parse the oid at the driver level without additional queries.
+      // We have the info from the schema to do it here without queriing the pg_type table.
+      if (resultType.inner.type === 'Enum') {
+        if (typeof value === 'string' && value.startsWith('{') && value.endsWith('}')) {
+          return mapEnumArray(value, resultType.inner.inner, enums)
+        }
+      }
+
       const values = value as unknown[]
       return values.map((v, i) => mapValue(v, `${columnName}[${i}]`, resultType.inner, enums))
     }
@@ -245,20 +254,24 @@ function mapValue(
     }
 
     case 'Enum': {
-      const enumDef = enums[resultType.inner]
-      if (enumDef === undefined) {
-        throw new DataMapperError(`Unknown enum '${resultType.inner}'`)
-      }
-      const enumValue = enumDef[`${value}`]
-      if (enumValue === undefined) {
-        throw new DataMapperError(`Unknown enum value '${value}' for enum '${resultType.inner}'`)
-      }
-      return enumValue
+      return mapEnum(value, resultType.inner, enums)
     }
 
     default:
       assertNever(resultType, `DataMapper: Unknown result type: ${(resultType as PrismaValueType).type}`)
   }
+}
+
+function mapEnum(value: unknown, enumName: string, enums: Record<string, Record<string, string>>): string {
+  const enumDef = enums[enumName]
+  if (enumDef === undefined) {
+    throw new DataMapperError(`Unknown enum '${enumName}'`)
+  }
+  const enumValue = enumDef[`${value}`]
+  if (enumValue === undefined) {
+    throw new DataMapperError(`Unknown enum value '${value}' for enum '${enumName}'`)
+  }
+  return enumValue
 }
 
 // The negative lookahead is to avoid a false positive on a date string like "2023-10-01".
@@ -281,4 +294,20 @@ function ensureTimezoneInIsoString(dt: string): string {
   } else {
     return dt
   }
+}
+
+// The pg adapter returns enum arrays as a string like "{a,b,c}" when the enum array is the result of a join.
+// In our case we are only parsing enums out of an array
+// See the lib pg uses for parsing:
+// https://github.com/bendrucker/postgres-array/blob/master/index.js
+function mapEnumArray(value: string, columnName: string, enums: Record<string, Record<string, string>>): string[] {
+  let cleanValue = value
+  if (cleanValue.startsWith('{')) {
+    cleanValue = cleanValue.slice(1)
+  }
+  if (cleanValue.endsWith('}')) {
+    cleanValue = cleanValue.slice(0, -1)
+  }
+  const values = cleanValue.split(',')
+  return values.map((v) => mapEnum(v, columnName, enums))
 }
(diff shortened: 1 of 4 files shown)


baileywickham:
@aqrln Hi, tagging you on this PR because you show up on the blame for most of the file. Let me know if this is the correct direction and I can add tests or whatever else you need! If you would point me to the best location for the tests (I assume functional tests), that would be great too.

aqrln:
Hmm, I would've expected parsing the array to be handled on the driver adapter level (if not in `pg` itself) and for the data mapper to only handle the conversion between DB names and Prisma names of the enum values 🤔 I guess we could do it here as the last resort if that's not feasible for whatever reason but let's see if we can avoid it.

Some existing tests are currently failing with this change, could you please take a look?

baileywickham:
Hey, @aqrln, I looked at doing it inside the adapter or `pg` and I found [this](https://github.com/prisma/prisma/blob/main/packages/adapter-pg/src/conversion.ts#L259-L262) comment for `pg` specifically. That points to doing it in DataMapper, but given the failure only occurs when `include`ing a table, I'm not so sure. 

I'm working on getting a minimal reproduction up, from what I can tell right now it only fails when you `include` a table with an enum array with the query compiler enabled. I've only tested it with the postgres driver.  I'm not sure why it doesn't fail when querying the table with the enum array directly, I'll try to figure that out

aqrln:
> from what I can tell right now it only fails when you include a table with an enum array with the query compiler enabled

Oh, that's an interesting detail, are you using `relationJoins` preview feature as well?

aqrln:
I'm asking about `relationJoins` because it makes total sense that neither `pg` would be able to decode it inside a nested query result in that case, nor the driver adapter possibly could, because it would be inside a JSON string built on the database side. In that case we can only do it in the data mapper indeed.

baileywickham:
@aqrln That was my first thought (and [it's true pg can't parse nested queries](https://github.com/brianc/node-postgres/issues/1801)).  Here is the client block I'm using (note no relation join)
```
generator client {
  provider = "prisma-client"
  output   = "../generated/prisma"
  previewFeatures = ["queryCompiler", "driverAdapters"]
}
```

But it does look like the query is an inner join under the hood. 
```ts
const _ = await client.workspace_member.findMany({  
      include: {
        roles: true,
      }
})

2025-07-28 14:55:54.806 UTC [22459] LOG:  duration: 2.150 ms  parse <unnamed>: SELECT "t1"."id", "t1"."name", "t1"."permissions", "t0"."A" AS "workspace_memberToworkspace_role@workspace_member" FROM "public"."_workspace_memberToworkspace_role" AS "t0" INNER JOIN "public"."workspace_role" AS "t1" ON "t0"."B" = "t1"."id" WHERE "t0"."A" IN ($1)
```

baileywickham:
https://github.com/baileywickham/prisma-issue-repro here is a repro for the issue

baileywickham:
Yeah so it looks like the issue is how `pg` parses the result of the join here. I think for the enum array case I initially patched we can definitely implement support in DataMapper, but do you know if there are other cases where a join will result in a query not being parsed correctly? Like when using `relationJoins`?

I think if there are too many other cases where `pg` can't parse the result it might make sense to pull this logic out of DataMapper since it is driver specific.

aqrln:
> That was my first thought (and [it's true pg can't parse nested queries](https://github.com/brianc/node-postgres/issues/1801)).

Ah, that's very good to know but that's not exactly what I meant. When using `relationJoins`, Prisma uses lateral joins and JSON aggregations to build the lists of related records on the database side, so instead of multiple rows per a single parent record with data duplication there's only one row per parent with all children serialized as a JSON list in a single column.

Outside of data mapper, where we have the information about the expected output shape of the query and the Prisma types, it would not be possible to do any kind of transformations inside those JSON columns because there would be no way to distinguish them from normal JSON columns that shouldn't be touched (we wouldn't even have the information about the original database types on the driver adapter level).

jacek-prisma:
This is already fixed on the main branch

PR #27755, opened 2025-07-25
fix(migrate): don't throw when an adapter is used instead of a url

GauBen:
Hi!

Small PR to address a crash when an adapter is used to connect to the db. It turns out the comment `// this is safe, as if the protocol was unknown, we would have already exited the program with an error` is no longer true now that adapters exist.


jkomyno:
Hi @GauBen, thank you for this PR!

jkomyno (approved):

PR #27777, opened 2025-07-30
fix: ORM-850 improve client options type performance

FGoessler:
This PR improves the complex type resolution of the `PrismaClient` type in case `ClientOptions` are provided. e.g.:

```typescript
const client = new PrismaClient( /* Client Options */ { datasourceUrl: "..." })
```

## Why is providing client options causing type performance issues?

The `PrismaClient` interface changes dynamically based on the provided options. There are two options relevant for this behavior: `log` and `omit` configuration.

`log` configuration changes the signature of the `$on` method of the client. This provides type safety so one can only observe log messages/events that are actually activated through the client options.

```typescript
  log: [
    { level: 'query', emit: 'event' },
    { level: 'error', emit: 'stdout' },
  ],
})
// ok:
client.$on('query', (event) => { ... })
// type error:
client.$on('info', (event) => { ... })
```

`omit` configuration changes what fields on models are available by default. This has wide reaching type implications across the `PrismaClient` instance and also deep into potential client extensions.

```typescript
  omit: {
    user: {
      name: true,
    },
  },
})
```

Due to those customizations a `PrismaClient` is not equal to any other `PrismaClient`. To check whether one instance is assignable to another (e.g. when passing a `PrismaClient` into a helper method) TypeScript has to perform some very deep checks that run through deep and complex generics. If no client options are provided TypeScript can usually "short circuit" and is very efficient - once client options are provided way more work has to be done.

## Existing workarounds

We have a known workaround to use the `typeof` operator to avoid redundant type resolution. Through this the user can give a specific `PrismaClient` instance an "named alias" and use this in their code. TypeScript then does not have to repeatedly check full type assignability but can rely on the equality of the named alias.

```typescript
const client = new PrismaClient({ ... })

type MyPrismaClient = typeof client

const passClientAround = (prisma: MyPrismaClient) => {
  ...
}
```

## Approach of this PR

Currently any variant of client options triggers the full complexity type check and prevents TypeScript from taking the "short circuit" route. Even if the provided client options have no influence on the actual type interface when they specify no `omit` or `log` configuration.

The idea is now to modify the type of the PrismaClient such that the type performance penalty is only paid if the complex checks for `omit` and `log` are actually necessary. To achieve this `PrismaClient` itself was changed to not depend on a generic `ClientOptions` but on separate `LogOpts` and `OmitOpts`. These are extracted from the `ClientOptions` in the constructor. 

```typescript
// Before
export interface PrismaClientConstructor {
  new <
    ClientOptions extends Prisma.PrismaClientOptions = Prisma.PrismaClientOptions,
    const U = LogOptions<ClientOptions>,
    ExtArgs extends runtime.Types.Extensions.InternalArgs = runtime.Types.Extensions.DefaultArgs
  >(options?: Prisma.Subset<ClientOptions, Prisma.PrismaClientOptions>): PrismaClient<ClientOptions, U, ExtArgs>
}

export interface PrismaClient<
  ClientOptions extends Prisma.PrismaClientOptions = Prisma.PrismaClientOptions,
  U = LogOptions<ClientOptions>,
  ExtArgs extends runtime.Types.Extensions.InternalArgs = runtime.Types.Extensions.DefaultArgs
> {
  ...
}

// After
export interface PrismaClientConstructor {
  new <
    Options extends Prisma.PrismaClientOptions,
    LogOpts extends LogOptions<Options>,
    OmitOpts extends Prisma.PrismaClientOptions['omit'] = Options extends { omit: infer U } ? U : Prisma.PrismaClientOptions['omit'],
    ExtArgs extends runtime.Types.Extensions.InternalArgs = runtime.Types.Extensions.DefaultArgs
  >(options?: Options): PrismaClient<LogOpts, OmitOpts, ExtArgs>
}

export interface PrismaClient<
  in LogOpts extends Prisma.LogLevel = never,
  in out OmitOpts extends Prisma.PrismaClientOptions['omit'] = Prisma.PrismaClientOptions['omit'],
  in out ExtArgs extends runtime.Types.Extensions.InternalArgs = runtime.Types.Extensions.DefaultArgs
> {
  ...
} 
```

Now two type instances of `PrismaClient` are only different if their `LogOpts` or `OmitOpts` are actually different. For most cases the type check is now way simpler and can go the "short circuit" routes. As the dependencies are now smaller for `LogOpts`, just providing `log` configuration causes way less type check overhead than before. Only with `omit` configuration provided one still pays a high complexity price (although that also improved significantly!).

This can be an acceptable tradeoff as `omit` client options is a more rarely used feature and we can extend the documentation to encourage users to use the `typeof` workaround when using `omit` client options.

You can see the magnitude of improvements in the updated type benchmark instantiation counts. In some cases for large schemas we went from about 13 million to below 1k instantiations!

## Change in generic type signature

> [!WARNING]
> This change comes with one caveat: As the `PrismaClient` generic arguments type signature had to change its a technically breaking change.

Which should be okay as this generator is still in preview and I would not expect many users to explicitly specify generic type params on their `PrismaClient` type. Even if so it is easy to work around and resolve. Ideally by not specifying the generic type arguments but letting them get inferred, using the `typeof` operator, or in worst case updating the generic arguments as shown below.

Log Level Case:
```typescript
// Before
let prisma: PrismaClient<{ log: [{ emit: 'event'; level: 'query' }] }>
// or:
let prisma: PrismaClient<{ log: ['query'] }>
// After
let prisma: PrismaClient<'query'>
```

Omit Case:
```typescript
// Before
let prisma: PrismaClient<never, omit: { user: { password: true } }>
// After
let prisma: PrismaClient<never, { user: { password: true } }>
```

Might be bigger issue for some library developers though? 🤔 

It is also an issue in our test infrastructure as we have a few client tests that need explicitly specified `LogOpts`. The same tests run on the new and the old generator though. So with this change we had to introduce complexity to change the test code based on what generator it is supposed to run with. For the `globalOmit` test in particular I duplicated the test to have one test run on the new generator only and the other on the old generator only.


packages/cli/src/__tests__/incomplete-schemas.test.ts
```diff
@@ -39,7 +39,7 @@ describe('[wasm] incomplete-schemas', () => {
 
     it('format', async () => {
       const result = await Format.new().parse([], defaultTestConfig())
-      expect(result).toMatch(/^Formatted (.*) in \d+m?s 🚀$/)
+      expect(result).toMatch(/^Formatted (.*) in \d+(\.\d+)?m?s 🚀$/)
```
FGoessler, line 42: ℹ️ Accounting for `1.20s` like formatting that sometimes happens if the test is low.

packages/client-generator-ts/src/TSClient/PrismaClient.ts
```diff
@@ -292,17 +292,18 @@ export type LogOptions<ClientOptions extends Prisma.PrismaClientOptions> =
...
 }
 
 ${this.jsDoc}
 export interface PrismaClient<
-  ClientOptions extends Prisma.PrismaClientOptions = Prisma.PrismaClientOptions,
-  U = LogOptions<ClientOptions>,
-  ExtArgs extends runtime.Types.Extensions.InternalArgs = runtime.Types.Extensions.DefaultArgs
+  in LogOpts extends Prisma.LogLevel = never,
+  in out OmitOpts extends Prisma.PrismaClientOptions['omit'] = Prisma.PrismaClientOptions['omit'],
+  in out ExtArgs extends runtime.Types.Extensions.InternalArgs = runtime.Types.Extensions.DefaultArgs
```
FGoessler, line 306: ℹ️ The `in out` annotations here significantly help to further reduce the required type instantiations.

packages/client-generator-ts/src/TSClient/PrismaClient.ts
```diff
@@ -292,17 +292,18 @@ export type LogOptions<ClientOptions extends Prisma.PrismaClientOptions> =
...
   ${indent(this.jsDoc, TAB_SIZE)}
   new <
-    ClientOptions extends Prisma.PrismaClientOptions = Prisma.PrismaClientOptions,
-    const U = LogOptions<ClientOptions>,
+    Options extends Prisma.PrismaClientOptions,
+    LogOpts extends LogOptions<Options>,
+    OmitOpts extends Prisma.PrismaClientOptions['omit'] = Options extends { omit: infer U } ? U : Prisma.PrismaClientOptions['omit'],
     ExtArgs extends runtime.Types.Extensions.InternalArgs = runtime.Types.Extensions.DefaultArgs
-  >(options?: Prisma.Subset<ClientOptions, Prisma.PrismaClientOptions>): PrismaClient<ClientOptions, U, ExtArgs>
+  >(options?: Options): PrismaClient<LogOpts, OmitOpts, ExtArgs>
```
aqrln: `Prisma.Subset` was checking that no extra unknown properties were provided. Is removing it intentional? I don’t feel strongly about it, but want to make sure it was a conscious decision that was made for a reason.

FGoessler: Oh - good point. I'll have a look if adding it back in is problematic or not :+1:

packages/client/tests/functional/batching-compound/tests.ts
```diff
@@ -15,11 +15,11 @@ const user2 = {
   lastName: copycat.lastName(2),
 }
 
-declare let newPrismaClient: NewPrismaClient<typeof PrismaClient>
+declare const newPrismaClient: NewPrismaClient<PrismaClient, typeof PrismaClient>
 
 testMatrix.setupTestSuite(
   () => {
-    let prisma: PrismaClient<{ log: [{ emit: 'event'; level: 'query' }] }>
+    let prisma: PrismaClient
```
aqrln, line 22: Is there any reason why we can't use these `@only-js-generator`/`@only-ts-generator` annotations here and in most tests?

packages/client/tests/functional/globalOmit/_matrix.ts
```diff
@@ -3,11 +3,11 @@ import { Providers } from '../_utils/providers'
...
-    { provider: Providers.SQLITE },
-    { provider: Providers.POSTGRESQL },
-    { provider: Providers.MYSQL },
-    { provider: Providers.MONGODB },
-    { provider: Providers.COCKROACHDB },
-    { provider: Providers.SQLSERVER },
+    { provider: Providers.SQLITE, generatorType: 'prisma-client-ts' },
+    { provider: Providers.POSTGRESQL, generatorType: 'prisma-client-ts' },
+    { provider: Providers.MYSQL, generatorType: 'prisma-client-ts' },
+    // { provider: Providers.MONGODB, generatorType: 'prisma-client-ts' }, // TODO: we have type issues to fix with the new client generator
```
aqrln, line 9: is there a ticket for this already?

packages/type-benchmark-tests/basic/client-options.bench.ts
```diff
@@ -103,7 +130,7 @@ bench('passed around client then extend', () => {
 
   return passClientAround(client)
   // Apparently passing the client around and then extending it is way slower.
-}).types([48156, 'instantiations']) // TODO: we want to get this number down
+}).types([2151, 'instantiations'])
```
aqrln, line 133: 🎉

packages/client/tests/functional/globalOmit/_matrix.ts
FGoessler: yes -> https://linear.app/prisma-company/issue/ORM-1326/fix-type-issues-for-mongodb

packages/client/tests/functional/batching-compound/tests.ts
FGoessler: Yes we could but the other tests not really test the correctness of the log types but just use it as a means to get to the query logs. So I figured it's easier to just ignore the type there and update them consistently once we drop the old generator and have a uniform setup.

aqrln (approved):

PR #27779, opened 2025-07-30
test(cli): match fractional time in `incomplete-schemas.test.ts`

aqrln:
https://github.com/prisma/prisma/pull/27762 updated the regular expressions to support seconds in additional to milliseconds. This didn't fix the test flakiness because in the case of seconds, the numbers are not necessarily integers but can be something like `1.1s`, which the regex still didn't accept.

### DIFF
--- a/packages/cli/src/__tests__/incomplete-schemas.test.ts
+++ b/packages/cli/src/__tests__/incomplete-schemas.test.ts
@@ -39,7 +39,7 @@ describe('[wasm] incomplete-schemas', () => {
 
     it('format', async () => {
       const result = await Format.new().parse([], defaultTestConfig())
-      expect(result).toMatch(/^Formatted (.*) in \d+m?s 🚀$/)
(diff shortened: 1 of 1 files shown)


tensordreams:
Closing as obsolete. Rebasing onto `main` leaves nothing: all three assertions this PR patched have since been rewritten and no longer check the `Formatted ... in Xms 🚀` output at all — two now expect a config-load error, the third is a bare `resolves.not.toThrow()`. The string `Formatted` no longer appears in `incomplete-schemas.test.ts`.

The underlying flakiness is handled, and more broadly than this PR did. The snapshot serializers normalise fractional seconds globally:

```js
// packages/get-platform/src/test-utils/vitest-snapshot-serializer.ts (and the Jest twin)
str.replace(/ \d+ms/g, " XXXms").replace(/ \d+(\.\d+)?s/g, " XXXms")
```

That `(\.\d+)?` is the same fractional-seconds case this PR added, applied to every timing snapshot rather than three call sites.

PR #27781, opened 2025-07-30
feat(qc): chunking

jacek-prisma:
[ORM-1156](https://linear.app/prisma-company/issue/ORM-1156/implement-chunking-with-qc)

Second half: https://github.com/prisma/prisma-engines/pull/5570


packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
```diff
@@ -324,6 +342,36 @@ export class QueryInterpreter {
...
+      case 'mysql':
+        return 65535
+      case 'sqlite':
+        return 999
+      case 'sqlserver':
+        return 2098
+      case 'mongodb':
+        return undefined
+      default:
+        assertNever(this.#provider, `Unexpected provider: ${this.#provider}`)
```
FGoessler, line 371: Shouldn't this be ideally something coming from the driver adapter? 🤔

packages/client-engine-runtime/src/interpreter/renderQuery.test.ts
```diff
@@ -3,6 +3,8 @@ import { GeneratorRegistry } from './generators'
 import { renderQuery } from './renderQuery'
 import { ScopeBindings } from './scope'
 
+const TEXT_MAX_CHUNK_SIZE = 10
```
FGoessler: typo? 🤔 

```suggestion
const TEST_MAX_CHUNK_SIZE = 10
```

FGoessler (approved):

packages/client-engine-runtime/src/interpreter/QueryInterpreter.ts
jacek-prisma: driver adapters don't always expose `getConnectionInfo` and the existing behavior in the QE is to default it per flavour like I do here

we could change the adapters to always return `maxBindParams` though, but I thought that's a more controversial change than doing this, I don't mind changing the driver adapters to make `ConnectionInfo` non-optional though

jacek-prisma: @FGoessler let me know if you want me change this

FGoessler: I'd prefer this to come from the driver adapter only. But we can do this in a dedicated change and leave it like this for this PR.

PR #27799, opened 2025-08-01
feat(client): drop deprecated middleware feature

jkomyno:
This PR:
- closes [ORM-1308](https://linear.app/prisma-company/issue/ORM-1308/drop-prismadollaruse)
- gets rid of `prisma.$use` (Prisma middlewares), which was deprecated in Prisma 4.16.0


FGoessler (commented):
There are two minor occurrences in tests that can also get delete:
- https://github.com/prisma/prisma/blob/feat/drop-middlewares/packages/client/tests/functional/extensions/extends.ts#L14
- https://github.com/prisma/prisma/blob/feat/drop-middlewares/packages/client/tests/functional/extensions/itx.ts#L264

What about `MiddlewareParams` and `Middleware`? I still see this type in some generator code. E.g. here: https://github.com/prisma/prisma/blob/feat/drop-middlewares/packages/client-generator-js/src/TSClient/PrismaClient.ts#L596
Maybe we should also drop those from the built-in names list [here](https://github.com/prisma/prisma/blob/feat/drop-middlewares/packages/client/tests/functional/naming-conflict/_builtInNames.ts#L54-L55)?

FGoessler (approved):

PR #27861, opened 2025-08-08
fix(qc): align tx manager errors

jacek-prisma:
[ORM-1167](https://linear.app/prisma-company/issue/ORM-1167/align-transaction-manager-error-behavior-with-the-qe)


packages/client-engine-runtime/src/UserFacingError.ts
```diff
@@ -165,7 +165,7 @@ function renderErrorMessage(err: DriverAdapterError): string | undefined {
       return `The column \`${column}\` does not exist in the current database.`
     }
     case 'InvalidIsolationLevel':
-      return `Invalid isolation level \`${err.cause.level}\``
+      return `Error in connector: Conversion error: ${err.cause.level}`
```
jacek-prisma, line 168: this error is probably worse than what it used to be, but I'm keeping it like this for consistency between QE and QC for now

aqrln (approved):

PR #27899, opened 2025-08-13
feat(qc): implement query arg conversions

jacek-prisma:
[ORM-1327](https://linear.app/prisma-company/issue/ORM-1327/implement-conversions-for-query-arguments)


packages/adapter-mssql/src/conversion.ts
```diff
@@ -83,41 +83,128 @@ export function mapIsolationLevel(level: IsolationLevel): sql.IIsolationLevel {
   }
 }
 
-export function mapArg(arg: unknown): unknown {
-  if (arg instanceof Uint8Array) {
-    return Buffer.from(arg)
+export function mapArg<A>(arg: A | BigInt | Date, argType: ArgType): null | number | BigInt | string | Uint8Array | A {
+  if (arg === null) {
```
aqrln, line 87: should the type of `arg` also include `| null`?

jacek-prisma: the point of the union type on the `arg` in all of the `mapArg` functions is just to make it possible to assign another type to it that is not A because it's sometimes convenient, it's not meant to represent the allowed input types

jacek-prisma: I can update them to have all the types we explicitly handle but I am not sure if that's useful because it doesn't really help with type safety much

aqrln (approved):

PR #27905, opened 2025-08-14
fix(client): use `customDataProxyFetch` correctly

aqrln:
1. Revert a previous incorrect change: `customDataProxyFetch` is a decorator and not the `fetch` function itself.
2. Make sure we only use the subset of the fetch API supported by the Accelerate extension.

Closes: https://linear.app/prisma-company/issue/ORM-1353/make-sure-customdataproxyfetch-is-used
Closes: https://linear.app/prisma-company/issue/ORM-1354/fix-accelerate-caching-tests

### DIFF
--- a/packages/client/src/runtime/core/engines/common/Engine.ts
+++ b/packages/client/src/runtime/core/engines/common/Engine.ts
@@ -38,35 +38,30 @@ export type GraphQLQuery = {
 }
 
 /**
- * Custom fetch function for `DataProxyEngine`.
+ * A stripped down interface of `fetch` that `@prisma/extension-accelerate`
+ * relies on. It must be in sync with the corresponding definition in the
+ * Accelerate extension.
  *
- * We can't use the actual type of `globalThis.fetch` because this will result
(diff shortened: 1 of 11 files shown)


packages/client/src/runtime/core/engines/common/Engine.ts
```diff
@@ -38,43 +38,33 @@ export type GraphQLQuery = {
...
- *
- * Since `@prisma/extension-accelerate` redefines the type of
- * `__internalParams.customDataProxyFetch` to its own type anyway (probably for
- * exactly this reason), our definition is never actually used and is completely
- * ignored, so it doesn't matter, and we can just use `unknown` as the type of
- * `fetch` here.
+ * This is the actual interface exposed by the extension. We can't use the
+ * custom fetch function provided by it as normal fetch because the API is
+ * different. Notably, `headers` must be an object and not a `Headers`
+ * instance, and `url` must be a `string` and not a `URL`.
```
ellotheth, line 48: 😱 😭 🫠

jacek-prisma (approved):

PR #27906, opened 2025-08-15
fix(qc): avoid exposing a transaction closing error

jacek-prisma:
[ORM-1355](https://linear.app/prisma-company/issue/ORM-1355/do-not-expose-transactionclosingerror-to-users)

Instead of exposing the closing error directly, we await the close and propagate a nicer error.

### DIFF
--- a/packages/client-engine-runtime/src/transaction-manager/transaction-manager.test.ts
+++ b/packages/client-engine-runtime/src/transaction-manager/transaction-manager.test.ts
@@ -1,3 +1,5 @@
+import timers from 'node:timers/promises'
+
 import type { SqlDriverAdapter, SqlQuery, SqlResultSet, Transaction } from '@prisma/driver-adapter-utils'
 import { ok } from '@prisma/driver-adapter-utils'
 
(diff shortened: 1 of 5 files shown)


packages/client-engine-runtime/src/transaction-manager/transaction-manager.test.ts
```diff
@@ -141,9 +141,47 @@ test('getTransaction works while being rolled back', async () => {
...
+  const driverAdapter = new MockDriverAdapter()
+  const timeout = 200
+  const rollbackDelay = 200
+
+  driverAdapter.rollbackMock = jest.fn().mockImplementation(() => {
+    return new Promise((resolve) => {
+      setTimeout(() => {
+        resolve(undefined)
+      }, rollbackDelay)
+    })
```
aqrln: nit: you can
```ts
import timers from 'node:timers/promises`
```
and then replace this with
```suggestion
    return timers.setTimeout(rollbackDelay)
```

aqrln (approved):

packages/client-engine-runtime/src/transaction-manager/transaction-manager.test.ts
jacek-prisma: updated

PR #27907, opened 2025-08-15
feat(qc): always attach error listeners on drivers

jacek-prisma:
[ORM-1234](https://linear.app/prisma-company/issue/ORM-1234/add-missing-event-emitter-error-handlers)

### DIFF
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -130,7 +130,7 @@ export class PrismaMariaDbAdapter extends MariaDbQueryable<mariadb.Pool> impleme
 
     const conn = await this.client.getConnection().catch((error) => onError(error))
     conn.on('error', (err: mariadb.SqlError) => {
-      debug(`Error from connection: %O`, err)
+      debug(`Error from connection: ${err.message} %O`, err)
       this.options?.onConnectionError?.(err)
     })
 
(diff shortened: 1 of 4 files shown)


packages/adapter-mariadb/src/mariadb.ts
```diff
@@ -128,6 +129,10 @@ export class PrismaMariaDbAdapter extends MariaDbQueryable<mariadb.Pool> impleme
     debug('%s options: %O', tag, options)
 
     const conn = await this.client.getConnection().catch((error) => onError(error))
+    conn.on('error', (err: mariadb.SqlError) => {
+      debug(`Error from connection: ${err.message} %O`, err)
```
aqrln, line 133: Doesn't pretty-printing an error object with `%O` print its error message anyway? If not, it's probably better to use `%s` instead of interpolating `err.message` so we don't need to construct a new string when debug logs are disabled.

jacek-prisma: I just copied the existing debug line from `pg` here, I suppose it might be best to update both to just use %O

jacek-prisma: I've checked and the %O format doesn't always have the error message

jacek-prisma: I think we should keep it like this

aqrln (approved):

PR #27923, opened 2025-08-18
perf(size): remove all raw wasm binaries from `@prisma/client`

aqrln:
This change removes all raw wasm binaries from `@prisma/client` package
and only leaves the base64 encoded bundles which are used by the
generated client at run time (or at application-side bundling time) when
using the new `prisma-client` generator and generating code for
Node.js-like environments.

When generating a `prisma-client` client for edge runtimes or always
with `prisma-client-js`, we don't use the base64 bundles and need to
copy the raw wasm files to the generated client directory.

`prisma-client` generator copies the raw wasm files from the CLI
package, since it can only be used via the CLI and doesn't require the
`@prisma/client` package to be installed at generation time.

Conversely, `prisma-client-js` generator relies on the `@prisma/client`
package to be available at generation time, and in fact can be used
without the CLI at all, since it is also shipped as part of the 
`@prisma/client` package for legacy reasons. Therefore, it doesn't have
anywhere to copy the raw wasm files from but instead has access to the
base64 bundles, so it loads and decodes them at generation time and
writes the decoded wasm files to the generated client directory.

Closes: https://linear.app/prisma-company/issue/ORM-1292/querycompiler-investigate-too-many-wasm-binaries-in-node-modules

### DIFF
--- a/packages/cli/helpers/build.ts
+++ b/packages/cli/helpers/build.ts
@@ -63,14 +63,26 @@ const cliLifecyclePlugin: esbuild.Plugin = {
 }
 
 async function copyClientWasmRuntime() {
-  const clientRuntimePath = path.join(__dirname, '..', '..', 'client', 'runtime')
+  const clientPath = path.join(__dirname, '..', '..', 'client')
+  const clientRuntimePath = path.join(clientPath, 'runtime')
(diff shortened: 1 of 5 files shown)


jacek-prisma:
@aqrln lots of tests fail complaining about missing wasm

jacek-prisma (approved):

jkomyno (approved):

PR #27927, opened 2025-08-19
fix(qc): fix mariadb data encoding issues

jacek-prisma:
[ORM-1359](https://linear.app/prisma-company/issue/ORM-1359/run-qe-tests-against-mariadb-with-relationjoins)

See https://github.com/prisma/prisma-engines/pull/5600 for details.

Extra addition:
- changed date output date encoding for consistency with other drivers, covered with a test added in the engines PR

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/data-mapper.ts
+++ b/packages/client-engine-runtime/src/interpreter/data-mapper.ts
@@ -247,7 +247,7 @@ function mapValue(
               `Expected a base64-encoded byte array in column '${columnName}', got ${typeof value}: ${value}`,
             )
           }
-          return { $type: 'Bytes', value: value.replace('\n', '') }
+          return { $type: 'Bytes', value }
 
         case 'hex':
           if (typeof value !== 'string' || !value.startsWith('\\x')) {
--- a/packages/client-engine-runtime/src/interpreter/data-mapper.ts
+++ b/packages/client-engine-runtime/src/interpreter/data-mapper.ts
@@ -240,19 +240,36 @@ function mapValue(
     }
 
     case 'bytes': {
-      if (typeof value === 'string' && value.startsWith('\\x')) {
-        // Postgres bytea hex format. We have to handle it here and not only in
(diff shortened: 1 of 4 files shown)


packages/client-engine-runtime/src/interpreter/data-mapper.ts
```diff
@@ -240,19 +240,36 @@ function mapValue(
...
-      if (value instanceof Uint8Array) {
-        return { $type: 'Bytes', value: Buffer.from(value).toString('base64') }
+      switch (scalarType.encoding) {
+        case 'base64':
+          if (typeof value !== 'string') {
+            throw new DataMapperError(
+              `Expected a base64-encoded byte array in column '${columnName}', got ${typeof value}: ${value}`,
+            )
+          }
+          return { $type: 'Bytes', value: value.replace('\n', '') }
```
aqrln: Why is this necessary? I thought base64 decoding in `Buffer` skips any whitespace

jacek-prisma: it does, but the database returns newlines in the string and we don't use `Buffer.from` here, which results in a discrepancy in the test snapshot

jacek-prisma: we could maybe roundtrip it through Buffer.from in `normalizeTaggedValue`

aqrln: Is that test snapshot on the engine side or on the client side? I'd rather do this replacement in https://github.com/prisma/prisma/blob/d95f8a63128196d85e6a004c0db46b26c4335b4c/packages/client-engine-runtime/src/json-protocol.ts#L85-L86 if possible so we don't need to pay the cost of it in the client outside of tests (this is technically also used by QPE for Accelerate in addition to the QC test executor in prisma-engines by I think that's fine)

jacek-prisma: done @aqrln

aqrln (approved):

PR #27939, opened 2025-08-20
test(qc): remove test failure lists

jacek-prisma:
Removes the QC failure lists and related code


packages/adapter-pg/src/pg.ts
```diff
@@ -41,6 +41,7 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
     debug(`${tag} %O`, query)
 
     const { fields, rows } = await this.performIO(query)
+    debug(`${tag} rows: %O`, rows)
```
aqrln: intentional?

jacek-prisma: oops, leftover from something else

packages/client/helpers/functional-test/run-tests.ts
```diff
@@ -35,8 +35,6 @@ const jestArgs = {
   // Use this flag to show full diffs and errors instead of a patch.
   '--expand': Boolean,
   '-e': '--expand',
-  // Create json test report -a lso see --outputFile.
-  '--json': Boolean,
```
aqrln: I think it's fine to leave this but no strong opinion

packages/adapter-pg/src/pg.ts
jacek-prisma: removed it

packages/client/helpers/functional-test/run-tests.ts
jacek-prisma: added it back

aqrln (approved):

PR #27941, opened 2025-08-20
feat(qc): accept a udt parser for pg

jacek-prisma:
[ORM-1356](https://linear.app/prisma-company/issue/ORM-1356/fix-enum-arrays-in-query-raw)

Allows users to specify a custom parser for user-defined types. The parser is only used if all 3 are true:
- the parser is present
- the field type has an OID that falls within the standard postgres range for user-defined types
- the field type has an OID we have no custom parser for

One of the tests demonstrates its usage with enum arrays.


aqrln (approved):
LGTM. This needs documentation, maybe even a guide in the docs.

aqrln:
@jacek-prisma there are errors on CI:

```
    PrismaClientConstructorValidationError: "adapter" property can only be provided to PrismaClient constructor when "driverAdapters" preview feature is enabled.
```

aqrln (approved):

PR #27945, opened 2025-08-20
chore(client-generator-ts): Throw complile error for `undefined` in Json Filters

### LINKED ISSUE #25819: Using `undefined` in a JSON filter should be a TypeScript error
## Problem

As per https://github.com/prisma/prisma/issues/7617, using:

```ts
await client.myModel.findMany({
  where: {
    myJsonColumn: {
      path: ['bar'],
      equals: undefined,
    },
  },
})
```

gives an error:

```
Input error. A JSON path cannot be set without a scalar filter.
```

This is a run-time error, hit when executing the code

## Suggested solution

It would be great if attempting to pass `undefined` here instead gives a TS compilation error, so that it's caught at build time, not run-time.

Aijeyomah:
- fixes https://github.com/prisma/prisma/issues/25819

- throw a compile error if `undefined` is passed as a value for JSON filter 

Before:
```
  { path: ['petName'], equals: undefined }  // Throws runtime error
```
```
After:
  { path: ['petName'], equals: undefined }  // now throws compile error
```
Changes:
- Enforce a value for key value types (disallow `undefined`)

**Loom**
https://www.loom.com/share/2fbe50205e31473b8f5b19f41edb6202?sid=f6c56240-5973-450f-b1e4-1cd5799a7b91


Aijeyomah:
@jkomyno could you help review this PR? I tested the changes in `sandbox/my-repro` and things look good on my end.

I need some clarity on how to update the type declaration so the tests under `types/json_filtering_mysql` and `types/json_filtering_postgres` pick up the new types. I went through CONTRIBUTING.md and the testing docs, but I could not find the exact step to refresh the generated client types used by the tsd tests. It still seems to read the old types. Could you point me to the right command or workflow to regenerate the client type declarations that these tsd tests consume? Thanks!

aqrln:
Thanks for the PR @Aijeyomah. Does [`strictUndefinedChecks`](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/null-and-undefined#strict-undefined-checks-preview-feature) solve this issue by any chance or is this a completely separate thing that needs to be handled specially?

> I need some clarity on how to update the type declaration so the tests under types/json_filtering_mysql and types/json_filtering_postgres pick up the new types. I went through CONTRIBUTING.md and the testing docs, but I could not find the exact step to refresh the generated client types used by the tsd tests. It still seems to read the old types. Could you point me to the right command or workflow to regenerate the client type declarations that these tsd tests consume? Thanks!

These tests should generate a new client each time you run them so you shouldn't need to do anything special. Make sure to rebuild the generator itself though (run `pnpm -r dev` to build the whole repo).

Aijeyomah:
> Thanks for the PR @Aijeyomah. Does [`strictUndefinedChecks`](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/null-and-undefined#strict-undefined-checks-preview-feature) solve this issue by any chance or is this a completely separate thing that needs to be handled specially?
> 
> > I need some clarity on how to update the type declaration so the tests under types/json_filtering_mysql and types/json_filtering_postgres pick up the new types. I went through CONTRIBUTING.md and the testing docs, but I could not find the exact step to refresh the generated client types used by the tsd tests. It still seems to read the old types. Could you point me to the right command or workflow to regenerate the client type declarations that these tsd tests consume? Thanks!
> 
> These tests should generate a new client each time you run them so you shouldn't need to do anything special. Make sure to rebuild the generator itself though (run `pnpm -r dev` to build the whole repo).

Yes, strictUndefinedChecks solves the issue, so there’s no need for a separate change. I’m closing this PR now. Thanks for the guidance!

PR #27948, opened 2025-08-21
feat: add AI safety checks to migration commands

aqrln:
Implement AI safety checks for dangerous CLI commands.

Many stories where people lost their data because an AI agent recklessly decided to drop or reset a production database have been gaining visibility on social media recently.

We should be proactive and implement additional confirmation checks for AI agents that force it to stop and ask the human to review and confirm the action, rather than the agent making the decision.

The list of checks for common AI agents is in no way complete and can be expanded in the future. Some agents (specifically Zed) don't currently provide a way to easily detect them without false positives when using integrated terminal, we should open a feature request.

See [this thread on Slack](https://prisma-company.slack.com/archives/C058VM009HT/p1755724422234739).

### DIFF
--- a/packages/migrate/src/commands/DbDrop.ts
+++ b/packages/migrate/src/commands/DbDrop.ts
@@ -15,6 +15,7 @@ import {
 import { bold, dim, red, yellow } from 'kleur/colors'
 import prompt from 'prompts'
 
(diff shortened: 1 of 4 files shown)


FGoessler (approved):
Nice one! Great prompt imo! 🙌

PR #27960, opened 2025-08-22
feat(client-generator-ts): set default "runtime" value based on std-env

jkomyno:
Experiment: let [`unjs/std-env`](https://github.com/unjs/std-env) figure out the default `prisma-client` runtime value if not provided, rather than always default to `nodejs`.

/integration


packages/client-generator-ts/src/runtime-targets.ts
```diff
@@ -49,6 +51,21 @@ function parseRuntimeTarget(target: RuntimeTarget | (string & {})): RuntimeTarge
   }
 }
 
+export function defaultRuntimeTargetFromEnv(): RuntimeTargetInternal {
+  switch (unjsRuntime) {
+    case 'node':
+    case 'bun':
+    case 'deno':
```
aqrln, line 58: Shouldn't we use `deno` in the Deno case?

packages/client-generator-ts/src/runtime-targets.ts
```diff
@@ -49,6 +51,21 @@ function parseRuntimeTarget(target: RuntimeTarget | (string & {})): RuntimeTarge
...
+      return 'nodejs'
+    case 'edge-light':
+      return 'vercel-edge'
+    case 'workerd':
+      return 'workerd'
```
aqrln, line 63: How are these code paths possible? You can't run `prisma generate` in workerd or edge-light. Isn't `std-env/runtime` the runtime in which the code is currently running?

packages/client-generator-ts/src/generator.ts
```diff
@@ -57,7 +57,8 @@ export class PrismaClientTsGenerator implements Generator {
     const outputDir = getOutputPath(options.generator)
     const tsconfig = getTsconfig(outputDir)?.config
 
-    const target = config.runtime !== undefined ? parseRuntimeTargetFromUnknown(config.runtime) : 'nodejs'
+    const target =
+      config.runtime === undefined ? defaultRuntimeTargetFromEnv() : parseRuntimeTargetFromUnknown(config.runtime)
```
bricefriha, line 61: Did you mean to change the indent?

To be fair, it takes more room now that you added `defaultRuntimeTargetFromEnv()`

What I would have done is to break the line at the `?`


```TypeScript
    const target = config.runtime === undefined ?
     defaultRuntimeTargetFromEnv() : parseRuntimeTargetFromUnknown(config.runtime)
```

PS: it really doesn't matter, its the nittiest pick ever 😄

bricefriha:
Looks good, aside from @aqrln 's concerns.
Is there any way to write tests on this?

ineshbose:
Would this be re-targetted for v7?

I was happy that diff between nodejs and workerd generated clients decreased in v7, but I'm still finding minor differences and feel runtime-specific logic isn't great to maintain and maybe Prisma can offload some of that through exports from `std-env` (including `process`, `env`, etc)

PR #27961, opened 2025-08-24
fix(client-engine-runtime): update data-mapper to natively support bigint value

jimexist:
currently it only handles string and number

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/data-mapper.ts
+++ b/packages/client-engine-runtime/src/interpreter/data-mapper.ts
@@ -173,7 +173,7 @@ function mapValue(
     }
 
     case 'bigint': {
-      if (typeof value !== 'number' && typeof value !== 'string') {
+      if (typeof value !== 'number' && typeof value !== 'string' && typeof value !== 'bigint') {
         throw new DataMapperError(`Expected a bigint in column '${columnName}', got ${typeof value}: ${value}`)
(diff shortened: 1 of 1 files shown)


aqrln:
Hmm, in theory it shouldn't be possible right now because driver adapters must be converting all results to JSON-serializable values (it's necessary for driver adapters to be usable with the query engine, we can only get rid of these unnecessary conversions back and forth when we remove the native query engine in Prisma 7). If you get a native JS bigint at this point, technically that would be a bug in the driver adapter. Do you have a reproduction where you get a bigint here?

jimexist:
> Hmm, in theory it shouldn't be possible right now because driver adapters must be converting all results to JSON-serializable values (it's necessary for driver adapters to be usable with the query engine, we can only get rid of these unnecessary conversions back and forth when we remove the native query engine in Prisma 7). If you get a native JS bigint at this point, technically that would be a bug in the driver adapter. Do you have a reproduction where you get a bigint here?

good point

turns out our own code has

`pg.types.setTypeParser(20, 'text', BigInt)`

somewhere in the code.

not sure what to make of it, should this be made compatible?

jacek-prisma:
We currently can't support BigInt being returned from driver adapters, because their results are meant to be plain JSON values, although this might change in the future (it's a requirement of the query engine).

PR #27969, opened 2025-08-26
feat(mssql driver adapter): ORM-1137 add support for mssql connection strings in the driver adapter

FGoessler:
This PR adds support for MSSQL connection strings in the driver adapter as part of ORM-1137.

## Changes
- Added support for MSSQL connection strings in the adapter
- Updated adapter to handle connection string parsing and configuration

## Related Issue
Related to ORM-1137


packages/adapter-mssql/src/mssql.test.ts
```diff
@@ -0,0 +1,104 @@
...
+describe('PrismaMssqlAdapterFactory', () => {
+  describe('constructor with connection string', () => {
+    it('should extract schema from connection string and merge with options', () => {
+      const connectionString = 'sqlserver://localhost;database=testdb;schema=custom'
+      const options = { onPoolError: vi.fn() }
+
+      const factory = new PrismaMssqlAdapterFactory(connectionString, options)
+
+      // Access private properties for testing
+      const privateOptions = (factory as any).options
```
aqrln: I'm not a fan of testing the implementation details instead of the interface of the class like this. If we want to test parsing the connection string, then these should be unit tests for the `parseConnectionString` function itself IMO.

packages/adapter-mssql/src/connection-string.ts
```diff
@@ -0,0 +1,174 @@
...
+  const parts = withoutProtocol.split(';')
+
+  const config: sql.config = {
+    server: '',
+    options: {},
+    pool: {},
+  }
+
+  // Parse the first part which contains host and port
+  const [hostPart, ...paramParts] = parts
```
jacek-prisma: I'd inline`const [hostPart, ...paramParts]` in line 57 to make it more clear that we're assuming there should be at least one element generated by `split` (otherwise hostPart could be undefined)

packages/adapter-mssql/src/mssql.test.ts
FGoessler: Yeah let's drop those tests. They add little value and we already the connection string parsing unit tested explicitly.

packages/adapter-mssql/src/connection-string.ts
FGoessler: Inlined it. But split anyways always generates at least one element - even from an empty string. ;)

jacek-prisma: yeah I realized that, just wanted it to be explicit and visible that it's coming from a split

jacek-prisma (approved):

PR #27976, opened 2025-08-27
feat(qc): add a compile span

jacek-prisma:
[ORM-1365](https://linear.app/prisma-company/issue/ORM-1365/add-prismaclientcompile-span)

### DIFF
--- a/packages/client/src/runtime/core/engines/client/ClientEngine.ts
+++ b/packages/client/src/runtime/core/engines/client/ClientEngine.ts
@@ -1,3 +1,4 @@
+import { SpanKind } from '@opentelemetry/api'
 import { QueryCompiler, QueryCompilerConstructor, QueryEngineLogLevel } from '@prisma/client-common'
 import {
   BatchResponse,
@@ -461,7 +462,12 @@ export class ClientEngine implements Engine {
 
     let queryPlan: QueryPlanNode
     try {
-      queryPlan = this.#withLocalPanicHandler(() => queryCompiler.compile(queryStr), queryStr) as QueryPlanNode
+      queryPlan = this.#withLocalPanicHandler(() =>
+        this.#withCompileSpanAndEvent({
+          queries: [query],
+          execute: () => queryCompiler.compile(queryStr) as QueryPlanNode,
+        }),
+      )
     } catch (error) {
       throw this.#transformCompileError(error)
     }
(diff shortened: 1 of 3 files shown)


FGoessler (approved):

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -619,6 +628,20 @@ export class ClientEngine implements Engine {
         )
     }
   }
+
+  #withCompileSpanAndEvent<T>({ queries, execute }: { queries: JsonQuery[]; execute: () => T }): T {
```
aqrln:
```suggestion
  #withCompileSpan<T>({ queries, execute }: { queries: JsonQuery[]; execute: () => T }): T {
```

this method only produces a span, it doesn't emit an event on `PrismaClient`

jacek-prisma: oh, that's a leftover from a copypaste 👍

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -619,6 +628,20 @@ export class ClientEngine implements Engine {
...
+    return this.tracingHelper.runInChildSpan(
+      {
+        name: 'compile',
+        kind: SpanKind.CLIENT,
```
aqrln:
```suggestion
```

It shouldn't be `CLIENT` because it's not a database call. The appropriate span kind here is `INTERNAL`, which is the default.

packages/client/src/runtime/core/engines/client/ClientEngine.ts

packages/client/src/runtime/core/engines/client/ClientEngine.ts
jacek-prisma: removed

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -506,7 +511,10 @@ export class ClientEngine implements Engine {
 
     let batchResponse: BatchResponse
     try {
-      batchResponse = queryCompiler.compileBatch(request)
+      batchResponse = this.#withCompileSpan({
+        queries,
+        execute: () => queryCompiler.compileBatch(request),
+      })
```
aqrln: Unrelated to this PR, but this is probably missing `#withLocalPanicHandler`.

aqrln (approved):

packages/client/src/runtime/core/engines/client/ClientEngine.ts
jacek-prisma: added it now

PR #27978, opened 2025-08-27
feat: ORM-1375 create `browser.ts` entry point in new generator

FGoessler:
This PR adds a new entry point `browser.ts` to the generated output of the new prisma-client generator. This entry point is suitable to be used in browser environments that e.g. do not provide node.js apis. Resolves https://github.com/prisma/prisma/issues/27341.

It does NOT contain any actual `PrismaClient` - this is in contrast to the old generator's approach that exposes stubs for `PrismaClient` and other objects. 
Types (except for `PrismaClient` itself) are still all available to support scenarios for of overlapping frontend/backend user code.

Notable changes:
- Created new `browser.ts` and `prismaNamespaceBrowser.ts` generated outputs.
- Updated the `index-browser` runtime bundle to also be available as esm version.
- Simplified the enum exporting along the way (should resolve https://github.com/prisma/prisma/issues/27943 / ORM-1378)
- Added an E2E test to type-check and runtime test this new entry point.

I decided to call this entry point now simply `browser.ts` as opposed to the original `web-client.ts`. I think this better aligns with the ecosystems naming (e.g. in export maps) and is simpler to grasp for our users. Open to feedback and changing this to something else though. :) 

With this change users can import all types (e.g. for models, the various input types, but also things like `Prisma.JsonValue`) as well as utility objects like `Prisma.DbNull`.

Example:

```typescript
import { Quotes, QuoteKind, Prisma } from '../lib/generated/prisma/browser'

async function createQuote(
  quote: string,
  kind: QuoteKind, // use enum types
  jsonData?: Prisma.JsonValue, // You can also use helpers like `Prisma.JsonValue`
): Promise<Quotes> {
  const createQuote: Prisma.QuotesCreateInput = {
    quote,
    kind,
    jsonData
  }

  // e.g. send the data to your backend that performs the actual database operations and returns the created quote
  // ...

  return createdQuote
}

function printModelNames(): void {
  alert('Model names are: ' + JSON.stringify(Prisma.ModelName, null, 2)) // get all model names at runtime - helpful for some meta programming
}

function createDecimal() {
  return new Prisma.Decimal(1.23)
}

```

> [!WARNING]
> While `Prisma.DbNull`, `Prisma.JsonNull` and `Prisma.AnyNull` are available they are NOT (yet?) transferrable from the frontend to the backend as they are class based and not directly serializable. Using them on the frontend hence makes likely very little sense but their presence could help to resolve otherwise arising typecheck issues in shared backend/frontend code.
>
[...]

PR #27989, opened 2025-08-28
feat: ORM-1074 extension dev entry point

FGoessler:
This PR extends the `@prisma/client/extension` entry point such that it provides further relevant utilities for extension developers in a reliable import that does NOT require an explicit client to be generated.

To ensure that exported error classes and relevant singleton types (e.g. `Prisma.DbNull`) and utilities like `Prisma.Decimal` correctly interoperate for equality and `instanceof` checks we need to ensure that generated clients and this direct entry point (as well as all other entry points) actually reference the same shared utilities. 
The build process now creates a dedicated `utilities` file and ensures that all other bundles treat its contents as external. So instead of each entry point bundling e.g. its own version of `PrismaClientInitializationError` they all refer to the same version in the utility file.

Overall the structure of imports with this change becomes:

```mermaid
graph TD;
  uc[User Code]
  ec[User Extension Code]
  gcc[Generated Code client.ts]
  gcb[Generated Code browser.ts]
  pce["@prisma/client/extension"]
  pcri["@prisma/client/runtime/index-browser"]
  pcrc["@prisma/client/runtime/client"]
  pcru["@prisma/client/runtime/utilities"]

  uc --> gcc
  uc --> gcb
  ec --> pce
  pce --> pcru
  gcc --> pcrc
  gcb --> pcri
  pcri --> pcru
  pcrc --> pcru
```

> [!NOTE] 
> This changes assumes that any `@prisma/client/runtime/*` entry point is an internal API that is used only from generated code but NOT by users directly.

This enables extension developers to use the `@prisma/client/extension` entry point to be used like this:

```typescript
import {
  Prisma,
  PrismaClient,
  PrismaClientInitializationError,
  sqltag,
  Decimal,
} from '@prisma/client/extension'
```

## 🚧  What is left to do / remaining issues:
- [ ] Currently the utilities are exposed as top level exports. Not under the `Prisma` namespace. This is inconsistent with our regular structures.
- [ ] `DbNull`, `JsonNull` and `AnyNull` are still quite tricky. Currently they have to be exposed via the `objectEnumValues` and are class based. For better frontend/back interoperability they should likely be simpler to serialize and handle branded objects. This is however better done in a separate change. => I hence marked them deprecated for now so we can do breaking changes here.
  - [ ] Actually, @aqrln and @jkomyno suggest exposing singleton values and `is*` functions (e.g., `dbNull` and `isDbNull(x: any): boolean`), hiding the fact that `DbNull` and similar utilities are actually classes using unique symbols.
- [ ] The `@prisma/client/extension` export is largely based on the `scripts/default-index.ts` dummy export. Am not sure how sustainable this is or if this is just a hacky workaround. The typing of this are also not very helpful imo as `PrismaClient` is just an alias for `any` in there.
- [ ] We lack any documentation for `@prisma/client/extension`.
- [ ] Create and fix E2E test.

🚧 Note: This PR currently also includes the changes from https://github.com/prisma/prisma/pull/27978.

/integration


sandbox/query-compiler/my-extension.ts
```diff
@@ -0,0 +1,55 @@
+// import { Prisma } from './prisma/client/client'
+// import { Prisma } from '@prisma/client/extension'
+import { Prisma } from './prisma/client/client'
+import { PrismaClientInitializationError as ErrExt, Decimal as DecimalExt, Prisma as PrismaExt, DbNull, objectEnumValues } from '@prisma/client/extension'
```
jkomyno, line 4: - `objectEnumValues` shouldn't be "visible" to the user-land
- `DbNull` and other classes shouldn't be exported
- Singleton values and `is*` predicates should be made available instead

FGoessler:
Superseded by https://github.com/prisma/prisma/pull/28202

PR #27994, opened 2025-08-29
feat: support unixepoch-ms timestamps in the sqlite adapters

jacek-prisma:
[ORM-1390](https://linear.app/prisma-company/issue/ORM-1390/implement-backward-compatibility-for-qe-timestamps)

### DIFF
--- a/packages/adapter-better-sqlite3/src/better-sqlite3.ts
+++ b/packages/adapter-better-sqlite3/src/better-sqlite3.ts
@@ -43,7 +43,10 @@ class BetterSQLite3Queryable<ClientT extends StdClient> implements SqlQueryable 
   readonly provider = 'sqlite'
   readonly adapterName = packageName
 
-  constructor(protected readonly client: ClientT) {}
+  constructor(
+    protected readonly client: ClientT,
+    protected readonly adapterOptions?: PrismaBetterSQLite3Options,
+  ) {}
 
   /**
    * Execute a query given as SQL, interpolating the given parameters.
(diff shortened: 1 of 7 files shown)


packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -205,22 +212,34 @@ export class PrismaBetterSQLite3Adapter extends BetterSQLite3Queryable<StdClient
 
 type BetterSQLite3InputParams = BetterSQLite3Options & {
   url: ':memory:' | (string & {})
-  shadowDatabaseURL?: ':memory:' | (string & {})
+}
+
+export type PrismaBetterSQLite3Options = {
+  shadowDatabaseUrl?: string
+  timestampFormat?: 'iso8601' | 'unixepoch-ms'
```
jacek-prisma, line 219: I call it `unixepoch-ms` rather than `unixepoch` to make it explicit it uses millis rather than seconds, `unixepoch` as a time format specifier in sqlite means seconds not millis

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -134,32 +137,36 @@ class BetterSQLite3Queryable<ClientT extends StdClient> implements SqlQueryable
 }
 
 class BetterSQLite3Transaction extends BetterSQLite3Queryable<StdClient> implements Transaction {
+  readonly #unlockParent: () => void
```
aqrln, line 140: Unrelated, but I wonder if it's still necessary. If I recall correctly, it was introduced in the libSQL adapter to work around a certain bug or limitation of an early version of the driver.

jacek-prisma: I think it's necessary because the driver cannot be used directly during a transaction. Maybe a recent version has changed something, but we can take care of that in another PR.

aqrln (approved):

PR #28001, opened 2025-09-01
fix: update tsconfig to not use legacy moduleResolution strategy anymore

FGoessler:
E2E tests for typescript 6.0.0 failed due with:

```
error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

This change should fix this.


aqrln:
Fixed in https://github.com/prisma/prisma/pull/28013.

PR #28003, opened 2025-09-01
chore: remove strip-ansi

dulmandakh:
remove `strip-ansi` and use `stripVTControlCharacters` from node where possible. Below is a snippet from `strip-ansi` README.md file

> Node.js has this built-in now with [stripVTControlCharacters](https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr). The benefit of this package is consistent behavior across Node.js versions and faster improvements. The Node.js version is actually based on this package.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -299,6 +299,7 @@
     "source-map-support": "0.5.21",
     "sql-template-tag": "5.2.1",
     "stacktrace-parser": "0.1.11",
+    "strip-ansi": "7.1.0",
     "strip-indent": "4.0.0",
     "tempy": "3.0.0",
     "ts-node": "10.9.2",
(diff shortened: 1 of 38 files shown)


packages/client/package.json
```diff
@@ -299,7 +299,7 @@
     "source-map-support": "0.5.21",
     "sql-template-tag": "5.2.1",
     "stacktrace-parser": "0.1.11",
-    "strip-ansi": "6.0.1",
+    "strip-ansi": "7.1.0",
```
aqrln, line 302: I assume this should also be removed

aqrln (commented):
This is fantastic! Also thanks for changing some imports to use the `node:` prefix along the way. I left one comment and there are some merge conflicts right now, other than that it looks great.

packages/client/package.json
dulmandakh: wrangler is complaining about node prefix, so left this package alone with strip-ansi dependency.

aqrln (approved):

PR #28007, opened 2025-09-01
chore: adapter-pg - use vitest

dulmandakh:
run tests for `adapter-pg` with `vitest` instead of `jest`

### DIFF
--- a/packages/adapter-pg/vitest.config.mts
+++ b/packages/adapter-pg/vitest.config.mts
@@ -1,0 +1,8 @@
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    globals: true,
+    include: ['**/*.test.ts'],
+  },
+})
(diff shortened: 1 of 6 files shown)


aqrln (approved):

packages/adapter-pg/vitest.config.mts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
+  test: {
+    globals: true,
```
aqrln, line 5: fwiw in other packages where we are already using vitest we import `test`, `describe`, `expect` etc explicitly, but I don't have a strong opinion and we can easily change it to be consistent later, so I think it's fine

dulmandakh: Thank you. I was thinking the same, at first make it work then improve.

PR #28008, opened 2025-09-01
chore: move vitest to workspace root

dulmandakh:
move `vitest` to workspace root to be shared with packages.


dulmandakh:
@aqrln please review. I would like to help with migrating tests to vitest, because it's much faster than jest.

PR #28013, opened 2025-09-02
fix(client): use modern module resolution in e2e tests

aqrln:
Use `"module": "nodenext"` and `"moduleResolution": "nodenext"` in the `tsconfig.json` in e2e tests. `node10` resolution is deprecated in `typescript@next` which is now causing our e2e tests to fail.

Closes: https://linear.app/prisma-company/issue/ORM-1417/fix-failing-ecosystem-tests

### DIFF
--- a/packages/client/tests/e2e/typed-sql-query-compiler-adapter-libsql/prisma/schema.prisma
+++ b/packages/client/tests/e2e/typed-sql-query-compiler-adapter-libsql/prisma/schema.prisma
@@ -1,7 +1,9 @@
 generator client {
-  provider        = "prisma-client"
-  previewFeatures = ["typedSql", "queryCompiler", "driverAdapters"]
-  output          = "../src/generated/prisma"
+  provider            = "prisma-client"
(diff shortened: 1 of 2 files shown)


packages/client/tests/e2e/typed-sql-query-compiler-adapter-libsql/prisma/schema.prisma
```diff
@@ -1,7 +1,9 @@
 generator client {
-  provider        = "prisma-client"
-  previewFeatures = ["typedSql", "queryCompiler", "driverAdapters"]
-  output          = "../src/generated/prisma"
+  provider            = "prisma-client"
+  previewFeatures     = ["typedSql", "queryCompiler", "driverAdapters"]
+  output              = "../src/generated/prisma"
+  moduleFormat        = "commonjs"
+  importFileExtension = ""
```
aqrln, line 6: This change is necessary because our heuristics for detecting the default aren't checking the `type` property in `package.json` and always choose ESM by default when `module = "nodenext"`.

jkomyno (approved):

PR #28021, opened 2025-09-03
chore: bump wrangler to latest 3.x

dulmandakh:
this PR bumps wrangler to 3.114.14 or latest 3.x version. Also, adds wranger to workspace root then remove from packages to consolidate version management.

Initially I was planning to bump wrangler to 4.x, but it required Node.js 20 or above. I hope that Prisma would bump Node.js requirement to 20 with a next major version.

### DIFF
--- a/packages/client/tests/e2e/enum-import-in-edge/package.json
+++ b/packages/client/tests/e2e/enum-import-in-edge/package.json
@@ -10,6 +10,7 @@
   "devDependencies": {
     "@types/jest": "29.5.12",
     "@types/node": "18.19.76",
-    "prisma": "/tmp/prisma-0.0.0.tgz"
+    "prisma": "/tmp/prisma-0.0.0.tgz",
+    "wrangler": "3.114.14"
   }
 }
--- a/packages/client/tests/e2e/enum-import-in-edge/package.json
+++ b/packages/client/tests/e2e/enum-import-in-edge/package.json
@@ -10,7 +10,6 @@
   "devDependencies": {
     "@types/jest": "29.5.12",
     "@types/node": "18.19.76",
-    "prisma": "/tmp/prisma-0.0.0.tgz",
-    "wrangler": "3.114.14"
+    "prisma": "/tmp/prisma-0.0.0.tgz"
   }
 }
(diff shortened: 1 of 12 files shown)


dulmandakh:
@aqrln I find it best to consolidate cli tools in monorepo root instead of in packages because those packages won't be developed individually. Also it's easy to manage tools versions.

I'm looking into using Prisma in the future, thus taking care of it by opening these PRs.

packages/client/tests/e2e/enum-import-in-edge/package.json
```diff
@@ -4,8 +4,7 @@
   "main": "index.js",
   "scripts": {},
   "dependencies": {
-    "@prisma/client": "/tmp/prisma-client-0.0.0.tgz",
-    "wrangler": "3.109.3"
```
aqrln, line 8: This one should stay here. E2E tests are running in isolated Docker containers and aren't part of the workspace. Right now the test fails due to this:

```
$ pnpm wrangler dev --ip 127.0.0.1 --port 8787
undefined
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "wrangler" not found
```
https://github.com/prisma/prisma/actions/runs/17420280696/job/49457098246?pr=28021

aqrln:
> Initially I was planning to bump wrangler to 4.x, but it required Node.js 20 or above. I hope that Prisma would bump Node.js requirement to 20 with a next major version.

We will, yes, it's going to be pretty soon.

> I find it best to consolidate cli tools in monorepo root instead of in packages because those packages won't be developed individually. Also it's easy to manage tools versions.

Agree

> I'm looking into using Prisma in the future, thus taking care of it by opening these PRs.

Thanks so much, we really appreciate it!

packages/client/tests/e2e/enum-import-in-edge/package.json
dulmandakh: Ok, I'll fix it tomorrow

aqrln:
@dulmandakh there are now some git conflicts that need to be resolved

aqrln (approved):

aqrln:
Thanks!

PR #28023, opened 2025-09-03
chore: consolidate esbuild dependency

dulmandakh:


aqrln (approved):

packages/client/tests/e2e/engine-not-found-error/bundler-tampered-with-engine-copy/package.json
```diff
@@ -8,7 +8,6 @@
   },
   "devDependencies": {
     "@types/node": "18.19.76",
-    "esbuild": "0.21.4",
```
aqrln, line 11: btw, here this works, unlike wrangler, because esbuild is globally installed in the docker image

PR #28024, opened 2025-09-03
chore: use turborepo for parallel builds

dulmandakh:
I propose to use `turborepo` for the monorepo setup and depend on its dependency resolution algorithm instead of custom `esbuild` plugin. Hence we could save around 3 minutes on clean builds. And below are some numbers for comparison and output if nothing is changed.

```sh
 Tasks:    33 successful, 33 total
Cached:    33 cached, 33 total
  Time:    346ms >>> FULL TURBO
```

### On M1 Pro

Before 

```sh
  Tasks:    33 successful, 33 total
 Cached:    0 cached, 33 total
   Time:    4m53.178s 
```

After

```sh
 Tasks:    33 successful, 33 total
Cached:    0 cached, 33 total
  Time:    2m5.925s 
```

### On GitHub Actions

Before

```sh
 Tasks:    33 successful, 33 total
Cached:    0 cached, 33 total
  Time:    8m43.245s 
```

After

```sh
 Tasks:    33 successful, 33 total
Cached:    0 cached, 33 total
  Time:    5m37.205s 
```


dulmandakh:
@aqrln what do you think about this change?

I think that turbo would allow us further improve developer experience by reducing time required to build, run tests both on local machine and CI/CD. Also we could remove some manual interventions such as dependency checks by relying on monorepo tools.

aqrln:
Amazing, this sounds great. I'll share this with the rest of the team and review it a little bit later.

aqrln (approved):
Thank you!

PR #28033, opened 2025-09-04
chore: simplify standard.dockerfile

dulmandakh:
This PR simplifies `standard.dockerfile`, thus fixes below warnings. I find that it's unnecessary to use multi-staged builds for such simple use case.

```sh
#2 WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
#2 WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 3)
#2 WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 23)
```


dulmandakh:
@aqrln please review and merge

aqrln (approved):
Yeah, there's no point in doing a multi-staged build like this. Thanks!

PR #28035, opened 2025-09-04
chore: watch using turbo

dulmandakh:
This PR changes `watch` script to use `turbo`, which will watch for changes, then compile packages and their dependent packages. Also, removes watch logic from build scripts.


dulmandakh:
@aqrln what do you think about this? this is working just fine for me

aqrln:
@dulmandakh could you please add a `dev` task in addition to `build` in `turbo.json` (make sure to also [add the `DEV` env var to the hash inputs](https://github.com/prisma/prisma/pull/28043)) so that `watch` can continue using `dev` instead of `build`? Or perhaps add separate `watch:dev` and `watch:build` tasks.

Even with Turborepo, it takes at least 1.5 min for me to build a production build of `@prisma/client` using `pnpm build`, which is too much when working on this package itself and frequently making changes, so the significantly faster `dev` tasks are still very useful.

I have a [draft PR](https://github.com/prisma/prisma/pull/28028) that introduces intra-package concurrency in the build scripts but it doesn't make `build` as much faster as I hoped it would, and it also makes esbuild crash on CI somehow (I guess the deadlock bug gets triggered by esbuild plainly having less CPU time for each individual process, especially in CI where there are a lot less cores than bundles to build and there's a lot of preemption going on). While it makes sense that running multiple esbuild processes in parallel isn't helpful (or even actively harmful) since esbuild itself is already highly parallel and utilizes all cores, I hoped at least building `.d.ts` bundles in parallel would've been beneficial. The build script would need some refactoring though to allow first running esbuild for each bundle sequentially, and then tsc and api-extractor in parallel.

dulmandakh:
@aqrln ok, I'll make dev task for turbo. Any suggestions on removing watch code?

aqrln:
@dulmandakh removing the watch code and replacing it with `turbo watch` sounds great, no objections to this. I don't believe this watch code worked well anyway, I personally just used [`watchexec`](https://github.com/watchexec/watchexec) instead.

PR #28046, opened 2025-09-05
feat(qc): stabilize driver adapters and the query compiler

jacek-prisma:
[ORM-1420](https://linear.app/prisma-company/issue/ORM-1420/remove-querycompiler-and-driveradapter-feature-flags)


packages/client-generator-js/src/generateClient.ts
```diff
@@ -236,18 +235,13 @@ export async function buildClient({
     // In short: A lot can be simplified, but can only happen in GA & P6.
     fileMap['default.js'] = JS(trampolineTsClient)
     fileMap['default.d.ts'] = TS(trampolineTsClient)
-    if (usesClientEngine) {
```
jacek-prisma, line 239: I think this was unreachable because of this validation:
https://github.com/prisma/prisma/blob/68fdbb40a1e0f7876ae5a0e677559682b104a71a/packages/client-generator-ts/src/generateClient.ts#L216

packages/client-generator-ts/src/generateClient.ts
```diff
@@ -213,10 +213,6 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
 
   const clientEngineType = getClientEngineType(generator)
```
bricefriha, line 214: It doesn't seem like we are using `clientEngineType` in this scope after removing  the queryCompiler check, are we? 🤔
```suggestion
```

packages/migrate/src/commands/DbPull.ts
```diff
@@ -237,14 +237,16 @@ Set composite types introspection depth to 2 levels
...
+            )
             const validationWarning = `Without the ${bold(
-              'driverAdapters',
+              'schemaEngineDriverAdapters',
             )} preview feature, the schema introspected via the ${bold('--local-d1')} flag will not work with ${bold(
               '@prisma/client',
             )}.`
 
-            if (hasDriverAdaptersPreviewFeature) {
+            if (hasSchemaEngineeDriverAdaptersPreviewFeature) {
```
bricefriha: I know it not specific to this PR. Does anyone know why we went for a `if.. else` instead of a one liner?
like:

```TS
return hasSchemaEngineeDriverAdaptersPreviewFeature ?   result : { ...result, validationWarning };
```

I personally think it's shorter and not necessarily harder to read

bricefriha (commented):
I like that!! 
I added a few questions. No biggies.

packages/client-generator-ts/src/generateClient.ts
jacek-prisma: it is still used in line 263

bricefriha: Sorry, my bad, don't know how I missed that. Thanks for pointing it out!

packages/client/tests/e2e/driver-adapters-accelerate/tests/no-config.ts
```diff
@@ -2,13 +2,14 @@ test('error is shown when both driver adapters and accelerate are not configured
...
+      const client = new PrismaClient({})
+      client.$connect()
+    }
 
     expect(newClient).toThrowErrorMatchingInlineSnapshot(`
-"PrismaClient failed to initialize because it wasn't configured to run in this environment (Node.js).
-In order to run Prisma Client in an edge runtime, you will need to configure one of the following options:
-- Enable Driver Adapters: https://pris.ly/d/driver-adapters
-- Enable Accelerate: https://pris.ly/d/accelerate"
+"PrismaClient is unable to run in this browser environment, or has been bundled for the browser (running in \`Node.js\`).
```
aqrln: I'm not sure this is the correct error message here, the previous one seemed to be more helpful or accurate. I believe this error should be shown only when attempting to use a stub client from the browser bundle.

packages/migrate/src/__tests__/DbPull/sqlite.test.ts
```diff
@@ -69,7 +68,7 @@ describeMatrix({ providers: { d1: true }, driverAdapters: {} }, 'D1', () => {
       ✔ Introspected 2 models and wrote them into prisma/schema.prisma in XXXms
             
       Run prisma generate to generate Prisma Client.
-      Without the driverAdapters preview feature, the schema introspected via the --local-d1 flag will not work with @prisma/client.
+      Without the schemaEngineDriverAdapters preview feature, the schema introspected via the --local-d1 flag will not work with @prisma/client.
```
aqrln: Since this is talking about `@prisma/client`, I don't think `schemaEngineDriverAdapters` is relevant here

aqrln: My understanding is this message just tells that a D1 adapter is necessary in the Client, or am I misunderstanding what's meant here?

packages/migrate/src/commands/DbPull.ts
aqrln: > Does anyone know why we went for a if.. else instead of a one liner?

I don't think it matters much one way or the other.

> I personally think it's shorter and not necessarily harder to read

It's certainly harder to read in the GitHub PR review UI:

Image

bricefriha: Thanks for the response. 
Again no biggies, I don't think it's wrong or anything, I just try to understand why we did it in the first place. 🙂

> I don't think it matters much one way or the other.

Fair, the reason I think it matters, is because, as it stands the code is taking a few unnecessary lines and is just about as wide. (counting the if statement indent)

> It's certainly harder to read in the GitHub PR review UI:

I don't think anybody is using the GitHub PR review comment UI to read the entire code. This matters when we open the code in a code editor, IDE or even when opening the entire file from GitHub.

To be honest, if I were to write the existing code in the comment, it would be just as hard to read.

packages/client/tests/e2e/driver-adapters-accelerate/tests/no-config.ts
jacek-prisma: Makes sense, changed it to be the same as the exception thrown in `getEngineInstance`

packages/migrate/src/__tests__/DbPull/sqlite.test.ts
jacek-prisma: That's true, I've misread the warning initially and thought it's about using the DA for migrations, but it's about the client, I think this warning is no longer needed, so I removed it

aqrln (dismissed):
Looks great 🚢

aqrln (approved):

PR #28047, opened 2025-09-05
chore: client-engine-runtime uses vitest

dulmandakh:


aqrln (dismissed):
LGTM (pending CI), thank you!

dulmandakh:
@aqrln a test is timing out on CI, but it's passing just fine on my local machine. Any suggestions?

aqrln:
@dulmandakh no ideas off the top of my head

tensordreams:
Thanks for this, @dulmandakh, and sorry it sat open so long.

This has since landed on `main` via #29234 ("chore: migrate client-engine-runtime to vitest") — `packages/client-engine-runtime` now runs `vitest run`, its `jest.config.js` is gone, and the test files carry explicit `vitest` imports, which is exactly what this PR proposed. Closing as a duplicate of that one; thank you for pushing the vitest migration forward.

PR #28057, opened 2025-09-07
fix(adapter-pg): error event listener leak

Zhincore:
Fixes an issue where when executing enough transactions in quick succession and pg-pool starts to reuse clients, error handlers are re-attached to them but old ones aren't removed, causing the `MaxListenersExceededWarning` and supposedly a memory leak.

Closes #28037

### DIFF
--- a/packages/adapter-pg/src/__tests__/pg.test.ts
+++ b/packages/adapter-pg/src/__tests__/pg.test.ts
@@ -51,11 +51,25 @@ describe('PrismaPgAdapterFactory', () => {
     const config: pg.PoolConfig = { user: 'test', password: 'test', database: 'test', port: 5432, host: 'localhost' }
     const factory = new PrismaPgAdapterFactory(config)
     const adapter = await factory.connect()
+
+    const mockConnection = {
+      on: vi.fn(),
+      removeListener: vi.fn(),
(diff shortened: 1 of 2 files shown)


jacek-prisma (approved):
LGTM! Thanks for the contribution

afonsomatos:
@jacek-prisma How do we use this? The final version is 6.15.0.

clemens:
Does this actually fix the problem? Just from looking at the code, it seems to me that it just makes it more unlikely, right? Wouldn't the error still appear if there were more concurrent transactions than the respective max listeners setting allows?

jacek-prisma:
@afonsomatos it'll be in the next minor release which will release soon
@clemens it should fix the problem, because concurrent transactions will not re-use connections, so each connection should have 1 listener - this problem would occur when connections were re-used

PatrickRogg:
@jacek-prisma I just quickly tried out v6.16.0-dev.37 and the issue still happening. Is this commit included in there already?

Zhincore:
@PatrickRogg It helped for me, the issue is gone. Make sure you updated the `adapter-pg` package specifically

PatrickRogg:
Ah, my bad. The package wasn't updated in my turbo repo. Works now. Thanks for fixing this @Zhincore

PR #28059, opened 2025-09-08
feat(client-generator-ts): removed "Prisma.Validator" from "prisma-client" generator

jkomyno:
This PR:
- closes [ORM-1376](https://linear.app/prisma-company/issue/ORM-1376/drop-prismavalidator-in-new-generator)


aqrln (approved):

CHC383:
Hi @jkomyno, can you share more context on why `validator` is removed, and what's the alternative on `prisma-client` generator? Shall we import from `@prisma/client/runtime/client` directly?
```ts
import { Public as PrismaPublic } from "@prisma/client/runtime/client";

PrismaPublic.validator...
```

arthur-ver:
Hi! This change broke some code without any notice after updating to the latest Prisma version. Is there some info in the docs what we should use instead? Also, why was it removed from `prisma-client`, but not `prisma-client-js`?

PR #28062, opened 2025-09-08
feat(client-generator-ts): re-export enums directly in `prisma-client` generator

### LINKED ISSUE #27943: Reexport enums directly when using `prisma-client` provider
### Feature Summary

Hi,
I'd like the prisma enums in the `client.ts` of the new `prisma-client` provider to be re-exported directly.

I'm using the new `prisma-client` provider (instead of prisma-client-js`) and [tsoa](https://github.com/lukeautry/tsoa) for building a rest api with an OpenAPI 3 spec. tsoa has a problem with the Prisma enums being imported in the `$Enums` namespace and then re-exported one by one. Why can't we just use
```ts
export * from './enums.js';
```

instead of 
```ts
import * as $Enums from "./enums.js"

[...]

export type OrderOrigin = $Enums.OrderOrigin
export const OrderOrigin = $Enums.OrderOrigin

export type PaymentType = $Enums.PaymentType
export const PaymentType = $Enums.PaymentType

export type PrintJobStatus = $Enums.PrintJobStatus
export const PrintJobStatus = $Enums.PrintJobStatus

export type TaxRate = $Enums.TaxRate
export const TaxRate = $Enums.TaxRate
```

tsoa currently thinks those are separate enums. While this could be seen as a tsoa issue, I was wondering why this import and re-export is necessary via the separate `$Enums` namespace.

### Proposed Solution

```ts
export * from './enums.js';
```

in `client.ts`

### Alternative Solutions

I'm currently using a script modifying the files after `prisma generate` ran. This feels hacky.

### Potential Considerations

As far as I can see, this does not break anything.

### Prisma Version
```
prisma                : 6.14.0
@prisma/client        : 6.14.0
Computed binaryTarget : darwin-arm64
Architecture          : arm64
Node.js               : v22.17.0
TypeScript            : 5.6.3
PSL                   : @prisma/prisma-schema-wasm 6.14.0-25.717184b7b35ea05dfa71a3236b7af656013e1e49
Schema Engine         : schema-engine-cli 717184b7b35ea05dfa71a3236b7af656013e1e49 (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Default Engines Hash  : 717184b7b35ea05dfa71a3236b7af656013e1e49
Preview Features      : driverAdapters, queryCompiler, typedSql
```

Prisma Client

### Additional Context

_No response_

### Pre-Submission Checklist

jkomyno:
This PR:
- closes [ORM-1378](https://linear.app/prisma-company/issue/ORM-1378/reexport-enums-more-conveniently)
- closes https://github.com/prisma/prisma/issues/27943


packages/client-generator-ts/src/TSClient/file-generators/ClientFile.ts
```diff
@@ -29,6 +29,7 @@ export function createClientFile(context: GenerateContext, options: TSClientOpti
 
   const exports = [
     ts.moduleExportFrom(context.importFileName('./enums')).asNamespace('$Enums'),
+    ts.moduleExportFrom(context.importFileName('./enums')),
```
jacek-prisma, line 32: I may have missed some conversation, but I thought we wanted them to be namespaced to avoid amibguities?

jkomyno: Namespacing enums is precisely what caused https://github.com/prisma/prisma/issues/27943

jacek-prisma: It seems the issue is caused by enums being exposed both via namespace and individual names, this doesn't really make sense to me, I'd expect them to only be exported via namespace

aqrln: I don't have a very strong opinion here and I think re-exporting enums in any way (either directly or as a namespace) isn't strictly necessary or useful because in the new client there's a separate `enums.ts` file you can just import directly.

I'd keep this for now though to make the migration easier from the old client since it does expose user's enums on top level.

aqrln (approved):

packages/client-generator-ts/src/TSClient/file-generators/ClientFile.ts
```diff
@@ -29,6 +29,7 @@ export function createClientFile(context: GenerateContext, options: TSClientOpti
 
   const exports = [
     ts.moduleExportFrom(context.importFileName('./enums')).asNamespace('$Enums'),
```
aqrln, line 31: Is this still necessary? If it's not used anywhere, and it doesn't mirror a pre-existing API in `prisma-client-js`, it can probably be gotten rid of.

jkomyno: It mirrors things from `prisma-client-js`, actually

jkomyno: So I'm not removing it

PR #28066, opened 2025-09-09
feat(client): remove node_modules caching check when generator has custom output

jkomyno:
This PR:
- skips the `node_modules` cache check when `generator { output = "<path>" }` is used
- solves [ORM-1429](https://linear.app/prisma-company/issue/ORM-1429/remove-node-modules-caching-check-when-generator-has-custom-output)
- Closes: https://linear.app/prisma-company/issue/ORM-1428/remove-vercel-cached-client-validation-with-custom-output-path

### DIFF
--- a/packages/client/src/runtime/core/init/checkPlatformCaching.ts
+++ b/packages/client/src/runtime/core/init/checkPlatformCaching.ts
@@ -15,20 +15,33 @@ const cachingPlatforms = {
   'Netlify CI': 'netlify',
 } as const
 
-type Config = Pick<GetPrismaClientConfig, 'postinstall' | 'ciName' | 'clientVersion'>
+type Config = Pick<GetPrismaClientConfig, 'postinstall' | 'ciName' | 'clientVersion' | 'generator'>
 
 /**
  * Throws an error if the client has been generated via auto-install and the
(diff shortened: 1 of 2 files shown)


packages/client/src/runtime/core/init/checkPlatformCaching.ts
```diff
@@ -15,21 +15,34 @@ const cachingPlatforms = {
...
-export function checkPlatformCaching({ postinstall, ciName, clientVersion }: Config) {
+export function checkPlatformCaching({ postinstall, ciName, clientVersion, generator }: Config) {
   debug('checkPlatformCaching:postinstall', postinstall)
   debug('checkPlatformCaching:ciName', ciName)
 
   // if client was not generated manually
   if (postinstall !== true) return
 
+  // check if a custom output directory is used
+  if (generator && generator.output) {
```
bricefriha: Perhaps it would be better to simplify this:
```suggestion
  if (generator?.output) {
```

unless I'm missing something

aqrln (dismissed):
LGTM modulo @bricefriha's suggestion above

aqrln (approved):

Issue #28089, opened 2025-09-12
`unref()` is not a function error when using `$transaction` with rust-free client on Cloudflare Workers


aiji42:
### Bug description

When using Prisma v6.16.1 with the rust-free client configuration (`engineType = "client"` and `runTime = "workerd"`) on Cloudflare Workers, executing queries with $transaction causes a TypeError because `unref()` is not available on the `setTimeout` return value in the Workers environment.

### Severity

⚠️ Major: Breaks core functionality (e.g., migrations fail)

### Reproduction

1. Update Prisma to v6.16.1
2. Configure schema.prisma with rust-free client for Cloudflare Workers:
```
  generator client {
    provider        = "prisma-client"
    engineType      = "client"
    runtime         = "workerd"
    previewFeatures = ["relationJoins"]
    output          = "./src/generated/prisma"
  }
```
3. Execute any query using $transaction
4. Error occurs: TypeError: i.unref is not a function

### Expected vs. Actual Behavior

`$transaction should` execute successfully without errors on Cloudflare Workers runtime.

### Frequency

Consistently reproducible

### Does this occur in development or production?

Both development and production

### Is this a regression?

Yes - This is a regression introduced in Prisma v6.16.0.

The issue was introduced by PR #28030 (fix(client-engine-runtime): unref transaction timers) which added `unref()` calls to transaction timers. This change works correctly in Node.js environments but breaks compatibility with edge runtimes like Cloudflare Workers.

Last known working version: Prisma versions prior to v6.16.0 should work without this issue, as they didn't include the `unref()` calls on transaction timers.

### Workaround

Manually remove the two `unref()` calls from `node_modules/@prisma/client/runtime/wasm-compiler-edge.mjs`. This fixes the issue but needs to be reapplied after each npm install.

### Prisma Schema & Queries

```prisma
generator client {
  provider        = "prisma-client"
  engineType      = "client"
  runtime         = "workerd"
  previewFeatures = ["relationJoins"]
  output          = "./src/generated/prisma"
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}
```

### Prisma Config

```ts
// Add your `prisma.config.ts`
```

### Logs & Debug Info

```
  [dev:main  ] ✘ [ERROR] TypeError: i.unref is not a function
  [dev:main  ]
  [dev:main  ]       at #s (file:///Users/....../node_modules/@prisma/client-engine-runtime/src/transaction-manager/transaction-manager.ts:216:15)
[...]

PR #28091, opened 2025-09-12
fix: handle missing timer unref

jacek-prisma:
timer.unref is not available in some environments like CF workers

[ORM-1431](https://linear.app/prisma-company/issue/ORM-1431/fix-transaction-manager-in-workerd-environment)

Fixes http://github.com/prisma/prisma/issues/28089


wmadden (approved):
Looks good 👍🏻

PR #28122, opened 2025-09-17
chore: CI - build - reusable workflow artifacts

dulmandakh:
This PR introduces build workflow that other workflows should depend to re-use build artifacts as cache to save time. Once this PR lands confirming that we can save time using build artifacts, I'll work on CI optimization utilizing turbo caching.


dulmandakh:
@aqrln please approve workflows

PR #28174, opened 2025-09-27
fix(cache): follow the XDG Base Directory specification

t-webber:


packages/fetch-engine/src/utils.ts
```diff
@@ -30,6 +30,9 @@ export async function getRootCacheDir(): Promise<string | null> {
       return null
     }
   }
+  if (process.env.XDG_CACHE_HOME) {
+    return path.join(process.env.XDG_CACHE_HOME, 'prisma')
+  }
   return path.join(os.homedir(), '.cache/prisma')
```
bricefriha: Great stuff!

just one thing: unless we are planning on adding more scenarios, I feel like it may as well be a one liner:

```suggestion
return process.env.XDG_CACHE_HOME ?
              path.join(process.env.XDG_CACHE_HOME, 'prisma') :
              path.join(os.homedir(), '.cache/prisma')
```

packages/fetch-engine/src/utils.ts
```diff
@@ -30,7 +30,9 @@ export async function getRootCacheDir(): Promise<string | null> {
       return null
     }
   }
-  return path.join(os.homedir(), '.cache/prisma')
+return process.env.XDG_CACHE_HOME ?
+              path.join(process.env.XDG_CACHE_HOME, 'prisma') :
+              path.join(os.homedir(), '.cache/prisma')
```
aqrln: Formatting is off here, please format using Prettier (see also the lint job failing on CI)

aqrln (commented):
The change itself looks good to me, just one stylistic issue

packages/fetch-engine/src/utils.ts
bricefriha: The mistake is on me sorry, I suggested the code

t-webber:
I think the CI error is unrelated to my change

aqrln (approved):
Thank you!

PR #28177, opened 2025-09-29
fix(adapter-mariadb): error event listener leak

Tiaansu:
This PR fixes the same event listener leak for `adapter-mariadb` that was resolved in `adapter-pg` (#28057).


packages/adapter-mariadb/src/mariadb.ts
```diff
@@ -151,7 +159,7 @@ export class PrismaMariaDbAdapter extends MariaDbQueryable<mariadb.Pool> impleme
       return tx
     } catch (error) {
       await conn.end()
```
jacek-prisma, line 161: should probably invoke `cleanup` here, because neither `commit` or `rollback` would be reached in case of an exception at this point

jacek-prisma (approved):

PR #28178, opened 2025-09-29
feat: extensions schema entrypoint

jacek-prisma:
[TML-1411](https://linear.app/prisma-company/issue/TML-1411/expose-an-entrypoint-for-configuring-the-extensions)


packages/config/src/PrismaConfig.ts
```diff
@@ -234,6 +234,7 @@ const PrismaConfigShape = Shape.Struct({
   enums: Shape.optional(EnumsConfigShape),
   views: Shape.optional(ViewsConfigShape),
   typedSql: Shape.optional(TypedSqlConfigShape),
+  extensions: Shape.optional(Shape.Any),
```
aqrln, line 242: Should we perhaps still have a validation schema defined for this even if it will change in the future?

Also, we probably need to update `ExperimentalConfig` and `ExperimentalConfigShape` to include `extensions`, and throw an error if `extensions` are used without `{ experimental: { extensions: true } }`.

packages/integration-tests/src/__tests__/integration/postgresql/__snapshots__/introspection.test.ts.snap
```diff
@@ -998,7 +998,7 @@ model b {
   id  Int @id @default(autoincrement())
   one Int
   two Int
-  a   a   @relation(fields: [one, two], references: [one, two], onDelete: NoAction, onUpdate: NoAction, map: "b_one_fkey")
+  a   a   @relation(fields: [one, two], references: [one, two], onDelete: NoAction, onUpdate: NoAction)
```
aqrln, line 1001: The change looks unrelated

packages/migrate/src/commands/DbExecute.ts
```diff
@@ -199,7 +199,7 @@ See \`${green(getCommandWithExecutor('prisma db execute -h'))}\``,
     }
 
     const adapter = await config.adapter?.()
-    const migrate = await Migrate.setup({ adapter })
+    const migrate = await Migrate.setup({ adapter, extensions: config['extensions'] })
```
aqrln, line 202:
```suggestion
    const migrate = await Migrate.setup({ adapter, extensions: config.extensions })
```

packages/integration-tests/src/__tests__/integration/postgresql/__snapshots__/introspection.test.ts.snap
jacek-prisma: it's because I bumped postgres from 10 to 12 in the docker compose file, I think introspection doesn't pick up the map field in newer postgres versions

I had to bump it because postgres 10 image is so old that its debian repos have been archived so I can't it install extensions on it

packages/config/src/PrismaConfig.ts
jacek-prisma: I think it's not worth validating it yet since it'll be changing a lot, I agree about making it experimental though

packages/migrate/src/commands/DbExecute.ts
jacek-prisma: I've omitted the extension object from `_PrismaConfigInternal` because its definition is coupled with effect validation types, so I can't define it without also defining the validations
So it needs to either be omitted there or have a validation added

packages/integration-tests/src/__tests__/integration/postgresql/__snapshots__/introspection.test.ts.snap
jacek-prisma: will create a ticket to keep track of this

packages/config/src/PrismaConfig.ts
jacek-prisma: will create a ticket to add validation once we have a shape for this

aqrln (dismissed):

packages/config/src/PrismaConfig.ts
jacek-prisma: https://linear.app/prisma-company/issue/TML-1484/implement-validation-for-the-extensions-config

packages/integration-tests/src/__tests__/integration/postgresql/__snapshots__/introspection.test.ts.snap
jacek-prisma: https://linear.app/prisma-company/issue/TML-1485/investigate-changed-introspection-behavior-in-postgres-12

aqrln (dismissed):

aqrln (approved):

PR #28195, opened 2025-09-30
fix: fix randomly failing tests

jacek-prisma:
[TML-1479](https://linear.app/prisma-company/issue/TML-1479/fix-randomly-failing-e2e-tests)

The tests seem to pass consistently now, run them 3 times.

### DIFF
--- a/packages/client/tests/e2e/_utils/retry.ts
+++ b/packages/client/tests/e2e/_utils/retry.ts
@@ -1,0 +1,13 @@
+import timers from 'node:timers/promises'
+
+export async function retry<A>(fn: () => Promise<A>, max: number, delay: number = 1000): Promise<A> {
+  for (let i = 0; i < max; i++) {
+    try {
+      return await fn()
+    } catch (e) {
+      if (i === max - 1) throw e
+    }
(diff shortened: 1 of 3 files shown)


packages/client/tests/e2e/_utils/retry.ts
```diff
@@ -0,0 +1,13 @@
+import timers from 'node:timers/promises'
+
+export async function retry<A>(fn: () => Promise<A>, max: number, delay: number = 1000): Promise<A> {
+  for (let i = 0; i < max; i++) {
+    try {
+      return await fn()
+    } catch (e) {
+      if (i === max - 1) throw e
+    }
+    await timers.setTimeout(delay)
```
aqrln, line 10: should we add exponential backoff or should we not overcomplicate if it works as is?

aqrln (approved):

packages/client/tests/e2e/_utils/retry.ts
jacek-prisma: I think exponential backoff isnt needed here, we don't expect something to go down for a prolonged period of time during the test, it's just random race conditions

PR #28200, opened 2025-10-01
chore: run QPE tests on SQL Server

aqrln:
Closes: https://linear.app/prisma-company/issue/TML-1418/run-tests-with-sql-server


.github/workflows/test-template.yml
aqrln, line 447: mssql was already present here

PR #28202, opened 2025-10-01
feat: TML-1074 extract client utilities to @prisma/client-runtime-utils package

FGoessler:
This PR introduces a new `@prisma/client-runtime-utils` package. This package contains various runtime helpers. These helpers were previously only exposed as part of the generated output. Making it hard to use them without a generated client or across different build artifacts like `browser.ts` and `client.ts` in the `prisma-client` generator output.

Now these utilities exist in a single point of truth and can be referenced either through direct imports from `@prisma/client-runtime-utils` or through their reexports in the generated client code.

This is especially relevant for classes and singleton instances like `Prisma.DBNull`.

Overall the structure of imports with this change becomes:

```mermaid
graph TD;
  uc[User Code]
  gcc[Generated Code client.ts]
  gcb[Generated Code browser.ts]
  pcri["@prisma/client/runtime/index-browser"]
  pcrc["@prisma/client/runtime/client"]
  pcru["@prisma/client-runtime-utils"]

  uc --> gcc
  uc --> gcb
  uc --> pcru
  gcc --> pcrc
  gcb --> pcri
  pcri --> pcru
  pcrc --> pcru
```

The following utilities are exposed through this change:

- Errors (`PrismaClientInitializationError`, `PrismaClientKnownRequestError`, `PrismaClientRustError`, `PrismaClientRustPanicError`, `PrismaClientUnknownRequestError`, `PrismaClientValidationError`)
- `DbNull`, `JsonNull`, `AnyNull`
- `Decimal`
- `sql-template-tag` helpers

For better dealing with `DbNull`, `JsonNull` & `AnyNull` this change also introduces new `isDbNull`, `isJsonNull` & `isAnyNull` helper functions.

## This is a breaking change! => Prisma 7 change.

> [!WARNING]
>
> The generated output now depends on the new `@prisma/client-runtime-utils` package which needs to be installed. 
> 
> For the new `prisma-client` generator this is not a problem as all dependencies in the generated code only refer to `@prisma/client` exports and `@prisma/client` pulls in `@prisma/client-runtime-utils` as transitive dependency.
> 
> For the old `prisma-client-js` generator when generating into `node_modules` this is also no issue.
> 
> But when the old `prisma-client-js` generator is used with a custom output path it copies over runtime dependencies from the original `@prisma/client` package but does not depend on that package anymore. hence the transitive dependency gets lost and code that lives directly in the generated output tries to import the  `@prisma/client-runtime-utils` package. To solve this the place that it is generated into needs to depend on the new `@prisma/client-runtime-utils` package itself.
> 
> The generated code for a custom output path also contains a `package.json`. It now contains a dependency field specifying this new dependency. If users use this `package.json` depending on their use-case they need to ensure `npm install` is executed on it. If user don't use this `package.json` but just import the code directly their surrounding package needs to declare the dependency on the matching `@prisma/client-runtime-utils` package in its `package.json`.

We have to go with a dedicated package as creating only a new entry point into the `@prisma/client` package proved to be not feasible due to complications with the generated `.d.ts` files. Their generation process cannot account for external files and hence cause type definition clashes.

---

/integration


packages/client-runtime-utils/src/errors/log.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+    if (log.fields?.reason) {
+      str += `\n${log.fields?.reason}`
+    }
+    return str
+  }
+
+  return 'Unknown error'
+}
```
jkomyno, line 45: Don't we have any existing utility for this?

packages/client-runtime-utils/src/nullTypes.ts
```diff
@@ -37,56 +37,74 @@ export abstract class ObjectEnumValue {
...
+ * Check if a value is the DBNull singleton instance.
  */
-function setClassName(classObject: Function, name: string) {
-  Object.defineProperty(classObject, 'name', {
-    value: name,
-    configurable: true,
-  })
+export function isDbNull(value: unknown): value is DbNullClass {
+  return value === DbNull
+}
```
jkomyno, line 96: Nice, we finally have public `is*` predicate helpers!

packages/client-runtime-utils/src/errors/log.ts
jkomyno: Edit: this was copied token-for-token from `packages/client/src/runtime/core/engines/common/utils/log.ts`

PR #28238, opened 2025-10-07
Create generator-generic-ossf-slsa3-publish.yml

psanogo:


.github/workflows/generator-generic-ossf-slsa3-publish.yml
```diff
@@ -0,0 +1,66 @@
...
+      # ========================================================
+      #
+      # Step 1: Build your artifacts.
+      #
+      # ========================================================
+      - name: Build artifacts
+        run: |
+            # These are some amazing artifacts.
+            echo "artifact1" > artifact1
+            echo "artifact2" > artifact2
```
aqrln, line 37: ...

aqrln (changes requested):
Our releases already have provenance. You can check the attestation on npm or on sigstore.dev, e.g.: https://search.sigstore.dev/?logIndex=589094569

What you probably attempted to do would've made sense if we published the build artifacts attached to GitHub releases, but we don't, we only use GitHub releases for release notes.

However, I can't tell for sure if my assumption about what the idea behind the PR was is indeed correct because the PR is not only completely lacking a description, but also lacking an implementation: the created workflow file is just a template (probably taken from docs or generated with an LLM) that does nothing and isn't integrated with the build system in any way.

Given the extremely low effort put into the PR, I'm going to go ahead and close it right away.

PR #28253, opened 2025-10-09
feat(qpe): add an option to disable logging to console

aqrln:
Necessary for `prisma dev`: https://github.com/prisma/team-expansion/pull/1903/files#r2322689633

Part of https://linear.app/prisma-company/issue/TML-1317/add-qc-support-in-local-ppg-with-accelerate-api


packages/query-plan-executor/src/log/sink.test.ts
```diff
@@ -20,6 +20,15 @@ describe('CapturingSink', () => {
   })
 })
 
+describe('DroppingSink', () => {
+  it('drops events without side effects', () => {
+    const sink = new DroppingSink()
+    const event = new LogEvent('info', 'ignored message')
+
+    expect(() => sink.write(event)).not.toThrow()
```
jacek-prisma: this test doesn't really check for side effects

aqrln: I think this test is rather useless. Checking for side effects doesn't make sense here, that's what the tests for `createConsoleLogger` do. I wanted to remove it, idk why I left it.

jacek-prisma: yeah, it makes sense to remove it

aqrln: done

jacek-prisma (dismissed):

jacek-prisma (approved):

Issue #9187, opened 2021-09-09
Add support for MultiSubnetFailover


andrewl913:
## Problem
When using the new SQL server DB connector.  We would like to connect to the always-on availability group.

## Suggested solution

Add support as a connection string property for the SQL server to include`multiSubnetFailover=true`

## Additional context
https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/sqlclient-support-for-high-availability-disaster-recovery

Thanks for all your hard work and support!

kihonq:
Would love to see this option available too ! Thanks !

salam-seek:
Facing the same issue as well, is there any workaround anyone can suggest ?

finn-orsini:
👋 Hoping to get some eyes on this, the lack of this config is impacting our apps ability to connect reliably to our DB. 

Is it as easy as updating [the options here](https://github.com/prisma/prisma/blob/main/packages/client/src/utils/setupMSSQL.ts#L21-L24) to accept this value from the connection string?

dariusj18:
Bump, seems a pretty important enterprise feature to be lacking.

pip8786:
@andrewl913 or others on this thread, what do you see when you connect prisma to a server with multi subnet failover turned on? We're getting timeouts getting a connection seemingly randomly since we switched to a server with this on and are trying to determine if this is the issue.

dariusj18:
@pip8786 yes, timeouts are the issue.  without multisubnetfailover the driver (given multiple results in a DNS query) will attempt to connect to the first one identified.  This will timeout if the first server it is trying is not accessible.  The multisubnetfailover flag tells the driver to attempt to connect to each server in parallel and go with whichever it succeeds connecting to first.

pip8786:
@dariusj18 Thanks for clarifying. I've added the issue to the driver repo in case that gains more traction. Looks like you saw it too. Hopefully they can fix it so we don't have to move off of prisma.

hirdstclendenin:
Now that Prisma has a client adapter for mssql using tedious, this should be as easy as allowing the parameter MultiSubnetFailover through from the adapter connection string. (I think this is working with the config passed as options)

Feature supported via connection options 

https://github.com/tediousjs/tedious/blob/ebb023ed90969a7ec0e4b036533ad52739d921f7/src/connection.ts#L756

[...]

PR #28261, opened 2025-10-09
feat(adapter-mssql): support multiSubnetFailover flag in connection string

tylerclendenin:
MultiSubnetFailover is a SQL Server connection property that enables faster failovers for Always On Availability Groups and Failover Cluster Instances, whether across single or multiple subnets. When enabled, clients connecting to a SQL Server listener will attempt to connect to all its IP addresses simultaneously, choosing the first one that responds, which significantly reduces reconnection time after a failover event. 

The Tedious package [supports it as an option](https://github.com/tediousjs/tedious/blob/ebb023ed90969a7ec0e4b036533ad52739d921f7/src/connection.ts#L756), however to use it from a connection string it needs to be extracted.

Ref https://github.com/prisma/prisma/issues/9187

### DIFF
--- a/packages/adapter-mssql/src/connection-string.test.ts
+++ b/packages/adapter-mssql/src/connection-string.test.ts
@@ -100,16 +100,12 @@ describe('parseConnectionString', () => {
     })
   })
 
-  it('should parse multiSubnetFailover parameter correctly', () => {
-    const testCases = [
-      { input: 'sqlserver://localhost;database=testdb;multiSubnetFailover=true', expected: true },
-      { input: 'sqlserver://localhost;database=testdb;multiSubnetFailover=false', expected: false },
-    ]
-
-    testCases.forEach(({ input, expected }) => {
-      const config = parseConnectionString(input)
-      expect(config.options?.multiSubnetFailover).toBe(expected)
-    })
+  it.each([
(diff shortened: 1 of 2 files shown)


packages/adapter-mssql/src/connection-string.test.ts
```diff
@@ -100,6 +100,18 @@ describe('parseConnectionString', () => {
...
+    const testCases = [
+      { input: 'sqlserver://localhost;database=testdb;multiSubnetFailover=true', expected: true },
+      { input: 'sqlserver://localhost;database=testdb;multiSubnetFailover=false', expected: false },
+    ]
+
+    testCases.forEach(({ input, expected }) => {
+      const config = parseConnectionString(input)
+      expect(config.options?.multiSubnetFailover).toBe(expected)
+    })
+  })
```
jacek-prisma: could use `it.each`
```suggestion
  it.each([
    { input: 'sqlserver://localhost;database=testdb;multiSubnetFailover=true', expected: true },
    { input: 'sqlserver://localhost;database=testdb;multiSubnetFailover=false', expected: false },
  ])('should parse multiSubnetFailover parameter correctly for %o', ({ input, expected }) => {
    const config = parseConnectionString(input)
    expect(config.options?.multiSubnetFailover).toBe(expected)
  })
```

jacek-prisma (approved):
LGTM

PR #28262, opened 2025-10-09
feat(adapter-mssql): make migration aware for use in prisma config adapter

tylerclendenin:
Update the MSSQL adapter to be compatible with the prisma config file's defineConfig adapter option by implementing `SqlMigrationAwareDriverAdapterFactory`.

I added the `connectToShadowDb` method and updated the `PrismaMssqlAdapter` class to accept a release function which is called during `dispose`.

I used the adapter-pg as a template and adapted it.

Ref https://github.com/prisma/prisma/issues/9187

### DIFF
--- a/packages/adapter-mssql/src/mssql.ts
+++ b/packages/adapter-mssql/src/mssql.ts
@@ -4,7 +4,7 @@ import {
   DriverAdapterError,
   IsolationLevel,
   SqlDriverAdapter,
-  SqlDriverAdapterFactory,
+  SqlMigrationAwareDriverAdapterFactory,
   SqlQuery,
   SqlQueryable,
   SqlResultSet,
@@ -119,6 +119,7 @@ class PrismaMssqlAdapter extends MssqlQueryable implements SqlDriverAdapter {
   constructor(
     private pool: sql.ConnectionPool,
     private readonly options?: PrismaMssqlOptions,
+    private readonly release?: () => Promise<void>,
   ) {
     super(pool)
   }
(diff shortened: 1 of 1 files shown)


jacek-prisma:
Hi @tylerclendenin,
Making the adapter conform to the interface is not enough to enable its usage in the CLI - we also need to support it in our schema engine and this work is incomplete. I think it's better to not implement the interface for the time being until the MSSQL adapter is actually fully supported. Right now users attempting to use it in the CLI would see panics coming from the schema engine.

tylerclendenin:
> Hi @tylerclendenin, Making the adapter conform to the interface is not enough to enable its usage in the CLI - we also need to support it in our schema engine and this work is incomplete. I think it's better to not implement the interface for the time being until the MSSQL adapter is actually fully supported. Right now users attempting to use it in the CLI would see panics coming from the schema engine.

So it's not just the MSSQL that is not supported, but none of the adapters will currently work because the engine only works with the Rust engine currently?

tylerclendenin:
@jacek-prisma depending on how long the integration for adapters takes for the prisma engine, this PR https://github.com/prisma/tiberius/pull/378 for Tiberius should give parity for the MultiSubnetFailover feature.  Do you know who might be able to review and integrate it? There does not seem to be much attention paid to it.

Or do you know a simple way I could add my own step to a pipeline to build my own Tiberius to be used by my own Prisma deployment?

tensordreams:
Hi @tylerclendenin — thanks for this, and apologies for the slow reply.

We are closing this one: as noted in review, conforming to the migration-aware interface is not sufficient without schema-engine support, and that work is incomplete.

PR #28266, opened 2025-10-09
feat(config): add support for `engine?: 'js' | 'classic'`

jkomyno:
This PR:
- closes [TML-1332](https://linear.app/prisma-company/issue/TML-1332/prisma-config-alternative-to-url-directurl-shadowdatabaseurl)
- this PR depends on https://github.com/prisma/prisma-engines/pull/5645.


packages/cli/jest.setup.js
```diff
@@ -0,0 +1,19 @@
...
+   */
+  process.env['JITI_ALIAS'] = JSON.stringify({
+    'src/': path.join(__dirname, 'src'),
+    'test-utils/': path.join(__dirname, 'src', '__tests__', '_utils'),
+  })
+})
+
+global.afterEach(() => {
+  delete process.env['JITI_ALIAS']
+})
```
jkomyno, line 19: Needed due to the changes in `Version.test.ts`

packages/config/src/PrismaConfig.ts
```diff
@@ -345,7 +431,7 @@ function validateExperimentalFeatures(config: PrismaConfig): Either.Either<Prism
  * Parse a given input object to ensure it conforms to the `PrismaConfig` type Shape.
  * This function may fail, but it will never throw.
  */
-function parsePrismaConfigShape(input: unknown): Either.Either<PrismaConfig, Error> {
+export function parsePrismaConfigShape(input: unknown): Either.Either<PrismaConfig, Error> {
```
jkomyno:
```suggestion
function parsePrismaConfigShape(input: unknown): Either.Either<PrismaConfig, Error> {
```

packages/internals/src/cli/schemaContext.ts
```diff
@@ -40,13 +43,16 @@ export type SchemaContext = {
...
+  /**
+   * @deprecated Only used during the refactoring for backwards compatibility. Use `schemaFiles` instead or determine needed file paths otherwise.
+   */
   schemaPath: string
 }
 
 type LoadSchemaContextOptions = {
   schemaPathFromArg?: string
   schemaPathFromConfig?: string
+  schemaEngineConfig?: SchemaEngineConfigInternal & { loadedFromFile: string | null }
```
jkomyno: TODO: check if we still need `loadedFromFile` here.

packages/internals/src/cli/schemaContext.ts
```diff
@@ -18,6 +20,7 @@ export type SchemaContext = {
   /**
    * The directory of the schema.prisma file that contains the datasource block.
    * Some relative paths like SQLite paths or SSL file paths are resolved relative to it.
+   * TODO: consider whether relative paths should be resolved relative to `prisma.config.ts` instead.
```
jkomyno:
```suggestion
   * TODO(prisma7): consider whether relative paths should be resolved relative to `prisma.config.ts` instead.
```

packages/config/src/__tests__/loadConfigFromFile.test.ts
```diff
@@ -33,14 +61,51 @@ describe('loadConfigFromFile', () => {
...
+
+        const expectedAdapter = mockMigrationAwareAdapterFactory('postgres')
+        assertConfigWithEngineJs(config)
+        expect(config.adapter).toStrictEqual(expect.any(Function))
+
+        const { adapter: adapterFactory } = config
+        expect(adapterFactory).toBeDefined()
+
+        const adapter = await adapterFactory()
+        expect(JSON.stringify(adapter)).toEqual(JSON.stringify(bindMigrationAwareSqlAdapterFactory(expectedAdapter)))
```
jacek-prisma: it might be cleaner to assert that some method was called on the mock (and maybe call something from the test) rather then comparing JSON serialized objects

packages/migrate/src/Migrate.ts
```diff
@@ -58,18 +58,20 @@ export class Migrate {
     this.shadowDbInitScript = shadowDbInitScript ?? ''
   }
 
-  static async setup({ adapter, schemaContext, ...rest }: MigrateSetupInput): Promise<Migrate> {
-    const engine = await (async () => {
-      if (adapter) {
+  static async setup({ schemaContext, schemaEngineConfig, ...rest }: MigrateSetupInput): Promise<Migrate> {
+    const schemaEngine = await (async () => {
+      if (schemaEngineConfig?.engine === 'js') {
+        const adapter = await schemaEngineConfig.adapter()
```
jacek-prisma, line 64: not necessarily a problem but keep in mind that we're now evaluting the `adapter` function multiple times and if it has any side effects, they're going to happen multiple times

jkomyno: Yes, I'm aware. I originally wanted to evaluate the adapter only once, when loading the config, and throw a hard error if its initialization failed. Florian talked me out of it, but perhaps that was the right design decision to make all along. I'm keeping that in mind, but I'm not refactoring that now.

packages/config/src/__tests__/loadConfigFromFile.test.ts
jkomyno: Calling a method would result in yet another JSON-serialized object comparison, so it would be the same with extra steps.

I can however compare the two values by ensuring they have the same symbol, which I just committed.

jacek-prisma: sounds good, I don't think it's been commited though

jkomyno: https://github.com/prisma/prisma/pull/28266/commits/62080a3b9fdb99a026505eed9998a5ffec04e766

jacek-prisma: I don't see it being used in the tests

jkomyno: [`bindMigrationAwareSqlAdapterFactory`](https://github.com/prisma/prisma/blob/62080a3b9fdb99a026505eed9998a5ffec04e766/packages/driver-adapter-utils/src/mock.ts#L50) is used [here](https://github.com/prisma/prisma/blob/93f557c368bc771322bbed8fc0c0ba060bc6cb30/packages/config/src/__tests__/loadConfigFromFile.test.ts#L82)

packages/config/src/__tests__/loadConfigFromFile.test.ts
```diff
@@ -33,14 +61,60 @@ describe('loadConfigFromFile', () => {
...
+        Object.entries(expectedAdapter).forEach(([key, value]) => {
+          if (typeof value === 'function') {
+            // avoid `Compared values have no visual difference` error on functions
+            expect(adapter[key]).toBeInstanceOf(Function)
+            expect(JSON.stringify(adapter[key])).toEqual(JSON.stringify(value))
+          } else {
+            expect(adapter[key]).toEqual(value)
+          }
+        })
+      })
```
jacek-prisma, line 100: I think this comparison isn't very useful because the adapter gets transformed via `bindMigrationAwareSqlAdapterFactory` and most of its properites are functions that get overwritten

For the purposes of this test I think it's good enough to just check if the `adapterName` matches what we expect instead of checking all properties

This test aside, I am a bit surprised that the error binding is happening when we are creating the config, the error handling behavior is related to the schema engine and IMO shouldn't be used in the config code

jkomyno: Please consider that the whole reason I spent time changing the comparison above is because of [this comment](https://github.com/prisma/prisma/pull/28266#discussion_r2426599599).
Notice how this test pattern was already consolidated 7 months ago (https://github.com/prisma/prisma/blob/e2e3522f4d4c4bb3e3833720620eb41e9557b97b/packages/config/src/__tests__/defineConfig.test.ts#L108), and isn't the scope of this PR.

---

> For the purposes of this test I think it's good enough to just check if the adapterName matches what we expect

Comparing a single field doesn't cut it, as we want to ensure the two sides of the comparison have actually the same shape.

> I am a bit surprised that the error binding is happening when we are creating the config

The users passes a `SqlMigrationAwareDriverAdapterFactory` value to their user-facing PrismaConfig, which is converted into a `ErrorCapturingSqlMigrationAwareDriverAdapterFactory` when saving the value in the internal PrismaConfig. This, again, was something established months ago. I don't recall all the details, but I suppose this was done to centralize the binding in a single place, rather than repeating the same operation in all `Migrate*` and `Db*` command classes in `@prisma/migrate`. We can have a separate conversation about this, but I don't see how this is blocking this PR.

jacek-prisma: That's fair enough, but keep in mind that this doesn't check the whole shape of the object, for example it doesn't confirm that the factory connect returns the right object and so on
Ideally we could inject some kind of mock into the loaded config and we could just verify that the mock was created from the bind function, but it might be tricky to do that here

jacek-prisma: I think for the time being we can move on and merge it as is to avoid being stuck on a single test

jacek-prisma (approved):

packages/config/src/__tests__/loadConfigFromFile.test.ts
jkomyno: Let's talk about it when you're back 💪🏻

clemens:
I've been wondering about these changes: I would expect that when I'm adding `engine` and `url` (and `shadowDatabaseUrl`) to `prisma.config.ts`, then I could remove their counterparts from `schema.prisma`. But that doesn't seem to be the case. Is this intentional or an oversight? Does preparing for Prisma v7 mean (for now) that we have to keep these values duplicated in both files?

PR #28271, opened 2025-10-10
fix: revert "chore: use pnpm catalog (#28236)"

jacek-prisma:
This reverts commit 10fab6b8ffb71d53e7af8c44c03df91df9a39aaa.

This reverts the inclusion and usages of pnpm catalog. The reason for reverting the change is that it breaks the prisma-engines test pipeline (see https://github.com/prisma/prisma-engines/actions/runs/18382915995/job/52373895545) and we have no simple fix for it.


aqrln (approved):

aqrln:
fyi @dulmandakh — we're gonna give it another try after we merge both repos together (which is going to happen in the foreseeable future but not immediately soon, sometime after Prisma 7)

PR #28283, opened 2025-10-13
chore: bump minimum Node version to 20.19

jacek-prisma:
[TML-1337](https://linear.app/prisma-company/issue/TML-1337/prisma-7-update-minimum-nodejs-versions)

Updates Node constraints to be:
- `^20.19 || ^22.12 || ^24.0` - for the engines in package.json (should allow using a newer minor versions)
- `~20.19.0` - for the typings (we should NOT rely on minor added types)
- `['20.19.0', '22.12.0', '24.0.1']` - for CI (we want to test our behavior on our lowest supported versions)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -39,7 +39,7 @@
   "bugs": "https://github.com/prisma/prisma/issues",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=18.18"
+    "node": ">=20.19"
   },
   "prisma": {
     "prismaCommit": "placeholder-for-commit-hash-replaced-during-publishing-in-publish-ts"
@@ -126,7 +126,7 @@
     "@swc/jest": "0.2.37",
     "@types/fs-extra": "11.0.4",
     "@types/jest": "29.5.14",
-    "@types/node": "18.19.76",
+    "@types/node": "20.19.0",
     "async-listen": "3.1.0",
     "checkpoint-client": "1.1.33",
     "chokidar": "4.0.3",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -39,7 +39,7 @@
   "bugs": "https://github.com/prisma/prisma/issues",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=20.19"
+    "node": "^20.19 || ^22.12 || ^24.0"
   },
   "prisma": {
     "prismaCommit": "placeholder-for-commit-hash-replaced-during-publishing-in-publish-ts"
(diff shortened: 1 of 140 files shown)


packages/cli/package.json
```diff
@@ -126,7 +126,7 @@
     "@swc/jest": "0.2.37",
     "@types/fs-extra": "11.0.4",
     "@types/jest": "29.5.14",
-    "@types/node": "18.19.76",
+    "@types/node": "20.19.0",
```
aqrln: There's `20.19.21` available

jkomyno:
> Bumps the min Node version to `20.19` everywhere.

The Linear ticket states:
> In all of our CIs, we should use:
> [Node.js 20.19.0](https://nodejs.org/en/blog/release/v20.19.0)
> [Node.js 22.12.0](https://nodejs.org/en/blog/release/v22.12.0)
> [Node.js 24.0.0](https://nodejs.org/en/blog/release/v24.0.0)

Basically, as described [in our docs](https://www.prisma.io/docs/orm/reference/system-requirements#software-requirements), we support 3 release lines of Node.js at the same time. This PR does a good job at bumping the minimum required Node.js version to 20.19.0, but what about 22.12.0 ad 24.0.0 (the other two lines)?

Should we perhaps update the `node: [20, 22, 24]` yaml to `node: ['20.19.0', '22.12.0', '24.0.0']`?

jacek-prisma:
@jkomyno I wasn't sure if the action supports specifying precise versions, but it looks like it does, I can update it

https://github.com/actions/setup-node?tab=readme-ov-file#supported-version-syntax

packages/cli/package.json
jacek-prisma: updated to ~20.19.0

packages/cli/scripts/preinstall.ts
```diff
@@ -23,17 +23,24 @@ function extractSemanticVersionParts(version: MajorMinor | MajorMinorPatch) {
...
+    '20': '19',
+    '22': '12',
+    '24': '0',
+  }
 
   if (
-    nodeMajorVersion < MIN_NODE_MAJOR_VERSION ||
-    (nodeMajorVersion === MIN_NODE_MAJOR_VERSION && nodeMinorVersion < MIN_NODE_MINOR_VERSION)
+    !(nodeMajorVersion in MIN_NODE_VERSION_MATRIX) ||
+    nodeMinorVersion < parseInt(MIN_NODE_VERSION_MATRIX[nodeMajorVersion], 10)
```
aqrln: > `parseInt`

why not store numbers in the map to begin with?

jacek-prisma: makes sense, updated

aqrln (approved):
lgtm

Issue #19519, opened 2023-05-30
importing prisma client pollutes `process.env` without using dotenv


jacobakkerboom:
### Bug description

If you put MY_TEST_VAR in a .env file and import a prisma client, then `process.env.MY_TEST_VAR` will be defined, despite the fact that we are not using e.g. dotenv.

### How to reproduce

prisma/schema.prisma:
```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mysql"
  url      = env("DATABASE_URL_MAIN")
}

model Locations {
  LocationID                   Int                            @id
  LocationName                 String                         @db.VarChar(45)
}
```

package.json:
```json
{
  "type": "module",
  "scripts": {
    "start": "node app.js"
  },
  "dependencies": {
    "@prisma/client": "^4.14.1",
    "prisma": "^4.14.1"
  }
}
```

.env:
```bash
MY_TEST_VAR="THIS IS THE VALUE OF MY_TEST_VAR"
```

app.js:
```javascript
import { PrismaClient } from '@prisma/client'
console.log(process.env.MY_TEST_VAR);
```

`npx prisma generate`
`npm start`
(logs the variable, not undefined)

### Expected behavior

The expected behaviour is that undefined is logged

So what is said [here](https://www.prisma.io/docs/guides/development-environment/environment-variables#using-environment-variables-in-your-code) is not a bit misleading, because the variables will be loaded anyway

"If you want environment variables to be evaluated at runtime, you need to load them manually in your application code (for example, by using [dotenv](https://github.com/motdotla/dotenv))"

### Prisma information

(same as above)
prisma/schema.prisma:
```
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "mysql"
  url      = env("DATABASE_URL_MAIN")
}

model Locations {
  LocationID                   Int                            @id
  LocationName                 String                         @db.VarChar(45)
}
```

### Environment & setup

Mac and linux

### Prisma Version

```
prisma                  : 4.15.0
@prisma/client          : 4.15.0
Current platform        : darwin-arm64
Query Engine (Node-API) : libquery-engine 8fbc245156db7124f997f4cecdd8d1219e360944 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli 8fbc245156db7124f997f4cecdd8d1219e360944 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
[...]

PR #28298, opened 2025-10-14
feat!: remove env loading and dotenv support

aqrln:
Since Prisma Client won't be aware of connection strings anymore (except when using Accelerate), this logic is mostly obsolete.

We wanted to remove loading `dotenv` regardless, and have people load their environment variables themselves, as it shouldn't be Prisma's responsibility and Prisma shouldn't be mutating `process.env`. However, this PR takes it one step further, and disables evaluating and validating the value of `env()` for the `url` property in PSL altogether (except in `LibraryEngine`, where the environment variables are still passed to the engine).

Ref: https://github.com/prisma/prisma-engines/pull/5681
Ref: https://linear.app/prisma-company/issue/TML-1459/remove-processenv-loading-from-prismaclient
Closes: https://linear.app/prisma-company/issue/TML-309/fix-leaky-env-logic-→-stop-mutating-processenv
Closes: https://github.com/prisma/prisma/issues/19519


packages/client/tests/functional/_utils/config-builder.ts
```diff
@@ -0,0 +1,41 @@
...
+    ...defaultTestConfig(),
+    engine: 'classic',
+    schema: schemaPath,
+    datasource: {
+      url: absolutizeSqliteUrl({ schemaPath, databaseUrl }),
+    },
+  }
+}
+
+const SQLITE_PROTOCOL = 'file:'
```
jacek-prisma:
```suggestion
const FILE_PROTOCOL = 'file:'
```
right now it'll only happen for sqlite, but there's nothing sqlite specific about this code, so I'd just call it `FILE_PROTOCOL`

jacek-prisma (approved):
lgtm with one very minor comment

packages/client/tests/functional/_utils/config-builder.ts
aqrln: makes sense, updated

Issue #13987, opened 2022-06-23
Possibility to generate types with `undefined` instead of `null`


alielkhateeb:
## Problem

It would be great if we can replace `null` properties with `undefined` and treat it the same way we treat `null` when it comes to database queries, for example:

If we have this model
```
model Booking {
  resourceId           String                  @id @map("resource_id") @db.Uuid
  requestedDate    DateTime            @map("requested_delivery_date") @db.Date
  confirmedDate    DateTime?          @map("confirmed_delivery_date") @db.Date
}
```

The prisma client generated will be:
```
export type Booking = {
  resourceId: string
  requestedDate: Date
  confirmedDate: Date | null
}
```

That makes it not so reusable because code like this will not work as the property `confirmedDate` is typed as `Date |  null` which is not equivalent to "optional property", for example this block of code will complain because `confirmedDate` is missing.
```
import { Booking } from '@prisma/client';

const booking: Booking = {
    resourceId: '1',
    requestedDate: new Date()
}
```

## Suggested solution

I understand that I could create a mapped type based on the prisma type (to replace all `null` properties to optional ones) 
But it would be really great if we can configure this in prisma with a flag (something like `useUndefined`) which internally will provide us (export) this mapped type, but keeping the logic internally the as-is (with the `null` type), that way it will be backward compatible and also there won't be any changes in typescript safety for queries (like the value of where clause and so on), but at least it's a single flag and we get real "optional" typescript properties.

jhwheeler:
This would be a huge benefit to our team as well, because otherwise there are lots of transformations of `null` to `undefined` in order for TypeScript to not complain with optional values being either `value | undefined` rather than ` value | null`.

demedos:
Related: https://github.com/prisma/prisma/issues/11927

dielsalder:
This would be much appreciated. I understand that `null` required for databases, but there should be a way to do this so that the type exposed uses `undefined` while still using `null` under the hood. It would make working with Typescript so much easier.

Issue #11927, opened 2022-02-20
prisma optional fields are not optional


dcsan:
### Bug description

created a schema with some optional fields:

```
model Player {
    id        Int     @id @default(autoincrement())
    name      String?
    discordId String?
    px        Float?  @default(0)
    py        Float?  @default(0)
}
```

```
// create a method that takes an instance of the generated object
import { Prisma, PrismaClient, Player } from '@prisma/client'
static async findOrCreate(query: Player):  { ... } 

// and try to call it
const p = await CPlayer.findOrCreate({ name: 'alice' })
```

but the "optional" properties are actually required:

```
Argument of type '{ name: string; }' is not assignable to parameter of type 'Player'.
  Type '{ name: string; }' is missing the following properties from type 'Player': id, discordId, px, py
```

looking into the generated definitions the fields are null values are ok
but they are still *required* params.

```
export type Player = {
  id: number
  name: string | null
  discordId: string | null
  px: number | null
  py: number | null
}
```

so this means I can't really use optional params, all code has to define ALL possible properties.

### How to reproduce

create the functions as above.

### Expected behavior

a way to use truly optional params/properties in prisma objects

### Prisma information

```
generator client {
    provider = "prisma-client-js"
}

datasource db {
    provider = "postgresql"
    url      = env("DATABASE_URL")
}

model Player {
    id        Int     @id @default(autoincrement())
    name      String?
    discordId String?
    px        Float?  @default(0)
    py        Float?  @default(0)

    // fish      PlayerFish[]
    // playerFish PlayerFish[]
}
```

### Environment & setup

macOS

this affects all versions

### Prisma Version

```
prisma                  : 3.9.2
@prisma/client          : 3.9.2
Current platform        : darwin-arm64
Query Engine (Node-API) : libquery-engine bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core bcc2ff906db47790ee902e7bbc76d7ffb1893009 (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
[...]

PR #28299, opened 2025-10-14
feat(client-generator-ts)!: Add support for generating optional TypeScript properties from nullable schema fields

Aijeyomah:
**Description**
-  Make nullable fields in schema generate optional properties in TypeScript types

closes #27914

### DIFF
--- a/packages/client-generator-ts/src/TSClient/Output.test.ts
+++ b/packages/client-generator-ts/src/TSClient/Output.test.ts
@@ -1,0 +1,38 @@
+import * as DMMF from '@prisma/dmmf'
+import * as ts from '@prisma/ts-builders'
+import { beforeEach, describe, expect, test } from 'vitest'
+
+import { DMMFHelper } from '../dmmf'
+import { buildModelOutputProperty } from './Output'
+
+describe('buildModelOutputProperty', () => {
+  let dmmf: DMMFHelper
+
+  beforeEach(() => {
+    dmmf = { isComposite: () => false } as DMMFHelper
+  })
+
+  test('should add optional modifier to non-required fields', () => {
+    const field: DMMF.Field = {
+      name: 'age',
+      kind: 'scalar',
+      isList: false,
+      isRequired: false,
+      isUnique: false,
+      isId: false,
+      isReadOnly: false,
+      type: 'String',
+      hasDefaultValue: false,
+      relationName: undefined,
+      relationFromFields: undefined,
+      relationToFields: undefined,
+      relationOnDelete: undefined,
+      documentation: undefined,
+    }
+
+    const property = buildModelOutputProperty(field, dmmf)
+
+    expect(property.isOptional).toBe(true)
+    expect(ts.stringify(property.type)).toEqual('string | null')
+  })
+})
--- a/packages/client-generator-ts/src/TSClient/Output.test.ts
+++ b/packages/client-generator-ts/src/TSClient/Output.test.ts
@@ -5,11 +5,48 @@ import { beforeEach, describe, expect, test } from 'vitest'
 import { DMMFHelper } from '../dmmf'
 import { buildModelOutputProperty } from './Output'
 
+function dmmf(inputObjectTypes: DMMF.InputType[] = [], models: DMMF.Model[] = []) {
+  return new DMMFHelper({
+    schema: {
+      inputObjectTypes: {
+        prisma: inputObjectTypes,
+      },
+      outputObjectTypes: {
+        prisma: [
+          { name: 'Query', fields: [] },
+          { name: 'Mutation', fields: [] },
+        ],
+        model: [],
+      },
+      enumTypes: {
+        prisma: [],
+      },
+      fieldRefTypes: {
+        prisma: [],
+      },
+    },
+    datamodel: {
+      models: models,
+      types: [],
+      enums: [],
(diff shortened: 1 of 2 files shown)


Aijeyomah:
@jacek-prisma @FGoessler can you help review this PR? Thanks!

aqrln:
@Aijeyomah thank you for the pull request!

Jacek and Florian are currently out of office, either me or @jkomyno can review it.

The code change by itself looks good to me, but I'm not sure about the reasoning and all consequences of this change. 

One problem that I can see is that these types describe _outputs_, not _inputs_ of Prisma operations, and making fields optional in them will encode states that aren't actually possible at run time (since Prisma returns these fields as nullable, not optional) and may force any userland code that processes these outputs to handle those states (for example, check that fields are defined), essentially introducing dead code and unnecessary boilerplate. From quickly skimming through https://github.com/prisma/prisma/issues/27914 and https://github.com/prisma/prisma/issues/11927, it sounds to me like a lot of those problems were caused by misusing the model types in the context where different types (also exported by the Prisma Client) were expected instead.

At the same time, I imagine there could potentially be some cases where the existing types that we export might not be ergonomic enough, especially when chaining different operations together, but I can't imagine any concrete scenarios myself. Would you mind preparing a motivating example for this change, something that's not possible with our types right now and this change would fix?

I read the description of https://github.com/prisma/prisma/issues/27914 and I'm not sure I'm following it. The issue talks about things like partial updates, but:
- model types have nothing to do with updates
- in partial updates _all_ fields can be omitted, not just optional fields, so I don't see how this change would've helped (ignoring the fact that it's the wrong type to use)
- the relevant types that we export like `...CreateInput`/`...UpdateInput` (e.g. `UserCreateInput`) do have the appropriate fields (optional fields in create inputs and all fields in update inputs) marked as optional properties in TypeScript types

Aijeyomah:
@aqrln Thanks for the thoughtful review and for pointing this out!

You’re right that the current change affects output types, and Prisma always returns `null` for missing values rather than `undefined`. My motivation for the change came from practical usage. In some parts of my code, I need to pass Prisma results or some property of results (which is always the moodel type e.g User) into external systems or UI components that expect `undefined` for “not present” values. This mismatch forces me to introduce small utilities to normalize Prisma’s `null` to `undefined`, for example:

```ts
/** Used to resolve type conflicts between Prisma (null) and the rest of the world (undefined) */
export const undef = <T>(val: T | null): T | undefined => {
  return val ?? undefined
}
```

This isn’t something I do frequently, but it adds boilerplate and extra conversions in situations where TypeScript would otherwise already support the ergonomics of treating optional values as `undefined`.

I also noticed that the related issue (#27914) was re-tagged as `kind/feature`, which I interpreted as an invitation to explore a contribution here. Since there’s overlap with earlier discussions like #13987, I thought addressing it at the type level could be a useful improvement, even if only for certain scenarios.

That said, if the concern is that this introduces states that Prisma never produces at runtime, I completely understand why we’d want to be careful here. If the team would prefer, I’m happy to close this PR and continue using small utilities like above to normalize outputs before passing them along.

aqrln:
That makes a lot sense, however:

> This mismatch forces me to introduce small utilities to normalize Prisma’s `null` to `undefined`, for example:

Wouldn't still be necessary even with this change? This change only expands the type but those properties will still be `null` at run time and will have to be mapped to `undefined` in the same way. Besides, `T | null | undefined` is still not assignable to `T | undefined` on the type level as well.

I think to solve your problem you'd need to change the generated types to make those properties optional _instead_ of being nullable (and not in addition to that), and change the actual Prisma Client logic to return those values as `undefined` instead of `null`, making sure that certain edge cases work as expected, like nullable JSON fields, where this change would actually allow to distinguish between database null values (which will become `undefined`) and JSON null values (which will still be encoded as `null`).

That's a pretty large change but I can see how it can be useful. This will probably have to be an opt-in via a feature flag or a generator option, and it will also be necessary to make it possible to run the tests with this option either enabled or disabled.

Would you be interested in working on that? I can see value in it but I don't know if there will be buy-in from the rest of the team, so if you want to implement this, I'd suggest writing an RFC first to gather feedback.

Aijeyomah:
> Wouldn't still be necessary even with this change? This change only expands the type but those properties will still be `null` at run time and will have to be mapped to `undefined` in the same way. Besides, `T | null | undefined` is still not assignable to `T | undefined` on the type level as well.
> 
> I think to solve your problem you'd need to change the generated types to make those properties optional _instead_ of being nullable (and not in addition to that), and change the actual Prisma Client logic to return those values as `undefined` instead of `null`, making sure that certain edge cases work as expected, like nullable JSON fields, where this change would actually allow to distinguish between database null values (which will become `undefined`) and JSON null values (which will still be encoded as `null`).
> 
> That's a pretty large change but I can see how it can be useful. This will probably have to be an opt-in via a feature flag or a generator option, and it will also be necessary to make it possible to run the tests with this option either enabled or disabled.
> 
> Would you be interested in working on that? I can see value in it but I don't know if there will be buy-in from the rest of the team, so if you want to implement this, I'd suggest writing an RFC first to gather feedback.

Thanks for the explanation, that clears things up. I’ll close this PR since I see now it doesn’t solve the core issue. I might put together an RFC around an opt-in flag for undefined vs null in the future, but I’ll pause on it for now.

PR #28305, opened 2025-10-16
feat(config): add `env` helper

aqrln:
Add a helper that safely checks if an environment variable is defined and is not empty, and returns its value if so, otherwise throws an error.

This allows users to avoid having boilerplate code related to environment variables in their config files or using unsafe assertions like `process.env.DATABASE_URL!`.

Usage:

```ts
import { defineConfig, env } from '@prisma/config'

export default defineConfig({
  engine: 'classic',
  datasource: {
    url: env('DATABASE_URL'),
  },
})
```

Closes: https://linear.app/prisma-company/issue/TML-1508/env-helper-in-prismaconfig

### DIFF
--- a/packages/config/src/env.ts
+++ b/packages/config/src/env.ts
@@ -1,0 +1,14 @@
+export class PrismaConfigEnvError extends Error {
+  constructor(name: string) {
+    super(`Missing required environment variable: ${name}`)
+    this.name = 'PrismaConfigEnvError'
+  }
+}
+
+export function env(name: string): string {
+  const value = process.env[name]
+  if (!value) {
+    throw new PrismaConfigEnvError(name)
+  }
+  return value
+}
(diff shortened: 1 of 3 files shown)


jkomyno (dismissed):

packages/config/src/env.ts
```diff
@@ -0,0 +1,7 @@
+export function env(name: string): string {
+  const value = process.env[name]
+  if (!value) {
+    throw new Error(`Missing required environment variable: ${name}`)
```
jkomyno: Nit: would it make sense for this to be a subclass of `Error` with a more distinguished name?
Or to include e.g. `[ValidationEnvError] ` as a prefix to the error message?

aqrln: good idea

aqrln: done

jkomyno (approved):

PR #28342, opened 2025-10-21
test: refactor before Prisma 7

jkomyno:
This PR:
- gets rid of `mini-proxy` in tests
- gets rid of `prisma generate --data-proxy` and `prisma generate --accelerate`
- temporarily skips `prisma studio` tests
- contributes to [TML-1535](https://linear.app/prisma-company/issue/TML-1535/get-rid-of-dataproxyengine-studio)


packages/cli/src/Generate.ts
```diff
@@ -174,7 +173,6 @@ ${bold('Examples')}
         typedSql,
         noEngine:
           Boolean(args['--no-engine']) ||
-          Boolean(args['--data-proxy']) || // legacy, keep for backwards compatibility
           Boolean(args['--accelerate']) || // legacy, keep for backwards compatibility
```
aqrln: These two should also go away (both `--data-proxy` and `--accelerate` are aliases for `--no-engine`)

aqrln: if you don't want to deal with removing the `--no-engine` flag in this PR, remove just `--accelerate`

packages/client/tests/functional/_utils/setupTestSuiteEnv.ts
```diff
@@ -365,7 +364,6 @@ export type DatasourceInfo = {
  */
 export function setupTestSuiteDbURI({
   suiteConfig,
-  clientMeta,
 }: {
   suiteConfig: NamedTestSuiteConfig['matrixOptions']
   clientMeta: ClientMeta
```
aqrln:
```suggestion
```

aqrln: you'll also need to adjust the call sites i guess

packages/cli/src/Generate.ts
aqrln: Ooh, i see why you removed the `--data-proxy` flag here, you probably `grep`ed the corresponding flag from the test runner across the codebase and found this one, although it's a different thing with the same name.

In that case maybe keep them all here, and remove all three in the PR where you'll be removing the `DataProxyEngine` (or any other follow up PR).

aqrln (approved):

PR #28345, opened 2025-10-21
feat(client): remove `DataProxyEngine`, remove `/edge` import

jkomyno:
This PR:
- is built on top of https://github.com/prisma/prisma/pull/28342
- removes `DataProxyEngine`
- removes `AccelerateEngine`
- removes `@prisma/accelerate-contract`
- removes `/edge` from `@prisma/client`
- removes `in-memory` tests (which were using `library` query engine)
- removes support for `wasm-engine-edge` and `edge` from `TARGET_BUILD_TYPE`
- adjusts some functional tests so that we no longer have to care about the line numbers in snapshot tests
- closes [TML-1535](https://linear.app/prisma-company/issue/TML-1535/get-rid-of-dataproxyengine-studio)


medz:
I'm curious about what will happen to the accelerate extension after the removal of `/edge`. I'm currently using accelerate with `/edge` extensively. When I import from `/`, it complains and tells me to import from `/edge`. Furthermore, `/` imports don't work in Cloudflare.

I've noticed that Prisma has been actively collaborating with Cloudflare, but currently, compatibility with the workerd environment is the worst.

1. For the MySQL adapter, Prisma chose `mariadb-connector-nodejs`, but `mariadb-connector-nodejs` doesn't work in a workerd environment at all.

2. Because of 1, I was forced to subscribe to Prisma Postgres Enterprise Edition, which uses accelerate extensively. Therefore, I'm concerned about the migration difficulty of using `/edge` in the hundreds of places in my existing applications.

I upgrade to Prisma as soon as a new version is released, to reduce the workload of upgrading to Prisma 7 in the future.

I watch two hours of Prisma Issues and PR content every day, and the recent Prisma work has me very worried.

ankur-arch:
Hey @medz ! 👋 Really appreciate the thoughtful feedback.

Could you please check your DMs on X? We reached out there about today’s AMA on our [X account](https://x.com/prisma/status/1980914588892922271)
 (starting at 4 p.m. CEST). 
 
It would be great if you could bring up some of these insights and questions in the thread. I think they’d add a lot of value to the discussion.

packages/client/src/runtime/core/engines/common/Engine.ts
```diff
@@ -160,14 +160,14 @@ export interface EngineConfig {
...
    * @remarks this field is used internally by Policy, do not rename or remove
    */
   inlineDatasources: GetPrismaClientConfig['inlineDatasources']
 
   /**
    * The string hash that was produced for a given schema
-   * @remarks only used by DataProxyEngine.ts
+   * @remarks only used by AccelerateEngine.ts
    */
   inlineSchemaHash: string
```
jkomyno: We should strip this out.

packages/client/src/runtime/core/engines/common/Engine.ts
```diff
@@ -160,14 +160,14 @@ export interface EngineConfig {
 
   /**
    * The contents of the datasource url saved in a string
-   * @remarks only used by DataProxyEngine.ts
+   * @remarks only used by RemoteExecutor.ts
    * @remarks this field is used internally by Policy, do not rename or remove
    */
   inlineDatasources: GetPrismaClientConfig['inlineDatasources']
```
jkomyno, line 166: Double-check this one.

packages/client/src/runtime/core/engines/library/LibraryEngine.ts
```diff
@@ -353,15 +343,6 @@ You may have to run ${green('prisma generate')} for your changes to take effect.
...
       })
-    } else if (isPanicEvent(event) && TARGET_BUILD_TYPE !== 'wasm-engine-edge') {
-      // The error built is saved to be thrown later
-      this.loggerRustPanic = new PrismaClientRustPanicError(
-        getErrorMessageWithLink(
-          this,
-          `${event.message}: ${event.reason} in ${event.file}:${event.line}:${event.column}`,
-        ),
-        this.config.clientVersion!,
-      )
```
jkomyno, line 364: I think we have to reinstate this code.

packages/client/src/runtime/core/engines/common/Engine.ts
jkomyno: I confirm, it's used by `RemoteExecutor`.

aqrln (approved):

PR #28351, opened 2025-10-22
feat(cli): add `@bomb.sh/tab` completions

AmirSa12:
Context: https://x.com/prisma/status/1980198095825809692

This PR introduces [tab](https://github.com/bombshell-dev/tab/) completion functionality for the prisma cli, improving developer experience by allowing shell(zsh, powershell, fish, bash) autocompletion for commands, options, values, and flags. This also comes with npm, pnpm, yarn and bun copmpletions! ( `npm exec prisma <TAB>`, `pnpm prisma <TAB>` ) With this PR, users can navigate available commands, options and values more efficiently and faster with speeding up workflow.

a small video of how this looks:

https://github.com/user-attachments/assets/1f916a60-ee4d-40be-bb7d-74035d48ca83

NOTE: The provided option-value completions (datasource providers, ports, etc.) are examples and may be customized or removed as needed. let me know if you want them changed!


packages/cli/src/completions/Completions.ts
```diff
@@ -0,0 +1,253 @@
...
+    init.option('datasource-provider', 'Define the datasource provider', (complete) => {
+      complete('postgresql', 'PostgreSQL')
+      complete('mysql', 'MySQL')
+      complete('sqlite', 'SQLite')
+      complete('mongodb', 'MongoDB')
+      complete('sqlserver', 'SQL Server')
+      complete('cockroachdb', 'CockroachDB')
+    })
+    init.option('generator-provider', 'Define the generator provider', (complete) => {
+      complete('prisma-client-js', 'Prisma Client for JavaScript/TypeScript')
```
aqrln: This should also complete the more modern and preferred `prisma-client` generator first

AmirSa12: Hey @aqrln ! Thanks for the review!
I've just added that!
let me know if you think there are more missing option-value completions! prisma cli docs was my source for these option-value completions.

AmirSa12:
resolved conflicts and updated tab to version 0.0.16.

also added completions for the new `bootstrap` command

@ankur-arch

packages/cli/package.json
```diff
@@ -106,6 +106,7 @@
     }
   },
   "devDependencies": {
+    "@bomb.sh/tab": "0.0.16",
```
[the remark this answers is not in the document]
AmirSa12: I think this is wrong? 
cc @aqrln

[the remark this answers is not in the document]
aqrln: yeah CodeRabbit is wrong here

packages/cli/src/completions/Completions.ts
```diff
@@ -0,0 +1,272 @@
...
+        return Promise.resolve('')
+      } catch (e) {
+        return Promise.resolve(new Error(`Failed to setup completions: ${e}`))
+      }
+    }
+
+    return Promise.resolve(new HelpError('Invalid shell type. Must be one of: zsh, bash, fish, powershell'))
+  }
+
+  private setupCompletions() {
```
aqrln: Could we somehow colocate the completion definitions together with the corresponding command definitions and just gather them here from all the modules? The way it's defined here in a single place is going to be pretty hard to maintain.

aqrln:
@AmirSa12 could you please [sign the CLA](https://github.com/prisma/prisma/pull/28351#issuecomment-3433102569)?

AmirSa12:
> @AmirSa12 could you please [sign the CLA](https://github.com/prisma/prisma/pull/28351#issuecomment-3433102569)?

Singed!

packages/cli/src/completions/Completions.ts
AmirSa12: Great point. 
Let me know your thoughts about the new approach.
the completion definitions now live next to each command. and in that completions.ts file we only gather and register them.

AmirSa12:
I also did update tab to version `0.0.17 `

packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -10,12 +10,12 @@ import { dim } from 'kleur/colors'
 
 const PRISMA_EXAMPLES_TARBALL_URL = 'https://api.github.com/repos/prisma/prisma-examples/tarball/latest'
 
-interface TemplateEntry {
+export interface TemplateEntry {
```
aqrln: I don't think this type is referred by name in `bootstrapCompletion`, we don't have to export it

packages/cli/src/completions/Completions.ts
```diff
@@ -0,0 +1,157 @@
...
+const devCompletion: CommandCompletion = {
+  name: 'dev',
+  description: 'Start a local Prisma Postgres server for development',
+  options: [
+    {
+      name: 'name',
+      description: 'Target a specific database instance',
+      values: [{ value: 'my-db', description: 'Example database name' }],
+    },
+    { name: 'n', description: 'Short for --name' },
```
aqrln: Should the short variants also have `values` completions?

packages/cli/src/completions/Completions.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+function registerOption(command: ReturnType<typeof t.command>, option: CompletionOption): void {
+  if (option.values === undefined) {
+    command.option(option.name, option.description)
+    return
+  }
+
+  command.option(option.name, option.description, (complete) => {
+    const values = typeof option.values === 'function' ? option.values() : option.values!
```
aqrln: Didn't some of the options (e.g. above in `dev` completions) have `values` undefined? Should it be `option.values ?? []` instead of `option.values!`?

packages/cli/src/Generate.ts
```diff
@@ -366,3 +367,17 @@ function getCurrentClientVersion(): string | null {
...
+  name: 'generate',
+  description: 'Generate artifacts (e.g. Prisma Client)',
+  options: [
+    { name: 'schema', description: 'Custom path to your Prisma schema' },
+    { name: 'config', description: 'Custom path to your Prisma config file' },
+    { name: 'watch', description: 'Watch the Prisma schema and rerun after a change' },
+    { name: 'generator', description: 'Generator to use (may be provided multiple times)' },
+    { name: 'no-hints', description: 'Hides the hint messages but still outputs errors and warnings' },
+    { name: 'require-models', description: 'Do not allow generating a client without models' },
+    { name: 'sql', description: 'Generate typed sql module' },
```
aqrln:
```suggestion
    { name: 'sql', description: 'Generate TypedSQL module' },
```

packages/cli/src/completions/Completions.ts
AmirSa12: Thanks for the review @aqrln 

I added/removed some options and flags based on [primsa cli docs](https://www.prisma.io/docs/cli)

AmirSa12: For the values, its following what's already in the docs.

tensordreams (commented):
Thanks for this — nice DX addition! I verified every completion descriptor against the actual command `arg()` specs and help text (and against `@prisma/cli-dev` / `@prisma/dev` for the externally-loaded `dev` command). The large majority are accurate, including the trickier value hints:

- ✅ Studio default port `51212` matches `DEFAULT_PORT` in `Studio.ts`.
- ✅ `dev` ports `51213` / `51214` / `51215` match `DEFAULT_SERVER_PORT` / `DEFAULT_DATABASE_PORT` / `DEFAULT_SHADOW_DATABASE_PORT` in `@prisma/dev`, and the `ls`/`rm`/`start`/`stop` subcommand names + descriptions match `@prisma/cli-dev`.
- ✅ `init --datasource-provider` values (all 6) and `--generator-provider` values match the help text.
- ✅ `migrate diff` correctly lists only the active flags and omits the removed ones (`--from-url`, `--to-url`, `--from-schema-datasource`, …).
- ✅ All `db` / `migrate` option names and descriptions match their `arg()` specs and help.

A few issues to address, flagged inline. The most important is the `platform` subcommand tree, which doesn't correspond to any command in this codebase.

Minor (no inline comment): the new `complete` command itself isn't in `ALL_COMPLETIONS`, so `prisma complete <TAB>` won't suggest the shells it accepts (`zsh`, `bash`, `fish`, `powershell`). Optional, but easy to add for consistency. (Also, the PR description has a typo: "copmpletions".)

packages/cli/src/platform/_Platform.ts
```diff
@@ -1 +1,58 @@
+import type { CommandCompletion } from '@prisma/internals'
+
 export * as Platform from './_'
+
+export const platformCompletion: CommandCompletion = {
+  name: 'platform',
+  description: 'Manage Prisma Platform resources',
+  subcommands: [
```
tensordreams: **This subcommand tree doesn't match the actual `platform` command.** In this codebase `prisma platform` has exactly one subcommand — `status` ("Show Prisma Data Platform service status"). It's wired in `bin.ts` as `platform: Platform.$.new({ status: Status.new() })` and is the only entry in `platform/$.ts`'s help.

None of `auth`, `workspace`, `environment`, `project`, `pulse`, `accelerate` (or their leaf commands like `login`, `enable`, …) exist here — there are no source files or command classes for them. They look like they were copied from an older Prisma Platform CLI. As written, every `platform …` completion points at a non-existent command, and the only real subcommand (`status`) is missing.

Please replace this whole tree with a single `status` entry, e.g.:
```ts
export const platformCompletion: CommandCompletion = {
  name: 'platform',
  description: 'Prisma Data Platform commands',
  subcommands: [{ name: 'platform status', description: 'Show Prisma Data Platform service status' }],
}
```

packages/cli/src/Init.ts
```diff
@@ -772,3 +773,42 @@ function assertDatasourceProvider(thing: unknown): asserts thing is ConnectorTyp
...
+  name: 'init',
+  description: 'Set up Prisma for your app',
+  options: [
+    {
+      name: 'url',
+      description: 'Define a custom datasource url',
+      values: [
+        { value: 'postgresql://', description: 'PostgreSQL connection string' },
+        { value: 'mysql://', description: 'MySQL connection string' },
+        { value: 'sqlite:', description: 'SQLite connection string' },
```
tensordreams: `sqlite:` is not a valid Prisma SQLite connection-string protocol — Prisma uses `file:` (e.g. `file:./dev.db`). See `protocolToConnectorType()` in `packages/internals/src/convertCredentials.ts`: it maps `file:` → `sqlite` and throws `Unknown protocol sqlite:` for `sqlite:`. So a user who tab-completes this value and runs `prisma init --url sqlite:…` would hit an error. This should be:
```ts
{ value: 'file:', description: 'SQLite connection string' },
```

packages/cli/src/completions/Completions.ts
```diff
@@ -0,0 +1,180 @@
...
+    { name: 'debug', description: 'Enable debug mode' },
+  ],
+}
+
+const devLsCompletion: CommandCompletion = { name: 'dev ls', description: 'List available servers' }
+const devRmCompletion: CommandCompletion = { name: 'dev rm', description: 'Remove servers' }
+const devStartCompletion: CommandCompletion = { name: 'dev start', description: 'Start one or more stopped servers' }
+const devStopCompletion: CommandCompletion = { name: 'dev stop', description: 'Stop servers' }
+
+const ALL_COMPLETIONS: CommandCompletion[] = [
```
tensordreams: **`postgres` is missing from the registry.** It's a documented top-level command — `bin.ts` has `postgres: PostgresCommand.new({ link: PostgresLink.new() })`, and the root CLI help lists it as "Manage Prisma Postgres databases". It has a `postgres link` subcommand ("Link a local project to a Prisma Postgres database") with options `--api-key`, `--database`, `--force`.

Without a descriptor here, `prisma postgres <TAB>` won't complete. Consider adding a `postgresCompletion` (command + `postgres link` subcommand) alongside the others.

packages/cli/src/completions/Completions.ts
```diff
@@ -0,0 +1,180 @@
...
+
+// `dev` is implemented in `@prisma/cli-dev` (loaded via `SubCommand` in bin.ts),
+// so its descriptor lives here rather than colocated with the command.
+const devCompletion: CommandCompletion = {
+  name: 'dev',
+  description: 'Start a local Prisma Postgres server for development',
+  options: [
+    {
+      name: 'name',
+      description: 'Target a specific database instance',
```
tensordreams: Minor accuracy: this describes `--name` as targeting "a specific database instance" (and the value hint says "Example database name"), but in `@prisma/cli-dev` `--name` is *"Name of the server. Helps keep state isolated between different projects."* (default `default`). It names the dev **server**, not a database instance. Suggest something like `description: 'Name of the server'` and a value hint that reflects a server name.

AmirSa12:
Reviews addressed. also fixed the way he passed aliases.

AmirSa12:
I moved all the completion suggestions into a separate file.

AmirSa12:
@aqrln let me know if there are other things I need to do for this PR

cc @ankur-arch

aqrln:
Hey @AmirSa12, sorry for the delay! I was trying to do some manual QA before merging but I can't get this to work in any of the shells (tested in fish, bash and zsh). I loaded the completions using `eval (pnpm prisma complete fish)`, `eval $(pnpm prisma complete bash)` and `eval $(pnpm prisma complete zsh)` respectively and I'm not getting any tab completions, neither for `pnpm prisma ...`/`npx prisma ...`, nor for bare `prisma ...`. I also tried `source <(pnpm prisma complete zsh)` in zsh just in case as a sanity check to follow your docs exactly but that also doesn't work for me. Am I holding it wrong? What may be the issue?

aqrln:
shell versions just in case:

```
aqrln@lima-dev ~/w/p/s/p/s/basic-sqlite (feat/tab-completions)> fish --version
fish, version 4.7.1
aqrln@lima-dev ~/w/p/s/p/s/basic-sqlite (feat/tab-completions)> bash --version
GNU bash, version 5.3.15(1)-release (aarch64-suse-linux)
Copyright (C) 2025 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
aqrln@lima-dev ~/w/p/s/p/s/basic-sqlite (feat/tab-completions)> zsh --version
zsh 5.9 (aarch64-suse-linux-gnu)
```

aqrln:
Also tried `pnpm prisma complete fish >  ~/.config/fish/completions/prisma.fish` and opening a new fish session, with the same result. `pnpm prisma ...` prints autocompletions for `pnpm` itself, `prisma`/`npx prisma` completes the file names in the current directory.

aqrln:
Okay, figured this out! The generated completion scripts require a global installation of Prisma (which is a problem because having a globally installed Prisma CLI is almost certainly a mistake and no one should have that; it also means that even if someone installs a global Prisma CLI as a workaround to have autocompletions, they will get wrong autocompletions for the wrong Prisma version in a project using a different version of Prisma).

After symlinking the built CLI from this branch to `~/.local/bin/prisma` the completions do work, mostly, although I immediately ran into another bug: `prisma complete` command attempts to load the config file if it exists in the current directory, which makes the autocompletion slower than they should be, and completely breaks them if there's an issue in the config file.

aqrln:
Ah here's at least one thing that I was holding wrong: for autocompletion for non-global installations to work, [package-manager specific completions must be installed](https://github.com/bombshell-dev/tab/#completing-locally-installed-clis) which makes a lot of sense.

I tried `tab pnpm fish > ~/.config/fish/completions/tab-pnpm.fish` but I still get no completions for `pnpm prisma ` :(
Only the irrelevant and incorrect on this level completions for subcommands of pnpm itself.

```
aqrln@lima-dev ~/w/p/s/p/s/basic-sqlite (feat/tab-completions)> pnpm prisma 
add                (pnpm)  dlx             (pnpm)  la            (pnpm)  rebuild      (pnpm)  up                              (pnpm)  --no-include-workspace-root  (pnpm)
approve-builds     (pnpm)  doctor          (pnpm)  licenses      (pnpm)  recursive    (pnpm)  update                          (pnpm)  --no-parseable               (pnpm)
audit              (pnpm)  env             (pnpm)  link          (pnpm)  remove       (pnpm)  upgrade                         (pnpm)  --no-stream                  (pnpm)
bin                (pnpm)  exec            (pnpm)  list          (pnpm)  restart      (pnpm)  why                             (pnpm)  --no-use-stderr              (pnpm)
c                  (pnpm)  fetch           (pnpm)  ll            (pnpm)  rm           (pnpm)  --aggregate-output              (pnpm)  --no-workspace-root          (pnpm)
cache              (pnpm)  find-hash       (pnpm)  ln            (pnpm)  root         (pnpm)  --changed-files-ignore-pattern  (pnpm)  --parseable                  (pnpm)
cat-file           (pnpm)  get             (pnpm)  ls            (pnpm)  run          (pnpm)  --color                         (pnpm)  --prefix                     (pnpm)
cat-index          (pnpm)  help            (pnpm)  m             (pnpm)  run-script   (pnpm)  --dir                           (pnpm)  --reporter                   (pnpm)
ci                 (pnpm)  i               (pnpm)  multi         (pnpm)  self-update  (pnpm)  --fail-if-no-match              (pnpm)  --stream                     (pnpm)
clean-install      (pnpm)  ic              (pnpm)  outdated      (pnpm)  server       (pnpm)  --filter                        (pnpm)  --test-pattern               (pnpm)
completion         (pnpm)  ignored-builds  (pnpm)  pack          (pnpm)  set          (pnpm)  --filter-prod                   (pnpm)  --use-stderr                 (pnpm)
completion-server  (pnpm)  import          (pnpm)  patch         (pnpm)  setup        (pnpm)  --ignore-workspace              (pnpm)  --version                    (pnpm)
config             (pnpm)  init            (pnpm)  patch-commit  (pnpm)  store        (pnpm)  --include-workspace-root        (pnpm)  --workspace-packages         (pnpm)
create             (pnpm)  install         (pnpm)  patch-remove  (pnpm)  un           (pnpm)  --loglevel                      (pnpm)  --workspace-root             (pnpm)
dedupe             (pnpm)  install-clean   (pnpm)  prune         (pnpm)  uni          (pnpm)  --no-aggregate-output           (pnpm)
deploy             (pnpm)  install-test    (pnpm)  publish       (pnpm)  uninstall    (pnpm)  --no-fail-if-no-match           (pnpm)
dislink            (pnpm)  it              (pnpm)  rb            (pnpm)  unlink       (pnpm)  --no-ignore-workspace           (pnpm)
```

aqrln:
`tab pnpm complete -- prisma` works but it takes a lot of time to finish (around two seconds). Maybe something in fish times out and discards the completions? I'll try with a different shell.

aqrln:
Yeah, autocompletion in bash works but is unusably slow. `fish` probably just times out and cancels indeed. I assume it's likely related to the unnecessary config loading, and fixing that should resolve the issue.

AmirSa12:
Hey @aqrln Thanks for digging into this so thoroughly.
Yes, so as you may have noticed, there are two different completion mechanisms with different requirements:

1. `prisma complete <shell> + source` generates a script that calls prisma by program name via PATH, so it only works with a global/aliased prisma ( excluding the fact that this is indeed a mistake, and works best for clis that can be installed globally )

2. `pnpm prisma <TAB>` (after `source <(tab pnpm zsh)`), which forwards completion to the local prisma and no global install is needed. and this works best for clis like prisma

I saw the changes with the lightweight entry point perf commit, it no longer boots the full cli and load the prisma config, but as I was testing that, I wasn't satisfied with the results and it was still slow and it made me profile the timings of the delegation path (in a project with prisma as a local dep) and I noticed a performance gap upstream on tab's side specifically in this delegation route. so I'm addressing these inefficiencies upstream, and I can speed this up by 20x

I will update this PR with the new version that addresses this perf issue shortly after testing all shells.

Thanks, again. 
PLease let me know if there are any questions!

aqrln:
Sounds great @AmirSa12. I'll merge this as is for now, looking forward to the perf improvements upstream. Please open a PR as soon as you have the new version ready!

PR #28366, opened 2025-10-24
feat(cli): remove "prisma generate --no-engine" leftovers

jkomyno:


packages/client-generator-js/src/generateClient.ts
```diff
@@ -402,10 +393,6 @@ export async function generateClient(options: GenerateClientOptions): Promise<vo
     throw new DenylistError(message)
   }
 
-  if (!copyEngine) {
-    await deleteOutputDir(outputDir)
-  }
```
aqrln, line 407: Shouldn't it be unconditional
```ts
await deleteOutputDir(outputDir)
```
now instead?

aqrln: though i'm not sure i understand why it was conditional on `copyEngine` to begin with

packages/client/src/runtime/core/init/getEngineInstance.ts
```diff
@@ -14,7 +14,7 @@ import { validateEngineInstanceConfig } from './validateEngineInstanceConfig'
  * @param engineConfig
  * @returns
  */
-export function getEngineInstance({ copyEngine = true }: GetPrismaClientConfig, engineConfig: EngineConfig): Engine {
+export function getEngineInstance(_: GetPrismaClientConfig, engineConfig: EngineConfig): Engine {
```
aqrln, line 17: let's maybe remove the first argument if it's not used anymore

it's not an implementation of a generic interface, no reason to keep a parameter around if it's not used in a concrete function

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
```diff
@@ -91,8 +66,6 @@ describe('validateEngineInstanceConfig', () => {
     test('do not recommend using `--no-engine` for ClientEngine', () => {
```
aqrln: this test should be removed completely i think

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
```diff
@@ -106,8 +79,6 @@ describe('validateEngineInstanceConfig', () => {
     test('do not recommend using `--no-engine` for edge build of ClientEngine', () => {
```
aqrln: ditto

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
```diff
@@ -121,23 +92,11 @@ describe('validateEngineInstanceConfig', () => {
     test('recommend using `--no-engine` for LibraryEngine if it was not run already', () => {
```
aqrln: ditto

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
```diff
@@ -146,8 +105,6 @@ describe('validateEngineInstanceConfig', () => {
     test('do not recommend using `--no-engine` if it was run already', () => {
```
aqrln: ditto

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
```diff
@@ -238,8 +170,6 @@ describe('validateEngineInstanceConfig', () => {
     test('do not recommend using `--no-engine` for ClientEngine', () => {
```
aqrln: ditto

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
```diff
@@ -268,33 +196,19 @@ describe('validateEngineInstanceConfig', () => {
     test('recommend using `--no-engine` for LibraryEngine if it was not run already', () => {
```
aqrln: ditto

packages/client/src/runtime/core/init/validateEngineInstanceConfig.ts
```diff
@@ -76,32 +63,14 @@ export function validateEngineInstanceConfig({
...
-    targetBuildType !== 'wasm-compiler-edge'
-  ) {
-    pushWarning([
-      'recommend--no-engine',
-      'In production, we recommend using `prisma generate --no-engine` (See: `prisma generate --help`)',
-    ])
-  }
-
-  const isAccelerateConfigured = isAccelerateUrlScheme || !copyEngine
+  const isAccelerateConfigured = isAccelerateUrlScheme
```
aqrln:
```suggestion
  const isAccelerateConfigured = isUsingPrismaAccelerate || isUsingPrismaPostgres
```

packages/client/src/runtime/core/init/validateEngineInstanceConfig.ts
```diff
@@ -76,32 +63,14 @@ export function validateEngineInstanceConfig({
...
 
   if (isUsingDriverAdapters && isAccelerateConfigured) {
     if (isAccelerateUrlScheme) {
```
aqrln: redundant `if`, it is now covered by the condition above (`isAccelerateConfigured` and `isAccelerateUrlScheme` are the same variable)

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
jkomyno: Why? `--no-engine` no longer exists in the codebase

packages/client/src/runtime/core/init/getEngineInstance.ts
jkomyno: Will do in a follow-up PR.

packages/client/src/runtime/core/init/validateEngineInstanceConfig.test.ts
aqrln: exactly, so there's no need to have any tests related to it

packages/client-generator-js/src/generateClient.ts
jkomyno: ~~fixed~~

I had to revert this change because of https://github.com/prisma/prisma/actions/runs/18840988010/job/53753018023?pr=28366#step:8:35.
Getting rid of it is the correct thing to do.

aqrln (approved):

PR #28368, opened 2025-10-24
feat(client): remove `BinaryEngine`

jkomyno:
This PR:
- removes `BinaryEngine`
- removes `TARGET_BUILD: 'binary'`

/integration


packages/cli/src/__tests__/fixtures/custom-generator/prisma/schema.prisma
```diff
@@ -1,6 +1,6 @@
 generator min {
-  // https://github.com/timsuchanek/minimal-generator
-  provider = "npx @timsuchanek/minimal-generator"
+  // https://github.com/prisma/minimal-generator
+  provider = "npx github:prisma/minimal-generator"
```
aqrln, line 3: (non blocking)

could we instead have the source code of the generator live in this directory and use a `node ...` command instead of `npx ...` here?

i think it's a little bit out of scope of this PR, but i'm not a huge fan of this doing network requests and relying on npx to work

packages/client/src/runtime/core/engines/common/types/Transaction.ts
```diff
@@ -19,8 +19,7 @@ export type InteractiveTransactionInfo<Payload = unknown> = {
 
   /**
    * Arbitrary payload the meaning of which depends on the `Engine` implementation.
-   * For example, `DataProxyEngine` needs to associate different API endpoints with transactions.
-   * In `LibraryEngine` and `BinaryEngine` it is currently not used.
+   * It is currently not used in `LibraryEngine`.
    */
   payload: Payload
```
aqrln, line 24: `ClientEngine` also isn't using it, it was only used by `DataProxyEngine`. We can remove it in a follow up PR.

packages/client/src/runtime/core/engines/common/resolveEnginePath.ts
```diff
@@ -2,7 +2,7 @@ import { PrismaClientInitializationError } from '@prisma/client-runtime-utils'
 import { Debug } from '@prisma/debug'
 import { getEnginesPath } from '@prisma/engines'
 import { BinaryTarget, getBinaryTargetForCurrentPlatform, getNodeAPIName } from '@prisma/get-platform'
-import { chmodPlusX, ClientEngineType } from '@prisma/internals'
```
aqrln, line 5: `chmodPlusX` should probably be completely removed from `@prisma/internals`

aqrln (approved):
amazing work, feels so good to see this diff

packages/cli/src/__tests__/fixtures/custom-generator/prisma/schema.prisma
jkomyno: We already have other tests with the generator living close to the directory it was invoked from.
This is afaik the only test that explicitly relies on `npx` resolution, and tests that transitively. I'd be tempted to keep that there. (I didn't even know we supported this feature before seeing this test :D )

packages/client/src/runtime/core/engines/common/resolveEnginePath.ts
jkomyno: Actually, this is still needed by the SchemaEngine

PR #28375, opened 2025-10-25
feat(client): remove library engine

jkomyno:
This PR:
- gets rid of `LibraryEngine`
- gets rid of `TARGET_BUILD = 'library'`
- gets rid of `binaryTargets` and `engineType` attributes in `prisma-client-js`
- gets rid of outdated tests

We also had to (temporarily?) disable the following tests:
- **integration tests**
- **client type tests**
- **functional tests**
  - All tests when using `--generator-type prisma-client-js`
  - extensions/defineExtensions when using `driverAdapters` and `--remote-executor`
  - invalid-sqlite-isolation-level
  - issues/21136-extensions-mocking-library
- **e2e tests**
  - `accelerate-types`
  - `unsupported-edge-error`
  - `connection-limit-reached`
  - `mongodb-notablescan`
  - `issues/17303-interactive-transaction-errors`
  - `nextjs-schema-not-found/7_monorepo-noServerComponents-noCustomOutput-noReExport`
  - `nextjs-schema-not-found/8_monorepo-serverComponents-noCustomOutput-noReExport`
  - `nextjs-schema-not-found/11_monorepo-noServerComponents-noCustomOutput-noExportIndirect`
  - `nextjs-schema-not-found/15`
  - `nextjs-schema-not-found/16`
  - `nextjs-schema-not-found/17`
  - `nextjs-schema-not-found/18`

We should re-enable them in follow-up PRs.

Other TODOs for follow-up PRs:
- Change "chunking query" message in `packages/client-engine-runtime/src/interpreter/render-query.ts`
- Figure out why functional/omit/test.ts showcases has different snapshots depending on the generator used
- Test `wasm-compiler-edge` runtime
---

Discoveries:
- `wasm-compiler-edge` was never tested before. Adding it to the test matrix, causes failures that are out of scope for this PR.
- Testing the `prisma-client-js` generator along with `prisma-client-ts` generator caused some Jest snapshots to fail. For this reason, I raised this issue internally with Alexey and temporarily stopped testing `prisma-client-js`. This should be fixed in a follow-up PR.

/integration

PR #28411, opened 2025-10-29
feat!: introduce `accelerateUrl`

aqrln:
This PR introduces a new way to configure the Accelerate URL in preparation to removing `url` from PSL:

```ts
new PrismaClient({
  accelerateUrl: "prisma://...",
})
```

This property is mutually exclusive with `adapter`, and exactly one of them needs to be provided when using `engineType = "client"`.

It also makes a few related changes along the way or as a side effect:

- Contributes towards [TML-1459](https://linear.app/prisma-company/issue/TML-1459/remove-processenv-loading-from-prisma-cli-and-prismaclient): the logic related to environment variables was completely removed from the `RemoteExecutor` code path, along with the `resolveDatasourceUrl` function.

- Fixes the `@prisma/extension-accelerate` extension which was inadvertently broken in https://github.com/prisma/prisma/pull/28345: while most of `accelerateUtils` (which were meant to be used by the removed `AccelerateEngine`) were indeed never used, one specific method is actually currently used: `_accelerateEngineConfig.accelerateUtils.resolveDatasourceUrl`. This PR now restores this function and re-implements it in a different way.
  - As a result of this, we now have a new validation and a DX improvement we didn't use to: now if you apply the Accelerate extension to a client that's not actually configured to use Accelerate, you will get a very clear error with a code suggestion.

- Removes `null` support for the `adapter` constructor option, which was introduced in https://github.com/prisma/prisma/pull/21315. It was the only option that required `null` instead of `undefined` to disable it explicitly and it was inconsistent with the others. The [motivation for introducing `null` and making it different from `undefined`](https://github.com/prisma/prisma/pull/21163#discussion_r1332724218) made total sense for the `LibraryEngine` (where accidentally setting the `adapter` to `undefined` meant silently using completely different drivers before that change) but is now obsolete for the `ClientEngine`. Since we are removing `LibraryEngine` in https://github.com/prisma/prisma/pull/28375, we don't need this distinction anymore since there are no longer any pitfalls from using `adapter: undefined`, and removing it here in this PR made the validation around the combinations of `adapter` and `accelerateUrl` easier.

- Fixed the incorrect error message when Accelerate URL cannot be parsed.

Closes: https://linear.app/prisma-company/issue/TML-1507/env-less-accelerate-url-configuration


medz:
@aqrln Fantastic! I'm a Prisma accelerate user! This is exactly what I needed. The adapter was preventing me from switching to `clientType = 'client'`! With this, it means I only need `const prisma = new PrismaClient({accelerateUrl: "..."})`!

The current client adapter is extremely unfriendly to Prisma Postgres users! Not only are they forced to continue using `prisma-client-js`, but they also can't prepare for migrating to Prisma 7 in advance.

You're a superhuman! You've completely solved my pain points!

aqrln:
@medz you can use Accelerate with `engineType = "client"` with Prisma 6 too, the way to do this is to not configure any driver adapter. It should work as long as the `url` property in the datasource block points at a `prisma://` or `prisma+postgres://` URL (alternatively, you can use `datasourceUrl` constructor option). But yeah, it's a little confusing.

jkomyno (approved):

PR #28432, opened 2025-11-03
feat: apply breaking changes for the diff command parameters

jacek-prisma:
[TML-1538](https://linear.app/prisma-company/issue/TML-1538/remove-from-schema-datasource-schema-etc-from-migrate-commands)


jacek-prisma:
Seems like the tests are failing because it's not using the schema engine CLI from my branch, the tests pass locally.

packages/adapter-d1/src/d1.ts
```diff
@@ -33,3 +36,33 @@ export class PrismaD1<Args extends D1Database | D1HttpParams = D1Database> imple
...
+ *
+ * @example
+ * ```typescript
+ * import { defineConfig } from '@prisma/config'
+ * import { listLocalDatabases } from '@prisma/adapter-d1'
+ *
+ * export default defineConfig({
+ *   engine: 'classic',
+ *   datasource: {
+ *     url: `file://${listLocalDatabases().pop()}`,
```
aqrln: without `//`

packages/migrate/src/__tests__/DbPull/sqlite.test.ts
```diff
@@ -53,16 +53,16 @@ describeMatrix({ providers: { d1: true }, driverAdapters: {} }, 'D1', () => {
...
 
     const introspect = new DbPull()
-    const result = introspect.parse(['--local-d1'], await ctx.config())
+    const result = introspect.parse([], await ctx.config())
 
     await expect(result).resolves.toMatchInlineSnapshot(`""`)
     expect(ctx.normalizedCapturedStdout()).toMatchInlineSnapshot(`
       "Prisma schema loaded from prisma/schema.prisma
-      Datasource "db": SQLite database "dev.db" <location placeholder>
+      Datasource "db": SQLite database "5d11bcce386042472d19a6a4f58e40041ebc5932c972e1449cbf404f3e3c4a7a.sqlite" <location placeholder>
```
aqrln, line 65: Is this stable across test runs?

packages/migrate/src/__tests__/MigrateDiff.test.ts
```diff
@@ -839,13 +574,29 @@ describe('migrate diff', () => {
...
+      await expect(result).rejects.toThrowErrorMatchingSnapshot()
+    })
   })
 
   describeMatrix(cockroachdbOnly, 'cockroachdb', () => {
     if (!process.env.TEST_SKIP_COCKROACHDB && !process.env.TEST_COCKROACH_URI_MIGRATE) {
       throw new Error('You must set a value for process.env.TEST_COCKROACH_URI_MIGRATE. See TESTING.md')
     }
-    const connectionString = process.env.TEST_COCKROACH_URI_MIGRATE?.replace('tests-migrate', 'tests-migrate-diff')
+    const connectionString = process.env.TEST_COCKROACH_URI_MIGRATE
```
aqrln, line 599: why was this changed?

packages/migrate/src/__tests__/fixtures/cloudflare-d1-one-db/prisma.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from '@prisma/config'
+import { listLocalDatabases } from '@prisma/adapter-d1'
+
+export default defineConfig({
+  engine: 'classic',
+  datasource: {
+    url: `file://${listLocalDatabases().pop()}`,
```
aqrln: `//` shouldn't be necessary here, although if it also works then it's fine i guess

packages/migrate/src/__tests__/DbPull/sqlite.test.ts
jacek-prisma: yes, the database file is part of the git repo

packages/migrate/src/__tests__/MigrateDiff.test.ts
jacek-prisma: it's unnecessary as far as I can tell and it was causing problems with some tests where the connection string would have the database set to `tests-migrate` in the config, this `connectionString` is used to create and tear down the databases so it needs to be aligned with the config

it wasn't a problem before because it was being passed as a `url` CLI argument, but since that's no longer the case I thought it's simpler to just use one shared database name

packages/adapter-d1/src/d1.ts
jacek-prisma: updated

packages/migrate/src/__tests__/fixtures/cloudflare-d1-one-db/prisma.config.ts
jacek-prisma: updated

aqrln (approved):

PR #28440, opened 2025-11-04
chore(ci): create tags on release

aqrln:
Previously we relied on the git tags being created manually (via GitHub UI) when creating a new release. This led to occasional human errors where wrong commits were tagged when new commits were pushed right after the release, as well as to a delay in creating the tags compared to publishing the release on npm.

As previously [agreed][], we're changing the process to have the tags pushed automatically from CI, and then a GitHub release with release notes being created manually afterwards.

[agreed]: https://prisma-company.slack.com/archives/C9D3H0P1A/p1761136642568979?thread_ts=1761119656.960289&cid=C9D3H0P1A

Furthermore, previously we'd create the tags automatically in the `prisma-engines` repo for all Prisma releases, including dev and integration. However, we did not create tags for dev and integration releases in the main repo, which sometimes was slightly annoying when bisecting or reproducing issues. Now every release will always be tagged, regardless of the release type.

---

Demonstration of the tags created from this branch:
- https://github.com/prisma/prisma/tree/6.19.0-integration-release-tags.4
- https://github.com/prisma/prisma-engines/tree/6.19.0-integration-release-tags.4


.github/actions/create-git-tag/action.yml
```diff
@@ -0,0 +1,102 @@
...
+            })
+          } catch (error) {
+            if (error.status === 422 && error.message?.includes('Reference already exists')) {
+              core.info(`Tag reference refs/tags/${tagName} already exists on ${owner}/${repo}, skipping.`)
+              return
+            }
+            throw error
+          }
+
+          core.info(`Created tag ${tagName} on ${owner}/${repo} at ${commitSha}.`)
```
jacek-prisma: can this be extracted into a file somewhere in .github directory?

aqrln: Probably yeah, the inline script here could import the file in `.github` directory.

.github/workflows/release-ci.yml
```diff
@@ -90,34 +90,23 @@ jobs:
...
+        if: ${{ steps.publish.outputs.prismaVersion != '' && !(github.event_name == 'workflow_dispatch' && github.event.inputs.dryRun == 'true') }}
+        uses: ./.github/actions/create-git-tag
+        with:
+          token: ${{ secrets.BOT_TOKEN }}
+          tag-name: ${{ steps.publish.outputs.prismaVersion }}
+          commit-sha: ${{ github.sha }}
+
       - name: Create a tag on prisma-engines repository
-        uses: actions/github-script@v7
+        uses: ./.github/actions/create-git-tag
```
jacek-prisma, line 103: shouldn't this have the same `if` field as `tag-prisma`?

aqrln: yes, it should

jacek-prisma (dismissed):

jacek-prisma (approved):

PR #28446, opened 2025-11-05
feat(adapters): PPg Serverless Driver Adapter + smoke tests

pcan:
- Added new `adapter-ppg` which leverages `@prisma/ppg` to perform queries (via HTTP) and interactive transactions (via WebSocket).
- Added smoke test for ppg, verified it works against a test PPg database.


packages/adapter-ppg/src/ppg.ts
```diff
@@ -0,0 +1,241 @@
...
+  readonly #client: Client
+
+  readonly adapterName = '@prisma/ppg-adapter'
+  readonly provider = 'postgres'
+
+  constructor(config: PrismaPostgresAdapterConfig) {
+    const existingParsers = 'parsers' in config && config.parsers instanceof Array ? config.parsers : []
+    this.#client = client({
+      ...config,
+      parsers: [...existingParsers, ...builtinParsers],
```
aqrln, line 82: Is it intentional that users can't override the bulit-in parsers?

aqrln: also, that parsers are hidden in TS types?

sandbox/driver-adapters/.envrc.example
```diff
@@ -1,3 +1,4 @@
 export JS_PLANETSCALE_DATABASE_URL="mysql://USER:PASSWORD@aws.connect.psdb.cloud/DATABASE?sslaccept=strict"
 export JS_NEON_DATABASE_URL="postgres://USER:PASSWORD@DATABASE-pooler.eu-central-1.aws.neon.tech/neondb?pgbouncer=true&connect_timeout=10"
 export JS_PG_DATABASE_URL="postgres://prisma:prisma@localhost:5432/tests"
+export JS_PPG_DATABASE_URL="postgres://identifier:key@db.prisma.io:5432/postgres?sslmode=require"
```
aqrln: nit: missing LF at the end of file

sandbox/driver-adapters/package.json
```diff
@@ -38,4 +40,4 @@
     "tsx": "4.1.2",
     "typescript": "5.2.2"
   }
-}
+}
```
aqrln: likewise, it was unnecessarily changed here

aqrln (dismissed):
A few non-blocking comments, looks great to me overall.

We might find something missing once we can run the full client test suite with this adapter but for now it looks solid and totally usable.

packages/adapter-ppg/src/ppg.ts
pcan: 1. The built-in always override the configured one, to prevent users from breaking the contract with the main prisma engine. The additional ones if specified, should only allow covering cases that the default ones don't cover.
2. You're right. This is intentionally an _hidden param_ for special use cases which we don't want to expose to the public (in general). 

I think we could even drop this altogether in the long term.

sandbox/driver-adapters/.envrc.example
pcan: fixing in the next commit

sandbox/driver-adapters/package.json
pcan: I think the IDE took some liberty here :D

do4gr (approved):

aqrln (approved):

PR #28451, opened 2025-11-06
feat!: remove `directUrl`

aqrln:
Remove `directUrl` from `prisma.config.ts`. There's no semantic difference between `url` and `directUrl` in `prisma.config.ts` and there is no need for any distinction between them. The runtime connection string is not configured in `prisma.config.ts`, and the `datasource.url` property there is exactly what `directUrl` in schema files used to mean.

This commit also removes some legacy remains of `directUrl` in the PSL datasource block, which is not allowed by PSL anymore, as well as some invalid usage of `shadowDatabaseUrl` property in the PSL which didn't even exist in the TypeScript representation of the `DataSource` type and only remained uncaught by TypeScript due to the usage of type assertions.

Closes: https://linear.app/prisma-company/issue/TML-1539/remove-directurl-from-prismaconfigts


packages/internals/src/cli/schemaContext.ts
```diff
@@ -117,14 +117,11 @@ export async function processSchemaResult({
 
   const primaryDatasource = match(schemaEngineConfig)
     .with({ engine: 'classic' }, ({ datasource }) => {
-      const { url, directUrl, shadowDatabaseUrl } = datasource
```
jacek-prisma, line 120: doesn't`shadowDatabaseUrl` still need to be passed through here?

aqrln: I don't think so, it's not used anywhere and it's not present in types (this only worked due to the `as DataSource` cast below).

FWIW, it's not the structure that's passed to the schema engine as datasource overrides, in case there's any confusion.

jacek-prisma: ok makes sense

jacek-prisma (approved):

jkomyno (approved):

PR #28453, opened 2025-11-06
feat: extract a shared release workflow

jacek-prisma:
Needed for [TML-1543](https://linear.app/prisma-company/issue/TML-1543/chore-migrate-npm-tokens-to-oidc). We need at most 1 publishing workflow.

Tested in https://github.com/prisma/prisma/actions/runs/19132603441/job/54676827671.

### DIFF
--- a/.github/workflows/release-ci.yml
+++ b/.github/workflows/release-ci.yml
@@ -36,8 +36,8 @@ jobs:
     uses: ./.github/workflows/release.yml
     with:
       timeout-minutes: 75
-      dry-run: ${{ github.event.inputs.dryRun || false }}
-      force-integration-release: ${{ github.event.inputs.forceIntegrationRelease || false }}
+      dry-run: ${{ github.event.inputs.dryRun == 'true' }}
+      force-integration-release: ${{ github.event.inputs.forceIntegrationRelease == 'true' }}
       skip-tagging-on-dry-run: ${{ github.event_name == 'workflow_dispatch' }}
       include-failure-notifications: false
     secrets: inherit
--- a/.github/workflows/release-ci.yml
+++ b/.github/workflows/release-ci.yml
@@ -38,6 +38,5 @@ jobs:
       timeout-minutes: 75
       dry-run: ${{ github.event.inputs.dryRun == 'true' }}
       force-integration-release: ${{ github.event.inputs.forceIntegrationRelease == 'true' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,182 @@
...
+          THE_INPUT: '${{ toJson(inputs) }}'
+        run: |
+          echo "Workflow inputs:"
+          echo "$THE_INPUT"
+
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/setup
+        with:
+          node-version: 18
```
aqrln: maybe let's use 20 here so we don't miss it when merging it back to main

.github/workflows/release-ci.yml
```diff
@@ -30,101 +30,14 @@ on:
...
-          SLACK_TITLE: 'prisma/prisma Release ${{ needs.release.outputs.prismaVersion }} succeeded :white_check_mark:'
-          SLACK_COLOR: '#55ff55'
-          SLACK_WEBHOOK: ${{ secrets.SLACK_RELEASE_FEED_WEBHOOK }}
-          SLACK_CHANNEL: feed-prisma-publish
+    uses: ./.github/workflows/release.yml
+    with:
+      timeout-minutes: 75
+      dry-run: ${{ github.event.inputs.dryRun == 'true' }}
+      force-integration-release: ${{ github.event.inputs.forceIntegrationRelease == 'true' }}
+      skip-tagging-on-dry-run: ${{ github.event_name == 'workflow_dispatch' }}
```
aqrln: I'm actually not sure it is still necessary, I didn't touch it when adding `prisma/prisma` tags (`prisma-engines` tagging step already had these conditions as they were) but I feel like it's unnecessary (especially now it's obvious because your refactoring made the logic easier to understand). We should not be creating git tags if we're not publishing the packages to npm, regardless of the event which triggered this. Besides, a dry run implies `workflow_dispatch` anyway because we don't do dry runs automatically.

.github/workflows/release.yml
jacek-prisma: updated

.github/workflows/release-ci.yml
jacek-prisma: updated

aqrln (approved):

PR #28455, opened 2025-11-06
feat(config): remove `engine` and `adapter`, make `datasource` mandatory

jkomyno:
This PR:
- removes the `engine` and `adapter` properties from `prisma.config.ts`
- makes the `datasource` property required in `prisma.config.ts`
- closes [TML-1563](https://linear.app/prisma-company/issue/TML-1563/prismaconfig-get-rid-of-engine-and-adapter)

/integration


packages/config/package.json
```diff
@@ -20,6 +20,7 @@
   "devDependencies": {
     "@prisma/driver-adapter-utils": "workspace:*",
     "@prisma/get-platform": "workspace:*",
+    "dotenv": "17.2.3",
```
jkomyno, line 23: Without this, some test fixtures that used `import 'dotenv/config'` were failing.

packages/internals/src/cli/schemaContext.ts
```diff
@@ -115,16 +112,17 @@ export async function processSchemaResult({
 
   const datasourceFromPsl = configFromPsl.datasources.at(0)
 
+  // Note: once @aqrln gets rid of `url` in `datasource`, he'll have to reconcile this code here.
```
jkomyno, line 115: @aqrln

aqrln: Yeah, on my other branch where I remove `url` I have this block completely removed

packages/cli/src/__tests__/commands/Version.test.ts
```diff
@@ -10,23 +10,22 @@ describe('version', () => {
   describe('shows @prisma/schema-engine-wasm when config.migrate.adapter is set', () => {
     describe('bypassing query engine env vars', () => {
       test('does not download query-engine when engine type is client', async () => {
```
aqrln: the test name is probably a bit out of date

aqrln: i mean, it's technically true, but it's not what we're testing here anymore

PR #28460, opened 2025-11-06
ci: merge the release workflows

jacek-prisma:
It seems like my worry was correct and we have to have a single top-level workflow for the OIDC publish to work, so I merged the three release files together.

I didn't notice that it was broken because I mistakenly run the test while the NPM token was still being populated in CI. Now that the token is gone the publish job is failing on `main`. I've run an integration publish from this branch and it has worked: https://github.com/prisma/prisma/actions/runs/19141641946/job/54708053029.


.github/workflows/release.yml
```diff
@@ -1,58 +1,73 @@
...
+    ${{
+      github.event.inputs.targetVersion &&
+      format('prisma/prisma Release {0} from {1} on latest tag failed :x:', github.event.inputs.targetVersion, github.ref_name) ||
+      'prisma/prisma Release failed :x:'
+    }}
 
 jobs:
   release:
-    timeout-minutes: ${{ inputs.timeout-minutes }}
+    timeout-minutes: ${{ github.event.inputs.targetVersion && 45 || 75 }}
```
aqrln, line 66: If I were to guess, they probably diverged at some point inadvertently. I'm fine with it either way though.

aqrln (approved):

PR #28461, opened 2025-11-06
feat(cli): new studio.

igalklebanov:
Hey 👋 

This PR replaces the old Studio with the new.

We'll add more provider support as we introduce new adapters in `@prisma/studio-core`.

We'll add a happy path E2E/component test in a follow up PR.


tsconfig.build.bundle.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "./tsconfig.build.regular.json",
   "compilerOptions": {
+    "module": "preserve",
+    "moduleResolution": "bundler",
```
igalklebanov, line 5: This is to enable resolution of `package.json#exports` of dependencies.

We build with `esbuild` anyway.

igalklebanov: It is done here and not in `tsconfig.build.regular.json` to not break legacy types tests that rely on `commonjs` and `classic`.

packages/dmmf/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './convert'
 export * from './dmmf'
+export * from './util'
```
igalklebanov, line 3: resolves a typescript error.

packages/cli/package.json
```diff
@@ -136,16 +135,18 @@
...
+    "hono": "4.9.4",
     "jest": "29.7.0",
     "jest-junit": "16.0.0",
     "kleur": "4.1.5",
     "line-replace": "2.0.1",
     "log-update": "6.1.0",
     "node-fetch": "3.3.2",
     "npm-packlist": "5.1.3",
-    "open": "7.4.2",
+    "open": "8.4.2",
```
igalklebanov, line 147: this is the last version before sindre went esm-only.
tried latest version and it doesn't work even when dynamically imported.

the newer versions support more browsers (e.g. brave). so cjs is holding back dx.

packages/cli/package.json
```diff
@@ -136,16 +135,18 @@
     "env-paths": "2.2.1",
     "execa": "5.1.1",
     "fs-extra": "11.3.0",
-    "get-port": "7.1.0",
+    "get-port-please": "3.2.0",
```
igalklebanov, line 138: unjs 👑 .

packages/cli/src/Studio.ts
```diff
@@ -90,85 +128,236 @@ ${bold('Examples')}
...
-    const schemaContext = await loadSchemaContext({
-      schemaPathFromArg: args['--schema'],
-      schemaPathFromConfig: config.schema,
-      schemaEngineConfig: config,
-    })
+    if (config.engine !== 'js') {
+      return new Error(
+        'Prisma Studio is not supported when using the "classic" query engine. Please switch to the "js" engine in your Prisma config.',
+      )
+    }
```
aqrln: `engine` in `prisma.config.js` doesn't refer to query engine, it refers to schema engine (i.e. Prisma Migrate). We're removing this property before launch because it's super confusing, and `engine === 'js'` is not usable by end users since almost nothing is implemented there yet.

igalklebanov: The types currently require `engine: 'js'` narrowing to access `adapter`.

packages/cli/package.json
jkomyno: I think we should bump Prisma to be esm-first @aqrln

packages/cli/src/Studio.ts
jkomyno: >The types currently require engine: 'js' narrowing to access adapter.

Feel free to override them, as Alexey said, we're removing the `engine` property from Prisma Config in Prisma 7.

aqrln: Theres also `config.studio?.adapter` which is a separate thing and available unconditionally.

igalklebanov: We talked about dropping `config.studio?.adapter` because it makes no sense to have a separate adapter just for Studio iirc.

aqrln: I would also be fine with not supporting custom driver adapters for studio for now and bundling a limited set of default driver adapters that will be automatically used with `config.datasource.url`.

I actually think that we should do it anyway, even if we allow to override the default driver adapter with a custom one using `prisma.config.ts`.

igalklebanov: Won't the user always define an adapter, for `PrismaClient` purposes?

jkomyno: >Won't the user always define an adapter, for PrismaClient purposes?

`prisma.config.ts` has nothing to do with PrismaClient. It regulates the CLI only

igalklebanov: OK, made some changes:

It tries reading `config.datasource.url`. Can override with `--url`.
It currently only accepts valid URLs. File paths will probably require getting `provider` from schema file - e.g. SQLite file vs. PGlite datadir.

packages/cli/package.json
```diff
@@ -166,7 +167,9 @@
   },
   "dependencies": {
     "@prisma/config": "workspace:*",
-    "@prisma/engines": "workspace:*"
+    "@prisma/engines": "workspace:*",
+    "@prisma/studio-core-licensed": "0.6.0",
```
jkomyno, line 171: Interesting, I thought we didn't need the `-licensed` version here?

packages/cli/src/Studio.ts
```diff
@@ -90,84 +152,153 @@ ${bold('Examples')}
...
-      schemaPathFromConfig: config.schema,
-    })
+    const connectionString = args['--url'] || (config as { datasource?: { url?: string } }).datasource?.url
 
-    const hostname = args['--hostname']
-    const port = args['--port'] || (await getPort({ port: portNumbers(5555, 5600) }))
-    const browser = args['--browser'] || process.env.BROWSER
+    if (!connectionString) {
+      return new Error('No database URL found.')
+    }
```
jkomyno, line 159: We can improve the way this is validated later, à la https://github.com/prisma/prisma/pull/28499

sandbox/studio/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "scripts": {
+    "studio": "dotenvx run -- tsx ../../packages/cli/build/index.js studio --config ./prisma.config.ts",
+    "studio:debug": "DEBUG=prisma:* pnpm studio"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@prisma/adapter-pg": "../../packages/adapter-pg",
+    "@prisma/client": "../../packages/client",
```
jkomyno, line 15: Are these deps actually needed?

jkomyno (approved):
Approved with a few comments. We'll likely need a few follow-up PRs here

packages/cli/package.json
igalklebanov: Following what was done in `language-tools`.

sandbox/studio/package.json
igalklebanov: not really. claude just plopped this for me and i never bothered because it's a sandbox.

PR #28464, opened 2025-11-07
feat(client): [breaking] remove `postinstall` hook, remove "auto-install" on `prisma generate`

jkomyno:
This PR:
- closes [TML-1549](https://linear.app/prisma-company/issue/TML-1549/remove-generate-on-postinstall)
- closes [TML-1574](https://linear.app/prisma-company/issue/TML-1574/remove-auto-install-of-prismaclient-if-not-already-installed)
- removes the `postinstall` hook on `@prisma/client`
- removes the "auto-install" mechanism on `prisma-client-js`
- removes support for the following env vars:
  - `PRISMA_SKIP_POSTINSTALL_GENERATE`
  - `PRISMA_GENERATE_SKIP_AUTOINSTALL`
  - `PRISMA_GENERATE_IN_POSTINSTALL`


packages/client/scripts/postinstall.js
```diff
@@ -72,24 +38,18 @@ async function main() {
...
   debug({
     localPath,
     installedGlobally,
     init_cwd: process.env.INIT_CWD,
-    PRISMA_GENERATE_IN_POSTINSTALL: process.env.PRISMA_GENERATE_IN_POSTINSTALL,
   })
   try {
     if (localPath) {
-      await run('node', [localPath, 'generate', '--postinstall', doubleQuote(getPostInstallTrigger())])
+      await run('node', [localPath, 'generate', doubleQuote(getPostInstallTrigger())])
```
jacek-prisma: I'm pretty sure the `doubleQuote(getPostInstallTrigger())` parameter is a value passed to `--postinstall`, so it should no longer be present

Also I thought that this PR would get rid of the generate call entirely from the postinstall hook, what's the reason to still have it?

jkomyno: This `postinstall` script is no longer ran on user-land. It's still used on some local workflows internal to Prisma, when running `pnpm run generate` in `@packages/client`. It can probably be removed, I'll see how the tests go.

packages/client-generator-js/src/generateClient.ts
```diff
@@ -295,26 +290,10 @@ function getTypedSqlRuntimeBase(runtimeBase: string) {
 }
 
 // TODO: explore why we have a special case for excluding pnpm
```
wmadden:
```suggestion
```

Looks like that special case no longer exists?

wmadden (approved):
Nice work! I'm surprised how much infra was related to this postinstall hook. Happy to see it gone! 👍🏻

packages/client-generator-js/src/generateClient.ts
jkomyno: Right, got rid of it.

packages/cli/src/__tests__/commands/DebugInfo.test.ts
```diff
@@ -93,69 +90,64 @@ describe('debug', () => {
...
-      For Prisma Studio
-      - BROWSER:
-
-      -- Terminal is interactive? --
-      false
-
-      -- CI detected? --
-      false
-      "
+        Path: REDACTED_PATH
```
aqrln: Wrong formatting, everything after the first line is now shifted to the right

packages/cli/src/DebugInfo.ts
```diff
@@ -71,68 +71,63 @@ export class DebugInfo implements Command {
...
-For Prisma Studio
-${formatEnvValue('BROWSER')}
-
-${underline('-- Terminal is interactive? --')}
-${isInteractive()}
-
-${underline('-- CI detected? --')}
-${isCi()}
-`
+  Path: ${schemaPath}
```
aqrln: This must be shifted back to the left

packages/cli/src/__tests__/commands/DebugInfo.test.ts
jkomyno: I fixed it once already in https://github.com/prisma/prisma/pull/28464/commits/20eaa435ebfd87f6672b6f2ec06a7305a2806fe1, looks like there's something wrong with either the way `mergiraf` was called, or something else. Thanks

aqrln: could be a bug in mergiraf!

aqrln: I've had it shuffle around methods in a class before (though that at least doesn't change the behaviour, unlike here)

aqrln (approved):

Issue #17303, opened 2023-01-13
Prisma interactive transaction ignores DB exception on commit (with triggers)


Siegrift:
### Bug description

I have created a repro in https://github.com/Siegrift/prisma-interactive-tx-bug-repro/tree/trigger

I have two models, `User` and `Deposit`. I remember `balance` field in the `User` table and I want it to be consistent with the sum of deposits. I am using PSQL constraint trigger for this. I noticed that the trigger failed, but Prisma silently resolved the promise and there were no errors in the logs. I was only able to find the error inside the logs of the database.

### How to reproduce

See: https://github.com/Siegrift/prisma-interactive-tx-bug-repro/tree/trigger

### Expected behavior

Prisma should notice the DB commit failed and gracefully reject the promise.

### Prisma information

See: https://github.com/Siegrift/prisma-interactive-tx-bug-repro/tree/trigger

### Environment & setup

- OS: macOS
- Database: PostgreSQL
- Node.js version: v18.12.1

### Prisma Version

```
prisma                  : 4.8.1
@prisma/client          : 4.8.1
Current platform        : darwin-arm64
Query Engine (Node-API) : libquery-engine d6e67a83f971b175a593ccc12e15c4a757f93ffe (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Migration Engine        : migration-engine-cli d6e67a83f971b175a593ccc12e15c4a757f93ffe (at node_modules/@prisma/engines/migration-engine-darwin-arm64)
Introspection Engine    : introspection-core d6e67a83f971b175a593ccc12e15c4a757f93ffe (at node_modules/@prisma/engines/introspection-engine-darwin-arm64)
Format Binary           : prisma-fmt d6e67a83f971b175a593ccc12e15c4a757f93ffe (at node_modules/@prisma/engines/prisma-fmt-darwin-arm64)
Format Wasm             : @prisma/prisma-fmt-wasm 4.8.0-61.d6e67a83f971b175a593ccc12e15c4a757f93ffe
Default Engines Hash    : d6e67a83f971b175a593ccc12e15c4a757f93ffe
Studio                  : 0.479.0
```

Siegrift:
This is pretty crucial and can lead to critical errors. E.g. you can implement a feature correctly, and want to execute some custom logic on tx rollback (e.g. log to some monitoring service). With this bug, that kind of logic cannot be executed and is it silently fails.

[...]

PR #28468, opened 2025-11-07
fix: fix hanging transactions #17303

jacek-prisma:
[TML-1562](https://linear.app/prisma-company/issue/TML-1562/investigate-and-fix-issues17303-interactive-transaction-errors-e2e)

Simply adds a `finally` block to make sure we always dispose of the transactions

### DIFF
--- a/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
+++ b/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
@@ -243,10 +243,7 @@ export class TransactionManager {
             const query = COMMIT_QUERY()
             await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query)).then(
               () => tx.transaction!.commit(),
-              async (err) => {
-                await tx.transaction!.rollback()
-                throw err
-              },
+              (err) => tx.transaction!.rollback().catch(() => Promise.reject(err)),
             )
           }
         } else if (tx.transaction) {
--- a/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
+++ b/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
@@ -243,7 +243,10 @@ export class TransactionManager {
             const query = COMMIT_QUERY()
             await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query)).then(
               () => tx.transaction!.commit(),
-              (err) => tx.transaction!.rollback().finally(() => Promise.reject(err)),
+              async (err) => {
+                await tx.transaction!.rollback()
+                throw err
+              },
             )
           }
         } else if (tx.transaction) {
--- a/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
+++ b/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
@@ -241,8 +241,11 @@ export class TransactionManager {
             await this.#withQuerySpanAndEvent(PHANTOM_COMMIT_QUERY(), tx.transaction, () => tx.transaction!.commit())
           } else {
             const query = COMMIT_QUERY()
-            await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
-            await tx.transaction.commit()
+            try {
+              await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
+            } finally {
+              await tx.transaction.commit()
+            }
           }
         } else if (tx.transaction) {
           if (tx.transaction.options.usePhantomQuery) {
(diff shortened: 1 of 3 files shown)


packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -241,8 +241,11 @@ export class TransactionManager {
             await this.#withQuerySpanAndEvent(PHANTOM_COMMIT_QUERY(), tx.transaction, () => tx.transaction!.commit())
           } else {
             const query = COMMIT_QUERY()
-            await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
-            await tx.transaction.commit()
+            try {
+              await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
+            } finally {
+              await tx.transaction.commit()
+            }
```
aqrln: should it rather be something like this?

```suggestion
            try {
              await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
            } catch (err) {
              await tx.transaction.rollback().finally(() => Promise.reject(err))
            }
            await tx.transaction.commit()
```

jacek-prisma: hmm good point, we should call roll back on the transaction with a failing commit,
in practice these commit and rollback only release the transaction, so there's no difference between them but semantically it makes more sense to use `rollback`

jacek-prisma: I think something like this might be a bit more readable

```ts
            await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query)).then(
              () => tx.transaction!.commit(),
              (err) => tx.transaction!.rollback().finally(() => Promise.reject(err)),
            )
```

jacek-prisma: updated

jacek-prisma: changed the wrong branch. corrected it now

jacek-prisma: extra change: changed from Promise.reject to plain `throw`, because`finally` doesn't support async callbacks

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -241,8 +241,13 @@ export class TransactionManager {
             await this.#withQuerySpanAndEvent(PHANTOM_COMMIT_QUERY(), tx.transaction, () => tx.transaction!.commit())
           } else {
             const query = COMMIT_QUERY()
-            await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
-            await tx.transaction.commit()
+            await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query)).then(
+              () => tx.transaction!.commit(),
+              async (err) => {
+                await tx.transaction!.rollback()
+                throw err
```
aqrln: If another error happens in `tx.transaction!.rollback()`, it will replace the original error, which is not ideal since the error in `executeRaw` is more important and salient.

We should either throw `err` even if `tx.transaction!.rollback()` fails:

```ts
try {
  await tx.transaction!.rollback()
} finally {
  throw err
}
```

or catch it and build a new error that contains both (e.g., maybe append the original error as the rollback error's [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause))

jacek-prisma: ok, changed it to `(err) => tx.transaction!.rollback().catch(() => Promise.reject(err)),`

aqrln: It needs to be `finally` and not `catch` because otherwise the whole operation will succeed if `rollback` doesn't throw, but we're already in the error path here and need to throw regardless

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
```diff
@@ -241,8 +241,10 @@ export class TransactionManager {
             await this.#withQuerySpanAndEvent(PHANTOM_COMMIT_QUERY(), tx.transaction, () => tx.transaction!.commit())
           } else {
             const query = COMMIT_QUERY()
-            await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
-            await tx.transaction.commit()
+            await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query)).then(
+              () => tx.transaction!.commit(),
+              (err) => tx.transaction!.rollback().catch(() => Promise.reject(err)),
```
aqrln: https://github.com/prisma/prisma/pull/28468#discussion_r2505151827

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
jacek-prisma: that's true, forgot about the non failing rollback case

the `throw` in a finally block gives a linting error, but I think something like this also does the job:
```ts
                const fail = () => Promise.reject(err)
                tx.transaction!.rollback().then(fail, fail)
```

jacek-prisma: updated

packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
jacek-prisma: updated

aqrln (approved):

PR #28474, opened 2025-11-07
refactor: update `checkUnsupportedDataProxy` to use config

aqrln:
Check datasource from config instead of schema.


wmadden (approved):
Looks good 👍🏻

PR #28475, opened 2025-11-07
refactor: don't use schema datasource in Studio

aqrln:
Load datasource URL from config file in Studio.


wmadden (approved):
This looks harmless to me

PR #28476, opened 2025-11-07
refactor: use config in `ensureDatabaseExists`

aqrln:
Use config datasource instead of schema datasource in `ensureDatabaseExists` and `ensureCannectToDatabase`.

### DIFF
--- a/packages/migrate/src/utils/ensureDatabaseExists.ts
+++ b/packages/migrate/src/utils/ensureDatabaseExists.ts
@@ -1,3 +1,6 @@
+import path from 'node:path'
+
+import { PrismaConfigInternal } from '@prisma/config'
 import { DataSource } from '@prisma/generator'
 import type { DatabaseCredentials } from '@prisma/internals'
 import {
@@ -8,7 +11,6 @@ import {
   uriToCredentials,
 } from '@prisma/internals'
 import { bold } from 'kleur/colors'
-import path from 'path'
 
 import { ConnectorType } from './printDatasources'
 import { getSocketFromDatabaseCredentials } from './unixSocket'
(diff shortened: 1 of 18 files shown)


packages/migrate/src/utils/ensureDatabaseExists.ts
wmadden, line 187: There are a few violations of SRP here 😅 

- `ensureDatabaseExists()` returns a formatted message or `undefined`
    - It should return an indicator of success or failure, formatting and printing a message should happen at the callsite
    - It may be preferable for the message construction logic to be reusable in which case it belongs in a separate function (which would take e.g. success/failure + datasource provider name as inputs)
- This function should not depend on DataSource
    - What it actually needs is the path resolution root, which is `datasource.sourceFilePath` (aka `schemaDir`)
    - That makes `ensureDatabaseExists()` responsible for selecting the path resolution root
    - The path resolution root ought to be passed in as a parameter, rather than `dataSource`

I would understand if you didn't want to make big changes here right now but at a minimum I would:
- accept `pathResolutionRoot` as an explicit parameter, use it instead of `datasource.sourceFilePath`
- add an explicit return type
- accept `providerName` and decouple from `DataSource`
```
type SuccessMessage = string;
export async function ensureDatabaseExists(
  pathResolutionRoot: string, 
  providerName: string, 
  config: PrismaConfigInternal
): Promise<SuccessMessage | undefined> { ... }
```

wmadden (approved):
Your changes look structurally fine but I left you a suggestion to decouple `ensureDatabaseExists()` from DataSource. I'm preemptively approving it to unblock you - and because the changes are trivial and don't need further review 👍🏻

PR #28477, opened 2025-11-07
refactor: use config in `parseDatasourceInfo`

aqrln:
Use datasource URL from config instead of from schema in `parseDatasourceInfo` function.


wmadden (approved):
Looks fine. Gross but fine 👍🏻

PR #28479, opened 2025-11-08
refactor: remove unused `inlineDatasources`

aqrln:
They aren't used since the introduction of `accelerateUrl`, and they can't really be used now that we are removing `url` from the schema.

It has been confirmed on Slack and in another PR that the comment about Policy is no longer relevant. Besides, it wouldn't work anyway even if we kept the field.

I thought we had already removed this, but it was either removed incompletely or inadvertently reintroduced when merging something.


wmadden (commented):
Good job 🧹

wmadden (approved):

PR #28481, opened 2025-11-08
fix(client): reference the correct `edge.js` file

aqrln:
`edge.mjs` doesn't exist, it should be `edge.js` for both `require` and `import`, like index, `default` and `extension` entrypoints do.

Well, it technically shouldn't, but let's make it consistent with the normal entrypoint for now, and change all of them later if needed.

Follow up to https://github.com/prisma/prisma/pull/28480

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -74,8 +74,8 @@
     "./edge": {
       "types": "./edge.d.ts",
       "require": "./edge.js",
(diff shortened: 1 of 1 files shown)


jkomyno (approved):

wmadden (approved):
I can't really validate whether this change is correct or not but structurally it looks fine 🤷🏻

PR #28484, opened 2025-11-08
fix(adapter-d1): split node/workerd entrypoints

aqrln:
`enum-import-in-edge` e2e test failed because importing `@prisma/adapter-d1` always pulled `node:fs`/`node:path` via the `listLocalDatabases` helper, which workerd doesn't support without a compatibility flag.

This change moves `listLocalDatabases` into a node-only module and adds separate `index-node`/`index-workerd` entrypoints wired through the exports map, build config, and TypeScript path alias so worker bundles stay free of `node:*` deps while Node consumers still get the helper.

### DIFF
--- a/packages/adapter-d1/src/index-node.ts
+++ b/packages/adapter-d1/src/index-node.ts
@@ -1,0 +1,2 @@
+export * from './index-workerd'
+export { listLocalDatabases } from './node/list-local-databases'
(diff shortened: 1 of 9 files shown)


packages/adapter-d1/src/index-node.ts
```diff
@@ -0,0 +1,4 @@
+export { PrismaD1 } from './d1'
+// exported for backwards compatibility
+export { PrismaD1HttpAdapterFactory as PrismaD1Http } from './d1-http'
```
aqrln:
```suggestion
export * from './index-workerd.ts'
```

jkomyno (approved):

PR #28486, opened 2025-11-08
refactor: remove `url` from `DataSource` structure

aqrln:
Remove `url` from the JS representation of PSL datasource.

### DIFF
--- a/packages/internals/src/cli/schemaContext.ts
+++ b/packages/internals/src/cli/schemaContext.ts
@@ -1,8 +1,8 @@
+import path from 'node:path'
+
 import { SchemaEngineConfigInternal } from '@prisma/config'
 import { ActiveConnectorType, DataSource, GeneratorConfig } from '@prisma/generator'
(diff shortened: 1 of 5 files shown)


packages/internals/src/cli/schemaContext.ts
```diff
@@ -113,22 +107,9 @@ export async function processSchemaResult({
 
   const configFromPsl = await getConfig({ datamodel: schemaResult.schemas })
```
jacek-prisma, line 108: isn't the `configFromPsl` name incorrect now?

jacek-prisma: might be a good idea to rename it

aqrln: No, it is correct. That's why this config doesn't have any URLs anymore, and the URL-related logic was removed from this function.

jacek-prisma (approved):

PR #28491, opened 2025-11-09
chore(examples): Change Examples folder to submodule

jy-joung-bitor:


aqrln:
I'm not convinced this is a good idea. A git submodule points at a specific commit and needs to be regularly updated and maintained. Submodules are normally used for things like dependencies, not just to link to another repo. Neither of these two repositories needs the other in order to build.

jy-joung-bitor:
Sorry about mistake

PR #28492, opened 2025-11-09
test(client): fix flaky `enum-import-in-edge` e2e test

aqrln:
## Summary

The `enum-import-in-edge` Cloudflare Workers regression test has been intermittently (and, lately, most of the time) red on CI with `write EPIPE` errors coming from Wrangler. The failure rate spiked whenever Wrangler happened to log anything after the test declared the dev server ready, causing the child process to crash. This PR keeps Wrangler’s stdout alive, waits for readiness without destroying the stream, and shuts the process down gracefully so the test is quiet and stable again.

## Investigation

CI logs (`packages/client/tests/e2e/enum-import-in-edge/LOGS.txt:60-87`) showed Wrangler starting successfully, the HTTP assertion passing, and then an immediate `write EPIPE` stack trace originating from `wrangler-dist/cli.js`. Re-running locally reproduced the same stack trace even on "green" runs, confirming we were exiting the child process in an unhealthy way.

`_steps.ts:13-38` captured Wrangler’s output via `for await (const line of wranglerProcess.stdout) { … break }`. Breaking out of an async iterator that reads from a stream destroys the underlying stream in Node.js. Wrangler (and my reproduction script using `child_process.spawn`) always crashed with `EPIPE` after the parent broke the loop while the child was still writing.

After tearing down stdout, the test immediately called `wranglerProcess.kill()`. Wrangler thus lost both its stdout pipe *and* received SIGTERM, which explains why the log noise happened even when the HTTP request already succeeded and why reruns sometimes passed: whether `curl` completed before the crash was a race.

## Fix

- Replace the `for await` loop with a dedicated `waitForWranglerReady()` helper that uses `readline` to watch stdout for the “Ready on …” banner without closing the stream. Once readiness is detected the helper resumes the stream so Node.js continues to drain data silently.
- Add a 30 s timeout plus exit/error guards so we still fail fast if Wrangler never becomes ready or exits early.
- Introduce `stopProcess()` cleanup that sends `SIGINT`, waits for the process to exit, and swallows races if it was already gone. This removes the redundant kill calls and prevents noisy stack traces.
- Confirmed with a standalone `child_process.spawn` script that the new watcher no longer causes child processes to throw `EPIPE` even when they keep logging.

## Verification

`pnpm --filter @prisma/client test:e2e --runInBand --testPathPattern enum-import-in-edge` passes and results in clean `LOGS.txt` without errors.

## Follow-ups

1. Audit other e2e fixtures that watch long-running services’ stdout and ensure they do not break out of `for await` loops prematurely.
2. Extract `waitForServiceReady` into `packages/client/tests/e2e/_utils/` if we need it in more places.

### DIFF
--- a/packages/client/tests/e2e/enum-import-in-edge/_steps.ts
+++ b/packages/client/tests/e2e/enum-import-in-edge/_steps.ts
@@ -23,7 +23,7 @@ void executeSteps({
 
         return await $`curl http://localhost:8787/ -s`
       } finally {
-        await stopProcess(wranglerProcess)
+        await stopProcess(wranglerProcess).catch(() => {})
       }
     }, 3)
 
@@ -81,15 +81,6 @@ async function waitForWranglerReady(processPromise: ProcessPromise) {
 }
 
 async function stopProcess(processPromise: ProcessPromise) {
-  try {
-    await processPromise.kill('SIGINT')
-  } catch {
-    // ignore – process might have already exited
-  }
-
-  try {
-    await processPromise
-  } catch {
-    // swallow to keep cleanup best-effort
-  }
+  await processPromise.kill('SIGINT')
+  await processPromise
 }
--- a/packages/client/tests/e2e/enum-import-in-edge/_steps.ts
+++ b/packages/client/tests/e2e/enum-import-in-edge/_steps.ts
@@ -1,5 +1,7 @@
+import { createInterface } from 'node:readline/promises'
 import timers from 'node:timers/promises'
 
+import type { ProcessPromise } from 'zx'
 import { $ } from 'zx'
 
 import { executeSteps } from '../_utils/executeSteps'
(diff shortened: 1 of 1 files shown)


packages/client/tests/e2e/enum-import-in-edge/_steps.ts
```diff
@@ -41,3 +40,56 @@ void executeSteps({
...
+    await processPromise.kill('SIGINT')
+  } catch {
+    // ignore – process might have already exited
+  }
+
+  try {
+    await processPromise
+  } catch {
+    // swallow to keep cleanup best-effort
+  }
```
jacek-prisma: I'd probably just add `.catch(() => {})` on the callsite, that way the error handling logic is in one function instead of being scattered and we don't want to handle any stopProcess errors anyway

jacek-prisma (approved):
left 1 minor non blocking comment

PR #28501, opened 2025-11-10
fix(ci): update FORCE_INTEGRATION_RELEASE input handling

Aryaman1792:
This PR fixes a minor bug in the `.github/workflows/release.yml` workflow:
- The environment variable `FORCE_INTEGRATION_RELEASE` was referencing `inputs.forceIntegrationRelease` instead of `github.event.inputs.forceIntegrationRelease`. This led to the value always being empty or false when the workflow was dispatched.
- This PR corrects the context so that workflow dispatch input for `forceIntegrationRelease` is properly passed to jobs and scripts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,8 @@ env:
   SKIP_PACKAGES: ${{ github.event.inputs.skipPackages }}
   ONLY_PACKAGES: ${{ github.event.inputs.onlyPackages }}
   DRY_RUN: ${{ github.event.inputs.dryRun || 'false' }}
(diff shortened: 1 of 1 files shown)


aqrln (approved):
Looks good to me, thanks so much for catching this.

Aryaman1792:
Hi @aqrln 👋

Thanks for reviewing and approving! Could you please merge this when convenient?
The branch is out of date with main, but it can be cleanly merged. I’ve confirmed that all checks have passed and the fix works as expected.

Appreciate your help! 🙏

Aryaman1792:
Hi @aqrln — looks like the Client (memory tests) [v24] job failed, but it seems unrelated to this workflow change.
Could you please re-run that check when you get a chance? Everything else passed successfully. 🙏

PR #28507, opened 2025-11-11
fix(client): preserve schema JSDoc for select/include and resolve TS2536

GoldFoylee:
## Summary

**Fixes a TypeScript issue in Prisma Client where schema comments (JSDoc) were missing from fields queried with select or include. Also resolves TS2536: Type 'K' cannot be used to index type 'S & I' by introducing a safe indexing helper.**

### Root Cause

The mapped type in GetFindResult previously iterated over keyof S | keyof I and directly indexed (S & I)[K].

- This lost access to $ModelPayload['scalars'], dropping JSDoc metadata.
- TypeScript 5.8 introduced stricter checks, making direct indexing invalid.

**Fix**
- Iterate over `keyof P['scalars']` instead of `keyof S | keyof I`, ensuring payload-based metadata is preserved.
- Introduce a helper to safely access intersected keys:
```
type SafeKeyAccess<S, I, K extends PropertyKey> =
  K extends keyof (S & I) ? (S & I)[K] : undefined;

```
- Replace all (S & I)[K] references with SafeKeyAccess<S, I, K>.

**Result**
- IntelliSense now consistently shows schema comments for all selected fields.
- All packages build cleanly (pnpm -r run build).
- No runtime or API surface changes — type-only fix.

### Example

**Before:**

`user2.username // (property) username: string | undefined`

**After:**

`user2.username // (property) username: string | undefined — "user's name"`

**Verification**
- [x] Issue reproduced and validated in sandbox.
- [x] Monorepo builds successfully.
- [x] IntelliSense verified for both findUnique and findUnique(select).


tensordreams:
Hi @GoldFoylee — thanks for this, and apologies for the slow reply.

We are closing this one: changing the mapped key source from `keyof S | keyof I` to `keyof P['scalars']` drops every relation field from `select`/`include` results, which is a wider regression than the JSDoc and TS2536 issues it targets.

The underlying problem is real, and a fix that keeps relation fields intact would be welcome.

PR #28508, opened 2025-11-11
chore: update readme

AmanVarshney01:


README.md
```diff
@@ -71,7 +71,8 @@ datasource db {
 
 // Generator
 generator client {
-  provider = "prisma-client-js"
+  provider = "prisma-client"
+  output   = "./generated"
```
aqrln: Output is relative to the schema file, so, assuming the schema file is in `prisma/schema.prisma`, this should be
```suggestion
  output   = "../generated"
```
for imports below to work

README.md
```diff
@@ -177,46 +125,106 @@ Once the data model is defined, you can [generate Prisma Client](https://www.pri
...
 ### Accessing your database with Prisma Client
 
-#### Generating Prisma Client
+#### Step 1: Install Prisma
 
-The first step when using Prisma Client is installing its npm package:
+First, install Prisma CLI as a development dependency and Prisma Client:
 
 ```
+npm install prisma --save-dev
```
aqrln, line 133: did we really not have it here before lol
incredible

README.md
```diff
@@ -177,46 +125,106 @@ Once the data model is defined, you can [generate Prisma Client](https://www.pri
...
+  migrations: {
+    path: "prisma/migrations",
+  },
+  datasource: {
+    url: env<Env>("DATABASE_URL"),
+  },
+});
+```
+
+**Note**: Environment variables from `.env` files are not automatically loaded when using `prisma.config.ts`. You can use `dotenv` by importing `dotenv/config` at the top of your config file, or when using `tsx`, you can pass a `--env-file` flag to load environment variables. For Bun, `.env` files are automatically loaded.
```
aqrln: > when using `tsx`, you can pass a `--env-file` flag to load environment variables

I don't think this would help for `prisma.config.ts`. For an application using Prisma Client, if you want to use env vars there, sure. But you're not running Prisma CLI with tsx, and `prisma.config.ts` is only used by Prisma CLI.

README.md
```diff
@@ -177,46 +125,106 @@ Once the data model is defined, you can [generate Prisma Client](https://www.pri
...
 ```
 
 or
 
 ```js
-const { PrismaClient } = require('@prisma/client')
+const { PrismaClient } = require('./generated/client')
 
 ```
```
aqrln: I'm not sure if we should still have this in 2025. If anyone still uses CommonJS (with raw CommonJS syntax and not fake ESM like many people still do), they'll figure out how to translate the import statement to a require call. I'd personally just delete this but I'm fine with it either way.

README.md
```diff
@@ -177,46 +125,106 @@ Once the data model is defined, you can [generate Prisma Client](https://www.pri
...
 
+**Note**: Depending on your database, you may need to use a [driver adapter](https://www.prisma.io/docs/orm/overview/databases/database-drivers#driver-adapters). For example, when using PostgreSQL with a driver adapter:
+
+```ts
+import { PrismaClient } from './generated/client'
+import { PrismaPg } from '@prisma/adapter-pg'
+
+const adapter = new PrismaPg({ connectionString: process.env.DATABASE_URL })
```
aqrln, line 216: Should we also suggest `dotenv`/`tsx --env-file`/`node --env-file`/Bun here?

README.md
AmanVarshney01: yep i agree to remove this, ESM is the way!!!

README.md
```diff
@@ -177,46 +125,100 @@ Once the data model is defined, you can [generate Prisma Client](https://www.pri
...
+```ts
+import { PrismaClient } from './generated/client'
+import { PrismaPg } from '@prisma/adapter-pg'
 
-const prisma = new PrismaClient()
+const adapter = new PrismaPg({ connectionString: process.env.DATABASE_URL })
+const prisma = new PrismaClient({ adapter })
 ```
 
+To load environment variables, you can use `dotenv` by importing `dotenv/config`, use `tsx --env-file=.env`, `node --env-file=.env` (Node.js 20+), or Bun (which loads `.env` automatically).
```
aqrln: No need to specify "(Node.js 20+)" since we don't support any older Node.js versions

aqrln (approved):

PR #28511, opened 2025-11-11
feat(cli): Studio + SQLite baaaaaaaby!

igalklebanov:
Hey 👋 

This PR adds SQLite support in `prisma studio` command.

It tries `node:sqlite` and if doesn't exist (Node <22 users), falls back to `better-sqlite3` - an optional peer dependency.
Uses the new Studio `node:sqlite` executor and SQLite adapter.

Sandbox has been enhanced to enable manual testing of multiple providers.


packages/cli/package.json
```diff
@@ -168,13 +170,17 @@
   "dependencies": {
     "@prisma/config": "workspace:*",
     "@prisma/engines": "workspace:*",
-    "@prisma/studio-core-licensed": "0.6.0",
+    "@prisma/studio-core-licensed": "0.0.0-dev.202511120235",
```
igalklebanov: stalling on releasing v0.7 until SQLite adapter is finished - currently only does introspection.

follow up PRs will bump the version.

packages/cli/src/Studio.ts
```diff
@@ -287,6 +327,8 @@ const INDEX_HTML =
       });
 
       const onEvent = (event) => {
+        console.log(event);
```
jkomyno: Still needed?

packages/cli/src/Studio.ts
```diff
@@ -212,8 +249,11 @@ ${bold('Examples')}
       return ctx.json([null, results])
     })
 
-    app.post('/telemetry', (ctx) => {
+    app.post('/telemetry', async (ctx) => {
       // TODO: ...
+
+      console.log(await ctx.req.json())
+
```
jkomyno, line 258: Are you sure this is still needed?

packages/cli/src/Studio.ts
```diff
@@ -152,7 +186,7 @@ ${bold('Examples')}
       return this.help()
     }
 
-    const connectionString = args['--url'] || (config as { datasource?: { url?: string } }).datasource?.url
+    const connectionString = args['--url'] || config.datasource?.url
 
     if (!connectionString) {
       return new Error('No database URL found.')
```
jkomyno: Can we modify the error message so that people are reminded of how to specify a URL for Studio, like we do for `migrate dev`? Can be done in a follow-up PR.

jkomyno (approved):
Approved with a couple of comments

packages/cli/src/Studio.ts
igalklebanov: done.

packages/cli/src/Studio.ts
igalklebanov: yeah, i wanna keep em around til telemetry is implemented.

packages/cli/src/Studio.ts
igalklebanov: removed.

PR #28516, opened 2025-11-13
fix(fetch-engine): skip native binary downloads when using Rust-free/…

gothamsidd:
Fixes prisma/prisma#28503.

When `PRISMA_CLI_QUERY_ENGINE_TYPE=wasm` or `PRISMA_USE_WASM=true`, the CLI should
use the Rust-free engine and must not attempt to resolve or download native engine
binaries.

This PR adds an early return in `packages/fetch-engine/src/download.ts` to skip all
binary resolution, validation, and download logic when WASM mode is enabled.

This allows `prisma generate` to work properly in Rust-free mode on systems such as
NixOS where native engines are not available.

### DIFF
--- a/packages/fetch-engine/src/download.ts
+++ b/packages/fetch-engine/src/download.ts
@@ -70,6 +70,16 @@ export async function download(options: DownloadOptions): Promise<BinaryPaths> {
   if (!options.binaries || Object.values(options.binaries).length === 0) {
     return {} // we don't download anything if nothing is provided
   }
+  // Skip native engine binaries when running with the Rust-free WASM engine
+const useRustFreeEngine =
+process.env.PRISMA_CLI_QUERY_ENGINE_TYPE === 'wasm' ||
+process.env.PRISMA_USE_WASM === 'true';
+
+if (useRustFreeEngine) {
+debug('Rust-free/WASM engine enabled — skipping native engine downloads.');
+return {};
+}
+
 
   if (enginesOverride?.['branch'] || enginesOverride?.['folder']) {
     // if this is true the engines have been fetched before and already cached


aqrln (changes requested):
Thanks for the enthusiasm and the contribution, but there's a number of issues with this PR:

* This attempts to work around the problem instead of solving it. Making sure that unnecessary engines aren't downloaded doesn't require introducing any new environment variables, we have all the information we need in the CLI. We should fix whatever is wrong in the existing code and not add extra complexity and hacks on top of it.

* `PRISMA_CLI_QUERY_ENGINE_TYPE` env var is a legacy that doesn't make much sense anymore because the CLI never uses the query engine, and hasn't in a long time even before the introduction of the native binary free client. `wasm` is also not a valid value for it, and it might break other code that still reads it for whatever legacy reason. Besides, this variable has been removed completely on the `next` branch together with other query engine related stuff.

* Code formatting is broken.

* Variable names must use precise and correct terms and not the "Rust-free" marketing moniker inconsistent with the terminology used in the codebase. `useWasmEngine`/`useWasmEngines` is better but also not ideal: while it would describe the experimental schema-engine-wasm and the legacy query-engine-wasm well, part of the new Client story (and the most important one at that) is the change in _architecture_, not languages or compilation targets. There's no single "query engine" in the new query compiler based client, Wasm or not, and many parts that used to belong to the query engine are now in pure TypeScript and not in Wasm.

* The usefulness of this environment variable is somewhat limited, since it will break all commands that rely on the schema engine (which is still native-only, even in Prisma 7, and has no full-featured Wasm equivalent yet), which is all Migrate commands and `prisma generate --sql`.

PR #28520, opened 2025-11-13
test(cli): fix broken `incomplete-schemas` test suite

aqrln:
The tests were broken with incorrect snapshots that expect incorrect errors because the paths to the schema files were wrong in the configs. This was recently broken in one of the previous pull requests and missed during code review.

This commit fixes the paths in the test fixtures and updates the snapshots to restore the correct error messages, with many of them being different from before because the error being test is now in the config file and not in the schema itself. Because of that, I also needed to update the snapshot serializer to support the temporary directory paths on macOS and Windows, since some of those error messages now contain the full path to the config file directory.

Discovered when working on https://linear.app/prisma-company/issue/TML-1580/resolve-sqlite-and-client-certs-paths-relative-to-configcwd-and-not-to

### DIFF
--- a/packages/cli/src/__tests__/incomplete-schemas.test.ts
+++ b/packages/cli/src/__tests__/incomplete-schemas.test.ts
@@ -2,28 +2,15 @@
 /* eslint-disable jest/no-identical-title */
 import { stripVTControlCharacters } from 'node:util'
 
-import { defaultTestConfig, loadConfigFromFile } from '@prisma/config'
 import { jestContext } from '@prisma/get-platform'
 import { DbExecute, DbPull, DbPush, MigrateDev, MigrateReset } from '@prisma/migrate'
 import fs from 'fs'
 
 import { Format } from '../Format'
 import { Validate } from '../Validate'
+import { configContextContributor } from './_utils/config-context'
 
-const ctx = jestContext.new().assemble()
-
-async function loadFixtureConfig(configFile?: string) {
-  const { config, error } = await loadConfigFromFile({ configFile })
-
-  if (error) {
-    if ('error' in error && error.error instanceof Error) {
-      throw error.error
-    }
-    throw new Error(`Failed to load Prisma config: ${error._tag}`)
-  }
-
-  return config ?? defaultTestConfig()
-}
+const ctx = jestContext.new().add(configContextContributor()).assemble()
 
 /**
  * Commands:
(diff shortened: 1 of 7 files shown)


packages/cli/src/__tests__/incomplete-schemas.test.ts
```diff
@@ -49,33 +36,10 @@ describe('[wasm] incomplete-schemas', () => {
       ctx.fixture('incomplete-schemas/datasource-block-url-env-set-invalid')
     })
 
-    it('validate', async () => {
```
aqrln, line 52: Schema validation doesn't care about URLs anymore

packages/cli/src/__tests__/incomplete-schemas.test.ts
```diff
@@ -137,32 +106,32 @@ describe('[wasm] incomplete-schemas', () => {
     it('validate', async () => {
       expect.assertions(1)
       try {
-        await Validate.new().parse([], await loadFixtureConfig())
+        await Validate.new().parse([], await ctx.config())
       } catch (e) {
         expect(stripVTControlCharacters(e.message)).toMatchInlineSnapshot(
-          `"Failed to load Prisma config: ConfigLoadError"`,
+          `"Failed to load config file "/tmp/dir" as a TypeScript/JavaScript module. Error: PrismaConfigEnvError: Missing required environment variable: SOME_UNDEFINED_DB"`,
```
aqrln, line 112: `load config file "/tmp/dir"` may look wrong — and indeed it is — but the snapshot is correct, there's nothing wrong with the test setup or snapshot serialization. The implementation of `loadConfig` does actually print the config directory and not the file path in this error message. We should fix it but it's out of scope of this PR.

packages/cli/src/__tests__/incomplete-schemas.test.ts
```diff
@@ -359,58 +284,45 @@ describe('[normalized library/binary] incomplete-schemas', () => {
...
-
-      try {
-        const config = await loadFixtureConfig()
-        await DbExecute.new().parse(['--file=./script.sql'], config)
-      } catch (e) {
-        expect(stripVTControlCharacters(e.message)).toMatchInlineSnapshot(
-          `"Failed to load Prisma config: ConfigLoadError"`,
-        )
-      }
+      await expect(DbExecute.new().parse(['--file=./script.sql'], await ctx.config())).resolves.not.toThrow()
```
jacek-prisma: this test case seems pointless since it no longer uses the schema

jacek-prisma (approved):
lgtm with 1 minor comment

PR #28522, opened 2025-11-13
fix(ci): add missing github-token input to github-script steps

Aryaman1792:
This fixes a workflow failure caused by missing github-token inputs in actions/github-script@v7 steps inside .github/workflows/update-engines-version.yml.

Changes made:

Added github-token: ${{ secrets.GITHUB_TOKEN }} to:

Extract Engine Commit hash from version

Extract patch branch name from version

This ensures the GitHub Script steps can authenticate properly with the repository context and prevents the “Input required and not supplied: github-token” error.


tensordreams:
Hi @Aryaman1792 — thanks for this, and apologies for the slow reply.

We are closing this one: both scripts are pure string manipulation and make no API calls, and `actions/github-script` already defaults `github-token` to `${{ github.token }}` — so this does not change behaviour.

PR #28528, opened 2025-11-14
feat!: remove `datasources` from PrismaClient options

aqrln:
Remove the datasources option used to override the datasource used by the query engine. The logic that uses it had been previously removed, but the user-facing option (which currently does nothing) wasn't.

Closes: https://linear.app/prisma-company/issue/TML-1589/remove-datasources-prismaclient-constructor-option


jacek-prisma (approved):

jkomyno:
Oops, good catch.

PR #28531, opened 2025-11-14
fix(generator-js): fix potentially missing `adapter` option for edge builds

aqrln:
The edge build of the old JS client could be missing the `adapter` option in the client constructor. It actually wasn't because at the moment we only generate the types for the default `client` runtime and write stubs that re-export types from it for other targets, but it could be easily changed and break when refactoring code. This code was wrong locally even if the offending code path wasn't used in practice.

### DIFF
--- a/packages/client-generator-js/src/TSClient/PrismaClient.ts
+++ b/packages/client-generator-js/src/TSClient/PrismaClient.ts
@@ -620,7 +620,7 @@ export type TransactionClient = Omit<Prisma.DefaultPrismaClient, runtime.ITXClie
     )
 
     if (
-      ['library.js', 'client.js'].includes(this.runtimeNameTs) &&
+      ['client.js', 'wasm-compiler-edge.js'].includes(this.runtimeNameTs) &&
       // We don't support a custom adapter with MongoDB for now.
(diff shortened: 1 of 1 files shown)


packages/client-generator-js/src/TSClient/PrismaClient.ts
```diff
@@ -620,7 +620,7 @@ export type TransactionClient = Omit<Prisma.DefaultPrismaClient, runtime.ITXClie
     )
 
     if (
-      ['library.js', 'client.js'].includes(this.runtimeNameTs) &&
+      ['client.js', 'wasm-compiler-edge.js'].includes(this.runtimeNameTs) &&
```
jacek-prisma, line 623: shouldn't this be
```
['library.js', 'client.js', 'wasm-compiler-edge.js']
```

aqrln: `library.js` doesn't exist anymore

jacek-prisma: oh I didn't notice it's on the next branch

jacek-prisma (approved):

PR #28533, opened 2025-11-14
fix(client): remove and fix remains of QE runtime bundles

aqrln:
- Fix incorrect references to `library`/`binary` runtime bundles in various places and replace them with the correct QC bundles.

- Remove dead code related to QE runtime bundles in generators.

- Fix broken `wasm-compiler-edge.d.ts` file in the `@prisma/client` package (it re-exported everything from `./library`).


packages/client-generator-js/src/TSClient/PrismaClient.ts
```diff
@@ -612,7 +604,6 @@ export type TransactionClient = Omit<Prisma.DefaultPrismaClient, runtime.ITXClie
     )
 
     if (
-      ['client.js', 'wasm-compiler-edge.js'].includes(this.runtimeNameTs) &&
```
jkomyno, line 615: Why?

jkomyno (approved):

packages/client-generator-js/src/TSClient/PrismaClient.ts
aqrln: For symmetry with `client-generator-ts` which doesn't have this check. It doesn't really make sense to have it anymore IMO: previously there were runtime bundles that didn't support driver adapters, now all of them do.

PR #28538, opened 2025-11-17
feat(cli): re-introduce init as an eager command.

igalklebanov:
Hey 👋 

We're moving `prisma init` back to being an eager command (not a sub-command). This is to freeze Prisma ORM <= 6 users to the last `@prisma/cli-init` version and to also reduce spin up time significantly for Prisma ORM >=7 users.

This is in preparation for the changes we want to do with the command, some of which are breaking.

This effort is a copy paste from the latest state of `@prisma/cli-init` code, with some dependency installs, and revert back to `jest`.


packages/cli/src/Init.ts
```diff
@@ -0,0 +1,724 @@
...
+    })
+
+    if (internals.isError(args) || args['--help']) {
+      return this.help()
+    }
+
+    const urlArg = args['--url']
+
+    if (urlArg) {
+      internals.checkUnsupportedDataProxy({
```
igalklebanov: this function changed it seems, no longer accepts `urls` so I guess-worked around the changes.

packages/cli/src/Init.ts
```diff
@@ -0,0 +1,736 @@
...
+${
+  previewFeatures.length > 0
+    ? `  previewFeatures = [${previewFeatures.map((feature) => `"${feature}"`).join(', ')}]\n`
+    : ''
+}  output   = "${output}"
+}
+
+datasource db {
+  provider = "${datasourceProvider}"
+  url      = env("DATABASE_URL")
```
aqrln:
```suggestion
```

packages/cli/src/Init.ts
aqrln: this looks good

aqrln (approved):

PR #28545, opened 2025-11-18
fix: clarify the message about supported Node.js versions

aqrln:
* The error message states 'Node.js versions 20.19, 22.12, 24.0' but the test cases check versions like '20.19.5', '22.12.0', '22.20.0', '24.0.1', '24.10.0'. The message should clarify whether it means minimum versions (e.g., '20.19+, 22.12+, 24.0+') or be more specific about supported version ranges.

* If the current version is higher than supported (i.e. the major version is 25+), it shouldn't prompt the user to "upgrade" their version. It should also be a warning and not a hard error because while it's not officially supported, it will likely work and there can be legitimate reasons to use it. We've never prevented users from using newer than supported versions before and even accepted community pull requests to improve compatibility with them.

Closes: https://linear.app/prisma-company/issue/TML-1600/clarify-the-message-about-supported-nodejs-versions

### DIFF
--- a/packages/cli/scripts/preinstall.ts
+++ b/packages/cli/scripts/preinstall.ts
@@ -24,25 +24,40 @@ export function printMessageAndExitIfUnsupportedNodeVersion(nodeVersion: MajorMi
   const [nodeMajorVersion, nodeMinorVersion] = extractSemanticVersionParts(nodeVersion)
 
   // Minimum Node.js versions supported by Prisma
-  const MIN_NODE_VERSION_MATRIX: Record<string, number> = {
-    '20': 19,
-    '22': 12,
-    '24': 0,
+  const MIN_NODE_VERSION_MATRIX: Record<number, number> = {
+    20: 19,
+    22: 12,
+    24: 0,
   }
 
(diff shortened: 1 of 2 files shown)


jkomyno (dismissed):

jacek-prisma (dismissed):

packages/cli/scripts/preinstall.ts
```diff
@@ -24,25 +24,39 @@ export function printMessageAndExitIfUnsupportedNodeVersion(nodeVersion: MajorMi
...
       .join(', ')
+    const highestSupportedMajor = Math.max(...Object.keys(MIN_NODE_VERSION_MATRIX).map((major) => Number(major)))
+    const isNodeVersionTooNew = nodeMajorVersion > highestSupportedMajor
+
+    const messageLines = [
+      `Prisma only supports Node.js versions ${supportedVersions}.`,
+      isNodeVersionTooNew ? 'Please use a supported Node.js version.' : 'Please upgrade your Node.js version.',
+    ]
 
     console.error(
```
jacek-prisma: maybe should be a warning instead if `isNodeVersionTooNew`

jkomyno (approved):

PR #28551, opened 2025-11-18
feat(cli): update `prisma init --db` output

aqrln:
Update `prisma init --db` to output the direct connection string to the database.

Since there's no way to obtain a direct TCP URL in the legacy GraphQL Platform API, this also updates `prisma init --db` to use the Management API instead. It also introduces a new OAuth login flow for the Management API, so users won't be affected by the token expiration issue that exists in the Platform CLI anymore.

The output of the command is updated to match https://www.notion.so/prismaio/Refreshed-prisma-init-29a9e8aecef780e1838eff4627741d19?source=copy_link#2a99e8aecef780759e8ec893017c7c2a

Closes: https://linear.app/prisma-company/issue/TML-1593/update-prisma-init-db-to-print-direct-tcp-url


packages/cli/src/management-api/auth.ts
aqrln, line 1: Most of this is copied from https://github.com/prisma/language-tools/blob/ce55bef3c68618e59c42f4e17dd97b9c71d63bb4/packages/vscode/src/plugins/prisma-postgres-manager/management-api/auth.ts

packages/cli/src/management-api/auth.ts
```diff
@@ -0,0 +1,182 @@
+import crypto from 'node:crypto'
+import events from 'node:events'
+import http from 'node:http'
+import { AddressInfo } from 'node:net'
+
+import open from 'open'
+import z from 'zod'
+
+const CLIENT_ID = 'cmi4ttoor03pv2wco4526rnin'
```
aqrln, line 9: Please remind me to move the OAuth app credentials to a shared 1Password vault

packages/cli/src/management-api/client.ts
aqrln, line 1: copied from https://github.com/prisma/language-tools/blob/ce55bef3c68618e59c42f4e17dd97b9c71d63bb4/packages/vscode/src/plugins/prisma-postgres-manager/management-api/client.ts

packages/cli/src/Init.ts
```diff
@@ -478,39 +477,29 @@ export class Init implements Command {
...
+          throw new Error('Missing connection string in response')
+        }
 
-        const serviceToken = await PlatformCommands.ServiceToken.createOrThrow({
-          token: platformToken,
-          environmentId: project.defaultEnvironment.id,
-          displayName: `database-setup-prismaPostgres-api-key`,
-        })
+        const { host, user, pass } = project.database.directConnection
+        prismaPostgresDatabaseUrl = `postgres://${user}:${pass}@${host}`
```
nurul3101: We should add the database name and sslmode parameter to make sure the connection string format remains same as what is returned by Prisma Console, Create DB CLI.

```suggestion
        prismaPostgresDatabaseUrl = `postgres://${user}:${pass}@${host}/postgres?sslmode=require`
```

jkomyno (approved):

PR #28552, opened 2025-11-18
fix: fix sqlite node 20 studio issues

jacek-prisma:
[TML-1603](https://linear.app/prisma-company/issue/TML-1603/bug-node-20-users-cant-use-even-when-better-sqlite3-installed)

Node 20 is supported by running studio via: `npx -p better-sqlite3 -p prisma@dev prisma studio`. Better-sqlite3 is not available by default and needs to be added explicitly.


igalklebanov (approved):
LGTM! 🚀

PR #28567, opened 2025-11-19
feat(fetch-engine): add mirror fallback and strict flag

Meet00028:
Summary
This PR enhances engine downloads by introducing a fallback mechanism from custom mirrors to the official Prisma CDN if the mirror download fails.
We introduce a new environment variable, PRISMA_ENGINES_STRICT_MIRROR, to disable this fallback and enforce strict mirror usage. This variable is now surfaced in $ prisma debug.

Testing
- pnpm --filter @prisma/fetch-engine test download
- pnpm --filter prisma test -- DebugInfo


Meet00028:
Windows All pkgs job exits with -1073741502; macOS/Linux + targeted suites pass locally. Looks like a runner issue—could someone rerun/approve?

aqrln:
Could you please tell us more about the motivating use case and how often you run into the issue this would solve? I'd like to avoid adding even more complexity and env vars in this area if possible, but I'm not strictly opposed to it if it brings real value.

Meet00028:
I’ve been seeing this happen whenever a hosting platform (like Vercel in #28564) injects PRISMA_ENGINES_MIRROR automatically. In those environments, the mirror points to a private S3 bucket, so every prisma generate fails with a 403 error. Users usually don’t realize the variable is set, so they’re stuck manually unsetting it or downgrading, which is pretty painful during CI.
I implemented a fallback that lets these deployments recover automatically by stepping down to binaries.prisma.sh. I also added a strict mode (PRISMA_ENGINES_STRICT_MIRROR=1) so teams that intentionally self-host or run offline can keep the 'fail-fast' behavior.
I considered just adding warnings, but that doesn’t help when you don’t control the platform environment variables. This change prevents those accidental regressions while preserving advanced mirror use cases.

PR #28574, opened 2025-11-19
sandbox: build CLI, copy env, fix imports

### LINKED ISSUE #7314: Missing import in docs (minor amends)
https://www.prisma.io/docs/orm/reference/prisma-config-reference

In the code section below "Using TypeScript's satisfies operator with the PrismaConfig type:"
you're missing `import { env } from "prisma/config";` for `env("DATABASE_URL")`.

Also, `import path from "node:path";` only works if `esModuleInterop` flag is set to `true`. So you might as well write `import * as path from "node:path";` which doesn't require the flag.

yureiblack:
**The original error was:** ```ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL```
 Command ```"prisma"``` not found when running ```pnpm prisma``` generate in the ```prisma/sandbox/query-compiler```directory. This indicated that the prisma CLI command was not available in the *PATH* or not properly linked by pnpm.
 Closes prisma/docs#7314 
### Approach 

- Ran `pnpm build` in root to compile all packages, including CLI.

- Retried `pnpm run generate`, still failed due to pnpm linking issues.

- Ran CLI directly with `node ../../packages/cli/build/index.js generate`, which worked but required `DATABASE_URL`.

- Checked `.db.env` for environment variables.

### Thought Process

- Root cause: CLI not built in monorepo setup.

- Pnpm didn't link binary correctly after build.

- Environment variables needed for config loading.

- No code changes required, just build and setup.

### Outcome
CLI is now built and functional. Use `dotenv -e ../../.db.env -- pnpm run generate` to run with environment variables.

### DIFF
--- a/sandbox/query-compiler/package.json
+++ b/sandbox/query-compiler/package.json
@@ -17,6 +17,7 @@
   "dependencies": {
     "@prisma/client": "../../packages/client",
     "@prisma/adapter-pg": "../../packages/adapter-pg",
+    "@prisma/config": "workspace:*",
     "pg": "../../packages/adapter-pg/node_modules/pg"
   },
   "devDependencies": {
--- a/sandbox/query-compiler/prisma.config.ts
+++ b/sandbox/query-compiler/prisma.config.ts
@@ -1,7 +1,8 @@
-import { defineConfig, env } from '@prisma/config'
+import { defineConfig, env } from 'prisma/config'
+import type { PrismaConfig } from 'prisma/config'
 
 export default defineConfig({
   datasource: {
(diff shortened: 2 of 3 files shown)


tensordreams:
Hi @yureiblack — thanks for the contribution, and sorry it sat for so long.

We're closing out the open PRs that are still blocked on an unsigned CLA, and this is one of them: https://cla-assistant.io/prisma/prisma?pullRequest=28574

If you'd like to pick it back up, signing there and reopening the PR is all it takes.

Thanks for taking the time either way.

PR #28596, opened 2025-11-20
Fix TypeScript syntax errors in EngineValidationError.ts

yureiblack:
# 🔧 Fix: Mapped Enum Handling in Prisma Create Operations

This PR fixes an issue where **mapped enums** in Prisma `create` operations caused `PrismaClientValidationError` exceptions.

## ❗ Problem
When using mapped enums, the Prisma Client was incorrectly generating **database-mapped enum strings** (e.g., `"pending"`) instead of the **enum key names** (e.g., `"PENDING"`).  
However, Prisma **7.0.0** validates incoming enum values using the **enum key names**, causing a mismatch and resulting in errors.

### 1. Correct Enum Value Generation
Updated logic in  
`prisma/packages/dmmf/src/convert.ts`  
so that the Prisma Client now generates **enum key names** instead of the mapped database values.

This ensures that the values passed to the validation layer are correct and consistent.

### 2. Added `InvalidEnumValueError` Support
Integrated support for `InvalidEnumValueError` into the error rendering system to provide clearer, more informative messages when invalid enum values are supplied.

## 🎯 Result
With these changes:

- Prisma Client now correctly passes enum key names to the engine.
- The engine continues mapping them to the appropriate database values.
- Validation issues are resolved.
- All existing functionality remains intact, as confirmed by the passing test suite.

## 🚀 Summary
This PR restores proper behavior for mapped enums in Prisma `create` operations and improves the developer experience through better error messaging.


yureiblack:
Kindly review the fix @devly-tal

tensordreams:
Hi @yureiblack — thanks for this, and apologies for the slow reply.

We are closing this one: this adds handling for an `InvalidEnumValue` error kind that the Wasm query compiler never emits, so the branch would be unreachable.

PR #28600, opened 2025-11-20
fix: allow any node version higher than 24

Sajito:
The current configuration prevents consumers from using higher node versions, while having strict-engine checks enabled.
Higher node versions might not be officially supported, but I think they shouldn't be blocked, unless they are known to now work.

Resolves #28594


aqrln (approved):
Thank you!

jacek-prisma (approved):

PR #28615, opened 2025-11-21
chore: update sandboxes for Prisma 7

FGoessler:
This PR cleans up some of our sandboxes so the `basic-sqlite` and `basic-postgres` ones are ready to use without any additional changes and use latest Prisma 7 config and generation.


sandbox/basic-postgres/package.json
```diff
@@ -14,14 +14,17 @@
...
+    "@prisma/adapter-pg": "../../packages/adapter-pg"
   },
   "devDependencies": {
-    "@types/node": "18.19.76",
-    "dotenv": "^16.4.7",
+    "@types/node": "~20.19.0",
+    "@prisma/config": "../../packages/config",
     "prisma": "../../packages/cli",
+    "pg": "8.13.3",
+    "@types/pg": "8.11.11",
```
jkomyno: Why?

FGoessler: Ah right we don't need that anymore. Left over from the times when pg was only a peer-dependency of our driver adapters.

sandbox/basic-postgres/prisma.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig, env } from "@prisma/config";
+import 'dotenv/config';
+
+export default defineConfig({
+  datasource: {
+    url: env('TEST_FUNCTIONAL_POSTGRES_16_URI'),
```
jacek-prisma: I think they should use `TEST_POSTGRES_URI`, `TEST_FUNCTIONAL_POSTGRES_URI` by default has a db name placeholder `postgres://prisma:prisma@localhost:5432/PRISMA_DB_NAME`

sandbox/basic-sqlite/package.json
```diff
@@ -15,12 +15,13 @@
   "license": "ISC",
   "dependencies": {
     "@prisma/client": "../../packages/client",
-    "@prisma/config": "../../packages/config"
+    "@prisma/adapter-libsql": "../../packages/adapter-libsql"
```
jacek-prisma: our docs point libsql users at `better-sqlite3`by default, so maybe the sandbox should use it too

jacek-prisma (approved):

PR #28672, opened 2025-11-23
feat(cli): Studio now warns when SQLite file did not exist and was created.

igalklebanov:
Hey :wave:

closes PTL-559.

This PR checks for the existence of the target SQLite file and if it didn't exist before, it prints a warning. This is to let users know that they might've provided a wrong path.

### DIFF
--- a/packages/cli/src/Studio.ts
+++ b/packages/cli/src/Studio.ts
@@ -89,7 +89,7 @@ const CONNECTION_STRING_PROTOCOL_TO_STUDIO_STUFF: Record<string, StudioStuff | n
         await access(resolvedPath, constants.F_OK).catch(() => {
           console.warn(
             yellow(
-              `database file at "${resolvedPath}" was not found. a new file was created. if this is an unwanted side effect, it might mean that the URL you have provided is incorrect.`,
+              `Database file at "${resolvedPath}" was not found. A new file was created. If this is an unwanted side effect, it might mean that the URL you have provided is incorrect.`,
             ),
           )
         })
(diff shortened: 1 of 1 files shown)


packages/cli/src/Studio.ts
```diff
@@ -81,7 +81,19 @@ const CONNECTION_STRING_PROTOCOL_TO_STUDIO_STUFF: Record<string, StudioStuff | n
...
-      const resolvedPath = path !== ':memory:' ? resolve(relativeTo, path) : path
+      const isInMemory = path === ':memory:'
+
+      const resolvedPath = isInMemory ? path : resolve(relativeTo, path)
+
+      if (!isInMemory) {
+        await access(resolvedPath, constants.F_OK).catch(() => {
+          console.warn(
+            yellow(
+              `database file at "${resolvedPath}" was not found. a new file was created. if this is an unwanted side effect, it might mean that the URL you have provided is incorrect.`,
```
FGoessler: Intentional lower case at sentence starts? Is that our cli output style? 🤔

FGoessler (dismissed):

packages/cli/src/Studio.ts
igalklebanov:
```suggestion
              `Database file at "${resolvedPath}" was not found. A new file was created. If this is an unwanted side effect, it might mean that the URL you have provided is incorrect.`,
```

igalklebanov: Corrected.

FGoessler (approved):

PR #28694, opened 2025-11-24
fix: allow env generic with interfaces

### LINKED ISSUE #28617: env() helper should allow interface generic
### Bug description

This works (copied from an example in the docs)

```ts
import 'dotenv/config'

type Env = {
  DATABASE_URL: string
}

export default defineConfig({
  migrations: {
    path: 'prisma/migrations',
  },
  datasource: {
    url: env<Env>('DATABASE_URL'),
  },
});
```

This, however, does not work
Notice how I changed from `type` to `interface` for `Env`

```ts
import 'dotenv/config'

interface Env {
  DATABASE_URL: string
}

export default defineConfig({
  migrations: {
    path: 'prisma/migrations',
  },
  datasource: {
    url: env<Env>('DATABASE_URL'),
  },
});
```

The code above gives the following error

```
Type 'Env' does not satisfy the constraint 'Record<string, string | undefined>'.
Index signature for type 'string' is missing in type 'Env'. (ts 2344)
```

I think the generic for `env` should be updated to also allow using an interface like the one in the example above

### Severity

### Reproduction

1. Follow your own docs here: https://www.prisma.io/docs/orm/reference/prisma-config-reference#type-safe-environment-variables
2. Change `Env` from being a `type` to being an `interface` instead.

Expected:

No TS errors

Actual:

Type 'Env' does not satisfy the constraint 'Record<string, string | undefined>'.
Index signature for type 'string' is missing in type 'Env'. (ts 2344)

### Frequency

Consistently reproducible

### Is this a regression?

No, this is new functionality

### Workaround

Add an index signature to the interface, like this:

```ts
interface Env {
  [key: string]: string | undefined
  DATABASE_URL: undefined
}
```

### Prisma Schema & Queries

N/A

### Prisma Config

```ts
import { defineConfig, env } from 'prisma/config'

interface Env {
  DATABASE_URL: undefined
}

const url = env<Env>('DATABASE_URL')

export default defineConfig({
  schema: 'db/schema.prisma',
  migrations: {
    path: 'db/migrations',
  },
  datasource: {
    url,
  },
})
```

### Logs & Debug Info

N/A

### Environment & Setup

- **OS:** macOS
- **Database:** Prisma Postgres
- **Node.js version:** v24.11.1

### Prisma Version

```
❯ npx prisma -v
prisma                  : 6.19.0
@prisma/client          : 6.19.0
Node.js                 : v24.11.1
TypeScript              : 5.9.2
Query Engine (Node-API) : libquery-engine 2ba551f319ab1df4bc874a89965d8b3641056773 (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
PSL                     : @prisma/prisma-schema-wasm 6.19.0-26.2ba551f319ab1df4bc874a89965d8b3641056773
Schema Engine           : schema-engine-cli 2ba551f319ab1df4bc874a89965d8b3641056773 (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Default Engines Hash    : 2ba551f319ab1df4bc874a89965d8b3641056773
```

SaubhagyaAnubhav:
### Problem

The `env` helper rejected interface-based environment variable types because the previous generic constraint  
`Env extends Record<string, string | undefined>` required an index signature that interfaces don’t have.  
This caused `TS2344` when following the documented example.

### Fix

The type definition of `env` was updated and a new `EnvKey` helper type was introduced to filter valid keys.  
This allows both interfaces and type aliases to work as expected without changing runtime behavior.

### Changes Made

- Updated `env` signature using a simplified generic pattern  
- Added `EnvKey` to restrict keys to `string | undefined` values  
- Added a type-level test ensuring interface + type alias compatibility  
- Kept all runtime logic untouched  

### Result

- Interface support now matches documentation  
- Type aliases continue to work  
- Compile-time validation preserved  
- Backwards-compatible with no runtime impact


SaubhagyaAnubhav:
@FGoessler  pls review this PR

FGoessler (approved):
Thanks for this tweak @SaubhagyaAnubhav 🎉

PR #28699, opened 2025-11-24
fix: add test coverage for Studio value column bug #28698

swarnim02:
- Add test fixture with tables containing 'value' columns
- Add Studio command tests for URL validation
- Create reproduction case for the bug
- Document root cause and proposed solutions


igalklebanov:
Hey 👋 

The issue is tracked in https://github.com/prisma/studio/issues/1392 and already fixed. Tests for this bug do not belong in the CLI distribution.

PR #28704, opened 2025-11-25
feat(client): add whereUnique support for relations with compound unique constraints

addy-2709genius:
## Description

Implements `whereUnique` option for relation selections when filtering by compound unique constraints, as requested in #28376.

## Problem

When selecting a relation with a compound unique constraint (e.g., `@@unique([creatorId, formId])`), using `where` returns an array even though the constraint guarantees at most 1 result.

## Solution

Adds `whereUnique` option that:
- Accepts partial compound unique constraint fields
- Returns `T | null` instead of `T[]` in types
- Converts to proper `where` clause with compound unique constraint at runtime
- Validates that `where` and `whereUnique` cannot be used together

## Example

// Before: returns FormResponse[]
const result = await prisma.form.findMany({
  select: {
    responses: {
      where: { creatorId: "..." },
    },
  },
});

// After: returns FormResponse | null
const result = await prisma.form.findMany({
  select: {
    responses: {
      whereUnique: { creatorId: "..." },
    },
  },
});## Changes

-  TypeScript generator: Added `whereUnique` to relation selection types
-  Runtime: Convert `whereUnique` to `where` with compound unique constraints
-  Tests: Comprehensive test suite added

## Note

Result processing to convert arrays to single items when `whereUnique` is used is still needed and will be added in a follow-up.

Closes #28376


tensordreams:
Hi @addy-2709genius — thanks for this, and apologies for the slow reply.

We are closing this one: it adds a new public client API surface that also needs query-compiler support, so it needs a product decision rather than a review.

PR #28710, opened 2025-11-25
ci: re-add memory tests

jacek-prisma:
[TML-1536](https://linear.app/prisma-company/issue/TML-1536/port-in-memory-tests-to-querycompiler)

### DIFF
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -420,7 +420,7 @@ jobs:
           timeout_minutes: 10
           max_attempts: 5
           retry_wait_seconds: 120
-          command: docker compose -f docker/docker-compose.yml up --wait --detach postgres # mysql mysql_isolated mssql mongo cockroachdb
+          command: docker compose -f docker/docker-compose.yml up --wait --detach postgres
 
       - name: Install & build
         uses: ./.github/actions/setup
--- a/.github/workflows/test-template.yml
+++ b/.github/workflows/test-template.yml
@@ -397,6 +397,97 @@ jobs:
           secret: ${{ secrets.BUILDPULSE_SECRET_ACCESS_KEY }}
 
   #
+  # CLIENT (memory tests)
+  #
+  client-memory:
+    name: Client (memory tests) [v${{ matrix.node }}]
+    timeout-minutes: ${{ inputs.jobTimeout }}
+    runs-on: ubuntu-latest
+    if: |
+      contains(inputs.jobsToRun, '-all-') ||
+      contains(inputs.jobsToRun, '-client-')
+    strategy:
+      fail-fast: false
+      matrix:
+        queryEngine: [client]
+        node: ['20.19.0', '22.12.0', '24']
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Start Docker Compose Services
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 10
+          max_attempts: 5
+          retry_wait_seconds: 120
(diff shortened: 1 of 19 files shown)


packages/client/tests/memory/big-schema/prisma.config.ts
```diff
@@ -4,4 +4,5 @@ export default defineConfig({
   datasource: {
     url: 'file:./dev.db',
   },
+  schema: 'prisma/schema.prisma',
```
jacek-prisma, line 7: These are only needed because of the schema loading bug which is fixed in https://github.com/prisma/prisma/pull/28734. They can be removed after the fix is merged.

packages/client/tests/memory/_utils/commonSettings.ts
```diff
@@ -3,4 +3,4 @@
 // Our tests incorrectly report this additional memory as leakage.
 // See: https://github.com/prisma/prisma/pull/25971#issuecomment-2572448506
 // TODO: Revisit this once Node.js 24 LTS is out.
-export const WARMUP_ITERATIONS = 1500
+export const WARMUP_ITERATIONS = 2000
```
jacek-prisma, line 5: needed to bump this for node 24

FGoessler: Can the TODO comment above be updated due to that when we now run on Node 24 anyways? 🤔

.github/workflows/test-template.yml
```diff
@@ -396,6 +396,97 @@ jobs:
...
+    timeout-minutes: ${{ inputs.jobTimeout }}
+    runs-on: ubuntu-latest
+    if: |
+      contains(inputs.jobsToRun, '-all-') ||
+      contains(inputs.jobsToRun, '-client-')
+    strategy:
+      fail-fast: false
+      matrix:
+        queryEngine: [client]
+        node: ['20.19.0', '22.12.0', '24']
```
FGoessler, line 413: Why not on a specific node 24 version? 🤔

.github/workflows/test-template.yml
```diff
@@ -396,6 +396,97 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Start Docker Compose Services
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 10
+          max_attempts: 5
+          retry_wait_seconds: 120
+          command: docker compose -f docker/docker-compose.yml up --wait --detach postgres # mysql mysql_isolated mssql mongo cockroachdb
```
FGoessler: ❓ Is the comment here still relevant?

FGoessler (dismissed):
Only nitpicks ;)

.github/workflows/test-template.yml
jacek-prisma: If you look at the rest of the file, these are the node versions we use for every test. It was decided a while back based on the fact that `24.0.0` had a nasty bug that actually breaks our CI and 24 was not LTS at the time. I think we could change it to a specific version now, but I think that's out of scope of this PR.

.github/workflows/test-template.yml
jacek-prisma: memory tests just don't need these containers, in theory the tests could use other databases but right now they're hardcoded to only use sqlite and postgres

I think the comment just indicates that we might add them later

packages/client/tests/memory/_utils/commonSettings.ts
jacek-prisma: removed it

.github/workflows/test-template.yml
jacek-prisma: removed it

FGoessler (approved):

PR #28724, opened 2025-11-25
fix(client-engine-runtime): make NowGenerator lazy to avoid synchronous new Date() calls

Priyank-Gaur:
The Issue: The GeneratorRegistry.snapshot() method creates a snapshot of all generators, including **NowGenerator**

```
class NowGenerator implements ValueGenerator {
  #now: Date = new Date()

  generate(): string {
    return this.#now.toISOString()
  }
}
```

The **NowGenerator** was initializing new Date() in its property declaration, meaning it was called synchronously during every query preparation, even if now() was not used.

The Fix: I modified  **NowGenerator** to be lazy. 

```
class NowGenerator implements ValueGenerator {
  #now: Date | undefined

  generate(): string {
    if (this.#now === undefined) {
      this.#now = new Date()
    }
    return this.#now.toISOString()
  }
}
```

The #now field is now initialized as undefined, and new Date() is only called inside the **generate()**

```
class NanoIdGenerator implements ValueGenerator {
  generate(arg: unknown): string {
    if (typeof arg === 'number') {
      return nanoid(arg)
    } else if (arg === undefined) {
      return nanoid()
    } else {
      throw new Error('Invalid Nanoid generator arguments')
    }
  }
} 
```

method if it hasn't been initialized yet.

Fixes #28588


aqrln:
I'm not opposed to this, but I'm not sure it will have any measurable impact to be honest. Is this a theoretical concern or an actual issue you profiled?

Also, if you'd like to get this merged, please sign the CLA.

Priyank-Gaur:
> I'm not opposed to this, but I'm not sure it will have any measurable impact to be honest. Is this a theoretical concern or an actual issue you profiled?
> 
> Also, if you'd like to get this merged, please sign the CLA.

Thank you for telling me to sign the CLA, This is my first time contributing to a big repository so I didn't knew that.
And about the issue I saw it from existing issues.
When using Prisma 7 together with Next.js Cache/Server Components, the eager new Date() call inside NowGenerator was being executed during generator snapshot creation. That triggers Next.js’ runtime warning:

“used new Date() before accessing either uncached data”

This happens even when now() is not used in the query.
By deferring the date initialization to the generate() method, the warning disappears and the page renders correctly without requiring additional workarounds.
So the change does have a measurable effect in real applications — it prevents premature time access during static evaluation.
If there is anything else you'd like to ask feel free.....it will help me to learn more.....Thanks

ftognetto:
> I'm not opposed to this, but I'm not sure it will have any measurable impact to be honest. Is this a theoretical concern or an actual issue you profiled?
> 
> Also, if you'd like to get this merged, please sign the CLA.

This is blocking all the Nextjs builds, for example. https://github.com/prisma/prisma/issues/28588

damischa1:
In our case prevents building the NextJS app. Any page that is making prisma queries is not possible to cache with cacheComponents. And actually those are the pages that we want to use caching. So the measurable impact is big.

mugiwaraya:
Any news follow-up on this?

cipriancaba:
This is why people move to @drizzle

jacek-prisma (approved):

PR #28734, opened 2025-11-26
refactor: refactor schema path loading and fix config relative handling

jacek-prisma:
[TML-1634](https://linear.app/prisma-company/issue/TML-1634/fix-the-default-schema-loading-location)

Refactors the code so that instead of passing 3 parameters everywhere we now pass a schema path that can only be one of the expected inputs:
```
export type SchemaPathInput = { cliProvidedPath: string } | { configProvidedPath: string } | { rootDir: string }
```
I also created a convenience function `createSchemaPathInput` for creating it.

This PR also fixes the default schema loading so that the schema is searched relative to the config file if it exists and `cwd` otherwise and added tests for the affected commands.


packages/cli/src/DebugInfo.ts
```diff
@@ -64,7 +78,14 @@ export class DebugInfo implements Command {
 
     let schemaPath
     try {
-      schemaPath = link((await getSchemaWithPath(args['--schema'], config.schema))?.schemaPath)
+      const schemaResult = await getSchemaWithPath({
+        schemaPath: createSchemaPathInput({
+          schemaPathFromArgs: args['--schema'],
+          schemaPathFromConfig: config.schema,
+          rootDir: configDir,
```
FGoessler: Sometimes this is called `configDir` and sometimes `rootDir`. Can we better align this or what's the difference between the terms? 🤔 
It is either the directory where the config file is located (if present) or as a fallback the cwd, right?

jacek-prisma: >  It is either the directory where the config file is located (if present) or as a fallback the cwd, right?

Yes

The reason it's named `configDir` in some places is because it was already named that way and I didn't happen to change it. I think the name is misleading though, because it's not always a config directory, so I'd be happy to rename it. I think `rootDir` is good in a context where we're unambiguously referring to a schema location, but in other places I think we should call it `schemaRootDir` or `prismaRootDir` or something like that. I'm open to suggestions.

FGoessler: How about `baseDir`? Conveys the point of being the directory we "base" ourselfes to and try to treat things relative to?

`schemaRootDir` I fidn confusing because one might think that it's the directory in which we search for multiple schema files when using the multi schema file feature. 🤔  Similar but not as drastic for `prismaRootDir`.

jacek-prisma: updated the code to use `baseDir` consistently

also removed schema paths from memory tests since they're no longer needed thanks to this PR

packages/internals/src/cli/getSchema.ts
```diff
@@ -36,30 +36,54 @@ type DefaultLookupResult =
...
+  schemaPath: SchemaPathInput
   cwd?: string
   argumentName?: string
 }
 
 type GetSchemaInternalOptions = Required<GetSchemaOptions>
 
+/**
+ * Creates SchemaPathInput based on a combination of possible inputs
+ * from CLI args, config file, or root directory.
```
FGoessler:
```suggestion
 * from CLI args, config file, or base directory. 
 * `baseDir` is either the directory containing the prisma config file or the cwd of the cli invocation if no config file is found.
```

FGoessler (dismissed):
One tiny addition to document the `baseDir` behavior. Besides that let's go! Nice work with the refactor! 🙌

FGoessler (approved):

PR #28743, opened 2025-11-26
doc: Add link to open source petition

Riiiad:
Hey hey,
there is a petition for Recognition of Open Source Work as Voluntary Service in Germany. Thought this would be a nice way to promote this nice petition.

The position in the readme is just an example, you can move it as you wish.


aqrln (changes requested):
While this petition is certainly worth being shared and supported, I don't believe this is an appropriate place for this, both due to the reasons stated above by Copilot, and because this project is maintained by a commercial company and salaried employees, not by volunteers.

If I were to review a similar change in a personal project, however, I would probably accept a similar change in principle but not in this form. I think the description is lacking context and is potentially confusing, and I find the QR code to be very intrusive and a potential security risk. Hope this feedback might be useful for getting this change accepted in more projects.

PR #28744, opened 2025-11-26
feat(migrate): suggest migration name based on schema changes

### LINKED ISSUE #28239: Auto-generate migration names
### Feature Summary

Suggest new migration name in the CLI based on schema changes.

Every time I make a change to schema and generate a migration, I have to think how to name the migration file. It usually ends up being something like `add_<field>_to_<table>` or `create_<index name>`. Could CLI suggest a name? Drizzle doesn't ask the user, but generates a random migration name (like elegant_elephant). 

### Proposed Solution

Examples:

- When one column is added: add_<column name>_to_<table>
- When multiple colums are added: add_fields_to_<table> or add_<column name>_and_other_fields_to_<table> or add_<column 1>_and_<column 2>_to_<table>
- When one column is removed: drop_<column name>_in_<table>

It could suggest names only for simple changes (one column added/removed) and require the user to provide a name for more complex one so not every edge case has to be handled.

CLI would prefill the input field (if this is possible in CLI) or let the user choose from 2 options - suggested name and custom name.

### Alternative Solutions

**Alternative 1: Random words**

Generate a random combination of words - what Drizzle does.

**Alternative 2: Optional name**

Let the user not name the migration file and identify it only based on its timestamp.

### Potential Considerations

Every team has different naming conventions so it might be helpful to make this optional or let the user override name templates.

### Prisma Version

_No response_

None

### Additional Context

Thinking about migration names seems like unnecessary burden when making a simple database change.

### Pre-Submission Checklist

VISHNU7KASIREDDY:
### Description
This PR introduces a new feature to the Prisma CLI that automatically suggests migration names based on your schema changes.
By analyzing the output of `migrate diff`, the CLI can now intelligently pre-fill the migration name prompt with suggestions like `add_email_to_User` or `create_table_Post`, saving you time and effort.

**Closes #28239**

### Key Changes
- **New Utility**: Added `getMigrationNameSuggestion` to parse schema diffs and generate human-readable names.
- **CLI Integration**: Updated the `MigrateDev` command to generate a suggestion and pass it to the prompt.
- **Improved UX**: The migration name prompt now accepts an initial value, allowing for the suggestion to be pre-filled.

### Verification
- **Unit Tests**: Added comprehensive unit tests for the suggestion utility to handle various schema change scenarios.
- **Integration Tests**: Verified the flow using SQLite integration tests to ensure the prompt is correctly populated.


aqrln:
This is a very cool idea but I think there should be a better way to implement it. There are two problems with this approach:

1. Parsing the output using regexes is brittle and may/will break in the future.
2. An extra `migrate diff` call adds extra latency which will negatively affect UX.

The first problem is easy to solve, e.g., by getting structured JSON output from the schema engine, but the second problem is trickier because SE needs a migration name beforehand (and needs to write it to the migration table after applying).

VISHNU7KASIREDDY:
@aqrln I will try to solve the isuues ,Thanks for the correction 😊.

VISHNU7KASIREDDY:
Thanks @aqrln for the feedback! You're absolutely right.

I dug into the code and realized there isn't a clean way to get the schema changes as data (JSON) right now. The engine only outputs text for humans or SQL, and parsing that text is fragile. Plus, making extra calls does slow things down.

Rather than adding a "band-aid" fix that might break later, I think the best path is to add this capability to the engine properly.

I'm going to close this PR and open a request in `prisma-engines` for a new feature that gives us the diff as structured data. This will make this feature robust and help with other tools too.

I'll post the link to the new issue shortly. Thanks for guiding me in the right direction!

PR #28762, opened 2025-11-28
fix: bump prisma dev to fix valibot security issue

FGoessler:


igalklebanov (approved):
it does fix it. saw it with me own eyes!

jkomyno (approved):

PR #28765, opened 2025-11-28
feat: TML-1618 add ppg local vs remote tracking for studio cli

FGoessler:
Detects whether studio connects to a local or remote ppg instance. Primarily by URL but to detect a local ppg we cross check a provided postgres url with the list of known local ppg instances to see if the ports match.

This information is then used in the telemetry data for studio. Tried to be analogous to the studio tracking implementation in the VSCode Extension.

### DIFF
--- a/packages/cli/src/utils/ppgInfo.ts
+++ b/packages/cli/src/utils/ppgInfo.ts
@@ -1,0 +1,31 @@
+import { ServerState } from '@prisma/dev/internal/state'
+
+/**
+ * Collects information about whether the given url connects to a Prisma Postgres remote or local database.
+ * Mimicking parts of the tracking logic in the Prisma VSCode extension.
+ */
+export async function getPpgInfo(connectionString: string) {
+  const url = new URL(connectionString)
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/ppgInfo.ts
```diff
@@ -0,0 +1,31 @@
+import { ServerState } from '@prisma/dev/internal/state'
+
+/**
+ * Collects information about whether the given url connects to a Prisma Postgres remote or local database.
+ * Mimicking parts of the tracking logic in the Prisma VSCode extension.
+ */
+export async function getPpgInfo(connectionString: string) {
+  const url = new URL(connectionString)
+  const isLocalhost = url.hostname === 'localhost' || url.hostname === '127.0.0.1'
```
aqrln: should it also check for ipv6 loopback address?

FGoessler: added it - although I think it's an unlikely edge case 😅

aqrln (approved):

Issue #28339, opened 2025-10-21
Missing `Decimal` type export on the `prismaNamespaceBrowser` file making the migration to `prisma-client` impossible for us.


dorbn4:
### Bug description

Hey there 👋

We are currently migrating to the latest prisma versions and the new `prisma-client` and are having a hard time as the Decimal type is missing from the browser export file.

`versions`
```json
"@prisma/adapter-neon": "^6.17.1",
"@prisma/adapter-pg": "^6.17.1",
"prisma": "6.17.1",
```

`schema.prisma`
```prisma
generator client {
  provider        = "prisma-client"
  previewFeatures = ["fullTextSearchPostgres"]
  output          = "../../generated/prisma"

  // Optional
  moduleFormat = "esm"
}

datasource db {
  provider     = "postgresql"
  url          = env("DATABASE_URL")
  relationMode = "prisma"
}
```

We are extensively using the `Prisma.Decimal` and as you guys do not really use the correct `Decimal` from `decimal.js` we are forced to import the `Prisma.Decimal` from the generated files. Currently the `export type Decimal = runtime.Decimal` is missing from the `prismaNamespaceBrowser` file making it impossible to import the `Prisma.Decimal` type on the client side. 

<img width="1512" height="784" alt="Image" src="https://github.com/user-attachments/assets/4d80147b-2768-4bff-82a3-43bd02bb8225" />

### Severity

⚠️ Major: Breaks core functionality (e.g., migrations fail)

### Reproduction

No reproduction needed, its literally a one line fix when you know the prisma generator codebase.

### Expected vs. Actual Behavior

We expected that the type would be exported and this is currently leading to quite a headache.

### Frequency

Consistently reproducible

### Does this occur in development or production?

Both development and production

### Is this a regression?

YES. It worked with `prisma-client-js` and (hopefully) the missing type export was just an oversight.

### Workaround

Not possible. It's completely making the migration impossible for us as we cant import the `Prisma.Decimal` from the non-browser file on the client due to the prisma client leading to FE crashes. We need to have the type export on the browser file. 

### Prisma Schema & Queries

n/a

### Prisma Config

see above

### Logs & Debug Info

n/a

### Environment & Setup

- **OS:** macOS
- **Database:** PostgreSQL
- **Node.js version:** 22

### Prisma Version

6.17.1

PrateekSingh43:
Hey everyone,

[...]

PR #28766, opened 2025-11-28
feat(client-generator-ts): `export import Decimal` so that consumers can fully use the Decimal.js package

brandon-julio-t:
### Summary

This PR updates the export strategy for the `Decimal` type in the Prisma client generator to allow consumers full access to the Decimal.js package API.

### Changes

**Before:**
```typescript
export const Decimal = runtime.Decimal
export type Decimal = runtime.Decimal
```

**After:**
```typescript
export import Decimal = runtime.Decimal
```

### Technical Details

The change uses TypeScript's type-only import/export syntax (`export import`) instead of value re-exports. This allows consumers to:

1. **Access static methods and properties** of Decimal.js (e.g., `Decimal.config()`, `Decimal.clone()`, etc.)
2. **Use Decimal.js constructors and utilities** directly through `Prisma.Decimal`
3. **Maintain full type safety** while having access to the complete Decimal.js API

### Files Modified

- `packages/client-generator-ts/src/TSClient/common.ts` - Updated Decimal export in main client file
- `packages/client-generator-ts/src/TSClient/file-generators/PrismaNamespaceBrowserFile.ts` - Updated Decimal export in browser bundle

### Testing

The change maintains backward compatibility for existing usage patterns like:
```typescript
const decimal = new Prisma.Decimal('12.34’)

interface RequestBody {
  amount: Prisma.Decimal.Value
}
```

While now also enabling access to Decimal.js static methods:
```typescript
Prisma.Decimal.config({ precision: 10 })
```

### Related Issue

Closes

- #28339

This change ensures that consumers can fully utilize the Decimal.js package when working with Prisma's Decimal fields, providing access to the complete API rather than just a re-exported reference.

## Summary by CodeRabbit

* **Refactor**
  * Simplified how a numeric/decimal type is exposed in the TypeScript client library. Internal declaration style changed for consistency; no functional or runtime differences for end users.


SamuelMS:
This issue is important; we make heavy use of `decimal.js` directly, and consumers of [prisma-zod-generator](https://github.com/omar-dulaimi/prisma-zod-generator) get hit with this out of the box since the validators assume Prisma's implementation – a fully separate type – rather than a type alias to `Decimal` under the hood.

brandon-julio-t:
Closing this from my side.

The original production need for this change is no longer active: the codebase that needed it has moved its active DB layer from Prisma to Drizzle, with Prisma now only legacy there. I do not want to keep a stale PR open or ask maintainers to spend review time on something I am no longer depending on.

Thanks for taking a look.

PR #28782, opened 2025-11-30
fix(client): preserve relation types with complex where clauses in GetFindResult

### LINKED ISSUE #28703: In Prisma 7, extending with accelerate drops relations in types under certain conditions
### Bug description

Selected relations are dropped from the type under some conditions in `where` clauses. So far, I've reproduced this with `in`, for example `where: { foo: { in: ['bar', 'baz'] } }`. The provided reproduction repository illustrates that.

Marking this as major since project cannot be built with Prisma 7 in these conditions, or I'd have to drop Accelerate, which is far from ideal. More importantly, it breaks core functionality.

### Severity

### Reproduction

[Reproduction repository](https://github.com/janoma/prisma-7-bug).

The following block is OK:
```ts
const users = await prisma.user.findMany({
  include: { posts: true },
  where: { id: 1 }
});

const total = users.map((user) => user.posts.length)
//                                       ^? (property) posts: { id: number; authorId: ...
```

Replace the `where` clause for the one that fails and see the error on `user.posts`:

```ts
const users = await prisma.user.findMany({
  include: { posts: true },
  where: { id: { in: [1]}}
});

const total = users.map((user) => user.posts.length)
//                                       ^? any
```

Image

Using `select` instead of `include` does not fix the issue. The type of the relation is dropped too.

#### Expected behavior

No type errors.

#### Actual behavior

`posts` is dropped from each user object in the typed response:

```sh
Property 'posts' does not exist on type '{ id: number; }'.ts(2339)
```

### Frequency

Consistently reproducible

### Is this a regression?

This worked with `@prisma/extension-accelerate@1.3.0`, `@prisma/client@6.19.0` and `prisma@6.19.0`.

### Workaround

1. Remove `$extends(withAccelerate())`, _or_
2. Downgrade to `prisma@6.19.0`, `@prisma/client@6.19.0` and `@prisma/extension-accelerate@1.3.0`

#### Model

```prisma
model User {
  id    Int    @id @default(autoincrement())
  posts Post[]
}

model Post {
  id        Int    @id @default(autoincrement())
  author    User   @relation(fields: [authorId], references: [id])
  authorId  Int
}
```

#### Query

```ts
  const users = await prisma.user.findMany({
    include: { posts: true },
    where: { id: { in: [1] }}
  });
```

#### Prisma Config

```ts
import "dotenv/config";

export default defineConfig({
  migrations: {
    path: "prisma/migrations",
  },
  datasource: {
    url: env("DATABASE_URL"),
  },
});
```

### Logs & Debug Info

Not applicable since this is a type error.

### Environment & Setup

- **OS:** macOS 15.6.1
- **Database:** PostgreSQL
- **Node.js version:** 22.19.0

### Prisma Version

```
prisma               : 7.0.0
@prisma/client       : 7.0.0
Operating System     : darwin
Architecture         : arm64
Node.js              : v22.19.0
TypeScript           : 5.9.3
PSL                  : @prisma/prisma-schema-wasm 6.20.0-16.next-0c19ccc313cf9911a90d99d2ac2eb0280c76c513
Schema Engine        : schema-engine-cli 0c19ccc313cf9911a90d99d2ac2eb0280c76c513 (at node_modules/.pnpm/@prisma+engines@7.0.0/node_modules/@prisma/engines/schema-engine-darwin-arm64)
Default Engines Hash : 0c19ccc313cf9911a90d99d2ac2eb0280c76c513
Studio               : 0.8.0
```

### Related issues

Possibly related to [issue 28661](https://github.com/prisma/prisma/issues/28661).

Meet00028:
## Summary
Fixes #28703

When using `include` with complex `where` clauses like `{ id: { in: [1] } }` in Prisma 7 with `$extends(withAccelerate())`, TypeScript was dropping relation types from the result.

## Changes
- Updated `GetFindResult` type to check for `include`/`select` properties using `keyof` instead of pattern matching
- This allows TypeScript to correctly infer relation types even when `where` clauses contain complex nested types
- Fixed TypeScript compilation error by safely accessing `A['omit']` only when it exists

## Testing
- TypeScript build passes
- No lint errors
- Fix verified with the scenario from issue #28703

### DIFF
--- a/packages/cli/src/DebugInfo.ts
+++ b/packages/cli/src/DebugInfo.ts
@@ -123,6 +123,7 @@ ${formatEnvValue('PRISMA_HIDE_UPDATE_MESSAGE')}
 For downloading engines
 ${formatEnvValue('PRISMA_ENGINES_MIRROR')}
 ${formatEnvValue('PRISMA_BINARIES_MIRROR', '(deprecated)')}
+${formatEnvValue('PRISMA_ENGINES_STRICT_MIRROR')}
 ${formatEnvValue('PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING')}
 ${formatEnvValue('BINARY_DOWNLOAD_VERSION')}
 
--- a/packages/cli/src/__tests__/commands/DebugInfo.test.ts
+++ b/packages/cli/src/__tests__/commands/DebugInfo.test.ts
@@ -53,6 +53,7 @@ const envVars = {
   PRISMA_HIDE_UPDATE_MESSAGE: 'true',
   PRISMA_ENGINES_MIRROR: 'http://localhost',
   PRISMA_BINARIES_MIRROR: 'http://localhost',
+  PRISMA_ENGINES_STRICT_MIRROR: 'true',
   PRISMA_ENGINES_CHECKSUM_IGNORE_MISSING: 'true',
   BINARY_DOWNLOAD_VERSION: 'true',
   PRISMA_SCHEMA_ENGINE_BINARY: 'some/path',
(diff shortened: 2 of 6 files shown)


aqrln:
Thanks so much! Could you please remove the unrelated first commit from this PR?

PR #28807, opened 2025-12-02
fix(cli): lazy load @prisma/dev in Init.ts to prevent Node <22 crash

devanshuprakash:
### Description
This PR fixes a crash that occurred when running `prisma generate` or `prisma -v` on Node.js versions < 22 (or Bun delegating to Node).  
The issue was caused by a top-level `require` of `@prisma/dev`, which depends on an ESM-only module (`zeptomatch`), causing incompatibility with older Node versions.

### Changes
- Lazy-loaded `@prisma/dev` and `@prisma/dev/internal/state` inside `packages/cli/src/Init.ts` within `defaultEnv`.
- Ensures `@prisma/dev` is only loaded when `prisma init` is executed and default environment variables need to be generated.
- Prevents unnecessary loading for other CLI commands, avoiding crashes.

### Fixes
- Fixes #28805

### DIFF
--- a/packages/cli/src/Init.ts
+++ b/packages/cli/src/Init.ts
@@ -3,8 +3,6 @@ import path from 'node:path'
 
 import { confirm, input, select } from '@inquirer/prompts'
 import { PrismaConfigInternal } from '@prisma/config'
-import { unstable_startServer } from '@prisma/dev'
-import { ServerState } from '@prisma/dev/internal/state'
 import type { ConnectorType } from '@prisma/generator'
 import {
   arg,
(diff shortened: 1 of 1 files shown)


aqrln:
> on Node.js versions < 22

`require(esm)` is supported on Node.js 20.19+, which is the minimal required version for Prisma.

tensordreams:
Hi @devanshuprakash — thanks for this, and apologies for the slow reply.

We are closing this one: the premise does not hold — `require(esm)` works on Node 20.19+, which is Prisma's minimum supported version, so the lazy import is not needed.

PR #28818, opened 2025-12-02
docs(sqlcommenter): add missing `sql` field information

aqrln:
Add a field missing in the docs.

### DIFF
--- a/packages/sqlcommenter/README.md
+++ b/packages/sqlcommenter/README.md
@@ -135,11 +135,10 @@ interface SqlCommenterContext {
 }
 ```
 
-- **`query`**: Information about the Prisma query being executed. See [`SqlCommenterQueryInfo`](#sqlcommenterqueryinfo).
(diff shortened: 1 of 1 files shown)


packages/sqlcommenter/README.md
```diff
@@ -131,11 +131,14 @@ A function that receives query context and returns key-value pairs. Return an em
...
 interface SqlCommenterContext {
   query: SqlCommenterQueryInfo
+  sql?: string
 }
 ```
 
 Context provided to plugins containing information about the query.
 
+- **`query`**: Information about the Prisma query being executed. See [`SqlCommenterQueryInfo`](#sqlcommenterqueryinfo).
+- **`sql`**: The SQL query being executed. It is only available when using driver adapters but not when using Accelerate.
```
aqrln, line 141:
```suggestion
- **`sql`**: The SQL query being executed. It is only available when using driver adapters but not when using Accelerate.

```

FGoessler (approved):

PR #28825, opened 2025-12-03
fix(adapter-mssql): prevent connection pool exhaustion and EREQINPROG errors by serializing commit/rollback

devanshuprakash:
#### Root Cause

The issue was caused by a race condition inside the `MssqlTransaction` class:

- The `commit` and `rollback` methods did not acquire the mutex that serializes access to the underlying MSSQL connection.
- As a result, `commit` or `rollback` could execute while a query was still in progress (e.g., a slow query).
- The MSSQL driver then threw:

EREQINPROG: Can't rollback transaction. There is a request in progress.

- In some cases, this left the connection in a bad state, eventually exhausting the connection pool.

## Fix

This PR ensures:

- `commit` acquires the mutex before executing.
- `rollback` acquires the mutex before executing.

This guarantees that these methods wait for any active queries to finish, removing the race condition.

### `packages/adapter-mssql/src/mssql.ts`

- Updated `MssqlTransaction.commit` to acquire the mutex.
- Updated `MssqlTransaction.rollback` to acquire the mutex.
- Exported `PrismaMssqlAdapter` to enable unit testing.

### `packages/adapter-mssql/src/transaction.test.ts`

- Added a regression test that reproduces the issue by:
  - Simulating a slow query,
  - Triggering a rollback while the query is active.
- Verified that the fix prevents the `EREQINPROG` error and allows the rollback to complete successfully after the query finishes.

## Verification

Tested via:

pnpm test packages/adapter-mssql/src/transaction.test.ts

Output:

✓ src/transaction.test.ts (1 test) 103ms
✓ PrismaMssqlAdapter reproduction > should reproduce EREQINPROG when rollback is called during active query 103ms

Fixes #28812


packages/adapter-mssql/src/transaction.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        } // Overridden in test
+        request() {
+          return {}
+        }
+      },
+    },
+  }
+})
+
+describe('PrismaMssqlAdapter reproduction', () => {
```
aqrln: the test suite name is confusing, it doesn't tell what exactly it reproduces (besides, if the issues is now fixed, it's not a reproduction anymore but a regression test)

packages/adapter-mssql/src/transaction.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  it('should reproduce EREQINPROG when rollback is called during active query', async () => {
+    // Setup mock behavior
+    let requestActive = false
+    let queryStartedResolve: () => void
+    const queryStarted = new Promise<void>((resolve) => (queryStartedResolve = resolve))
+
+    const rollbackMock = vi.fn().mockImplementation(async () => {
+      await Promise.resolve()
+      if (requestActive) {
+        const err: any = new Error("Can't rollback transaction. There is a request in progress.")
```
aqrln: please don't use `any` here

packages/cli/src/Init.ts
aqrln, line 1: please remove unrelated commit from this PR

aqrln (commented):
Thanks so much! The change in the logic itself looks good to me, but there are some issues in the test and unrelated changes that need to be moved out of this PR.

devanshuprakash:
I've addressed your feedback:
1. Renamed the test suite to `PrismaMssqlAdapter transaction regression` to better reflect its purpose.
2. Removed the `any` usage in the test and replaced it with a proper type assertion (`Error & { code: string }`).
Please review @aqrln

devanshuprakash:
@aqrln 
Done! I've updated the test names to be more descriptive, fixed the types, and added the requested assertions for mock calls and ordering.

arthur-ver:
@aqrln What's the status of this fix? Currently experiencing lots of `TransactionError: Can't rollback transaction. There is a request in progress.` with the latest Prisma 7.2.0...

trentonsnyder:
@devanshuprakash We patched our adapter with these changes about a week ago and haven't seen the `Transaction Error`since then. Thanks for putting this up 👍 

using v16.16.3 "rust-free"

tensordreams:
Merged latest `main` into the branch and resolved a stale-merge conflict (the branch had drifted far enough to revert the VarBinary-null fix, the savepoint methods, and the connection-string handling — all restored to match `main`).

Heads-up on scope: the commit/rollback serialization this PR introduces has since landed independently on `main` via #29141, so `mssql.ts` no longer differs there. The remaining delta on this branch is now:
- exporting `PrismaMssqlAdapter` so it can be constructed directly in tests, and
- `transaction.test.ts`, the regression test that guards the EREQINPROG race (verifies `rollback()` waits for an in-flight query instead of throwing "There is a request in progress").

All review feedback is addressed (test renamed, mock-call/ordering assertions added, `any` removed) and the unrelated `Init.ts` change is gone. `pnpm --filter @prisma/adapter-mssql test` passes (88 tests); prettier and eslint are clean on the changed files. Head is now 571e7960.

tensordreams:
Thank you for this, @devanshuprakash, and apologies for the very slow reply.

Your diagnosis was exactly right, and the fix has since landed on `main` via #29141: `MssqlTransaction.commit()` and `rollback()` now both `await this.#mutex.acquire()` — the same mutex `performIO()` holds — so they wait for an in-flight query instead of hitting `EREQINPROG` and stranding the connection. That PR also added `packages/adapter-mssql/src/mssql.test.ts`, which covers the same rollback-during-an-active-query scenario your `transaction.test.ts` reproduces.

With the behavioural change and the regression coverage both on `main`, all that remains in this branch is the `export` of `PrismaMssqlAdapter`, so I'm closing it as a duplicate. Thank you for tracking down the race and writing it up so clearly — the report was genuinely useful.

PR #28835, opened 2025-12-03
feat: fix enum array cast syntax in PostgreSQL adapter

Imposon:
This PR implements a workaround in the adapter-pg TypeScript runtime to correct incorrectly generated enum array cast syntax produced by the Rust query compiler (prisma-engines). The compiler currently produces SQL like:

cast($1 as "public"."Role[]")

which is invalid. The correct SQL is:

cast($1 as "public"."Role"[])

Because the Rust code that emits the SQL requires a separate change (and to avoid breaking the engines at once), this PR post-processes SQL strings in the runtime and fixes the specific incorrect pattern before executing the query.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SQL cast syntax for enum array types in PostgreSQL queries. The adapter now correctly transforms enum array cast patterns to ensure proper query execution.

* **Tests**
  * Added functional test coverage for enum array field updates across PostgreSQL and CockroachDB database providers.

### DIFF
--- a/packages/adapter-pg/src/__tests__/enum-array-cast.test.ts
+++ b/packages/adapter-pg/src/__tests__/enum-array-cast.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from 'vitest'
+
+describe('Enum array cast fix', () => {
+  it('should fix enum array cast syntax in SQL', () => {
+    // Test the SQL transformation indirectly by checking the behavior
+    // The actual fix happens in the private performIO method
+
+    // Test cases for the regex pattern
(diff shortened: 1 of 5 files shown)


Imposon:
solved for issue
prisma/studio#1426

tensordreams:
Hi @Imposon — thanks for this, and apologies for the slow reply.

We are closing this one: the SQL being corrected here is produced by the query compiler, and rewriting it with a regex inside the adapter can only mask that. The cast needs to be emitted correctly at the point it is generated.

PR #28837, opened 2025-12-03
fix(adapter-pg): Fix PostgreSQL Bytes upsert regression in v7

### LINKED ISSUE #28824: Prisma v7 breaks postgres Byte upserts
### Bug description

Prisma v7 breaks postgres Byte upserts.

### Severity

### Reproduction

Schema:
```prisma
datasource db {
  provider     = "postgresql"
  relationMode = "foreignKeys"
  extensions   = [pgcrypto]
}

generator client {
  previewFeatures = ["relationJoins", "typedSql", "postgresqlExtensions"]
  output          = "../__generated__/prisma"
  engineType      = "library"
}

generator pothos {
  provider = "prisma-pothos-types"
  output   = "../__generated__/pothos-types.ts"
}

model TestByteId {
  id    Bytes  @id
  value String
}

model TestStringId {
  id    String @id
  value String
}
```

script
```ts
await prisma.$executeRaw`DROP TABLE IF EXISTS "Test" CASCADE`;
await prisma.$executeRaw`DROP TABLE IF EXISTS "Test2" CASCADE`;

const stringId = 'hello world!';
const upsertStringRow = () =>
  prisma.testStringId.upsert({
    create: {
      id: stringId,
      value: 'hello world!',
    },
    where: {
      id: stringId,
    },
    update: {},
  });

await upsertStringRow();
console.log('created string row');
await upsertStringRow();
console.log('updated string row');

const byteId = new Uint8Array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
const upsertByteRow = () =>
  prisma.testByteId.upsert({
    create: {
      id: byteId,
      value: 'hello world!',
    },
    where: {
      id: byteId,
    },
    update: {},
  });

await upsertByteRow();
console.log('created byte row');
await upsertByteRow();
console.log('updated byte row');

await prisma.$disconnect();
```

Expected:
```
created string row
updated string row
created byte row
updated byte row
```

Actual:
```
created string row
updated string row
created byte row

[error]
An operation failed because it depends on one or more records that were required but not found. No record was found for an upsert.
```

### Frequency

Consistently reproducible

### Is this a regression?

Worked in 6.12.0
Broken in 7.0.1

### Workaround

Works in 6.12.0

### Prisma Schema & Queries

See repro

### Prisma Config

_No response_

### Logs & Debug Info

_No response_

### Environment & Setup

- **OS:** OSX 15.6.1
- **Database:** Postgres 17
- **Node.js version:** v22.14.0

### Prisma Version

```
prisma               : 7.0.1
@prisma/client       : 7.0.1
Operating System     : darwin
Architecture         : arm64
Node.js              : v22.14.0
TypeScript           : 5.8.3
PSL                  : @prisma/prisma-schema-wasm 7.1.0-2.f09f2815f091dbba658cdcd2264306d88bb5bda6
Schema Engine        : schema-engine-cli f09f2815f091dbba658cdcd2264306d88bb5bda6 (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Default Engines Hash : f09f2815f091dbba658cdcd2264306d88bb5bda6
Studio               : 0.8.2
```

dncoder14:
## Description
Fixes critical regression in Prisma v7 where PostgreSQL upserts with Bytes primary keys fail on the second call.

## Problem
- Works in v6.12.0, broken in v7.0.1
- String ID upserts work fine, only Bytes IDs affected
- Error: "No record was found for an upsert"

## Root Cause
Inconsistent `Uint8Array` serialization between create and where operations in upsert.

## Solution
Added explicit `Uint8Array` handling in `mapArg()` to ensure consistent Buffer conversion.

## Testing
- [x] Reproduces the exact issue from bug report
- [x] Verifies fix resolves the problem
- [x] No impact on existing functionality

Closes #28824

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures consistent handling of binary (Uint8Array) inputs for database upserts, preventing intermittent upsert failures.

* **Tests**
  * Added a regression test suite and test matrix with a PostgreSQL setup to validate bytes-based upserts, plus a test schema covering byte-array and string IDs.


aqrln:
Thanks a lot for the PR! Could you please also add a regression test in `packages/client/tests/functional`?

dncoder14:
Done! I've added a regression test in packages/client/tests/functional/postgres-bytes-upsert-regression/ that reproduces the exact issue and verifies the fix works correctly.

packages/client/tests/functional/postgres-bytes-upsert-regression/tests.ts
```diff
@@ -0,0 +1,51 @@
...
+    await upsertByteRow() // Should not throw
+    
+    // Verify the record exists and has correct data
+    const result = await prisma.testByteId.findUnique({
+      where: { id: byteId }
+    })
+    expect(result).toBeTruthy()
+    expect(result?.value).toBe('hello world!')
+  })
+})
```
jacek-prisma, line 51: An `optOut` field is required here:
```
{
    optOut: {
      from: ['sqlite', 'mysql', 'mongodb', 'cockroachdb', 'sqlserver'],
      reason: 'This test is only relevant for PostgreSQL as it tests PostgreSQL Bytes behavior.',
    },
  }
```
This is causing some test failures

jacek-prisma:
I think the reproduction is valid, but it looks like the fix doesn't work when I run the test on my machine.

I've got an alternative fix in https://github.com/prisma/prisma/pull/28913

PR #28850, opened 2025-12-04
fix(client): ensure log events are emitted in span context in RemoteE…

deepakpathik:
Problem

In RemoteExecutor, log events were being dispatched separately from spans. Because of this, logs were not properly linked to their corresponding trace spans, resulting in missing context such as trace IDs and span IDs. This gap was also noted by a FIXME in RemoteExecutor.ts.

Solution

This PR updates the TracingHelper interface and its ActiveTracingHelper implementation to support handling log events together with spans.

dispatchEngineSpans now accepts an optional logs argument.

ActiveTracingHelper correlates these logs with spans by matching their spanId.

Any logs that match a span are added to the OpenTelemetry span as events using span.addEvent().

Changes

@prisma/instrumentation-contract: Updated the TracingHelper interface so dispatchEngineSpans can receive logs.

@prisma/instrumentation: Updated ActiveTracingHelper to use the new signature and attach logs to spans.

@prisma/client: Updated RemoteExecutor to forward extensions.logs to dispatchEngineSpans.

Tests:

Added a new unit test ActiveTracingHelper.test.ts to ensure logs are correctly attached to spans.

Updated existing mocks to match the updated interface.

Verification

Added a dedicated test in packages/instrumentation/src/ActiveTracingHelper.test.ts.

Confirmed that span.addEvent is called with the correct message and attributes when logs are provided.

Successfully ran pnpm test in packages/instrumentation and all tests passed.

## Summary by CodeRabbit

* **New Features**
  * Logs can now be associated with distributed trace spans, improving correlation between logs and traces for clearer debugging and observability.

* **Tests**
  * Added unit tests verifying log events are attached to trace spans, covering cases with and without message fields and ensuring attributes and timestamps propagate correctly.

### DIFF
--- a/packages/client/src/runtime/core/engines/client/RemoteExecutor.ts
+++ b/packages/client/src/runtime/core/engines/client/RemoteExecutor.ts
@@ -227,7 +227,7 @@ export class RemoteExecutor implements Executor {
       // not use `dispatchEngineSpans` here at all and emit the spans directly.
       // The second option is probably better long term so we can get rid of
       // `dispatchEngineSpans` entirely when QC is in GA and the QE is gone.
-      this.#tracingHelper.dispatchEngineSpans(extensions.traces)
+      this.#tracingHelper.dispatchEngineSpans(extensions.traces, extensions.logs)
     }
   }
 
--- a/packages/client/src/runtime/core/tracing/TracingHelper.ts
+++ b/packages/client/src/runtime/core/tracing/TracingHelper.ts
@@ -1,11 +1,12 @@
 import type { Context } from '@opentelemetry/api'
-import {
+import type {
   EngineSpan,
+  EngineTraceEvent,
   ExtendedSpanOptions,
-  getGlobalTracingHelper,
   SpanCallback,
   TracingHelper,
 } from '@prisma/instrumentation-contract'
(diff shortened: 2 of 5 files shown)


tensordreams:
Thank you for this, @deepakpathik, and apologies for the long silence — this deserved a faster reply than it got.

The same bug (the `FIXME` in `RemoteExecutor` about log events being dispatched outside their span context) has since been fixed on `main` via #28892. That change took the same route yours did: `TracingHelper.dispatchEngineSpans` now receives the log events alongside the spans, and `ActiveTracingHelper` correlates them by `spanId` — it additionally emits each event while its span is active so `$on` handlers see the right context, maps `error` events to `recordException`, and makes sure events whose span was filtered out by `ignoreSpanTypes` are still emitted rather than dropped. It also landed `packages/instrumentation/src/ActiveTracingHelper.test.ts`, the very file you added here.

Since the fix is now on `main`, I'm closing this as a duplicate. Your diagnosis was correct and your approach was the right one — it simply lost the race. Thank you for the effort.

deepakpathik:
Thanks for the update, @tensordreams — appreciate you closing the loop on it. Good to know the fix that landed also handles the $on context and ignoreSpanTypes cases, which is more thorough than what I had here.

I'll take a look at #28892 to see the full implementation.

PR #28851, opened 2025-12-04
refactor: make auth client credential storage injectable

deepakpathik:
Problem

The createAuthenticatedManagementApiClient function in auth-client.ts was tightly coupled to the file-system–based loadCredentials and saveCredentials functions. A TODO comment noted the need to make credential storage injectable so that the auth client could be used with different storage mechanisms.

Solution

This PR refactors createAuthenticatedManagementApiClient to accept an optional CredentialStorage implementation. This introduces proper dependency injection, making it easier to swap out storage providers—for example, during testing or when running in non-file-system environments—while also decoupling the auth client from the underlying storage mechanism.

Changes

packages/cli/src/management-api/credentials.ts

Added a new CredentialStorage interface.

Exported a defaultCredentialsStorage that uses the existing file-based credential loading and saving logic.

packages/cli/src/management-api/auth-client.ts

Updated createAuthenticatedManagementApiClient to accept a storage parameter, defaulting to defaultCredentialsStorage.

packages/cli/src/management-api/auth-client.test.ts

Added new tests to verify that the auth client correctly calls the provided storage for loading and saving credentials.

Verification

Implemented new unit tests in packages/cli/src/management-api/auth-client.test.ts.

Confirmed that the injected storage methods are invoked as expected during the authentication flow.

Ran pnpm --filter prisma test src/management-api/auth-client.test.ts and all tests passed.

## Summary by CodeRabbit

* **New Features**
  * Tracing now supports attaching per-span log events for improved observability.

* **Refactor**
  * Credential storage made configurable to improve how authentication tokens are persisted.

* **Tests**
  * Added unit tests covering authentication flows and tracing behavior.

### DIFF
--- a/packages/cli/src/management-api/auth-client.test.ts
+++ b/packages/cli/src/management-api/auth-client.test.ts
@@ -1,0 +1,63 @@
+/* eslint-disable @typescript-eslint/unbound-method */
+import { login, refreshToken } from './auth'
+import { createAuthenticatedManagementApiClient } from './auth-client'
+import { createManagementApiClient } from './client'
+import { CredentialStorage } from './credentials'
+
+jest.mock('./auth')
+jest.mock('./client')
+
+describe('createAuthenticatedManagementApiClient', () => {
+  const mockLogin = login as jest.Mock
+  const mockRefreshToken = refreshToken as jest.Mock
+  const mockCreateClient = createManagementApiClient as jest.Mock
+
+  const mockStorage: CredentialStorage = {
+    load: jest.fn(),
+    save: jest.fn(),
+  }
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should use storage.load() to get credentials', async () => {
+    const credentials = { token: 'token', refreshToken: 'refresh' }
+    ;(mockStorage.load as jest.Mock).mockResolvedValue(credentials)
+
+    await createAuthenticatedManagementApiClient({ utmMedium: 'test' }, mockStorage)
+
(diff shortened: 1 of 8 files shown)


tensordreams:
Thank you for this, @deepakpathik, and sorry for the long wait.

Both halves of this branch have been overtaken by work on `main`:

- **Credential storage is now injectable.** #28878 moved the CLI onto `@prisma/management-api-sdk`, deleted `packages/cli/src/management-api/credentials.ts` entirely, and replaced the hard-wired `loadCredentials`/`saveCredentials` calls with a `TokenStorage` interface passed into `createManagementApiSdk({ tokenStorage })`. `FileTokenStorage` in `token-storage.ts` is now just one implementation of it, so the seam you were opening up here already exists — the file this PR edits is gone from `main`.
- **The tracing commits** riding along on this branch (`ActiveTracingHelper`, `TracingHelper`, `RemoteExecutor`, `instrumentation-contract`) landed via #28892, which threads log events through `dispatchEngineSpans` and emits them inside their span's context.

Since there is nothing left here that isn't on `main` in one form or another, I'm closing this as a duplicate. The instinct — decouple the client from a concrete credential store so it can be tested — was the right one, and it's the shape `main` ended up with. Thanks for the contribution.

deepakpathik:
Thanks for the thorough explanation, @tensordreams — no worries on the timing. Glad to hear the storage injection idea ended up landing in a cleaner form through the SDK init path; makes sense to tie it to createManagementApiSdk rather than the auth client directly.

I'll check out #28878 and #28892 to see how it was implemented. Thanks again for taking the time to look at this and explain the context — happy to contribute further if there are other areas where it'd be useful.

PR #28861, opened 2025-12-05
fix: isolate sqlcommenter plugins from each other

aqrln:
Make sure one sqcommenter plugin can't poison the context passed to all subsequent ones.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests validating SQL commenter context isolation and immutability across top-level, deeply nested properties, and arrays for both single and compacted queries.

* **Bug Fixes**
  * Ensure each plugin receives an isolated copy of the SQL commenter context so mutations in one plugin do not affect others or shared state.

### DIFF
--- a/packages/client-engine-runtime/src/sql-commenter.ts
+++ b/packages/client-engine-runtime/src/sql-commenter.ts
@@ -1,6 +1,22 @@
 import type { SqlCommenterContext, SqlCommenterPlugin } from '@prisma/sqlcommenter'
 
 /**
+ * Recursively freezes an object and all its nested properties.
+ * This prevents plugins from mutating the context in a way that affects other plugins.
+ */
+function deepFreeze<T extends object>(obj: T): Readonly<T> {
+  Object.freeze(obj)
+
+  for (const value of Object.values(obj)) {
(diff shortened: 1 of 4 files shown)


packages/client-engine-runtime/src/sql-commenter.ts
```diff
@@ -1,5 +1,21 @@
...
 
+/**
+ * Recursively freezes an object and all its nested properties.
+ * This prevents plugins from mutating the context in a way that affects other plugins.
+ */
+function deepFreeze<T extends object>(obj: T): Readonly<T> {
+  Object.freeze(obj)
+
+  for (const value of Object.values(obj)) {
+    if (value !== null && typeof value === 'object' && !Object.isFrozen(value)) {
```
jacek-prisma: if someone manually freezes an object, but not it's nested objects, this check would prevent this from recursing down to them, which seems incorrect

aqrln:
@jacek-prisma I changed the code to clone the context using `klona` instead of freezing it. This seems simpler and safer, handles all edge cases, and addresses your and coderabbit's comments.

jacek-prisma (approved):

PR #28886, opened 2025-12-08
fix(client): fix tracing tests failing with QPE

aqrln:
This PR fixes tracing tests broken with QPE for three reasons:

1. The mismatch between the expected and actual field names (`traces` vs `spans`) in the result extensions. `RemoteExecutor` used the outdated name previously used by QE, not the new one used by QPE.
2. Functional tests started the QPE server in the same Node.js process, which led to internal (non-exported) QPE spans being picked up by the `SpanProcessor` set up in tests. This PR isolates the QPE server in a worker thread.
3. QPE had unnecessary extra spans created at the top level in `app.ts`, although they are already created by `QueryInterpreter` and `Transactionmanager`, which led to duplicate spans (i.e. `db_query` within `db_query`, `start_transaction` within `span_transaction` etc).

Additionally, this change renames the QPE span to have `prisma:accelerate:` prefix instead of `prisma:engine:`.

Closes https://linear.app/prisma-company/issue/TML-1395/fix-missing-spans-in-tracing-tests-with-qcaccelerate

## Summary by CodeRabbit

* **New Features**
  * Test infrastructure now runs the query-plan executor in a worker for improved isolation and Node 20 compatibility.

* **Tests**
  * Standardized telemetry span names to the new accelerate prefix.
  * Span filtering updated so different spans are now captured or suppressed, affecting which spans appear in traces.

* **Refactor**
  * Removed redundant tracing wrappers around core executor methods.
  * Engine events are now dispatched as spans instead of traces.

### DIFF
--- a/packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
+++ b/packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
@@ -125,8 +125,17 @@ function setupTestSuiteMatrix(
         if (clientMeta.clientEngineExecutor === 'remote') {
           qpeWorker = new Worker(path.join(__dirname, 'qpe-worker-entry.cjs'))
 
+          const qpeStartupTimeoutMs = 10_000
+
           const { hostname, port } = await new Promise<QpeWorkerReadyResponse>((resolve, reject) => {
+            const timeoutId = setTimeout(async () => {
+              await qpeWorker!.terminate()
+              qpeWorker = undefined
+              reject(new Error(`QPE worker startup timed out after ${qpeStartupTimeoutMs}ms`))
+            }, qpeStartupTimeoutMs).unref()
+
             qpeWorker!.once('message', (response: QpeWorkerResponse) => {
+              clearTimeout(timeoutId)
               if (response.type === 'ready') {
                 resolve(response)
               } else if (response.type === 'error') {
--- a/packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
+++ b/packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
@@ -136,10 +136,15 @@ function setupTestSuiteMatrix(
 
             qpeWorker!.once('message', (response: QpeWorkerResponse) => {
               clearTimeout(timeoutId)
-              if (response.type === 'ready') {
-                resolve(response)
-              } else if (response.type === 'error') {
-                reject(new Error(response.message))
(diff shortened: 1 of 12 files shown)


packages/client/tests/functional/_utils/setupTestSuiteMatrix.ts
```diff
@@ -114,49 +115,33 @@ function setupTestSuiteMatrix(
...
+            qpeWorker!.postMessage({
+              type: 'start',
+              databaseUrl: datasourceInfo.databaseUrl,
+            } satisfies QpeWorkerStartMessage)
+          })
 
-          server = { qpe, net }
-          datasourceInfo.accelerateUrl = `prisma://${hostname}:${address.port}/?api_key=1&use_http=1`
+          datasourceInfo.accelerateUrl = `prisma://${hostname}:${port}/?api_key=1&use_http=1`
         }
```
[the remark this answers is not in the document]
aqrln: Since the message handler itself handles errors in the worker, we probably don't need to do this here in the first place. If an error event happens, it should be outside of the message handler, so handling it here won't help anyway.

packages/client/tests/functional/tracing/tests.ts
```diff
@@ -33,15 +33,17 @@ function buildTree(rootSpan: ReadableSpan, spans: ReadableSpan[]): Tree {
...
+          if (
+            [
+              `${prefix}:db_query`,
+              `${prefix}:start_transaction`,
+              `${prefix}:commit_transaction`,
+              `${prefix}:rollback_transaction`,
+            ].includes(name)
+          ) {
+            return 'prisma:engine:' + name.slice(`${prefix}:`.length)
+          }
```
jacek-prisma: instead of doing this, I'd define a const in the test
```ts
const clientSpanName = clientEngineExecutor === 'remote' ? 'accelerate' : 'client'
```
and then just have
```ts
          [
            `prisma:${clientSpanName}:db_query`,
             ...
          ].includes(name)
```

it could also simplify other code like replace `maybeRemoteSpan('db_query')` with just `prisma:${clientSpanName}:db_query`

jacek-prisma (dismissed):

packages/client/tests/functional/tracing/tests.ts
aqrln: that's much better, thanks

jacek-prisma (approved):

PR #28893, opened 2025-12-09
ci: allow publishing with custom NPM tag

jacek-prisma:
[TML-1637](https://linear.app/prisma-company/issue/TML-1637/modify-release-workflow-to-support-prisma-6)

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying a custom distribution tag during release, allowing explicit control over the publish tag.

* **Chores**
  * Release workflow updated to accept and propagate a custom dist-tag input.
  * Publishing flow now reads the custom tag input and applies it to the publish step, with clearer logging of the chosen tag.

### DIFF
--- a/scripts/ci/publish.ts
+++ b/scripts/ci/publish.ts
@@ -462,7 +462,7 @@ async function publish() {
     '--dry-run': Boolean,
     '--release': String, // TODO What does that do? Can we remove this? probably
     '--test': Boolean,
-    '--skip-npm-tag': Boolean,
+    '--custom-dist-tag': String,
   })
 
   if (!process.env.GITHUB_REF_NAME) {
@@ -487,8 +487,8 @@ async function publish() {
     // and then replace the args['--release'] with it
   }
 
-  if (process.env.SKIP_NPM_TAG === 'true') {
-    args['--skip-npm-tag'] = true
+  if (process.env.CUSTOM_DIST_TAG) {
+    args['--custom-dist-tag'] = process.env.CUSTOM_DIST_TAG
   }
 
   if (!args['--test'] && !args['--publish'] && !dryRun) {
@@ -558,8 +558,9 @@ async function publish() {
     tag = 'dev'
   }
 
-  if (args['--skip-npm-tag']) {
-    tag = undefined
+  if (args['--custom-dist-tag']) {
+    console.log(`Using custom dist tag: ${args['--custom-dist-tag']}`)
+    tag = args['--custom-dist-tag']
   }
 
   console.log({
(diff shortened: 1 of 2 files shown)


scripts/ci/publish.ts
```diff
@@ -583,8 +592,8 @@ async function publish() {
       }
       const passing = await areEcosystemTestsPassing(tagForEcosystemTestsCheck)
       if (!passing && process.env.SKIP_ECOSYSTEMTESTS_CHECK !== 'true') {
-        throw new Error(`We can't release, as the ecosystem-tests are not passing for the ${tag} npm tag!
-Check them out at https://github.com/prisma/ecosystem-tests/actions?query=workflow%3Atest+branch%3A${tag}`)
+        throw new Error(`We can't release, as the ecosystem-tests are not passing for the ${tag ?? 'N/A'} npm tag!
```
aqrln: What do we even check in this script if we don't have a tag to map to the branch?

scripts/ci/publish.ts
```diff
@@ -803,7 +812,7 @@ async function publishPackages(
          *  - Your working directory is clean (there are no uncommitted changes).
          *  - The branch is up-to-date.
          */
-        await run(pkgDir, `pnpm publish --no-git-checks --access public --tag ${tag}`, dryRun)
+        await run(pkgDir, `pnpm publish --no-git-checks --access public${tag ? ` --tag ${tag}` : ''}`, dryRun)
```
aqrln: I believe this will publish on the `latest` tag by default. I'm not sure if pnpm as the client and the npm registry allow it, but in npm CLI you can't actually publish a package without a dist-tag at all, you can only remove the dist-tag from the package after publishing.

I also think that we should, actually, have a dedicated dist-tag for the 6.x release line (`classic`/`old`/`prev`/whatever).

scripts/ci/publish.ts
jacek-prisma: This checks the the status of the latest dev build in the ecosystem tests (always `dev` or `patch-dev`), it doesn't actually use the `tag` value

jacek-prisma: All patch branches get published under one dist-tag, even if they're for different major/minor versions so this check doesn't seem very reliable, it'll always check the latest patch-dev build, not necessarily the one that's being published - that's how it's already been implemented though

scripts/ci/publish.ts
jacek-prisma: I've checked and you're right, we need to publish it under a custom tag, I'll modify to accept a custom tag

We could have some automated logic to figure out whether we're publishing the latest major version and automatically set the tag but I feel like that might be too automagical for the release script and it should just be an explicit argument

jacek-prisma: Changed it to accept a custom dist tag @aqrln, lmk what you think

aqrln (approved):

scripts/ci/publish.ts
aqrln: lgtm

PR #28895, opened 2025-12-10
feat(config): allow undefined urls in e.g. `prisma generate`

jkomyno:
This PR:
- fixes [TML-1673](https://linear.app/prisma-company/issue/TML-1673/prismaconfigts-do-not-throw-validation-errors-if-env-variable-isnt)
- adds a new e2e test, `prisma-config-generate-doesnt-need-valid-url`
- updates the `import { env } from 'prisma/config'`'s error message
  - before: `"Missing required environment variable: UNDEFINED_VARIABLE"`
  - after: `"Cannot resolve environment variable: UNDEFINED_VARIABLE."`

After this PR:

- `prisma generate` works when the `datasource.url` value in the Prisma config file is undefined:
  ```ts
  // prisma.config.ts
  
  export default defineConfig({
    datasource: {
      url: process.env['UNDEFINED_VARIABLE'],
    },
  })
  ```

- `prisma generate` still fails when the `datasource.url` value in the Prisma config file uses the `env` utility with an undefined variable:
  ```ts
  // prisma.config.ts
  import { defineConfig, env } from 'prisma/config'
  
  export default defineConfig({
    datasource: {
      url: env('UNDEFINED_VARIABLE'),
    },
  })
  ```

  The error looks like: `Failed to load config file "[path]/prisma.config.ts" as a TypeScript/JavaScript module. Error: PrismaConfigEnvError: Missing required environment variable: UNDEFINED_VARIABLE`.

- `prisma db push` still fails when  the `datasource.url` value in the Prisma config file is undefined.
  The error looks like: `Error: The datasource.url property is required in your Prisma config file when using prisma db push.`.

## Release Notes

* **New Features**
  * Datasource URL is now optional in Prisma configuration files, allowing prisma generate to succeed without a valid database URL.

* **Bug Fixes**
  * Enhanced error messages to specify exactly which configuration property is required (datasource.url instead of datasource).
  * Improved environment variable resolution error messages for clarity when variables cannot be resolved.


igalklebanov:
Hey 👋 

This still failing for `env(...)` usage, which is the main path according to docs, is not really fixing the problem (unless I'm missing something).

If a command doesn't require a config thing, it should work when the thing is not available - regardless how the thing is provided (e.g. the `env()` helper). Anything else is bad DX.

The responsibility of deciding what should throw should be the command's. The config loader should do best effort and report (not throw) missing things.

jkomyno:
> Hey 👋
> 
> This still failing for `env(...)` usage, which is the main path according to docs, is not really fixing the problem (unless I'm missing something).
> 
> If a command doesn't require a config thing, it should work when the thing is not available - regardless how the thing is provided (e.g. the `env()` helper). Anything else is bad DX.
> 
> The responsibility of deciding what should throw should be the command's. The config loader should do best effort and report (not throw) missing things.

That's by design. `env()`'s whole purpose is to raise the error `PrismaConfigEnvError: Missing required environment variable $VARIABLE_NAME` when `process.env[$VARIABLE_NAME]` resolves to `undefined`. And, since it was shipped in Prisma 7.0.0, we can't suddenly break this behavior.

Now, I've already shared with the docs team (in the past 10 days or so) that `env()` should NOT be the default option shown in our guides / tutorials.

What this PR does, is to allow the `process.env[$VARIABLE_NAME] == undefined` case not to trigger an error in `prisma.config.ts`' loader, which is an improvement over the previous DX, and the main point of [TML-1673](https://linear.app/prisma-company/issue/TML-1673/prismaconfigts-do-not-throw-validation-errors-if-env-variable-isnt) altogether.

igalklebanov:
> That's by design. env()'s whole purpose is to raise the error PrismaConfigEnvError: Missing required environment variable $VARIABLE_NAME when process.env[$VARIABLE_NAME] resolves to undefined.

But that doesn't matter for commands that don't require or use the config property it was invoked for. Nobody's expecting `env()` should throw for `datasource.url` if I run `prisma generate`. That's too rigid.

Sometimes people run these commands without providing the envars they normally would for others. e.g. a GitHub workflow that generates the client and then runs `tsc --noEmit` should not have access to a connection string or other secrets.

igalklebanov (approved):

PR #28903, opened 2025-12-11
feat(cli): customize `prisma init` based on the JS runtime (Bun vs other)

jkomyno:
This PR:
- closes [TML-1672](https://linear.app/prisma-company/issue/TML-1672/prisma-init-customize-bun-experience-add-bun-support-to-e2e-tests)
- updates the `prisma.config.ts` file generated by `prisma init`
- adds Bun support to `e2e` tests
- adds a new `prisma-init-bun` e2e test

**Before this PR**

`prisma init` would unconditionally generate the following `prisma.config.ts` file:

```typescript
// This file was generated by Prisma and assumes you have installed the following:
// npm install --save-dev prisma dotenv
import "dotenv/config";

export default defineConfig({
  migrations: {
    path: "prisma/migrations",
  },
  datasource: {
    url: env("DATABASE_URL"),
  },
});
```

**After this PR**

`prisma init` generates different `prisma.config.ts` files, depending on the JS runtime used.

- Bun, via `bun run --bun prisma [command]`

  ```ts
  // This file was generated by Prisma, and assumes you run Prisma commands using \`bun --bun run prisma [command]\`.

  export default defineConfig({
    migrations: {
      path: "prisma/migrations",
    },
    datasource: {
      url: env("DATABASE_URL"),
    },
  });
  ```

  Note that Bun automatically loads `.env` variables for us.

- Node.js / other, via e.g. `npx prisma [command]`, `pnpm prisma [command]`, `bunx run prisma [command]`, `bun run prisma [command]` etc

  ```ts
  // This file was generated by Prisma and assumes you have installed the following:
  // npm install --save-dev prisma dotenv
  import "dotenv/config";

  export default defineConfig({
    migrations: {
      path: "prisma/migrations",
    },
    datasource: {
      url: process.env["DATABASE_URL"],
    },
  });
  ```

  Notice how `env()` is no longer used by default.

---

Note: the only guaranteed way I found of running Prisma via Bun is `bun run --bun prisma [command]` (as per https://github.com/oven-sh/bun/issues/3417#issuecomment-1927842420). `bunx run prisma [command]` and `bun run prisma [command]` both use the Node.js runtime due to our [Node.js-specific shebang](https://github.com/prisma/prisma/blob/dbada5738682d577ea9b4f5c66ce435486a2f360/packages/cli/helpers/build.ts#L48) (see https://github.com/oven-sh/bun/issues/3417).

## Release Notes

* **New Features**
  * Added Bun runtime support to `prisma init` command, which now generates runtime-appropriate configuration for Bun environments.

* **Tests**
  * Added end-to-end tests to verify correct configuration generation when using Bun as the runtime.
  * Enhanced test infrastructure with Bun installation and configuration in Docker and test environments.


packages/cli/src/Init.ts
```diff
@@ -216,16 +216,26 @@ function normalizePath(configPath: string) {
...
+type DefaultConfigInput = {
+  /**
+   * The path to the `prisma` folder.
+   */
+  prismaFolder: string
+  /**
+   * The runtime to use.
+   * Currently, this is only used to customize the Bun experience.
+   */
+  runtime: 'bun' | (string & {})
```
aqrln: I'd probably prefer a closed enum like `'bun' | 'other'`, the current type is just a string with extra steps and no additional type safety.

packages/client/tests/e2e/_utils/standard.cmd.sh
```diff
@@ -19,6 +19,10 @@ export NODE_PATH=$(npm root --quiet -g) # make global packages available
 export NEXT_TELEMETRY_DISABLED=1
 export NO_COLOR=1
 
+# Ensure Bun is in PATH (in case it's not inherited from Dockerfile ENV)
```
aqrln: I believe the behavior of whether it's inherited or not should be well defined, so either this whole block is unnecessary, or the comment should be updated to be more assertive (e.g. "because it's not inherited...").

packages/cli/src/Init.ts
igalklebanov, line 46: nitpicky

consider outsourcing maintenance of this knowledge to [`std-env`](https://www.npmjs.com/package/std-env).

packages/cli/src/Init.ts
```diff
@@ -216,16 +216,26 @@ function normalizePath(configPath: string) {
...
 
+export const defaultConfig = ({ prismaFolder, runtime }: DefaultConfigInput) => {
   const schemaPath = path.relative(process.cwd(), path.join(prismaFolder, 'schema.prisma'))
   const migrationsPath = path.relative(process.cwd(), path.join(prismaFolder, 'migrations'))
 
-  return `\
-// This file was generated by Prisma and assumes you have installed the following:
-// npm install --save-dev prisma dotenv
-import "dotenv/config";
+  const configSrc = match({ runtime })
```
igalklebanov, line 235: nitpicky

do we really need `ts-pattern` here? seems forced.

packages/internals/src/cli/getSchema.ts
```diff
@@ -109,8 +109,7 @@ export async function getSchemaWithPathOptional({
 }
 
 export function printSchemaLoadedMessage(schemaPath: string) {
-  // TODO: this causes https://github.com/prisma/prisma/issues/27005
-  process.stdout.write(dim(`Prisma schema loaded from ${path.relative(process.cwd(), schemaPath)}`) + '\n')
+  process.stderr.write(dim(`Prisma schema loaded from ${path.relative(process.cwd(), schemaPath)}.`) + '\n')
```
igalklebanov: why is this written to `stderr`?

packages/cli/src/Init.ts
```diff
@@ -609,6 +640,7 @@ export class Init implements Command {
       path.join(outputDir, 'prisma.config.ts'),
       defaultConfig({
         prismaFolder,
+        runtime: isBun ? 'bun' : 'other',
```
igalklebanov, line 643: nitpicky

consider `std-env`'s `runtime` property.

igalklebanov (dismissed):

packages/internals/src/cli/getSchema.ts
aqrln: This looks like a change from [another Alberto's pull request](https://github.com/prisma/prisma/pull/28911), it looks like this one unintentionally includes commits from the other one.

jkomyno: Good catch. Sorry about that. I removed those merge commits from my other PR.

igalklebanov (approved):

PR #28909, opened 2025-12-12
feat(adapter-mariadb): Improve MariaDB adapter error reporting by surfacing underlying error causes

### LINKED ISSUE #28612: MySQL pool timeout using adapter (rust free engine)
### Bug description

I'm trying to upgrade prisma from version 6.17.0 to 7.0.0. 
I'm using MySQL with the PrismaMariaDb driver adapter

```typescript
const databaseUrl = process.env.DATABASE_URL;

const adapter = new PrismaMariaDb(databaseUrl);

``` 

I'm getting following error while trying to get data from the database:

`pool timeout: failed to retrieve a connection from pool after 10001ms (pool connections: active=0 idle=0 limit=10)`

### Severity

### Reproduction

Install the mariadb adapter, npx prisma generate, try a simple query, and timeout

No timeout error is thrown and I can query data from my database.

### Frequency

Consistently reproducible

### Is this a regression?

n/a

### Workaround

n/a

### Prisma Schema & Queries

```typescript
generator client {
  provider   = "prisma-client"
  output     = "../prisma/client"
  engineType = "client"
}

datasource db {
  provider = "mysql"
}
``` 

### Prisma Config

```ts
import dotenv from '@dotenvx/dotenvx';
import { defineConfig, env } from "@prisma/config";

dotenv.config();

export default defineConfig({
    datasource: {
        url: env('DATABASE_URL'),
    },
});
```

### Logs & Debug Info

No response

### Environment & Setup

- **Database:** MySQL
- **Node.js version:** v20.19.5

### Prisma Version

```
7.0.0
```

alireza-malek:
**Summary**
The `mariadb` driver wraps connection‑level failures in a pool timeout error, but it also includes a nested `cause` with the real reason. This PR updates `@prisma/adapter-mariadb` error mapping so MySQL/MariaDB driver errors include the underlying driver cause (when present), instead of only surfacing the top-level pool timeout. Closes #28612

**Why**
[Users were hitting errors like “pool timeout”](https://github.com/prisma/prisma/issues/28612) where the real root cause was attached by the MariaDB driver as a nested cause (e.g. socket connect timeout, RSA public key retrieval). Prisma surfaced only the outer message, so people couldn’t diagnose what to fix.

**Behavior before vs after**

- **Before:**

  ```text
  DriverAdapterError: pool timeout: failed to retrieve a connection from pool after 10008ms
      (pool connections: active=0 idle=0 limit=5)
    cause: {
      originalCode: '45028',
      originalMessage: 'pool timeout: failed to retrieve a connection from pool after 10008ms (pool connections: ...)',
      kind: 'mysql',
      code: 45028,
      message: 'pool timeout: failed to retrieve a connection from pool after 10008ms (pool connections: ...)',
      state: 'HY000'
    }
  ```

  The underlying message about RSA public key retrieval or socket timeout was completely hidden.

- **After:**

  ```text
  DriverAdapterError: pool timeout: failed to retrieve a connection from pool after 10008ms
      (pool connections: active=0 idle=0 limit=5)
    cause: {
      originalCode: '45028',
      originalMessage: 'pool timeout: failed to retrieve a connection from pool after 10008ms (pool connections: ...)',
      kind: 'mysql',
      code: 45028,
      message: 'pool timeout: failed to retrieve a connection from pool after 10008ms (pool connections: ...)',
      state: 'HY000',
      cause: 'SqlError: (conn:44, no: 45044, SQLState: 08S01) RSA public key is not available client side. Either set option `cachingRsaPublicKey` to indicate public key path, or allow public key retrieval with option `allowPublicKeyRetrieval`'
    }
  ```

  Also applies to network issues like:

  ```text
    ...
    cause: 'SqlError: (conn:-1, no: 45012, SQLState: 08S01) Connection timeout: failed to create socket after 1003ms'
  }
  ```

Now users can immediately see whether the problem is TLS/SSL, authentication (RSA key retrieval), connectivity (socket timeout), etc., instead of only a generic pool timeout.

---

**Notes**

- This follows the existing convention in `driver-adapter-utils` of exposing additional detail in a `cause: string` field.
- It avoids concatenating multiple messages into `message`, keeping the top-level error text focused while still surfacing full context via `cause`.
- It is **backwards compatible**: It specifically fixes the diagnosability issues described in [#28612](https://github.com/prisma/prisma/issues/28612) without changing how errors are raised or handled elsewhere.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MariaDB/MySQL adapter error reporting to include an optional underlying-cause field and to propagate nested error messages, providing richer diagnostic context in logs and error payloads.
* **Tests**
  * Added tests validating error-mapping behavior with and without nested driver causes to ensure cause propagation and omission behave as expected.

### DIFF
--- a/packages/adapter-mariadb/src/errors.ts
+++ b/packages/adapter-mariadb/src/errors.ts
@@ -1,4 +1,5 @@
 import { Error as DriverAdapterErrorObject, MappedError } from '@prisma/driver-adapter-utils'
+import type { SqlError } from 'mariadb/types/share'
 
 export function convertDriverError(error: unknown): DriverAdapterErrorObject {
   if (isDriverError(error)) {
@@ -136,6 +137,7 @@ export function mapDriverError(error: DriverError): MappedError {
         code: error.errno,
         message: error.sqlMessage ?? 'N/A',
         state: error.sqlState ?? 'N/A',
+        // eslint-disable-next-line @typescript-eslint/no-base-to-string
         cause: error.cause && 'toString' in error.cause ? error.cause.toString() : undefined,
       }
   }
@@ -145,7 +147,7 @@ type DriverError = {
   errno: number
   sqlMessage: string | null
   sqlState: string | null
-  cause?: any
+  cause?: SqlError
 }
 
 function isDriverError(error: any): error is DriverError {
--- a/packages/adapter-mariadb/src/errors.ts
+++ b/packages/adapter-mariadb/src/errors.ts
@@ -1,5 +1,5 @@
 import { Error as DriverAdapterErrorObject, MappedError } from '@prisma/driver-adapter-utils'
-import type { SqlError } from 'mariadb/types/share'
+import type { SqlError } from 'mariadb'
 
 export function convertDriverError(error: unknown): DriverAdapterErrorObject {
   if (isDriverError(error)) {
--- a/packages/adapter-mariadb/src/errors.ts
+++ b/packages/adapter-mariadb/src/errors.ts
@@ -137,8 +137,7 @@ export function mapDriverError(error: DriverError): MappedError {
         code: error.errno,
         message: error.sqlMessage ?? 'N/A',
         state: error.sqlState ?? 'N/A',
-        // eslint-disable-next-line @typescript-eslint/no-base-to-string
-        cause: error.cause && 'toString' in error.cause ? error.cause.toString() : undefined,
+        cause: error.cause?.message ?? undefined,
       }
   }
 }
--- a/packages/adapter-mariadb/src/errors.ts
+++ b/packages/adapter-mariadb/src/errors.ts
@@ -136,6 +136,7 @@ export function mapDriverError(error: DriverError): MappedError {
         code: error.errno,
         message: error.sqlMessage ?? 'N/A',
         state: error.sqlState ?? 'N/A',
+        cause: error.cause && 'toString' in error.cause ? error.cause.toString() : undefined,
       }
   }
 }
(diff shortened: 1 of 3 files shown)


packages/driver-adapter-utils/src/types.ts
```diff
@@ -173,6 +173,7 @@ export type MappedError =
       code: number
       message: string
       state: string
+      cause: string | undefined
```
jacek-prisma: this being a required field is causing type errors, it should probably be `cause?: string`

jacek-prisma:
Thanks for the contribution @alireza-malek!
I left one comment, the new field is causing some type errors, but it can be fixed pretty easily. Also I think the coderabbit comment is valid.

packages/adapter-mariadb/src/errors.ts
```diff
@@ -136,6 +137,7 @@ export function mapDriverError(error: DriverError): MappedError {
         code: error.errno,
         message: error.sqlMessage ?? 'N/A',
         state: error.sqlState ?? 'N/A',
+        cause: error.cause?.message ?? undefined,
```
jacek-prisma, line 140: it'd be nice to have a test case for this, either
- a functional test in `packages/client/tests/functional` that uses the real client and triggers an error
- or at least a unit test for `mapDriverError` in the adapter, analogous to stuff in `mariadb.test.ts` (this is slightly worse since it doesn't test the real client but easier)

jacek-prisma (approved):

packages/adapter-mariadb/src/errors.ts
alireza-malek: Added tests for `convertDriverError`. Functional is possible, but in this specific case it’s tricky and not worth it IMO to trigger a stable network failure error in CI.
If you can provide me with examples from other adapters, I would check that out.

delboy1978uk:
I'm looking forward to seeing this, can anyone say when this will be released? :-D

PR #28911, opened 2025-12-12
fix(cli): make `prisma version --json` emit JSON only to stdout

jkomyno:
This PR:
- closes https://github.com/prisma/prisma/issues/27005
- closes [TML-1257](https://linear.app/prisma-company/issue/TML-1257/bug-replace-stdout-diagnostic-with-stderr-fix-27005-which-is-similar)

---

I've also wasted time on `packages/migrate/src/__tests__/DbPull/postgresql-views.test.ts`, which doesn't absolutely work locally for me, and gets stuck in ELOOP errors.

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end test to verify `prisma version --json` produces valid, parseable JSON.

* **Chores**
  * Test image now includes jq to support JSON validation.
  * CLI logging observed in tests: schema-load messages moved to stderr, altering log ordering.

* **Tests**
  * Updated many CLI and migration test snapshots/expectations to reflect stdout/stderr changes.

### DIFF
--- a/packages/client/tests/e2e/prisma-version-json/readme.md
+++ b/packages/client/tests/e2e/prisma-version-json/readme.md
@@ -1,0 +1,15 @@
+# Prisma Version JSON E2E Test
+
+This test verifies that `prisma version --json` outputs valid JSON that can be parsed by `jq`.
+
+## What it tests
+
+- `prisma version --json` outputs valid JSON
+- The JSON output can be successfully parsed by `jq` (exit code 0)
+- The JSON contains expected fields (prisma, platform, engines)
+
+## Running the test
+
+```bash
+pnpm --filter @prisma/client test:e2e prisma-version-json
+```
--- a/packages/client/tests/e2e/prisma-version-json/readme.md
+++ b/packages/client/tests/e2e/prisma-version-json/readme.md
@@ -6,7 +6,7 @@ This test verifies that `prisma version --json` outputs valid JSON that can be p
 
 - `prisma version --json` outputs valid JSON
 - The JSON output can be successfully parsed by `jq` (exit code 0)
-- The JSON contains expected fields (prisma, platform, engines)
+- The JSON contains some expected fields
 
 ## Running the test
 
--- a/packages/client/tests/e2e/prisma-version-json/tests/main.test.ts
+++ b/packages/client/tests/e2e/prisma-version-json/tests/main.test.ts
@@ -1,0 +1,24 @@
+import { execSync } from 'node:child_process'
+
+import { expect, test } from 'vitest'
+
+test('prisma version --json outputs valid JSON', () => {
+  // Run prisma version --json | jq and check exit code
+  // jq will exit with code 0 if JSON is valid, non-zero otherwise
+  try {
+    const output = execSync('pnpm exec prisma version --json | jq .', {
+      encoding: 'utf-8',
+      shell: '/bin/sh',
+    })
+
+    // additional check: verify that the output is valid JSON by parsing it
+    const jsonOutput = output.trim()
+    expect(() => JSON.parse(jsonOutput)).not.toThrow()
+
+    // verify it contains expected fields
+    const parsed = JSON.parse(jsonOutput)
+    expect(parsed).toHaveProperty('prisma')
+  } catch (error: any) {
+    throw new Error(`Command failed with exit code ${error.status}: ${error.message}`)
+  }
+})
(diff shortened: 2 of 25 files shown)


packages/client/tests/e2e/prisma-version-json/readme.md
```diff
@@ -0,0 +1,15 @@
+# Prisma Version JSON E2E Test
+
+This test verifies that `prisma version --json` outputs valid JSON that can be parsed by `jq`.
+
+## What it tests
+
+- `prisma version --json` outputs valid JSON
+- The JSON output can be successfully parsed by `jq` (exit code 0)
+- The JSON contains expected fields (prisma, platform, engines)
```
jkomyno:
```suggestion
- The JSON contains some expected fields
```

packages/client/tests/e2e/prisma-version-json/tests/main.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  // jq will exit with code 0 if JSON is valid, non-zero otherwise
+  try {
+    const output = execSync('pnpm exec prisma version --json | jq .', {
+      encoding: 'utf-8',
+      shell: '/bin/sh',
+    })
+
+    // additional check: verify that the output is valid JSON by parsing it
+    const jsonOutput = output.trim()
+    expect(() => JSON.parse(jsonOutput)).not.toThrow()
```
aqrln: This is superfluous and doesn't assert any new properties. If JSON is not valid, it will also throw and fail the test in the other `JSON.parse` below.

Asserting `.not.toThrow()` only makes sense when you don't care about the result, it just makes the intent more clear than just calling a function for side effects without assigning the result to a variable, but is functionally identical.

aqrln: And if we didn't care about the result and didn't parse it below, this assertion still wouldn't be useful because you would be testing `jq` and not Prisma. It would only make sense if you did this with the raw output of Prisma CLI without feeding it to `jq` first.

packages/client/tests/e2e/prisma-version-json/tests/main.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    // verify it contains expected fields
+    const parsed = JSON.parse(jsonOutput)
+    expect(parsed).toHaveProperty('prisma')
+  } catch (error: any) {
+    throw new Error(`Command failed with exit code ${error.status}: ${error.message}`)
+  }
```
aqrln: I don't think this try/catch block brings any value; on the contrary, it only obscures the actual error and removes useful information like the stacktrace.

aqrln: it is also wrong because it's not limited in scope to the command invocation and will catch all assertion errors like `expect(parsed).toHaveProperty('prisma')`

packages/internals/src/cli/getSchema.ts
```diff
@@ -109,8 +109,7 @@ export async function getSchemaWithPathOptional({
 }
 
 export function printSchemaLoadedMessage(schemaPath: string) {
-  // TODO: this causes https://github.com/prisma/prisma/issues/27005
-  process.stdout.write(dim(`Prisma schema loaded from ${path.relative(process.cwd(), schemaPath)}`) + '\n')
+  process.stderr.write(dim(`Prisma schema loaded from ${path.relative(process.cwd(), schemaPath)}.`) + '\n')
```
igalklebanov, line 112: why is this writing to `stderr`?

aqrln: `stderr` is the correct stream for logs and diagnostics in Unix programs, `stdout` is for program results. Writing this log message to `stdout` made `prisma version --json` unusable, which is exactly the bug this pull request is fixing.

Writing logs to stdout is fine for interactive programs or those that are only used by humans, but anything that outputs data for machine consumption must always print messages for humans to stderr.

aqrln (dismissed):

aqrln (commented):
One existing e2e test is now failing, it currently expects parsing the output to fail and there's a todo comment to flip the assertion when the bug is fixed.

jkomyno:
> One existing e2e test is now failing, it currently expects parsing the output to fail and there's a todo comment to flip the assertion when the bug is fixed.

Right, fixed in https://github.com/prisma/prisma/commit/c3e02fb1d1cabd3086e0a4c7a92bc01321383773.

igalklebanov (approved):

PR #28913, opened 2025-12-12
fix: fix byte upserts by removing legacy byte array representation

### LINKED ISSUE #28824: Prisma v7 breaks postgres Byte upserts

jacek-prisma:
[TML-1683](https://linear.app/prisma-company/issue/TML-1683/fix-byte-array-upsert-issue)

Removes the legacy byte representation that used arrays. The old representation used arrays to make query arguments JSON serializable, which is something we no longer need. It was inefficient and it was also causing an edge case in the query renderer where it would interpret byte arrays as parameter lists.

Fixes https://github.com/prisma/prisma/issues/28824.

## Summary by CodeRabbit

* **Improvements**
  * Binary values are preserved as native binary (Buffer/Uint8Array) across adapters; byte arrays are no longer auto-converted to number arrays or re-encoded.
  * Result value types now accept Uint8Array and runtime truthiness treats Uint8Array like arrays.
  * Bytes/BytesArray no longer receive special serialization/transformation and pass through raw binary.

* **Tests**
  * Added functional tests for upserting records using byte-array IDs (skipped on SQL Server).

### DIFF
--- a/packages/adapter-better-sqlite3/src/conversion.ts
+++ b/packages/adapter-better-sqlite3/src/conversion.ts
@@ -150,11 +150,6 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
   for (let i = 0; i < row.length; i++) {
     const value = row[i]
 
-    if (value instanceof ArrayBuffer) {
-      result[i] = new Uint8Array(value)
-      continue
-    }
-
     // If an integer is required and the current number isn't one,
     // discard the fractional part.
     if (
--- a/packages/adapter-better-sqlite3/src/conversion.ts
+++ b/packages/adapter-better-sqlite3/src/conversion.ts
@@ -150,12 +150,8 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
   for (let i = 0; i < row.length; i++) {
     const value = row[i]
 
-    // Convert array buffers to arrays of bytes.
-    // Base64 would've been more efficient but would collide with the existing
-    // logic that treats string values of type Bytes as raw UTF-8 bytes that was
-    // implemented for other adapters.
-    if (value instanceof ArrayBuffer || value instanceof Buffer) {
-      result[i] = Array.from(new Uint8Array(value))
+    if (value instanceof ArrayBuffer) {
+      result[i] = Buffer.from(value)
       continue
     }
 
@@ -238,10 +234,6 @@ export function mapArg<A>(
 
   if (typeof arg === 'string' && argType.scalarType === 'bytes') {
     return Buffer.from(arg, 'base64')
-  }
-
-  if (Array.isArray(arg) && argType.scalarType === 'bytes') {
-    return Buffer.from(arg)
   }
 
   return arg
--- a/packages/adapter-better-sqlite3/src/conversion.ts
+++ b/packages/adapter-better-sqlite3/src/conversion.ts
@@ -4,7 +4,7 @@ import { PrismaBetterSqlite3Options } from './better-sqlite3'
 
 const debug = Debug('prisma:driver-adapter:better-sqlite3:conversion')
 
-type Value = null | string | number | bigint | ArrayBuffer | Buffer
+type Value = null | string | number | bigint | Buffer
 export type Row = {
   /** Number of columns in this row.
    *
--- a/packages/adapter-better-sqlite3/src/conversion.ts
+++ b/packages/adapter-better-sqlite3/src/conversion.ts
@@ -151,7 +151,7 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
     const value = row[i]
 
     if (value instanceof ArrayBuffer) {
-      result[i] = Buffer.from(value)
+      result[i] = new Uint8Array(value)
       continue
     }
 
--- a/packages/adapter-d1/src/conversion.ts
+++ b/packages/adapter-d1/src/conversion.ts
@@ -107,7 +107,7 @@ export function mapRow<A>(result: (A | ResultValue)[], columnTypes: ColumnType[]
     const value = result[i]
 
     if (value instanceof ArrayBuffer) {
-      result[i] = Buffer.from(value)
+      result[i] = new Uint8Array(value)
       continue
     }
 
--- a/packages/adapter-d1/src/conversion.ts
+++ b/packages/adapter-d1/src/conversion.ts
@@ -107,7 +107,7 @@ export function mapRow<A>(result: (A | ResultValue)[], columnTypes: ColumnType[]
     const value = result[i]
 
     if (value instanceof ArrayBuffer) {
-      result[i] = Array.from(new Uint8Array(value))
+      result[i] = Buffer.from(value)
       continue
     }
 
--- a/packages/adapter-libsql/src/conversion.ts
+++ b/packages/adapter-libsql/src/conversion.ts
@@ -139,7 +139,7 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
     // logic that treats string values of type Bytes as raw UTF-8 bytes that was
     // implemented for other adapters.
     if (value instanceof ArrayBuffer) {
-      result[i] = Array.from(new Uint8Array(value))
+      result[i] = Buffer.from(value)
       continue
     }
 
@@ -206,10 +206,6 @@ export function mapArg(arg: unknown, argType: ArgType, options?: PrismaLibSqlOpt
 
   if (typeof arg === 'string' && argType.scalarType === 'bytes') {
     return Buffer.from(arg, 'base64')
-  }
-
-  if (Array.isArray(arg) && argType.scalarType === 'bytes') {
-    return new Uint8Array(arg)
   }
 
   return arg as InValue
--- a/packages/adapter-libsql/src/conversion.ts
+++ b/packages/adapter-libsql/src/conversion.ts
@@ -134,6 +134,11 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
   for (let i = 0; i < row.length; i++) {
     const value = row[i]
 
+    if (value instanceof ArrayBuffer) {
+      result[i] = new Uint8Array(value)
+      continue
+    }
+
     // If an integer is required and the current number isn't one,
     // discard the fractional part.
     if (
--- a/packages/adapter-libsql/src/conversion.ts
+++ b/packages/adapter-libsql/src/conversion.ts
@@ -139,7 +139,7 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
     // logic that treats string values of type Bytes as raw UTF-8 bytes that was
     // implemented for other adapters.
     if (value instanceof ArrayBuffer) {
-      result[i] = Buffer.from(value)
+      result[i] = new Uint8Array(value)
       continue
     }
 
--- a/packages/adapter-libsql/src/conversion.ts
+++ b/packages/adapter-libsql/src/conversion.ts
@@ -134,11 +134,6 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
   for (let i = 0; i < row.length; i++) {
     const value = row[i]
 
-    if (value instanceof ArrayBuffer) {
-      result[i] = new Uint8Array(value)
-      continue
-    }
-
     // If an integer is required and the current number isn't one,
     // discard the fractional part.
     if (
--- a/packages/adapter-libsql/src/conversion.ts
+++ b/packages/adapter-libsql/src/conversion.ts
@@ -134,10 +134,6 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
   for (let i = 0; i < row.length; i++) {
     const value = row[i]
 
-    // Convert array buffers to arrays of bytes.
-    // Base64 would've been more efficient but would collide with the existing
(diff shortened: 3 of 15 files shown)


packages/client/tests/functional/bytes-upsert/_matrix.ts
```diff
@@ -0,0 +1,4 @@
+import { defineMatrix } from '../_utils/defineMatrix'
+import { allProviders, Providers } from '../_utils/providers'
+
+export default defineMatrix(() => [allProviders.filter(({ provider }) => provider !== Providers.SQLSERVER)])
```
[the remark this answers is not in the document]
jacek-prisma: Both are necessary in our tests

packages/client/tests/functional/bytes-upsert/tests.ts
```diff
@@ -0,0 +1,42 @@
+import { randomBytes } from 'crypto'
+
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './generated/prisma/client'
+
```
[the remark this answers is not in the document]
jacek-prisma: It is needed in our tests

packages/client/tests/functional/bytes-upsert/tests.ts
```diff
@@ -0,0 +1,42 @@
...
+      // This second call was failing in v7 with "No record was found for an upsert"
+      await upsertByteRow() // Should not throw
+
+      // Verify the record exists and has correct data
+      const result = await prisma.testByteId.findUnique({
+        where: { bytes: byteId },
+      })
+      expect(result).toBeTruthy()
+      expect(result?.bytes).toEqual(byteId)
+    })
```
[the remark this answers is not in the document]
jacek-prisma: doesn't matter in our tests

packages/adapter-neon/src/conversion.ts
```diff
@@ -370,18 +359,11 @@ const parsePgBytes = getTypeParser(ScalarColumnType.BYTEA) as (_: string) => Buf
 const parseBytesArray = getTypeParser(ArrayColumnType.BYTEA_ARRAY) as (_: string) => Buffer[]
 
 function normalizeByteaArray(serializedBytesArray) {
-  const buffers = parseBytesArray(serializedBytesArray)
-  return buffers.map((buf) => (buf ? encodeBuffer(buf) : null))
+  return parseBytesArray(serializedBytesArray)
 }
```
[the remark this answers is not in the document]
jacek-prisma: this line was not touched in the PR

packages/adapter-pg/src/conversion.ts
```diff
@@ -383,18 +372,11 @@ const parsePgBytes = getTypeParser(ScalarColumnType.BYTEA) as (_: string) => Buf
 const parseBytesArray = getTypeParser(ArrayColumnType.BYTEA_ARRAY) as (_: string) => Buffer[]
 
 function normalizeByteaArray(serializedBytesArray) {
-  const buffers = parseBytesArray(serializedBytesArray)
-  return buffers.map((buf) => (buf ? encodeBuffer(buf) : null))
+  return parseBytesArray(serializedBytesArray)
 }
```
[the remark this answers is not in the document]
jacek-prisma: this line was not touched in the PR

packages/adapter-d1/src/conversion.ts
```diff
@@ -107,7 +107,7 @@ export function mapRow<A>(result: (A | ResultValue)[], columnTypes: ColumnType[]
     const value = result[i]
 
     if (value instanceof ArrayBuffer) {
-      result[i] = Array.from(new Uint8Array(value))
+      result[i] = Buffer.from(value)
```
aqrln: I think it should be `new Uint8Array(value)`, we try to not use `Buffer` much anymore and we have even changed the public API in Prisma 6 to accept and return `Uint8Array`s. Especially here, since it's an adapter for Cloudflare D1, I'd expect a standard JS type and not a Node.js specific one.

packages/adapter-libsql/src/conversion.ts
```diff
@@ -139,7 +139,7 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
     // logic that treats string values of type Bytes as raw UTF-8 bytes that was
     // implemented for other adapters.
     if (value instanceof ArrayBuffer) {
```
aqrln: Outdated comment

packages/adapter-libsql/src/conversion.ts
```diff
@@ -139,7 +139,7 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
...
-      result[i] = Array.from(new Uint8Array(value))
+      result[i] = Buffer.from(value)
```
aqrln: likewise, let's not use `Uint8Array` and not `Buffer` unless absolutely necessary.

Even better, maybe we shouldn't even handle this case on this level, it should be the `DataMapper` that converts arbitrary `ArrayBuffer`s to `Uint8Array`s if necessary at all.

packages/adapter-neon/src/conversion.ts
```diff
@@ -370,18 +359,11 @@ const parsePgBytes = getTypeParser(ScalarColumnType.BYTEA) as (_: string) => Buf
...
-/**
- * Convert bytes to a JSON-encodable representation since we can't
- * currently send a parsed Buffer or ArrayBuffer across JS to Rust
- * boundary.
- */
-function convertBytes(serializedBytes: string): number[] {
-  const buffer = parsePgBytes(serializedBytes)
-  return encodeBuffer(buffer)
+function convertBytes(serializedBytes: string): Buffer {
+  return parsePgBytes(serializedBytes)
```
aqrln: Do we still need these wrapper functions? Should we maybe just use the underlying functions directly?

packages/adapter-libsql/src/conversion.ts
jacek-prisma: I think it's better to have one 'expected' type for byte arrays, we even enforce it a lot of the time via `ResultValue`, which right now does not permit `ArrayBuffer`

packages/adapter-d1/src/conversion.ts
jacek-prisma: updated all calls that used either an arraybuffer or a view to use `new Uint8Array`

packages/adapter-libsql/src/conversion.ts
jacek-prisma: 👍 
removed it

packages/adapter-neon/src/conversion.ts
jacek-prisma: I don't think so, removed them

packages/adapter-mssql/src/conversion.ts
```diff
@@ -125,12 +125,8 @@ export function mapArg<A>(arg: A | BigInt | Date, argType: ArgType): null | numb
...
   }
 
-  if (Array.isArray(arg) && argType.scalarType === 'bytes') {
-    return Buffer.from(arg)
-  }
-
   if (ArrayBuffer.isView(arg)) {
-    return Buffer.from(arg.buffer, arg.byteOffset, arg.byteLength)
+    return new Uint8Array(arg.buffer, arg.byteOffset, arg.byteLength)
   }
```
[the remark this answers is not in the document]
jacek-prisma: fixed

packages/adapter-neon/src/conversion.ts
```diff
@@ -444,13 +422,9 @@ export function mapArg<A>(arg: A | Date, argType: ArgType): null | unknown[] | s
...
 
-  if (Array.isArray(arg) && argType.scalarType === 'bytes') {
-    return Buffer.from(arg)
-  }
-
   // https://github.com/brianc/node-postgres/pull/2930
   if (ArrayBuffer.isView(arg)) {
-    return Buffer.from(arg.buffer, arg.byteOffset, arg.byteLength)
+    return new Uint8Array(arg.buffer, arg.byteOffset, arg.byteLength)
   }
```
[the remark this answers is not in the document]
jacek-prisma: I don't think this is accurate based on my search, and also the tests pass with `Uint8Array`

packages/adapter-better-sqlite3/src/conversion.ts
```diff
@@ -150,12 +150,8 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
...
     const value = row[i]
 
-    // Convert array buffers to arrays of bytes.
-    // Base64 would've been more efficient but would collide with the existing
-    // logic that treats string values of type Bytes as raw UTF-8 bytes that was
-    // implemented for other adapters.
-    if (value instanceof ArrayBuffer || value instanceof Buffer) {
-      result[i] = Array.from(new Uint8Array(value))
+    if (value instanceof ArrayBuffer) {
+      result[i] = new Uint8Array(value)
```
igalklebanov: `better-sqlite3` returns `Buffer` for `BLOB` affinity according to https://github.com/WiseLibs/better-sqlite3/blob/6209be238d6a1b181f516e4e636986604b0f62e1/docs/api.md?plain=1#L663.

`Buffer` is a subclass of `Uint8Array`.
`ArrayBuffer` is not an ancestor of `Uint8Array`.

This code seems to be unreachable.

packages/adapter-libsql/src/conversion.ts
```diff
@@ -134,12 +134,8 @@ export function mapRow(row: Row, columnTypes: ColumnType[]): ResultValue[] {
   for (let i = 0; i < row.length; i++) {
     const value = row[i]
 
-    // Convert array buffers to arrays of bytes.
-    // Base64 would've been more efficient but would collide with the existing
-    // logic that treats string values of type Bytes as raw UTF-8 bytes that was
-    // implemented for other adapters.
     if (value instanceof ArrayBuffer) {
```
igalklebanov, line 137: See https://github.com/prisma/prisma/pull/28913/changes#r2619173013, which might be relevant here as well.

packages/adapter-better-sqlite3/src/conversion.ts
jacek-prisma: the `ArrayBuffer` check was already there, maybe it's not needed, I'll remove it and see if tests still pass

packages/adapter-libsql/src/conversion.ts
jacek-prisma: `libsql` defines the type for column values as
```
export type Value = null | string | number | bigint | ArrayBuffer
```
which I think means that they can indeed return an ArrayBuffer (removing the check gives a type error)

igalklebanov: What you mentioned:

https://github.com/tursodatabase/libsql-client-ts/blob/45634442fb2e9c395daf49394839a64dd449aca9/packages/libsql-core/src/api.ts#L460-L478

You're right.

igalklebanov (approved):

PR #28958, opened 2025-12-23
perf(client): query plan caching PoC

aqrln:
Implements a proof-of-concept for query plan caching that uses a strawman non production ready parameterization algorithm.

This proves that query plan caching provides significant performance improvements for repeated queries with different parameters and measures the concrete impact.

Next steps:
- Finish implementation of query plan caching for batch queries.
- Mark parameterizable places in the query schema (Rust) and DMMF.
- Embed the parameter metadata in the generated client for schema-aware parameterization.
- We will also probably need new kinds of parameters for things like lists so we can parameterize more queries and make shapes more generics, and move chunking completely to the client side.

Ref: https://linear.app/prisma-company/issue/TML-1725/query-plan-caching

### DIFF
--- a/packages/client/src/__tests__/benchmarks/query-performance/caching.bench.ts
+++ b/packages/client/src/__tests__/benchmarks/query-performance/caching.bench.ts
@@ -1,0 +1,187 @@
+import fs from 'node:fs'
+import path from 'node:path'
+
+import { withCodSpeed } from '@codspeed/benchmark.js-plugin'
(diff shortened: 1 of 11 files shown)


egoldschmidt:
what's the ETA on landing this optimization?

aqrln:
@egoldschmidt it was planned for the next release but likely won't make it, so 7.4.0.

I'm closing this proof-of-concept PR because the actual work is happening on other branches.

Issue #8548, opened 2021-08-02
`prisma format` ends the file with a single CRLF on windows


binary64:
### Bug description

On windows, I run `prisma format` and note the unusual file ending. The lines are all LF, but the very last line is CRLF.

This causes issue on my Linux CI where it formats it ending in LF's only, causing a diff to occur and the build to fail.

### How to reproduce

1. On windows do prisma format
2. Open in HxD or similar
3. See attached:
 
![image](https://user-images.githubusercontent.com/1680627/127931864-fcd66391-9b31-4914-8553-f6c71b3a6fad.png)

### Expected behavior

LF's only - no CR's to exist at all

### Environment & setup

Node 16
Windows/Linux

### Prisma Version

```
$ C:\p\pab\monorepo\node_modules\.bin\prisma -v
prisma               : 2.26.0
@prisma/client       : 2.28.0
Current platform     : windows
Query Engine         : query-engine 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\query-engine-windows.exe)
Migration Engine     : migration-engine-cli 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\migration-engine-windows.exe)
Introspection Engine : introspection-core 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\introspection-engine-windows.exe)
Format Binary        : prisma-fmt 9b816b3aa13cc270074f172f30d6eda8a8ce867d (at node_modules\@prisma\engines\prisma-fmt-windows.exe)
Default Engines Hash : 9b816b3aa13cc270074f172f30d6eda8a8ce867d
Studio               : 0.408.0
```

Miladiir:
I just ran into the same problem. I check out files as-is, so without git's auto-clrf setting, even on windows.

DanielZavala11:
Hey @janpio, Guillermo (@GmoZen) and I are working on a fix for this bug as part of our CodeDay project (https://labs.codeday.org/).
I'm wondering if we can fix this by just changing the os.EOL line in this file to \n,  that way the file would consistently have the same line endings on every line, as opposed to having the platform-specific line ending on the last line?
https://github.com/prisma/prisma/blob/96dc3444b1439e3ab166a69c8290aec06c61bfba/packages/cli/src/Format.ts#L65
[...]

PR #28974, opened 2025-12-29
fix(cli): normalize line endings in prisma format output

eL1fe:
## Summary
Normalizes line endings to LF before writing formatted schema files to ensure consistent output across platforms.

## Problem
On Windows, `prisma format` produces files with mixed line endings: LF for most lines but CRLF for the last line. This causes:
- Inconsistent diffs between platforms in CI
- Files being reformatted on Linux after being formatted on Windows

## Solution
- Normalize line endings to LF before writing the formatted schema file
- Also normalize in `--check` mode to avoid false positives when comparing files with different line ending styles

## Changes
- `packages/cli/src/Format.ts`: Added line ending normalization using `.replace(/\r\n/g, '\n')`

Fixes #8548

## Release Notes

* **Bug Fixes**
  * Fixed cross-platform line ending inconsistencies to ensure consistent format checking and file output across Windows, macOS, and Linux environments.

### DIFF
--- a/packages/cli/src/Format.ts
+++ b/packages/cli/src/Format.ts
@@ -93,7 +93,10 @@ Or specify a Prisma schema path
           return new HelpError(`${bold(red(`!`))} The schema ${underline(filename)} is not found in the schema list.`)
         }
         const [, originalSchema] = originalSchemaTuple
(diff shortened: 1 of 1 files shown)


SevInf:

PR #29003, opened 2026-01-06
perf: port optimizations from perf branch

jacek-prisma:
[TML-1735](https://linear.app/prisma-company/issue/TML-1735/review-and-merge-query-interpreter-optimizations-from-perf-branch)
Applies a number of optimizations that seem sensible from the `perf` branch.

## Summary by CodeRabbit

* **Refactor**
  * Optimized object field mapping to reduce per-row memory allocations and improve serialization performance.
* **Refactor**
  * Improved tracing flow: added an enabled check to skip tracing work when disabled and consolidated query timing emission to avoid unnecessary overhead.

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/data-mapper.ts
+++ b/packages/client-engine-runtime/src/interpreter/data-mapper.ts
@@ -13,6 +13,18 @@ export class DataMapperError extends UserFacingError {
   }
 }
 
+// Optimization: Cache field entries to avoid repeated Object.entries() allocations per row
+const fieldEntriesCache = new WeakMap<Record<string, ResultNode>, [string, ResultNode][]>()
+
+function getFieldEntries(fields: Record<string, ResultNode>): [string, ResultNode][] {
+  let entries = fieldEntriesCache.get(fields)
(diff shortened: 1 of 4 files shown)


packages/client-engine-runtime/src/interpreter/data-mapper.ts
```diff
@@ -13,6 +13,18 @@ export class DataMapperError extends UserFacingError {
   }
 }
 
+// Optimization: Cache field entries to avoid repeated Object.entries() allocations per row
+const fieldEntriesCache = new WeakMap<Record<string, ResultNode>, [string, ResultNode][]>()
```
igalklebanov, line 17: we might need to double check this works as expected in Cloudflare Workers.

igalklebanov (approved):

packages/client-engine-runtime/src/interpreter/data-mapper.ts
jacek-prisma: cloudflare workers are covered by the ecosystem tests, we should see if it works there

PR #29004, opened 2026-01-06
perf: purified plans

jacek-prisma:
[TML-1735](https://linear.app/prisma-company/issue/TML-1735/review-and-merge-query-interpreter-optimizations-from-perf-branch)

A custom optimization based on something from the `perf` branch.


aqrln (approved):

tensordreams:
Rebased on top of latest `main` (resolving conflicts with #29262, #29251 and #29184 — `DeepReadonly` plan types, `canAssumeStrictEquality` on joins, parameterised in-memory ops) and pushed follow-up commits addressing the review comments:

- **`fix: stop query plan purification at transaction nodes`** — `transaction` fell through the default case of `findUniqueUnconditionalImpureNode` and was treated as if it contained no impure nodes. The switch is now exhaustive (`assertNever`), so any node kind added in the future has to be classified explicitly instead of silently falling into the same trap.
- **`fix: translate purified plan interpretation errors to user-facing errors`** — the pure path now goes through `rethrowAsUserFacing` like every other interpretation path.
- **`fix: avoid mutating shared query plans during purification`** — this one goes beyond the review comments: query plans are cached in `QueryPlanCache` and shared between requests, and `purifyQueryPlan` spliced the query result into the plan in place via `Object.assign`. The first execution poisoned the cached plan with its own result; subsequent runs of the same plan then found no impure node and returned the first run's data without ever hitting the database. Same class of bug as #29254 / #29262, which landed after this PR was opened. Purification now substitutes the evaluated node by copying only the nodes on the path to it, sharing all unaffected subtrees with the input plan. Added unit tests for both the interpreter-level behaviour (the same plan run twice hits the DB twice and stays deep-equal to its original shape) and the purifier itself.
- **`docs: document query plan purity types`** — doc comments for `PureQueryPlanNode` / `ImpureQueryPlanNode` / `QueryPlanNode` / `purifyQueryPlan` and the purification-only `lastInsertId` field.

A note on the benchmark numbers: the bench suites reuse the same plan constants across iterations, so under the in-place mutation the `findUnique` benchmark only hit the mock adapter on its first iteration and measured the poisoned-plan shortcut afterwards — the previously reported +44%/+53% gains were partly an artifact of the bug. Expect them to come down now that every iteration does honest work. Local wall-clock runs of `bench/interpreter.bench.ts` show no regression on `join (1:N)` (+0.8% vs `main`, well within noise), so the −29.57% CodSpeed report looks like a simulation artifact, but worth re-checking on the fresh CI run.

PR #29014, opened 2026-01-09
chore: sanitize QPE connection errors

jacek-prisma:
[TML-1815](https://linear.app/prisma-company/issue/TML-1815/security-sanitize-logged-connection-strings-from-qpe)

TLDR: we do not want anything resembling a connection string surfaced in the connection errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error messages now automatically redact sensitive connection information across various error scenarios and formats.
  * Improved error handling for adapter operations (transactions, queries, and raw commands) to prevent exposure of connection details in error output.

* **Refactor**
  * Adapter factory now supports configurable custom factories for enhanced flexibility.


packages/query-plan-executor/src/logic/app.ts
```diff
@@ -40,7 +41,7 @@ export class App {
   static async start(options: Options): Promise<App> {
     const connector = createAdapter(options.databaseUrl)
 
-    const db = await runInActiveSpan('connect', () => connector.connect())
+    const db = await runInActiveSpan('connect', () => connector.connect().catch(rethrowSanitizedError))
```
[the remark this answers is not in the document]
jacek-prisma: this comment is not related to the change here

packages/query-plan-executor/src/utils/error.test.ts
```diff
@@ -40,3 +40,76 @@ describe('extractErrorFromUnknown', () => {
...
+      expect(caughtError).toBeInstanceOf(Error)
+      expect((caughtError as Error).name).toBe('DatabaseError')
+      expect((caughtError as Error).stack).toBe(originalStack)
+      expect((caughtError as Error).message).toBe('[REDACTED] connection failed')
+    }
+  })
+
+  it('rethrows non-Error values unchanged', () => {
+    expect(() => rethrowSanitizedError('string error')).toThrow('string error')
+  })
```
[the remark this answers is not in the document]
jacek-prisma: not very valuable

packages/query-plan-executor/src/utils/error.ts
```diff
@@ -8,3 +8,20 @@ export function extractErrorFromUnknown(value: unknown): Error | string {
...
+
+/**
+ * Rethrows the given error after sanitizing its message by redacting
+ * any connection strings found within it.
+ */
+export function rethrowSanitizedError(error: unknown): never {
+  if (error instanceof Error) {
+    error.message = error.message.replaceAll(CONNECTION_STRING_REGEX, '[REDACTED]')
+  }
+  throw error
```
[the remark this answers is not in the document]
jacek-prisma: not very valuable

packages/query-plan-executor/src/utils/error.ts
```diff
@@ -8,3 +8,20 @@ export function extractErrorFromUnknown(value: unknown): Error | string {
...
+}
```
igalklebanov: this can easily not catch enough, and is a maintenance burden.

is there any reason not to redact all URL-like parts?

igalklebanov (dismissed):

packages/query-plan-executor/src/utils/error.ts
```diff
@@ -8,3 +8,20 @@ export function extractErrorFromUnknown(value: unknown): Error | string {
   }
   return String(value)
 }
+
+// A lenient regex to match connection strings in error messages.
+// It might match some false positives, but that's acceptable for errors in the QPE.
+// We do not want anything that looks like a connection string in the logs.
+const CONNECTION_STRING_REGEX =
+  /['"`]?(mysql|mariadb|postgresql|jdbc:sqlserver|sqlserver|sqlite|mongodb):\/\/[^\s]+['"`]?/gi
```
igalklebanov: `postgres` and `postgresql` are both valid for PostgreSQL.

igalklebanov: Is `prisma+postgres` still relevant?

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: This regex should always match whole URLs since URLs can't contain whitespace
Do you have a suggestion how it could be improved?

packages/query-plan-executor/src/utils/error.ts
igalklebanov: `mssql` is also valid for MS SQL Server.

jacek-prisma: I am not sure if it's possible for us to have a `prisma+postgres` URL in the QPE

jacek-prisma: yeah, it'd not even start, see `packages/query-plan-executor/src/logic/adapter.ts`

jacek-prisma: we don't allow `mssql://` in `packages/query-plan-executor/src/logic/adapter.ts`, but I guess it could be added for future proofing

packages/query-plan-executor/src/utils/error.ts
```diff
@@ -8,3 +8,20 @@ export function extractErrorFromUnknown(value: unknown): Error | string {
...
+
+/**
+ * Rethrows the given error after sanitizing its message by redacting
+ * any connection strings found within it.
+ */
+export function rethrowSanitizedError(error: unknown): never {
+  if (error instanceof Error) {
+    error.message = error.message.replaceAll(CONNECTION_STRING_REGEX, '[REDACTED]')
```
igalklebanov: do we need to also take care of `error.cause`?

jacek-prisma: I was thinking about that too, I think it'd make sense

igalklebanov: and `AggregateError`'s `errors`?

packages/query-plan-executor/src/utils/error.ts
igalklebanov: Not talking about the implementation, more about the strategy. 

Is this thing too strict? is there any reason not to redact anything that looks like a URL and call it a day? No need to maintain protocol names, no risk.

packages/query-plan-executor/src/utils/error.ts
igalklebanov: I think it's worth adding support for `mssql://` in another PR. Sounds like a small thing to do that might make someone's DX smoother.

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: I don't think this is strict though, it just expects one of our expected protocols and any sequence of non whitespace

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: added handling for both

jacek-prisma: wondering if we should recursively check if the error has a cause or is an AggregateError, I didn't implement it recursively though, because it could lead to infinite recursion if the errors end up having a cycle which is something I wouldn't be very surprised by in javascript

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: I've added it

packages/query-plan-executor/src/utils/error.ts
```diff
@@ -8,3 +8,36 @@ export function extractErrorFromUnknown(value: unknown): Error | string {
...
+
+  throw error
+}
+
+function sanitizeError(error: unknown): unknown {
+  if (error instanceof Error) {
+    error.message = error.message.replaceAll(CONNECTION_STRING_REGEX, '[REDACTED]')
+  }
+  return error
+}
```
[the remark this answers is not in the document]
jacek-prisma: We expect the underlying drivers to actually throw errors

packages/query-plan-executor/src/utils/error.ts
```diff
@@ -8,3 +8,36 @@ export function extractErrorFromUnknown(value: unknown): Error | string {
...
+  }
+
+  if (error instanceof AggregateError) {
+    for (const innerError of error.errors) {
+      sanitizeError(innerError)
+    }
+  }
+
+  throw error
+}
```
igalklebanov: LLM potentially drunk:

`pg` might attach the `config` object to errors, which might contain the connection string.

packages/query-plan-executor/src/utils/error.ts
jkomyno: Also, we have `prisma+postgres` and `prisma+postgresql`, afaik.

I don't think listing out every known protocol upfront is the way forward here, tbh.

jacek-prisma: we can expose a complete set of all allowed protocols from `adapter.ts`, by just exporting something like:
```ts
export const allAllowedProtocols: Set<string> = new Set(factories.flatMap((factory) => factory.protocols))
```
and use it in the sanitization

packages/query-plan-executor/src/utils/error.ts
igalklebanov: What about the other question?

packages/query-plan-executor/src/utils/error.ts
igalklebanov: I guess you could use a `WeakMap` or similar to catch "been here before".

packages/query-plan-executor/src/utils/error.ts
igalklebanov: I suspect having a module that knows all `factories` might not be the best idea.

igalklebanov: Can we do something at build time?

packages/query-plan-executor/src/utils/error.test.ts
```diff
@@ -40,3 +40,114 @@ describe('extractErrorFromUnknown', () => {
...
+
+    try {
+      rethrowSanitizedError(error)
+    } catch (caughtError) {
+      expect(caughtError).toBeInstanceOf(Error)
+      expect((caughtError as Error).message).toBe('Connection error occurred')
+      expect((caughtError as Error).cause).toBeInstanceOf(Error)
+      expect(((caughtError as Error).cause as Error).message).toBe('Failed to connect to [REDACTED]')
+    }
+  })
```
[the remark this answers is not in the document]
jacek-prisma: done

packages/query-plan-executor/src/utils/error.ts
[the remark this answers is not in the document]
jacek-prisma: done

jacek-prisma: I rewrote it to just sanitize all properties recursively while avoiding cycles

packages/query-plan-executor/src/utils/error.test.ts
```diff
@@ -40,3 +40,114 @@ describe('extractErrorFromUnknown', () => {
...
+      rethrowSanitizedError(aggregateError)
+    } catch (caughtError) {
+      expect(caughtError).toBeInstanceOf(AggregateError)
+      const caught = caughtError as AggregateError
+      expect(caught.message).toBe('Multiple connection errors')
+      expect(caught.errors).toHaveLength(2)
+      expect(caught.errors[0].message).toBe('Connection failed: [REDACTED]')
+      expect(caught.errors[1].message).toBe('Another failure: [REDACTED]')
+    }
+  })
```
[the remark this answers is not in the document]
jacek-prisma: done

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: I rewrote it to just sanitize all properties recursively while avoiding cycles

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: updated the code to dynamically generate the pattern for all supported protocols, so we don't have to maintain a list

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: updated the code to dynamically generate the pattern for all supported protocols, so we don't have to maintain a list

packages/query-plan-executor/src/utils/error.test.ts
```diff
@@ -40,3 +41,182 @@ describe('extractErrorFromUnknown', () => {
...
+  it.each([
+    ['basic mysql connection string', 'mysql://user:password@localhost:3306/database', '[REDACTED]'],
+    ['mariadb connection string', 'mariadb://user:password@localhost:3306/database', '[REDACTED]'],
+    ['postgres connection string', 'postgres://user:password@localhost:5432/database', '[REDACTED]'],
+    ['postgresql connection string', 'postgresql://user:password@localhost:5432/database', '[REDACTED]'],
+    [
+      'jdbc:sqlserver connection string',
+      'jdbc:sqlserver://localhost:1433;database=test;user=sa;password=pass',
+      '[REDACTED]',
+    ],
```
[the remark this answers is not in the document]
jacek-prisma: deleted the case

packages/query-plan-executor/src/utils/error.ts
```diff
@@ -8,3 +8,47 @@ export function extractErrorFromUnknown(value: unknown): Error | string {
   }
   return String(value)
 }
+
+/**
+ * Rethrows the given error after sanitizing its message by redacting
+ * any connection strings found within it.
+ */
+export function rethrowSanitizedError(error: unknown, protocols: string[]): never {
```
igalklebanov: nitpicky

consider

```suggestion
export function rethrowSanitizedError(error: unknown, protocols = allAllowedProtocols): never {
```

This simplify the callsite to just `.catch(rethrowSanitizedError)`, and won't force it to import this.

Then the test suite won't have to import it either.

or..

even not using injection at all, and hard-coding the usage of `allAllowedProtocols` in this function - given nothing currently passes anything other than `allAllowedProtocols`.

jacek-prisma: I didn't like the idea of `utils` importing from `logic`, we'd probably end up with cyclic imports, that's why I implemented it as a plain argument

igalklebanov: Pushing it as a "util" based on a single call site is premature.

Should be moved into "app" module for co-location. Tests should be around the "app" throwing an error that includes a connection string, not this very small and insignificant unit.

igalklebanov: > Should be moved into "app" module for co-location. Tests should be around the "app" throwing an error that includes a connection string, not this very small and insignificant unit.

The test setup should make `connector.connect()` fail with all sorts of connection strings.

igalklebanov: Also, "app" knowing all protocols is an anti-pattern.

The `connect()` function should have enough information to perform the replacements using the protocols relevant to it.

igalklebanov: A PostgreSQL connector should replace PostgreSQL connection strings.

jacek-prisma: > The test setup should make connector.connect() fail with all sorts of connection strings.

I intended to do that but currently there's no test setup for the `app.ts` and it'd require some extra work to make it testable
In order to avoid large changes in this PR I didn't do that since it's just a security fix

jacek-prisma: > Also, "app" knowing all protocols is an anti-pattern.

> A PostgreSQL connector should replace PostgreSQL connection strings.


Those are issues with the pre-existing code, I think refactoring existing code is out of scope of this PR

igalklebanov: "app" knowing all protocols is a new thing introduced in this PR.

Why is sanitization at the source out of scope of a PR titled "sanitize QPE connection errors" when connection errors happen at the connector layer?
It's not out of scope, we're just knowingly going for a hack downstream.

jacek-prisma: > "app" knowing all protocols is a new thing introduced in this PR.

The protocols have been hardcoded there previously, so "app" already knew all the protocols, I haven't added that

I just exposed it out of the app module for use in tests

igalklebanov: > The protocols have been hardcoded there previously, so "app" already knew all the protocols, I haven't added that

How is that possible? `packages/query-plan-executor/src/logic/app.ts` has a new import for protocols. The existing file has nothing related. 😕

jacek-prisma: I think there are ways to make this better from the PoV of tests and encapsulation, but it'd require either
a) making `app.ts` more testable, which I wanted to avoid to contain the scope of the changes, since it currently has no tests or test setup at all
b) avoid dealing with `app.ts`, we could contain these changes purely in the `adapter.ts` by introducing a wrapper for the adapters that automatically redacts all errors, this is more complex than what I am doing but can be more effectively tested

igalklebanov: This is good to go for now.

But we need to ack the hack (explicitly) and track the healthier solution for later. The abstraction either needs enhancement and/or take responsibility of sanitizing per-connector based on what each connector allows or expects.

igalklebanov (dismissed):

packages/query-plan-executor/src/utils/error.ts
jacek-prisma: I think I'll give the b) approach a go, since it's actually better than the current approach in a lot of ways, we can encapsulate error handling within the adapter module and wrap all the methods we care about there, also I think only handling `connect` errors is arguably insufficient, since many of the adapters can produce database connection errors after the `connect` call is made since they rely on connection pools

igalklebanov: Connection pools normally swallows these errors until a certain time/retry threshold is exceeded. Makes it harder to test probably.

If `pg` attaches the config object to errors, you could get connection strings in bad query errors too.

We should focus on the types of errors that were reported.

jacek-prisma: @igalklebanov I've refactored it now and it's all encapsulated in `adapter.ts`

packages/query-plan-executor/src/logic/adapter.ts
```diff
@@ -65,3 +65,78 @@ const factories: Factory[] = [
     },
   },
 ]
+
+const allSupportedProtocols = factories.flatMap((factory) => factory.protocols)
```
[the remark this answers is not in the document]
jacek-prisma: updated

packages/query-plan-executor/src/logic/adapter.ts
```diff
@@ -65,3 +65,78 @@ const factories: Factory[] = [
...
+  const pattern = [
+    `['"\`]?`, // Optional opening quote
+    `(${escapedProtocols})`, // Protocol group
+    `:\\/\\/`, // Protocol separator
+    `[^\\s]+`, // Connection string body
+    `['"\`]?`, // Optional closing quote
+  ].join('')
+
+  return new RegExp(pattern, 'gi')
+}
```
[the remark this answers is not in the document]
jacek-prisma: updated

packages/query-plan-executor/src/logic/adapter.ts
```diff
@@ -65,3 +65,78 @@ const factories: Factory[] = [
...
+
+  for (const key of Object.getOwnPropertyNames(error)) {
+    const value = error[key]
+    if (typeof value === 'string') {
+      error[key] = value.replaceAll(regex, '[REDACTED]')
+    } else if (typeof value === 'object' && value !== null) {
+      sanitizeError(value as object, regex, visited)
+    }
+  }
+}
```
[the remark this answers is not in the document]
jacek-prisma: updated

packages/query-plan-executor/src/logic/adapter.ts
```diff
@@ -65,3 +65,78 @@ const factories: Factory[] = [
...
+  const pattern = [
+    `['"\`]?`, // Optional opening quote
+    `(${escapedProtocols})`, // Protocol group
+    `:\\/\\/`, // Protocol separator
+    `[^\\s]+`, // Connection string body
+    `['"\`]?`, // Optional closing quote
+  ].join('')
+
+  return new RegExp(pattern, 'gi')
+}
```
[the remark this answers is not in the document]
jacek-prisma: the code has changed due to other comments

packages/query-plan-executor/src/logic/adapter.ts
```diff
@@ -65,3 +65,78 @@ const factories: Factory[] = [
...
+
+function wrapTransaction(tx: Transaction): Transaction {
+  return {
+    ...tx,
+    commit: () => tx.commit().catch(rethrowSanitizedError),
+    rollback: () => tx.rollback().catch(rethrowSanitizedError),
+    executeRaw: (query) => tx.executeRaw(query).catch(rethrowSanitizedError),
+    queryRaw: (query) => tx.queryRaw(query).catch(rethrowSanitizedError),
+  }
+}
```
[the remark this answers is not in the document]
jacek-prisma: updated

igalklebanov (approved):

PR #29016, opened 2026-01-09
chore: make the interpreter reusable

jacek-prisma:
Came out of the `perf` branch work, the `forSql` calls are often visible on profiles generated in benchmarks. We can avoid them by separating construction-time and run-time options and reusing the interpreter instance.

## Summary by CodeRabbit

* **Refactor**
  * Restructured internal runtime context flow for query interpreter to improve configuration management and reduce parameter passing overhead across interpretation operations.


aqrln (approved):
I like this, seems like a code improvement and a better separation of concerns regardless of the performance aspect.

PR #29088, opened 2026-01-24
fix(adapter-mariadb): handle object-based rows returned from driver

seoyeonjin:
### Problem
The MariaDB adapter assumed rows returned from the driver were always arrays.
In some cases (e.g. stored procedures), the driver may return object-based rows,
leading to a runtime `row.map is not a function` error.

### Solution
Normalize the row input by converting object-based rows to arrays before mapping,
while preserving existing behavior for array-based rows.

### Verification
The issue was reproduced with a local MariaDB instance using
`$queryRawUnsafe('CALL …')`. The fix was verified by patching the adapter runtime.

스크린샷 2026-01-25 오후 9 25 43

Fixes #29068

## Summary by CodeRabbit

* **New Features**
  * MariaDB adapter now accepts row data provided as either objects or arrays, improving input flexibility while remaining backward compatible.

* **Tests**
  * Added tests validating date/time formatting, large integer handling, and preservation of field order when mapping mixed rows.

### DIFF
--- a/packages/adapter-mariadb/src/conversion.test.ts
+++ b/packages/adapter-mariadb/src/conversion.test.ts
@@ -1,0 +1,38 @@
+import * as mariadb from 'mariadb'
+import { describe, expect, it } from 'vitest'
+
+import { mapRow } from './conversion'
+
+describe('mapRow', () => {
+  it('maps datetime columns from object rows to ISO strings', () => {
+    const row = { createdAt: '2024-01-02 03:04:05.000' }
+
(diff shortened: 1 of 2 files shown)


jacek-prisma:
Hi @seoyeonjin 
I think it'd be valuable to have a test in `packages/client/tests/functional` that invokes a stored procedure with the mariadb adapter to verify that stored procedure calls now work as expected.

jacek-prisma (approved):
The tests here are incomplete, but we can improve that in a follow-up change

PR #29119, opened 2026-02-02
fix(adapter-mssql): acquire mutex in commit/rollback to prevent EREQINPROG error

### LINKED ISSUE #28994: @prisma/adapter-mssql (Prisma 7) is unusable
### Bug description

Hi,

since Prisma 7 and the switch to `@prisma/adapter-mssql` I get `TransactionError: Can't rollback transaction. There is a request in progress` on almost every query, making it unusable in the current state. I had to rollback to version `6.19.1` without the Rust-free adapter.

Might be related:
#28812 #27990

### Severity

expected: upgrade to Prisma 7 should not break production code
actual: currently unusable

### Frequency

Consistently reproducible

### Workaround

Downgrading to Prisma 6.19.1 fixes it

### Prisma Schema & Queries

```prisma
```
```ts
```

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** Windows
- **Database:** MSSQL
- **Node.js version:** v22

### Prisma Version

```
6.19.1
```

MarkusLund:
## Description

Fix race condition in `MssqlTransaction` where `commit()` or `rollback()` can be called while a query is still in progress, causing the mssql driver to throw `EREQINPROG` error.

## Problem

The mssql driver throws `EREQINPROG` ("Can't rollback transaction. There is a request in progress.") when `commit()` or `rollback()` is invoked while a query is executing. This commonly occurs when a transaction times out during an in-flight query.

In the current implementation:
- `performIO()` acquires a mutex before executing queries
- `commit()` and `rollback()` do **not** acquire the mutex

This allows `commit()`/`rollback()` to race with active queries.

## Solution

Modified `commit()` and `rollback()` to acquire the same mutex that `performIO()` uses. This ensures they wait for any active query to complete before executing.

## Changes

- `packages/adapter-mssql/src/mssql.ts`: Added mutex acquisition to `commit()` and `rollback()` methods
- `packages/adapter-mssql/src/mssql.test.ts`: Added unit tests validating the mutex serialization behavior

Fixes #28994

## Summary by CodeRabbit

* **Bug Fixes**
  * Commit and rollback now reliably wait for in‑flight queries using serialization, preventing race conditions and reducing unexpected transaction failures; improved handling of aborted (EABORT) and other error cases.

* **Tests**
  * Added comprehensive tests validating transaction serialization, concurrent query behavior, correct ordering of commit/rollback relative to running queries, and proper handling/rethrowing of abort and non-abort errors.

### DIFF
--- a/packages/adapter-mssql/src/mssql.test.ts
+++ b/packages/adapter-mssql/src/mssql.test.ts
@@ -1,0 +1,183 @@
+import { Mutex } from 'async-mutex'
+import { beforeEach, describe, expect, Mock, test, vi } from 'vitest'
+
+/**
+ * Tests for MssqlTransaction race condition fix.
+ *
+ * Fixes https://github.com/prisma/prisma/issues/28994
+ *
+ * The mssql driver throws EREQINPROG when commit() or rollback() is called
+ * while a query is in progress. The fix ensures commit() and rollback()
+ * acquire the same mutex that performIO() uses.
+ */
+
+describe('MssqlTransaction mutex serialization', () => {
+  let events: string[]
+  let isQueryRunning: boolean
+  let mockTransaction: {
+    request: () => { arrayRowMode: boolean; input: Mock; query: Mock }
+    commit: Mock
+    rollback: Mock
+  }
+
+  beforeEach(() => {
+    events = []
+    isQueryRunning = false
+
+    mockTransaction = {
+      request: () => ({
+        arrayRowMode: true,
+        input: vi.fn(),
+        query: vi.fn(async () => {
+          isQueryRunning = true
+          events.push('query:start')
+          await new Promise((resolve) => setTimeout(resolve, 50))
+          events.push('query:end')
+          isQueryRunning = false
+          return { recordset: [[]], columns: [[{ name: 'id' }]] }
+        }),
+      }),
+      commit: vi.fn().mockImplementation(() => {
+        events.push(`commit:called (queryRunning=${isQueryRunning})`)
+        if (isQueryRunning) {
+          const error = new Error("Can't commit transaction. There is a request in progress.")
+          ;(error as any).code = 'EREQINPROG'
+          return Promise.reject(error)
+        }
+        events.push('commit:success')
+        return Promise.resolve()
+      }),
+      rollback: vi.fn().mockImplementation(() => {
+        events.push(`rollback:called (queryRunning=${isQueryRunning})`)
+        if (isQueryRunning) {
+          const error = new Error("Can't rollback transaction. There is a request in progress.")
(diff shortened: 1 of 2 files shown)


jacek-prisma:
Hi @MarkusLund, thanks for the pull request. Before we merge it though, could you sign the CLA?

packages/adapter-mssql/src/mssql.test.ts
```diff
@@ -0,0 +1,209 @@
...
+
+      const queryPromise = tx.queryRaw({ sql: 'SELECT 1' })
+      await new Promise((resolve) => setTimeout(resolve, 10))
+
+      await expect(tx.commit()).rejects.toThrow("Can't commit transaction. There is a request in progress.")
+      await queryPromise
+
+      expect(events).toContain('commit:called (queryRunning=true)')
+    })
+  })
```
jacek-prisma: I think we don't need these two cases - while they're useful to demonstrate the problem, I think all we need here are cases that catch possible future regressions preventing us from breaking it again and the test cases below cover that.

jacek-prisma:
The PR looks good, I think it can be merged when the CLA is signed and the one minor comment I left is addressed.

packages/adapter-mssql/src/mssql.test.ts
MarkusLund: Removed, and I rebased and squashed the commits.

MarkusLund:
CLA signed, @jacek-prisma 👍

packages/adapter-mssql/src/mssql.test.ts
```diff
@@ -0,0 +1,183 @@
...
+        try {
+          const req = mockTransaction.request()
+          return await req.query(query.sql)
+        } finally {
+          release()
+        }
+      },
+
+      async commit() {
+        if (useMutexForCommitRollback) {
```
jacek-prisma, line 78: `useMutexForCommitRollback` is now always true, so we can remove the parameter and the unreachable branches

packages/adapter-mssql/src/mssql.test.ts
```diff
@@ -0,0 +1,183 @@
...
+          await mockTransaction.rollback().catch((e: Error & { code?: string }) => {
+            if (e.code === 'EABORT') return
+            throw e
+          })
+        }
+      },
+    }
+  }
+
+  describe('with mutex (the fix)', () => {
```
jacek-prisma, line 111: probably don't need the nested `describe` here

jacek-prisma:
Included in https://github.com/prisma/prisma/pull/29141

PR #29177, opened 2026-02-12
fix(client-engine-runtime): use cross-realm safe Uint8Array check

### LINKED ISSUE #29129: `@prisma/adapter-pg` `Bytes` field fails in jsdom environment due to cross-realm `instanceof Uint8Array` check
### Bug description

When running tests with Vitest in `jsdom` environment, reading or writing `Bytes` fields fails with:
```
Expected a byte array in column 'value', got object: <content>
```

This happens because jsdom creates its own JavaScript realm with separate built-in classes. A `Uint8Array` created in jsdom's realm fails `instanceof Uint8Array` checks in Node's realm where `@prisma/adapter-pg` runs.

### Severity

### Reproduction

1. Create a Vitest config with `environment: 'jsdom'`
2. Use `@prisma/adapter-pg` with a model containing a `Bytes` field
3. Create/read a record with `Buffer.from('test')` or `new TextEncoder().encode('test')`
4. Observe the error

- **Expected:** Bytes fields work the same as in Node environment
- **Actual:** Fails with "Expected a byte array in column 'value', got object"

### Frequency

Consistently reproducible

### Is this a regression?

Likely introduced with Prisma 7

### Workaround

Add `// @vitest-environment node` directive to test files, or configure Vitest to use `node` environment for server tests

### Prisma Schema & Queries

```prisma
model ProjectContext {
  id        String   @id @default(cuid())
  value     Bytes
  // ...
}
```

```ts
await prisma.projectContext.create({
  data: { value: Buffer.from('test content') }
})
```

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

OS: macOS
Database: PostgreSQL
Node.js version: 24.x
Test runner: Vitest 4.x with jsdom

### Prisma Version

```
7.3.0
```

omair445:
Fixes #29129

## Problem

`value instanceof Uint8Array` fails in jsdom (and other cross-realm environments like iframes or vm contexts) because the `Uint8Array` constructor in the realm where the value was created is a different object than the one in the current realm.

This causes `@prisma/adapter-pg` to break when used with Bytes fields in jsdom-based test environments (vitest with jsdom, jest with jsdom, etc).

## Fix

Added a cross-realm safe `isUint8Array()` helper in `utils.ts` that uses `Object.prototype.toString.call(value)` instead of `instanceof`:

```ts
export function isUint8Array(value: unknown): value is Uint8Array {
  return ArrayBuffer.isView(value) && Object.prototype.toString.call(value) === "[object Uint8Array]"
}
```

Replaced all `instanceof Uint8Array` checks in:
- `packages/client-engine-runtime/src/interpreter/data-mapper.ts` (2 occurrences)
- `packages/client-engine-runtime/src/utils.ts` (2 occurrences)


omair445:
yeah good call — added an `isDate` helper using `Object.prototype.toString.call()` for consistency with the Uint8Array fix. makes sense to handle both the same way since they have the exact same cross-realm issue. pushed in 79d766a.

omair445:
updated — now using `isDate()` inside `asDate()` and also in `data-mapper.ts` where there was another `instanceof Date`. rebased on latest main too, should be clean now.

omair445:
Hey everyone, I am very very sorry I got stuck in work-related stuff and haven't been able to follow up on this. I am happy if anyone else can take care of this from here or if the maintainers want to push the final bits. Apologies again for the delay!

tensordreams:
Rebased on latest `main` (clean merge; both target files were touched upstream and the cross-realm fix reconciled without conflict). Verified no `instanceof Uint8Array` / `instanceof Date` remains on the hot path in `packages/client-engine-runtime/src`; `asDate()` now routes through `isDate()`, addressing CodeRabbit's critical note. Added `utils.test.ts` covering same-realm, Buffer subclass, and cross-realm (`node:vm`) values for `isUint8Array`/`isDate` plus `doKeysMatch`, asserting `instanceof` fails where the helpers succeed. Full suite green: `pnpm --filter @prisma/client-engine-runtime test` -> 20 files, 226 tests passing; prettier + eslint clean on changed files. Head: 9bae7dcd9adfd5e5a67576d98d6be023fd19d2fc.

aqrln (approved):

aqrln:
Thank you!

PR #29218, opened 2026-02-18
fix(client): apply result extensions for nested and fluent relations

LucianBuzzo:
Result extensions were always applied using the root request model/args in getPrismaClient._request. That works for regular top-level results, but fluent relation calls use dataPath unpacking first, so the payload handed to applyAllResultExtensions is already the nested relation object/list. In that case, walking extensions with the root model context skips nested computed fields (for example post extensions on user.findUnique(...).posts()).

This change resolves the effective extension context from dataPath before applying result extensions. The new resolveResultExtensionContext helper strips select/include tokens, follows relation fields through runtimeDataModel, and derives the matching nested args branch. If resolution is invalid it safely falls back to the root context.

I also added regression coverage for nested reads/writes and fluent relation calls in functional extensions tests, plus focused unit tests for context resolution (including fluent-style unwrapped payload behavior).

## Summary by CodeRabbit

* **New Features**
  * Result extensions can compute and expose additional nested fields (e.g., derived labels) on related models.

* **Improvements**
  * More reliable resolution of nested model context for result extensions across fluent select/include paths, including when relation fields are named "select" or "include".
  * Extension application now uses resolved nested context for more accurate behavior in nested queries.

* **Bug Fixes**
  * Prevents dropped nested selections when selector segments appear in relation paths.

* **Tests**
  * Added comprehensive unit and functional tests covering nested resolution and computed-field propagation.

### DIFF
--- a/packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
+++ b/packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
@@ -29,8 +29,8 @@ export function resolveResultExtensionContext({
     args: args ?? {},
   }
 
-  const relationPath = dataPath.filter((segment) => segment !== 'select' && segment !== 'include')
-  if (relationPath.length === 0) {
+  const relationPath = relationPathFromDataPath(dataPath)
+  if (!relationPath || relationPath.length === 0) {
     return rootContext
   }
 
@@ -56,6 +56,23 @@ export function resolveResultExtensionContext({
     modelName: currentModelName,
     args: currentArgs,
   }
+}
+
+function relationPathFromDataPath(dataPath: string[]): string[] | undefined {
+  const relationPath: string[] = []
+
+  for (let index = 0; index < dataPath.length; index += 2) {
+    const selector = dataPath[index]
+    const relationFieldName = dataPath[index + 1]
+
+    if ((selector !== 'select' && selector !== 'include') || relationFieldName === undefined) {
+      return undefined
+    }
+
+    relationPath.push(relationFieldName)
+  }
+
+  return relationPath
 }
 
 function resolveNextArgs(args: JsArgs, relationFieldName: string): JsArgs {
--- a/packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
+++ b/packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
@@ -16,7 +16,9 @@ type ResultExtensionContext = {
 
 /**
  * Resolves the model/args context used for result extensions from a query dataPath.
- * Falls back to the root model context when the relation path can not be traversed.
+ * Falls back to the root model context when traversal can not continue because the model is
+ * missing, the segment is not a relation, or the path format is invalid.
+ * Throws when a relation field segment from dataPath does not exist on the current model.
  */
 export function resolveResultExtensionContext({
   dataPath,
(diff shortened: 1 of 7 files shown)


packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
```diff
@@ -0,0 +1,81 @@
...
+  let currentArgs: JsArgs = args ?? {}
+
+  for (const relationFieldName of relationPath) {
+    const currentModel = runtimeDataModel.models[currentModelName]
+    if (!currentModel) {
+      return rootContext
+    }
+
+    const relationField = currentModel.fields.find((field) => field.name === relationFieldName)
+    if (!relationField || relationField.kind !== 'object' || !relationField.relationName) {
```
jacek-prisma: Shouldn't `!relationField` be impossible? If it's not ever expected I think it should be a runtime error

packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
```diff
@@ -0,0 +1,81 @@
...
+  modelName,
+  args,
+  runtimeDataModel,
+}: ResolveResultExtensionContextParams): ResultExtensionContext {
+  const rootContext = {
+    modelName,
+    args: args ?? {},
+  }
+
+  const relationPath = dataPath.filter((segment) => segment !== 'select' && segment !== 'include')
```
jacek-prisma: Can't `select` and `include` be valid names of relation fields? If they can this would incorrectly strip them

LucianBuzzo: Yeah you're right, I was thinking of single field select for some reason

LucianBuzzo: I'll update this and add tests to cover the case.

packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
LucianBuzzo: There's potential that a nosy dev (like myself) could mangle `dataPath` through a middleware. I can change this to a runtime error if you prefer.

jacek-prisma: If not a runtime error,  I think it could be worth to at least call `debug` (via `Debug` from `@prisma/driver-adapter-utils`)  here with some message to make it easier to spot this while debugging

packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
```diff
@@ -0,0 +1,103 @@
...
+  return {}
+}
+
+function asNestedArgs(value: unknown): JsArgs | undefined {
+  if (value && typeof value === 'object' && !Array.isArray(value)) {
+    return value as JsArgs
+  }
+
+  return undefined
+}
```
jacek-prisma, line 101: slightly more idiomatic
```suggestion
function resolveNextArgs(args: JsArgs, relationFieldName: string): JsArgs {
  const select = args.select?.[relationFieldName]
  if (isNestedArgs(select)) {
    return select
  }

  const include = args.include?.[relationFieldName]
  if (isNestedArgs(include)) {
    return include
  }

  return {}
}

function isNestedArgs(value: unknown): value is JsArgs {
  return Boolean(value) && typeof value === 'object' && !Array.isArray(value)
}
```

packages/client/src/runtime/core/extensions/resolve-result-extension-context.ts
LucianBuzzo: I thought a bit more about this and I think that as a consumer I would prefer the query to fail loudly and noisily then to quietly continue.

LucianBuzzo:
@jacek-prisma I've applied your suggestions and fixed a minor typescript error in a test file. Should be good to go now.

jacek-prisma (approved):

PR #29238, opened 2026-02-23
fix(adapter-mariadb,adapter-planetscale): return strings for text columns with binary collation in raw queries

sakurai-ryo:
Fixes #29237

### adapter-mariadb
Replace the `BINARY_FLAG` check with a collation index check (`collation.index === 63`), matching the approach used by `mariadb-connector-nodejs`.
https://github.com/mariadb-corporation/mariadb-connector-nodejs/blob/3.5.1/lib/cmd/decoder/text-decoder.js#L44

### adapter-planetscale
Add a charset-based collation check (`field.charset !== 63`) in `fieldToColumnType` to distinguish text columns from true binary columns. Vitess converts CHAR/VARCHAR/TEXT columns with a binary collation to BINARY/VARBINARY/BLOB respectively, so the charset value is used to detect this conversion.
https://github.com/planetscale/database-js/blob/de78eebfaec8cd88c670b8c644fc5a3fd69e664c/src/cast.ts#L92

## Root cause

Because the `mapColumnType` function in `adapter-mariadb` returns `ColumnTypeEnum.Bytes` when a column has a Binary Collation,
https://github.com/prisma/prisma/blob/7.4.1/packages/adapter-mariadb/src/conversion.ts#L86

and the `fieldToColumnType` function in `adapter-planetscale` does not inspect the charset at all for BLOB/BINARY/VARBINARY types,
https://github.com/prisma/prisma/blob/7.4.1/packages/adapter-planetscale/src/conversion.ts#L85

the `deserializeValue` function unintentionally converts even string data into a `Uint8Array`, treating it as a base64 string.
https://github.com/prisma/prisma/blob/7.4.1/packages/client/src/runtime/utils/deserializeRawResults.ts#L20

Since a Binary Collation can also apply to string data types like `CHAR`, the type determination logic in both `adapter-mariadb` and `adapter-planetscale` needs to be updated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved binary collation detection in MariaDB and PlanetScale database adapters. Columns with binary collation now correctly return as text data.

* **Tests**
  * Added MySQL column type tests to validate binary collation handling in query results.

### DIFF
--- a/packages/adapter-mariadb/src/conversion.ts
+++ b/packages/adapter-mariadb/src/conversion.ts
@@ -2,7 +2,8 @@ import { ArgType, ColumnType, ColumnTypeEnum, ResultValue } from '@prisma/driver
 import * as mariadb from 'mariadb'
 
 const UNSIGNED_FLAG = 1 << 5
-const BINARY_FLAG = 1 << 7
+// https://github.com/mariadb-corporation/mariadb-connector-nodejs/blob/be72ebf9fee6e0bd153b6ff6e0bb252f794dbf0e/lib/const/collations.js#L150
+const BINARY_COLLATION_INDEX = 63
 
 const enum MariaDbColumnType {
   DECIMAL = 'DECIMAL',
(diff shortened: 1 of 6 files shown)


jacek-prisma:
The newly added test is failing for `planetscale`

sakurai-ryo:
I added some fixes to adapter-planetscale as well.

jacek-prisma (approved):

PR #29245, opened 2026-02-24
fix(format): normalize line endings to LF when writing formatted schema

turanalmammadov:
Fixes #8548

## Problem

On Windows, `prisma format` writes the formatted schema with a trailing `CRLF` (`\r\n`) while every other line uses `LF` (`\n`). This causes an endless cycle on cross-platform teams:

1. Developer on Windows runs `prisma format` → file ends with `CRLF`
2. Linux CI runs `prisma format` → file ends with `LF`
3. Git sees a diff → CI fails

## Root Cause

The `prisma-schema` Wasm formatter appends a trailing `\r\n` on Windows. The raw output is written directly to disk with `fs.writeFile` without normalising line endings.

## Fix

**`packages/cli/src/Format.ts`**

1. **Write path** – strip `\r\n` and lone `\r` before calling `fs.writeFile`, guaranteeing LF-only output on all platforms.

2. **`--check` path** – normalise both the on-disk content _and_ the formatter output before comparing, so a repo that uses CRLF checkouts (Windows `git config core.autocrlf true`) is not spuriously flagged as unformatted.

## Normalisation Strategy

```typescript
const normalizedData = data.replace(/\r\n/g, '\n').replace(/\r/g, '\n')
```

- First pass: Windows `CRLF` → `LF`
- Second pass: Classic Mac `CR` → `LF`

## Impact

- No functional change on Linux / macOS (output is already LF-only)
- Fixes the Windows trailing-CRLF issue
- Fixes the `--check` false-positive on Windows git checkouts
- Zero additional dependencies

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed line ending inconsistencies (CRLF vs LF) that caused false formatting comparisons and prevented consistent behavior across different operating systems.

### DIFF
--- a/packages/cli/src/Format.ts
+++ b/packages/cli/src/Format.ts
@@ -93,7 +93,10 @@ Or specify a Prisma schema path
           return new HelpError(`${bold(red(`!`))} The schema ${underline(filename)} is not found in the schema list.`)
         }
         const [, originalSchema] = originalSchemaTuple
-        if (originalSchema !== formattedSchema) {
+        // Normalize line endings before comparing to avoid false positives when
+        // the file on disk uses CRLF (Windows) but the formatter emits LF.
+        const normalizeNewlines = (s: string) => s.replace(/\r\n/g, '\n').replace(/\r/g, '\n')
+        if (normalizeNewlines(originalSchema) !== normalizeNewlines(formattedSchema)) {
           return new HelpError(
             `${bold(red(`!`))} There are unformatted files. Run ${underline('prisma format')} to format them.`,
           )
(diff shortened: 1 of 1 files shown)


packages/cli/src/Format.ts
```diff
@@ -103,7 +106,14 @@ Or specify a Prisma schema path
     }
 
     for (const [filename, data] of formattedDatamodel) {
-      await fs.writeFile(filename, data)
+      // Normalize line endings to LF (\n) before writing.
+      // On Windows, the Wasm formatter appends a trailing CRLF (\r\n) which
+      // causes spurious diffs in cross-platform CI pipelines where git is
+      // configured to use LF. Stripping all \r characters here ensures the
+      // written file always uses Unix-style LF endings, matching the behaviour
+      // on Linux and macOS.
```
jacek-prisma, line 114: This shouldn't be necessary if you're using `core.autocrlf` (and most crossplatform projects should)

jacek-prisma: Also even if we did this change, it should probably happen in the `prisma-engines` repo and not here, ideally the formatter should just output the expected line endings

jacek-prisma (changes requested):
Left 1 important comment

SevInf:

PR #29255, opened 2026-02-25
feat(cli): update Studio to @prisma/studio-core 0.15.0

sorenbs:
## Summary
- bump `@prisma/studio-core` in CLI from `0.13.1` to `0.15.0`
- remove legacy Accelerate (`prisma+postgres`) URL transformation logic from `prisma studio`
- fail early for Accelerate URLs (`prisma://` and `prisma+postgres://`) with a clear user-facing error
- return Studio validation errors through `help()` so they render without stack traces

## Validation
- `node build/index.js studio --browser none --port 5557 --url "postgresql://...@db.prisma.io:5432/postgres"` starts successfully
- `node build/index.js studio --browser none --url "prisma+postgres://..."` exits with clean unsupported-Accelerate messaging

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized user-facing errors for unsupported/invalid connection URLs (including accelerate-like protocols) and clearer messaging.
  * Improved MySQL URL compatibility: normalizes sslaccept to mysql2 ssl options, maps connection_limit to connectionLimit, and strips Prisma-specific MySQL query params.
  * CLI now prints concise messages for user-facing errors and help.

* **Tests**
  * Added tests for MySQL URL normalization (sslaccept and connection_limit).

* **Chores**
  * Dependency updated: @prisma/studio-core bumped to 0.16.3.


packages/cli/src/Studio.ts
```diff
@@ -87,6 +82,14 @@ const PRISMA_ORM_SPECIFIC_QUERY_PARAMETERS = [
   'statement_cache_size',
 ] as const
 
+const PRISMA_ORM_SPECIFIC_MYSQL_QUERY_PARAMETERS = [
```
jacek-prisma, line 86: We no longer have these Prisma ORM specific parameters, since now users directly create driver configuration themselves

packages/cli/src/Studio.ts
```diff
@@ -323,21 +285,25 @@ ${bold('Examples')}
...
     if (!connectionString) {
-      return new Error(
+      return this.help(
         'No database URL found. Provide it via the `--url <url>` argument or define it in your Prisma config file as `datasource.url`.',
       )
     }
 
     if (!URL.canParse(connectionString)) {
-      return new Error('The provided database URL is not valid.')
+      return this.help('The provided database URL is not valid.')
```
jacek-prisma: I think showing CLI help when the `prisma.config.ts` is not a good idea, the user has to scroll a lot to see the error message and it makes it look like it's an issue with the way the CLI command is issues but it's not

I'd revert all of these `this.help` back to `return new Error`

sorenbs: The problem with `return new Error` is that it prints a stack trace, which I think is worse then printing the full help output. It looks like this:

```
Error: Prisma Studio no longer supports Accelerate URLs (prisma:// or prisma+postgres://). Use a direct database connection string instead.
    at e.parse (/Users/sorenschmidt/code/prisma/packages/cli/build/index.js:4542:579)
    at e.parse (/Users/sorenschmidt/code/prisma/packages/cli/build/index.js:1060:2327)
    at _or (/Users/sorenschmidt/code/prisma/packages/cli/build/index.js:4647:144)
```

packages/cli/src/Studio.ts
sorenbs: That's what I thought too! But as far as I can tell, other CLI commands (including prisma db pull) still do this magic, so I decided to add it here such that `prisma studio` will work on all DATABASE_URLs supported by other CLI commands.

We have the additional issue that Planetscale currently tell their customers to use this specific format. See this internal thread for details: https://prisma-company.slack.com/archives/CEYCG2MCN/p1772092389142929

packages/cli/src/Studio.ts
jacek-prisma: I see, I think it might make sense to add a subclass of `Error` and handle it specifically here:
https://github.com/prisma/prisma/blob/6b6908332025f76fc5182eb23fabedda64b7a5a6/packages/cli/src/bin.ts#L169

In order to avoid printing the extra stack trace or CLI help. Many of these errors shouldn't print the CLI help

packages/cli/src/Studio.ts
jacek-prisma: Okay, fair enough 👍

packages/cli/src/Studio.ts
```diff
@@ -184,49 +179,8 @@ Please use Node.js >=22.5, Deno >=2.2 or Bun >=1.0 or ensure you have the \`bett
...
-      connectionURL.protocol = 'postgres:'
-      connectionURL.searchParams.delete(ACCELERATE_API_KEY_QUERY_PARAMETER)
-      connectionURL.searchParams.set('sslmode', 'require')
-
-      return await POSTGRES_STUDIO_STUFF.createExecutor(connectionURL.toString(), relativeTo)
-    },
-    reExportAdapterScript: POSTGRES_STUDIO_STUFF.reExportAdapterScript,
-  },
+  prisma: null,
+  'prisma+postgres': null,
```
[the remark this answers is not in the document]
jacek-prisma: I think this comment is correct

packages/cli/src/Studio.ts
sorenbs: I like it. Addressed in latest commit.

jacek-prisma (approved):

PR #29272, opened 2026-02-27
feat(client): Allow client extension to change query action

Stevnbak:
Adds an optional argument to the query function available in [Client Extensions query](https://www.prisma.io/docs/orm/prisma-client/client-extensions/query) that if set changes the operation that is performed.

### Example:
Can for example be used to implement soft deletion by changing any delete operation to an update operation:
```ts
...
.$extends(
	Prisma.defineExtension({
		name: "Soft Delete Extension",
		query: {
			$allModels: {
				async delete({args, query, model, operation}) {
					return query({...args, data: { isDeleted: true }}, "update");
				}
			}
		}
	})
)
...
```

## Summary by CodeRabbit

* **New Features**
  * Query extension callbacks and public typings now accept an optional operation/context parameter so callbacks can receive the operation when provided.

* **Breaking Changes**
  * Public callback and extension type signatures changed to include the optional operation parameter — update any custom extensions or handlers to the new signature.

### DIFF
--- a/packages/client/src/runtime/core/extensions/applyQueryExtensions.ts
+++ b/packages/client/src/runtime/core/extensions/applyQueryExtensions.ts
@@ -37,7 +37,7 @@ function iterateAndCallQueryCallbacks(
       // @ts-expect-error because not part of public API
       __internalParams: params,
       query: (args, operation, __internalParams = params) => {
-        if (operation) params.action = operation
+        if (operation) __internalParams.action = operation
         // we need to keep track of the current customDataProxyFetch
         // this is to cascade customDataProxyFetch like a middleware
         const currCustomFetch = __internalParams.customDataProxyFetch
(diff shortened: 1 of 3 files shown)


themavik (commented):
The middle `query` argument is now `operation`; any extension that still calls `query(args, __internalParams)` will treat that object as an action name and blow up at runtime. nit: `applyQueryExtensions.ts` and `ExtensionArgs.ts` show mode `100755` in the diff—probably accidental chmod noise worth reverting.

tensordreams:
Hi @Stevnbak — thanks for this, and apologies for the slow reply.

We are closing this one: inserting `operation` as the second positional parameter of the public `query()` callback shifts `__internalParams`, so existing extensions written as `query(args, __internalParams)` would silently misread it at runtime. Reshaping the extension API also needs a design decision we cannot take in this window.

PR #29274, opened 2026-02-27
Fix DateTime fields returning Invalid Date with unixepoch-ms

connortessaro:
## Summary
   Fix DateTime fields returning `Invalid Date` when using `unixepoch-ms`.

   ## Changes
   - Correct DateTime handling for `unixepoch-ms` (millisecond timestamps).
   - Add regression test covering DateTime read/write with `unixepoch-ms`.

   ## Tests
   - Added/updated tests for this regression and verified they pass.

   Fixes #28890

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer handling of very large integers from the database: values that fit safely are returned as numbers, while out-of-range integers fall back to strings to prevent precision loss.

* **Tests**
  * Added functional tests for millisecond-precision timestamps to verify created and aggregated datetime values are returned as valid Date instances across create, find, and aggregate scenarios.


jacek-prisma:
This doesn't actually address https://github.com/prisma/prisma/issues/28890, because that issue is only reproducible by creating the date column manually with type set to `INTEGER`. In order to verify that the fix actually works you should write a test that manually creates a column with type `INTEGER`.

connortessaro:
Thanks for flagging this. I added a test that creates a DateTime column backed by INTEGER and stores unixepoch-ms values to reproduce the issue discussed in #28890. The test now verifies that Prisma Client returns a valid Date (not Invalid Date) for that DateTime column.

jacek-prisma (approved):

PR #29284, opened 2026-03-02
fix(client-engine-runtime): preserve Decimal precision in increment/decrement operations

### LINKED ISSUE #29160: Precision loss when using increment/decrement with large Decimal values
## Bug description

When using `increment` or `decrement` operations on `Decimal` fields with large values (>10^18), Prisma appears to lose precision, resulting in incorrect calculated values. Manual calculation using Decimal operations works correctly.

## Additional context

This appears to be a type coercion or floating-point conversion issue within Prisma's increment/decrement operations. The database field is defined as `Decimal(30, 0)` which should support arbitrary precision, but the increment/decrement operations seem to be losing precision with values larger than JavaScript's safe integer range (2^53 - 1).

This is critical for financial applications, blockchain applications, or any use case requiring precise arithmetic with large numbers.

### Severity

## How to reproduce

1. Define a schema with a large Decimal field:
```prisma
model AssetAccount {
  id      Int     @id @default(autoincrement())
  assetId Int
  address String
  amount  Decimal @default(0) @db.Decimal(30, 0)
  
  @@unique([assetId, address])
}

model AssetHistory {
  id          Int     @id @default(autoincrement())
  assetId     Int
  address     String
  amount      Decimal @default(0) @db.Decimal(30, 0)
  totalAmount Decimal @default(0) @db.Decimal(30, 0)
  // ... other fields
}
```

2. Perform sequential increment/decrement operations:
```typescript
// Initial deposit
const account1 = await prisma.assetAccount.update({
  where: { assetId_address: { assetId: 1, address: "0xabc..." } },
  data: { amount: { increment: new Decimal("5000000000000000000000000000") } }
});
// Result: 5000000000000000000000000000 ✅

// First withdrawal
const account2 = await prisma.assetAccount.update({
  where: { assetId_address: { assetId: 1, address: "0xabc..." } },
  data: { amount: { decrement: new Decimal("1000000000000000000000") } }
});
// Result: 4999999000000000000000000000 ✅

// Second withdrawal
const account3 = await prisma.assetAccount.update({
  where: { assetId_address: { assetId: 1, address: "0xabc..." } },
  data: { amount: { decrement: new Decimal("1000000000000000000000") } }
});
// Expected: 4999998000000000000000000000
// Actual:   4999997999999999000000000000 ❌ (precision loss)
```

## Expected behavior

Each decrement of `1000000000000000000000` should subtract exactly that amount:
- Start: 5,000,000,000,000,000,000,000,000,000
- After withdrawal 1: 4,999,999,000,000,000,000,000,000,000
- After withdrawal 2: 4,999,998,000,000,000,000,000,000,000
- After withdrawal 3: 4,999,997,000,000,000,000,000,000,000

## Actual behavior

Precision is lost on subsequent operations:
- Start: 5,000,000,000,000,000,000,000,000,000
- After withdrawal 1: 4,999,999,000,000,000,000,000,000,000
- After withdrawal 2: 4,999,997,999,999,999,000,000,000,000 ❌
- After withdrawal 3: 4,999,996,999,999,999,000,000,000,000 ❌

### Frequency

Consistently reproducible

### Is this a regression?

No

### Workaround

Manual calculation works correctly:
```typescript
const account = await prisma.assetAccount.findUnique({
  where: { assetId_address: { assetId: 1, address: "0xabc..." } }
});

const newAmount = new Decimal(account.amount).sub(decrementValue);

const updated = await prisma.assetAccount.update({
  where: { assetId_address: { assetId: 1, address: "0xabc..." } },
  data: { amount: newAmount }
});
```

### Prisma Schema & Queries

```prisma
```
```ts
```

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

## Environment & setup

**Prisma Version:**
```
prisma                  : 7.3.0
@prisma/client          : 7.3.0
```

**Database:**
- [ ] PostgreSQL (version: )
- [x] MySQL (version:8.0.40 )
- [ ] Other: 

**Operating System:**
- [x] macOS
- [ ] Windows
- [ ] Linux

**Node.js version:**
```
v24.13.0
```

### Prisma Version

```
prisma               : 7.3.0
@prisma/client       : 7.3.0
Operating System     : darwin
Architecture         : arm64
Node.js              : v24.13.0
TypeScript           : 5.9.3
PSL                  : @prisma/prisma-schema-wasm 7.3.0-16.9d6ad21cbbceab97458517b147a6a09ff43aa735
Schema Engine        : schema-engine-cli 9d6ad21cbbceab97458517b147a6a09ff43aa735 (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Default Engines Hash : 9d6ad21cbbceab97458517b147a6a09ff43aa735
Studio               : 0.13.1
```

Not-Sarthak:
Fixes https://github.com/prisma/prisma/issues/29160

## Problem

When using `increment` or `decrement` operations on `Decimal` fields with large values (>2^53), Prisma loses precision because the query interpreter's `evalFieldOperation()` converts all operands to JavaScript `number` via `Number()`. JavaScript numbers are IEEE 754 doubles and cannot safely represent integers larger than `Number.MAX_SAFE_INTEGER` (2^53 - 1).

For example, incrementing `5000000000000000000000000000` by `1000000000000000000000` produces an incorrect result due to floating-point precision loss. This is critical for financial and blockchain applications that require precise arithmetic with large numbers.

## Fix

Added `evalArithmetic()` which detects when either operand is a string (the representation used for Decimal database columns) and uses the `decimal.js` library (already a dependency via `@prisma/client-runtime-utils`) for the computation. When both operands are regular numbers (integer/float columns), the existing JavaScript arithmetic is preserved unchanged.

The key insight is that database drivers return Decimal column values as strings (because they can't be represented as JS numbers), while integer and float columns return numbers. This makes `typeof value === 'string'` a reliable indicator for when Decimal-safe arithmetic is needed.

## Tests

Added unit tests in `packages/client-engine-runtime/src/interpreter/query-interpreter.test.ts` covering:
- Add/subtract/multiply/divide with large decimal strings (precision preserved)
- Division by zero returns null for both decimal and number paths
- Mixed operands (string value + number operand) use Decimal path
- Regular number arithmetic is unchanged (integers, floats)
- All existing tests (100/100) continue to pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decimal arithmetic (add, subtract, multiply, divide) to preserve very large and long-decimal values and handle mixed string/number operands.
  * Division-by-zero now consistently returns null.
  * Preserved decimal precision when sending parameters to MySQL.

* **Tests**
  * Added comprehensive unit and functional tests covering field-level arithmetic, mixed-type scenarios, decimal precision edge cases, and division-by-zero across multiple providers.

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/query-interpreter.test.ts
+++ b/packages/client-engine-runtime/src/interpreter/query-interpreter.test.ts
@@ -144,6 +144,24 @@ describe('mapRecord field operations', () => {
       expect((result as Record<string, unknown>).amount).toBeNull()
     })
 
+    test('add preserves precision beyond 20 significant digits', async () => {
+      const adapter = new MockDriverAdapter()
+      adapter.setQueryResult('SELECT', {
+        columnNames: ['id', 'amount'],
+        columnTypes: [0, 1],
+        rows: [[1, '1234567890123456789012345678']],
+      })
+
+      const plan = makeMapRecordPlan('SELECT id, amount FROM Account WHERE id = 1', {
+        amount: { type: 'add', value: '1' },
+      })
+
+      const interpreter = QueryInterpreter.forSql(createInterpreterOptions())
+      const result = await interpreter.run(plan, createRuntimeOptions(adapter))
+
+      expect((result as Record<string, unknown>).amount).toBe('1234567890123456789012345679')
+    })
+
     test('add with string value and number operand uses Decimal', async () => {
       const adapter = new MockDriverAdapter()
       adapter.setQueryResult('SELECT', {
--- a/packages/client-engine-runtime/src/interpreter/query-interpreter.test.ts
+++ b/packages/client-engine-runtime/src/interpreter/query-interpreter.test.ts
@@ -1,0 +1,221 @@
+import { describe, expect, test } from 'vitest'
+
+import { MockDriverAdapter, mockTracingHelper } from '../../bench/mock-adapter'
+import type { QueryPlanNode } from '../query-plan'
+import { QueryInterpreter, QueryInterpreterOptions } from './query-interpreter'
+import { serializeSql } from './serialize-sql'
+
+function createInterpreterOptions(): QueryInterpreterOptions {
+  return {
+    tracingHelper: mockTracingHelper,
+    serializer: serializeSql,
+    provider: 'postgres',
+    connectionInfo: { supportsRelationJoins: false },
+  }
+}
+
+function createRuntimeOptions(adapter: MockDriverAdapter) {
+  return {
+    queryable: adapter,
+    scope: {},
+    transactionManager: { enabled: false },
+  } as const
+}
+
+/**
+ * Constructs a query plan that SELECTs a record and then applies a mapRecord
+ * field operation, simulating an update with increment/decrement.
+ */
+function makeMapRecordPlan(
+  selectSql: string,
+  fields: Record<string, { type: 'add' | 'subtract' | 'multiply' | 'divide'; value: unknown }>,
+): QueryPlanNode {
+  return {
+    type: 'mapRecord',
+    args: {
+      expr: {
+        type: 'unique',
+        args: {
+          type: 'query',
+          args: {
+            type: 'templateSql',
+            fragments: [{ type: 'stringChunk', chunk: selectSql }],
+            placeholderFormat: { prefix: '$', hasNumbering: true },
+            args: [],
+            argTypes: [],
+            chunkable: false,
+          },
+        },
+      },
+      fields,
+    },
+  }
+}
+
+describe('mapRecord field operations', () => {
+  describe('decimal precision', () => {
+    test('add preserves precision for large decimal strings', async () => {
+      const adapter = new MockDriverAdapter()
+      adapter.setQueryResult('SELECT', {
+        columnNames: ['id', 'amount'],
+        columnTypes: [0, 1],
+        rows: [[1, '5000000000000000000000000000']],
+      })
+
+      const plan = makeMapRecordPlan('SELECT id, amount FROM Account WHERE id = 1', {
+        amount: { type: 'add', value: '1000000000000000000000' },
+      })
+
+      const interpreter = QueryInterpreter.forSql(createInterpreterOptions())
+      const result = await interpreter.run(plan, createRuntimeOptions(adapter))
+
+      expect((result as Record<string, unknown>).amount).toBe('5000001000000000000000000000')
+    })
+
+    test('subtract preserves precision for large decimal strings', async () => {
+      const adapter = new MockDriverAdapter()
+      adapter.setQueryResult('SELECT', {
+        columnNames: ['id', 'amount'],
+        columnTypes: [0, 1],
+        rows: [[1, '5000000000000000000000000000']],
+      })
+
+      const plan = makeMapRecordPlan('SELECT id, amount FROM Account WHERE id = 1', {
+        amount: { type: 'subtract', value: '1000000000000000000000' },
+      })
+
+      const interpreter = QueryInterpreter.forSql(createInterpreterOptions())
+      const result = await interpreter.run(plan, createRuntimeOptions(adapter))
+
+      expect((result as Record<string, unknown>).amount).toBe('4999999000000000000000000000')
+    })
+
+    test('multiply preserves precision for large decimal strings', async () => {
+      const adapter = new MockDriverAdapter()
+      adapter.setQueryResult('SELECT', {
+        columnNames: ['id', 'amount'],
+        columnTypes: [0, 1],
+        rows: [[1, '1000000000000000000']],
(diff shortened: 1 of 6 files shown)


Not-Sarthak:
### Follow-up fix: High-precision Decimal clone

During thorough testing, I discovered that the initial implementation using `new Decimal()` had a subtle but critical issue: **Decimal.js defaults to only 20 significant digits of precision for arithmetic operations**. This means values with 21+ unique significant digits would silently lose precision during computation — the exact same class of bug we're trying to fix.

**Example of the problem with default precision:**
```
new Decimal('1234567890123456789012345678').plus('1').toFixed()
// Returns: '1234567890123456789000000000' (WRONG — last 9 digits zeroed out)
// Expected: '1234567890123456789012345679'
```

**The fix:** Use `Decimal.clone({ precision: 1000 })` to create a module-level high-precision Decimal constructor (`ArithmeticDecimal`). This:
- Is created once at module level (zero per-operation overhead)
- Supports up to 1000 significant digits (covers PostgreSQL NUMERIC's ~131K+16K range, MySQL DECIMAL's 65, and SQL Server's 38 in all practical scenarios)
- Added a dedicated test case for the >20 significant digit scenario

The original issue scenario (`5000000000000000000000000000 + 1000000000000000000000`) happened to work with default precision because the non-zero significant digits fit within 20, but any balance with many unique digits (like `1234567890123456789012345678`) would have been silently corrupted.

packages/client-engine-runtime/src/interpreter/query-interpreter.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      const plan = makeMapRecordPlan('SELECT id, credit FROM Account WHERE id = 1', {
+        credit: { type: 'add', value: 1.5 },
+      })
+
+      const interpreter = QueryInterpreter.forSql(createInterpreterOptions())
+      const result = await interpreter.run(plan, createRuntimeOptions(adapter))
+
+      expect((result as Record<string, unknown>).credit).toBe(12)
+    })
+  })
```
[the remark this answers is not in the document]
Not-Sarthak: Added this test in 04723e230. Thanks for the suggestion!

jacek-prisma:
This doesn't actually fix #29160, although it might still be a valid change.
The issue in #29160 is a loss of precision when serializing the number into the MySQL protocol (and I already opened a PR to fix it: https://github.com/prisma/prisma/pull/29285)
You should try adding a functional test (in `packages/client/tests/functional/`) that actually demonstrates a real loss of precision rather than adding query interpreter test.

Not-Sarthak:
ohh yes, you're right

the real issue is mysql treating `?` params as doubles in arithmetic expressions like `SET amount = amount + ?`, not the js side arithmetic.

fixed in ea8dd8c by wrapping decimal typed params with `CAST(? AS DECIMAL(65,30))` when provider is `mysql`, inside `renderFragment()` in `render-query.ts`. also added a functional test in `packages/client/tests/functional/decimal/increment-precision/` that reproduces the exact precision loss from #29160. verified locally against mysql with `--adapter js_mariadb`.

the earlier `evalArithmetic` changes are still valid for the mapRecord path where arithmetic happens in js rather than in the database.

packages/client-engine-runtime/src/interpreter/render-query.ts
```diff
@@ -116,11 +118,21 @@ function renderFragment<Type extends DeepReadonly<DynamicArgType> | undefined>(
...
-    case 'parameter':
-      return formatPlaceholder(placeholderFormat, ctx.placeholderNumber++)
+    case 'parameter': {
+      const placeholder = formatPlaceholder(placeholderFormat, ctx.placeholderNumber++)
+      // MySQL treats parameterized values as doubles in arithmetic expressions,
+      // losing precision for large decimals. Wrapping with CAST ensures MySQL
+      // interprets the value as DECIMAL with full precision.
+      // https://github.com/prisma/prisma/issues/29160
+      if (provider === 'mysql' && hasDecimalArgType(fragment)) {
+        return `CAST(${placeholder} AS DECIMAL(65,30))`
```
jacek-prisma, line 132: I think is no longer needed now that https://github.com/prisma/prisma/issues/29160 has been fixed already

packages/client/tests/functional/decimal/increment-precision/tests.ts
```diff
@@ -0,0 +1,73 @@
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { Prisma as PrismaNamespace, PrismaClient } from './generated/prisma/client'
+
+declare let prisma: PrismaClient
+declare let Prisma: typeof PrismaNamespace
+
+// Reproduces https://github.com/prisma/prisma/issues/29160
```
jacek-prisma, line 8: https://github.com/prisma/prisma/issues/29160 has already been fixed and covered with a test in another PR

PR #29287, opened 2026-03-02
feat(adapter-pg): accept connection string URL in PrismaPg constructor

Not-Sarthak:
## Summary

- Adds support for passing a raw connection string URL directly to `PrismaPg` (e.g., `new PrismaPg(process.env.DATABASE_URL)`)
- The constructor now accepts `pg.Pool | pg.PoolConfig | string`, converting strings to `{ connectionString: string }` internally
- This is a convenience feature — `pg.PoolConfig` already supports `{ connectionString: '...' }`, but accepting a plain string is more ergonomic

Related: #29151

## Test plan

- [x] Added unit test verifying `PrismaPgAdapterFactory` accepts a connection string URL and creates a working adapter
- [x] All existing adapter-pg tests pass (36/36)
- [x] Lint passes with no new warnings

> **Note:** The CodSpeed regression (`compile findUnique -97.77%`) is unrelated — this PR only touches `adapter-pg` constructor logic, not query compilation.

### DIFF
--- a/packages/adapter-pg/src/__tests__/pg.test.ts
+++ b/packages/adapter-pg/src/__tests__/pg.test.ts
@@ -34,6 +34,15 @@ describe('PrismaPgAdapterFactory', () => {
     await adapter.dispose()
   })
 
+  it('should accept a connection string URL', async () => {
+    const connectionString = 'postgresql://test:test@localhost:5432/test'
+    const factory = new PrismaPgAdapterFactory(connectionString)
+    const adapter = await factory.connect()
+    expect(adapter).toBeDefined()
+    expect(adapter.adapterName).toBeDefined()
+    await adapter.dispose()
+  })
+
   it('should add and remove error event listener when using an external Pool', async () => {
     const pool = new pg.Pool({ user: 'test', password: 'test', database: 'test', port: 5432, host: 'localhost' })
     pool.on('error', () => {})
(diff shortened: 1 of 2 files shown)


packages/adapter-pg/src/__tests__/pg.test.ts
```diff
@@ -34,6 +34,15 @@ describe('PrismaPgAdapterFactory', () => {
...
   })
 
+  it('should accept a connection string URL', async () => {
+    const connectionString = 'postgresql://test:test@localhost:5432/test'
+    const factory = new PrismaPgAdapterFactory(connectionString)
+    const adapter = await factory.connect()
+    expect(adapter).toBeDefined()
+    expect(adapter.adapterName).toBeDefined()
+    await adapter.dispose()
+  })
```
Not-Sarthak, line 46: Strengthened in 1f0b6e39c. Now asserts both the factory's internal config normalization and the pool's connectionString.

jacek-prisma:
I think the original issue was linked to missing Typescript typings (which has already been resolved on `main`), see https://github.com/prisma/prisma/issues/29151#issuecomment-3991005113, which means this is no longer a bug fix, although this change could be merged as a new feature.

Not-Sarthak:
> I think the original issue was linked to missing Typescript typings (which has already been resolved on `main`), see [#29151 (comment)](https://github.com/prisma/prisma/issues/29151#issuecomment-3991005113), which means this is no longer a bug fix, although this change could be merged as a new feature.

Thanks for the context! updated the PR title, please lmk if there's anything i am missing

jacek-prisma (approved):

PR #29306, opened 2026-03-05
feat(client): add initial support for polymorphic relations

ian:
This adds the foundation for polymorphic relations (e.g., item Post | Comment):

- Add isPolymorphic, relationTypes, relationDiscriminator to DMMF types
- Generate union types for polymorphic relation outputs in TS/JS generators
- Add { on: 'Post' } syntax support for include/select filtering
- Handle discriminator-based query routing in runtime
- Add test fixture for polymorphic relations

Note: Schema parser changes in prisma-engines are still needed to parse the Post | Comment syntax and `@polymorphic` attribute.

## Example Usage

###Schema

```
model Vote {
  id        String         @id @default(cuid())
  item      Post | Comment @polymorphic(discriminator: itemType)
  itemId    String
  itemType  String
}

model Post {
  id    String @id @default(cuid())
  title String
  votes Vote[]
}

model Comment {
  body String
  votes Vote[]
}
```

### Querying

```ts
// Union type (Post | Comment)
const vote = await prisma.vote.findUnique({
  where: { id: '1' },
  include: { item: true },
})
// vote.item: Post | Comment

// Filtered by `on` (Post only)
const voteWithPost = await prisma.vote.findUnique({
  where: { id: '1' },
  include: { item: { on: 'Post' } },
})
// voteWithPost.item: Post | null

// Filter by discriminator
const postVotes = await prisma.vote.findMany({
  where: { itemType: 'Post' },
  include: { item: true },
})
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -127,6 +127,50 @@
   - Test files named `*.test.ts` are excluded from build output via esbuild config; place tests alongside source files.
   - **Update this file** (`AGENTS.md`) whenever you learn something new about the codebase that would be valuable for future tasks.
 
+- **Polymorphic relations** (issue #1644):
+  - Syntax: `item Post | Comment @polymorphic(discriminator: "itemType")` in schema
+  - Grammar rule in `psl/schema-ast/src/parser/datamodel.pest`: `polymorphic_type = { identifier ~ ("|" ~ identifier)+ }`
+  - `FieldType::Polymorphic(Vec<Identifier>)` in AST (`psl/schema-ast/src/ast/field.rs`)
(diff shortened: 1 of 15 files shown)


aqrln:
Thanks a lot for giving this a shot. We're cleaning up the pull requests in preparation for moving the [Prisma Next](https://github.com/prisma/prisma-next) rewrite to the main repo.

This PR is missing the corresponding work on the PSL and the query compiler side (I see from the commits that you had changes locally but I don't believe I've seen them opened as a PR in prisma-engines repo) and has been stuck in draft for a while, so I'll close it for now.

Additionally, Prisma Next does implement polymorphic relations (both STI and MTI), so this feature is going to exist in a future major release.

PR #29317, opened 2026-03-07
refactor: extract 'prisma-client-js' into PRISMA_CLIENT_JS_PROVIDER constant

Felipeness:
## Summary

Closes #29316

Extracts the hardcoded `'prisma-client-js'` string literal into a shared `PRISMA_CLIENT_JS_PROVIDER` constant defined in `@prisma/internals`, replacing all occurrences across the codebase.

### Changes

- **New**: `packages/internals/src/prisma-client-js-provider.ts` — single source of truth for the constant
- **Modified** (8 files across 4 packages):
  - `packages/internals/src/index.ts` — export the constant
  - `packages/internals/src/cli/getGeneratorSuccessMessage.ts` — use constant
  - `packages/internals/src/utils/extractPreviewFeatures.ts` — use constant
  - `packages/cli/src/Generate.ts` — use constant (2 occurrences)
  - `packages/cli/src/utils/checkpoint.ts` — use constant
  - `packages/client-generator-js/src/generator.ts` — use constant
  - `packages/client-generator-js/src/utils/types/dmmfToTypes.ts` — use constant (2 occurrences)
  - `packages/client/src/utils/getTestClient.ts` — use constant

### Motivation

This was identified during the review of #29315 by @coderabbitai. Having a single constant:
- Eliminates risk of typos in the provider string
- Makes it easy to find all usages via "Find References"
- Provides a single place to update if the value ever changes

## Test plan

- [x] `pnpm build` passes (43 tasks)
- [x] `getGenerators` tests pass (40/40, 33 snapshots)
- [x] No behavioral changes — pure mechanical refactor

## Summary by CodeRabbit

* **Refactor**
  * Standardized provider identifiers across internal modules using a centralized constant.
  * No changes to behavior or public APIs; end-user experience remains unchanged.
  * Improved internal consistency and reliability of provider detection and generator selection logic.


packages/cli/src/utils/checkpoint.ts
```diff
@@ -7,6 +7,7 @@ import {
   isCurrentBinInstalledGlobally,
   loadSchemaContext,
   parseEnvValue,
+  PRISMA_CLIENT_JS_PROVIDER,
```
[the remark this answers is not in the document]
Felipeness: Good catch! Though `'prisma-client'` is actually a different provider — it's the alias for the newer TypeScript generator, not a variant of `'prisma-client-js'`. I kept the scope focused on what #29316 raised specifically.

If you think it's worth extracting the other provider names too, feel free to open a broader issue covering all hardcoded provider strings across the codebase — there are a few more like `'prisma-client-ts'` as well. Happy to help with that if needed!

Felipeness:
Hey! Just a heads-up that the "Milestone Check" is blocking this PR because no milestone has been assigned yet — and as an external contributor I can't set it myself. Would a maintainer be able to set the right milestone here (I'm guessing `7.5.0`)? Thank you!

packages/internals/src/prisma-client-js-provider.ts
```diff
@@ -0,0 +1 @@
+export const PRISMA_CLIENT_JS_PROVIDER = 'prisma-client-js'
```
jacek-prisma: How about
```suggestion
export const enum BuiltInProvider {
  PrismaClientJs = 'prisma-client-js',
  PrismaClientTs = 'prisma-client'
}
```
Or
```ts
export const BuiltInProvider = {
  'PrismaClientJs': 'prisma-client-js',
  'PrismaClientTs': 'prisma-client'
}
```

jacek-prisma (changes requested):
Suggested 1 change

packages/internals/src/built-in-provider.ts
```diff
@@ -1 +1,4 @@
-export const PRISMA_CLIENT_JS_PROVIDER = 'prisma-client-js'
+export const enum BuiltInProvider {
```
jacek-prisma: This file should be renamed to `built-in-provider.ts`

Felipeness:
Hey Jacek! Applied both suggestions — grouped the providers into a `BuiltInProvider` const enum and renamed the file to `built-in-provider.ts`. Also took the opportunity to replace the hardcoded `'prisma-client'` in `checkpoint.ts` with `BuiltInProvider.PrismaClientTs`. Let me know if it looks good!

packages/internals/src/prisma-client-js-provider.ts
Felipeness: Hey Jacek! I went with the `const enum` initially, but it turns out the project uses `isolatedModules: true` in `tsconfig.build.regular.json`, which makes `const enum` fail across package boundaries with `TS2748`:

```
Cannot access ambient const enums when 'isolatedModules' is enabled.
```

So I switched to the second option you suggested — the `as const` object. Same API, works with `isolatedModules`.

packages/internals/src/built-in-provider.ts
Felipeness: Done! Renamed in commit `1427d3b`.

Felipeness:
Hey @jacek-prisma! Quick update on the latest changes:

**Fixed the CI failures** — the `const enum` approach didn't work because the project uses `isolatedModules: true` in `tsconfig.build.regular.json`. Cross-package `const enum` access fails with `TS2748`. Switched to the `as const` object (your second suggestion), which works identically but is compatible with `isolatedModules`.

**Changes in the latest push:**
- Rebased on latest `main`
- `const enum` → `as const` object (fixes all 216 CI failures)
- Added `string[]` type annotation in `checkpoint.ts` (needed because `as const` creates literal types that are stricter with `.includes()`)
- Added JSDoc comment on `BuiltInProvider`

**Still need maintainer action:**
- Set milestone (guessing `7.5.0`?)
- Approve the CI workflows
- Re-review when you get a chance

Thanks!

jacek-prisma (approved):

PR #29320, opened 2026-03-09
fix(adapter-pg): transform prisma+postgres://localhost URLs for local...

Br1an67:
Fixes #29191

## Summary

Users using `npx prisma dev` to create a local Prisma Postgres database receive a connection string with the `prisma+postgres://` protocol (e.g., `prisma+postgres://localhost:51216/?api_key=xxx`). When they try to use this URL with `@prisma/adapter-pg`, the `pg` library fails to connect because it doesn't recognize the `prisma+postgres://` protocol, resulting in a "Connection terminated unexpectedly" error.

This PR adds URL transformation logic to the `PrismaPgAdapterFactory` constructor that:
1. Detects `prisma+postgres://` URLs pointing to localhost (local development servers)
2. Transforms them to standard `postgres://` URLs that the `pg` driver can use
3. Throws a clear error for remote `prisma+postgres://` URLs (Accelerate) with guidance to use `@prisma/adapter-ppg` or `accelerateUrl` instead

## Changes

- **packages/adapter-pg/src/pg.ts**: Added URL transformation logic in `PrismaPgAdapterFactory` constructor
  - Added `isPrismaPostgresDevUrl()` helper to detect local dev URLs
  - Added `transformPrismaPostgresToPgUrl()` to convert `prisma+postgres://` to `postgres://`
  - Transforms local dev URLs automatically, throws error for remote Accelerate URLs
- **packages/adapter-pg/src/__tests__/pg.test.ts**: Added 6 new tests for URL transformation

## Testing

All 41 tests pass including the new URL transformation tests:
- `should transform prisma+postgres://localhost URLs to postgres:// URLs`
- `should transform prisma+postgres://127.0.0.1 URLs to postgres:// URLs`
- `should transform prisma+postgres://[::1] URLs to postgres:// URLs`
- `should throw error for remote prisma+postgres:// URLs (Accelerate)`
- `should leave standard postgres:// URLs unchanged`
- `should leave standard postgresql:// URLs unchanged`

## Diff stats

```
 packages/adapter-pg/src/__tests__/pg.test.ts | 52 ++++++++++++++++++++++++
 packages/adapter-pg/src/pg.ts                | 61 ++++++++++++++++++++++++++++
 2 files changed, 113 insertions(+)
```

## Summary by CodeRabbit

* **New Features**
  * Local Prisma Postgres development URLs are automatically converted to standard PostgreSQL format for seamless local usage.
  * Remote Prisma Postgres URLs are explicitly rejected with clear error messages to prevent unsupported remote usage.

* **Tests**
  * Added comprehensive tests covering URL transformation, local credential decoding, fallback behavior, and rejection cases.

### DIFF
--- a/packages/adapter-pg/src/__tests__/pg.test.ts
+++ b/packages/adapter-pg/src/__tests__/pg.test.ts
@@ -4,6 +4,58 @@ import { describe, expect, it, vi } from 'vitest'
 
 import { PrismaPgAdapterFactory } from '../pg'
 
+describe('PrismaPgAdapterFactory URL transformation', () => {
+  it('should transform prisma+postgres://localhost URLs to postgres:// URLs', () => {
+    const config: pg.PoolConfig = {
+      connectionString: 'prisma+postgres://localhost:51216/?api_key=test123',
+    }
+    const factory = new PrismaPgAdapterFactory(config)
+    // The factory should transform the URL internally
+    expect(factory).toBeDefined()
+  })
+
+  it('should transform prisma+postgres://127.0.0.1 URLs to postgres:// URLs', () => {
+    const config: pg.PoolConfig = {
+      connectionString: 'prisma+postgres://127.0.0.1:5432/mydb',
+    }
+    const factory = new PrismaPgAdapterFactory(config)
+    expect(factory).toBeDefined()
+  })
+
+  it('should transform prisma+postgres://[::1] URLs to postgres:// URLs', () => {
+    const config: pg.PoolConfig = {
+      connectionString: 'prisma+postgres://[::1]:5432/mydb',
+    }
+    const factory = new PrismaPgAdapterFactory(config)
+    expect(factory).toBeDefined()
+  })
+
+  it('should throw error for remote prisma+postgres:// URLs (Accelerate)', () => {
+    const config: pg.PoolConfig = {
+      connectionString: 'prisma+postgres://accelerate.prisma-data.net/db?api_key=test123',
+    }
+    expect(() => new PrismaPgAdapterFactory(config)).toThrow(
+      'The "prisma+postgres://" protocol is not supported by @prisma/adapter-pg',
+    )
+  })
+
+  it('should leave standard postgres:// URLs unchanged', () => {
+    const config: pg.PoolConfig = {
+      connectionString: 'postgres://user:password@localhost:5432/mydb',
+    }
+    const factory = new PrismaPgAdapterFactory(config)
+    expect(factory).toBeDefined()
+  })
+
+  it('should leave standard postgresql:// URLs unchanged', () => {
+    const config: pg.PoolConfig = {
+      connectionString: 'postgresql://user:password@localhost:5432/mydb',
+    }
+    const factory = new PrismaPgAdapterFactory(config)
+    expect(factory).toBeDefined()
+  })
+})
+
 describe('PrismaPgAdapterFactory', () => {
   it('should subscribe to pool error events', async () => {
     const config: pg.PoolConfig = { user: 'test', password: 'test', database: 'test', port: 5432, host: 'localhost' }
--- a/packages/adapter-pg/src/__tests__/pg.test.ts
+++ b/packages/adapter-pg/src/__tests__/pg.test.ts
@@ -5,29 +5,44 @@ import { describe, expect, it, vi } from 'vitest'
 import { PrismaPgAdapterFactory } from '../pg'
 
 describe('PrismaPgAdapterFactory URL transformation', () => {
-  it('should transform prisma+postgres://localhost URLs to postgres:// URLs', () => {
+  // Helper to create a base64url-encoded api_key payload matching the local PPg format
+  function encodeApiKey(databaseUrl: string, shadowDatabaseUrl = databaseUrl): string {
+    return Buffer.from(JSON.stringify({ databaseUrl, shadowDatabaseUrl })).toString('base64url')
+  }
+
+  it('should extract databaseUrl from api_key for prisma+postgres://localhost URLs', () => {
+    const apiKey = encodeApiKey('postgres://localhost:51216/postgres')
     const config: pg.PoolConfig = {
-      connectionString: 'prisma+postgres://localhost:51216/?api_key=test123',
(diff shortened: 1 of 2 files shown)


packages/adapter-pg/src/__tests__/pg.test.ts
```diff
@@ -4,6 +4,58 @@ import { describe, expect, it, vi } from 'vitest'
...
 import { PrismaPgAdapterFactory } from '../pg'
 
+describe('PrismaPgAdapterFactory URL transformation', () => {
+  it('should transform prisma+postgres://localhost URLs to postgres:// URLs', () => {
+    const config: pg.PoolConfig = {
+      connectionString: 'prisma+postgres://localhost:51216/?api_key=test123',
+    }
+    const factory = new PrismaPgAdapterFactory(config)
+    // The factory should transform the URL internally
+    expect(factory).toBeDefined()
```
jacek-prisma: This isn't sufficiently testing the behavior, you can check what URL is being set by mocking parts of the adapter

packages/adapter-pg/src/pg.ts
```diff
@@ -271,6 +271,51 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
...
+    return false
+  }
+}
+
+/**
+ * Transforms a `prisma+postgres://` URL to a standard `postgres://` URL
+ * for direct TCP connection with the pg driver.
+ *
+ * Example: `prisma+postgres://localhost:51216/?api_key=xxx`
+ * becomes: `postgres://localhost:51216/postgres?api_key=xxx`
```
jacek-prisma:
```suggestion
 * becomes: `postgres://localhost:51216/postgres`
```

packages/adapter-pg/src/pg.ts
```diff
@@ -271,6 +271,51 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
...
+  const transformedUrl = new URL(url.href.replace(PRISMA_POSTGRES_PROTOCOL, 'postgres:'))
+
+  // Ensure there's a database name (pg driver requires one)
+  // Default to 'postgres' if not specified
+  if (!transformedUrl.pathname || transformedUrl.pathname === '/') {
+    transformedUrl.pathname = '/postgres'
+  }
+
+  // Remove the api_key parameter as it's not used by the pg driver
+  transformedUrl.searchParams.delete('api_key')
```
jacek-prisma, line 338: When using a local prisma postgres URL, the `api_key` actually contains a base64 encoded JSON object with the full local connection string of pglite.
I think it'd be better to extract it here and use it rather than creating a plain localhost URL.

jacek-prisma (changes requested):
Left 2 important comments

Br1an67:
Thanks for the review @jacek-prisma! All feedback has been addressed in the latest commit:

1. **Decode `api_key` to extract the actual connection string** — The `api_key` is now decoded as a base64url JSON payload (matching the `LocalPpgApiKey` structure used in `prisma-engines`) to extract the embedded `databaseUrl`, with a fallback to simple protocol replacement if decoding fails.

2. **Tests now verify actual transformed URLs** — Tests assert on `factory['config'].connectionString` to verify the exact transformation result, including a new test case for the fallback path.

3. **Fixed doc comment** — Updated to reflect the actual behavior.

packages/adapter-pg/src/__tests__/pg.test.ts
```diff
@@ -4,6 +4,73 @@ import { describe, expect, it, vi } from 'vitest'
...
+  it('should extract databaseUrl from api_key for prisma+postgres://[::1] URLs', () => {
+    const apiKey = encodeApiKey('postgres://[::1]:5432/mydb')
+    const config: pg.PoolConfig = {
+      connectionString: `prisma+postgres://[::1]:5432/mydb?api_key=${apiKey}`,
+    }
+    const factory = new PrismaPgAdapterFactory(config)
+    expect(factory['config'].connectionString).toBe('postgres://[::1]:5432/mydb')
+  })
+
+  it('should fall back to simple URL transformation when api_key is not a valid payload', () => {
```
jacek-prisma, line 40: I think this scenario should be treated as a hard error, since it should normally never occur

jacek-prisma:
Left 1 comment here: https://github.com/prisma/prisma/pull/29320#discussion_r2946410689

tensordreams:
Hi @Br1an67 — thanks for this contribution, and sorry for the slow response.

https://cla-assistant.io/prisma/prisma?pullRequest=29320

packages/adapter-pg/src/pg.ts
```diff
@@ -275,24 +271,106 @@ export class PrismaPgAdapter extends PgQueryable<StdClient> implements SqlDriver
...
+          }
+        } else if (poolOrConfig.connectionString.startsWith(`${PRISMA_POSTGRES_PROTOCOL}//`)) {
+          // Remote prisma+postgres:// URLs (Accelerate) are not supported by adapter-pg
+          throw new Error(
+            `The "prisma+postgres://" protocol is not supported by @prisma/adapter-pg. ` +
+              `For remote Prisma Postgres (Accelerate), use @prisma/adapter-ppg or configure accelerateUrl in PrismaClient options. ` +
+              `For local development with "prisma dev", use the DATABASE_URL with a direct postgres:// connection string.`,
+          )
+        }
+      }
```
aqrln, line 373: I think this error message is all we need, for both local and remote Accelerate connection strings. Let's not rely on the implementation details of `@prisma/dev` to decode the connection string and reconstruct the raw Postgres connection string, let the user just provide it themselves in the first place.

I would also rewrite this to more actively discourage the usage of legacy `prisma+postgres://` URLs and and recommend direct or pooled `postgres://` connection strings for remote Prisma Postgres instances too.

tensordreams:
Hi @Br1an67 — thanks for this, and for working through the review feedback earlier.

We're closing out the open PRs that are still blocked on an unsigned CLA, and this is one of them: https://cla-assistant.io/prisma/prisma?pullRequest=29320

That's the only thing standing in the way here, not the change itself. If you're still up for it, sign there and reopen the PR, and we'll pick the review back up from where it left off.

Thanks again for the contribution.

PR #29322, opened 2026-03-09
Bump studio-core dependency to 0.21.1

sorenbs:
Summary
- update `@prisma/studio-core` in `packages/cli` and the lockfile to the new `0.21.1` release so Studio pulls the latest bundle
- Adapted BFF to support SQL linting
- refresh lockfile metadata so downstream installs resolve the updated package version

Studio Features
- More intuitive filtering
- Filter by SQL
- fulltext search across all columns
- SQL Query tab

## BFF changes

Implemented the CLI-side BFF extension needed for newer Studio SQL linting.

What changed:
- Added support for the `sql-lint` BFF procedure in `packages/cli/src/Studio.ts`
- Forwarded `sql-lint` requests to `executor.lintSql(...)` when the executor supports it
- Kept an explicit fallback for executors that do not implement SQL lint
- Improved BFF error normalization so already-serialized upstream errors are passed through unchanged, including RPC envelopes like `{"@@error": ...}`

Why this is needed:
- `@prisma/studio-core` now sends SQL lint requests through the BFF as `procedure: "sql-lint"`
- Without handling that procedure in the CLI, Studio falls through to `Unknown procedure`
- Some upstream failures arrive as serialized RPC error envelopes rather than native `Error` instances, so the CLI needs to unwrap those instead of turning them into generic fallback errors

Result:
- SQL query linting works through the CLI BFF
- When linting fails, Studio now gets the real database error payload instead of a generic `Unknown procedure` or fallback error

Validation:
- Added focused Vitest coverage for:
  - routing `sql-lint` requests through the BFF
  - unwrapping `@@error`-serialized failures
- `pnpm --filter prisma test src/__tests__/Studio.vitest.ts` passes

## Summary by CodeRabbit

* **New Features**
  * Added server-side SQL linting for the Studio backend endpoint.
* **Bug Fixes**
  * Standardized BFF-style error formatting and unwrapping for more consistent backend error responses.
* **Tests**
  * New tests for the SQL lint flow, error-unwrapping behavior, and related server-side handling.
* **Chores**
  * Updated dependency versions for improved compatibility.

### DIFF
--- a/packages/cli/src/__tests__/Studio.vitest.ts
+++ b/packages/cli/src/__tests__/Studio.vitest.ts
@@ -6,10 +6,19 @@ const serveMock = vi.fn(() => ({ close: vi.fn() }))
 const createPostgresJSExecutorMock = vi.fn(() => ({
   execute: vi.fn(),
 }))
-const serializeErrorMock = vi.fn((error: Error) => ({
-  message: error.message,
-  name: error.name,
-}))
+const serializeErrorMock = vi.fn((error: unknown) => {
+  if (error instanceof Error) {
+    return {
+      message: error.message,
+      name: error.name,
+    }
+  }
+
+  return {
+    message: JSON.stringify(error),
+    name: 'UnknownError',
+  }
+})
 
 vi.mock('mysql2/promise', () => {
   return {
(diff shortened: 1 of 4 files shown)


packages/cli/src/__tests__/Studio.vitest.ts
```diff
@@ -120,3 +129,110 @@ describe('Studio MySQL URL compatibility', () => {
...
+    expect(await response.json()).toEqual([
+      null,
+      {
+        diagnostics: [{ from: 0, message: 'lint-ok', severity: 'info', to: 1 }],
+        schemaVersion: 'v1',
+      },
+    ])
+  })
+
+  test('unwraps RPC-serialized sql-lint errors', async () => {
```
jacek-prisma, line 186: I think it would be worth to use this test to cover more than just the `@@error` case (there's multiple branches in `getSerializedBffError`, for example the nested `error`):

```suggestion
  test.each([
    {
      error: {
        '@@error': {
          message: 'relation "missing_table" does not exist',
          name: 'PostgresError',
        },
      },
      expected: {
        message: 'relation "missing_table" does not exist',
        name: 'PostgresError',
      },
    },
    // ...other cases
  ])('converts sslaccept=accept_invalid_certs to mysql2 ssl JSON', async ({ error, expected }) => {
```

jacek-prisma (approved):

PR #29328, opened 2026-03-10
fix(adapter-better-sqlite3): support positional and named parameter b…

subhaushsingh:
# fix(adapter-better-sqlite3): support positional and named parameter bindings

## Problem

Fixes #29324

When using `$queryRawTyped` or `$queryRawUnsafe` with `PrismaBetterSqlite3`, positional (`?N`) and named (`:param`) parameter bindings fail with `Too many parameter values were provided`.

The root cause is in `better-sqlite3.ts` — both `performIO` and `executeIO` always pass args as a plain array to `better-sqlite3`'s `.bind()`:

```ts
const stmt = this.client.prepare(query.sql).bind(args) // always an array
```

This only works for anonymous `?` bindings. `better-sqlite3` requires different formats depending on the binding style:
- Positional `?N` → needs a numeric-keyed object: `{ 1: "alice" }`
- Named `:param` → needs a named object: `{ username: "alice" }`

## Fix

Added a `bindArgs` helper that detects the binding style from the SQL string and transforms args into the correct format before passing to `.bind()`:

- Plain `?` → array (existing behavior, unchanged)
- Positional `?N` → numeric-keyed object
- Named `:param` → named object, with parameter names extracted from the SQL in order

## Tests

Added `better-sqlite3.test.ts` following the pattern of `adapter-libsql/src/libsql.test.ts`, including test cases that directly reproduce all three scenarios from the issue. Note: `vitest` will need to be added as a dev dependency to this package.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for the SQLite adapter covering queries, statements, multi-row scripts, positional and named parameter bindings, error propagation, and rejection of unsupported transaction isolation levels.

* **Improvements**
  * Enhanced parameter binding to support multiple SQL placeholder styles and mixed bindings.
  * More consistent error reporting for query failures and invalid isolation level requests.

### DIFF
--- a/packages/adapter-better-sqlite3/src/better-sqlite3.ts
+++ b/packages/adapter-better-sqlite3/src/better-sqlite3.ts
@@ -39,36 +39,202 @@ type BetterSQLite3Meta = {
   lastInsertRowid: number | bigint
 }
 
-function getUnquotedSql(sql: string): string {
-  return sql
-    .replace(/'[^']*'/g, '')
-    .replace(/"[^"]*"/g, '')
-    .replace(/`[^`]*`/g, '')
-    .replace(/\/\*[\s\S]*?\*\//g, '')
-    .replace(/--[^\n]*/g, '')
+type State =
+  | 'normal'
+  | 'single'
+  | 'double'
+  | 'backtick'
+  | 'bracket'
+  | 'lineComment'
+  | 'blockComment'
+
+/**
+ * Scans the SQL string and extracts parameter placeholders while ignoring
+ * strings, identifiers, and comments using a state-machine approach.
+ */
+function scanPlaceholders(sql: string): string[] {
+  const tokens: string[] = []
+
+  let i = 0
+  let state: State = 'normal'
+
+  while (i < sql.length) {
+    const char = sql[i]
+    const next = sql[i + 1]
+
+    if (state === 'normal') {
+      if (char === "'") {
+        state = 'single'
+      } else if (char === '"') {
+        state = 'double'
+      } else if (char === '`') {
+        state = 'backtick'
+      } else if (char === '[') {
+        state = 'bracket'
+      } else if (char === '-' && next === '-') {
+        state = 'lineComment'
+        i++
+      } else if (char === '/' && next === '*') {
+        state = 'blockComment'
+        i++
+      }
+
+      else if (char === '?') {
+        if (next && /\d/.test(next)) {
+          let j = i + 1
+          while (j < sql.length && /\d/.test(sql[j])) j++
+          tokens.push(sql.slice(i, j))
+          i = j - 1
+          continue
+        } else {
+          tokens.push('?')
+          i++
+          continue
+        }
+      }
+
+      else if (char === '$') {
+        if (next && /\d/.test(next)) {
+          let j = i + 1
+          while (j < sql.length && /\d/.test(sql[j])) j++
+          tokens.push(sql.slice(i, j))
+          i = j - 1
+          continue
+        } else if (next && /[a-zA-Z_]/.test(next)) {
+          let j = i + 1
+          while (j < sql.length && /[a-zA-Z0-9_]/.test(sql[j])) j++
+          tokens.push(sql.slice(i, j))
+          i = j - 1
+          continue
+        }
+      }
+      else if ((char === ':' || char === '@') && next && /[a-zA-Z_]/.test(next)) {
+        let j = i + 1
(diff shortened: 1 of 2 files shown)


josephgeis:
This works great for the positional ones like `?1`, `?2`, etc. However, it does not work with named ones in TypedSQL.

With TypedSQL enabled, try putting this into a `prisma/sql/myQuery.sql`
```sql
-- @param {String} $1:first This argument will be passed first and mapped to :first
-- @param {String} $2:second This argument will be passed second and mapped to :second
select :second as second, :first as first;
```

Then, in `index.ts`:
```ts
let res = await prisma.$queryRawTyped(
  myQuery('first', 'second')
);

res.forEach((elem) => {
  console.log(elem);
})
```

The expected should be:
```js
{ second: 'second', first: 'first' }
```

But the actual is:
```js
{ second: 'first', first: 'second' }
```

The issue really lies in the introspection of the TypedSQL queries. From the `better-sqlite3` adapter, there is no way to access the array of declared parameters in order in the TypedSQL `.sql` query files.

subhaushsingh:
@josephgeis  Thank you for the detailed explanation. I agree the positional ?N fix works correctly, but the named parameter ordering for TypedSQL is a fundamental limitation at the adapter level since the args arrive as a plain ordered array with no name metadata.

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -39,14 +39,49 @@ type BetterSQLite3Meta = {
   lastInsertRowid: number | bigint
 }
 
+function getUnquotedSql(sql: string): string {
+  return sql
+    .replace(/'[^']*'/g, '')
+    .replace(/"[^"]*"/g, '')
+    .replace(/`[^`]*`/g, '')
+    .replace(/\/\*[\s\S]*?\*\//g, '')
+    .replace(/--[^\n]*/g, '')
```
jacek-prisma: This doesn't correctly handle any escape sequences, and SQLite supports a lot of them like single quote escapes, backslash escapes, double quoted identifiers, square bracket escapes, backtick escapes etc

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -39,14 +39,49 @@ type BetterSQLite3Meta = {
...
+  if (/[:$@][a-zA-Z_][a-zA-Z0-9_]*/.test(unquoted)) {
+    const names = [...unquoted.matchAll(/[:$@]([a-zA-Z_][a-zA-Z0-9_]*)/g)].map(m => m[1])
+    const unique = [...new Set(names)]
+    const obj: Record<string, unknown> = {}
+    unique.forEach((name, i) => { obj[name] = args[i] })
+    return obj
+  }
+
+  return args
+}
```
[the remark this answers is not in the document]
jacek-prisma: this looks like a valid issue with this code

jacek-prisma (changes requested):
Left 2 important comments

packages/adapter-better-sqlite3/src/better-sqlite3.ts
```diff
@@ -39,14 +39,85 @@ type BetterSQLite3Meta = {
   lastInsertRowid: number | bigint
 }
 
+/**
+ * Strips all string literals, quoted identifiers, and comments from a SQL string.
+ * Prevents false-positive matches when detecting parameter placeholder styles.
+ */
+function getUnquotedSql(sql: string): string {
+  return sql
+    .replace(/'(?:''|\\'|[^'])*'/g, '')
```
jacek-prisma: It's not valid to just replace any occurences of things that look like string literals in arbitrary SQL strings,
For example, this is valid SQLite code: `CREATE TABLE "a'string'" ( col int )`.
In order handle all the cases correctly, I think you would need to actually tokenize the SQL string

subhaushsingh:
Hi @jacek-prisma,

I’ve addressed the previous feedback:
- Replaced regex-based SQL stripping with a state-machine tokenizer
- Added support for quoted identifiers and comments
- Fixed placeholder detection for positional (?N, $N) vs named (:name, $name, @name)
- Ensured order-safe argument binding
- Added tests for edge cases (strings, identifiers, comments, and invalid mixing)

Would appreciate another review when you get time. Thanks!

tensordreams:
Hi @subhaushsingh — thanks for this, and apologies for the slow reply.

We are closing this one: making this work requires a hand-rolled SQL tokenizer inside the adapter to infer the binding style, and review surfaced escaping cases (quoted identifiers, backticks, brackets, comments) that make that approach fragile by construction.

PR #29329, opened 2026-03-10
fix(migrate): avoid recreating indexes for Prisma 6 m-n upgrade

nehaaprasad:
### description:

- Problem: m-n upgrade migration added a new PK and dropped the unique index → long locks on big tables.
- Change: For PostgreSQL, emit PRIMARY KEY USING INDEX "..." so the existing unique index is promoted.
- Where: migrate dev and migrate diff --script --output (PostgreSQL only).
- Result: Metadata-only change, near constant time, no index rebuild.

fix : #29326

## Release Notes

* **New Features**
  * Added database-aware migration script transformations to optimize PostgreSQL primary key constraint handling.

* **Bug Fixes**
  * Improved handling of PostgreSQL primary key statements in migration scripts when using output and script options in migration diff commands.

### DIFF
--- a/packages/migrate/src/Migrate.ts
+++ b/packages/migrate/src/Migrate.ts
@@ -101,6 +101,7 @@ export class Migrate {
       extension,
       migrationName: generatedMigrationName,
       script: migrationScript,
+      connectorType,
     }).catch((e: Error) => {
       throw new Error(`Failed to write migration script to ${directoryPath}: ${e.message}`)
     })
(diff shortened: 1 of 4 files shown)


jacek-prisma (changes requested):
This should be implemented in `prisma-engines` instead of a regex here. The generated SQL is fully controlled by `prisma-engines`.

tensordreams:
Hi @nehaaprasad — thanks for this, and apologies for the slow reply.

We are closing this one: the migration SQL is generated entirely by the schema engine, so post-processing the generated DDL with string transforms is not something we can maintain — it needs correcting where the DDL is produced.

PR #29332, opened 2026-03-11
chore: add pnpm overrides to fix audit

jacek-prisma:
## Summary by CodeRabbit

* **Chores**
  * Updated dependencies including hono, @hono/node-server, and @tootallnate/once to latest stable versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,10 +162,12 @@
   "pnpm": {
     "overrides": {
       "form-data": ">=4.0.4",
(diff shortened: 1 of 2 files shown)


aqrln (approved):
lgtm, but we also need to update those deps in `@prisma/dev`, otherwise users will still get these warnings

jacek-prisma:
@aqrln 
That's correct

PR #29337, opened 2026-03-12
perf(adapter-pg): use deterministic prepared statement names

jekh:
This adds a prepared statement name when executing SQL queries with the `pg` driver. It addresses the issue identified in this discussion: https://github.com/prisma/prisma/discussions/28683

I created a repro here: https://github.com/jekh/prisma-stmt-test

It shows a dramatic difference in plan reuse and a corresponding reduction in network traffic, in line with the behavior of the legacy rust engine.

This intentionally doesn't add any sort of eviction or LRU for the client-side statement cache, because it doesn't look like the Rust adapter ever called DEALLOCATE, even when entries were evicted from the cache. Theoretically DEALLOCATE could be called when an entry is evicted, but that involves adding some kind of LRU cache and running DEALLOCATE on underlying pg, which is rather complicated. This change is basically a return to the old Rust behavior.

## Summary by CodeRabbit

* **New Features**
  * Optional statement caching for the PostgreSQL adapter to enable prepared-statement reuse and improve query performance.
  * New public option to customize statement naming and a public helper that produces stable, short prepared-statement names.

* **Tests**
  * Expanded test coverage for statement name generation, caching behavior, transaction (BEGIN/COMMIT/ROLLBACK) handling, name propagation into query execution, and stability across identical query shapes.

### DIFF
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -39,7 +39,7 @@ const debug = Debug('prisma:driver-adapter:pg')
  */
 export function generateStatementName(query: SqlQuery): string {
   const hashInput = query.sql + '\0' + JSON.stringify(query.argTypes)
-  return 'p_' + crypto.hash('sha1', hashInput, 'hex').slice(0, 24)
+  return 'p_' + crypto.hash('sha1', hashInput, 'base64url').slice(0, 16)
 }
 
 type StdClient = pg.Pool
@@ -124,7 +124,9 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
         {
           text: sql,
           values,
-          name: this.pgOptions?.enableStatementCaching ? generateStatementName(query) : undefined,
+          name: this.pgOptions?.enableStatementCaching
+            ? (this.pgOptions.getStatementName ?? generateStatementName)(query)
+            : undefined,
           rowMode: 'array',
           types: {
             // This is the error expected:
@@ -214,6 +216,16 @@ export interface PrismaPgOptions {
    * acceptable for Prisma workloads because the set of query shapes is bounded.
    */
   enableStatementCaching?: boolean
+  /**
+   * Overrides the default statement name generator used for PostgreSQL named
+   * prepared statements. The function receives a {@link SqlQuery} and must return
+   * a stable name for queries with the same SQL text and parameter types.
+   *
+   * Defaults to {@link generateStatementName} for hash-based naming.
+   *
+   * Requires {@link enableStatementCaching} to be set to `true`.
+   */
+  getStatementName?: (query: SqlQuery) => string
 }
 
 export type UserDefinedTypeParser = (oid: number, value: unknown, adapter: SqlQueryable) => Promise<unknown>
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -37,12 +37,9 @@ const debug = Debug('prisma:driver-adapter:pg')
  * Named statements live for the lifetime of the connection, which is likely acceptable for
  * Prisma workloads because the set of query shapes is typically bounded.
  */
-export function getStatementName(query: SqlQuery): string {
-  const hashInput =
-    query.argTypes.length > 0
-      ? query.sql + '\0' + query.argTypes.map((t) => `${t.scalarType}:${t.arity}`).join(',')
-      : query.sql
-  return 'p_' + crypto.hash('sha1', hashInput, 'hex').slice(0, 16)
+export function defaultStatementNameGenerator(query: SqlQuery): string {
+  const hashInput = query.sql + '\0' + JSON.stringify(query.argTypes)
+  return 'p_' + crypto.hash('sha1', hashInput, 'hex').slice(0, 24)
 }
 
 type StdClient = pg.Pool
@@ -127,7 +124,7 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
         {
           text: sql,
           values,
-          name: getStatementName(query),
+          name: this.pgOptions?.statementNameGenerator?.(query),
           rowMode: 'array',
           types: {
             // This is the error expected:
@@ -209,6 +206,15 @@ export type PrismaPgOptions = {
   onPoolError?: (err: Error) => void
   onConnectionError?: (err: Error) => void
   userDefinedTypeParser?: UserDefinedTypeParser
+  /**
+   * When provided, enables PostgreSQL named prepared statements for query plan caching.
+   * The function receives a {@link SqlQuery} and must return a stable name for queries
+   * with the same SQL text and parameter types.
+   *
+   * Use {@link defaultStatementNameGenerator} for hash-based naming, or provide a custom
+   * implementation.
+   */
+  statementNameGenerator?: (query: SqlQuery) => string
 }
 
 export type UserDefinedTypeParser = (oid: number, value: unknown, adapter: SqlQueryable) => Promise<unknown>
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -200,7 +200,7 @@ class PgTransaction extends PgQueryable<TransactionClient> implements Transactio
   }
 }
 
-export type PrismaPgOptions = {
+export interface PrismaPgOptions {
   schema?: string
   disposeExternalPool?: boolean
   onPoolError?: (err: Error) => void
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -37,7 +37,7 @@ const debug = Debug('prisma:driver-adapter:pg')
  * Named statements live for the lifetime of the connection, which is likely acceptable for
  * Prisma workloads because the set of query shapes is typically bounded.
  */
-export function defaultStatementNameGenerator(query: SqlQuery): string {
+export function generateStatementName(query: SqlQuery): string {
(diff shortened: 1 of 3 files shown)


packages/adapter-pg/src/pg.ts
```diff
@@ -25,6 +27,24 @@ const types = pg.types
...
+ * Derive a stable prepared statement name from the SQL text and argument types.
+ *
+ * Passing a `name` makes `pg` use PostgreSQL named prepared statements, so repeated
+ * query shapes can reuse server-side plans on the same connection. We key the name
+ * by SQL text and parameter types, mirroring the Rust implementation.
+ *
+ * Named statements live for the lifetime of the connection, which is likely acceptable for
+ * Prisma workloads because the set of query shapes is typically bounded.
+ */
+export function getStatementName(query: SqlQuery): string {
```
jacek-prisma: I'd avoid using a custom hash here to avoid performance and collision issues.
I think we could have a `Map<string, number>` which maps `JSON.stringify`-ied `query` to plain indices which get incremented for every new entry and then get converted into a statement name via `p_{index}`.
I think it'd also be worth make this behavior configurable, so users of the adapter can opt in or out of this behavior via `PrismaPgOptions`.
We could go even further and let users provide a custom `getStatementName?: (q: SqlQuery) => string` and have users implement it however they like.

jacek-prisma (changes requested):
Left 1 important comment

packages/adapter-pg/src/pg.ts
jekh: Thanks for reviewing. A few thoughts:
- Collisions: This is a birthday problem with 64 bits, so the collision probability with 100,000 distinct queries on a single connection is about 1 in 3,700,000. 64 bits is also what postgres's pg_stat_statements uses. But we could increase the hex length to 24 characters (96 bits) and the odds of a collision at 100,000 distinct queries would be about 1 in 1.6 * 10^19.
- Performance: The actual hash time of node's sha1 is likely going to be in the microsecond range for a few hundred or thousand bytes of SQL. We could benchmark it, but `Map<sql string, number>` is also going to have to hash the input, so it's going to be a wash in terms of hash time. A map will use more memory though.
- Hash string: I don't think we can `JSON.stringify(query)` because it contains args, but we could probably do `const hashInput = query.sql + '\0' + JSON.stringify(query.argTypes)`?
- For the `PrismaPgOptions`, how about accepting `statementNameGenerator?: (q: SqlQuery) => string` and then providing `defaultStatementNameGenerator` as a default implementation to opt in:
```ts
export function defaultStatementNameGenerator(query: SqlQuery): string {
    const hashInput = query.sql + '\0' + JSON.stringify(query.argTypes)
    return 'p_' + crypto.hash('sha1', hashInput, 'hex').slice(0, 24)
  }
```

In the 6.x rust version, prepared statements were the default, but given pg driver adapters have never had them enabled by default it probably makes sense to make them opt in.

jacek-prisma: Two comments:
- It's true that the collisions are very unlikely here but using `Map` would prevent the collision issue entirely. And also, `Map` would use a faster non-cryptographic hash function which is more suitable for this
- I think providing a `statementNameGenerator?: (q: SqlQuery) => string` option and exporting a `defaultStatementNameGenerator` is fine, but if we provide a robust default implementation, I think we can just skip that and make it a boolean toggle `enableStatementCaching?: boolean`. After some thinking, it's hard for me to imagine a valid use case for generating custom statement names given just the query contents and arguments

jacek-prisma (changes requested):
One more comment here: https://github.com/prisma/prisma/pull/29337#discussion_r2947090426

packages/adapter-pg/src/pg.ts
jekh: > It's true that the collisions are very unlikely here

That's an understatement... To put the 1 in 1.6 * 10^19 collision risk in perspective, you would need 400 million unique SQL statements on a single connection to have a 0.0001% chance of collision. Getting even 10000 unique SQL statements on a single connection would be extraordinary in a typical application.

> Map would use a faster non-cryptographic hash function which is more suitable for this

I'm not sure Map would actually perform better; node's sha-1 is native and is going to be hardware-optimized almost everywhere, and while v8 Map's hash algorithm might be faster at hashing, the overhead associated with Map (dealing with key collisions, incrementing and storing the counter) will, I suspect, outweigh any algorithmic  hash advantage.

Ultimately though, the difference is going to be on the order of microseconds. I'm happy to put together a benchmark, but microseconds either way will pale in comparison to the milliseconds it takes to get a query to the DB. We could really, really push performance to the max and use md5 as the hash algorithm, or even pull in a dependency on a blake3 or xxh3 hash implementation. If I had to guess, we're looking at no more than hundreds-of-nanoseconds difference here.

In addition to being more complicated, the Map-and-counter approach also takes more memory, since it has to store the full SQL in memory. In an application where you're dealing with 2k+-byte SQL statements, it's not huge (1000 stmts * 2k = 2 MB in the Map per connection), but it's not nothing either, especially compared to the no-memory-impact stateless sha-1 approach.

Perhaps the fatal flaw of the Map-with-counter approach is that it will result in real collisions and errors if a pg.Pool is shared among adapter-pg instances. I'm sure it's rare to run 2 or more adapter-pg instances with the same underlying pg.Pool, but it's certainly possible, and would result in node-postgres throwing as soon as the 2nd adapter-pg instance attempted to execute a query with a named statement, since they would both start their counters at 0. This is different from the v6 rust design, which didn't allow sharing connection pools (AFAIK).

So to sum it all up, using a sha-1 (or md5 if you prefer) hash, rather than a monotonically incrementing counter tracked in a Map, is:
- Simple (a stateless pure function)
- More memory efficient
- Negligible performance difference (microseconds; it really doesn't even qualify as a rounding error compared to I/O)
- Will not break if the connection pool is shared between adapter instances

>  if we provide a robust default implementation, I think we can just skip that and make it a boolean toggle enableStatementCaching?: boolean

👍

jacek-prisma: I think using SHA-1 as the default is problematic, especially considering that the resulted hash is truncated by 40% in order to get the statement name, which degrades a lot of its properties. Also, SHA-1 as a cryptographic is much slower than some alternatives, for example the `xxhash` benchmark reports it having 40 times higher throughput than SHA-1. Additionally, SHA-1 is known to be fragile to many collision attacks.

I've looked at some other hashing solutions and they all have some some drawbacks. I think the best way to handle this is to expose a `getStatementName?: (q: SqlQuery) => string` after all and export one or more implementations.

I think ideally we would have a very fast non-cryptographic hash implementation that doesn't rely on hash truncation (`xxhash` made available as an NPM peer dependency looks like a good candidate).

I don't mind having the SHA1 algorithm as another exported implementation as long as we document the pros (requires no dependencies) and cons (slower than a non-cryptographic hash, known to be fragile to some collision attacks, the implementation truncates 40% of the hash). I think the current implementation might also be improved by using `base64` instead of `hex` to reduce the impact of truncation on the resulting hash.

jekh: I think maybe there's some misunderstanding about the purpose of SHA-1 here. Its purpose is to generate a unique string, quickly. We're not verifying digital signatures, we're generating unique strings. This is not a cryptographic application of SHA-1.

> SHA-1 as the default is problematic, especially considering that the resulted hash is truncated by 40% in order to get the statement name, which degrades a lot of its properties. 

The property we care about is its ability to generate deterministic, unique strings. 96 bits provides a 1 in 1.6 * 10^19 collision risk after 100,000 unique SQL queries, which is far, far more than a single connection could ever have. This is not degraded, in fact it's overkill.

> Also, SHA-1 as a cryptographic is much slower than some alternatives, for example the xxhash benchmark reports it having 40 times higher throughput than SHA-1.

Scale is important here. While this is true, we're talking about microseconds... at the most, probably sub-microsecond for typical queries. In I/O context where text is being put on the wire, this does not even amount to noise.

> I think ideally we would have a very fast non-cryptographic hash implementation

As soon as node adds a native non-cryptographic hash, I would be happy to submit another PR to switch. But I'll just point out the contradiction of simultaneously asserting the downside of being "fragile to some collision attacks", while advocating for a non-cryptographic hash implementation.

>  I think the current implementation might also be improved by using base64 instead of hex to reduce the impact of truncation on the resulting hash.

This is a good idea. Not to reduce the truncation of the hash, but because there's no harm doing base64url and we can go back to 16 characters while maintaining the 96-bits of collision resistance, so we might as well.

These updates are now in, both an enabled flag (enabling this default implementation), and an override.

jacek-prisma: > This is not a cryptographic application of SHA-1.

I'm not arguing for a cryptographic hash here

> This is not degraded, in fact it's overkill.

By degraded I just mean it's not an efficient use of SHA-1, it generates a full hash and gets a fraction of its intended properties

> In I/O context where text is being put on the wire, this does not even amount to noise.

I haven't benchmarked it, so it's difficult to say the effect, but even if that were true, not all database interactions involve I/O, for example dealing some SQlite adapters

> As soon as node adds a native non-cryptographic hash

We can add a peer-dependency-based implementation like I suggested earlier

> But I'll just point out the contradiction of simultaneously asserting the downside of being "fragile to some collision attacks", while advocating for a non-cryptographic hash implementation.

Collision attacks aren't a problem exclusive to cryptographic hashes, most popular modern non-cryptographic hashes are designed to be resistant to hash flooding etc

jacek-prisma:
Implemented a subset of this in https://github.com/prisma/prisma/pull/29395

tensordreams:
Hi @jekh — thanks for this, and apologies for the slow reply.

We are closing this one: the prepared-statement name generator was implemented and merged in #29395, so this is superseded by the landed change.

PR #29340, opened 2026-03-13
WIP: fix(client): avoid excessive stack depth in extension contexts

### LINKED ISSUE #27059: Excessive stack depth comparing types
### Bug description

I'm getting an excessive stack depth when I try to extend prisma.

### Goal
I have a few temporal objects that have a currentVersion property. By default I want all queries to have
```ts
where: {
  currentVersion: true,
}
```

But there are cases where I need to query against all versions so I need a way to override that default.

### Solution
My idea was to extend prisma to automatically apply the where clause to all queries. I would then add new methods that would let me query all versions if needed. So something like this:

```ts
await prisma.employeeBenefit.findMany({
  where: {
    employeeId: 1,
  },
});
```
This would automatically add a `currentVersion: true` where clause.

If I wanted to query all I would do
```ts
await prisma.employeeBenefit.findManyAllVersions({
  where: {
    employeeId: 1,
  },
});
```
This wouldn't include the currentVersion where clause.

### Extension
This is what my extension looks like

```ts
new PrismaClient({
    log: ['query'],
  }).$extends({
    model: {
      employeeBenefit: {
        async findManyAllVersions(args?: Prisma.EmployeeBenefitFindManyArgs) {
          const normalizedArgs = {
            ...(args || {}),
          } as Prisma.EmployeeBenefitFindManyArgs & {
            allVersions?: boolean;
          };

          normalizedArgs.allVersions = true;

          const context = Prisma.getExtensionContext(this);
          return context.findMany(normalizedArgs);
        },
        foo() {
          return this;
        },
      },
    },
    query: {
      employeeBenefit: {
        async findMany({ args, query }) {
          const normalizedArgs = {
            ...(args || {}),
          } as Prisma.EmployeeBenefitFindManyArgs & {
            allVersions?: boolean;
          };

          // Filter by currentVersion unless allVersions is true
          if (normalizedArgs.allVersions !== true) {
            normalizedArgs.where = {
              AND: [
                normalizedArgs.where || {},
                {
                  currentVersion: true,
                },
              ],
            };
          }

          // Remove our custom property
          if ('allVersions' in normalizedArgs) {
            delete normalizedArgs.allVersions;
          }

          return query(normalizedArgs);
        },
      },
    },
  });
```

### Error
I get this error
```
src/lib/prisma.ts:23:18 - error TS2321: Excessive stack depth comparing types 'EmployeeBenefitSelect<DefaultArgs>' and 'EmployeeBenefitSelect<InternalArgs & DefaultArgs>'.

23           return context.findMany(normalizedArgs);
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in src/lib/prisma.ts:23
```

Note: This works fine with a simpler database schema. If I just have one user model I don't get the max depth error. 

Maybe an extension isn't the right approach? I have tried creating a repository for the model and proxying the methods but got the same issue. 

### Severity

### Reproduction

I have created a repo with the schema here.

https://github.com/dmaksimov/prisma-max-depth

Expected inferred types to work fine but I get a maximum depth exceeded error

### Frequency

Consistently reproducible

### Is this a regression?

No

### Workaround

I have not found any workarounds yet. I have tried to create a repository for the model with proxy methods for findMany and findManyAllVersions but got the same error.

### Prisma Schema & Queries

All included above

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** macOS
- **Database:** PostgreSQL
- **Node.js version:** v20.11.0

### Prisma Version

```
// Prisma version output
prisma                  : 6.7.0
@prisma/client          : 6.7.0
Node.js                 : v20.11.0
TypeScript              : 5.8.3
Query Engine (Node-API) : libquery-engine 3cff47a7f5d65c3ea74883f1d736e41d68ce91ed (at node_modules/@prisma/engines/libquery_engine-darwin-arm64.dylib.node)
Schema Engine           : schema-engine-cli 3cff47a7f5d65c3ea74883f1d736e41d68ce91ed (at node_modules/@prisma/engines/schema-engine-darwin-arm64)
Schema Wasm             : @prisma/prisma-schema-wasm 6.7.0-36.3cff47a7f5d65c3ea74883f1d736e41d68ce91ed
Default Engines Hash    : 3cff47a7f5d65c3ea74883f1d736e41d68ce91ed
```

sorenbs:
## Summary
- add an e2e regression fixture for #27059 based on the public reproduction
- keep the Prisma 7.x `TS2321` failure documented in the fixture
- avoid `Exact`-based deep argument comparison for model operations exposed through extension contexts

## Testing
- manually reproduced the failure with `node packages/cli/build/index.js generate --config packages/client/tests/e2e/issues/27059-excessive-stack-depth-get-extension-context/prisma.config.ts`
- manually verified the fix with `./node_modules/.bin/tsc --noEmit -p packages/client/tests/e2e/issues/27059-excessive-stack-depth-get-extension-context/tsconfig.json`
- manually sanity-checked that `context.findMany({ badInput: true })` still type-errors in the fixture

Fixes #27059.


dbrxnds:
How can I help move this along?

tensordreams (commented):
I investigated whether dropping `Exact` here is the right call, verified the fix and its trade-offs empirically (generated the fixture's 41-model client and A/B-flipped `Exact` in the built runtime types), and pushed two changes to this branch: a rebase onto current `main` (the branch was conflicting) and a fix for the e2e fixture that was failing the Client E2E job.

## Verification

- With `Exact` (before this PR): `src/index.ts:20` of the fixture fails with `TS2321: Excessive stack depth comparing types 'EmployeeBenefitSelect<DefaultArgs>' and 'EmployeeBenefitSelect<InternalArgs & DefaultArgs>'` — the exact error from #27059.
- With this PR: clean. Wall-clock is ~3 s in both states at this schema size, so what trips is the instantiation-depth limiter, not a timeout.

## Scope note

This signature types every model operation on **any `$extends`-ed client** (via `DynamicClientExtensionThis` → `DynamicModelExtensionThis`), not just `Prisma.getExtensionContext(this)`. Anyone using e.g. the Accelerate extension is on this code path. Worth reflecting in the PR description and, ideally, a release-notes callout.

## What changes in practice

Verified against the generated fixture client (extended client):

| args passed to `xprisma.model.op(...)` | with `Exact` | with this PR |
|---|---|---|
| inline literal, only unknown keys | error | error |
| inline literal, unknown key next to a valid one | error | **accepted** |
| inline literal, nested typo (`where: { employeeId, typoField }`) | error | **accepted** |
| variable with extra keys (the #27059 forwarding pattern) | accepted | accepted |
| wrong value type | error | error |
| `select`/`include` narrowing, `orderBy`/`groupBy` inference, checked/unchecked create XOR | works | works, unchanged |

The base (non-extended) client is unaffected: its generated `SelectSubset` signatures are stricter than the extension path ever was, in both states. Unknown keys that survive to the engine still fail at runtime with a validation error (`UnknownArgument`). No test in the repo asserts the two "accepted" rows, which is why nothing else goes red.

Mechanism of the loss: with a bare `<A extends Args>(args: A)`, TS infers `A` as the literal's own type; if that candidate satisfies the constraint (any object sharing one valid key does), excess property checking never engages. It only kicks in when the candidate *fails* the constraint (unknown-keys-only case), where inference falls back to the constraint itself.

## Options considered, and why this is the least bad

1. **Keep `Exact` (status quo)** — stays broken. `Exact` recurses per key of the inferred type and distributes over unions on both sides, and cross-instantiation comparisons like `XSelect<DefaultArgs>` vs `XSelect<InternalArgs & DefaultArgs>` have unmeasurable variance (conditional-type-bearing aliases), so every compare is structural all the way down. It is also a per-call type-instantiation tax on every extended-client operation.
2. **This PR: plain `<A extends Args>(args: A)`** — the surviving constraint check performs the same cross-instantiation comparison once, plainly, with relation caching; verified it does not trip the limiter. Cost: the two "accepted" rows above.
3. **`SelectSubset`-style top-level exactness (align with the base client)** — cheap (maps unknown top-level keys to `never`, no deep recursion into the wanted type), and would actually catch *more* than `Exact` did (it rejects extra keys on non-fresh objects too). But that is exactly why it cannot be used here: it would reject the forwarding pattern from #27059 itself (`Prisma.EmployeeBenefitFindManyArgs & { allVersions?: boolean }` passed through the context so a `query` extension can strip the custom key). It would turn the reported repro from a compiler error into a type error rather than fixing it.
4. **Fix the root cause in the generated types** — restructure the `*Select`/`GetSelect` machinery so instantiations of the same generic relate via measurable variance instead of deep structure. The principled long-term fix, but a large restructuring with its own perf risks, and it still leaves `Exact`'s per-node instantiation cost in place. Worth a follow-up issue, not this PR.
5. **Depth-bounded exactness** (exact for the first level, plain below) — arbitrary semantics (typos caught at depth 1 but not depth 2), still costly, more complexity in an already subtle file.

TypeScript cannot distinguish a fresh inline literal from a forwarded variable at the type level — freshness exists only inside the checker (excess property checking). So no signature can simultaneously accept the forwarding pattern and reject inline typos. Given that, dropping `Exact` is the only option that fixes the bug, preserves the issue's use-case, keeps result-type inference intact, and leaves the base client untouched.

## Pushed to this branch

- Rebase onto current `main` (conflict in `AGENTS.md` resolved by keeping both sides; `Extensions.ts` merged clean).
- `test(client)`: added the missing `@prisma/client` tarball dependency to the fixture. The generated client imports `@prisma/client/runtime/client`, and in the e2e container each test installs as a standalone package root, so the import could not resolve — that is the failing **Client E2E** job. (It passes when `tsc` runs from the repo root, where resolution walks up into the workspace — which is how the fix was originally verified.)
- Verified after the rebase: the fixture generates and typechecks clean against the rebuilt runtime, and the isolated Docker e2e run now passes (`pnpm --filter @prisma/client test:e2e 27059-excessive-stack-depth-get-extension-context` → All 1/1 tests passed).

## Remaining suggestions (non-blocking)

- `Exact` still guards the fluent API (`DynamicModelExtensionFluentApi`) and `Prisma.validator`, so the same TS2321 class remains reachable there on large schemas. Worth a follow-up issue.
- Drop the WIP from the title, and state the full scope and the intentional loosening in the PR description so release notes can mention it.

PR #29357, opened 2026-03-19
fix(adapter-pg): update @types/pg

### LINKED ISSUE #29356: Update @types/pg dependency in @prisma/adapter-pg to support pg@8.18+ type changes
### Bug description

`@prisma/adapter-pg` currently depends on `@types/pg@8.11.11`, which is incompatible with projects using `@types/pg@8.18.0` or later. This causes TypeScript compilation errors when passing a `Pool` instance to the `PrismaPg` constructor.

## Error Message
Type error: Argument of type 'Pool' is not assignable to parameter of type 'Pool | PoolConfig'. Type 'import("node_modules/@types/pg/index").Pool' is not assignable to type 'import("node_modules/@prisma/adapter-pg/node_modules/@types/pg/index").Pool'. The types of 'options.Client' are incompatible between these types. [...] Construct signature return types 'ClientBase' and 'ClientBase' are incompatible. The types returned by 'connect()' are incompatible between these types. Type 'Promise' is not assignable to type 'Promise'.

## Root Cause
In **pg@8.18.0** (released Jan 27, 2026), the `Client.connect()` method was changed to return the client instance instead of `void` ([brianc/node-postgres#3564](https://github.com/brianc/node-postgres/pull/3564)).
**@types/pg@8.18.0** (released Feb 28, 2026) updated the type definitions accordingly ([DefinitelyTyped/DefinitelyTyped#74427](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74427)):
- **Old:** `connect(): Promise<void>`
- **New:** `connect(): Promise<ClientBase>` (or `Promise<Client>` for Client class)
Since `@prisma/adapter-pg` pins `@types/pg@8.11.11` as a direct dependency (added in #29277), there's now a type mismatch when consumers have `@types/pg@8.18.0+` in their dependency tree.

## Proposed Solution
Update @prisma/adapter-pg's dependency on @types/pg to accept ^8.18.0:

```json
{
  "dependencies": {
    "@types/pg": "^8.18.0"
  }
}
```

This change is backwards compatible since:

- The runtime behavior hasn't changed (pg@8.18.0 is backwards compatible)
- The type change only affects the return value of connect(), which adapter-pg doesn't appear to use directly
- Consumers using older @types/pg versions will continue to work via TypeScript's structural typing

## Environment
- **@prisma/adapter-pg**: 7.5.0 (also affects 7.3.0+)
- **@types/pg** in adapter-pg: 8.11.11
- **@types/pg** in consumer project: 8.18.0
- **pg**: 8.13.1 (but affects any version when paired with @types/pg@8.18.0+)

### Severity

### Reproduction

**package.json:**
```json
{
  "dependencies": {
    "@prisma/adapter-pg": "^7.3.0",
    "@prisma/client": "^7.0.0",
    "pg": "^8.13.1"
  },
  "devDependencies": {
    "@types/pg": "^8.11.10"
  }
}
```

After npm update, @types/pg upgrades to 8.18.0, causing build failures.

Code:

```ts
import { Pool } from 'pg';
const pool = new Pool({ connectionString: process.env.DATABASE_URL });
const adapter = new PrismaPg(pool); // TypeScript error here
```

Build should pass. It does not.

### Frequency

Consistently reproducible

### Is this a regression?

Not a Prisma regression, but an upstream types change

### Workaround

Using resolutions or pinning an old version of @types/pg

### Prisma Schema & Queries

NA

### Prisma Config

_No response_

### Logs & Debug Info

_No response_

### Environment & Setup

- **OS:** Windows
- **Database:** PostgreSQL
- **Node.js version:** 24

### Prisma Version

```
Prisma schema loaded from prisma\schema.prisma.
prisma               : 7.5.0
@prisma/client       : 7.5.0
Operating System     : win32
Architecture         : x64
Node.js              : v24.11.1
TypeScript           : 5.9.3
PSL                  : @prisma/prisma-schema-wasm 7.5.0-15.280c870be64f457428992c43c1f6d557fab6e29e
Schema Engine        : schema-engine-cli 280c870be64f457428992c43c1f6d557fab6e29e (at node_modules\@prisma\engines\schema-engine-windows.exe)
Default Engines Hash : 280c870be64f457428992c43c1f6d557fab6e29e
Studio               : 0.21.1
```

dcaffull:
Supports pg@8.18+ type changes

Closes #29356

## Summary by CodeRabbit

* **Chores**
  * Updated a PostgreSQL-related dependency to a newer semver range for improved compatibility and stability.
* **Style**
  * Minor internal code cleanup with no functional changes to runtime behavior.

### DIFF
--- a/packages/adapter-pg/package.json
+++ b/packages/adapter-pg/package.json
@@ -39,7 +39,7 @@
     "@prisma/driver-adapter-utils": "workspace:*",
     "pg": "^8.16.3",
     "postgres-array": "3.0.4",
-    "@types/pg": "8.11.11"
+    "@types/pg": "^8.18.0"
   },
   "devDependencies": {
     "@prisma/debug": "workspace:*"
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -109,18 +109,6 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
           values,
           rowMode: 'array',
           types: {
-            // This is the error expected:
-            // No overload matches this call.
-            // The last overload gave the following error.
-            // Type '(oid: number, format?: any) => (json: string) => unknown' is not assignable to type '{ <T>(oid: number): TypeParser<string, string | T>; <T>(oid: number, format: "text"): TypeParser<string, string | T>; <T>(oid: number, format: "binary"): TypeParser<...>; }'.
-            //   Type '(json: string) => unknown' is not assignable to type 'TypeParser<Buffer, any>'.
-            //     Types of parameters 'json' and 'value' are incompatible.
-            //       Type 'Buffer' is not assignable to type 'string'.ts(2769)
-            //
-            // Because pg-types types expect us to handle both binary and text protocol versions,
-            // where as far we can see, pg will ever pass only text version.
-            //
-            // @ts-expect-error
             getTypeParser: (oid: number, format?) => {
               if (format === 'text' && customParsers[oid]) {
                 return customParsers[oid]
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -306,8 +306,8 @@ importers:
         specifier: workspace:*
         version: link:../driver-adapter-utils
       '@types/pg':
-        specifier: 8.11.11
-        version: 8.11.11
+        specifier: ^8.18.0
+        version: 8.18.0
       pg:
         specifier: ^8.16.3
         version: 8.16.3
@@ -4008,6 +4008,9 @@ packages:
 
   '@types/pg@8.11.6':
     resolution: {integrity: sha512-/2WmmBXHLsfRqzfHW7BNZ8SbYzE8OSk7i3WjFYvfgRHj7S1xj+16Je5fUKv3lVdVzk/zn9TXOqf+avFCFIE0yQ==}
+
+  '@types/pg@8.18.0':
+    resolution: {integrity: sha512-gT+oueVQkqnj6ajGJXblFR4iavIXWsGAFCk3dP4Kki5+a9R4NMt0JARdk6s8cUKcfUoqP5dAtDSLU8xYUTFV+Q==}
 
   '@types/pluralize@0.0.33':
     resolution: {integrity: sha512-JOqsl+ZoCpP4e8TDke9W79FDcSgPAR0l6pixx2JHkhnRjvShyYiAYw2LVsnA7K08Y6DeOnaU6ujmENO4os/cYg==}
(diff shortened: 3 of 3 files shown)


themavik (commented):
pnpm lock pulls `@types/pg@8.18` with `pg-types` 2.x in that snapshot while other importer paths still reference 4.x-style stacks—if TS starts disagreeing on `QueryResult` / field shapes across packages, align the type tree. nit: run adapter-pg typecheck against the new `@types/pg` minor.

jacek-prisma (approved):

jacek-prisma (changes requested):
Looks like the type update makes the `@ts-expect-error` in `packages/adapter-pg/src/pg.ts` no longer valid (which broke the CI)

dcaffull:
@jacek-prisma thanks. Looks like the upgrade fixed the expected issue. I've removed that directive now.

packages/adapter-pg/package.json
```diff
@@ -39,7 +39,7 @@
     "@prisma/driver-adapter-utils": "workspace:*",
     "pg": "^8.16.3",
     "postgres-array": "3.0.4",
-    "@types/pg": "8.11.11"
+    "@types/pg": "^8.18.0"
```
jacek-prisma, line 42: Why not just set both `pg` and `@types/pg` to `^8.16.3`? That should allow you to use `8.18.0` of both anyway

dcaffull: The type definitions changed in 8.18.0. Setting it lower than that potentially causes transitive dependency mismatches for applications that upgrade pg (the issue outlined in #29356).

jacek-prisma: From my manual tests, setting it to `^8.16.0` appears to work fine for applications that upgraded to newer pg (like `8.18.0`). Setting it to `^8.18.0` however causes a problem for consumers that use `@types/pg@8.16.0` (aligned with our current `pg`) or `@types/pg@8.11.11` (aligned with the `@pg/types` added in Prisma 7.5.0).

I've run a quick script to find which `npm i && tsc` combinations produce an error and generated a table:
| Prisma @types/pg | App @types/pg | Result | Description |
|------------------|---------------|--------|-------------|
| `^8.16.0` | `8.18.0` | ✅ | `^8.16` types in Prisma vs recent pg |
| `^8.16.0` | `8.16.0` | ✅ | `^8.16` types in Prisma vs pg aligned with prisma 7.5 pg dep |
| `^8.16.0` | `8.11.11` | ❌ | `^8.16` types in Prisma vs pg aligned with prisma 7.5 `@pg/types` dep |
| `^8.18.0` | `8.18.0` | ✅ | `^8.18` types in Prisma vs recent pg |
| `^8.18.0` | `8.16.0` | ❌ | `^8.18` types in Prisma vs pg aligned with prisma 7.5 pg dep |
| `^8.18.0` | `8.11.11` | ❌ | `^8.18` types in Prisma vs pg aligned with prisma 7.5 `@pg/types` dep |

I think the `^8.16.0` constraint is better since it doesn't break the case when the user has `8.16`.

jacek-prisma:
Superseded by https://github.com/prisma/prisma/pull/29390

PR #29370, opened 2026-03-23
fix(adapter-mariadb): fix prepared statement leak

### LINKED ISSUE #29364: adapter-mariadb 7.5.0: prepared statement leak after switch to binary protocol
## Bug description

  After upgrading `@prisma/adapter-mariadb` from 7.3.0 to 7.5.0, prepared statements accumulate
  unboundedly in MySQL/MariaDB, eventually hitting `max_prepared_stmt_count`.

  7.5.0 switched the adapter from text protocol (`.query()`) to binary protocol
  (`.execute()` / `COM_STMT_PREPARE`) in PR #29285 to fix lossy number conversions.
  This means every query now creates a server-side prepared statement.

  The mariadb Node.js driver caches prepared statements per connection
  (`prepareCacheLength` defaults to 256). Because Prisma generates many distinct SQL strings
  (e.g. varying `IN (?)` clause sizes), the cache fills with unique entries.
  Evicted statements are not reliably closed via `COM_STMT_CLOSE` on the server,
  causing `prepared_statements_instances` to grow indefinitely.

  ## Evidence from production

  - `max_prepared_stmt_count = 16382` (hit the limit)
  - `Com_stmt_prepare = 40,802,294` vs `Com_stmt_close = 3,279,759` (massive gap)
  - `performance_schema.prepared_statements_instances` shows the same Prisma-generated
    query repeated 300+ times
  - Queries confirmed to be Prisma-style SQL (backtick-quoted identifiers, `AND 1=1` pattern)
  - `performance_schema.threads` confirms all statements come from Node.js connections
  - **Did not happen on 7.3.0** — only after upgrading to 7.5.0

  ## How to reproduce

  1. Use `@prisma/adapter-mariadb` 7.5.0 with the `mariadb` Node.js driver
  2. Run a workload with many Prisma queries (especially varying `IN` clause sizes)
  3. Monitor: `SHOW GLOBAL STATUS LIKE 'Com_stmt%'` and
     `SELECT COUNT(*) FROM performance_schema.prepared_statements_instances`
  4. Observe `Prepared_stmt_count` growing continuously until it hits `max_prepared_stmt_count`

  ## Expected behavior

  Prepared statements should be closed after use (`COM_STMT_CLOSE`), or the adapter
  should continue using text protocol (`.query()` / `COM_QUERY`) which worked correctly in 7.3.0.

  ## Workarounds

  - **Pin `@prisma/adapter-mariadb` to 7.3.0** (uses text protocol, no prepared statements)
  - **Set `prepareCacheLength: 0`** on the mariadb pool config to disable caching

  ## Environment

  - `@prisma/adapter-mariadb`: 7.5.0
  - `@prisma/client`: 7.5.0
  - `mariadb` (Node.js driver): 3.5.2
  - Database: MariaDB (RDS)

  ## Related

  - PR #29285 (switched adapter to binary protocol)
  - #6872 (historical prepared statement ballooning)
  - #18975 (prepared statement leak with OpenTelemetry)

Goldyvaiiii:
Fixes: #29364

Description

This PR fixes a prepared statement leak in @prisma/adapter-mariadb that could cause the database to hit the max_prepared_stmt_count limit over time.

Problem

In version 7.5.0, the adapter switched to using the binary protocol via client.execute(). The MariaDB Node.js driver internally caches prepared statements, but these statements are not always reliably closed on the server when evicted from the cache.

Since Prisma frequently generates dynamic SQL queries (for example, queries with varying IN clause sizes), this resulted in a large number of unique prepared statements accumulating on the server, eventually leading to resource exhaustion.

Solution

This PR reverts query execution back to the text protocol using client.query() inside MariaDbQueryable.performIO.

Avoids creating server-side prepared statements
Prevents unbounded growth of prepared statements
Aligns behavior with existing transaction handling in the adapter
Impact
Fixes server-side prepared statement leaks
Prevents hitting max_prepared_stmt_count
Improves stability for applications with dynamic query patterns
Checklist
 Bug fix (non-breaking change)
 Tested with MariaDB/MySQL
 No regression in existing functionality

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents accumulation of server-side prepared statements in the MariaDB adapter, reducing connection/server resource leaks.
  * Improves reliability of transaction operations (BEGIN/COMMIT/ROLLBACK/SAVEPOINT).
  * Preserves large-number precision during queries to avoid lossy numeric conversions.

### DIFF
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -315,6 +315,12 @@ function withNoPrepareCache(config: mariadb.PoolConfig | string): mariadb.PoolCo
     return { ...config, prepareCacheLength: 0 }
   }
 
-  const separator = config.includes('?') ? '&' : '?'
-  return `${config}${separator}prepareCacheLength=0`
+  try {
+    const url = new URL(config)
+    url.searchParams.set('prepareCacheLength', '0')
+    return url.toString()
+  } catch {
+    const separator = config.includes('?') ? '&' : '?'
+    return `${config}${separator}prepareCacheLength=0`
+  }
 }
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -63,8 +63,10 @@ class MariaDbQueryable<Connection extends mariadb.Pool | mariadb.Connection> imp
         typeCast,
       }
       const values = args.map((arg, i) => mapArg(arg, query.argTypes[i]))
-      // We intentionally use `execute` here, because it uses the binary protocol, unlike `query`.
-      return await this.client.execute(req, values)
+      // We use `query` here instead of `execute` because `execute` uses the binary protocol, which
+      // creates server-side prepared statements that are cached by the driver but not always closed,
+      // leading to a leak in Prisma because Prisma generates many distinct SQL strings.
+      return await this.client.query(req, values)
     } catch (e) {
       const error = e as Error
       this.onError(error)
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -22,7 +22,7 @@ class MariaDbQueryable<Connection extends mariadb.Pool | mariadb.Connection> imp
   readonly provider = 'mysql'
   readonly adapterName = packageName
 
-  constructor(protected client: Connection) {}
+  constructor(protected client: Connection) { }
 
   async queryRaw(query: SqlQuery): Promise<SqlResultSet> {
     const tag = '[js::query_raw]'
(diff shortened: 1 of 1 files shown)


Goldyvaiiii:
Hi maintainers 👋
This PR fixes a prepared statement leak in the MariaDB adapter (#29364).

Would appreciate a review when you get time. Thanks!

jacek-prisma (changes requested):
Hi @Goldyvaiiii 
The adapter was changed to use the binary protocol in order fix some bugs present when passing large numbers via the text protocol.
Which means we cannot just switch back to the text protocol without re-introducing these bugs and therefore breaking tests.

packages/adapter-mariadb/src/mariadb.ts
```diff
@@ -219,7 +222,14 @@ export class PrismaMariaDbAdapterFactory implements SqlDriverAdapterFactory {
   async connect(): Promise<PrismaMariaDbAdapter> {
     let pool: mariadb.Pool
     try {
-      pool = mariadb.createPool(this.#config)
+      // We set `prepareCacheLength: 0` to prevent a server-side prepared statement leak.
+      // When using `execute()` (binary protocol), the driver caches prepared statements per
+      // connection. Since Prisma generates many distinct SQL strings (e.g. varying IN clause sizes),
+      // the cache fills and evicted statements are not reliably closed via COM_STMT_CLOSE.
+      // Setting `prepareCacheLength: 0` disables this cache, ensuring every statement is closed
+      // after use. Binary protocol is still used to avoid lossy conversions for large numbers (#29285).
```
jacek-prisma, line 230: The implementation is inconsistent with this comment, it uses an unsupported `prepareCacheLength` URL parameter rather than simply passing it as a config field

jacek-prisma:
Superseded by https://github.com/prisma/prisma/pull/29392

PR #29371, opened 2026-03-23
Fix/cli no models msg

Goldyvaiiii:
Description

This PR improves the developer experience when running prisma generate on a schema with no models defined.

Changes
Introduced a new --allow-no-models flag for the prisma generate command
Updated the error message when no models are present but required
Made the message clearer, more concise, and actionable by suggesting the new flag
Problem

Previously, running prisma generate on a schema without models resulted in a confusing or unhelpful error message, leaving users unsure how to proceed.

Solution
Added an explicit --allow-no-models flag to support valid use cases where no models are defined
Improved the error messaging to clearly guide users toward the correct action
Impact
Better developer experience and clearer guidance
Supports edge cases where schemas intentionally have no models
Reduces confusion for new users
Checklist
 Feature / UX improvement
 Backward compatible
 No breaking changes

## Summary by CodeRabbit

* **New Features**
  * Added `--allow-no-models` CLI flag to support running `prisma generate` without any models defined.

* **Documentation**
  * Updated guidance to recommend using `--allow-no-models` for projects using raw queries only.

### DIFF
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -63,8 +63,10 @@ class MariaDbQueryable<Connection extends mariadb.Pool | mariadb.Connection> imp
         typeCast,
       }
       const values = args.map((arg, i) => mapArg(arg, query.argTypes[i]))
-      // We intentionally use `execute` here, because it uses the binary protocol, unlike `query`.
-      return await this.client.execute(req, values)
+      // We use `query` here instead of `execute` because `execute` uses the binary protocol, which
+      // creates server-side prepared statements that are cached by the driver but not always closed,
+      // leading to a leak in Prisma because Prisma generates many distinct SQL strings.
+      return await this.client.query(req, values)
     } catch (e) {
       const error = e as Error
       this.onError(error)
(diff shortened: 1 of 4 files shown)


themavik (commented):
MariaDB: switching client.execute to query drops the binary protocol—worth a changelog note if anyone tuned for that path. CLI: missingModelMessage now only mentions --allow-no-models, but prisma generate without flags still allows empty schemas when --require-models is off; nit: mention both so raw-SQL users are not nudged into a flag they do not need.

themavik (commented):
MariaDB: swapping `client.execute` for `query` drops the binary protocol—worth a release note if anyone tuned for that path. CLI: `missingModelMessage` pushes `--allow-no-models`, but bare `prisma generate` still works when `--require-models` is off; nit: mention both so raw-SQL users are not pushed into a flag they do not need.

jacek-prisma (commented):
What's the DX case this improves? No models are already allowed by default, so I don't see what case this is aiming to improve exactly

aqrln:
This PR has multiple unrelated things mixed together, including a feature with unclear purpose, previous unanswered and unresolved comments, and a failed CLA signature (GitHub account doesn't match the commit author). I'll go ahead and close it.

Feel free to open more targeted PRs for the specific fixes you wanted to make, open issues or discuss your problem with us on Discord first!

PR #29373, opened 2026-03-23
fix(internals): preserve original line endings and ensure single trailing newline in formatSchema

cyphercodes:
This PR addresses issue #8548 by ensuring that `prisma format` preserves the original line ending style (CRLF or LF) from the input schema. It also ensures the output always ends with a single trailing newline of the correct type, fixing the issue where an extra CRLF was added on Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve each schema's original line-ending style (CRLF or LF) when formatting and ensure exactly one trailing newline of that style to avoid cross-platform inconsistencies.
  * Leave formatted content unchanged for files not present in the original input.
  * Maintain existing linting, diagnostics, and error-handling behavior so validation and reporting remain unchanged.


jacek-prisma (changes requested):
The proper place to implement this change is `prisma-engines` which actually writes the formatted files rather than here

SevInf:

PR #29374, opened 2026-03-23
feat(cli): add prisma bootstrap command

kristof-siket:
## Purpose of change

Adds `prisma bootstrap` as a root-level CLI command that orchestrates the full Prisma Postgres activation flow, enabling users to go from zero to a running, connected Prisma Postgres app in a single command.

Closes PTL-1279, PTL-1280, PTL-1281, PTL-1284, PTL-1286, PTL-1287

## Command flow

```mermaid
flowchart TD
    Start([prisma bootstrap]) --> CheckPkg{package.json<br/>exists?}

    CheckPkg -->|No| AskTemplate{Offer template?}
    AskTemplate -->|Yes| Scaffold[Download & scaffold template]
    AskTemplate -->|No| EmptyGuard[Show instructions:<br/>npm init / pnpm init / bun init]
    EmptyGuard --> Exit1([Exit — user must init first])

    CheckPkg -->|Yes| DetectState[Detect project state<br/>schema, config, models, seed]

    Scaffold --> InstallDeps[Install dependencies]
    InstallDeps --> Link

    DetectState --> HasSchema{schema.prisma<br/>exists?}
    HasSchema -->|No| AskModel{Add sample User model?}
    AskModel -->|Yes| RunInit[prisma init + User model]
    AskModel -->|No| RunInitPlain[prisma init]
    RunInit --> CheckDeps
    RunInitPlain --> CheckDeps
    HasSchema -->|Yes| CheckDeps

    CheckDeps{dotenv & prisma<br/>installed?}
    CheckDeps -->|No| DepsGate[Show: install deps, then re-run]
    DepsGate --> Link
    CheckDeps -->|Yes| Link

    Link{Already linked?}
    Link -->|Yes, no --force| SkipLink[Skip link]
    Link -->|No, or --force| RunLink[Link to Prisma Postgres<br/>--api-key or browser auth]
    RunLink --> SetEnv[Write DATABASE_URL to .env]
    SetEnv --> Migrate

    SkipLink --> Migrate

    Migrate{Models in schema?}
    Migrate -->|Yes| AskMigrate{Apply with<br/>prisma migrate dev?}
    AskMigrate -->|Yes| RunMigrate[Run prisma migrate dev]
    AskMigrate -->|No| Generate
    Migrate -->|No| Generate

    RunMigrate --> Generate

    Generate[Run prisma generate]
    Generate --> HasSeed{Seed script<br/>defined?}

    HasSeed -->|Yes| AskSeed{Run seed?}
    AskSeed -->|Yes| RunSeed[Run prisma db seed]
    AskSeed -->|No| Summary
    HasSeed -->|No| Summary
    RunSeed --> Summary

    Summary([Completion summary<br/>+ next steps])

    style Start fill:#4C51BF,color:#fff
    style Summary fill:#38A169,color:#fff
    style Exit1 fill:#E53E3E,color:#fff
    style DepsGate fill:#ED8936,color:#fff
```

### Key design decisions

1. **Orchestrator, not installer.** Bootstrap sequences existing Prisma commands — it doesn't replace them. Each step (init, link, migrate, generate, seed) delegates to its real implementation.

2. **Consent gates at every side effect.** Interactive `confirm()` prompts before migration and seeding. Users always know what will happen before it does.

3. **Deps gate without package manager opinions.** When `dotenv` or `prisma` aren't installed, bootstrap pauses and tells the user to install them — it doesn't make package manager decisions for them.

4. **Local binary delegation.** Migrate and generate shell out to the local `prisma` binary when present. This means Prisma 6 and Prisma 7 projects both work correctly — each version uses its own schema resolution.

5. **Idempotent re-runs.** Running bootstrap again in an already-linked project skips the link step and proceeds to migrate/generate. No `--force` needed for normal re-runs.

### New: `prisma bootstrap` command
- **Command registration** (`bin.ts`): Registered as root-level command alongside `init`, `generate`, etc.
- **Project state detection** (`bootstrap/project-state.ts`): Detects `package.json`, `schema.prisma`, `prisma.config.ts`, `.env`, models, and seed scripts.
- **Orchestration** (`bootstrap/Bootstrap.ts`): Sequences init → link → migrate → generate → seed with consent gates.
- **Template scaffolding** (`bootstrap/template-scaffold.ts`): Downloads and extracts starter templates, detects package managers, installs deps.
- **Completion summary** (`bootstrap/completion-output.ts`): Structured output showing step status and next steps.
- **PostHog telemetry** (`bootstrap/telemetry.ts`): Emits activation lifecycle events. Respects `CHECKPOINT_DISABLE`. Sends error codes only (no raw messages).

### Refactored: `Link` returns structured result
- `Link.ts`: Added public `link()` method returning `LinkResult` with `workspaceId`, `projectId`, `environmentId`, `databaseId`, `connectionString`.
- `resolveDatabase()`: Returns `DatabaseSelection` including workspace/project context.
- Added JWT-based `extractWorkspaceIdFromToken()` and best-effort `resolveProjectForDatabase()`.
- `parse()` behavior unchanged — backward compatible.

### Flags
- `--api-key`: Workspace API key (CI / non-interactive)
- `--database`: Database ID to link to
- `--template`: Starter template name (e.g. `nextjs`, `express`)
- `--force`: Re-link even if already linked
- `-h, --help`: Help text

### Automated tests (36 passing)
- `project-state.vitest.ts`: 10 tests — file detection combinations
- `template-scaffold.vitest.ts`: 16 tests — package manager detection, template download
- `Bootstrap.vitest.ts`: 10 tests — help/validation, new project flow, existing project flow, consent gates, error handling

### Integration tests (12/12 passing)
See [test report comment](https://github.com/prisma/prisma/pull/29374#issuecomment-4143481678) for full details covering: empty dirs, npm/pnpm/bun projects, Next.js apps, templates, Prisma 6 + 7 compat, idempotency, `--force`, and `CHECKPOINT_DISABLE`.

/integration


packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (!header) break
+
+    offset += 512
+    const paddedSize = Math.ceil(header.size / 512) * 512
+
+    if (header.type === '0' || header.type === '') {
+      const relPath = stripFirstComponent(header.name)
+      if (relPath?.startsWith(templatePrefix)) {
+        const destPath = path.resolve(targetDir, relPath.slice(templatePrefix.length))
+        if (!destPath.startsWith(path.resolve(targetDir))) continue
```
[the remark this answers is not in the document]
kristof-siket: Fixed in 2cf6dce — now uses `path.resolve(targetDir) + path.sep` to prevent sibling-prefix false positives.

packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+const execFileAsync = promisify(execFile)
+
+export async function installDependencies(baseDir: string): Promise<void> {
+  const pm = detectPackageManager(baseDir)
+  await execFileAsync(pm, ['install'], {
+    cwd: baseDir,
+    env: { ...process.env },
+  })
```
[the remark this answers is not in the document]
kristof-siket: Fixed in 2cf6dce — added `shell: process.platform === 'win32'` to the execFileAsync call.

packages/cli/src/bootstrap/Bootstrap.ts
```diff
@@ -0,0 +1,563 @@
...
+            steps.migrate = 'completed'
+            stepsCompleted.push('migrate')
+            await emitStepCompleted(telemetryCtx, 'migrate', performance.now() - migrateStart)
+          }
+        } catch (err) {
+          const msg = err instanceof Error ? err.message : String(err)
+          console.log(`${yellow('warn')} Migration failed: ${sanitizeErrorMessage(msg)}`)
+          steps.migrate = 'failed'
+          await emitStepFailed(telemetryCtx, 'migrate', sanitizeErrorMessage(msg))
+        }
```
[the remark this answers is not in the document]
kristof-siket: Fixed in 2cf6dce — all three call sites now use `if (useLocalBin && localPrismaBin)` instead of `localPrismaBin!`.

packages/cli/src/bootstrap/Bootstrap.ts
```diff
@@ -0,0 +1,563 @@
...
+    try {
+      return await this.run(apiKey, databaseId, templateName, force, config, baseDir)
+    } catch (err) {
+      if (err instanceof LinkApiError) {
+        return new HelpError(`\n${bold(red('!'))} ${sanitizeErrorMessage(err.message)}`)
+      }
+      const message = err instanceof Error ? err.message : String(err)
+      return new HelpError(`\n${bold(red('!'))} ${sanitizeErrorMessage(message)}`)
+    }
+  }
```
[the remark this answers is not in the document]
kristof-siket: Acknowledged — this is a valid follow-up but out of scope for the initial PR. The config is already accepted as a parameter in `run()` and passed through from the CLI framework. We'll add the explicit `--config` flag in a subsequent PR.

packages/cli/src/bootstrap/Bootstrap.ts
```diff
@@ -0,0 +1,563 @@
...
+      if (fs.existsSync(envPath)) {
+        const envContent = fs.readFileSync(envPath, 'utf-8')
+        const match = envContent.match(/^DATABASE_URL=["']?([^"'\n]+)["']?/m)
+        if (match) databaseUrl = match[1]
+      }
+    }
+    if (databaseUrl) {
+      process.env.DATABASE_URL = databaseUrl
+    }
+    const subprocessEnv: Record<string, string> = databaseUrl ? { DATABASE_URL: databaseUrl } : {}
```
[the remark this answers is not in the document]
kristof-siket: The regex handles the connection strings produced by `prisma postgres link` which have a predictable format. A full dotenv parser would add a dependency for marginal gain here — the primary path already uses `telemetryCtx.linkResult?.connectionString` directly. Will revisit if edge cases surface.

packages/cli/src/bootstrap/__tests__/project-state.vitest.ts
```diff
@@ -0,0 +1,231 @@
...
+  })
+
+  test('returns empty array for schema without models', () => {
+    const prismaDir = path.join(tmpDir, 'prisma')
+    fs.mkdirSync(prismaDir)
+    fs.writeFileSync(path.join(prismaDir, 'schema.prisma'), `datasource db { provider = "postgresql" }`, 'utf-8')
+
+    expect(getModelNames(tmpDir)).toEqual([])
+  })
+})
```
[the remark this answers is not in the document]
kristof-siket: Good suggestion for hardening. The functions already handle missing files gracefully — malformed content tests are a nice-to-have we'll add as follow-up test coverage.

packages/cli/src/bootstrap/__tests__/template-scaffold.vitest.ts
```diff
@@ -0,0 +1,54 @@
...
+  test('defaults to npm when no lockfile exists', () => {
+    expect(detectPackageManager(tmpDir)).toBe('npm')
+  })
+
+  test('prefers pnpm over yarn when both exist', () => {
+    fs.writeFileSync(path.join(tmpDir, 'pnpm-lock.yaml'), '', 'utf-8')
+    fs.writeFileSync(path.join(tmpDir, 'yarn.lock'), '', 'utf-8')
+    expect(detectPackageManager(tmpDir)).toBe('pnpm')
+  })
+})
```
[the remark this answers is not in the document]
kristof-siket: Agreed — adding bun/deno detection test cases is a good follow-up. The detection logic itself is straightforward (lockfile existence check) and unlikely to regress, but we'll add the coverage.

packages/cli/src/postgres/link/Link.ts
```diff
@@ -67,17 +86,43 @@ async function resolveDatabase(client: ManagementApiClient): Promise<string> {
...
+      const databases = await listDatabases(client, project.id)
+      if (databases.some((db) => db.id === databaseId)) {
+        return project.id
+      }
+    }
+  } catch {
+    // best-effort; telemetry should not block the command
+  }
+  return null
 }
```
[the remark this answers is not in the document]
kristof-siket: The function already returns early once found, so in practice it's usually 1-2 API calls. This is best-effort telemetry context — it never blocks the command and failures are silently caught. A server-side lookup endpoint would be the proper optimization if this becomes a concern.

packages/cli/src/bootstrap/project-state.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+function checkSeedInPackageJson(baseDir: string): boolean {
+  const packageJsonPath = path.join(baseDir, 'package.json')
+  if (!fs.existsSync(packageJsonPath)) return false
+
+  try {
+    const pkg = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'))
+    return typeof pkg.prisma?.seed === 'string' && pkg.prisma.seed.length > 0
+  } catch {
```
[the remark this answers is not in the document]
kristof-siket: The schema read in `getModelNames` is called within the bootstrap flow which already handles exceptions at the top level. A local try/catch would silently swallow IO errors — the current behavior (letting it propagate) is acceptable for an initial implementation. Will consider adding graceful handling as follow-up.

packages/cli/src/bootstrap/telemetry.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+export async function emitStepFailed(ctx: TelemetryContext, stepName: string, error: string): Promise<void> {
+  try {
+    await eventCapture.capture(ctx.distinctId, 'activation:cli_step_failed', {
+      ...baseProperties(ctx),
+      step_name: stepName,
+      error_message: error,
+    })
```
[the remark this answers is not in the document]
kristof-siket: Fixed in eed8045 — `emitStepFailed` now sends `error_code` (extracted Prisma error codes like P1012, system errors like ENOENT, or categorized identifiers) instead of the raw `error_message`. No sensitive values leave the client.

packages/cli/src/bootstrap/project-state.ts
```diff
@@ -0,0 +1,104 @@
...
+    return false
+  }
+}
+
+export function getModelNames(baseDir: string): string[] {
+  const schemaPath = findSchemaPath(baseDir)
+  if (!schemaPath) return []
+  const content = fs.readFileSync(schemaPath, 'utf-8')
+  const matches = content.matchAll(/^\s*model\s+(\w+)/gm)
+  return Array.from(matches, (m) => m[1])
```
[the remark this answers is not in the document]
kristof-siket: Fixed in 3afb12a — wrapped `readFileSync` in `getModelNames` and `detectProjectState` with try/catch, consistent with all other read functions in the module.

kristof-siket:
## Integration Test Report: `prisma bootstrap`

**Version:** `prisma@7.7.0-integration-feat-prisma-bootstrap.2`  
**API:** Production (`https://api.prisma.io`)  
**Date:** 2026-03-27  
**Database region:** `eu-west-3` (Europe — Paris)

### What was tested

Comprehensive end-to-end testing of the `prisma bootstrap` command across 12 scenarios covering: empty directories, fresh projects, existing apps (Next.js), multiple package managers (npm, pnpm, bun), template scaffolding, Prisma 6 and Prisma 7 compatibility, idempotent re-runs, `--force` re-linking, and telemetry consent (`CHECKPOINT_DISABLE`).

### Results: 12/12 PASS

#### Fresh project scenarios

| ID | Scenario | Init | Link | Deps Gate | Migrate | Generate | Status |
|----|----------|:----:|:----:|:---------:|:-------:|:--------:|--------|
| T-01 | Empty dir (no `package.json`) | Shows message + offers template | — | — | — | — | **PASS** |
| T-02 | npm project, deps **not** installed | done | done | Detects missing `dotenv` + `prisma`, pauses | — | — | **PASS** |
| T-03 | npm project, deps pre-installed, clean DB | done | done | — | done | done | **PASS** |

#### Template scaffolding

| ID | Scenario | Template | Link | Deps | Migrate | Generate | Seed | Status |
|----|----------|:--------:|:----:|:----:|:-------:|:--------:|:----:|--------|
| T-05 | `--template nextjs` | done (2 models) | done | done | done | done | failed¹ | **PASS** |

¹ Seed failed because the template's seed command references `tsx` which isn't in the template's dependencies. This is a template authoring issue, not a bootstrap issue. Bootstrap correctly reports it as failed and continues.

#### Existing app scaffolders

| ID | Scenario | Init | Link | Generate | Status |
|----|----------|:----:|:----:|:--------:|--------|
| T-06 | Next.js (`create-next-app`) | done | done | done | **PASS** |
| T-07 | pnpm project | done | done | done | **PASS** |
| T-08 | bun project | done | done | done | **PASS** |

#### Existing Prisma projects (Prisma 6 vs 7 compatibility)

| ID | Scenario | Local Prisma | Config file | Link | Migrate | Generate | Status |
|----|----------|:------------:|:-----------:|:----:|:-------:|:--------:|--------|
| T-09a | Existing project, **Prisma 7** (7.6.0) | `prisma.config.ts` present | ✓ | done | done (2 models) | done | **PASS** |
| T-09b | Existing project, **Prisma 6** (6.19.2) | `datasource url` in schema | — | done | done (2 models) | done | **PASS** |

Both versions work correctly because bootstrap shells out to the **local** `prisma` binary for `migrate` and `generate`:
- Prisma 7 reads config from `prisma.config.ts` → works
- Prisma 6 reads `datasource url` from schema + `.env` → works

**Edge case covered:** An earlier test used Prisma 7 installed locally but without `prisma.config.ts` (an invalid setup). This correctly fails at migration — but it's not a bootstrap bug, it's a broken local environment that wouldn't occur organically (either `prisma init` on v7 creates the config, or a v6 project has `url` in the schema).

#### Idempotency & flags

| ID | Scenario | Behavior | Status |
|----|----------|----------|--------|
| T-04 | Re-run in already-bootstrapped dir | "Already linked to Prisma Postgres" — skips link, proceeds | **PASS** |
| T-10 | `--force` re-links to different database | Correctly re-linked from EU → Paris db | **PASS** |
| T-11 | `CHECKPOINT_DISABLE=1` | Command works identically, telemetry silently skipped | **PASS** |

### Key observations

1. **Happy path is solid.** Fresh npm project + deps + clean DB completes init → link → migrate → generate in ~30s.
2. **Template path works well.** Scaffolding, dependency installation, and migration all succeed. Only the seed step can fail if the template has missing transitive deps.
3. **Package manager diversity confirmed.** npm, pnpm, and bun all work correctly. Lockfile detection picks up the right manager.
4. **Prisma 6 ↔ 7 compat is clean.** Both major versions work because bootstrap delegates to the local binary for schema-aware commands.
5. **Idempotency works.** Re-running after a successful bootstrap skips the link step and proceeds to migrate/generate without requiring `--force`.
6. **Telemetry consent respected.** `CHECKPOINT_DISABLE=1` silently disables all PostHog events without affecting functionality.

### Cleanup

All test databases created during testing were deleted via the Management API.

packages/cli/src/bootstrap/Bootstrap.ts
```diff
@@ -0,0 +1,563 @@
...
+    // version installed locally (e.g., Prisma 6 with `url` in schema.prisma instead of
+    // prisma.config.ts). Running migrate in-process would force this CLI's engine version
+    // on their project, causing version mismatches or hard failures.
+    //
+    // When a local `prisma` binary exists in node_modules, we shell out to it so that
+    // migrate/seed run with the user's own Prisma version and configuration. We only fall
+    // back to in-process execution for fresh projects where `init` just scaffolded Prisma 7
+    // files and no local binary exists yet.
+    const localPrismaBin = findLocalPrismaBin(baseDir)
+    const useLocalBin = localPrismaBin !== null && (initialState.hasSchemaFile || templateScaffolded)
```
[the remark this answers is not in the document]
kristof-siket: This is intentional. The scenario described (project has `prisma` installed but no schema, init creates schema) is one where the local binary may be Prisma 6 while init just created Prisma 7-style `prisma.config.ts`. Shelling out to the v6 binary would fail because it can't parse the config. Running in-process (current behavior) is correct — the bootstrap CLI that just created the config is the right one to run migrate against it. Using `updatedState` here would cause a regression for Prisma 6 users.

packages/cli/src/bootstrap/project-state.ts
```diff
@@ -0,0 +1,110 @@
...
+const SCHEMA_CANDIDATES = ['prisma/schema.prisma', 'schema.prisma']
+const MODEL_PATTERN = /^\s*model\s+\w+/m
+
+function findSchemaPath(baseDir: string): string | null {
+  for (const candidate of SCHEMA_CANDIDATES) {
+    const full = path.join(baseDir, candidate)
+    if (fs.existsSync(full)) return full
+  }
+  return null
+}
```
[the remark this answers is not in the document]
kristof-siket: Acknowledged as a known limitation. `loadSchemaContext` is async, invokes the schema engine, and requires valid schema content — too heavy for the lightweight "does a schema exist?" heuristic we need here. The real gap is non-standard schema paths configured in `prisma.config.ts` (e.g. `src/db/schema.prisma`) — our check won't find those. That's an uncommon setup and acceptable for the initial release. We'll revisit with a lighter config-aware lookup as follow-up.

packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -0,0 +1,161 @@
...
+    offset += 512
+    const paddedSize = Math.ceil(header.size / 512) * 512
+
+    if (header.type === '0' || header.type === '') {
+      const relPath = stripFirstComponent(header.name)
+      if (relPath?.startsWith(templatePrefix)) {
+        const destPath = path.resolve(targetDir, relPath.slice(templatePrefix.length))
+        if (!destPath.startsWith(path.resolve(targetDir) + path.sep)) continue
+        fs.mkdirSync(path.dirname(destPath), { recursive: true })
+        fs.writeFileSync(destPath, tarBuffer.subarray(offset, offset + header.size))
```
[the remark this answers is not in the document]
kristof-siket: Templates are only offered when the project doesn't have a Prisma schema, and the user explicitly confirms the selection. Overwriting is the expected behavior — the template replaces the project setup. Adding collision detection would make the flow fail awkwardly since `package.json` always exists at this point (we require it). If we want a warning for non-empty directories, that's a UX decision for a follow-up, not a safety issue.

packages/cli/src/bootstrap/Bootstrap.ts
```diff
@@ -0,0 +1,563 @@
...
+
+    try {
+      const init = Init.new()
+      const initArgs = ['--datasource-provider', 'postgresql']
+      if (withModel) initArgs.push('--with-model')
+      const initResult = await init.parse(initArgs, config)
+
+      if (initResult instanceof Error) {
+        await emitStepFailed(telemetryCtx, 'init', sanitizeErrorMessage(initResult.message))
+        throw new LinkApiError(`Init failed: ${initResult.message}`)
```
wmadden: **F01 [Blocking]** `LinkApiError` is semantically wrong here — it represents Management API errors (with an optional `statusCode`), not init failures. The top-level catch in `parse()` already handles both `LinkApiError` and generic errors identically, so a plain `Error` would be correct:

```suggestion
        throw new Error(`Init failed: ${initResult.message}`)
```

Same issue on L553.

packages/cli/src/postgres/link/Link.ts
```diff
@@ -67,17 +86,46 @@ async function resolveDatabase(client: ManagementApiClient): Promise<string> {
...
+    if (parts.length < 2 || !parts[1]) return null
+    const payload = JSON.parse(Buffer.from(parts[1], 'base64url').toString())
+    if (typeof payload !== 'object' || payload === null) return null
+    return payload.workspace_id ?? payload.workspaceId ?? null
+  } catch {
+    return null
+  }
+}
+
+async function resolveProjectForDatabase(client: ManagementApiClient, databaseId: string): Promise<string | null> {
```
wmadden, line 121: **F04 [Non-blocking]** This function exists solely to populate the `'project-id'` PostHog telemetry property — the resolved `projectId` has no effect on user-visible output or control flow. It makes N+1 API calls (list all projects, then list databases for each), which adds latency before `createDevConnection` runs on L257.

The catch block (L125) also swallows *all* errors, not just telemetry-related ones. The comment is misleading — this function is project resolution, not telemetry itself.

Consider:
1. Simply shipping `project-id: null` in the `--api-key` + `--database` path (most pragmatic)
2. Running this concurrently with `createDevConnection` via `Promise.allSettled` to avoid blocking
3. At minimum, clarifying the comment to "best-effort resolution for telemetry context"

packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -0,0 +1,161 @@
...
+  if (fs.existsSync(path.join(baseDir, 'bun.lock')) || fs.existsSync(path.join(baseDir, 'bun.lockb'))) return 'bun'
+  if (fs.existsSync(path.join(baseDir, 'deno.lock'))) return 'deno'
+  return 'npm'
+}
+
+const execFileAsync = promisify(execFile)
+
+export async function installDependencies(baseDir: string): Promise<void> {
+  const pm = detectPackageManager(baseDir)
+  await execFileAsync(pm, ['install'], {
```
wmadden: **F05 [Non-blocking]** No timeout on `execFileAsync`. A stuck `npm install` (registry issue, hung postinstall script) would block bootstrap indefinitely. Consider adding `timeout: 300_000` (5 min) to the options.

packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/**
+ * Downloads the prisma-examples tarball and extracts only `orm/<templateName>/`
+ * into `targetDir`. Buffers the decompressed tar and walks it entry-by-entry.
+ *
+ * The prisma-examples tarball is typically <10 MB decompressed, so buffering
+ * is acceptable and avoids a streaming tar parser dependency.
+ */
+export async function downloadAndExtractTemplate(templateName: string, targetDir: string): Promise<void> {
```
wmadden: **F15 [Non-blocking, acknowledged]** Template scaffolding downloads and extracts code from `prisma/prisma-examples` then runs `npm install` (executing postinstall scripts) without checksum or pinned commit verification. This is the same trust model as `create-next-app` / `create-prisma` — not a novel risk. The URL is hardcoded, template names are allowlisted, and path traversal is guarded. Just flagging as an acknowledged trust boundary.

packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -0,0 +1,161 @@
...
+  })
+}
+
+interface TarEntry {
+  name: string
+  size: number
+  type: string
+}
+
+function parseTarHeader(buf: Buffer, offset: number): TarEntry | null {
```
wmadden, line 122: **F07 [Non-blocking]** The tar header parser doesn't validate checksums (bytes 148–155). A corrupted or truncated download could extract partial/invalid files silently. Consider adding a post-extraction sanity check (e.g., verify `package.json` exists in the extracted template).

packages/cli/src/bootstrap/template-scaffold.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * GitHub tarballs add a `<owner>-<repo>-<sha>/` prefix to all paths.
+ */
+function stripFirstComponent(filePath: string): string | null {
+  const idx = filePath.indexOf('/')
+  if (idx === -1) return null
+  return filePath.slice(idx + 1)
+}
+
+type PackageManager = 'npm' | 'pnpm' | 'yarn' | 'bun' | 'deno'
```
wmadden: **F08 [Nit]** `PackageManager` type is not exported but `detectPackageManager` is. Consumers (including tests) can't type the result without re-declaring the union.

packages/cli/src/bootstrap/__tests__/Bootstrap.vitest.ts
```diff
@@ -0,0 +1,296 @@
...
+  FileTokenStorage: class {
+    getTokens() {
+      return Promise.resolve({ accessToken: 'stored_token', workspaceId: 'wksp_1' })
+    }
+  },
+}))
+
+vi.mock('@inquirer/prompts', () => ({
+  select: vi.fn(),
+  confirm: vi.fn(),
```
wmadden, line 45: **F09 [Non-blocking]** `confirm` is mocked to return the same value for all calls in each test, but the orchestrator calls it for multiple distinct prompts (template, sample model, migrate, seed). Consider adding a test using `mockResolvedValueOnce()` to verify mixed accept/decline scenarios — e.g., accepting migration but declining seed.

packages/cli/src/bootstrap/telemetry.ts
```diff
@@ -0,0 +1,102 @@
+import type { LinkResult } from '../postgres/link/Link'
+import { PosthogEventCapture } from '../utils/nps/capture'
+import type { ProjectState } from './project-state'
+
+function isTelemetryDisabled(): boolean {
+  return Boolean(process.env.CHECKPOINT_DISABLE)
+}
+
+const eventCapture = new PosthogEventCapture()
```
wmadden, line 9: **F10 [Nit]** Module-level singleton — importing this module always creates a `PosthogEventCapture` even when telemetry is disabled. Currently harmless (constructor is trivial), just noting.

packages/cli/src/bootstrap/__tests__/Bootstrap.vitest.ts
```diff
@@ -0,0 +1,296 @@
...
+    const result = await Bootstrap.new().parse(
+      ['--api-key', 'bad_key', '--database', 'db_abc123'],
+      defaultTestConfig(),
+      tmpDir,
+    )
+
+    expect(result).toBeInstanceOf(HelpError)
+    expect((result as HelpError).message).toContain('Invalid credentials')
+  })
+})
```
wmadden, line 506: **F11 [Non-blocking]** Missing unit test coverage for two significant orchestration paths:
- **Deps gate**: missing `dotenv`/`prisma` in `node_modules` → early exit with "Missing dependencies" message
- **Seed step**: seed script present + user accepts/declines

These are likely covered by integration tests, but unit tests would increase confidence in the orchestration logic.

packages/cli/src/bootstrap/project-state.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  try {
+    const pkg = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'))
+    return typeof pkg.prisma?.seed === 'string' && pkg.prisma.seed.trim().length > 0
+  } catch {
+    return false
+  }
+}
+
+const SEED_PATTERN = /seed\s*[:=]\s*['"`]/
```
wmadden, line 41: **F12 [Nit]** This regex could match `seed` in comments or unrelated code (e.g., `// Don't seed: "this"`). Acceptable heuristic for real `prisma.config.ts` files, but consider tightening if false positives become a concern.

packages/cli/src/bootstrap/Bootstrap.ts
```diff
@@ -0,0 +1,563 @@
...
+    force: boolean,
+    config: PrismaConfigInternal,
+    baseDir: string,
+  ): Promise<string | HelpError> {
+    const flowStart = performance.now()
+    const stepsCompleted: string[] = []
+    const steps: BootstrapStepStatus = {
+      init: 'skipped',
+      template: 'not-applicable',
+      link: 'failed',
```
wmadden: **F13 [Nit]** Initializing `link` status as `'failed'` is unusual. While functionally correct (link either succeeds or throws), consider `'skipped'` as the default and setting `'failed'` explicitly in the catch block, matching the pattern used for the other steps.

packages/cli/src/bootstrap/project-state.ts
```diff
@@ -0,0 +1,110 @@
...
+  hasPackageJson: boolean
+  hasSchemaFile: boolean
+  hasPrismaConfig: boolean
+  hasEnvFile: boolean
+  hasModels: boolean
+  hasSeedScript: boolean
+}
+
+const SCHEMA_CANDIDATES = ['prisma/schema.prisma', 'schema.prisma']
+const MODEL_PATTERN = /^\s*model\s+\w+/m
```
wmadden, line 14: **F06 [Nit]** The model-detection regex and schema-path search logic is duplicated in `Link.ts` (`schemaHasModels`, L289–300). Consider sharing the implementation to avoid future divergence.

wmadden (approved):
Pre-emptive approval. Please address the comments (or note why you won't) before merging.

Overall this is well-structured — clean orchestrator pattern, good separation of concerns, consent gates on every side effect, and smart local-binary delegation for Prisma 6 compat.

**1 blocking issue**: `LinkApiError` misused for init failures (F01).

**Non-blocking concerns** (6): N+1 API calls for telemetry-only data (F04), no timeout on `installDependencies` (F05), no checksum in tar parser (F07), confirm mock always returns same value (F09), missing unit test coverage for deps gate and seed paths (F11), template scaffolding trust assumption (F15).

**Nits** (5): Duplicated `schemaHasModels` logic (F06), `PackageManager` type not exported (F08), module-level `PosthogEventCapture` instantiation (F10), imprecise `SEED_PATTERN` regex (F12), default link status is `'failed'` (F13).

packages/cli/src/bootstrap/Bootstrap.ts
kristof-siket: Fixed

packages/cli/src/postgres/link/Link.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/template-scaffold.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/template-scaffold.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/template-scaffold.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/__tests__/Bootstrap.vitest.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/__tests__/Bootstrap.vitest.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/Bootstrap.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/project-state.ts
kristof-siket: Fixed

packages/cli/src/bootstrap/project-state.ts
kristof-siket: acknowledged, didn't fix this since the config file is usually not hand-written so i don't think it is worth the extra complexity

packages/cli/src/bootstrap/telemetry.ts
kristof-siket: acknowledged but left as-is for now (since it is pretty trivial)

packages/cli/src/bootstrap/template-scaffold.ts
kristof-siket: acknowledged, existing pattern applied so not fixing this one

PR #29382, opened 2026-03-24
fix(client): prevent caching of createMany queries to avoid cache bloat and possible node.js crashes

### LINKED ISSUE #29331: Memory consumption increased significantly in Prisma 7 in comparison to Prisma 6
### Bug description

We observe significant memory consumption increase after migration from Prisma 6 to Prisma 7 for Postgres.

### Severity

### Reproduction

Following example can be considered:

```schema.prisma
generator client {
  provider = "prisma-client"
  output   = "../src/generated/prisma"
}

datasource db {
  provider = "postgresql"
  url      = env("PRISMA_TEST_POSTGRES_URL")
}

model Sample {
  time     DateTime
  sensorId String

  pressure    Float
  temperature Float

  @@unique([time, sensorId])
  @@index([time])
}

```

The test script performing following steps:
1. generate batch of data
2. insert batch into database (Postgres)
3. log memory usage
4. repeat steps 1-3

```main.ts
import 'dotenv/config';
import { PrismaClient, type Sample } from './generated/prisma/client';
import { StatsWriter } from './stats-writer';

const DB_URL = process.env['PRISMA_TEST_POSTGRES_URL'];
const NUM_ITERATIONS = 1_000;
const BATCH_SIZE = 10_000;
const STATS_FILE = 'prisma-6.csv';

const stats = new StatsWriter(STATS_FILE);

  datasourceUrl: DB_URL,
});

const nextBatch = (iteration: number, batchSize: number): Sample[] => {
  return Array.from({ length: batchSize }, (_, i) => ({
    time: new Date(iteration * batchSize + i),
    sensorId: Math.random().toString(36).substring(2, 15),
    pressure: Math.random() * 1000,
    temperature: Math.random() * 10,
  }));
};

const runTest = async () => {
  await prisma.sample.deleteMany();

  for (let i = 0; i < NUM_ITERATIONS; i++) {
    const batch = nextBatch(i, BATCH_SIZE);

    const start = process.hrtime.bigint();
    await prisma.sample.createMany({ data: batch });
    const end = process.hrtime.bigint();
    const executionTimeMillis = Number(end - start) / 1_000_000;

    await stats.write({
      iteration: i,
      executionTimeMillis,
      memory: process.memoryUsage(),
    });
  }
};

await runTest();
```

With the test script executed with Prisma 6 and Prisma 7 following results obtained:

- [prisma-6.csv](https://github.com/user-attachments/files/25885783/prisma-6.csv)
- [prisma-7.csv](https://github.com/user-attachments/files/25885782/prisma-7.csv)

Visual charts:

Image

Image

Image

Image

Image

In our production workload, we have "bigger" data model (approx 40 attributes) and have NodeJS OOM crash in case of intensive data ingestion.

### Frequency

Consistently reproducible

### Prisma Schema & Queries

```prisma
generator client {
  provider = "prisma-client"
  output   = "../src/generated/prisma"
}

datasource db {
  provider = "postgresql"
  url      = env("PRISMA_TEST_POSTGRES_URL")
}

model Sample {
  time     DateTime
  sensorId String

  pressure    Float
  temperature Float

  @@unique([time, sensorId])
  @@index([time])
}

```
```ts
const runTest = async () => {
  await prisma.sample.deleteMany();

  for (let i = 0; i < NUM_ITERATIONS; i++) {
    const batch = nextBatch(i, BATCH_SIZE);

    const start = process.hrtime.bigint();
    await prisma.sample.createMany({ data: batch });
    const end = process.hrtime.bigint();
    const executionTimeMillis = Number(end - start) / 1_000_000;

    await stats.write({
      iteration: i,
      executionTimeMillis,
      memory: process.memoryUsage(),
    });
  }
};
```

### Prisma Config

```ts
import 'dotenv/config';

export default defineConfig({
  migrations: {
    path: 'prisma/migrations',
  },
});

```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** macOS, Linux
- **Database:** PostgreSQL
- **Node.js version:** v24.14.0

### Prisma Version

```
7.4.2
```

shakim:
`createMany` operations inherently have unstable cache keys due to varying row counts and parameter de-duplication.

`createMany` operations are characterized by large cache keys and very large query plans. In large batches, the cache will consume several GB of memory that will not be evicted by the GC even after the query is complete because the plans are still referenced by the cache itself.

This fix opts out of caching such queries with the following rationale:
1. Such queries will most likely create cache-misses and so caching them is ineffective
2. When running batch load/update workloads, these query plans will evict valid queries from the cache thus creating an advert effect on performance

Fixes: prisma#29331

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced query-plan cache growth by skipping single-query caching for high-volume bulk create operations, lowering memory use and improving bulk write efficiency.

* **Tests**
  * Added a comprehensive functional test suite (matrix, provider-specific schema setup, and stress tests) that seeds data and runs parameterized create-many scenarios to validate performance and correctness under load.

### DIFF
--- a/packages/client/src/runtime/core/engines/client/ClientEngine.ts
+++ b/packages/client/src/runtime/core/engines/client/ClientEngine.ts
@@ -475,14 +475,17 @@ export class ClientEngine implements Engine {
       const cacheKey = JSON.stringify(parameterizedQuery)
       placeholderValues = extractedValues
 
-      const cached = this.#queryPlanCache.getSingle(cacheKey)
+      const isCacheable = query.action !== 'createMany' && query.action !== 'createManyAndReturn'
+      const cached = isCacheable ? this.#queryPlanCache.getSingle(cacheKey) : undefined
       if (cached) {
         debug('query plan cache hit')
         plan = cached
       } else {
         debug('query plan cache miss')
         plan = this.#compileQuery(parameterizedQuery, cacheKey, queryCompiler)
-        this.#queryPlanCache.setSingle(cacheKey, plan)
+        if (isCacheable) {
+          this.#queryPlanCache.setSingle(cacheKey, plan)
+        }
       }
     }
 
--- a/packages/client/src/runtime/core/engines/client/ClientEngine.ts
+++ b/packages/client/src/runtime/core/engines/client/ClientEngine.ts
@@ -475,6 +475,9 @@ export class ClientEngine implements Engine {
       const cacheKey = JSON.stringify(parameterizedQuery)
       placeholderValues = extractedValues
 
+      // We do not cache `createMany` and `createManyAndReturn` queries as they are very unlikely
+      // to benefit from caching due to their high variability in parameters, which leads to a very
+      // high cache miss rate and potential cache bloat.
       const isCacheable = query.action !== 'createMany' && query.action !== 'createManyAndReturn'
       const cached = isCacheable ? this.#queryPlanCache.getSingle(cacheKey) : undefined
       if (cached) {
--- a/packages/client/tests/functional/issues/29331-query-plan-cache-bloat/tests.ts
+++ b/packages/client/tests/functional/issues/29331-query-plan-cache-bloat/tests.ts
@@ -1,0 +1,96 @@
+import { faker } from '@snaplet/copycat'
+
+import testMatrix from './_matrix'
+// @ts-ignore
+import type { PrismaClient } from './generated/prisma/client'
+
+declare let prisma: PrismaClient
+
+function daysBefore(days: number): Date {
+  const date = new Date()
+  date.setDate(date.getDate() - days)
+  return date
+}
+
+testMatrix.setupTestSuite(() => {
+  const contactId1 = faker.database.mongodbObjectId()
+  const contactId2 = faker.database.mongodbObjectId()
+  const contactId3 = faker.database.mongodbObjectId()
+
+  afterEach(async () => {
+    await prisma.contactAnalytics.deleteMany()
+    await prisma.contact.deleteMany()
+  })
+
+  beforeEach(async () => {
+    await prisma.contact.createMany({
+      data: [{ id: contactId1 }, { id: contactId2 }, { id: contactId3 }],
+    })
+    await prisma.contactAnalytics.createMany({
+      data: [
+        {
+          contactId: contactId1,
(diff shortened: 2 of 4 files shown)


packages/client/tests/functional/issues/29331-query-plan-cache-bloat/tests.ts
```diff
@@ -0,0 +1,96 @@
...
+        {
+          contactId: contactId3,
+          date5: daysBefore(3),
+          val3: 10,
+        },
+      ],
+    })
+  })
+
+  const fullStress = process.env.PRISMA_CREATE_MANY_STRESS === 'true'
```
jacek-prisma, line 53: This test doesn't fail on current Prisma version unless this flag is set, I understand that this is done to avoid having a very long running test, but I think it undermines the value of having this test in the first place, since it won't let us discover regressions in CI. I have reproduced this issue, I think we can just remove the test and just document the reasoning.

packages/client/src/runtime/core/engines/client/ClientEngine.ts
```diff
@@ -475,14 +475,17 @@ export class ClientEngine implements Engine {
       const cacheKey = JSON.stringify(parameterizedQuery)
       placeholderValues = extractedValues
 
-      const cached = this.#queryPlanCache.getSingle(cacheKey)
+      const isCacheable = query.action !== 'createMany' && query.action !== 'createManyAndReturn'
```
jacek-prisma, line 481: Can we leave a comment explaining the reasoning
```suggestion
      // We do not cache `createMany` and `createManyAndReturn` queries as they are very unlikely
      // to benefit from caching due to their high variability in parameters, which leads to a very
      // high cache miss rate and potential cache bloat.
      const isCacheable = query.action !== 'createMany' && query.action !== 'createManyAndReturn'
```

packages/client/tests/functional/issues/29331-query-plan-cache-bloat/tests.ts
shakim: Correct.
The short test is useful with debug logs to demonstrate the cache misses problem whereas the long test demonstrates the actual node crash.

I should have documented that in the PR

jacek-prisma (approved):

PR #29389, opened 2026-03-25
Pre-bundle Studio frontend assets and replace Hono

sorenbs:
## Summary
- pre-bundle the Studio frontend into `build/studio.js` and `build/studio.css` as part of the CLI build
- simplify the Studio HTML shell and static asset serving to use explicit local routes only
- replace Hono with a minimal runtime-specific `Request -> Response` server binding and add listener-level coverage for the Node adapter
- include a small follow-up fix for Node response streaming so `GET` bodies are served correctly from `scripts/run-studio.ts`

## Testing
- `pnpm exec vitest run src/__tests__/Studio.vitest.ts src/__tests__/studio-server.vitest.ts`
- `pnpm --filter prisma build`

## Summary by CodeRabbit

* **New Features**
  * Studio frontend is now pre-bundled into single JS/CSS assets and served directly for consistent, faster loading.

* **Improvements**
  * Streamlined server routing with proper CORS/preflight handling, clearer error responses, and embedded runtime config for adapter selection.
  * Telemetry wiring added for Studio events.

* **Tests**
  * Expanded server and asset tests to improve reliability and prevent regressions.


packages/cli/src/Studio.ts
```diff
@@ -577,16 +438,19 @@ function prismaSslAcceptToMySQL2Ssl(sslAccept: string): { rejectUnauthorized: bo
...
+function getIndexHtml(adapter: StudioAdapterType): string {
+  return `<!doctype html>
 <html lang="en" style="height: 100%">
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="icon" href="${PRISMA_LOGO_SVG_DATA_URL}" type="image/svg+xml">
-    <script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4.1.17"></script>
-    <link rel="stylesheet" href="/ui/index.css">
+    <link rel="stylesheet" href="/${STUDIO_CSS_FILE_NAME}">
```
wmadden, line 435: **F01 (blocking): Tailwind browser script removal not verified**

The `@tailwindcss/browser` CDN script was removed from the HTML template. If `@prisma/studio-core`'s UI components use Tailwind utility classes that were being processed at runtime by this script (rather than being pre-compiled into `/ui/index.css`), Studio's styling will be broken.

Before merging, please test Studio against a real database and verify that the UI renders correctly without the Tailwind browser script. If styling is broken, either:
- Ensure studio-core ships fully compiled CSS (preferred; coordinate with the studio-core team), or
- Bundle the Tailwind browser script as a local asset instead of removing it.

packages/cli/src/studio-server.ts
```diff
@@ -0,0 +1,158 @@
...
+function startNodeStudioServer({ handler, onListen, port }: StartStudioServerOptions): StudioServer {
+  const server = createServer(async (nodeRequest, nodeResponse) => {
+    try {
+      const request = createNodeRequest(nodeRequest, port)
+      const response = await handler(request)
+      await writeNodeResponse(nodeResponse, response, nodeRequest.method)
+    } catch {
+      nodeResponse.statusCode = 500
+      nodeResponse.setHeader('Access-Control-Allow-Origin', '*')
+      nodeResponse.end('Internal Server Error')
```
wmadden: **F02 (blocking): Silent error swallowing**

This catch block returns a bare 500 with no logging and no error detail. This is the CLI's user-facing server — if a request fails, the user sees "Internal Server Error" in their browser with zero diagnostic information, and nothing appears in the terminal where they ran `prisma studio`. They have no way to debug the problem.

Please log the error to the terminal and include the error message in the response body:

```typescript
} catch (error) {
  console.error('[Prisma Studio]', error)
  nodeResponse.statusCode = 500
  nodeResponse.setHeader('Access-Control-Allow-Origin', '*')
  nodeResponse.end(error instanceof Error ? error.message : 'Internal Server Error')
}
```

packages/cli/src/Studio.ts
```diff
@@ -598,47 +462,240 @@ const INDEX_HTML =
...
+      }
+
+      await sendEvent(input).catch(() => {
+        // noop
+      })
+
+      return emptyResponse(200)
+    }
+
+    return textResponse('Not Found', 404)
```
wmadden, line 539: **F03 (non-blocking): No OPTIONS / CORS preflight handler**

The old Hono `cors()` middleware responded to `OPTIONS` preflight requests. The new handler has no `OPTIONS` case — unknown methods/paths fall through to this 404. Since Studio is same-origin, this is unlikely to cause issues, but if any client makes a cross-origin request (e.g. a browser extension), it would fail preflight.

Consider adding a minimal `OPTIONS` handler returning 204 with CORS headers.

wmadden (changes requested):
## Summary

Clean implementation that addresses both the frontend bundling and Hono removal. The code is well-structured, the test updates are thorough, and the new modules have clear boundaries. Three items need to be addressed before merge.

## What looks solid

- **`studio-entry.ts`** is clean and type-safe. The `satisfies` constraint on `ADAPTER_FACTORIES` ensures the adapter map stays in sync with the type. The `getStudioConfig` validation guards against malformed injected config.
- **`studio-server.ts`** correctly handles the Node `Request`/`Response` translation, including streaming bodies via `pipeline`, HEAD semantics, and request body handling (only attaches body for non-GET/HEAD methods).
- **Test coverage is well-structured**: handler logic tested via mock (fast, no I/O), server binding tested with real HTTP (catches Node-specific regressions like the HEAD-drops-GET-body bug documented in the AGENTS.md update).
- **The HTML template** is radically simpler — 30 lines down from 65, no import map, no CDN URLs.
- **AGENTS.md** field notes correctly updated: removed the now-obsolete esm.sh and import map notes, added the new pre-bundle serving note.

## Non-blocking concerns

### F03: No OPTIONS / CORS preflight handler

The old Hono `cors()` middleware responded to `OPTIONS` preflight requests. The new handler has no `OPTIONS` case — unknown methods/paths fall through to 404. Since Studio is same-origin (browser loads from the same `localhost` that serves the BFF), this is unlikely to cause issues. But if any client makes a cross-origin request (e.g. a browser extension, or a frontend running on a different port during development), it would fail preflight.

Suggestion: add a minimal `OPTIONS` handler that returns 204 with CORS headers, or document that cross-origin requests to the BFF are not supported.

### F04: `FILE_EXTENSION_TO_CONTENT_TYPE` map is oversized

The content-type map at L32–49 still contains entries for `.png`, `.jpg`, `.jpeg`, `.gif`, `.woff`, `.woff2`, `.ttf`, `.otf`, `.eot` — none of which are served anymore. After the bundling change, the server only serves `.html`, `.js`, `.css`, and `.svg`. Consider trimming it to only the content types actually used.

## Acceptance-criteria traceability

| Criterion | Status |
|-----------|--------|
| No `importmap`, `esm.sh`, `cdn.jsdelivr.net`, `/adapter.js` in HTML | ✅ Verified in `getIndexHtml` + test assertions |
| `build/studio.js` and `build/studio.css` produced by build | ✅ `buildStudioFrontend` in `build.ts` |
| `hono` and `@hono/node-server` removed from `package.json` | ✅ |
| Server handles Node, Bun, Deno | ✅ Runtime switch in `studio-server.ts`; Node tested, Bun/Deno untested (no CI runtime) |
| Tests pass with updates | ✅ |
| CORS on all responses | ✅ `withCors` wrapper + test assertions |
| AGENTS.md updated | ✅ |

wmadden (approved):
Eagerly approving, please address the open comments before merging 👍🏻

packages/cli/src/Studio.ts
sorenbs: Verified against the real Postgres URL from this thread by running `npx tsx scripts/run-studio.ts` locally and opening `http://localhost:5555` with Playwright. The UI renders correctly without the Tailwind browser script: the page loads with `link[rel=stylesheet] = /studio.css`, the script list is `[inline, /studio.js]`, and the Studio shell/table grid render with the expected styling. No code change needed here.

packages/cli/src/studio-server.ts
sorenbs: Addressed in `922f5d429`. The Node binding now logs `[Prisma Studio]` errors to the terminal, returns the error message in the 500 response body, and guards the fallback path when headers have already been sent. I also added a real-listener regression test for the error path in `studio-server.vitest.ts`.

packages/cli/src/Studio.ts
sorenbs: Addressed in `3014ea409`. The handler now responds to `OPTIONS` with `204` plus the minimal CORS preflight headers (`Access-Control-Allow-Origin`, `Access-Control-Allow-Methods`, and `Access-Control-Allow-Headers`), and `Studio.vitest.ts` covers that path.

sorenbs:
Addressed F04 from the review summary in `fc1bd1c01` by trimming `FILE_EXTENSION_TO_CONTENT_TYPE` down to the extensions the bundled Studio server actually uses after this change (`.html`, `.js`, `.css`).

jacek-prisma (approved):

PR #29391, opened 2026-03-25
chore(client): use typescript 6 in e2e tests

aqrln:
Follow up to https://github.com/prisma/prisma/pull/29383


jacek-prisma:
See https://github.com/prisma/prisma/pull/29385

PR #29409, opened 2026-03-28
fix(client-generator-ts): preserve Cloudflare ts imports in native Deno projects

raashish1601:
## Summary
- preserve generated TypeScript import/export extensions for `runtime = "cloudflare"` when Prisma is generating into a native Deno project
- keep the existing bundler-style bare import behavior for non-Deno Cloudflare projects
- add focused regression coverage for the Deno-vs-non-Deno inference split

Fixes #29403

## Validation
- `corepack pnpm exec vitest run packages/client-generator-ts/tests/file-extensions.test.ts`
- `corepack pnpm exec prettier --check packages/client-generator-ts/src/file-extensions.ts packages/client-generator-ts/src/generator.ts packages/client-generator-ts/tests/file-extensions.test.ts`
- `corepack pnpm exec eslint packages/client-generator-ts/src/file-extensions.ts packages/client-generator-ts/src/generator.ts packages/client-generator-ts/tests/file-extensions.test.ts`
- direct local generator repro against the built package: without `deno.json`, generated Cloudflare imports stay bare; with `deno.json`, `models.ts` and `models/User.ts` keep explicit `.ts` suffixes

## Summary by CodeRabbit

* **New Features**
  * Improved import file extension selection for workerd targets by detecting nearby Deno configuration and adjusting generated extensions accordingly.

* **Tests**
  * Added tests covering import-extension behavior for workerd with and without a Deno config present, and tests verifying the Deno-config detection across directory layouts.

PR #29408, opened 2026-03-27
fix(client-generator-ts): type browser null singleton exports

raashish1601:
## Summary
- add explicit `typeof runtime.*` annotations to the generated browser-side `DbNull`, `JsonNull`, and `AnyNull` singleton exports
- keep the emitted declarations portable in pnpm monorepos instead of leaving declaration emit to infer names that trigger TS2742
- add a focused generator regression that checks the emitted `internal/prismaNamespaceBrowser.ts` text directly

## Testing
- `pnpm.cmd install --frozen-lockfile`
- `pnpm.cmd --filter @prisma/client-generator-ts... build`
- `pnpm.cmd --filter @prisma/client-generator-ts exec vitest run tests/generator.test.ts -t "emits portable typed null singleton exports in the browser namespace"`
- `pnpm.cmd exec eslint packages/client-generator-ts/src/TSClient/NullTypes.ts packages/client-generator-ts/tests/generator.test.ts`
- `git diff --check HEAD~1..HEAD`

Fixes #28581

## Summary by CodeRabbit

* **Refactor**
  * Enhanced TypeScript type annotations for null singleton constants for stricter typing.

* **Tests**
  * Added coverage for typed null singleton exports in browser builds.
  * Improved test cleanup to ensure generator shutdown runs even on failures.

PR #29411, opened 2026-03-28
fix: keep the prisma package root export resolvable

raashish1601:
## Summary
- write the missing `build/types.js` stub during the CLI build so the published root export path exists
- keep the root `prisma` package export pointing at a no-op module instead of aliasing it to the CLI entrypoint
- add a focused Jest regression that simulates `require.resolve(''prisma'')` from a consumer project

## Repro
`prisma@7.5.0` currently exports `./build/types.js` from the package root, but the published tarball does not include that file, so `require.resolve(''prisma'')` throws `MODULE_NOT_FOUND`.

## Validation
- `packages/cli/node_modules/.bin/jest.cmd --runInBand --testPathPattern package-exports.test.ts`
- reproduced the breakage from the published `prisma-7.5.0.tgz` tarball where `require.resolve(''prisma'')` fails on missing `build/types.js`
- packed the local CLI package after writing the stub and verified the tarball now contains `package/build/types.js`, and that an extracted consumer install resolves `prisma` successfully

## Notes
- `corepack pnpm install --frozen-lockfile` on this Windows host still hits the pre-existing `packages/engines` postinstall failure because it expects `dist/scripts/postinstall.js` during install.
- `corepack pnpm --dir packages/cli build` also remains non-diagnostic on this checkout because workspace package type declarations are not available until the broader repo build graph is prepared.

## Summary by CodeRabbit

* **New Features**
  * CLI build now generates type definitions to support CommonJS consumers.

* **Tests**
  * Added test to verify CommonJS package exports resolve correctly.


jacek-prisma:
The changes seemingly broke the tests:
https://github.com/prisma/prisma/actions/runs/23677757152/job/69174954346?pr=29411

tensordreams:
Hi @raashish1601 — thanks for this, and sorry for the long silence on our side.

We're closing out the open PRs that are still blocked on an unsigned CLA, and this is one of them: https://cla-assistant.io/prisma/prisma?pullRequest=29411

Nothing against the change itself — we just can't merge a contribution without a signed CLA. If you'd like it to land, signing there and reopening the PR is all that's needed, and one signature covers #29408 and #29409 too.

Thanks for taking the time either way.

PR #29417, opened 2026-03-31
fix(cli): move generate hint below matching client output

### LINKED ISSUE #1737: `generate`: Usage instructions should be next to generator
If you use a custom generator, the output currently looks like this:

```
✔ Generated Prisma Client to .\node_modules\@prisma\client in 287ms

✔ Generated Prisma Test Utils to .\prisma\node_modules\@generated\prisma-test-utils in 3.50s

You can now start using Prisma Client in your code:

```
// or const { PrismaClient } = require('@prisma/client')

```

Explore the full API: http://pris.ly/d/client
```

I think the output would make more sense with the "usage instructions" next to the generator that caused them:

```
✔ Generated Prisma Client to .\node_modules\@prisma\client in 287ms

You can now start using Prisma Client in your code:

```
// or const { PrismaClient } = require('@prisma/client')

```

Explore the full API: http://pris.ly/d/client

✔ Generated Prisma Test Utils to .\prisma\node_modules\@generated\prisma-test-utils in 3.50s
```

That could also lead to generators being able to add their own output.

canaokar:
## Summary

This updates `prisma generate` so the Prisma Client usage hint is printed directly below the corresponding generated Prisma
Client entry, instead of being appended after all generator output.

This makes multi-generator output easier to follow and aligns the hint with the generator that produced it.

Fixes #1737 

## Changes

- collect generator log entries before rendering the final output
- attach the Prisma Client usage hint to the first matching Prisma Client generator entry
- update `Generate` command snapshots to cover the new ordering

## Testing

- `pnpm test -- src/__tests__/commands/Generate.test.ts --runInBand`

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -35,6 +35,11 @@ import { simpleDebounce } from './utils/simpleDebounce'
 
 const pkg = eval(`require('../package.json')`)
 
+type GeneratorLogEntry = {
+  text: string
+  attachesClientHint: boolean
+}
+
 /**
  * $ prisma generate
  */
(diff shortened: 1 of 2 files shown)


canaokar:
Closing this as duplicate of #29315, which already addresses #1737

Issue #29354, opened 2026-03-18
Batch inserts error with overflow error, despite target database field being comfortably wide enough to handle value


dfoley112:
### Bug description

When using a batch insert, if some of the values of a field exceed 32bits, and some don't, Prisma will parameterize the values that are < 32bit as ints and those > 32bits as nvarchar. SQL Server will then try to cast these nvarchars to int 32, since that data type has precedence over nvarchar, leading to an "Value out of range for the type: The conversion of the nvarchar value '3850000000' overflowed an int column." error.

No combination of Prisma data types vs. db mapped types in my schema for the offending field seem to matter.

### Severity

🚨 Critical: Data loss, app crash, security issue

### Reproduction

RELEVANT SCHEMA: 
  model PolicyRiskCategory {
    PolicyRiskCategoryId        Int          @id(map: "PK__PolicyRi__7BFDC33F117DE65F") @default(autoincrement())
    PolicyId                    Int
    RiskCategoryId              Int
    SubjectPremium              Float
    MaxIndividualLoss           Float
    RateChange                  Float
    AverageUnderlyingLine       Float?

    Policy                      Policy       @relation(fields: [PolicyId], references: [PolicyId], onDelete: Cascade, onUpdate: 
  }

INSERT: 
PRISMA.policy.create({
            data: {
              .... irrelevant fields
              PolicyRiskCategory: {
                create: p.policyRiskCategories.map((rc) => ({
                  RiskCategoryId: rc.riskCategoryId,
                  SubjectPremium: rc.subjectPremium,
                  MaxIndividualLoss: rc.maxLine,
                  RateChange: rc.rateChange,
                  AverageUnderlyingLine: rc.averageUnderlyingLine,
                })),
              },
}

GENERATED SQL:
[...]

PR #29418, opened 2026-03-31
test(mssql): add reproduction test for batch insert type inconsistency

yangbobo2021:
This PR adds a reproduction test for https://github.com/prisma/prisma/issues/29354.

## Problem

When performing batch inserts with Float field values, the mssql adapter produces inconsistent parameter types depending on the value range:
- Values within int32 range → SQL \`int\`
- Values exceeding int32 range → SQL \`nvarchar\`

## Root Cause

\`packages/adapter-mssql/src/mssql.ts:56\` calls \`req.input()\` without an explicit SQL type:
\`\`\`typescript
req.input(\`P\${i + 1}\`, mapArg(query.args[i], query.argTypes[i]))
\`\`\`

This allows the mssql package to auto-detect types based on values, causing type inconsistency in batch operations.

## Test Case

The test demonstrates that with values \`[450000000, 3850000000, 1600000000]\`:
- \`450000000\` → \`int\` (within int32 range)
- \`3850000000\` → \`nvarchar\` (exceeds int32 max: 2147483647)
- \`1600000000\` → \`int\` (within int32 range)

### Execution
\`\`\`bash
export TEST_MSSQL_URI="mssql://SA:Pr1sm4_Pr1sm4@localhost:1433/master"
pnpm --filter @prisma/adapter-mssql test issue-29354-batch-insert
\`\`\`

### Suggested Fix
Add explicit SQL type to \`req.input()\`:
\`\`\`typescript
req.input(\`P\${i + 1}\`, sql.Float, mapArg(query.args[i], query.argTypes[i]))
\`\`\`

Would you like me to create a follow-up PR with the fix?

### DIFF
--- a/packages/adapter-mssql/src/issue-29354-batch-insert.test.ts
+++ b/packages/adapter-mssql/src/issue-29354-batch-insert.test.ts
@@ -1,0 +1,186 @@
+/**
+ * Issue #29354 - Batch Insert Type Inconsistency Bug
+ *
+ * BUG DESCRIPTION (from ISSUE):
+ * When using batch insert with Float field values that exceed 32-bit int range
+ * mixed with values within 32-bit range, Prisma parameterizes them differently:
+ * - Values < 32bit: parameterized as SQL int
+ * - Values > 32bit: parameterized as SQL nvarchar
+ *
+ * ROOT CAUSE (from ISSUE):
+ * In packages/adapter-mssql/src/mssql.ts line 56:
+ *   req.input(`P${i + 1}`, mapArg(query.args[i], query.argTypes[i]))
+ *
+ * The req.input() is called WITHOUT explicit SQL type, allowing mssql package
+ * to auto-detect types based on values.
+ *
+ * ISSUE SCENARIO:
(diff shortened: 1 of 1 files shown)


packages/adapter-mssql/src/issue-29354-batch-insert.test.ts
```diff
@@ -0,0 +1,186 @@
...
+      const hasInconsistentTypes = uniqueTypes.size > 1
+
+      console.log('\nResult:')
+      if (hasInconsistentTypes) {
+        console.log('  ❌ BUG CONFIRMED: Parameter types are INCONSISTENT')
+        console.log(`  ❌ Types detected: ${Array.from(uniqueTypes).join(', ')}`)
+        console.log('  ❌ This causes SQL Server type precedence issues in batch insert')
+      } else {
+        console.log('  ✅ Parameter types are consistent')
+      }
```
jacek-prisma, line 140: we should avoid spamming `console.log` from tests, this should just have an concise and meaningful failing assertion

aqrln:

PR #29420, opened 2026-03-31
test(migrate): add reproduction test for RUST_LOG dependency in migrate deploy

yangbobo2021:
This PR adds a reproduction test for a bug related to https://github.com/prisma/prisma/issues/29355.

## Problem

The \`prisma migrate deploy\` command fails when \`RUST_LOG\` is set to \`warn\` or \`error\`, but succeeds when set to \`info\` or left unset. This demonstrates that migrate deploy behavior is incorrectly affected by the RUST_LOG environment variable.

**Note:** The failure pattern discovered differs from the original issue report:
- **Issue #29355**: Default fails, \`RUST_LOG=info\` succeeds
- **Test discovered**: Default succeeds, \`RUST_LOG=warn/error\` fails

Both patterns demonstrate the same underlying problem: migrate deploy behavior is affected by RUST_LOG setting.

## Root Cause

\`packages/migrate/src/SchemaEngineCLI.ts:408-415\` allows \`processEnv\` to override the default \`RUST_LOG='info'\` setting via object spread syntax:

\`\`\`typescript
env: {
  RUST_LOG: 'info',      // Default setting
  RUST_BACKTRACE: '1',
  ...processEnv,         // User's env vars override the defaults
},
\`\`\`

## Test Results

| RUST_LOG Setting | Result |
|-----------------|--------|
| undefined (default) | ✅ Succeeds |
| \`info\` | ✅ Succeeds |
| \`warn\` | ❌ Fails with "Schema engine error:" |
| \`error\` | ❌ Fails with "Schema engine error:" |

### Execution
\`\`\`bash
GITHUB_REF_NAME=main pnpm --filter @prisma/migrate test MigrateDeploy.issue29355
\`\`\`

### Suggested Fix
Prevent \`processEnv.RUST_LOG\` from overriding the default value:
\`\`\`typescript
env: {
  RUST_LOG: 'info',
  RUST_BACKTRACE: '1',
  ...Object.fromEntries(
    Object.entries(processEnv).filter(([key]) => key !== 'RUST_LOG')
  ),
},
\`\`\`

Would you like me to create a follow-up PR with the fix?


packages/migrate/src/__tests__/MigrateDeploy.issue29355.test.ts
```diff
@@ -0,0 +1,233 @@
...
+      error2 = e as Error
+      result2 = (e as Error).message
+    }
+    const dbCreated2 = fs.exists(dbPath) === 'file'
+    results.push({ setting: 'info', result: result2, error: error2, dbCreated: dbCreated2 })
+
+    // Clean up database for next test
+    if (fs.exists(dbPath) === 'file') {
+      fs.remove(dbPath)
+    }
```
jacek-prisma, line 142: I think this section of the test is unnecessary

packages/migrate/src/__tests__/MigrateDeploy.issue29355.test.ts
```diff
@@ -0,0 +1,233 @@
...
+      return
+    }
+
+    // If some failed and some succeeded, we have reproduced the BUG
+    if (failedResults.length > 0 && succeededResults.length > 0) {
+      const failedSettings = failedResults.map((r) => r.setting).join(', ')
+      const succeededSettings = succeededResults.map((r) => r.setting).join(', ')
+
+      // This is a BUG related to ISSUE #29355, but with a different failure pattern
+      throw new Error(
```
jacek-prisma, line 202: We should try to structure the tests so that there is a failing assertion rather than an exception being thrown

aqrln:

Issue #29263, opened 2026-02-27
Prisma 7.4 Partial Indexes are bugged during migrations


brian-lou:
### Bug description

I'm using "partialIndexes" as previewFeatures in my prisma schema, and have a few postgres partial indexes.

For example,
```
  @@unique([project_id], where: { default_branch: true }, map: "unique_default_branch_per_project")
```

I've already applied the corresponding migration to my db; however, when I try to run any `prisma migrate dev --name "some_other_migration"`, prisma will also drop and re-generate the corresponding partial indexes on the db:
```
/*
  Warnings:

  - A unique constraint covering the columns `[project_id]` on the table `branches` will be added. If there are existing duplicate values, this will fail.

*/
-- DropIndex
DROP INDEX "unique_default_branch_per_project";

-- CreateIndex
CREATE UNIQUE INDEX "unique_default_branch_per_project" ON "branches"("project_id") WHERE ("default_branch" = true);

```
This happens every time I run `prisma migrate dev --name`

### Severity

🔹 Minor: Unexpected behavior, but does not block development

### Reproduction

Create a partial index on a table, and then run `prisma migrate dev --name <any_string>` a few times.

### Expected vs. Actual Behavior

Duplicate migrations should not be created.

### Frequency

Consistently reproducible

### Does this occur in development or production?

Only in development (e.g., CLI tools, migrations, Prisma Studio)

### Is this a regression?

Kind of - previous versions didn't support previewFeatures

### Workaround

No

### Prisma Schema & Queries

```prisma
  @@unique([project_id], where: { default_branch: true }, map: "unique_default_branch_per_project")
```

### Prisma Config

```ts
generator client {
  provider = "prisma-client"
  output   = "../../src/generated"
  previewFeatures = ["typedSql", "partialIndexes"]
  moduleFormat = "esm"
  compilerBuild = "fast" // default to fast
}
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** macOs
- **Database:** PostgreSQL
- **Node.js version:** v24.13.0

### Prisma Version

```
7.4.1
```

jacek-prisma:
This might already be fixed on the dev version, `7.5.0-dev.28`

harrisoncramer:
Hey @jacek-prisma I'm running that version and still unable to get Prisma to recognize that the partial indexes have already been applied.

[...]

Issue #5795, opened 2021-02-23
PrismaClient is unable to be run in the browser.


pramit-marattha:
## Bug description

![image](https://user-images.githubusercontent.com/37651620/108820687-a9ff3480-75e4-11eb-9c84-b1de6c96ec79.png)
![image](https://user-images.githubusercontent.com/37651620/108820722-b5eaf680-75e4-11eb-8eb9-0dbc0471aa6e.png)

## Expected behavior
A user authentication page should be displayed.

## Prisma information
![image](https://user-images.githubusercontent.com/37651620/108821227-63f6a080-75e5-11eb-8ea8-4ba5efb67afb.png)
### schema.prisma
![image](https://user-images.githubusercontent.com/37651620/108821298-7c66bb00-75e5-11eb-88a4-2557e1f26a0e.png)
![image](https://user-images.githubusercontent.com/37651620/108821337-88527d00-75e5-11eb-94a6-e101283e0372.png)

## Environment & setup

- OS: windows 10  v1909
![image](https://user-images.githubusercontent.com/37651620/108822071-a8cf0700-75e6-11eb-9c4b-903c351b5756.png)

- Database: Heroku postgresql 
![image](https://user-images.githubusercontent.com/37651620/108821744-2e9e8280-75e6-11eb-8a08-400187a5e735.png)

- Node.js version: v 15.4.0
![image](https://user-images.githubusercontent.com/37651620/108822124-bc7a6d80-75e6-11eb-936e-1a86937aaced.png)

- Prisma version: 2.17.0
![image](https://user-images.githubusercontent.com/37651620/108821949-7a512c00-75e6-11eb-9c6c-adfc6f135596.png)

```

```

matthewmueller:
Hey there, could you post the Next.js code where this is happening? It's true that Prisma can't run in the browser, but we have documentation on how to make it work with next.js

https://www.prisma.io/nextjs

pantharshit00:
@pramit-marattha 

Please paste the code instead of the screenshots. We can't copy and debug things if you don't paste it because it will waste a lot of time if will copy it from the screenshot. 

Ideally share a reproduction inside of a git repository.

pantharshit00:
I am going to close this as its been 3 months now and there is no activity here. 

Please reply if you want us to take another look.

ninest:
I'm not sure if anyone is still facing this issue, but I solved it by initializing prisma once and on the **server only**.


```ts
[...]

Issue #5792, opened 2021-02-22
Performance issues after 2.14


GCastilho:
## Bug description

After upgrading prisma do a version higher than 2.14 I'm experiencing performance issues with the queries. Mocha reports tests that usually took less than 60ms to now be taking more than 150ms, sometimes more than 200ms to complete

I commented about this in #5485 and analyzed the logs, but they're identical (or very similar) so it does not seems like it's the same as #5274

## Expected behavior

Average performance should remain the same or improve between Prisma versions.

## Prisma information

I've picked some queries to show comparing 2.14 with 2.15 and 2.17, but they were _exactly_ the same, the only difference was the time it took to run them (54ms -> 141ms, 65ms -> 193ms) so I don't thing it make sense to show them (since they're the same). But it's not a problem to share that part of the schema and queries if needed

## Environment & setup
- OS: Debian 10 buster x64
- Database: PostgreSQL 12 (localhost)
- Node.js version: v14.15.5
- Prisma version: >= 2.15.0

Please let me know if you need more information or tests that you need. Thanks!

timsuchanek:
Thanks for reporting @GCastilho!
Could you share the specific query with us plus a subset of the schema that got slower?

GCastilho:
For sure, @timsuchanek 

This is the schema:

```prisma
model User {
  @@map("users")
  id              String   @id @default(uuid())
  email           String   @unique
  password        String
  role            UserRole @default(USER)
  created_at      DateTime @default(now())
  updated_at      DateTime @updatedAt
  lotes           Lote[]
  criadores       Criador[]
  granjas         Granja[]
}

model Criador {
  @@map("criadores")
  id          Int
  userId      String
  user        User      @relation(fields: [userId])
  nome        String
  telefone    String
  created_at  DateTime  @default(now())
  updated_at  DateTime  @updatedAt
  granjas     Granja[]
  lotes       Lote[]
  @@id([userId, id])
}
```

And the queries logged by prisma:

From the first test, prisma@2.15:

```sql
prisma:info  Starting a postgresql pool with 5 connections.
prisma:info  Fetched a connection from the pool
prisma:info  Started http server on http://127.0.0.1:34325
prisma:info  Fetched a connection from the pool
prisma:query DELETE FROM users
prisma:info  Fetched a connection from the pool
prisma:query BEGIN
[...]

Issue #5790, opened 2021-02-22
Rust client


ArcticSpaceFox:
Currently there are only Js, Ts and Go clients. Can we have a rust version as well?

pantharshit00:
Checkout our SQL abstraction for Rust for now: https://github.com/prisma/quaint/

It is more primitive than prisma but it might help you in your projects if you don't want to use SQLx

ArcticSpaceFox:
Thanks I just love prisma sooo much, the schema and everything just makes db stuff so easy. Just want it for rust, thanks for the tip. I will keep an eye on this issue tho 😄👍

auderer:
Prisma would be amazing for Rust. Are there any plans to build this out in the future?

Milo123459:
Any updates?

fivethreeo:
There is the unofficial https://github.com/seunlanlege/prisma-client-rs

Brendonovich:
There is also my library [prisma-client-rust](https://github.com/Brendonovich/prisma-client-rust) which has an alternative syntax to [prisma-client-rs](https://github.com/seunlanlege/prisma-client-rs)

merelinguist:
It seems there is now no longer a Go client either https://github.com/prisma/prisma-client-go/issues/707#issuecomment-1000239342

VaZark:
With the team currently focussing exclusively on the JS/TS environment, I don't see this happening anytime soon. 

That being said, I'd love to see if it'd possible to have a primitive rust crate that handles migrations and re-exposes quaint as a querybuilder for the moment.

JonathanWilbur:
One thing I just thought of that I wanted to share: I know the Prisma engine itself is written in Rust. Would there be any way to just support rust clients by binding directly / statically linking the Prisma engine? Forgive me if that's a stupid question. I only barely understand that the TS client sends GraphQL queries to the engine. It just seems like this could be bypassed entirely by a rust client.

JonathanWilbur:
Not to mention, if a single compiled programming language is supported, it would be easier for all other compiled languages (including Go) to just link to it as well, so you'd kind of have support for many languages all for the cost of one.

jkomyno:
Hi there,

As part of our effort to clean up our GitHub repo, we’re closing this issue to help focus on features and improvements that align with our focus areas and those of our community.

If you want the familiar Prisma experience in Rust, I'd recomment checking out the third-party projects that were mentioned above.

---

[...]

PR #29421, opened 2026-03-31
test(migrate): add reproduction test for varchar partial unique index migration loop

yangbobo2021:
This PR adds a reproduction test for https://github.com/prisma/prisma/issues/29386.

## Problem

When using \`@db.VarChar\` with a partial unique index using a negation predicate
(\`where: { status: { not: "value" } }\`), \`prisma migrate dev\` incorrectly
generates the same DROP + CREATE migration on every run.

## Root Cause

PostgreSQL stores index predicates with \`::text\` casts and \`<>\` operators,
while Prisma generates them without:

| Prisma generates | PostgreSQL stores |
|-----------------|-------------------|
| \`("status" != 'superseded')\` | \`(status)::text <> 'superseded'::text\` |

The schema differ in \`prisma-engines/migration-engine/src/schema_differ.rs\`
doesn't recognize these as semantically equivalent, causing spurious migrations.

This is similar to the fix for #29263 (equality predicates), but affects
negation predicates with VarChar columns.

## Test Results

Running \`migrate dev\` twice:
- **First run**: Creates index successfully
- **Second run**: BUG - Creates spurious DROP + CREATE migration

### Execution
\`\`\`bash
TEST_POSTGRES_URI_MIGRATE="postgres://prisma:prisma@localhost:5432/tests-migrate" \\
TEST_POSTGRES_SHADOWDB_URI_MIGRATE="postgres://prisma:prisma@localhost:5432/tests-migrate-shadowdb" \\
pnpm --filter @prisma/migrate run test -- src/__tests__/MigrateDev.issue29386.test.ts
\`\`\`

### Schema
\`\`\`prisma
model Document {
  id      String @id @default(uuid())
  groupId String
  status  String @db.VarChar(20)

  @@unique([groupId], where: { status: { not: "superseded" } })
}
\`\`\`

### Suggested Fix
In \`prisma-engines/migration-engine/src/schema_differ.rs\`, update
\`pg_predicates_semantically_equal()\` to handle:
- PostgreSQL's automatic \`::text\` casts for varchar columns
- Operator normalization (\`!=\` vs \`<>\`)

This is similar to the fix for #29263 (pr-engines #5788, #5790, #5795, #5792).

Would you like me to create a follow-up PR with the fix in prisma-engines?

### DIFF
--- a/packages/migrate/src/__tests__/MigrateDev.issue29386.test.ts
+++ b/packages/migrate/src/__tests__/MigrateDev.issue29386.test.ts
@@ -1,0 +1,133 @@
+/**
+ * Issue #29386: prisma migrate infinite generation loop with varchar partial unique index
+ *
+ * BUG SUMMARY:
+ * When using @db.VarChar type with @@unique partial unique constraints and a
+ * negation predicate (where: { status: { not: "superseded" } }), prisma migrate
+ * dev generates a spurious DROP + CREATE migration on EVERY run.
+ *
+ * ROOT CAUSE:
+ * PostgreSQL normalizes the WHERE clause of a partial index on a varchar column
+ * by adding ::text casts:
+ * - Prisma generates: ("status" != 'superseded')
+ * - PostgreSQL stores: (status)::text <> 'superseded'::text
+ *
+ * The schema differ doesn't account for these normalizations, causing it to
+ * detect a "difference" and generate a spurious migration on every run.
+ *
+ * This test reproduces the bug by running migrate dev twice and verifying that
+ * the second run should NOT generate a new migration (but the bug causes it to).
+ */
+
+import fs from 'fs-jetpack'
+import path from 'path'
+import prompt from 'prompts'
+
+import { MigrateDev } from '../commands/MigrateDev'
+import { setupPostgres, tearDownPostgres } from '../utils/setupPostgres'
+import { createDefaultTestContext } from './__helpers__/context'
+
+const testIf = (condition: boolean) => (condition ? test : test.skip)
+
+const ctx = createDefaultTestContext()
+
+describe('Issue #29386: Varchar Partial Unique Index Migration Loop', () => {
+  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


packages/migrate/src/__tests__/MigrateDev.issue29386.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        if (migrationDirsAfterSecond!.length > initialMigrationDirs!.length) {
+          // BUG REPRODUCED: A new migration was created when it shouldn't have been
+          const newMigrationDir = migrationDirsAfterSecond![migrationDirsAfterSecond!.length - 1]
+          const newMigrationPath = path.join(migrationsDir, newMigrationDir, 'migration.sql')
+          const newMigrationContent = fs.read(newMigrationPath)
+
+          // The spurious migration should contain DROP + CREATE for the same index
+          expect(newMigrationContent).toContain('DROP INDEX')
+          expect(newMigrationContent).toContain('CREATE UNIQUE INDEX')
```
jacek-prisma, line 112: The test should expect the correct behavior here, not the bug (rather than throwing an error later)

packages/migrate/src/__tests__/MigrateDev.issue29386.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        // Read the first migration to verify it contains the expected index
+        const firstMigrationContent = fs.read(firstMigrationPath)
+        expect(firstMigrationContent).toContain('CREATE UNIQUE INDEX')
+        expect(firstMigrationContent).toContain('Document_groupId_key')
+        expect(firstMigrationContent).toMatch(/WHERE.*status.*!=.*superseded/)
+
+        // Ensure CI mode is still enabled
+        process.env.GITHUB_ACTIONS = '1'
+        process.env.CI = '1'
```
jacek-prisma, line 87: this should be properly mocked/unmocked to avoid polluting other tests

aqrln:

PR #29441, opened 2026-04-07
fix: fix stack trace capture in PrismaClientInitializationError

ChALkeR:
It passed the constructor function instead of `this` as the first argument, setting the `.stack` property on the class itself rather than on the error instance

That said, I'm unsure what that served for, apparently it wasn't needed as it was unnoticed

This is how it behaved:
```console
> x = new PrismaClientInitializationError()
...
> x.stack
'PrismaClientInitializationError\n' +
  '    at REPL31:1:3\n' +
...
> PrismaClientInitializationError.stack
'PrismaClientInitializationError\n' +
  '    at new PrismaClientInitializationError (REPL29:8:11)\n' +
  '    at REPL31:1:3\n' +
...
```

PR #29442, opened 2026-04-07
fix: fix condition check for log level precedence

ChALkeR:
From `const currentLevel = typeof curr === 'string' ? curr : curr.level`, curr could be an object `{ level: 'info' }`.

Comparing it to string seems like a mistake.

### DIFF
--- a/packages/client/src/runtime/getLogLevel.ts
+++ b/packages/client/src/runtime/getLogLevel.ts
@@ -16,7 +16,7 @@ export function getLogLevel(log: LogLevel | Array<LogLevel | LogDefinition>): Lo
     if (!acc) {
       return currentLevel
     }
-    if (curr === 'info' || acc === 'info') {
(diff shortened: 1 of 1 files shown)


aqrln (approved):

aqrln:
@ChALkeR thanks you! Could you please [sign the CLA](https://github.com/prisma/prisma/pull/29442#issuecomment-4195978613)?

tensordreams:
Hi @ChALkeR — thanks again for this, and sorry for how long it sat.

We're closing out the open PRs that are still blocked on an unsigned CLA, and this is one of them: https://cla-assistant.io/prisma/prisma?pullRequest=29442

This isn't a judgement on the change itself — we simply can't merge a contribution we don't have a CLA for. If you'd still like it to land, signing there and reopening this PR is all it takes, and one signature covers #29441 as well. We'd be glad to pick it back up.

Thanks for taking the time to send it either way.

PR #29445, opened 2026-04-07
fix(build): use require.resolve to find Wasm files

lazuee:
When using pnpm, dependencies can get hoisted and won't always show up in local `node_modules`. The build scripts were using hardcoded paths to find Wasm files, which broke in these setups. Using `require.resolve(...)` lets NodeJS find the files on its own, so the build works regardless of how the package manager structures the dependencies.

**Changes:**
- Updated Wasm file lookups in `packages/cli/helpers/build.ts` and `packages/internals/helpers/build.ts` to use `require.resolve` instead of hardcoded paths
- Cleaned up old path variables that were no longer doing anything useful
- Noted the `require.resolve` pattern in `AGENTS.md` so future contributors know why it's done this way


jacek-prisma:
Hi @lazuee, thank you for the contribution!
Before I merge this, can you provide an example where the current `main` code doesn't work while this does?

lazuee:
@jacek-prisma 
In this screenshot, I’ve taken a fresh clone at the [latest current commit](https://github.com/prisma/prisma/commit/c10c58747cfba2ae3dbef35872bc6a9b6813e21c), I ran: `pnpm install && pnpm run dev`
Before that, I also cleared the cache with: `pnpm cache delete && pnpm store prune --force`

I get an error and it seems like indicating that the WASM files cannot be found.

Before this PR

But after merging this PR, the error is resolved, as the PR ensures the WASM files are found by using `require.resolve()`:

After merging

jacek-prisma:
Hi @lazuee, is there something specific to your setup? I don't get the same error on my machine

lazuee:
Hi @jacek-prisma This turned out to be on pnpm’s side. Because pnpm uses a [symlinked `node_modules` structure](https://pnpm.io/symlinked-node-modules-structure), the WASM file path was being resolved to `~/Documents/Projects/prismas/packages/internals/node_modules/@prisma/schema-engine-wasm/schema_engine_bg.wasm`, which doesn’t actually exist in this setup.
not-exist

So, this PR switches to using `require.resolve` to locate the WASM file via the package path, so it now resolves to the real file path: `~/Documents/Projects/prismas/node_modules/@prisma/schema-engine-wasm/schema_engine_bg.wasm`.
real-path

tensordreams:
Merged latest `main` into the branch (no conflicts) and verified the fix by building the affected packages. `pnpm --filter @prisma/internals build` and `pnpm --filter prisma build` both succeed; the CLI build exercises all three `require.resolve`-based Wasm paths and copies `schema_engine_bg.wasm`, `prisma_schema_build_bg.wasm`, and all 10 `query_compiler_*` variants into `packages/cli/build/`. prettier and eslint are clean on the changed files.

aqrln (approved):
I'm not sure if we necessarily want to support non-standard pnpm configs but I think this is an improvement regardless and looks cleaner, so I'll merge it. Thank you!

PR #29455, opened 2026-04-10
fix(migrate): remove dead --shadow-database-url reference in MigrateDiff

### LINKED ISSUE #29454: migrate diff error message references removed --shadow-database-url CLI flag
### Bug description

When running `prisma migrate diff --from-migrations` without a shadow database URL configured, the error message suggests passing a `--shadow-database-url` CLI flag that **was removed in Prisma 7** (PR #28432, commit `96d81975`):

```
Error: You must pass the `--shadow-database-url` flag or set `datasource.shadowDatabaseUrl`
in your `prisma.config.ts` if you want to diff a migrations directory.
```

Passing `--shadow-database-url` causes an "unknown flag" error because the flag is no longer in the arg spec. Only the second option (`datasource.shadowDatabaseUrl` in `prisma.config.ts`) works.

Additionally, `MigrateDiff.ts` line ~286 contains dead code: `shadowDatabaseUrl: args['--shadow-database-url'] ?? null` — this always evaluates to `null` since the flag was removed from the arg spec.

### Severity

### Reproduction

```bash
# 1. Create a project with prisma.config.ts that has NO shadowDatabaseUrl
# prisma.config.ts:
# export default defineConfig({
#   datasource: { url: process.env['DATABASE_URL'] },
# })

# 2. Run migrate diff with --from-migrations
npx prisma migrate diff --from-migrations ./prisma/migrations --to-schema ./prisma/schema.prisma --exit-code
# → Error suggests --shadow-database-url flag

# 3. Try to follow the error message's suggestion
npx prisma migrate diff --from-migrations ./prisma/migrations --to-schema ./prisma/schema.prisma --shadow-database-url "postgresql://user:pass@localhost:5432/shadow" --exit-code
# → Shows usage/help and exits with code 1 (unknown flag)
```

**Expected:** The error message should only suggest the working option:
```
Error: Set `datasource.shadowDatabaseUrl` in your `prisma.config.ts` if you want to diff
a migrations directory.
```

**Actual:** The error message suggests `--shadow-database-url` CLI flag which was removed in Prisma 7 and causes a separate error when used.

### Frequency

Consistently reproducible

### Is this a regression?

Yes. In Prisma 6, `--shadow-database-url` was a valid CLI flag. It was removed in Prisma 7 via PR #28432, but the Rust engine error message (`schema-engine/core/src/commands/diff_cli.rs`) was not updated to reflect this removal.

### Workaround

Set `datasource.shadowDatabaseUrl` in `prisma.config.ts` instead of using the CLI flag:

```ts
export default defineConfig({
  datasource: {
    url: process.env['DATABASE_URL'],
    shadowDatabaseUrl: process.env['SHADOW_DATABASE_URL'],
  },
});
```

### Prisma Schema & Queries

```prisma
// Not schema-specific — affects any project using prisma migrate diff --from-migrations
```
```ts
// No Prisma Client query involved — this is a CLI issue
```

### Prisma Config

```ts
import 'dotenv/config';

export default defineConfig({
  datasource: {
    url: process.env['DATABASE_URL'],
    // shadowDatabaseUrl is NOT set — triggers the misleading error
  },
});
```

### Logs & Debug Info

```
# Without the flag:
$ npx prisma migrate diff --from-migrations ./prisma/migrations --to-schema ./prisma/schema.prisma --exit-code
Loaded Prisma config from prisma.config.ts.
Error: You must pass the `--shadow-database-url` flag or set `datasource.shadowDatabaseUrl` in your `prisma.config.ts` if you want to diff a migrations directory.

# With the flag (following the error's advice):
$ npx prisma migrate diff --from-migrations ./prisma/migrations --to-schema ./prisma/schema.prisma --shadow-database-url "postgresql://..." --exit-code
Usage
  $ prisma migrate diff [options]
  ...
(exits with code 1 — flag not recognized)
```

### Environment & Setup

- **OS:** macOS 15.4 (darwin arm64) / Linux Ubuntu 24.04 (GitHub Actions self-hosted runner)
- **Database:** PostgreSQL 17
- **Node.js version:** v22.16.0

### Prisma Version

```
prisma               : 7.3.0
@prisma/client       : 7.3.0
Operating System     : darwin / linux
Architecture         : arm64 / x64
Node.js              : v22.16.0
TypeScript           : 5.9.3
Schema Engine        : schema-engine-cli 9d6ad21cbbceab97458517b147a6a09ff43aa735
Default Engines Hash : 9d6ad21cbbceab97458517b147a6a09ff43aa735
```

happinessee:
## Changes

**Commit 1** — Remove dead code referencing the removed `--shadow-database-url` CLI flag in `packages/migrate/src/commands/MigrateDiff.ts`.

**Before (line 286):**
```ts
shadowDatabaseUrl: args['--shadow-database-url'] ?? null,
```

**After (interim):**
```ts
shadowDatabaseUrl: null,
```

**Commit 2** — Following reviewer feedback, completely remove the deprecated `shadowDatabaseUrl` field from `MigrateDiffInput`:

- Removed `shadowDatabaseUrl: string | null` from `MigrateDiffInput` interface in `types.ts`
- Removed the `shadowDatabaseUrl: null` line from the `migrate.engine.migrateDiff()` call

The matching `DiffParams.shadow_database_url` field has also been removed from the schema engine in the companion PR: prisma/prisma-engines#5803.

## Context

The `--shadow-database-url` flag was removed from the arg spec in #28432 (Prisma 7 breaking changes), but:
1. Line 286 still referenced `args['--shadow-database-url']` which always evaluated to `undefined` → `null` via `??`
2. The `shadowDatabaseUrl` field in `MigrateDiffInput` was still passed to the engine (as `null`)

Other removed flags (`--from-url`, `--to-url`, etc.) were properly cleaned up in #28432, but `shadowDatabaseUrl` was missed in both the function body and the type definition.

Shadow database URL is now exclusively configured via `datasource.shadowDatabaseUrl` in `prisma.config.ts`. Tests using `ctx.setDatasource({ url, shadowDatabaseUrl })` and `shadowdb.config.ts` fixtures are unaffected — those represent the `datasource` configuration path, which is the correct mechanism.

Fixes #29454


jacek-prisma (approved):
The changes look good, although we could go even further and completely remove `shadowDatabaseUrl` from the diff parameters on both sides, since it no longer has any function

happinessee:
@jacek-prisma 
I'll check on that and get back to you. Thank you for your suggestion!

afurm:
Removing dead code is always good. Quick question - is this a breaking change for anyone still passing --shadow-database-url? Should we at least warn users if they provide it? Would help with migration instead of silently ignoring.

happinessee:
> Removing dead code is always good. Quick question - is this a breaking change for anyone still passing --shadow-database-url? Should we at least warn users if they provide it? Would help with migration instead of silently ignoring.

 Thanks for the great question!

  This is not a new breaking change. The --shadow-database-url flag was already removed from the CLI arg spec in
   PR #28432 (Prisma 7). Anyone passing it today gets an "unknown flag" error at the CLI parser level — it never
   reaches the engine. So the break already happened when users upgraded to Prisma 7.

  Our PRs only clean up what was left behind after that removal:
  1. The misleading error message that still suggested using the flag (prisma-engines#5803)
  2. The dead code in both TypeScript and Rust that referenced the flag but never received a value
  (prisma#29455)

  Adding a deprecation warning wouldn't help here either, since the CLI parser rejects the flag before any
  application code runs.

Issue #29458, opened 2026-04-11
Could not load query compiler bg from postgresql


Caioohv:
### Bug description

after executing `npx prisma generate`, it creates the `prisma/generated/internal/class.ts` which has this code:

```ts
getRuntime: async () =>
    await import("@prisma/client/runtime/query_compiler_bg.postgresql.mjs"),

  getQueryCompilerWasmModule: async () => {
    const { wasm } =
      await import("@prisma/client/runtime/query_compiler_bg.postgresql.wasm-base64.mjs");
    return await decodeBase64AsWasm(wasm);
  },
```

I am creating a Nuxt 4 app using Prisma 7.7.0
When running the project, i get this log:

```
[nitro 12:40:22 AM]  ERROR  Error: Could not load /home/viier/vps/apps/template/node_modules/@prisma/client/runtime/query_compiler_bg.postgresql.wasm-base64.mjs (imported by prisma/generated/internal/class.ts): ENOENT: no such file or directory, open '/home/viier/vps/apps/template/node_modules/@prisma/client/runtime/query_compiler_bg.postgresql.wasm-base64.mjs'
```

I found out that these two files does not exist, but there are files with similar names, with the postfix 'fast':
`@prisma/client/runtime/query_compiler_fast_bg.postgresql.mjs` and `@prisma/client/runtime/query_compiler_fast_bg.postgresql.wasm-base64.mjs`

So, i manually changed this and it worked.

### Severity

🚨 Critical: Data loss, app crash, security issue

### Reproduction

Create Nuxt project using `prisma bootstrap`
Set up env values
Execute `npm install && npx prisma generate`
Then, `npm run dev`

The error should appear.

### Expected vs. Actual Behavior

prisma generate should create a functional client

it imports wrong files

### Frequency

Consistently reproducible

### Does this occur in development or production?

Both development and production

### Is this a regression?

Dont know

### Workaround

I found out that these two files does not exist, but there are files with similar names, with the postfix 'fast':
`@prisma/client/runtime/query_compiler_fast_bg.postgresql.mjs` and `@prisma/client/runtime/query_compiler_fast_bg.postgresql.wasm-base64.mjs`

So, i manually changed this and it worked.

### Prisma Schema & Queries

```prisma
// Add relevant schema.prisma snippet
```
```ts
// Add relevant Prisma Client query
```

generator client {
  provider   = "prisma-client"
  output     = "./generated"
}

datasource db {
  provider = "postgresql"
}

model Visita {
[...]

PR #29459, opened 2026-04-11
fix(generator): uses small or fast packages dinamically

Caioohv:
issue #29458


packages/client-generator-ts/src/utils/wasm.ts
```diff
@@ -155,3 +155,7 @@ export function buildWasmFileMap({ activeProvider, runtimeName, compilerBuild }:
 
   return fileMap
 }
+
+function normalizeCompilerBuild(compilerBuild: string): 'fast' | 'small' {
```
jacek-prisma, line 159: the type of `compilerBuild` in `BuildWasmModuleOptions` is `'fast' | 'small'`, so this function is never going to do anything other than just forwarding its input, since the argument can never hold anything other than `small` or `fast`

PR #29468, opened 2026-04-15
fix(adapter-pg): serialize queries within a transaction

matingathani:
## Problem

When a `update` (or any write operation) is executed with `include` relations, the query plan produces a `join` node whose children are fetched via `Promise.all`. Each child eventually calls `queryRaw` on `context.queryable`. Inside a transaction, `context.queryable` is a `PgTransaction` backed by a single `pg.PoolClient` — not a pool. Dispatching concurrent `client.query()` calls on a single pg Client triggers:

```
DeprecationWarning: Calling client.query() when the client is already executing
a query is deprecated and will be removed in pg@9.0. Use async/await or an
external async flow control mechanism instead.
```

**This becomes a hard error in `pg@9.0`.** Users are already hitting it in Prisma 7.5 (since `pg` 8.20.0 introduced the warning). The only current workaround is to downgrade `pg` to 8.18.

Reported in #29407.

## Root cause

`query-interpreter.ts`, `case 'join'` (line ~243):

```typescript
const children = await Promise.all(
  node.args.children.map(async (joinExpr) => ({
    joinExpr,
    childRecords: (await this.interpretNode(joinExpr.child, context)).value,
  })),
)
```

`Promise.all` fires all child queries concurrently. Fine when `context.queryable` is a `Pool` (each query gets its own connection), but not when it is a `Transaction` (single connection).

## Fix

Detect whether `context.queryable` is a `Transaction` via the `commit` property (present on the `Transaction` interface from `@prisma/driver-adapter-utils`). When inside a transaction, replace `Promise.all` with a sequential `for...of` loop. When using a pool the existing parallel behaviour is unchanged.

```typescript
function isTransaction(queryable: SqlQueryable): queryable is Transaction {
  return 'commit' in queryable
}
```

## Test plan
- [ ] `update` with multiple `include` relations inside `$transaction([...])` no longer emits the pg deprecation warning
- [ ] `update` with multiple `include` relations outside a transaction continues to fire relation queries in parallel (no performance regression)
- [ ] Existing tests pass

Fixes #29407


jacek-prisma:
Since this is a `pg` specific issue, I think it might be better to address this in the `pg` adapter, for example by wrapping the client with a mutex like we do in some other adapters

matingathani:
Thanks for the direction — agreed, the adapter is the right place for this.

I looked at how `adapter-planetscale` and `adapter-mssql` handle it: both override `performIO` in their `Transaction` subclass and gate each call behind a `#mutex = new Mutex()` from `async-mutex`. The `PgTransaction` class in `adapter-pg/src/pg.ts` already extends `PgQueryable<TransactionClient>` and overrides `commit`/`rollback`, so the same pattern fits naturally — just add the mutex field and override `performIO` to acquire it before delegating to `super.performIO`.

That approach is strictly better than what I did in the runtime: the runtime doesn't need to know whether a queryable can handle concurrent calls, and keeping it pg-specific means other adapters are unaffected. I'll update the PR to revert the `query-interpreter.ts` change and apply the mutex in `PgTransaction` instead.

matingathani:
@jacek-prisma could you approve the CI run when you get a chance? The workflows are blocked waiting for maintainer approval.

matingathani:
Addressed in the latest commit (7765742) — replaced the manual `acquire()`/try/finally/`release()` pattern with `runExclusive(() => super.performIO(query))` as suggested. Serialisation behaviour is identical; `runExclusive` handles acquire and release internally, removing the boilerplate.

jacek-prisma:
Looks good, but I think the [4a9f3df](https://github.com/prisma/prisma/pull/29468/commits/4a9f3dffab43e9f95c5cec72eac4f7df42a34c8b) commit is unrelated and shouldn't be in this PR

matingathani:
Good catch — that commit was unintentionally included. I'll squash it out in the next push.

matingathani:
Done — removed the unrelated commit. The branch now only contains the mutex-based serialization fix.

tensordreams:
Merged latest `main` and normalized the lockfile: the branch had carried unrelated lockfile churn (TypeScript/turbo/globals/engines-version downgrades and a stray `tslib` bump). `pnpm-lock.yaml` now differs from `main` by only the `async-mutex@0.5.0` entry in the `packages/adapter-pg` importer block.

The mutex is scoped to `PgTransaction` only — the standard pooled client (`PrismaPgAdapter`) is unaffected — and `runExclusive` releases the lock even when a query throws. Added test coverage for both serialization and mutex-release-on-error. `pnpm --filter @prisma/adapter-pg test` passes (53 tests), prettier and eslint clean on the changed files.

leoamato10:
We're hitting this too, and can add two data points that widen the scope of the bug.

**1. It's not limited to `update` with multiple includes — batch `$transaction` triggers it as well.**

This fires the warning on every call:

```ts
const [courses, totalCount] = await prisma.$transaction([
  prisma.course.findMany({
    where,
    include: {
      teacher: { select: { name: true } },
      schedules: true,
      coursePriceAdjustment: {   // nested include with per-parent take
        include: { adjustment: true },
        orderBy: { adjustment: { appliedAt: "desc" } },
        take: 1,
      },
    },
    skip,
    take: pageSize,
  }),
  prisma.course.count({ where }),
]);
```

Captured with `node --trace-deprecation`:

```
DeprecationWarning: Calling client.query() when the client is already executing a query is deprecated and will be removed in pg@9.0.
    at Client.query (node_modules/pg/lib/client.js:715:7)
    at PgTransaction.performIO (@prisma/adapter-pg/dist/index.mjs)
    at PgTransaction.queryRaw (@prisma/adapter-pg/dist/index.mjs)
    at <query interpreter> (@prisma/client/runtime/client.js)
    at Array.map (<anonymous>)
    at e.interpretNode (@prisma/client/runtime/client.js)
```

The `Array.map` inside `interpretNode` dispatches the relation sub-queries concurrently onto the single `pg` Client held by `PgTransaction`, exactly as the OP suspected. Since paginated `findMany` + `count` inside a batch transaction is a very common pattern, this likely affects far more users than the `update`-with-includes case in the original report.

**2. The "downgrade pg" workaround doesn't work.** The deprecation warning is not new in pg 8.20 — we reproduced it on `pg@8.19.0` as well (same message, with the old "asycn" typo). Pinning pg only trades warning text; the deprecated internal queueing is still what keeps things working, and that's what goes away in pg@9.

**Confirming the fix in #29468 works:** we applied an equivalent local patch via `pnpm patch` — serializing `performIO` on `PgTransaction` with a promise chain (same idea as the mutex in the PR, and as what `adapter-mssql` / `adapter-planetscale` already do):

```js
// PgTransaction
constructor(...) { ...; this.queryChain = Promise.resolve(); }
async performIO(query) {
  const result = this.queryChain.then(() => super.performIO(query));
  this.queryChain = result.then(() => {}, () => {});
  return result;
}
```

With this in place the warning is gone across our full app (verified with `--trace-deprecation` while exercising transactions with nested includes, batch transactions, and interactive transactions). No perf downside in practice — Postgres executes queries on a single connection sequentially anyway, so pg was already queueing them internally.

Environment: `prisma@7.9.0`, `@prisma/adapter-pg@7.9.0`, `pg@8.22.0` (also repro'd on 8.19.0), Node 24, PostgreSQL 17.

It would be great to see #29468 land — happy to test a dev release.

packages/adapter-pg/src/pg.ts
```diff
@@ -132,6 +133,10 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
 }
 
 class PgTransaction extends PgQueryable<TransactionClient> implements Transaction {
+  // pg.PoolClient does not support concurrent queries on the same connection,
+  // so we serialize all performIO calls with a mutex.
+  #mutex = new Mutex()
```
aqrln, line 138: I think we should move the mutex to the base `PgQueryable` and do the serialization logic there instead of overriding `performIO` here.

As the comment says, `pg.PoolClient` doesn't allow this generally, doesn't matter if it's within a transaction or not. Batch transactions are just how you happen to be able to observe the issue right now in practice but the issue is more general.

gadcam:
Hey @matingathani & @tensordreams 
Thanks for the work on this, and sorry to step on your toes: this is also showing up in my logs and this lessen trust in my DB operations since I upgraded to the latest version.
So I opened #29979 implementing @aqrln's suggestion (serialization in the base `PgQueryable`, with a carve-out so the pool path stays parallel) plus the regression tests requested in review.
If you'd rather carry this PR forward yourself, I'm happy to close mine or contribute the tests here instead.

One data point supporting the approach: the same fix, per-client serialization with the pool unaffected, already shipped in the next-gen driver on main in #29839

PR #29471, opened 2026-04-15
fix(adapters): serialize JSON args with Decimal-aware replacer to restore numeric type

matingathani:
## Problem

Fixes #29387 — a regression introduced in Prisma 7.4.0 where `Decimal` values nested inside a JSON/JSONB field are persisted as a JSON *string* instead of a JSON *number*.

### Root cause

`Decimal.prototype.toJSON` is aliased to `Decimal.prototype.toString`, so when a driver library (e.g. `node-postgres`) calls `JSON.stringify` on the argument object, Decimal instances are serialized as quoted strings (`"-11000"`) rather than numeric literals (`-11000`).

**Example:**

```ts
// before 7.4.0 → stored as {"someDecimal": -11000}  ✅
// after  7.5.0 → stored as {"someDecimal": "-11000"} ❌
await prisma.modelWithJson.create({
  data: {
    id: '1',
    properties: { someDecimal: new Decimal('-11000') },
  },
})
```

## Fix

Each adapter's `mapArg` now pre-serialises JSON arguments using a custom `JSON.stringify` replacer that:

- Converts **Decimal-like instances** (detected via `constructor.name === 'Decimal'` + presence of `toNumber()`) to their numeric value, restoring the JSON number representation.
- Keeps the existing handling for `bigint` (→ string) and `ArrayBuffer`/`ArrayBufferView` (→ base64).

All four adapters (`adapter-pg`, `adapter-mariadb`, `adapter-mssql`, `adapter-libsql`) receive the same fix.

**Bonus:** also adds the missing `'JSON'` decltype mapping in `adapter-libsql` (`mapDeclType` only handled `'JSONB'`, not `'JSON'`).

## Test plan

- [ ] Create a model with a `Json` field in PostgreSQL/MariaDB/SQL Server/SQLite (libSQL)
- [ ] Write a record containing `new Decimal("-11000")` inside the JSON object
- [ ] Read it back and verify the stored value is the JSON number `-11000`, not the string `"-11000"`
- [ ] Verify existing JSON-field tests pass

## Checklist

- [x] `adapter-pg/src/conversion.ts` — `serializeJsonArg` + `mapArg` guard
- [x] `adapter-mariadb/src/conversion.ts` — same
- [x] `adapter-mssql/src/conversion.ts` — same
- [x] `adapter-libsql/src/conversion.ts` — same + `'JSON'` decltype case

### DIFF
--- a/packages/adapter-libsql/src/conversion.ts
+++ b/packages/adapter-libsql/src/conversion.ts
@@ -51,6 +51,7 @@ function mapDeclType(declType: string): ColumnType | null {
       return ColumnTypeEnum.Bytes
     case 'BOOLEAN':
       return ColumnTypeEnum.Boolean
+    case 'JSON':
     case 'JSONB':
       return ColumnTypeEnum.Json
     default:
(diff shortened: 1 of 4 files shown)


matingathani:
Thanks for the review! Fixed in the latest commit (3336607). The issue was exactly as described: `JSON.stringify` calls `toJSON()` *before* invoking the replacer, so `Decimal` instances were already converted to strings by the time our constructor check ran.

Fixed by switching from an arrow function to a regular function as the replacer, so `this` is bound to the parent object. We now inspect `this[key]` (the original pre-toJSON value) for nested properties, and close over the outer `value` argument for the root call (`key === ''`).

matingathani:
Fixed in the latest commit (84e1284):

1. **Double-cast removed** — dropped `as unknown as A` in `adapter-pg/mapArg`'s JSON branch. `serializeJsonArg` returns `string`, which is already in the declared return type union.

2. **NaN/Infinity guard added** — all four adapters (adapter-pg, adapter-mssql, adapter-mariadb, adapter-libsql) now guard `Decimal.toNumber()` with `Number.isFinite()` and fall back to `String(original)` for non-finite values, preventing silent `null` corruption in JSON payloads.

tensordreams:
Hi @matingathani — thanks for this, and apologies for the slow reply.

We are closing this one: there is no repro or linked issue and no tests, and the same `serializeJsonArg` is copied into four adapters relying on `constructor?.name === 'Decimal'` duck-typing, which does not hold across realms or after minification.

PR #29473, opened 2026-04-15
fix(cli): route bootstrap telemetry to Prisma Web Properties project

kristof-siket:
## Summary

- Routes bootstrap activation telemetry (`activation:*` events) to the correct PostHog project for console-related tracking.
- Makes `PosthogEventCapture` accept an optional API key parameter, defaulting to the existing key for backward compatibility.
- NPS feedback events remain unaffected.

## Test plan

- [ ] Verify bootstrap activation events appear in the target PostHog project after running `prisma bootstrap`
- [ ] Verify NPS feedback events are unaffected

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -1,0 +1,3 @@
+# False positive: publishable PostHog client keys (PR #29473).
+1b2a3e437c30e29989fdeac51c15906b37f91b8d:packages/cli/src/utils/nps/capture.ts:generic-api-key:14
+1b2a3e437c30e29989fdeac51c15906b37f91b8d:packages/cli/src/utils/nps/capture.ts:generic-api-key:17
--- a/packages/cli/src/bootstrap/telemetry.ts
+++ b/packages/cli/src/bootstrap/telemetry.ts
@@ -1,12 +1,12 @@
 import type { LinkResult } from '../postgres/link/Link'
-import { PosthogEventCapture } from '../utils/nps/capture'
+import { PosthogEventCapture, PUBLIC_POSTHOG_BOOTSTRAP_ACTIVATION_PROJECT_KEY } from '../utils/nps/capture'
 import type { ProjectState } from './project-state'
 
 function isTelemetryDisabled(): boolean {
   return Boolean(process.env.CHECKPOINT_DISABLE)
 }
 
-const eventCapture = new PosthogEventCapture()
+const eventCapture = new PosthogEventCapture(PUBLIC_POSTHOG_BOOTSTRAP_ACTIVATION_PROJECT_KEY)
 
 interface TelemetryContext {
   distinctId: string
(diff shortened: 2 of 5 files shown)


packages/cli/src/utils/nps/capture.ts
wmadden: Should this be deleted since you use `#apiKey` now?

packages/cli/src/bootstrap/telemetry.ts
```diff
@@ -6,7 +6,7 @@ function isTelemetryDisabled(): boolean {
   return Boolean(process.env.CHECKPOINT_DISABLE)
 }
 
-const eventCapture = new PosthogEventCapture()
+const eventCapture = new PosthogEventCapture('phc_cmc85avbWyuJ2JyKdGPdv7dxXli8xLdWDBPbvIXWJfs')
```
wmadden: This looks like it ought to be an env var, shouldn't it?

packages/cli/src/utils/nps/capture.ts
kristof-siket: This is a default value. History:
 - The posthog integration was originally only used for NPS surveys - that is a separate project from the one we use now
 - I introduced the new events for the new experiment. I left the default as-is but made it configurable on-demand

Having said that, I tend to agree - it could be a mandatory argument since defaulting to a project can have harmful consequences. I'm going to fix it 👍

packages/cli/src/bootstrap/telemetry.ts
kristof-siket: The old API key was also stored in the code like this one (`posthogKey` variable). Regarding env variable, I don't think that would work since the capture runs client-side (this "key" is meant for client-side use), and we want to control from the code what Posthog project to use, not let the user set it with an env var.

.gitleaksignore
```diff
@@ -0,0 +1,3 @@
+# Publishable PostHog client project keys (false positive generic-api-key). See PR #29473.
+af01985181411e09f8ed6280de6660a70adbe0a5:packages/cli/src/utils/nps/capture.ts:generic-api-key:14
+af01985181411e09f8ed6280de6660a70adbe0a5:packages/cli/src/utils/nps/capture.ts:generic-api-key:17
```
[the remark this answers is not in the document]
kristof-siket: GitLeaks comment says:

> 🛑 Gitleaks has detected a secret with rule-id generic-api-key in commit https://github.com/prisma/prisma/commit/af01985181411e09f8ed6280de6660a70adbe0a5.
If this secret is a true positive, please rotate the secret ASAP. If this secret is a false positive, you can add the fingerprint below to your .gitleaksignore file and commit the change to this branch.

What exactly do you mean when you say "Gitleaks is detecting these findings under rule posthog-project-api-key, but the .gitleaksignore entries use generic-api-key."?

wmadden (approved):
Good work @kristof-siket

PR #29474, opened 2026-04-15
feat(generator): add enabled field to toggle generators via env vars

quinnjr:
At Lexmata we maintain a single shared Prisma schema as the source of truth across all our repos and microservices, both TypeScript and Go, and one thing that's been bugging us for a while is that we can't selectively turn generators on or off without maintaining separate schema files or doing hacky workarounds. We have generators for things like Ent modules that we need in our Go projects but absolutely do not want dragging into our TypeScript services and adding unnecessary dependency bloat, and right now there's no clean way to handle that from a single schema.

This PR adds an `enabled` field to generator blocks that works with `env()` so you can do something like `enabled = env("GENERATE_ENT")` and just control what runs at generate time through environment variables. Falsy values (false, 0, no, off, empty string) disable the generator, anything else or unset keeps it enabled so existing schemas aren't affected at all.

### What changed

- Added `enabled?: EnvValue` to `GeneratorConfig` in `@prisma/generator`
- `resolveEnabledField` in `getConfig.ts` resolves the env var value during config loading, same pattern as `resolveBinaryTargets`
- `isGeneratorEnabled` utility checks the resolved value and `filterGenerators` skips disabled ones with an info log
- Unit tests for the utility and integration tests for the full `getGenerators` flow

Happy to adjust anything, this has been working well for us internally and figured it might be useful upstream too.


tensordreams:
Hi @quinnjr — thanks for this, and apologies that it sat unanswered for so long. We've decided not to take it, but I'd rather give you the real reasoning than close it silently.

The mechanism you've built works. `enabled = env("X")` parses fine today: PSL keeps unrecognised generator properties in a catch-all config map and preserves `env()` references unresolved, so the client side can read them. Nothing here is broken.

The trouble is that the catch-all is untyped, so nothing validates the *value*. `enabled = 5` is accepted by schema validation — PSL coerces numerics to strings — and `isGeneratorEnabled` treats anything outside its five-item deny-list (`false`, `0`, `no`, `off`, empty) as enabled. So `enabled = 5` runs the generator, and so does `enabled = "disabled"`. For a property whose entire job is to switch a generator off, defaulting to *on* for every unrecognised value is the wrong failure mode. Making it safe means promoting it to a first-class generator property with real validation in `prisma-engines`, which is a separate repository and release cycle.

For the underlying need, `prisma generate --generator <name>` already covers a good deal of it. It's repeatable, filters to the generators you name, and errors if you name one that doesn't exist — so it validates in exactly the place the `enabled` property could not. Driving that from a variable in your build script gets you the same toggle with a safer failure mode.

Thanks again for the effort you put in here.

PR #29475, opened 2026-04-15
fix: normalize line endings in prisma format output

yosofbadr:
## Summary

- Normalizes CRLF (`
`) line endings to LF (`
`) in the output of `formatSchema`, which powers the `prisma format` command.
- On Windows, the Wasm formatter can produce a trailing CRLF at the end of the file even when all other lines use LF. This causes spurious diffs when the same file is formatted on Linux CI.
- Adds a test to verify that CRLF input is normalized to LF in the formatted output.

Fixes #8548

## Test plan

- [x] Added unit test: `should normalize CRLF to LF in formatted output` in `packages/internals/src/__tests__/engine-commands/formatSchema.test.ts`
- [x] All existing `formatSchema` tests pass (12/12)
- [x] Verify on Windows that `prisma format` no longer produces CRLF at end of file


SevInf:

PR #29476, opened 2026-04-15
feat(cli): warn when local Prisma installation exists

### LINKED ISSUE #2738: Global Prisma could check for local Prisma and warn or require explicit confirmation

yosofbadr:
## Summary

- When the global `prisma` CLI is invoked in a project that also has a local installation in `node_modules`, a warning is now printed to stderr advising the user to use `npx prisma` instead.
- Leverages the existing `isCurrentBinInstalledGlobally()` utility from `@prisma/internals` to detect global execution, and checks for `node_modules/.bin/prisma` or `node_modules/prisma` to detect a local installation.
- Includes vitest tests covering all branches: no warning when local, no warning when global without local, warning when global with local.

Closes #2738

## Test plan

- [x] `findLocalPrismaInstallation` returns `undefined` when no local installation exists
- [x] `findLocalPrismaInstallation` detects `node_modules/.bin/prisma`
- [x] `findLocalPrismaInstallation` detects `node_modules/prisma`
- [x] `warnIfLocalInstallationShadowed` does not warn when CLI is not globally installed
- [x] `warnIfLocalInstallationShadowed` does not warn when globally installed but no local installation
- [x] `warnIfLocalInstallationShadowed` warns with correct message when globally installed and local installation exists

### DIFF
--- a/packages/cli/src/bin.ts
+++ b/packages/cli/src/bin.ts
@@ -55,6 +55,7 @@ import { redactCommandArray } from './utils/checkpoint'
 import { loadOrInitializeCommandState } from './utils/commandState'
 import { UserFacingError } from './utils/errors'
 import { loadConfig } from './utils/loadConfig'
+import { warnIfLocalInstallationShadowed } from './utils/warn-local-installation'
 import { Validate } from './Validate'
 import { Version } from './Version'
 
@@ -93,6 +94,9 @@ const args = arg(
  * Main function
  */
 async function main(): Promise<number> {
+  // https://github.com/prisma/prisma/issues/2738
+  warnIfLocalInstallationShadowed()
+
   // create a new CLI with our subcommands
 
   const cli = CLI.new(
--- a/packages/cli/src/utils/__tests__/warn-local-installation.vitest.ts
+++ b/packages/cli/src/utils/__tests__/warn-local-installation.vitest.ts
@@ -1,0 +1,117 @@
+import fs from 'node:fs'
+import os from 'node:os'
+import path from 'node:path'
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('@prisma/internals', () => ({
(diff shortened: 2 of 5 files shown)


yosofbadr:
Addressed the two CodeRabbit findings in 18f26e4:

1. **Windows shim** — `findLocalPrismaInstallation` now also probes `node_modules/.bin/prisma.cmd` in addition to the POSIX `prisma` shim, so Windows users with a local install are detected correctly. Added a test case.
2. **yarn / pnpm global installs** — extended `isCurrentBinInstalledGlobally()` to return `'npm' | 'yarn' | 'pnpm' | false` instead of `'npm' | false`. Yarn is detected via the existing `global-directory` package; pnpm is detected via `$PNPM_HOME` and the default global locations pnpm uses on macOS / Linux / Windows. `printUpdateMessage` now emits the correct `yarn global add` / `pnpm add -g` command when the CLI was installed through those managers.

`warn-local-installation.vitest.ts` (7 tests) and `printUpdateMessage.test.ts` (2 snapshot tests) both pass locally.

tensordreams:
Thanks — the extension of `isCurrentBinInstalledGlobally()` to yarn/pnpm globals is genuinely careful work (the `global-directory` fallbacks are correct).

Blocking: there's no version comparison anywhere. The warning fires on the mere presence of a local install — including when global and local versions match exactly, which is the false-positive the issues warn about. Please read the installed version from `node_modules/prisma/package.json`, compare against the running CLI's version, and warn only on a real mismatch. Also consider an opt-out and quieter behavior in non-interactive runs — as written this prints on every command in every project with a local install, including CI.


---

PR #29481, opened 2026-04-17
fix(adapter-pg): correctly handle TIMESTAMPTZ reads and writes on non-UTC servers

### LINKED ISSUE #26786: `@db.Timestamptz()` not handled correctly with `driverAdapters` when PostgreSQL timezone is not UTC
### Bug description

I followed the official Prisma documentation below to enable driverAdapters and use node-postgres:

- https://www.prisma.io/docs/orm/overview/databases/postgresql#using-the-node-postgres-driver

When using a Date field with `@db.Timestamptz()` in the Prisma schema, we noticed that values are not retrieved correctly if the PostgreSQL database is configured with a timezone other than `UTC` (e.g., `Asia/Tokyo`).

- This issue does not occur when not using driverAdapters.
- This issue also does not occur when querying directly with node-postgres without PrismaClient.

### Severity

### Reproduction

Here is a reproduction project:

- https://github.com/onozaty/problem-prisma-driver-adapter-date

You can reproduce the issue by following these steps:

- Open the project in a DevContainer or GitHub Codespaces.
    - The timezone for both PostgreSQL and the execution environment is `Asia/Tokyo`.
- Run `pnpm test`.
    - Only the case using driverAdapters results in incorrect datetime values.

The correct datetime, adjusted for the configured timezone, should be returned.

### Frequency

Consistently reproducible

### Is this a regression?

I do not know if it is a regression.

### Workaround

No workaround found.

### Prisma Schema & Queries

```prisma
generator client {
}

datasource db {
  provider = "postgresql"
}

model User {
  email     String
  createdAt DateTime @db.Timestamptz()
}
```
```ts
import "dotenv/config";
import { Pool } from "pg";

const connectionString = `${process.env.DATABASE_URL}`;

const pool = new Pool({ connectionString });
const adapter = new PrismaPg(pool);
export const prismaAdapter = new PrismaClient({ adapter });

await prismaAdapter.$connect();
```

```ts
  const adapterResult = await prismaAdapter.user.findUnique({
    where: {
      id: created.id,
    },
  });
```

### Logs & Debug Info

```
node@561ea9e59a68:/workspaces/problem-prisma-driver-adapter-date$ pnpm test

> problem-prisma-driver-adapter-date@1.0.0 test /workspaces/problem-prisma-driver-adapter-date
> vitest run


 RUN  v3.0.9 /workspaces/problem-prisma-driver-adapter-date

stdout | prisma.test.ts > timestamp with timezone
Prisma(default): {
  id: 4,
  email: 'user@example.com',
  createdAt: 2025-01-01T00:00:00.000Z
}

stdout | prisma.test.ts > timestamp with timezone
Prisma(adapter): {
  id: 4,
  email: 'user@example.com',
  createdAt: 2025-01-01T09:00:00.000Z
}

stdout | prisma.test.ts > timestamp with timezone
pg: {
  id: 4,
  email: 'user@example.com',
  createdAt: 2025-01-01T00:00:00.000Z
}

 ❯ prisma.test.ts (1 test | 1 failed) 77ms
   × timestamp with timezone 76ms
     → expected 2025-01-01T09:00:00.000Z to deeply equal 2025-01-01T00:00:00.000Z

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  prisma.test.ts > timestamp with timezone
AssertionError: expected 2025-01-01T09:00:00.000Z to deeply equal 2025-01-01T00:00:00.000Z

- Expected
+ Received

- 2025-01-01T00:00:00.000Z
+ 2025-01-01T09:00:00.000Z

 ❯ prisma.test.ts:29:41
     27|   });
     28|   console.log("Prisma(adapter):", adapterResult);
     29|   expect.soft(adapterResult?.createdAt).toEqual(createdAt);
       |                                         ^
     30| 
     31|   const pgResult = await pgPool.query('SELECT * FROM "User" WHERE id = $1', [

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


 Test Files  1 failed (1)
      Tests  1 failed (1)
   Start at  04:47:18
   Duration  706ms (transform 63ms, setup 0ms, collect 323ms, tests 77ms, environment 0ms, prepare 95ms)

 ELIFECYCLE  Test failed. See above for more details.
```

### Environment & Setup

- **OS:** Debian GNU/Linux 12 (bookworm)
- **Database:**  PostgreSQL 16.3
- **Node.js version:** v20.15.1

### Prisma Version

```
prisma                  : 6.5.0
@prisma/client          : 6.5.0
Computed binaryTarget   : debian-openssl-3.0.x
Operating System        : linux
Architecture            : x64
Node.js                 : v20.15.1
TypeScript              : 5.8.2
Query Engine (Node-API) : libquery-engine 173f8d54f8d52e692c7e27e72a88314ec7aeff60 (at node_modules/.pnpm/@prisma+engines@6.5.0/node_modules/@prisma/engines/libquery_engine-debian-openssl-3.0.x.so.node)
Schema Engine           : schema-engine-cli 173f8d54f8d52e692c7e27e72a88314ec7aeff60 (at node_modules/.pnpm/@prisma+engines@6.5.0/node_modules/@prisma/engines/schema-engine-debian-openssl-3.0.x)
Schema Wasm             : @prisma/prisma-schema-wasm 6.5.0-73.173f8d54f8d52e692c7e27e72a88314ec7aeff60
Default Engines Hash    : 173f8d54f8d52e692c7e27e72a88314ec7aeff60
Studio                  : 0.510.0
```

### LINKED ISSUE #28629: adapter-pg: Date objects incorrectly stored in Timestamptz fields
### Bug description

Follow up to #26786 with some precisions,
it's not a retrieval issue, dates are not stored correctly in db if db timezone isn't set to UTC.

Kinda severe for a pushed up feature which is now the default with v7, and a clear regression.
That's why I create a new issue.
It's mitigated though as DBs are commonly set to UTC.

**Issue:**
Date object intended to be stored in a Timestamptz field is stored as UTC with db timezone.

### Severity

### Reproduction

With a database set to a timezone other than UTC, here America/New_York (-5), see docker-compose.

schema.prisma
```
generator client {
  provider = "prisma-client"
  output   = "generated"
}

datasource db {
  provider = "postgresql"
}

model Test {
  createdAt DateTime @map("created_at") @db.Timestamptz(6)
}
```

test.js
```js
import 'dotenv/config';
import { PrismaClient } from './prisma/generated/client.js';

const adapter = new PrismaPg({ connectionString: process.env.DATABASE_URL })

await prisma.test.create({
  data: {
    createdAt: new Date(),
  },
});
```

Prisma logs an UTC date "2025-11-21T10:23:41.021Z"
But db receives and stores that time as if was in its own timezone
```
SELECT * FROM "Test";
 id |        created_at
----+----------------------------
  1 | 2025-11-21 10:23:41.021-05
(1 row)
```

Without PrismaPg, old v6 way using 'prisma-client-js' generator,
it's correctly stored as UTC
```
SELECT * FROM "Test";
 id |         created_at
----+----------------------------
  1 | 2025-11-21 05:23:41.021-05
(1 row)
```

### Frequency

Consistently reproducible

### Is this a regression?

Yes, works correctly the old way, without adapters.

### Environment & Setup

docker-compose.yml
```
---
services:
  db:
    image: postgres:15-alpine
    ports:
      - "5432:5432"
    environment:
      POSTGRES_DB: db
      POSTGRES_USER: db
      POSTGRES_PASSWORD: db
      TZ: America/New_York
    command:
      - "postgres"
      - "-c"
      - "log_statement=all"
    volumes:
      - db-data:/var/lib/postgresql/data
volumes:
  db-data:
```

### Prisma Version

7.0.0, but every version using Query Compiler is concerned.

matingathani:
## Summary

Fixes two bugs in `packages/adapter-pg/src/conversion.ts` that cause `@db.Timestamptz` fields to return or store wrong values when the PostgreSQL session timezone is not UTC.

### Read bug — closes #26786

`normalize_timestamptz` swapped the offset label to `+00:00` without adjusting the time:

```ts
// Before — WRONG: '09:26:34.887-06' → '09:26:34.887+00:00' (6 h off)
return time.replace(' ', 'T').replace(/[+-]\d{2}(:\d{2})?$/, '+00:00')

// After — correct UTC conversion
const withFullOffset = time.replace(' ', 'T').replace(/([+-]\d{2})$/, '$1:00')
return new Date(withFullOffset).toISOString().replace(/Z$/, '+00:00')
```

### Write bug — closes #28629

`formatDateTime` emitted no timezone suffix, so PostgreSQL treated the value as being in the server's local timezone:

```ts
// Before — WRONG: '1999-12-31 23:59:59.999' (no timezone — DB-local)
// After  — correct: '1999-12-31 23:59:59.999+00:00' (explicitly UTC)
```

For `TIMESTAMP WITHOUT TIME ZONE` columns PostgreSQL silently ignores timezone suffixes, so the `+00:00` addition is a safe no-op for those columns.

## Test plan

- [ ] Unit tests in `packages/adapter-pg/src/__tests__/conversion.test.ts` updated and extended — cover UTC-6, UTC+5:30, UTC+0, and cross-midnight precision cases for `normalize_timestamptz`
- [ ] Functional regression test `packages/client/tests/functional/issues/26786-pg-timestamptz-non-utc/` — sets the PostgreSQL session to `America/New_York` and verifies the full write → read round-trip produces the original UTC instant
- [ ] CI `driver-adapter-unit-tests` / `client-functional-tests` jobs will validate against a live database


UgaTheDev (commented):
Independently reproduced both halves of this on the current `v7` and `7.9.x` tips, and I think the diagnosis and the branch targeting are right. One finding below that I believe blocks merge — it's the same line CodeRabbit flagged, but I'd resolve it the other way, and I have the numbers for why.

**Confirming the mechanism.** The two bugs are symmetric, which is why they hid for so long:

- `formatDateTime` emits a naive string with no offset, so PostgreSQL resolves it in the **server session timezone** — the write-side shift in #28629.
- `normalize_timestamptz` relabels the offset rather than converting it, so `2025-11-24 09:26:34.887-06` becomes `...T09:26:34.887+00:00` — a different instant. That's #26786.

Both err in the same direction, so a round-trip against a single non-UTC server looks correct. The corruption only surfaces when a value crosses a boundary — a relation query taking a different code path, a second client, or a UTC-configured staging environment. Worth calling out because it explains the "works locally, wrong in prod" reports in the issue thread.

Both are still present verbatim on `v7` and `7.9.x` today: `normalize_timestamptz` is still the one-line `.replace(/[+-]\d{2}(:\d{2})?$/, '+00:00')`, and `formatDateTime` still ends at the milliseconds with no offset suffix.

**The regex offset handling is genuinely necessary here**, which I want to state explicitly since it looks like over-engineering on first read. Two cases justify it: 6-digit microseconds (PostgreSQL's full precision, which JS `Date` truncates to milliseconds) and sub-minute LMT-era offsets like `+05:53:20`. Neither survives a naive `new Date(...)` round-trip, so on a `Date`-based adapter the manual handling has to exist.

---

### Blocking: the LMT test fails, and I think the code is what's wrong

I ran `pnpm run test` in `packages/adapter-pg` against this branch. 12 of 13 pass; the LMT case fails:

```
FAIL  src/__tests__/conversion.test.ts > normalize_timestamptz
      > normalises LMT-era offsets with seconds component (e.g. +05:53:20)

AssertionError: expected '1900-01-01T04:07:00.000+00:00'
                to be     '1900-01-01T04:06:40.000+00:00'
```

Same failure under `TZ=UTC`, `TZ=Asia/Kolkata` and `TZ=America/Chicago`, so it isn't a local-timezone artifact. It's also already red in this PR's CI — `DA Unit Tests [v20.19.0]` and `[v24]` both fail on exactly this assertion, so this isn't a suite that's being skipped.

The cause is this line in `conversion.ts`:

```js
.replace(/([+-]\d{2}):(\d{2}):\d{2}$/, '$1:$2')
```

`+05:53:20` is applied as `+05:53`, so the result lands 20 seconds late — precisely the seconds component being discarded.

**On the fix:** I'd push back on relaxing the assertion to `04:07:00`. `Temporal` — which is what `main` uses for this now — agrees with the test, not the code. Running the PR's own wire formats through the pinned `temporal-polyfill@1.0.4`:

| input | `Temporal.Instant.from` | this PR |
|---|---|---|
| `2025-11-24 09:26:34.887-06` | `2025-11-24T15:26:34.887Z` | correct |
| `2024-01-01 20:30:00+05:30` | `2024-01-01T15:00:00Z` | correct |
| `2025-11-24 09:26:34.887654-06` | `2025-11-24T15:26:34.887654Z` | correct |
| `1900-01-01 10:00:00+05:53:20` | `1900-01-01T04:06:40Z` | **off by 20s** |
| `1900-01-01 10:00:00-05:53:20` | `1900-01-01T15:53:20Z` | **off by 20s** |

The error is symmetric across the sign, and it survives the microsecond path too (`...10:00:00.123456+05:53:20` comes back as `04:07:00.123456`). So relaxing the assertion would codify a silent whole-second error in the one code path whose entire purpose is returning a correct instant — and the comment justifying it ("the fractional-second error is negligible") describes a smaller error than the one actually occurring.

The seconds are already captured by the regex, so folding them back in is a few lines. This passes all 13 tests plus the two extra cases above, under all three timezones:

```js
const offsetSecondsMatch = withSeparator.match(/([+-])\d{2}:\d{2}:(\d{2})$/)
const offsetSecondsMs = offsetSecondsMatch
  ? (offsetSecondsMatch[1] === '-' ? 1 : -1) * Number(offsetSecondsMatch[2]) * 1000
  : 0
```

then apply it at both construction sites:

```js
const d = new Date(new Date(trimmed).getTime() + offsetSecondsMs)
// ...and...
const d = new Date(new Date(withFullOffset).getTime() + offsetSecondsMs)
```

If that's more than this PR wants to take on, the honest alternative is to drop the LMT test and the `:\d{2}` branch entirely and document the limitation — but silently rounding is the one option I'd avoid.

---

### Non-blocking

- Appending `+00:00` unconditionally in `formatDateTime` is correct for `TIMESTAMP WITHOUT TIME ZONE` (PostgreSQL ignores the offset there), and the inline comment saying so is the right call — that's the kind of thing that gets "cleaned up" by someone later.
- The `skipDriverAdapter: { from: ['js_neon'] }` with a written reason and a removal condition is the right shape for a temporary skip. Nice to see the follow-up named rather than left implicit.
- Scope note, not a request: `@db.Timetz` has the same shape of bug and this PR doesn't touch it. `mapArg` routes `TIMETZ` through `formatTime`, which emits no offset, and `normalize_timez` strips the offset and assumes UTC. That's a deliberate pre-existing convention (the comment cites quaint), but after this PR the two paths disagree — `timestamptz` converts, `timetz` still relabels. Might be worth a linked issue so it doesn't get rediscovered as a regression.

---

### Context: this is a v7-lifetime fix

Worth noting for scoping. On `main` `packages/adapter-pg/src/conversion.ts` no longer exists at all — the repo has been restructured and `timestamptz` now runs through the `pg/timestamptz-temporal@1` codec, which decodes with `Temporal.Instant.from()` in `packages/3-targets/3-targets/postgres/src/core/temporal-codec-helpers.ts`. That's offset-correct by construction rather than by regex, which is why the table above uses it as the reference.

So this fix is needed for the v7 line but has a natural sunset at v8 — which seems like the right reason to keep the diff conservative rather than reaching for a larger refactor. Given both bugs are still live on `v7` and `7.9.x` and the issue thread has people blocked on 7.x, this seems worth landing once the LMT case is settled.

---

## Notes for review (not part of the comment)

**What was actually run**

- Worktree at PR head `acc4657`, `pnpm install --filter=@prisma/adapter-pg... --ignore-scripts`, then `vitest run packages/adapter-pg/src/__tests__/conversion.test.ts`. Workspace packages weren't built in the scratch checkout, so `@prisma/driver-adapter-utils` and `@prisma/debug` were aliased to their sources in a local-only `vitest.config.ts` edit. That affects module resolution only, not the code under test.
- Result: **12 passed, 1 failed** — the LMT test — identically under `TZ=UTC`, `TZ=Asia/Kolkata`, `TZ=America/Chicago`.
- The same failure is present in the PR's CI: run 30017685132, jobs `DA Unit Tests [v20.19.0]` (89963003612) and `[v24]`, byte-identical assertion message. `.github/workflows/test-template.yml:919-921` runs `pnpm run test` in `packages/adapter-pg`, so the suite is wired up and simply red. The earlier speculation that the suite might not be running is **disproved** — do not repeat it.
- The proposed patch was applied locally and re-run: **13/13 pass** under all three timezones, plus two ad-hoc cases (negative LMT offset, LMT offset with microseconds) that also pass with the patch and fail without it.
- Temporal reference values produced with `temporal-polyfill@1.0.4` in a throwaway project, not inferred.

**Corrections to the previous draft**

- The "dead `optOut` block listing sqlite/mysql/mongodb" finding was **stale** and has been removed. It was true of the PR's first commit `d0af4bd71` only; the current head replaced it with a documented `skipDriverAdapter` for `js_neon`. Do not post that bullet.
- "This test should be failing CI; worth checking whether the timestamptz suite is actually running" replaced with the confirmed CI evidence.

**Still true**
- The rest of the diff checked out. No other defects found.
- Not posted. No PR opened, no branch, no commits, nothing pushed.

PR #29484, opened 2026-04-17
fix(adapters): use the reported column for NullConstraintViolation meta (P2011)

matingathani:
## Problem

When a PostgreSQL NOT NULL constraint is violated (error code `23502`), the Prisma client returns `P2011` with an empty `meta.target`. Users see:

```
PrismaClientKnownRequestError: Null constraint violation on the (not available)
  code: 'P2011',
  meta: { target: [] }
```

instead of the expected:

```
meta: { target: ['email'] }
```

## Root Cause

`packages/adapter-pg/src/errors.ts` was applying a regex designed for **unique-violation** (`23505`) errors to **NOT NULL** (`23502`) errors:

```typescript
// WRONG: this is the 23505 unique-violation detail format
const fields = error.detail?.match(/Key \(([^)]+)\)/)?.at(1)?.split(', ')
```

PostgreSQL's `23502` error detail looks like `"Failing row contains (null, null, null)"` — which never matches `/Key \(([^)]+)\)/`. PostgreSQL provides the violating column name via `error.column` instead, which was being ignored.

## Fix

Read `error.column` directly for `23502` errors:

```typescript
case '23502': {
  return {
    kind: 'NullConstraintViolation',
    constraint: error.column !== undefined ? { fields: [error.column] } : undefined,
  }
}
```

## Test Plan

- [x] Updated the existing `23502` test to use real PostgreSQL error shape (`detail: "Failing row contains ..."`, `column: "foo"`) — it previously passed with synthetic data that accidentally matched the broken regex
- [x] Added a test for the `error.column`-absent fallback
- [x] All 34 tests in `packages/adapter-pg/src/__tests__/errors.test.ts` pass


jacek-prisma:
I think this should be tested by triggering a real database error in one of the functional tests to ensure we got the error format right

matingathani:
Agreed, a functional test that triggers a real NOT NULL violation would be more robust. I'll add one.

matingathani:
Done — added a functional test at `packages/client/tests/functional/driver-adapters/null-constraint-target/`. It forces a NOT NULL violation by omitting the required `title` field and asserts that `meta.target` contains `'title'` (the column name, sourced from `error.column`). This directly exercises the `errors.ts` fix end-to-end against a real PostgreSQL DB via the js_pg adapter.

tensordreams:
Rebased onto `main` (resolved the `errors.test.ts` conflict against the newly landed 23505/23001 tests) and refined the tests.

The adapter fix itself stands: for `23502` PostgreSQL populates `error.column` (its `error.detail` is `Failing row contains (...)`, not the `Key (...)` format of unique violations), so reading `error.column` is correct. Unit tests now cover column-present (with and without `detail`) and column-absent.

One correction to the functional test: it asserted `meta.target` for P2011, but that is not right on two counts. prisma-engines exposes the P2011 constraint under `meta.constraint`, not `meta.target` (`NullConstraintViolation { constraint: DatabaseConstraint }`), and the client-engine-runtime does not populate that meta field on `main` yet — that plumbing is #29349, still open. So the assertion could not pass here. The test now asserts the rendered P2011 message names the violating column ("Null constraint violation on the fields: (\`title\`)"), which is exactly the effect of this change (without it the message reads "on the (not available)") and is self-contained. This adapter fix and #29349 are complementary: once #29349 lands, reading `error.column` here is what makes `meta.constraint` non-null for Postgres NOT NULL violations.

`pnpm --filter @prisma/adapter-pg test` → 54 passed; prettier and eslint clean on the changed files. The DB-backed functional test was not run locally.

tensordreams:
Pushed a fix for the CI failures (14452e8fc).

All 12 non-lint failures were the same test, `driver-adapters/null-constraint-target`, failing with `PrismaClientValidationError` instead of `PrismaClientKnownRequestError`. With `title String` required in the schema, `create({ data: {} })` was rejected by client-side validation before any SQL reached the database, so the 23502 path was never exercised. The test now declares `title String?` and adds the NOT NULL constraint via `ALTER TABLE` in `beforeAll` (same approach as `issues/TML-1664-invalid-enum-value-error`). It also asserts the rendered `Null constraint violation on the fields: (`title`)` rather than a bare `title` substring, which the callsite argument preview would have satisfied on its own.

The `error.column` change itself was correct. Verified against a live PostgreSQL: a 23502 reports `column: "title"` while `detail` is `Failing row contains (...)`, so the `Key (...)` regex never matched and P2011 always rendered `(not available)`.

`adapter-neon` and `adapter-ppg` carried the identical 23502 bug (both copied the 23505 `detail` parsing), which is why the js_neon jobs failed too. Both now read the reported column, with unit tests added. Adapter suites are green locally: pg 54, neon 10, ppg 39.

PR #29485, opened 2026-04-17
fix(adapter-neon): use error.column for NullConstraintViolation and fix ColumnNotFound for unquoted names

matingathani:
### P2011 always shows empty `meta.target`

When a PostgreSQL NOT NULL constraint is violated (error code `23502`), Prisma returns `P2011` with an empty `meta.target`:

```
PrismaClientKnownRequestError: Null constraint violation on the (not available)
  code: 'P2011',
  meta: { target: [] }
```

The root cause: `adapter-neon` applied the unique-constraint regex `/Key \(([^)]+)\)/` to `error.detail` for NOT NULL errors. PostgreSQL's `23502` errors always have `error.detail = "Failing row contains (...)"`, which never matches that pattern — so `fields` is always `undefined` and `meta.target` is always empty.

PostgreSQL correctly sets `error.column` to the violating column name for NOT NULL errors. The fix reads that field directly.

### ColumnNotFound misses unquoted column names (42703)

The previous implementation — `error.message.split(' ').at(1)?.split('"').at(1)` — only extracted column names surrounded by double-quotes. For unquoted names (`column foo does not exist`), qualified names (`column users.first_name does not exist`), and escaped-quote names (`column "a""b" does not exist`), it returned `undefined` or the wrong value.

The same fix was applied to `adapter-pg` in #29307. This PR brings parity to `adapter-neon`.

## Fix

**`errors.ts`**

- `23502`: use `error.column` directly instead of the wrong regex.
- `42703`: use the same improved regex (`/^column (.+) does not exist$/`) plus the de-quoting replacer already in `adapter-pg`.

## Tests

Added `src/__tests__/errors.test.ts` with full coverage of all error cases including:
- NullConstraintViolation with and without `error.column`
- ColumnNotFound with unquoted, quoted, qualified, spaced, and escaped-quote column names

All 27 tests pass.

## Related

- Fixes the same bugs as adapter-pg PR #29484 (NullConstraintViolation) and commit #29307 (ColumnNotFound).

### DIFF
--- a/packages/adapter-neon/src/__tests__/errors.test.ts
+++ b/packages/adapter-neon/src/__tests__/errors.test.ts
@@ -1,0 +1,218 @@
+import { describe, expect, it } from 'vitest'
+
+import { convertDriverError } from '../errors'
+
+describe('convertDriverError', () => {
+  it('should handle LengthMismatch (22001)', () => {
+    const error = { code: '22001', column: 'foo', message: 'msg', severity: 'ERROR' }
+    expect(convertDriverError(error)).toEqual({
+      kind: 'LengthMismatch',
+      column: 'foo',
+      originalCode: error.code,
+      originalMessage: error.message,
+    })
+  })
+
+  it('should handle ValueOutOfRange (22003)', () => {
+    const error = { code: '22003', message: 'out of range', severity: 'ERROR' }
+    expect(convertDriverError(error)).toEqual({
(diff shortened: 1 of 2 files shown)


matingathani:
Superseded by #29489 which includes these same 23502/42703 fixes alongside the TLS and socket error handling additions. Closing to avoid conflicting changes to the same file.

PR #29498, opened 2026-04-21
fix(client): align postgres driver adapter error mapping with raw path

### LINKED ISSUE #29496: plpgsql RAISE EXCEPTION (P0001) surfaces as bare DriverAdapterError on ORM calls but as PrismaClientKnownRequestError(P2010) on $queryRaw
### Bug description

When `@prisma/adapter-pg`'s `mapDriverError` falls through to `{ kind: 'postgres', ... }` — which it does for every Postgres SQLSTATE without a specific mapping, notably `P0001` from plpgsql `RAISE EXCEPTION` — the resulting error surfaces to user code with different shapes depending on whether it originated on the ORM path or the raw-query path.

Reference: `@prisma/client@7.7.0` runtime bundle.

```js
// ORM path
function Be(e) {
  if (!isDriverAdapterError(e)) throw e;
  let t = ap(e);   // P-code mapper — returns undefined for kind:'postgres'
  let r = ns(e);
  throw !t || !r ? e : new UserFacingError(r, t, { driverAdapterError: e });
}

// $queryRaw / $executeRaw path
function bn(e) {
  throw isDriverAdapterError(e)
    ? new UserFacingError(
        `Raw query failed. Code: \`${e.cause.originalCode ?? "N/A"}\`. Message: \`${e.cause.originalMessage ?? ns(e)}\``,
        "P2010",
        { driverAdapterError: e },
      )
    : e;
}
```

`ap()` contains `case "postgres": ... return;` (no P-code), so the ORM path rethrows the raw `DriverAdapterError` unchanged. The raw-query path wraps the identical underlying cause as `PrismaClientKnownRequestError` (via the UserFacingError → PrismaClientKnownRequestError conversion at the dispatcher) with `code: "P2010"`.

### Repro

```sql
CREATE OR REPLACE FUNCTION reject_account_id_change() RETURNS TRIGGER AS $$
BEGIN
  IF NEW.account_id IS DISTINCT FROM OLD.account_id THEN
    RAISE EXCEPTION 'workspaces.account_id is immutable';
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER t_workspaces_no_account_change
  BEFORE UPDATE ON workspaces
  FOR EACH ROW EXECUTE FUNCTION reject_account_id_change();
```

- `prisma.workspace.update({ where: {...}, data: { accountId: otherId } })` → bare `DriverAdapterError`, name `"DriverAdapterError"`, no `code` property. **(verified against adapter-pg 7.7.0 + pg 18 + client 7.7.0)**
- `prisma.$executeRaw\`UPDATE workspaces SET account_id = ${otherId} WHERE id = ${id}\`` → based on the `bn()` code above, this should surface as `PrismaClientKnownRequestError` with `code: "P2010"` and the pg message on `meta.driverAdapterError.cause.originalMessage`.

### Why this matters

Error-translation layers end up needing two different code paths for the same underlying pg error. `err instanceof PrismaClientKnownRequestError` is false on the ORM path; checking `err.code` doesn't work either. The only portable signal across both is the nested `err.cause.kind === 'postgres'` / `err.cause.code === 'P0001'` shape — and that structure sits at different depths in the two wrappers (`err.cause` for ORM, `err.meta.driverAdapterError.cause` for raw), because per #28281 `meta.driverAdapterError` isn't public API anyway.

### Expected

Two reasonable options:

1. Wrap `kind: 'postgres'` in `ap()` with a stable P-code — P2010 matches what `bn()` already does on the raw path, and P2010's existing definition ("Raw query failed") is a reasonable fit for "adapter-level pg error that Prisma didn't classify further." This removes the asymmetry.
2. Explicitly document that callers must inspect both wrapped and bare shapes for any non-specifically-mapped SQLSTATE, and expose a stable accessor for `{kind, code, message}` that works in both cases.

Related: #28810 covers deferred-trigger timing under adapter-pg, which is a different symptom.

### Versions

- `@prisma/client` 7.7.0
- `@prisma/adapter-pg` 7.7.0
- Postgres 18

subhaushsingh:
### Problem
When a plpgsql trigger fires `RAISE EXCEPTION` (SQLSTATE P0001), the error
surfaces with different shapes depending on the call path:

- ORM call (`prisma.model.update(...)`) → bare `DriverAdapterError`, no `.code`
- `$queryRaw` / `$executeRaw` → `PrismaClientKnownRequestError` with `code: "P2010"`

Root cause: `ap()` (the ORM path's P-code mapper) has `case "postgres": return`
which returns `undefined`, causing the ORM handler `Be()` to rethrow the raw
DriverAdapterError. The raw-query handler `bn()` unconditionally wraps as P2010.

### Fix
Add `return "P2010"` to the `case "postgres"` branch in `ap()`, mirroring
what `bn()` already does. This makes `instanceof PrismaClientKnownRequestError`
reliable on both paths for any Postgres error that isn't more specifically mapped.

### Testing
- Added unit tests verifying ORM path wraps as `PrismaClientKnownRequestError`
- Added symmetry test verifying `.code` matches between ORM and raw paths

### No breaking changes
Callers already catching `PrismaClientKnownRequestError` P2010 are unaffected.
Callers catching the raw `DriverAdapterError` by name will need to update —
this is the intended behavior change.

### DIFF
--- a/packages/client-engine-runtime/src/user-facing-error.test.ts
+++ b/packages/client-engine-runtime/src/user-facing-error.test.ts
@@ -31,16 +31,18 @@ test('rethrowAsUserFacing wraps postgres errors as P2010 with formatted message'
     originalMessage: 'Trigger violation',
   })
 
+  expect.assertions(3)
+
   try {
     rethrowAsUserFacing(error)
   } catch (e: any) {
-    // Verify the Prisma P-code
+    // 1. Verify the Prisma P-code
     expect(e.code).toBe('P2010')
     
-    // Verify the formatted message matches the $queryRaw path
+    // 2. Verify the formatted message matches the $queryRaw path (via the new helper)
     expect(e.message).toBe('Raw query failed. Code: `P0001`. Message: `Trigger violation`')
     
-    // Verify the original error is attached in meta
+    // 3. Verify the original error is attached in meta
     expect(e.meta).toMatchObject({
       driverAdapterError: error,
     })
--- a/packages/client-engine-runtime/src/user-facing-error.test.ts
+++ b/packages/client-engine-runtime/src/user-facing-error.test.ts
@@ -23,3 +23,26 @@ test('rethrowAsUserFacing re-throws the original error for unknown cause kinds',
   // Crucially the message must NOT degrade to the useless "[object Object]"
   expect(() => rethrowAsUserFacing(error)).not.toThrowError('[object Object]')
 })
+
+test('rethrowAsUserFacing wraps postgres errors as P2010 with formatted message', () => {
+  const error = makeDriverAdapterError({
+    kind: 'postgres',
+    originalCode: 'P0001',
+    originalMessage: 'Trigger violation',
+  })
+
+  try {
+    rethrowAsUserFacing(error)
+  } catch (e: any) {
+    // Verify the Prisma P-code
+    expect(e.code).toBe('P2010')
+    
+    // Verify the formatted message matches the $queryRaw path
+    expect(e.message).toBe('Raw query failed. Code: `P0001`. Message: `Trigger violation`')
+    
+    // Verify the original error is attached in meta
+    expect(e.meta).toMatchObject({
+      driverAdapterError: error,
+    })
+  }
+})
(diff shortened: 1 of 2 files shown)


tensordreams:
Hi @subhaushsingh — thanks for this, and apologies for the slow reply.

We are closing this one: this would revert a deliberate decision: unmapped database errors on the non-raw path map to P2039 precisely so they do not claim "Raw query failed" for an operation that was not a raw query. Routing postgres back to P2010 reintroduces that.

PR #29535, opened 2026-05-05
fix(cli): force LF line endings on Windows

sunhimuka:
This PR ensures that `prisma format` uses LF line endings on Windows, preventing CRLF inconsistencies.


SevInf:
Hi @sunhimuka 
Our intention is for prisma-format to output CRLF endings on Windows.
This pull request will be a big breaking change for every windows user we have right now.

What we might accept instead is an off-by-default option to switch force LF endings on Windows. This repo would be a wrong place for such feature as well — it should be done in https://github.com/prisma/prisma-engines as a part of `prisma-fmt`

PR #29538, opened 2026-05-06
chore(deps): bump postgres-bytea to 1.0.1 to drop Buffer() deprecation

kolia-zamnius:
## Problem

Reading a `Bytes` column through `@prisma/adapter-neon`, `@prisma/adapter-pg`, or `@prisma/adapter-ppg` emits Node.js `DEP0005` because `pg-types`' `BYTEA` parser is [`postgres-bytea@1.0.0`](https://www.npmjs.com/package/postgres-bytea/v/1.0.0), which still uses the deprecated `new Buffer(...)` constructor.

## Fix

[`postgres-bytea@1.0.1`](https://www.npmjs.com/package/postgres-bytea/v/1.0.1) (released 2025-12-17) swaps `new Buffer(...)` for `Buffer.from || Buffer` and stays inside the `~1.0.0` range that `pg-types` declares, so a lockfile refresh is enough — no `package.json` or source changes are required.

This commit updates the three `pnpm-lock.yaml` references (resolution block, snapshot block, and the single consumer entry under `pg-types@2.x`) from `1.0.0` to `1.0.1`. `pnpm install --lockfile-only --frozen-lockfile` round-trips cleanly.

The previous version of this PR inlined a hand-written replacement parser across the three adapters. Thanks to @aqrln for [pointing out](https://github.com/prisma/prisma/pull/29538#issuecomment-4485884504) that the upstream fix had already landed and the change reduces to a transitive bump.

Refs [`brianc/node-postgres#2426`](https://github.com/brianc/node-postgres/issues/2426).

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7434,8 +7434,8 @@ packages:
     resolution: {integrity: sha512-nAUSGfSDGOaOAEGwqsRY27GPOea7CNipJPOA7lPbdEpx5Kg3qzdP0AaWC5MlhTWV9s4hFX39nomVZ+C4tnGOJQ==}
     engines: {node: '>=12'}
 
-  postgres-bytea@1.0.0:
-    resolution: {integrity: sha512-xy3pmLuQqRBZBXDULy7KbaitYqLcmxigw14Q5sj8QBVLqEwXfeybIKVWiqAXTlcvdvb0+xkOtDbfQMOf4lST1w==}
+  postgres-bytea@1.0.1:
+    resolution: {integrity: sha512-5+5HqXnsZPE65IJZSMkZtURARZelel2oXUEO8rH83VS/hxH5vv1uHquPg5wZs8yMAfdv971IU+kcPUczi7NVBQ==}
     engines: {node: '>=0.10.0'}
 
   postgres-bytea@3.0.0:
@@ -14676,7 +14676,7 @@ snapshots:
     dependencies:
       pg-int8: 1.0.1
       postgres-array: 2.0.0
-      postgres-bytea: 1.0.0
+      postgres-bytea: 1.0.1
       postgres-date: 1.0.7
       postgres-interval: 1.2.0
 
@@ -14760,7 +14760,7 @@ snapshots:
 
   postgres-array@3.0.4: {}
 
-  postgres-bytea@1.0.0: {}
+  postgres-bytea@1.0.1: {}
 
   postgres-bytea@3.0.0:
     dependencies:


kolia-zamnius:
@coderabbitai thanks for the review. Pushed `063fc98` addressing the actionable findings:

**Fixed:**
- **Infinite loop on lone trailing backslash** (legacy octal-escape branch). `postgres-bytea@1.0.0` carried this same defect — paired backslashes were consumed but a remainder odd one left `i` unmoved. The parser now advances by the full run, swallowing any odd leftover. Applied to all three adapters.
- **Renamed `parseByteaTextWire` → `parseBytea`** in `adapter-ppg` to match the sibling adapters.
- **Added `'{\\x68656c6c6f,NULL}'` array test** in `adapter-ppg/src/conversion.test.ts` exercising the hex-format BYTEA_ARRAY path.

**Not addressed (with reason):**
- **Extracting `parseBytea` into a shared helper package**. Three call sites of ~25 lines each; introducing a new workspace package or shared utils module to dedupe seems heavier than the duplication itself, and the function's wire format is genuinely Postgres-specific (won't be reused by mssql / sqlite / etc.). Happy to extract if maintainers prefer that direction.
- **Docstring coverage warning (50%)**. The function name + the inline comment about the deprecation pin already cover the "what / why"; pure-formatting JSDoc on each branch felt like noise. Skipping unless a maintainer wants different.

Build clean across the three packages, `adapter-pg` (47) + `adapter-ppg` (26, +1 new) tests green.

kolia-zamnius:
Pushed `bb533ad` — addresses the second-round findings:

**Fixed:**
- **`\\\012` data corruption (all 3 adapters).** The previous run-counting + full-advance was wrong — a 3-backslash run consumed 3 chars instead of 2, swallowing the `\` that anchors the next octal triple, so e.g. bytes `[0x5C, 0x0A]` (encoded as `\\\012`) decoded to `[0x5C, 0x30, 0x31, 0x32]`. Replaced with the pair-by-pair approach proposed in the comment: emit one `\` per iteration, advance 2 on a paired backslash, advance 1 on a lone leftover (no stall).
- **`normalizeByteaArray` return type** → `(Buffer | null)[]` in `adapter-neon` and `adapter-pg`. `parseArray(input, parseBytea)` passes SQL NULL elements through as `null`, the previous `Buffer[]` was lying to callers. (`adapter-ppg` was already correct since it had the explicit map+null-guard.)
- **Regression test** added in `adapter-ppg/src/conversion.test.ts` covering the `\\\012` case so future changes can't silently regress this path again.

**Still skipping:**
- **Extracting `parseBytea` into a shared helper.** Three call sites of ~22 lines, all in Postgres-flavoured adapters. Setting up a new shared package (or routing through an existing utils one) feels heavier than the duplication, and the wire format is genuinely Postgres-specific (won't be reused by mssql / sqlite / etc.). Happy to extract if a maintainer prefers that direction — open question for review.
- **Docstring coverage 50% warning.** The rest of `conversion.ts` in these packages is mostly `normalize_*` helpers without docstrings, so following local convention.

Build clean across the three packages, `adapter-pg` (47) + `adapter-ppg` (27, +1 new) tests green.

kolia-zamnius:
Pushed `d371eb2` — extracted `parseBytea` into `@prisma/driver-adapter-utils` as suggested.

`packages/driver-adapter-utils/src/parse-bytea.ts` is the single source of truth; the three adapters now `import { parseBytea } from '@prisma/driver-adapter-utils'` and only carry their own thin `convertBytes` / `normalizeByteaArray` wrappers (which differ between adapters — `adapter-ppg` has a separate base64 path for the singleton column, the others go through `parseBytea` directly). All three adapters already depend on `@prisma/driver-adapter-utils`, so no new package wiring.

Net diff is now smaller (+63 / -26), all three adapters build clean, `pnpm -r run build` succeeds across the workspace, `adapter-pg` (47) + `adapter-ppg` (27) tests green.

Marking the consolidation thread as resolved from my side, but happy to revisit naming or location if there's a preferred home.

kolia-zamnius:
Hi @jacek-prisma, sorry to ping — would you have a moment to give this a glance when it fits your queue? CI is also blocked on first-time-contributor approval, so it can't even run until someone with permissions kicks it off. No rush at all if there's other higher-priority work.

kolia-zamnius:
Friendly bump — branch is now up to date with main (just merged in #29553). Pinging @jacek-prisma / @aqrln in case this slipped through the queue. CI is still gated on first-time-contributor approval, so happy to wait on whoever has bandwidth to kick it off. No urgency, just keeping it visible.

packages/driver-adapter-utils/src/parse-bytea.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Decodes the PostgreSQL BYTEA wire format. Two formats are supported: `\x`-prefixed hex
+ * (default since PostgreSQL 9.0) and the legacy octal-escape format.
+ *
+ * Inlined here instead of using `pg-types`' `getTypeParser(BYTEA)` — that path goes through
+ * `postgres-bytea@1.0.0`'s `new Buffer(...)` and emits DEP0005 on every Bytes column. Shared
+ * across the Postgres-flavoured adapters (`adapter-neon`, `adapter-pg`, `adapter-ppg`) so a
+ * single source of truth covers all of them.
+ */
+export function parseBytea(input: string): Buffer {
+  if (/^\\x/.test(input)) {
```
aqrln: nit:
```suggestion
  if (input.startsWith('\\x')) {
```

packages/driver-adapter-utils/src/parse-bytea.ts
```diff
@@ -0,0 +1,32 @@
...
+    return Buffer.from(input.slice(2), 'hex')
+  }
+  let output = ''
```
aqrln: don't use an intermediate string, push the bytes to a `number[]` array or allocate a `Buffer` upfront

aqrln (commented):
Left a few comments.

If I'm reading [this comment](https://github.com/brianc/node-postgres/issues/2426#issuecomment-3663711810) in the referenced issue correctly, the issue seems to be already fixed upstream and we just need to update the transitive dependency, or am I missing something?

kolia-zamnius:
You're right, and thanks for catching it. I just verified against the published tarball:

- [`postgres-bytea@1.0.1`](https://www.npmjs.com/package/postgres-bytea/v/1.0.1) was released 2025-12-17 and switches to `Buffer.from || Buffer`, so DEP0005 is gone.
- It's within the `~1.0.0` range that `pg-types` declares, so no breaking change for consumers.
- The only reason DEP0005 still fires for Prisma users is that the current `pnpm-lock.yaml` on `main` (as of `ab2b847`) still resolves `postgres-bytea` to `1.0.0` — a `pnpm update postgres-bytea` (or a `pnpm.overrides` entry if you want it pinned strictly) closes the gap.

Out of respect for the review time you already spent, I also pushed your two nits on `parse-bytea.ts` in `aef173b` (`startsWith('\\x')` + `number[]` instead of the intermediate string) — but the file itself is moot now, so feel free to disregard.

Happy to close this PR, or to repurpose the branch into the lockfile bump if that's more convenient — whichever fits your workflow. No urgency on my end.

aqrln:
I'd personally just repurpose the branch, but whichever you prefer. Ping me when the lockfile patch is ready, whether in this PR or in a new one, and I'll approve and merge. Thank you!

kolia-zamnius:
@aqrln branch is repurposed and ready — the PR now contains a single commit (7ea21cb76) that just bumps `postgres-bytea` from 1.0.0 to 1.0.1 in `pnpm-lock.yaml` (resolution + integrity + the single consumer ref). No `package.json` or source changes; `pnpm install --lockfile-only --frozen-lockfile` round-trips cleanly. Ready for your review whenever it fits.

aqrln (approved):

aqrln:
Thank you!

Issue #29478, opened 2026-04-16
Slow relational queries due to duplicate IN


barbeque-squared:
How to reproduce: https://github.com/barbeque-squared/prisma-duplicate-in

Note that my repro is specifically for MariaDB because prisma can't generate Join queries there, but this can likely be reproduced on any other database as well by simply turning off that feature.

On my computer it outputs something like
```
100 x 1 duplicate elements took 193ms, deduplicated took 91ms
100 x 1000 duplicate elements took 257ms, deduplicated took 85ms
100 x 15000 duplicate elements took 2279ms, deduplicated took 188ms
100 x 1 unique elements took 102ms, deduplicated unique took 57ms
100 x 1000 unique elements took 214ms, deduplicated unique took 212ms
100 x 15000 unique elements took 2140ms, deduplicated unique took 2135ms
```

The entire crux is that the `deduplicated` simply does `in: [...new Set(theValues)]`

Context:
Consider a simple `users` and `posts` database, where posts have a title, content, and user_id:
```
await prisma.post.findMany({
  include: {
    user: true,
  },
})
```

On MariaDB (because we don't have any kind of join) this results in 2 queries:
```
select posts.title, posts.content, posts.user_id from posts;

select users.user_id, users.user_name from users
where users.user_id IN (1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1);
```
The issue is that second query. This is fine for a handful of values, but not if you get into 1000+ posts that all have the same 1-2 users. Here it's numbers, my reproduction already uses strings, in our production we have _longer_ strings.

I can accept:
* this is a non-issue if MariaDB ever gets proper join queries
* this might confuse if people _explicitly_ query for duplicate values using `IN`

However:
* `IN` is literally a `Set` anyway?
* Specifically in the case of relational queries, these _aren't_ queries done by the user directly, and it 100% behaves like a Set at this point.

I tried searching for things like `performance` and `join` or `relation` or `duplicate`, but all of them are about "just do joins pls" while none of them (that I could find) point towards this seemingly trivial issue that would, at the very least, do _something_ to improve the situation for MariaDB (though simply deduplicating will help every database). In the best case there is probably 1 specific spot in the prisma code where it's recursing into the relations, but I don't know where that is.

sharon77242:
[...]

PR #29544, opened 2026-05-10
fix(query-engine): deduplicate IN clause values in relational queries (#29478)

wahajahmed010:
## Problem
When Prisma generates relational queries with IN clauses (especially on MariaDB where joins are not supported), the IN clause contains duplicate foreign key values. This causes unnecessarily large queries and poor performance when many rows share the same foreign key.

Example from #29478:
```sql
SELECT ... FROM users WHERE user_id IN (1,1,1,1,1,1,1,1,1,1,...);
```

## Solution
Deduplicate the IN clause values in `pairFragmentsWithParams()` before generating the SQL query.

## Changes
- Modified `packages/client-engine-runtime/src/interpreter/render-query.ts` to deduplicate `parameterTuple` values using `Array.from(new Set(value))`
- Added test case `deduplicates duplicate values in IN clause (#29478)` to verify the fix

**AI-assisted contribution.** Human review requested.


tensordreams:
Maintainer cleanup ahead of the v7 branch cut: rebased/merged latest `main` into the branch and pushed (head `a4050e40c`). The dedup now lives in the single `pairFragmentsWithParams` choke point, so the chunking passes and the final render pass all operate on the same deduplicated `parameterTuple` list; it allocates a fresh array rather than mutating the (DeepReadonly) query plan. Added coverage for first-occurrence order preservation and for chunking a large deduplicated IN list. All 218 `@prisma/client-engine-runtime` tests pass locally; prettier and eslint are clean on the changed files.

PR #29546, opened 2026-05-11
feat(cli): warn on global local version mismatch

Iineman2:
/claim #1911

## Summary
- add a global/local Prisma version mismatch warning to `prisma generate`
- compare the running global CLI version against local `prisma` and `@prisma/client` versions when the CLI is installed globally
- avoid noisy false positives for non-exact specifiers such as `workspace:*` or dist tags
- include project-package-manager-specific guidance (`npx`, `pnpm`, `yarn`, or `bun`) so users can run the local Prisma CLI

## Verification
- `nvm use 20.19.4; pnpm --filter prisma test -- src/__tests__/version-mismatch-warning.test.ts`
- `nvm use 20.19.4; pnpm exec turbo build --filter=prisma`
- `git diff --check HEAD~1 HEAD`

Note: I also added focused `Generate.test.ts` coverage for the user-facing `generate` output path, but the full `Generate.test.ts` fixture suite is blocked on this Windows environment by symlink permissions (`EPERM: operation not permitted, symlink ... node_modules/@prisma/client`). The helper suite and filtered CLI build pass locally.

### DIFF
--- a/packages/cli/src/Generate.ts
+++ b/packages/cli/src/Generate.ts
@@ -32,6 +32,7 @@ import { Watcher } from './generate/Watcher'
 import { breakingChangesMessage } from './utils/breakingChanges'
 import { handleNpsSurvey } from './utils/nps/survey'
 import { simpleDebounce } from './utils/simpleDebounce'
+import { getVersionMismatchWarning, type VersionMismatchWarningOptions } from './utils/version-mismatch-warning'
 
 const pkg = eval(`require('../package.json')`)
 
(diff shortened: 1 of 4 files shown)


Iineman2:
CLA is now signed, and CodeRabbit is clean after the follow-up simplification commit. Latest local targeted verification remains green:

- `pnpm --filter prisma test -- src/__tests__/version-mismatch-warning.test.ts`
- `pnpm exec turbo build --filter=prisma`

Happy to adjust quickly if maintainers prefer different wording or placement for the warning.

tensordreams:
Thanks — this is close: gating on `isCurrentBinInstalledGlobally()`, reading `@prisma/client` via `getInstalledPrismaClientVersion()` (node_modules first), and rejecting `workspace:*`/dist-tags/ranges in `normalizeVersion()` are all right.

One blocking ask: the `prisma` self-version side reads the `package.json` specifier (via `getPackageJsonVersion`/`package-up`) rather than the installed `node_modules/prisma/package.json`. `normalizeVersion()` mitigates the worst of it, but a `^7.0.0` specifier with a newer installed version still misreports. Please read the installed version for both packages, and get a full CI run on the branch (we saw the note about the Windows symlink issue blocking local verification).


---

aqrln:
Merged in https://github.com/prisma/prisma/pull/29593

PR #29556, opened 2026-05-15
fix(query-engine): deduplicate IN clause values in relational queries

### LINKED ISSUE #29478: Slow relational queries due to duplicate IN

wahajahmed010:
## Problem

When Prisma generates relational queries that use IN clauses (especially on MariaDB where joins are not supported), the IN clause can contain duplicate values. For example, a `findMany` with `include` on a relation generates queries like:

```sql
SELECT ... FROM users WHERE user_id IN (1,1,1,1,...)
```

This causes:
- Unnecessary query payload bloat
- Reduced performance as the database engine processes redundant values
- Potentially incorrect result counts when the database interprets duplicates differently

Referenced in issue [#29478](https://github.com/prisma/prisma/issues/29478).

## Solution

Deduplicate `parameterTuple` and `parameterTupleList` values in `render-query.ts` before building SQL query fragments:

- **parameterTuple**: Deduplicates scalar values in IN clauses using a Map-based approach
- **parameterTupleList**: Deduplicates tuple lists using JSON.stringify-based comparison

The deduplication:
- Preserves insertion order (first occurrence is kept)
- Works for primitive values (numbers, strings) using direct comparison
- Works for complex objects using JSON.stringify
- Only affects IN clause values, not other query parameters

## Files Changed

- `packages/client-engine-runtime/src/interpreter/render-query.ts` (+30 lines, -4 lines)

## Testing

Existing tests cover IN clause rendering with `parameterTuple` and `parameterTupleList`. The fix is backward compatible as it only removes duplicate values that would have been included anyway.

Closes #29478

### DIFF
--- a/packages/client-engine-runtime/src/interpreter/render-query.ts
+++ b/packages/client-engine-runtime/src/interpreter/render-query.ts
@@ -27,7 +27,7 @@ export function renderQuery(
     case 'rawSql':
       return [renderRawSql(dbQuery.sql, args, dbQuery.argTypes)]
     case 'templateSql': {
-      const chunks = dbQuery.chunkable ? chunkParams(dbQuery.fragments, args, maxChunkSize) : [args]
+      const chunks = chunkParams(dbQuery.fragments, args, maxChunkSize)
       return chunks.map((params) => {
         const rendered = renderTemplateSql(dbQuery.fragments, dbQuery.placeholderFormat, params, dbQuery.argTypes)
         if (maxChunkSize !== undefined && rendered.args.length > maxChunkSize) {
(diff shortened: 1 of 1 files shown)


SevInf:
@wahajahmed010 could you sign a CLA?

tensordreams:
Hi @wahajahmed010 — thanks for this, and apologies for the slow reply.

We are closing this one: this duplicates #29544 from the same branch. We'll keep the earlier PR as the single place for that work.

PR #29561, opened 2026-05-17
fix(format): normalize schema line endings

devnull37:
Fixes #8548.\n\nThis normalizes formatter output to LF line endings in the shared formatSchema internals path before linting/returning formatted schemas. It also adds a regression test with CRLF input to assert formatted schema content contains no carriage returns.\n\nVerification: not run locally; Prisma requires pnpm >=10.15 <11, while the available installed pnpm is 11.0.8.


devnull37:
Addressed CodeRabbits CR-only line-ending test request in c80e20a.

SevInf:

PR #29562, opened 2026-05-17
fix(cli): confirm global Prisma usage

devnull37:
/claim #1911

Implement a confirmation flow when the global Prisma CLI detects a local Prisma install.

Issue: https://github.com/prisma/prisma/issues/2738
Amount: $150

Verification:
- git diff --check
- Focused pnpm test could not be executed here because Corepack cannot fetch pnpm from npm registry on this machine (certificate chain failure)

Changed paths:
- packages/cli/src/CLI.ts
- packages/cli/src/bootstrap/Bootstrap.ts
- packages/cli/src/utils/localPrisma.ts
- packages/cli/src/__tests__/commands/CLI.test.ts

### DIFF
--- a/packages/cli/src/CLI.ts
+++ b/packages/cli/src/CLI.ts
@@ -6,6 +6,7 @@ import { arg, drawBox, format, HelpError, isError, link, unknownCommand } from '
 import { bold, dim, green, red } from 'kleur/colors'
 
 import { runCheckpointClientCheck } from './utils/checkpoint'
+import { confirmUsingGlobalPrisma } from './utils/localPrisma'
 import { printUpdateMessage } from './utils/printUpdateMessage'
 import { Version } from './Version'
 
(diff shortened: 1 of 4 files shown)


tensordreams:
Thanks — and to be fair to this approach: #2738 does mention "warn or even require explicit confirmation", so a prompt isn't inherently wrong. But as implemented it can't ship:

1. The check fires on the mere existence of `node_modules/.bin/prisma` — it never checks whether the *running* CLI is the global one (`isCurrentBinInstalledGlobally()` from `@prisma/internals`). Since the local bin always exists when you correctly run the local CLI, this prompts on every legitimate local/npx invocation.
2. In non-interactive environments `canPrompt()` is false, the function returns `false`, and the command aborts with a `HelpError` — every CI pipeline or script in a project with a local install breaks. Whatever shape this takes, it must never block non-interactive runs.
3. There's no version comparison; the issues are specifically about a version mismatch.

If you rework it, the shape we'd review: gate on `isCurrentBinInstalledGlobally()`, read the installed local version from `node_modules/prisma/package.json`, and print a non-blocking advisory on real mismatch. Also, please sign the CLA (link in the thread above).


---

tensordreams:
Hi @devnull37 — thanks for this, and apologies for the slow reply.

PR #29563, opened 2026-05-17
feat(cli): show current prisma install path

devnull37:
/claim #7771

Adds the current Prisma install path to `prisma -v` output and covers it with a regression snapshot.

Verification: `corepack pnpm --filter prisma test -- Version.test.ts` was blocked locally because this checkout has no `node_modules`, and Corepack could not fetch `pnpm` until TLS was relaxed.

### DIFF
--- a/packages/cli/src/Version.ts
+++ b/packages/cli/src/Version.ts
@@ -119,8 +119,8 @@ export class Version implements Command {
   }
 
   private async getPrismaInstallPath(): Promise<string> {
-    const prismaEntryPoint = process.argv[1] ? fs.realpathSync(process.argv[1]) : undefined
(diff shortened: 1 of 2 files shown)


tensordreams:
Thanks — this is a solid take on #7771: realpathing the entrypoint, walking to the package root, and falling back through `resolvePkg` is robust, and both the snapshot and the sanitizer in `Version.test.ts` are correctly updated.

A few asks before we can consider it:

1. Please sign the CLA (link in the thread above) — we can't merge without it.
2. Repo convention prefers native `#private` members over the TypeScript `private` keyword for the new helpers (see AGENTS.md).
3. Please run `pnpm lint` locally — the new `fs`/`path` import placement looks like it may trip `simple-import-sort`.


---

tensordreams:
Hi @devnull37 — thanks for this, and apologies for the slow reply.

PR #29570, opened 2026-05-20
fix(format): normalize schema line endings

daniel-deblas:
Fixes #8548.

This normalizes CRLF line endings returned by the formatter to LF before writing schema files and before comparing formatted output in --check mode.


daniel-deblas:
Added a focused regression test for --check mode where the formatter returns CRLF line endings while the original schema uses LF. The command now still reports the schema as formatted.

SevInf:
No

PR #29574, opened 2026-05-21
fix(format): normalize schema formatter line endings

kkk67-hub:
## Summary
- Normalize formatted Prisma schemas to LF before validation, check, and write.
- Keep prisma format --check and written schema output stable if formatter output contains CRLF line endings.
- Add regression coverage for CRLF formatter output.

Fixes #8548.

## Test
- npx pnpm@10.15.1 --filter prisma... build
- npx pnpm@10.15.1 --filter prisma exec jest --silent src/__tests__/commands/Format.test.ts
- npx pnpm@10.15.1 exec prettier --check packages/cli/src/Format.ts packages/cli/src/__tests__/commands/Format.test.ts

Note: npx pnpm@10.15.1 --filter prisma test -- src/__tests__/commands/Format.test.ts ran the target Jest suite successfully, then failed in unrelated package-wide Vitest tests (studio-server/Init) due local environment behavior.

### DIFF
--- a/packages/cli/src/Format.ts
+++ b/packages/cli/src/Format.ts
@@ -13,6 +13,7 @@ import {
   HelpError,
   printSchemaLoadedMessage,
   validate,
+  type MultipleSchemas,
 } from '@prisma/internals'
 import { bold, dim, red, underline } from 'kleur/colors'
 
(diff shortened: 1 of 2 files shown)


SevInf:
No

PR #29591, opened 2026-05-26
feat(cli): show Prisma CLI path in version output

symbiosis-institute:
/claim #7771
Fixes #7771

## Summary

- add the running Prisma CLI package path to `prisma -v` / `prisma version`
- resolve the path from the CLI package metadata so the reported location matches the CLI that is actually executing
- add test coverage for both the human-readable output and the JSON output

## Why

Issue #7771 asks for a way to tell which Prisma CLI installation is actually running. This helps when `prisma` resolves to a parent or otherwise unexpected install and the current version output does not reveal that.

## Implementation

- compute `prismaCliPath` from `path.dirname(require.resolve('../package.json'))`
- append a `Prisma CLI Path` row to the version output table
- update the inline snapshot for the version command
- add a JSON regression test asserting the `prisma-cli-path` field

## Verification

- `pnpm --filter "prisma..." build`
- `pnpm test src/__tests__/commands/Version.test.ts`
- `pnpm exec eslint packages/cli/src/Version.ts packages/cli/src/__tests__/commands/Version.test.ts`
- `node packages/cli/build/index.js -v`
- `node packages/cli/build/index.js version --json`

This contribution was prepared with AI assistance under human direction and review. I inspected the change, verified the behavior described here, and am responsible for the submission.

### DIFF
--- a/packages/cli/src/Version.ts
+++ b/packages/cli/src/Version.ts
@@ -1,3 +1,5 @@
+import path from 'node:path'
+
 import type { PrismaConfigInternal } from '@prisma/config'
 import { enginesVersion } from '@prisma/engines'
 import {
@@ -78,6 +80,7 @@ export class Version implements Command {
 
     const prismaClientVersion = await getInstalledPrismaClientVersion()
     const typescriptVersion = await getTypescriptVersion()
+    const prismaCliPath = path.dirname(require.resolve('../package.json'))
 
     const rows = [
       [packageJson.name, packageJson.version],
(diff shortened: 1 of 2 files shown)


tensordreams:
Thanks — this looks correct end to end: the resolution reuses the existing `require('../package.json')` pattern, the snapshot alignment matches `formatTable`'s padding, the sanitizer is right, and the JSON test asserting the slugified `prisma-cli-path` key is a nice touch. No blocking issues from our review; we'll get full CI running on it.

tensordreams:
Hi @symbiosis-institute — thanks for this, and apologies for the slow reply.

PR #29593, opened 2026-05-26
feat(cli): warn on global local version mismatch

### LINKED ISSUE #1911: Warn on mismatch between global `prisma2` and local `prisma2` or `@prisma/client`

jamilahmadzai:
/claim #1911
Fixes #1911

## Summary
- warn during `prisma generate` when the current Prisma CLI is globally installed and differs from local installed `prisma` or `@prisma/client`
- compare against package versions found in local `node_modules` rather than package.json specifiers, so locked/workspace installs are not misread
- keep generation non-blocking; `--no-hints` still hides normal hints but keeps this warning visible

## Coverage
- added unit coverage for non-global runs, missing packages, matching versions, single/both package mismatches, empty global version, and node_modules resolution
- added Generate coverage for `--no-hints` and non-`prisma-client-js` generator output

## Validation
- `corepack pnpm --filter prisma test src/__tests__/globalLocalVersionMismatch.test.ts`
- `corepack pnpm --filter prisma test src/__tests__/commands/Generate.test.ts`
- `corepack pnpm exec eslint packages/cli/src/Generate.ts packages/cli/src/utils/globalLocalVersionMismatch.ts packages/cli/src/__tests__/globalLocalVersionMismatch.test.ts packages/cli/src/__tests__/commands/Generate.test.ts`
- `corepack pnpm exec prettier --check packages/cli/src/Generate.ts packages/cli/src/utils/globalLocalVersionMismatch.ts packages/cli/src/__tests__/globalLocalVersionMismatch.test.ts packages/cli/src/__tests__/commands/Generate.test.ts`
- `corepack pnpm --filter prisma build`

Algora payout: @jamilahmadzai via Algora platform.


tensordreams:
Thanks — this is the strongest implementation among the open PRs for #1911: real `isCurrentBinInstalledGlobally()` gate, installed-version reads from `node_modules/<pkg>/package.json` for both packages, the warning reaches all generators (not just `prisma-client-js`), and the failure paths are guarded so generation never breaks.

Asks before merge:

1. Anchor the local lookup at the project root of the schema (`schemaContext.schemaRootDir`) rather than `baseDir` — `baseDir` is the shell cwd unless `--config` is used, so `prisma generate --schema /path/to/app/...` from another directory misses `/path/to/app/node_modules/prisma`.
2. Rename `globalLocalVersionMismatch.ts` to kebab-case (`global-local-version-mismatch.ts`) per repo conventions.
3. Nice-to-have: package-manager-aware suggestion (`pnpm`/`yarn`/`bun`) instead of `npx` only.


---

jamilahmadzai:
Thanks, pushed follow-up commit 2c302a3e4 addressing the merge asks:

- anchored the local package-version lookup at `schemaContext.schemaRootDir` instead of `baseDir`, so `prisma generate --schema /path/to/app/...` checks the app-local `node_modules`
- renamed `globalLocalVersionMismatch.ts` to `global-local-version-mismatch.ts`
- added a regression test covering the parent-directory/schema-root lookup case

Local validation: `git diff --check` passed. Focused Jest did not complete in my fresh checkout because the workspace package build outputs for `@prisma/*` were missing after dependency setup; I’ll watch the PR checks for the full result.

packages/cli/src/utils/global-local-version-mismatch.ts
```diff
@@ -0,0 +1,109 @@
...
+    })),
+  )
+
+  const mismatches = localVersions.reduce<LocalPackageVersionMismatch[]>((acc, { packageName, version }) => {
+    if (version && version !== globalVersion) {
+      acc.push({ packageName, globalVersion, localVersion: version })
+    }
+
+    return acc
+  }, [])
```
aqrln: this should be a `map` and not `reduce`

tensordreams: Updated the mismatch collection to use `map` followed by a typed filter, as suggested. Pushed in 952d96e73.

packages/cli/src/__tests__/commands/Generate.test.ts
```diff
@@ -266,6 +267,54 @@ it('should hide hints with --no-hints', async () => {
...
+it('should not fail generate when the global/local version warning lookup fails', async () => {
+  ctx.fixture('example-project')
+  const generate = new Generate(jest.fn(), jest.fn().mockResolvedValue({ prompted: false }), {
+    getGlobalLocalVersionMismatchWarning: () => Promise.reject(new Error('version lookup failed')),
+  })
+  const output = stripAnsi((await generate.parse([], await ctx.config())).toString())
+
+  expect(output).toContain('Generated Prisma Client')
+  expect(output).not.toContain('version lookup failed')
+}, 60_000)
```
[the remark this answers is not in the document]
tensordreams: Added a named rejecting mock and asserted that the warning lookup is called exactly once, while retaining the successful-generation and swallowed-error assertions. Pushed in 985724d2a.

PR #29594, opened 2026-05-26
feat(cli): show prisma cli path in version output

jamilahmadzai:
## Summary
- add `Prisma CLI Path` to `prisma -v` and `prisma version --json`
- resolve the Prisma CLI package root directly with `path.dirname(require.resolve('../package.json'))`
- keep JSON output coverage and snapshot sanitization for the machine-specific path

## Validation
- `pnpm --filter prisma... build`
- `pnpm --filter prisma test src/__tests__/commands/Version.test.ts`
- `git diff --check`

/claim #7771
Fixes #7771
Algora payout: @jamilahmadzai via Algora platform.

### DIFF
--- a/packages/cli/src/Version.ts
+++ b/packages/cli/src/Version.ts
@@ -15,8 +15,10 @@ import {
   resolveEngine,
   wasm,
 } from '@prisma/internals'
+import fs from 'fs'
 import { bold, dim, red } from 'kleur/colors'
 import os from 'os'
+import path from 'path'
 
 import { getInstalledPrismaClientVersion } from './utils/getClientVersion'
 
@@ -78,6 +80,7 @@ export class Version implements Command {
 
     const prismaClientVersion = await getInstalledPrismaClientVersion()
     const typescriptVersion = await getTypescriptVersion()
+    const prismaCliPath = getPrismaCliPath()
 
     const rows = [
       [packageJson.name, packageJson.version],
@@ -92,6 +95,7 @@ export class Version implements Command {
 
       ['Default Engines Hash', enginesVersion],
       ['Studio', packageJson.dependencies['@prisma/studio-core']],
+      ['Prisma CLI Path', prismaCliPath],
     ]
 
     /**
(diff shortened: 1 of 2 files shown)


TejasAmle (commented):
Solid, useful addition — `prisma version` is one of the first things people paste in bug reports, and the CLI path often makes the difference between debugging a global install vs a local one vs a symlink. Having it in the output removes a manual lookup step.

A few thoughts:

**1. `findPrismaPackageRoot` walk — missing a max-depth guard**
The `while (true)` loop stops only when a `package.json` with `name: prisma` is found, or when `path.dirname(current) === current` (filesystem root). On most systems this is fine, but on unusual mounts or deeply nested symlink trees an explicit depth cap (e.g. 20 iterations) would make the function resilient against edge cases where the root condition is never reached.

**2. Fallback to `path.dirname(realEntryPoint)` — is this always a useful path?**
If `findPrismaPackageRoot` returns `null`, the function returns the directory containing the entry point script. This could be a temp directory (e.g. from `npx`) rather than the installed package root. It might be worth logging or labeling this differently — e.g. `path.dirname(realEntryPoint) + ' (entry point dir)'` — so users know it is an approximation.

**3. Test snapshot uses `sanitized_cli_path`**
Good call to sanitize the path in snapshots — CI paths are always machine-specific. The `getPrismaCliPath` unit tests with `tmpdir` are well-structured and cover the happy path (found), the fallback (not a prisma package), and the `undefined` entry point case.

**4. `fs` import style inconsistency**
The existing file uses `import os from 'os'` (default import) and this PR adds `import fs from 'fs'` and `import path from 'path'` in the same style — consistent, good.

**5. JSON output key naming**
The JSON output uses `prisma-cli-path` (kebab-case) while other keys in the JSON output use... what convention? Worth checking that the casing matches the existing JSON output format for consistency with tools that parse `prisma version --json`.

tensordreams:
Thanks — this works and is carefully engineered (symlink resolution, package-root walk with a `name === 'prisma'` check, unit-tested helper, JSON coverage). Honest feedback: it's heavier than the problem needs — the same output can be achieved with `path.dirname(require.resolve('../package.json'))` in one line. If we move forward with this one we'd likely ask to slim the helper down; no action needed from you right now beyond keeping CI green.


---

tensordreams:
Hi @jamilahmadzai — thanks for this, and apologies for the slow reply.

PR #29598, opened 2026-05-27
feat(cli): warn when global prisma CLI differs from locally installed version

### LINKED ISSUE #1911: Warn on mismatch between global `prisma2` and local `prisma2` or `@prisma/client`

Rome-1:
Closes #1911.

## What

When `prisma generate` runs from a globally-installed `prisma` CLI in a project
that has a different `prisma` version under `node_modules`, the CLI now emits
a single advisory line — same visual style as the existing
`prisma`/`@prisma/client` version-mismatch warning — pointing the user toward
running the local version (`npx prisma generate` or your package manager's
equivalent).

## How

- Add `getInstalledPrismaCliVersion(cwd)` in
  `packages/cli/src/utils/getClientVersion.ts` — walks up from `cwd` looking
  for `node_modules/prisma/package.json`. Explicit filesystem walk (not
  `require.resolve`) because the running prisma CLI is itself the `prisma`
  package, so Node's resolver would otherwise find the CLI's own
  `package.json`.
- Add `shouldWarnGlobalLocalCliMismatch({ ... })` — pure predicate. Only
  triggers when (a) `isCurrentBinInstalledGlobally()` is truthy, (b) a
  locally-installed prisma version is found under `node_modules`, and (c)
  both versions match exact semver and differ. Skips `workspace:*`, dist-tags
  (`latest`, `next`), ranges (`^x`, `~x`, `>=x`), `file:`/`link:` specifiers,
  and partial versions to avoid false positives.
- In `packages/cli/src/Generate.ts`, after the existing `versionsWarning`
  block, append a `globalLocalCliWarning` to the `hint` (same style; same
  `--no-hints` and `logger.should.warn()` gating as the existing warning).

## Diff size

74 lines of production code added, 2 lines modified across 2 files. No new
source files (helpers live in the existing `getClientVersion.ts`). 126 lines
of new tests in `src/utils/getClientVersion.test.ts` (22 cases), placed
alongside the source file, matching the convention of existing util tests
(`client-output-path.test.ts`, `loadConfig.test.ts`, `ppgInfo.test.ts`).

## Tests

22 unit tests on the two new helpers (`pnpm --filter prisma test src/utils/getClientVersion.test.ts`):

- exact-semver mismatch → warn
- exact-semver match → no warn
- not globally installed → no warn
- no local prisma installed → no warn
- 13 non-exact specifier cases (`workspace:*`, `latest`, `next`, `^7.0.0`,
  `~7.0.0`, `>=7.0.0`, `file:../prisma`, `link:../prisma`,
  `npm:prisma@7.0.0`, `7`, `7.0`, empty string, `workspace:^7.0.0`) → all
  no-warn
- prerelease semver `7.0.0-beta.1` → warn
- non-exact global version (e.g. placeholder strings) → no warn
- helper reads `node_modules/prisma/package.json` correctly
- returns null when prisma is not installed
- returns null when version field is absent

## Documentation

This change adds a runtime advisory line to the existing `prisma generate`
hint block; it does not introduce new commands, flags, or configuration. The
existing `prisma@x` ↔ `@prisma/client@x` mismatch warning is similarly
surfaced only as runtime output and is not documented in the CLI reference,
so I haven't filed a docs-update PR. Happy to open one if a maintainer would
prefer the warning behavior described in the reference.

## Disclosure

Patch drafted with AI assistance (Claude Code); reviewed line-by-line by the
author, locally tested, and CLA-signed by the author.

/claim #1911

### DIFF
--- a/packages/cli/src/utils/getClientVersion.test.ts
+++ b/packages/cli/src/utils/getClientVersion.test.ts
@@ -45,26 +45,22 @@ describe('shouldWarnGlobalLocalCliMismatch', () => {
     ).toBe(false)
   })
 
+  // Defensive: `installedLocalPrismaCliVersion` is always read from an installed
+  // `node_modules/prisma/package.json` `version` field, so in practice it is a
+  // well-formed version. These cases lock in that a malformed / non-exact-semver
+  // `version` value never produces a (misleading) warning.
   test.each([
-    ['workspace:*', 'workspace:*'],
-    ['workspace alias', 'workspace:^7.0.0'],
-    ['dist-tag latest', 'latest'],
-    ['dist-tag next', 'next'],
-    ['caret range', '^7.0.0'],
-    ['tilde range', '~7.0.0'],
-    ['gte range', '>=7.0.0'],
-    ['file specifier', 'file:../prisma'],
-    ['link specifier', 'link:../prisma'],
-    ['npm alias', 'npm:prisma@7.0.0'],
+    ['empty string', ''],
     ['partial version', '7'],
     ['partial version with minor', '7.0'],
-    ['empty string', ''],
-  ])('does not warn for non-exact-semver local specifier (%s)', (_label, specifier) => {
+    ['leading v', 'v7.0.0'],
+    ['non-version string', 'not-a-version'],
+  ])('does not warn for a non-exact-semver installed version (%s)', (_label, version) => {
     expect(
       shouldWarnGlobalLocalCliMismatch({
         isGlobalInstall: true,
         globalCliVersion: '7.0.0',
-        installedLocalPrismaCliVersion: specifier,
+        installedLocalPrismaCliVersion: version,
       }),
     ).toBe(false)
   })
@@ -119,6 +115,15 @@ describe('getInstalledPrismaCliVersion', () => {
     const prismaDir = path.join(tmpDir, 'node_modules', 'prisma')
     fs.mkdirSync(prismaDir, { recursive: true })
     fs.writeFileSync(path.join(prismaDir, 'package.json'), JSON.stringify({ name: 'prisma' }))
+
+    const result = await getInstalledPrismaCliVersion(tmpDir)
+    expect(result).toBeNull()
+  })
+
+  test('returns null when the version field is not a string', async () => {
+    const prismaDir = path.join(tmpDir, 'node_modules', 'prisma')
+    fs.mkdirSync(prismaDir, { recursive: true })
+    fs.writeFileSync(path.join(prismaDir, 'package.json'), JSON.stringify({ name: 'prisma', version: 7 }))
 
     const result = await getInstalledPrismaCliVersion(tmpDir)
     expect(result).toBeNull()
--- a/packages/cli/src/utils/getClientVersion.test.ts
+++ b/packages/cli/src/utils/getClientVersion.test.ts
@@ -1,0 +1,126 @@
+import fs from 'node:fs'
+import os from 'node:os'
+import path from 'node:path'
+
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/getClientVersion.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    ['dist-tag next', 'next'],
+    ['caret range', '^7.0.0'],
+    ['tilde range', '~7.0.0'],
+    ['gte range', '>=7.0.0'],
+    ['file specifier', 'file:../prisma'],
+    ['link specifier', 'link:../prisma'],
+    ['npm alias', 'npm:prisma@7.0.0'],
+    ['partial version', '7'],
+    ['partial version with minor', '7.0'],
+    ['empty string', ''],
```
aqrln: How can any of these possibly be in `node_modules/prisma/package.json` (or the global equivalent)?

tensordreams:
Thanks — the helper itself is well done: the strict exact-semver check on both sides correctly rejects `workspace:*`, ranges, dist-tags, and `file:`/`link:`/`npm:` specifiers, and the walk-up read of `node_modules/prisma/package.json` is the right way to get the installed version.

Two blocking integration issues:

1. The warning is built inside the `if (prismaClientJSGenerator)` branch and the final message is gated by `hasJsClient` — so projects on the Prisma 7 default `prisma-client` (PrismaClientTs) generator, or custom generators, never see it. The warning needs to be emitted independent of which generator ran (the message gate needs an `else` path / loosened condition).
2. `getInstalledPrismaCliVersion(baseDir)` anchors at the shell cwd; `prisma generate --schema /path/to/app/...` from another directory misses `/path/to/app/node_modules/prisma`. Anchor at `schemaContext.schemaRootDir` (or check it first).


---

packages/cli/src/utils/getClientVersion.test.ts
Rome-1: Good catch — you're right, those were dependency *specifiers*, not values that can appear in an installed `node_modules/prisma/package.json` `version` field. `shouldWarnGlobalLocalCliMismatch` only ever receives concrete installed versions (the running CLI's own `pkg.version` and the version read by `getInstalledPrismaCliVersion`), so testing `workspace:*` / `^7.0.0` / `file:` / `npm:` overstated the real coverage.

Fixed in 4ac898a:
- Replaced the specifier cases with realistic malformed-*version* values (`''`, `7`, `7.0`, `v7.0.0`, `not-a-version`) and added a comment noting the check is defensive.
- Clarified the predicate's JSDoc so it no longer claims to handle specifiers.
- Also type-guarded the read `version` (a non-string field now yields `null`), with a test.

I kept the comparison as exact-string (so a `+build` metadata difference would still surface) since this is an identity advisory — happy to switch to parsed-semver comparison or a stricter semver regex if you'd prefer.

Rome-1:
Thanks — both fixed in `6f30eef`:

1. The warning was coupled to the `prisma-client-js` hint path (built inside `if (prismaClientJSGenerator)` and gated by `hasJsClient`). It's now computed and appended independent of which generator ran, so it fires for the `prisma-client` default, custom/non-JS generators, and `--generator` runs too. Still gated by `logger.should.warn()` and suppressed on the error path.
2. The local-install lookup now anchors at `schemaContext.schemaRootDir` rather than the shell cwd, so `prisma generate --schema <dir>` resolves that project's `node_modules/prisma`.

One more I noticed while in here: the warning also never fires in `--watch` mode (it's built only in the `!watchMode` branch). Happy to extend it to the watch path too if you'd like — let me know and I'll add it in this PR.

tensordreams:
Hi @Rome-1 — thanks for this, and apologies for the slow reply.

PR #29606, opened 2026-05-28
fix(cli): avoid CRLF in prisma format output

EUM27:
Fixes #8548.

On Windows, prisma format could write a final CRLF even when all other lines are LF. This normalizes formatter output to LF before writing schema files, and adds a regression test.


tensordreams:
Hi @EUM27 — thanks for this, and apologies that it sat for so long.

We are closing this one as inactive: the CLA has not been signed, no CI has ever run on it, and the Windows-only CRLF behaviour it describes was never substantiated against `Format.ts`.

PR #29607, opened 2026-05-28
fix(cli): normalize trailing CRLF to LF in prisma format

nagyminyo-cmyk:
Fixes #8548 — prisma format produces trailing CRLF instead of LF on Windows.

The Wasm formatter (Rust) preserves CRLF on Windows. The fix normalizes the trailing CRLF to LF in the TypeScript CLI layer before writing the file.

```typescript
  const normalizedData = data.endsWith("\r\n") ? data.replace(/\\r\n$/, "\n") : data
  await fs.writeFile(filename, normalizedData)
}
```

### DIFF
--- a/packages/cli/src/Format.ts
+++ b/packages/cli/src/Format.ts
@@ -104,9 +104,10 @@ Or specify a Prisma schema path
 
     for (const [filename, data] of formattedDatamodel) {
       const normalizedData = data.endsWith('
-') ? data.replace(/\r
+') ? data.replace(/
 $/, '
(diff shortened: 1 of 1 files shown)


nagyminyo-cmyk:
Bounty: https://algora.io/prisma/prisma-bounty-8548 — $50

nagyminyo-cmyk:
Closing this. I missed the pinned maintainer guidance: this targets the CLI and forces LF, both explicitly out of scope, and the bounty is inactive. Sorry for the noise.

PR #29612, opened 2026-05-30
fix(adapter-mariadb): release pooled connections and harden transaction cleanup

andychinghk01:
## Summary

Fixes #28964.

The MariaDB adapter previously called `mariadb.Connection#end()` to dispose of a transaction connection. For a `PoolConnection` `end()` happens to be aliased to `release()`, so in practice it does return the connection to the pool — but the intent is ambiguous, and the same code path is reachable with a plain `mariadb.Connection` (e.g. in tests), where `end()` permanently closes the connection. The transaction failure path also removed the connection's `error` listener *after* releasing the connection and without a `try/finally`, so a throw during teardown would leak the listener onto the next caller.

## Changes

- Introduce a small `releaseConnection()` helper that prefers `PoolConnection#release()` and falls back to `Connection#end()`.
- Use it from `MariaDbTransaction#commit` and `#rollback`.
- In `startTransaction`'s catch path, detach the `error` listener **before** releasing the connection and wrap the release in `try/finally` so the connection is always returned to the pool, even if cleanup throws.

## Tests

Added unit tests in `packages/adapter-mariadb/src/mariadb.test.ts`:

- `release()` is used for pool connections; `end()` fallback for non-pool connections.
- The transaction `error` listener is detached on `commit`.
- The transaction `error` listener is detached and the connection is released on `startTransaction` failure.

All 26 adapter unit tests pass (`pnpm --filter @prisma/adapter-mariadb test`).

## Out-of-band verification

I also ran an end-to-end harness with Docker (`mariadb:10.11`) covering:

1. `connectionLimit=100`, 100 concurrent transactions → 100 idle conns after release ✅
2. Two sequential rounds of 100 → pool reuses, stays at 100 ✅
3. 200 concurrent vs `connectionLimit=100` → capped at 100, no leak ✅
4. `idleTimeout=10s` → pool drains to `minimumIdle` after the window ✅
5. After drain, a new burst grows the pool again ✅

Closes #28964

### DIFF
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -397,8 +397,6 @@ type ArrayModeResult = unknown[][] & { meta?: mariadb.FieldInfo[]; affectedRows?
  * which breaks pool reuse and the `idleTimeout` / `minimumIdle` lifecycle.
  * Falls back to `end()` when `release` is not available (e.g. unit tests with
  * a non-pool connection).
- *
- * See https://github.com/prisma/prisma/issues/28964
  */
 async function releaseConnection(conn: mariadb.Connection): Promise<void> {
   const poolConn = conn as mariadb.PoolConnection
--- a/packages/adapter-mariadb/src/mariadb.ts
+++ b/packages/adapter-mariadb/src/mariadb.ts
@@ -103,7 +103,7 @@ class MariaDbTransaction extends MariaDbQueryable<mariadb.Connection> implements
       this.onError(err)
     } finally {
       this.cleanup?.()
-      await this.client.end()
+      await releaseConnection(this.client)
     }
   }
 
@@ -116,7 +116,7 @@ class MariaDbTransaction extends MariaDbQueryable<mariadb.Connection> implements
       this.onError(err)
     } finally {
       this.cleanup?.()
-      await this.client.end()
+      await releaseConnection(this.client)
     }
   }
 
@@ -188,7 +188,7 @@ export class PrismaMariaDbAdapter extends MariaDbQueryable<mariadb.Pool> impleme
     conn.on('error', onError)
 
     const cleanup = () => {
-      conn.removeListener('error', onError)
+      conn.off('error', onError)
     }
 
     try {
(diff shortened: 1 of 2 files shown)


aqrln (approved):
LGTM, thank you!

tensordreams:
Hi @andychinghk01 — thanks for this contribution, and sorry for the slow response.

https://cla-assistant.io/prisma/prisma?pullRequest=29612

andychinghk01:
Hi @tensordreams , thanks for reaching out. I've just signed the CLA and rebased onto the latest main to resolve the merge conflicts. Let me know if there's anything else needed.

packages/adapter-mariadb/src/mariadb.ts
```diff
@@ -384,3 +390,21 @@ export function rewriteConnectionString(url: URL): URL {
...
+ * See https://github.com/prisma/prisma/issues/28964
+ */
+async function releaseConnection(conn: mariadb.Connection): Promise<void> {
+  const poolConn = conn as mariadb.PoolConnection
+  if (typeof poolConn.release === 'function') {
+    await poolConn.release()
+  } else {
+    await conn.end()
+  }
+}
```
aqrln, line 408: This is just cosmetic, right? It doesn't actually fix the issue, the `try/finally` above does. We should remove the confusing issue reference in the doc comment since it's unrelated to the issue.

packages/adapter-mariadb/src/mariadb.ts
```diff
@@ -384,3 +390,21 @@ export function rewriteConnectionString(url: URL): URL {
...
 type ArrayModeResult = unknown[][] & { meta?: mariadb.FieldInfo[]; affectedRows?: number; insertId?: BigInt }
+
+/**
+ * Returns a pool connection back to its pool via `PoolConnection#release()`.
+ * Using `end()` instead would permanently close a plain `mariadb.Connection`,
+ * which breaks pool reuse and the `idleTimeout` / `minimumIdle` lifecycle.
+ * Falls back to `end()` when `release` is not available (e.g. unit tests with
+ * a non-pool connection).
+ *
+ * See https://github.com/prisma/prisma/issues/28964
```
aqrln:
```suggestion
```

aqrln (approved):

PR #29613, opened 2026-05-30
[codex] show prisma cli path in version output

kitwongpixel:
/claim #7771
Fixes #7771

## Summary

- Add the installed Prisma CLI path to `prisma -v` / `prisma version` output.
- Resolve the path from the CLI package root so the reported location matches the CLI that is actually executing.
- Add regression coverage for both the human-readable table output and the JSON output.

## Validation

- `git diff --check`
- `pnpm --filter prisma... build`
- `pnpm --filter prisma exec jest --silent --passWithNoTests --testTimeout=30000 src/__tests__/commands/Version.test.ts`

### DIFF
--- a/packages/cli/src/Version.ts
+++ b/packages/cli/src/Version.ts
@@ -22,6 +22,7 @@ import path from 'path'
 import { getInstalledPrismaClientVersion } from './utils/getClientVersion'
 
 const packageJson = require('../package.json')
+
 const prismaCliPath = path.dirname(require.resolve('../package.json'))
 
 /**
--- a/packages/cli/src/Version.ts
+++ b/packages/cli/src/Version.ts
@@ -16,8 +16,8 @@ import {
   wasm,
 } from '@prisma/internals'
 import { bold, dim, red } from 'kleur/colors'
-import path from 'path'
 import os from 'os'
+import path from 'path'
 
 import { getInstalledPrismaClientVersion } from './utils/getClientVersion'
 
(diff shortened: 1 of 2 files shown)


tensordreams:
Thanks — this is correct and complete: row, table snapshot, JSON snapshot, and sanitizers all check out, and the JSON key matches the real `slugify()` behavior.

One blocking one-liner: `import path from 'path'` is inserted after `kleur/colors` and before `import os from 'os'`; `simple-import-sort` requires `os` before `path`, so `pnpm lint` fails. Reorder the import (or run `pnpm lint --fix`) and this is in good shape.


---

kitwongpixel:
Thanks for the review. I pushed a follow-up commit that keeps `os` before `path` for simple-import-sort and also satisfies the newline-after-import lint rule around the package.json require.

Validated locally:
- `pnpm exec eslint packages/cli/src/Version.ts packages/cli/src/__tests__/commands/Version.test.ts`
- `pnpm --dir packages/cli exec jest --silent --passWithNoTests src/__tests__/commands/Version.test.ts --testTimeout=60000`

The default test command timed out at Jest's 5s default locally, but the same Version test file passes with a 60s timeout.

tensordreams:
Thank you for this, @kitwongpixel, and my apologies for the slow response — this sat in the queue far longer than it should have.

The same feature has since landed on `main` via #29573: `prisma version` / `prisma -v` now prints a `Prisma CLI Path` row resolved with `path.dirname(require.resolve('../package.json'))`, and the JSON output carries the same `prisma-cli-path` key. Your implementation matched it almost line for line, right down to the resolution strategy and the JSON key name — it simply lost the race, so I'm closing this as a duplicate.

Sorry for the wasted effort, and thanks for taking a swing at #7771.

PR #29614, opened 2026-05-30
Show Prisma CLI path in version output

oathis:
## Summary
- add the resolved Prisma CLI script path to `prisma -v` / `prisma version` output
- sanitize the path in the existing version command snapshot

## Tests
- `npx prettier --check packages/cli/src/Version.ts packages/cli/src/__tests__/commands/Version.test.ts`
- `git diff --check`

Note: I attempted `corepack pnpm --filter prisma test -- Version.test.ts`, but this local Windows checkout could not build/link workspace packages such as `@prisma/engines` after install. The changed snapshot covers the intended output shape.

Fixes #7771

@algora-pbc /claim #7771

### DIFF
--- a/packages/cli/src/Version.ts
+++ b/packages/cli/src/Version.ts
@@ -16,6 +16,7 @@ import {
   wasm,
 } from '@prisma/internals'
 import { bold, dim, red } from 'kleur/colors'
+import path from 'node:path'
 import os from 'os'
 
 import { getInstalledPrismaClientVersion } from './utils/getClientVersion'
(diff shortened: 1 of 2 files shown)


tensordreams:
Thanks for the focused change — the integration point (the `rows` array) and the sanitizer are right.

Blocking: the hand-written snapshot line is misaligned by one space. `formatTable` pads each label to the longest one (`Default Engines Hash`, 20 chars), so `Prisma CLI path` (15 chars) gets exactly 5 trailing spaces + ` : ` — 6 spaces before the colon; the snapshot line has 7. Since your sanitizer preserves the real padding, the snapshot can never match and the test fails. The fix is to actually run the test and let Jest write the snapshot rather than hand-editing it (the PR notes the local environment couldn't build — that's the missing step).

Also worth considering: `path.resolve(process.argv[1])` points at the bin shim; `fs.realpathSync` would surface the actual install location.


---

tensordreams:
Hi @oathis — thanks for this, and apologies for the slow reply.

Issue #29605, opened 2026-05-28
@prisma/dev pins vulnerable @hono/node-server@1.19.11 (GHSA-92pp-h63x-v22m) across all prisma 7.x — forces downstream npm overrides


oleksiidonetskyi:
## Bug description

Every published `@prisma/dev` version from `0.13.0` to the current `0.24.8` pins `@hono/node-server` to a release below `1.19.13`. All pinned versions are affected by [GHSA-92pp-h63x-v22m](https://github.com/advisories/GHSA-92pp-h63x-v22m) — middleware bypass via repeated slashes in `serveStatic`, fixed upstream in `@hono/node-server@1.19.13`.

Since `@prisma/dev` is a regular (non-optional) dependency of `prisma` starting in 7.x, every prisma 7.x project inherits the vulnerable transitive. The CLI also `require()`s `@prisma/dev/internal/state` at startup for non-`dev` subcommands (`generate`, `migrate deploy`, etc.), so the package cannot be removed or stubbed locally without breaking the whole CLI.

The only available downstream workaround is an `npm overrides` entry pinning `@hono/node-server: ^1.19.13`. For service templates consumed by many downstream projects, that override propagates and never goes away on its own.

## Severity

🔹 Minor: Unexpected behavior, but does not block development

## Reproduction

```bash
mkdir prisma-cve-repro && cd prisma-cve-repro
npm init -y
npm install prisma@7.8.0
npm audit
# → 3 moderate severity vulnerabilities
#   (all trace to @prisma/dev → @hono/node-server@1.19.11)
```

## Expected vs. Actual Behavior

**Expected:** `npm install prisma@latest` produces a clean `npm audit` (0 vulnerabilities), since the upstream advisory for GHSA-92pp-h63x-v22m has been fixed in `@hono/node-server@1.19.13`.

**Actual:** `npm install prisma@7.8.0` pulls `@prisma/dev@0.24.3` which pins `@hono/node-server@1.19.11`. `npm audit` reports 3 moderate vulnerabilities. The findings persist across every prisma 7.x version because no `@prisma/dev` release has bumped the hono pin.

## Frequency

Consistently reproducible

## Does this occur in development or production?

Only in development (e.g., CLI tools, migrations, Prisma Studio)

## Is this a regression?

[...]

PR #29615, opened 2026-06-01
fix(dev): bump @hono/node-server to ^1.19.13 to resolve GHSA-92pp-h63…

adityakrmishra:
**Description:**
Bumps `@hono/node-server` to `^1.19.13` in `@prisma/client` and `@prisma/query-plan-executor` to patch the middleware bypass vulnerability (GHSA-92pp-h63x-v22m). This removes the need for downstream consumers to use npm overrides in Prisma 7.x.

**Testing:**
- Monorepo dependencies were updated using `pnpm install` and the lockfile was verified to resolve the secure version.
- The dev build (`pnpm -r run dev`) completed successfully with zero TypeScript compilation errors.
- Internal test suites (`vitest` for the query-plan-executor) passed locally.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -167,7 +167,7 @@
     "@codspeed/benchmark.js-plugin": "4.0.0",
     "@faker-js/faker": "9.6.0",
     "@fast-check/jest": "2.0.3",
-    "@hono/node-server": "1.19.0",
+    "@hono/node-server": "^1.19.13",
     "@inquirer/prompts": "7.3.3",
     "@jest/create-cache-key-function": "29.7.0",
     "@jest/globals": "29.7.0",
--- a/packages/query-plan-executor/package.json
+++ b/packages/query-plan-executor/package.json
@@ -11,7 +11,7 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
-    "@hono/node-server": "1.19.0",
+    "@hono/node-server": "^1.19.13",
     "@hono/zod-validator": "0.7.2",
     "@opentelemetry/api": "1.9.0",
     "@opentelemetry/context-async-hooks": "2.1.0",


aqrln (commented):
Thanks. This does not close #29605 because it only updates the dependency in non user facing code and not in the package that issue is about but it's still a useful change. Could you please update the lockfile to make the PR mergeable?

tensordreams:
Thanks for the PR and for caring about the advisory! Unfortunately this is a no-op, so I'm going to close it.

`@hono/node-server` is already pinned to a patched version across the whole workspace via a `pnpm.overrides` entry in the root `package.json`:

```jsonc
"pnpm": {
  "overrides": {
    "hono": ">=4.12.14",
    "@hono/node-server": ">=1.19.13",
    ...
  }
}
```

Because of that override, `@hono/node-server` already resolves to `1.19.13` everywhere in `pnpm-lock.yaml` — it's the only version present, including for transitive consumers — so GHSA-92pp-h63x-v22m is already remediated on `main` and `pnpm audit` is clean.

This also explains why CI doesn't complain about an outdated lockfile: pnpm applies the override when recording the importer specifier, so the lockfile already stores `specifier: '>=1.19.13'` for both of these packages (not `1.19.0`). The override wins over the `package.json` range regardless of whether it says `1.19.0` or `^1.19.13`, so the effective specifier — and therefore the lockfile — is identical either way, and the frozen install in CI stays green. The net effect of this change on the installed dependency tree is zero.

One more note: these edits are in the `devDependencies` of `@prisma/client` and `@prisma/query-plan-executor`, which aren't shipped to consumers. The actual concern in #29605 is the *published* `@prisma/dev` declaring a vulnerable `@hono/node-server` as a regular dependency for downstream `prisma@7.x` users — that lives outside this repo and isn't affected by these changes.

Closing as a no-op, but thanks again for flagging the area. 🙏

PR #29619, opened 2026-06-03
ci(e2e): shard client e2e suite across 3 jobs to fix chronic timeouts

tensordreams:
## Problem

The `client-e2e` job chronically hit `timeout-minutes` and got cancelled — the ~90 Docker-based e2e tests (each a fresh `pnpm install`) don't finish within one job at `--maxWorkers 3` (it was already bumped 25 → 35 in `e8b5825fe` and still timed out).

## Part 1 — shard the suite

- `tests/e2e/_utils/run.ts`: add `--shard <index>/<count>` (sorted, round-robin → stable, disjoint, balanced; exact `^(\d+)\/(\d+)$` validation).
- `test-template.yml`: run `client-e2e` as a 3-way matrix (`1/3`, `2/3`, `3/3`).

Each shard finishes in ~16–22 min, well under timeout.

## Part 2 — fix the failures sharding exposed

Letting the suite finish surfaced failures that were pre-existing on main (the timeout was hiding them):

1. **Node version.** The 8 `_steps.cts` files import ESM-only `zx`; as CommonJS that's `require('zx')`, which throws `ERR_VM_MODULE_LINK_FAILURE` on Node 20. `custom-output-yarn` fails because `@prisma/streams-local` needs Node `>=22`. → Bump the e2e Docker image `node:20.19 → node:22`.
2. **Remove the dead `generator-build` shim.** `prisma-client-imports-sqlite` ENOENT'd on `@prisma/client/generator-build`'s sibling `prisma_schema_build_bg.wasm`. That entrypoint (`src/generation/generator.ts`) is a JSON-RPC adapter for the prisma-client-js generator that, per its own doc comment, *"is only used by Studio … will be removed"*. Prisma 7's Studio no longer uses it, it's never been user-facing, and it had silently been broken for a while — so rather than ship a 2.9 MB wasm to keep it alive, **remove it entirely**: the entrypoint, its esbuild config, the `./generator-build` export + `files` entry, the jest ignore, and the e2e import-assertions.
3. **Run the ESM `_steps` files as `.ts`.** The `_steps` files are ESM but were executed as `.cts`, so `import { $ } from 'zx'` (ESM-only) blew up. Run them as `.ts`, and since a named import of the CommonJS-transpiled `executeSteps` helper isn't statically resolvable under ESM, import the module default and destructure `executeSteps` from it. Exclude `_steps.ts` (alongside `_steps.cts`) from the fixture tsconfigs so the steps are run directly by the runtime rather than type-compiled.

## Part 3 — bugs the green suite caught / runner portability

1. **SQL commenter leaked user data (product bug).** For non-raw client-engine queries, `ClientEngine` built the `SqlCommenterQueryInfo` from the original query (single, batched, and compacted paths), so plugins such as `query-insights` received real user data values instead of `Param` placeholders. Pass the **parameterized** query payloads into the SQL commenter context so user data never reaches the SQL comments. The `query-insights` e2e assertions are hardened to check for the `Param` placeholder shape instead of exact JSON equality.
2. **Make the Docker runner work on SELinux-enabled Linux hosts.** Bind-mounted e2e files were unreadable inside the container on SELinux hosts, and `chmod +x`-then-exec of the mounted startup script could fail. Append the `:z` relabel option to bind volumes on Linux, invoke the startup script through `bash` instead of relying on the executable bit, and tag the built image (`prisma-client-test-e2e`) so it can be reused.

## ⚠️ Required status checks

Sharding renames the e2e check to `Client E2E … (shard 1/3 | 2/3 | 3/3)`. If branch protection requires the old `🧪 / Client E2E ubuntu-latest [v20.19.0]` by name, repoint it to the three shard checks on merge.

## Notes

- e2e no longer runs on the minimum supported Node 20.19; functional/unit suites still cover Node 20 (and `@prisma/streams-local`'s `>=22` engine, defined in another repo, makes a Node-20-green e2e unattainable from here anyway).
- A stale `// TODO … @prisma/client/generator-build …` comment remains in `internals/resolveBinary.ts` (a historical engine-path note, not code); left out of scope.


aqrln (approved):

packages/client/helpers/build.ts
```diff
@@ -198,13 +198,31 @@ function wasmEdgeRuntimeBuildConfig(format: ModuleFormat, name: string): BuildOp
...
+      // its bundled JS at runtime, so esbuild does not inline it. Copy it into
+      // the output directory so it is present in the published package (and not
+      // an ENOENT at runtime). `generator-build` is in `files` in package.json.
+      name: 'copy-schema-build-wasm',
+      setup(build) {
+        build.onEnd(() => {
+          const wasmFileName = 'prisma_schema_build_bg.wasm'
+          const schemaWasmDir = path.dirname(require.resolve('@prisma/prisma-schema-wasm/package.json'))
+          fs.copyFileSync(path.join(schemaWasmDir, 'src', wasmFileName), path.join(generatorBuildDir, wasmFileName))
+        })
```
aqrln: Why would the client need `@prisma/prisma-schema-wasm`? It shouldn't be necessary at run time in Client. Is that the legacy generator entrypoint we haven't removed? Or maybe the e2e test is doing something weird?

packages/client/tests/e2e/_utils/standard.dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM node:20.19
+# Node 22.12 is the minimum version where `require()` of an ES module is
+# enabled by default. The e2e `_steps.cts` files run as CommonJS and
+# `require('zx')` (ESM-only), which throws ERR_VM_MODULE_LINK_FAILURE on
+# Node 20. Node 22 also satisfies `@prisma/streams-local`'s `engines: >=22`.
+FROM node:22.12
```
aqrln: Use the latest minor of `node:22`

packages/client/helpers/build.ts
tensordreams: Good instinct — it's the legacy entrypoint, and the test is exercising it intentionally (not doing anything weird):

- `generator-build` is `src/generation/generator.ts`, whose own doc comment says *"A JSON-RPC adapter for the `prisma-client-js` generator. It is only used by Studio. It will be removed in the future."*
- `@prisma/client/generator-build` is a declared public export (in `package.json#exports`), and `prisma-client-imports-sqlite/src/default.ts` deliberately does `import * as G1 from '@prisma/client/generator-build'` to assert that entrypoint is importable.
- Importing it pulls `@prisma/client-generator-js` → `@prisma/internals` → `@prisma/prisma-schema-wasm`, which reads its `.wasm` from disk at module init. That `.wasm` only became a sibling file in a recent prisma-schema-wasm change (previously inlined), so the bundle started ENOENTing — which is why this was "broken for a while".

So the client doesn't need prisma-schema-wasm for normal runtime use — only this legacy shim does. Copying the wasm is the minimal correct fix while `generator-build` exists; the cleaner fix is deleting the shim (Studio-coupled, slated for removal), which I kept out of scope here. I've also wrapped the copy in a try/catch with a clear error.

packages/client/tests/e2e/_utils/standard.dockerfile
tensordreams: Done — switched to `node:22` (latest minor).

packages/client/helpers/build.ts
tensordreams: Agreed — removed it entirely in 6a00a4e rather than shipping the wasm. Confirmed nothing in the repo imports `@prisma/client/generator-build` outside these e2e assertions (Prisma 7 Studio doesn't use it; the entrypoint's own comment says it's Studio-only and slated for removal). Dropped the entrypoint, its esbuild config + the wasm-copy workaround, the `./generator-build` export and `files` entry, the jest ignore, and the three e2e import-assertions. Client build verified to no longer emit `generator-build/`.

(One stale `// TODO … generator-build …` comment remains in `internals/resolveBinary.ts` — a historical engine-path note, not code; left it out of scope, happy to clean it up if you'd like.)

aqrln (approved):

PR #29621, opened 2026-06-04
fix(adapter-mariadb): return Uint8Array for binary rows

### LINKED ISSUE #29597: @prisma/adapter-mariadb: DEP0005 `Buffer() is deprecated` is emitted on every query that returns BLOB/BINARY columns
## Summary

When using `@prisma/adapter-mariadb` with Prisma Client, every query whose result set contains a `BLOB` / `BINARY` / `VARBINARY` column emits the following warning from Node.js:

```
(node:1) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues.
Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

The warning is emitted **once per binary value** returned from the database, so a single `findMany()` can produce many lines of noise. Running Node.js with `--throw-deprecation` causes the process to throw.

## Root cause (as far as we could narrow it down)

`@prisma/adapter-mariadb` returns binary column values as Node.js `Buffer` instances (which is what the underlying `mariadb` driver returns). Somewhere inside Prisma Client / engine, the value is then passed to the legacy `Buffer()` constructor (rather than `Buffer.from()` / `Buffer.alloc()`, or accepting the `Uint8Array` view directly). That triggers `DEP0005` for each value.

We confirmed this on the application side by intercepting the driver adapter's `queryRaw` and converting every `Buffer` in `SqlResultSet.rows` to a `Uint8Array` view of the same `ArrayBuffer` (no copy) before returning it to Prisma core. With that wrapper in place the `DEP0005` warning disappears completely and query results remain correct. This strongly suggests the `new Buffer()` call happens on the Prisma side, after the driver adapter has returned its result set.

## Workaround for other users

Wrap the factory returned by `new PrismaMariaDb(...)` and convert `Buffer` values in `SqlResultSet.rows` to `Uint8Array` views before they reach Prisma core. Both `adapter.queryRaw` and `transaction.queryRaw` need to be intercepted (`executeRaw` is unaffected because it does not return rows). Reference implementation:

https://github.com/88labs/andpad-greenfile/pull/1382/files

Key snippet:

```ts
const normalizeResultSet = (result: SqlResultSet): SqlResultSet => ({
  ...result,
  rows: result.rows.map((row) =>
    row.map((value) =>
      Buffer.isBuffer(value)
        ? new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
        : value,
    ),
  ),
});
```

## Suggested fix

Either:

1. Stop calling the deprecated `Buffer()` constructor on the Prisma side. Use `Buffer.from(...)` / accept `Uint8Array` directly, or
2. Have `@prisma/adapter-mariadb` return `Uint8Array` instead of `Node.js Buffer` for binary columns, since the driver-adapter contract is supposed to be a `Uint8Array`-friendly boundary.

### Severity

### Steps to reproduce

1. Create a table with a binary column in MariaDB / MySQL:
   ```sql
   CREATE TABLE bin_test (id INT PRIMARY KEY, data VARBINARY(16));
   INSERT INTO bin_test VALUES (1, 0x0102030405);
   ```
2. Define the matching Prisma model with `Bytes`:
   ```prisma
   model BinTest {
     id   Int   @id
     data Bytes
     @@map("bin_test")
   }
   ```
3. Configure Prisma Client with the MariaDB driver adapter:
   ```ts
   import { PrismaClient } from '@prisma/client';
   import { PrismaMariaDb } from '@prisma/adapter-mariadb';

   const adapter = new PrismaMariaDb({ /* host, user, password, database, ... */ });

   await prisma.binTest.findMany();
   ```
4. Run Node.js normally (or with `--trace-deprecation` to see the stack) and observe:
   ```
   (node:1) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues.
   Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
   ```
   The warning is emitted for every binary value in the result set. With `--throw-deprecation`, the process throws.

### Note
The issue does **not** happen with `executeRaw` (no rows returned). It only happens on code paths that return rows containing `Buffer` values (i.e. `queryRaw` / `findMany` / `findUnique` / etc. on models with `Bytes` columns).

**Expected:** No `DEP0005` warning. Prisma should not invoke the deprecated `Buffer()` constructor on values coming from a driver adapter. Either it should call `Buffer.from(...)` / accept `Uint8Array` directly, or the driver adapter should hand `Uint8Array` over to core.

**Actual:** A `[DEP0005] DeprecationWarning: Buffer() is deprecated` line is printed to stderr for every binary value returned by `@prisma/adapter-mariadb`. On a process started with `--throw-deprecation` this becomes a fatal error. In log aggregators (e.g. Datadog) this floods error/warning channels in production.

### Frequency

Consistently reproducible

### Is this a regression?

Unknown — this is the first time we have used `@prisma/adapter-mariadb` (driver adapters preview), so we cannot confirm whether earlier versions were free of this warning. We are reporting against the latest published versions (7.8.0).

### Workaround

Yes. Wrap the factory returned by `new PrismaMariaDb(...)` and convert every `Buffer` in `SqlResultSet.rows` to a `Uint8Array` view (`new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength)` — no copy) before returning the result set to Prisma core. Both `adapter.queryRaw` and `transaction.queryRaw` need to be intercepted. `executeRaw` is unaffected.

Reference implementation (TypeScript, with proxies that keep `this` bound on the wrapped adapter / transaction): https://github.com/88labs/andpad-greenfile/pull/1382/files

This is a user-land workaround and should not be necessary; it is included here so that other users hitting `DEP0005` can unblock themselves while the upstream fix is being considered.

### Prisma Schema & Queries

```prisma
// schema.prisma (minimal)
datasource db {
  provider = "mysql"
}

generator client {
  provider = "prisma-client-js"
}

model BinTest {
  id   Int   @id
  data Bytes
  @@map("bin_test")
}
```

```ts
// query that triggers the DEP0005 warning
import { PrismaClient } from '@prisma/client';
import { PrismaMariaDb } from '@prisma/adapter-mariadb';

const adapter = new PrismaMariaDb({
  host: 'localhost',
  port: 3306,
  user: 'app',
  password: '***',
  database: 'app',
});

const rows = await prisma.binTest.findMany();
// → [DEP0005] DeprecationWarning: Buffer() is deprecated ...
```

### Prisma Config

```ts
export default defineConfig({
  migrations: {
    path: 'prisma/migrations',
    seed: 'npm run seed',
  },
  ...(databaseUrl
    ? {
        datasource: {
          url: databaseUrl,
        },
      }
    : {}),
});

```

### Environment & Setup

- OS: Linux (containerized runtime)
- Node.js: 24.x (the project pins `engines.node` to `>=24.0.0` and uses `@types/node@^24.12.4`)
- Database: MariaDB, accessed via `@prisma/adapter-mariadb` (driver-adapters preview), not the native Rust engine.
- Framework: Next.js application running in production (warning is observed in Datadog logs).

The warning is reproducible on any Node.js version that still emits `DEP0005` for `new Buffer()`.

### Prisma Version

```
prisma: 7.8.0
@prisma/client: 7.8.0
@prisma/adapter-mariadb: 7.8.0
@prisma/instrumentation: 7.8.0
@prisma/extension-read-replicas: 0.5.0
```

puneetdixit200:
Fixes #29597.

MariaDB returns binary column values as Node Buffers, but the driver-adapter result contract uses Uint8Array for binary values. This normalizes Buffers in mapRow before they reach Prisma Client, preserving the original byte offset and length for sliced buffers.

Verification:
- pnpm --config.store-dir=/tmp/prisma-29597-pnpm-store --filter @prisma/adapter-mariadb test conversion.test.ts
- pnpm --config.store-dir=/tmp/prisma-29597-pnpm-store --filter @prisma/adapter-mariadb test
- pnpm --config.store-dir=/tmp/prisma-29597-pnpm-store --filter @prisma/adapter-mariadb build
- git diff --check

### DIFF
--- a/packages/adapter-mariadb/src/conversion.test.ts
+++ b/packages/adapter-mariadb/src/conversion.test.ts
@@ -26,10 +26,14 @@ describe('mapRow', () => {
     const row = { data: value }
 
     const result = mapRow(row, [{ type: 'BLOB' } as unknown as mariadb.FieldInfo])
+    const mapped = result[0] as Uint8Array
 
-    expect(Buffer.isBuffer(result[0])).toBe(false)
-    expect(result[0]).toBeInstanceOf(Uint8Array)
-    expect(result[0]).toEqual(Uint8Array.from([1, 2, 3]))
+    expect(Buffer.isBuffer(mapped)).toBe(false)
+    expect(mapped).toBeInstanceOf(Uint8Array)
+    expect(mapped.buffer).toBe(value.buffer)
+    expect(mapped.byteOffset).toBe(value.byteOffset)
+    expect(mapped.byteLength).toBe(value.byteLength)
+    expect(mapped).toEqual(Uint8Array.from([1, 2, 3]))
   })
 
   it('preserves property order for mixed object rows', () => {
--- a/packages/adapter-mariadb/src/conversion.test.ts
+++ b/packages/adapter-mariadb/src/conversion.test.ts
@@ -20,6 +20,18 @@ describe('mapRow', () => {
     expect(result).toEqual(['123'])
   })
 
+  it('maps buffer columns to Uint8Array views', () => {
+    const source = Buffer.from([0, 1, 2, 3, 4])
+    const value = source.subarray(1, 4)
+    const row = { data: value }
+
+    const result = mapRow(row, [{ type: 'BLOB' } as unknown as mariadb.FieldInfo])
+
+    expect(Buffer.isBuffer(result[0])).toBe(false)
+    expect(result[0]).toBeInstanceOf(Uint8Array)
+    expect(result[0]).toEqual(Uint8Array.from([1, 2, 3]))
+  })
+
   it('preserves property order for mixed object rows', () => {
     const row = {
       createdAt: '2024-01-02 03:04:05.123',
--- a/packages/adapter-mariadb/src/conversion.ts
+++ b/packages/adapter-mariadb/src/conversion.ts
@@ -166,6 +166,10 @@ export function mapRow<A>(row: A[] | Record<string, A>, fields?: mariadb.FieldIn
       return value.toString()
     }
 
+    if (Buffer.isBuffer(value)) {
+      return new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
+    }
+
     return value
   })
 }


packages/adapter-mariadb/src/conversion.test.ts
```diff
@@ -20,6 +20,22 @@ describe('mapRow', () => {
     expect(result).toEqual(['123'])
   })
 
+  it('maps buffer columns to Uint8Array views', () => {
+    const source = Buffer.from([0, 1, 2, 3, 4])
+    const value = source.subarray(1, 4)
+    const row = { data: value }
+
+    const result = mapRow(row, [{ type: 'BLOB' } as unknown as mariadb.FieldInfo])
```
aqrln, line 28: why the unsafe double cast?

aqrln (commented):
I'm not against the change but I'm also not convinced. What's the motivation? `Buffer` is a subclass of `Uint8Array` so the contract is still upheld one way or the other, this is not strictly necessary.

aqrln:
Also @puneetdixit200 please sign the CLA

aqrln:
I now noticed that you marked the PR as fixing #29597. How does it fix it? Have you verified that it actually stops printing the warning?

The diff does not remove any deprecated `Buffer` constructor invocations and therefore doesn't address the root cause. If the issue is upstream in the `mariadb` driver, then this change does nothing to prevent it. If the issue is on our side, then at worst this also does nothing and at best it only incidentally works around the issue by skipping the affected code path, while the correct fix would have been to replace that constructor invocation with a non-deprecated API.

daltino (commented):
Nice, focused fix — the implementation is clean and the intent is clear. A few things worth discussing before merging.

---

## Overall

The core idea is sound: `Buffer` is a subclass of `Uint8Array`, so `instanceof Uint8Array` checks would actually pass already, but the problem is that `Buffer.isBuffer()` would also return `true`, and some downstream code (or serialization) may behave differently for `Buffer` vs plain `Uint8Array`. Stripping the `Buffer` identity via a view over the same `ArrayBuffer` is the right move.

---

## Issues & Suggestions

### 1. 🟡 Sliced buffer — `byteOffset` is preserved but is that what callers expect?

```ts
return new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
```

This is correct for a *view*, but worth being explicit about in the PR description (you already mentioned it — good). The subtle risk: if anything downstream calls `.buffer` and iterates the whole `ArrayBuffer`, it will see bytes outside the intended slice. This is standard `TypedArray` semantics, but it's a potential footgun if Prisma's serialization layer ever does `uint8Array.buffer` without respecting `byteOffset`/`byteLength`.

**Suggestion:** Consider adding a comment here explaining the intent:

```ts
// Preserve the slice's offset/length rather than copying, so this is O(1).
// Callers must respect byteOffset + byteLength (standard TypedArray contract).
return new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
```

If there's *any* uncertainty about whether Prisma's internals honour `byteOffset`, the safer (but O(n)) alternative is:

```ts
return new Uint8Array(value.buffer.slice(value.byteOffset, value.byteOffset + value.byteLength))
// or simply:
return new Uint8Array(value)  // copies, always offset-0
```

It's worth a quick audit of how `Uint8Array` values are consumed downstream before settling on the zero-copy path.

---

### 2. 🟡 Test doesn't assert byte *values*

The test verifies the structural identity of the view (same `buffer`, correct `byteOffset`) but never checks that the actual bytes are correct:

```ts
expect(mapped[0]).toBe(1)
expect(mapped[1]).toBe(2)
expect(mapped[2]).toBe(3)
expect(mapped.length).toBe(3)
```

This would catch a regression where `byteOffset` or `byteLength` is passed incorrectly (e.g. if someone accidentally omitted `byteOffset`). Strongly recommend adding it.

---

### 3. 🟢 Missing test: non-sliced / whole-buffer case

The test only covers `subarray()` (a sliced buffer). It's worth adding a simple case for a non-sliced `Buffer`:

```ts
it('maps a plain Buffer to Uint8Array', () => {
  const value = Buffer.from([10, 20, 30])
  const result = mapRow({ data: value }, [{ type: 'BLOB' } as unknown as mariadb.FieldInfo])
  const mapped = result[0] as Uint8Array
  expect(Buffer.isBuffer(mapped)).toBe(false)
  expect(mapped).toBeInstanceOf(Uint8Array)
  expect(Array.from(mapped)).toEqual([10, 20, 30])
})
```

---

### 4. 🟢 `FieldInfo` cast — minor

```ts
[{ type: 'BLOB' } as unknown as mariadb.FieldInfo]
```

The `Buffer.isBuffer` check in `conversion.ts` doesn't actually look at the field type at all — the `BLOB` type hint is unused by the new branch. The test would pass identically with `{ type: 'VARCHAR' }`. Not a bug, just slightly misleading. A small comment like `// field type doesn't affect Buffer detection` would clarify intent, or you could drop the field descriptor entirely and pass `undefined` (if the function allows it).

---

### 5. 🟢 Consider placement of the `Buffer.isBuffer` check

Currently the check sits *after* the `bigint` and string-coercion guards. Given that `Buffer` is a `Uint8Array` subclass, it can't be a `bigint`, so ordering is fine. But it sits *before* the final `return value` fallthrough. Just making sure there's no earlier branch that could intercept a `Buffer` — looks like there isn't, so this is fine as-is.

---

## Summary

| | |
|---|---|
| Core fix | ✅ Correct |
| Zero-copy view | ⚠️ Verify downstream consumers honour `byteOffset` |
| Test: structural | ✅ Good |
| Test: byte values |

tensordreams:
Hi @puneetdixit200 — thanks for this, and apologies that it sat for so long.

We are closing this one as inactive: the CLA has not been signed, and the question raised in review — whether this actually fixes #29597, given the diff removes no deprecated `Buffer` constructor — has gone unanswered for several weeks.

PR #29623, opened 2026-06-05
fix(cli): skip URL validation for sqlserver protocol in Studio

### LINKED ISSUE #29620: Prisma Studio fails on MSSQL database URLs
### Bug description

Starting Prisma Studio using an MSSQL connection URL in the format
```
sqlserver://host:1433;user={MyServer/User};password={Pass:Word;};database=db
```
as described in the [Prisma docs](https://www.prisma.io/docs/orm/core-concepts/supported-databases/sql-server#connection-details) fails with the error:
```
Loaded Prisma config from prisma.config.ts.


! The provided database URL is not valid.
```

The problem is due to *all* URLs being passed through `URL.canParse()`: https://github.com/prisma/prisma/blob/3937779e617c743966ebb50901d013167dd9171e/packages/cli/src/Studio.ts#L281-L283

That works for Postgres & MySQL connection URLs with these formats (all taken from different driver adapter pages in the docs):
```
postgresql://user:pass@host:5432/database
postgres://user:pass@host-pooler.region.aws.neon.tech:5432/db
postgresql://user:pass@host:5432/db?sslmode=require&sslcert=./cert.pem
mysql://user:pass@localhost/db?socket=/var/run/mysqld/mysqld.sock
```
but it doesn't work specifically with MSSQL URLs with this format (note the non-standard semicolons and braces):
```
sqlserver://HOST:PORT;database=DATABASE;user=USER;password=PASSWORD;encrypt=true
sqlserver://host:1433;user={MyServer/User};password={Pass:Word;};database=db
```

To quote the [docs](https://www.prisma.io/docs/orm/core-concepts/supported-databases/sql-server#connection-details):
> SQL Server uses **JDBC-style connection strings**:
> `sqlserver://HOST[:PORT];database=DATABASE;user=USER;password=PASSWORD;encrypt=true`

Trying it out in the Node CLI:
```
> URL.canParse("postgresql://user:pass@host:5432/db?sslmode=require&sslcert=./cert.pem")
true
> URL.canParse("sqlserver://HOST:PORT;database=DATABASE;user=USER;password=PASSWORD;encrypt=true")
false
```

### Severity

### Reproduction

1. Setup the following files for a basic Prisma project
    - `.env` with `DATABASE_URL` set to an MSSQL connection URL
    - `schema.prisma`
    - `prisma.config.ts` that uses the `.env`
2. Run `prisma studio`

## Expected behavior
The Prisma Studio server starts with a web URL.

## Actual behvaior
Prisma Studio doesn't start.

### Frequency

Consistently reproducible

### Is this a regression?

It worked before the Studio rewrite in Prisma 7.0.0.

### Workaround

Using a Prisma version < 7.0.0 and a pre-v7 schema with the MSSQL connection URL set in the `datasource` block works.

### Prisma Schema & Queries
```
generator client {
  previewFeatures     = ["nativeDistinct"]
  binaryTargets       = ["native"]
  output              = "./generated"
  importFileExtension = "js"
}

datasource db {
  provider = "sqlserver"
  schemas  = [..., ..., ...]
}

<misc. models>
```

### Prisma Config
```

export default defineConfig({
  schema: "schema.prisma",
  migrations: {
    path: "migrations",
    seed: "tsx seed.ts",
  },
  datasource: {
    url: process.env.DATABASE_URL!, // env is inserted using dotenvx
  },
});
```

### Logs & Debug Info

```
<misc. corepack lines>

prisma:config:defineConfig [default]: '{\n  "loadedFromFile": null\n}' +270ms
prisma:config:defineConfig [config.schema]: 'schema.prisma' +1ms
prisma:config:defineConfig [config.datasource]: '{\n' +
  '  "url": "sqlserver://<host>:<port>;database=<database>;user=<user>;password=<password>;encrypt=<bool>;trustServerCertificate=<bool>"\n' +
  '}' +0ms
prisma:config:defineConfig [config.migrations]: '{\n  "path": "migrations",\n  "seed": "tsx seed.ts"\n}' +0ms
prisma:config:loadConfigFromFile Config file loaded in 25.86ms +0ms
prisma:config:PrismaConfig Parsing PrismaConfigInternal: '{\n' +
  '  "loadedFromFile": null,\n' +
  '  "schema": "schema.prisma",\n' +
  '  "datasource": {\n' +
  '    "url": "sqlserver://<host>:<port>;database=<database>;user=<user>;password=<password>;encrypt=<bool>;trustServerCertificate=<bool>"\n' +
  '  },\n' +
  '  "migrations": {\n' +
  '    "path": "migrations",\n' +
  '    "seed": "tsx seed.ts"\n' +
  '  }\n' +
  '}' +1ms
prisma:config:PrismaConfig Short-circuit: input is already a PrismaConfigInternal object +0ms
Loaded Prisma config from prisma.config.ts.

prisma:getSchema Reading schema from single file <path>/schema.prisma +366ms
prisma:getSchema Reading schema from single file <path>/schema.prisma +0ms
prisma:getConfig Using getConfig Wasm +0ms
prisma:getConfig config data retrieved without errors in getConfig Wasm +8ms
prisma:cli:checkpoint runCheckpointClientCheck(): Execution time for getting info: 9.139040999999963 ms +0ms
prisma:cli:checkpoint runCheckpointClientCheck(): Execution time for "await checkpoint.check(data)": 5.104540999999983 ms +5ms
prisma:cli:bin Execution time for executing "await cli.parse(commandArray)": 14.743499999999983 ms +0ms


! The provided database URL is not valid.
```

### Environment & Setup

- **OS:** macOS
- **Database:** MSSQL
- **Node.js version:** v26.0.0

### Prisma Version

```
prisma               : 7.8.0
Operating System     : darwin
Architecture         : arm64
Node.js              : v26.0.0
TypeScript           : unknown
PSL                  : @prisma/prisma-schema-wasm 7.8.0-6.3c6e192761c0362d496ed980de936e2f3cebcd3a
Schema Engine        : schema-engine-cli 3c6e192761c0362d496ed980de936e2f3cebcd3a (at <schema engine path>)
Default Engines Hash : 3c6e192761c0362d496ed980de936e2f3cebcd3a
Studio               : 0.27.3
Preview Features     : nativeDistinct
```

sijie-Z:
Fixes #29620

I traced the Prisma Studio URL validation issue on the current `main` branch:

In `packages/cli/src/Studio.ts` (Line 281), `URL.canParse(connectionString)` rejects MSSQL URLs because they use semicolons (`;`) as parameter delimiters instead of the standard URL query parameters.

The fix extracts the protocol before performing the URL check and skips validation for `sqlserver://`. This ensures that users see the correct "not supported" message instead of a misleading "not valid" error.


daltino (commented):
## Review: `fix(cli): skip URL validation for sqlserver protocol in Studio`

Nice fix for a real pain point — MSSQL connection strings like `sqlserver://server;database=mydb;user=sa;password=secret` are genuinely not valid WHATWG URLs, so skipping `URL.canParse` for them makes sense. The linked issue and inline comment are appreciated. Here are some suggestions to tighten it up before merging.

---

### 🐛 Correctness / Edge Cases

**1. Protocol extraction is duplicated and brittle**

The patch shows the `protocol` variable is extracted early (before validation), but the original code also had `const protocol = new URL(connec...` (truncated in the diff). This means there are now likely **two** `protocol` declarations in the function, which would be a compile/runtime error. Please verify the full diff removes or consolidates the second one.

**2. `split('://')` edge cases**

```ts
const protocol = connectionString.split('://')[0]?.toLowerCase() ?? ''
```

- If `connectionString` is an empty string or contains no `://`, `split` still returns a one-element array, so `[0]` is `''` — that's fine.
- If `connectionString` is something malformed like `://foo`, `[0]` is `''` — also fine, falls through to `URL.canParse`.
- However, if the input is `null` or `undefined` (possible if the env var is unset), this will throw before the existing guard. Double-check that `connectionString` is guaranteed to be a string at this point.

**3. Consider `sqlserver` variants**

The MSSQL connector in Prisma also accepts `jdbc:sqlserver://...` in some contexts. It's worth confirming whether that variant can ever reach this code path, or adding a note that it's intentionally excluded.

---

### 🧪 Test Coverage

There don't appear to be any new tests added. At minimum, consider adding a unit/integration test in the Studio test suite for:

- A valid `sqlserver://` string → should not return `UserFacingError`
- An invalid non-sqlserver URL → should still return `UserFacingError`
- A valid `postgresql://` URL → unchanged behaviour

Even a simple snapshot or assertion test would prevent regression here.

---

### 💅 Code Style

**4. Minor: `?? ''` is redundant**

```ts
const protocol = connectionString.split('://')[0]?.toLowerCase() ?? ''
```

`String.prototype.split` always returns a non-empty array, so `[0]` is never `undefined`. The optional chaining (`?.`) and nullish coalescing (`?? ''`) are unnecessary noise. This works fine:

```ts
const protocol = connectionString.split('://')[0].toLowerCase()
```

**5. Consider a named constant or helper**

If other protocols are ever added to this bypass list (e.g., some future connector), a small set or array would be cleaner and more readable:

```ts
const PROTOCOLS_SKIPPING_URL_VALIDATION = new Set(['sqlserver'])
const protocol = connectionString.split('://')[0].toLowerCase()

if (!PROTOCOLS_SKIPPING_URL_VALIDATION.has(protocol) && !URL.canParse(connectionString)) {
  return new UserFacingError('The provided database URL is not valid.')
}
```

Not required for this PR, but worth a thought.

---

### ⚡ Performance

No concerns — this is a one-time startup check, not a hot path.

---

### Summary

| Area | Status |
|---|---|
| Core logic | ✅ Correct approach |
| Duplicate `protocol` var | ⚠️ Needs verification |
| Test coverage | ❌ Missing |
| Code style | 🟡 Minor cleanup suggested |

The fix direction is right. The main asks before merging are: (1) confirm there's no duplicate `protocol` variable in the final file, and (2) add at least one test case for the `sqlserver://` path. Great work tracking this down!

sijie-Z:
Thanks for the review @daltino!

I've pushed a follow-up commit addressing the feedback:

* Added test coverage for the sqlserver path and regression cases.
* Cleaned up the protocol extraction logic.
* Refactored the bypass list into a module-level Set.
* Confirmed there is no duplicate protocol declaration in the final implementation.

Thanks again for the suggestions!

aqrln (approved):

aqrln (changes requested):
Tests are failing on CI

tensordreams:
Pushed a fix for the failing `CLI commands` and `Lint` jobs.

The two Studio URL-validation tests failed because `UserFacingError` wraps its message with a leading newline, a bold red `!` prefix and ANSI color codes (see `packages/cli/src/utils/errors.ts`), so the exact `.toBe('Prisma Studio is not supported for the "sqlserver" protocol.')` / `.toBe('The provided database URL is not valid.')` assertions never matched the actual `message`. They now assert on `error.name === 'UserFacingError'` plus a substring match via `.toContain(...)`. Collapsing the multi-line `expect()` and the mysql arg array to single lines also clears the Prettier failure.

Verified locally (targeted, no full build):
- `vitest run src/__tests__/Studio.vitest.ts` -> 18 passed
- `vitest run src/__tests__/studio-server.vitest.ts` -> 4 passed
- `prettier --check` and `eslint` on both changed files -> clean

Also merged the latest `main` into the branch. The functional change remains scoped to `Studio.ts` (skip WHATWG URL validation for the `sqlserver` protocol) plus its test.

sijie-Z:
Thanks for the reviews! Here's what was addressed:

@daltino's review:
- ✅ Duplicate `protocol` variable — already resolved in the latest code (single declaration at L289)
- ✅ `split('://')` edge cases — safe: `connectionString` is guarded by the truthiness check at L283
- ✅ `?? ''` redundancy — removed, using `.toLowerCase()` directly
- ✅ Named constant — `PROTOCOLS_SKIPPING_URL_VALIDATION` (Set) already added
- ✅ Test coverage — three tests added in "Studio URL validation" describe block
- ℹ️ `jdbc:sqlserver://` variant — not applicable here. The protocol matched against `CONNECTION_STRING_PROTOCOL_TO_STUDIO_STUFF` is the URL scheme (before `://`), and `jdbc` would already return "not supported" from the protocol map lookup, which is correct behavior.

**@aqrln's "Tests are failing on CI":**
- ✅ All 18 tests pass locally with the latest commit (`a921a74`).

**@coderabbitai's suggestion (line 129-130):**
- ✅ Added assertions to verify the URL components (protocol, hostname, pathname) that `createPoolMock` received, confirming the MySQL URL is forwarded correctly rather than just checking call count.
- Used URL component assertions to avoid brittleness from incidental URL normalization by `new URL()`.

All tests pass: `18 passed | 0 failed`.

aqrln (approved):

PR #29628, opened 2026-06-08
fix(p2002): correct modelName in nested create unique constraint errors #29595

### LINKED ISSUE #29595: P2002 error reports incorrect `modelName` for nested create operations
## Bug description

When a nested `create` operation causes a unique constraint violation (P2002), the `modelName` in `error.meta` incorrectly reports the **top-level model** (the one `.create()` / `.update()` was called on) instead of the model where the constraint violation actually occurred.

## How to reproduce

Given the following schema:

```prisma
model AppMajorVersion {
  id       String       @id @default(cuid())
  appId    String
  number   Int
  versions AppVersion[]

  @@unique([appId, number])
}

model AppVersion {
  id             String          @id @default(cuid())
  majorVersionId String
  number         Int
  majorVersion   AppMajorVersion @relation(fields: [majorVersionId], references: [id])

  @@unique([majorVersionId, number])
}
```

When executing a nested create:

```ts
await prisma.appVersion.create({
  data: {
    majorVersion: {
      create: {
        appId: "some-app-id",
        number: 1, // duplicate — violates AppMajorVersion@@unique([appId, number])
      },
    },
    number: 1,
  },
});
```

If the nested `create` on `AppMajorVersion` violates `@@unique([appId, number])`, the resulting `PrismaClientKnownRequestError` has:

```json
{
  "code": "P2002",
  "meta": {
    "modelName": "AppVersion",  // ❌ WRONG — should be "AppMajorVersion"
    "target": ["appId", "number"]
  }
}
```

The `target` fields are correct (they reflect the actual database constraint), but `modelName` points to the top-level model (`AppVersion`) rather than the nested model (`AppMajorVersion`) where the violation occurred.

## Expected behavior

`error.meta.modelName` should be `"AppMajorVersion"` — the model that owns the violated constraint.

## Prisma information

**Environment:** Prisma 6.x with driver adapters (`@prisma/adapter-pg`)

## Root cause analysis

The issue is in `packages/client/src/runtime/RequestHandler.ts` in `handleRequestError()`:

```ts
const meta = modelName ? { modelName, ...error.meta } : error.meta
```

The `modelName` is always the top-level model from the client operation (passed via `params.modelName` from `getPrismaClient.ts`). It does not account for which model in a nested write actually triggered the constraint violation.

The engine-level error (`UniqueKeyViolation` in `libs/user-facing-errors/src/query_engine/mod.rs`) only carries a `DatabaseConstraint` (the fields), not a model name. The client injects `modelName` based solely on which model the operation was called on, regardless of nesting depth.

This was introduced in #20724 which added `modelName` to error meta but only handles top-level operations correctly.

## Environment & setup

- OS: Linux
- Database: PostgreSQL (via driver adapter)
- Node.js version: 22.x
- Prisma version: 6.x

AnupamKumar-1:
Fixes #29595

**Problem**

When a nested `create` caused a unique constraint violation (P2002), `error.meta.modelName` always reported the top-level model (e.g. `AppVersion`) instead of the model where the violation actually occurred (e.g. `AppMajorVersion`).

**Root cause**

The adapters were stripping the table name when parsing DB errors. By the time the error reached `RequestHandler.ts`, there was no model info left, so it just stamped the top-level operation's model name onto every P2002.

**Fix**

- Added `table?` to `UniqueConstraintViolation` in `driver-adapter-utils`
- Extract table name in `better-sqlite3` and `libsql` from the SQLite error message
- Use native `error.table` in `pg` and `neon` (already provided by the driver)
- Set `meta.modelName` from `table` in `user-facing-error.ts` before `RequestHandler` can overwrite it
- `RequestHandler.ts` now prefers `error.meta.modelName` if already set

**Test**

Added functional regression test in `packages/client/tests/functional/issues/29595-p2002-nested-create-modelname/`

---


daltino (changes requested):
## Review: `fix(p2002): correct modelName in nested create unique constraint errors`

Great direction on this fix — the root cause analysis is correct, and the approach of plumbing the table name through from the adapter layer up to the user-facing error is the right architectural move. However, there are several issues that need addressing before this is mergeable.

---

### 🔴 Blockers

#### 1. `table` is extracted but never returned (SQLite adapters)

In both `adapter-better-sqlite3/src/errors.ts` and `adapter-libsql/src/errors.ts`, you extract `table` but never include it in the returned object:

```ts
const table = afterColon?.split(', ').at(0)?.split('.').at(0)
return {
  // `table` is never included here!
  kind: 'UniqueConstraintViolation',
  ...
}
```

The fix needs to actually pass `table` through, e.g.:

```ts
return {
  kind: 'UniqueConstraintViolation',
  constraint: fields !== undefined ? { fields } : undefined,
  table,
}
```

This is a complete omission that means the SQLite fix doesn't actually work at all.

---

#### 2. `table` isn't part of the `MappedError` / `DriverAdapterErrorObject` type

The `table` property is being added to returned objects, but unless the union type for `UniqueConstraintViolation` errors already includes `table?: string`, TypeScript may either silently drop it (excess property checking), give a type error, or — depending on how the type is structured — not propagate it correctly. The patch doesn't show any type definition updates. Please confirm and show the type changes, or add them.

---

#### 3. Raw table name vs. Prisma model name — naming mismatch may confuse users

In `user-facing-error.ts`:

```ts
if (error.cause.kind === 'UniqueConstraintViolation' && error.cause.table) {
  meta.modelName = error.cause.table
}
```

The database table name (e.g. `app_major_versions`) is being set as `modelName`. But Prisma users expect `modelName` to be the Prisma model name (e.g. `AppMajorVersion`), not the raw SQL table name. For models with custom `@@map` names, this will give the wrong value and could confuse users or break downstream tooling that depends on `modelName`.

This is actually the crux of the issue. The existing code path in `RequestHandler.ts` presumably does a table→model name mapping. You should either:
- Do the same mapping here, or
- Pass the raw table name under a different key (e.g. `table`) and let `RequestHandler.ts` continue to handle the model name resolution

---

### 🟡 Concerns / Suggestions

#### 4. SQLite table extraction is fragile for multi-column unique constraints

```ts
const table = afterColon?.split(', ').at(0)?.split('.').at(0)
```

For a multi-column constraint like `UNIQUE constraint failed: foo.a, bar.b`, this assumes all fields belong to the same table by only looking at the first field. That's probably fine in practice (SQLite constraints span one table), but it's worth a comment explaining the assumption. Also, what if the column name itself contains a dot? Unlikely but defensive coding helps.

#### 5. `error.table` availability on `DatabaseError` for pg/neon

For `adapter-pg` and `adapter-neon`, `error.table` is the table name from the PostgreSQL error detail. This should be reliable for single-table violations, but it's worth confirming this is always populated for `23505` (unique violation) errors and documenting the assumption.

#### 6. No tests added

There are no test cases in this PR. Given the specificity of this bug (nested create, unique constraint), this is a high-value regression test target. Please add:
- A test for nested `create` causing P2002 that asserts `error.meta.modelName` matches the *nested* model's name (not the top-level model)
- Tests for both SQLite and PostgreSQL paths
- A test with a `@@map` model to ensure the table→model mapping works correctly (see blocker #3)

---

### ✅ What's good

- The diagnosis of where the table name was being lost is correct.
- The approach of extracting `table` at the adapter layer and threading it up is clean.
- Keeping the `fields` extraction logic intact (just refactored to share the `afterColon` intermediate) is a nice cleanup.
- The pg/neon changes are minimal and correct in structure (just need the type fix).

---

### Summary

The core idea is right, but there are two critical bugs: the SQLite `table` is extracted but never returned, and there's no type update. There's also a semantic concern about raw table names vs. Prisma model names that could cause incorrect behavior for models with `

packages/client-engine-runtime/src/user-facing-error.ts
```diff
@@ -34,7 +34,11 @@ export function rethrowAsUserFacing(error: any): never {
   const code = getErrorCode(error)
   const message = renderErrorMessage(error)
   if (code !== undefined && message !== undefined) {
-    throw new UserFacingError(message, code, { driverAdapterError: error })
+    const meta: Record<string, unknown> = { driverAdapterError: error }
+    if (error.cause.kind === 'UniqueConstraintViolation' && error.cause.table) {
+      meta.modelName = error.cause.table
```
aqrln: What if model name and table name are different?

AnupamKumar-1: I changed the implementation so user-facing-error.ts no longer sets modelName from the raw table name. Instead, it forwards the table as meta.table.

The table → Prisma model resolution now happens in RequestHandler, where we have access to the runtime data model. This allows mapped models (@@map) to be resolved back to their Prisma model names before exposing error.meta.modelName.

I also added a regression test using a mapped model (AppMajorVersion → app_major_versions) to cover this case. The test verifies that error.meta.modelName is AppMajorVersion rather than the mapped table name.

AnupamKumar-1:
Hi @aqrln , I’ve addressed the latest CI failures and pushed the fixes. When you have time, could you please take another look? Thank you!

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -258,6 +258,41 @@ export class RequestHandler {
     throw error
   }
 
+  private resolveErrorMeta(errorMeta: Record<string, unknown> | undefined, topLevelModelName?: string) {
+    if (!errorMeta) {
+      return topLevelModelName ? { modelName: topLevelModelName } : errorMeta
+    }
+
+    const { table, ...rest } = errorMeta as { table?: string; [key: string]: unknown }
+    let resolvedModelName = typeof rest.modelName === 'string' ? rest.modelName : topLevelModelName
```
aqrln: nit: something like

```suggestion
    const { table } = errorMeta
    let resolvedModelName = typeof errorMeta.modelName === 'string' ? rest.modelName : topLevelModelName
```

since `errorMeta` is already `Record<string, unknown>` and we check the type of `table` below

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -258,6 +258,41 @@ export class RequestHandler {
...
+      const unqualifiedTable = table.split('.').pop()
+
+      const models = Object.entries(this.client._runtimeDataModel.models)
+
+      // Prefer an exact (schema-qualified) match first. Only fall back to matching
+      // on the unqualified table name if no exact match exists - otherwise, when two
+      // models in different schemas share the same bare table name, `find` could
+      // return the wrong model depending on iteration order, before ever reaching
+      // the correct exact match later in the list.
+      const exactMatch = models.find(([key, model]: [string, any]) => (model.dbName ?? key) === table)
```
aqrln: why `any` here?

packages/client/src/runtime/RequestHandler.ts
AnupamKumar-1: Thanks! Updated accordingly.

packages/client/src/runtime/RequestHandler.ts
AnupamKumar-1: Fixed, thanks!

tensordreams:
Resolved the blocking review question about \`meta.modelName\` and \`@@map\`.

The table→model mapping is done in \`RequestHandler.resolveErrorMeta\`/\`modelNameForTable\`: the driver adapters surface the violated table name on the \`UniqueConstraintViolation\` \`MappedError\` (new \`table?\` field), \`user-facing-error.ts\` forwards it as \`meta.table\`, and the client maps it back to the Prisma model via \`_runtimeDataModel.models\` (comparing \`model.dbName ?? name\`, i.e. the \`@@map\` target, against the reported table, unqualified). The lookup returns no match on ambiguity (two models sharing a bare table name under \`@@schema\`), falling back to the top-level operation's model name. The added functional test (\`AppMajorVersion\` mapped to \`app_major_versions\`) asserts \`meta.modelName === 'AppMajorVersion'\`.

Also merged \`main\` and reconciled with #29587: adapter-pg keeps the constraint index name for \`meta.target\` while still deriving the table for the modelName mapping. The mssql 2627 constraint target is left unchanged, so existing sqlserver snapshots are unaffected; table extraction there is purely additive.

Verified locally: \`@prisma/client-engine-runtime\` (217), adapter unit tests for pg/mssql/libsql/mariadb/neon/planetscale/d1 all pass; prettier clean. The functional suite was not run locally (the harness requires \`db push --force-reset\`).

packages/adapter-mssql/src/errors.ts
```diff
@@ -70,17 +73,23 @@ export function mapDriverError(error: DriverError): MappedError {
...
     case 1505: {
-      const index = error.message.split("'").at(3)
+      const segments = error.message.split("'")
+      const table = segments.at(3)?.split('.').pop()
+      const index = segments.at(5)
       return {
         kind: 'UniqueConstraintViolation',
         constraint: index ? { index } : undefined,
+        table,
       }
```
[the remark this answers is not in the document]
tensordreams: Already addressed in the current revision. Case 1505 now reads `const table = segments.at(1)?.split('.').pop()` and `const index = segments.at(3)` — the object name comes from the first quoted segment and the index from the second, exactly as suggested. (The `at(3)` for `index` also matches the value on `main`, so that part is unchanged.)

On the follow-up point about preserving the schema-qualified name: the `.split('.').pop()` is intentional and consistent with the other cases here (2627, 2601). It's also harmless downstream — `RequestHandler.modelNameForTable` normalizes the reported table to its bare last segment before matching, and `dbName` in the runtime data model is always the bare table name, so there is no exact schema-qualified match path to preserve. Keeping the extraction additive here avoids adapter-snapshot churn that we can't regenerate in this environment.

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -258,6 +258,34 @@ export class RequestHandler {
...
+      // the table name schema-qualified (e.g. `public.app_major_versions`)
+      // instead of bare (e.g. `app_major_versions`), so compare using only
+      // the last `.`-separated segment on both sides.
+      const unqualifiedTable = table.split('.').pop()
+
+      const models = this.client._runtimeDataModel.models
+      const match = Object.entries(models).find(([key, model]: [string, any]) => {
+        const dbName = model.dbName ?? key
+        return dbName === table || dbName.split('.').pop() === unqualifiedTable
+      })
```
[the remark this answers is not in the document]
tensordreams: Already addressed. This lookup was refactored into `modelNameForTable`, which no longer does a single `find` with an OR fallback. It normalizes the reported table to its bare last segment (`table.split('.').pop()`) and then `filter`s all models whose `dbName ?? name` equals it. When more than one model matches — which is exactly the schema-reuse case this comment worries about — it returns `undefined` so the caller falls back to the top-level operation's model name rather than guessing.

The suggested 'exact qualified match first' can't apply here: the runtime data model carries no `@@schema` information and `dbName` is always the bare table name, so a qualified exact match against `public.app_major_versions` is impossible. Ambiguity-detection (return `undefined`, fall back) is the safe resolution given that constraint.

packages/client-engine-runtime/src/user-facing-error.test.ts
```diff
@@ -186,5 +186,29 @@ test('rethrowAsUserFacing still wraps specifically-mapped kinds with their dedic
...
+  // top-level operation.
+  const error = makeDriverAdapterError({
+    kind: 'UniqueConstraintViolation',
+    constraint: { fields: ['appId', 'number'] },
+    table: 'app_major_versions',
+  })
+
+  try {
+    rethrowAsUserFacing(error)
+    throw new Error('expected rethrowAsUserFacing to throw')
```
aqrln: this test must use `expect(...).toThrow` instead of this hack

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -258,6 +258,55 @@ export class RequestHandler {
...
+   * Returns `undefined` when no model, or more than one model, matches.
+   */
+  private modelNameForTable(table: string): string | undefined {
+    // `dbName` is always the bare table name, while a driver adapter may
+    // report the table schema-qualified (e.g. `public.users`), so compare
+    // only the last `.`-separated segment of the reported name.
+    const tableName = table.split('.').pop()
+    const matches = Object.entries(this.client._runtimeDataModel.models).filter(
+      ([name, model]) => (model.dbName ?? name) === tableName,
+    )
```
aqrln: Why don't we check the schema? What if we have tables with identical names in different schemas?

packages/client/src/runtime/RequestHandler.ts
```diff
@@ -258,6 +258,55 @@ export class RequestHandler {
...
+    // With `@@schema`, models in different schemas can share a table name.
+    // The runtime data model carries no schema information, so an ambiguous
+    // match cannot be resolved — report no match instead of guessing, which
+    // makes the caller fall back to the top-level operation's model name.
```
aqrln: Please add the schema information to the runtime data model (on the prisma-engines side if necessary)

packages/client-engine-runtime/src/user-facing-error.test.ts
tensordreams: Done — this test now asserts via `expect(() => rethrowAsUserFacing(error)).toThrow(...)` instead of the manual try/catch: one call checks it throws `UserFacingError`, and a second matches the thrown error with `expect.objectContaining({ code: 'P2002', meta: expect.objectContaining({ table: 'app_major_versions', driverAdapterError: error }) })`.

Verified with `pnpm --filter @prisma/client-engine-runtime test` (user-facing-error.test.ts): 9 passed. I also confirmed the matcher actually fails (not a vacuous pass) by temporarily flipping the expected table — it errored with `expected error to match asymmetric matcher` — then reverted.

packages/client/src/runtime/RequestHandler.ts
tensordreams: Done. The runtime data model now carries each model's DB schema, and `modelNameForTable` uses it.

It turned out no prisma-engines change was necessary: the query compiler's DMMF already emits per-model `schema` (`query-compiler/dmmf/src/serialization_ast/datamodel_ast.rs` `pub schema: Option<String>`, populated from `model.schema()` in `datamodel_ast_builder.rs`), and `dmmfToRuntimeDataModel` already preserves it. The only thing dropping it was `pruneRuntimeDataModel` (used for the `client` / `wasm-compiler-edge` runtimes). I added `schema` to `PrunedRuntimeModel` and to the pruning, so it now reaches `_runtimeDataModel.models[*].schema` at runtime.

`modelNameForTable` (extracted to a testable helper at `packages/client/src/runtime/utils/modelNameForTable.ts`) now disambiguates by `(schema, table)`: when the reported table is schema-qualified (`<schema>.<table>`) it matches on both the reported schema and the mapped table name; when it isn't, it keeps the bare-table behavior and still returns `undefined` on genuine ambiguity so the caller falls back to the top-level model. Added unit tests for two models `@@map`-ped to the same table under different `@@schema` (resolving correctly per schema, and staying ambiguous → `undefined` for a bare table name).

One honest caveat / natural follow-up: the current adapters report the bare table name for P2002 (adapter-pg uses `error.table`; adapter-mssql strips the schema via `.split('.').pop()`), so for cross-schema-ambiguous tables the resolver still falls back safely today rather than firing the new schema path. Surfacing the schema from the adapters (schema-qualified table or a dedicated field) would light this up for real errors — I kept that out of this PR since it ripples through `MappedError` and the adapter snapshots, but happy to do it here or separately if you'd prefer.

AnupamKumar-1:
Hi @aqrln  , @tensordreams ! I synced to the latest branch and ran the prisma-dot-dmmf (provider=postgresql, js_pg) test locally. It passes, and the snapshot is already up to date (no changes with -u). If possible, could you rerun the failed jobs? I also noticed the Lint failure doesn’t seem related to the changes in this PR. Thanks!

PR #29630, opened 2026-06-09
fix(adapter-mssql): set VarBinary type when null Bytes? argument is passed #29557

### LINKED ISSUE #29557: [MSSQL] Setting Bytes? (@db.VarBinary(Max)) to null throws "Implicit conversion from nvarchar to varbinary(max)"
## Bug description

When updating a `Bytes?` field mapped to `@db.VarBinary(Max)` with `null` using `@prisma/adapter-mssql`, the query fails with an implicit type conversion error. The adapter is sending the `null` parameter typed as `nvarchar` instead of respecting the `varbinary(max)` column type.

## How to reproduce

**Schema:**
```prisma
model Pessoa {
  idPessoa              String  @id
  assinaturaBic         Bytes?  @db.VarBinary(Max)
  assinaturaBicExtensao String?
  assinaturaBicNoS3     Boolean
}
```

**Code:**
```typescript
await prisma.pessoa.update({
  where: { idPessoa },
  data: {
    assinaturaBic: null,
    assinaturaBicExtensao: null,
    assinaturaBicNoS3: false,
  },
});
```

## Expected behavior

The field should be set to `NULL` in the database without errors.

## Actual behavior

```
DriverAdapterError: Implicit conversion from data type nvarchar to varbinary(max) is not allowed. Use the CONVERT function to run this query.
    at MssqlTransaction.onError (.../node_modules/@prisma/adapter-mssql/dist/index.js:709:11)
    at MssqlTransaction.performIO (.../node_modules/@prisma/adapter-mssql/dist/index.js:704:12)
    at MssqlTransaction.performIO (.../node_modules/@prisma/adapter-mssql/dist/index.js:722:14)
    at MssqlTransaction.executeRaw (.../node_modules/@prisma/adapter-mssql/dist/index.js:692:13)
  cause: {
    originalCode: 'EREQUEST',
    originalMessage: 'Implicit conversion from data type nvarchar to varbinary(max) is not allowed. Use the CONVERT function to run this query.',
    kind: 'mssql',
    code: 257
  },
  clientVersion: '7.8.0'
```

## Workaround

Running raw SQL directly works correctly:

```sql
UPDATE Pessoa SET assinaturaBic = CONVERT(varbinary(max), NULL) WHERE idPessoa = '...'
```

## Environment

```
prisma                : 7.8.0
@prisma/client        : 7.8.0
@prisma/adapter-mssql : 7.8.0
OS                    : Windows
Node.js               : 24.15.0
Database              : Microsoft SQL Server 2019
```

AnupamKumar-1:
Fixes #29557 

When setting a `Bytes?` (`@db.VarBinary(Max)`) field to `null`, the adapter was sending the parameter without a type, causing MSSQL to default to `nvarchar` and throw an implicit conversion error.

Fixed by explicitly passing `sql.VarBinary` as the type for null bytes arguments in `performIO`.

All 87 tests passing locally.

### DIFF
--- a/packages/adapter-mssql/src/mssql.test.ts
+++ b/packages/adapter-mssql/src/mssql.test.ts
@@ -1,6 +1,8 @@
 import { Mutex } from 'async-mutex'
 import { beforeEach, describe, expect, Mock, test, vi } from 'vitest'
 
+import { PrismaMssql } from './index'
+
 /**
  * Tests for MssqlTransaction race condition fix.
  *
@@ -166,5 +168,69 @@ describe('MssqlTransaction mutex serialization', () => {
     const commitIndex = events.findIndex((e) => e.startsWith('commit:called'))
     const lastQueryEndIndex = events.lastIndexOf('query:end')
     expect(commitIndex).toBeGreaterThan(lastQueryEndIndex)
+  })
+})
+
+describe('null Bytes? field handling', () => {
+  const connectionString = process.env.TEST_MSSQL_URI
+
+  test('setting a VarBinary(Max) field to null does not throw implicit conversion error', async () => {
+    // This test requires a live MSSQL connection. It is intentionally skipped
+    // in CI environments that do not provide TEST_MSSQL_URI. Integration tests
+    // against a real MSSQL instance are run separately in the MSSQL CI pipeline.
+    if (!connectionString) {
+      console.log('Skipping integration test: TEST_MSSQL_URI not set')
+      return
+    }
+
+    const factory = new PrismaMssql(connectionString)
+    const adapter = await factory.connect()
+
+    try {
+      await adapter.executeRaw({
+        sql: `IF OBJECT_ID('dbo.PrismaTest') IS NOT NULL DROP TABLE dbo.PrismaTest`,
+        args: [],
+        argTypes: [],
+      })
+
+      await adapter.executeRaw({
+        sql: `CREATE TABLE dbo.PrismaTest (id NVARCHAR(36) PRIMARY KEY, bic VARBINARY(MAX))`,
+        args: [],
+        argTypes: [],
+      })
+
+      await adapter.executeRaw({
+        sql: `INSERT INTO dbo.PrismaTest VALUES (@P1, @P2)`,
+        args: ['test-id', 'AQID'],
+        argTypes: [
+          { scalarType: 'string', arity: 'scalar' },
+          { scalarType: 'bytes', arity: 'scalar' },
+        ],
+      })
+
+      const result = await adapter.executeRaw({
+        sql: `UPDATE dbo.PrismaTest SET bic = @P1 WHERE id = @P2`,
+        args: [null, 'test-id'],
+        argTypes: [
+          { scalarType: 'bytes', arity: 'scalar' },
+          { scalarType: 'string', arity: 'scalar' },
+        ],
+      })
+
+      expect(result).toBeGreaterThanOrEqual(0)
+    } finally {
+      // Use IF OBJECT_ID so cleanup remains idempotent even if
+      // the test fails before CREATE TABLE completes.
+      // Always dispose the connection, even if DROP TABLE fails.
+      try {
+        await adapter.executeRaw({
+          sql: `IF OBJECT_ID('dbo.PrismaTest') IS NOT NULL DROP TABLE dbo.PrismaTest`,
+          args: [],
+          argTypes: [],
+        })
+      } finally {
+        await adapter.dispose()
+      }
+    }
   })
 })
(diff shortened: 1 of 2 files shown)


daltino (commented):
Great fix for a real pain point — implicit conversion errors in MSSQL are notoriously cryptic, and this tackles the root cause cleanly. The approach is correct and the test is a welcome addition. I have a few suggestions worth considering before merging.

---

## Review

### `packages/adapter-mssql/src/mssql.ts`

**The core fix is correct.** When `mssql` receives a `null` value with no type hint, it falls back to `nvarchar`, which can't be implicitly converted to `varbinary`. Explicitly passing `sql.VarBinary` resolves the ambiguity.

**Suggestion 1 — generalise the null-type pattern**

The same implicit-conversion problem *could* affect other binary/typed columns when they're `null`. Right now you're special-casing `bytes`, but consider whether other scalar types need the same treatment. At minimum, a comment explaining *why* `bytes` is special-cased (vs. relying on `mapArg` for everything else) would help future maintainers:

```ts
// mssql cannot infer the correct type for a null binary parameter and
// defaults to nvarchar, causing an implicit conversion error. We must
// supply the type explicitly.
if (query.args[i] === null && argType.scalarType === 'bytes') {
  req.input(`P${i + 1}`, sql.VarBinary, null)
} else {
  req.input(`P${i + 1}`, mapArg(query.args[i], argType))
}
```

**Suggestion 2 — check whether `mapArg` already receives `null` and handles it elsewhere**

It's worth verifying (and documenting) that `mapArg` does *not* already attempt to set a type when the value is `null`. If it silently passes through `null` without a type for other scalar types too, there may be a latent issue with e.g. `DateTime?` or `Float?` fields. A brief comment or a link to the `mapArg` implementation would add confidence.

**Suggestion 3 — minor: extract the condition**

Purely cosmetic, but pulling the special-case into a helper makes the loop body easier to scan:

```ts
function inputParam(req, name, arg, argType) {
  if (arg === null && argType.scalarType === 'bytes') {
    req.input(name, sql.VarBinary, null)
  } else {
    req.input(name, mapArg(arg, argType))
  }
}
```

Not blocking — just a readability nudge.

---

### `packages/adapter-mssql/src/mssql.test.ts`

The test validates the right behaviour (no throw on `null` assignment) and has a sensible skip-if-no-connection guard.

**Suggestion 4 — the test requires a live DB connection**

The existing mutex tests in this file are fully mocked, so they run anywhere. The new test silently skips when `TEST_MSSQL_URI` is unset, which is fine for local dev but means CI without a real MSSQL instance will never exercise this path. Consider:

- Adding a comment explaining the skip is intentional and pointing to wherever integration tests are run with a live DB.
- Or — if this can be mocked — verifying that `req.input` is called with `(name, sql.VarBinary, null)` rather than relying on a full round-trip. A unit-style mock test would run in every CI environment.

**Suggestion 5 — schema/teardown robustness**

```ts
await prisma.$executeRawUnsafe(`
  CREATE TABLE TestVarBinary (id INT PRIMARY KEY, data VARBINARY(MAX))
`)
```

If the test fails before the `DROP TABLE`, subsequent runs will fail on the `CREATE TABLE` with "object already exists". Wrap the setup/teardown in a `try/finally`, or use `IF OBJECT_ID(...) IS NOT NULL DROP TABLE ...` in the teardown.

**Suggestion 6 — disconnect on failure**

If any assertion throws before `await prisma.$disconnect()`, the connection leaks. Wrap the body in `try/finally`:

```ts
try {
  // ... test body
} finally {
  await prisma.$disconnect()
}
```

---

### Summary

| | |
|---|---|
| Bug fix correctness | ✅ |
| Handles the described regression | ✅ |
| Test added | ✅ |
| Test isolation (teardown safety) | ⚠️ needs `try/finally` |
| Coverage in always-on CI | ⚠️ worth discussing |
| Generalisation / future-proofing | 💡 optional but worth a comment |

The fix itself is solid. Addressing the teardown robustness (Suggestion 5 & 6) would be the main thing I'd want to see before merging. Everything else is optional polish.

AnupamKumar-1:
@daltino Thanks for the thorough review!

* Addressed suggestions 5 & 6: cleanup now uses IF OBJECT_ID(...) IS NOT NULL DROP TABLE to make teardown idempotent, and dispose() is executed in a nested finally block so the connection is always released even if cleanup fails.
* Addressed suggestion 4: added a comment clarifying that the test is intentionally skipped when TEST_MSSQL_URI is not provided and that MSSQL integration tests run separately against a real MSSQL instance.
* Addressed suggestion 1: added a comment explaining why bytes requires special handling when the value is null (mssql defaults to nvarchar rather than VARBINARY, which causes the implicit conversion error).

Tested locally against a real MSSQL instance running in Docker; all tests pass, including the new integration test (87/87).

Thanks again for the review.

aqrln (approved):
Thanks!

PR #29634, opened 2026-06-10
fix(adapter-pg): preserve error.constraint on 23505 unique violations

### LINKED ISSUE #29495: adapter-pg: mapDriverError drops error.constraint for 23505 but uses it for 23503

tsushanth:
Closes #29495.

## Summary

`@prisma/adapter-pg`'s `mapDriverError` treated `error.constraint` (populated by node-postgres from `PG_DIAG_CONSTRAINT_NAME`) asymmetrically between unique-violation (23505) and foreign-key-violation (23503) branches:

- `case '23503'` already falls back to `{ index: error.constraint }` when `error.column` isn't populated.
- `case '23505'` ignored `error.constraint` entirely and only emitted `{ fields }` parsed out of `error.detail`.

`@prisma/adapter-mysql` already preserves the constraint name as `constraint.index` on 1062 (see #29344), so Postgres was the outlier here.

## Why it matters

For **partial unique indexes**, the parsed column list is ambiguous — two unique constraints on the same table can share key columns. For example, a full unique on `(user_id, account_id)` and a partial unique on `(account_id) WHERE role = 'owner'` both surface as 23505 with the same parsed `fields`. The constraint name is the only signal that disambiguates which one fired, and callers shouldn't have to regex-match `originalMessage` (not public API per #28281) to know.

## Fix

Prefer `error.constraint` (mapped to `{ index }`, matching the 23503 shape) when it's set, falling back to the existing `{ fields }` parse when it isn't:

```ts
case '23505': {
  let constraint: { fields: string[] } | { index: string } | undefined
  if (error.constraint) {
    constraint = { index: error.constraint }
  } else {
    const fields = error.detail?.match(/Key \(([^)]+)\)/)?.at(1)?.split(', ')
    if (fields !== undefined) {
      constraint = { fields }
    }
  }
  return { kind: 'UniqueConstraintViolation', constraint }
}
```

`MappedError.UniqueConstraintViolation` already accepts `{ index: string }` in `@prisma/driver-adapter-utils`, so this is a non-breaking widening of what callers may already see from the MySQL adapter.

## Tests

Three tests under `src/__tests__/errors.test.ts`:

- `should handle UniqueConstraintViolation (23505) with detail only` — existing behaviour, unchanged.
- `should handle UniqueConstraintViolation (23505) with constraint` — new branch.
- `should handle UniqueConstraintViolation (23505) with both constraint and detail (constraint wins)` — pins the precedence rule.

All 35 tests in `errors.test.ts` pass.

### DIFF
--- a/packages/adapter-pg/src/__tests__/errors.test.ts
+++ b/packages/adapter-pg/src/__tests__/errors.test.ts
@@ -33,11 +33,41 @@ describe('convertDriverError', () => {
     })
   })
 
-  it('should handle UniqueConstraintViolation (23505)', () => {
+  it('should handle UniqueConstraintViolation (23505) with detail only', () => {
     const error = { code: '23505', message: 'msg', severity: 'ERROR', detail: 'Key (id)' }
     expect(convertDriverError(error)).toEqual({
       kind: 'UniqueConstraintViolation',
       constraint: { fields: ['id'] },
+      originalCode: error.code,
+      originalMessage: error.message,
+    })
+  })
+
+  it('should handle UniqueConstraintViolation (23505) with constraint', () => {
+    const error = { code: '23505', message: 'msg', severity: 'ERROR', constraint: 'users_email_partial_idx' }
+    expect(convertDriverError(error)).toEqual({
+      kind: 'UniqueConstraintViolation',
+      constraint: { index: 'users_email_partial_idx' },
+      originalCode: error.code,
+      originalMessage: error.message,
+    })
+  })
+
+  it('should handle UniqueConstraintViolation (23505) with both constraint and detail (constraint wins)', () => {
+    // Two unique constraints can share key columns (e.g. a full unique on
+    // `(user_id, account_id)` and a partial unique on `(account_id)
+    // WHERE role = 'owner'`); the constraint name is the only signal that
+    // disambiguates which one fired, so we prefer it over the parsed columns.
+    const error = {
+      code: '23505',
+      message: 'msg',
+      severity: 'ERROR',
+      constraint: 'users_email_partial_idx',
+      detail: 'Key (email)',
+    }
+    expect(convertDriverError(error)).toEqual({
+      kind: 'UniqueConstraintViolation',
+      constraint: { index: 'users_email_partial_idx' },
       originalCode: error.code,
       originalMessage: error.message,
     })
--- a/packages/adapter-pg/src/errors.ts
+++ b/packages/adapter-pg/src/errors.ts
@@ -75,13 +75,23 @@ function mapDriverError(error: DatabaseError): MappedError {
         message: error.message,
       }
     case '23505': {
-      const fields = error.detail
-        ?.match(/Key \(([^)]+)\)/)
-        ?.at(1)
-        ?.split(', ')
+      let constraint: { fields: string[] } | { index: string } | undefined
+
+      if (error.constraint) {
+        constraint = { index: error.constraint }
+      } else {
+        const fields = error.detail
+          ?.match(/Key \(([^)]+)\)/)
+          ?.at(1)
+          ?.split(', ')
+        if (fields !== undefined) {
+          constraint = { fields }
+        }
+      }
+
       return {
         kind: 'UniqueConstraintViolation',
-        constraint: fields !== undefined ? { fields } : undefined,
+        constraint,
       }
     }
     case '23502': {
--- a/packages/adapter-pg/src/errors.ts
+++ b/packages/adapter-pg/src/errors.ts
@@ -57,6 +57,13 @@ export function convertDriverError(error: unknown): DriverAdapterErrorObject {
   throw error
 }
 
+function parseKeyFields(detail: string | undefined): string[] | undefined {
+  return detail
+    ?.match(/Key \(([^)]+)\)/)
+    ?.at(1)
+    ?.split(', ')
+}
+
 function mapDriverError(error: DatabaseError): MappedError {
   switch (error.code) {
     case '22001':
@@ -80,10 +87,7 @@ function mapDriverError(error: DatabaseError): MappedError {
       if (error.constraint) {
         constraint = { index: error.constraint }
       } else {
-        const fields = error.detail
-          ?.match(/Key \(([^)]+)\)/)
-          ?.at(1)
-          ?.split(', ')
+        const fields = parseKeyFields(error.detail)
         if (fields !== undefined) {
           constraint = { fields }
         }
@@ -95,10 +99,7 @@ function mapDriverError(error: DatabaseError): MappedError {
       }
     }
     case '23502': {
-      const fields = error.detail
-        ?.match(/Key \(([^)]+)\)/)
-        ?.at(1)
-        ?.split(', ')
+      const fields = parseKeyFields(error.detail)
       return {
         kind: 'NullConstraintViolation',
         constraint: fields !== undefined ? { fields } : undefined,


tsushanth:
Thanks @coderabbitai, good DRY catch. Pushed e201c7b extracting `parseKeyFields` so the regex lives in one place and both the 23505 detail-only branch and the 23502 case read at the intent layer. Existing tests at `errors.test.ts:36, 47, 62, 77` already exercise both paths, so behaviour is unchanged.

On the outside-diff edge-case suggestion (empty constraint string, malformed detail, neither present): I'd rather not add those. The constraint-empty-string case is unreachable through node-pg (an empty `constraint` is parsed as undefined upstream), the malformed-detail and both-absent cases collapse into the same `constraint: undefined` branch already covered by the existing tests, and `fields !== undefined` already guards the consumers. Adding them would test the regex/optional-chain language semantics rather than the adapter contract.

daltino (commented):
Great fix — the asymmetry between `23503` and `23505` was a real bug and the PR addresses it cleanly. The extracted `parseKeyFields` helper is a nice bonus since it eliminates duplicated regex logic. A few things worth discussing before merging:

---

## Review

### Correctness

The core logic looks right. Falling back to `{ index: error.constraint }` when `detail` is absent (or the regex doesn't match) mirrors the existing `23503` branch and is the correct behaviour — `error.constraint` is always populated by libpq when the error comes from a constraint violation.

One subtle concern worth double-checking: **what happens when both `detail` and `constraint` are present?** The current precedence (detail → constraint) seems intentional, since `detail` gives us the actual column names rather than just the constraint name. Make sure this is documented in a comment so the next reader doesn't wonder why we're ignoring `constraint` in that branch.

```ts
case '23505': {
  // Prefer detail (gives column names); fall back to constraint name.
  const fields = parseKeyFields(error.detail) ?? (error.constraint ? [error.constraint] : undefined)
  ...
}
```

### Edge Cases

**1. Composite key with spaces in column names**

`parseKeyFields` splits on `', '`. A column name containing a comma (unusual but technically valid in quoted identifiers) would silently produce wrong field names. This is an edge case Postgres itself handles weirdly, so it may not be worth solving now, but a comment acknowledging the limitation would be helpful.

**2. `detail` present but regex doesn't match**

If `detail` exists but has an unexpected format (e.g., a translated locale message or a future Postgres change to the format), `parseKeyFields` returns `undefined` and the code falls back to `constraint`. That's good defensive behaviour — just confirm the fallback path is exercised by a test.

**3. Both `detail` and `constraint` absent**

What does the returned object look like when both are `undefined`? If `fields` ends up `undefined`, is `{ fields: undefined }` or `{}` returned? Check that downstream consumers handle the missing `fields` key gracefully.

### Test Coverage

The new test for `constraint`-only is great. A couple of gaps:

- **No test for the "both `detail` and `constraint` present" case** — verify that `detail` takes precedence and `constraint` is ignored.
- **No test for the "neither `detail` nor `constraint` present" case** — what does the function return? This is currently untested for `23505`, whereas `23503` presumably has similar gaps. Consider adding it.
- The renamed existing test (`with detail only`) is a nice clarification — good call.

```ts
it('should handle UniqueConstraintViolation (23505) with both detail and constraint', () => {
  const error = {
    code: '23505',
    message: 'msg',
    severity: 'ERROR',
    detail: 'Key (email)=(foo@bar.com) already exists.',
    constraint: 'users_email_key',
  }
  expect(convertDriverError(error)).toEqual({
    kind: 'UniqueConstraintViolation',
    fields: ['email'], // detail wins
  })
})
```

### Code Style

`parseKeyFields` is a good extraction. One small nit: the name is slightly generic — `parseDetailFields` or `parseUniqueViolationFields` would tie it more clearly to its purpose. Not a blocker at all.

The optional-chaining chain:

```ts
detail?.match(...)?.at(1)?.split(', ')
```

is idiomatic modern TypeScript and readable. ✅

### Performance

Not a concern here — this is only called on error paths.

---

### Summary

| Area | Status |
|---|---|
| Core bug fix | ✅ Correct |
| `parseKeyFields` extraction | ✅ Nice cleanup |
| Fallback precedence (detail → constraint) | ✅ Mirrors 23503 |
| Test: constraint-only path | ✅ Added |
| Test: both present (detail wins) | ⚠️ Missing |
| Test: neither present | ⚠️ Missing |
| Composite key with spaces comment | 💬 Nice-to-have |

Overall this is a solid, minimal fix. Add the two missing test cases and optionally a short comment about precedence, and it's ready to merge. 🎉

nahrinoda (commented):
Small correction on @daltino's review — the precedence is **constraint → detail**, not detail → constraint. The code checks `error.constraint` first:

```ts
if (error.constraint) {
  constraint = { index: error.constraint }
} else {
  const fields = parseKeyFields(error.detail)
  ...
}
```

This is the right call for the partial unique index case the PR describes — two constraints sharing the same key columns can only be disambiguated by name, so the constraint name should take precedence over parsed column names.

The "both present" test is also already included (`'with both constraint and detail (constraint wins)'`) and correctly pins this precedence. So daltino's suggested test would actually fail against the current code since it asserts `fields: ['email']` (detail wins).

The "neither present" case (no `constraint`, no `detail`) is still untested — `constraint` would be `undefined`, giving `{ kind: 'UniqueConstraintViolation', constraint: undefined }`. Might be worth a one-liner test if the author wants completeness, but not a blocker.

Clean fix overall. The `parseKeyFields` extraction and the 23503/23505 symmetry are solid.

aqrln (approved):
This looks good to me, thank you!

Could you please [sign the CLA](https://github.com/prisma/prisma/pull/29634#issuecomment-4673851934) so we can merge the PR?

tensordreams:
Hi @tsushanth — thanks for this contribution, and sorry for the slow response.

https://cla-assistant.io/prisma/prisma?pullRequest=29634

aqrln:
Closing in favor of https://github.com/prisma/prisma/pull/29587

PR #22239, opened 2023-12-04
fix(adapter-neon,adapter-pg): treat all custom types as text

aqrln:
Treat all custom Postgres types as text in the result set.

This is an alternative to https://github.com/prisma/prisma/pull/21918.

Thanks to driver adapters using Postgres text protocol, unlike our native driver that uses the binary protocol, we can get away with treating all unknown extension types as text if we want to expose them to users as such — we (or the driver) don't need to know how to decode them and convert to strings if they are already strings in the first place.

However, we cannot distinguish between known types added by extensions and enums without querying the catalog for type names. This may look like a problem at first, but in reality we can get away with treating enums as text in result set too, the response IR layer in the query engine uses the query schema to know it's an enum, not the type hint that comes from `quaint::Value`/`PrismaValue`. If the output type in the query schema says it's an enum, it will treat `PrismaValue::String` and `PrismaValue::Enum` equally: https://github.com/prisma/prisma-engines/blob/12d71c4263690f3990d590fbff611b2a24ef2dbc/query-engine/core/src/response_ir/internal.rs#L813. So mapping Postgres enums to `quaint::ValueType::Text` is fine, and is already supported by QE.

Engines test suite passing: https://github.com/prisma/prisma-engines/pull/4531

Fixes: https://github.com/prisma/prisma/issues/21807
Closes: https://github.com/prisma/prisma/pull/21918

PR #29648, opened 2026-06-18
Fix raw query enum array decoding

### LINKED ISSUE #28623: Rust free client does not infer types with raw queries
### Bug description

I have a raw query like this:

```
  const results = await tx.$queryRaw<
    {
      id: string;
      name: string;
      description: string;
      isDefault: boolean;
      permissions: UserPermission[];
      rolesCount: number;
    }[]
  >`
    SELECT
      roles.id,
      name,
      description,
      "isDefault",
      permissions,
      COUNT(users_clients.user_id)::integer as "userCount"
    FROM
      roles
    LEFT JOIN users_clients ON users_clients."roleId" = roles.id
    WHERE
      roles."clientId" = ${clientId}
      AND roles."deletedAt" IS NULL
      AND users_clients."deletedAt" IS NULL
    GROUP BY roles.id
  `;
```

Where before, permissions was an array, it now returns a string.

### Severity

Expected: returns an array like before

### Frequency

Consistently reproducible

### Is this a regression?

Regression for v7 rust free

### Workaround

To parse the array afterward but that's a shame it has not been tested :/ 

### Prisma Schema & Queries

```prisma
```
```ts
```

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

- **OS:** 
- **Database:** 
- **Node.js version:** 

### Prisma Version

```
// Prisma version output
```

sawirricardo:
## Summary
- detect PostgreSQL user-defined enum array OIDs in adapter-pg raw query results
- parse enum array literals into arrays and report them as enum-array columns
- add adapter unit coverage and a functional raw-query regression test

## Testing
- `mise x node@22.12 -- pnpm --filter @prisma/adapter-pg test`
- `mise x node@22.12 -- pnpm --filter @prisma/adapter-pg build`
- `TEST_FUNCTIONAL_POSTGRES_URI='postgres://prisma:prisma@localhost:15432/PRISMA_DB_NAME' mise x node@22.12 -- pnpm run test:functional:code tests/functional/enum-array/tests.ts --provider postgresql --adapter js_pg --client-runtime client --runInBand`

Fixes #28623


tensordreams (commented):
Thank you for digging into this, @sawirricardo — the test coverage is appreciated, and the regression you are addressing (#28623) is real: the Rust engine in v6 did decode enum arrays in raw query results, and the driver-adapter-based client currently does not.

That said, we are going to pass on this approach: it reintroduces something we have deliberately decided against before, namely **implicit type-introspection queries issued by the adapter by default**.

Some history for context:

- In the Rust engine era, automatic type resolution for unknown OIDs (tokio-postgres's `typeinfo` catalog queries) was a persistent source of operational pain with connection poolers. It is the reason Prisma maintained a [fork of rust-postgres](https://github.com/prisma/rust-postgres/tree/pgbouncer-mode) whose `pgbouncer-mode` disables the prepared-statement and type caches behind transaction-mode poolers (so introspection effectively re-ran per query), and why users had to know to set `pgbouncer=true` on their connection strings.
- When driver adapters were introduced, we intentionally mapped all user-defined types to text and deferred enum handling to layers that have schema knowledge, precisely to avoid catalog lookups at this level — see the comment right above the `default:` branch this PR modifies in `conversion.ts`, and #22239.
- When this specific enum-array gap came up, we considered a catalog lookup along these lines and instead shipped an explicit opt-in escape hatch: the `userDefinedTypeParser` adapter option (#27941). It receives the queryable as its third argument specifically so that a parser can perform its own catalog lookups when the application is fine with the extra queries.

For completeness, a few implementation-level issues we noticed, which any default-on variant would also need to solve:

- The OID cache lives on `PgQueryable`, but `startTransaction()` creates a fresh `PgTransaction` (which extends `PgQueryable`) per transaction, so every interactive transaction pays the catalog round-trip again — and the lookup runs on the transaction's connection, inside the user's open transaction.
- The lookup fires for any OID >= 16384 (`vector`, `citext`, PostGIS types, `hstore`, ...), but only enum arrays benefit; a `citext[]` column, for example, still comes back as an unparsed literal, so the broader problem in #28623 is only partially addressed.
- It runs even when `userDefinedTypeParser` is configured, adding a hidden extra query for users of the existing opt-in mechanism and changing their reported column types from `Text` to `EnumArray`.
- Negative cache entries are seeded before the catalog query runs, so if that query fails, the affected OIDs stay misclassified as `Text` for the lifetime of the queryable — and the error fails a user query whose own execution had already succeeded.
- Two concurrent first queries race: the second one sees the pre-seeded `null` entries and returns strings while the first returns arrays.
- Only `adapter-pg` is covered, so the same SQL would behave differently on `@prisma/adapter-neon` and `@prisma/adapter-ppg`; the catalog query would also run against CockroachDB (which uses `adapter-pg`), where it has not been validated.
- Nit: `t.oid::int4` is binary-coercible and goes negative for OIDs >= 2^31, so such an OID would never match the unsigned `dataTypeID` reported in the row description.

If you are running into this today, there are two supported ways to get proper arrays back:

1. Cast in SQL — no configuration needed:

   ```sql
   SELECT "plans"::text[] FROM "User"
   ```

2. Configure `userDefinedTypeParser` once on the adapter. There is a complete, tested example — including the catalog lookup for OID-precise detection, performed on your own terms — in [`packages/client/tests/functional/enum-array/tests.ts`](https://github.com/prisma/prisma/blob/main/packages/client/tests/functional/enum-array/tests.ts).

If we later conclude that default-on parity with the old engine is worth it here — a fair discussion, especially now that PgBouncer >= 1.21 supports protocol-level prepared statements — we would want to make that call deliberately and across all Postgres adapters at once, rather than land it as a side effect of a bug fix.

Closing this one, but thank you again for the contribution — we hope it does not discourage you from future ones!

Issue #29646, opened 2026-06-15
DeprecationWarning with @prisma/adapter-pg v7.8.0


chrinoreEsambu:
### Bug description

**Environment:**

* Prisma: 7.8.0
* @prisma/client: 7.8.0
* @prisma/adapter-pg: 7.8.0
* pg: latest compatible version
* Node.js: (your version)

Issue:

When using @prisma/adapter-pg v7.8.0, a deprecation warning is emitted by the pg driver:

DeprecationWarning: Calling client.query() when the client is already executing a query is deprecated and will be removed in pg@9.0.

After investigating, the warning appears to originate from the internal implementation of @prisma/adapter-pg.

In node_modules/@prisma/adapter-pg/dist/index.js, the performIO method calls:

const result = await this.client.query(
{
text: sql,
values,
rowMode: "array",
...
},
values
);

The values parameter is passed **twice**:

1. Inside the query config object (`values`)
2. Again as the second argument to `client.query()`

This seems to trigger pg's deprecated query(config, values) path and results in the warning, even though application code uses proper async/await patterns.

Expected behavior:
The adapter should call client.query() using only the query config object when values are already provided inside the configuration.

Impact:
No functional issue observed, but a deprecation warning is emitted on every startup/query and may become problematic when pg 9.0 removes the deprecated behavior.

### Severity

⚠️ Major: Breaks core functionality (e.g., migrations fail)

### Steps

1. Create a new project:

```bash
mkdir prisma-pg-warning
cd prisma-pg-warning
npm init -y
```

2. Install dependencies:

```bash
npm install prisma@7.8.0 @prisma/client@7.8.0 @prisma/adapter-pg@7.8.0 pg
```

3. Create a PostgreSQL database.

4. Create `schema.prisma`:

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

model Test {
  id String @id @default(uuid())
}
```

5. Generate the client:

```bash
npx prisma generate
```

6. Create `index.js`:

```js
import { PrismaPg } from "@prisma/adapter-pg";
import { PrismaClient } from "@prisma/client";
import { Pool } from "pg";

const pool = new Pool({
  connectionString: process.env.DATABASE_URL,
});

const adapter = new PrismaPg(pool);
const prisma = new PrismaClient({ adapter });

await prisma.test.findMany();
```

7. Run:

```bash
node index.js
```

### Observed behavior

[...]

PR #29650, opened 2026-06-18
chore(adapter-pg): remove duplicate values parameter

Swapanrishi:
## What

Removes the redundant second `values` argument from the `this.client.query(...)` call in `PgQueryable.performIO`. The `values` array is already provided inside the query config object, so passing it again as the second positional argument was a no-op. Also updates the corresponding unit tests to match the single-argument call signature.

## Why

Small cleanup. The duplicate argument has no runtime effect: `pg`'s `normalizeQueryConfig` simply reassigns the same array onto `config.values`, so removing it changes no behavior.

### DIFF
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -103,24 +103,21 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
     const values = args.map((arg, i) => mapArg(arg, query.argTypes[i]))
 
     try {
-      const result = await this.client.query(
-        {
-          name: this.pgOptions?.statementNameGenerator?.(query),
-          text: sql,
-          values,
-          rowMode: 'array',
-          types: {
-            getTypeParser: (oid: number, format?) => {
-              if (format === 'text' && customParsers[oid]) {
-                return customParsers[oid]
-              }
+      const result = await this.client.query({
+        name: this.pgOptions?.statementNameGenerator?.(query),
+        text: sql,
+        values,
+        rowMode: 'array',
+        types: {
+          getTypeParser: (oid: number, format?) => {
+            if (format === 'text' && customParsers[oid]) {
+              return customParsers[oid]
+            }
 
-              return types.getTypeParser(oid, format)
-            },
+            return types.getTypeParser(oid, format)
           },
         },
-        // Removed the duplicate `values` argument here!
-      )
+      })
 
       return result
     } catch (e) {
--- a/packages/adapter-pg/src/pg.ts
+++ b/packages/adapter-pg/src/pg.ts
@@ -119,7 +119,7 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
             },
           },
         },
-        values,
+        // Removed the duplicate `values` argument here!
       )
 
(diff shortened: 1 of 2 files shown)


packages/adapter-pg/src/pg.ts
```diff
@@ -119,7 +119,7 @@ class PgQueryable<ClientT extends StdClient | TransactionClient> implements SqlQ
             },
           },
         },
-        values,
+        // Removed the duplicate `values` argument here!
```
aqrln: please remove this comment

tensordreams:
Thanks for the contribution, @Swapanrishi, and for digging into this! Unfortunately, after investigating I don't think this change fixes the warning from #29646, though it's a harmless cleanup on its own. Sharing the details so the analysis is on record.

### What the warning actually is

The warning was introduced in `pg` **8.19.0** and is emitted from a single condition in `pg`'s `client.js`:

```js
const queryQueueLengthDeprecationNotice = nodeUtils.deprecate(
  () => {},
  'Calling client.query() when the client is already executing a query is deprecated and will be removed in pg@9.0. ...'
)
// ...
if (this._queryQueue.length > 0) {
  queryQueueLengthDeprecationNotice()
}
```

It fires whenever `client.query()` is called on a client that is **already executing a query** (i.e. concurrent queries on one connection). It is not related to how many times `values` is passed.

### Why the duplicate `values` argument isn't the cause

Passing `values` both inside the config object and again as the second positional argument is redundant, but harmless. In `pg`'s `normalizeQueryConfig`, the second argument (an array) simply reassigns the same array onto `config.values`:

```js
if (values) {
  if (typeof values === 'function') {
    config.callback = values
  } else {
    config.values = values // the identical array
  }
}
```

There is no separate "deprecated `query(config, values)` path", and the query is only enqueued once (`this.queryQueue.push(query)`), so this never affects `queryQueue.length` and never emits the warning.

### The real root cause

This is the concurrency issue already tracked in #29407. The query interpreter dispatches parallel branches onto a single client, e.g. in `packages/client-engine-runtime/src/interpreter/query-interpreter.ts`:

```js
await Promise.all(node.args.map((arg) => this.interpretNode(arg, context).then((res) => res.value)))
```

Those parallel branches issue concurrent `performIO` -> `client.query()` calls on the same client, which is exactly what trips the `pg` warning. A real fix needs to serialize queries on a single client (or avoid concurrent dispatch onto one), not change the call site's argument list.

### Two additional notes

- This repo pins `pg@^8.16.3`, which predates 8.19.0 and doesn't contain the warning at all, so the updated unit tests can't actually exercise the reported behavior.
- The change itself is fine as a small cleanup: removing the redundant argument is a no-op at runtime, and the test updates correctly mirror the single-argument signature. If you'd like to keep it, we can retain it as a pure cleanup with an accurate description, but it shouldn't be linked as closing #29646.

Swapanrishi:
Thank you so much for the deep dive into the concurrency issue and for updating the PR description! I've updated the PR title to chore since this is just a cleanup now. Let me know if you need me to squash any commits or if it's good to go as-is

aqrln:
It's fine as is, it'll get squashed on merge anyway.

aqrln (approved):

PR #29664, opened 2026-06-28
ci: pin GitHub Actions to full commit SHAs

XananasX7:
Pin unpinned GitHub Actions to immutable commit SHAs. Defense against supply-chain attacks via mutable tags. Version tags retained as inline comments. See: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

### DIFF
--- a/.github/workflows/auto-close-github-discussions.yml
+++ b/.github/workflows/auto-close-github-discussions.yml
@@ -15,15 +15,15 @@ jobs:
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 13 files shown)


tensordreams:
Hi @XananasX7 — thanks for this, and apologies for the slow reply.

We are closing this one: this is identical to #29663, which is already closed, and we are not taking this workflow-wide pinning change in the branch-transition window. Closing the duplicate avoids splitting any future discussion across two PRs.

PR #29683, opened 2026-07-02
fix(client): keep fluent relation fields named select or include when unpacking results

spokodev:
`RequestHandler.unpack` derives the path to read a fluent-API result out of the
response with:

    const pathForGet = dataPath.filter((key) => key !== 'select' && key !== 'include')

`dataPath` is a sequence of `[selector, relationField]` pairs, where the selector
is always `select` or `include` and the relation field names are at the odd
positions. Filtering by the literal string value also strips any relation field
that is itself named `select` or `include`, so `deepGet` receives the wrong path
and returns the wrong slice of the response (often the whole response object).

Both `select` and `include` are valid Prisma relation field names. For example a
dataPath of `['select', 'select']` produced `[]` and returned the entire object
instead of `response.select`. The sibling module `resolve-result-extension-context`
already reads the same dataPath positionally (`index += 2`) and ships tests for
fields named `select` and `include`; only `unpack` was left on the naive filter.

The fix derives the path positionally through a small `dataPathToGetPath` helper,
keeping the odd-position field names. Behavior is identical for every field name
except `select` and `include`.

### DIFF
--- a/packages/client/src/runtime/RequestHandler.ts
+++ b/packages/client/src/runtime/RequestHandler.ts
@@ -278,7 +278,7 @@ export class RequestHandler {
     }
     const operation = Object.keys(data)[0]
     const response = Object.values(data)[0]
-    const pathForGet = dataPath.filter((key) => key !== 'select' && key !== 'include')
+    const pathForGet = dataPathToGetPath(dataPath)
     const extractedResponse = deepGet(response, pathForGet)
     const deserializedResponse =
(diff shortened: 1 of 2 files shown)


Ziiyodullayevv (commented):
Good fix for fluent relation fields. Keeping select or include named fields when unpacking results ensures query results maintain their expected structure.

aqrln (approved):
Thank you!

PR #29684, opened 2026-07-02
chore(migrate): modernize AI agent detection in the safety checkpoint

tensordreams:
## Overview

The AI safety checkpoint (`packages/migrate/src/utils/ai-safety.ts`) guards `prisma db drop`, `prisma db push --force-reset`, and `prisma migrate reset` by detecting AI coding agents and refusing to proceed without explicit user consent. Its detection list dated from 2025 and missed most of the current agent landscape. Every marker in this PR was verified against primary sources (the agent's official docs or source code).

## Changes

- **Cross-platform Codex CLI detection.** `CODEX_SANDBOX=seatbelt` is only set under the macOS Seatbelt sandbox, so Codex on Linux was undetected. Now also checks `CODEX_THREAD_ID`, `CODEX_CI`, and `CODEX_SANDBOX_NETWORK_DISABLED`.
- **New agents detected:** Qwen Code (`QWEN_CODE`), GitHub Copilot CLI (`COPILOT_CLI`), OpenCode (`OPENCODE`/`OPENCODE_CLIENT`), Cline (`CLINE_ACTIVE`), Goose (`AGENT=goose`), Amp (`AGENT=amp`), Crush (`CRUSH`), Augment Code (`AUGMENT_AGENT`), Antigravity (`ANTIGRAVITY_AGENT`), Replit Agent (`REPLIT_SESSION` with an `agent-` prefix, observed empirically; the human session format does not match), and Devin (the `/opt/.devin` marker file — Devin sets no env var).
- **Generic conventions as a fallback:** `AI_AGENT=<name>` (promoted by `@vercel/detect-agent`) and `AGENT=<name|1>`, so future agents adopting these conventions are caught without code changes.
- **Removed `REPLIT_CLI`:** no tool appears to actually set it.
- **Restructured as a lazy matcher chain.** Detection is a module-level list of `AgentMatcher` objects evaluated in order until one matches; each matcher declares the env vars it inspects, and the tests derive their env cleanup list from that instead of hardcoding it.
- **Error prompt:** fixed typos and added guidance to use structured user-question tools where available and to abort when running unattended.
- **Documented why `@vercel/detect-agent` is not used:** it targets deployment attribution and tolerates human false positives (`REPL_ID`, `CURSOR_TRACE_ID`, user-set `COPILOT_*` vars), which is unacceptable for a check that blocks a command; it also misses several agents covered here.
- **Added unit tests** (26 cases). The suite clears inherited agent markers per test, since the test process itself may run under an AI agent.

https://claude.ai/code/session_01AR81vnkqSuBTikM8xXU6jv


packages/migrate/src/__tests__/utils/ai-safety.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  'PRISMA_USER_CONSENT_FOR_DANGEROUS_AI_ACTION',
+]
+
+beforeEach(() => {
+  // The tests themselves may be running under an AI agent, so the inherited
+  // markers must be cleared before each test.
+  for (const name of agentEnvVars) {
+    delete process.env[name]
+  }
+})
```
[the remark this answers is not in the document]
aqrln: Good point. We should have `AgentMarker` expose the list of files it reads in the same way as the list of env vars.

packages/migrate/src/utils/ai-safety.ts
```diff
@@ -47,18 +54,90 @@ export function aiAgentConfirmationCheckpoint(): void {
...
+  // Generic conventions: `AI_AGENT=<name>` (promoted by @vercel/detect-agent)
+  // and `AGENT=<name or 1>` (set by e.g. Goose, Amp, Crush, OpenCode).
+  {
+    envVars: ['AI_AGENT', 'AGENT'],
+    detect: (env) => {
+      const generic = env.AI_AGENT || env.AGENT
+      if (!generic) {
+        return undefined
+      }
+      return generic === '1' ? 'an unidentified AI agent' : `an AI agent identifying itself as "${generic}"`
```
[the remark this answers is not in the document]
tensordreams: Done in 17d2ec0d: the generic value now collapses non-printable characters to spaces, is trimmed and capped at 64 characters, and is interpolated via JSON.stringify. A value with no printable characters falls back to "an unidentified AI agent" (presence of the marker still counts as detection).

packages/migrate/src/__tests__/utils/ai-safety.test.ts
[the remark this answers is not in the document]
tensordreams: Implemented in 17d2ec0d: `AgentMatcher` now has a `files` field mirroring `envVars` (populated with `/opt/.devin` for Devin), and the test setup mocks `node:fs.existsSync` to report all matcher-declared files as absent by default, so the suite is stable inside agent VMs. Also added a positive test for Devin detection. Superseded #29685, closing it.

packages/migrate/src/utils/ai-safety.ts
```diff
@@ -47,18 +54,111 @@ export function aiAgentConfirmationCheckpoint(): void {
...
+      if (generic === '1') {
+        return 'an unidentified AI agent'
+      }
+      // The value is ambient free-form text interpolated into instructions
+      // addressed to the agent, so it must not be able to reshape them:
+      // non-printable characters collapse to spaces and the length is capped.
+      const sanitized = generic
+        .replace(/[^\x20-\x7e]+/g, ' ')
+        .trim()
+        .slice(0, 64)
```
aqrln: Wouldn't it be better to slice first, then replace?

packages/migrate/src/__tests__/utils/ai-safety.test.ts
```diff
@@ -0,0 +1,110 @@
+import { existsSync } from 'node:fs'
+
+import { agentMatchers, aiAgentConfirmationCheckpoint } from '../../utils/ai-safety'
+
+jest.mock('node:fs', () => ({
+  ...jest.requireActual('node:fs'),
+  existsSync: jest.fn(),
+}))
```
aqrln: Get rid of this mock. Instead pass a `fileExists` function as part of the context to `AgentMarker#detect`.

packages/migrate/src/utils/ai-safety.ts
tensordreams: Done in 4fee2a2c — the value is now sliced to 64 characters first, so the sanitization regex never processes more than that.

packages/migrate/src/__tests__/utils/ai-safety.test.ts
tensordreams: Done in 4fee2a2c: `detect` now takes an `AgentDetectionContext` (`{ env, fileExists }`), `aiAgentConfirmationCheckpoint` accepts an optional context defaulting to the real environment, and the tests pass a stub `fileExists` instead of mocking `node:fs`.

aqrln (approved):

PR #29689, opened 2026-07-03
feat(cli): install Prisma agent skills during prisma init

tensordreams:
`prisma init` now installs the [prisma/skills](https://github.com/prisma/skills) catalog into freshly scaffolded projects, so AI agents working in them start with current, version-relevant Prisma knowledge instead of stale training data. First slice (S1) of the Agent-Native Prisma ORM project — spec and plan in `docs/plans/agent-native/` and `projects/agent-native/` (#29688).

## Changes

- **Skill-install runner** (`packages/cli/src/init/skill-install.ts`): executes `<runner> skills@1.5.14 add prisma/skills --agent cursor claude-code codex windsurf --skill '*' --copy -y` via execa, streaming output. The runner adapts to the invoking package manager (`npx --yes` / `pnpm dlx` / `yarn dlx` / `bunx`), detected from `npm_config_user_agent`, then the Bun runtime, then a lockfile sniff, defaulting to npm. Yarn 1 routes through npx — classic yarn has no `dlx`. `installSkills` never throws: any failure resolves to `{ ok: false, manualCommand }`.
- **Init wiring** (`packages/cli/src/Init.ts`): the install runs after file scaffolding (both default and Prisma Postgres new-project flows); `--no-skills` opts out (arg spec + help). Success appends `.claude/skills/`, `.windsurf/skills/`, `.agents/skills/`, `skills-lock.json` to the init summary; failure prints a warning with the copy-pasteable manual command and init still exits 0. The PPG existing-project path (which scaffolds nothing) skips the install.
- **Tests**: 43 passing in `packages/cli` — 17 runner unit tests (command assembly per package manager, detection fallbacks, failure shapes; injected exec, no network, no module mocks) and 26 Init tests including 4 new integration cases (default invokes the runner, `--no-skills` skips, failure is non-fatal, help lists the flag). Snapshot updates are exactly the new summary lines.

## Why

- **Default-on with a clean opt-out** implements the project spec's distribution goal (zero-configuration agent readiness), following the prisma-next precedent — with two deliberate deviations: the `skills` CLI version is **pinned** (supply-chain hygiene; bumped via Renovate) and install failure is **non-fatal** (the stable CLI must never fail scaffolding over an auxiliary network step).
- **`--copy` instead of the default symlink layout**: skills@1.5.14 with multiple `--agent` values writes only the universal `.agents/skills/` tree and silently skips the per-agent symlinks its own plan output promises (verified empirically; upstream bug report prepared). `--copy` materializes real `.claude/skills/` and `.windsurf/skills/` copies so Claude Code and Windsurf find skills where they actually look.
- **Default-branch install for now**: prisma/skills has no release tags yet; a prepared ask requests per-ORM-minor tags so the install can pin `prisma/skills#v<minor>` in a follow-up.

## Verification (live runs, built CLI, clean env)

- **Default `prisma init`**: exit 0; all 8 catalog skills present in each of `.claude/skills/`, `.windsurf/skills/`, `.agents/skills/` (24 `SKILL.md` files, zero symlinks); `skills-lock.json` written; summary matches disk.
- **Forced failure** (unreachable registry): exit 0; scaffolding intact; warning printed with the manual command.
- **`--no-skills`**: exit 0; no agent artifacts.

Full evidence and the two prepared external filings (tagging ask for prisma/skills; symlink bug for vercel-labs/skills) live in `projects/agent-native/slices/init-skill-install/verification.md` on the project branch (#29688).

**Linear:** [TML-2968](https://linear.app/prisma-company/issue/TML-2968/s1-install-prisma-skills-during-prisma-init)


packages/cli/src/init/skill-install.ts
```diff
@@ -0,0 +1,162 @@
...
+}: InstallSkillsOptions): Promise<InstallSkillsResult> {
+  let runner = runners.npm
+  try {
+    runner = detectRunner(cwd, detectOptions)
+    await exec(runner.command, installArgs(runner), { cwd })
+    return { ok: true }
+  } catch {
+    return { ok: false, manualCommand: buildManualCommand(runner) }
+  }
+}
```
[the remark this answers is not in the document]
tensordreams: Fixed in 65eac77df — `defaultExec` now passes `timeout: 60_000` to execa with a why-comment; the kill rejects into the existing `{ ok: false, manualCommand }` path, so a hung registry can no longer hang `prisma init`.

packages/cli/src/__tests__/skill-install.vitest.ts
```diff
@@ -0,0 +1,224 @@
...
+  `skills@${SKILLS_CLI_VERSION}`,
+  'add',
+  SKILLS_SOURCE,
+  '--agent',
+  ...SKILL_AGENTS,
+  '--skill',
+  '*',
+  '--copy',
+  '-y',
+]
```
aqrln: You could export installArgs from the skil-install.ts module too

packages/cli/src/init/skill-install.ts
```diff
@@ -0,0 +1,165 @@
...
+  ['pnpm-lock.yaml', 'pnpm'],
+  ['yarn.lock', 'yarn'],
+  ['bun.lock', 'bun'],
+  ['bun.lockb', 'bun'],
+  ['package-lock.json', 'npm'],
+]
+
+/** Indicates if the CLI runs on the Bun runtime. */
+export const isBun: boolean =
+  'Bun' in globalThis || typeof (process.versions as Record<string, string | undefined>).bun === 'string'
```
aqrln: This doesn't belong here since it's now shared. Put it into a separate helper module.

packages/cli/src/__tests__/skill-install.vitest.ts
tensordreams: Fixed in dd9fd3153. `installArgs` is now exported from `packages/cli/src/init/skill-install.ts`, and the command-assembly tests import it instead of duplicating the skills CLI args array.

packages/cli/src/init/skill-install.ts
tensordreams: Fixed in dd9fd3153. Bun runtime detection now lives in `packages/cli/src/init/is-bun.ts`, and both `Init.ts` and `skill-install.ts` import it from that helper.

aqrln (approved):

PR #29690, opened 2026-07-03
feat(cli): offer agent skills install once on prisma generate

tensordreams:
> **Stacked PR** — based on #29689 (S1, the skill-install runner); will be retargeted to `main` once #29689 merges. The diff here is only S2's two commits.

Existing projects (which never ran the new `prisma init`) get exactly one polite, interactive, time-limited offer to install the [prisma/skills](https://github.com/prisma/skills) catalog, shown after a successful `prisma generate`. Second slice (S2) of the Agent-Native Prisma ORM project (#29688).

## Changes

- **Offer module** (`packages/cli/src/utils/skills/skills-offer.ts`): `handleSkillsOffer()` runs the full NPS-style gate chain — prior acknowledgement, already-installed detection, interactive TTY, Deno guard, CI, git hook, npm lifecycle hook, container, and "first Prisma command ≥ 1 day ago" — then asks one yes/no question with a 30-second timeout defaulting to No. Every outcome (`accepted` / `declined` / `timeout` / `already-installed`) persists `skills-offer.json` in the OS config dir, so the offer fires **once ever per machine** (unlike the NPS survey's once-per-timeframe). Accepting runs S1's pinned skill-install runner; failure prints the manual command, non-fatally. The whole body never throws — an offer failure can never fail `generate`.
- **Generate wiring** (`packages/cli/src/Generate.ts`): the offer runs in the existing `!hideHints` non-watch success block, before the NPS survey; if the offer prompted, the survey is skipped for that run — at most one prompt per generate. Injectable handler mirroring the existing `surveyHandler` seam.
- **Shared helper** (`packages/cli/src/utils/prompt-timeout.ts`): the `timeout()` racer extracted verbatim from `nps/survey.ts`; both prompts now share it.
- **Telemetry**: a single `skills_offer_resolved` `{ outcome, cliVersion }` event through the existing PostHog capture path, only when a prompt was actually shown.
- **Tests**: 27 unit tests for the module (every gate short-circuit, input variants, fake-timer timeout, four failure-isolation cases) + 4 Generate-level tests (ordering, mutual exclusion both ways, `--no-hints` and watch suppression). Injection over module mocking throughout.

## Why

- **Once-ever, not once-per-timeframe**: a declined offer should never nag again; `prisma init` (S1) and docs remain the discovery paths.
- **Acknowledgement only after a prompt resolves**: gated-out runs (CI, non-TTY) write nothing, so a developer whose first runs are in CI still gets the offer later on a real TTY.
- **Ordering inside the resolution path** (acknowledge → install → telemetry) makes once-ever survive a crash mid-install and keeps an offline telemetry endpoint from robbing an accepting user of the install.
- **Mutual exclusion with NPS** keeps the post-generate moment to a single question, ever.

## Verification (live pty runs, built CLI, isolated config dirs)

- First interactive run: offer shown; `n` persists `outcome: "declined"`; second run silent.
- Accept run: skills land via the S1 runner (24 SKILL.md across `.claude`/`.windsurf`/`.agents` + `skills-lock.json`); `outcome: "accepted"`.
- Non-TTY run: no prompt, no acknowledgement written, generate output normal.
- In prompted runs the offer was the only prompt (NPS skipped; mechanism also pinned by unit tests).

Full captures: `projects/agent-native/slices/generate-skill-offer/verification.md` on the project branch (#29688).

**Linear:** [TML-2971](https://linear.app/prisma-company/issue/TML-2971/s2-one-time-skill-offer-on-prisma-generate)


packages/cli/src/get-cli-version.ts
```diff
@@ -0,0 +1,14 @@
+import Debug from '@prisma/debug'
+
+const debug = Debug('prisma:cli:get-cli-version')
+
+export function getCliVersion(): string {
+  try {
+    // eval keeps esbuild from inlining the require; the path resolves relative
+    // to src in development and to the built CLI bundle in production.
+    return eval(`require('../package.json')`).version
```
aqrln: There's no need for this because we replace the version in `package.json` before `pnpm build` on CI in release builds. Just `import { version } from '../package.json'` should suffice.

packages/cli/src/utils/prompt-timeout.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    return promise.then(
+      (result) => {
+        clearTimeout(timeoutId)
+        resolve(result)
+      },
+      () => {
+        clearTimeout(timeoutId)
+        resolve(undefined)
+      },
```
aqrln: This doesn't look right, this should reject with the original error.

```ts
return promise.then(resolve).finally(() => clearTimeout(timeoutId));
```

packages/cli/src/utils/skills/skills-offer.ts
```diff
@@ -0,0 +1,232 @@
...
+    // Closes the underlying interface directly: the proxy throws once the
+    // stream is closed, and close must stay callable from the finally block.
+    close: () => rl.close(),
+  }
+}
+
+function fileExists(filePath: string): Promise<boolean> {
+  return fs.promises.access(filePath).then(
+    () => true,
+    () => false,
```
aqrln: Shouldn't this check what error occurred exactly?

packages/cli/src/utils/skills/skills-offer.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  if (await skillsAlreadyInstalled(ctx.cwd)) {
+    await writeAcknowledgement(ackPath, 'already-installed')
+    return { prompted: false }
+  }
+
+  if (
+    !ctx.isInteractive() ||
+    ctx.isDeno ||
```
aqrln: Can you find what was the original reason the survey was skipped on Deno and verify if it's still an issue?

packages/cli/src/get-cli-version.ts
tensordreams: Done — dropped the `eval` wrapper and the `get-cli-version` module entirely, importing `version` from `../package.json` directly at both call sites (matching what `MCP.ts` already does). The `'unknown' fallback went with it, since there is no longer a way to trigger it; `versionsOutOfSync` is back to the shape it had before this PR.

Verified the import bundles and inlines correctly via esbuild — it resolves to `0.0.0` locally, which is exactly the value CI replaces before `pnpm build`.

packages/cli/src/utils/prompt-timeout.ts
tensordreams: Agreed, fixed. Used your shape with the rejection wired through:

```ts
promise.then(resolve, reject).finally(() => clearTimeout(timeoutId))
```

Note `then(resolve, reject)` rather than `then(resolve)` — with the bare form the rejection escapes into the `.finally()` chain, whose result nothing awaits, so it surfaces as an unhandled rejection instead of reaching the caller.

Two call-site consequences:

- **NPS survey**: a prompt failure now propagates to the existing `.catch()` in `handleNpsSurvey`, so the survey aborts instead of silently recording `{}` and writing an acknowledgement for a survey the user never answered. It gets retried next run.
- **Skills offer**: still maps an unreadable answer to "no answer", so the offer is recorded as resolved and `generate` never shows two prompts in one run. That is now an explicit `.catch()` with a comment rather than a side effect of the helper swallowing everything.

Added `packages/cli/src/__tests__/prompt-timeout.vitest.ts` covering value/rejection/timeout plus timer cleanup on both settle paths — a leaked timer would hold the event loop open for the full 30s and delay the command exit.

packages/cli/src/utils/skills/skills-offer.ts
tensordreams: It should, fixed. `fileExists` now returns `false` only for `ENOENT`/`ENOTDIR` and rethrows anything else, so e.g. an `EACCES` on the config dir is no longer read as "no acknowledgement recorded" (which would have re-offered on every run). It follows the same idiom `readConfig` in `survey.ts` already uses.

`ENOTDIR` is in there deliberately: it means a path segment is not a directory, so the file genuinely cannot be there. The `unwritable config dir` test exercises exactly that — it points `configDir` inside a regular file — and it caught me when I first checked only for `ENOENT`.

Applied the same treatment to the adjacent `readdir(...).catch(() => [])` in `skillsAlreadyInstalled`, which had the same issue: a permission error on `.claude/skills` was being read as "no skills installed". Happy to split that back out if you would rather keep it out of scope.

packages/cli/src/utils/skills/skills-offer.ts
tensordreams: Traced it: b7d1639fa / #27133 (May 2025), your own commit. The symptom was that on Deno, merely creating the readline interface kept `prisma generate` alive until Enter was pressed, and the message says the root cause should ideally be fixed before the next survey.

I could not reproduce it on **Deno 2.9.3**. Tested under a real pty with no stdin, comparing against Node 24:

| scenario | Node | Deno 2.9.3 |
|---|---|---|
| create + close | exits | exits |
| create + async work + close | exits | exits |
| full question + 5s timeout + close | exits at 5s | exits at 5s |
| create, never close | hangs | hangs |

The last row hangs on Node too, so it is not Deno-specific — it is just an unclosed interface holding the event loop, and both code paths always close in a `finally`.

The third row is the one that matters, and I ran it against the real `timeout()` implementation (esbuild-bundled from `prompt-timeout.ts`) with `createSafeReadlineProxy` and the `createReadlinePrompt` construction copied verbatim. Deno printed the prompt, timed out, wrote `No response received`, and exited cleanly without Enter — identical to Node.

So the guard is removed from both `skills-offer.ts` and `survey.ts`.

One caveat worth naming: this only proves current Deno is fixed. Anyone on a Deno old enough to still carry the original bug would see the hang return. That seemed acceptable given the fix is upstream, but say the word and I will restore the guard with the finding recorded in the comment instead.

packages/cli/src/utils/prompt-timeout.ts
tensordreams: Update: `prompt-timeout.ts` is gone entirely, so this thread is now moot — flagging it so the earlier reply does not mislead.

Both prompts moved to `@clack/prompts`, whose prompts accept an `AbortSignal`. That is strictly better than racing a promise against a timer: the deadline now *cancels* the prompt and restores the terminal, rather than walking away from a prompt that still holds stdin. The helper and its test are deleted.

The rejection concern you raised still holds and is still handled — a throwing prompt is caught at the skills-offer call site and recorded as a timeout, covered by `a throwing prompt resolves to the timeout outcome`.

One behaviour change worth your eye: an aborted signal distinguishes a lapsed deadline from a user dismissal, so Ctrl+C / Esc is now recorded as `declined` rather than `timeout`. That felt more honest than calling a deliberate dismissal a timeout, but it is a telemetry-visible change, so say the word if you would rather keep them merged.

packages/cli/src/utils/nps/survey.ts
```diff
@@ -118,41 +76,48 @@ export async function handleNpsSurveyImpl(
...
+  const result = await collectFeedback(prompts)
 
   if (result.rating) {
     await submitSurveyEvent({ rating: result.rating, ...result }, eventCapture)
-    rl.write('Thanks for your feedback!\n')
+    prompts.message('Thanks for your feedback!')
   }
 
   await writeConfig({ acknowledgedTimeframe: status.currentTimeframe })
 }
```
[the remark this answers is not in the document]
tensordreams: Good catch, and a real one. Fixed in 4b62468 — the check is now `result.rating !== undefined`.

Worth recording that this bug predates the PR: the same `if (result.rating)` is on `main` today (`survey.ts:123`), and this branch only carried it across when the survey moved off readline. So it is a live bug in the shipped CLI, not a regression introduced here. Fixing it as part of this PR since the surrounding code is already being touched.

The consequence was as you describe and slightly worse in combination: `writeConfig` runs regardless, so a user who rated Prisma 0 had their score dropped *and* the timeframe marked acknowledged, meaning they were never asked again.

packages/cli/src/utils/nps/survey.ts
```diff
@@ -118,41 +76,48 @@ export async function handleNpsSurveyImpl(
...
     return {}
   }
 
-  const feedbackAnswer = await rl.question(
-    'Optional: Provide additional feedback or press Enter to skip.\n' + 'Additional feedback: ',
-  )
-  const feedback = feedbackAnswer.trim() === '' ? undefined : feedbackAnswer
+  const feedbackAnswer = await prompts.text({
+    message: 'Optional: Provide additional feedback, or leave blank to skip.',
+  })
```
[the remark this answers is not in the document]
tensordreams: Agreed, fixed in 4b62468 — the feedback prompt now carries the same `promptTimeoutSecs` deadline as the rating prompt.

This one also predates the PR: on `main` the follow-up is a bare `rl.question(...)` with no timeout wrapper, so the survey could already hang there indefinitely. It is cheap to fix now that both prompts take a deadline.

A rating already given is still submitted when the follow-up lapses — the timeout resolves the feedback to `undefined` rather than discarding the rating. Covered by `should give the feedback prompt a deadline of its own`.

packages/cli/src/__tests__/nps.test.ts
```diff
@@ -300,22 +309,26 @@ describe('nps survey', () => {
...
     expect(mockRead).toHaveBeenCalled()
     expect(mockWrite).toHaveBeenLastCalledWith(
       expect.anything(),
       JSON.stringify({ acknowledgedTimeframe: currentTimeframe }),
     )
     expect(status).toHaveBeenCalledTimes(1)
-    expect(readline.question).toHaveBeenCalledTimes(2)
-    expect(readline.write).toHaveBeenCalledWith('Thanks for your feedback!\n')
+    expect(prompts.text).toHaveBeenCalledTimes(2)
+    expect(prompts.message).toHaveBeenCalledWith('Thanks for your feedback!')
```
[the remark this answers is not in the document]
tensordreams: Added in 4b62468 as `should submit a rating of 0`, which asserts both the captured event (`rating: 0`, with feedback) and the thank-you message.

I checked it actually earns its place rather than just passing: reverting the fix to `if (result.rating)` makes it fail, and restoring the fix makes it pass.

```
✕ should submit a rating of 0     <- with `if (result.rating)`
✓ should submit a rating of 0     <- with `if (result.rating !== undefined)`
```

packages/cli/src/utils/prompt-timeout.ts
```diff
@@ -0,0 +1,16 @@
...
+    const timeoutId = setTimeout(() => {
+      resolve(undefined)
+    }, ms)
+
+    return promise.then((result) => {
+      clearTimeout(timeoutId)
+      resolve(result)
+    })
+  })
+}
```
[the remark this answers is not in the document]
tensordreams: Resolved, though by a different route than suggested: `prompt-timeout.ts` no longer exists.

The rejection gap you identified was real and was fixed first (rejecting with the original error rather than swallowing it). Both prompts have since moved to `@clack/prompts`, whose prompts accept an `AbortSignal`, so the deadline now cancels the prompt and restores the terminal instead of racing a promise against a timer and leaving the prompt holding stdin. The helper and its test were deleted as superseded.

packages/cli/src/utils/skills/skills-offer.ts
```diff
@@ -0,0 +1,238 @@
...
+    if (answer === undefined) {
+      prompt.write(`No response received within ${promptTimeoutSecs} seconds.\n`)
+      return 'timeout'
+    }
+
+    return /^y(es)?$/i.test(answer.trim()) ? 'accepted' : 'declined'
+  } finally {
+    prompt.close()
+  }
+}
```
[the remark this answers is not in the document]
tensordreams: Resolved. As you predicted, fixing the shared helper covered this call site — and it is now moot in any case, since `timeout()` is gone and the deadline is an `AbortSignal` passed to the prompt itself.

The underlying concern is still handled explicitly: a throwing prompt is caught in `promptForInstall` and recorded as a timeout, so `handleSkillsOffer` still reports `prompted: true` and the NPS survey cannot fire in the same run. Covered by `a throwing prompt resolves to the timeout outcome`.

packages/cli/src/utils/skills/skills-offer.ts
```diff
@@ -0,0 +1,238 @@
...
+    await submitOfferEvent(outcome, ctx)
+  } catch (err) {
+    // A telemetry failure after the prompt was shown must not report the run
+    // as not prompted, or the caller would show a second prompt via the NPS
+    // survey.
+    debug(`Failed to submit the skills offer telemetry event: ${err}`)
+  }
+
+  return { prompted: true }
+}
```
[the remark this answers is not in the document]
tensordreams: Fixed in bbd64a2 / 275ec9f — the `ctx.installSkills` call has its own try/catch inside `handleSkillsOfferImpl`, logs the failure, and falls through to `submitOfferEvent`, so the run still reports `prompted: true`.

Your reasoning about the consequence was correct and is what the test now pins: `a throwing skill install still reports the prompt as shown` asserts `{ prompted: true }`, the `accepted` acknowledgement, and that the telemetry event is still submitted — which together prevent the NPS survey from firing as a second prompt in the same run.

packages/cli/src/utils/skills/skills-offer.ts
```diff
@@ -0,0 +1,238 @@
...
+function getCliVersion(): string {
+  try {
+    // eval keeps esbuild from inlining the require; the path resolves
+    // relative to the built CLI bundle, next to which package.json lives.
+    return eval(`require('../package.json')`).version
+  } catch (err) {
+    debug(`Failed to read the CLI version from package.json: ${err}`)
+    return 'unknown'
+  }
+}
```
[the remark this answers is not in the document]
tensordreams: Resolved in 6b2d081, though by removing the duplication rather than centralising it.

Release builds replace the version in `package.json` before `pnpm build`, so the `eval(require(...))` workaround was unnecessary in the first place. Both call sites now use a plain `import { version } from "../package.json"`, matching what `MCP.ts` already did, and `get-cli-version.ts` is deleted along with its `unknown` fallback — which no longer has any way to trigger.

PR #29700, opened 2026-07-08
fix(client): throw on invalid Date in raw query parameters instead of binding "null"

### LINKED ISSUE #29696: Invalid Date passed to $queryRaw/$executeRaw is silently serialized as the string "null" instead of throwing
### Description
Raw-parameter serialization tags dates via `parameter.toJSON()`, which returns `null` for an invalid `Date` (spec behavior). So an invalid `Date` becomes `{ prisma__type: 'date', prisma__value: null }`; on deserialize it becomes the literal string `"null"` typed as `datetime`, silently corrupting the value — whereas the structured query builder throws a clear `"Provided Date object is invalid"`.

### Steps to reproduce
```js
serializeRawParameters([ new Date('not a date'), new Date('2024-01-15T00:00:00Z') ])
// -> [{"prisma__type":"date","prisma__value":null}, {"prisma__type":"date","prisma__value":"2024-01-15T00:00:00.000Z"}]
// after deserialize -> args ["null", "2024-01-15T00:00:00.000Z"], both typed datetime
```
Observed: the invalid date is coerced to the string `"null"` and bound as a datetime. Expected: a validation error, as the non-raw path throws.

### Root cause
`packages/client/src/runtime/utils/serializeRawParameters.ts` uses `prisma__value: parameter.toJSON()` in the `isDate` branch; `deserializeRawParameters.ts` then produces the literal `"null"`.

### Suggested fix
In the `isDate` branch, validate with the existing `isValidDate` helper and throw a clear client error for invalid dates instead of emitting `toJSON()`.

thribhuvan003:
Fixes #29696

Passing an invalid Date to $queryRaw/$executeRaw currently does not fail — it quietly gets serialized as the string "null" and bound as a datetime, which can corrupt the value that ends up in the database. The regular query path already throws "Provided Date object is invalid" for the same input, so this makes the raw path throw the same error using the existing isValidDate helper. Added two unit tests, one for a top-level invalid date and one nested in an array.

### DIFF
--- a/packages/client/src/__tests__/serializeRawParameters.test.ts
+++ b/packages/client/src/__tests__/serializeRawParameters.test.ts
@@ -27,6 +27,18 @@ describe('serializeRawParameters', () => {
     ])
   })
 
+  test('invalid date', () => {
+    const data = [new Date('not a date')]
+
+    expect(() => serialize(data)).toThrow('Provided Date object is invalid')
+  })
+
+  test('invalid date nested in array', () => {
+    const data = [[new Date('invalid')]]
+
+    expect(() => serialize(data)).toThrow('Provided Date object is invalid')
+  })
+
   test('BigInt', () => {
     const data = [BigInt('321804719213721')]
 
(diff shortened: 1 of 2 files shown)


thribhuvan003:
looks like #29697 landed the same fix (reject invalid raw dates) — closing this in favor of that. thanks!

PR #29701, opened 2026-07-08
fix(client): copy Bytes values out of Node's shared Buffer pool on deserialization

### LINKED ISSUE #29694: Bytes fields deserialize to a Uint8Array backed by Node's shared Buffer pool (wrong .buffer, cross-value exposure)
### Description
Every `Bytes` value returned by Prisma Client is decoded as a **view into Node's shared 8 KB `Buffer` pool** rather than a standalone buffer. For payloads ≤ 4096 bytes (the common case — hashes, tokens, keys, small blobs), `Buffer.from(base64)` allocates from the process-global pool, and the returned `Uint8Array` is a *view* into that pool. So `field.buffer` is the 8192-byte pool — which contains **other decoded `Bytes` values from other columns/rows/queries** — and `field.buffer.byteLength` (8192) does not equal the value's length.

Two call sites:
- `packages/client-engine-runtime/src/json-protocol.ts:157-158` — structured query results (runs for every query on every engine).
- `packages/client/src/runtime/utils/deserializeRawResults.ts:20-21` — `$queryRaw` Bytes columns.

```ts
const { buffer, byteOffset, byteLength } = Buffer.from(value, 'base64')
return new Uint8Array(buffer, byteOffset, byteLength)
```

### Steps to reproduce
```js
function decodeBytes(v) {
  const { buffer, byteOffset, byteLength } = Buffer.from(v, 'base64')
  return new Uint8Array(buffer, byteOffset, byteLength)
}
const a = decodeBytes(Buffer.from([1, 2, 3]).toString('base64'))       // value from row/column A
const b = decodeBytes(Buffer.from([9, 9, 9, 9]).toString('base64'))    // value from row/column B

console.log(a.length, a.buffer.byteLength)   // 3 8192
console.log(a.buffer === b.buffer)           // true  — both alias the same pool
console.log(Array.from(new Uint8Array(a.buffer).slice(a.byteOffset, a.byteOffset + 16)))
// [1,2,3, 0,0,0,0,0, 9,9,9,9, 0,0,0,0]  — B's bytes are visible through A.buffer
```

### Expected
Each `Bytes` field is a standalone `Uint8Array` whose `.buffer.byteLength === .length` and which does not expose unrelated values — e.g. `new Uint8Array(user.avatar.buffer)` should equal `user.avatar`.

### Actual
`.buffer` is the 8192-byte shared pool, and distinct `Bytes` values alias the same `ArrayBuffer`. Any code that reads `.buffer` — `new Uint8Array(x.buffer)`, `new Blob([x.buffer])`, `crypto.subtle.digest(algo, x.buffer)`, structured-clone / `postMessage`, or sending `.buffer` over the wire — gets oversized/garbage data, and one row's `.buffer` contains other rows'/queries' decoded `Bytes` within the same process.

### Suggested fix
Return an owned copy at both sites, e.g.:
```ts
return Uint8Array.prototype.slice.call(Buffer.from(value, 'base64'))
// or:
const b = Buffer.from(value, 'base64')
return new Uint8Array(b.buffer.slice(b.byteOffset, b.byteOffset + b.byteLength))
```

thribhuvan003:
Fixes #29694

Bytes values come back as a Uint8Array that is a view into Node's shared 8 KB Buffer pool, so field.buffer is the whole pool — including bytes from other decoded values — and field.buffer.byteLength is 8192 instead of the value's real length. This changes both decode sites (structured results and $queryRaw) to copy into a standalone buffer instead of keeping the pool view. Added a test in each package asserting the returned array owns its buffer.

### DIFF
--- a/packages/client-engine-runtime/src/json-protocol.test.ts
+++ b/packages/client-engine-runtime/src/json-protocol.test.ts
@@ -28,6 +28,14 @@ describe('deserializeJsonObject', () => {
     expect(value).toEqual(new Uint8Array(Buffer.from('hello world')))
   })
 
+  test('Bytes is backed by a standalone buffer', () => {
+    const value = deserializeJsonObject({ $type: 'Bytes', value: 'aGVsbG8gd29ybGQ=' }) as Uint8Array
+    // must not be a view into Node's shared Buffer pool, which would expose
+    // unrelated pool data through `value.buffer`
+    expect(value.byteOffset).toBe(0)
+    expect(value.buffer.byteLength).toBe(value.byteLength)
+  })
+
   test('Decimal', () => {
     const value = deserializeJsonObject({ $type: 'Decimal', value: '123.45' })
     expect(value).toBeInstanceOf(Decimal)
--- a/packages/client-engine-runtime/src/json-protocol.ts
+++ b/packages/client-engine-runtime/src/json-protocol.ts
@@ -153,10 +153,9 @@ function deserializeTaggedValue({ $type, value }: JsonInputTaggedValue | JsonOut
   switch ($type) {
     case 'BigInt':
       return BigInt(value)
-    case 'Bytes': {
-      const { buffer, byteOffset, byteLength } = Buffer.from(value, 'base64')
-      return new Uint8Array(buffer, byteOffset, byteLength)
-    }
+    case 'Bytes':
+      // copy instead of returning a view into Node's shared Buffer pool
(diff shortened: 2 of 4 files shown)


aqrln (approved):
LGTM. This is unfortunate that we have to do this but safety and correctness come first.

If this causes measurable performance issues for large buffers, we could change this to only clone the buffers up to 4 KB which are actually pooled.

PR #29703, opened 2026-07-08
fix(client,schema-files-loader): fix source slice and schema loader edge cases

ShiroKSH:
## Summary

- fix SourceFileSlice range checks so line transforms stop at the actual last line
- make slicing an existing SourceFileSlice use line numbers relative to that slice
- make schema file resolver case folding locale-independent

## Why

SourceFileSlice keeps original source line numbers after slicing. Two helpers still mixed original line numbers with array indexes, which allowed one line past the end in mapLineAt and produced incorrect results when slicing an already sliced source.

The schema files loader also used locale-sensitive lowercasing for case-insensitive file keys. That can produce unstable keys in locales with special casing rules, such as Turkish.

## Tests

- corepack pnpm --dir packages/client test -- src/runtime/utils/SourceFileSlice.test.ts
- corepack pnpm --dir packages/schema-files-loader exec vitest run src/resolver/caseSensitivity.test.ts src/resolver/InMemoryFilesResolver.test.ts src/resolver/CompositeFilesResolver.test.ts
- corepack pnpm exec prettier --check packages/client/src/runtime/utils/SourceFileSlice.ts packages/client/src/runtime/utils/SourceFileSlice.test.ts packages/schema-files-loader/src/resolver/caseSensitivity.ts packages/schema-files-loader/src/resolver/caseSensitivity.test.ts
- git diff --check

### DIFF
--- a/packages/client/src/runtime/utils/SourceFileSlice.test.ts
+++ b/packages/client/src/runtime/utils/SourceFileSlice.test.ts
@@ -27,6 +27,11 @@ test('sliced toString', () => {
   expect(slice.toString()).toBe('2\n3')
 })
 
+test('sliced toString after slice', () => {
+  const slice = SourceFileSlice.fromContent('1\n2\n3\n4\n5').slice(2, 4).slice(3, 4)
+  expect(slice.toString()).toBe('3\n4')
+})
+
 test('lineAt', () => {
   const slice = SourceFileSlice.fromContent('1\n2\n3')
   expect(slice.lineAt(2)).toBe('2')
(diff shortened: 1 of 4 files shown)


EvilDeaD306 (changes requested):
T

ShiroKSH:
@aqrln - could you take a look at this client and schema-loader fix, or route it to the right reviewer? The slice-bounds suggestion was addressed in 029ab40, and the inline thread is resolved. I can make any further changes or add tests. Thanks!

aqrln (approved):
Thank you!

PR #29708, opened 2026-07-09
fix(client): resolve esbuild "Dynamic require of node:path" for the ESM runtime

Develop-KIM:
Fixes #28126.

## Problem

With the `prisma-client` generator (`engineType = "client"`), bundling an app with esbuild for **ESM** on Node crashes at runtime:

```
Error: Dynamic require of "node:path" is not supported
    at node_modules/@prisma/client/runtime/client.js
```

## Cause

The `./runtime/client` export in `packages/client/package.json` has a `node` condition that lists only `require`/`default`, both pointing at the CJS `client.js`:

```json
"node": {
  "require": "./runtime/client.js",
  "default": "./runtime/client.js"
},
```

Export conditions match top-to-bottom. esbuild resolves an `import`-syntax dependency with the `import` condition regardless of its output format, so on Node it enters `node`, finds no `import`, and falls through to `node.default` = `client.js` (CJS). Bundling that CJS module into ESM wraps its top-level `require("node:path")` in esbuild's `__require` shim, which throws at import time.

The `node` override was added in #27610 to fix the opposite case (esbuild bundling to **CJS** was picking the `.mjs` and breaking on `import.meta.url`). That fixed CJS and left ESM without an `import` target.

## Fix

1. Add `import` → `client.mjs` to the `node` condition, so ESM consumers (esbuild ESM and native `import`) get the ESM build instead of falling through to CJS.
2. esbuild resolves both its ESM and CJS output with the `import` condition, so (1) also routes esbuild-CJS to the `.mjs`. Guard the ESM banner's `import.meta.url` (which esbuild leaves `undefined` when it emits the `.mjs` as CommonJS) so the esbuild-to-CJS case #27610 fixed keeps working.

`node.default` stays on the CJS build, so `require()` is unchanged.

## Verification

Built `@prisma/client` and bundled a consumer of `@prisma/client/runtime/client` against the real artifacts, in every resolver:

| resolver | before | after |
| --- | --- | --- |
| esbuild ESM + `platform: node` | ❌ `Dynamic require of "node:path"` | ✅ loads |
| esbuild CJS + `platform: node` (the #27610 case) | ✅ | ✅ |
| native Node `import` | ❌ | ✅ |
| native Node `require` | ✅ (`client.js`) | ✅ (`client.js`) |

I confirmed all four end-to-end with the built runtime, but couldn't run the Dockerized `tests/e2e` harness in my environment. I'd be glad to add an e2e under `tests/e2e/` that bundles a generated client with `esbuild --format=esm --platform=node` and runs it (mirroring `bundler-detection-error`) — happy to include it here if you tell me which provider flavor you'd prefer.

---

Written with AI assistance (disclosed via the commit's `Co-Authored-By` trailer); I've reviewed and verified every line.


jenbonilla-create (approved):

tensordreams:
**Correction to my earlier comment on this PR.**

An earlier comment here claimed the `import.meta.url` banner guard in `helpers/build.ts` was dropped "per the maintainer decision." That was wrong on two counts, and I want to set it straight:

1. **There was no maintainer decision** — that framing was mistaken, and I should not have stated it.
2. **Dropping the guard was the wrong call.** As this PR's description already explains, the two halves are coupled: adding `node.import → ./runtime/client.mjs` also makes a consumer's esbuild-CJS bundle resolve the runtime to `client.mjs`, where esbuild leaves `import.meta.url` undefined. Without the guard, the banner's `fileURLToPath(import.meta.url)` throws at import time — reintroducing exactly the esbuild-to-CJS crash that #27610 fixed.

I've restored the banner guard (commit `41ca6838a`), so both halves of the fix now ship together as you originally designed them. Apologies for the incorrect earlier comment and for the churn.

packages/client/helpers/build.ts
```diff
@@ -35,9 +35,13 @@ import * as __banner_node_module from "node:module";
...
 import * as __banner_node_url from "node:url";
-const __filename = __banner_node_url.fileURLToPath(import.meta.url);
+// \`import.meta.url\` is only defined when this ESM build is loaded as ESM. A bundler
+// that emits this file as CommonJS (e.g. esbuild \`format: 'cjs'\`) leaves it undefined,
+// so fall back to the entrypoint path to avoid crashing at import time.
+const __banner_esm_url = import.meta.url || __banner_node_url.pathToFileURL(process.argv[1] ?? process.cwd()).href;
+const __filename = __banner_node_url.fileURLToPath(__banner_esm_url);
 globalThis['__dirname'] = __banner_node_path.dirname(__filename);
-const require = __banner_node_module.createRequire(import.meta.url);`
+const require = __banner_node_module.createRequire(__banner_esm_url);`
```
aqrln: Can't we skip creating `require` completely and instead use existing `require` if `import.meta.url` is undefined (assuming one implies the other)?

Develop-KIM: Yes — that's the better shape, done in 364247f.

Your assumption holds: `import.meta.url` is only undefined here because a bundler re-emitted this `.mjs` as CommonJS, and that wrapper already gives us the real `require`/`__filename`/`__dirname`. So the banner now only synthesizes them in the ESM branch (`if (import.meta.url)`) and lets the ambient CJS ones through otherwise — no `createRequire`, and the `process.argv[1] ?? process.cwd()` fallback is gone.

That also settles the anchoring point CodeRabbit raised: the old fallback anchored `require` to the entrypoint, so when the bundle was loaded as a non-entry module `require()` resolved from the wrong directory. I checked it against a standalone esbuild build across native `import`, esbuild ESM, esbuild CJS as the entry, and esbuild CJS required from another module — the last one is exactly where the old fallback failed (`MODULE_NOT_FOUND`) and the ambient version resolves correctly.

aqrln: Thank you, Claude 😂

This looks good, I've updated the branch and will merge it if the CI is green.

aqrln:
@Develop-KIM `packages/client/tests/e2e/nextjs-schema-not-found/10_monorepo-serverComponents-customOutput-noReExport` e2e test consistently fails with these changes (restarted twice)

Develop-KIM:
Thanks for running it down. I think I see the mechanism: the `node.import` → `runtime/client.mjs` entry I added to `packages/client/package.json` doesn't only reach esbuild — Next.js resolves the runtime with the `import` condition too, so in the `output: 'standalone'` + server-components build it now pulls the ESM `client.mjs` where it previously fell through to `node.default` (`client.js`, CJS). These `nextjs-schema-not-found` tests assert exactly that path: with the workaround the query must return 200, and without it Prisma must surface the friendly `engine-not-found-nextjs` message rather than a raw crash. Swapping the bundled runtime from CJS to ESM there is very likely what shifts the outcome.

Could you share which half fails and the actual stderr — the `WORKAROUND=true` 200 case, or the `WORKAROUND=false` friendly-error case? That tells me whether the engine-not-found detection stopped firing or the working path itself broke.

The tension is that export conditions can't tell esbuild and Next.js apart — both are `node` + `import` — so scoping the `.mjs` to esbuild alone isn't really possible through conditions. If it's the friendly-error path that regressed, the fix likely has to keep Next.js on the CJS runtime while still getting esbuild-ESM off the dynamic-`require` crash, which probably means a different approach than the `node.import` entry (e.g. addressing the `__require` shim in the CJS build directly). Happy to explore that direction — just let me know what the log shows.

Develop-KIM:
I'm closing this as I'm stepping back from open-source contributions. One heads-up: I'm also deleting my fork, so this branch and the commits you pushed to it will go away with it. If you want to keep 364247f or anything else on the branch, please grab it first. Sorry for the churn, and thanks for the time you put in.

Develop-KIM:
@aqrln One correction to what I said above: your commits weren't lost. GitHub keeps `refs/pull/29708/head` after a fork is deleted, so 364247f and your merges are all still there — I pulled the branch back from it.

This is reopened as #29915 with the exact same 9 commits (head 79593e5). Sorry for the noise.

PR #29712, opened 2026-07-10
fix(client): infer moduleFormat correctly for nodenext/node16 in CJS projects

### LINKED ISSUE #29710: prisma-client generator infers moduleFormat esm for CommonJS projects using module nodenext
### Bug description

The `prisma-client` generator decides the default `moduleFormat` by looking only at `compilerOptions.module` in tsconfig.json: anything that isn't literally `"commonjs"` gets treated as ESM. But `"module": "nodenext"` doesn't mean the project is ESM. With nodenext, TypeScript decides the format per file from the nearest package.json `type` field, and when there's no `"type": "module"` everything compiles as CommonJS.

So in a CJS project using nodenext (a pretty common setup for NestJS backends), the generated client contains `import.meta.url`, tsc compiles it as CommonJS, and the compiled output blows up the moment it gets required:

```
ReferenceError: exports is not defined in ES module scope
    at .../src/prisma/client.ts:48:23
```

Same crash as #27556, which was closed with the `moduleFormat = "cjs"` workaround. The workaround works, but the default is still wrong.

What makes it nasty is that the generated files start with `// @ts-nocheck`, which suppresses the TS1470 error (`'import.meta' is not allowed in a CommonJS file`) that would otherwise catch the mismatch at compile time. So there is zero signal until it crashes at runtime, and the stack trace points at a generated file with a sourcemapped position in the middle of a doc comment, which makes it fun to debug.

### How to reproduce

1. package.json **without** `"type": "module"`
2. tsconfig.json with `"module": "nodenext"`, `"moduleResolution": "nodenext"`
3. generator block without an explicit moduleFormat:

```prisma
generator client {
  provider = "prisma-client"
  output   = "../src/prisma"
}
```

4. `prisma generate`, compile with tsc, `require()` the generated client → crash above.

### Expected behavior

When `module` is `nodenext`/`node16`, infer the way tsc does: look for `"type": "module"` in the nearest package.json and default to `"cjs"` when it's absent. Failing that, refusing to generate and asking for an explicit `moduleFormat` would also be fine. Anything but silently emitting ESM-only code into a CommonJS compilation.

### Prisma information

```prisma
datasource db {
  provider = "sqlite"
}

generator client {
  provider = "prisma-client"
  output   = "../../src/prisma"
}
```

### Environment & setup

- OS: Windows 11 / ubuntu-latest (GitHub Actions), reproduced on both
- Database: SQLite (`@prisma/adapter-better-sqlite3`)
- Node.js: 24.x
- TypeScript: 7.0.2

### Prisma Version

7.7.0 (checked the inference logic in the 7.8.0 bundle as well, same behavior)

ShauryaaSharma:
## Summary

- The `prisma-client` generator inferred `moduleFormat` by checking whether `tsconfig.json`'s `compilerOptions.module` was literally `"commonjs"`, treating anything else (including `"nodenext"`/`"node16"`) as ESM.
- Under `nodenext`/`node16`, TypeScript actually decides CJS vs ESM per file based on the nearest `package.json`'s `"type"` field, not the `module` setting alone. A CJS project (no `"type": "module"`) using `nodenext` therefore got an ESM-only client containing `import.meta.url`, which crashes at runtime once `require()`'d (`ReferenceError: exports is not defined in ES module scope`).
- This PR makes `inferModuleFormat` mirror `tsc`'s own resolution for `nodenext`/`node16`: look up the nearest `package.json` to the generator's output directory and infer `"esm"` only when its `"type"` field is `"module"`, defaulting to `"cjs"` otherwise (including when no `package.json` is found). Other `module` settings (`commonjs`, `esnext`, etc.) are unaffected.

Closes #29710

## Test plan

- [x] Added `packages/client-generator-ts/tests/module-format.test.ts` — unit tests covering `commonjs`/`esnext` unconditional behavior, `node16`/`nodenext` (with casing variants) against `package.json` with `type: module`, `type: commonjs`, no `type` field, and no `package.json` found, plus the pre-existing extension-based fallback.
- [x] `pnpm run build` in `packages/client-generator-ts` — clean, no type errors.
- [x] Full package test suite (50 tests, including `generator.test.ts`) passes with no regressions.
- [x] Manually reproduced the exact bug report: CJS project (no `"type"` field) + `tsconfig.json` with `module: nodenext` + no explicit `moduleFormat` → generated client now has no `import.meta.url`, compiles cleanly with `tsc`, and runs under `node` without crashing.
- [x] Verified no regression for genuine ESM projects: `"type": "module"` + `module: nodenext` → generated client still correctly gets `import.meta.url` and runs under Node's ESM loader.

### DIFF
--- a/packages/client-generator-ts/src/generator.ts
+++ b/packages/client-generator-ts/src/generator.ts
@@ -66,6 +66,7 @@ export class PrismaClientTsGenerator implements Generator {
             tsconfig,
             generatedFileExtension,
             importFileExtension,
+            outputDir,
           })
 
     await generateClient({
--- a/packages/client-generator-ts/src/module-format.ts
+++ b/packages/client-generator-ts/src/module-format.ts
@@ -1,4 +1,7 @@
+import fs from 'node:fs'
+
 import { TsConfigJson, TsConfigJsonResolved } from 'get-tsconfig'
+import { packageUpSync } from 'package-up'
 
 import { GeneratedFileExtension } from './file-extensions'
 
@@ -30,15 +33,17 @@ type InferModuleFormatOptions = {
   tsconfig: TsConfigJsonResolved | undefined
   generatedFileExtension: GeneratedFileExtension
   importFileExtension: GeneratedFileExtension
+  outputDir: string
 }
 
 export function inferModuleFormat({
   tsconfig,
   generatedFileExtension,
   importFileExtension,
+  outputDir,
 }: InferModuleFormatOptions): ModuleFormat {
   if (tsconfig?.compilerOptions?.module) {
-    return fromTsConfigModule(tsconfig.compilerOptions.module)
+    return fromTsConfigModule(tsconfig.compilerOptions.module, outputDir)
   }
 
   if (generatedFileExtension === 'cts' || importFileExtension === 'cjs') {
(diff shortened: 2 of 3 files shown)


aqrln (approved):
Perfect, thank you!

PR #29717, opened 2026-07-12
fix(adapter-pg): map PostgreSQL deadlocks to P2034

### LINKED ISSUE #29716: @prisma/adapter-pg does not map PostgreSQL deadlock SQLSTATE 40P01 to P2034
# Bug description

PostgreSQL uses SQLSTATE `40P01` for `deadlock_detected`. `@prisma/adapter-pg`
maps `40001` to `TransactionWriteConflict`, but does not map `40P01`.

In an isolated PostgreSQL 16 reproduction using `@prisma/adapter-pg`,
`@prisma/client`, and Prisma CLI 6.19.2, an ordinary model mutation whose
database trigger deterministically raises `40P01` does not surface
`PrismaClientKnownRequestError` with code `P2034`. The same behavior occurs
inside an interactive transaction.

Without `P2034`, callers cannot safely apply application retry logic keyed to
Prisma's structured deadlock code without parsing provider text or retrying
unrelated failures.

Severity: Major. This breaks the structured retry contract for PostgreSQL
deadlocks.

## Reproduction

Self-contained reproduction:
https://gist.github.com/neliro/b659bb6d1bf388422129e36a9f54b213

The reproduction uses a disposable loopback-only PostgreSQL 16 database and a
trigger to raise the exact SQLSTATE deterministically. On the exercised path,
output is limited to the error classification, attempt count, outcome, and row
count; all created database objects are removed.

## Expected vs. actual behavior

Expected:

```json
{
  "clientVersion": "6.19.2",
  "directClassification": "known:P2034",
  "transactionClassification": "known:P2034",
  "attempts": 2,
  "outcome": "inserted",
  "rowCount": 1
}
```

Actual with the unmodified 6.19.2 adapter:

```json
{
  "clientVersion": "6.19.2",
  "directClassification": "unknown_request",
  "transactionClassification": "unknown_request",
  "attempts": 1,
  "outcome": "not_retryable",
  "rowCount": 0
}
```

At Prisma main commit
[`cda80a4488b7b551c36bf09ca2e8303ef9509da4`](https://github.com/prisma/prisma/commit/cda80a4488b7b551c36bf09ca2e8303ef9509da4),
[`packages/adapter-pg/src/errors.ts`](https://github.com/prisma/prisma/blob/cda80a4488b7b551c36bf09ca2e8303ef9509da4/packages/adapter-pg/src/errors.ts#L128-L133)
maps `40001` to `TransactionWriteConflict`, while `40P01` falls through to the
generic PostgreSQL error path. Prisma translates `TransactionWriteConflict` to
`P2034`, whose message covers write conflicts and deadlocks.

The minimal source change is to give `40P01` the same mapping as `40001` and
extend the existing adapter unit test to cover both codes.

## Frequency

Consistently reproducible for PostgreSQL `40P01` through `@prisma/adapter-pg`.

## Development or production

Both. This is a Prisma Client runtime error-classification path.

## Regression

Unknown. I have not confirmed a previously working adapter version.

## Workaround

There is no safe application-level workaround. Parsing provider messages or
retrying every generic failure would broaden retries beyond deadlocks. A local
package patch that maps `40P01` to `TransactionWriteConflict` restores `P2034`.

## Prisma schema and query

The complete schema and executable query are in the reproduction gist. The
relevant operation is an ordinary model mutation inside an interactive
transaction:

```ts
await prisma.$transaction((tx) =>
  tx.probe.create({ data: { value: 'transaction' } }),
)
```

The trigger raises `40P01` only on the first attempt so a correctly classified
retry commits exactly one row.

## Prisma config

Not used. The Prisma 6 reproduction uses the datasource in `schema.prisma` and
passes the connection string to `PrismaPg`.

## Environment and setup

- OS: macOS
- Database: PostgreSQL 16
- Node.js: 25.7.0
- Prisma CLI: 6.19.2
- Prisma Client: 6.19.2
- `@prisma/adapter-pg`: 6.19.2

No credentials, provider logs, or raw provider messages are included.

neliro:
## Summary

- map PostgreSQL `40P01` (`deadlock_detected`) to `TransactionWriteConflict` in
  `@prisma/adapter-pg`
- extend the existing mapping test to cover both `40001` and `40P01`

PostgreSQL categorizes both codes as transaction-rollback conditions. Prisma
translates `TransactionWriteConflict` to `P2034`, its structured retry signal
for write conflicts and deadlocks. Before this change, `40P01` fell through to
the generic PostgreSQL error path.

Fixes #29716

## Verification

- `pnpm --filter @prisma/adapter-pg test src/__tests__/errors.test.ts -t 40P01`
- `pnpm -r --filter @prisma/adapter-pg... build`
- `pnpm --filter @prisma/adapter-pg test` (48 tests passed)

### DIFF
--- a/packages/adapter-pg/src/errors.ts
+++ b/packages/adapter-pg/src/errors.ts
@@ -142,6 +142,7 @@ function mapDriverError(error: DatabaseError): MappedError {
         user: error.message.split(' ').pop()?.split('"').at(1),
       }
     case '40001':
+    case '40P01':
       return {
         kind: 'TransactionWriteConflict',
       }
--- a/packages/adapter-neon/src/__tests__/errors.test.ts
+++ b/packages/adapter-neon/src/__tests__/errors.test.ts
@@ -1,0 +1,14 @@
+import { describe, expect, it } from 'vitest'
+
+import { convertDriverError } from '../errors'
+
+describe('convertDriverError', () => {
+  it.each(['40001', '40P01'])('should handle TransactionWriteConflict (%s)', (code) => {
+    const error = { code, message: 'msg', severity: 'ERROR' }
+    expect(convertDriverError(error)).toEqual({
+      kind: 'TransactionWriteConflict',
+      originalCode: error.code,
+      originalMessage: error.message,
+    })
+  })
+})
--- a/packages/adapter-neon/src/errors.ts
+++ b/packages/adapter-neon/src/errors.ts
@@ -98,6 +98,7 @@ function mapDriverError(error: DatabaseError): MappedError {
         user: error.message.split(' ').pop()?.split('"').at(1),
       }
     case '40001':
+    case '40P01':
       return {
         kind: 'TransactionWriteConflict',
       }
--- a/packages/adapter-pg/src/__tests__/errors.test.ts
+++ b/packages/adapter-pg/src/__tests__/errors.test.ts
@@ -127,8 +127,8 @@ describe('convertDriverError', () => {
     })
   })
 
-  it('should handle TransactionWriteConflict (40001)', () => {
-    const error = { code: '40001', message: 'msg', severity: 'ERROR' }
+  it.each(['40001', '40P01'])('should handle TransactionWriteConflict (%s)', (code) => {
+    const error = { code, message: 'msg', severity: 'ERROR' }
     expect(convertDriverError(error)).toEqual({
       kind: 'TransactionWriteConflict',
       originalCode: error.code,
--- a/packages/adapter-ppg/src/errors.test.ts
+++ b/packages/adapter-ppg/src/errors.test.ts
@@ -23,4 +23,13 @@ describe('convertDriverError', () => {
       originalMessage: error.message,
     })
   })
+
+  it.each(['40001', '40P01'])('should handle TransactionWriteConflict (%s)', (code) => {
+    const error = { code, message: 'msg', details: { severity: 'ERROR' } }
+    expect(convertDriverError(error)).toEqual({
+      kind: 'TransactionWriteConflict',
+      originalCode: error.code,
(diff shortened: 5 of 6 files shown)


packages/adapter-pg/src/errors.ts
```diff
@@ -128,6 +128,7 @@ function mapDriverError(error: DatabaseError): MappedError {
         user: error.message.split(' ').pop()?.split('"').at(1),
       }
     case '40001':
+    case '40P01':
```
aqrln, line 145: neon and ppg adapters should also have this

tensordreams: Done in 9a40b49 — both adapters now map `40P01` alongside `40001` to `TransactionWriteConflict`, in the same switch position as adapter-pg.

Neither adapter had error-mapping tests before, so each gains the same two-code `it.each` case adapter-pg extended here: `adapter-neon/src/__tests__/errors.test.ts` and `adapter-ppg/src/errors.test.ts` (placed to match each package's local test layout). Note the fixtures differ deliberately — ppg's `isDriverError` expects `severity` nested under `details`, unlike pg/neon.

All three adapters' suites pass: pg 34, ppg 27, neon 6.

tensordreams:
Hi @neliro — thanks for this contribution, and sorry for the slow response.

We're doing a cleanup pass to land what we can before the current codebase moves to a `v7` branch, and this PR is one we'd like to include — CI is fully green, so the one thing blocking it on our side is the CLA, which still shows as unsigned:

https://cla-assistant.io/prisma/prisma?pullRequest=29717

neliro:
Thanks for the follow-up — I’ve signed the CLA, and CLA Assistant now shows all committers as signed. I’m still interested in moving this forward. Please let me know if anything else is needed.

PR #29723, opened 2026-07-14
fix(client): mask each quoted string in maskQuery independently

### LINKED ISSUE #17931: Query sanitization function for panic reports almost always sanitizes it down to `{ "X": {} }`
[`maskQuery`](https://github.com/prisma/prisma/blob/main/packages/engine-core/src/common/errors/utils/maskQuery.ts) function attempts to mask users private data from a GraphQL query that is posted to github issue. It has two problems:

1. First regexp for sanitizing the quotes is overly greedy and replaces all the content between the first and the last quote in the string. 
2. Query data is often JSON string, not plain GraphQL query, so even if regexp is corrected it will replace all the keys in the query as well, not only values.

We have two ways to go about it:
1. Fix the thing.
2. Remove "Engine Query" section from panic report. One thing in favor of this is: GraphQL/JSON query still discloses some information about users schema and we should not really include it into issue template without users approval.

santichausis:
## Description

`maskQuery` sanitizes string values in the engine query before it's included in panic/error reports. The regex used to strip strings (`/".*"/g`) is greedy, so on a line with more than one quoted string it matches from the first quote to the last quote and collapses everything in between into a single masked value.

For example:
```
name_in: ["hans", "peter", "schmidt"]
```
was masked to:
```
name_in: ["X"]
```
instead of:
```
name_in: ["X", "X", "X"]
```

This changes the regex to non-greedy (`/".*?"/g`) so each quoted string is matched and masked independently, and updates the existing snapshot test to reflect the corrected output.

Closes #17931

## Test plan

- Updated the `maskQuery.test.ts` snapshot; both existing test cases pass with the fix.
- `pnpm run test src/runtime/core/engines/common/utils/maskQuery.test.ts` in `packages/client` — passing.
- `eslint` on the changed files — no errors.

### DIFF
--- a/packages/client/src/runtime/core/engines/common/utils/maskQuery.test.ts
+++ b/packages/client/src/runtime/core/engines/common/utils/maskQuery.test.ts
@@ -101,3 +101,9 @@ test('aggregate', () => {
       }"
   `)
 })
+
+test('string with an escaped quote', () => {
+  const query = `email_endsWith: "safe\\"secret", next: "other"`
+
+  expect(maskQuery(query)).toMatchInlineSnapshot(`"email_endsWith: "X", next: "X""`)
+})
--- a/packages/client/src/runtime/core/engines/common/utils/maskQuery.test.ts
+++ b/packages/client/src/runtime/core/engines/common/utils/maskQuery.test.ts
@@ -48,7 +48,7 @@ test('big query', () => {
         where: {
           age_gt: 5
           age_in: [5, 5, 5]
-          name_in: ["X"]
+          name_in: ["X", "X", "X"]
           OR: [
             {
               age_gt: 5
--- a/packages/client/src/runtime/core/engines/common/utils/maskQuery.ts
+++ b/packages/client/src/runtime/core/engines/common/utils/maskQuery.ts
@@ -5,7 +5,7 @@ export function maskQuery(query?: string): string {
   return (
     query
       // replace all strings with X
-      .replace(/".*?"/g, '"X"')
+      .replace(/"(?:[^"\\]|\\.)*"/g, '"X"')
       // replace all numbers with 5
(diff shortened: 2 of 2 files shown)


santichausis:
Good catch on the escaped-quote case — pushed a fix using `/"(?:[^"\\]|\\.)*"/g` instead, which treats `\"` as part of the string content rather than a terminator, plus a regression test for it.

wbizmo (approved):
The change looks correct and appropriately scoped.

The previous `".*"` matcher was greedy, so a query containing multiple quoted values could be consumed from the first opening quote through the final closing quote. That masked the sensitive values, but it also collapsed the structure between them and reduced the usefulness of the resulting diagnostic query.

The replacement:

```regex
"(?:[^"\\]|\\.)*"
```

keeps each match bounded by an unescaped closing quote while allowing escaped characters within the string. With the global replacement, adjacent quoted values are therefore masked independently without altering the surrounding query structure.

I also validated the matcher independently against multiple quoted values, escaped quotes, escaped backslashes, empty strings, multiline content, and unterminated strings. The intended cases behaved correctly, and the previous greedy matcher reproduced the structural collapse addressed by this PR.

The updated array snapshot covers the original regression, while the escaped-quote test covers the most important boundary condition introduced by the new matcher. A broader parser or refactor would add complexity without being proportionate to this utility.

This is a focused and sufficient fix.

aqrln (approved):
This change looks good in isolation but I don't believe it fixes the referenced issue because it still makes the escaped queries largely useless with JSON protocol in Prisma 5+, as it will replace the keys as well. I'll merge this since it's improving the status quo a little bit (at least some structure will be preserved now), but it's not really the solution.

aqrln:
For what it's worth, given that this has been broken for so long and we don't make use of that data, I think the best course of action would be to just delete this legacy code completely.

PR #29727, opened 2026-07-16
fix(client-engine-runtime): send ROLLBACK when discarding a timed-out transaction start

### LINKED ISSUE #29719: Driver adapters: connection returned to pool with an open transaction (ROLLBACK never sent) when maxWait expires during startTransaction — "there is already a transaction in progress"
### Bug description

When an interactive transaction's `maxWait` expires while `adapter.startTransaction()` is still in flight, the client-engine `TransactionManager` discards the late-arriving transaction with a bare `tx.rollback()`. For `usePhantomQuery: false` adapters (`@prisma/adapter-pg`, `@prisma/adapter-neon`), `rollback()` sends **no SQL** — it only releases the client back to the pool — while `BEGIN` has already been executed on the wire. The pooled connection therefore re-enters `pg.Pool` **inside an open transaction**.

Consequences on the next checkouts of that connection:

- Postgres logs `WARNING: there is already a transaction in progress` when the next transaction BEGINs on it.
- The next `COMMIT` on that connection **commits the leaked transaction's work along with its own** (atomicity violation), or — if the leaked transaction had errored — every subsequent statement fails with `25P02 current transaction is aborted` until something sends ROLLBACK/COMMIT.
- Until reuse, `pg_stat_activity` shows the backend `idle in transaction` with `query = 'BEGIN'`.

We hit this in production (Prisma 7.8.0, Next.js, node-postgres): host CPU-stall episodes froze the event loop past `maxWait` (default 2000 ms), producing exactly this triad — client-side P2028 `Unable to start a transaction in the given time`, an `idle in transaction` backend stuck on `BEGIN`, the postgres warning on reuse, and one real casualty: an INSERT that ran inside a leaked aborted transaction and failed with `25P02`. Note the trigger doesn't require a stall: **pool exhaustion alone** (all slots busy for longer than `maxWait`, so `pool.connect()` resolves late) takes the same code path.

#### Where it happens

`packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts` (`7.8.0`), `#startTransactionImpl`:

```ts
const startTransactionPromise = this.driverAdapter
  .startTransaction(options.isolationLevel)
  .catch(rethrowAsUserFacing)

transaction.transaction = await Promise.race([
  startTransactionPromise.finally(() => clearTimeout(startTimer)),
  once(abortController.signal, 'abort').then(() => undefined),
])
...
    if (abortController.signal.aborted) {
      // The startTransaction promise may still be running in the background.
      // If it eventually succeeds, we need to release the connection to avoid
      // leaking it and exhausting the connection pool. ...
      void startTransactionPromise
        .then((tx) => tx.rollback())          // <-- releases WITHOUT sending ROLLBACK
        .catch((e) => debug('error in discarded transaction:', e))

      await this.#closeTransaction(transaction, 'timed_out')  // transaction.transaction === undefined, sends nothing

      throw new TransactionStartTimeoutError()
    }
```

The manager's own `#closeTransaction` knows the contract for `usePhantomQuery: false` adapters — *it* sends the explicit `ROLLBACK` query and only then calls `tx.rollback()`:

```ts
} else if (tx.transaction) {
  if (tx.transaction.options.usePhantomQuery) {
    ...
  } else {
    const query = ROLLBACK_QUERY()
    try {
      await this.#withQuerySpanAndEvent(query, tx.transaction, () => tx.transaction!.executeRaw(query))
    } finally {
      await tx.transaction.rollback()
    }
  }
}
```

…but the discard path above skips that and calls `rollback()` bare. And `@prisma/adapter-pg`'s `PgTransaction.rollback()` (`packages/adapter-pg/src/pg.ts`, 7.8.0) is a pure resource release:

```ts
async rollback(): Promise<void> {
  debug(`[js::rollback]`)
  this.cleanup?.()
  this.client.release()   // no ROLLBACK statement; BEGIN already ran in startTransaction()
}
```

`client.release()` with no error argument returns the client to pg-pool's idle list with **no session reset** — mid-transaction.

The existing unit test encodes the bug: `transaction-manager.test.ts` ("transaction start timeout cleans up connection if transaction eventually starts") builds a `usePhantomQuery: false` mock but asserts only `expect(rollbackMock).toHaveBeenCalled()` — it never asserts that a `ROLLBACK` statement was executed on the discarded transaction.

#### History

- 7.1.0 (#28723) introduced the prompt maxWait abort; the raced-away promise had no continuation at all (connection never released — pool exhaustion).
- 7.2.0 (#28831, "don't leak connections in timed out transactions") added `.then((tx) => tx.rollback())`, fixing the pool-slot leak but introducing this release-without-ROLLBACK corruption for `usePhantomQuery: false` adapters.
- Affected: 7.2.0 → 7.8.0 inclusive; `transaction-manager.ts` and `adapter-pg/src/pg.ts` are unchanged on `main` as of 2026-07-13.
- Likely the mechanism behind at least some reports in #27990 (P2028 spikes on the new client engine).
- The stalled refactor #28768 (abortable `startTransaction`) would structurally fix this; #29611 touches the adjacent execution-timeout rollback path.

### How to reproduce

Deterministic, no database needed — an in-memory `pg.Pool` subclass stands in for the server and records what happens to the connection (a real Postgres shows the same thing, just less deterministically; hold the event loop or saturate the pool instead of gating BEGIN):

```ts
import { PrismaClient } from '@prisma/client' // any generated client with a postgresql datasource
import { EventEmitter } from 'node:events'
import { Pool } from 'pg'

const emptyResult = (command: string) => ({ rows: [], fields: [], rowCount: 0, command })

class FakeClient extends EventEmitter {
  sql: string[] = []
  inTx = false
  beginWhileInTx = 0
  releasedInTx = 0
  holdBegin: Promise<void> | null = null
  query(cfg: any): Promise<any> {
    const text: string = typeof cfg === 'string' ? cfg : (cfg.text ?? '')
    this.sql.push(text)
    const kw = text.trim().split(/\s+/)[0]?.toUpperCase()
    const settle = () => {
      if (kw === 'BEGIN') {
        if (this.inTx) this.beginWhileInTx++ // = postgres "already a transaction in progress"
        this.inTx = true
      } else if (kw === 'COMMIT' || (kw === 'ROLLBACK' && !/^ROLLBACK\s+TO/i.test(text))) this.inTx = false
      return emptyResult(kw)
    }
    if (kw === 'BEGIN' && this.holdBegin) {
      const gate = this.holdBegin
      this.holdBegin = null
      return gate.then(settle)
    }
    return Promise.resolve().then(settle)
  }
  release(err?: Error | boolean) {
    if (this.inTx) this.releasedInTx++
    if (!err) pool.free.push(this) // reuse, like pg-pool's idle list
  }
}

class FakePool extends Pool {
  clients: FakeClient[] = []
  free: FakeClient[] = []
  nextHold: Promise<void> | null = null
  override connect(): Promise<any> {
    const c = this.free.pop() ?? (this.clients.push(new FakeClient()), this.clients.at(-1)!)
    if (this.nextHold) { c.holdBegin = this.nextHold; this.nextHold = null }
    return Promise.resolve(c)
  }
}

const pool = new FakePool()
const prisma = new PrismaClient({ adapter: new PrismaPg(pool) })

let thaw!: () => void
pool.nextHold = new Promise<void>((r) => (thaw = r)) // BEGIN reply delayed past maxWait — like an event-loop stall / slow server

await prisma.$transaction(async () => {}, { maxWait: 100, timeout: 2000 }).catch((e) => console.log(e.code)) // P2028
thaw() // the stall ends; TransactionManager's discard cleanup runs now
await new Promise((r) => setTimeout(r, 50))

const c = pool.clients[0]
console.log('SQL sent:', c.sql)                 // [ 'BEGIN' ]  — no ROLLBACK, ever
console.log('released in-tx:', c.releasedInTx)  // 1           — pool got a mid-transaction connection

await prisma.$transaction(async () => {}, { maxWait: 1000, timeout: 2000 }) // reuses the connection
console.log('BEGIN on in-tx conn:', c.beginWhileInTx) // 1     — the postgres WARNING; this COMMIT adopts the leaked tx
```

Observed output with `@prisma/client` / `@prisma/adapter-pg` 7.8.0:

```
P2028
SQL sent: [ 'BEGIN' ]
released in-tx: 1
BEGIN on in-tx conn: 1
```

### Expected behavior

The discarded-start path should honor the same contract as `#closeTransaction`: for `usePhantomQuery: false` adapters, execute `ROLLBACK` on the transaction before `rollback()` releases the connection, e.g.

```ts
void startTransactionPromise
  .then((tx) =>
    tx.options.usePhantomQuery
      ? tx.rollback()
      : tx.executeRaw(ROLLBACK_QUERY()).finally(() => tx.rollback()),
  )
  .catch((e) => debug('error in discarded transaction:', e))
```

(or give adapters a dispose-with-rollback primitive / land #28768). A connection must never re-enter the pool with an open transaction: the current behavior silently merges unrelated queries into a stale transaction, which is data-corruption grade, not just noise.

### Prisma information

```prisma
generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
}
```

```ts
const adapter = new PrismaPg({ connectionString: process.env.DATABASE_URL })
// any interactive transaction whose start outlives maxWait:
await prisma.$transaction(async (tx) => { ... }) // default maxWait 2000ms
```

### Environment & setup

- OS: Linux (Ubuntu 22.04 VM, prod) / Windows 11 (repro)
- Database: PostgreSQL 16 (mechanism is server-version-independent; also shown with the fake pool above)
- Node.js version: v20.x (prod), v24.16 (repro)
- pg: 8.13.x

### Prisma version

```
prisma            : 7.8.0
@prisma/client    : 7.8.0
@prisma/adapter-pg: 7.8.0
```

Also applies to `@prisma/adapter-neon` (same release-only `rollback()`, `usePhantomQuery: false`). 7.1.0 has the older never-released variant of the leak; ≤ 7.0.x waited for `startTransaction` to finish and closed via the full path, so it is not affected this way.

lazerg:
## Description

When an interactive transaction's `maxWait` expires while `adapter.startTransaction()` is still in flight, `TransactionManager` discards the transaction once it eventually resolves. It did this with a bare `tx.rollback()`, but for adapters that don't use phantom queries (`@prisma/adapter-pg`, `@prisma/adapter-neon`) `rollback()` only releases the connection back to the pool without sending any SQL. `BEGIN` has already run on the wire at that point, so the connection goes back into the pool mid-transaction. The next query that picks it up then hits `there is already a transaction in progress`, or worse, commits the leaked transaction's work along with its own.

`#closeTransaction` already handles this correctly: for non-phantom-query adapters it sends an explicit `ROLLBACK` before calling `rollback()`. This applies the same branch to the discard path, so the connection is always rolled back before it's released.

Closes #29719

## Test plan

The existing `transaction start timeout cleans up connection if transaction eventually starts` test already used a `usePhantomQuery: false` adapter, but only checked that `rollback()` ran, not that `ROLLBACK` was actually sent on the discarded transaction. I captured its transaction's `executeRaw` mock and asserted the discarded transaction executes `ROLLBACK`; it fails on `main` and passes with the fix.

`pnpm --filter @prisma/client-engine-runtime test` passes (202 tests).

### DIFF
--- a/packages/client-engine-runtime/src/transaction-manager/transaction-manager.test.ts
+++ b/packages/client-engine-runtime/src/transaction-manager/transaction-manager.test.ts
@@ -610,6 +610,7 @@ test('transaction start timeout cleans up connection if transaction eventually s
   const REMAINING_TIME_FOR_START = SLOW_START_TRANSACTION_TIME - TIME_PAST_MAX_WAIT
 
   const rollbackMock = vi.fn().mockResolvedValue(undefined)
+  const executeRawMock = vi.fn().mockResolvedValue(1)
 
   const driverAdapter = {
     adapterName: 'slow-adapter',
@@ -628,7 +629,7 @@ test('transaction start timeout cleans up connection if transaction eventually s
                 provider: 'postgres',
                 options: { usePhantomQuery: false },
                 queryRaw: vi.fn(),
-                executeRaw: vi.fn(),
+                executeRaw: executeRawMock,
                 commit: vi.fn(),
                 rollback: rollbackMock,
               }),
@@ -660,8 +661,10 @@ test('transaction start timeout cleans up connection if transaction eventually s
   // Now advance time to let the startTransaction promise resolve
   await vi.advanceTimersByTimeAsync(REMAINING_TIME_FOR_START)
 
-  // After the background startTransaction completes, rollback should be called
-  // to release the connection and avoid pool exhaustion
+  // After the background startTransaction completes, the discarded transaction must send an
+  // explicit ROLLBACK before releasing the connection. Without it, a non-phantom-query adapter
+  // (e.g. pg/neon) returns a connection to the pool while it is still mid-transaction.
+  expect(executeRawMock).toHaveBeenCalledWith(expect.objectContaining({ sql: 'ROLLBACK' }))
   expect(rollbackMock).toHaveBeenCalled()
 })
 
--- a/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
+++ b/packages/client-engine-runtime/src/transaction-manager/transaction-manager.ts
@@ -172,12 +172,25 @@ export class TransactionManager {
       case 'waiting':
         if (abortController.signal.aborted) {
           // The startTransaction promise may still be running in the background.
-          // If it eventually succeeds, we need to release the connection to avoid
-          // leaking it and exhausting the connection pool. We ignore any errors
-          // that happen during startup or rollback here because we will have
-          // already returned our own `TransactionStartTimeoutError` error to the user.
+          // If it eventually succeeds, we need to roll back and release the connection to
+          // avoid leaking it and exhausting the connection pool. For adapters that don't use
+          // phantom queries (e.g. pg/neon), `rollback()` only releases the connection without
+          // sending SQL, so we send an explicit ROLLBACK first; otherwise the connection returns
+          // to the pool mid-transaction because `BEGIN` already ran on the wire during startup.
+          // We ignore any errors here because we will have already returned our own
+          // `TransactionStartTimeoutError` error to the user.
           void startTransactionPromise
-            .then((tx) => tx.rollback())
+            .then(async (tx) => {
+              if (tx.options.usePhantomQuery) {
+                await tx.rollback()
+              } else {
+                try {
+                  await tx.executeRaw(ROLLBACK_QUERY())
+                } finally {
+                  await tx.rollback()
+                }
+              }
+            })
             .catch((e) => debug('error in discarded transaction:', e))
 
           // Call `#closeTransaction` to update internal state. It won't actually attempt


aqrln (approved):
LGTM, thanks a lot!

Could you please [sign the CLA](https://github.com/prisma/prisma/pull/29727#issuecomment-4996135534) so we can merge it?

lazerg:
@aqrln done ✔️

aqrln:
Fantastic, thanks!

PR #29729, opened 2026-07-17
fix(config): add directUrl to datasource type definition for prisma.config.ts

### LINKED ISSUE #29726: Missing directUrl in datasource type definition for prisma.config.ts
### Bug description
When using the new `prisma.config.ts` configuration, passing a `directUrl` to the `datasource` object causes a TypeScript error because `directUrl` is missing from the type definition. 

Since `directUrl` is required when using connection poolers (like Supabase), it should be a supported property inside `defineConfig` just as it is inside `schema.prisma`.

### How to reproduce
1. Create a `prisma.config.ts` file.
2. Import `defineConfig` from `prisma/config`.
3. Add a `datasource` object with `directUrl`:
```typescript

export default defineConfig({
  datasource: {
    url: process.env["DATABASE_URL"],
    directUrl: process.env["DIRECT_URL"], // <-- TypeScript Error
  },
});
```

### Severity

### Steps to reproduce
1. Initialize a new project with Prisma `7.8.0` (or similar version supporting `prisma.config.ts`).
2. Create a `prisma.config.ts` file.
3. Import `defineConfig` from `prisma/config`.
4. Add a `datasource` object and provide a `directUrl`:
```typescript
export default defineConfig({
  datasource: {
    url: process.env["DATABASE_URL"],
    directUrl: process.env["DIRECT_URL"], // <-- TypeScript Error
  },
});

Expected Behavior: `directUrl` should be recognized as a valid, optional string property in the datasource configuration object (alongside `url` and `shadowDatabaseUrl`).

Actual Behavior: TypeScript throws the following error indicating the property is missing from the types: Object literal may only specify known properties, and `directUrl` does not exist in type `{ url?: string | undefined; shadowDatabaseUrl?: string | undefined; }`.

### Frequency

Consistently reproducible

### Is this a regression?

No, this is not a regression. It appears to be an omission in the type definitions for the newly introduced prisma.config.ts feature.

### Workaround

You can bypass the TypeScript compilation error using a` // @ts-ignore` comment, and Prisma will still correctly process the URL at runtime

### Prisma Config

```ts
```

### Logs & Debug Info

```
// Debug logs here
```

### Environment & Setup

OS: Windows / macOS / Linux
Database: PostgreSQL (with connection pooling)
Node.js version: 22

### Prisma Version

```
7.8.0
```

AmariahAK:
## Description

Closes #29726

### What was the issue

When using the `prisma.config.ts` configuration, passing a `directUrl` to the `datasource` object caused a TypeScript error because `directUrl` was missing from the `Datasource` type definition. Since `directUrl` is required when using connection poolers like Supabase, it should be a supported property inside `defineConfig` just as it is inside `schema.prisma`.

### Where was the issue

`packages/config/src/PrismaConfig.ts` — both the runtime validator (`DatasourceShape`) and the exported TypeScript type (`Datasource`) were missing `directUrl` as an optional property.

### How it was fixed

Added `directUrl` as an optional string (`directUrl?: string`) to both the `DatasourceShape` effect/Schema validator and the `Datasource` type. Updated the corresponding test fixtures and inline snapshots in `defineConfig.test.ts` and `loadConfigFromFile.test.ts`.

### Why this approach

- Follows the exact same pattern used for `shadowDatabaseUrl` — optional string, same level in both the Shape and the type
- No runtime or engine changes needed — the Rust schema engine already supports `directUrl` natively (it's a standard `schema.prisma` datasource field). The engine receives the datasource as raw JSON via `--datasource`, so only the TypeScript type needed updating
- Minimal diff: 4 files, +8/-2 lines

### How to test locally

```bash
cd packages/config
pnpm run test
```

All 71 tests pass. You can also verify manually with a `prisma.config.ts`:

```ts

export default defineConfig({
  datasource: {
    url: process.env['DATABASE_URL'],
    directUrl: process.env['DIRECT_URL'], // no more TypeScript error
  },
})
```

### DIFF
--- a/packages/config/src/PrismaConfig.ts
+++ b/packages/config/src/PrismaConfig.ts
@@ -21,11 +21,13 @@ export type ExperimentalConfig = {
 const DatasourceShape = Shape.Struct({
   url: Shape.optional(Shape.String),
   shadowDatabaseUrl: Shape.optional(Shape.String),
+  directUrl: Shape.optional(Shape.String),
 })
 
 export type Datasource = {
   url?: string
   shadowDatabaseUrl?: string
+  directUrl?: string
 }
 
 export type SchemaEngineConfigInternal = {
--- a/packages/config/src/__tests__/defineConfig.test.ts
+++ b/packages/config/src/__tests__/defineConfig.test.ts
@@ -54,11 +54,13 @@ describe('defineConfig', () => {
         datasource: {
           url: 'postgresql://DATABASE_URL',
           shadowDatabaseUrl: 'postgresql://SHADOW_DATABASE_URL',
+          directUrl: 'postgresql://DIRECT_URL',
         },
       })
       expect(config.datasource).toMatchObject({
         url: 'postgresql://DATABASE_URL',
         shadowDatabaseUrl: 'postgresql://SHADOW_DATABASE_URL',
+        directUrl: 'postgresql://DIRECT_URL',
       })
     })
   })
--- a/packages/config/src/__tests__/fixtures/loadConfigFromFile/datasource/prisma.config.ts
+++ b/packages/config/src/__tests__/fixtures/loadConfigFromFile/datasource/prisma.config.ts
@@ -2,5 +2,6 @@ export default {
   datasource: {
     url: 'postgresql://DATABASE_URL',
     shadowDatabaseUrl: 'postgresql://SHADOW_DATABASE_URL',
+    directUrl: 'postgresql://DIRECT_URL',
   },
 }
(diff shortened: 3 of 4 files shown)


aqrln:
I think you either misunderstand what `directUrl` is, or what `prisma.config.ts` configures. This PR does nothing of value because there's nothing in the code that would ever read that `directUrl` value out of the config, and there's no reason why we'd need a `directUrl` option in addition to `url` here.

`directUrl` in Prisma 6 and below was the URL that Migrate and CLI would use, as opposed to to the `url` used by Prisma Client at run time. There's no such thing as `directUrl` in Prisma 7 precisely because the CLI configuration (`prisma.config.ts`) and the runtime driver configuration (driver adapter arguments) are cleanly separated. In other words, the `url` property in `prisma.config.ts` **is** exactly what `directUrl` used to be.

AmariahAK:
Thanks for the detailed context @aqrln . I hadn't clocked that in Prisma 7 the `url` in `prisma.config.ts` *is* the old `directUrl`, since the CLI config and the runtime driver args are now cleanly separated. Appreciate you taking the time to explain the P7 model rather than just closing it.

aqrln:
No worries! I think this indicates that it isn't explained well in our docs, which is something we should fix.

Issue #9988, opened 2021-10-27
Update does not enforce type on `data`


austincrim:
### Discussed in https://github.com/prisma/prisma/discussions/9957

> <div type='discussions-op-text'>
> 
> <sup>Originally posted by **GuiSim** October 26, 2021</sup>
> Hello! 
> 
> I'm new to Prisma and testing it out. I encountered an issue today where an `update` statement would not raise any type issue on an invalid `data` object.
> 
> ```
> prisma.mything.update({
>   where: { id: '1' },
>   data: { invalidField: 42 }, // Does not compile as `invalidField` is not a valid field on `mything`
> });
> 
> const data = { invalidField: 42 };
> prisma.mything.update({
>   where: { id: '1' },
>   data: data, // Compiles without issues. Why?
> });
> ```
> 
> Can you help me understand why?</div>

Jolg42:
See https://github.com/prisma/prisma/issues/22445

PR #29735, opened 2026-07-19
fix(client): reject non-object values in XOR type helper

### LINKED ISSUE #22445: No typing error when passing a string or a number to data for update()
### Bug description

When calling update, any type can be passed to data:

```
prisma.user.update({
  where: {
    id: 1,
  },
  data: 10, // should trigger a type error
})
```

Null or Undefined trigger Typescript errors but strings or numbers don't.

There might be something wrong with the XOR type definition:

```
import { Prisma } from '@prisma/client'

type XOR<T, U> = T extends object
  ? U extends object
    ? (Without<T, U> & U) | (Without<U, T> & T)
    : U
  : T

type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never }

type TestType = XOR<Prisma.UserUpdateInput, Prisma.UserUncheckedUpdateInput>

const fn = (a: TestType) => {}
const fn1 = (a: Prisma.UserUpdateInput) => {}
const fn2 = (a: Prisma.UserUncheckedUpdateInput) => {}

fn(5) // No errors
fn1(5) // Type '5' has no properties in common with type 'UserUpdateInput'.ts(2559)
fn2(5) // Type '5' has no properties in common with type 'UserUncheckedUpdateInput'.ts(2559)

```

### How to reproduce

I just followed the Getting started.
1. Create model as below
2. `npx prisma generate`

### Expected behavior

A type error should be raised when anything but an object matching the table type is passed.

### Prisma information

```prisma

generator client {
  provider = "prisma-client-js"
}

datasource db {
  provider = "postgresql"
}

model User {
  email String  @unique
  name  String?
}


```

### Environment & setup

- OS: Any
- Database: Any
- Node.js version: 18.17.1

### Prisma Version

```
prisma                  : 5.7.0
@prisma/client          : 5.7.0
Computed binaryTarget   : darwin
Architecture            : x64
Node.js                 : v18.17.1
Query Engine (Node-API) : libquery-engine 79fb5193cf0a8fdbef536e4b4a159cad677ab1b9 (at node_modules/@prisma/engines/libquery_engine-darwin.dylib.node)
Schema Engine           : schema-engine-cli 79fb5193cf0a8fdbef536e4b4a159cad677ab1b9 (at node_modules/@prisma/engines/schema-engine-darwin)
Schema Wasm             : @prisma/prisma-schema-wasm 5.7.0-41.79fb5193cf0a8fdbef536e4b4a159cad677ab1b9
Default Engines Hash    : 79fb5193cf0a8fdbef536e4b4a159cad677ab1b9
Studio                  : 0.495.0

```

kyungseopk1m:
## Problem

Passing a primitive to an XOR-typed argument such as `update({ data: 5 })` is accepted at compile time even though the runtime rejects it. Only `null` and `undefined` are caught today. Reported in #22445 (duplicate of #9988).

```ts
prisma.user.update({
  where: { id: 1 },
  data: 5, // no type error, but there should be one
})
```

## Cause

`XOR<T, U>` expands to `(Without<T, U> & U) | (Without<U, T> & T)`. For update inputs the checked type's keys are a subset of the unchecked type's keys, so `Without<T, U>` reduces to `{}`. The member `{} & U` then loses the weak-type check that would otherwise reject a primitive against an all-optional object, so the union accepts numbers and strings.

## Change

Intersect the union with `object` once:

```ts
type XOR<T, U> =
  T extends object ?
  U extends object ?
    ((Without<T, U> & U) | (Without<U, T> & T)) & object
  : U : T
```

`object` excludes primitives, `null` and `undefined` while keeping every object shape, including empty objects, index signatures, branded objects, arrays and `Date`. Mutual exclusivity between the two sides is unchanged.

Applying `& object` to the whole union rather than to each member matters for type performance. Intersecting each member raised type instantiations by 4 to 9.5% on the type benchmarks, while the single outer intersection leaves them effectively unchanged (basic +0.002%, huge-schema +0.003%, lots-of-relations 0%).

The helper is emitted by both the TypeScript and JavaScript client generators, so both `common.ts` files are updated.

## Verification

- Generated a client with both generators and confirmed `update({ data: 5 })` now fails to type-check, while normal create/update/where and nested `connectOrCreate` still compile.
- Unit tests for `@prisma/client-generator-ts` and `@prisma/client-generator-js` pass.
- Ran the type benchmarks (basic, huge-schema, lots-of-relations) with the original and updated helper on the same generated output; instantiation counts are unchanged.

Closes #22445

### DIFF
--- a/packages/client-generator-js/src/TSClient/common.ts
+++ b/packages/client-generator-js/src/TSClient/common.ts
@@ -329,7 +329,7 @@ type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 type XOR<T, U> =
   T extends object ?
   U extends object ?
-    (Without<T, U> & U) | (Without<U, T> & T)
+    ((Without<T, U> & U) | (Without<U, T> & T)) & object
   : U : T
 
 
--- a/packages/client-generator-ts/src/TSClient/common.ts
+++ b/packages/client-generator-ts/src/TSClient/common.ts
@@ -155,7 +155,7 @@ type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 export type XOR<T, U> =
   T extends object ?
   U extends object ?
-    (Without<T, U> & U) | (Without<U, T> & T)
+    ((Without<T, U> & U) | (Without<U, T> & T)) & object
   : U : T
 
 


aqrln (approved):
Fantastic, thanks! This may need to have some snapshots updated in tests when running the full suite, let's see what CI says.

PR #29737, opened 2026-07-20
fix(adapter-ppg): handle both quoted and unquoted column names in ColumnNotFound

kyungseopk1m:
Postgres reports a missing column in more than one way. The identifier is
quoted only when it needs to be, and it can be qualified with the table
name:

```
column foo does not exist
column "foo" does not exist
column users.first_name does not exist
column "users"."first name" does not exist
```

`adapter-ppg` splits the message on double quotes, which only handles the
second form. The unquoted forms produce `undefined`, and an identifier
containing spaces is truncated at the first space, so `P2022` ends up
without a usable column name.

The same bug was fixed in `adapter-pg` in #29307. This applies the same
parsing to `adapter-ppg` and adds the test cases from that fix, which the
package did not have before.

### DIFF
--- a/packages/adapter-ppg/src/errors.test.ts
+++ b/packages/adapter-ppg/src/errors.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from 'vitest'
+
+import { convertDriverError } from './errors'
+
+describe('convertDriverError', () => {
+  it.each([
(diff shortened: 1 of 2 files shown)


aqrln (approved):
Thank you!

PR #29738, opened 2026-07-20
fix(client-generator-js): do not emit undefined for schemas without models

kyungseopk1m:
A schema without models generates declarations containing a bare
`undefined` statement:

```ts
  /**
   * Deep Input Types
   */

  undefined
```

`inputObjectTypes.prisma` is undefined when there are no models, and the
optional chain result is interpolated into the template as is. The
interpolation for `inputObjectTypes.model` two lines below already
handles this.

Generating without models needs no flag since `--allow-no-models` was
replaced by `--require-models`, so any schema that only declares enums or
types hits this.

The added test generates from a model-less schema and asserts the
declarations contain no such statement.

### DIFF
--- a/packages/client-generator-js/src/TSClient/TSClient.ts
+++ b/packages/client-generator-js/src/TSClient/TSClient.ts
@@ -232,8 +232,8 @@ ${fieldRefs.join('\n\n')}`
  * Deep Input Types
  */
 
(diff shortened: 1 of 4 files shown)


aqrln (approved):

aqrln:
Thanks!

PR #29742, opened 2026-07-20
fix(client): avoid literal DATABASE_URL in generated JSDoc examples

### LINKED ISSUE #29724: Generated client JSDoc examples embed process.env.DATABASE_URL, tripping security scanners
The `prisma-client` generator's JSDoc usage examples in `internal/class.ts` hardcode `process.env.DATABASE_URL`. Combined with the single-line `inlineSchema` (which makes the file look like a minified bundle to heuristic scanners), this triggers secret-leak findings in security tooling (e.g. react-doctor's `artifact-env-leak`) even though nothing is leaked. Consider a placeholder like `process.env.YOUR_DATABASE_URL_ENV_VAR` in the generated examples, or an option to omit doc examples from generated output.

santichausis:
## Description

The generated client's JSDoc/comment usage examples hardcode `process.env.DATABASE_URL` as the connection string:

```ts
  adapter: new PrismaPg({ connectionString: process.env.DATABASE_URL })
})
```

Combined with the single-line inlined schema in the current generator's output (`internal/class.ts`), this makes the generated file look like a minified bundle to heuristic secret-scanning tools, and the literal `process.env.DATABASE_URL` pattern trips false-positive "env leak" findings — even though nothing is actually leaked, since it's just referencing an env var by name.

This swaps the example to a clearly placeholder-style env var name (`YOUR_DATABASE_URL_ENV_VAR`) in both the current TS-based generator (`client-generator-ts`) and the legacy JS one (`client-generator-js`), since both produce the same kind of generated example.

Intentionally **not** changed:
- `packages/client/src/runtime/utils/validatePrismaClientOptions.ts` — this is a real runtime error message shown to users when client options are invalid, not generated example code.
- `packages/client/src/runtime/getPrismaClient.ts` — this is `@prisma/client`'s own shipped source/JSDoc, not per-project generated output.

Both show the real, correct `DATABASE_URL` convention, which is what users should actually see there.

Closes #29724

## Test plan

- `pnpm run test` in `packages/client-generator-ts` — 39 tests passing, including `tests/generator.test.ts` which generates a real client from several schemas and validates the resulting TypeScript.
- `pnpm run test` in `packages/client-generator-js` — 18 tests passing.
- `eslint` on the changed files — no errors.
- Confirmed via `grep` that no `.snap` files reference the changed string, and that all other `process.env.DATABASE_URL` occurrences in the codebase (outside the 3 changed files) are either the two files above or unrelated test usages, not generated example code.

### DIFF
--- a/packages/client-generator-js/src/TSClient/PrismaClient.ts
+++ b/packages/client-generator-js/src/TSClient/PrismaClient.ts
@@ -410,7 +410,7 @@ export class PrismaClientClass implements Generable {
  * @example
  * \`\`\`
  * const prisma = new PrismaClient({
- *   adapter: new PrismaPg({ connectionString: process.env.DATABASE_URL })
+ *   adapter: new PrismaPg({ connectionString: process.env.YOUR_DATABASE_URL_ENV_VAR })
  * })
  * // Fetch zero or more ${capitalize(example.plural)}
  * const ${uncapitalize(example.plural)} = await prisma.${uncapitalize(example.model)}.findMany()
@@ -632,7 +632,7 @@ export type TransactionClient = Omit<Prisma.DefaultPrismaClient, ${transactionCl
             import { PrismaPg } from '@prisma/adapter-pg'
             import { PrismaClient } from './generated/prisma/client'
 
-            const adapter = new PrismaPg({ connectionString: process.env.DATABASE_URL })
+            const adapter = new PrismaPg({ connectionString: process.env.YOUR_DATABASE_URL_ENV_VAR })
             const prisma = new PrismaClient({ adapter })
             \`\`\`
           `),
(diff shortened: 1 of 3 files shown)


aqrln:
Thank you for the PR @santichausis! Did you run into the issue yourself as well, or did you pick it up because it was available? I'm asking because that issue frankly sounds a bit weird, and doesn't contain any reproduction to prove the claims. At best that sounds like a bug in those linter rules which will certainly get fixed properly if it is reported in the appropriate place and not here, at worst like the user holding it wrong and misconfiguring those rules.

I would prefer not to have unnecessary churn and deviate these JSDoc comments from the official examples, docs, actual env var used by default by Prisma Compute + Prisma Postgres, and so on, unless it is actually a pain point. Since we only have one report with no reproduction, and no upvotes in that issue, I'm leaning towards just closing both the issue and the PR. Happy to revisit this if I'm wrong and it does affect many people.

PR #29747, opened 2026-07-21
fix(adapter-neon): map args through mapArg in PrismaNeonHttpAdapter

### LINKED ISSUE #29743: adapter-neon: PrismaNeonHttp passes query args unmapped — Bytes params silently corrupted, DateTime params get local-timezone skew
### Bug description

`PrismaNeonHttpAdapter.performIO` (in `@prisma/adapter-neon`) passes query arguments to the Neon HTTP client **without running them through `mapArg`**, while the WebSocket adapter (`NeonWsQueryable.performIO`) maps every argument:

```js
// NeonWsQueryable.performIO (WebSocket) — correct
const result = await this.client.query(
  { text: sql, rowMode: 'array', types: { ... } },
  args.map((arg, i) => mapArg(arg, query.argTypes[i]))   // ✅
);

// PrismaNeonHttpAdapter.performIO (HTTP) — args passed through raw
const { sql, args: values } = query;
return await this.client(sql, values, { arrayMode: true, fullResults: true, types: { ... } });  // ❌ no mapArg
```

The query engine hands the adapter `Bytes` arguments as **base64 strings** and `DateTime` arguments as ISO strings / `Date` objects — `mapArg` is what converts them to driver-ready values. Skipping it causes two silent data-corruption bugs on the HTTP adapter:

1. **`Bytes` parameters are stored and compared as literal base64 text.** The base64 string reaches Postgres as a plain text param on a `bytea` column, which interprets it in escape format — so the *text characters* of the base64 encoding are stored (e.g. 32 bytes stored for a 24-byte value). Writes are silently corrupted, and any filter on a `Bytes` column never matches rows containing correct raw bytes (rows written by anything other than the HTTP adapter itself). No error is raised anywhere.

   Real-world impact: this broke all WebAuthn/passkey logins on our production site after migrating to Prisma — `credential.findUnique({ where: { credentialId } })` silently returned `null` for every stored credential, and newly registered credentials were stored with corrupted `credentialId`/`publicKey`.

2. **`Date` parameters are serialized with the process's local timezone offset.** Unmapped `Date` objects fall through to the Neon driver's pg-style `prepareValue`, which renders local time with an offset suffix (e.g. `2026-07-20T20:08:28.709-07:00`); a `timestamp` (without time zone) column silently drops the offset, storing local wall-clock time while reads assume UTC. Every `DateTime` write is skewed by the process's UTC offset in any non-UTC runtime (e.g. `expiresAt` values land 7 hours in the past in a PDT dev environment).

Reproduced on `@prisma/adapter-neon` 7.8.0 and 7.9.0.

### Severity

(Silent corruption of every `Bytes` write and silent mismatch of every `Bytes` filter through the HTTP adapter; no error surfaced.)

### Reproduction

Schema:

```prisma
model Thing {
  id   Int   @id
  blob Bytes
}
```

Script (any Neon database URL, or Neon's serverless protocol via the [local-neon-http-proxy](https://github.com/TimoWilhelm/local-neon-http-proxy) docker image in front of a local Postgres):

```ts
import { PrismaNeonHttp } from '@prisma/adapter-neon';
import { PrismaClient } from './generated/client.js';

const db = new PrismaClient({ adapter: new PrismaNeonHttp(process.env.DATABASE_URL, {}) });

// 1. Bytes writes are corrupted:
const bytes = Buffer.from('hello-raw-bytes'); // 15 bytes
await db.thing.create({ data: { id: 1, blob: bytes } });
const [row] = await db.$queryRaw`
  SELECT octet_length("blob") AS len, encode("blob", 'hex') AS hex FROM "Thing" WHERE id = 1`;
console.log(Number(row.len));                       // 20  (base64 text length) — expected 15
console.log(row.hex === bytes.toString('hex'));     // false — stored bytes are the base64 *text*

// 2. Bytes filters never match correctly-stored rows (e.g. written before a migration to Prisma):
await db.$executeRaw`INSERT INTO "Thing" (id, blob) VALUES (2, '\x0102030405'::bytea)`;
console.log(await db.thing.findFirst({ where: { blob: Buffer.from([1, 2, 3, 4, 5]) } })); // null — expected the row

// 3. DateTime skew (run with a non-UTC TZ, e.g. TZ=America/Los_Angeles):
//    create a row with a DateTime field and read it back — the value shifts by the UTC offset.
```

All three behave correctly with the WebSocket adapter (`PrismaNeon`), and all three are fixed by adding the same `mapArg` call the WebSocket path uses:

```diff
   async performIO(query) {
-    const { sql, args: values } = query;
+    const { sql, args } = query;
+    const values = args.map((arg, i) => mapArg(arg, query.argTypes[i]));
     return await this.client(sql, values, {
```

We validated this fix end-to-end (bytea round-trip byte-exact, cross-written rows found by `findUnique`, timestamps stored in UTC).

**Expected:** `Bytes` values round-trip byte-exact and match `bytea` values written by other clients; `DateTime` values are stored in UTC regardless of process timezone — i.e. the HTTP adapter behaves like the WebSocket adapter.

**Actual:** `Bytes` values are stored as the base64 text of the intended bytes and never match correctly-stored rows in filters; `Date` params are rendered in local time, which naive `timestamp` columns keep after dropping the offset. No errors are raised.

### Frequency

Consistently reproducible

### Is this a regression?

Not a regression within Prisma 7 as far as we can tell — present in both 7.8.0 and 7.9.0 (older versions not tested). The mapping exists and works in the WebSocket adapter in the same file, so it appears the HTTP path was simply missed. Possibly related history: #25514 fixed the *result*-decoding direction for `PrismaNeonHTTP` by adding custom type parsers; this issue is the same asymmetry in the *parameter*-encoding direction.

### Workaround

Patch the package (e.g. `pnpm patch @prisma/adapter-neon`) to add the missing mapping in `PrismaNeonHttpAdapter.performIO`, in both `dist/index.js` and `dist/index.mjs`:

```diff
   async performIO(query) {
-    const { sql, args: values } = query;
+    const { sql, args } = query;
+    const values = args.map((arg, i) => mapArg(arg, query.argTypes[i]));
     return await this.client(sql, values, {
```

Alternatively, use the WebSocket adapter (`PrismaNeon`) instead of the HTTP one, at the cost of per-request connection setup in serverless environments.

### Prisma Schema & Queries

```prisma
model Credential {
  id           String @id @default(uuid(7)) @db.Uuid
  credentialId Bytes  @unique
  publicKey    Bytes
  // ...
}
```

```ts
// silently returns null for any row whose bytea was written with correct raw bytes:
await db.credential.findUnique({ where: { credentialId: someBuffer } });

// silently stores base64 text bytes instead of the given bytes:
await db.credential.create({ data: { credentialId: idBuffer, publicKey: keyBuffer /* ... */ } });
```

### Prisma Config

```ts
// prisma.config.ts — nothing unusual; client generated with the `prisma-client` generator.
// Adapter constructed as:
import { PrismaNeonHttp } from '@prisma/adapter-neon';
const adapter = new PrismaNeonHttp(process.env.DATABASE_URL, {});
const db = new PrismaClient({ adapter });
```

### Logs & Debug Info

```
No errors are emitted anywhere — that is part of the problem. Writes succeed
(with corrupted bytes) and filters succeed (returning null / no rows).
```

### Environment & Setup

- **OS:** macOS (dev) / Cloudflare Workers (prod) — bytea corruption reproduces on both; the DateTime skew needs a non-UTC process TZ
- **Database:** PostgreSQL (Neon serverless, HTTP transport)
- **Node.js version:** v22 / v24

### Prisma Version

```
prisma               : 7.8.0
@prisma/client       : 7.8.0
Operating System     : darwin
Architecture         : arm64
Node.js              : v22.23.1
TypeScript           : 6.0.3

(also reproduced against @prisma/adapter-neon 7.9.0)
```

santichausis:
## Description

`PrismaNeonHttpAdapter.performIO` passes query arguments to the Neon HTTP client without running them through `mapArg`, while `NeonWsQueryable.performIO` (used by the WebSocket adapter, same file) maps every argument:

```ts
// NeonWsQueryable.performIO (WebSocket) — correct
args.map((arg, i) => mapArg(arg, query.argTypes[i]))

// PrismaNeonHttpAdapter.performIO (HTTP) — args passed through raw, before this PR
const { sql, args: values } = query
```

`mapArg` is what converts the engine's wire representation of arguments into driver-ready values — e.g. `Bytes` arguments arrive as base64 strings and need decoding to a `Buffer`, and `Date` arguments need to be formatted in UTC rather than left to the driver's default (local-timezone) serialization. Skipping it on the HTTP path caused two silent data-corruption bugs, with no error raised anywhere:

1. **`Bytes` values are stored/compared as literal base64 text** instead of raw bytes — writes are corrupted, and filters on `Bytes` columns never match rows written by anything other than the HTTP adapter itself.
2. **`Date` values are serialized with the process's local timezone offset** instead of UTC, so a `timestamp` (without time zone) column silently stores skewed wall-clock time.

This applies the same `mapArg` call the WebSocket path already uses, in `PrismaNeonHttpAdapter.performIO`.

Closes #29743

## Test plan

- Added a regression test in `packages/adapter-neon/src/__tests__/neon.test.ts` (there were no existing tests for `PrismaNeonHttpAdapter`) that constructs the adapter with a mocked query function and asserts a base64-encoded `Bytes` argument is decoded to a `Buffer` before being passed to the client, rather than passed through as the raw base64 string.
- Confirmed the new test fails on the pre-fix code with the exact corruption described in the issue (`expected 'aGVsbG8tcmF3LWJ5dGVz' to be an instance of Buffer`), and passes with the fix.
- `pnpm run test` in `packages/adapter-neon` — 5/5 tests passing.
- `eslint` on the changed files — no new errors (3 pre-existing warnings elsewhere in the file, unrelated to this change).

### DIFF
--- a/packages/adapter-neon/src/__tests__/neon.test.ts
+++ b/packages/adapter-neon/src/__tests__/neon.test.ts
@@ -1,8 +1,9 @@
 import * as neon from '@neondatabase/serverless'
 import { getLogs } from '@prisma/debug'
+import type { SqlQuery } from '@prisma/driver-adapter-utils'
 import { describe, expect, it, vi } from 'vitest'
 
-import { PrismaNeonAdapterFactory } from '../neon'
+import { PrismaNeonAdapterFactory, PrismaNeonHttpAdapter } from '../neon'
 
 describe('PrismaNeonAdapterFactory', () => {
   it('should subscribe to pool error events', async () => {
@@ -74,5 +75,29 @@ describe('PrismaNeonAdapterFactory', () => {
     expect(mockConnection.removeListener).toHaveBeenCalledWith('error', expect.any(Function))
 
     await adapter.dispose()
+  })
+})
+
+describe('PrismaNeonHttpAdapter', () => {
+  it('maps args through mapArg before passing them to the HTTP client', async () => {
+    const mockClient = vi.fn().mockResolvedValue({ fields: [], rows: [], rowCount: 0 })
+    const adapter = new PrismaNeonHttpAdapter(mockClient as unknown as neon.NeonQueryFunction<any, any>)
+
+    // a base64-encoded `Bytes` arg, as the query engine hands it to the adapter
+    const bytesBase64 = Buffer.from('hello-raw-bytes').toString('base64')
+
+    const query: SqlQuery = {
+      sql: 'INSERT INTO "Thing" ("blob") VALUES ($1)',
+      args: [bytesBase64],
+      argTypes: [{ scalarType: 'bytes', arity: 'scalar' }],
+    }
+
+    await adapter.performIO(query)
+
+    const [, values] = mockClient.mock.calls[0]
+    expect(values[0]).toBeInstanceOf(Buffer)
+    expect((values[0] as Buffer).toString('utf-8')).toBe('hello-raw-bytes')
+    // the unmapped base64 string must not be what's sent to the driver
+    expect(values[0]).not.toBe(bytesBase64)
   })
 })
--- a/packages/adapter-neon/src/__tests__/neon.test.ts
+++ b/packages/adapter-neon/src/__tests__/neon.test.ts
@@ -100,4 +100,25 @@ describe('PrismaNeonHttpAdapter', () => {
     // the unmapped base64 string must not be what's sent to the driver
     expect(values[0]).not.toBe(bytesBase64)
   })
+
+  it('maps a DateTime arg to a UTC timestamp string, not the process-local offset', async () => {
+    const mockClient = vi.fn().mockResolvedValue({ fields: [], rows: [], rowCount: 0 })
+    const adapter = new PrismaNeonHttpAdapter(mockClient as unknown as neon.NeonQueryFunction<any, any>)
+
+    // the query engine hands DateTime args over as ISO strings
+    const isoDate = '2023-01-01T12:00:00.000Z'
+
+    const query: SqlQuery = {
+      sql: 'INSERT INTO "Thing" ("date") VALUES ($1)',
+      args: [isoDate],
+      argTypes: [{ scalarType: 'datetime', arity: 'scalar' }],
+    }
+
+    await adapter.performIO(query)
+
+    const [, values] = mockClient.mock.calls[0]
+    // formatDateTime renders the UTC wall-clock value with no offset suffix, unlike the
+    // driver's default `Date#toISOString`/`prepareValue`-style local-offset serialization
+    expect(values[0]).toBe('2023-01-01 12:00:00')
+  })
 })
--- a/packages/adapter-neon/src/neon.ts
+++ b/packages/adapter-neon/src/neon.ts
@@ -288,7 +288,8 @@ export class PrismaNeonHttpAdapter extends NeonQueryable implements SqlDriverAda
   }
 
   override async performIO(query: SqlQuery): Promise<PerformIOResult> {
-    const { sql, args: values } = query
+    const { sql, args } = query
+    const values = args.map((arg, i) => mapArg(arg, query.argTypes[i]))
     return await this.client(sql, values, {
       arrayMode: true,
       fullResults: true,


santichausis:
Good call — added a DateTime test alongside the Bytes one, asserting the exact UTC-formatted string `formatDateTime` produces rather than just checking it changed.

aqrln (approved):

PR #29751, opened 2026-07-22
fix(client-engine-runtime): avoid spreading row/param-sized arrays onto the stack

### LINKED ISSUE #29746: RangeError: Maximum call stack size exceeded in query-interpreter (client-engine-runtime) for findMany above ~59-60K    rows/relations/in-filter values
### Bug description

`findMany` throws an unhandled RangeError: Maximum call stack size exceeded from @prisma/client-engine-runtime's query interpreter once a query
  processes roughly **_59,000–60,000 items_** in one go. Reproduced with two independent, minimal query shapes that don't share any obvious code path
  except "many items processed in one query":

  1. A `findMany` whose select includes a to-many relation that actually has matching child rows for enough parent rows.
  2. A `findMany` with a single `in: [...]` filter containing more than ~59-60K values, with **no relations selected at all.**

  Both fail at the same threshold, which suggests a shared recursive routine rather than two separate bugs. Below the threshold both queries succeed
  and return correct data. This is not specific to cursor-based pagination — it reproduces with plain `take`/`orderBy`, no `cursor`/`skip` involved.

### Severity

### Reproduction

No public repository yet — happy to put one together if useful, but the repro is small enough to include in full here (single schema, two models,
  no app-specific dependencies).

  Schema:
```prisma
 model Group {
    id       Int              @id @default(autoincrement())
    members  GroupMember[]
  } 
  
  model GroupMember {
    id         Int      @id @default(autoincrement())
    groupId Int
    userId     Int
    group   Group @relation(fields: [groupId], references: [id], onDelete: Cascade)
  }
  
```
  Seed: 100,000 `Group` rows, each with exactly 4 `GroupMember` rows (400,000 total).

  Case 1 — nested to-many relation:
```js
  const adapter = new PrismaPg({ connectionString: DATABASE_URL });

  // works
  await prisma.group.findMany({
    select: { id: true, members: { select: { userId: true } } },
    take: 59000,
    orderBy: { id: 'asc' },
  });
  
  // throws
  await prisma.group.findMany({
    select: { id: true, members: { select: { userId: true } } },
    take: 60000,
    orderBy: { id: 'asc' },
  });
```

Binary-searched the exact boundary: fails starting at `take: 60000`, consistently succeeds at `take: 59000` (4 children per parent, ~236K total rows attached across the batch). The same shape one relation deeper reproduces it too, at a similar order of magnitude (fails at `take: 80000`, succeeds at `take: 50000`).

  Case 2 — flat in filter, no relations at all:
```js
  // works
  await prisma.groupMember.findMany({
    where: { groupId: { in: Array.from({ length: 59000 }, (_, i) => i + 1) } },
    select: { groupId: true, userId: true },
  });
  
  // throws
  await prisma.groupMember.findMany({
    where: { groupId: { in: Array.from({ length: 60000 }, (_, i) => i + 1) } },
    select: { groupId: true, userId: true, turnOrder: true }, 
  });
```

Fails at `60,000` values in the in array, succeeds at `59,000`. No relation selected at all in this query.

Ruled out as necessary: `_count` in the select, cursor pagination (`cursor`/`skip`), multi-level relation nesting.
Ruled out as sufficient: a nested to-many relation with zero matching children never crashes, even at `take: 100000` — the recursion scales with rows actually attached, not just parent rows returned.

Expected: a query returning 60K+ rows should succeed the same way it does at 59K — or if there's an intentional limit, surface as a typed Prisma
  error, not an unhandled stack overflow.

  Actual:

  RangeError: Maximum call stack size exceeded
      at e.interpretNode (/node_modules/@prisma/client-engine-runtime/src/interpreter/query-interpreter.ts:196:26)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async /node_modules/@prisma/client-engine-runtime/src/interpreter/query-interpreter.ts:246:28
      at async e.interpretNode (/node_modules/@prisma/client-engine-runtime/src/interpreter/query-interpreter.ts:243:26)
      at async /node_modules/@prisma/client-engine-runtime/src/interpreter/query-interpreter.ts:246:28
      at async e.interpretNode (/node_modules/@prisma/client-engine-runtime/src/interpreter/query-interpreter.ts:243:26)
      at async e.interpretNode (/node_modules/@prisma/client-engine-runtime/src/interpreter/query-interpreter.ts:272:41)
      at async e.run (/node_modules/@prisma/client-engine-runtime/src/interpreter/query-interpreter.ts:94:23) {
    clientVersion: '7.9.0'
  }

### Frequency

Consistently reproducible

### Is this a regression?

Yes, confirmed. Ran the identical query, against the identical seeded data (same Postgres database, untouched), on the same machine/Node version, against Prisma 6.19.3 instead of 7.9.0:

  - Prisma 6.19.3, library engine (@prisma/client, default prisma-client-js generator, no adapter): take: 60000 → OK (994ms), take: 100000 → OK
  (1599ms)
  - Prisma 6.19.3, binary engine (engineType = "binary"): take: 60000 → OK (1143ms), take: 100000 → OK (1792ms)
  - Prisma 7.9.0, client-engine-runtime + @prisma/adapter-pg: take: 60000 → RangeError: Maximum call stack size exceeded

  Both Prisma 6 engine variants (library and binary — the two Rust-engine code paths) handle 100,000 rows with the nested relation without any issue.
  Only the new Prisma 7 TypeScript interpreter fails, and it fails consistently above ~59-60K. This is specific to the new engine introduced in 7.x,
  not present in 6.19.3.

### Workaround

Yes: keep every findMany under ~59,000 items total (parent rows with attached relation data, or values in a single in filter). We chunk large in filters to 32,000, and fetch to-many relations via a separate, chunked query instead of nesting them in a large-batch findMany, merging in application code. No way found to exceed the threshold in a single query safely.

### Prisma Schema & Queries

see Reproduction section

### Prisma Config

```js
  import * as dotenv from 'dotenv';
  
  dotenv.config();

  export default defineConfig({
    migrations: { path: 'prisma/migrations' },
    datasource: { url: process.env.DIRECT_URL },
  });
 // Client instantiation:
  import { PrismaClient } from './generated/client/client';

  const adapter = new PrismaPg({ connectionString: process.env.DATABASE_URL });
```

### Logs & Debug Info

Not yet captured with DEBUG=* — the failure is a synchronous JS stack overflow inside the interpreter rather than a query/connection-level failure, so debug/DB logs aren't expected to add anything beyond the query being issued; can capture if useful.

### Environment & Setup

  - OS: macOS (Darwin, arm64) — local repro; production target is Debian (node:20-bookworm-slim), not expected to be OS-specific
  - Database: PostgreSQL 15.17 (Alpine, linux/aarch64)
  - Node.js version: v22.22.0

### Prisma Version

```
  prisma               : 7.9.0
  @prisma/client       : 7.9.0
  Operating System     : darwin
  Architecture         : arm64
  Node.js              : v22.22.0
  TypeScript           : 5.9.3
  PSL                  : @prisma/prisma-schema-wasm 7.9.0-1.e922089b7d7502aff4249d5da3420f6fa55fc6ad
  Schema Engine        : schema-engine-cli e922089b7d7502aff4249d5da3420f6fa55fc6ad
  Default Engines Hash : e922089b7d7502aff4249d5da3420f6fa55fc6ad
  Studio               : 0.33.0
  (@prisma/adapter-pg: 7.9.0, pg: 8.22.0 — not in this output, adding since the driver adapter is central to the repro)
```

Hprogram:
## Problem

Closes #29746.

`findMany` throws `RangeError: Maximum call stack size exceeded` once a query processes roughly 59-60K items — both for a select with a to-many relation over enough parent rows, and for a flat `in: [...]` filter with ~60K values.

Both shapes funnel into the same routine: `renderTemplateSql()` appends every parameter behind an `IN (...)` clause by spreading the whole array as call arguments (`flattenedParams.push(...flattenedFragmentParams(fragment))`), which exceeds V8's argument-count/stack limit. Relation-loading subqueries are compiled with `chunkable: false`, so `chunkParams` — the intended mitigation — is bypassed entirely and the full parent-key array reaches the spread. When chunking does run, the per-chunk row merge uses the same pattern (`results.rows.push(...result.rows)` in `query-interpreter.ts`), which can still overflow on high-fanout relations (chunk size × fanout rows).

The exact threshold depends on stack already consumed when the query runs (a bare `push(...arr)` fails above ~125K elements; with a few thousand frames of app/framework machinery it fails around 60K), which matches the reported binary-searched boundary and its variation at deeper nesting.

## Changes

- `render-query.ts`: append flattened fragment params with a loop instead of spreading them as call arguments; keeps the `added` count used by the tuple-arity check
- `query-interpreter.ts`: merge chunked query result rows with a loop instead of a spread
- Regression tests for both sites (200K-element `parameterTuple` render with `chunkable: false`; chunked-result merge preserving 200K rows) — both fail with `RangeError` on the previous implementation and pass now

A repo-wide sweep found no other spread sites that scale with row/parameter count.

## Test plan

- [x] `pnpm --filter @prisma/client-engine-runtime test` — 209 passed (includes the 2 new regression tests)
- [x] New tests verified to fail (`RangeError` at the exact fixed lines) against the previous implementation
- [x] Build, lint, and prettier clean


aqrln (approved):
LGTM, thank you!

Hprogram:
Addressed the CodSpeed regression on `interpreter: findUnique`. The plain per-element loop replaced the previous single-spread call even for small inputs, and the added iterator overhead in the hot render path is what the instrument-based benchmark picked up.

Switched to a shared `appendToArray` helper: a single `push(...source)` for arrays up to 8,192 elements (restoring the original fast path — this covers virtually all real queries), and batched spreads above that (preserving the stack-safety fix; the regression tests with 200K elements still pass). The 8,192 batch size stays far below V8's argument limit even with a deep call stack.

Hprogram:
Follow-up on the updated report: the original `interpreter: findUnique` regression is resolved by the fast-path restore (it's back in the untouched list).

The newly flagged `interpreter: sequence` entry looks like a cross-environment artifact rather than a real change:

- CodSpeed marks the comparison itself as **"compared across different runtime environments"** — BASE ran on an EPYC 7763 (64-core), HEAD on an EPYC 9V74 (80-core).
- The benchmark is flagged for containing system calls that cannot be consistently instrumented (~47% of the profile is `node:internal/crypto/random`).
- The same run reports an ×86 "improvement" on `compile findUnique (uncached baseline)` (79,875 µs → 929 µs), which this diff certainly didn't cause — pointing to the same environment variance.

Code-wise, `SEQUENCE_PLAN` carries only 1-2 scalar parameters per statement, so it takes the same single-spread fast path as before this PR. Happy to dig further if useful, but I believe both entries are measurement variance from the runner change.

aqrln:
Yes, I think this is just noise. The PR looks good to me as is, and is ready to be merged as soon as the CI passes.

